PR #520, opened 2022-09-07
test: e2e balances

francovenica:
- [x] Editing selectors
- [x] Fixing regex for balances validations
- [x] Changing pagination test (used to be 5 rows by default, now is 25)
~Changing the test of the "Receive" modal test (no more Receive button in the assets tab)~

### DIFF
--- a/cypress/e2e/smoke/balances.cy.js
+++ b/cypress/e2e/smoke/balances.cy.js
@@ -165,31 +165,31 @@ describe('Assets > Coins', () => {
       // Open the Safe used for testing pagination
       cy.visit(`/${PAGINATION_TEST_SAFE}/balances`)
       cy.contains('button', 'Accept selection').click()
-  
+
       //Ensure the table is fully loaded
-      cy.contains('Rows per page:', {timeout:10000})
-    })  
+      cy.contains('Rows per page:', { timeout: 10000 })
+    })
     it('should should allow 25 rows per page and navigate to next and previous page', () => {
       // Click on the pagination select dropdown
       cy.get('[aria-haspopup="listbox"]').contains('25').click()
-  
+
       // Select 25 rows per page
       cy.get('[aria-haspopup="listbox"]').get('li[role="option"]').contains('25').click()
-  
+
       // Table should have 25 rows
       cy.contains('1–25 of 27')
       cy.get(balanceSingleRow).should('have.length', 25)
-  
+
       // Click on the next page button
       cy.get('button[aria-label="Go to next page"]').click()
-  
+
       // Table should have 1 rows
       cy.contains('26–27 of 27')
       cy.get(balanceSingleRow).should('have.length', 2)
-  
+
       // Click on the previous page button
       cy.get('button[aria-label="Go to previous page"]').click()
-  
+
       // Table should have 25 rows
       cy.contains('1–25 of 27')
       cy.get(balanceSingleRow).should('have.length', 25)
@@ -198,29 +198,29 @@ describe('Assets > Coins', () => {
 
   describe('should open assets modals', () => {
     const receiveAssetsModalTestId = '[aria-labelledby=":r1i:"]'
-    
+
     it('should open the Receive assets modal', () => {
       // Assets table container should exist
       cy.get('[data-track="overview: Show Safe QR code"]').should('be.visible').click()
 
       // The Receive Assets modal should be present
       cy.get(receiveAssetsModalTestId).should('be.visible')
-      
+
       // Receive assets should be present
       cy.get(receiveAssetsModalTestId).findByText('Receive assets').should('be.visible')
-      
+
       // The Receive screen should have the correct address
       const [, safeAddress] = PAGINATION_TEST_SAFE.split(':')
       cy.get(receiveAssetsModalTestId).findByText(safeAddress).should('be.visible')
-      
+
       // Checking and unchecking the QR code
       cy.get(receiveAssetsModalTestId).find('[type="checkbox"]').should('be.checked')
-      
+
       cy.get(receiveAssetsModalTestId).find('[type="checkbox"]').uncheck().should('not.be.checked')
-      
+
       // Click in the Done button
       cy.get(receiveAssetsModalTestId).find('[data-testid="CloseIcon"]').click()
-      
+
       // The Receive screen should be hidden
       cy.get(receiveAssetsModalTestId).should('not.exist')
     })
--- a/cypress/e2e/smoke/balances.cy.js
+++ b/cypress/e2e/smoke/balances.cy.js
@@ -178,7 +178,8 @@ describe('Assets > Coins', () => {
       cy.get(balanceSingleRow).should('have.length', 25)
 
       // Change to 10 rows per page
-      cy.contains('Rows per page:').next().contains('25').click()
+      // force click because the center of the element is hidden from the view
+      cy.contains('Rows per page:').next().contains('25').click({ force: true })
       cy.get('ul[role="listbox"]').contains('10').click()
 
       // Table should have 10 rows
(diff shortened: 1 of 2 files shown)


cypress/e2e/smoke/balances.cy.js
```diff
@@ -0,0 +1,228 @@
...
+    })
+  })
+
+  describe('fiat currency can be changed', () => {
+    it('should have ETH as default currency', () => {
+      // First row Fiat balance should not contain EUR
+      cy.get(balanceSingleRow).first().find('td').eq(FIAT_AMOUNT_COLUMN).should('not.contain', '€')
+      // First row Fiat balance should contain DAI
+      cy.get(balanceSingleRow).first().find('td').eq(TOKEN_AMOUNT_COLUMN).contains('DAI')
+    })
```
DiogoSoaress, line 147: What are we trying to test here?

Looking at `safe-react` we were testing that the default fiat currency was `$`

cypress/e2e/smoke/balances.cy.js
```diff
@@ -0,0 +1,228 @@
...
+
+    it('should allow changing the currency to EUR', () => {
+      // Click on balances currency dropdown
+      cy.get('[id="currency"]').click()
+
+      // Select EUR
+      cy.get('ul[role="listbox"]').findByText('EUR').click()
+
+      // First row Fiat balance should not contain USDC
```
DiogoSoaress, line 156:
```suggestion
      // First row Fiat balance should not contain USD
```

cypress/e2e/smoke/balances.cy.js
```diff
@@ -0,0 +1,228 @@
...
+      cy.get(balanceSingleRow).first().find('td').eq(FIAT_AMOUNT_COLUMN).should('not.contain', 'USDC')
```
DiogoSoaress, line 157:
```suggestion
      cy.get(balanceSingleRow).first().find('td').eq(FIAT_AMOUNT_COLUMN).should('not.contain', 'USD')
```

cypress/e2e/smoke/balances.cy.js
```diff
@@ -0,0 +1,228 @@
...
+          cy.get('td').eq(TOKEN_AMOUNT_COLUMN).contains('1.003 ZRX')
+          cy.get('td').eq(FIAT_AMOUNT_COLUMN).contains(fiatRegex)
+        })
+    })
+  })
+
+  describe('fiat currency can be changed', () => {
+    it('should have ETH as default currency', () => {
+      // First row Fiat balance should not contain EUR
+      cy.get(balanceSingleRow).first().find('td').eq(FIAT_AMOUNT_COLUMN).should('not.contain', '€')
```
DiogoSoaress, line 144:
```suggestion
      cy.get(balanceSingleRow).first().find('td').eq(FIAT_AMOUNT_COLUMN).should('not.contain', 'EUR')
```

DiogoSoaress: A PR was merged today with this change

cypress/e2e/smoke/balances.cy.js
```diff
@@ -0,0 +1,228 @@
...
+      cy.get(balanceSingleRow).first().find('td').eq(FIAT_AMOUNT_COLUMN).should('not.contain', 'USDC')
+      // First row Fiat balance should contain EUR
+      cy.get(balanceSingleRow).first().find('td').eq(FIAT_AMOUNT_COLUMN).contains('€')
+    })
+  })
+
+  describe('pagination should work', () => {
+    before(() => {
+      // Open the Safe used for testing pagination
+      cy.visit(`/${PAGINATION_TEST_SAFE}/balances`, { failOnStatusCode: false })
```
DiogoSoaress, line 166: Refer to my first comment: 
> I suggest removing it and going with the default behaviour.

cypress/e2e/smoke/balances.cy.js
```diff
@@ -0,0 +1,228 @@
...
+      cy.contains('button', 'Accept selection').click()
+
+      //Ensure the table is fully loaded
+      cy.contains('Rows per page:', { timeout: 10000 })
+    })
+    it('should should allow 25 rows per page and navigate to next and previous page', () => {
```
DiogoSoaress, line 172:
```suggestion
    it('should allow 25 rows per page and navigate to next and previous page', () => {
```

cypress/e2e/smoke/balances.cy.js
```diff
@@ -0,0 +1,228 @@
...
+      // Click on the pagination select dropdown
+      cy.get('[aria-haspopup="listbox"]').contains('25').click()
+
+      // Select 25 rows per page
+      cy.get('[aria-haspopup="listbox"]').get('li[role="option"]').contains('25').click()
```
DiogoSoaress, line 177: Why clicking on the already selected value?

cypress/e2e/smoke/balances.cy.js
```diff
@@ -0,0 +1,228 @@
...
+
+      cy.get(receiveAssetsModalTestId).find('[type="checkbox"]').uncheck().should('not.be.checked')
+
+      // Click in the Done button
+      cy.get(receiveAssetsModalTestId).find('[data-testid="CloseIcon"]').click()
+
+      // The Receive screen should be hidden
+      cy.get(receiveAssetsModalTestId).should('not.exist')
+    })
+  })
```
DiogoSoaress, line 227: this button does not exist on the Balances table. If you want to keep this test I suggest moving it out of this file.

DiogoSoaress (changes requested):

cypress/e2e/smoke/balances.cy.js
```diff
@@ -0,0 +1,228 @@
...
+const TOKEN_AMOUNT_COLUMN = 1
+const FIAT_AMOUNT_COLUMN = 2
+
+describe('Assets > Coins', () => {
+  // Fiat balance regex
+  const fiatRegex = new RegExp(`([0-9]{1,3},)*[0-9]{1,3}.[0-9]{2}`)
+
+  before(() => {
+    // Open the Safe used for testing
+    cy.visit(`/${TEST_SAFE}/balances`, { failOnStatusCode: false })
```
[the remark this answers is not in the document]
francovenica: I don't know the full explanation but simply without that it doesn't work. Check the other smoke test, all of them have it. I cannot remove it

cypress/e2e/smoke/balances.cy.js
francovenica: dunno what happened here. I'll change it to first checks that there is $ and not € and after the balance selector change there is € and not $

cypress/e2e/smoke/balances.cy.js
francovenica: So is no longer "€" and "$"? currently if I put USD and EUR it will fail

cypress/e2e/smoke/balances.cy.js
francovenica: Same as before, can't remove that

cypress/e2e/smoke/balances.cy.js
francovenica: True, no need to click the selector. Now is just there to check that the element exists before continuing

cypress/e2e/smoke/balances.cy.js
francovenica: You can still open the modal on the "QR" code button on the side. If someday we make a test about the sidebar I'll move it there. I don't want to lose the steps.

francovenica:
I tried to merge the current main, the one with the change to the balances symbol into my branch, but there is something currenlty broken in main so my branch is now broken as well (until main is fixed). So I changed the validation of the balances symbols directly on the github editor)

cypress/e2e/smoke/balances.cy.js
DiogoSoaress: I added a test for clicking a different value (10) to see if the pagination worked and the table adjusted accordingly

cypress/e2e/smoke/balances.cy.js
DiogoSoaress: I suggest you create a test for the Sidebar Header then. This test here is misplaced

DiogoSoaress (approved):

PR #523, opened 2022-09-07
fix: copy prefix when showing it is disabled

iamacook:
## What it solves

Resolves address prefixes not copying

## How this PR fixes it

`showPrefix`/`copyPrefix` props have been added as per `safe-react` and the `<PrefixedEthHashInfo>` references settings to enable these accordingly.

## How to test it

Disable rendering of prefixes. Observe that the prefix is, however, copied.

### DIFF
--- a/src/components/common/EthHashInfo/index.tsx
+++ b/src/components/common/EthHashInfo/index.tsx
@@ -11,6 +11,7 @@ import { useAppSelector } from '@/store'
 import { selectSettings } from '@/store/settingsSlice'
 import { selectChainById } from '@/store/chainsSlice'
 import useChainId from '@/hooks/useChainId'
+import { ethers } from 'ethers'
 
 type EthHashInfoProps = {
   address: string
@@ -41,7 +42,7 @@ const EthHashInfo = ({
   hasExplorer,
 }: EthHashInfoProps): ReactElement => {
   const [fallbackToIdenticon, setFallbackToIdenticon] = useState(false)
-  const isAddress = address.length === 42
+  const shouldPrefix = ethers.utils.isAddress(address)
 
   return (
     <div className={css.container}>
@@ -70,12 +71,12 @@ const EthHashInfo = ({
 
         <Box className={css.addressRow}>
           <Typography variant="body2" fontWeight="inherit" component="div" className={css.address}>
-            {isAddress && showPrefix && prefix && <b>{prefix}:</b>}
+            {showPrefix && shouldPrefix && prefix && <b>{prefix}:</b>}
             {shortAddress && shortAddress ? shortenAddress(address) : address}
           </Typography>
 
           {showCopyButton && (
-            <CopyAddressButton prefix={prefix} address={address} copyPrefix={isAddress && copyPrefix} />
+            <CopyAddressButton prefix={prefix} address={address} copyPrefix={shouldPrefix && copyPrefix} />
           )}
 
           {hasExplorer && <ExplorerLink address={address} />}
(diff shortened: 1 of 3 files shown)


src/components/common/EthHashInfo/index.tsx
```diff
@@ -37,6 +42,7 @@ const EthHashInfo = ({
   hasExplorer,
 }: EthHashInfoProps): ReactElement => {
   const [fallbackToIdenticon, setFallbackToIdenticon] = useState(false)
+  const shouldPrefix = ethers.utils.isAddress(address)
```
DiogoSoaress, line 45:
```suggestion
  const isAddress = ethers.utils.isAddress(address)
```

src/components/common/EthHashInfo/index.tsx
```diff
@@ -65,11 +71,13 @@ const EthHashInfo = ({
 
         <Box className={css.addressRow}>
           <Typography variant="body2" fontWeight="inherit" component="div" className={css.address}>
-            {prefix && <b>{prefix}:</b>}
-            {shortAddress ? shortenAddress(address) : address}
+            {showPrefix && shouldPrefix && prefix && <b>{prefix}:</b>}
+            {shortAddress && shortAddress ? shortenAddress(address) : address}
```
DiogoSoaress, line 75: Why this change?

src/components/common/EthHashInfo/index.tsx
```diff
@@ -88,9 +96,16 @@ const PrefixedEthHashInfo = ({
...
 
   const name = showName ? props.name || addressBook[props.address] : undefined
-  const prefix = props.address.length === 42 && settings.shortName.show ? chain?.shortName : undefined
 
-  return <EthHashInfo prefix={prefix} {...props} name={name} />
+  return (
+    <EthHashInfo
+      prefix={chain?.shortName}
+      showPrefix={settings.shortName.show}
+      copyPrefix={settings.shortName.copy}
```
DiogoSoaress, line 104: We are passing these `settings` as props because eventually the `<EthHashInfo />` will come from a component library, right?

DiogoSoaress (approved):
👍 looks good. Couple of questions though.

francovenica:
I see a bunch of questions still. I'll put it back in peer review. Send it to QA ready when merged

src/components/common/EthHashInfo/index.tsx
iamacook: I would argue that this is more declarative as we don't want to prefix hashes. What do you think?

src/components/common/EthHashInfo/index.tsx
iamacook: This was an oversight on my behalf. Good catch!

src/components/common/EthHashInfo/index.tsx
iamacook: Exactly. I separated the local settings from the component.

src/components/common/EthHashInfo/index.tsx
DiogoSoaress: I think it doesn't reflect our real intention in the logic. Perhaps `canPrefix` ? Once it can only prefix if the string is an address. Your call though

src/components/common/EthHashInfo/index.tsx
DiogoSoaress: Super!

francovenica:
Works fine.

PR #526, opened 2022-09-07
fix: older Safe upgrades

### LINKED ISSUE #522: Safes 1.0.0 - pages crashes when trying to update the version
## Bug description
Just clicking on the Update button makes the page crash and says the 1.0.0 is not a valid version

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby

## Steps to reproduce
 1. Go to a safe 1.0.0
 2. Click the upgrade button

## Expected result
The popUp to review the tx shows up

## Obtained result
Page crashes

## Screenshots
image

iamacook:
## What it solves

Resolves #522

## How this PR fixes it

On Safes unsupported by the Core SDK, the contract interface is retrieved directly.

## How to test it

Upgrading a 1.0.0 Safe should succeed.


src/services/tx/safeUpdateParams.ts
```diff
@@ -1,11 +1,41 @@
...
 import { LATEST_SAFE_VERSION } from '@/config/constants'
+import { isValidSafeVersion } from '@/hooks/coreSDK/safeCoreSDK'
+import { Gnosis_safe__factory } from '@/types/contracts/factories/Gnosis_safe__factory'
+import { getWeb3 } from '@/hooks/wallets/web3'
 
 export const CHANGE_MASTER_COPY_ABI = 'function changeMasterCopy(address _masterCopy)'
 export const CHANGE_FALLBACK_HANDLER_ABI = 'function setFallbackHandler(address handler)'
 
+const getContractInterface = (chain: ChainInfo, version: string) => {
+  if (isValidSafeVersion(version)) {
```
iamacook, line 17: This check happens in `getSafeContractInstance` itself. I considered using `try { ... } catch { ... }` instead but thought this was more declarative.

src/services/tx/safeUpdateParams.ts
```diff
@@ -14,13 +44,17 @@ export const CHANGE_FALLBACK_HANDLER_ABI = 'function setFallbackHandler(address
  */
 export const createUpdateSafeTxs = (safe: SafeInfo, chain: ChainInfo): MetaTransactionData[] => {
   const latestMasterCopy = getGnosisSafeContractInstance(chain, LATEST_SAFE_VERSION)
-  const safeContractInstance = getGnosisSafeContractInstance(chain, safe.version)
+  const safeContractInterface = getContractInterface(chain, safe.version)
 
   // @ts-expect-error this was removed in 1.3.0 but we need to support it for older safe versions
-  const changeMasterCopyCallData = safeContractInstance.encode('changeMasterCopy', [latestMasterCopy.getAddress()])
+  const changeMasterCopyCallData = safeContractInterface.encodeFunctionData('changeMasterCopy', [
```
usame-algan, line 50: Can we also call `.encodeFunctionData` if the contract is returned from the core-sdk? Previously it was changed to just `.encode`

iamacook: Via the `interface`, yes.

usame-algan, line 50: It looks like `changeMasterCopy` is not found on the 1.0.0 contract even though the function exists.

Screenshot 2022-09-08 at 10 20 36

src/services/tx/safeUpdateParams.ts
```diff
@@ -1,11 +1,41 @@
...
+    throw new Error(`GnosisSafe contract not found for chainId: ${chain.chainId}`)
+  }
+
+  const web3 = getWeb3()
+
+  if (!web3) {
+    throw new Error('No provider instance found')
+  }
+
+  return Gnosis_safe__factory.connect(contractAddress, web3).interface
```
usame-algan, line 36: If we only want to encode the function call we can use the interface directly without relying on `web3` i.e. `return Gnosis_safe__factory.createInterface()`

iamacook:
This now successfully creates the required transaction(s) to upgrade a Safe, however, <1.1.1 Safes are not supported by the Core SDK. When attempting to upgrade directly from a 1.0.0 Safe, the Core SDK doesn't initialise as the version is [not in the `safeDeploymentsVersion`](https://github.com/safe-global/safe-core-sdk/blob/main/packages/safe-core-sdk/src/contracts/config.ts#L16-L38). I verified that the transactions are correctly created by hardcoding '1.0.0' in a newer Safe. They were added to the queue successfully.

@germartinez's confirmed that the Core SDK-supported versions mirror that of the `safe-react` UI. I think we should assess whether we even allow interaction with legacy Safes at all in `web-core`.

src/services/contracts/safeContracts.ts
```diff
@@ -21,7 +21,7 @@ export const _getValidatedGetContractProps = (
   safeVersion: string,
 ): Pick<GetContractProps, 'chainId' | 'safeVersion'> => {
   if (!isValidSafeVersion(safeVersion)) {
-    throw new Error(`${safeVersion} is not a valid Safe version`)
+    throw new Error(`${safeVersion} is not a Safe version supported by the safe-core-sdk`)
```
katspaugh, line 24: `safe-core-sdk` -> `Safe Core SDK`

src/services/tx/safeUpdateParams.ts
```diff
@@ -1,41 +1,86 @@
 import { MetaTransactionData, OperationType } from '@gnosis.pm/safe-core-sdk-types'
 import { ChainInfo, SafeInfo } from '@gnosis.pm/safe-react-gateway-sdk'
-import { getFallbackHandlerContractInstance, getGnosisSafeContractInstance } from '@/services/contracts/safeContracts'
+import {
+  getFallbackHandlerContractInstance,
+  getGnosisSafeContractInstance,
+  _getSafeContractDeployment,
```
katspaugh, line 6: If you import it from another module, it shouldn't have an underscore.

src/services/tx/safeUpdateParams.ts
```diff
@@ -1,41 +1,86 @@
...
-      operation: OperationType.Call,
-    },
-  ]
-
-  return txs
+  const safeContractInterface = getContractInterface(chain, safe.version)
+
+  const changeMasterCopyTx = getChangeMasterCopyTransaction(safe, chain, safeContractInterface)
+
+  const hasSetFallbackHandler = semverSatisfies(safe.version, '>=1.1.0')
```
katspaugh, line 77: Please don't hardcode strings and numbers like this.

iamacook:
As we intend to host `safe-react` as well, I will close this for now. We can use it as a reference should we revisit this in the future.

PR #527, opened 2022-09-08
fix: remove date label timezone adjustment from queue

iamacook:
## What it solves

Resolves broken transaction grouping

## How this PR fixes it

Date label timezone adjustment was inadventedly added to the transaction queue. The queue doesn't return date labels and it has therefore been removed.

image

## How to test it

Open `rin:0xFfDC1BcdeC18b1196e7FA04246295DE3A17972Ac` and observe that no date labels are present and that nonce 395/396 transactions are correctly grouped with no date label separation.

## Screenshots

image

### DIFF
--- a/src/components/common/PaginatedTxns/index.tsx
+++ b/src/components/common/PaginatedTxns/index.tsx
@@ -43,7 +43,7 @@ const TxPage = ({
 
   const isQueue = useTxns === useTxQueue
 
-  const localizedItems = useMemo(() => {
+  const txListPageItems = useMemo(() => {
     if (!page?.results) {
       return
     }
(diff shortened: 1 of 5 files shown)


src/components/common/PaginatedTxns/index.tsx
```diff
@@ -40,20 +41,29 @@ const TxPage = ({
   const { page, error } = useTxns(pageUrl)
   const [filter] = useTxFilter()
 
-  if (page?.results) {
-    const isQueue = useTxns === useTxQueue
+  const isQueue = useTxns === useTxQueue
 
+  const localizedItems = useMemo(() => {
```
DiogoSoaress, line 46: Doesn't sound like a naming improvement IMO. 
I suggest `txListPageItems`

DiogoSoaress (approved):
just a naming suggestion

francovenica:
LGMT. Looks like the snapshot in the ticket
image

PR #528, opened 2022-09-08
feat: Safe Apps Sign and Execute transactions

yagopv:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/454

## How this PR fixes it
We are adding the required components to allow Safe Apps to send transactions and sign messages
  - Transaction Modal components
  - Sign Message Modal components
  - Communicator support for `sendTransactions`, `signMessage` and `signTypedMessage`
  - Adding support for Safe Apps request Id's in the current EventBus. Required for closing requests from the SDK when proposing transactions
  - Adding support for components as Stepper titles (label)

## How to test it
1) When trying to interact with a contract doing an unique operation it should show the SafeTxModal with the details of the operation (Single Call)

image

2) When trying to interact with a contract with several operations it should show the SafeTxModal with the details of the operation actions (and create a MultiSend delegate call)

image

3) When trying to sign a message `signMessage` it should show the SignMessage Modal with the text message and the `signMessage` method

image

4) When trying to sign a typed message `signTypedMessage` it should show the SignMessage Modal with the text message and the `signTypedMessage`method

image


src/components/safe-apps/SafeAppsTxModal/ReviewSafeAppsTx.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const chain = useCurrentChain()
+  const { safe } = useSafeInfo()
+
+  const isMultiSend = txs.length > 1
+  const canExecute = safe.threshold === 1
+  const [safeTx, safeTxError] = useAsync<SafeTransaction>(async () => {
+    const tx = await createMultiSendCallOnlyTx(txs)
+
+    if (params?.safeTxGas) {
+      // @ts-expect-error safeTxGas readonly
```
yagopv, line 37: Readonly prop. Are there another way to specify the gas coming from a parameter in an SDK request?

src/services/contracts/safeContracts.ts
```diff
@@ -170,3 +172,16 @@ export const getFallbackHandlerContractInstance = (chainId: string): Compatibili
 
   return new Contract(contractAddress, new Interface(fallbackHandlerDeployment.abi)) as Compatibility_fallback_handler
 }
+
+// Sign messages deployment
+// TODO: Should this be implemented in Core SDK?
```
yagopv, line 177: Good idea to add it the core?

yagopv:
> I think that the data structure of the legacy version (the one on the right) would suit us better.

Yes, I think we should format this but need to change some web-core components. Don't know if the format is intended @iamacook @usame-algan

JagoFigueroa:
Found a small issue when trying to vote on a snapshot proposal on a 2/x safe: 
Screenshot 2022-09-09 at 09 51 44
I am prompted to execute the signing tx when it should only be a signing one (also there seems to be an issue as well with the app logo on the pop up).

Cheers!

yagopv:
> I am prompted to execute the signing tx when it should only be a signing one (also there seems to be an issue as well with the app logo on the pop up).

Fixed this!

iamacook:
>     1. When there is any issue on the estimation/review of the transaction the Submit button is disabled. In the legacy web the user could to send the transaction anyway. Is this intentional?

I don't think so. Semantically, the warning suggests that transaction submission is possible. I see no issue in enabling it if it mirrors that of `safe-react`.

>     2. The simulation is not working for me because an GS020 error (Signatures are not added) but the transaction works if I submit it. Do you know if you have any issue with the simulation?

We haven't enountered any issues. Can you share the Tenderly simulation?

src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -44,6 +59,36 @@ const AppFrame = ({ appUrl }: AppFrameProps): ReactElement => {
...
+        }
+
+        txModalState.isOpen ? closeTxModal() : closeSignMessageModal()
+      }
+    })
+
+    return () => {
+      unsubscribe()
+    }
+  }, [chainId, closeTxModal, closeSignMessageModal, communicator, txModalState, signMessageModalState])
```
katspaugh, line 80: Instead of passing `txModalState, signMessageModalState` you could just pass the id.
```
const requestId = txModalState.requestId || signMessageModalState.requestId
```

It would make it a little more clear why this hook depends on those two states.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -28,6 +30,7 @@ type SignOrExecuteProps = {
   onSubmit: (data: null) => void
   children?: ReactNode
   error?: Error
+  requestId?: RequestId
```
katspaugh, line 33: `safeAppRequestId` maybe?

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -18,6 +19,7 @@ import { type ConnectedWallet } from '@/hooks/wallets/useOnboard'
 import { useCurrentChain } from '@/hooks/useChains'
 import { getTxOptions } from '@/utils/transactions'
 import { TxSimulation } from '@/components/tx/TxSimulation'
+import { txDispatch, TxEvent } from '@/services/tx/txEvents'
```
katspaugh, line 22: As you can see, this component wasn't previously dispatching any tx events directly.
`txSender` does it. Let's create a function in txSender for this. Something like `dispatchSafeAppsTx`.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -129,11 +133,18 @@ const SignOrExecuteForm = ({
       return
     }
 
+    // For Safe Apps:
+    // - Avoid redirection
+    // - Dispatch an event for return the safeTxHash to the caller
+    if (requestId) {
+      txDispatch(TxEvent.SAFE_APPS_REQUEST, { txId: id, requestId })
+      return
```
katspaugh, line 141: I'd suggest not to return here and instead let if call `onSubmit` further down.

To avoid redirecting, just add a requestId check to the if below.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -129,11 +133,18 @@ const SignOrExecuteForm = ({
       return
     }
 
+    // For Safe Apps:
+    // - Avoid redirection
+    // - Dispatch an event for return the safeTxHash to the caller
```
katspaugh, line 138:
```suggestion
    // Dispatch a Safe Apps SDK transaction request
```

src/components/tx/TxStepper/index.tsx
```diff
@@ -12,11 +12,15 @@ const TxStepper = ({ steps, initialData, initialStep, onClose }: TxStepperProps)
...
 
+  const activeLabel = steps[activeStep].label
+
   return (
     <div className={css.container}>
       <ModalDialogTitle onClose={onClose}>
         <Grid container px={1} alignItems="center" gap={2}>
-          <Grid item>{steps[activeStep].label}</Grid>
+          <Grid item>
+            {typeof activeLabel === 'string' ? activeLabel : activeLabel(stepData[Math.max(0, activeStep)])}
```
katspaugh, line 22: What's this for?

src/services/contracts/safeContracts.ts
katspaugh: cc @germartinez

germartinez: There is a refactor that I want to do regarding how the contract instances are returned by the Core SDK and I think it makes sense to add the SignMessageLib there as well

src/services/contracts/safeContracts.ts
```diff
@@ -170,3 +172,16 @@ export const getFallbackHandlerContractInstance = (chainId: string): Compatibili
...
+export const getSignMessageLibDeploymentContractInstance = (chainId: string): Sign_message_lib => {
+  const signMessageLibDeployment = getSignMessageLibDeployment({ network: chainId }) || getSignMessageLibDeployment()
```
mikhailxyz, line 179: Not directly related to the PR, but I wonder why we need this fallback logic. Do we not trust that `({ network: chainId })` filter works? cc @katspaugh @iamacook

src/components/tx/SignOrExecuteForm/index.tsx
mikhailxyz: > Dispatch a Safe Apps SDK transaction request

IMO this comment just repeats what the code already says. The original comment is not perfect, but it at least tries to describe why a specific code is needed.

src/services/contracts/safeContracts.ts
mikhailxyz: I agree, but also at the same time, it looks like removing it won't break anything

src/services/contracts/safeContracts.ts
yagopv: Great!!. For now we leave it like this. Make us know when you will be on it

Thanks!!

iamacook: Maybe it'd help putting it in an issue? I've just moved over the other Core SDK requests into issues [here](https://github.com/safe-global/web-core/pull/565#issuecomment-1250728045).

src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -44,6 +59,36 @@ const AppFrame = ({ appUrl }: AppFrameProps): ReactElement => {
...
+
+  const onSafeAppsModalClose = () => {
+    if (txModalState.isOpen) {
+      communicator?.send(REJECT_TRANSACTION_MESSAGE, txModalState.requestId, true)
```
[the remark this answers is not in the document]
yagopv: But even creating another method, i would maintain the message as a parameter because perhaps i want to change it if the rejection is for another cause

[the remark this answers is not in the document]
katspaugh: I see. Let's at least move the message constant itself to the communicator module then.

src/components/safe-apps/AppFrame/index.tsx
yagopv: yes but even with that i have the `txModalState.isOpen` inside the effect so will need to create another effect for handling this

src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -44,6 +59,36 @@ const AppFrame = ({ appUrl }: AppFrameProps): ReactElement => {
...
+          communicator?.send({ safeTxHash: detailedExecutionInfo.safeTxHash }, requestId)
+        }
+
+        txModalState.isOpen ? closeTxModal() : closeSignMessageModal()
+      }
+    })
+
+    return () => {
+      unsubscribe()
+    }
```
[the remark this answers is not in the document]
yagopv: done, thanks!

src/components/safe-apps/AppFrame/index.tsx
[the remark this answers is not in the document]
yagopv: make sense, done!

src/components/safe-apps/SafeAppsSignMessageModal/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+import TxModal, { TxModalProps } from '@/components/tx/TxModal'
+import SafeAppsModalLabel from '../SafeAppsModalLabel'
+import ReviewSafeAppsSignMessage from './ReviewSafeAppsSignMessage'
+
+export type SafeAppsSignMessageParams = {
+  appId?: string
+  app?: SafeAppData
+  requestId: RequestId
+  message: string | EIP712TypedData
+  method: string
```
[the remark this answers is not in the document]
yagopv: What do you think to simply change it to `method: Methods.signMessage | Methods.signTypedMessage`

src/components/safe-apps/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+    return false
+  }
+
+  const isAddressValid = validateAddress(t.to) === undefined
+  return isAddressValid && !!t.data && typeof t.data === 'string'
+}
+
+export const isTxValid = (txs: BaseTransaction[]) => txs.length && txs.every((t) => validateTransaction(t))
+
+export const isNativeTransfer = (encodedData: string) => encodedData && isNaN(parseInt(encodedData, 16))
```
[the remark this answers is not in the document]
yagopv: Found a similar one here https://github.com/safe-global/web-core/blob/fed4dbb802aad9989aadc9fa251f02928d7b1cbf/src/components/tx/DecodedTx/index.tsx#L25

@katspaugh do you think is a good idea to add this as an utility function ?

src/components/safe-apps/AppFrame/index.tsx
katspaugh: Ah, I haven't noticed that one.

src/components/safe-apps/utils.ts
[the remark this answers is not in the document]
katspaugh: Sure!

src/components/tx/SignOrExecuteForm/index.tsx
yagopv: In this case I agree with @mikhailxyz. The intention is to make clear we need to avoid the redirection to the transaction history and we need to get the hash for the awaiting caller

src/components/tx/TxStepper/index.tsx
yagopv: This is in order to allow custom components on the modal header. We need it for showing the app icon with the app name (both dynamic)

src/components/safe-apps/AppFrame/useAppCommunicator.ts
```diff
@@ -23,7 +28,16 @@ type JsonRpcResponse = {
   error?: string
 }
 
-const useAppCommunicator = (iframeRef: MutableRefObject<HTMLIFrameElement | null>, app?: SafeAppData) => {
+type UseAppCommunicatorConfig = {
+  app?: SafeAppData
+  onConfirmTransactions: (txs: BaseTransaction[], requestId: RequestId, params?: SendTransactionRequestParams) => void
+  onSignTransactions: (message: string | EIP712TypedData, requestId: string, method: Methods) => void
```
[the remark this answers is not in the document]
yagopv: Done!! Make (super) sense 😉

src/components/safe-apps/SafeAppsSignMessageModal/index.tsx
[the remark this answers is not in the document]
yagopv: did it

src/components/tx/SignOrExecuteForm/index.tsx
yagopv: ok done

src/components/tx/TxStepper/index.tsx
yagopv: Done!

src/components/tx/SignOrExecuteForm/index.tsx
yagopv: I prefer to take control of the modal because when user press the cancel button or the cross icon i need to send a message to the Safe App

katspaugh: You can still do it onSubmit, can't you? I don't like that the Safe App request behaves different than other types of modals. IMHO they should all call onSubmit on the final step.

yagopv: Ok, refactored as suggested leaving the SignOrExecuteForm as before but with the id in the Submit. Adjusted other components using it

katspaugh: Cool, thanks Yago!

katspaugh (approved):
LGTM & LFG!

src/components/safe-apps/SafeAppsSignMessageModal/ReviewSafeAppsSignMessage.tsx
```diff
@@ -0,0 +1,132 @@
...
+      ])
+    }
+
+    return createTx({
+      to: signMessageAddress,
+      value: '0',
+      data: txData || '0x',
+      operation: OperationType.DelegateCall,
+    })
+  }, [])
```
usame-algan, line 66: We might be missing some dependencies here as well e.g. `message`

JagoFigueroa:
I have been creating and executing a few transactions on different chains and safes and I think it is all bueno to be merged. We can continue with the testing for this on web core moving forward. Cheers!

src/components/safe-apps/SafeAppsSignMessageModal/ReviewSafeAppsSignMessage.tsx
yagopv: You are right!

src/components/safe-apps/SafeAppsModalLabel/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+const SafeAppsModalLabel = ({ app }: { app?: SafeAppData }) => {
+  if (!app) {
+    return <Typography variant="h3">Safe Apps Transaction</Typography>
+  }
+
+  return (
+    <Box display="flex" alignItems="center">
+      <img src={app?.iconUrl || APP_LOGO_FALLBACK_IMAGE} alt={app?.name} className={css.modalLabel} />
+      <Typography variant="h4">{app?.name}</Typography>
+    </Box>
```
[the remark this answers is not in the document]
yagopv: Thanks!

src/components/safe-apps/SafeAppsTxModal/ReviewSafeAppsTx.tsx
```diff
@@ -0,0 +1,100 @@
...
+            </InfoDetails>
+
+            {safeTx && (
+              <Box pb={2}>
+                <Typography mt={2} color="secondary.light">
+                  Data (hex encoded)
+                </Typography>
+                {generateDataRowValue(safeTx.data.data, 'rawData')}
+              </Box>
+            )}
```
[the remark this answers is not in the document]
yagopv: yes thanks. Previously I had more conditions

src/components/safe-apps/SafeAppsTxModal/ReviewSafeAppsTx.tsx
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    return tx
+  }, [])
+
+  const [decodedData] = useAsync<DecodedDataResponse | undefined>(async () => {
+    if (!safeTx || isEmptyHexData(safeTx.data.data)) return
+
+    return getDecodedData(chainId, safeTx.data.data)
+  }, [safeTx])
```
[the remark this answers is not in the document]
yagopv: ok!

src/components/tx/TxStepper/index.tsx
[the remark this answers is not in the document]
yagopv: ok!

src/services/safe-apps/manifest.ts
```diff
@@ -17,14 +17,34 @@ export type AppManifest = {
...
+const chooseBestIcon = (icons: AppManifestIcon[]): string => {
+  const svgIcon = icons.find((icon) => icon?.sizes?.includes('any') || icon?.type === 'image/svg+xml')
+
+  if (svgIcon) {
+    return svgIcon.src
+  }
+
+  for (const icon of icons) {
+    for (const size of icon.sizes.split(' ')) {
+      if (Number(size?.split('x')[0]) >= MIN_ICON_WIDTH) {
```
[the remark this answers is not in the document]
yagopv: thanks!

yagopv:
Thanks @usame-algan, useful corrections!

PR #529, opened 2022-09-08
fix: load conflicted transactions across pages

iamacook:
## What it solves

Resolves conflicts not rendering

## How this PR fixes it

If the final item in the transaction list has a "HasNext" conflict, the next page is automatically loaded.

## How to test it

Queue enough transactions to fill the page and then group the final trasaction so that it spreads onto the next page (example Safe: `rin:0xF5A2915982BC8b0dEDda9cEF79297A83081Fe88f`). Observe that there are two queue network requests when opening the queue.

## Screenshots
conflicts

### DIFF
--- a/src/components/common/PaginatedTxns/index.tsx
+++ b/src/components/common/PaginatedTxns/index.tsx
@@ -44,6 +44,10 @@ const TxPage = ({
   const isQueue = useTxns === useTxQueue
 
   useEffect(() => {
+    if (!isQueue) {
+      return
+    }
+
     // If conflict transactions split across pages, we must fetch the next page
     const lastItem = page?.results[page.results.length - 1]
     const hasSplitConflict = lastItem && isHasNextConflictTransactionListItem(lastItem)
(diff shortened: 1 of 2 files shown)


src/components/common/PaginatedTxns/index.tsx
```diff
@@ -43,6 +43,20 @@ const TxPage = ({
...
   const isQueue = useTxns === useTxQueue
 
+  useEffect(() => {
+    if (!isQueue) {
+      return
+    }
+
+    // If conflict transactions split across pages, we must fetch the next page
+    const lastItem = page?.results[page.results.length - 1]
+    const hasSplitConflict = lastItem && isHasNextConflictTransactionListItem(lastItem)
```
usame-algan, line 53: What happens if we don't fetch the next page in those cases and what does the list look like when paginating then?

iamacook: They don't appear as a full group.

iamacook:
It seems that our current pagination actually interferes with this. I'm going to close this for the time being but we can use this as reference.

PR #530, opened 2022-09-08
feat: Add smart contract wallet support when signing txs

### LINKED ISSUE #503: Wallet connect: can not sign if the wallet is connected via WC
The same issue for the wallets connected via WC:
Steps:
1. open Metamask on the mobile device
2. connect Metamask to the safe via WC
3. try to sign tx
Current result: The error appears, and signing is not possible 

Image

usame-algan:
## What it solves

Resolves #503 

## How this PR fixes it

Being connected to the safe via a smart contract wallet is handled when signing a transaction.

- Adds a new `txEvent` called `AWAITING_ON_CHAIN_SIGNATURE`

## How to test it

1. Open a Safe
2. Connect to safe 1 via WC safe app on another safe 2
3. Create a new transaction on safe 1
4. Observe a popup via WC on safe 2
5. Confirm the transaction on safe 2
6. Observe a notification on safe 1
7. Observe being navigated to a proposed tx on safe 1
8. Observe the proposed tx having 1 signature after the tx was executed in safe 2

## Open issues

~The UI currently doesn't handle the response from mobile and is "stuck" in the modal flow. The new tx appears correctly in the queue of the safe that is connected but `sdk.approveTransactionHash` and `sdk.executeTransaction` are not resolving when the tx is confirmed on mobile.~
This has been solved by using an [unchecked signer](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#UncheckedJsonRpcSigner) which is similar to web3js `PromiEvent.once('transactionHash')`


src/services/tx/txSender.ts
```diff
@@ -189,6 +189,23 @@ export const dispatchTxSigning = async (
...
+  const sdk = getAndValidateSafeSDK()
+
+  try {
+    const safeTxHash = await sdk.getTransactionHash(safeTx)
+    await sdk.approveTransactionHash(safeTxHash)
+  } catch (err) {
+    txDispatch(TxEvent.SIGN_FAILED, { txId: undefined, error: err as Error })
+    throw err
+  }
+
```
iamacook, line 205: We should dispatch a success event here.

src/hooks/wallets/wallets.ts
```diff
@@ -77,6 +78,18 @@ export const isHardwareWallet = (wallet: ConnectedWallet): boolean => {
...
+export const isSmartContractWallet = async (wallet: ConnectedWallet) => {
+  const provider = getWeb3ReadOnly()
+
+  if (!provider) {
+    throw new Error('Provider not found')
+  }
+
+  const code = await provider.getCode(wallet.address)
+
+  return code !== '0x'
```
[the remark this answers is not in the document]
usame-algan: We can reuse `EMPTY_DATA` from the core-sdk

src/services/tx/txSender.ts
```diff
@@ -192,17 +195,29 @@ export const dispatchTxSigning = async (
 /**
  * On-Chain sign a transaction
  */
-export const dispatchOnChainSigning = async (safeTx: SafeTransaction) => {
+export const dispatchOnChainSigning = async (safeTx: SafeTransaction, chain?: ChainInfo, txId?: string) => {
   const sdk = getAndValidateSafeSDK()
+  const provider = getWeb3()
+
+  if (!provider || !chain) return safeTx
```
katspaugh, line 202: So the `chain` param is only passed for this check? Can we do it outside of the call instead?

usame-algan: It was a leftover from a previous implementation. I removed it and also moved the `provider` check outside. Thanks!

src/services/tx/txSender.ts
usame-algan: Dispatching the SUCCESS event makes sense only after the transaction has been executed. In this case we don't execute or sign the transaction. It is only being proposed to the gateway. I would suggest we only dispatch a FAILED event in case the transaction is being rejected in the wallet.

usame-algan: I also added an EXECUTING event dispatch and use a `groupKey` since we don't have a `txId` at that point.

PR #531, opened 2022-09-08
test: unit test formatNumber

DiogoSoaress:
## What it solves
Unit tests for number formatting -> https://github.com/5afe/safe/wiki/How-to-format-amounts

_Note:_ I've used underscore as numeric separator for readability.

## How to test it
Run `yarn test formatNumber`

### DIFF
--- a/src/utils/__tests__/formatNumber.test.ts
+++ b/src/utils/__tests__/formatNumber.test.ts
@@ -138,11 +138,13 @@ describe('formatAmount', () => {
   })
 
   it('should use < -0.00001 or < +0.00001 when the Eucledian distance of the amount is smaller than 0.00001', () => {
-    // explicitly positive numbers: keep the sign
-    expect(formatAmount('+0.000001', true)).toEqual('< +0.00001')
-    expect(formatAmount('+0.000009', true)).toEqual('< +0.00001')
+    // to keep the '+' sign the amount shall be passed as a string
+    expect(formatAmount('+0.000001')).toEqual('< +0.00001')
+    expect(formatAmount('+0.000009')).toEqual('< +0.00001')
 
-    // negative numbers
+    // negative numbers will keep the sign either way
+    expect(formatAmount(-0.000001)).toEqual('< -0.00001')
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/formatNumber.test.ts
```diff
@@ -1,13 +1,149 @@
-describe('Appease Jest', () => {
-  it('adds at least one test to the test file', () => {
-    expect(true).toBe(true)
+import { formatAmount } from '@/utils/formatNumber'
```
iamacook, line 1: Great tests! We should add another that checks that the "Thousands and decimal separators are used according to user's locale."

src/utils/formatNumber.ts
```diff
@@ -57,31 +58,29 @@ const getCurrencyFormatterOptions = (float: number, currency: string): Intl.Numb
   }
 }
 
-// TODO: Add signs in reference to final point of number formatting
-const formatNumber = (float: number, options: Intl.NumberFormatOptions): string => {
-  const VISIBLE_UPPER_LIMIT = 999 * 10 ** 12 // 999T
-
+const formatNumber = (float: number, options: Intl.NumberFormatOptions, keepSign?: boolean): string => {
```
iamacook, line 61: We can determine whether the "Eucledian distance of the amount is smaller than 0.00001" without the `keepSign` argument. I don't think it's necessary as we format numbers the same app wide.

src/utils/__tests__/formatNumber.test.ts
DiogoSoaress: We've decided before to go with literal assertions. Please refer to https://github.com/safe-global/safe-react/pull/3920/commits/09a66b717f763b821719e34edcb7a1ceb33b7c9b#r893219929

src/utils/formatNumber.ts
DiogoSoaress: I removed the argument and preferred to check if the string starts with the plus sign

iamacook (approved):

PR #536, opened 2022-09-11
Refactor: move batch-able txns from props to a hook call

katspaugh:
<s>
## What it solves
Tx batching and grouping-by-nonce both need to operate on txs across paginated payloads.
This PR aggregates tx pages into one flat array for this purpose.

## How this PR fixes it
Previously each paginated group was rendered as a separate component with its own loading hook.
Now all pages are loaded in the parent component and rendered as a single flat list.

Please note that we unfortunately lose the automatic re-rendering of the first page when polling, if the the user has scrolled past the initial page. This could be solved tho, if we refactor the useTxns hook. To be discussed. IMHO, not a blocker however.
</s>

Since the queue is usually within one page, the cons of a flat list outweighs the pros. I've restored it how it was + just added some small improvements.

## How to test it
Load the tx list, scroll to the bottom. Filter. Switch the queue/history tabs back and forth.

## Analytics changes
None.


src/hooks/useTxHistory.ts
```diff
@@ -22,6 +22,7 @@ const useTxHistory = (
     () => {
       if (!safeLoaded || !(filter || pageUrl)) return
 
+      // TODO: pass timezone_offset param to the API
```
katspaugh, line 25: @iamacook passing timezone_offset to CGW should solve the labels being in the wrong place. It probably also localizes the timestamps in the tx list, need to check.

src/components/common/PaginatedTxns/index.tsx
```diff
@@ -9,93 +11,78 @@ import InfiniteScroll from '../InfiniteScroll'
...
-    return <ErrorMessage>Error loading transactions</ErrorMessage>
-  }
-
-  return <SkeletonTxList />
-}
+  // Reset the pages when the filter changes
+  useEffect(() => {
+    setPages([])
+    setPageUrl(undefined)
+  }, [filter, useTxns])
```
katspaugh, line 41: I've noticed this doesn't always reset the pageUrl quick enough. Sometimes a filtered page loads quickly before pageUrl is reset, and it's not the first page.

src/components/common/PaginatedTxns/index.tsx
```diff
@@ -9,93 +11,78 @@ import InfiniteScroll from '../InfiniteScroll'
...
-  isFirstPage: boolean
-}): ReactElement => {
-  const { page, error } = useTxns(pageUrl)
+const PaginatedTxns = ({ useTxns }: { useTxns: typeof useTxHistory | typeof useTxQueue }): ReactElement => {
+  const [pages, setPages] = useState<Array<{ pageUrl?: string; data: TransactionListPage }>>([])
+  const [pageUrl, setPageUrl] = useState<string>()
   const [filter] = useTxFilter()
-
   const isQueue = useTxns === useTxQueue
+  const allItems = useMemo(() => flatten(pages.map((page) => page.data.results)), [pages])
```
iamacook, line 35: I suggest we use [`flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) instead of `lodash`.

src/hooks/useTxHistory.ts
iamacook: Sadly not. The `timezone_offset` is [ignored by the gateway](https://github.com/safe-global/safe-client-gateway/blob/da658a17153d88ce4964bf3889d70ae0bf79ddba/src/routes/transactions/routes.rs#L128) as it's yet to be implemented.

iamacook (commented):
> This could be solved tho, if we refactor the useTxns hook. To be discussed. IMHO, not a blocker however.

I totally agree that we should look into this at some point. Nice adjustment nonetheless!

src/components/common/PaginatedTxns/index.tsx
katspaugh: Oh wow, didn’t know we had this in JS, thanks!

src/components/common/PaginatedTxns/index.tsx
usame-algan: Is there a reason we are relying on `useTxns` as a dependency here? It seems to update correctly when I remove it.

katspaugh: I wasn't sure If it re-mounts the entire component when you switch tabs between Queue and History.

src/components/common/PaginatedTxns/index.tsx
```diff
@@ -9,93 +10,78 @@ import InfiniteScroll from '../InfiniteScroll'
...
-  const [pages, setPages] = useState<string[]>([''])
-  const [filter] = useTxFilter()
+  // Load the current page
+  const { page: lastPage, error, loading } = useTxns(pageUrl)
 
+  // When a tx page is loaded, update the list of pages
   useEffect(() => {
-    setPages([''])
-  }, [filter])
+    if (!lastPage) return
```
usame-algan, line 47: Is this meant to be the current page? The name could be confused with being the last page in the pages array i.e. pages.length - 1

katspaugh:
Pros of this approach:
* We group conflicting txns in the queue correctly
* You can batch more than the first page worth of queued txns (provided that your queue has conflicting txns, otherwise it's limited to the page size anyway)

Cons:
* The queue is (supposedly) rarely more than one page, so in most cases the pros above don't apply
* We lose the auto-updating of the tip of the history/queue when you scroll past page 1
* The code is more complex

src/components/common/PaginatedTxns/index.tsx
```diff
@@ -9,93 +10,78 @@ import InfiniteScroll from '../InfiniteScroll'
...
-  }
+    setPages((prevPages) => {
+      // If we're on the first page), "refresh" that page because it's polled.
+      // Otherwise, append the new page to the list of pages.
+      let pageIndex = prevPages.findIndex((page) => page.pageUrl === pageUrl)
+      if (pageIndex === -1) pageIndex = prevPages.length
+      const newPages = prevPages.slice()
+      newPages[pageIndex] = { pageUrl, data: lastPage }
+      return newPages
+    })
```
usame-algan, line 57: Was there an issue with the previous approach of always appending the new page? e.g. `setPages((prevPages) => [...prevPages, { pageUrl, data: lastPage }])`

src/components/common/PaginatedTxns/index.tsx
usame-algan: Makes total sense, `useTxns` also doesn't seem to change when applying the filter so there must be another reason.

src/components/common/PaginatedTxns/index.tsx
katspaugh: Will rename 👍

src/components/common/PaginatedTxns/index.tsx
katspaugh: Yes, actually the bug you found. If you're on the first page, the page updates every 15 sec and we don't won't to append it every time.

katspaugh: I've just realized there might be another bug. When a tx is executed, the first history page updates and there are more txns, then the pageUrl might be also different (a new offset).

usame-algan: Good find! I was also able to reproduce that issue. We only update the history if there is no filter applied or the user hasn't scrolled yet. Not sure if we should "force" an update in those cases.

katspaugh: Fixed it by using a count.

katspaugh:
@iamacook I've made the filter form popup in an overlay. Otherwise it jumps when switching the filter type.

Screenshot 2022-09-15 at 12 02 11

iamacook:
> @iamacook I've made the filter form popup in an overlay. Otherwise it jumps when switching the filter type.

The former was agreed upon with @liliiaorlenko as it displays all transactions. We could alternatively preserve space for the extra fields?

DiogoSoaress (commented):
~It seems you can keep scrolling down after applying a filter~

src/components/common/PaginatedTxns/index.tsx
katspaugh: I ended up restoring the previous approach with separate TxPage components each with its own loading hook.

katspaugh:
@iamacook gotcha, reverted.

src/components/common/PaginatedTxns/index.tsx
```diff
@@ -9,18 +10,17 @@ import InfiniteScroll from '../InfiniteScroll'
...
+import { isTransactionListItem } from '@/utils/transaction-guards'
 
 const NoQueuedTxns = () => (
   <Box mt="5vh">
     <PagePlaceholder imageUrl="/images/no-transactions.svg" text="Queued transactions will appear here" />
   </Box>
 )
 
-const getResultCount = (filter: TxFilter, page: TransactionListPage) => {
+const getFilterResultCount = (filter: TxFilter, page: TransactionListPage) => {
```
DiogoSoaress, line 23: Screenshot 2022-09-15 at 12 59 15
 is not necessarily true if the filtered transactions are more than 1 page

katspaugh: Yeah, it just refers to the first page, hence the `>` sign. I haven't changed anything with this component.

DiogoSoaress: Oh! The `>` was not obvious at first sight. 

>  I haven't changed anything with this component.

Yeah ik, was just a small remark

DiogoSoaress (approved):
Why is it better to make a hook call instead of the props?

I would rather leave the component dumber but not strongly opinionated in this case.

katspaugh:
@DiogoSoaress because we render txns in separate components per page, and were passing items only from the first page w/o it being very obvious. The result is the same – the batch button only sees the latest page, but it's at least decoupled from the rendered list.

src/components/common/PaginatedTxns/index.tsx
```diff
@@ -34,57 +34,59 @@ const TxPage = ({
...
+      {isFirstPage && filter && page && (
+        <Box display="flex" flexDirection="column" alignItems="flex-end" pt={[2, 0]} pb={3}>
+          {getFilterResultCount(filter, page)}
+        </Box>
+      )}
 
-  if (error) {
-    return <ErrorMessage>Error loading transactions</ErrorMessage>
-  }
+      {page && <TxList items={page.results} />}
```
usame-algan, line 52: Not critical but `TxList` is now being rendered as an empty element even if there are no results. We could still check for `page.results.length > 0` before rendering the list.

usame-algan (approved):
🚀

PR #540, opened 2022-09-14
test: E2E Transaction history 

francovenica:
Editing selectors.
I didn't need to change the test at all

NOTE: I had to change my timezone for these test to work, because they expect specific dates that don't show up if you are not in GMT +2. 
This could be related to this issue https://github.com/safe-global/web-core/issues/539
I don't know what timezone the CI uses


iamacook:
> I don't know what timezone the CI uses

The CI should use UTC. We set the `TZ` [here](https://github.com/safe-global/web-core/blob/main/package.json#L17).

usame-algan (approved):
LGTM

PR #541, opened 2022-09-15
WIP E2e test landing

francovenica:
Change selectors to test the preferences banner

The beamer test is commented because it doesn't work on my local (check that)


liliya-soroka:
Let's add the same TestId('cookies-banner-form') instead of .parent()  and reuse the full "landing.spec.js" test from the safe-react

cypress/e2e/smoke/landing.cy.js
```diff
@@ -0,0 +1,34 @@
...
+  it('redirects to welcome page', () => {
+    cy.visit('/')
+
+    cy.url().should('include', '/welcome')
+  })
+
+  it('features discoverability cookies', () => {
+    cy.visit('/')
+
+    let preferencesModal = cy.contains("We use cookies to provide").parent()
```
liliya-soroka: Let's add the same data-testid="cookies-banner-form"  for the "class="MuiPaper-root MuiPaper-elevation MuiPaper-rounded MuiPaper-elevation3 styles_container__12f0T css-c28ola" and reuse the full "landing.spec.js" test from the safe-react with testId instead of .parent()

iamacook: I would only advise we not `data-testId`s unless we really have to, e.g. if we were to alter the content of the banner, the tests won't break.

DiogoSoaress: Opted to use `cy.contains('We use cookies to provide').parent('div')` to find the Cookies banner parent element.

cypress/e2e/smoke/landing.cy.js
```diff
@@ -0,0 +1,35 @@
...
+    cy.visit('/')
+
+    cy.url().should('include', '/welcome')
+  })
+
+  it('should require accept "Updates" cookies to display Beamer', () => {
+    cy.visit('/')
+
+    // Way to select the cookies banner without an id
+    cy.contains('We use cookies to provide').parent('div').should('be.visible')
```
usame-algan, line 12: Is there a reason we are not checking for the existence of the content directly without going through the parent element? i.e. `cy.contains('We use cookies to provide').should('be.visible')`

cypress/e2e/smoke/landing.cy.js
```diff
@@ -0,0 +1,35 @@
...
+    cy.contains('We use cookies to provide').parent('div').contains('Accept selection').click()
```
usame-algan, line 13: Similar here, could we do `cy.contains('Accept selection').click()` like we do with other buttons/links?

cypress/e2e/smoke/landing.cy.js
DiogoSoaress: fair enough. I think I blindly replaced the old code 👍

usame-algan (approved):

PR #543, opened 2022-09-15
fix: queue/wallet events over tracking

iamacook:
## What it solves

Resolves queue/wallet events over tracking.

## How this PR fixes it

1. Transaction queue is now compared for equality before being tracked. Only actual transactions are tracked, now not including `DATE_LABEL`s or `CONFLICT_HEADER`s.
2. Connected wallet/WalletConnect peer labels are cached/compared before tracking.

## How to test it

1. Open the dashboard/transaction queue. Observe that only one event is tracked for the queue length, unless a queued transaction is added or executed.
2. Connect a wallet/WalletConnect peer and observe that only one event is tracked.

## Analytics changes

1. Queue length is tracked once for current queue data.
2. Wallet connection events are tracked once upon connection.

### DIFF
--- a/src/hooks/wallets/useOnboard.ts
+++ b/src/hooks/wallets/useOnboard.ts
@@ -79,16 +79,38 @@ const getWalletConnectLabel = async ({ label, provider }: ConnectedWallet): Prom
   return peerWallet ?? UNKNOWN_PEER
 }
 
-const trackWalletType = async (wallet: ConnectedWallet) => {
-  trackEvent({ ...WALLET_EVENTS.CONNECT, label: wallet.label })
+const prevTrackedWallet = {
+  label: '',
+  wcLabel: '',
+}
+
+const trackWalletType = async (wallet: ReturnType<typeof getConnectedWallet>) => {
+  if (!wallet) {
+    prevTrackedWallet.label = ''
+    prevTrackedWallet.wcLabel = ''
+    return
+  }
+
+  if (wallet.label !== prevTrackedWallet.label) {
+    trackEvent({ ...WALLET_EVENTS.CONNECT, label: wallet.label })
+
+    prevTrackedWallet.label = wallet.label
+  }
 
   const wcLabel = await getWalletConnectLabel(wallet)
 
-  if (wcLabel) {
+  if (!wcLabel) {
+    prevTrackedWallet.wcLabel = ''
+    return
+  }
+
+  if (wcLabel !== prevTrackedWallet.wcLabel) {
     trackEvent({
       ...WALLET_EVENTS.WALLET_CONNECT,
       label: wcLabel,
     })
+
+    prevTrackedWallet.wcLabel = wcLabel
   }
 }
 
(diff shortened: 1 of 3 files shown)


usame-algan:
The onboard wallet subscription seems to run multiple times which makes sense if certain values like ENS are updating but it also makes it difficult to detect an initial connection without having to rely on a cache.

Alternatively, we could track the connection event when calling `onboard.connectWallet` which is more explicit. Wdyt?

iamacook:
> Alternatively, we could track the connection event when calling `onboard.connectWallet` which is more explicit. Wdyt?

I do like this approach. `connectWallet` is used in three separate places. I will create a custom wrapper that follows this logic.

src/hooks/wallets/useOnboard.ts
```diff
@@ -79,16 +79,38 @@ const getWalletConnectLabel = async ({ label, provider }: ConnectedWallet): Prom
...
+}
+
+const trackWalletType = async (wallet: ReturnType<typeof getConnectedWallet>) => {
+  if (!wallet) {
+    prevTrackedWallet.label = ''
+    prevTrackedWallet.wcLabel = ''
+    return
+  }
+
+  if (wallet.label !== prevTrackedWallet.label) {
```
katspaugh, line 94: In what cases does it trigger a connection event w/o this check? Not just once per actual connection?

src/store/txQueueSlice.ts
```diff
@@ -24,18 +25,26 @@ export const selectQueuedTransactionsByNonce = createSelector(
...
+      if (isEqual(txQueue.data?.results, payload.data?.results)) {
+        return
+      }
+
+      const transactions = payload.data?.results.filter(isTransactionListItem) || []
 
       trackEvent({
         ...TX_LIST_EVENTS.QUEUED_TXS,
-        label: payload.data.results.length.toString(),
+        label: transactions.length.toString(),
```
katspaugh, line 47: Maybe extract this into a function "trackQueueSize"?

src/hooks/wallets/useOnboard.ts
iamacook: The subscription fires again when ENS domains resolve, etc.

image

iamacook:
> The onboard wallet subscription seems to run multiple times which makes sense if certain values like ENS are updating but it also makes it difficult to detect an initial connection without having to rely on a cache.
> 
> Alternatively, we could track the connection event when calling `onboard.connectWallet` which is more explicit. Wdyt?

I've created a `connectWallet` wrapper that includes the tracking/caching of the last used wallet, as well as error logging as `onboard.connectWallet` is present in three places. It reduces a lot of code.

francovenica:
The queued tx even looks fine. It triggers by creating a tx (or a tx rejection) or by executing a tx (basically anything that makes the queue have more or less tx in the list).
Note that visiting the queue does not trigger this event (it only triggers the pageVisit)

francovenica:
I have issues with the "wallet connecting" event:
The only time I was able to trigger it is by using the mobile pairing. Anything else (MM, WalletConnect, Trezor, Ledger..) did not trigger the event.

image

iamacook:
> Note that visiting the queue does not trigger this event (it only triggers the pageVisit)

This is as expected. As you found, it only tracks the queue when it changes so we don't over track.

> The only time I was able to trigger it is by using the mobile pairing. Anything else (MM, WalletConnect, Trezor, Ledger..) did not trigger the event.

I've fixed that [here](https://github.com/safe-global/web-core/pull/551).

francovenica:
LGTM now. The event tracker for wallets was fixed in that other ticket

PR #549, opened 2022-09-15
fix: validate upper amount of spending limit

iamacook:
## What it solves

Resolves "value out-of-bounds" error

## How this PR fixes it

The [upper limit of a spending limit amount (`uint96`)](https://github.com/safe-global/safe-modules/blob/master/allowances/contracts/AlowanceModule.sol#L52) is now validated against.

## How to test it

Attempt to create a spending limit with a value > 2^96 - 1 and observe the new validation error that prevents you from submitting.

## Screenshots
image


katspaugh (approved):
👍

francovenica:
Never thought going so far up. 
LGTM
image

PR #550, opened 2022-09-15
Chore: update obsolete todos

katspaugh:
Some of the TODOs have been already fixed but the comments remained.

### DIFF
--- a/src/utils/date.ts
+++ b/src/utils/date.ts
@@ -1,5 +1,5 @@
-// TODO: Remove date-fns and covert to use @date-io/date-fns
-// https://github.com/dmtrKovalenko/date-io#projects
+// TODO: Remove date-fns and convert to use the native JS formatter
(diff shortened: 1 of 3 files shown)


src/utils/date.ts
```diff
@@ -1,5 +1,5 @@
-// TODO: Remove date-fns and covert to use @date-io/date-fns
-// https://github.com/dmtrKovalenko/date-io#projects
+// TODO: Remove date-fns and convert to use the native JS formatter
+// @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat
```
iamacook, line 2: It seems like the [`@date-io/date-fns`](https://github.com/dmtrKovalenko/date-io#interface) interface doesn't expose all of the features of `date-fns` that we use. I don't think we can remove it entirely.

`Intl.RelativeTimeFormat` looks like a good replacement for `formatDistanceToNow` (in `formatTimeInWords`) though.

iamacook (approved):

PR #551, opened 2022-09-16
fix: add missing wallet tracking

iamacook:
## What it solves

Resolves https://github.com/safe-global/web-core/pull/543#issuecomment-1248684913

## How this PR fixes it

The connection call when manually connecting is now wrapped in the tracking function.

## How to test it

Connect to a wallet via the header and observe the tracking event.

## Analytics changes

Connecting to a wallet via the header tracks the "Connect wallet" event

## Screenshots

connect


usame-algan (approved):
Good catch!

francovenica:
Looks good. Tried with WC, MM, trezor and ledger.
image

PR #552, opened 2022-09-16
refactor: currency formatting

iamacook:
## What it solves

Resolves currency style

## How this PR fixes it

The `Intl.NumberFormat` logic for amounts been refactored so that currency/amounts adhere to the same style.

Currency formatting now displays the currency code at the end of the amount which matches our desired design.

## How to test it

Observe no differences in amounts across the Safe, apart from the sidebar header showing currencies appended with the relevant currency code.

The currencies should be rounded to the country-specific precision, e.g. JPY has no decimals, EUR has two, IQD has three.

## Screenshots

image
image
image

### DIFF
--- a/src/utils/__tests__/formatNumber.test.ts
+++ b/src/utils/__tests__/formatNumber.test.ts
@@ -1,151 +1,313 @@
-import { formatAmount } from '@/utils/formatNumber'
+import { formatAmount, formatCurrency } from '@/utils/formatNumber'
 
-describe('formatAmount', () => {
-  it('should remove trailing zeroes', () => {
-    expect(formatAmount('0.10000')).toEqual('0.1')
-    expect(formatAmount('0.100000000000')).toEqual('0.1')
+describe('formatNumber', () => {
+  describe('formatAmount', () => {
+    it('should remove trailing zeroes', () => {
+      expect(formatAmount('0.10000')).toEqual('0.1')
+      expect(formatAmount('0.100000000000')).toEqual('0.1')
+    })
+
+    it('should use maximum of 5 decimals', () => {
+      expect(formatAmount('0.123456789')).toEqual('0.12346')
+    })
+
+    it('should use five decimals for numbers up until 999.99999', () => {
+      expect(formatAmount('345.123456789')).toEqual('345.12346') // 9 decimals
+      expect(formatAmount('999.99999')).toEqual('999.99999') // 5 decimals
+
+      // rounds above the specified limit
+      expect(formatAmount('999.999992')).toEqual('999.99999') // 6 decimals
+      expect(formatAmount('999.999996')).toEqual('1,000')
+    })
+
+    it('should use four decimals for numbers between 1,000.0001 until 9,999.9999', () => {
+      // rounds down past the specified precision
+      expect(formatAmount(1_000.00001)).toEqual('1,000')
+
+      expect(formatAmount(1_000.0001234)).toEqual('1,000.0001')
+      expect(formatAmount(1_234.123456789)).toEqual('1,234.1235')
+      expect(formatAmount(9_999.9999)).toEqual('9,999.9999')
+
+      // rounds above the specified limit
+      expect(formatAmount(9_999.99992)).toEqual('9,999.9999')
+      expect(formatAmount(9_999.99996)).toEqual('10,000')
+    })
+
+    it('should use three decimals for numbers between 10,000.001 until 99,999.999', () => {
+      // rounds down past the specified precision
+      expect(formatAmount(10_000.00001)).toEqual('10,000')
+
+      expect(formatAmount(10_000.001)).toEqual('10,000.001')
+      expect(formatAmount(12_345.123456789)).toEqual('12,345.123')
+      expect(formatAmount(99_999.999)).toEqual('99,999.999')
+
+      // rounds above the specified limit
+      expect(formatAmount(99_999.9992)).toEqual('99,999.999')
+      expect(formatAmount(99_999.9996)).toEqual('100,000')
+    })
+
+    it('should use two decimals for numbers between 100,000.01 until 999,999.99', () => {
+      // rounds down past the specified precision
+      expect(formatAmount(100_000.00001)).toEqual('100,000')
+
+      expect(formatAmount(100_000.01)).toEqual('100,000.01')
+      expect(formatAmount(123_456.123456789)).toEqual('123,456.12')
+      expect(formatAmount(999_999.99)).toEqual('999,999.99')
+
+      // rounds above the specified limit
+      expect(formatAmount(999_999.992)).toEqual('999,999.99')
+      expect(formatAmount(999_999.996)).toEqual('1,000,000')
+    })
+
+    it('should use one decimal for numbers between 1,000,000.1 until 9,999,999.9', () => {
+      // rounds down past the specified precision
+      expect(formatAmount(1_000_000.00001)).toEqual('1,000,000')
+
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/formatNumber.test.ts
```diff
@@ -1,151 +1,313 @@
...
+      expect(formatCurrency(amount1, 'BHD')).toBe('1.000 BHD')
+
+      const amount2 = '0.123456789'
+
+      expect(formatCurrency(amount2, 'JPY')).toBe('0 JPY')
+      expect(formatCurrency(amount2, 'IQD')).toBe('0 IQD')
+      expect(formatCurrency(amount2, 'USD')).toBe('0.12 USD')
+      expect(formatCurrency(amount2, 'EUR')).toBe('0.12 EUR')
+      expect(formatCurrency(amount2, 'GBP')).toBe('0.12 GBP')
+      expect(formatCurrency(amount2, 'BHD')).toBe('0.123 BHD')
```
DiogoSoaress, line 172: why has this currency 3 decimals?

iamacook: All Dinar have three decimal places.

src/utils/__tests__/formatNumber.test.ts
```diff
@@ -1,151 +1,313 @@
...
+
+      expect(formatCurrency(amount1, 'JPY')).toBe('0 JPY')
+      expect(formatCurrency(amount1, 'IQD')).toBe('0 IQD')
+      expect(formatCurrency(amount1, 'USD')).toBe('0.00 USD')
+      expect(formatCurrency(amount1, 'EUR')).toBe('0.00 EUR')
+      expect(formatCurrency(amount1, 'GBP')).toBe('0.00 GBP')
+      expect(formatCurrency(amount1, 'BHD')).toBe('0.000 BHD')
+    })
+
+    it("doesn't use < -0.00001 or < +0.00001 when the Eucledian distance of the amount is smaller than 0.00001", () => {
```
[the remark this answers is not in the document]
iamacook: Currencies are formatted to their native precision.

src/utils/__tests__/formatNumber.test.ts
```diff
@@ -1,151 +1,313 @@
...
+
+      expect(formatCurrency(amount3, 'JPY')).toBe('1T JPY')
+      expect(formatCurrency(amount3, 'IQD')).toBe('1T IQD')
+      expect(formatCurrency(amount3, 'USD')).toBe('1T USD')
+      expect(formatCurrency(amount3, 'EUR')).toBe('1T EUR')
+      expect(formatCurrency(amount3, 'GBP')).toBe('1T GBP')
+      expect(formatCurrency(amount3, 'BHD')).toBe('1T BHD')
+    })
+
+    it('should use T notation for numbers between 999,999,500,000 and 999,000,000,000', () => {
```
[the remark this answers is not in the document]
iamacook: Different currencies have varying decimal precision.

src/utils/__tests__/formatNumber.test.ts
```diff
@@ -1,151 +1,313 @@
...
+
+      expect(formatCurrency(amount3, 'JPY')).toBe('1B JPY')
+      expect(formatCurrency(amount3, 'IQD')).toBe('1B IQD')
+      expect(formatCurrency(amount3, 'USD')).toBe('1B USD')
+      expect(formatCurrency(amount3, 'EUR')).toBe('1B EUR')
+      expect(formatCurrency(amount3, 'GBP')).toBe('1B GBP')
+      expect(formatCurrency(amount3, 'BHD')).toBe('1B BHD')
+    })
+
+    it('should use B symbol for numbers between 999,999,500 and 999,999,500,000', () => {
```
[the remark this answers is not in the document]
iamacook: Different currencies have varying decimal precision.

src/utils/formatNumber.ts
```diff
@@ -5,7 +5,42 @@ const LOWER_LIMIT = 0.00001
...
+
+/**
+ * Numbers either with a +/- sign or that are negative are prepended with their relevant sign (when not 0)
+ * @param number Number to prepend sign to
+ */
+const getNumberFormatSignDisplay = (number: string | number): Intl.NumberFormatOptions['signDisplay'] => {
+  const float = Number(number)
+
+  const shouldDisplaySign =
+    typeof number === 'string' ? number.includes('+') || number.includes('-') : Math.sign(float) === -1
```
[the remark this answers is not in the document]
iamacook: > I think it's safer to trim the string and use `startsWith`

`includes` checks all indexes of a string.

>In general I feel it's cleaner to pass that check to outside the ternary. We should check for the - sign with something like

If we wanted to check against `float`, we would still need to use `Math.abs(float)`. I feel like using the native sign function is the most declarative.

src/utils/__tests__/formatNumber.test.ts
```diff
@@ -1,151 +1,262 @@
...
+      const amount1 = 1
+
+      expect(formatCurrency(amount1, 'JPY')).toBe('1 JPY')
+      expect(formatCurrency(amount1, 'IQD')).toBe('1 IQD')
+      expect(formatCurrency(amount1, 'USD')).toBe('1.00 USD')
+      expect(formatCurrency(amount1, 'EUR')).toBe('1.00 EUR')
+      expect(formatCurrency(amount1, 'GBP')).toBe('1.00 GBP')
+      expect(formatCurrency(amount1, 'BHD')).toBe('1.000 BHD')
+
+      const amount2 = '0.123456789'
```
DiogoSoaress, line 165: The `amount2` rounds always down for the tested currencies. I would add a `amount3 = 1.7777` (per example) to test if the rounding up is working as expected as well.

src/utils/__tests__/formatNumber.test.ts
```diff
@@ -1,151 +1,262 @@
...
+    })
+
+    it('should use B symbol for numbers between 999,999,500 and 999,999,500,000', () => {
+      const amount1 = 1_000_000_000
+
+      expect(formatCurrency(amount1, 'JPY')).toBe('1B JPY')
+
+      const amount2 = 1_234_567_898.123456789
+
+      expect(formatCurrency(amount2, 'JPY')).toBe('1.2B JPY')
```
DiogoSoaress, line 196: shouldn't it display `1.235B JPY` ?

src/utils/__tests__/formatNumber.test.ts
```diff
@@ -1,151 +1,262 @@
...
+    })
+
+    it('should use T notation for numbers between 999,999,500,000 and 999,000,000,000', () => {
+      const amount1 = 1_000_100_000_000
+
+      expect(formatCurrency(amount1, 'JPY')).toBe('1T JPY')
+
+      const amount2 = 1_234_567_898_765.123456789
+
+      expect(formatCurrency(amount2, 'JPY')).toBe('1.2T JPY')
```
DiogoSoaress, line 210: `1.234T JPY` ?

DiogoSoaress (changes requested):

src/utils/__tests__/formatNumber.test.ts
DiogoSoaress: Resolving as the number of fraction digits comes from the currency formatter

src/utils/__tests__/formatNumber.test.ts
DiogoSoaress: Resolving as the number of fraction digits comes from the currency formatter

DiogoSoaress (approved):
👍

francovenica:
LGTM

PR #555, opened 2022-09-16
refactor: use CGW SDK enums

DiogoSoaress:
## What it solves
Updates to CGW SDK latest version using the added types:

- ConflictType
- DetailedExecutionInfoType
- TransactionInfoType
- TransactionListItemType

## How to test it
History and Queue transaction lists should work as before
Expanding any transaction type details should work as before

### DIFF
--- a/src/utils/transaction-guards.ts
+++ b/src/utils/transaction-guards.ts
@@ -10,7 +10,9 @@ import {
   DetailedExecutionInfoType,
   Erc20Transfer,
   Erc721Transfer,
+  ExecutionInfo,
   Label,
+  ModuleExecutionDetails,
   ModuleExecutionInfo,
   MultiSend,
   MultisigExecutionDetails,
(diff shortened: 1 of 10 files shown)


src/utils/transaction-guards.ts
```diff
@@ -47,32 +54,15 @@ export const isOwner = (safeOwners: AddressEx[] | NamedAddress[] = [], walletAdd
...
-export const isModuleExecutionInfo = (
-  value: TransactionSummary['executionInfo'] | DetailedExecutionInfo,
-): value is ModuleExecutionInfo => value?.type === EXECUTION_INFO_TYPES.MODULE
-
-enum EXECUTION_INFO_TYPES {
-  MULTISIG = 'MULTISIG',
-  MODULE = 'MODULE',
-}
+export const isModuleDetailedExecutionInfo = (value?: DetailedExecutionInfo): value is ModuleExecutionDetails =>
+  value?.type === DetailedExecutionInfoType.MODULE
```
DiogoSoaress, line 63: I've renamed these methods to improve understanding of its use

src/utils/transaction-guards.ts
```diff
@@ -47,32 +54,15 @@ export const isOwner = (safeOwners: AddressEx[] | NamedAddress[] = [], walletAdd
   return safeOwners.some((owner) => sameAddress(owner.address, walletAddress))
 }
 
-export const isMultisigExecutionDetails = (value?: DetailedExecutionInfo): value is MultisigExecutionDetails => {
-  return value?.type === 'MULTISIG'
+// Narrows `TransactionDetails` -> when requesting a single transaction
```
iamacook, line 57: The guards are for `TransactionDetails['detailedExecutionInfo']`.

src/utils/transaction-guards.ts
```diff
@@ -129,6 +112,13 @@ export const isDateLabel = (value: TransactionListItem): value is DateLabel => {
   return value.type === TransactionListItemType.DATE_LABEL
 }
 
+// Narrows `Transaction`s -> when request a page of `TransactionListItem`
```
iamacook, line 115: The guards are for `TransactionSummary['executionInfo']`.

iamacook (commented):
I don't think the following comments are necessary but if we want to keep them, they aren't quite correct.

src/utils/transaction-guards.ts
DiogoSoaress: maybe the comments are unnecessary after all. I thought they would help clarifying the situations where the code deals with `ExecutionInfo` vs `DetailedExecutionInfo`

iamacook (approved):

Issue #559, opened 2022-09-16
Simulation not working if own signature is missing for 1/n safes


schmanu:
## Bug description
If a newly created transaction is being simulated it currently fails with an GS020 error because no signatures are provided.
This is caused because we only add prevalidated Signatures if the transaction is not executable.

Example simulation:
https://dashboard.tenderly.co/public/safe/safe-apps/simulator/8de7a235-e340-4acb-a8be-eb3613ad16c1?trace=0.0.1.0.0

## Environment
Does not matter

## Steps to reproduce
 1. Open a 1/n safe
 2. create any new transaction 
 3. try to simulate it

## Expected result
No error during simulation.

## Obtained result
GS020 Error

## Screenshots
See tenderly link in description for example

Issue #560, opened 2022-09-16
Simulation ignores gas limits


schmanu:
## Bug description
If a transaction is fully signed or only the current owner's signature is missing we do not need and should not overwrite the safe's threshold and pass the manual selected or estimated gas limit.

Reason: The number of signatures matters for the transactions gas limit. If the tx will only check 1 signature instead of `n>1` signatures it can occur that a successfully simulated tx reverts when being executed.

## Steps to reproduce
 1. Create a new ERC20 transfer
 2. Go to review step
 3. change the gas limit to 21000 (minimum)
 4. click simulate
 5. Observe simulation success
 6. try to execute with that gas limit
 7. Observe error: `Your transaction was unsuccessful. [ethjs-query] while formatting outputs from RPC '{"value":{"code":-32000,"message":"intrinsic gas too low"}}'`

## Expected result
The simulation should fail with "out of gas"

## Obtained result
The simulation always uses 30 million gas as limit and always succeeds.

PR #561, opened 2022-09-16
WIP: fix issues around tx simulation

schmanu:
- add prevalidated signatures if more signatures are needed and the owner's sig is missing
- pass gasLimit to txSimulation if estimated / manually set
- only overwrite threshold if necessary (threshold won't be reached otherwise)

closes issue #559 & issue #560

## What it solves

Resolves #

### DIFF
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -175,7 +175,14 @@ const SignOrExecuteForm = ({
           onFormSubmit={onAdvancedSubmit}
         />
 
-        {safeTx && <TxSimulation transactions={safeTx} canExecute={canExecute} disabled={submitDisabled} />}
+        {safeTx && (
+          <TxSimulation
+            manualGasLimit={advancedParams.gasLimit?.toNumber()}
+            transactions={safeTx}
+            canExecute={canExecute}
+            disabled={submitDisabled}
+          />
+        )}
 
         {(error || (willExecute && gasLimitError)) && (
           <ErrorMessage error={error || gasLimitError}>
--- a/src/components/tx/TxSimulation/useSimulation.ts
+++ b/src/components/tx/TxSimulation/useSimulation.ts
@@ -9,7 +9,7 @@ type UseSimulationReturn =
   | {
       simulationRequestStatus: FETCH_STATUS.NOT_ASKED | FETCH_STATUS.ERROR | FETCH_STATUS.LOADING
       simulation: undefined
-      simulateTransaction: (params: Omit<SimulationTxParams, 'gasLimit'>) => void
+      simulateTransaction: (params: Omit<SimulationTxParams, 'gasLimit'> & { manualGasLimit?: number }) => void
       simulationLink: string
       requestError?: string
       resetSimulation: () => void
@@ -17,7 +17,7 @@ type UseSimulationReturn =
   | {
       simulationRequestStatus: FETCH_STATUS.SUCCESS
       simulation: TenderlySimulation
-      simulateTransaction: (params: Omit<SimulationTxParams, 'gasLimit'>) => void
+      simulateTransaction: (params: Omit<SimulationTxParams, 'gasLimit'> & { manualGasLimit?: number }) => void
       simulationLink: string
       requestError?: string
       resetSimulation: () => void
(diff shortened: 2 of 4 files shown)


iamacook:
@yagopv, I think this may solve [your issue](https://github.com/safe-global/web-core/pull/528#issuecomment-1241751248).

src/components/tx/TxSimulation/useSimulation.ts
```diff
@@ -38,17 +38,25 @@ export const useSimulation = (): UseSimulationReturn => {
...
       setRequestError(undefined)
 
       try {
-        const { gasLimit } = await web3ReadOnly.getBlock('latest')
+        const manualGasLimit = params.manualGasLimit
+        let gasLimit: number
+        if (manualGasLimit) {
+          gasLimit = manualGasLimit
+        } else {
+          const latestBlock = await web3ReadOnly.getBlock('latest')
```
katspaugh, line 53: I think it's unlikely that there's no gas limit passed? Do we need this fallback?

schmanu: Yes for batch transactions we don't have gas estimation there. See [ReviewBatchExecute.tsx](https://github.com/safe-global/web-core/blob/main/src/components/tx/modals/BatchExecuteModal/ReviewBatchExecute.tsx#L96)

katspaugh: Gotcha 👍

src/components/tx/TxSimulation/useSimulation.ts
```diff
@@ -38,17 +38,25 @@ export const useSimulation = (): UseSimulationReturn => {
   }, [])
 
   const simulateTransaction = useCallback(
-    async (params: Omit<SimulationTxParams, 'gasLimit'>) => {
+    async (params: Omit<SimulationTxParams, 'gasLimit'> & { manualGasLimit?: number }) => {
```
iamacook, line 41: You can set `SingleTransactionSimulationParams['gasLimit']` and `MultiSendTransactionSimulationParams['gasLimit']` to optional and then use `params` directly.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -175,7 +175,14 @@ const SignOrExecuteForm = ({
           onFormSubmit={onAdvancedSubmit}
         />
 
-        {safeTx && <TxSimulation transactions={safeTx} canExecute={canExecute} disabled={submitDisabled} />}
+        {safeTx && (
+          <TxSimulation
+            manualGasLimit={advancedParams.gasLimit?.toNumber()}
```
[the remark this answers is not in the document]
schmanu: Done!

src/components/tx/TxSimulation/useSimulation.ts
schmanu: Done!

katspaugh (approved):
👍

src/components/tx/TxSimulation/utils.ts
```diff
@@ -129,19 +137,45 @@ const isSingleTransactionSimulation = (params: SimulationTxParams): params is Si
...
+  const tx = params.transactions
+  const hasOwnerSig = tx.signatures.has(params.executionOwner)
+  const effectiveSigs = tx.signatures.size + (hasOwnerSig ? 0 : 1)
+  return params.safe.threshold > effectiveSigs
+}
+
+const getLatestBlockGasLimit = async (): Promise<number> => {
+  const web3 = getWeb3ReadOnly()
+  const latestBlock = await web3?.getBlock('latest')
+  return latestBlock?.gasLimit.toNumber() || 30_000_000
```
iamacook, line 157: We should normally put this in a constant, but in this case I would argue against hardcoding a value at all. We want the simulation to remain as 'true' as possible.

src/components/tx/TxSimulation/useSimulation.ts
```diff
@@ -38,17 +38,15 @@ export const useSimulation = (): UseSimulationReturn => {
...
       if (!web3ReadOnly) return
 
       setSimulationRequestStatus(FETCH_STATUS.LOADING)
       setRequestError(undefined)
 
       try {
-        const { gasLimit } = await web3ReadOnly.getBlock('latest')
-        const simulationPayload = getSimulationPayload({
+        const simulationPayload = await getSimulationPayload({
           ...params,
```
iamacook, line 49: You don't need to spread anymore.

src/components/tx/TxSimulation/utils.ts
schmanu: The problem here is that `getBlock()` can return `undefined` this should never happen but for type safety I used this as a legit fallback.

iamacook: If there is no `gasLimit` and then `web3ReadOnly` is `undefined`, let's throw.

schmanu: okay

francovenica:
I'm always getting a GS020 error when simulating the 2nd tx in the queue:

For this queue 2 tx in a safe 2/x
The first tx is fully signed ready to be executed (I don't think this matters, but is the current state of my safe)
The second tx (the one that fails) has only be signed once, and this is the 2nd signature. Creating this 2nd tx didn't throw any error.

image
Also is using 30 million gas limit for some reason
image
https://dashboard.tenderly.co/public/safe/safe-apps/simulator/8cdca21f-d4bf-4c67-b98a-03cdd7da0680

francovenica:
The issue was fixed. LGTM

PR #562, opened 2022-09-19
Refactor: extract redirect + encode URL params

katspaugh:
## What it solves
* Encodes the safeApp URL param so that its own query params don't mix with the outer URL query
* Extracts the redirection logic in Safe creation into a function

Ideally we should generate URLs using the `url` library's formatter. But for now this fixes the issue.


mikhailxyz (approved):
Oops, sorry 🤦

PR #564, opened 2022-09-19
fix: remove pairing TODOs

iamacook:
## What it solves

Removes TODOs

## How this PR fixes it

TODOs have been moved to an issue https://github.com/safe-global/web-core/issues/563 instead as we intend to improve this in the future.


schmanu (approved):
👍

PR #566, opened 2022-09-19
fix: remove unnecessary queue dependency

iamacook:
## What it solves

Remove unnecessary queue loading dependency.

## How this PR fixes it

Queue polling does not rely on the `txHistoryTag` anymore.

Both the `txQueueTag` and `txHistoryTag` are [updated at the same time by the gateway](https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/safes/handlers/safes.rs#L29-L33), using the [latest transaction timestamps](https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/safes/handlers/safes.rs#L80-L143).

## How to test it

Queue a transaction then execute it. Observe that the transaction moves to the history as before.

### DIFF
--- a/src/hooks/loadables/useLoadTxQueue.ts
+++ b/src/hooks/loadables/useLoadTxQueue.ts
@@ -6,7 +6,7 @@ import { Errors, logError } from '@/services/exceptions'
 
 export const useLoadTxQueue = (): AsyncResult<TransactionListPage> => {
   const { safe, safeAddress, safeLoaded } = useSafeInfo()
-  const { chainId, txQueuedTag, txHistoryTag } = safe
+  const { chainId, txQueuedTag } = safe
 
   // Re-fetch when chainId/address, or txQueueTag change
   const [data, error, loading] = useAsync<TransactionListPage | undefined>(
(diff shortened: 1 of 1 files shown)


katspaugh:
Nice, thank you!

I've remembered now why/how we added it:
Originally, there was only the queueTag + a subscription to TxEvents.SUCCESS which reloaded the queue faster than the polling would. And then we looked at it and thought it was too complex and (mistakenly) added the historyTag as a replacement.

katspaugh (approved):

usame-algan:
We added the `historyTag` because the `queueTag` was/is not updating after a tx has been executed. Waiting for the next poll doesn't change it either so the only way to update the Queue is a hard reload of the page.

katspaugh:
I'll restore a tx event subscription then. I think that's the most robust.

iamacook:
> We added the `historyTag` because the `queueTag` was/is not updating after a tx has been executed. Waiting for the next poll doesn't change it either so the only way to update the Queue is a hard reload of the page.

This was not the case when I tested it. The gateway code suggests otherwise as well.

katspaugh:
Aren't we firing success when a tx is indexed?

iamacook:
> Aren't we firing success when a tx is indexed?

Very true, when a transaction has moved from the pending cache to historical list. I thought it was post execution.

@usame-algan and I tested the `txQueueTag` just now and it's certainly inconsistent (on staging, not production and predominantly with multisig transactions). I'm not entirely sure if relying on the subscription is solid though - if the tags don't update then the list won't have updated (referencing the gateway code).

We didn't notice any inconsistencies when including both tags. Wdyt about reverting to those instead? We should raise this with the backend team as well.

katspaugh:
Sounds good. 🆗

iamacook:
> A subscription, though, was also supposed to speed up the update, as otherwise you'd have to wait for up to 15 seconds.

I think we initially relied on the subscription and then altered to using the tags instead but I can't remember why. Does it index in time?

katspaugh:
Because it was "too complicated", AFAIR. But it did work in my tests.

PR #568, opened 2022-09-19
refactor: abstract context menu

iamacook:
## What it solves

Resolves TODO/duplicate code

## How this PR fixes it

The context menus from the transaction list and sidebar Safe list have been abstracted into one `ContextMenu` component.

## How to test it

- Expand a transaction and observe that the context menu is as before.
- Observe that the context menu in the sidebar Safe list is as before.

## Screenshots

image

image

### DIFF
--- a/src/components/transactions/TxDetails/TxData/Transfer/index.tsx
+++ b/src/components/transactions/TxDetails/TxData/Transfer/index.tsx
@@ -31,7 +31,7 @@ const TransferTxInfo = ({ txInfo, txStatus }: TransferTxInfoProps) => {
   return (
     <Box>
       <TransferTxInfoSummary txInfo={txInfo} txStatus={txStatus} />
-      <Box display="flex" alignItems="flex-end">
+      <Box display="flex" alignItems="center">
         <EthHashInfo address={address} shortAddress={false} hasExplorer showCopyButton />
         <TransferActions address={address} txInfo={txInfo} />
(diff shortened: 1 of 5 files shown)


src/components/transactions/TxDetails/TxData/Transfer/index.tsx
```diff
@@ -31,7 +31,7 @@ const TransferTxInfo = ({ txInfo, txStatus }: TransferTxInfoProps) => {
   return (
     <Box>
       <TransferTxInfoSummary txInfo={txInfo} txStatus={txStatus} />
-      <Box display="flex" alignItems="flex-end">
+      <Box display="flex" alignItems="center">
```
DiogoSoaress, line 34: Why this change? I suggest leaving it as it was

DiogoSoaress (approved):
I suggest revert the "ellipsis" button. It looks weird to be centered

src/components/transactions/TxDetails/TxData/Transfer/index.tsx
iamacook: The button is otherwise misaligned:

image

iamacook: It looks misaligned on my end without the styling adjustment. I'll revert my changes though and we can look at it separately.

PR #569, opened 2022-09-19
Chore: add txSender tests

katspaugh:
## What it solves
Some missing coverage for dispatchTxExecution + assertions for tx events in existing txSender tests.

### DIFF
--- a/src/services/tx/__tests__/txSender.test.ts
+++ b/src/services/tx/__tests__/txSender.test.ts
@@ -1,8 +1,10 @@
 import { setSafeSDK } from '@/hooks/coreSDK/safeCoreSDK'
 import Safe from '@gnosis.pm/safe-core-sdk'
+import { type TransactionResult } from '@gnosis.pm/safe-core-sdk-types'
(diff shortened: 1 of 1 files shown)


src/services/tx/__tests__/txSender.test.ts
```diff
@@ -130,6 +144,40 @@ describe('txSender', () => {
...
+
+      const proposedTx = await dispatchTxProposal('4', '0x123', '0x456', tx, '345')
+
+      expect(proposeTx).toHaveBeenCalledWith('4', '0x123', '0x456', tx, '0x1234567890')
+      expect(proposedTx).toEqual({ txId: '123' })
+
+      expect(txEvents.txDispatch).toHaveBeenCalledWith('SIGNATURE_PROPOSED', { txId: '123' })
+    })
+
+    it('should fail to propose', async () => {
```
iamacook, line 166: We should also test when new transactions fail to propose.

katspaugh: Added 👍

iamacook (approved):

PR #570, opened 2022-09-19
Chore: cloudflare redirects

katspaugh:
## What it solves
CloudFlare Pages have redirects – https://developers.cloudflare.com/pages/platform/redirects/ – but I'm afraid in our case this wouldn't work, since `/safe/balances` and `/rin:0x1234/balances` look the same to the matcher.

So I removed the `src/pages/safe` folder and moved the contents to just `src/pages`. That way we can avoid an infinite redirect.


DiogoSoaress:
Is there a way to test without merging the PR?

katspaugh:
Yes, @DiogoSoaress, and it seems to work.

DiogoSoaress (approved):

DiogoSoaress (approved):
Working now!

PR #571, opened 2022-09-19
fix: abstract trailing zero removal logic

iamacook:
## What it solves

Resolves `safeFormatUnits` FIXMEs

## How this PR fixes it

Trailing zero removal logic has been abstracted into its own function and tests added.

## How to test it

The unit tests should pass and all token amount display, i.e. Safe token amount, spending limit amount/spent and assets to send should display as before.

### DIFF
--- a/src/utils/formatters.ts
+++ b/src/utils/formatters.ts
@@ -5,6 +5,7 @@ import { formatAmount } from './formatNumber'
 const GWEI = 'gwei'
 
 export const _removeTrailingZeros = (value: string): string => {
+  // Match `.0` or `0*` after non-zero decimals by "Positive Lookbehind Assertion"
   return value.replace(/(\.0*|(?<=(\..*))0*)$/, '')
 }
 
--- a/src/utils/formatters.ts
+++ b/src/utils/formatters.ts
@@ -5,17 +5,7 @@ import { formatAmount } from './formatNumber'
 const GWEI = 'gwei'
 
 export const _removeTrailingZeros = (value: string): string => {
-  let [integer, fractions] = value.split('.')
-
-  if (!fractions) {
-    return value
-  }
-
-  while (fractions[fractions.length - 1] === '0') {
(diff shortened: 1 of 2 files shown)


src/utils/formatters.ts
```diff
@@ -4,6 +4,20 @@ import { formatAmount } from './formatNumber'
...
+export const _removeTrailingZeros = (value: string): string => {
+  let [integer, fractions] = value.split('.')
+
+  if (!fractions) {
+    return value
+  }
+
+  while (fractions[fractions.length - 1] === '0') {
+    fractions = fractions.substring(0, fractions.length - 1)
+  }
```
katspaugh, line 16: 😱 
Please avoid a loop to cut zeroes off the end of a string.

src/utils/formatters.ts
```diff
@@ -5,17 +5,7 @@ import { formatAmount } from './formatNumber'
...
-  if (!fractions) {
-    return value
-  }
-
-  while (fractions[fractions.length - 1] === '0') {
-    fractions = fractions.substring(0, fractions.length - 1)
-  }
-
-  return fractions ? `${integer}.${fractions}` : integer
+  return value.replace(/(\.0*|(?<=(\..*))0*)$/, '')
```
katspaugh, line 8: This regex looks complicated. Can you explain what it matches?
If you just need to trim a dot and zeros at the end, `/\.0+$/` should do the trick.

iamacook: We don't want to just trim a dot and zeros but also lookbehind the dot and any non-zero decimals for trailing zeros as well, e.g. `0.001000` => `0.001`.

It's taken from [here](https://stackoverflow.com/a/65172703/6585368). Is adding a summarised comment sufficient?

katspaugh: Ah, gotcha. A dot somewhere before the zeros but not necessary directly adjacent.

katspaugh (approved):

PR #573, opened 2022-09-19
fix: use app id instead of name

schmanu:
## What it solves
Uses the app ID from staging `61` and production `95` to fetch the app URL for the claiming app.

## How to test it
Click the token widget.
Hint: For prod the id exists but its disabled and not yet returned by the config service. But as soon as enabled it should work.

## Analytics changes
None


src/components/common/SafeTokenWidget/index.tsx
```diff
@@ -16,7 +17,8 @@ import SafeTokenIcon from './safe_token.svg'
 
 import css from './styles.module.css'
 
-const CLAIMING_APP_NAME = '$SAFE Claiming App'
+const isStaging = !IS_PRODUCTION && !cgwDebugStorage.get()
```
katspaugh, line 20: Good catch to include the debug toggle, it's easy to miss. We should probably move `cgwDebugStorage.get()` to the definition of `IS_PRODUCTION` itself.

katspaugh (approved):
👍

francovenica:
I can only check the change in ID in stg and is fine . I don't have access to django prod for now
image

francovenica:
Ok, we simply don't have access to django prod. No need to keep this ticket around.
QA done

PR #574, opened 2022-09-19
feat: Cypress wallet module

iamacook:
## What it solves

Missing E2E wallet

## How this PR fixes it

A custom `@web3-onboard` module using the `@truffle/hdwallet-provider` provider has been added to E2E environmeents. The `create_safe.cy.js` tests has been updated to utilise it.

## How to test it

Run the `create_safe` E2E test and observe that it passes.


iamacook:
The broken E2E test will be fixed by https://github.com/safe-global/web-core/pull/576.

src/tests/e2e-wallet.ts
```diff
@@ -0,0 +1,35 @@
+import type { ChainInfo } from '@gnosis.pm/safe-react-gateway-sdk'
+import type { WalletInit } from '@web3-onboard/common'
+
+import { CYPRESS_MNEMONIC } from '@/config/constants'
+import { getRpcServiceUrl } from '@/hooks/wallets/web3'
+
+const WALLET_NAME = 'E2E Wallet'
+
+const e2eWalletModule = ({ rpcUri }: ChainInfo): WalletInit => {
+  return () => {
```
DiogoSoaress, line 10: would opt for the simpler syntax
```
return () => ({
  ...
})
```

src/hooks/wallets/useOnboard.ts
```diff
@@ -124,16 +124,16 @@ export const useInitOnboard = () => {
...
     const enableWallets = async () => {
       const { getSupportedWallets } = await import('@/hooks/wallets/wallets')
-      const supportedWallets = getSupportedWallets(chain.disabledWallets)
+      const supportedWallets = getSupportedWallets(chain)
       onboard.state.actions.setWalletModules(supportedWallets)
     }
 
     enableWallets()
-  }, [chain?.disabledWallets, onboard])
+  }, [chain, onboard])
```
DiogoSoaress, line 136: The previous logic is better as it narrows more the dependency array, no?

DiogoSoaress (commented):
The fix for the failing E2E in `loading_safe` was merged into `main` so updating this branch should fix it.

src/tests/e2e-wallet.ts
iamacook: I'll keep it as is to follow the general style of other modules.

src/hooks/wallets/useOnboard.ts
iamacook: We are now not just relying on the `disabledWallets` but also the `rpcUri` within `getSupportedWallets`. The `chain` only changes when switching chain anyway. It should act the same

DiogoSoaress (approved):
Great!!! 🚀

PR #577, opened 2022-09-20
fix: formatting no currency

iamacook:
## What it solves

Resolves formatting 0 amounts

## How this PR fixes it

Instead of comparing against the minimum denominator of a currency, the current 'part' of the formatted currency is compared against `0` as that could be the value.

The minimum denominator of a currency is now derived from `1` as well.

## How to test it

All currencies should display as before, especially when no amount is present in a Safe and the `formatNumber` test should pass.

### DIFF
--- a/src/utils/formatNumber.ts
+++ b/src/utils/formatNumber.ts
@@ -114,15 +114,13 @@ export const formatAmountWithPrecision = (
 
 // Fiat formatting
 
-const getMinimumCurrencyDenominator = (number: string | number, currency: string): number => {
-  const float = Number(number)
-
+const getMinimumCurrencyDenominator = (currency: string): number => {
   const formatter = new Intl.NumberFormat(undefined, {
     style: 'currency',
     currency,
   })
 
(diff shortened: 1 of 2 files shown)


src/utils/formatNumber.ts
```diff
@@ -114,19 +116,17 @@ export const formatAmountWithPrecision = (
...
-  const float = Number(number)
-
+const getMinimumCurrencyDenominator = memoize((currency: string): number => {
   const formatter = new Intl.NumberFormat(undefined, {
     style: 'currency',
     currency,
   })
 
-  const fraction = formatter.formatToParts(float).find(({ type }) => type === 'fraction')
+  const fraction = formatter.formatToParts(1).find(({ type }) => type === 'fraction')
```
DiogoSoaress: would be easier to understand if
`fraction` -> `hasFraction`
or
`currencyHasFraction`

DiogoSoaress (approved):

src/utils/formatNumber.ts
iamacook: This isn't a flag. We require `fraction.value.length` on line 130.

PR #578, opened 2022-09-20
fix: Hide result length from tx list when there are no results

usame-algan:
## What it solves

Small fix for the tx list displaying "0" when there are no results.

## How to test

1. Open the Queue on a safe with no transactions
2. Observe no "0" visible in content area

## Screenshots

Before:
Screenshot 2022-09-20 at 10 36 43

After:
Screenshot 2022-09-20 at 10 36 34


katspaugh (approved):
Oops, my bad. Thanks!

PR #585, opened 2022-09-20
Fix: recommended gas limit

### LINKED ISSUE #584: Reset to recomended gas is not re-estimating
## Bug description
In the advanded options if you change the gas limit, a button shows up to reset that value to "Recomended gas limit". But what the button does is just setting the last valid value it had, even if that value was modified

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Any

## Steps to reproduce
 1. Go to a safe 1/x
 2. Create any tx
 3. Edit adv parameters and take notes on the value (lets say 65k for a regular send funds tx)
 4. Edit the Gas limit field to 25k and click the button (goes back to 65k)
 5. Edit the value again to 25k confirm
 6. Enter into the adv params again
 7. Edit the value gas limit to 35k and click the button once more 

## Expected result
Given what the tool tip says, is expected to give you the initial estimation (for this example 65k)

## Obtained result
It sets to the last valid value saved, in this example 25k

## Screenshots
image

katspaugh:
## What it solves

Resolves #584

## How this PR fixes it

Apparently we lost the right gas limit param at some point. I'm now passing it explicitly to the form.

### DIFF
--- a/src/components/tx/AdvancedParams/AdvancedParamsForm.tsx
+++ b/src/components/tx/AdvancedParams/AdvancedParamsForm.tsx
@@ -16,7 +16,7 @@ type AdvancedParamsFormProps = {
   params: AdvancedParameters
   onSubmit: (params: AdvancedParameters) => void
   recommendedNonce?: number
-  estimatedGasLimit?: string
+  recommendedGasLimit?: AdvancedParameters['gasLimit']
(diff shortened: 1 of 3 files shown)


usame-algan (approved):
Nice catch 👍

PR #588, opened 2022-09-21
Fix: delay validation error in AddressInput

katspaugh:
## What it solves

Suppress validation errors when an ENS domain is entered and is being resolved.
If ENS failed to resolve, display that specific error.

## How this PR fixes it

Adds a debounce for regular validation errors, but there's no delay when the input becomes valid or an ENS name has been entered and is resolving.

## How to test it
**Case 1:**

* Open the Send Funds form
* Enter a valid ENS domain (type quickly!)
* Observe that the field remains green the whole time

If you type slowly, it will still show a validation error eventually.

**Case 2:**
* Open the Send Funds form
* Enter a bogus ENS domain
* Observe that the field remains green while the ENS is resolving but becomes red and shows an ENS error after a couple seconds

## Analytics changes
None

## Screenshots
Screenshot 2022-09-21 at 09 26 46

### DIFF
--- a/src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx
+++ b/src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx
@@ -67,6 +67,7 @@ const SendAssetsForm = ({ onSubmit, formData }: SendAssetsFormProps): ReactEleme
   const formMethods = useForm<SendAssetsFormData>({
     defaultValues: { ...formData, [SendAssetsField.type]: SendTxType.multiSig },
     mode: 'onChange',
+    delayError: 500,
   })
   const {
     register,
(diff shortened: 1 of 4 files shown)


src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx
```diff
@@ -67,6 +67,7 @@ const SendAssetsForm = ({ onSubmit, formData }: SendAssetsFormProps): ReactEleme
   const formMethods = useForm<SendAssetsFormData>({
     defaultValues: { ...formData, [SendAssetsField.type]: SendTxType.multiSig },
     mode: 'onChange',
+    delayError: 500,
```
katspaugh, line 70: @iamacook FYI this worked for other fields, but not for the address input. So I had to add a manual debounce.

iamacook (commented):
Looks good to me and works as expected in the following places (where the component is used):

- Create/load Safe flows
- Send assets/NFTs form
- Adding address book entries
- Adding spending limits

I think we should adjust the tests a little bit though.

iamacook (approved):

PR #590, opened 2022-09-21
Feature/safe apps transaction list

kiarashghasemi:
## What it solves

Resolves # 1


mikhailxyz (changes requested):
I ran the PR locally and could see some misalignment in the bar, could you please fix it?

PR #598, opened 2022-09-21
fix: simulation of queued transactions

### LINKED ISSUE #597: Simulation: doesn't work with partially signed txns
For partially signed transactions in the queue we need to overwrite the current nonce of the safe to match with the transaction nonce.
For instance if you simulate a transaction with nonce 4 but there are still transactions with nonce 3 and 2 before that in the queue.

schmanu:
## What it solves
-(partially) signed transactions, which are not at the top of the queue. Could not be simulated successfully. They would run into an GS026 error.

Resolves #597 

## How this PR fixes it
- If a transaction has signatures we add those to the simulated transactions signatures
- We check if the simulated txs nonce is higher than the safe's nonce.
If this is true, we overwrite the nonce for the simulation.

## How to test it
- Queue 2 or more transactions and sign them.
- now try to simulate the transaction with the higher nonce

## Analytics changes
None

### DIFF
--- a/src/components/tx/TxSimulation/utils.ts
+++ b/src/components/tx/TxSimulation/utils.ts
@@ -166,6 +166,7 @@ const getNonceOverwrite = (params: SimulationTxParams): number | undefined => {
   Safe storage layout can be found here:
   https://github.com/gnosis/safe-contracts/blob/main/contracts/libraries/GnosisSafeStorage.sol */
 export const THRESHOLD_STORAGE_POSITION = hexZeroPad('0x4', 32)
+export const THRESHOLD_OVERWRITE = hexZeroPad('0x1', 32)
 /* We need to overwrite the nonce if we simulate a (partially) signed transaction which is not at the top position of the tx queue.
   The nonce can be found in storage slot 5 and uses a full 32 bytes slot. */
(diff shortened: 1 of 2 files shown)


src/components/tx/TxSimulation/utils.ts
```diff
@@ -151,6 +149,43 @@ const isOverwriteThreshold = (params: SimulationTxParams) => {
...
+export const NONCE_STORAGE_POSITION = hexZeroPad('0x5', 32)
+
+const getStateOverwrites = (params: SimulationTxParams) => {
+  const nonceOverwrite = getNonceOverwrite(params)
+  const isThresholdOverwrite = isOverwriteThreshold(params)
+
+  const storageOverwrites: Record<string, string> = {} as Record<string, string>
+
+  if (isThresholdOverwrite) {
+    storageOverwrites[THRESHOLD_STORAGE_POSITION] = hexZeroPad('0x1', 32)
```
katspaugh, line 180: `const FALLBACK_THRESHOLD = hexZeroPad('0x1', 32)` maybe?

francovenica:
LGTM, tx simulation works fine for the tx with the higher nonce now.

PR #600, opened 2022-09-21
Feature: Safe apps transactions bar

### LINKED ISSUE #453: Use transaction list from Safe Apps
## What is the feature about
Migrate the transaction list review feature for Safe Apps

## The list of requirements

- [ ] Show transaction list
- [ ] Allow to use the transaction list without leaving the Safe App

## Links
Old epic ticket https://github.com/safe-global/safe-react/issues/3940
Current epic https://github.com/safe-global/safe-react/issues/4044

### LINKED ISSUE #594: Incorrect routing when adding a custom app -> https://gnosis-safe.io/app/
## Bug description
When a user of https://web-core.pages.dev/ clicks to Add a Custom Safe App they are routed to https://gnosis-safe.io/app/ 

## Environment
 - Browser: Chrome | Firefox
 - Wallet: MetaMask
 - Chain: Any

## Steps to reproduce
 1. Go to https://web-core.pages.dev/rin:0x28924a93f932ED76a72fadA69d23587FAe406912/apps
 2. Click Add custom app

## Expected result
The Add custom app modal should appear with a form to enter the App URL and App Name.

## Obtained result
The user is routed to a default hosted version of the app at https://gnosis-safe.io/app/rin:0x28924a93f932ED76a72fadA69d23587FAe406912/apps

mikhailxyz:
## What it solves

Resolves #453 
Resolves #594 

## How this PR fixes it
For the transaction bar, it simply reuses the transaction page in MUI's accordion component in the app frame
For the redirect thing, it completely removes the references to safe-react in the app

## How to test it
Play around with the bar and transactions in your favourite app

## Analytics changes
None

## Screenshots
image
image
image


src/components/common/PaginatedTxns/index.tsx
```diff
@@ -49,7 +49,7 @@ const TxPage = ({
         </Box>
       )}
 
-      {page && page.results.length > 0 && <TxList items={page.results} />}
+      {!!page?.results.length && <TxList items={page.results} />}
```
mikhailxyz, line 52: i fixed it on my branch before it got fixed on main and i think it looks tidier with optional chaining, lmk if I should revert the fix

katspaugh: Chaining for sure, but we should avoid `!!` as it's somewhat cryptic.
Please use `page?.results.length > 0`

src/components/safe-apps/TransactionQueueBar/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const [dismissedByUser, setDismissedByUser] = useState(false)
+
+  const toggleQueueBar = (): void => {
+    setExpanded((prev) => !prev)
+  }
+
+  const { page = { results: [] } } = useTxQueue()
+  const queuedTxCount = getQueuedTransactionCount(page)
+
+  return !dismissedByUser ? (
```
katspaugh, line 23: Code style: I would use an early return here.

src/components/safe-apps/TransactionQueueBar/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+                sx={{ transform: expanded ? 'rotate(180deg)' : undefined }}
+              >
+                <ExpandLessIcon />
+              </IconButton>
+              <IconButton onClick={() => setDismissedByUser(true)} aria-label="dismiss transaction queue bar">
+                <CloseIcon />
+              </IconButton>
+            </AccordionSummary>
+            <AccordionDetails>
+              <PaginatedTxns useTxns={useTxQueue} disableTopActionMargins />
```
katspaugh, line 64: > disableTopActionMargins

Can we just give the list a top margin here to compensate for the negative margin inside? Passing a prop just for a minor style detail seems a bit excessive.

(I realize we probably shouldn't have used negative margins in the first place).

katspaugh (approved):
Looks great! Left some minor comments.

src/components/common/PaginatedTxns/index.tsx
mikhailxyz: It seems like `page?.results.length > 0` check is not enough for typescript to ensure that `page` is defined, same for `Boolean(page?.results.length)`. I'll just revert the change then

src/components/safe-apps/TransactionQueueBar/index.tsx
mikhailxyz: Theoretically, we could, but this would require making the transaction bar component heavily dependent on the implementation details of `PaginatedTxns`, so if you change the margins in there, you also have to adjust them in the bar. So I think this is no good.

An alternative solution could be explicitly passing the styles for the top action, so each component consumer sets their own. What do you think?

katspaugh: True, valid point. Let's keep the prop for now then, I'll see if we can get rid of the negative margins.

src/components/safe-apps/AppFrame/TransactionQueueBar/index.tsx
```diff
@@ -24,12 +24,14 @@ const TransactionQueueBar = (): ReactElement | null => {
     return null
   }
 
+  // if you inline the expression, it will split put the `queuedTxCount` on a new line
+  // and make it harder to find this text for matchers in tests
+  const barTitle = `(${queuedTxCount}) Transaction Queue`
```
katspaugh, line 29: Btw, please don't Capitalize Nouns. We decided against it.

iamacook (approved):

JagoFigueroa:
Espectacular trabajo for both implementations bandido ;)

I found this issue where executing the first tx of a queue will go back to its original state after indexing but as this is happening in the og queue as well I will let señor Franco know to handle it from his side (maybe it is a known issue): 

https://user-images.githubusercontent.com/92332710/191903075-140b0b34-5834-40de-a13e-84b228f373a8.mp4

Cheers!

katspaugh:
@safe-global/safe-web ☝️ that's why we probably need to refresh the queue when the success tx event is fired.

iamacook:
> @safe-global/safe-web ☝️ that's why we probably need to refresh the queue when the success tx event is fired.

Whilst I see that it is the most "correct" solution. If we rely on the subscription, a batch would trigger a mass number of requests. Readding the history tag, whilst not the most ideal solution, seems like a good middle ground until we discuss this with backend.

mikhailxyz:
I fixed using @iamacook's invention

PR #601, opened 2022-09-21
fix: use processing instead of mining

iamacook:
## What it solves

Accomodating statuses for PoS

## How this PR fixes it

All instances of mining/mined have been changed to processing/processed in order to accomodate for PoS.

## How to test it

- Create and execute a transaction. Observe the notification regards it as "processing" the transaction and the status is now also "Processing".
- Adding a spending limit now says "Spending limit transactions only appear in the interface once they are successfully processed and indexed." in the modal.
- Transactions that timeout (by underpricing the gas) should throw "Transaction not processed in 6.5 minutes. Be aware that it might still be processed."

### DIFF
--- a/src/components/create-safe/status/CreationStatus.tsx
+++ b/src/components/create-safe/status/CreationStatus.tsx
@@ -35,7 +35,7 @@ const getStep = (status: SafeCreationStatus) => {
         description: 'Waiting for wallet connection',
         instruction: 'Please make sure your wallet is connected on the correct network.',
       }
-    case SafeCreationStatus.MINING:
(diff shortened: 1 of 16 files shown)


DiogoSoaress (approved):
PoS proof

francovenica:
The "Processing" and the new "Spending limit" message are fine.

I waited over 8 minutes and I cannot get the "Transaction not processed in 6.5 mins" message. I know this ticket is only about the wording, but still this won't allow me to be sure the change has been done. I've seen the change of the message in the file so I'll pass this ticket and report the lack of message as a different issue

PR #605, opened 2022-09-22
style: update light/dark illustrations

iamacook:
## What it solves

Resolves outdated illustrations

## How this PR fixes it

The app-wide illustrations have been updated to match the new theme in light/dark mode, as described [here](https://5afe.slack.com/archives/C03DAGWJCR5/p1663773208255179).

## How to test it

The following should be visibly updated and different in light/dark mode:

- No pending transactions in the dashboard
- Spending limit settings (when none are set)
- Adding a custom app
- No NFTs (same image in light/dark mode)


katspaugh:
As I mentioned on Slack, we should update the SVGs' code to use CSS vars to avoid having two sets of same images for the light/dark modes.

katspaugh:
They look great!

This one also needs to be updated but it can be done in another PR. (Also something happened with the spacing there.)

Screenshot 2022-09-22 at 12 09 14

katspaugh (approved):
🚀

katspaugh:
Screenshot 2022-09-22 at 12 12 24 – custom app looks no bueno

iamacook:
> Screenshot 2022-09-22 at 12 12 24 – custom app looks no bueno

Am already on it. Something is up with overriding `stroke` via classes.

src/styles/globals.css
```diff
@@ -63,14 +63,4 @@ input[type='number'] {
...
-}
-
-.illustration-light-stroke {
-  stroke: var(--color-border-main);
-}
-
-.illustration-background-stroke {
-  stroke: var(--color-logo-background);
-}
+/* It's not possible to set `stroke` via a class as it overrides other properties */
```
katspaugh, line 66: What do you mean? Which other properties does it override?

iamacook: By setting `stroke` via a class, it overrides all other associated stroke properties, i.e. `stroke-width`, `stroke-linecap` etc.

iamacook: It seems like we've found a way. `stroke` needs to exists _as well as_ the class. I'll revert my change above.

src/components/common/PagePlaceholder/index.tsx
```diff
@@ -3,17 +3,17 @@ import { Typography } from '@mui/material'
...
 
-const PagePlaceholder = ({ imageUrl, text, children }: PagePlaceholderProps): ReactElement => {
+const PagePlaceholder = ({ img, text, children }: PagePlaceholderProps): ReactElement => {
   return (
     <div className={css.container}>
-      <img src={imageUrl} alt="background" />
+      {img}
 
-      <Typography variant="h3" m={3}>
+      <Typography variant="body1" color="primary.light">
```
schmanu, line 16: I think in the designs we always have 16pxs margin-top:

```suggestion
      <Typography variant="body1" mt={2} color="primary.light">
```


localhost_3000_

iamacook:
> In the PR description you mention the empty address book screen.
> 
> But that does not exist. Also I see no changes in this PR in the address book. Is it out of scope? 

Sorry, I copied the description over from my other PR. That won't be in this.

francovenica:
I see the icons in the spending limit to be missaligned relative to their text, the first and last icon. The icons looks like they are too high up
image

francovenica:
I'll report the aligment issue in a new ticket. Closing this one

PR #607, opened 2022-09-22
chore: add typescript-css-module-plugin

schmanu:
## What it solves
Currently if non existing css class is being used from a css module import we do not get any error / warning about it.

## How this PR fixes it
Adds `typescript-css-module-plugin` to enable compile errors if non-existing css classes are being referenced from css modules.

## How to test it
- install the new package
- restart vs code and choose the workspace version of typescript as the used typescript version
- change a imported css module class name to something invalid and wait for the compilation error


DiogoSoaress (approved):
Noice!

PR #609, opened 2022-09-22
fix: display better single tx error message

DiogoSoaress:
## What it solves
1. _Error details color_
Screenshot 2022-09-22 at 11 21 23

2. _the query parameters (`txId`) were malformed and consequently the page couldn't fetch the txDetails_
Screenshot 2022-09-23 at 14 25 31

3. _long transaction id string was not wrapped and was overflowing_ & 4. _transaction not contained in the current Safe_
Screenshot 2022-09-22 at 17 32 28

## How this PR fixes it
Changes the color used in the `<Error />` details
Allow word-wrap in `<Error />`

## How to test it
Visit https://web-core.pages.dev/rin:0x87a57cBf742CC1Fc702D0E9BF595b1E056693e2f/transactions/tx?id%3F=multisig_0x87a57cBf742CC1Fc702D0E9BF595b1E056693e2f_0x236da79434c398bf98b204e6f3d93d4ec76719f0b25814a537872098d3cc0d99


src/pages/transactions/tx.tsx
```diff
@@ -62,7 +62,9 @@ const SingleTransaction: NextPage = () => {
       <Breadcrumbs Icon={TransactionsIcon} first="Transactions" second="Details" firstLink={breadcrumbsLink} />
 
       {(error || !isCurrentSafeTx) && !loading ? (
-        <ErrorMessage error={error}>Failed to load transaction {transactionId}</ErrorMessage>
+        <ErrorMessage error={!isCurrentSafeTx ? new Error('Error: tx Id not found in this Safe') : error}>
```
katspaugh, line 65: `tx Id` > `transaction with this id`

DiogoSoaress, line 65: Not totally sure if we should be overwriting the `error` here but we were getting
> undefined: undefined

src/components/tx/ErrorMessage/styles.module.css
```diff
@@ -4,6 +4,7 @@
   padding: var(--space-2);
   margin: var(--space-2) 0;
   border-radius: 4px;
+  word-break: break-all;
```
katspaugh, line 7: Instead of this, I would move the echoing of the id to the details of the error.

src/pages/transactions/tx.tsx
katspaugh: I.e. here you could say `transaction with id ${txId} not found in this Safe`

DiogoSoaress:
> here you could say transaction with id ${txId} not found in this Safe

If the transaction is not found in the Safe by the tx service, we were displaying it correctly. This message was for the case of a malformed query parameter.
For example instead of `id=multisig_0x87...`
if we pass `foo=multisig_0x87...`

Which makes this Error message incorrect... Can we say "Not able to retrieve transaction details" ?

DiogoSoaress:
I've addressed the comments but I will also include tests to make it more robust

usame-algan (approved):
Lets tackle the tests in a separate PR

src/components/transactions/SingleTx/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+      </ErrorMessage>
+    )
+  }
+
+  if (txDetails) {
+    return <SingleTxGrid txDetails={txDetails} />
+  }
+
+  return (
+    <ErrorMessage error={error || Error('Resource not found. Please review the url.')}>
```
katspaugh, line 64: What resource? Please write a more specific message here.

Btw URL is spelled in all caps.

src/components/transactions/SingleTx/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const isCurrentSafeTx = sameAddress(txDetails?.safeAddress, safeAddress)
+
+  if (loading) {
+    return <CircularProgress />
+  }
+
+  if (!isCurrentSafeTx && transactionId) {
+    return (
+      <ErrorMessage error={Error(`Transaction with id ${transactionId} not found in this Safe `)}>
```
katspaugh, line 53: `Error` -> `new Error`

src/components/transactions/SingleTx/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+const SingleTx = () => {
+  const chainId = useChainId()
+  const router = useRouter()
+  const { id } = router.query
+  const transactionId = Array.isArray(id) ? id[0] : id
+  const { safe, safeAddress } = useSafeInfo()
+
+  const [txDetails, error, loading] = useAsync<TransactionDetails>(
+    () => {
+      if (!id || !transactionId) return
```
katspaugh, line 38: `id` shouldn't be part of this hook it it's not passed in the dependencies.
`transactionId` is enough in any case.

src/components/transactions/SingleTx/index.tsx
katspaugh, line 53: This is code duplication. Just define the error like so:

```
const [txDetails, txDetailsError, loading] = useAsync...
const error = transactionId && !isCurrentSafeTx ? new Error('Transaction not found') : txDetailsError
```

katspaugh (changes requested):
See comments

src/components/transactions/SingleTx/index.tsx
DiogoSoaress: from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

> Note: Error() can be called with or without [new](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new). Both create a new Error instance.

that's why I didn't include the `new`. I'll change it for consistency across the codebase.

src/components/transactions/SingleTx/index.tsx
DiogoSoaress: Changed to 'Couldn't retrieve the transaction details. Please review the URL.'

src/components/transactions/SingleTx/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    },
+    [transactionId, safe.txQueuedTag, safe.txHistoryTag],
+    false,
+  )
+  const isCurrentSafeTx = sameAddress(txDetails?.safeAddress, safeAddress)
+
+  const error =
+    transactionId && !isCurrentSafeTx
+      ? new Error(`Transaction  with id ${transactionId} not found in this Safe`)
+      : txDetailsError || new Error("Couldn't retrieve the transaction details. Please review the URL.")
```
katspaugh, line 47: Why are you falling back to `new Error` unconditionally? If there's no error, it should display any message. It's either loading, or there's an error.

DiogoSoaress: Reason is when the URL is malformed there is no TransactionDetails request but we could still inform that it is worth rechecking the URL. Should I simply remove this fallback? 

Screenshot 2022-09-23 at 14 58 32

katspaugh: What is a "malformed" URL? Missing id? If you know the id is missing, you can use it is as a condition. As it stands, you're just saying "show data if there's data, otherwise show this made up error".

katspaugh (approved):
Thanks! 🙏

PR #613, opened 2022-09-22
fix: Use eth_sign when connected via WC

### LINKED ISSUE #503: Wallet connect: can not sign if the wallet is connected via WC

usame-algan:
## What it solves

Resolves #503 

## How this PR fixes it

Uses `eth_sign` when signing a transaction via WalletConnect

## How to test it

1. Open the Safe
2. Connect to the safe via mobile MM app
3. Create/Sign a transaction
4. Observe no error
5. Observe a window popping up on the mobile MM app

### DIFF
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -15,7 +15,7 @@ import useGasLimit from '@/hooks/useGasLimit'
 import useSafeInfo from '@/hooks/useSafeInfo'
 import ErrorMessage from '@/components/tx/ErrorMessage'
 import AdvancedParams, { type AdvancedParameters, useAdvancedParams } from '@/components/tx/AdvancedParams'
-import { isHardwareWallet, isSafeMobileWallet, isSmartContractWallet } from '@/hooks/wallets/wallets'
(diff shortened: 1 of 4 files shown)


src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -101,7 +101,8 @@ const SignOrExecuteForm = ({
   const onSign = async (): Promise<string> => {
     const [connectedWallet, createdTx, provider] = assertSubmittable()
 
-    const shouldEthSign = isHardwareWallet(connectedWallet) || isSafeMobileWallet(connectedWallet)
+    const shouldEthSign =
+      isHardwareWallet(connectedWallet) || isSafeMobileWallet(connectedWallet) || isWalletConnect(connectedWallet)
```
iamacook, line 105: I suggest we create a helper for this with a brief note as to why each requires it. Would be nice to add a test case for it as well.

iamacook (approved):

PR #617, opened 2022-09-23
Chore: serve from /app

katspaugh:
## What it solves
In preparation for a potential temporary deployment to the old domain in a subfolder rather than the root, I've added `/app` as the base path.

We could parametrize it as a BASE_PATH var, but since it's temporary, I think it's easier to keep the string and then just grep and clean it up once we move to a new domain w/o a subfolder.


katspaugh:
CloudFlare deployment doesn't work properly but that's expected because the root path needs to be set there as well. I'll leave it as `/` for now, so as to not break the other branches.

PR #621, opened 2022-09-23
fix: adjust SDK error message + fix onboard

iamacook:
## What it solves

Resolves unclear error message on older Safes

## How this PR fixes it

When the Core SDK is not initialized (it only supports >= v1.1.1 Safes), the error message is now "The Safe SDK could not be initialized. Please be aware that we only support >= v1.1.1 Safes." when transacting.

Note: the temporary `@web3-onboard` fix for `joi` schemas has been installed as well. This should be upgraded once they next publish a release.

## How to test it

- Open a v1.0/1.1.0 Safe and create a transaction. Observe the error message.
- Open a >= v1.1.1 Safe then switch to a < v1.1.1 Safe and repeat the above. Observe the error message.


package.json
```diff
@@ -52,16 +52,16 @@
...
+    "@web3-onboard/coinbase": "^2.1.3-alpha.1",
+    "@web3-onboard/core": "2.8.5-alpha.2",
+    "@web3-onboard/fortmatic": "^2.0.14-alpha.1",
+    "@web3-onboard/injected-wallets": "^2.2.4-alpha.2",
+    "@web3-onboard/keystone": "^2.3.1-alpha.1",
+    "@web3-onboard/ledger": "^2.3.1-alpha.1",
+    "@web3-onboard/portis": "^2.1.3-alpha.1",
+    "@web3-onboard/torus": "^2.1.3-alpha.1",
+    "@web3-onboard/trezor": "^2.3.1-alpha.1",
+    "@web3-onboard/walletconnect": "^2.1.3-alpha.1",
```
iamacook, line 64: An official release will be available [soon](https://github.com/blocknative/web3-onboard/issues/1287). Although "alpha", it [only updates a peer dependency](https://github.com/blocknative/web3-onboard/pull/1290).

usame-algan: Should we wait with the update until the official release is published and its not alpha anymore?

iamacook: As seen in [their PR](https://github.com/blocknative/web3-onboard/pull/1290), they are only updating `joi` to the same version across packages. If we prefer to not use the alpha versions then we need to have static dependency versions instead.

usame-algan: I would prefer we don't upgrade to pre-releases unless its a critical fix that is needed. Which command did you run to upgrade the versions? I tried `yarn upgrade` but it doesn't change anything so maybe the caret already excludes postfix versions.

iamacook: The inclusion of the caret automatically upgrades to all minor/patch versions and the current version is broken. I referenced their PR for these packages.

usame-algan (approved):

PR #622, opened 2022-09-23
Fix: Filter out the EIP712Domain type from `eth_signTypedData` RPC payload before passing to the hash function

### LINKED ISSUE #363: Migrate search and pin Safe App
## Overview
Migrate bookmark and Safe Apps search to the new codebase.

## Requirements

- [ ] Search in the Safe Apps list
- [ ] Pin Safe App as favourite
- [ ] Adapt to new Safe Apps list design, provided in desings section

## Designs
https://www.figma.com/file/qfg41YAO8mLqphHl2Pbp4C/Apps-Main?node-id=1223%3A2694

## Links
- https://github.com/safe-global/safe-react/issues/4044

mikhailxyz:
## What it solves

From ethers.js discord:
> The types are not allowed to be recursive, so ever type must either be used by another type, or be the primary type. And there must only be one type that is not used by any other type.

Resolves #363

### DIFF
--- a/src/components/safe-apps/SafeAppsSignMessageModal/ReviewSafeAppsSignMessage.tsx
+++ b/src/components/safe-apps/SafeAppsSignMessageModal/ReviewSafeAppsSignMessage.tsx
@@ -53,6 +53,11 @@ const ReviewSafeAppsSignMessage = ({
       txData = signMessageDeploymentInstance.interface.encodeFunctionData('signMessage', [hashMessage(message)])
     } else if (isTypedMessage) {
       const typesCopy = { ...message.types }
+
+      // We need to remove the EIP712Domain type from the types object
+      // Because it's a part of the JSON-RPC payload, but for the `.hash` in ethers.js
+      // The types are not allowed to be recursive, so ever type must either be used by another type, or be
+      // the primary type. And there must only be one type that is not used by any other type.
       delete typesCopy.EIP712Domain
       txData = signMessageDeploymentInstance.interface.encodeFunctionData('signMessage', [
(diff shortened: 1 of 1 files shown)


katspaugh (approved):

katspaugh:
Can we somehow make it strict using types? Or maybe add a unit test.

mikhailxyz:
> Can we somehow make it strict using types? Or maybe add a unit test.

We can add a unit test to ensure that we're removing it from the `types`, yeah. Regarding types, I don't think it's possible because, technically, the type is correct, it's the hashing algorithm implementation that is incompatible with the JSON-RPC payload type.

JagoFigueroa:
Good news is that I was able to vote señores:

Screenshot 2022-09-29 at 09 27 15


But unfortunately It did not reflect in real time on snapshot's side as I got an error:

Screenshot 2022-09-29 at 09 23 34

My guess is that this is not on our side so if you confirm that that is the case we can close this ticket 😄

mikhailxyz:
@bonustrack @samuveth could you please check this on your side?

samuveth:
@mikhailxyz where to test this? I tested it on https://gnosis-safe.io but I'm still getting the old error.

mikhailxyz:
> @mikhailxyz where to test this? I tested it on https://gnosis-safe.io but I'm still getting the old error.

Yes. Interesting, I don't get an error. Could you try to perform a hard refresh?
image

samuveth:
@mikhailxyz I tried to hard refresh and I tried using WalletConnect instead of Snapshot safe app, but I get the same error either way.

PR #623, opened 2022-09-23
feat: invalid mastercopy notification

iamacook:
## What it solves

Notifying users of invalid mastercopies

## How this PR fixes it

When switching Safes, the mastercopy is verified for validity and a notification dispatched if it is invalid.

## How to test it

Open a Safe with an invalid mastercopy (`matic:0x5B7e9a88237Ca67591e751186b10623Db5653045`) and observe the invalid mastercopy notification. Switching between Safes should trigger this only when the master copy is invalid. The CLI link should open in a new window.

Any currently present links, i.e. to update the Safe, should open in the current tab.

## Screenshots

image

### DIFF
--- a/src/components/common/Notifications/index.tsx
+++ b/src/components/common/Notifications/index.tsx
@@ -8,7 +8,7 @@ import NextLink from 'next/link'
 import ChevronRightIcon from '@mui/icons-material/ChevronRight'
 import { OVERVIEW_EVENTS } from '@/services/analytics/events/overview'
 import Track from '../Track'
-import { isRelativeUrlWithoutProtocol } from '@/utils/url'
+import { isRelativeUrl } from '@/utils/url'
 
 const toastStyle = { position: 'static', margin: 1 }
 
@@ -23,7 +23,7 @@ export const NotificationLink = ({
     return null
   }
 
-  const isExternal = !isRelativeUrlWithoutProtocol(link.href)
+  const isExternal = !isRelativeUrl(link.href)
 
   return (
     <Track {...OVERVIEW_EVENTS.NOTIFICATION_INTERACTION} label={link.title}>
--- a/src/components/common/Notifications/index.tsx
+++ b/src/components/common/Notifications/index.tsx
@@ -8,6 +8,7 @@ import NextLink from 'next/link'
 import ChevronRightIcon from '@mui/icons-material/ChevronRight'
 import { OVERVIEW_EVENTS } from '@/services/analytics/events/overview'
 import Track from '../Track'
+import { isRelativeUrlWithoutProtocol } from '@/utils/url'
 
 const toastStyle = { position: 'static', margin: 1 }
 
@@ -22,10 +23,16 @@ export const NotificationLink = ({
     return null
   }
 
+  const isExternal = !isRelativeUrlWithoutProtocol(link.href)
+
   return (
     <Track {...OVERVIEW_EVENTS.NOTIFICATION_INTERACTION} label={link.title}>
       <NextLink href={link.href} passHref>
-        <Link className={css.link} onClick={onClick}>
+        <Link
+          className={css.link}
+          onClick={onClick}
+          {...(isExternal && { target: '_blank', rel: 'noopener noreferrer' })}
+        >
           {link.title} <ChevronRightIcon />
         </Link>
       </NextLink>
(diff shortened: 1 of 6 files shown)


src/components/common/Notifications/index.tsx
```diff
@@ -8,6 +8,7 @@ import NextLink from 'next/link'
 import ChevronRightIcon from '@mui/icons-material/ChevronRight'
 import { OVERVIEW_EVENTS } from '@/services/analytics/events/overview'
 import Track from '../Track'
+import { isRelativeUrlWithoutProtocol } from '@/utils/url'
```
katspaugh, line 11: Jesus, who called this function that.

src/hooks/useSafeNotifications.ts
```diff
@@ -37,6 +42,44 @@ const useSafeNotifications = (): void => {
       dispatch(closeNotification({ id }))
     }
   }, [dispatch, chainId, safeAddress, implementationVersionState, version, chain?.shortName])
+
+  /**
+   * Show a notification when the Safe master copy is not supported
+   */
+
+  const masterCopy = safe.implementation.value
```
schmanu, line 50: We should add a test for this

src/services/contracts/safeContracts.ts
```diff
@@ -12,11 +12,17 @@ import { LATEST_SAFE_VERSION } from '@/config/constants'
...
 import semverSatisfies from 'semver/functions/satisfies'
-import { SafeInfo, type ChainInfo } from '@gnosis.pm/safe-react-gateway-sdk'
+import { getMasterCopies, SafeInfo, type ChainInfo } from '@gnosis.pm/safe-react-gateway-sdk'
 import type { GetContractProps, SafeVersion } from '@gnosis.pm/safe-core-sdk-types'
 import { type Compatibility_fallback_handler } from '@/types/contracts/Compatibility_fallback_handler'
 import { type Sign_message_lib } from '@/types/contracts/Sign_message_lib'
 import { createEthersAdapter, isValidSafeVersion } from '@/hooks/coreSDK/safeCoreSDK'
+import { sameAddress } from '@/utils/addresses'
+
+export const isValidMasterCopy = async (chainId: string, address: string): Promise<boolean> => {
```
schmanu, line 22: You could copy the small test cases from safe-react: `safeVersion.test.ts`

schmanu (commented):
Looking good. I also prefer it as notification over banner! 

The unit tests are currently failing, though.

src/hooks/useSafeNotifications.ts
```diff
@@ -37,6 +42,44 @@ const useSafeNotifications = (): void => {
...
+
+  const [validMasterCopy] = useAsync(async () => {
```
usame-algan, line 52:
```suggestion
  const [validMasterCopy] = useAsync<boolean>(async () => {
```

src/hooks/useSafeNotifications.ts
```diff
@@ -37,6 +42,44 @@ const useSafeNotifications = (): void => {
...
+    if (masterCopy) {
+      return await isValidMasterCopy(chainId, masterCopy)
+    }
+  }, [chainId, masterCopy])
+
+  useEffect(() => {
+    if (validMasterCopy === undefined || validMasterCopy) {
+      return
+    }
```
usame-algan, line 61: I suggest to use the `loading` flag from `useAsync` here instead of checking if the return value is undefined

src/hooks/useSafeNotifications.ts
iamacook: As `useAsync` uses generics, we need not strictly type this as we only call `getMasterCopies` which returns a `boolean`:

image

src/hooks/useSafeNotifications.ts
iamacook: The `loading` flag is not reliabl:

loading flag

I even tried moving `setLoading` to the top of the `useEffect` to no avail. I think it's related to React 18's batched updates but `flushSync` didn't help either.

It's something we should look into at some point as I agree that it is a better flag to reference.

francovenica:
It looks good for me
Tried with 2 other invalid safes
avax:0x31d7a5194Fe60AC209Cf1Ce2d539C9A60662Ed6b
matic:0x6fB53537015FA77Dd0132D6532A1A0d392e67A99

PR #626, opened 2022-09-23
feat: page headers

iamacook:
## What it solves

Add page headers

## How this PR fixes it

The app-wide page headers have been added, as shown [here](https://5afe.slack.com/archives/C03DAGWJCR5/p1663773208255179).

`PageLayout`'s `.container main` styling adds the default padding for pages and the `main` element have therefore been moved outside of the page headers, sometimes into the page components themselves.

No extensive style adjustments have been made in order to keep this PR concise.

## How to test it

The following pages should have interactable (sticky) headers:

- Assets: tabs and currency selector
- Transactions: tabs and working batch execute/filter buttons respectively (the open filter should stick below the header on long lists)
- Address book: import, export and entry creation buttons (search function will be implemented later)
- Safe Apps: as before
- Settings: tabs

Safe Apps queue bar should retain batch execute functionality.

## Screenshots

image

image

image

image

image

image


src/components/address-book/AddressBookHeader/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+            >
+              Export
+            </Button>
+          </Track>
+
+          <Track {...ADDRESS_BOOK_EVENTS.CREATE_ENTRY}>
+            <Button
+              onClick={handleOpenModal(ModalType.ENTRY)}
+              variant="text"
+              sx={{ color: 'text.primary' }}
```
usame-algan, line 55:
```suggestion
              color="primary"
```
It also makes the button color green in dark-mode which is in line with [Figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=876%3A38473)

usame-algan (approved):
New page headers looks amazing! Good job 💯  🚀

katspaugh:
Screenshot 2022-09-26 at 10 54 31 – a small 1px mismatch with the sidebar header

The gap between the description and actions is a bit too big IMHO. It's even more evident on pages w/o tabs:
Screenshot 2022-09-26 at 10 56 14

On the NFTs page, we could get rid of the info block and move that text to the header, WDYT?

iamacook:
> a small 1px mismatch with the sidebar header

This seems to be alright on my end (left Firefox, right Chrome):

Named Safe:

image

Unknown Safe:

image

---

> The gap between the description and actions is a bit too big IMHO. It's even more evident on pages w/o tabs

If we reduce the height, it won't line up as nicely with the sidebar header. The address book will have a search bar in below the header so I think it's fine. I can implement it as soon as we get this merged.

---

> On the NFTs page, we could get rid of the info block and move that text to the header, WDYT?

I like the idea but there isn't enough room:

image

---

Do you have any thoughts on the above @TanyaEfremova?

katspaugh:
Reg. the NFT info, I meant we could move only the text, not the whole blue banner.

iamacook:
> Reg. the NFT info, I meant we could move only the text, not the whole blue banner.

What do you think of this?

nft header

katspaugh:
It should be the same font size.

src/components/common/PageHeader/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const PageHeader = ({
+  title,
+  subtitle,
+  helperText,
+  action,
+}: {
+  title: string
+  subtitle: string
+  helperText?: string
```
katspaugh, line 16: Instead of these two params, I would accept arbitrary children.

iamacook:
> It should be the same font size.

image

katspaugh (approved):
LFGTM!

TanyaEfremova:
@iamacook  
1. Address Book gap: there is a search missing, will it be implemented any time soon? if not, then we could reduce the gap by moving the buttons container to the left. In general, what's the top and bottom margin? Looks that the bottom-margin is a bit different than in the [design](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=5%3A1571)
2. I would not move the info banner to the header on the NFT page, to avoid creating inconsistencies, also it doesn't belong there semantically. The texts should use the same font size, as @katspaugh mentioned.

iamacook:
>     1. Address Book gap: there is a search missing, will it be implemented any time soon?

I will do so as soon as we merge this.

>     2. I would not move the info banner to the header on the NFT page, to avoid creating inconsistencies, also it doesn't belong there semantically. The texts should use the same font size, as @katspaugh mentioned.

So, you would prefer we keep the `Alert` or is this as expected:

image

katspaugh:
Sorry for the wild goose chase!

francovenica:
Looks good. 

Remove the horizontal scroll from the fiat selector in the "assets" 
image

francovenica:
Add some padding between the alert message and the NFT category
image

iamacook:
> Remove the horizontal scroll from the fiat selector in the "assets"

This is now fixed.

> Add some padding between the alert message and the NFT category

This screenshot was taken on the 'old' style. I've just confirmed that there is padding on the this PR.

> Note: Idk if this ticket is related to this, but since I've seen that the alert message was discussed in this ticket I'd assume is part of it. If not I'll make another ticket for this fix

The `Alert` will stay as is.

PR #627, opened 2022-09-23
fix: Display short address on smaller screens, add sticky table cells

### LINKED ISSUE #614: Actions within tables not visible on smaller screens
## Bug description

Although tables are scrollable/swipeable on smaller screens, important actions within the tables are not immediately visible to the user.

## Environment
 - Browser: any
 - Wallet: any
 - Chain: any

## Steps to reproduce
 1. Open the Safe on a smaller screen
 2. Go to Settings / Address Book / Assets
 3. Observe actions not being visible in the table

## Expected result

- Actions should be visible at all times

Possible solutions: https://www.notion.so/gnosis-safe/Responsive-Tables-5dba28401551448f889c9881e80d38a1

## Obtained result

- Actions are only visible after horizontally scrolling the table

## Screenshots

Screenshot 2022-09-22 at 17 44 21
Screenshot 2022-09-22 at 17 44 42

usame-algan:
## What it solves

Resolves #614 

## How this PR fixes it

- Table "Actions" are always visible on smaller screens
- Addresses are short on smaller screens

## Other changes

- Adjusted table button sizes to be small according to the Designs
- Change the Spending limit table variant to be consistent with other tables

## How to test it

1. Open the Safe on a smaller screen (< 600px)
2. Navigate to Assets
3. Observe "Send" Buttons in the Table being visible
4. Navigate to Address Book
5. Observe Buttons within the Table being visible
6. Navigate to Settings/Setup
7. Observe Buttons within the Table being visible
8. Navigate to Settings/Spending limits
9. Observe Buttons within the Table being visible

## Screenshots
Screenshot 2022-09-23 at 15 54 29
Screenshot 2022-09-23 at 15 54 49


iamacook (commented):
Looks great!

We could reduce some of the `TableCell` size to gain event more space:

image
image

src/components/common/EthHashInfo/index.tsx
```diff
@@ -74,6 +74,10 @@ const EthHashInfo = ({
             {showPrefix && shouldPrefix && prefix && <b>{prefix}:</b>}
             {shortAddress ? shortenAddress(address) : address}
           </Typography>
+          <Typography variant="body2" fontWeight="inherit" component="div" className={css.mobileAddress}>
+            {showPrefix && shouldPrefix && prefix && <b>{prefix}:</b>}
+            {shortenAddress(address)}
```
katspaugh, line 79: I would add only this part instead of the whole Typography. It could be wrapped in a span.

usame-algan: Makes sense, thanks!

iamacook (approved):

francovenica:
> Looks great!
> 
> We could reduce some of the `TableCell` size to gain event more space:

I assume that's a fixed width, since there are table that has 3 items. If the cell can change the width depending on how many itmes it has then yeah, it would be great for it to adjust, but I see it as nish fix.

PR #632, opened 2022-09-26
test: add tests to `<SingleTx />`

DiogoSoaress:
## What it solves
<SingleTx /> tests (part of https://github.com/safe-global/web-core/pull/609)

Test cases:
1. Safe contains transaction (happy path)
2. Url is wrong
3. txId is wrong
4. Safe does not contain the transaction


usame-algan (approved):
👍

PR #641, opened 2022-09-26
fix: Parse transaction values to string in safe apps modal

### LINKED ISSUE #631: Safes as owners cannot sign transactions
## Bug description
Using a safe as the owner of another safe, when trying to create/sign/execute a tx it always will show an error in the review step and won't allow the user to sign

## Environment
 - Browser: Chrome
 - Wallet: WC (Connecting with the WC app from another safe)
 - Chain: rinkeby

## Steps to reproduce
 1. Have 2 safes and their owners, can be any policy
 2. Safe1 must have Safe2 as owner
 3. Open each safe in a different tab
 4. Connect in Safe1 and pick the "wallet connect" option and copy the QR code
 5. Open the Safe2 apps, go to wallet connect and paste the QR code
 6. Also in Safe2 connect with any of its owners
 7. In Safe1 create any tx (add owner for example)
 8. In Safe2 the prompt to sign the tx pops up

## Expected result
The tx can be signed normally

## Obtained result
An error 1337: a valid integer is required pops up. This is for any type of signature (on-chain or off-chain)

## Screenshots
image

usame-algan:
## What it solves

Resolves #631 

## How this PR fixes it

Transaction values in the communicator are parsed with BigNumber to make sure that they are strings and not hex to avoid incompatibility issues down the line like `safeTxGas` estimation. This is in line with how it is handled in [safe-react](https://github.com/safe-global/safe-react/blob/5625253248eec5bcb08b25242eebca110dd4afdd/src/routes/safe/components/Apps/components/ConfirmTxModal/ReviewConfirm.tsx#L58).

## How to test it

Follow steps from the issue: #631 

Other cases to test:

1. Open any safe app
2. Create a transaction
3. Observe no errors in the modal

### DIFF
--- a/src/components/safe-apps/AppFrame/useAppCommunicator.ts
+++ b/src/components/safe-apps/AppFrame/useAppCommunicator.ts
@@ -20,6 +20,7 @@ import useSafeInfo from '@/hooks/useSafeInfo'
 import useIsGranted from '@/hooks/useIsGranted'
 import { useCurrentChain } from '@/hooks/useChains'
 import { createSafeAppsWeb3Provider } from '@/hooks/wallets/web3'
+import { BigNumber } from '@ethersproject/bignumber'
 
 export enum CommunicatorMessages {
   REJECT_TRANSACTION_MESSAGE = 'Transaction was rejected',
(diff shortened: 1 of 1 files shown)


iamacook (approved):

yagopv (approved):

francovenica:
I think the e2e test is failing because of what it was reported in this ticket https://github.com/safe-global/web-core/issues/629

usame-algan:
> I think the e2e test is failing because of what it was reported in this ticket https://github.com/safe-global/web-core/issues/629

Should work again now

PR #650, opened 2022-09-26
fix: jumpy review tx modal

DiogoSoaress:
## What it solves
- do not display nonce `0` while waiting for the estimations endpoint
- renders a Skeleton while waiting for the Simulation parameters

## How to test it
1. Go to a Safe and create a transaction
2. On the review modal, while the nonce is being retrieved, a Skeleton is displayed
3. In the same modal, a Skeleton component displays before the Simulation's accordion

## Screenshots
Screenshot 2022-09-26 at 15 33 17

### DIFF
--- a/src/components/tx/GasParams/index.tsx
+++ b/src/components/tx/GasParams/index.tsx
@@ -69,7 +69,7 @@ const GasParams = ({ params, isExecution, onEdit }: GasParamsProps): ReactElemen
         ) : (
           <Typography>
             Signing the transaction with nonce&nbsp;
-            {nonce !== 0 || safeNonce === 0 ? (
+            {nonce !== undefined ? (
               nonce
             ) : (
               <Skeleton variant="text" sx={{ display: 'inline-block', minWidth: '2em' }} />
--- a/src/components/tx/GasParams/index.tsx
+++ b/src/components/tx/GasParams/index.tsx
@@ -4,7 +4,6 @@ import { useCurrentChain } from '@/hooks/useChains'
 import { formatVisualAmount } from '@/utils/formatters'
 import { type AdvancedParameters } from '../AdvancedParams/types'
 import { trackEvent, MODALS_EVENTS } from '@/services/analytics'
-import useSafeInfo from '@/hooks/useSafeInfo'
 
 const GasDetail = ({ name, value, isLoading }: { name: string; value: string; isLoading: boolean }): ReactElement => {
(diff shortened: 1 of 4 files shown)


src/components/tx/GasParams/index.tsx
```diff
@@ -66,7 +69,7 @@ const GasParams = ({ params, isExecution, onEdit }: GasParamsProps): ReactElemen
         ) : (
           <Typography>
             Signing the transaction with nonce&nbsp;
-            {nonce !== undefined ? (
+            {nonce !== 0 || safeNonce === 0 ? (
```
katspaugh, line 72: 👀

src/components/tx/TxSimulation/index.tsx
```diff
@@ -86,5 +86,15 @@ export const TxSimulation = (props: TxSimulationProps): ReactElement | null => {
     return null
   }
 
+  if (!props.transactions) {
+    return (
+      <Box sx={{ mt: 2 }}>
```
katspaugh, line 91: First of all, why do you need a box if you can just assign the margin to the skeletonWrapper? Secondly, for future: `<Box mt={2}`.

src/components/tx/GasParams/index.tsx
katspaugh: I don't get this condition.

DiogoSoaress: I should have left a comment.... We were displaying 0 while waiting for the estimation. It only make sense to display the zero while waiting for the estimation if the `safeNonce` is indeed 0

src/components/tx/TxSimulation/index.tsx
DiogoSoaress: I added the wrapper after the `<Box />`. thank you for the catch

src/components/tx/GasParams/index.tsx
DiogoSoaress: on it ser! I was just testing

katspaugh (approved):
👍

francovenica:
Is still jumpy, since there are many other fields besides the one for the estimation, there is the simulation and the tx data (this for safes 1/x) . Should we make a skeleton of each different block?
jumpy

katspaugh:
The problem is we don't know ahead of time if the tx is executable (nonce can be not the earliest etc). So even if we show skeletons, they might disappear. So it will just jump in more cases than now.

francovenica:
Fair enough.

LGTM. Can be merged when possible

PR #651, opened 2022-09-26
fix: show total of queued transactions by nonce

### LINKED ISSUE #624: Dashboard Transaction queue counter
## Bug description
The counter is also counting the tx in parallel. The tx counter should only count the amount of different nonces

## Environment
 - Browser: Chrome
 - Wallet: Any
 - Chain: Any

## Steps to reproduce
 1. Go to a safe
 2. Queue any tx
 3. Queue a rejection for that transaction
 4. Go to the dashboard and check the counter in the Transaction Queue widget

## Expected result
It just counts the amount of different nonces present in queue (in this example only 1)

## Obtained result
The counter counts also the Rejection (because is in parallel) so it shows a "2" in this example

## Screenshots
image

image

iamacook:
## What it solves

Resolves #624

## How this PR fixes it

The queue count is calculated based on differing nonces.

## How to test it

Queue a transaction and create a rejection for it. Observe that the dashboard does not count two, but one.


katspaugh (approved):
👍

PR #660, opened 2022-09-26
fix: add nonce of transaction to execute to label

### LINKED ISSUE #655: [Queued list] the text above queued about tx in the next section is missing
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby

## Steps to reproduce
 Current result:
image

## Expected result
QUEUE - TRANSACTION WITH NONCE 303 NEEDS TO BE EXECUTED FIRST

iamacook:
## What it solves

Resolves #655 

## How this PR fixes it

" - transaction with nonce {{nonce}} needs to be executed first" has been added to "Queued" labels.

## How to test it

Queue two transactions. Observe the extra text on the second `LABEL`.

## Screenshots

image

### DIFF
--- a/src/components/transactions/GroupLabel/index.tsx
+++ b/src/components/transactions/GroupLabel/index.tsx
@@ -5,35 +5,41 @@ import useTxQueue from '@/hooks/useTxQueue'
 import { isLabelListItem, isMultisigExecutionInfo, isTransactionListItem } from '@/utils/transaction-guards'
 import useSafeInfo from '@/hooks/useSafeInfo'
 
-const GroupLabel = ({ item }: { item: Label }): ReactElement => {
-  let label: string | LabelValue = item.label
-
+export const useGroupLabel = (item: Label): string => {
   const { page } = useTxQueue()
   const { safe } = useSafeInfo()
 
-  /**
-   * 'Next' label is returned when the first transaction has a correct nonce
-   * 'Queued' label is returned after 'Next' OR when the first transaction has an out of order nonce
-   *
-   * We want to append to the 'Queue' label if:
-   * - There is a 'Next' label on the first page
-   * - The first transaction has an out of order nonce
-   */
+  const { label } = item
 
-  if (label === LabelValue.Queued && page) {
-    const hasNext = page.results.some((tx) => isLabelListItem(tx) && tx !== item)
-    const firstTx = page.results.find(isTransactionListItem)
-
-    if (
-      firstTx &&
-      isMultisigExecutionInfo(firstTx.transaction.executionInfo) &&
-      (hasNext || firstTx.transaction.executionInfo.nonce !== safe.nonce)
-    ) {
-      const nextNonce = hasNext ? firstTx.transaction.executionInfo.nonce : safe.nonce
-
-      label = `${label} - transaction with nonce ${nextNonce} needs to be executed first`
-    }
+  if (label !== LabelValue.Queued || !page) {
+    return label
   }
+
+  const firstTx = page.results.find(isTransactionListItem)
+
+  if (!firstTx || !isMultisigExecutionInfo(firstTx.transaction.executionInfo)) {
+    return label
+  }
+
+  const getQueueLabel = (nonce: number) => {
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
👍

src/components/transactions/GroupLabel/index.tsx
```diff
@@ -1,9 +1,23 @@
...
+import { isMultisigExecutionInfo, isTransactionListItem } from '@/utils/transaction-guards'
 
 const GroupLabel = ({ item }: { item: Label }): ReactElement => {
-  return <div className={css.container}>{item.label}</div>
+  let label: string | LabelValue = item.label
+
+  const { page } = useTxQueue()
+
+  if (label === LabelValue.Queued && !page?.next) {
+    const firstTx = page?.results.find(isTransactionListItem)
```
katspaugh, line 13: Maybe we should display the `safeNonce` instead?

katspaugh: Also how is `!page?.next` related to this label?

iamacook: It's possible to set a transaction nonce higher than that of the Safe nonce though. This mirrors that of `safe-react` as well.

iamacook: If there is `!page?.next`, there will be no 'Next' transaction label to append the nonce to.

iamacook: In hindsight, that wouldn't be valid for long queues. I'm looking into this further.

src/components/transactions/GroupLabel/index.tsx
```diff
@@ -1,9 +1,28 @@
...
 
 const GroupLabel = ({ item }: { item: Label }): ReactElement => {
-  return <div className={css.container}>{item.label}</div>
+  let label: string | LabelValue = item.label
+
+  const { page } = useTxQueue()
+
+  const labels = page?.results.filter(isLabelListItem)
+
+  // Current page has both 'Queued' and 'Next' labels
```
katspaugh, line 14: `First page` would be more correct.

src/components/transactions/GroupLabel/index.tsx
```diff
@@ -1,9 +1,28 @@
...
+  const hasBothLabels = labels && labels.length > 1
+
+  if (hasBothLabels && label === LabelValue.Queued) {
```
katspaugh, line 17: Code style: I would move the bothLabels thing into the if like so:

```
// We're dealing with the "Queued" transactions that can be executed only after the "Next" transactions
if (label === LabelValue.Queued) {
  const hasNext = page?.results.some((tx) => isLabelListItem(tx) && tx !== item)
  const firtstTx...
}
```

katspaugh (commented):
Worth extracting into a hook and adding tests?

src/components/transactions/GroupLabel/index.tsx
```diff
@@ -1,9 +1,47 @@
...
+
+  // There is also a 'Next' label on the page of the queue
+  if (page.results.some((tx) => isLabelListItem(tx) && tx !== item)) {
+    return getQueueLabel(firstTx.transaction.executionInfo.nonce)
+  }
+
+  // First transaction has an out of order nonce
+  if (firstTx.transaction.executionInfo.nonce !== safe.nonce) {
+    return getQueueLabel(safe.nonce)
+  }
```
katspaugh, line 36: That's a bit hard to read.
Suggestion:
```
const hasNextLabel = page.results.some((tx) => isLabelListItem(tx) && tx !== item)
const nonceInFuture = firstTx.transaction.executionInfo.nonce !== safe.nonce

if (!hasNextLabel && !nonceInFuture) return label

const nextNonce = nonceInFuture ? safe.nonce : firstTx.transaction.executionInfo.nonce

return `${label} - transaction with nonce ${nextNonce} needs to be executed first`
```

katspaugh (approved):
✅

PR #664, opened 2022-09-26
fix: Redirect to single tx view when rejecting a tx

### LINKED ISSUE #656: [tx creation] Tx details is not opened for the rejection tx
## Bug description
[tx creation] Tx details is not opened for the rejection tx
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby

## Steps to reproduce
 2. try to reject any tx in the queued list
 Current result: rejection tx is created but tx details for this tx is not opened

## Expected result
Open tx details for the rejection tx too

usame-algan:
## What it solves

Resolves #656 

## How this PR fixes it

Redirects to the single tx view when rejecting an existing transaction.

## How to test it

1. Open the Safe
2. Create a reject tx
3. Observe being navigated to the single tx view

### DIFF
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -149,7 +149,7 @@ const SignOrExecuteForm = ({
 
     // If txId isn't passed in props, it's a newly created tx
     // Redirect to the single tx view
-    if (redirectToTx && !txId) {
+    if ((redirectToTx && !txId) || isRejection) {
       router.push({
         pathname: AppRoutes.transactions.tx,
         query: { safe: router.query.safe, id },


src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -149,7 +149,7 @@ const SignOrExecuteForm = ({
 
     // If txId isn't passed in props, it's a newly created tx
     // Redirect to the single tx view
-    if (redirectToTx && !txId) {
+    if ((redirectToTx && !txId) || isRejection) {
```
katspaugh, line 152: Should be
`if (redirectToTx && (!txId || isRejection)) {

usame-algan: We could possibly simplify the condition to `if (redirectToTx) ...` and just set the prop explicitly in those cases e.g. when using `SignOrExecuteForm` in `ConfirmProposedTx` we could pass `redirectToTx={false}` and could ignore the `txId` prop for the redirection.

katspaugh: Good idea.

katspaugh (approved):

PR #666, opened 2022-09-26
fix: don't adjust timestamps + show exact time on history

### LINKED ISSUE #657: [txs list] Wrong calculation for the N days ago
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
image
2. the actual date and time - 
image

Current result: 6 days ago is displayed for the 5 days ago action

## Expected result
Make sure that the valid information is displayed about N days ago in the txs list

iamacook:
## What it solves

Resolves #657 & #673

## How this PR fixes it

- The `timestamp` of a transaction is formatted as is.
- The exact time is shown on the history list.

## How to test it

Open `rin:0xf2565317F3Ae8Ae9EA98E9Fe1e7FADC77F823cbD/transactions/queue` and observe that the queued transaction have valid "n days ago" labels. The exact time will, however, been shown on the history list.

### DIFF
--- a/src/components/common/DateTime/index.tsx
+++ b/src/components/common/DateTime/index.tsx
@@ -1,11 +1,16 @@
 import { ReactElement } from 'react'
 import { Tooltip } from '@mui/material'
 import { formatDateTime, formatTimeInWords } from '@/utils/date'
+import { useRouter } from 'next/router'
+import { AppRoutes } from '@/config/routes'
 
 const DAYS_THRESHOLD = 60
 
 const DateTime = ({ value }: { value: number }): ReactElement => {
-  const displayExactDate = Math.floor((Date.now() - value) / 1000 / 60 / 60 / 24) > DAYS_THRESHOLD
+  const router = useRouter()
+  const isHistory = router.pathname.includes(AppRoutes.transactions.history)
+
+  const displayExactDate = Math.floor((Date.now() - value) / 1000 / 60 / 60 / 24) > DAYS_THRESHOLD || isHistory
 
   return (
     <Tooltip title={displayExactDate ? '' : formatDateTime(value)} placement="top">
--- a/src/components/common/DateTime/index.tsx
+++ b/src/components/common/DateTime/index.tsx
@@ -7,19 +7,25 @@ import { useTxFilter } from '@/utils/tx-history-filter'
 
 const DAYS_THRESHOLD = 60
 
+/**
+ * If queue, show relative time until threshold then show full date and time
+ * If history, show time (as date labels are present)
+ * If filter, show full date and time
+ */
+
 const DateTime = ({ value }: { value: number }): ReactElement => {
   const [filter] = useTxFilter()
-
   const router = useRouter()
-  const isHistory = router.pathname === AppRoutes.transactions.history
+
+  // (non-filtered) history is the endpoint that returns date labels
+  const showTime = router.pathname === AppRoutes.transactions.history && !filter
 
   const isOld = Math.floor((Date.now() - value) / 1000 / 60 / 60 / 24) > DAYS_THRESHOLD
-
-  const displayExactDate = isOld || isHistory
+  const showDateTime = isOld || filter
 
   return (
-    <Tooltip title={displayExactDate ? '' : formatDateTime(value)} placement="top">
-      <span>{filter ? formatDateTime(value) : displayExactDate ? formatTime(value) : formatTimeInWords(value)}</span>
+    <Tooltip title={showDateTime ? '' : formatDateTime(value)} placement="top">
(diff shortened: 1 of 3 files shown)


usame-algan (approved):

src/components/common/DateTime/index.tsx
```diff
@@ -1,11 +1,16 @@
...
 import { formatDateTime, formatTimeInWords } from '@/utils/date'
+import { useRouter } from 'next/router'
+import { AppRoutes } from '@/config/routes'
 
 const DAYS_THRESHOLD = 60
 
 const DateTime = ({ value }: { value: number }): ReactElement => {
-  const displayExactDate = Math.floor((Date.now() - value) / 1000 / 60 / 60 / 24) > DAYS_THRESHOLD
+  const router = useRouter()
+  const isHistory = router.pathname.includes(AppRoutes.transactions.history)
```
katspaugh, line 11: This will match filters as well.

src/components/common/DateTime/index.tsx
```diff
@@ -1,11 +1,16 @@
...
+
+  const displayExactDate = Math.floor((Date.now() - value) / 1000 / 60 / 60 / 24) > DAYS_THRESHOLD || isHistory
```
katspaugh, line 13:
```
const isOld = Math.floor((Date.now() - value) / 1000 / 60 / 60 / 24) > DAYS_THRESHOLD
const displayExactDate = isOld || isHistory
```

src/components/common/DateTime/index.tsx
iamacook: Yes, filtered results are also historical. This matches `safe-react`.

katspaugh: I guess what I meant is unrelated to this line. I meant that maybe we don't need to show the full date when it's in the date label.

katspaugh: Also why is it `includes` and not just a strict comparison?

iamacook: > I meant that maybe we don't need to show the full date when it's in the date label.

Do you mean something like this?

```tsx
<span>{filter ? formatDateTime(value) : displayExactDate ? formatTime(value) : formatTimeInWords(value)}</span>
```

>Also why is it `includes` and not just a strict comparison?

I'll adjust it.

katspaugh: Filter and queue, yes. They both don't have date labels.

src/components/common/DateTime/index.tsx
```diff
@@ -1,20 +1,25 @@
...
+  const isHistory = router.pathname === AppRoutes.transactions.history
+
+  const isOld = Math.floor((Date.now() - value) / 1000 / 60 / 60 / 24) > DAYS_THRESHOLD
+
+  const displayExactDate = isOld || isHistory
 
   return (
     <Tooltip title={displayExactDate ? '' : formatDateTime(value)} placement="top">
-      <span>{displayExactDate ? formatDateTime(value) : formatTimeInWords(getRelevantTimestamp(value))}</span>
+      <span>{filter ? formatDateTime(value) : displayExactDate ? formatTime(value) : formatTimeInWords(value)}</span>
```
katspaugh, line 22: What about the queue?

iamacook: As per https://github.com/safe-global/web-core/issues/673, the queue will use "textual" time up until the threshold.

iamacook: The queue and filter have no date labels.

- As per #673, the queue should show "n days ago" until the `DAYS_THRESHOLD` is met and the history show "Jan 1, 1970 - 00:00:00 AM".
- However, as per your suggestion, the history should use "00:00:00 AM" as it has date labels and the filter use "Jan 1, 1970 - 00:00:00 AM" as it doesn't.

katspaugh: Yes. From your code however, I see the following logic:
* If filter, show full date-time
* If history or over 60 days, show time when over 60 days
* If queue, it will show **only time** when over 60 days, otherwise "N days ago"

katspaugh (approved):

francovenica:
Question:
The date in the tooltip and the lable that group the tx in a specific day are off by 1 day, but that should be because of my timezone, but we discussed this was mostly a backend issue. Is it the same for you in your timezone?

https://tx-dates.web-core.pages.dev/rin:0x9913B9180C20C6b0F21B6480c84422F6ebc4B808/transactions/history

dates

iamacook:
> Question: The date in the tooltip and the lable that group the tx in a specific day are off by 1 day, but that should be because of my timezone, but we discussed this was mostly a backend issue. Is it the same for you in your timezone?

No, it is a timezone issue.

PR #670, opened 2022-09-26
test: prefer Goerli for E2E

DiogoSoaress:
## What it solves
- Ports the E2E test suite from Rinkeby to the Goerli testnet

_Note:_ I have also update one of the non-smoking tests (`add_owner`) but IMO it is not worth the effort to update all tests if they are not being triggered in the CI. Main reason is: the tests will simply get outdated if not run regularly.

Migrated tests (smoking):
- [x] address book
- [x] balances
	- [ ] transfer more Dai and USDC to test the M notation
	- [ ] missing a Goerli Safe to test pagination
- [x] dashboard
- [x] landing (nothing to port)
- [x] load safe
    - [ ] create a goerli ENS name for the Safe
    - [ ] create a goerli ENS name for the Safe owner
- [x] tx history	(_ps: the previous tests logic was never run_)


katspaugh (approved):
Nice work! 🙏

package.json
```diff
@@ -24,7 +24,7 @@
     "generate-types": "yarn generate-types:safeDeployments && yarn generate-types:spendingLimit",
     "postinstall": "yarn generate-types && yarn css-vars",
     "analyze": "cross-env ANALYZE=true yarn build",
-    "cypress:open": "cypress open",
+    "cypress:open": "cross-env TZ=UTC cypress open",
```
iamacook, line 27: The CI won't be set to `UTC` unless you also prepend the `cypress:run` command.

Issue #682, opened 2022-09-27
Transaction data overlapping on narrow resolutions


JagoFigueroa:
## Bug description
Transaction information overlaps on smaller window resolutions causing visual issues on UI

## Environment
 - Browser: Any
 - Wallet: Any
 - Chain: Any

## Steps to reproduce
Go to either Transaction queue or [Transaction history](https://web-core.pages.dev/rin:0x91dB860c37E83dab0A4A005E209577E64c61EEfA/transactions/history)

in my 2k monitor, presentation issues start at around 1600px X xxxxpx resolution

## Expected result
No visual issues (at least until a reasonable resolution, like on og web)

## Obtained result
A couple examples: 
![Screenshot 2022-09-27 at 10 59 54](https://user-images.githubusercontent.com/92332710/192487217-a0ff59bc-967a-42a4-a8f3-764b1007425e.png)
Revamp (left) vs legacy (right)
![Screenshot 2022-09-27 at 11 14 15](https://user-images.githubusercontent.com/92332710/192487345-6f77f90e-4470-4de6-a944-941a14f6b436.png)

francovenica:
This ticket is the same as #675

liliya-soroka:
Verified

PR #684, opened 2022-09-27
style: Adjust tx summary columns to have fixed width

### LINKED ISSUE #675: [History] The list is "dancing"
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to https://web-core.pages.dev/rin:0xf2565317F3Ae8Ae9EA98E9Fe1e7FADC77F823cbD/transactions/history
 2. scroll down to see the txs with long descriptions and short at the same time

Current result:
It's hard to read and looks messy
image

## Expected result
Make sure that the columns are seen quite well, so it's easy to read the information 
image

usame-algan:
## What it solves

Resolves #675 #682 

## How this PR fixes it

Adds a min-width to the dynamic columns and breaks the content into multiple lines if there is not enough space. This change comes with the trade-off that there will be more white space if a column has very little content e.g. a short method name.

## How to test it

1. Open the Safe
2. Navigate to the transaction history
3. Adjust the viewport width
4. Observe each transaction column to be on the same line
5. Observe columns with a lot of content breaking into multiple lines if necessary

## Screenshots
Screenshot 2022-09-27 at 11 52 52


DiogoSoaress (commented):
Looks so much more organized! Good job here.
Is there a way to not break the token symbol?
Screenshot 2022-09-27 at 17 08 11

iamacook (approved):

usame-algan:
> Is there a way to not break the token symbol?

Noticed that too but I think the value can also get quite large depending on the amount that is transfered so we have to break it like we do with the other fields.

francovenica:
The alignment looks good, but we have so much empty space between the date and the status. I know that that space is there cuz is where the accept/reject would be for tx in queue, but can't we do an exception in the history tab?
image

usame-algan:
> The alignment looks good, but we have so much empty space between the date and the status. I know that that space is there cuz is where the accept/reject would be for tx in queue, but can't we do an exception in the history tab?

Good idea! I adjusted the layout for the tx history to use more of the available space

Screenshot 2022-09-29 at 17 51 45

liliya-soroka:
Verified, can be merged 👍

PR #685, opened 2022-09-27
Fix: reconnect wallet only if unlocked

### LINKED ISSUE #615: Modal to connect to the app via physical device keeps popping up
## Bug description
When connected with a trezor or ledger, if you refresh the page the modal pops up so you can connect again. The issue is that the only way to make it not pop up again is by connecting with a wallet and disconnecting manually
In safe react, the behavior was that you would just disconnect if the page is refreshed and no popUp would show up

## Environment
 - Browser: Chrome
 - Wallet: Trezor/ledger
 - Chain: Any

## Steps to reproduce
 1. Go to the safe app
 2. Connect with trezor or ledger
 3. Refresh the page

## Expected result
The connection to the wallet is lost and nothing more

## Obtained result
The connection is lost and the popup keeps showing everytime you refresh the page, even if you are not connected

## Screenshots
no screenshots

katspaugh:
## What it solves

Resolves #615

## How this PR fixes it
Avoid attempting to reconnect if the wallet is not verifiably unlocked.

## How to test it
MetaMask:
* Connect to MM
* Refresh the page – it should auto-connect again
* Lock MM
Screenshot 2022-09-27 at 12 03 31

* Refresh the page
* It should _not_ reconnect

WalletConnect/Safe mobile pairing:
* Connect
* Refresh the page – it should auto-connect again
* Initiate a disconnect from the WC side (not via our app!)
* Refresh the page
* It should _not_ reconnect

## Analytics changes
None.

### DIFF
--- a/src/components/common/SafeTokenWidget/index.tsx
+++ b/src/components/common/SafeTokenWidget/index.tsx
@@ -53,7 +53,7 @@ const SafeTokenWidget = () => {
       <Tooltip title={url ? `Open ${claimingApp?.name}` : ''}>
         <span>
           <Track {...OVERVIEW_EVENTS.SAFE_TOKEN_WIDGET}>
-            <Link href={url || '#'} passHref>
+            <Link href={url || ''} passHref>
               <ButtonBase
                 aria-describedby={'safe-token-widget'}
                 sx={{ alignSelf: 'stretch' }}
--- a/src/utils/wallets.ts
+++ b/src/utils/wallets.ts
@@ -1,3 +1,4 @@
+import { ProviderLabel } from '@web3-onboard/injected-wallets'
 import { hasStoredPairingSession } from '@/services/pairing/connector'
 import { PAIRING_MODULE_LABEL } from '@/services/pairing/module'
 
@@ -23,7 +24,7 @@ export const isWalletRejection = (err: Error & { code?: number }): boolean => {
 }
 
 const WalletNames = {
-  METAMASK: 'MetaMask',
+  METAMASK: ProviderLabel.MetaMask,
   WALLET_CONNECT: 'WalletConnect',
   SAFE_MOBILE_PAIRING: PAIRING_MODULE_LABEL,
 }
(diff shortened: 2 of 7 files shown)


src/components/common/SafeTokenWidget/index.tsx
```diff
@@ -53,7 +53,7 @@ const SafeTokenWidget = () => {
       <Tooltip title={url ? `Open ${claimingApp?.name}` : ''}>
         <span>
           <Track {...OVERVIEW_EVENTS.SAFE_TOKEN_WIDGET}>
-            <Link href={url || '#'} passHref>
```
katspaugh, line 56: This fixes a hydration warning.

src/utils/wallets.ts
```diff
@@ -18,3 +21,31 @@ const isMMRejection = (err: Error & { code?: number }): boolean => {
 export const isWalletRejection = (err: Error & { code?: number }): boolean => {
   return isMMRejection(err) || isWCRejection(err) || isKeystoneError(err)
 }
+
+const WalletNames = {
+  METAMASK: 'MetaMask',
```
iamacook, line 26: You can import `ProviderLabel.MetaMask` from `'@web3-onboard/injected-wallets'` to get this.

iamacook (approved):
Works well! We should check other wallet modules/injected wallets at some point as well.

francovenica:
This works fine
With MM extension: it will connect when refreshing as long the extension is not locked 
With physical devices: It just disconnects on refresh and you have to manually open the modal again to connect
With paring: It reconnects on refresh only if you are still connected in the mobile app 

For WC using the MM phone app:
There is no way to "disconnect" from the MM mobile app, more than locking the wallet the same way that is done in the MM extension on web browser, but even locking it like that won't disconnect it from the web app. Unlocking the MM phone app and trying to execute a tx works just fine. I also tried to connect the MM phone app to another page at the same time and again, you can still sign tx in the gnosis safe app.

Long story short: it seems that the WC connection is never lost unless you manually disconnect from the safe app, which for me is fine.

francovenica:
I'll pass the ticket since the issue of the modal constantly showing up was solved
Also connection to wallets have not been affected (nothing broke)

PR #687, opened 2022-09-27
fix: replacement tx error

### LINKED ISSUE #620: Speeding up tx causes a duplicate
## Bug description
After speeding up a tx that was taking too long ~showed that the tx end up being replaced, but now there is the same tx in history and queue at the same time~ got an error saying that the tx was unsucessful and was also replaced.
The safe: https://web-core.pages.dev/rin:0x9913B9180C20C6b0F21B6480c84422F6ebc4B808/transactions/history

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Rinkeby

## Steps to reproduce
 1. Go to a Safe 2/x
 2. Create a tx to queue it
 3. Sign it with the 2nd owner but in adv options set a low gas price and fee (0.0000001)
 4. Wait for a bit, the tx will get "stuck" in processing
 5. After a few mins open your MM extension and Speed up the tx with a high gas fee value

## Expected result
The original tx gets executed and no error is shown.

## Obtained result
* An error saying that the tx was unsucessful and was also replaced.
* The tx is now both in the history and queue tabs (same nonce)

## Screenshots
tx queue

DiogoSoaress:
## What it solves
Do not dispatch a replacement transaction as an error

Partially resolves #620 

## How this PR fixes it
`ethers` throws an error upon speeding up a transaction. The error is handled now and does not dispatch a FAILED event.

### DIFF
--- a/src/services/tx/txSender.ts
+++ b/src/services/tx/txSender.ts
@@ -17,7 +17,7 @@ import extractTxInfo from '@/services/tx/extractTxInfo'
 import proposeTx from './proposeTransaction'
 import { txDispatch, TxEvent } from './txEvents'
 import { getSafeSDK } from '@/hooks/coreSDK/safeCoreSDK'
-import { didRevert, EthersError } from '@/utils/ethers-utils'
+import { didReprice, didRevert, EthersError } from '@/utils/ethers-utils'
 import Safe, { RemoveOwnerTxParams } from '@gnosis.pm/safe-core-sdk'
 import { AddOwnerTxParams, SwapOwnerTxParams } from '@gnosis.pm/safe-core-sdk/dist/src/Safe'
 import MultiSendCallOnlyEthersContract from '@gnosis.pm/safe-ethers-lib/dist/src/contracts/MultiSendCallOnly/MultiSendCallOnlyEthersContract'
@@ -26,7 +26,6 @@ import { ContractTransaction, ethers } from 'ethers'
 import { SpendingLimitTxParams } from '@/components/tx/modals/TokenTransferModal/ReviewSpendingLimitTx'
 import { getSpendingLimitContract } from '@/services/contracts/spendingLimitContracts'
 import EthersAdapter from '@gnosis.pm/safe-ethers-lib'
-import { ErrorCode } from '@ethersproject/logger'
 
 const getAndValidateSafeSDK = (): Safe => {
   const safeSDK = getSafeSDK()
@@ -266,12 +265,11 @@ export const dispatchTxExecution = async (
     .catch((err) => {
       const error = err as EthersError
 
-      // Meaning the transaction was "repriced" or "replaced"
-      if (error.code === ErrorCode.TRANSACTION_REPLACED && error.reason !== 'cancelled') {
-        // TxEvent.SUCCESS is handled by txHistoryMiddleware
-        return
+      if (didReprice(error)) {
+        txDispatch(TxEvent.PROCESSED, { txId, receipt: error.receipt })
+      } else {
+        txDispatch(TxEvent.FAILED, { txId, error: error as Error })
       }
-      txDispatch(TxEvent.FAILED, { txId, error: error as Error })
     })
 
   return result.hash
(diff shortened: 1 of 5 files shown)


src/services/tx/txSender.ts
```diff
@@ -262,7 +263,14 @@ export const dispatchTxExecution = async (
...
       }
     })
-    .catch((error) => {
+    .catch((err) => {
+      const error = err as EthersError
+
+      // Meaning the transaction was "repriced" or "replaced"
+      if (error.code === ErrorCode.TRANSACTION_REPLACED && error.reason !== 'cancelled') {
+        // TxEvent.SUCCESS is handled by txHistoryMiddleware
+        return
```
usame-algan, line 272: Should we also dispatch a `PROCESSED` event here?

src/services/tx/txSender.ts
```diff
@@ -262,7 +263,14 @@ export const dispatchTxExecution = async (
...
+      }
       txDispatch(TxEvent.FAILED, { txId, error: error as Error })
```
iamacook, line 274: We should abstract this and use it for the multisend dispatcher as well.

src/utils/ethers-utils.ts
```diff
@@ -1,4 +1,7 @@
 import type { TransactionReceipt } from '@ethersproject/abstract-provider/lib'
+import { ErrorCode } from '@ethersproject/logger'
+
+export type EthersError = Error & { code: ErrorCode; reason: string }
```
iamacook, line 4: Let's create an `enum` for the `reason`s here.

src/services/tx/txSender.ts
```diff
@@ -262,7 +263,14 @@ export const dispatchTxExecution = async (
         txDispatch(TxEvent.PROCESSED, { txId, receipt })
       }
     })
-    .catch((error) => {
+    .catch((err) => {
+      const error = err as EthersError
+
+      // Meaning the transaction was "repriced" or "replaced"
+      if (error.code === ErrorCode.TRANSACTION_REPLACED && error.reason !== 'cancelled') {
```
katspaugh, line 270: Would this work?
```
if (error.reason === 'repriced') {
```

katspaugh (commented):
What I don't get: how does fixing the error dispatching help with duplicate txns?

src/services/tx/txSender.ts
DiogoSoaress: added

src/services/tx/txSender.ts
DiogoSoaress: I've created a `isReprice` util

DiogoSoaress:
> What I don't get: how does fixing the error dispatching help with duplicate txns?

This PR handles the Notification error from the bug [description](https://github.com/safe-global/web-core/issues/620#issue-1383492352) 
Screenshot 2022-09-27 at 16 56 08


I was not able to see or reproduce the duplicated transaction so I also left a comment in the original PR to Franco -> https://github.com/safe-global/web-core/issues/620#issuecomment-1259473078

src/utils/ethers-utils.ts
```diff
@@ -1,8 +1,19 @@
...
+}
+
+export type EthersError = Error & { code: ErrorCode; reason: EthersTxReplacedReason; receipt: TransactionReceipt }
 
 export const didRevert = (receipt: TransactionReceipt): boolean => {
   return receipt.status === 0
 }
+
+export const didReprice = (error: EthersError): boolean => {
+  return error.code === ErrorCode.TRANSACTION_REPLACED && error.reason === EthersTxReplacedReason.repriced
```
katspaugh, line 18: Why do you need to check both the code and the reason? Just the reason seems enough.

katspaugh:
> This PR handles the Notification error from the bug https://github.com/safe-global/web-core/issues/620#issue-1383492352

Then it should say it _resolves_ #620?

src/utils/ethers-utils.ts
```diff
@@ -1,5 +1,19 @@
 import type { TransactionReceipt } from '@ethersproject/abstract-provider/lib'
+import { ErrorCode } from '@ethersproject/logger'
+
+// https://docs.ethers.io/v5/api/providers/types/#providers-TransactionResponse
+export enum EthersTxReplacedReason {
+  repriced = 'repriced',
+  cancelled = 'cancelled',
+  replaced = 'replaced',
```
katspaugh, line 8: @iamacook do you know if the ethers package exports these?

iamacook: They're [just strings](https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/base-provider.ts#L1372-L1377) sadly.

src/utils/ethers-utils.ts
DiogoSoaress: I've updated the condition as per your comment.

PR #690, opened 2022-09-27
Fix: reset error after ENS resolution

### LINKED ISSUE #686: Recipient input - Using ENS name removes the network prefix after convertion
## Bug description
Writing and ENS name translates it to its address. The issue is that the prefix is removed after this convetion

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Rinkeby

## Steps to reproduce
 1. Go to a safe you own in Rinkeby
 2. Open the send funds modal (no matter if you don't have funds in the safe)
 3. Write "francotest.eth" and should translate to rin:0x61a0c717d18232711bC788F19C9Cd56a43cc8872

## Expected result
The address is set with the prefix

## Obtained result
The prefix is removed after a few seconds

## Screenshots
prefix

katspaugh:
## What it solves

Resolves #686

## How this PR fixes it

Some sort of bug in RHF caches the error state somewhere. I blamed our code at first, but it doesn't cache the error anywhere. It's not the debounce or anything else on our side. `trigger` seems to help.

### DIFF
--- a/src/components/common/AddressInput/index.tsx
+++ b/src/components/common/AddressInput/index.tsx
@@ -43,17 +43,17 @@ const AddressInput = ({ name, validate, required = true, ...props }: AddressInpu
   // Update the input value
   const setAddressValue = useCallback(
     (value: string) => {
-      setValue(name, value)
-      // Workaround for a bug in react-hook-form that it restores a cached error state on blur
-      trigger(name)
+      setValue(name, value, { shouldValidate: true })
     },
-    [setValue, trigger, name],
+    [setValue, name],
   )
 
   // On ENS resolution, update the input value
   useEffect(() => {
-    address && setAddressValue(address)
-  }, [address, setAddressValue])
+    if (address) {
+      setAddressValue(`${currentShortName}:${address}`)
+    }
+  }, [address, currentShortName, setAddressValue])
 
   return (
     <Grid container alignItems="center" gap={1}>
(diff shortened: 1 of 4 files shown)


src/components/common/AddressInput/index.tsx
```diff
@@ -29,10 +30,8 @@ const AddressInput = ({ name, validate, required = true, ...props }: AddressInpu
   const isDomainLookupEnabled = !!currentChain && hasFeature(currentChain, FEATURES.DOMAIN_LOOKUP)
   const { address, resolverError, resolving } = useNameResolver(isDomainLookupEnabled ? watchedValue : '')
 
-  // errors[name] doesn't work with nested field names like 'safe.address'.
-  // But getFieldState doesn't trigger a re-render, which breaks tests.
-  // So both are needed to get the error state and pass tests.
-  const fieldError = resolverError || getFieldState(name).error || errors[name]
+  // errors[name] doesn't work with nested field names like 'safe.address', need to use the lodash get
+  const fieldError = resolverError || _get(errors, name)
```
katspaugh, line 34: This is unrelated, just fixed along the way. Seems to still works with complex field names (tested with owners in Safe creation).

iamacook: RHF exports it's own version of `lodash`' `get` function.

src/components/common/AddressInput/index.tsx
```diff
@@ -45,9 +44,10 @@ const AddressInput = ({ name, validate, required = true, ...props }: AddressInpu
   // Update the input value
   const setAddressValue = useCallback(
     (value: string) => {
-      setValue(name, value, { shouldValidate: true })
+      setValue(name, value)
+      trigger(name)
```
iamacook, line 48: Can you add a comment regarding the blur bug here for reference.

iamacook (approved):

francovenica:
I still see the issue:
It works fine if after the name is translated you touch nothing, but as soon as you click anywhere else to keep going with the form the prefix is removed again

prefix

Issue #703, opened 2022-04-28
Add Tally Ho as a wallet option


johannesmoormann:
## Overview

In the product sync we have decided to add Tally Ho as a wallet option. In terms of further maintenance & support we limit our focus on the most used wallet options. 

## Requirements

- [ ] Install onboard module

francovenica:
Looks good. 
I created a wallet to test this. I was able to create a safe and add a new owner. (that's it since I had to use ETH mainnet)

PR #711, opened 2022-09-28
fix: filter injected wallets per chain

iamacook:
## What it solves

Reimplementation of #703

## How this PR fixes it

Injected wallets are now adjusted according to whether they are present in the `disabledWallets` array.

All wallets are initially loaded and then [onboard's `setWalletModules` is called](https://github.com/safe-global/web-core/blob/main/src/hooks/wallets/useOnboard.ts#L131-L132) with the [supported wallets](https://github.com/safe-global/web-core/blob/main/src/hooks/wallets/wallets.ts#L81-L88). The `ChainInfo` is now passed to wallet modules to allow for [filtering](https://docs.blocknative.com/onboard/injected-wallets#excluding-wallets).

## How to test it

Open the Safe and view the available wallets. Disable an inject wallet (tally, frame, opera, trust or metamask) via the CGW and refresh the page. Observe the disabled wallet is no longer available.


iamacook:
I'm aware of the failing tests.

katspaugh (approved):
LGTM.
FYI, I also modified the jest setup in a similar way in my PR.

katspaugh:
It looks like in the config-service admin, you need to explicitly enable each wallet. It means, we lose a lot of flexibility with the "detected wallet" module being filtered based on an allow-list. I suggest we keep it open-ended for now, and think about this if we encounter a problem with some of the injected wallets.

iamacook:
> I suggest we keep it open-ended for now, and think about this if we encounter a problem with some of the injected wallets.

I'll close this for now then. We can use it as reference when we eventually tackle this in the future.

PR #714, opened 2022-09-28
fix: Display multisend actions when executing batch

### LINKED ISSUE #662: [Batch execution] More user friendly txs info is missing on the Batch execution review screen
## Bug description
More user friendly txs info is missed on the Batch execution review screen
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby

## Steps to reproduce
 2. click execute batch txs
 3. check review screen
 Current result: 
There is tx details but no proper decoded info about each tx 
image

## Expected result
we should keep proper user-friendly tx details for batch txs, as we have on safe-react
image

usame-algan:
## What it solves

Resolves #662 

## How this PR fixes it

Reuses the `Multisend` component to display individual actions of a multisend tx when executing a batch.

Some actions are missing information in the decoded data response. These include On-chain rejections and Native token transfers. This is a general issue with the `Multisend` component and should be fixed separately.

## How to test it

1. Open the Safe
2. Queue up a couple transactions
3. Click the Batch Execute button
4. Observe loading skeletons
5. Observe actions being displayed the same way as they are in Safe Apps and Tx Details

## Screenshots
Screenshot 2022-09-28 at 11 47 52


DiogoSoaress:
> Some actions are missing information in the decoded data response. These include On-chain rejections and Native token transfers. This is a general issue with the Multisend component and should be fixed separately.

Can you point or open an issue to keep track of this?

DiogoSoaress (changes requested):
By using the `<Multisend />` we are changing the Accordion style we had previously.

usame-algan:
> By using the <Multisend /> we are changing the Accordion style we had previously.

Good point, I extracted the accordion styles into the theme as a variant and also slightly refactored `Multisend` so we don't have to prop drill as much.

DiogoSoaress (approved):
Great job here! Nice to see all the Accordion customization in the theme 🙏 

One small remark, couldn't we show the account blockie inside the Accordions as well? It identifies immediately the address

Screenshot 2022-09-28 at 19 03 41

liliya-soroka:
Returned back:
Now I don't see the list of the included txs at all
image
https://0b585a55.web-core.pages.dev/rin:0xf2565317F3Ae8Ae9EA98E9Fe1e7FADC77F823cbD/transactions/queue

usame-algan:
> One small remark, couldn't we show the account blockie inside the Accordions as well? It identifies immediately the address

I like the idea and its a small change. Should we do it in this PR? It would also affect every tx detail view.

> Now I don't see the list of the included txs at all

Thanks for testing! The issue was that we were using the multiSendCallOnly contract even though one of the transaction is a delegate call. I switched to the normal multiSend contract which solved the issue.

liliya-soroka:
@DiogoSoaress , could you check this set of the tx , please - https://display-batch-execute-action.web-core.pages.dev/rin:0x938bae50a210b80EA233112800Cd5Bc2e7644300/transactions/queue ( staging CGW)
The fix worked when there were 3 tx, but doesn't work with 4 txs in the batch 
image

DiogoSoaress:
> Should we do it in this PR? It would also affect every tx detail view.

Fair point. I think we can address it separately if it is really a go.

src/components/transactions/TxDetails/TxData/DecodedData/SingleTxDecoded/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { isDeleteAllowance, isSetAllowance } from '@/utils/transaction-guards'
+import { Accordion, AccordionDetails, AccordionSummary, Typography } from '@mui/material'
+import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
+import css from './styles.module.css'
+import CodeIcon from '@mui/icons-material/Code'
+import { DelegateCallWarning } from '@/components/transactions/Warning'
+import { InfoDetails } from '@/components/transactions/InfoDetails'
+import EthHashInfo from '@/components/common/EthHashInfo'
+
+type SingleTxDecoded = {
```
DiogoSoaress, line 16:
```suggestion
type SingleTxDecodedProps = {
```

DiogoSoaress:
The block explorer icon is cut here 🤯 any idea where to get some more width?

Screenshot 2022-10-04 at 15 17 00

DiogoSoaress:
Can you please update the branch? We already merged this https://github.com/safe-global/web-core/pull/747 and we are not using MUI icons anymore. Plus, there are some conflicts that must be resolved

src/components/transactions/TxDetails/TxData/DecodedData/SingleTxDecoded/index.tsx
DiogoSoaress: nit

usame-algan:
> The block explorer icon is cut here 🤯 any idea where to get some more width?

Looks like it only affects some addresses depending on the character width. We looked at this problem in the past where one proposed solution was to use a monospaced font for the addresses. For now, I suggest we increase the modal width slightly even though its not an optimal solution.

PR #715, opened 2022-09-28
Fix: page headers w/o Safe name

katspaugh:
## What it solves
Screenshot 2022-09-28 at 12 04 16

## How this PR fixes it
Adds the same fixed offset when sticky as is added to the height.

## Screenshots
Screenshot 2022-09-28 at 12 06 32


iamacook (approved):

katspaugh:
I'll go ahead and merge this directly, as it's just a small CSS fix.

PR #719, opened 2022-09-28
fix: handle gas estimation error in createTx

### LINKED ISSUE #644: [Send NFT] No simulation and Submit is disabled for the tx that will most likely fail
## Bug description
No simulation and Submit is disabled for the tx that will most likely fail
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby

## Steps to reproduce
 1. Go to https://web-core.pages.dev/rin:0x1230B3d59858296A31053C1b8562Ecf89A2f888b/balances/nfts
 2.  try to send NFT that can not be send from the safe
 Current result: The error is displayed on the review screen, 0 is displayed instead of safe nonce, there is no Simulate option and submit button is disabled
https://web-core.pages.dev/rin:0x1230B3d59858296A31053C1b8562Ecf89A2f888b/balances/nfts

## Expected result
the same tx from the safe-react: safe nonce is not 0, simulate and submit options are available
image

image

DiogoSoaress:
## What it solves

Resolves #644

## How this PR fixes it
The SafeTx gas estimation call is now inside of a `try..catch`. Because the estimation failed with the initial param, both `recommendedNonce` and `safeTxGas` were not fetched.
If the estimation fails with the initial params, a second request with '0x' data is issued to be able to retrieve the recommendedNonce for the "Review Modal"

## How to test it
See steps of #644 
The Review Modal is expected to be operable now.

## Screenshots
Screenshot 2022-09-28 at 13 35 27

### DIFF
--- a/src/services/tx/txSender.ts
+++ b/src/services/tx/txSender.ts
@@ -59,7 +59,8 @@ export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: num
   // Set the recommended nonce and safeTxGas if not provided
   if (nonce === undefined) {
     const chainId = await safeSDK.getChainId()
-    const safeAddress = await safeSDK.getAddress()
+    const safeNonce = await safeSDK.getNonce()
+    const safeAddress = safeSDK.getAddress()
     let estimation: SafeTransactionEstimation
     try {
       estimation = await estimateSafeTxGas(String(chainId), safeAddress, txParams)
@@ -72,7 +73,7 @@ export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: num
         estimation = await estimateSafeTxGas(String(chainId), safeAddress, minTxParams)
       } catch (e) {
         logError(Errors._616, (e as Error).message)
-        estimation = { currentNonce: 0, recommendedNonce: 0, safeTxGas: '0' }
+        estimation = { currentNonce: safeNonce, recommendedNonce: safeNonce, safeTxGas: '0' }
       }
     }
     txParams = { ...txParams, nonce: estimation.recommendedNonce, safeTxGas: Number(estimation.safeTxGas) }
--- a/src/services/tx/txSender.ts
+++ b/src/services/tx/txSender.ts
@@ -57,11 +57,18 @@ export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: num
   // Set the recommended nonce and safeTxGas if not provided
   if (nonce === undefined) {
     const chainId = await safeSDK.getChainId()
+    const safeAddress = await safeSDK.getAddress()
     let estimation: SafeTransactionEstimation
     try {
       estimation = await estimateSafeTxGas(String(chainId), safeSDK.getAddress(), txParams)
     } catch (e) {
-      estimation = { currentNonce: 0, recommendedNonce: 0, safeTxGas: '0' }
+      try {
+        const minTxParams = { ...txParams, data: '0x', to: safeAddress }
+        estimation = await estimateSafeTxGas(String(chainId), safeSDK.getAddress(), minTxParams)
+      } catch (e) {
+        console.error('Error estimating SafeTx gas', e)
+        estimation = { currentNonce: 0, recommendedNonce: 0, safeTxGas: '0' }
+      }
     }
     txParams = { ...txParams, nonce: estimation.recommendedNonce, safeTxGas: Number(estimation.safeTxGas) }
   } else {
--- a/src/services/tx/txSender.ts
+++ b/src/services/tx/txSender.ts
@@ -50,30 +50,33 @@ const estimateSafeTxGas = async (
   })
 }
 
-const getEstimation = async (txParams: SafeTransactionDataPartial): Promise<SafeTransactionEstimation | undefined> => {
+const getRecommendedTxParams = async (
+  txParams: SafeTransactionDataPartial,
+): Promise<{ nonce: number; safeTxGas: number } | undefined> => {
   const safeSDK = getAndValidateSafeSDK()
-
   const chainId = await safeSDK.getChainId()
   const safeAddress = safeSDK.getAddress()
-
   let estimation: SafeTransactionEstimation | undefined
 
   try {
     estimation = await estimateSafeTxGas(String(chainId), safeAddress, txParams)
   } catch (e) {
     try {
-      /* If the initial transaction data causes the estimation to fail
-         we retry the request with a cancellation transaction to get the
-         recommendedNonce even if the original transaction will likely fail */
+      // If the initial transaction data causes the estimation to fail,
+      // we retry the request with a cancellation transaction to get the
+      // recommendedNonce, even if the original transaction will likely fail
       const cancellationTxParams = { ...txParams, data: EMPTY_DATA, to: safeAddress, value: '0' }
-
       estimation = await estimateSafeTxGas(String(chainId), safeAddress, cancellationTxParams)
     } catch (e) {
       logError(Errors._616, (e as Error).message)
+      return
     }
   }
 
-  return estimation
+  return {
+    nonce: estimation.recommendedNonce,
+    safeTxGas: Number(estimation.safeTxGas),
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/services/tx/txSender.ts
```diff
@@ -57,7 +57,19 @@ export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: num
...
+    let estimation: SafeTransactionEstimation
+    try {
+      estimation = await estimateSafeTxGas(String(chainId), safeSDK.getAddress(), txParams)
+    } catch (e) {
+      try {
+        const minTxParams = { ...txParams, data: '0x', to: safeAddress }
+        estimation = await estimateSafeTxGas(String(chainId), safeSDK.getAddress(), minTxParams)
+      } catch (e) {
+        console.error('Error estimating SafeTx gas', e)
+        estimation = { currentNonce: 0, recommendedNonce: 0, safeTxGas: '0' }
```
katspaugh, line 70: As discussed, let's not pass `recommendedNonce: 0` here, and instead take the current safe nonce (I think the SDK does it automatically if you pass `nonce: undefined`).

src/services/tx/txSender.ts
```diff
@@ -57,7 +57,19 @@ export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: num
...
+    const safeAddress = await safeSDK.getAddress()
+    let estimation: SafeTransactionEstimation
+    try {
+      estimation = await estimateSafeTxGas(String(chainId), safeSDK.getAddress(), txParams)
+    } catch (e) {
+      try {
+        const minTxParams = { ...txParams, data: '0x', to: safeAddress }
+        estimation = await estimateSafeTxGas(String(chainId), safeSDK.getAddress(), minTxParams)
+      } catch (e) {
+        console.error('Error estimating SafeTx gas', e)
```
katspaugh, line 69: `console.error` -> `logError(ErrorCodes.xxx)`

src/services/tx/txSender.ts
```diff
@@ -57,7 +57,19 @@ export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: num
...
     const chainId = await safeSDK.getChainId()
-    const estimation = await estimateSafeTxGas(String(chainId), safeSDK.getAddress(), txParams)
+    const safeAddress = await safeSDK.getAddress()
+    let estimation: SafeTransactionEstimation
+    try {
+      estimation = await estimateSafeTxGas(String(chainId), safeSDK.getAddress(), txParams)
+    } catch (e) {
+      try {
+        const minTxParams = { ...txParams, data: '0x', to: safeAddress }
+        estimation = await estimateSafeTxGas(String(chainId), safeSDK.getAddress(), minTxParams)
```
katspaugh, line 67: Please add a comment here that we're falling back to an empty transaction just to get the recommendedNonce even if the transaction will likely revert.

src/services/tx/txSender.ts
iamacook: It doesn't, hence setting it to either the recommended one or that which is passed as an argument:

image

In `safe-react` we use 0. I am unsure as to when a rejection estimation could fail though? The issue says we return `0` sometimes though.

src/services/tx/txSender.ts
iamacook, line 67: You call the SDK to get the address multiple times here. I'd just use the `safeAddress` variable.

src/services/tx/txSender.ts
iamacook: It was a standard transaction creation. Disregard my comment then.

src/services/tx/txSender.ts
DiogoSoaress: We have 3 potentially fitting error descs:
```
_612 = '612: Error estimating gas'
...
_616 = '616: Failed to retrieve recommended nonce',
_617 = '617: Error fetching safeTxGas',
```

I will add 617

katspaugh: IMHO 616 fits better.

src/services/tx/txSender.ts
DiogoSoaress: I am not sure how to approach this suggestion. Should I pass the currentSafeNonce to the createTx?
I can't call the hook to retrieve the currentNonce but I need a value in `SafeTransactionEstimation` type

src/services/tx/txSender.ts
```diff
@@ -57,7 +58,22 @@ export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: num
...
+    const safeAddress = await safeSDK.getAddress()
+    let estimation: SafeTransactionEstimation
+    try {
+      estimation = await estimateSafeTxGas(String(chainId), safeAddress, txParams)
+    } catch (e) {
+      try {
+        /* If the initial transaction data causes the estimation to fail
+         we retry the request with an empty transaction to obtain the
+         recommendedNonce even if the original transaction will likely fail */
+        const minTxParams = { ...txParams, data: '0x', to: safeAddress, value: '0' }
```
iamacook, line 70: Can you be explicit that this mimics a transcation rejection cancellation?

src/services/tx/txSender.ts
```diff
@@ -57,7 +58,22 @@ export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: num
   // Set the recommended nonce and safeTxGas if not provided
   if (nonce === undefined) {
     const chainId = await safeSDK.getChainId()
-    const estimation = await estimateSafeTxGas(String(chainId), safeSDK.getAddress(), txParams)
+    const safeAddress = await safeSDK.getAddress()
```
iamacook, line 61:
```suggestion
    const safeAddress = safeSDK.getAddress()
```

iamacook (approved):

src/services/tx/txSender.ts
```diff
@@ -57,7 +59,23 @@ export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: num
...
+         we retry the request with a cancellation transaction to get the
+         recommendedNonce even if the original transaction will likely fail */
+        const cancellationTxParams = { ...txParams, data: EMPTY_DATA, to: safeAddress, value: '0' }
+        estimation = await estimateSafeTxGas(String(chainId), safeAddress, cancellationTxParams)
+      } catch (e) {
+        logError(Errors._616, (e as Error).message)
+        estimation = { currentNonce: safeNonce, recommendedNonce: safeNonce, safeTxGas: '0' }
+      }
+    }
     txParams = { ...txParams, nonce: estimation.recommendedNonce, safeTxGas: Number(estimation.safeTxGas) }
```
katspaugh, line 79: I would leave the estimation as undefined in the final catch, and not modify the txParams in that case.
The txParams should already contain the current safe nonce, so you can remove `const safeNonce = await safeSDK.getNonce()` too.

DiogoSoaress: That works well! Just tried to remove the first catch to enter directly in this case and it is handled elegantly 👌 

👇 
```ts
try {
  estimation = await estimateSafeTxGas(String(chainId), safeAddress, txParams)
} catch (e) {
  logError(Errors._616, (e as Error).message)
  return safeSDK.createTransaction({ safeTransactionData: txParams })
}
```

DiogoSoaress: Just pushed then change

src/services/tx/txSender.ts
```diff
@@ -57,7 +59,22 @@ export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: num
...
+    } catch (e) {
+      try {
+        /* If the initial transaction data causes the estimation to fail
+         we retry the request with a cancellation transaction to get the
+         recommendedNonce even if the original transaction will likely fail */
+        const cancellationTxParams = { ...txParams, data: EMPTY_DATA, to: safeAddress, value: '0' }
+        estimation = await estimateSafeTxGas(String(chainId), safeAddress, cancellationTxParams)
+      } catch (e) {
+        logError(Errors._616, (e as Error).message)
+        return safeSDK.createTransaction({ safeTransactionData: txParams })
```
katspaugh, line 75: That's code duplication...

DiogoSoaress: Totally. Next commit fixes it. Thanks for the thorough review

liliya-soroka:
Verified, can be merged

PR #720, opened 2022-09-28
Fix: onboard theme colors

### LINKED ISSUE #672: Onboard styling doesn't match new theme
## Bug description

Some elements of the onboard modal don't match that of the new theme. The CSS variables of onboard need to be better mapped to our new theme

## Steps to reproduce
 1. Connect varying wallets and observe the colours do not match perfectly to our new theme.

## Expected result

- The stepper is `primary.main` in light mode.
- THe checkmark is a darker color in light mode.

## Obtained result

- The stepper is green in light mode.
- The checkmark of a connected wallet is white (and hard to see) in light mode.

## Screenshots

image

katspaugh:
## What it solves

Resolves #672

## How this PR fixes it
Adjusts onboard CSS vars.

## Screenshots
Screenshot 2022-09-28 at 13 38 03

Screenshot 2022-09-28 at 13 38 21

Screenshot 2022-09-28 at 13 38 36

Screenshot 2022-09-28 at 13 38 47


iamacook (approved):
Looks good!

@TanyaEfremova, are you alright with the way in which we highlight wallets here or should it match that of apps and transactions?

TanyaEfremova:
@iamacook nice improvement on the borders in the normal state! 

I think, in both modes (if possible given our palette) we can highlight the cards the same way as dApp cards for consistency, unless it would bring additional changes to the colour palette.

If that's the case, then it's not a priority.

katspaugh:
Added the same bg but the border color unfortunately can't be made exactly like our cards (because onboard allows to override only the box-shadow, not border).

Screenshot 2022-09-29 at 16 22 05

PR #721, opened 2022-09-28
Fix: avoid multiple safe-apps requests

### LINKED ISSUE #698: Request optimisation when accessing sections of the web interface
## Bug description
When accessing sections on the UI like the safe apps page the same GET request is performed 3 times instead of 1 to load resources.

## Steps to reproduce
Go to the safe-apps section (for example) and check the network tab on the developer tools.

## Expected result
Request performed once to load resources on the interface.

## Obtained result
Request performed 3 times to load resources on the interface.

https://user-images.githubusercontent.com/92332710/192573728-b7ed6e00-c0d4-43e4-a6aa-ec790ddaaa0a.mp4

katspaugh:
## What it solves

Resolves #698

Screenshot 2022-09-28 at 14 35 29

## How this PR fixes it
Multiple components were requesting remote Safe Apps almost simultaneously. Probably assuming they come from a centralized local state.

E.g. the two Dashboard widgets (Featured Apps and Safe Apps), and the SAFE token widget in the header.

I've added some primitive request caching to avoid this. We should probably consider storing apps in Redux.

### DIFF
--- a/src/hooks/safe-apps/useRemoteSafeApps.ts
+++ b/src/hooks/safe-apps/useRemoteSafeApps.ts
@@ -4,12 +4,27 @@ import { Errors, logError } from '@/services/exceptions'
 import useChainId from '@/hooks/useChainId'
 import useAsync, { AsyncResult } from '../useAsync'
 
+// To avoid multiple simultaneous requests (e.g. the Dashboard and the SAFE header widget),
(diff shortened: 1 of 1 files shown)


mikhailxyz (approved):

JagoFigueroa:
All bueno, thanks! 🚀

PR #725, opened 2022-09-28
fix: show newest conflict tx on dashboard

### LINKED ISSUE #669: Dashboard Tx queue widget - Tx shown should be the last one created with the same nonce
## Bug description
The list shows the first tx created with that nonce. It should be the last one or the one ready to be executed

## Environment
 - Browser: any
 - Wallet: any
 - Chain: any

## Steps to reproduce
 1. Go to a 2/x safe
 2. Create a tx and queue it
 3. Check is in the Tx queue widget
 4. Wait a couple of minutes
 5. Create the rejection to that tx (do not execute it).
 6. Check the Tx queue widget in the dashboard (1st expected result)
 7. Fully sign the fist tx created
 8. Check the Tx queue widget (2nd expected result)

## Expected result
1 - The rejection should be shown since it was the latest tx created
2 - Now the fully signed and ready to be executed tx should be the one shown in the widget.

## Obtained result
The first tx created is shown in ever scenario

Note:
It should be the same logic as the "Execute batch" button, where when you hover the txs that will be executed are highlighted. Those tx are the ones that should be shown in the widget.

Screenshot:
image
image

iamacook:
## What it solves

Resolves #669

## How this PR fixes it

The latest `timestamp`ed transaction is picked from conflict groups.

## How to test it

1. Queue a transaction.
2. Observe the queued transaction in the pending list.
3. Queue a transaction rejection.
4. Observe the transaction rejection replaces the transaction in the pending list.

## Screenshots
image

image

image

image


src/components/dashboard/PendingTxs/PendingTxsList.tsx
```diff
@@ -43,12 +42,22 @@ const PendingTxsList = ({ size = 4 }: { size?: number }): ReactElement | null =>
   const router = useRouter()
   const url = `${AppRoutes.transactions.queue}?safe=${router.query.safe}`
 
-  const queuedTxns: Transaction[] = (page?.results || []).filter(isTransactionListItem)
+  const queuedTxns = useMemo(() => {
+    return groupConflictingTxs(page?.results || [])
+      .map((group) => {
+        if (Array.isArray(group)) {
+          const latestTx = group.reduce((acc, tx) => {
+            return tx.transaction.timestamp > acc.transaction.timestamp ? tx : acc
```
usame-algan, line 50: I would suggest naming them `prev` and `curr` or something similar ([mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#syntax) uses `previousValue` and `currentValue` in their docs) since we are not accumulating a value in this case.

src/components/dashboard/PendingTxs/PendingTxsList.tsx
```diff
@@ -43,12 +42,22 @@ const PendingTxsList = ({ size = 4 }: { size?: number }): ReactElement | null =>
...
   const url = `${AppRoutes.transactions.queue}?safe=${router.query.safe}`
 
-  const queuedTxns: Transaction[] = (page?.results || []).filter(isTransactionListItem)
+  const queuedTxns = useMemo(() => {
+    return groupConflictingTxs(page?.results || [])
+      .map((group) => {
+        if (Array.isArray(group)) {
+          const latestTx = group.reduce((prev, curr) => {
+            return curr.transaction.timestamp > prev.transaction.timestamp ? curr : prev
+          })
```
katspaugh, line 51: Aren't they already sorted by date by the backend?

In any case, `group.sort((a, b) => a.timestamp - b.timestamp).pop()` would be probably much simpler?

katspaugh (approved):
Neat!

DiogoSoaress (approved):

liliya-soroka:
Returned:
doesn't work for me
1. reject tx in the next section
Current result: the previous tx is displayed instead of the latest one
image
image

iamacook:
> doesn't work for me

It turns out we weren't ordering our grouping. I've adjusted it and it now works, and as a bonus grouped transactions in the transaction are ordered as well:

image

image

PR #726, opened 2022-09-28
style: make `button` element inherit parent font globally

### LINKED ISSUE #668: "Hide All" text has a default font-family
## Bug description
Most text in the site has the DM sans font family. The "Hide All" text in signed tx has not font-family so is just using a default one

## Environment
 - Browser: any
 - Wallet: any
 - Chain: any

## Steps to reproduce
 1. Go to any safe with tx
 2. Open the tx details

Expected result
The text should have DM sans as font-family

## Screenshots
image

DiogoSoaress:
## What it solves

Resolves #668

## How this PR fixes it
Sets a global style for `button` in order to keep the agent stylesheet from overwriting the font property

## How to test it
1. Go to the Tx Signers in the history list.
2. The "Hide All" button shall have `DM Sans,sans-serif` font-family

## Screenshots
Screenshot 2022-09-28 at 16 43 32

### DIFF
--- a/src/styles/theme.ts
+++ b/src/styles/theme.ts
@@ -162,13 +162,6 @@ const initTheme = (darkMode: boolean) => {
           sizeLarge: { fontSize: '16px' },
         },
       },
-      MuiLink: {
-        styleOverrides: {
-          button: ({ theme }) => ({
-            fontFamily: theme.typography.fontFamily,
-          }),
-        },
-      },
       MuiAccordion: {
         styleOverrides: {
           root: ({ theme }) => ({
--- a/src/styles/theme.ts
+++ b/src/styles/theme.ts
@@ -162,6 +162,13 @@ const initTheme = (darkMode: boolean) => {
           sizeLarge: { fontSize: '16px' },
         },
       },
+      MuiLink: {
+        styleOverrides: {
+          button: ({ theme }) => ({
+            fontFamily: theme.typography.fontFamily,
+          }),
+        },
+      },
       MuiAccordion: {
         styleOverrides: {
           root: ({ theme }) => ({
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -32,6 +32,10 @@ a {
   text-decoration: none;
 }
 
+button {
+  font: inherit;
+}
+
 * {
   box-sizing: border-box;
 }


src/styles/theme.ts
```diff
@@ -162,6 +162,13 @@ const initTheme = (darkMode: boolean) => {
           sizeLarge: { fontSize: '16px' },
         },
       },
+      MuiLink: {
+        styleOverrides: {
+          button: ({ theme }) => ({
+            fontFamily: theme.typography.fontFamily,
```
iamacook, line 168: Can we not globally apply this to the entire theme?

DiogoSoaress: I think the component MuiLink overrides it

usame-algan: In the [docs](https://mui.com/material-ui/react-link/#accessibility) they also apply `variant="body2"` to the Link and it overrides the native browser stylesheet thats causing the issue with font-family.

mikhailxyz: Why do we have to override the font family for this component in particular while we have the global font family set?

https://github.com/safe-global/web-core/blob/main/src/styles/theme.ts#L68-L69

Wouldn't it be a bug in the material UI if it doesn't work just for this component?

iamacook: > overrides the native browser stylesheet thats causing the issue with font-family

The thing is, [we're MUI's `CSSBaseline`](https://github.com/safe-global/web-core/blob/main/src/pages/_app.tsx#L101).

iamacook: [We need to override the `CssBaseline` as well](https://mui.com/material-ui/customization/typography/#self-hosted-fonts).

DiogoSoaress: Even having the self-hosted fonts in the `MuiCssBaseline` override, the `button` font-family has to be overwritten it seems. Tried it in Chrome and Firefox.

DiogoSoaress:
I've moved the font imports to the theme file as per the [documentation](https://mui.com/material-ui/customization/typography/#self-hosted-fonts)

src/styles/theme.ts
```diff
@@ -109,6 +109,23 @@ const initTheme = (darkMode: boolean) => {
       },
     },
     components: {
+      MuiCssBaseline: {
+        styleOverrides: `
+          @font-face {
```
katspaugh, line 114: Why does it have to be in the MUI theme and not in the global CSS? What's the difference?

DiogoSoaress: I don't think there is a difference. They state
> In order to globally define Raleway as a font face, the [CssBaseline](https://mui.com/material-ui/react-css-baseline/) component can be used (or any other CSS solution of your choice).

DiogoSoaress: Should I revert it? Also find it more readable in `globals.css`

iamacook: If the font is defined in the theme, it can be used by others, e.g. when we export it in SRC.

DiogoSoaress: Oh yes. Didn't think of that.

katspaugh: Yes, no unnecessary changes is our motto.

katspaugh (approved):
👍

PR #727, opened 2022-09-28
Fix: page placeholder style

katspaugh:
## What it solves

We lost the image size after the illustrations were updated.

Screenshot 2022-09-28 at 18 28 01

### DIFF
--- a/src/components/common/PagePlaceholder/styles.module.css
+++ b/src/components/common/PagePlaceholder/styles.module.css
@@ -7,3 +7,10 @@
   text-align: center;
   flex: 1;
 }
+
+.container svg,
+.container img {
+  min-height: 200px;
+  max-height: 100%;
+  width: auto;
+}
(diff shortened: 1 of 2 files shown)


iamacook (commented):
The size was reduced to [match the Figma design](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=1%3A5), although it should admittedly be moved further down the page.

katspaugh:
I see, thanks @iamacook. Removed the height and added a top padding instead.

katspaugh:
Screenshot 2022-09-28 at 18 40 02

iamacook (commented):
Looks good now but we need to adjust the pending transaction list widget as it is using this placeholder:

image

katspaugh:
Hmm, I distinctly remember removing it and using an image there. Will do it again.

katspaugh:
Will merge w/o QA since it's just a few lines of CSS.

PR #731, opened 2022-09-29
fix: `Backdrop` `backgroundColor` based on theme

iamacook:
## What it solves

Resolves theme-based `Backdrop` `backgroundColor`

## How this PR fixes it

The `Backdrop` `backgroundColor` is set from the theme `palette`, thus reacting to light/dark mode.

## How to test it

1. Open the Safe in light mode and open the wallet popper. Observe a light background.
2. Open the Safe in dark mode and open the wallet popper. Observe a dark background.

## Screenshots

image

image


src/utils/style.ts
```diff
@@ -0,0 +1,4 @@
+export const setOpacity = (hex: string, alpha: number) =>
```
iamacook, line 1: Taken from [here](https://stackoverflow.com/a/66143374/6585368).

usame-algan:
I would suggest we go with a neutral color in light mode instead of green. Especially when multiple backdrops stack on top of each other its very distracting imo.
Screenshot 2022-09-29 at 09 41 08

iamacook:
> I would suggest we go with a neutral color in light mode instead of green.

I've altered it to `background.main`.

src/styles/theme.ts
```diff
@@ -428,10 +429,10 @@ const initTheme = (darkMode: boolean) => {
       },
       MuiBackdrop: {
         styleOverrides: {
-          root: {
+          root: ({ theme }) => ({
             backdropFilter: 'blur(1px)',
-            backgroundColor: 'rgba(228, 232, 241, 0.75)',
-          },
+            backgroundColor: setOpacity(theme.palette.background.main || '', 0.75),
```
usame-algan, line 434: MUI ships with a util function called `alpha` which we can use: `alpha(theme.palette.background.main, 0.75)`. Do we need the fallback to ''? Afaik `background.main` is defined in both color palettes.

usame-algan (changes requested):

usame-algan:
After discussion and comparing with the [designs](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=1056%3A41070) it looks like we need a new color mapping that is #303033 in light-mode and #000000 in dark-mode as the backdrop color. Imo using `background.main` already brings an improvement in dark-mode but doesn't separate the background from the modals enough in light-mode.

I suggest we merge this PR and tackle the new color mapping separately as it requires syncing with design on the color scheme. cc @TanyaEfremova

usame-algan (approved):

francovenica:
Question:
Do the backgrounds have to overlap if you open multiple modals? like if you click "new transactions > Send funds", because 2 modals were opened 2 background blur are stuck on each other. 
I'd assume that there is no way to know if there is already a background blur active.

Issue:
Opening the onboard modal to select a wallet has a different background color than any other modal like send funds.
In light mode and dark mode the background is the same color it seems
image

I'd expect to have this background color 
image

iamacook:
> Question: Do the backgrounds have to overlap if you open multiple modals?

Yes, they stack.

> Issue: Opening the onboard modal to select a wallet has a different background color than any other modal like send funds. 

It is currently not possible to customise the backdrop of onboard either.

I will therefore merge this as it is a simple CSS tweak. I'll get in touch with Blocknative and maybe we can revisit their modal backdrop in the future.

PR #737, opened 2022-09-29
Fix: add safe nonce tooltip

### LINKED ISSUE #736: [Settings] Tooltip with nonce info is missed near safe nonce
## Bug description
Tooltip with nonce info is missed near safe nonce
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby

## Steps to reproduce
 1. Go to https://web-core.pages.dev/rin:0xf2565317F3Ae8Ae9EA98E9Fe1e7FADC77F823cbD/settings/setup
 2.  scroll down to the safe info

Current result: 
image

## Expected result
The tooltip should contain the same info as we have in the rsafe-eact app for the safe nonce
image

katspaugh:
## What it solves

Resolves #736

## How this PR fixes it

Adds a tooltip to the Safe nonce in settings.

Screenshot 2022-09-29 at 12 08 03

### DIFF
--- a/src/pages/settings/setup.tsx
+++ b/src/pages/settings/setup.tsx
@@ -35,7 +35,10 @@ const Setup: NextPage = () => {
             <Grid item lg={4} xs={12}>
               <Typography variant="h4" fontWeight={700}>
                 Safe nonce
-                <Tooltip placement="top" title="For security reasons, transactions made with Safe need to be executed in order. The nonce shows you which transaction will be executed next. You can find the nonce for a transaction in the transaction details.">
+                <Tooltip
+                  placement="top"
+                  title="For security reasons, transactions made with Safe need to be executed in order. The nonce shows you which transaction will be executed next. You can find the nonce for a transaction in the transaction details."
+                >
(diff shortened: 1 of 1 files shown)


src/pages/settings/setup.tsx
```diff
@@ -34,8 +35,17 @@ const Setup: NextPage = () => {
             <Grid item lg={4} xs={12}>
               <Typography variant="h4" fontWeight={700}>
                 Safe nonce
+                <Tooltip title="For security reasons, transactions made with Safe need to be executed in order. The nonce shows you which transaction will be executed next. You can find the nonce for a transaction in the transaction details.">
```
iamacook, line 38: I'd add `placement="top"` to the `Tooltip` so you can still read the current nonce. Most `Tooltip`s in the app use this.

katspaugh: Good idea!

iamacook (approved):

liliya-soroka:
Verified and approved for merging

PR #760, opened 2022-09-29
fix: add missing `required` prop to relevant fields

### LINKED ISSUE #742: * is missing near the enter fields for the mandatory data all over the app
## Bug description
"*" is missing near the enter fields for the mandatory data all over the app
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby

## Steps to reproduce
 1. Go through the create safe flow or replace owner ( any other flow in the app)
Current result: there is no indication of mandatory data. So the only way is to leave a field empty and check an error message 

## Expected result
Could you add again "*" for all mandatory fields?

iamacook:
## What it solves

Resolves #742

## How this PR fixes it

Fields that are required now append an asterisk to their label:

- Asset/amount when creating a spending limit
- Gas limit/nonce/safeTxGas (on older Safes) when editing transaction parameters
- Collection/asset when sending an NFT
- Asset/amount when sending an asset

## How to test it

View the aforementioned fields and observe the asterisk.

## Screenshots

image
image
image
image

### DIFF
--- a/src/components/common/AddressInput/index.tsx
+++ b/src/components/common/AddressInput/index.tsx
@@ -83,6 +83,7 @@ const AddressInput = ({ name, validate, required = true, ...props }: AddressInpu
             ...(props.InputLabelProps || {}),
             shrink: !!watchedValue || props.focused,
           }}
+          required={required}
           {...register(name, {
             required,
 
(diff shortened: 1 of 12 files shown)


katspaugh (approved):
👍

PR #761, opened 2022-09-29
fix(safe-apps): Return empty app when manifest not loading properly

yagopv:
## What it solves
Resolves https://github.com/safe-global/web-core/issues/738

## How this PR fixes it
By adding an empty app initializer just in case the retrieval of the manifest fails (Current approach in safe-react)


mikhailxyz:
While I understand it fixes the issue, something doesn't add up. We list the manifest file as a hard requirement for Safe Apps - see 2) at https://docs.gnosis-safe.io/build/sdks/safe-apps/releasing-your-safe-app

If the manifest is not working or cannot be loaded, why should we fall back to an empty one?

yagopv:
yep but for release we want this to work as we talked in the daily right?

yagopv:
We will follow up this after the release in order to fix the manifest problems by contacting affected apps

https://github.com/safe-global/web-core/issues/782

After the apps start to work fine again we can remove the workaround

mikhailxyz (approved):

JagoFigueroa:
All bueno. From the list on the original ticket only 2 do not load now.

- BarnBridge which also does not load in legacy web.
- Plasma bridge, with the SDK limited to the legacy web and present in the other ticket.

I think we can close this ticket. Thanks!

PR #765, opened 2022-09-29
feat: notify of tx awaiting confirmation

iamacook:
## What it solves

Notification of awaiting confirmations

## How this PR fixes it

As in `safe-react` a notification is shown when a transaction is awaiting confirmation from the user on Safe load.

## How to test it

1. Queue a transaction on a 2/n Safe.
2. Switch to the other owner of the Safe and observe the notification.

Note: switching back and forth should *not* show the notification again until page refresh.

## Screenshots

awaiting confirmation notification

### DIFF
--- a/src/hooks/useTxNotifications.ts
+++ b/src/hooks/useTxNotifications.ts
@@ -1,10 +1,14 @@
 import { useEffect } from 'react'
-import { showNotification } from '@/store/notificationsSlice'
-import { useAppDispatch } from '@/store'
+import { selectNotifications, showNotification } from '@/store/notificationsSlice'
+import { useAppDispatch, useAppSelector } from '@/store'
 import { TxEvent, txSubscribe } from '@/services/tx/txEvents'
 import { AppRoutes } from '@/config/routes'
 import useSafeInfo from './useSafeInfo'
 import { useCurrentChain } from './useChains'
+import useTxQueue from './useTxQueue'
+import { isTransactionListItem } from '@/utils/transaction-guards'
+import { TransactionStatus } from '@gnosis.pm/safe-react-gateway-sdk'
+import { selectPendingTxs } from '@/store/pendingTxsSlice'
 
 const TxNotifications = {
   [TxEvent.SIGN_FAILED]: 'Signature failed. Please try again.',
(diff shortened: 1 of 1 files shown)


src/hooks/useTxNotifications.ts
```diff
@@ -68,6 +76,44 @@ const useTxNotifications = (): void => {
...
+
+    if (!items || items.length === 0) {
+      return
+    }
+
+    const { transaction } = items[0]
+
+    if (transaction.txStatus !== TransactionStatus.AWAITING_CONFIRMATIONS || pendingTxs[transaction.id]) {
+      return
+    }
```
katspaugh, line 99: I would extract this part out into a function called `getFirstQueuedTransaction`.

src/hooks/useTxNotifications.ts
```diff
@@ -68,6 +76,44 @@ const useTxNotifications = (): void => {
       unsubFns.forEach((unsub) => unsub())
     }
   }, [dispatch, safeAddress, chain?.shortName])
+
+  /**
+   * Show a notification when a transaction is awaiting confirmation on Safe load
```
katspaugh, line 81: Thanks for the comment. Please add some more details, e.g. "if there's at least one tx awaiting a confirmation, show a notification for it".

src/hooks/useTxNotifications.ts
```diff
@@ -68,6 +76,44 @@ const useTxNotifications = (): void => {
...
+  useEffect(() => {
+    const items = page?.results?.filter(isTransactionListItem)
+
+    if (!items || items.length === 0) {
+      return
+    }
+
+    const { transaction } = items[0]
+
+    if (transaction.txStatus !== TransactionStatus.AWAITING_CONFIRMATIONS || pendingTxs[transaction.id]) {
```
katspaugh, line 97: Don't we also need to check that it's awaiting _your_ signature?

iamacook: Good point. I'll adjust it now.

src/hooks/useTxNotifications.ts
iamacook: I've extracted it to a memoised variable because it would otherwise have a lot of arguments with the new checks. What do you think?

katspaugh (approved):
👍

src/hooks/useTxNotifications.ts
katspaugh: Sure, as long as it's encapsulated 👍

liliya-soroka:
Verified

PR #766, opened 2022-09-29
fix: do not allow threshold greater than owners

### LINKED ISSUE #758: [Safe creation] Possible to set a safe with a higher threshold than owners
## Bug description
Playing around with safe creation it is possible to set an inconsistent status of a higher threshold than owners the safe has.

## Environment
 - Browser: Any
 - Wallet: MetaMask
 - Chain: Any

## Steps to reproduce
 1. create a 3/3 safe (for example)
 2. Remove a couple of the owners
 3. Continue with the creation flow

## Expected result
Safe reflects a valid threshold or at least not possible to continue the flow without re-setting a valid one.

## Obtained result
https://user-images.githubusercontent.com/92332710/193040555-06302827-ce73-4658-9599-0671b481f3f7.mp4

DiogoSoaress:
## What it solves

Resolves #758

## How this PR fixes it
Watches the threshold value on the Safe creation Owners policy step.
Adds a useEffect to guarantee that the threshold selected does not exceed the number of owners.

## How to test it
1. Add 3 owners when Safe creation.
2. Set threshold of 3
3. Delete one owner
4. The threshold adjusts automatically
5. Going to the next step you see the correct owners policy

## Analytics changes
None

## Screenshots
threshold fixed


src/components/create-safe/steps/OwnerPolicyStep.tsx
```diff
@@ -30,7 +30,7 @@ const OwnerPolicyStep = ({ params, onSubmit, setStep, onBack }: Props): ReactEle
     address: wallet?.address || '',
   }
 
-  const defaultThreshold = params.threshold || 1
+  const defaultThreshold = 1
```
iamacook, line 33: We need to use `params.threshol` in order to preserve the value between steps. Try going forward then back a step and you'll see it remains as 1.

DiogoSoaress: I saw that no `params.threshold` were being kept between steps. Perhaps it's a different bug to fix. I will revert this line then

DiogoSoaress: reverted

iamacook (approved):

src/components/create-safe/steps/OwnerPolicyStep.tsx
```diff
@@ -53,6 +53,11 @@ const OwnerPolicyStep = ({ params, onSubmit, setStep, onBack }: Props): ReactEle
     append({ name: '', address: '' })
   }
 
+  const threshold = watch('threshold')
+  useEffect(() => {
+    setValue('threshold', threshold > fields.length ? fields.length : threshold)
+  }, [fields, setValue, threshold])
```
katspaugh, line 59: Can't you just validate the select instead?

katspaugh: Also, if we populate the select options, how can it even have a threshold bigger than owners?

katspaugh: If it doesn't re-render, add a `key={fields.length}` property to the select

katspaugh (changes requested):

src/components/create-safe/steps/OwnerPolicyStep.tsx
DiogoSoaress: The form keeps the previously selected threshold.

DiogoSoaress: > Can't you just validate the select instead?

But you will have to set the threshold anyway no? My thought here is that if we select a "N out of N" policy that we keep it

katspaugh: The last selected threshold could be passed in defaultValue. Math max of that and owners.

DiogoSoaress: I was trying it out but didn't quite work as you suggest: setting the value in the `<Select />` does not set the form value unfortunately.

```tsx
<Select
  {...register('threshold')}
  defaultValue={threshold || defaultThreshold}
  value={threshold > fields.length ? fields.length : threshold}
  key={fields.length}
>

Comment (iamacook) [reply src/components/create-safe/steps/OwnerPolicyStep.tsx:59]: `setValue` is required because the RHF state has to be updated as well.

Comment (DiogoSoaress) [reply src/components/create-safe/steps/OwnerPolicyStep.tsx:59]: Yes, because of that I couldn't get around the `useEffect`

Comment (katspaugh) [reply src/components/create-safe/steps/OwnerPolicyStep.tsx:59]: Let me play with it, I'll get back to you.

**Review by katspaugh — APPROVED** [review]
I've made a commit to your owner/threshold PR. You were right there’s not much we can do other than setValue. I moved it to on remove tho.

File: src/components/create-safe/steps/OwnerPolicyStep.tsx [hunk 3]
```diff
@@ -41,18 +46,33 @@ const OwnerPolicyStep = ({ params, onSubmit, setStep, onBack }: Props): ReactEle
...
 
   const addOwner = () => {
     append({ name: '', address: '' })
   }
 
+  const removeOwner = (index: number) => {
+    remove(index)
+
+    // Make sure the threshold is not higher than the number of owners
+    if (currentThreshold >= ownerLength) {
```
iamacook, line 71: Shouldn't this just be `currentThreshold > ownerLength`?

JagoFigueroa:
Can confirm that this is working as expected now. Thanks!

PR #767, opened 2022-09-29
fix: check if transaction is next

### LINKED ISSUE #745: [Tx confirmation]Confusing text when the last confirmation can be received but the execution is not available
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to the queued list
 2. click confirm to add the last required confirmation for any tx in the queued list except the tx in the next section

Current result: "Sign or immediately execute this transaction." is displayed to the user but actually, the user can not execute tx as it's not the next for the execution tx in the queued list
image

## Expected result
Don't mention execution if the "execution" operation is not available

iamacook:
## What it solves

Resolves #745

## How this PR fixes it

The transaction is deemed executable if it is next. The queue list is first checked for checked for the presence of a "Next" label and then that the first transaction is that of the chosen one.

## How to test it

1. Queue two transactions on a 2/n Safe.
2. Switch to another owner and confirm the second transaction. Observe the correct, "Sign this transaction.", text.

## Screenshots

check next


src/components/tx/modals/ConfirmTxModal/ConfirmProposedTx.tsx
```diff
@@ -24,8 +25,10 @@ const ConfirmProposedTx = ({ txSummary, onSubmit }: ConfirmProposedTxProps): Rea
   const wallet = useWallet()
   const safeAddress = useSafeAddress()
   const chainId = useChainId()
+  const { page } = useTxQueue()
+
   const txId = txSummary.id
-  const canExecute = isExecutable(txSummary, wallet?.address || '')
+  const canExecute = isExecutable(txSummary, wallet?.address || '') && isNextTx(txId, page?.results || [])
```
usame-algan, line 31: Can we make the `isNextTx` function part of the `isExecutable` function? Currently, `isExecutable` returns true even if a tx is not executable which seems like a bug.

iamacook: True. I'll look into it.

src/utils/transaction-guards.ts
```diff
@@ -124,8 +125,24 @@ export const isSignableBy = (txSummary: TransactionSummary, walletAddress: strin
   return !!executionInfo?.missingSigners?.some((address) => address.value === walletAddress)
 }
 
-export const isExecutable = (txSummary: TransactionSummary, walletAddress: string): boolean => {
-  if (!txSummary.executionInfo || !isMultisigExecutionInfo(txSummary.executionInfo)) {
+export const _isNextTx = (txId: string, items: TransactionListItem[]) => {
+  const hasNextLabel = items.find(isLabelListItem)?.label === LabelValue.Next
```
katspaugh, line 129: Relying on a label seems hacky tbh.
Can’t we just check that it’s first in the queue and that the nonce is >= safe nonce?

iamacook: I didn't think of that - thanks! All we need to do is compare the nonce. I've adjusted it accordingly.

src/utils/transaction-guards.ts
```diff
@@ -124,8 +125,12 @@ export const isSignableBy = (txSummary: TransactionSummary, walletAddress: strin
   return !!executionInfo?.missingSigners?.some((address) => address.value === walletAddress)
 }
 
-export const isExecutable = (txSummary: TransactionSummary, walletAddress: string): boolean => {
-  if (!txSummary.executionInfo || !isMultisigExecutionInfo(txSummary.executionInfo)) {
+export const isExecutable = (txSummary: TransactionSummary, walletAddress: string, safe: SafeInfo): boolean => {
+  if (
+    !txSummary.executionInfo ||
+    !isMultisigExecutionInfo(txSummary.executionInfo) ||
+    safe.nonce !== txSummary.executionInfo.nonce
```
katspaugh, line 132: What if it's lower than the safe nonce?

katspaugh: We use the same strict comparison in SignOrExecute tho.

katspaugh: Maybe it's good to not allow to skip nonces.

katspaugh (approved):
Looks good 👍

src/utils/transaction-guards.ts
iamacook: Just to summarise why we resolved this: you can't set a nonce lower than the Safe nonce and when it's +2> than the Safe nonce, it's not executable.

Issue #755, opened 2022-09-29
[Txs list]Addressbook name is not taken into account in the txs list


liliya-soroka:
## Bug description
Known addresses( back-end or address book) are not taken into account in the txs list
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby

## Steps to reproduce
 1. Add the spending limit address to the address book - rin:0x10A7EC8D10CD175dC33781fB9Cf3394220Fac78c
 2.  go to your safe and delete any limit 
 
Current result: The tx is created but "contract interaction" is used as a name for the tx instead of the spending limit address name from the addressbook

## Expected result
In the old app, we implemented known addresses with the rule that the address book name has higher priority over any other name in txs list 
It should be implemented in the new app too  to the better tx decoding

PR #770, opened 2022-09-29
fix: only use address book/known names

### LINKED ISSUE #749: [Settings-module] Addressbook name has the priority for address names on UI
## Bug description
Addressbook name has the priority for address names on UI or make sure that the same name is used in the modules table and on the remove module review screen
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain:rinkeby

## Steps to reproduce
 1. make sure that a spending limits module is enabled in your safe
 2. add the spending limit module address in the address book with a custom name
 3. go to the settings->modules
 Current result: you will see that the address has a "spending limits module" name ( probably from the back-end known addresses list

4. click remove spending limit module
Current result: on the review screen you will see the address name from the addressbook

## Expected result
Make sure that the address book name has priority over the known address name or any other sourth.

iamacook:
## What it solves

Resolves #749, #775 & #755

## How this PR fixes it

- 749/775: We no longer hardcode any module names, including "Unknown module".
- 755: Contract interactions in the transaction list return the address book entry name if it is present as their type.

## How to test it

- 749/775: Ensure the spending limit or other module is enabled on your Safe and observe that they have no name.
- 755: Interact with a contract then add it's address to your address book. Observe the transaction list type now returns the address book name

## Screenshots

module name

image

### DIFF
--- a/src/components/settings/SafeModules/index.tsx
+++ b/src/components/settings/SafeModules/index.tsx
@@ -14,7 +14,7 @@ const NoModules = () => {
   )
 }
 
-const ModuleDisplay = ({ moduleAddress, chainId, name }: { moduleAddress: string; chainId: string; name: string }) => {
+const ModuleDisplay = ({ moduleAddress, chainId, name }: { moduleAddress: string; chainId: string; name?: string }) => {
   const isGranted = useIsGranted()
 
   return (
@@ -63,7 +63,7 @@ const SafeModules = () => {
                   key={module.value}
                   chainId={safe.chainId}
                   moduleAddress={module.value}
-                  name={module.name || ''}
+                  name={module.name}
                 />
               ))
             )}
--- a/src/components/settings/SafeModules/index.tsx
+++ b/src/components/settings/SafeModules/index.tsx
@@ -5,10 +5,6 @@ import { Paper, Grid, Typography, Box, Link } from '@mui/material'
 import css from './styles.module.css'
 import { RemoveModule } from '@/components/settings/SafeModules/RemoveModule'
 import useIsGranted from '@/hooks/useIsGranted'
-import { getSpendingLimitModuleAddress } from '@/services/contracts/spendingLimitContracts'
-import { sameAddress } from '@/utils/addresses'
-
-export const DEFAULT_MODULE_NAME = 'Unknown module'
 
 const NoModules = () => {
   return (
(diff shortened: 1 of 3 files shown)


src/components/settings/SafeModules/index.tsx
```diff
@@ -18,14 +14,6 @@ const NoModules = () => {
   )
 }
 
-const getModuleName = (chainId: string, address: string): string => {
-  if (sameAddress(getSpendingLimitModuleAddress(chainId), address)) {
-    return 'Spending limit module'
-  }
```
katspaugh, line 24: Wasn't this quite useful?

src/hooks/useTransactionType.ts
```diff
@@ -84,20 +85,14 @@ const getTxType = (tx: TransactionSummary): TxType => {
...
       }
     }
   }
 }
-
-export const useTransactionType = (tx: TransactionSummary): TxType => {
-  return useMemo(() => {
-    return getTxType(tx)
-  }, [tx])
-}
```
DiogoSoaress, line 103: won't we lose performance for dismissing this memoization?

src/components/settings/SafeModules/index.tsx
katspaugh: Maybe we could initialize the AB with some preset address like this?

iamacook: I like the idea but I think it's best backend decide the known addresses. It means there isn't a populated address book from the get go as well. What do you think?

src/hooks/useTransactionType.ts
iamacook: The performance increased seemed neglibile when looking at the dependency array but I'll put it back.

DiogoSoaress (approved):
Thanks! It LGTM 👍

src/components/settings/SafeModules/index.tsx
katspaugh: Backend having it would be best of course.

liliya-soroka: we only need somebody with permissions on prod to add the module to the known addresses list

JagoFigueroa:
All of the fixes look good to me, thanks! Moving the related tickets to QA done.

PR #776, opened 2022-09-30
fix: spending limit amount validation

### LINKED ISSUE #774: [Spending limits] Not possible to set one with decimals
## Bug description
I was trying to create a new spending limit of 0.5 eth for one of my owners and it is not possible, we should fix this.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Rinkeby

## Steps to reproduce
 1. On any owned safes, go to Settings -> Spending Limits and try to set one with decimals.

## Expected result
(legacy)
Screenshot 2022-09-30 at 08 55 33

## Obtained result
(revamp)
Screenshot 2022-09-30 at 08 54 48

iamacook:
## What it solves

Resolves #774

## How this PR fixes it

Numbers <1 are validated for their decimal length, wheres those >=1 that they're a uint96.

## How to test it

Create a new spending limit and test the lower limit and higher limit.

## Screenshots

image

image


src/components/settings/SpendingLimits/NewSpendingLimit/steps/SpendingLimitForm.tsx
```diff
@@ -110,9 +110,9 @@ export const SpendingLimitForm = ({ data, onSubmit }: Props) => {
               {...register('amount', {
                 required: true,
                 validate: (val) =>
-                  validateAmount(val) ||
-                  validateDecimalLength(val, selectedToken?.tokenInfo.decimals) ||
-                  validateUpperSpendingLimitAmount(val),
+                  validateAmount(val) || Number(val) < 1
```
katspaugh, line 113: Why does the < 1 check has to be here? Can't you do it in `validateDecimalLength` and `validateUpperSpendingLimitAmount`?

iamacook: Good point. I will have to rework `validateUpperSpendingLimitAmount` as it doesn't like decimals at all.

katspaugh (approved):

francovenica:
I'm having a issue in this PR, The networks are not loading, blocking me from connecting or loading safes.
I've checked in https://web-core.pages.dev/welcome and I have no issues there, so I assume the issue is just in ths PR.

image

iamacook:
> I'm having a issue in this PR, The networks are not loading,

I believe this is related to the CGW migration. I've merged `main` (and on all my other opened PRs) and it should work now.

PR #777, opened 2022-09-30
Refactor: simplify SignOrExecute props

katspaugh:
## What it solves

`isExecutable={safe.threshold === 1}` was passed to SignOrExecuteForm in every modal except for the ConfirmTxModal. So I moved it into the SignOrExecuteForm itself if txId isn't passed (meaning it's a new tx).

Also I removed the redirect to tx on submit, because we now have a notification with the transaction link popping up after tx creation.

Also fixed that txId wasn't passed to the tx events in the on-chain signing flow.

## How to test

- Create a transaction in a 1/1 Safe, it should be immediately executable
- Create a tx in a 2/2 Safe, it should not be executable
- Confirm an existing signed transaction, it should be executable
- Confirm a signed tx with an lower/higher nonce than Safe nonce – it should not be executable


yagopv:
@JagoFigueroa I'm not fixing:

```
4. Pop up covering the whole screen on smaller resolutions
```

As I think this is ok seeing for example the send transaction popup. I like it indeed ☺️

image

image

@katspaugh can you confirm that the intention for mobile popups is covering the whole screen ?

iamacook (approved):
Nice! Logic reads correctly and I've tested the redirection working.

francovenica:
LGTM

Signed with safes 1/x
safes 2/x
safes 3/x
safes 1/x with queuing first and executing later
Different order nonces
With more signatures than needed: Safe with several queued tx with 2+/x signatures needed, all signed, one of those tx changes the policy to 1/x, after that all tx require 1 signature even tho they have 2 or more. The tx executed just fine

PR #778, opened 2022-09-30
fix(safe-apps):  Add custom app popup visual issues

yagopv:
## What it solves
Resolves https://github.com/safe-global/web-core/issues/697, https://github.com/safe-global/web-core/issues/785

## How this PR fixes it
We are making some adjustments in the UI/UX of the popup


yagopv:
@JagoFigueroa I'm not fixing:

```
4. Pop up covering the whole screen on smaller resolutions
```

As I think this is ok seeing for example the send transaction popup. I like it indeed ☺️

image

image

@katspaugh can you confirm that the intention for mobile popups is covering the whole screen ?

src/components/safe-apps/AddCustomAppModal.tsx
```diff
@@ -73,7 +75,7 @@ const AddCustomAppModal = ({ open, onClose, onSave, safeAppsList }: Props) => {
     })
   }, [chainId, debouncedUrl])
 
-  const appLogoUrl = safeApp?.iconUrl || APP_LOGO_FALLBACK_IMAGE
+  const appLogoUrl = safeApp?.iconUrl || (isDarkMode ? APP_LOGO_FALLBACK_DARK_IMAGE : APP_LOGO_FALLBACK_IMAGE)
```
katspaugh, line 78: Please use the same image for both modes. You need to add special classes from global.css to the SVG elements. See how we did it for all other icons.

yagopv: Was using the img fallback component but changing to the SvgIcon seems to work just fine. done , thanks

src/components/safe-apps/AddCustomAppModal.tsx
```diff
@@ -114,13 +111,20 @@ const AddCustomAppModal = ({ open, onClose, onSave, safeAppsList }: Props) => {
...
+            <SvgIcon component={SafeAppIcon} inheritViewBox sx={{ height: TEXT_FIELD_SIZE, width: TEXT_FIELD_SIZE }} />
+            <TextField
+              label="App name"
+              disabled
+              sx={({ palette }) => ({
+                width: '100%',
+                ml: 2,
+                WebkitTextFillColor: palette.text.secondary,
+                '&& input': {
+                  WebkitTextFillColor: safeApp?.name && palette.text.primary,
```
katspaugh, line 123: What's this for? And why only webkit?

yagopv: hey @katspaugh !! 

This configuration is in order to make the app name title standout a little as the textfield is disabled. So it looks like an enabled one in both themes

image

image


Webkit is because is a non standard property. Don't know if you know a better way to do it

katspaugh: If it's disabled, it should look like one. Otherwise the user might try to edit it, resulting in a bad UX.
Maybe you can use `readOnly` instead of `disabled`?

yagopv: In this concrete case for me it looks better without leaving the default disable colors

image

The problem with readOnly is that the adornment is going up and down when clicking over the field and the colors change on blur. Disable for me fits better

yagopv: Hey @katspaugh we were talking in the daily about this and we agree about bringing @liliiaorlenko into the conversation for make some design change and substitute this input for a label/text/any. It doesn't make sense to have a disabled input there and is more an historical thing

Lila, could you provide us with some guidelines to change The App Name from an input to something fancier 😁 ?

liliiaorlenko: I made a designs with different state for this screen https://www.figma.com/file/rEKs2w2zMdmhNtkgTQDEWD/Add-custom-app?node-id=2%3A2757
- start screen
- app can be added
- error

yagopv: Let's try then! Thanks

Will ping you again when implemented

yagopv: Hey guys, we are going to create a new ticket for the redesign of the popup and send this PR to QA as it fix another things

katspaugh (approved):

JagoFigueroa (approved):

PR #780, opened 2022-09-30
fix: add `shortName` attribute to Beamer user

iamacook:
## What it solves

Allows for Beamer top bar announcements

## How this PR fixes it

A custom `chain` attribute has been added to Beamer users which we can leverage as a filter. Each `chain` has been added to "segment filters" on Beamer, ready for top bar announcements as follows:

image

The colours of Beamer have been updated to match the new theme. The top bar is set to that of `info.main` as it best matches light/dark mode. All colours can be easily adjusted via the Beamer configuration panel, however.

## How to test it

1. Open the Safe NOT on Rinkeby (do so via the URL so that it doesn't default to Rinkeby). Observe no top bar.
2. Switch to Rinkeby and observe the new top bar.

## Screenshots

image

image

image

image


src/components/sidebar/SidebarFooter/index.tsx
```diff
@@ -17,21 +17,23 @@ import DebugToggle from '../DebugToggle'
...
   const cookies = useAppSelector(selectCookies)
+  const chain = useCurrentChain()
 
   const hasBeamerConsent = useCallback(() => cookies[CookieType.UPDATES], [cookies])
 
   useEffect(() => {
     // Initialise Beamer when consent was previously given
-    if (hasBeamerConsent() && !isBeamerLoaded()) {
-      loadBeamer()
+    if (hasBeamerConsent() && chain?.shortName) {
```
iamacook, line 33: The loading flag is checking inside `loadBeamer`.

katspaugh (approved):
Code looks good!

The blue color, though, doesn't work in our design IMHO. Please discuss with @TanyaEfremova

francovenica:
Looks fine, I think the blue color is fine, but I agree we can discuss it with Tanya.

image
image

iamacook:
> Looks fine, I think the blue color is fine, but I agree we can discuss it with Tanya.

Sounds good. I will merge this, however, as the colour is altered via the Beamer admin panel.

PR #784, opened 2022-09-30
Hide the transaction queue in the app view when there is no transactions 

### LINKED ISSUE #753: [Safe apps tx queue] hide queue when there are no transactions on it
## description
The current behaviour for the tx queue is to be always present, even if there are no txs on it. We should change this to keep the same behaviour as in legacy web and only show it when there are txs to be shown or when a user with an empty queue would create a tx from a safe app that would end up in the queue.

mikhailxyz:
## What it solves

Resolves #753 

## How this PR fixes it

By hiding the queue when there are no queued transactions

## How to test it
Open a Safe App with a Safe that doesn't have queued transactions


src/components/safe-apps/AppFrame/TransactionQueueBar/index.tsx
```diff
@@ -22,7 +22,7 @@ const TransactionQueueBar = ({ expanded, visible, setExpanded, onDismiss }: Prop
   const { page = { results: [] } } = useTxQueue()
   const queuedTxCount = getQueuedTransactionCount(page)
 
-  if (!visible) {
+  if (!visible || queuedTxCount === '0') {
```
iamacook, line 25: I would suggest we compare against the `results?.length` instead. We could change the return of `getQueuedTransactionCount` (to i.e. "No transactions found") and this would break.

mikhailxyz: Thanks ser, adjusted it now

iamacook (approved):

JagoFigueroa:
Something is off for me bandido :(

When the queue goes away because there is no longer a tx on it will leave this behind:

Screenshot 2022-10-04 at 14 38 42

JagoFigueroa:
All good after the latest changes. Thank you señor!

PR #787, opened 2022-09-30
Chore: @typescript-eslint + type imports

katspaugh:
## What it solves
Adds the typescript eslint plugin and a rule for it to automatically import types with the `type` keyword.

See https://github.com/safe-global/safe-react-gateway-sdk/pull/91 for context.


src/components/address-book/ExportDialog/index.tsx
```diff
@@ -4,7 +4,8 @@ import Button from '@mui/material/Button'
 import Link from '@mui/material/Link'
 import Typography from '@mui/material/Typography'
 import { useCSVDownloader } from 'react-papaparse'
-import { SyntheticEvent, useMemo, type ReactElement } from 'react'
+import type { SyntheticEvent } from 'react'
+import { useMemo, type ReactElement } from 'react'
```
iamacook, line 8: Doesn't seem to have worked here.

src/components/common/CopyButton/index.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { type ReactElement, ReactNode, type SyntheticEvent, useCallback, useState } from 'react'
+import type { ReactNode } from 'react'
+import React, { type ReactElement, type SyntheticEvent, useCallback, useState } from 'react'
```
iamacook, line 2: Here either.

iamacook (approved):
I'll approve it as it was fixed by the linter but there are multiple places where it didn't find "nested" imported types for some reason.

src/components/common/CopyButton/index.tsx
iamacook: I won't mark the others.

katspaugh: Looks like it extracts them into their own import. Can't say I like it.

Issue #694, opened 2022-09-27
[Share app, web core] Safe field not auto-populated


JagoFigueroa:
## Description
When a EOA address with created safes receives the shared app link, it is not auto-populated as in the previous iteration where the current one will be automatically selected for the user.

## Steps to reproduce
While in the web core and conected to a safe on Rinkeby, open this [magic link](https://web-core.pages.dev/share/safe-app?appUrl=https%3A%2F%2Ftokentool.bitbond.com%2F&chain=rin)

## Expected result
Current connected safe should be detected and be auto selected for the user.

## Obtained result
No safe is selected for the user. A video with a quick comparison between revamp (left) and legacy (right):

https://user-images.githubusercontent.com/92332710/192525971-8957bbc6-e767-4537-80bc-d4cd0b397f50.mp4

JagoFigueroa:
All bueno with this one, thanks! Checking the other ticket fixed in the pr :pepe-naruto:

PR #791, opened 2022-10-03
fix: fixed Safe Selector and redirection in the Safe Apps

### LINKED ISSUE #696: [Share app] redirection not performed to the current chain on create safe flow
## Description
Trying to create a safe from the share apps flow is not performing the redirection correctly for me.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Safe: None, trying to create one
 - Environment: web-core prod

## Steps to reproduce
 1. Via this [link](https://web-core.pages.dev/share/safe-app?appUrl=https%3A%2F%2Fapps.gnosis-safe.io%2Fdrain-safe&chain=matic) (for polygon, feel free to change the chainId) connect with a EOA that has no safes on the current chain plus any saved safe on the local storage for any other.
 2. Try to create a safe.

## Expected result
Redirection performed to the current chain and process can be continued. Switching to the current one ignores the flow of safe creation from the share link.

## Obtained result
Redirection is performed to [mainnet](https://web-core.pages.dev/open?safeViewRedirectURL=%2Fapps%3FappUrl%3Dhttps%3A%2F%2Fapps.gnosis-safe.io%2Fdrain-safe)
Screenshot 2022-09-27 at 13 17 39

DaniSomoza:
## What it solves

Resolves #696 #694

### DIFF
--- a/src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
+++ b/src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
@@ -1,38 +1,59 @@
 import { Box, Button, MenuItem, Select, Typography, Grid, FormControl, InputLabel } from '@mui/material'
-import type { ChainInfo } from '@gnosis.pm/safe-react-gateway-sdk'
-import { useState } from 'react'
+import type { ChainInfo, SafeAppData } from '@gnosis.pm/safe-react-gateway-sdk'
+import { useEffect, useMemo, useState } from 'react'
 import type { ConnectedWallet } from '@/hooks/wallets/useOnboard'
 import { useAppSelector } from '@/store'
-import { selectAddressBookByChain } from '@/store/addressBookSlice'
+import { selectAllAddressBooks } from '@/store/addressBookSlice'
+import { selectChains } from '@/store/chainsSlice'
 import useLastSafe from '@/hooks/useLastSafe'
 import { parsePrefixedAddress } from '@/utils/addresses'
 import SafeIcon from '@/components/common/SafeIcon'
 import EthHashInfo from '@/components/common/EthHashInfo'
 import { AppRoutes } from '@/config/routes'
+import useOwnedSafes from '@/hooks/useOwnedSafes'
 import { CTA_BUTTON_WIDTH, CTA_HEIGHT } from '@/components/safe-apps/SafeAppLandingPage/constants'
 
 type Props = {
   appUrl: string
   wallet: ConnectedWallet | null
   onConnectWallet: () => Promise<void>
-  safes: string[]
   chain: ChainInfo
+  app: SafeAppData
 }
 
-const AppActions = ({ wallet, onConnectWallet, safes, chain, appUrl }: Props): React.ReactElement => {
+type CompatibleSafesType = { address: string; chainId: string; shortName?: string }
+
+const AppActions = ({ wallet, onConnectWallet, chain, appUrl, app }: Props): React.ReactElement => {
   const lastUsedSafe = useLastSafe()
-  const lastUsedSafeAddress =
-    lastUsedSafe && safes.includes(lastUsedSafe) ? parsePrefixedAddress(lastUsedSafe).address : ''
-  const [safeToUse, setSafeToUse] = useState<string | undefined>(lastUsedSafeAddress || undefined)
-  const addressBook = useAppSelector((state) => selectAddressBookByChain(state, chain.chainId))
+  const ownedSafes = useOwnedSafes()
+  const addressBook = useAppSelector(selectAllAddressBooks)
+  const chains = useAppSelector(selectChains)
+  const compatibleChains = app.chainIds
+
+  const compatibleSafes = useMemo(
+    () => getCompatibleSafes(ownedSafes, compatibleChains, chains.data),
+    [ownedSafes, compatibleChains, chains.data],
+  )
+
+  const [safeToUse, setSafeToUse] = useState<CompatibleSafesType>()
+
+  useEffect(() => {
+    const selectedSafe = getDefaultSafe(compatibleSafes, chain.chainId, lastUsedSafe)
+    if (selectedSafe) {
+      setSafeToUse(selectedSafe)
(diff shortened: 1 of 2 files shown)


src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
```diff
@@ -1,38 +1,59 @@
...
   const hasWallet = !!wallet
-  const hasSafes = safes.length > 0
+  const hasSafes = compatibleSafes.length > 0
   const shouldCreateSafe = hasWallet && !hasSafes
 
   let button: React.ReactNode
   switch (true) {
     case hasWallet && hasSafes:
-      const href = `${AppRoutes.apps}?appUrl=${encodeURIComponent(appUrl)}&safe=${chain.shortName}:${safeToUse}`
+      const href = `${AppRoutes.apps}?appUrl=${encodeURIComponent(appUrl)}&safe=${safeToUse?.shortName}:${
```
katspaugh, line 54: We decided to avoid string URLs and use URL Objects as much as possible. Could you please convert this one?

src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
```diff
@@ -111,3 +135,45 @@ const AppActions = ({ wallet, onConnectWallet, safes, chain, appUrl }: Props): R
...
+
+  // as a second option, we use any user Safe in the provided chain
+  const safeInTheSameChain = compatibleSafes.find((safe) => safe.chainId === chainId)
+
+  if (safeInTheSameChain) {
+    return safeInTheSameChain
+  }
+
+  // as a fallback we salect a random compatible user Safe
+  return compatibleSafes[0]
```
mikhailxyz, line 178: Why do we fall back to arbitrary choices? Looking at the original [epic](https://github.com/safe-global/safe-pm/issues/60) it clearly defines what should be the pre-selected value:
> User with at least one Safe defined in this Chain
If the user has defined at least a safe in the safe app chain we display:

> A selector with all the safes compatibles with the safe app (**the last safe used as a default selected option** if it is present).
A button to open the safe app with a safe demo.

src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
```diff
@@ -1,38 +1,59 @@
...
+
+  const compatibleSafes = useMemo(
+    () => getCompatibleSafes(ownedSafes, compatibleChains, chains.data),
+    [ownedSafes, compatibleChains, chains.data],
+  )
+
+  const [safeToUse, setSafeToUse] = useState<CompatibleSafesType>()
+
+  useEffect(() => {
+    const selectedSafe = getDefaultSafe(compatibleSafes, chain.chainId, lastUsedSafe)
```
mikhailxyz, line 41: if the function is named `getDefaultSafe`, would it be better to name the variable `defaultSafe`?

src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
DaniSomoza: It seems that [`new URL(..)`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) requires an absolute url. But in our case we have a relative one:

Captura de pantalla 2022-10-04 a las 15 47 04


For internal redirection we are using string URLs:

Captura de pantalla 2022-10-04 a las 15 42 49
Captura de pantalla 2022-10-04 a las 15 43 25

src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
mikhailxyz: Also the current ticket says that the expected outcome should be the last used Safe:
>Expected result
Last connected safe should be detected and be auto selected for the user.

src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
katspaugh: It takes simple objects, not necessarily URLObject objects, like so:
```
<Link href={{
  path: '/welcome',
  { query: { chain: 'rin' } }
}} />
```

katspaugh: See https://github.com/safe-global/web-core/pull/783

src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
DaniSomoza: I think that we can keep the same behavior that we have in the legacy codebase to facilitate QA work.

In the future we can create an improvement to change this.

src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
DaniSomoza: done!

src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
mikhailxyz: Could you please document this in the ticket for future reference? So if someone revisits this later they can track down the change. At the moment, all tickets say that it should select the last used safe.

mikhailxyz: Also, I'm unsure if including additional behaviour facilitates QA's work. At first glance, I'd say the QA has more cases to test versus a single case with the last used safe

DaniSomoza: I updated[ the epic](https://github.com/safe-global/safe-pm/issues/60) for future reference:

Captura de pantalla 2022-10-06 a las 10 38 24

mikhailxyz (approved):

src/components/safe-apps/SafeAppLandingPage/AppActions.tsx
mikhailxyz: gracias señor

PR #795, opened 2022-10-03
fix: fixed small visual issues in our Safe Apps Cards (logo + descriptions)

### LINKED ISSUE #779: Some app icons not fitting entirely on safe apps screen
## Bug description
We have some small visual issues with some of our safe apps where their icon does not fit completely on the safe apps list.

This is also present on the safe apps widget in the home screen.

## Steps to reproduce
Go to safe-apps main screen

## Expected result
All icons fitting completely.

## Obtained result
Some examples:

safe apps list
Screenshot 2022-09-30 at 12 13 08
Screenshot 2022-09-30 at 12 13 22
Screenshot 2022-09-30 at 12 14 15

safe apps widget in /home
Screenshot 2022-09-30 at 12 23 35

DaniSomoza:
## What it solves

Resolves #779

## How this PR fixes it

Added new styles to fix the small issues in our Safe Apps Cards (Logo + Description)

```

.truncateDescription {
  display: -webkit-box;
  -webkit-line-clamp: 3;
  -webkit-box-orient: vertical;
  overflow: hidden;
}
```

## Screenshots

Captura de pantalla 2022-10-03 a las 18 33 04
Captura de pantalla 2022-10-03 a las 18 33 29

### DIFF
--- a/src/components/safe-apps/AppCard/index.tsx
+++ b/src/components/safe-apps/AppCard/index.tsx
@@ -87,7 +87,7 @@ const CompactAppCard = ({ url, safeApp, onPin, pinned, shareUrl }: CompactSafeAp
   <AppCardContainer url={url} variant="compact">
     <div className={styles.compactCardContainer}>
       {/* App logo */}
-      <img src={safeApp.iconUrl} alt={`${safeApp.name} logo`} className={styles.safeAppLogo} />
+      <img src={safeApp.iconUrl} alt={`${safeApp.name} logo`} className={styles.safeAppLogoCompact} />
 
       {/* Share button */}
       <CopyButton
@@ -134,7 +134,7 @@ const AppCard = ({ safeApp, pinned, onPin, onDelete, variant = 'default' }: AppC
   return (
     <AppCardContainer url={url}>
       <CardHeader
-        avatar={<img src={safeApp.iconUrl} alt={`${safeApp.name} logo`} className={styles.safeAppLogo} />}
+        avatar={<img src={safeApp.iconUrl} alt={`${safeApp.name} logo`} className={styles.safeAppLogoDefault} />}
         action={
           <>
             {/* Share button */}
--- a/src/components/safe-apps/AppCard/index.tsx
+++ b/src/components/safe-apps/AppCard/index.tsx
@@ -1,7 +1,6 @@
 import type { ReactElement, ReactNode, SyntheticEvent } from 'react'
 import { useRouter } from 'next/router'
 import Link from 'next/link'
-import Avatar from '@mui/material/Avatar'
 import Card from '@mui/material/Card'
 import CardContent from '@mui/material/CardContent'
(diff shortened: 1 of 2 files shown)


src/components/safe-apps/AppCard/index.tsx
```diff
@@ -88,12 +87,7 @@ const CompactAppCard = ({ url, safeApp, onPin, pinned, shareUrl }: CompactSafeAp
   <AppCardContainer url={url} variant="compact">
     <div className={styles.compactCardContainer}>
       {/* App logo */}
-      <Avatar
-        src={safeApp.iconUrl}
-        alt={`${safeApp.name} logo`}
-        variant="square"
-        sx={{ objectFit: 'contain', width: 52, height: 52 }}
-      />
+      <img src={safeApp.iconUrl} alt={`${safeApp.name} logo`} className={styles.safeAppLogoCompact} />
```
katspaugh, line 90: This won't have a fallback when the iconUrl is missing, like Avatar does.

src/components/safe-apps/AppCard/styles.module.css
```diff
@@ -12,3 +12,22 @@
   top: 2px;
   left: 2px;
 }
+
+.truncateDescription {
+  display: -webkit-box;
+  -webkit-line-clamp: 3;
+  -webkit-box-orient: vertical;
```
katspaugh, line 19: Please don't use browser-specific hacks. It should work in all browsers.

katspaugh (changes requested):

src/components/safe-apps/AppCard/styles.module.css
DaniSomoza: We have the same styles [in our old codebase](https://github.com/safe-global/safe-react/blob/dev/src/routes/safe/components/Apps/components/SafeAppCard/SafeAppCard.tsx#L217) for this and it seems that it has a good browser support (`97.47%`):

```
https://caniuse.com/?search=-webkit-line-clamp
```

What is your opinion about it?

katspaugh: Prefixed CSS rules are by definition browser-specific/temporary.
Did you test in Firefox?
It might work (for now), but we should in general avoid them and stick to the standard CSS spec.

katspaugh (approved):

src/components/safe-apps/AppCard/styles.module.css
DaniSomoza: it seems that[ only the prefixed css rule is available](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp).

For me it's working fine in Chrome, Firefox and Safari.

```
https://css-tricks.com/line-clampin/#aa-the-standardized-way
```

JagoFigueroa:
Espectacular trabajo, thanks!

PR #803, opened 2022-10-04
fix: display SignOrExecuteForm on opening Change threshold modal

### LINKED ISSUE #751: [Settings]Submit button is hidden on change policy review screen when  selected policy is the same as the current one
## Bug description
Submit button is hidden on the change policy review screen when the selected policy is the same as the current one
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby

## Steps to reproduce
 1. Go to the setting ->set up
 2. click change policy
 Current result: The "submit" button is hidden till the diff from the current policy is selected
image

## Expected result
The submit button should be disabled instead of hiding

DiogoSoaress:
## What it solves

Resolves #751

## How this PR fixes it
The `<ChangeThresholdDialog />` renders `<SignOrExecuteForm />` unconditionally and passes `undefined` safeTx to disable the "Submit" button.

## How to test it
1. Open the Safe settings and click to change threshold.
2. Submit button should be visible and disabled until the selected threshold differs from the current one.

## Analytics changes
None

## Screenshots
Screenshot 2022-10-04 at 15 55 21


src/components/settings/owner/ChangeThresholdDialog/index.tsx
```diff
@@ -90,27 +91,24 @@ const ChangeThresholdStep = ({ data, onSubmit }: { data: ChangeThresholdData; on
...
-            isExecutable={safe.threshold === 1}
-            onSubmit={onChangeTheshold}
-            error={safeTxError}
-          />
-        </Box>
-      )}
+      <Box mt={-5}>
+        <SignOrExecuteForm
+          safeTx={safeTx}
+          isInitiallySubmittable={selectedThreshold !== data.threshold}
```
katspaugh, line 106: Can't we leave it as submittable but exit or throw on submit if the form is invalid?

DiogoSoaress:
> Can't we leave it as submittable but exit or throw on submit if the form is invalid?
It's the only modal where this prop would be used and I'd very much like to avoid adding more props to SignOrExecute.

I agree to avoid passing the extra prop. On the latest commit I am passing an `error` and making the transaction "unsubmittable".

I would like to take a step back tho and would reconsider if this is really a behaviour we want to fix (displaying the submit button).
Alternatively I think we could provide an improved message to the user explaining why the "Submit" button is not visible in first place.

Screenshot 2022-10-05 at 11 45 48

_OR_

Screenshot 2022-10-05 at 11 44 23

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -201,11 +203,13 @@ const SignOrExecuteForm = ({
           <ErrorMessage error={submitError}>Error submitting the transaction. Please try again.</ErrorMessage>
         )}
 
-        <Typography variant="body2" color="border.main" textAlign="center" mt={3}>
-          You&apos;re about to {txId ? '' : 'create and '}
-          {willExecute ? 'execute' : 'sign'} a transaction and will need to confirm it with your currently connected
-          wallet.
-        </Typography>
+        {safeTx && (
```
katspaugh, line 206: This will make it jump as `safeTx` is async.

katspaugh: I would just leave it even if the submit button is disabled.

DiogoSoaress: I would leave the `<AdvancedParams />` conditional however. We also have couple lines above the 
```
{tx && <DecodedTx tx={tx} txId={txId} />}
```
Wdyt? Otherwise if I render the `<AdvancedParams />` without the nonce, the Skeleton will be a bit misleading
Screenshot 2022-10-05 at 14 00 21

katspaugh: It will make all other modals jump

katspaugh (approved):

katspaugh:
Fixed, thanks @francovenica!

PR #805, opened 2022-10-04
Chore: deploy HTML files w/o extension

katspaugh:
## What it solves
Copies HTML files to S3 w/o an extension but with the right content-type, so that URLs like `/welcome` work along with `/welcome.html`.

Also I moved the 404 redirect code to the 404 page itself, I think it makes more sense than a common hook.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,8 +30,7 @@
     "serve": "npx -y http-server out"
   },
   "pre-commit": [
-    "lint",
-    "prettier"
+    "lint"
   ],
   "dependencies": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -30,8 +30,7 @@
     "serve": "npx -y http-server out"
   },
   "pre-commit": [
-    "lint",
-    "prettier"
```
usame-algan, line 34: Does `prettier` not work in the pre-commit hook?

usame-algan (approved):

PR #807, opened 2022-10-05
Fix: CSS-only dark mode

### LINKED ISSUE #804: Light theme flashes before the dark mode loads
## Bug description
We add `data-theme="dark"` via JS, based on the dark mode setting, and the CSS switches depending on this attribute.
And although we by default follow the OS setting, it has to go thru JS before it's applied.

The white background flashes briefly when the page loads and hurts the eyes.

## Expected result
I suggest we use a dark mode media query for CSS variables, and the `data-theme` should just override it.

So the logic should be the following:
* Use dark mode CSS vars if OS is in dark mode
* Override that if `data-theme="light"`

katspaugh:
## What it solves

Resolves #804 

## How this PR fixes it
Makes the dark CSS vars kick in before the theme is initialized.
However, if the light mode is forced via the user toggle, they'll be overridden.

## How to test it
Switch the OS to dark mode and load/reload the site. It should not flash the light theme.
Switch the dark mode toggle to light and try the same.


src/pages/_app.tsx
```diff
@@ -83,18 +83,28 @@ const WebCoreApp = ({ Component, pageProps, emotionCache = clientSideEmotionCach
...
 
+        {/* Mobile tags */}
         <meta name="viewport" content="width=device-width" />
         <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent" />
-        <meta httpEquiv="Content-Security-Policy" content={ContentSecurityPolicy} />
-        {IS_PRODUCTION && <meta httpEquiv="Strict-Transport-Security" content={StrictTransportSecurity} />}
+
+        {/* PWA primary color and manifest */}
+        <meta name="theme-color" content="#ffffff" media="(prefers-color-scheme: light)" />
+        <meta name="theme-color" content="#121312" media="(prefers-color-scheme: dark)" />
```
iamacook, line 100: I'd suggest we import these colours.

iamacook (approved):

PR #808, opened 2022-10-05
Fix: add TallyHo

### LINKED ISSUE #703: Add Tally Ho as a wallet option

katspaugh:
## What it solves

Resolves #703

## How this PR fixes it
Adds the TallyHo module.

## Screenshots
Screenshot 2022-10-05 at 09 11 55


src/config/constants.ts
```diff
@@ -15,10 +16,10 @@ export const BEAMER_ID = process.env.NEXT_PUBLIC_BEAMER_ID || ''
 
 // Wallets
 export const WC_BRIDGE = process.env.NEXT_PUBLIC_WC_BRIDGE || 'https://safe-walletconnect.gnosis.io/'
-export const FORTMATIC_KEY = process.env.NEXT_PUBLIC_FORTMATIC_KEY || ''
-export const PORTIS_KEY = process.env.NEXT_PUBLIC_PORTIS_KEY || ''
```
iamacook, line 19: We should check whether we're paying for these at some point as well. We should cancel them if so.

src/config/constants.ts
```diff
@@ -28,10 +29,10 @@ export const SAFE_TOKEN_ADDRESSES: { [chainId: string]: string } = {
 }
 
 // Safe Apps
-export const SAFE_APPS_POLLING_INTERVAL = process.env.NODE_ENV === 'test' ? 4500 : 15000
+export const SAFE_APPS_INFURA_TOKEN = process.env.NEXT_PUBLIC_SAFE_APPS_INFURA_TOKEN || INFURA_TOKEN
+export const SAFE_APPS_POLLING_INTERVAL = 15_000
```
iamacook, line 33: Why remove the period when polling?

src/config/constants.ts
```diff
@@ -15,10 +16,10 @@ export const BEAMER_ID = process.env.NEXT_PUBLIC_BEAMER_ID || ''
 
 // Wallets
 export const WC_BRIDGE = process.env.NEXT_PUBLIC_WC_BRIDGE || 'https://safe-walletconnect.gnosis.io/'
-export const FORTMATIC_KEY = process.env.NEXT_PUBLIC_FORTMATIC_KEY || ''
-export const PORTIS_KEY = process.env.NEXT_PUBLIC_PORTIS_KEY || ''
-export const TREZOR_APP_URL = 'gnosis-safe.io'
-export const TREZOR_EMAIL = 'safe@gnosis.io'
+export const TREZOR_APP_URL = 'app.safe.global'
```
iamacook, line 19: Have you tested that this works on staging? Maybe it should be dynamic.

src/config/constants.ts
katspaugh: It's still there, I just moved the infura var up.

src/config/constants.ts
katspaugh: It's just for links I believe. As you can see `gnosis-safe.io` has been working fine all along.

src/config/constants.ts
katspaugh: We're definitely not paying for them.

src/config/constants.ts
iamacook: `SAFE_APPS_POLLING_INTERVAL` was dependent on the `NODE_ENV`.

katspaugh: We don't have that set anywhere, and it seems it didn't affect anything.

iamacook: It's used in `useAppIsLoading` for tests but I see that they're still working.

iamacook:
I'll hold off on approving this until we head back from product [here](https://5afe.slack.com/archives/C03DAGWJCR5/p1664952542007989).

src/config/constants.ts
katspaugh: I meant that NODE_ENV isn't set anywhere.

iamacook: Jest [sets it to `"test"` by default](https://jestjs.io/docs/environment-variables#node_env).

yagopv: This came from safe-react. Was renamed here (https://github.com/safe-global/safe-react/commit/918d68eb01b918f2994cb26bdc3de912d414560c). I think we can remove the process.env check without problems as there is no specific test for apps not loading.

yagopv: We can rename the constant as well as the name perhaps does not make sense

katspaugh: Should I rename it to `SAFE_APPS_LOAD_TIMEOUT`?

yagopv: For me fine if you want to do it. If not we can do it in one of our PRs

katspaugh: I'll leave it as is for now then.

iamacook (approved):

PR #809, opened 2022-10-05
Fix: scroll body when modal is too big

### LINKED ISSUE #744: Two scrolls when the modal window is scrollable
## Bug description
Two scrolls when the modal window is scrollable
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby

## Steps to reproduce
 1. Open the review screen in a replace owner modal window in the safe with many owners
 2. scroll the modal window context 

Current result: https://drive.google.com/file/d/1EECYYXXqVmiRRGPOVlVBezQP9jKoGV4J/view?usp=sharing
## Expected result
1. maybe it makes sense to scroll inside the modal, so the header and the bottom part with buttons of the modal window are always visible?
2.  can the scroll of the main screen be disabled when the modal window is opened?

katspaugh:
## What it solves

Resolves #744

## How this PR fixes it
Adds scoll=body to the MUI Dialog.

## How to test it
Open a Safe with many owners and try to add a new one.
The whole page should now scroll.

### DIFF
--- a/src/components/common/ModalDialog/index.tsx
+++ b/src/components/common/ModalDialog/index.tsx
@@ -70,7 +70,7 @@ const ModalDialog = ({
   const fullScreen = useMediaQuery(theme.breakpoints.down('sm'))
 
   return (
-    <StyledDialog {...restProps} fullScreen={fullScreen} onClick={(e) => e.stopPropagation()}>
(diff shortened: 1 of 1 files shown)


iamacook (approved):

usame-algan (approved):
This is much nicer than an inner scroll in the modal!

PR #818, opened 2022-10-05
style: address rebranding feedback

iamacook:
## What it solves

Resolves [rebranding feedback](https://www.notion.so/gnosis-safe/Rebranding-Feedback-d5561d3c88fc456cb15e93f867e0c971)

#### General
- [x] No `box-shadow` when hovering over buttons

image
image

- [x] Borders of tables

image
image

- [x] Border widths of 1px (check transaction details)

image
image

- [x] Dark mode reactive pairing icon

image
image

- [x] Switch `box-shadow`

image
image

- [x] `EthHashInfo` copy/share icon to use ours

image
image

- [x] Backdrop adjustment

image
image

#### Dashboard
- [x] View assets alignment

image
image

- [x] Hover color of "Connect & transact"/"Use WalletConnect" titles (and all other links)

image
image

- [x] Pending transaction list hover background color

image
image

#### Transactions
- [x] Transaction summary execution rocket

image
image

#### Apps
- [x] Open/close chevron icons colours match section text
- [x] Section text is bold with `letter-spacing`

image
image

#### Notes

- Sticky modal buttons will be tackled in another issue.
- Share app icon too small on dashboard has been asked about.

Apps no result page has been removed per @katspaugh's suggestion of requiring a new tag.
image
image


src/styles/theme.ts
```diff
@@ -46,8 +46,6 @@ declare module '@mui/material/Button' {
   }
 }
 
-export const theme = createTheme()
```
iamacook, line 49: We should always import the theme via the `useTheme` hook.

src/components/safe-apps/SafeAppList.tsx
```diff
@@ -28,6 +36,24 @@ const SafeAppList = () => {
...
+    }
+
+    return {
+      pathname: AppRoutes.apps,
+      query: {
+        safe: router.query.safe,
+        appUrl: wc.url,
+      },
+    }
+  }, [allSafeApps, router.query.safe])
```
katspaugh, line 55: What's this for?
Even if this is really necessary, we shouldn't search apps by name.

iamacook: I was cautious of hardcoding it because of the `id` changing across chains but it seems like it's the same (and across environments because we don't have multiple). I've channged it accordingly.

src/components/safe-apps/SafeAppList.tsx
```diff
@@ -28,6 +36,24 @@ const SafeAppList = () => {
 
   const debouncedSearchQuery = useDebounce(searchQuery, 500)
 
+  const wcLink: LinkProps['href'] = useMemo(() => {
+    const WC_APP_ID = 11
```
katspaugh, line 40: First of all, the id is different on staging and prod. Secondly, hardcoding it here is no bueno.

katspaugh: I would add a new tag, e.g. `search-fallback` or `walletconnect` and use that.
Let's not add for now, and leave the search how it was.

katspaugh (approved):

katspaugh:
Tested, everything looks good! ✅ 

The only problem I noticed is a double border between nested transactions:
Screenshot 2022-10-07 at 08 48 30

iamacook:
I will merge this as it is stylistic-based and has been verified.

PR #819, opened 2022-10-05
Fix: tooltip colors

katspaugh:
## What it solves

A small fix for the tooltip colors that I accidentally broke in #807.


iamacook (commented):
Looks good but the arrow is not styled correctly:

image
image

katspaugh:
@iamacook good catch, fixed 👍

iamacook (approved):

PR #820, opened 2022-10-05
test: fix e2e test for tx simulation

### LINKED ISSUE #817: [Simulation] simulation result should reset when changing the gas limit
## Bug description

When the gas limit gets edited after simulating a transaction the simulation result should reset and show the "Simulate" Button again.
This bug was discovered by our simulation E2E tests.

## Steps to reproduce

1. Create a executable transaction. 
2. Simulate it. 
3. Now change the gas limit to something too low (21000)
4. The simulation result should now disappear and show the "simulate" button again.

schmanu:
- fix selectors for web-core
- fix simulation issue after changing the gas limit
- fix E2E Test wallet injection

## What it solves
It changes a issue where simulations do not reset when the gas limit parameter is changed.

Resolves #817 

## How to test it
The e2e tests:
1. set IS_PRODUCTION to true using .env variables
2. Use the correct CYPRESS_MNEMONIC
3. run the tx simulation e2e tests

The simulation issue:
1. Simulate a executable tx
4. change the gas limit to something else
5. the simulation result gets closed automatically and shows the simulate button again

## Analytics changes
none

### DIFF
--- a/cypress/e2e/tx_simulation.cy.js
+++ b/cypress/e2e/tx_simulation.cy.js
@@ -6,24 +6,24 @@ describe('Tx Simulation', () => {
     cy.connectE2EWallet()
 
     // Open the Safe used for testing
-    cy.visit(`/${TEST_SAFE}`)
-    cy.contains('a', 'Accept selection').click()
+    cy.visit(`/${TEST_SAFE}/home`, { failOnStatusCode: false })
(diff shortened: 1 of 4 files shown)


iamacook (commented):
Looks good to me but it fails as it is trying to open a Rinkeby Safe:

image

PR #823, opened 2022-10-05
fix: validate owners in parallel

### LINKED ISSUE #789: [Safe creation] Improve error handling for safe owner configuration
## Bug description
The error handling seen on screen when trying to create a safe with repeated owners or invalid addresses needs a little bit of tweaking as after some adding / removing of safe owner configurations that should be valid are stuck with an error on ui.

## Steps to reproduce
Play around on safe creation by adding the same owner address a couple of times and then remove one of the 2 (see video for more examples of this).

## Expected result
UI refreshes as expected when adding / removing owners.

## Obtained result
UI gets a little bit stuck as seen on this video:

https://user-images.githubusercontent.com/92332710/193516532-da39643a-1a72-4bca-8b77-67db422c3f63.mp4

iamacook:
## What it solves

Resolves #789

## How this PR fixes it

The `owners` group are now dependencies of each other. When one is validated, the others are as well.

## How to test it

1. Open Safe creation and navigate to the owner step.
2. Add the same owner twice and observe the error on the second field.
3. Edit the first owner entry and observe the error on the second field disappear.

## Screenshots

owner validation


katspaugh (approved):
LGTM but there's lint

francovenica:
The validation is not being done when QR codes are involved. 

Have row 1 and row 2 with owner 1 and owner 2
Use the QR code in row 2 to set the address owner 1. Now is duplicated and an error is shown in row 2 (this is correct)
Now in row 1 use the QR code to set a different address

Expected result:
Now that both of them are different no error should be shown

Obtained result:
Row 2 still shows the "duplicated owner" error.
safe owners

iamacook:
> The validation is not being done when QR codes are involved.

I've adjusted the validation to trigger after a small timeout. RHF seemingly returns a cached error object. As we are reworking the Safe creation process in the forthcoming roadmap, we should reassess how this is done then. cc @usame-algan

francovenica:
I still have the same error reported

Steps:
OwnerA in row 1 
Add a new row2 and copy OwnerA, the error shows up
in row1 use the QR code feature to change to another address different than OwnerA

Result: The row 2 still shows that the address is duplicated

duplicated owner

katspaugh:
I suggest we ignore this as an edge case. It's due to a bug in RHF, and it will be eventually fixed on their side.

francovenica:
Agree, we can skip it for now.

PR #824, opened 2022-10-05
fix: remove Address Book entry for edited address

### LINKED ISSUE #790: [Addressbook] Editing name + address creating a new entry
## Bug description
Editing the name & address of an entry will create a new one instead of replacing it.

## Steps to reproduce
1. Go to the addressbook
2. Replace the name and address of an existing entry

## Expected result
On legacy I see that it is not possible to edit the address, only the name. 

I would keep that behaviour or at least don't create a new one when editing it fully.

## Obtained result

https://user-images.githubusercontent.com/92332710/193535401-c78ae2c3-ac05-45cc-b644-e824b6414b87.mp4

DiogoSoaress:
## What it solves

Resolves #790

## How this PR fixes it
When editing a AB entry and the address is modified, remove the clicked entry.

## How to test it
1. On AB click to edit an entry.
2. Modify the name and the address.
3. Save
4. The clicked entry should be removed
5. The new entry should be in the AB

## Analytics changes
None

### DIFF
--- a/src/components/address-book/EntryDialog/index.tsx
+++ b/src/components/address-book/EntryDialog/index.tsx
@@ -10,7 +10,7 @@ import ModalDialog from '@/components/common/ModalDialog'
 import NameInput from '@/components/common/NameInput'
 import useChainId from '@/hooks/useChainId'
 import { useAppDispatch } from '@/store'
-import { upsertAddressBookEntry } from '@/store/addressBookSlice'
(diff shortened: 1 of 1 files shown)


iamacook (commented):
I think disabling the field would make more sense instead of removing it and readding it. Wdyt?

DiogoSoaress:
So only editing the name?

iamacook (approved):

PR #828, opened 2022-10-05
fix: tx details header

### LINKED ISSUE #813: [Transaction details] Wrong header for the tx detailspage
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: rinkeby prod

## Steps to reproduce
 1. Go to tx details

Current result: 
1. There is no information in the header that you are located on the tx details page now and the header is too big for the small screens
image

2. information about section where the tx is located is missed

## Expected result
The design was added for the tx details page - https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=1109%3A47999

schmanu:
## What it solves
The header of tx details was taking too much space and displaying a wrong label.

Resolves #813 

## How this PR fixes it
- Displays a simple label for transaction details.
- Adds a hint which tx needs to be executed before the displayed transaction to explain why the execute button is disabled.

## How to test it
- Queue multiple transactions.
- Click share button on one of the transactions with high nonce and open the copied link
- Witness that a hint is displayed that another tx needs to be executed first.

## Analytics changes
None

## Screenshots
localhost_3000_gor_0x954cD69f0E902439f99156e3eeDA080752c08401_transactions_tx_id=multisig_0x954cD69f0E902439f99156e3eeDA080752c08401_0x9a5eb65ec51701fe1178a8ff40d43ad9deed0b7cf3c3d6633e9b804edf896469


src/components/transactions/SingleTx/index.tsx
```diff
@@ -1,59 +1,48 @@
...
+
+  const { safe } = useSafeInfo()
 
-  const error = !transactionId
-    ? new Error("Couldn't retrieve the transaction details. Please review the URL.")
-    : !isCurrentSafeTx
-    ? new Error(`Transaction with id ${transactionId} not found in this Safe`)
-    : txDetailsError
+  let nonceWarning: string | undefined = undefined
+  if (txDetails?.detailedExecutionInfo?.type === DetailedExecutionInfoType.MULTISIG) {
```
katspaugh, line 29: I think there's a guard function for this.

src/components/transactions/SingleTx/index.tsx
```diff
@@ -1,59 +1,48 @@
...
+    if (txDetails.detailedExecutionInfo.nonce > safe.nonce) {
+      nonceWarning = `Transaction with nonce ${safe.nonce} needs to be executed first`
+    }
+  }
```
katspaugh, line 33: There's very similar logic in `<GroupLabel>`, maybe we can extract it into a hook from there?

src/components/transactions/SingleTx/index.tsx
```diff
@@ -1,59 +1,50 @@
...
 
   if (error) {
     return <ErrorMessage error={error}>Failed to load transaction</ErrorMessage>
   }
 
   if (txDetails) {
-    return <SingleTxGrid txDetails={txDetails} />
+    return (
+      <Box display="flex" flexDirection="column" gap={1}>
+        {nonceWarning && <GroupLabelTypography label={nonceWarning} />}
```
iamacook, line 44: Maybe we should show the `DateLabel` if there is no nonce warning, wdyt?

src/hooks/useTxDetails.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @returns TransactionDetails for transaction id in url
+ */
+export const useTxDetails = (): [TransactionDetails | undefined, Error | undefined, boolean] => {
+  const chainId = useChainId()
+  const router = useRouter()
+  const { id } = router.query
+  const transactionId = Array.isArray(id) ? id[0] : id
+  const { safe, safeAddress } = useSafeInfo()
```
iamacook, line 17: I'd suggest we make this hook more reusable and pass these as arguments. It's very reliant on the URL otherwise.

There are a few other places we could then replace current logic as well (out of scope for the time being though).

src/components/transactions/SingleTx/index.tsx
schmanu: I asked Lila and she said it's not necessary as we show the executionDate in the Accordion.

src/components/transactions/GroupLabel/index.tsx
```diff
@@ -29,15 +35,17 @@ export const useGroupLabel = (item: Label): string => {
     return label
   }
 
-  const nextNonce = nonceInFuture ? safe.nonce : firstTx.transaction.executionInfo.nonce
```
usame-algan, line 32: It looks like now we are always using `safe.nonce` in `futureNonceLabel`. Is the fallback to `firstTx.transaction.executionInfo.nonce` not needed anymore?

schmanu: Before we also always used `safe.nonce` but jut expressed in a unnecessary condition:
```ts  
const nonceInFuture = firstTx.transaction.executionInfo.nonce !== safe.nonce
const nextNonce = nonceInFuture ? safe.nonce : firstTx.transaction.executionInfo.nonce
```

which basically sets `nextNonce `to `safe.nonce` if the first transaction nonce is not equal to `safe.nonce` and if they are equal we set it to the first transaction nonce. But it doesn't change anything because they are equal.

src/hooks/useTxDetails.ts
schmanu: I will change it so that we can pass an optional `transactionID` and if this isn't passed we fallback to the id from the router. 

Of course I will also document it as a comment.

src/components/transactions/GroupLabel/index.tsx
usame-algan: True, it should be equal. Nice catch! 💯

iamacook (approved):

src/components/transactions/GroupLabel/index.tsx
```diff
@@ -29,15 +35,17 @@ export const useGroupLabel = (item: Label): string => {
...
+  return `${label} - ${futureNonceLabel}`
+}
 
-  return `${label} - transaction with nonce ${nextNonce} needs to be executed first`
+export const GroupLabelTypography = ({ label }: { label: string }) => {
+  return <div className={css.container}>{label}</div>
```
DiogoSoaress, line 42: Couldn't we try our `overline` typography?

schmanu: It sadly looks different :( I used it at first. I kept this typography as it was and just extracted it to its own component. So if we adjust our overline typography enough we could change this here at some point.

DiogoSoaress: Hmmm that's a pitty :/ we would benefit from sticking with the already implemented styles. But could be addressed in other PR as you didn't really touch it 👍

src/components/transactions/SingleTx/index.tsx
```diff
@@ -1,59 +1,53 @@
...
-    ? new Error("Couldn't retrieve the transaction details. Please review the URL.")
-    : !isCurrentSafeTx
-    ? new Error(`Transaction with id ${transactionId} not found in this Safe`)
-    : txDetailsError
+  const { safe } = useSafeInfo()
+
+  let nonceWarning: string | undefined = undefined
+  const executionInfo = txDetails?.detailedExecutionInfo
+  if (isMultisigDetailedExecutionInfo(executionInfo)) {
+    if (executionInfo.nonce > safe.nonce) {
```
katspaugh, line 35: This is much simpler (I like it) than the [DateLabel logic.](https://github.com/safe-global/web-core/blob/af182a29562c00a50b8e9eb83aa0e1bc6f4bab44/src/components/transactions/GroupLabel/index.tsx#L25-L32)

Why do we need that complicated logic there and not here?

src/components/transactions/SingleTx/index.tsx
```diff
@@ -1,59 +1,53 @@
...
-import { getTransactionDetails } from '@gnosis.pm/safe-react-gateway-sdk'
-import { sameAddress } from '@/utils/addresses'
+import type { Transaction, TransactionDetails } from '@gnosis.pm/safe-react-gateway-sdk'
 import type { ReactElement } from 'react'
-import { makeDateLabelFromTx, makeTxFromDetails } from '@/utils/transactions'
+import { makeTxFromDetails } from '@/utils/transactions'
 import { TxListGrid } from '@/components/transactions/TxList'
-import TxDateLabel from '@/components/transactions/TxDateLabel'
 import ExpandableTransactionItem from '@/components/transactions/TxListItem/ExpandableTransactionItem'
+import { useTxDetails } from '@/hooks/useTxDetails'
```
katspaugh, line 9: If this hook is used only here, I would put it in the same folder as this component.

src/hooks/useTxDetails.ts
```diff
@@ -0,0 +1,32 @@
...
+import type { TransactionDetails } from '@gnosis.pm/safe-react-gateway-sdk'
+import { getTransactionDetails } from '@gnosis.pm/safe-react-gateway-sdk'
+import useAsync from './useAsync'
+import useChainId from './useChainId'
+import useSafeInfo from './useSafeInfo'
+
+/**
+ * @returns TransactionDetails for transaction id in url
+ */
+export const useTxDetails = (txId?: string): [TransactionDetails | undefined, Error | undefined, boolean] => {
```
iamacook, line 11: Heads up: https://github.com/safe-global/web-core/pull/853 will require you tweak this hook a bit.

src/components/transactions/GroupLabel/index.tsx
```diff
@@ -1,43 +1,22 @@
...
-  }
-
-  const hasNextLabel = page.results.some((tx) => isLabelListItem(tx) && tx !== item)
-  const nonceInFuture = firstTx.transaction.executionInfo.nonce !== safe.nonce
-
-  if (!hasNextLabel && !nonceInFuture) {
-    return label
-  }
-
-  const nextNonce = nonceInFuture ? safe.nonce : firstTx.transaction.executionInfo.nonce
```
katspaugh, line 32: This logic was added in #660. I replaced it with just always display the current safe nonce next to the "Queued" label.
It's possible that the actual next tx will have a nonce higher than the current safe none, but that's an edge case.

src/pages/transactions/tx.tsx
```diff
@@ -11,9 +11,12 @@ const SingleTransaction: NextPage = () => {
         <title>Safe – Transaction details</title>
       </Head>
 
-      <TxHeader />
-
       <main>
+        <div>
```
iamacook, line 15: Why do we need an extra `div`?

iamacook (commented):
Looks good but the label is incorrect when the transaction nonce matches that of the Safe nonce. It should be "Next", without a warning:

image

src/pages/transactions/tx.tsx
katspaugh: Removed.

iamacook (approved):

iamacook:
Tested working for "Next", "Queued" and out of order nonce "Queued" transactions.

PR #783, opened 2022-09-30
refactor: use `UrlObject`s for internal links

iamacook:
## What it solves

Difficult to understand link structure

## How this PR fixes it

All internal links have been reformatted to `UrlObject`s.

## How to test it

The following links should work as before:

- Safe token widget
- Dashboard: assets, NFTs, pending and view all transactions
- Safe Apps: opening internal/external apps, sharing and demoing them
- Opening Safes via the sidebar
- Notifications: updating the Safe, transaction lifecycles and invalid mastercopy (which should open in a new tab).

PR #839, opened 2022-10-06
fix: wrap relative links in next/link

### LINKED ISSUE #794: Creating a safe from share app flow loading connect wallet screen first
## Bug description
When trying to create a safe for a EOA that was shared a safe app link, there is a small loading issue that would first load the 'connect wallet' screen instead of the creation one.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Rinkeby

## Steps to reproduce
With a EOA that has no safes to trigger the create safe flow from the shared app link, click on it

## Expected result
Mucho bueno transition to the creation screen.

## Obtained result
Transition is a bit off, going through the connect wallet flow for 1-2 seconds first then loading the create safe flow. 

(The video below shows this in a very dramatic way due to throttling)

https://user-images.githubusercontent.com/92332710/193602535-24f88c56-ad7c-40e7-8cdc-7f15a8610aaf.mp4

DiogoSoaress:
## What it solves

Resolves #794

## How this PR fixes it
Wraps relative links in `next/link` avoiding a page reload

## How to test it
1. Connect with an account that does NOT own a Safe
2. Navigate to a shared Transaction Builder link
3. Click "Create new Safe"
4. The navigation to the Create Safe screen should be smooth and not lose your wallet connection

## Analytics changes
None

## Screenshots
Screenshot 2022-10-06 at 11 35 40


iamacook (commented):
Looks good to me. We should be careful with this before QAing it though as there is a [refactor PR open](https://github.com/safe-global/web-core/pull/783/files) that will interfere with most of this.

Perhaps we should leave this in "Peer Review" for the time being until that is merged. Wdyt? cc @francovenica

DiogoSoaress:
Merge after #783

katspaugh:
Please pull from dev.

PR #845, opened 2022-10-06
fix: rename Safe Apps settings route

iamacook:
## What it solves

Adjusting the Safe Apps settings route

## How this PR fixes it

The Safe Apps settings router has been renamed to `/settings/safe-apps` and the title adjusted to match our lowercase styling.

## How to test it

Open the Safe Apps permissions settings and observe the title be "Safe Apps permissions" and route `/settings/safe-apps`

## Screenshots

image

### DIFF
--- a/src/config/routes.ts
+++ b/src/config/routes.ts
@@ -17,7 +17,7 @@ export const AppRoutes = {
     modules: '/settings/modules',
     index: '/settings',
     appearance: '/settings/appearance',
-    safeAppsPermissions: '/settings/safe-apps-permissions',
+    safeApps: '/settings/safe-apps',
   },
   share: {
     safeApp: '/share/safe-app',
--- a/src/config/routes.ts
+++ b/src/config/routes.ts
@@ -17,7 +17,7 @@ export const AppRoutes = {
     modules: '/settings/modules',
     index: '/settings',
     appearance: '/settings/appearance',
-    'safe-apps': {
+    safeApps: {
       index: '/settings/safe-apps',
     },
   },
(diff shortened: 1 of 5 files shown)


src/config/routes.ts
```diff
@@ -17,7 +17,7 @@ export const AppRoutes = {
     modules: '/settings/modules',
     index: '/settings',
     appearance: '/settings/appearance',
-    safeAppsPermissions: '/settings/safe-apps-permissions',
```
katspaugh, line 20: Did you change this by hand? Because it's supposed to generate a nested object here.

src/config/routes.ts
```diff
@@ -17,7 +17,7 @@ export const AppRoutes = {
...
+    safeApps: '/settings/safe-apps',
```
yagopv, line 20: Not sure why `settings/safe-apps/` is better than `settings/safe-apps-permissions`, perhaps `settings/safe-apps/permissions` if you don't like the current route?

yagopv: Or you think about if we add more settings in the future we can create subsections there?

yagopv (approved):

src/config/routes.ts
iamacook: Exactly that. For the time being, I think `/settings/safe-apps` is sufficient (until we add more).

src/config/routes.ts
katspaugh: Could you please mod the script to camelCase dash-separated paths?

katspaugh (approved):

iamacook:
I'll merge this as it's a route change, confirmed working.

liliya-soroka:
noted

PR #850, opened 2022-10-06
fix: Adjust page header width

### LINKED ISSUE #849: Page header content is cut off
## Bug description

The page header is cut off on the right side of the screen

## Steps to reproduce
 1. Go to a safe
 2. Navigate to transaction history or address book
 3. Observe the safe header is cut off on the right side of the screen

## Expected result

Everything in the safe header is visible on the screen

## Obtained result

Content inside the safe header is cut off

## Screenshots
Screenshot 2022-10-06 at 16 36 56
Screenshot 2022-10-06 at 16 37 07

usame-algan:
## What it solves

Resolves #849 

## How this PR fixes it

Since we are using `box-sizing: content-box` we have to subtract the padding from the total width.

## How to test it

1. Open the Safe
2. Navigate to Tx history, AB and any other page with a page header
3. Observe that the page header content is visible

## Screenshots
Screenshot 2022-10-06 at 17 21 09

### DIFF
--- a/src/components/common/PageHeader/styles.module.css
+++ b/src/components/common/PageHeader/styles.module.css
@@ -7,7 +7,7 @@
   border-bottom: 1px solid var(--color-border-light);
   background-color: var(--color-background-paper);
   z-index: 1;
-  width: 100%;
+  width: calc(100% - var(--space-6));
   position: sticky !important;
 }
 
--- a/src/services/tx/txSender.ts
+++ b/src/services/tx/txSender.ts
@@ -253,18 +253,18 @@ export const dispatchOnChainSigning = async (safeTx: SafeTransaction, provider: 
   const sdkUnchecked = await getUncheckedSafeSDK(provider)
   const safeTxHash = await sdkUnchecked.getTransactionHash(safeTx)
 
-  txDispatch(TxEvent.EXECUTING, { txId, groupKey: safeTxHash })
+  txDispatch(TxEvent.EXECUTING, { groupKey: safeTxHash })
 
   try {
     // With the unchecked signer, the contract call resolves once the tx
     // has been submitted in the wallet not when it has been executed
     await sdkUnchecked.approveTransactionHash(safeTxHash)
   } catch (err) {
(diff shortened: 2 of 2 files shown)


src/services/tx/txSender.ts
```diff
@@ -253,18 +253,18 @@ export const dispatchOnChainSigning = async (safeTx: SafeTransaction, provider:
...
     // has been submitted in the wallet not when it has been executed
     await sdkUnchecked.approveTransactionHash(safeTxHash)
   } catch (err) {
-    txDispatch(TxEvent.FAILED, { txId, groupKey: safeTxHash, error: err as Error })
+    txDispatch(TxEvent.FAILED, { groupKey: safeTxHash, error: err as Error })
     throw err
   }
 
-  txDispatch(TxEvent.AWAITING_ON_CHAIN_SIGNATURE, { txId, groupKey: safeTxHash })
+  txDispatch(TxEvent.AWAITING_ON_CHAIN_SIGNATURE, { groupKey: safeTxHash })
```
usame-algan, line 267: Somehow Github didn't show any conflicts for this when we merged https://github.com/safe-global/web-core/pull/740 so I had to fix it here 🤯

iamacook (approved):

src/components/common/PageHeader/styles.module.css
```diff
@@ -7,7 +7,7 @@
   border-bottom: 1px solid var(--color-border-light);
   background-color: var(--color-background-paper);
   z-index: 1;
-  width: 100%;
+  width: calc(100% - var(--space-6));
```
katspaugh, line 10: My bad! I've added box-sizing to the container and forgot about the horizontal padding.

katspaugh (approved):
👍

katspaugh:
Tested, it's fixed. ✅

src/components/common/PageHeader/styles.module.css
usame-algan: I tried it without the box-sizing but then the 0.5px padding we add doesn't work anymore so we would have to find another way.

PR #852, opened 2022-10-07
Fix: wrong error code for load chains

katspaugh:
## What it solves

Simply fixed an inconsistent error code choice.

### DIFF
--- a/src/hooks/useMasterCopies.ts
+++ b/src/hooks/useMasterCopies.ts
@@ -1,5 +1,6 @@
 import useAsync from '@/hooks/useAsync'
 import useChainId from '@/hooks/useChainId'
+import { Errors, logError } from '@/services/exceptions'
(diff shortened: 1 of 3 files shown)


src/hooks/useMasterCopies.ts
```diff
@@ -35,7 +36,7 @@ export const useMasterCopies = () => {
       const res = await getMasterCopies(chainId)
       return res.map(extractMasterCopyInfo)
     } catch (error) {
-      console.error('Fetching data from master-copies errored', error)
+      logError(Errors._620, (error as Error).message)
```
iamacook, line 39:
```suggestion
      logError(Errors._619, (error as Error).message)
```

katspaugh: Thanks!

iamacook (approved):

PR #853, opened 2022-10-07
Fix: prevent error flashing in SingleTx

### LINKED ISSUE #851: [Single tx] It flashes "Failed to load tx details" before loading the tx details
It should not show an error while loading.

katspaugh:
## What it solves

Resolves #851

## How this PR fixes it

`query.id` is undefined at first because the router is client-side only. So the page is pre-generated with an error message.

I removed the handling of a missing query.id. It will just show an infinite loader if you open a page w/o the query param. I think it's fine because why would you open it like this?

### DIFF
--- a/src/components/transactions/SingleTx/index.tsx
+++ b/src/components/transactions/SingleTx/index.tsx
@@ -20,7 +20,7 @@ const SingleTxGrid = ({ txDetails }: { txDetails: TransactionDetails }): ReactEl
   return (
     <TxListGrid>
       <TxDateLabel item={dateLabel} />
-      <ExpandableTransactionItem item={tx} txDetails={txDetails} testId="single-tx" />
+      <ExpandableTransactionItem item={tx} txDetails={txDetails} />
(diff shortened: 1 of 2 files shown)


src/components/transactions/SingleTx/index.tsx
```diff
@@ -35,21 +35,23 @@ const SingleTx = () => {
   const [txDetails, txDetailsError] = useAsync<TransactionDetails>(
     () => {
       if (!transactionId) return
-      return getTransactionDetails(chainId, transactionId)
+
+      console.log(safeAddress, transactionId)
```
iamacook, line 39: 👀

iamacook (commented):
There still seems to be a very brief error shown:

single tx error

katspaugh:
@iamacook fixed, it was because safeAddress is async, it was undefined on first request (and also it would be doing two requests).

iamacook (approved):

PR #856, opened 2022-10-07
fix: rank Claiming App first on dashboard

### LINKED ISSUE #829: Dashboard safe apps - Claim app must be the first one in the list
## Bug description
The $Safe claimin app is not the first one in the list. This is for ETH mainnet and is how it is currenlty working in safe-react prod

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet
 - Env: https://web-core.pages.dev/eth:0xAAb5A43af6599845f2D97EC3c4E23b23528bAaB1/home

## Steps to reproduce
 1. Go to any safe
 2. Check the dashboard > Apps widget

## Expected result
The claiming app should be always the first one, regardless of amount of visits or if it is pinned or not.

## Screenshots
image

iamacook:
## What it solves

Resolves #829

## How this PR fixes it

The Claiming App is now positioned first in the dashboard featured apps.

Note: the Claiming App now has a `"safe-claiming-app"` tag and is instead found by this instead of it's `id` (which changes depending on environment).

## How to test it

Open the dashboard on a network that supports the Claiming App and observe it in first position. If necessary, add the `"safe-claiming-app"` to a network temporarily.

## Screenshots

image

### DIFF
--- a/src/hooks/safe-apps/useRankedSafeApps.ts
+++ b/src/hooks/safe-apps/useRankedSafeApps.ts
@@ -13,9 +13,10 @@ const useRankedSafeApps = (safeApps: SafeAppData[], pinnedSafeApps: SafeAppData[
     const mostUsedApps = rankSafeApps(safeApps)
     const rankedPinnedApps = rankSafeApps(pinnedSafeApps)
     const randomApps = safeApps.slice().sort(() => Math.random() - 0.5)
+    const safeClaimingApp = safeApps?.filter((app) => app.tags?.includes(SafeAppsTag.SAFE_CLAIMING_APP)) || []
 
-    const allRankedApps = rankedPinnedApps
-      .concat(pinnedSafeApps, mostUsedApps, randomApps)
+    const allRankedApps = safeClaimingApp
+      .concat(rankedPinnedApps, pinnedSafeApps, mostUsedApps, randomApps)
       // Filter out Featured Apps because they are in their own section
       .filter((app) => !app.tags.includes(SafeAppsTag.DASHBOARD_FEATURED))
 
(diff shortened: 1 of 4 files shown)


src/hooks/safe-apps/useRankedSafeApps.ts
```diff
@@ -17,13 +19,22 @@ const useRankedSafeApps = (safeApps: SafeAppData[], pinnedSafeApps: SafeAppData[
...
-    const numberOfRandomSafeApps = NUMBER_OF_SAFE_APPS - mostUsedSafeApps.length
     const nonRankedApps = safeApps.filter((app) => !mostUsedSafeAppsIds.includes(String(app.id)))
-    const randomSafeApps = sampleSize(nonRankedApps, numberOfRandomSafeApps)
+    const randomSafeApps = sampleSize(nonRankedApps, NUMBER_OF_SAFE_APPS)
 
-    const rankedSafeApps = [...mostUsedSafeApps, ...randomSafeApps]
+    const rankedSafeApps = mostUsedSafeApps.concat(randomSafeApps)
 
-    return rankedSafeApps
+    // Display the safe-claiming-app at the first position
```
katspaugh, line 27: Can we do it only for Safes that actually have something to claim/already claimed?
Might be not very interesting or even annoying to show it to not eligible users.

iamacook: I think we should leave it as is to match the widget. We'd otherwise have to fetch the total allocation. We can always remove the tag if/when needed to remove it from the top position.

katspaugh (approved):

katspaugh:
I've tested, it works. ✅ 

Will merge into #857.

francovenica:
Works fine in 857. LGTM

PR #857, opened 2022-10-07
fix: Always display pinned apps on top in dashboard

### LINKED ISSUE #831: Dashboard Safe apps - Pinned apps are not the first in the list
## Bug description
The pinned apps are not being set first in the Dashboard safe apps widget

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to any safe > Safe apps
 2. Pin a specific app
 3. Visit another app several times (just enter in the app, thats considered a visit)

## Expected result
The pinned app, even tho has never been visited should be first, the app you visited several times should be second

## Obtained result
The app with more visits is first and the pinned app is second

## Screenshots
N/A

usame-algan:
## What it solves

Resolves #831 

## How this PR fixes it

- Adjusts ranking function to exclude pinned apps
- Always shows pinned apps before ranked and random apps

## How to test it

1. Open the Safe
2. Navigate to the dashboard
3. It is possible to pin safe apps
4. Pinned apps always appear in the first position
5. The list still only shows 5 apps
6. Random apps at the end of the list are not reordering anymore when pinning/unpinning apps
7. Reloading the page keeps the pinned state of the apps

## Screenshots
Screenshot 2022-10-07 at 12 18 05


src/hooks/safe-apps/useRemoteSafeApps.ts
```diff
@@ -8,17 +8,18 @@ import useAsync from '../useAsync'
 
 // To avoid multiple simultaneous requests (e.g. the Dashboard and the SAFE header widget),
 // cache the request promise for 100ms
-let cache: Promise<SafeAppsResponse> | undefined
-const cachedGetSafeApps = (chainId: string): ReturnType<typeof getSafeApps> => {
-  if (!cache) {
-    cache = getSafeApps(chainId, { client_url: window.location.origin })
+let cache: Record<string, Promise<SafeAppsResponse> | undefined> = {}
```
usame-algan, line 11: Had to adjust the cache because it would only do one request on the default chains (gor, mainnet) and remove safes that were previously pinned if those didn't exist on either of the default chains

src/hooks/safe-apps/useRankedSafeApps.ts
```diff
@@ -1,30 +1,24 @@
...
+import { difference } from 'lodash'
 import type { SafeAppData } from '@gnosis.pm/safe-react-gateway-sdk'
 import { getAppsUsageData, rankSafeApps } from '@/services/safe-apps/track-app-usage-count'
 
-// number of ranked Safe Apps that we include in the array
+// number of ranked Safe Apps that we want to display
 const NUMBER_OF_SAFE_APPS = 5
 
 const useRankedSafeApps = (safeApps: SafeAppData[], pinnedSafeApps: SafeAppData[]): SafeAppData[] => {
-  return useMemo(() => {
```
usame-algan, line 10: The random items still keep their previous order even after removing `useMemo`.

DiogoSoaress: It doesn't even seem to be randomizing as the "random" apps are displayed in alphabetical order

DiogoSoaress: and per "Safe Apps" page order

DiogoSoaress:
Is it expected to allow "WalletConnect" and "Transaction Builder" to appear in the _Safe Apps section_?

Screenshot 2022-10-07 at 12 48 39

src/hooks/safe-apps/useRankedSafeApps.ts
usame-algan: Oh, I accidently removed the `sampleSize` call. It should be random again now.

src/hooks/safe-apps/useRankedSafeApps.ts
```diff
@@ -1,30 +1,27 @@
...
 
-    // we add random Safe Apps if no enough Safe Apps are present
-    const numberOfRandomSafeApps = NUMBER_OF_SAFE_APPS - mostUsedSafeApps.length
-    const nonRankedApps = safeApps.filter((app) => !mostUsedSafeAppsIds.includes(String(app.id)))
-    const randomSafeApps = sampleSize(nonRankedApps, numberOfRandomSafeApps)
+  const usageSafeAppsData = getAppsUsageData()
+  const mostUsedSafeAppsIds = rankSafeApps(usageSafeAppsData)
+  const mostUsedSafeApps = mostUsedSafeAppsIds
+    .map((id) => nonFeaturedApps.find((safeApp) => String(safeApp.id) === id))
+    .filter(Boolean) as SafeAppData[]
```
katspaugh, line 18: I would replace this with 
```
.filter((app) => !app.tags?.includes(FEATURED_APPS_TAG))
```

usame-algan: I tried removing the filter but it seems like `difference` is not filtering out `undefined` items. What did you have in mind?

katspaugh: I meant the filter by featured.

usame-algan: I moved the filter call to the end of the function as discussed but noticed that it could lead to some edge cases e.g. if there are no pinned and ranked safe apps and we randomly select 5 safe apps and by chance one of them is a featured app it would lead to fewer than 5 random apps being displayed.

src/hooks/safe-apps/useRankedSafeApps.ts
```diff
@@ -1,30 +1,22 @@
...
-    const nonRankedApps = safeApps.filter((app) => !mostUsedSafeAppsIds.includes(String(app.id)))
-    const randomSafeApps = sampleSize(nonRankedApps, numberOfRandomSafeApps)
-
-    const rankedSafeApps = [...mostUsedSafeApps, ...randomSafeApps]
-
-    return rankedSafeApps
-  }, [safeApps, pinnedSafeApps])
+  return pinnedSafeApps
+    .concat(mostUsedApps)
+    .concat(randomApps)
```
iamacook, line 17:
```suggestion
    .concat(mostUsedApps, randomApps)
```

src/hooks/safe-apps/useRankedSafeApps.ts
katspaugh: Well, we should select 7 random apps then, not 5.

katspaugh (approved):

francovenica:
The main issue of apps not being able to be pinned from the dashboard and put them in at the top of the list was fixed.
I have 2 issues, but these can be their own tickets if you feel they are outside of the scope of this ticket. Let me know

1 - In prod we also have this "most visited apps first" among the pinned ones (so the most visited pinned app comes before a less visited pinned app). In this PR it seems that the pinned apps are fixed once you pin them, they seem to completely ignore this "most visited" feature.

2 - Also in prod we have an animation that comes win pin/unpin apps:
pin apps

katspaugh:
@francovenica I've added ranking to pinned apps as well.
Reg. animation, it's a feature that they don't reorder now.

iamacook (approved):

francovenica:
LGTM

There is a ranking system for the pinned apps as well now, the most visited ones are put first in the list.
Noted then that the animation that reorders the app won't play anymore.

PR #858, opened 2022-10-07
Fix: don't display EIP-1559 params for non-EIP-1559 chains

### LINKED ISSUE #855: [L2- Arbitrum] Gas price  is bigger than in safe-react
## Bug description
Gas price  is bigger than in safe-react 

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain:Arbitrum prod

## Steps to reproduce
 1. Go to https://web-core.pages.dev/arb1:0x861099a9FD94b03471a1002efCa662c3ECAdB10C/transactions/tx?id=multisig_0x861099a9FD94b03471a1002efCa662c3ECAdB10C_0x40b0ca2628b39d7f46d67575f4f3d733e9891062119f00994244581903110407
 2. Click execute tx on web-core and safe-react -> check gas estimation
Current result:
web-core : gas price = 1.7
safe-react : gas price = 0.1

## Expected result
Make sure that the price is proper on web-core

katspaugh:
## What it solves

Resolves #855

## How this PR fixes it

Adds a check in GasParams and AdvancedParamsForm to show the old `gas price` if the chain is configured as not supporting EIP-1559.

## How to test it
* Make a tx on Arbitrum (no EIP-1559) – the gas price should be around 0.1 Gwei.
* Make a tx on Mainnet – it should have max priority fee set to 1.5 Gwei.

## Screenshots
Screenshot 2022-10-07 at 13 28 37

### DIFF
--- a/src/components/tx/GasParams/index.tsx
+++ b/src/components/tx/GasParams/index.tsx
@@ -32,11 +32,11 @@ const GasParams = ({ params, isExecution, isEIP1559, onEdit }: GasParamsProps): 
   }
 
   const chain = useCurrentChain()
-  const isLoading = !gasLimit || !maxFeePerGas || !maxPriorityFeePerGas
+  const isLoading = !gasLimit || !maxFeePerGas
 
   // Total gas cost
   const totalFee = !isLoading
-    ? formatVisualAmount(maxFeePerGas.add(maxPriorityFeePerGas).mul(gasLimit), chain?.nativeCurrency.decimals)
+    ? formatVisualAmount(maxFeePerGas.mul(gasLimit), chain?.nativeCurrency.decimals)
     : '> 0.001'
 
   // Individual gas params
--- a/src/hooks/useGasPrice.ts
+++ b/src/hooks/useGasPrice.ts
@@ -2,14 +2,11 @@ import { useMemo } from 'react'
 import { BigNumber } from 'ethers'
 import type { FeeData } from '@ethersproject/providers'
 import type { GasPrice, GasPriceOracle } from '@gnosis.pm/safe-react-gateway-sdk'
-import { FEATURES } from '@gnosis.pm/safe-react-gateway-sdk'
 import { GAS_PRICE_TYPE } from '@gnosis.pm/safe-react-gateway-sdk'
-import type { JsonRpcProvider } from '@ethersproject/providers'
 import useAsync from '@/hooks/useAsync'
 import { useCurrentChain } from './useChains'
 import useIntervalCounter from './useIntervalCounter'
 import { useWeb3ReadOnly } from '../hooks/wallets/web3'
-import { hasFeature } from '@/utils/chains'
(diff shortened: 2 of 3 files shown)


src/hooks/useGasPrice.ts
```diff
@@ -59,31 +81,42 @@ const useGasPrice = (): {
...
+  )
+
+  // Fetch the gas fees from the blockchain itself
+  const [feeData, feeDataError, feeDataLoading] = useAsync<FeeData>(
+    () => {
+      if (!chain || !provider) return
+      return _getFeeData(provider, hasFeature(chain, FEATURES.EIP1559))
+    },
+    [chain, provider, counter],
+    false,
```
iamacook, line 102: Why do we now opt to clear the data?

src/hooks/useGasPrice.ts
```diff
@@ -59,31 +81,42 @@ const useGasPrice = (): {
...
+  )
+
+  // Prepare the return values
+  const maxFee = gasPrice || feeData?.maxFeePerGas || undefined
```
iamacook, line 106: If it's EIP-1559 capable, shouldn't we preferentially return `maxFeePerGas` first?

iamacook (commented):
Works well when testing it but I have a couple of questions and there are failing tests.

src/hooks/useGasPrice.ts
katspaugh: False means NOT to clear the data. It's how it's been.

src/hooks/useGasPrice.ts
katspaugh: No, `gasPrice` is the oracle price. `maxFeePerGas` from the chain is just a fallback.

iamacook (approved):

src/components/tx/GasParams/index.tsx
```diff
@@ -32,11 +32,11 @@ const GasParams = ({ params, isExecution, isEIP1559, onEdit }: GasParamsProps):
   }
 
   const chain = useCurrentChain()
-  const isLoading = !gasLimit || !maxFeePerGas || !maxPriorityFeePerGas
+  const isLoading = !gasLimit || !maxFeePerGas
 
   // Total gas cost
   const totalFee = !isLoading
-    ? formatVisualAmount(maxFeePerGas.add(maxPriorityFeePerGas).mul(gasLimit), chain?.nativeCurrency.decimals)
+    ? formatVisualAmount(maxFeePerGas.mul(gasLimit), chain?.nativeCurrency.decimals)
```
katspaugh, line 39: [Ethers.js already includes maxPrioFee in the maxFee](https://github.com/ethers-io/ethers.js/blob/ec1b9583039a14a0e0fa15d0a2a6082a2f41cf5b/packages/abstract-provider/src.ts/index.ts#L253), so no need to sum it here.

Issue #722, opened 2022-09-28
No warning is shown when trying to sign tx as non owner


JagoFigueroa:
## Bug description
When trying to sign a tx on a safe without being an owner no warning is shown to the user.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Mainnet

## Steps to reproduce
a) As a non owner of a 1.1.1 mainnet safe (that I will share privately when tackling this ticket) try to sign a tx on any safe app.

b) As a non owner of a 1.3.0 mainnet safe (that I will share privately when tackling this ticket) try to sign a tx on any safe app.

## Expected result
Warning is shown to the user blocking the flow to continue in either case.

## Obtained result
a) No warning is show to the user and can click submit button, showing this error:
<img width="414" alt="Screenshot 2022-09-28 at 14 37 15" src="https://user-images.githubusercontent.com/92332710/192780401-d7a32833-569d-4c59-911e-05090a606d10.png">

b) Generic error warning is shown:

<img width="586" alt="Screenshot 2022-09-28 at 14 43 27" src="https://user-images.githubusercontent.com/92332710/192781508-64f02de0-a88d-4466-b80e-4a8e5aa957f5.png">

and clicking submit button will show an error.

JagoFigueroa:
I edited the ticket a little bit after the latest info and lowered the priority. 

I think it would be nice to show the good old 'You are currently not an owner of this Safe and won't be able to submit this transaction.' info in the pop up itself blocking the non owners to continue from there but nothing major.

Gracias guys!

PR #860, opened 2022-10-07
fix: display specific error message for wrong chain and non-owner wallets

### LINKED ISSUE #769: [Tx modal] Show specific error when trying to submit a tx while on a different chain
## Bug description
When trying to submit a tx on a safe while connected to a different chain no accurate description is shown to the user as in legacy.

## Steps to reproduce
Try to perform a transaction on a mainnet safe while being connected to Rinkeby on Metamask (for example).

## Expected result
(legacy)
Screenshot 2022-09-29 at 17 42 12

## Obtained result
(revamp)
Screenshot 2022-09-29 at 17 43 57

niklasnatter:
## What it solves

This PR displays a specific error message in the transaction modal if the connected wallet is connected to the wrong chain or if it is not an owner of the wallet. 

Resolves #769, #722

## How this PR fixes it

If the connected wallet is connected to the wrong chain or is not an owner, an explicit error message that explains the problem is displayed above the generic `This transaction will most likely fail.` message. Additionally, it disables the submit button in these cases. I used the error messages from #769 and #722.

## How to test it

1. Create a transaction in the transaction builder
2. Switch the connected wallet to a different chain or connect an unrelated address
3. Open the transaction overlay via the `Send Batch` button

## Screenshots
Screenshot 2022-10-07 at 16 08 36
Screenshot 2022-10-07 at 16 09 14

### DIFF
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -166,7 +166,9 @@ const SignOrExecuteForm = ({
     setAdvancedParams(data)
   }
 
-  const submitDisabled = !isSubmittable || isEstimating || !tx || disableSubmit
+  const walletIsWrongChain = currentChain?.chainId != wallet?.chainId
+  const walletIsNoOwner = !safe.owners.some((owner) => owner.value === wallet?.address)
+  const submitDisabled = !isSubmittable || isEstimating || !tx || walletIsWrongChain || walletIsNoOwner || disableSubmit
 
   return (
     <form onSubmit={handleSubmit}>
@@ -192,6 +194,14 @@ const SignOrExecuteForm = ({
           canExecute={canExecute}
           disabled={submitDisabled}
         />
+
+        {walletIsWrongChain && <ErrorMessage>Your wallet is connected to the wrong chain.</ErrorMessage>}
+
+        {walletIsNoOwner && (
+          <ErrorMessage>
+            You are currently not an owner of this Safe and won&apos;t be able to submit this transaction.
+          </ErrorMessage>
+        )}
 
         {(error || (willExecute && gasLimitError)) && (
           <ErrorMessage error={error || gasLimitError}>
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -166,9 +166,8 @@ const SignOrExecuteForm = ({
     setAdvancedParams(data)
   }
 
-  const walletIsWrongChain = currentChain?.chainId != wallet?.chainId
-  const walletIsNoOwner = !safe.owners.some((owner) => owner.value === wallet?.address)
-  const submitDisabled = !isSubmittable || isEstimating || !tx || walletIsWrongChain || walletIsNoOwner || disableSubmit
+  const isWalletAllowed = willExecute || safe.owners.some((owner) => owner.value === wallet?.address)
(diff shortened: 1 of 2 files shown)


src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -164,7 +164,9 @@ const SignOrExecuteForm = ({
     setAdvancedParams(data)
   }
 
-  const submitDisabled = !isSubmittable || isEstimating || !tx
+  const walletIsWrongChain = currentChain?.chainId != wallet?.chainId
```
katspaugh, line 167: We've just merged another PR that adds `const isWrongChain = useIsWrongChain()`. Please pull from dev, you can then use this `isWrongChain` var.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -164,7 +164,9 @@ const SignOrExecuteForm = ({
...
+  const walletIsNoOwner = !safe.owners.some((owner) => owner.value === wallet?.address)
```
katspaugh, line 168: Non-owners can execute transactions, thought. So I would suggest to do smth like 
```
  const walletCanSubmit = willExecute || isSignableBy(transaction, wallet?.address || '')
```

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -191,6 +193,14 @@ const SignOrExecuteForm = ({
           disabled={submitDisabled}
         />
 
+        {walletIsWrongChain && <ErrorMessage>Your wallet is connected to the wrong chain.</ErrorMessage>}
+
+        {walletIsNoOwner && (
+          <ErrorMessage>
+            You are currently not an owner of this Safe and won&apos;t be able to submit this transaction.
+          </ErrorMessage>
+        )}
```
katspaugh, line 202: That's nice 👍

katspaugh (commented):
Many thanks Niklas! 🙏

src/components/tx/SignOrExecuteForm/index.tsx
katspaugh: However, if it's a new tx (no `txId` passed), your check will work fine.
So to break down:

* New tx – must be an owner
* Existing tx to be signed – needs to be an owner who hasn't signed yet
* Existing tx to be executed – can be anyone

niklasnatter: Thanks for the feedback! I added the suggestions and fixed the testcases.

I am not familiar enough with the domain logic here- does your check already cover all 3 requirements or do i need to check if `txId` is passed additionally?

katspaugh: I think your current version should work as it is 👍

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -164,7 +165,9 @@ const SignOrExecuteForm = ({
     setAdvancedParams(data)
   }
 
-  const submitDisabled = !isSubmittable || isEstimating || !tx
+  const isWrongChain = useIsWrongChain()
```
katspaugh, line 168: I've just realized this branch is based off of `main` but we've recently switched to `dev` as our default branch.
Could you please rebase to dev and also change the base branch of the PR? Then this `useIsWrongChain` will be there, it's on dev.

katspaugh (approved):
LGTM, thank you Niklas!

PR #861, opened 2022-10-07
fix: add Sentry to the CSP allowed domains

### LINKED ISSUE #827: Safe upgrade - Error message too dark for the dark theme
## Bug description
When trying to update a safe 1.0.0 an error shows up (the console shows that safe 1.0.0) are not compatible
This error message is has the font-color black even in for the dark theme

## Environment
 - Browser: Chrome
 - Wallet: any
 - Chain: any

## Steps to reproduce
 1. Try to update a safe 1.0.0

## Screenshots
image

DiogoSoaress:
## What it solves

Resolves #827 

## How this PR fixes it
Adds `https://*.ingest.sentry.io` to our CSP allowed src so the Core SDK error is handled without crashing the app.

## How to test it
1. Try to update a 1.0.0 Safe (which will throw an uncatched error)
2. The app should handle the error elegantly and not crash (i.e display our ErrorBoundary component)

## Analytics changes
None

## Screenshots
N/A

### DIFF
--- a/src/config/securityHeaders.ts
+++ b/src/config/securityHeaders.ts
@@ -12,7 +12,7 @@ import { CYPRESS_MNEMONIC, IS_PRODUCTION } from '@/config/constants'
 export const ContentSecurityPolicy = `
  default-src 'self';
  connect-src 'self' *;
- script-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com 'unsafe-inline' https://*.getbeamer.com https://www.googletagmanager.com ${
+ script-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com 'unsafe-inline' https://*.getbeamer.com https://www.googletagmanager.com https://*.ingest.sentry.io ${
    !IS_PRODUCTION || /* TODO: remove after moving cypress to görli and testing in staging again!! */ CYPRESS_MNEMONIC
      ? "'unsafe-eval'"
      : ''


src/config/securityHeaders.ts
```diff
@@ -12,7 +12,7 @@ import { CYPRESS_MNEMONIC, IS_PRODUCTION } from '@/config/constants'
 export const ContentSecurityPolicy = `
  default-src 'self';
  connect-src 'self' *;
- script-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com 'unsafe-inline' https://*.getbeamer.com https://www.googletagmanager.com ${
+ script-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com 'unsafe-inline' https://*.getbeamer.com https://www.googletagmanager.com https://*.ingest.sentry.io ${
```
iamacook, line 15: Are all Sentry logs using the `*.ingest` subdomain or should we allow all?

DiogoSoaress: Good point. Should I change to `https://*.sentry.io` ?

DiogoSoaress: I was trying to not be so loose with it

iamacook: I glossed over their docs and it seems that `*.ingest` is correct.

iamacook (approved):

DiogoSoaress:
Merging as it needs to be verified in a deployed env

src/config/securityHeaders.ts
DiogoSoaress: I think I also need to add `https://sentry.io` given in `https://web-core.pages.dev/` the domain is different.

Screenshot 2022-10-07 at 17 12 07

PR #862, opened 2022-10-07
chore: extend CSP domains

### LINKED ISSUE #827: Safe upgrade - Error message too dark for the dark theme

DiogoSoaress:
## What it solves

Resolves #827 

## How this PR fixes it
Adds `https://*.ingest.sentry.io` and `https://sentry.io`  to our CSP allowed src so the Core SDK error is handled without crashing the app.

## How to test it
1. Try to update a 1.0.0 Safe (which will throw an uncatched error)
2. The app should handle the error elegantly and not crash (i.e display our ErrorBoundary component)

## Analytics changes
None

## Screenshots
N/A

### DIFF
--- a/src/config/securityHeaders.ts
+++ b/src/config/securityHeaders.ts
@@ -12,7 +12,7 @@ import { CYPRESS_MNEMONIC, IS_PRODUCTION } from '@/config/constants'
 export const ContentSecurityPolicy = `
  default-src 'self';
  connect-src 'self' *;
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
👍

PR #873, opened 2022-10-10
Fix: auto-choose to WalletConnect on mobile

### LINKED ISSUE #871: [Mobile] Wallet connection – use WalletConnect instead of onboard
## Bug description
Most wallets we offer in the onboard popup don't work on mobile, so it's better to just open the WalletConnect popup right away.

katspaugh:
## What it solves

Resolves #871

## How this PR fixes it
When the user presses "Connect wallet" on mobile, it automatically pops up the WalletConnect modal instead the onboard modal.

### DIFF
--- a/src/components/common/ConnectWallet/WalletDetails.tsx
+++ b/src/components/common/ConnectWallet/WalletDetails.tsx
@@ -1,13 +1,23 @@
-import { Button, Typography } from '@mui/material'
+import { Button, Typography, useMediaQuery } from '@mui/material'
 import type { ReactElement } from 'react'
-
+import { useTheme } from '@mui/material/styles'
 import useOnboard, { connectWallet } from '@/hooks/wallets/useOnboard'
 import { OVERVIEW_EVENTS } from '@/services/analytics/events/overview'
 import KeyholeIcon from '@/components/common/icons/KeyholeIcon'
 import { trackEvent } from '@/services/analytics'
+import { WalletNames } from '@/utils/wallets'
 
 const WalletDetails = ({ onConnect }: { onConnect?: () => void }): ReactElement => {
   const onboard = useOnboard()
+  const theme = useTheme()
+  const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
+
+  // On mobile, automatically choose WalletConnect
+  const options = isMobile
+    ? {
+        autoSelect: WalletNames.WALLET_CONNECT,
+      }
+    : undefined
 
   const handleConnect = async () => {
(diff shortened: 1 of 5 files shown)


src/components/common/ConnectWallet/WalletDetails.tsx
```diff
@@ -1,13 +1,23 @@
...
 import useOnboard, { connectWallet } from '@/hooks/wallets/useOnboard'
 import { OVERVIEW_EVENTS } from '@/services/analytics/events/overview'
 import KeyholeIcon from '@/components/common/icons/KeyholeIcon'
 import { trackEvent } from '@/services/analytics'
+import { WalletNames } from '@/utils/wallets'
 
 const WalletDetails = ({ onConnect }: { onConnect?: () => void }): ReactElement => {
   const onboard = useOnboard()
+  const theme = useTheme()
+  const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
```
iamacook, line 13: Larger tablets won't be deemed as a mobile device. I'd recommend we use `@web3-onboard`'s `getDevice` function instead.

iamacook (commented):
It's possible to install plugins on Firefox mobile, e.g. MetaMask. We should probably take this into account. What do you think?

After looking through the code, let's switch `onboard.connectWallet` with `connectWallet` on `SafeAppLanding` in this PR as well.

katspaugh:
I haven't checked mobile browser extensions, but you can go back to the onboard modal if you close the WC modal.

src/components/common/ConnectWallet/WalletDetails.tsx
katspaugh: It cannot be imported, AFAICT. I added a simple user agent regex.

iamacook: `getDevice` is indeed exported from a declaration file. You [can use `bowser` directly](https://github.com/safe-global/web-core/blob/dev/src/services/pairing/connector.ts#L32-L34) but the regex seems sufficient for now.

It's somewhat out of scope but we can disable hardware wallets directly, e.g. [Trezor](https://github.com/blocknative/web3-onboard/blob/v2-web3-onboard-develop/packages/trezor/src/index.ts#L26) and [Ledger](https://github.com/blocknative/web3-onboard/blob/v2-web3-onboard-develop/packages/ledger/src/index.ts#L116). We could have a standard `filter` we add to all apart from WalletConnect.

iamacook (approved):

PR #875, opened 2022-10-10
fix: add signing to pending tx states

### LINKED ISSUE #872: [Queued list] The loading state  during tx signing is not displayed after switching between the tabs
## Bug description
The loading state  during tx signing is not displayed after switching between the tabs
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: goerli

## Steps to reproduce
 1. Go to the queued list in the 2 of N safe
 2. creat tx
 3.  confirm tx as the second use and go to the history and back to queued before the "Need your confirmation" is changed to the " Need execution."
Current result: There is no "loading state, and the user is lost on what should be done next.
In some time "Need your confirmation" is updated to "Need execution" without any action from the user

## Expected result
Make sure that the loading state is displayed after the tab switch 

## Screenshots
https://drive.google.com/file/d/1sY0_hJa7k5nYw9DlRoxB-iGKDN5ndpfO/view?usp=sharing

schmanu:
## What it solves
There is a delay of a few seconds between signing a queued tx and receiving the new signature from the tx services. 
Currently the pending signature is only reflected by the signing button and not reflected in the redux store causing the loading spinner to disappear when switching views.

Resolves #872 

## How this PR fixes it
Adds a new `PendingStatus.SIGNING` which receives the pending `signerAddress` as payload.
The `txQueueSlice` middleware will update the pending status once the pending signature is returned by the tx service queue.

## How to test it
1. Queue any tx is a 2/n safe
2. add the second signature without executing the tx
3. witness the new "Signing" status which is persisted when switching to the history tab and back

## Analytics changes
None

## Screenshots
pendingSignature

### DIFF
--- a/src/hooks/useTransactionStatus.ts
+++ b/src/hooks/useTransactionStatus.ts
@@ -20,6 +20,7 @@ const STATUS_LABELS: Record<TxLocalStatus, string> = {
   [PendingStatus.SUBMITTING]: 'Submitting',
   [PendingStatus.PROCESSING]: 'Processing',
   [PendingStatus.INDEXING]: 'Indexing',
+  [PendingStatus.SIGNING]: 'Signing',
   [ReplacedStatus]: 'Transaction will be replaced',
 }
 
--- a/src/hooks/useTxPendingStatuses.ts
+++ b/src/hooks/useTxPendingStatuses.ts
@@ -58,7 +58,6 @@ const useTxPendingStatuses = (): void => {
   useEffect(() => {
     const unsubFns = Object.entries(pendingStatuses).map(([event, status]) =>
       txSubscribe(event as TxEvent, (detail) => {
-        console.log('Tx Event:', detail)
         // All pending txns should have a txId
         const txId = 'txId' in detail && detail.txId
         if (!txId) return
@@ -66,11 +65,9 @@ const useTxPendingStatuses = (): void => {
         // Clear the pending status if the tx is no longer pending
         const isFinished = status === null
         if (isFinished) {
-          console.log('Tx finished')
           dispatch(clearPendingTx({ txId }))
           return
         }
-        console.log('Tx in status', status)
 
         // Or set a new status
         dispatch(
--- a/src/hooks/useTxPendingStatuses.ts
+++ b/src/hooks/useTxPendingStatuses.ts
@@ -7,6 +7,8 @@ import { waitForTx } from '@/services/tx/txMonitor'
 import { useWeb3ReadOnly } from '@/hooks/wallets/web3'
 
 const pendingStatuses: Partial<Record<TxEvent, PendingStatus | null>> = {
+  [TxEvent.SIGNATURE_PROPOSED]: PendingStatus.SIGNING,
+  [TxEvent.SIGNATURE_INDEXED]: null,
   [TxEvent.EXECUTING]: PendingStatus.SUBMITTING,
   [TxEvent.PROCESSING]: PendingStatus.PROCESSING,
   [TxEvent.PROCESSED]: PendingStatus.INDEXING,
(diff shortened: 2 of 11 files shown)


src/hooks/useTransactionStatus.ts
```diff
@@ -20,6 +20,7 @@ const STATUS_LABELS: Record<TxLocalStatus, string> = {
   [PendingStatus.SUBMITTING]: 'Submitting',
   [PendingStatus.PROCESSING]: 'Processing',
   [PendingStatus.INDEXING]: 'Indexing',
+  [PendingStatus.SIGNING]: 'Signing',
```
schmanu, line 23: Question for reviewer: Is `Signing` a good label for this status?
Or should it be something like `Adding confirmation` or `Indexing confirmation`?

katspaugh: I think it's a good name!

src/hooks/useTxPendingStatuses.ts
```diff
@@ -75,6 +77,7 @@ const useTxPendingStatuses = (): void => {
             status,
             txHash: 'txHash' in detail ? detail.txHash : undefined,
             groupKey: 'groupKey' in detail ? detail.groupKey : undefined,
+            signerAddress: `signerAddress` in detail ? detail.signerAddress : undefined,
```
katspaugh, line 80: Why back-tick quotes?

katspaugh (approved):
Looks good!

It's basically the same logic we have for when an executed tx is being indexed, right?

iamacook (commented):
Have you tested this with on-chain signatures?

schmanu:
> Looks good!
> 
> It's basically the same logic we have for when an executed tx is being indexed, right?

Yes it's using the our `pendingTxsSlice` and similar to how we remove successfully executed transactions once they appear in the history we remove the new 'SIGNING' state once the missing signature is added.

src/services/tx/txSender.ts
```diff
@@ -217,7 +217,10 @@ export const dispatchTxProposal = async (
     throw error
   }
 
-  txDispatch(txId ? TxEvent.SIGNATURE_PROPOSED : TxEvent.PROPOSED, { txId: proposedTx.txId })
+  txDispatch(txId ? TxEvent.SIGNATURE_PROPOSED : TxEvent.PROPOSED, {
+    txId: proposedTx.txId,
+    signerAddress: txId ? sender : undefined,
```
iamacook, line 222: Should we not always dispatch with `sender`?

francovenica:
Issue:
For safes 3/3 or higher, during the "signing" status the amount of signatures goes to 3/3 even if that signature is just the 2nd one, so it should show 2/3 during that signature
Tested this on the main branch and doens't happen there, so is just this PR.

signing

schmanu:
> > Issue: For safes 3/3 or higher, during the "signing" status the amount of signatures goes to 3/3 even if that signature is just the 2nd one, so it should show 2/3 during that signature Tested this on the main branch and doens't happen there, so is just this PR.
> 
> Imo, this should show 1/3 whilst "Signing" until the backend confirmed the user is no longer a `missingSigner`, then show 2/3.

I didn 't really touch that logic but with the new state it would be very consistent to only show the confirmations once they are reflected by the tx service. I will look into it.

schmanu:
> > Issue: For safes 3/3 or higher, during the "signing" status the amount of signatures goes to 3/3 even if that signature is just the 2nd one, so it should show 2/3 during that signature Tested this on the main branch and doens't happen there, so is just this PR.
> 
> Imo, this should show 1/3 whilst "Signing" until the backend confirmed the user is no longer a `missingSigner`, then show 2/3.

@iamacook I removed the the overwrites for the number of confirmations to remove this bug.

francovenica:
Looks good now. The amount of signatures updates after the singing status is done

PR #877, opened 2022-10-10
Feat: contract interaction (TxBuilder)

### LINKED ISSUE #876: Replacing the contract interaction feature with the transaction builder
## Overview

The transaction builder has overtaken the contract interaction feature in UX and feature-depth with all its latest updates and since we do not want to maintain both in the future, it is time to sunset the contract interaction feature. 

## Requirements

- [x] link button to the transaction builder Safe App instead of opening the contract interaction module 

## Designs

- [x] design: replace <> symbol with transaction builder logo @liliiaorlenko wdyt?

katspaugh:
## What it solves

Resolves #876 

## How this PR fixes it

Adds a Contract interaction button with the TxBuilder icon and link.

## Screenshots
Screenshot 2022-10-10 at 19 05 22

### DIFF
--- a/src/components/dashboard/FeaturedApps/FeaturedApps.tsx
+++ b/src/components/dashboard/FeaturedApps/FeaturedApps.tsx
@@ -7,8 +7,7 @@ import { useRouter } from 'next/router'
 import { useRemoteSafeApps } from '@/hooks/safe-apps/useRemoteSafeApps'
 import NextLink from 'next/link'
 import { AppRoutes } from '@/config/routes'
-
-export const FEATURED_APPS_TAG = 'dashboard-widgets'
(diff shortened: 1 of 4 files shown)


mikhailxyz (approved):

francovenica:
Before reporting
The "Send" button in the AB also shows the "Tx builder", but it should only be available for Contract addresses, and not wallets. Also it used to autofill the address you click "Send" from in the recipient input.

Idk if any of those things are possible. At least I think we can just remove the tx builder button if it is accessed from the AB

katspaugh:
@francovenica it's not possible to pass an address to the app, so I removed the button in AB, like you suggested. 👍

katspaugh:
I'll merge this because it blocks a couple other Safe Apps-related PRs.

PR #890, opened 2022-10-12
Fix: rm QR camera facing mode

katspaugh:
## What it solves

@francovenica:
> For mobile. When you try to use the QR code feature the system will turn on your frontal camera and there is no way to change it to the back camera, making it really hard to read a QR code that way. I'd like to discuss it here first since idk if there is anything we can do about it

I've removed `facingMode: user` so that it choses whichever camera is the default.


katspaugh:
Tested on the iPhone and MacBook ✅ 
The phone now uses the back camera.

iamacook:
Can confirm that this also works on Android as well.

PR #891, opened 2022-10-12
Fix: hide error details in notifications

### LINKED ISSUE #884: Too much info in the error notification
## Bug description
Too much info in the error notification.

Screenshot 2022-10-11 at 17 41 16

## Suggestion
Hide the JSON trace under a collapsible "Details" section, just like the inline `<ErrorMessage>` does.

If it's an Ethers.js error, it has a `reason` field which should be still displayed uncollapsed, but the initial letter needs to be capitalized.

So in the end it should look like this:

```
ℹ️ Your transaction was unsuccessful.
   User rejected the transaction.

   View transaction
```

katspaugh:
## What it solves

Resolves #884

## How this PR fixes it
Screenshot 2022-10-12 at 12 01 56
Screenshot 2022-10-12 at 12 02 00

### DIFF
--- a/src/components/tx/ErrorMessage/index.tsx
+++ b/src/components/tx/ErrorMessage/index.tsx
@@ -3,7 +3,6 @@ import { Link, Typography, SvgIcon } from '@mui/material'
 import WarningIcon from '@/public/images/notifications/warning.svg'
 import css from './styles.module.css'
 
-// TODO: Replace with notification
 const ErrorMessage = ({
   children,
   error,
--- a/src/components/common/Notifications/index.tsx
+++ b/src/components/common/Notifications/index.tsx
@@ -45,6 +45,7 @@ export const NotificationLink = ({
 
 const Toast = ({
   message,
+  detailedMessage,
   variant,
   link,
   onClose,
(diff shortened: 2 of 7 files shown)


src/components/tx/ErrorMessage/index.tsx
```diff
@@ -3,7 +3,6 @@ import { Link, Typography, SvgIcon } from '@mui/material'
 import WarningIcon from '@/public/images/notifications/warning.svg'
 import css from './styles.module.css'
 
-// TODO: Replace with notification
```
iamacook, line 6: Out of scope but I think it's worthwhile showing these as notifications. The `ErrorMessage` toast seems to be a duplicate of our notification system. It even handles the `reason` property. Wdyt?

iamacook (approved):

iamacook:
Do we want to display the `detailedMessage` in the notification centre? It looks good without it imo:

image

Nit: I noticed that there's no full stop after the capitalised `reason`.

src/components/tx/ErrorMessage/index.tsx
katspaugh: Sometimes it's redundant, but it also handles non-tx-lifecycle errors in the modal (e.g. gas limit).
I think it's an advantage that it's inline, because it gives the error an immediate context.

iamacook: Is the toast specified anywhere in the design? It's not related to this ticket but it would be good to align with them, cc @liliiaorlenko

iamacook: I've looked deeper into this and I take back my comment. I was getting confused with the error toast from NextJS.

PR #892, opened 2022-10-12
Fix: encode executor sig only if owner

### LINKED ISSUE #886: Tx signing - Error estimating gas for non-owner tx execution
## Bug description
This issue was initially reported in this ticket https://github.com/safe-global/web-core/pull/863.

When a non-owner wants to execute a tx the gas estimation doesn't work and throws an error.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Goerli
 - Env: https://error-messages.web-core.pages.dev/gor:0x0Bf3C9DA222Fb3e77Ea66C5D04130d35F6eEf9F4/transactions/tx?id=multisig_0x0Bf3C9DA222Fb3e77Ea66C5D04130d35F6eEf9F4_0xf033b4511064a358a2a9d2a391571da7937c977aac2a4d6406f73840b637cefb

## Steps to reproduce
 1. With an owner, create a tx in a safe 1/x and queue it
 2. Switch to a non-owner and try to execute the tx.

## Expected result
The owner excutes the tx

## Obtained result
The estimation cannot be done, an error shows up in the modal

## Screenshots
image

katspaugh:
## What it solves

Resolves #886

## How this PR fixes it

The problem was that we were adding the executing wallet's address as a "fake" signature.
It should only add the "fake" signature if the wallet is an owner and their signature is still missing.

## How to test it
Try to execute a tx as a non-owner – gas estimation should work

### DIFF
--- a/src/hooks/useGasLimit.ts
+++ b/src/hooks/useGasLimit.ts
@@ -8,6 +8,7 @@ import { useWeb3ReadOnly } from '@/hooks/wallets/web3'
 import useSafeAddress from './useSafeAddress'
 import useWallet from './wallets/useWallet'
 import { useSafeSDK } from './coreSDK/safeCoreSDK'
+import useIsSafeOwner from './useIsSafeOwner'
 
 export const getPreValidatedSignature = (from: string): string => {
   return `0x000000000000000000000000${from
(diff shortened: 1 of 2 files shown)


src/hooks/useGasLimit.ts
```diff
@@ -8,20 +8,21 @@ import { useWeb3ReadOnly } from '@/hooks/wallets/web3'
...
-  const owner = from.toLowerCase()
-  const needsOwnerSig = !safeTx.signatures.has(owner)
+export const _encodeSignatures = (safeTx: SafeTransaction, from?: string): string => {
+  const owner = from?.toLowerCase()
+  const needsOwnerSig = owner && !safeTx.signatures.has(owner)
 
   // https://docs.gnosis.io/safe/docs/contracts_signatures/#pre-validated-signatures
   if (needsOwnerSig) {
-    const ownerSig = getPreValidatedSignature(from)
+    const ownerSig = getPreValidatedSignature(owner)
```
iamacook, line 25: You can use `generatePreValidatedSignature` from the `safe-core-sdk` here. We then don't even need `getPreValidatedSignature` in the app, i.e.

```ts
if (needsOwnerSig) {
  const signature = generatePreValidatedSignature(owner)
  safeTx.addSignature(signature)
}
```

iamacook (approved):

src/hooks/useGasLimit.ts
katspaugh: Good call, didn't know this fn existed!

PR #893, opened 2022-10-12
fix: improve `EnhancedTable` + hide send on mobile

### LINKED ISSUE #888: Mobile viewport - Being able to scroll horizontally in tables is not intuitive
## Bug description
Tables can be scrolled horizontally to see all the columns, but is not obvious, there is nothing that tells the user he can do that, so it looks broken

## Environment
 - Browser: Chrome
 - Env: this safe has assets and spending limits created: https://web-core.pages.dev/rin:0x9913B9180C20C6b0F21B6480c84422F6ebc4B808/settings/spending-limits

## Steps to reproduce
 1. Go to This safe

## Expected result
There should be a visible scroll bar or something that tells the user that horizontal scrolling is possible

## Obtained result
Without any indication that horizontal scrolling is possible in tables, it looks broken

## Screenshots
image
image

iamacook:
## What it solves

Resolves #888

## How this PR fixes it

- The "Send" button in the assets table is hidden when `!isGranted || isMobile`
- If actions are hidden, i.e. because the user is not granted, the header is now hidden as well. `EnhancedTable` now hides them via CSS instead of JS.

## How to test it

Open the Safe on mobile and observe that the assets "Send" button is never visible, including when connected. All other tables according to the following conditions:

- Address book: always, (no "Send" on mobile)
- Owners: edit always, others when granted
- Spending limit: when granted

## Screenshots

image

image

image

image


src/components/balances/AssetsTable/index.tsx
```diff
@@ -13,38 +13,46 @@ import useIsSafeOwner from '@/hooks/useIsSafeOwner'
...
-    label: '',
-    width: '20%',
-    sticky: true,
-  },
-]
-
 const AssetsTable = ({ items }: AssetsTableProps): ReactElement => {
   const [selectedAsset, setSelectedAsset] = useState<string | undefined>()
   const isSafeOwner = useIsSafeOwner()
+  const isMobile = useIsMobile()
```
katspaugh, line 25: Can't you just hide the last column via CSS?

iamacook: There are [two dependent flags for hiding the actions](https://github.com/safe-global/web-core/blob/improve-enhanced-tables/src/components/balances/AssetsTable/index.tsx#L24-L27) so I thought keeping them together made the most sense. The `EnhancedTable` then hides it via CSS [here](https://github.com/safe-global/web-core/blob/improve-enhanced-tables/src/components/common/EnhancedTable/index.tsx#L77) and [here](https://github.com/safe-global/web-core/blob/improve-enhanced-tables/src/components/common/EnhancedTable/index.tsx#L144).

katspaugh: They are not dependent. A user can be a non-owner on both desktop and mobile.

katspaugh (approved):
👍

PR #906, opened 2022-10-14
fix: replace "New transaction" button for "Switch to ..." button if on wrong chain

### LINKED ISSUE #897: [Mobile] - Change to current network not intuitive
## Bug description
(technically not a bug, more of a suggestion)
In desktop, if you are in the wrong network we have a big button in the topbar that asks you to switch, but in mobile is only if you open the Connect display that you see the button, so is pretty hidden
This is an issue specially with the "New Transaction" button that only says "wrong network"

Suggestion: Make that New Transaction button to also have the feature of switching to the expected network.

image
image
image

DiogoSoaress:
## What it solves

Resolves #897

## How this PR fixes it
If the connected wallet is in a different chain that the opened Safe, the "Switch to ..." button is displayed instead of the "New transaction" button.

## How to test it
1. Connect your wallet on Goerli
2. Open a Ethereum Safe
3. The "New transaction" button is replaced by the "Switch to ..." button
4. "Switch to ..." button is removed from the Header

## Screenshots
_Mobile resolution_
Screenshot 2022-10-14 at 10 31 20
_Web_
Screenshot 2022-10-14 at 10 31 52

### DIFF
--- a/src/components/common/Header/index.tsx
+++ b/src/components/common/Header/index.tsx
@@ -4,6 +4,7 @@ import { IconButton, Paper } from '@mui/material'
 import MenuIcon from '@mui/icons-material/Menu'
 import classnames from 'classnames'
 import css from './styles.module.css'
+import ChainSwitcher from '@/components/common/ChainSwitcher'
 import ConnectWallet from '@/components/common/ConnectWallet'
 import NetworkSelector from '@/components/common/NetworkSelector'
 import SafeTokenWidget, { getSafeTokenAddress } from '@/components/common/SafeTokenWidget'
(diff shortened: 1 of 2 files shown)


src/components/common/Header/index.tsx
```diff
@@ -42,10 +41,6 @@ const Header = ({ onMenuToggle }: HeaderProps): ReactElement => {
         </Link>
       </div>
 
-      <div className={classnames(css.element, css.hideMobile)}>
-        <ChainSwitcher />
-      </div>
```
katspaugh, line 47: I would keep it here as well, more prominent.

katspaugh (approved):

PR #907, opened 2022-10-14
Fix: timeout in safe creation

### LINKED ISSUE #905: [Safe creation] Aurora - error during Safe creation
## Bug description
When you create a safe in aurora, almost immediately you get a message that your safe wasn't created properly. 
After a few seconds it tells you that the safe has been created and is being indexed. The safe ends up being created sucessfully, but in the history tab the "Safe creation" tx is missing

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Aurora
 - Env: https://web-core.pages.dev/

## Steps to reproduce
 1. Go to the app in aurora
 2. Create a safe 
 3. Go to the history tab once created

## Expected result
During the safe creation no errors are shown
in the history tab the "Safe creation" tx is present

Safes:
https://web-core.pages.dev/aurora:0x9e4e47F2C383499d35DF6478FFB5fCee16E19EDE/transactions/history
https://web-core.pages.dev/aurora:0x7a0B114660b74a27C029f08B3acd14460092C542/transactions/history

## Screenshots
image
image

katspaugh:
## What it solves

Resolves #905

## How this PR fixes it
Pending Safe creation was failing early because of a poorly handled error.

I've added a condition that falls back to a timeout when that unexpected error happens.

## How to test it
Create a Safe on Aurora.

### DIFF
--- a/src/components/create-safe/status/usePendingSafeCreation.ts
+++ b/src/components/create-safe/status/usePendingSafeCreation.ts
@@ -10,6 +10,7 @@ import { getSafeInfo } from '@gnosis.pm/safe-react-gateway-sdk'
 import { getProxyFactoryContractInstance } from '@/services/contracts/safeContracts'
 import useChainId from '@/hooks/useChainId'
 import { ErrorCode } from '@ethersproject/logger'
+import { Errors, logError } from '@/services/exceptions'
 
 export const pollSafeInfo = async (chainId: string, safeAddress: string): Promise<SafeInfo> => {
   // exponential delay between attempts for around 4 min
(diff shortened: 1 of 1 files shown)


src/components/create-safe/status/usePendingSafeCreation.ts
```diff
@@ -24,11 +25,24 @@ export const pollSafeInfo = async (chainId: string, safeAddress: string): Promis
...
+  provider: JsonRpcProvider,
+  txHash: string,
+  chainId: string,
+): Promise<SafeCreationStatus> => {
+  const TIMEOUT_TIME = 6.5 * 60 * 1000 // 6.5 minutes
 
   try {
     const txResponse = await provider.getTransaction(txHash)
+
+    if (!txResponse) {
```
katspaugh, line 38: It was previously throwing because `txResponse` was undefined, and it assumed the Safe creation had failed.

Now it will return a timeout instead.

DiogoSoaress: makes sense

DiogoSoaress (approved):
Looks good! I don't have funds on Aurora to test though.

PR #908, opened 2022-10-14
Fix: display allocated SAFE amount

katspaugh:
## What it solves
Displays SAFE token allocation properly.
Ported from https://github.com/safe-global/safe-react/pull/4081.

### DIFF
--- a/cypress/e2e/smoke/address_book.cy.js
+++ b/cypress/e2e/smoke/address_book.cy.js
@@ -4,7 +4,6 @@ import { format } from 'date-fns'
 
 const NAME = 'Owner1'
 const EDITED_NAME = 'Edited Owner1'
-const ENS_NAME = 'test20.eth'
 const ENS_ADDRESS = '0xE297437d6b53890cbf004e401F3acc67c8b39665'
 const GOERLI_TEST_SAFE = 'gor:0x97d314157727D517A706B5D08507A1f9B44AaaE9'
 const GNO_TEST_SAFE = 'gno:0xB8d760a90a5ed54D3c2b3EFC231277e99188642A'
@@ -30,8 +29,7 @@ describe('Address book', () => {
       // Add a new entry manually
       cy.contains('Create entry').click()
       cy.get('input[name="name"]').type(NAME)
-      cy.get('input[name="address"]').type(ENS_NAME)
-      // cy.wait(5000)
+      cy.get('input[name="address"]').type(`gor:${ENS_ADDRESS}`)
       cy.contains('button', 'Save').click()
 
       cy.contains(NAME).should('exist')
--- a/src/components/common/SafeTokenWidget/__tests__/SafeTokenWidget.test.tsx
+++ b/src/components/common/SafeTokenWidget/__tests__/SafeTokenWidget.test.tsx
@@ -1,24 +1,22 @@
-import { SAFE_TOKEN_ADDRESSES } from '@/config/constants'
 import * as useBalances from '@/hooks/useBalances'
 import * as nextRouter from 'next/router'
 import useChainId from '@/hooks/useChainId'
 import { render, waitFor } from '@/tests/test-utils'
-import { SafeAppAccessPolicyTypes, TokenType } from '@gnosis.pm/safe-react-gateway-sdk'
-import { ethers } from 'ethers'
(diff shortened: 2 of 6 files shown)


src/components/common/SafeTokenWidget/index.tsx
```diff
@@ -44,10 +44,7 @@ const SafeTokenWidget = () => {
       }
     : undefined
 
-  const safeBalance = balances.balances.items.find((balanceItem) => balanceItem.tokenInfo.address === tokenAddress)
-
-  const safeBalanceDecimals = Number(safeFormatUnits(safeBalance?.balance || 0, safeBalance?.tokenInfo.decimals))
-  const flooredSafeBalance = formatAmountWithPrecision(safeBalanceDecimals, 2)
+  const flooredSafeBalance = formatVisualAmount(allocation, DECIMALS)
```
DiogoSoaress, line 47: Don't we want to keep the 2 decimals precision? It broke the tests

DiogoSoaress (changes requested):

src/components/common/SafeTokenWidget/index.tsx
katspaugh: It's the default, and it's not what broken them.

DiogoSoaress (approved):

cypress/e2e/smoke/address_book.cy.js
```diff
@@ -30,8 +29,7 @@ describe('Address book', () => {
       // Add a new entry manually
       cy.contains('Create entry').click()
       cy.get('input[name="name"]').type(NAME)
-      cy.get('input[name="address"]').type(ENS_NAME)
-      // cy.wait(5000)
+      cy.get('input[name="address"]').type(`gor:${ENS_ADDRESS}`)
```
DiogoSoaress, line 32: This seems to be a bug in the form. I was investigating on it but I will fill in an issue

francovenica:
The widget is there and it shows the correct value

Some details (I can move these into a new ticket necessary)

- The widget is not clickable, maybe because the app is not available in this PR
- The number has more than 2 decimals. In prod is capped at 2 decimals.
- Is not visible for a mobile view anywhere. I think we should show it somewhere for a while until more people claim their tokens

src/components/common/SafeTokenWidget/index.tsx
katspaugh: @DiogoSoaress you were actually right, it needs the precision, as Franco also pointed out. I've passed it explicitly.

katspaugh:
@francovenica I've fixed the decimals.
The claiming app isn't tagged on staging, please use the prod toggle.
Regarding mobile, there's no place for it in the header. I will be solved in near future by a dedicated Dashboard widget.

src/components/common/SafeTokenWidget/__tests__/SafeTokenWidget.test.tsx
```diff
@@ -81,50 +65,21 @@ describe('SafeTokenWidget', () => {
...
-      error: undefined,
-    }))
+  it('Should display 0 if Safe has no SAFE token', async () => {
+    ;(useSafeTokenAllocation as jest.Mock).mockImplementation(() => BigNumber.from(0))
 
     const result = render(<SafeTokenWidget />)
     await waitFor(() => expect(result.baseElement).toHaveTextContent('0'))
   })
 
   it('Should display the value formatted correctly', async () => {
```
DiogoSoaress, line 75: This test should cover the precision nuance.
I.e, without explicitly passing the precision it renders with a wrong number of decimals **and** when explicitly passing the precision, it formats the number according to that

DiogoSoaress (changes requested):

src/components/common/SafeTokenWidget/__tests__/SafeTokenWidget.test.tsx
katspaugh: I've set [a better mock value](https://github.com/safe-global/web-core/pull/908/commits/55a334fe0f2c1d615777b48873f10f06428a5c34#diff-cfe34aabfe1308977e7ecf7c7128682037147b447afa6badc9035bf46c10f983R76) for the allocation amount, that results in 4 decimals if formatted badly.

DiogoSoaress (approved):

PR #913, opened 2022-10-14
Fix: update the WC bridge URL

katspaugh:
## What it solves

A new WC URL.


DiogoSoaress (approved):
Same changes as in `safe-react` 👌

mikhailxyz (approved):

PR #918, opened 2022-10-17
Refactor: clean up unused error codes

katspaugh:
## What it solves
There were some error codes left over from safe-react that we don't log anymore (largely because they are displayed in the UI).


DiogoSoaress (approved):
Nice cleanup!

PR #920, opened 2022-10-17
fix: trigger form validate when resolving ENS

### LINKED ISSUE #909: Address Book - Create entry form does not validate after translating ENS
## Bug description
When adding an entry to the Address Book, the form is invalid right after translating the ENS.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Goerli

## Steps to reproduce
 1. Go to Address Book
 2. Click to create an entry
 3. Insert `safe20.eth` in the address
 4. Observe the form is invalid at first

## Expected result
The form is valid after translating the ENS name

The E2E test 'should add an entry' works with a ENS name

## Obtained result
The form shows an error and "Save" button is disabled

## Screenshots
Screenshot 2022-10-14 at 11 21 00

DiogoSoaress:
## What it solves

Resolves #909 

## How this PR fixes it
Triggers a form validation when the input changes. This will catch the case where ENS resolves and the form validation was not triggered

## How to test it
1. Go to AB > Create entry
2. Insert a ENS name
3. Form should be valid when the name is resolved

## Analytics changes
N/A

## Screenshots
Screenshot 2022-10-17 at 16 39 42

### DIFF
--- a/src/components/address-book/EntryDialog/index.tsx
+++ b/src/components/address-book/EntryDialog/index.tsx
@@ -3,6 +3,7 @@ import Button from '@mui/material/Button'
 import DialogActions from '@mui/material/DialogActions'
 import DialogContent from '@mui/material/DialogContent'
 import type { ReactElement } from 'react'
+import { useEffect } from 'react'
 import { FormProvider, useForm } from 'react-hook-form'
 
 import AddressInput from '@/components/common/AddressInput'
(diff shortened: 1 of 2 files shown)


src/components/address-book/EntryDialog/index.tsx
```diff
@@ -38,7 +39,16 @@ const EntryDialog = ({
...
+
+  const { handleSubmit, formState, watch, trigger } = methods
+
+  const address = watch('address')
+
+  useEffect(() => {
+    if (address) {
+      trigger('address')
+    }
+  }, [address, trigger])
```
DiogoSoaress, line 51: I was trying to avoid this use effect but RHF listens only to onChange and onBlur and doesn't seem to provide another way to trigger the validation. See https://react-hook-form.com/api/useform

katspaugh (approved):
👍

src/components/address-book/EntryDialog/index.tsx
usame-algan, line 51: I would suggest doing the revalidation inside the `AddressInput` component with the `shouldValidate` option when setting the value.

```
const setAddressValue = useCallback(
    (value: string) => setValue(name, value, { shouldValidate: true }),
    [setValue, name],
  )
```

usame-algan (changes requested):

usame-algan (approved):
👍

PR #922, opened 2022-10-18
Fix: pinned Safe Apps per chain

### LINKED ISSUE #814: [Pinned safe apps] Pins performed globally instead of per chain
## Description
We should keep the same behaviour we had in legacy for pinned apps where the users would see their pinned apps per chain and not globally.

## Steps to reproduce
1. Pin a couple of safe apps on the revamp and check local storage.

2. Do the same in legacy for comparison.

## Expected result
Pinned apps stored per chain:
Screenshot 2022-10-05 at 11 32 52

## Obtained result
Pinned apps are stored globally:

Screenshot 2022-10-05 at 11 44 04

katspaugh:
## What it solves

Resolves #814 

## How this PR fixes it
Makes the pinned apps slice an object per chain.

## How to test it
Pin apps on different chain, they should be independent.

## Screenshots
Screenshot 2022-10-18 at 09 22 59


src/components/dashboard/FeaturedApps/FeaturedApps.tsx
```diff
@@ -23,14 +23,14 @@ const StyledGridItem = styled(Grid)`
 `
 
 export const FeaturedApps = (): ReactElement | null => {
-  const [allApps = [], , isLoading] = useRemoteSafeApps()
```
mikhailxyz, line 26: Why replace this with `useSafeApps`? It doesn't look like any non-remote apps are used

mikhailxyz (approved):

src/components/dashboard/FeaturedApps/FeaturedApps.tsx
katspaugh: I grepped, and it was the only place where this hook was used apart from `useSafeApps`.
I think it's better if `useSafeApps` is the only entry to safe apps, and the remote apps hook should remain as a raw fetcher.

mikhailxyz: > I grepped, and it was the only place where this hook was used apart from `useSafeApps`. I think it's better if `useSafeApps` is the only entry to safe apps, and the remote apps hook should remain as a raw fetcher.

`useSafeApps` wasn't meant to be the only entry point, it just happens that most of the places need apps from multiple sources 🤷 If you know that you only need the remote apps, in my opinion, it's better to use the raw hook because `useSafeApps` also requests other resources which are unnecessary in this case

PR #923, opened 2022-10-18
Fix: remove filetype check in AB import (breaks Android)

### LINKED ISSUE #889: [Android] - Uploading a csv in the AB errors out and displays [Object object]
## Bug description
Trying to upload a csv in the Address book doesn't work.
Tested it in Android

## Environment
 - Browser: Chrome

## Steps to reproduce
 1. Need to have downloaded a csv in your phone already. Export csv feature works fine, you can use that
 2. Try to upload the csv

## Expected result
The AB gets updated

## Obtained result
An error is displayed.

katspaugh:
## What it solves

Resolves #889

## How this PR fixes it
Removes the extraneous content type validation.

## How to test it
Import an exported Safe address book on Android.

### DIFF
--- a/src/components/address-book/ImportDialog/validation.ts
+++ b/src/components/address-book/ImportDialog/validation.ts
@@ -2,11 +2,7 @@ import type { ParseResult } from 'papaparse'
 
 import { validateAddress, validateChainId } from '@/utils/validation'
 
-export const abCsvReaderValidator = ({ type, size }: File): string[] | undefined => {
-  if (type !== 'text/csv') {
-    return ['Address book must be a CSV file']
-  }
-
+export const abCsvReaderValidator = ({ size }: File): string[] | undefined => {
   if (size > 1_000_000) {
     return ['Address book cannot be larger than 1MB']
   }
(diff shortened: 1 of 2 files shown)


katspaugh:
@francovenica please see if this branch works for you, I don't have an Android to test, but hope this helps.

src/components/address-book/ImportDialog/validation.ts
```diff
@@ -2,11 +2,7 @@ import type { ParseResult } from 'papaparse'
 
 import { validateAddress, validateChainId } from '@/utils/validation'
 
-export const abCsvReaderValidator = ({ type, size }: File): string[] | undefined => {
-  if (type !== 'text/csv') {
-    return ['Address book must be a CSV file']
-  }
-
+export const abCsvReaderValidator = ({ size }: File): string[] | undefined => {
```
DiogoSoaress, line 5: This is not a fix, right? Either we investigate deeper or I'd say let's not fix this.

katspaugh: Let's see what Franco says. If it fixes it, I'd say it's better not to do the validation that gives false negatives on some OSes.

DiogoSoaress: Ok, if we are having false negatives in some OSes. Waiting for Franco on this.

francovenica:
I tried several times with my android and I always get an [Object Object] error. I tried the same csv using chrome in PC windows and it works fine, so the .csv is a valid one

Tried to debug with the dev console but it shows no errors there

image

katspaugh:
OK, I'll close this PR then and we'll debug properly on Android or an emulator.
Thanks @francovenica

PR #924, opened 2022-10-18
fix: Cache transaction info on safe creation

### LINKED ISSUE #917: [Safe Creation] Instant Timeout error is displayed
Related to https://github.com/safe-global/web-core/issues/905

When creating a safe we poll the transaction for 6.5 minutes via the `waitForTransaction` call. Since we also want to check for a replacement transaction we are calling `getTransaction` beforehand to get the necessary info. However, this can return a null response if the transaction hasn't been picked up yet by the provider which then displays an immediate TIMEOUT error even before the 6.5 minutes have passed.

## Steps to reproduce

- The `getTransaction` RPC call needs to be blocked/mocked to return null
- Observe the TIMEOUT error is displayed immediately

## Expected behaviour

The TIMEOUT error should only be displayed if the transaction can not be found within the time limit

## Current behaviour

The TIMEOUT error is immediately displayed if the `getTransaction` call returns null.

## ToDos

- [x] Cache tx related info to use for `_waitForTransaction`
- [x] Write more tests for `checkSafeCreationTx`

usame-algan:
## What it solves

Resolves #917 

## How this PR fixes it

- Caches relevant info of the safe creation tx for the `_waitForTransaction` call to detect a replacement tx
- Only watches an ongoing tx if the original promise is lost (after a page reload)

### Other changes

- Adds the fallback handler address to the safe creation call, not sure about the impact but we forgot to port it from safe-react

## How to test it

- Safe creation on different chains should work as before and there are no instant timeout errors anymore

### DIFF
--- a/src/components/create-safe/status/hooks/useWatchSafeCreation.ts
+++ b/src/components/create-safe/status/hooks/useWatchSafeCreation.ts
@@ -68,7 +68,7 @@ const useWatchSafeCreation = ({
       setPendingSafe(undefined)
 
       // Asynchronously wait for Safe creation
-      if (safeAddress && pendingSafe) {
+      if (safeAddress) {
         pollSafeInfo(chainId, safeAddress)
           .then(() => setStatus(SafeCreationStatus.INDEXED))
           .catch(() => setStatus(SafeCreationStatus.INDEX_FAILED))
@@ -81,7 +81,7 @@ const useWatchSafeCreation = ({
       status === SafeCreationStatus.TIMEOUT
     ) {
       if (pendingSafe?.txHash) {
-        setPendingSafe((prev) => (prev ? { ...prev, txHash: undefined } : undefined))
+        setPendingSafe((prev) => (prev ? { ...prev, txHash: undefined, tx: undefined } : undefined))
       }
     }
   }, [router, safeAddress, setPendingSafe, status, pendingSafe, setStatus, chainId])
--- a/src/components/create-safe/status/hooks/useWatchSafeCreation.ts
+++ b/src/components/create-safe/status/hooks/useWatchSafeCreation.ts
@@ -6,7 +6,7 @@ import { type UrlObject } from 'url'
 import { pollSafeInfo } from '@/components/create-safe/status/usePendingSafeCreation'
 import { AppRoutes } from '@/config/routes'
 import { SafeCreationStatus } from '@/components/create-safe/status/useSafeCreation'
-import { trackEvent, CREATE_SAFE_EVENTS, SAFE_APPS_EVENTS } from '@/services/analytics'
+import { CREATE_SAFE_EVENTS, SAFE_APPS_EVENTS, trackEvent } from '@/services/analytics'
 import chains from '@/config/chains'
 
 const getRedirect = (chainId: string, safeAddress: string, redirectQuery?: string | string[]): UrlObject | string => {
@@ -75,7 +75,11 @@ const useWatchSafeCreation = ({
       }
     }
 
-    if (status === SafeCreationStatus.ERROR || status === SafeCreationStatus.REVERTED) {
+    if (
+      status === SafeCreationStatus.ERROR ||
+      status === SafeCreationStatus.REVERTED ||
+      status === SafeCreationStatus.TIMEOUT
+    ) {
       if (pendingSafe?.txHash) {
         setPendingSafe((prev) => (prev ? { ...prev, txHash: undefined } : undefined))
       }
--- a/src/components/create-safe/status/usePendingSafeCreation.ts
+++ b/src/components/create-safe/status/usePendingSafeCreation.ts
@@ -25,6 +25,27 @@ export const pollSafeInfo = async (chainId: string, safeAddress: string): Promis
   })
 }
 
+export const _getTransactionByHash = (provider: JsonRpcProvider, txHash: string) => {
+  return provider.getTransaction(txHash).then((resp) => {
+    if (resp === null) {
+      throw new Error('Transaction not found')
+    }
+    return resp
+  })
+}
+
+const pollTransaction = async (provider: JsonRpcProvider, txHash: string) => {
+  return backOff(() => _getTransactionByHash(provider, txHash), {
+    startingDelay: 750,
+    maxDelay: 10000,
+    numOfAttempts: 10,
+    retry: (e) => {
+      console.info('waiting for transaction', e)
+      return true
+    },
+  })
+}
+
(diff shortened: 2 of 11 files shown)


src/components/create-safe/status/usePendingSafeCreation.ts
```diff
@@ -33,15 +54,7 @@ export const checkSafeCreationTx = async (
...
-    const txResponse = await provider.getTransaction(txHash)
-
-    if (!txResponse) {
-      return new Promise((resolve) => {
-        setTimeout(() => {
-          resolve(SafeCreationStatus.TIMEOUT)
-        }, TIMEOUT_TIME)
-      })
-    }
+    const txResponse = await pollTransaction(provider, txHash)
```
usame-algan, line 57: Not an ideal solution to poll `getTransaction` imo since `waitForTransaction` also calls `getTransaction` internally. Alternatively, we could encode and cache the original tx info on the client side and reuse that to build the `replacement` object but it requires a bigger change.

src/components/create-safe/status/usePendingSafeCreation.ts
```diff
@@ -25,6 +25,27 @@ export const pollSafeInfo = async (chainId: string, safeAddress: string): Promis
   })
 }
 
+export const _getTransactionByHash = (provider: JsonRpcProvider, txHash: string) => {
+  return provider.getTransaction(txHash).then((resp) => {
+    if (resp === null) {
```
katspaugh, line 30: I start to dislike ethers 😅

src/components/create-safe/status/usePendingSafeCreation.ts
```diff
@@ -25,6 +25,27 @@ export const pollSafeInfo = async (chainId: string, safeAddress: string): Promis
...
+    }
+    return resp
+  })
+}
+
+const pollTransaction = async (provider: JsonRpcProvider, txHash: string) => {
+  return backOff(() => _getTransactionByHash(provider, txHash), {
+    startingDelay: 750,
+    maxDelay: 10000,
+    numOfAttempts: 10,
```
katspaugh, line 41: Based on these settings, what's the maximum wait in seconds?

katspaugh (approved):

src/components/create-safe/status/usePendingSafeCreation.ts
usame-algan: Yep, the function type is definitely an issue as it can return `null`, `undefined` or `Promise<TransactionResponse>` but its only typed as `Promise<TransactionResponse>` 🤦

francovenica:
I know the ticket was moved back
Still, during testing I found that for Poly still has the problem of showing the error for a few secs and then loading the safe just fine
safe error

usame-algan:
I moved this issue back because I identified the root cause of #905 which we need to fix. Essentially, we are checking the pending transaction twice. Once with the `createSafe` call and then also with `usePendingSafeCreation` even though the hook is only meant to pick up a pending transaction if we lose the original promise e.g. if the user reloads the page after tx submission.

So I see 2 things that need to be fixed:

1. Don't run `usePendingSafeCreation` if we still have the `createSafe` promise i.e. user did not reload the page after submission/cancellation/replacement
2. Encode and cache the transaction info to be used with `_waitForTransaction` inside `usePendingSafeCreation` instead of calling `getTransaction` because as soon as the user cancels/replaces the tx in their wallet, the original `txHash` will be gone from the pool and `getTransaction` will always return `null`. Right now we let the user artificially wait for 6.5 minutes before we display a TIMEOUT error even though we are able to know that the tx was cancelled or replaced much sooner (by calling `waitForTransaction` with the cached data)

src/components/create-safe/status/hooks/useWatchSafeCreation.ts
```diff
@@ -75,7 +75,11 @@ const useWatchSafeCreation = ({
       }
     }
 
-    if (status === SafeCreationStatus.ERROR || status === SafeCreationStatus.REVERTED) {
+    if (
+      status === SafeCreationStatus.ERROR ||
+      status === SafeCreationStatus.REVERTED ||
+      status === SafeCreationStatus.TIMEOUT
```
usame-algan, line 81: Since we also display the Retry button on `TIMEOUT` it makes sense to clear the `txHash` then as well.

src/components/create-safe/useEstimateSafeCreationGas.ts
```diff
@@ -24,24 +22,21 @@ export const useEstimateSafeCreationGas = ({
...
-  const proxyContract = getProxyFactoryContractInstance(chainId)
-  const proxyContractAddress = proxyContract.getAddress()
-
-  const encodedSafeCreationTx = getSafeCreationTx({ owners, threshold, saltNonce, chain: currentChain })
-
   const [gasLimit, gasLimitError, gasLimitLoading] = useAsync<BigNumber>(() => {
-    if (!wallet?.address || !encodedSafeCreationTx || !web3ReadOnly) return
+    if (!wallet?.address || !chain || !web3ReadOnly) return
+
+    const proxyContract = getProxyFactoryContractInstance(chain.chainId)
```
katspaugh, line 31: We should rename this to `proxyFactoryContract`. Proxy contract is the Safe instance itself.

francovenica:
Issue:

Sefe creation flow fails if you speed up the tx.
The safe is created once the sped up tx is done, but the UI will still show the error message and ask you to retry.

In goerli, create a safe but in MM set the gas price and fee to 0.0001
As soon as you submit open the MM extension, find the tx and open it in the block explorer (to see the tx hash)
Open the MM extension again and speed up the tx to the max value
Open MM one last time and open the sped up tx, and open it in the block explorer (the hash value must be different)
Wait a bit and you will see that the UI tells you that the safe creation failed. 
Copy the safe address you see in the UI, it should be already created by the sped up tx
Open it in /gor:ADDRESS/home

usame-algan:
> Sefe creation flow fails if you speed up the tx.

I've adjusted the code to detect this. The following things should be working now:

- Submit tx and detect replacement
- Submit tx and detect cancellation
- Submit tx, reload and detect replacement
- Submit tx, reload and detect cancellation
- Submit tx, reload and timeout after ~6.5 minutes

Submitting a tx and detecting a timeout is not working (I've waited for around 15 minutes). The Core SDK call `deploySafe` that is running `.wait()` doesn't seem to time out. There is also an issue about this on the [ethers repository](https://github.com/ethers-io/ethers.js/issues/945). We will rework the way we watch the tx in #932 so it should be fixed there.

PR #933, opened 2022-10-18
refactor: extract reused buttons to separate components

mikhailxyz:
## What it solves
Resolves https://github.com/safe-global/web-core/issues/567

## How this PR fixes it
- It extracts buttons used in different versions of the app card component into separate components

## How to test it
- Check if the looks are identical to the production app and make sure buttons still work


katspaugh (approved):
Nice, thank you!

katspaugh:
I've tested, apps and pinned apps look the same as before. ✅

PR #935, opened 2022-10-18
fix: omit connection center in mobile devices

### LINKED ISSUE #912: [Mobile] - Remove mobile paring QR code and App store button
## Bug description
Mobile pairing is just if you are in a computer, you won't be using 2 phones to connect to the app, so remove the QR code for mobile pairing if in mobile viewport
Also remove the "App store" button if the user is on an android phone

## Environment
 - Browser: Chrome Mobile
 - Chain: Any
 - https://web-core.pages.dev/welcome

## Steps to reproduce
 1. Go to the app using your phone
 2. Open the connect display on the top right

## Expected result
Just the connect button should be present. Everything bellow "connect to mobile" should not be present

## Screenshots
image

DiogoSoaress:
## What it solves

Resolves #912

## How this PR fixes it
Renders PairingDetails conditionally on the viewport width.

## How to test it
1. Emulate a mobile viewport in the browser and observe there is no Pairing in the ConnectionCenter nor in the "connect wallet & select network" (step 1) from create Safe flow.
2. In non-mobile viewports the layout should remain the same.

## Analytics changes
N/A

## Screenshots
Screenshot 2022-10-18 at 16 55 10
Screenshot 2022-10-19 at 17 57 57

### DIFF
--- a/src/components/common/ConnectWallet/ConnectionCenter.tsx
+++ b/src/components/common/ConnectWallet/ConnectionCenter.tsx
@@ -1,4 +1,4 @@
-import { Popover, ButtonBase, Typography, Paper, Divider } from '@mui/material'
+import { Popover, ButtonBase, Typography, Paper, Divider, Box } from '@mui/material'
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
 import ExpandLessIcon from '@mui/icons-material/ExpandLess'
 import { useState, type MouseEvent, type ReactElement } from 'react'
@@ -63,11 +63,11 @@ const ConnectionCenter = (): ReactElement => {
           <WalletDetails onConnect={handleClose} />
 
           {isSupported && (
-            <>
+            <Box className={css.pairingDetails}>
               <Divider flexItem />
 
               <PairingDetails vertical />
-            </>
+            </Box>
           )}
         </Paper>
       </Popover>
--- a/src/components/common/ConnectWallet/ConnectionCenter.tsx
+++ b/src/components/common/ConnectWallet/ConnectionCenter.tsx
@@ -10,16 +10,9 @@ import PairingDetails from '@/components/common/PairingDetails'
 import css from '@/components/common/ConnectWallet/styles.module.css'
 import { useCurrentChain } from '@/hooks/useChains'
 import { isPairingSupported } from '@/services/pairing/utils'
-import AppStoreButton from '@/components/common/AppStoreButton'
-import { isAppleMobileDevice } from '@/utils/detectMobileDevices'
 
 const ConnectionCenter = (): ReactElement => {
   const chain = useCurrentChain()
-  const [showDownload, setShowDownload] = useState(false)
-
-  setTimeout(() => {
-    setShowDownload(isAppleMobileDevice())
-  })
 
   const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>(null)
   const open = !!anchorEl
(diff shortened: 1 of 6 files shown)


katspaugh (changes requested):
Please just hide the QR code with a CSS media query.
We try to avoid using JS for mobile detection.

iamacook (commented):
If you hide it in `PairingDetails` directly, you won't need to add the duplicate CSS.

DiogoSoaress:
> If you hide it in `PairingDetails` directly, you won't need to add the duplicate CSS.

I would prefer to leave this logic in the parent components. Plus, we would have to add more logic to hide the `<Divider />` in the ConnectionCenter.

francovenica:
In the issue i made the comment "Everything bellow "connect to mobile" should not be present", but actually for iPhones the "App store" button should be displayed. If it is possible to know if the user is using an android or iphone then I'd like to have that distintion.

DiogoSoaress:
> In the issue i made the comment "Everything bellow "connect to mobile" should not be present", but actually for iPhones the "App store" button should be displayed. If it is possible to know if the user is using an android or iphone then I'd like to have that distintion.

@francovenica now that I was breaking the code down, I would argue that the whole "Connect to mobile" makes more sense all together.
There was a ticket (https://github.com/safe-global/safe-react/issues/3003) that would prompt you to use the native mobile app if you'd be accessing the web app through your browser. Although unrelated with this ticket I think that implementation is better. WDYT?

francovenica:
I think removing the QR code is necessary.
For the App button I think it should be there and we could add some flavor text to it at most

src/components/common/ConnectWallet/ConnectionCenter.tsx
```diff
@@ -10,9 +10,16 @@ import PairingDetails from '@/components/common/PairingDetails'
...
+import AppStoreButton from '@/components/common/AppStoreButton'
+import { isAppleMobileDevice } from '@/utils/detectMobileDevices'
 
 const ConnectionCenter = (): ReactElement => {
   const chain = useCurrentChain()
+  const [showDownload, setShowDownload] = useState(false)
+
+  setTimeout(() => {
+    setShowDownload(isAppleMobileDevice())
+  })
```
DiogoSoaress, line 22: `navigator` was not defined in the first render so I couldn't detect the device based on `navigator.userAgent`

src/utils/detectMobileDevices.ts
```diff
@@ -0,0 +1,3 @@
+export const isMobileDevice = () => /iPhone|iPad|iPod|Android/i.test(navigator.userAgent)
+
+export const isAppleMobileDevice = () => /iPhone|iPad|iPod/i.test(navigator.userAgent)
```
katspaugh, line 3: I'm going to block this.
We should only use CSS media queries for responsiveness, and avoid checks based on the device.
I would rather have the download button visible on all devices.

katspaugh (changes requested):

src/utils/detectMobileDevices.ts
katspaugh: Yes, that'd be best 👍

DiogoSoaress: Ok, reverted to the commit you had previously approved.

DiogoSoaress:
@francovenica we are pushing ommiting the whole pairing + download button in mobile viewports. The way to call for the native iOS app download should be done through https://developer.apple.com/documentation/webkit/promoting_apps_with_smart_app_banners

francovenica:
my iphone would not let me open the env in safari
image

katspaugh (approved):

PR #937, opened 2022-10-19
Fix: space out page header tabs

katspaugh:
## What it solves

Makes the tabs a bit bigger and easier to click on.

As per @liliiaorlenko:

> Yes maybe we can widen the area for 5pm and lower the tabs 15px, may this can visually improve it

Screenshot 2022-10-19 at 07 38 57
Screenshot 2022-10-19 at 07 38 48


usame-algan (approved):
👍

iamacook (approved):

PR #940, opened 2022-10-19
fix(safe-apps): Remove the trailing slash check

yagopv:
## What it solves
Resolves https://github.com/safe-global/web-core/issues/929

## How this PR fixes it
By removing the trailingSlash check as it cleans the custom error onChange

### DIFF
--- a/src/components/safe-apps/AddCustomAppModal/index.tsx
+++ b/src/components/safe-apps/AddCustomAppModal/index.tsx
@@ -23,7 +23,7 @@ import useAsync from '@/hooks/useAsync'
 import useDebounce from '@/hooks/useDebounce'
 import { fetchSafeAppFromManifest } from '@/services/safe-apps/manifest'
 import { SAFE_APPS_EVENTS, trackEvent } from '@/services/analytics'
-import { trimTrailingSlash, isSameUrl } from '@/utils/url'
+import { isSameUrl } from '@/utils/url'
 import { AppRoutes } from '@/config/routes'
 import CustomAppPlaceholder from './CustomAppPlaceholder'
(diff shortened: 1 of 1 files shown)


katspaugh:
I vaguely recall there was a reason were trimming slashes? @mikhailxyz do you know?

yagopv:
> I vaguely recall there was a reason we were trimming slashes? @mikhailxyz do you know?

I guess it was for avoid a method call in case user adds and remove the slash. That's fine but causes this issue as we validate the manifest and when user change the url after the first error pops up it is being removed because the validation is not triggered again. 

Another concern I have is that the domain with or without trailing slash could point to a different resource (It can have subpaths) 🤔

Anyway. I fixed this without being 100% sure about if it better o worse. WDYT?

katspaugh:
You could alternatively add the trimming to the validate function maybe.

https://github.com/safe-global/web-core/blob/4b73b8699d30179253bbf3874859473315cf35f6/src/components/safe-apps/AddCustomAppModal/index.tsx#L111

yagopv:
> You could alternatively add the trimming to the validate function maybe.

The validation is a custom one and the error is showed in a different place from the form ones

https://github.com/safe-global/web-core/blob/4b73b8699d30179253bbf3874859473315cf35f6/src/components/safe-apps/AddCustomAppModal/index.tsx#L72-L79

But ... perhaps I can create an error field separated from the react-hook-form logic because that is basically the problem as it clear the custom error onChange. Let's try

yagopv:
Updated this. Ready to review again

JagoFigueroa:
Mucho bueno!  Muchas gracias!  Mucho approved 😉

PR #946, opened 2022-10-20
feat(safe-app-e2e): Add tests for permissions settings section

yagopv:
## What it solves

- Adds e2e test for the Permissions settings section
- Adds global configuration for [Test Retries](https://docs.cypress.io/guides/guides/test-retries)

### DIFF
--- a/cypress/e2e/safe-apps/permissions_settings.cy.js
+++ b/cypress/e2e/safe-apps/permissions_settings.cy.js
@@ -44,103 +44,107 @@ describe('The Safe Apps permissions settings section', () => {
     cy.visit(`${TEST_SAFE}/settings/safe-apps`, { failOnStatusCode: false })
   })
 
-  it('should show the dApps with allowed permissions', () => {
-    cy.get('main .MuiPaper-root > .MuiGrid-item')
-      .should('have.length', 4)
-      .then((items) => {
+  it('should show the permissions configuration for each stored app', () => {
+    cy.findAllByRole('heading', { level: 5 }).should('have.length', 4)
+  })
+
+  describe('For each app', () => {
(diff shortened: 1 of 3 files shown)


cypress/e2e/safe-apps/permissions_settings.cy.js
```diff
@@ -0,0 +1,146 @@
...
+          ],
+        }),
+      )
+    })
+
+    cy.visit(`${TEST_SAFE}/settings/safe-apps`, { failOnStatusCode: false })
+  })
+
+  it('should show the dApps with allowed permissions', () => {
+    cy.get('main .MuiPaper-root > .MuiGrid-item')
```
mikhailxyz, line 48: Not a big fan of getting hooked on implementation details. Is there a different approach to getting these than depending on Mui classes?

yagopv: I strongly agree but couldn't figure out because the web team don't want to add test ids or ids I think. At least in safe-react

Perhaps an article tag?

iamacook: Can't you target the title and traverse elements from there?

mikhailxyz: >  don't want to add test ids or ids I think

Why? I agree it should be the last resort, but we shouldn't ban them

yagopv: i don't get the difference, i'm going to query directly for the app and then going to the parent, sibling ... etc At the end going to implementation details isn't it?

mikhailxyz (approved):

PR #948, opened 2022-10-20
test: E2E create queued transactions

DiogoSoaress:
Creates a transaction and verifies it is queued


iamacook (commented):
Is it not possible to target these elements via the UI, e.g. find text and traverse? Targetting via classes is on par with test IDs. I would suggest we only do so as a final option.

DiogoSoaress:
> e.g. find text and traverse

This approach is more prone to break IMO. I tried to minimize relying on class selectors or test ids but I also tried to be pragmatic on evaluating the specificity and the code added to select a certain DOM element.

cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -0,0 +1,87 @@
+const SAFE = 'gor:0x04f8b1EA3cBB315b87ced0E32deb5a43cC151a91'
+const EOA = '0xE297437d6b53890cbf004e401F3acc67c8b39665'
+
+describe('Queue a transaction on N/1', () => {
```
katspaugh, line 4: `N/1` means N owners out of 1? Why not just 1/1 then?

cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -0,0 +1,87 @@
...
+
+    // Open the new transaction modal
+    cy.contains('Execute transaction').click()
+
+    cy.get('@modal').within(() => {
+      cy.get('input[type="checkbox"]')
+        .parent('span')
+        .should(($div) => {
+          // Turn the classList into a string
+          const classListString = Array.from($div[0].classList).join()
```
katspaugh, line 65: Why this instead of just checking that the input itself is checked=true?

cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -0,0 +1,87 @@
...
+
+  it('should display the queued tx on Dashboard', () => {
+    cy.contains('h2', 'Transaction queue').parents('section').as('txQueueSection')
+
+    cy.get('@txQueueSection').within(() => {
+      // There should be queued transactions
+      cy.contains('This Safe has no queued transactions').should('not.exist')
+
+      // Created transaction should be queued
+      cy.contains(`a[href="/transactions/queue?safe=${SAFE}"]`, '3' + 'Send' + '-' + '0.00003 GOR' + '1/1').should(
```
katspaugh, line 82: This check doesn't make sense if you constantly create the same transaction. `0.00003 GOR` will always match a tx from previous test runs. You need to check something unique, like the nonce, or send a unique amount each time.

cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -0,0 +1,87 @@
...
+    cy.get('input[name="amount"]').type('0.00003')
+
+    cy.contains('Next').click()
+  })
+
+  it('should create a queued transaction', () => {
+    // Alias for New transaction modal
+    cy.contains('h2', 'Review transaction').parents('div').as('modal')
+
+    cy.contains('Signing the transaction with nonce 4').click()
```
katspaugh, line 37: `4` will be 5, 6, 7, etc in the next test runs.

katspaugh (changes requested):

cypress/e2e/smoke/create_tx.cy.js
DiogoSoaress: I meant 1/N

cypress/e2e/smoke/create_tx.cy.js
DiogoSoaress: I tried to rely on input's **checked** attribute first but the HTML input is always `checked`. The only thing that changes on toggle is the MUICheckbox class list.

cypress/e2e/smoke/create_tx.cy.js
DiogoSoaress: I agree, then I will change the sent amount. Reason being we show the first 3 queued transactions so anything queued past 3 transactions won't be displayed in the Dashboard and if we would to verify in the Transaction List, we would also have to deal with the pagination limit.

cypress/e2e/smoke/create_tx.cy.js
DiogoSoaress: The test keeps queuing with nonce 3 for the reason I explained in the previous comment.

katspaugh: I don't understand how your other comment is related. The dashboard displays the oldest txns, I get that. How does the nonce of a new tx depend on that?

DiogoSoaress: Because all the test is doing is rewriting the tx with the nonce 3.

katspaugh: I see, thanks.

katspaugh (approved):

cypress/e2e/smoke/create_tx.cy.js
DiogoSoaress: I mean, I've taken this approach of not increasing the queued nonce so the test could keep veryfying the Pending queued on the Dashboard

PR #950, opened 2022-10-21
fix: disable address book export when no entries exist on all chains

### LINKED ISSUE #949: [Address book] - Export button enabled when there are no address for export
## Bug description
Deleting the last address in your list is not disabling the Export button

## Environment
 - Browser: Chrome
 - Chain: any

## Steps to reproduce
 1. Go to the app and clear the storage (at least the Safe_v2_addressBook)
 2. Create an entry, (use this address: 0xFD71c1ABadBD37F60E4C8F208386dDFC4d2Bf01f)
 3. Click the export button, it should only have 1 address to export
 4. Delete the entry in the table
 5. Check the export button

## Expected result
the button should be disabled

## Obtained result
The button is enabled. My guess is that the object still exist, but is not checking if it is empty or not

## Screenshots
image
image

iamacook:
## What it solves

Resolves #949

## How this PR fixes it

The length of chain-specific address books are checked to ensure exporting is possible as they can be an empty object.

The "No entries found" placeholder has been updated to include the chain name as well, e.g. "No entries found on Polygon".

## How to test it

1. Remove the `SAFE_v2__addressBook` `localStorage` entry.
2. Add an address book entry on chain X, and another on a chain Y.
3. Observe the export button enabled when on chain Y, even after deleting the entry (as well as "No entries found on Polygon" shown).
4. Switch to chain Y and observe that the address book is still exportable. Deleting the entry disables the button.

## Screenshots

address book export

### DIFF
--- a/src/components/address-book/AddressBookHeader/index.tsx
+++ b/src/components/address-book/AddressBookHeader/index.tsx
@@ -56,7 +56,7 @@ type Props = {
 
 const AddressBookHeader = ({ handleOpenModal, searchQuery, onSearchQueryChange }: Props): ReactElement => {
   const allAddressBooks = useAppSelector(selectAllAddressBooks)
-  const canExport = Object.values(allAddressBooks).length > 0
+  const canExport = Object.values(allAddressBooks).some((addressBook) => Object.keys(addressBook || {}).length > 0)
 
   return (
     <PageHeader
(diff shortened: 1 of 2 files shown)


usame-algan (approved):
As a future task, it might be worth exploring the option to only export a chain specific address book as it could be confusing that we display "No entries found" but its still possible to export other entries that the user can not see.

iamacook:
> As a future task, it might be worth exploring the option to only export a chain specific address book as it could be confusing that we display "No entries found" but its still possible to export other entries that the user can not see.

I agree. I've added an issue [here](https://github.com/safe-global/web-core/issues/951).

francovenica:
> As a future task, it might be worth exploring the option to only export a chain specific address book as it could be confusing that we display "No entries found" but its still possible to export other entries that the user can not see.

I think we could show in the modal "You are about to export: 2 entries from GNO chain, 3 entries from ETH chain..." so even if there are no entries and the user tries the export button at least he will understand what is he exporting

iamacook:
> I think we could show in the modal "You are about to export: 2 entries from GNO chain, 3 entries from ETH chain..." so even if there are no entries and the user tries the export button at least he will understand what is he exporting

I've adjusted the issue I created to include this.

PR #953, opened 2022-10-21
[PoC] Reusable RHF inputs

iamacook:
## What it solves

Simplified RHF inputs

## How this PR fixes it

Custom input components have been made that follow MUI/RHF APIs as purely/closely as possible. View `/pages/demo.tsx` as an example.

Note: `AddressInput`, `AddressBookInput`, `NameInput` and `DatePickerInput` would still need to be migrated if we like this implementation (although we can continue using them as is for now).

## How to test it

Open '/demo' for a preview of each element.

### DIFF
--- a/src/pages/demo.tsx
+++ b/src/pages/demo.tsx
@@ -1,0 +1,82 @@
+import { FormControlLabel, MenuItem, Radio, Grid, Button } from '@mui/material'
+import { useForm } from 'react-hook-form'
+import type { NextPage } from 'next'
+
+import { SafeRadioGroup, SafeSelect, SafeSwitch, SafeTextField } from '@/components/common/inputs'
+
+type DemoForm = {
+  text: string
+  select: number
+  radio: boolean
+  switch: boolean
+}
+
+const Demo: NextPage = () => {
+  const { control, handleSubmit, setValue } = useForm<DemoForm>({
+    defaultValues: { text: 'Test', select: 10, radio: true, switch: true },
+  })
+
+  const onSubmit = (data: DemoForm) => {
+    console.log(data)
+  }
+
+  return (
+    <main>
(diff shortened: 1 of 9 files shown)


src/pages/demo.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <main>
+      <form onSubmit={handleSubmit(onSubmit)}>
+        <Grid container spacing={3}>
+          <Grid item xs={6}>
+            <SafeTextField<DemoForm> label="Text field" name="text" control={control} required fullWidth />
+          </Grid>
+
+          <Grid item xs={6}>
+            <SafeSelect<DemoForm>
```
iamacook, line 32: This results in type-safety for things like the `name` and all `rules`.

DiogoSoaress (commented):
Although I see some of the advantages of streamlining the form components merging with MUI I found it difficult to grasp the greater upside of opting for the use of these components. 

Maybe one way to put it is saying that RHF props are taking precedence over MUI when calling one of these new components.

After talking with you, I think it makes things clearer to replace a current RHF implementation by this one and then compare both to weigh in the tradeoffs of this implementation. and if it really simplifies our work and codebase.

iamacook:
> Although I see some of the advantages of streamlining the form components merging with MUI I found it difficult to grasp the greater upside of opting for the use of these components.
> 
> Maybe one way to put it is saying that RHF props are taking precedence over MUI when calling one of these new components.
> 
> After talking with you, I think it makes things clearer to replace a current RHF implementation by this one and then compare both to weigh in the tradeoffs of this implementation. and if it really simplifies our work and codebase.

After the Safe creation revamp is finished, I will migrate all of the inputs and we can discuss this again.

iamacook:
I will close this for now and use it as a reference for when I tackle the aforementioned comment.

PR #954, opened 2022-10-21
fix: add refresh to pairing modal QR code

### LINKED ISSUE #936: [Mobile pairing]No reload QR code option when pairing is opened from the onboard popup
## Bug description
No reload QR code option when pairing is opened from the onboard popup

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Open mobile pairing from the onboard popup
 Current result: There is no reload option, but the QR code can be used only once. If there was a problem with the first connection, the user will not be able to connect

## Expected result
it should be an easy option to reload the QR code
## Screenshots
image

iamacook:
## What it solves

Resolves #936

## How this PR fixes it

The pairing QR code has been extracted to a `PairingQRCode` component, which includes the refresh functionality. This is now used in the `QRModal`, allowing for refreshing of the QR code.

## How to test it

1. Open the pairing QR modal and scan the QR, refusing the connection.
2. Observe the refresh button, which refreshes the QR on click.

## Screenshots

qr code

### DIFF
--- a/src/hooks/useChainId.ts
+++ b/src/hooks/useChainId.ts
@@ -31,7 +31,7 @@ const getLocationQuery = (): ParsedUrlQuery => {
 export const useUrlChainId = (): string | undefined => {
   const router = useRouter()
   // Dynamic query params are available only in an effect
-  const query = router.query.safe || router.query.chain ? router.query : getLocationQuery()
+  const query = router && (router.query.safe || router.query.chain) ? router.query : getLocationQuery()
   const chain = query.chain?.toString() || ''
   const safe = query.safe?.toString() || ''
 
--- a/src/components/common/PairingDetails/PairingQRCode.tsx
+++ b/src/components/common/PairingDetails/PairingQRCode.tsx
@@ -1,0 +1,53 @@
+import { useEffect, useState } from 'react'
+import { IconButton, Box } from '@mui/material'
+import RefreshIcon from '@mui/icons-material/Refresh'
+import type { ReactElement } from 'react'
+
+import { getPairingConnector, usePairingConnector, WalletConnectEvents } from '@/services/pairing/connector'
+import usePairingUri from '@/services/pairing/hooks'
+import useChainId from '@/hooks/useChainId'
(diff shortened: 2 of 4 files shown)


src/hooks/useChainId.ts
```diff
@@ -31,7 +31,7 @@ const getLocationQuery = (): ParsedUrlQuery => {
 export const useUrlChainId = (): string | undefined => {
   const router = useRouter()
   // Dynamic query params are available only in an effect
-  const query = router.query.safe || router.query.chain ? router.query : getLocationQuery()
+  const query = router && (router.query.safe || router.query.chain) ? router.query : getLocationQuery()
```
iamacook, line 34: `router` is initially `null` when the `QRModal` mounts.

usame-algan (approved):
Nice refactor to reuse the same component as well 👍

francovenica:
The refresh works once, but if the person rejects again the "Reload" icon doesn't show up anymore, the QR code remains the same and it doesn't work a second time. The reload button should show everytime the user rejects the connection (or there is any other issue detected by the onboard)

Notes:
It happened to me once that the QR code would refresh itself everytime I rejecte the connection. Keept trying a couple of times but I wasn't able to reproduce it again.
Also we should check the code in the top right corner to work the same as the one shown by the onboard modal.

iamacook:
> The refresh works once, but if the person rejects again the "Reload" icon doesn't show up anymore, the QR code remains the same and it doesn't work a second time. The reload button should show everytime the user rejects the connection (or there is any other issue detected by the onboard)
> 
> Notes: It happened to me once that the QR code would refresh itself everytime I rejecte the connection. Keept trying a couple of times but I wasn't able to reproduce it again. Also we should check the code in the top right corner to work the same as the one shown by the onboard modal.

After discussing this with the team, we agree that this is an edge case not worth fixing. The wallet popper works as expected.

I'll merge this for now and if we get reports of it being an issue in the future, we'll address it then. (For future reference: the `disconnect` event is not fired on second rejection).

PR #957, opened 2022-10-21
fix: Use fixed color for mobile pairing icon

### LINKED ISSUE #934: [Mobile pairing] Mobile pairing icons is missing in the header
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to https://safe-web-core.staging.5afe.dev/welcome
 2. connect your owner using r mobile pairing
Current result: The icon is missing
## Expected result
Make sure that the mobile pairing and WC are displayed in the header if it was used
## Screenshots
image

usame-algan:
## What it solves

Resolves #934 

## How this PR fixes it

- Uses a fixed color (#0FDA6D) for the mobile pairing icon

## How to test it

1. Open the safe
2. Connect via mobile app
3. Observe the icon is visible in light and dark mode

## Screenshots
Screenshot 2022-10-21 at 15 54 18

Screenshot 2022-10-21 at 15 54 10


usame-algan:
I picked a color that has good visibility in both light and dark mode. Also looked into rendering the wallet icons as inline svg but I think we decided against that when we implemented #215 because we would have to use `dangerouslySetInnerHtml` since they are stringified.

iamacook (commented):
Instead of using a fixed colour, maybe we can embed it as an `SvgIcon` instead of an `img`? It [seems possible to convert a stringified svg to an svg object](https://stackoverflow.com/a/38718224/6585368).

usame-algan:
> Instead of using a fixed colour, maybe we can embed it as an SvgIcon instead of an img? It [seems possible to convert a stringified svg to an svg object](https://stackoverflow.com/a/38718224/6585368).

I tested using the `DOMParser` but it parses the svg as a DOM element which we can't render in JSX. With the `img` we also have an `alt` tag for added accessibility. The only downside of this approach is that we can't use css variables for the wallet icons but I think its fine for the wallet icons to look the same in light and dark mode since usually these are some kind of product logos with a fixed coloring.

iamacook (approved):

PR #959, opened 2022-10-24
fix: use declarative warnings instead of symbols

iamacook:
## What it solves

Removes use of `>=` in warnings

## How this PR fixes it

- `Nonce must be >= ${safeNonce}` ->`Nonce can't be lower than ${safeNonce}`
- `The Safe SDK could not be initialized. Please be aware that we only support >= v1.1.1 Safes.` -> `The Safe SDK could not be initialized. Please be aware that we only support v1.1.1 Safes and up.`

## How to test it

- Create a transaction and edit the nonce to be below that of the Safe nonce. Observe the new warning.
- Open a v1.0.0 Safe and try to create a transaction. Observe the new warning.

## Screenshots

image

image


katspaugh (approved):
Good little improvement 👍

PR #967, opened 2022-10-25
feat: new creation design + step 1

### LINKED ISSUE #961: [Create Safe] Implement redesign Network + Name step
## What is the feature about

Implement the redesign for the Network + Name step

## The list of requirements

- Has fields for changing the network and adding a name for the safe
- The name field has an icon that shows a tooltip on hover
- Shows the progress bar

## Designs/sketches
Screenshot 2022-10-24 at 11 55 30

## Links

Figma: https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=466%3A10273

iamacook:
## What it solves

Resolves #961

## How this PR fixes it

This implements the new creation/load flow layout, as well as the first step of creation. The progress bar is a placeholder as it will rely on all steps/progress and the stepper hook.

Note: the info widget is yet to be included as [it has not been QA'd](https://github.com/safe-global/web-core/pull/960).

## How to test it

Open `/demo` and observe the new design.

## Screenshots

image

image


katspaugh (commented):
Creating custom components just to add a couple of lines of CSS to MUI seems...excessive?

I would just make one Card module, and either pass props for the header and content or export them as named exports.

src/components/open/CreateSafe/styles.module.css
```diff
@@ -0,0 +1,7 @@
+.back {
+  margin: var(--space-4) 0;
+}
+
+.title {
+  padding-bottom: var(--space-2);
+}
```
katspaugh, line 7: These one-liners could be `my={4}` and `pb={2}` respectively.

katspaugh (commented):
Since we're building a new file structure anyway, I suggest we don't use the old "open" folder and put stuff under `src/components/new-safe` (to be used by both Create and Add Safe).

src/components/open/CreateSafe/styles.module.css
iamacook: `Button` does not accept shorthands so I've added an `sx` to it.

src/components/new-safe/steps/Step1/styles.module.css
```diff
@@ -0,0 +1,24 @@
...
+  padding-left: var(--space-1);
+  border-left: 1px solid var(--color-border-main);
+}
+
+.chain {
+  min-width: 90px;
+  min-height: var(--space-2);
+  border-radius: 4px;
+  font-size: 12px;
+  font-weight: 700px;
```
usame-algan, line 23:
```suggestion
  font-weight: 700;
```

src/components/new-safe/steps/Step1/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+              className: css.select,
+            }}
+            defaultValue={chainId}
+          >
+            {configs.map((chain) => (
+              <MenuItem key={chain.chainId} value={chain.chainId}>
+                <ChainIndicator chainId={chain.chainId} className={css.chain} />
+              </MenuItem>
+            ))}
+          </Select>
```
usame-algan, line 80: The Menu items look different than the network selector in the header. Could we reuse `NetworkSelector` here? The main difference seems to be the label "Network" in front.

Screenshot 2022-10-25 at 11 42 50

src/components/new-safe/CreateSafe/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const onBack = () => {
+    router.back()
+
+    // Logic to be handled by stepper hook
+  }
+
+  // TODO: Improve layout when other widget/responsive design is ready
+  return (
+    <Grid container spacing={3}>
+      <Grid item xs={1} />
```
usame-algan, line 30: I would suggest not adding these offsets here. From the designs it looks like the form will be centered and the sidebar will be hidden.

schmanu (approved):
Nice!

PR #977, opened 2022-10-25
Refactor: sync useLocalStorage across components and tabs

katspaugh:
## What it solves

Although the localStorage is global and was shared across all calls of the `useLocalStorage` hook, it wasn't triggering an update when one of the instances sets a new value.

This PR subscribes to the global `storage` event and syncs the hooks across components and browser tabs.


src/services/local-storage/useLocalStorage.ts
```diff
@@ -1,32 +1,67 @@
...
   )
 
+  // Subscribe to changes in local storage and update the cache
+  // This will work across tabs
+  useEffect(() => {
+    const onStorageEvent = (event: StorageEvent) => {
+      if (event.key === local.getPrefixedKey(key)) {
+        const newValue = local.getItem<T>(key)
+        setCache((oldValue) => {
+          return newValue === undefined || newValue === oldValue ? oldValue : newValue
```
usame-algan, line 43: Do we need to check anything here or can we simply return `newValue`?

For example if we use the hook twice for the same local storage key and one hook updates its value to `undefined`, should the other hook not also update its cache to `undefined`?

katspaugh: The current usage of this hook unfortunately expects that the value is never undefined. I think it would be more correct if you could set it to undefined, but I kept the previous behavior.

usame-algan: In safe creation we set the `pendingSafe` to `undefined` when the user cancels the transaction. With the change it is being set to `{}` which causes unintended behaviour. Maybe we could call `local.removeItem` when undefined is passed?

katspaugh: OK, makes sense, let me allow undefined and remove the initialValue.

src/services/local-storage/useLocalStorage.ts
```diff
@@ -1,32 +1,74 @@
...
-    [setCache, key],
+    [key],
   )
 
+  // On mount, sync the cache with the local storage
+  useEffect(() => {
+    const lsValue = local.getItem<T>(key)
+    if (lsValue !== null) {
+      setCache(lsValue)
+    }
```
iamacook, line 43: Just as a note, I refactored this logic in https://github.com/safe-global/web-core/pull/1001 to happen in `cache` initiation, e.g.

```ts
const [cache, setCache] = useState(() => {
  const value = local.getItem<T>(key)
  return value !== undefined ? lsValue : initialValue
})
```

katspaugh: This’ll cause hydration issues.

iamacook: I see. I'll hold off on adjusting my PR until this has been merged then.

src/services/local-storage/useLocalStorage.ts
```diff
@@ -1,32 +1,74 @@
...
+
+    const onStorageEvent = (event: StorageEvent) => {
+      if (event.key === local.getPrefixedKey(key)) {
+        syncCache()
+      }
+    }
+
+    // Set the initial value
+    syncCache()
```
usame-algan, line 63: This seems to do the same as the other useEffect that sets the initial value on mount. Can we remove one or the other?

iamacook (approved):

src/components/create-safe/status/CreationStatus.tsx
```diff
@@ -12,11 +12,11 @@ import { useCallback } from 'react'
...
 import type { StepRenderProps } from '@/components/tx/TxStepper/useTxStepper'
 import type { PendingSafeData } from '@/components/create-safe/types.d'
-import useLocalStorage from '@/services/local-storage/useLocalStorage'
 import StatusMessage from '@/components/create-safe/status/StatusMessage'
 import useWallet from '@/hooks/wallets/useWallet'
 import useIsWrongChain from '@/hooks/useIsWrongChain'
 import useStatus from '@/components/create-safe/status/useStatus'
+import usePendingCreation from '@/components/create-safe/usePendingCreation'
 
 export const SAFE_PENDING_CREATION_STORAGE_KEY = 'pendingSafe'
```
usame-algan, line 21: This is not being used anymore

usame-algan (approved):
💯

PR #978, opened 2022-10-25
Feat: notify about unsupported Safe versions

katspaugh:
## What it solves

We don't support making transactions with <=1.0.0 Safes, so I've added a notification suggesting using the legacy desktop app for these Safes.

Screenshot 2022-10-25 at 16 17 39

## How to test
Can be tested on https://legacy-notif.web-core.pages.dev/eth:0xcd2E72aEBe2A203b84f46DEEC948E6465dB51c75

### DIFF
--- a/src/hooks/useSafeNotifications.ts
+++ b/src/hooks/useSafeNotifications.ts
@@ -29,9 +29,9 @@ const useSafeNotifications = (): void => {
       return
     }
 
-    const isOldest = isOldestVersion(version)
+    const isOldSafe = isOldestVersion(version)
 
-    const link = isOldest
+    const link = isOldSafe
       ? {
           href: LEGACY_DESKTOP_APP,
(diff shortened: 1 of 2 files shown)


src/hooks/useSafeNotifications.ts
```diff
@@ -26,18 +29,29 @@ const useSafeNotifications = (): void => {
...
           },
           title: 'Update Safe',
-        },
+        }
+
+    const id = dispatch(
+      showNotification({
+        variant: 'warning',
+        message: isOldest
+          ? `Safe version ${version} not supported. Please use the legacy desktop app.`
```
usame-algan, line 51:
```suggestion
          ? `Safe version ${version} is not supported. Please use the legacy desktop app.`
```

JagoFigueroa:
Looks good to me 😉

PR #1001, opened 2022-10-28
feat: Safe creation AB test

### LINKED ISSUE #968: [Safe Creation] A/B test the new design
## What is the feature about

## The list of requirements

## The A/B hook

```
  
  useEffect(() => {
  }, [setB])
  
  return isB
}
```

## Links

iamacook:
## What it solves

Resolves #968

## How this PR fixes it

An "coin toss" decides which Safe creation flow to use (new or old) and stores this flag locally. The chosen flow is also sent with every analytics event.

## How to test it

Open Safe creation (from `/welcome`) and note which flow is chosen. Return to `/welcome`, delete the `SAFE_v2__AB__newSafe` `localStorage` entry and refresh the page. Creating a Safe again may choose a different flow. Repeat until the other is taken.

## Analytics changes

`{ abTest?: string }` is sent with events in Safe load/creation, labelled "New" or "Old" accordingly.


src/config/routes.ts
```diff
@@ -5,13 +5,16 @@ export const AppRoutes = {
...
-  createSafe: '/create-safe',
   apps: '/apps',
   addressBook: '/address-book',
   balances: {
     nfts: '/balances/nfts',
     index: '/balances',
   },
+  newSafe: {
+    create: '/new-safe/create',
+    add: '/new-safe/add',
```
iamacook, line 16: This was added in anticipation of the add Safe revamp.

src/hooks/useNewSafeRoutes.ts
```diff
@@ -0,0 +1,15 @@
...
+import { AppRoutes } from '@/config/routes'
+
+export const NEW_SAFE_AB_TEST_NAME = 'newSafe'
+
+const useNewSafeRoutes = () => {
+  const shouldUseNewRoute = useABTest(NEW_SAFE_AB_TEST_NAME)
+
+  return {
+    createSafe: shouldUseNewRoute ? AppRoutes.newSafe.create : AppRoutes.open,
+    addSafe: shouldUseNewRoute ? AppRoutes.newSafe.add : AppRoutes.load,
```
iamacook, line 11: `AppRoutes.newSafe.add` uses the `AppRoutes.load` component for now.

katspaugh:
It would be much simpler to just add a conditional redirect on the old URL.

katspaugh (changes requested):
To make my point about the redirect clear, instead of the `useNewSafeRoutes` spread across many components, the A/B test should have just one entry point: the old Safe creation route.
On that page, do a simple URL redirect to the alternative URL if the user is in B.

As a side note, I would split this PR into two:
* A/B test system implementation
* The Safe creation test

iamacook:
> As a side note, I would split this PR into two:
> 
>     * A/B test system implementation
> 
>     * The Safe creation test

I'll work on this now.

PR #1006, opened 2022-10-31
[Create Safe] feat: Connect wallet step

### LINKED ISSUE #1003: [Safe Creation] Connect wallet step
## What is the feature about

Implements the connect wallet + mobile QR as first step in safe creation.

## The list of requirements

- Should be the first step in safe creation
- Should display a button to connect with the wallet
- Should display a QR code for mobile pairing
- Should disable the Continue button until a wallet is connect
- If the wallet is disconnected at any step of the safe creation it should either jump back to this step or disable the continue button

## Designs/sketches
Screenshot 2022-10-28 at 15 54 45

## Links

Figma: https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=51%3A5661

usame-algan:
## What it solves

Resolves #1003 

## How this PR fixes it

- Adds a first step to the create safe flow for connecting a wallet
- Disables the Continue button if the wallet is disconnected

## How to test it

1. Open /create-safe while disconnected
2. A button to connect the wallet exists
3. A QR code for mobile pairing exists
4. After connection it should automatically move to the next step

## Screenshots
Screenshot 2022-10-31 at 08 44 00


schmanu:
Whenever I connect a wallet from the new Dialog Step I run into this error:
Screenshot 2022-10-31 at 09 42 17

src/components/new-safe/CreateSafe/useCreateSafe.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const useCreateSafe = (setStep: StepRenderProps['setStep']) => {
+  const [pendingSafe] = useLocalStorage<PendingSafeData | undefined>(SAFE_PENDING_CREATION_STORAGE_KEY, undefined)
+  const wallet = useWallet()
+  const isWrongChain = useIsWrongChain()
+
+  useEffect(() => {
+    // Jump to the status screen if there is already a tx submitted
+    if (pendingSafe) {
+      setStep(4)
```
schmanu, line 17: Currently Step 4 is not included in the stepper which causes an error after I connect a wallet. 
In my local storage there is an empty object `{}` in the localStorage for `pendingSafe` which causes the dialog to always jump to step 4. Probably we should check for `pendingSafe?.txHash` ?

src/components/new-safe/steps/Step0/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+              <Grid item xs={12} md={6} display="flex" flexDirection="column" alignItems="center" gap={2}>
+                <PairingQRCode />
+                <Typography variant="h6" fontWeight="700">
+                  Connect to mobile
+                </Typography>
+                <PairingDescription />
+              </Grid>
+            )}
+          </Grid>
+        </>
```
DiogoSoaress, line 69: Why aren't we reusing the code we have in `ConnectWalletStep` ?
```javascript
<div className={css.connectionOptions}>
  <div className={css.wallet}>
    <WalletDetails />
  </div>

  {isSupported && (
    <div className={css.pairing}>
      <PairingDetails />
    </div>
  )}
</div>

Comment (usame-algan) [reply src/components/new-safe/steps/Step0/index.tsx:69]: I tried this first but the [new design](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=120%3A6753) looks too different for them to be reused e.g. different text in the button, different title sizes. Maybe we could update the design for the header as well? cc @liliiaorlenko

**Comment by usame-algan** [comment]
> Whenever I connect a wallet from the new Dialog Step I run into this error:

I can't reproduce this. Can you try again?

File: src/components/new-safe/CreateSafe/useCreateSafe.ts (hunk 1 above)
Comment (usame-algan) [reply src/components/new-safe/CreateSafe/useCreateSafe.ts:17]: Lets remove this part for now until the status step is implemented.

File: src/components/new-safe/steps/Step0/index.tsx [hunk 3]
```diff
@@ -0,0 +1,104 @@
...
+  const isSupported = isPairingSupported(chain?.disabledWallets)
+  const onboard = useOnboard()
+
+  const handleConnect = async () => {
+    if (!onboard) return
+
+    trackEvent(OVERVIEW_EVENTS.OPEN_ONBOARD)
+
+    await connectWallet(onboard)
+    setStep(1)
```
iamacook, line 36: If this throws, you shouldn't progress to the next step.

src/components/new-safe/steps/Step0/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+export const ConnectWalletContent = ({
+  wallet,
+  isWrongChain,
+  setStep,
+}: {
+  wallet: ConnectedWallet | null
+  isWrongChain: boolean
+  setStep: StepRenderProps<NewSafeFormData>['setStep']
+}) => {
+  const chain = useCurrentChain()
```
schmanu, line 26: Why do we pass the connected wallet here?
```suggestion
  isWrongChain,
  setStep,
}: {
  isWrongChain: boolean
  setStep: StepRenderProps<NewSafeFormData>['setStep']
}) => {
  const wallet = useWallet()
  const chain = useCurrentChain()
```

src/components/new-safe/steps/Step0/index.tsx
schmanu, line 36: Probably we should call do something like 
```ts
onSubmit({ owners: [{ address: wallet.address }]
```
instead such that the owners step will have the connected wallet as owner. 
Otherwise it will not be prefilled with the connected wallet as the initialData of the stepper was already created.

schmanu (commented):
The error when connecting is gone.
But if I open the create safe dialog without a wallet connected and then connect a wallet in Step0 the connected wallet won't be added to the prefilled owners.

src/components/new-safe/steps/Step0/index.tsx
usame-algan: Because `wallet` is also used in the parent component but after looking at it again we can actually reuse the `useSafeCreation` there and access `wallet` and `isWrongChain` in this component directly like you suggest. Thanks!

src/components/new-safe/steps/Step0/index.tsx
usame-algan: Good idea! I reused the `WalletDetails` component which allows us to pass an `onConnect` handler maybe we can solve it that way too.

src/components/new-safe/steps/Step0/index.tsx
usame-algan: I changed it back to reuse both components

src/components/new-safe/steps/Step0/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+import useCreateSafe from '@/components/new-safe/CreateSafe/useCreateSafe'
+import type { ConnectedWallet } from '@/services/onboard'
+
+export const ConnectWalletContent = ({ onSubmit }: { onSubmit: StepRenderProps<NewSafeFormData>['onSubmit'] }) => {
+  const isWrongChain = useIsWrongChain()
+  const wallet = useWallet()
+  const chain = useCurrentChain()
+  const isSupported = isPairingSupported(chain?.disabledWallets)
+
+  const onConnect = (wallet?: ConnectedWallet) => {
```
schmanu, line 22: I feel like the wallet should not be allowed to be undefined or if we have to allow this we should check for this before we call  `onSubmit`
```
if (wallet) {
  onSubmit({ owners: [{ address: wallet.address || '', name: wallet.ens || '' }] })
}
```

schmanu (approved):
Looking good now 🥳 . I just have one small left over concern.

src/components/new-safe/steps/Step0/index.tsx
usame-algan: Added a check for the wallet before calling onSubmit

src/services/pairing/QRModal.tsx
```diff
@@ -56,7 +56,7 @@ const Modal = ({ uri, cb }: { uri: string; cb: () => void }) => {
   return (
     <StoreHydrator>
       <AppProviders>
-        <Dialog open onClose={handleClose} disablePortal>
+        <Dialog open onClose={handleClose} disablePortal sx={{ zIndex: 1301 }}>
```
iamacook, line 59: We can use onboard's CSS var here.

iamacook (approved):

PR #1007, opened 2022-10-31
Fix: show a disabled Execute checkbox in the Execute modal

katspaugh:
## What it solves

A small UX improvement.

When executing a transaction, a checked Execute checkbox signals that it's an execution, not a signature.

In the "Execute" modal, the checkbox was completely hidden because it's the only option, but with this, we're losing this UI signal. So I made it a disabled checkbox instead of completely hiding it.

Screenshot 2022-10-31 at 09 05 43


usame-algan (approved):
There is a failing test

iamacook (approved):

PR #1008, opened 2022-10-31
feat: AB testing system

### LINKED ISSUE #968: [Safe Creation] A/B test the new design

iamacook:
## What it solves

Partially resolves #968

## How this PR fixes it

A new AB testing system has been implemented, adding `abTest` to GTM events (if `_abTest` is currently set).

## How to test it

`gtmSetAbTest` should be called in `useEffect` and the returned unsubscribe function called on cleanup. When set, the `abTest` key should exist in the GTM events and be visible on debug view.

## Analytics changes

`abTest` string should now be present in events.

## Screenshots

image


src/services/analytics/gtm.ts
```diff
@@ -48,6 +48,23 @@ export const gtmSetChainId = (chainId: string): void => {
   _chainId = chainId
 }
 
+export const enum AbTest {
+  SAFE_CREATION = 'safe-creation',
+}
+
+let _abTest: AbTest | null = null
```
iamacook, line 55: Using a hashmap would require us to manually map every property manually. As such, I've opted for a localised setter like we use with `chainId`.

The example use case is explained in the testing example: set this in a `useEffect` on the given route, resetting it on cleanup.

iamacook:
I've branched this incorrectly. Creating a new PR.

PR #1014, opened 2022-10-31
Refactor: external store in useLocalStorage

katspaugh:
## What it solves
Caches the localStorage in a globally shared in-memory store (ExternalStore).
This will make `useLocalStorage` slightly faster if the cache is already initialized.
Important for use-cases like the A/B test hook which we'll initialize on app start but it can be read later on another route.


iamacook (approved):
Tested working cached values and cross-tab updating.

PR #1017, opened 2022-11-01
feat: add AB testing to Safe creation

### LINKED ISSUE #968: [Safe Creation] A/B test the new design

iamacook:
## What it solves

Resolves #968

## How this PR fixes it

`useABTesting` now decides whether the new or old Safe creation flow should be used (caching the decision locally). If the new is "chosen", the "decision" is attached to all analytics events.

## How to test it

Open the Safe and attempt to create a Safe, observe either the new or old creation flow. Calling `localStorage.removeItem("SAFE_v2__AB_safe_creation")` and trying again should choose the other, although this may have to be repeated a few times.

When the new Safe creation flow was "chosen", all analytics events should appear as follows (and be visible in the GA debug view).

```ts
  {
    ...event,
    abTest: "safe_creation"
  }
```

When the old creation flow was "chosen", the `abTest` property will not exist.

## Notes

- The creation route has been changed to `/new-safe/create`.
- We need to release the new GTM container when testing is completed.

## Analytics changes

A new `abTest` property is populated when a user is chosen to test the desired feature.

## Screenshots

image


katspaugh (approved):
👍

katspaugh:
Tested, I'm assigned correctly to A or B randomly, and it opens the new flow on B. ✅ 
GA events are augmented accordingly.

PR #1019, opened 2022-11-01
fix: hide widget when no Safe address exists

iamacook:
## What it solves

Resolves empty widget container in header.

## How this PR fixes it

If no Safe address exists, the widget is not rendered.

## How to test it

Open Safe creation and observe no empty widget container:

image


katspaugh (approved):
Thanks!

PR #1024, opened 2022-11-01
Fix: tiny icons in tx-list

katspaugh:
## What it solves
Screenshot 2022-11-01 at 15 25 02

### DIFF
--- a/src/components/common/EthHashInfo/styles.module.css
+++ b/src/components/common/EthHashInfo/styles.module.css
@@ -27,6 +27,11 @@
   white-space: nowrap;
 }
 
+.addressRow svg {
+  height: 1.2em;
+  width: auto;
+}
+
 .nameRow {
   overflow: hidden;
 }
--- a/src/components/common/EthHashInfo/styles.module.css
+++ b/src/components/common/EthHashInfo/styles.module.css
@@ -27,11 +27,6 @@
   white-space: nowrap;
 }
 
-.addressRow svg {
-  height: 1.2em;
-  width: auto;
-}
-
 .nameRow {
   overflow: hidden;
 }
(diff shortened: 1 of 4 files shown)


usame-algan:
There are a few more icons that appear smaller than they used to be:

Address Book:
Screenshot 2022-11-01 at 16 14 34
Screenshot 2022-11-01 at 16 14 54

Owner Settings:
Screenshot 2022-11-01 at 16 15 14

Assets:
Screenshot 2022-11-01 at 16 20 29

I think the reason is that some icons are exported in their large size from [Figma](https://www.figma.com/file/HKW5iarMST9rL7Kxkf59qY/MUI-for-Figma?node-id=10964%3A70434) and then rendered with `fontSize="small"` which looks inconsistent. Instead they should be exported in their small size and not be modified within `SvgIcon`

src/components/common/EthHashInfo/styles.module.css
```diff
@@ -27,6 +27,11 @@
   white-space: nowrap;
 }
 
+.addressRow svg {
+  height: 1.2em;
+  width: auto;
+}
+
```
usame-algan, line 34: We can achieve the same result by removing the `fontSize="small"` prop from the `TokenExplorerLink` and `CopyButton`

katspaugh: Good suggestion, it fixes it indeed. I grepped, and it looks like we have quite a bunch of places where this `fontSize="small"` is used on icons.

katspaugh: I'm going to fix it only for copy and explorer in this PR to stay within the scope. Will open an issue to fix and check the rest.

usame-algan (approved):
👍

PR #1026, opened 2022-11-01
Chore: GitHub action to deploy releases

katspaugh:
## What it solves
Adds a GitHub action partially copied from safe-react.
When a GitHub release is published, it will build a production version of the app and upload it to a special folder on the staging (!) bucket.
It will then trigger a web-hook specified in the secrets.
After that DevOps will be able to deploy the code to the production bucket.


scripts/github/prepare_production_deployment.sh
```diff
@@ -0,0 +1,17 @@
...
+set -ev
+
+# Only:
+# - Tagged commits
+# - Security env variables are available.
+if [ -n "$VERSION_TAG" ] && [ -n "$PROD_DEPLOYMENT_HOOK_TOKEN" ] && [ -n "$PROD_DEPLOYMENT_HOOK_URL" ]
+then
+  curl --silent --output /dev/null --write-out "%{http_code}" -X POST \
+     -F token="$PROD_DEPLOYMENT_HOOK_TOKEN" \
+     -F ref=master \
```
katspaugh, line 12: Raul, does this need to be "master"?

scripts/github/prepare_production_deployment.sh
```diff
@@ -0,0 +1,17 @@
...
+     -F "variables[TRIGGER_RELEASE_COMMIT_TAG]=$VERSION_TAG" \
+      $PROD_DEPLOYMENT_HOOK_URL
+else
+  echo "[ERROR] Production deployment could not be prepared"
```
katspaugh, line 16: Btw, should we add `exit 1` here?

scripts/github/prepare_production_deployment.sh
luarx: Yes, that is ok as it does not refer to the Github repo branch

scripts/github/prepare_production_deployment.sh
luarx: I would say that just leave as it is configured in the old repo, probably we should add it but must be tested before

.github/workflows/release.yml
```diff
@@ -0,0 +1,62 @@
...
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+
+      # Script to upload release files
+      - name: 'Upload release build files for production'
+        run: |
+          cd out
+
+          aws s3 sync . s3://${{ secrets.AWS_STAGING_BUCKET_NAME }}/releases/${{ github.event.release.tag_name }}
```
luarx, line 50: If the only difference between this and the following line is the "--delete" argument, then this line is not needed.

scripts/github/prepare_production_deployment.sh
katspaugh: OK.

.github/workflows/release.yml
katspaugh: Added a comment about why it's needed.

luarx (approved):

PR #1031, opened 2022-11-02
fix: use default viewbox for custom icons

### LINKED ISSUE #1025: Some icons are too small
Some icons need `fontSize="small"` removed.

From @usame-algan's comment:

There are a few more icons that appear smaller than they used to be:

Address Book:
Screenshot 2022-11-01 at 16 14 34
Screenshot 2022-11-01 at 16 14 54

Owner Settings:
Screenshot 2022-11-01 at 16 15 14

Assets:
Screenshot 2022-11-01 at 16 20 29

There is some inconsistency with the icon viewboxes (some have a larger viewbox but overall smaller icon). Could this be related to how they are exported from [Figma](https://www.figma.com/file/HKW5iarMST9rL7Kxkf59qY/MUI-for-Figma?node-id=10964%3A70434)?

iamacook:
## What it solves

Resolves #1025

## How this PR fixes it

All custom SVG icons are now `width="24" height="24" viewBox="0 0 24 24"` by default and `fontSize="small"` sets them to `16x16`.

Two notable fixes include the rocket icon for execution and (un-)pinning Safe Apps icons.

## How to test it

Open the Safe and observe all icons are correctly sized when compared to [Figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File).


usame-algan:
The sizes look much more consistent now 👍 

A few things I noticed while testing:

Safe Apps icon color is too dark and the link icon should have all circles outlined
Screenshot 2022-11-02 at 17 45 29

Owner Settings icons need to be a bit more spaced out
Screenshot 2022-11-02 at 17 46 49

Whats new icon
Screenshot 2022-11-02 at 17 47 17

The info icon is too dark and should be outlined
Screenshot 2022-11-02 at 17 48 09

In dark mode, the icons should have the color #A1A3A7 instead of #636669
Screenshot 2022-11-02 at 17 48 58

There are two slightly different versions for the qr, copy and external link icon in Figma. Which ones should we use? cc @liliiaorlenko 
Screenshot 2022-11-02 at 17 51 34
Screenshot 2022-11-02 at 17 51 48

iamacook:
Thank you for the thorough review. I've ticked all the comments that I've addressed as follows:

> - Safe Apps icon color is too dark and the link icon should have all circles outlined

The share button is now correct and bookmark icon is `"border.main"` (although the design uses `"primary.main"`) as I agree it looks much more consistent:

image
image

> - Owner Settings icons need to be a bit more spaced out

I've centralised the `action` spacing to be in `EnhancedTable`. This means we no longer have to style the row in other places, i.e. in `AddressBookTable`:

image

> - Whats new icon

This now renders correctly:

image

> - The info icon is too dark and should be outlined

This is now our custom icon (not MUI's) and the correct colour:

image

> - There are two slightly different versions for the qr, copy and external link icon in Figma.

The `SidebarHeader` icons seem to be a lot bolder that those in [our icon collection](https://www.figma.com/file/HKW5iarMST9rL7Kxkf59qY/MUI-for-Figma). I've kept the often used QR, copy and link icons in `./public/images/common` and added `{{name}}-bold.svg` versions in `./public/images/sidebar` for now:

image

---

The following seems to be an inconsistency in the design as all other buttons of a similar action also use `"border.main"`. I think we should keep it as is. cc @liliiaorlenko 

> - In dark mode, the icons should have the color #A1A3A7 instead of #636669:
>
> Screenshot 2022-11-02 at 17 48 58

usame-algan:
The Tooltip is not being displayed anymore when hovering the info icon next to Safe Nonce in Settings.
Screenshot 2022-11-03 at 11 50 12

The icon colors in dark mode are not #A1A3A7 but rgb(99, 102, 105). It looks like some default fallback of MUI
Screenshot 2022-11-03 at 11 52 43

The new Send/Receive Icons appear a bit larger than other icons. In Figma they are 10px instead of 16px like the rest. We might be able to solve it by scaling them down while still keeping the same viewbox. This would have to be done before exporting though cc @liliiaorlenko 
Screenshot 2022-11-03 at 11 52 59

iamacook:
> The Tooltip is not being displayed anymore when hovering the info icon next to Safe Nonce in Settings.

This is now fixed:

image

> The icon colors in dark mode are not #A1A3A7 but rgb(99, 102, 105).

`rgb(99, 102, 105)` is `#636669` which is `"border.main"`. This is the standard colour we use for all basic icons, e.g. the chevrons in the header.

> The new Send/Receive Icons appear a bit larger than other icons. In Figma they are 10px instead of 16px like the rest. We might be able to solve it by scaling them down while still keeping the same viewbox.

They appear larger but they are all the same:

image

usame-algan:
> rgb(99, 102, 105) is #636669 which is "border.main". This is the standard colour we use for all basic icons, e.g. the chevrons in the header.

You are right, most icons I can find in Figma have #636669 but there are some with #A1A3A7 and even #B2B5B2 in dark mode. @liliiaorlenko could you clarify which color to go with for all icons or if there is a system behind the different colors?

liliiaorlenko:
So we have to exclude colour #636669, and only use #A1A3A7 in light mode. In dark mode #636669.
As for the sizes I can export all the icons in 16px if needed so we would have only 16px and 24px

iamacook:
> What I mean is that the send and receive Icons should be smaller than the other icons. In Figma they are 10x10px instead of 16x16px like the rest of the icons.

The icons are now in the correct resolutions:

image

usame-algan (approved):
💯 🚀

iamacook:
I'm going to merge this directly as it is purely image-based.

PR #1032, opened 2022-11-02
fix: initial pageview event

iamacook:
## What it solves

Resolves initial pageviews not being tracked.

## How this PR fixes it

The cookie policy is checked before attempting to dispatch a pageview (which ensures GTM is enabled).

## How to test it

Open the Safe in GTM debug view and refresh. Observe the intial pageview being tracked.

## Analytics changes

Intial pageview is now successfully tracked.


katspaugh (approved):

francovenica:
It seems that I don't have access to the GTM. I requested it

francovenica:
LGTM.
This is the pageview event you have as soon as you enable the analytics in the application
image

PR #1033, opened 2022-11-02
refactor: GTM

iamacook:
## What it solves

Simplifies GTM loading and adds event stacking.

## How this PR fixes it

- `TagManager.initialize` and `TagManager.dataLayer` have been simplified. Most notably, the `dataLayer` is **always** pushed to as GTM will send these events after initialisation.
- GTM is now fully removed on `destroy` as well as cookies being removed. (Before it was not unmounted).
- Logging now reflects what is being sent to GTM.
- An unnecessary `isGtmLoaded` flag was also removed.

### Refactor

- GTM should initialise and events should be sent as before. (make sure the intial `pageview` works especially)
- No errors/warnings should be shown in the console if the cookie preferences are changed,
- When disabling "Analytics" preferences, the GTM tag should be removed. (no tracking visible in Network tab)

### Event stacking

Note: ensure this is being tested on a cold cache. (We need to check the legality of this).

1. Open the Safe and DO NOT agree to "Analytics" cookies but keep the banner open.
2. Click around the Safe, navigate to different pages, etc. Observe no events in GTM.
3. Accept the "Analytics" cookies and observe all of the events that _should've_ been sent now appear in GTM.

## Analytics changes

Events are now stacked regardless of cookie policy. They are NOT sent to GTM unless agreed to.

### DIFF
--- a/src/services/analytics/TagManager.ts
+++ b/src/services/analytics/TagManager.ts
@@ -1,3 +1,5 @@
+import Cookies from 'js-cookie'
+
 // Based on https://github.com/alinemorelli/react-gtm
 
 type DataLayer = Record<string, unknown>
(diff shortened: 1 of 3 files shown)


iamacook:
Closing in favour of #1047.

iamacook:
Closing in favour of #1047.

PR #1041, opened 2022-11-03
fix: Anonymize safe address on gtm init

### LINKED ISSUE #1034: GA is tracking user_engamenet
## Bug description
We should somehow disable the tracking of the `user_engagement` metric because it captures the full non-anonymized page URL.
Or, alternatively, make it not track the full URL.

Screenshot 2022-11-02 at 13 22 04

According to @usame-algan, it should be possible by setting an anonymized URL in the initial data layer.

usame-algan:
## What it solves

Resolves #1034 

## How this PR fixes it

Sets anonymized path and location in dataLayer when initializing GTM

## How to test it

1. Open the Safe with DebugView
2. Reload the page
3. Scroll around
4. Observe a `user_engagement` event with anonymized path and location values

## Screenshots
Screenshot 2022-11-03 at 09 10 30
Screenshot 2022-11-03 at 09 10 33

### DIFF
--- a/src/services/analytics/gtm.ts
+++ b/src/services/analytics/gtm.ts
@@ -20,6 +20,7 @@ import {
 } from '@/config/constants'
 import type { AnalyticsEvent, EventLabel, SafeAppEvent } from './types'
 import { EventType } from './types'
+import { addressRe, prefixedAddressRe } from '@/utils/url'
 
 type GTMEnvironment = 'LIVE' | 'LATEST' | 'DEVELOPMENT'
 type GTMEnvironmentArgs = Required<Pick<TagManagerArgs, 'auth' | 'preview'>>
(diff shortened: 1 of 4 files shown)


src/services/analytics/gtm.ts
```diff
@@ -56,10 +76,14 @@ export const gtmInit = (): void => {
     return
   }
 
+  const pagePath = getAnonymizedPathname()
+
   TagManager.initialize({
     gtmId: GOOGLE_TAG_MANAGER_ID,
     ...GTM_ENVIRONMENT,
     dataLayer: {
+      pageLocation: `${location.origin}${pagePath}`,
```
iamacook, line 85: In my tests, `router.pathname` doesn't include our rewrites. We could pass it as an argument to `gtmInit` and then not need anonymise the `pathname`.

usame-algan: Good idea! It works nicely although we have to exclude it from the dependency array but I think it should be fine since it is only needed on the initial load.

src/services/analytics/useGtm.ts
```diff
@@ -19,7 +19,8 @@ const useGtm = () => {
 
   // Initialize GTM, or clear it if analytics is disabled
   useEffect(() => {
-    isAnalyticsEnabled ? gtmInit() : gtmClear()
+    isAnalyticsEnabled ? gtmInit(router.pathname) : gtmClear()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
iamacook, line 23: Can you add a comment why we need to do this please.

src/utils/url.ts
```diff
@@ -5,7 +5,8 @@ const trimTrailingSlash = (url: string): string => {
 const isSameUrl = (url1: string, url2: string): boolean => {
   return trimTrailingSlash(url1) === trimTrailingSlash(url2)
 }
-
+export const prefixedAddressRe = /[a-z0-9-]+\:0x[a-f0-9]{40}/i
+export const addressRe = /0x[a-f0-9]{40}/i
```
iamacook, line 9: This is not being used anymore.

iamacook (approved):

PR #1042, opened 2022-11-03
feat(safe-apps): Analytics refactor

yagopv:
## What it solves

This PR add a refactor to the Safe Apps analytics current behavior as part of the [analytics epic](https://github.com/safe-global/web-core/issues/942)

## How this PR fixes it

We are making changes to reflect the status described [here](https://www.notion.so/gnosis-safe/Tag-Manager-configuration-d5a4a4e3a97e4bec9eabfb3f8fe653a0)

## How to test it

We need to validate all the event actions described [here](https://github.com/safe-global/web-core/blob/b106b67c563dad193efcf672e6883e2a37c7b50d/src/services/analytics/events/safeApps.ts#L11-L65) are being sent

@JagoFigueroa the preference is validate this using GTM and GA4 tools. We can align on this

## Analytics changes

- We now are sending all the analytics from Safe Apps using the concrete `safeApp`  event or the custom `pageview`. Previously it was a combination of Safe App events and Safe Web Custom events with `metadata` category. Is easier and less prone to errors to use only the Safe App event. We now have under the `safeApp` event two categories (`safe-apps` and `safe-apps-sdk`)

- We now are sending the app name always in the `safeAppName` diimension. If the app belongs to the default list we send the name and if not the url (custom app). Previously the dimension was only being used for SDK events and we were sending the names in the `eventLabel` with no clear consensus. Doing this, we know clearly if the app is known and we free a dimension (`eventLabel`) for include any other data

- We send `pageview` events for the Safe Apps list and for the concrete app being loaded in order to generate reports as the time elapsed in Safe apps sections or apps


src/pages/share/safe-app.tsx
```diff
@@ -28,9 +29,15 @@ const ShareSafeApp = () => {
   }
 
   return (
-    <main>
-      <SafeAppLanding appUrl={appUrl} chain={chain} />
-    </main>
+    <>
+      <Head>
+        <title>Safe Apps – Share</title>
```
mikhailxyz, line 34: Perhaps we can include the app name in the title as a future enhancement 🤔

src/components/safe-apps/SafeAppLandingPage/index.tsx
```diff
@@ -30,17 +30,19 @@ const SafeAppLanding = ({ appUrl, chain }: Props) => {
...
-        label: safeApp.name,
-      })
-      trackEvent({
-        ...SAFE_APPS_EVENTS.SHARED_APP_CHAIN_ID,
-        label: chain.chainId,
-      })
+    if (!isLoading && safeApp?.chainIds.length) {
+      const appName = backendApp ? backendApp.name : safeApp.url
+
+      trackSafeAppEvent(SAFE_APPS_EVENTS.SHARED_APP_LANDING, appName)
```
mikhailxyz, line 36: Do we really need 2 separate events for this? Can the chainid be included in the `SHARED_APP_LANDING`?

mikhailxyz (approved):

src/pages/share/safe-app.tsx
yagopv: yep, we sent the app name or url (if custom) [here](https://github.com/safe-global/web-core/blob/39776b65ac58cc0b71a720fcabd62a66eb3f883e/src/components/safe-apps/SafeAppLandingPage/index.tsx#L36)

src/components/safe-apps/SafeAppLandingPage/index.tsx
yagopv: yep, changed it. Does not make sense

PR #1038, opened 2022-11-02
fix: footer release link

iamacook:
## What it solves

Broken footer link

## How this PR fixes it

The `homepage` in `package.json` has been updated to the correct repository, as well as the link corrected to match our release naming format.

## How to test it

Click the version number in the footer and observe the latest release opens.

PR #1045, opened 2022-11-03
Release/1.0.3

usame-algan:
## Includes

- #1038 
- #1032 
- #1041

### DIFF
--- a/src/services/analytics/useGtm.ts
+++ b/src/services/analytics/useGtm.ts
@@ -19,7 +19,10 @@ const useGtm = () => {
 
   // Initialize GTM, or clear it if analytics is disabled
   useEffect(() => {
-    isAnalyticsEnabled ? gtmInit() : gtmClear()
+    // router.pathname doesn't contain the safe address
+    // so we can override the initial dataLayer
+    isAnalyticsEnabled ? gtmInit(router.pathname) : gtmClear()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isAnalyticsEnabled])
 
   // Set the chain ID for GTM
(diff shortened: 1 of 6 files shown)


iamacook (approved):

src/services/analytics/useGtm.ts
```diff
@@ -19,7 +19,10 @@ const useGtm = () => {
 
   // Initialize GTM, or clear it if analytics is disabled
   useEffect(() => {
-    isAnalyticsEnabled ? gtmInit() : gtmClear()
+    // router.pathname doesn't contain the safe address
+    // so we can override the initial dataLayer
+    isAnalyticsEnabled ? gtmInit(router.pathname) : gtmClear()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
katspaugh, line 25: Not the best solution, we should use a state variable here that would be assigned to the initial pathname only once.

katspaugh (approved):
🔥

francovenica:
The link in the footer is fine. The 1.0.3 page in github doesn't exist yet, but when it does it will be fine
The page view is fine

Regarding the Address in the tracker, there is the referral still showing the full address.
I understand that referrals could be any page in the world, but since is being tracked as we move through the site we should parse it and check there is no address present
image

usame-algan:
> Regarding the Address in the tracker, there is the referral still showing the full address.

Afaik we had the referrer overwrite in GTM at some point but removed it again. Testing on localhost, the referrer doesn't seem to be there in the events but on prod it is. I suggest we merge this as it anonymizes the path and location and look at the referrer issue separately.

PR #1047, opened 2022-11-03
refactor: simplify GTM logic + unmount on permission removal

iamacook:
## What it solves

Refactors GTM and adds `disale` functionality.

## How this PR fixes it

- `TagManager.initialize` and `TagManager.dataLayer` have been simplified.
- GA cookies are removed and the session is reloaded on `TagManager,disable`
- Logging now reflects what is being sent to GTM.
- An unnecessary `isGtmLoaded` flag was also removed.

### Refactor

- GTM should initialise and events should be sent as before. (make sure the intial `pageview` works especially)
- No errors/warnings should be shown in the console if the cookie preferences are changed, A reload should occur when revoking permission.
- When disabling "Analytics" preferences, the GTM tag should be removed. (no tracking visible in Network tab)

## Analytics changes

The GTM scripts are all removed on revocation.


src/services/analytics/TagManager.ts
```diff
@@ -10,85 +13,100 @@ export type TagManagerArgs = {
...
+    if (gtmScriptRef) {
+      return
+    }
 
-    if (dataLayer) {
-      const gtmDataLayerScript = _getGtmDataLayerScript(dataLayer)
-      document.head.appendChild(gtmDataLayerScript)
+    // Push configuration to dataLayer
+    if (args.dataLayer) {
+      window[DATA_LAYER_NAME] = []
```
katspaugh, line 61:
```suggestion
      window[DATA_LAYER_NAME] = window[DATA_LAYER_NAME] || []
```

src/services/analytics/TagManager.ts
```diff
@@ -10,85 +13,100 @@ export type TagManagerArgs = {
...
+    // Initialize GTM. This pushes the default dataLayer event:
+    // { "gtm.start": new Date().getTime(), event: "gtm.js" }
+    document.head.insertBefore(gtmScriptRef, document.head.childNodes[0])
   },
   dataLayer: (dataLayer: DataLayer) => {
-    if (window[DATA_LAYER_NAME]) {
-      return window[DATA_LAYER_NAME].push(dataLayer)
+    // Push to dataLayer if mounted
+    if (!gtmScriptRef || !window[DATA_LAYER_NAME]) {
+      return
```
katspaugh, line 74:
```suggestion
    window[DATA_LAYER_NAME] = window[DATA_LAYER_NAME] || []
    window[DATA_LAYER_NAME].push(dataLayer)

    if (!gtmScriptRef) {
      return
```

katspaugh (commented):
Left suggestions in the code to allow stacking pre-init events as you had in another PR.

iamacook:
> Left suggestions in the code to allow stacking pre-init events as you had in another PR.

We want to discuss event stacking with legal as the team is concerned regarding consent.

katspaugh (approved):
Looks good!
Let's merge _after_ 1.0.4, tho.

iamacook:
> We want to discuss event stacking with legal as the team is concerned regarding consent.

Just for future reference: @johannesmoormann confirmed that we *don't* want to stack event prior to cookie agreement.

francovenica:
If I enable the "Analytic" checkbox, track a few actions and then I disable the checkbox this error start to show in the console for every action.
image

The error goes away once I reload the page manually

francovenica:
Tried this again and I can still see the same issue. Can u check if the PR updated correctly?

iamacook:
> Tried this again and I can still see the same issue. Can u check if the PR updated correctly?

Despite unmounting the scripts, they remain in memory. @usame-algan and I will investigate this further.

src/config/constants.ts
```diff
@@ -45,6 +45,9 @@ export const GOOGLE_TAG_MANAGER_AUTH_LIVE = process.env.NEXT_PUBLIC_GOOGLE_TAG_M
 export const GOOGLE_TAG_MANAGER_AUTH_LATEST = process.env.NEXT_PUBLIC_GOOGLE_TAG_MANAGER_LATEST_AUTH || ''
 export const GOOGLE_TAG_MANAGER_DEVELOPMENT_AUTH = process.env.NEXT_PUBLIC_GOOGLE_TAG_MANAGER_DEVELOPMENT_AUTH || ''
 
+// Google Analytics
+export const GOOGLE_ANALYTICS_MEASUREMENT_ID = process.env.NEXT_PUBLIC_GOOGLE_ANALYTICS_MEASUREMENT_ID || ''
```
iamacook, line 49: We need to add this to our secrets.

src/services/analytics/TagManager.ts
```diff
@@ -1,94 +1,114 @@
...
+    // const GTM_COOKIE_LIST = ['_ga', '_gat', '_gid']
 
-    const gtmDataLayerScript = _getGtmDataLayerScript(dataLayer)
-    document.head.insertBefore(gtmDataLayerScript, document.head.childNodes[0])
+    // GTM_COOKIE_LIST.forEach((cookie) => {
+    //   Cookies.remove(cookie, {
+    //     path: '/',
+    //     domain: `.${location.host.split('.').slice(-2).join('.')}`,
+    //   })
+    // })
```
iamacook, line 111: These need not be removed anymore but we need to determine if it's legal to keep them.

src/services/analytics/__tests__/TagManager.test.ts
```diff
@@ -1,90 +1,180 @@
...
+  get: jest.fn(),
+}))
+
+jest.mock('@/config/constants', () => ({
+  get GOOGLE_ANALYTICS_MEASUREMENT_ID() {
+    return MOCK_MEASUREMENT_ID
+  },
+}))
 
 describe('TagManager', () => {
```
iamacook, line 23: I am aware of the broken tests.

src/config/constants.ts
iamacook: We would just need to set both as disabled in the `window` object.

src/services/analytics/TagManager.ts
```diff
@@ -1,94 +1,89 @@
...
-
-// Data layer scripts
+    return script
+  },
+  isInitialized: () => {
+    const GTM_SCRIPT = 'https://www.googletagmanager.com/gtm.js'
 
-export const _getGtmDataLayerScript = (dataLayer: DataLayer) => {
-  const script = document.createElement('script')
+    return !!document.querySelector(`[src^="${GTM_SCRIPT}"]`)
```
usame-algan, line 44: Would it work if we checked that `window.dataLayer` exists instead? It would be a bit simpler than querying the DOM.

usame-algan (approved):
Great idea to reload the page! 💯

src/services/analytics/TagManager.ts
iamacook: I would sooner check the DOM. We could manually push to `window.dataLayer` and then this would return `true` when GTM isn't initialised.

francovenica:
I agree with the solution. I don't think people are enabling/disabling the analytics every 5 mins for the reload to be an inconvenience.
LGTM

PR #1049, opened 2022-11-03
fix(safe-apps): Communication with the Safe App starts before Safe Info is loaded

yagopv:
## What it solves

Fix a timing issue where 20-30% of the times the communication starts before having the safe information


mikhailxyz (approved):

schmanu (changes requested):
The unit test for `AppFrame` are failing.
Otherwise the fix looks good :)

yagopv:
> The unit test for `AppFrame` are failing.

Fixed!

schmanu (approved):
Lets goooo!

PR #1052, opened 2022-11-03
chore: Add action to deploy to dev

usame-algan:
The same action we use to deploy to staging on main merge but instead it should deploy to a different bucket on AWS when merging into dev.


.github/workflows/deploy-dev.yml
```diff
@@ -0,0 +1,58 @@
+name: Deploy to Dev
```
dasanra, line 1: There is no need to create a whole new file. Almost all this code is exactly the same as on `deploy-staging.yml`

You can detect push to different branches for the same action
https://github.com/safe-global/safe-react/blob/0d09d940100ac2f0e37b371a6273b51cb490f081/.github/workflows/deploy.yml#L6-L9

And the you can filter with an `if` in the last step to select to upload to dev or staging
https://github.com/safe-global/safe-react/blob/0d09d940100ac2f0e37b371a6273b51cb490f081/.github/workflows/deploy.yml#L103-L112

It will make easier to keep the workflows whenever you add new env vars or update some build flows

dasanra (changes requested):

.github/workflows/deploy-dev.yml
usame-algan: Thanks! I combined them in one file as per your suggestion.

luarx (approved):

.github/workflows/deploy-dev.yml
luarx: Look good to me!

dasanra (approved):

.github/workflows/deploy-dev.yml
katspaugh: It could be further improved by reusing the same upload script, just by setting a different bucket from the script.
```
env:
  BUCKET: ${{ staging ? STAGING_BUCKET : DEV_BUCKET }}
```

luarx: Remember that dev and staging use different upload paths.
- Dev: /
- Staging: /current

usame-algan: I tried a two ways to solve this and used `deploy-dev-action` instead of `dev` for testing in this PR:

1.
```
on:
  push:
    branches:
      - main
      - dev
      - deploy-dev-action
env:
    BUCKET: ${{ github.ref == 'refs/heads/deploy-dev-action' && secrets.AWS_DEVELOPMENT_BUCKET_NAME || format('{0}/current', secrets.AWS_STAGING_BUCKET_NAME }}
```

2. (These steps were always skipped in the CI for some reason)
```
on:
  push:
    branches:
      - main
      - dev
      - deploy-dev-action
...
- name: Set staging bucket
  run: echo "BUCKET=${{ secrets.AWS_STAGING_BUCKET_NAME }}/current" >> $GITHUB_ENV
  if: github.ref == 'refs/head/main'

- name: Set dev bucket
  run: echo "BUCKET=${{ secrets.AWS_DEVELOPMENT_BUCKET_NAME }}" >> $GITHUB_ENV
  if: github.ref == 'refs/head/deploy-dev-action'
```

In both cases `BUCKET` seems to be an empty string so the deployment fails. Any ideas what could be the cause?

katspaugh: How are you accessing that BUCKET var afterward?
Have you tried this in the deploy step itself?
```
name: Deploy to the dev S3
  env:
    BUCKET: ${{ github.ref == 'refs/head/main' ? STAGING_BUCKET + '/current' : DEV_BUCKET }}
```

usame-algan: Using ternaries like that is corrupting the workflow file thats why I had to opt for `&& ||`. Using `+` to concatenate strings also doesn't seem to be allowed. I was using the variable like so: `${{ env.BUCKET }}` but I believe the issue is that the step to set the variable never runs. I tested the deployment without the optimization on this Branch and it worked fine: https://safe-web-core.dev.5afe.dev/welcome

katspaugh (approved):
Neat!

PR #1053, opened 2022-11-03
feat: replace specific transactions

### LINKED ISSUE #972: [Transaction replacement] Replace button
## What is the feature about
* Replace the "Reject" button with a "Replace" button.
Screenshot%202022-10-11%20at%2012 26 27

* Clicking on the Replace button opens a New Tx modal

Screenshot%202022-10-11%20at%2012 29 51

* New transaction has the nonce of the original transaction pre-filled

## Links
Epic: #971

iamacook:
## What it solves

Resolves #972

## How this PR fixes it

Clicking "Replace" now opens the transaction creation modal in "replacement mode". It is possible to create an asset/NFT transfer with the nonce of the chosen transaction, or alternatively a rejection transaction.

## How to test it

Queue a transaction then click "Replace". Observe that the transaction-to-be-created will be of the same nonce as that which will be replaced.

## Analytics changes

Clicking "Rejection transaction" in the creation modal emits:

```ts
  {
    event: "customClick",
    chainId: "",
    eventCategory: "modals",
    eventAction: "Reject transaction"
  }
```

## Screenshots

Note: the button style has changed since recording this GIF.

replacement

New variant:

image

### DIFF
--- a/src/components/transactions/RejectTxButton/index.tsx
+++ b/src/components/transactions/RejectTxButton/index.tsx
@@ -1,17 +1,19 @@
 import type { TransactionSummary } from '@gnosis.pm/safe-react-gateway-sdk'
 import { Button, Tooltip, SvgIcon } from '@mui/material'
 
-import type { SyntheticEvent } from 'react'
-import { useState, type ReactElement } from 'react'
+import type { SyntheticEvent, ReactElement } from 'react'
+import { useState, Suspense } from 'react'
 import { useQueuedTxByNonce } from '@/hooks/useTxQueue'
 import { isCustomTxInfo, isMultisigExecutionInfo } from '@/utils/transaction-guards'
-import RejectTxModal from '@/components/tx/modals/RejectTxModal'
+import dynamic from 'next/dynamic'
 import useIsPending from '@/hooks/useIsPending'
 import IconButton from '@mui/material/IconButton'
 import ErrorIcon from '@/public/images/notifications/error.svg'
 import useIsSafeOwner from '@/hooks/useIsSafeOwner'
 import Track from '@/components/common/Track'
 import { TX_LIST_EVENTS } from '@/services/analytics/events/txList'
+
+const NewTxModal = dynamic(() => import('@/components/tx/modals/NewTxModal'))
 
 const RejectTxButton = ({
   txSummary,
@@ -42,7 +44,7 @@ const RejectTxButton = ({
     <>
       <Track {...TX_LIST_EVENTS.REJECT}>
         {compact ? (
-          <Tooltip title="Reject" arrow placement="top">
+          <Tooltip title="Replace" arrow placement="top">
             <span>
               <IconButton onClick={onClick} color="error" size="small" disabled={isDisabled}>
                 <SvgIcon component={ErrorIcon} inheritViewBox fontSize="small" />
@@ -51,12 +53,16 @@ const RejectTxButton = ({
           </Tooltip>
         ) : (
           <Button onClick={onClick} variant="danger" disabled={isDisabled} size="stretched">
-            Reject
+            Replace
           </Button>
         )}
       </Track>
 
-      {open && <RejectTxModal onClose={() => setOpen(false)} initialData={[txSummary]} />}
+      {open && (
+        <Suspense>
+          <NewTxModal onClose={() => setOpen(false)} nonce={txNonce} />
+        </Suspense>
+      )}
     </>
   )
 }
--- a/src/components/transactions/RejectTxButton/index.tsx
+++ b/src/components/transactions/RejectTxButton/index.tsx
@@ -60,7 +60,7 @@ const RejectTxButton = ({
 
       {open && (
         <Suspense>
-          <NewTxModal onClose={() => setOpen(false)} nonce={txNonce} />
+          <NewTxModal onClose={() => setOpen(false)} txNonce={txNonce} />
         </Suspense>
       )}
     </>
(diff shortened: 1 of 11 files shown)


katspaugh:
Design-wise, "Rejection transaction" could also be just grey, not red, and say "Empty transaction"? Might be more transparent to the user.
@liliiaorlenko WDYT?

src/components/tx/modals/NewTxModal/index.tsx
```diff
@@ -50,14 +59,21 @@ const NewTxModal = ({ onClose, recipient }: { onClose: () => void; recipient?: s
...
+    trackEvent(MODALS_EVENTS.REJECT_TX)
+    setRejectModalOpen(true)
+  }
+
   const onContractInteraction = () => {
     trackEvent(MODALS_EVENTS.CONTRACT_INTERACTION)
     onClose()
   }
 
+  const dialogTitle = txNonce ? `Replace transaction #${txNonce}` : 'New transaction'
```
schmanu, line 72: Do we use the notation `#<nonce>` anywhere else? Or should we just write `Replace transaction with nonce <nonce>`

src/components/transactions/RejectTxButton/index.tsx
```diff
@@ -42,7 +44,7 @@ const RejectTxButton = ({
     <>
       <Track {...TX_LIST_EVENTS.REJECT}>
         {compact ? (
-          <Tooltip title="Reject" arrow placement="top">
+          <Tooltip title="Replace" arrow placement="top">
             <span>
               <IconButton onClick={onClick} color="error" size="small" disabled={isDisabled}>
                 <SvgIcon component={ErrorIcon} inheritViewBox fontSize="small" />
```
schmanu, line 50: I feel like the `ErrorIcon` does not capture the replacement feature as much as it did the rejection.

src/components/tx/modals/NewTxModal/index.tsx
```diff
@@ -50,14 +59,21 @@ const NewTxModal = ({ onClose, recipient }: { onClose: () => void; recipient?: s
...
+
   return (
     <>
-      <ModalDialog open dialogTitle="New transaction" onClose={onClose}>
+      <ModalDialog open dialogTitle={dialogTitle} onClose={onClose}>
         <DialogContent>
           <Box display="flex" flexDirection="column" alignItems="center" gap={2} pt={7} pb={4} width={240} m="auto">
```
schmanu, line 78: Should we add some info-text here if the nonce is set explaining how replacement works?

Something along the lines of `Replacing a transaction will create a transaction with  the same nonce. To actually replace / reject this new transaction has to get executed to use that nonce`

schmanu (commented):
It works very well and I like the implementation.

I have some small suggestions / questions.

src/components/tx/modals/NewTxModal/index.tsx
iamacook: We don't. I'll adjust this.

src/components/transactions/RejectTxButton/index.tsx
iamacook: @johannesmoormann mentioned that we will work on these designs some more so I think we can revisit this at a later stage.

image

src/components/tx/modals/NewTxModal/index.tsx
iamacook: It looked somewhat odd with text inside the modal. I've added a tooltip. What do you think?

tooltip

iamacook:
> Design-wise, "Rejection transaction" could also be just grey, not red, and say "Empty transaction"? Might be more transparent to the user. 

I much prefer the wording. I've adjusted the `"variant"` to match that of the "Contract interaction" button. I think it looks much better:

image

image

"Contract interaction" for reference:

image

src/components/tx/modals/NewTxModal/index.tsx
schmanu: Nice! I like it.

schmanu (approved):
Nice work! 🚀

src/components/tx/modals/NewTxModal/index.tsx
```diff
@@ -68,8 +107,14 @@ const NewTxModal = ({ onClose, recipient }: { onClose: () => void; recipient?: s
               Send NFTs
             </TxButton>
 
+            {txNonce && (
+              <TxButton onClick={onRejectModalOpen} variant="outlined">
+                Empty transaction
+              </TxButton>
+            )}
```
usame-algan, line 114:
```suggestion
            {txNonce ? (
              <TxButton onClick={onRejectModalOpen} variant="outlined">
                Empty transaction
              </TxButton>
            ) : null}
```

usame-algan: When replacing a tx with nonce 0

src/components/tx/modals/NewTxModal/index.tsx
```diff
@@ -68,8 +107,14 @@ const NewTxModal = ({ onClose, recipient }: { onClose: () => void; recipient?: s
...
+
             {/* Contract interaction via Transaction Builder */}
-            {txBuilder.app && !recipient && (
+            {txBuilder.app && !recipient && !txNonce && (
```
usame-algan, line 117: Same here, this would fail for nonce 0
Screenshot 2022-11-07 at 12 13 09

usame-algan (changes requested):

usame-algan (approved):

iamacook:
I will merge this into the epic branch.

elandrea07 (commented):
approve

PR #1059, opened 2022-11-04
feat: add autocomplete to `NonceForm` nonce

### LINKED ISSUE #973: [Transaction replacement] Make the editable nonce field an autocomplete field with existing nonces
## What is the feature about
In the Advanced Params of a new tx, we allow editing the tx nonce.
Currently, it's just a text field but we want to make it a "free solo" autocomplete field.
That is, you can type a number, or you can choose one of the existing tx nonces.

## Designs/sketches
Screenshot%202022-10-11%20at%2012 33 40

## Links
Epic: #971

iamacook:
## What it solves

Resolves #973

## How this PR fixes it

When creating a transaction, the Safe nonce field now shows autocomplete options with the transaction type which sets the associated nonce on click. Conflicting transactions show the type of the most recently queued one (as we do in the dashboard).

Note: the transaction "info", e.g "withdraw", "n action(s)" is not implemented as it would require a big refactor of `TxInfo` to extract component logic into functions. Should we want to include this, we should include it in a separate PR. After discussion with the team, we think the current implementation is a sufficient MVP.

## How to test it

Ensure transactions are already queued and create a new transaction. When editing the nonce an autocomplete dropdown should allow selection of previous nonces, as well as allowing manual entry. All validation should work as before.

## Screenshots

image

### DIFF
--- a/src/components/dashboard/PendingTxs/PendingTxsList.tsx
+++ b/src/components/dashboard/PendingTxs/PendingTxsList.tsx
@@ -1,12 +1,11 @@
 import type { ReactElement } from 'react'
 import { useMemo } from 'react'
 import { useRouter } from 'next/router'
-import { groupConflictingTxs } from '@/utils/tx-list'
+import { getLatestTransactions } from '@/utils/tx-list'
 import styled from '@emotion/styled'
 import { Box, Skeleton, Typography } from '@mui/material'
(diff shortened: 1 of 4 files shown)


katspaugh (commented):
Looks good!
Can we show what you're sending, e.g. "Send 100 USDT"?

usame-algan (approved):

iamacook:
I will merge this into the epic branch.

PR #1062, opened 2022-11-04
feat: add `SIGNING` event + `safeTxHash` details

### LINKED ISSUE #704: Display `safeTxHash` when signing
## Overview
When signing, the `safeTxHash` is not displayed and it is therefore not possible to verify the validity of it. Currently, you would either have to calculate it yourself or call `getTransactionHash` directly from the contract. Therefore, it would be beneficial to display it in the UI when requesting a signature.

## Notes

- Although it would be nice to display this in the transaction list and used from there as reference, immediately executed transactions do not appear there.
- This is helpful for hardware wallets/wallets which do not support the signing of typed data.

## Requirements
- Emit `safeTxHash` from `TxSender` alongisde a signature request.
- Display it in the UI (likely in the "Please sign the transaction" notification.
- Show the hash in bold if possible

## Designs

Preliminary mock, awaiting final design from @liliiaorlenko

image

## Links
- [Post initial implementation discussion thread](https://5afe.slack.com/archives/C03DAGWJCR5/p1668503195855029)

iamacook:
## What it solves

Resolves #704

## How this PR fixes it

- A new `TxEvent.SIGNING` event is now dispatched before signing a transaction (mirroring `safe-react`).
- The `generateSafeTxHash` function allows for hash generation from transaction data. It is passed to `TxEvent.SIGNING`/`TxEvent.EXECUTING` events and displayed in notification `"Details"` (if present).
- Notifications with `"Details"` now persist

## How to test it

Observe that when signing/executing a transaction, `"Details"` now exist in the dispatched notifications with the `safeTxHash`. Both notifications persist.

## Screenshots

image

image

### DIFF
--- a/src/services/tx/generateSafeTxHash.ts
+++ b/src/services/tx/generateSafeTxHash.ts
@@ -7,9 +7,9 @@ import { isValidSafeVersion } from '@/hooks/coreSDK/safeCoreSDK'
 import ErrorCodes from '../exceptions/ErrorCodes'
 import { logError } from '../exceptions'
 
-export const generateSafeTxHash = (safe: SafeInfo, data: SafeTransactionData): string => {
+export const generateSafeTxHash = (safe: SafeInfo, data: SafeTransactionData): string | undefined => {
   if (!isValidSafeVersion(safe.version)) {
-    return ''
+    return
   }
 
   const typedData = generateTypedData({
@@ -22,7 +22,7 @@ export const generateSafeTxHash = (safe: SafeInfo, data: SafeTransactionData): s
   // `ethers` generates `EIP712Domain` automatically
   const { EIP712Domain: _, ...types } = typedData.types
 
-  let safeTxHash = ''
+  let safeTxHash: string | undefined
 
   try {
     const typedDataHash = utils._TypedDataEncoder.encode(
--- a/src/services/tx/generateSafeTxHash.ts
+++ b/src/services/tx/generateSafeTxHash.ts
@@ -19,6 +19,7 @@ export const generateSafeTxHash = (safe: SafeInfo, data: SafeTransactionData): s
     safeTransactionData: data,
   })
 
+  // `ethers` generates `EIP712Domain` automatically
   const { EIP712Domain: _, ...types } = typedData.types
 
   let safeTxHash = ''
@@ -26,7 +27,7 @@ export const generateSafeTxHash = (safe: SafeInfo, data: SafeTransactionData): s
   try {
     const typedDataHash = utils._TypedDataEncoder.encode(
       typedData.domain,
-      // @ts-ignore - typedData.types.SafeTxData key is not recognised as a string
+      // @ts-ignore - the typedData.types.SafeTxData key is not recognised as a string
       types,
       typedData.message,
     )
--- a/src/services/tx/generateSafeTxHash.ts
+++ b/src/services/tx/generateSafeTxHash.ts
@@ -1,0 +1,39 @@
+import { utils } from 'ethers'
+import { generateTypedData } from '@gnosis.pm/safe-core-sdk-utils'
+import type { SafeTransactionData } from '@gnosis.pm/safe-core-sdk-types'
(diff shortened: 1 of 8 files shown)


src/services/tx/txEvents.ts
```diff
@@ -24,6 +25,7 @@ export enum TxEvent {
 type Id = { txId: string; groupKey?: string } | { txId?: string; groupKey: string }
 
 interface TxEvents {
+  [TxEvent.SIGNING]: { txId?: string; safeTxHash?: string }
```
iamacook, line 28: We could leverage the `safeTxHash` as a more stable `id`. Out of scope here but worth revisiting in the future.

src/services/tx/generateSafeTxHash.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  const typedData = generateTypedData({
+    safeAddress: safe.address.value,
+    safeVersion: safe.version,
+    chainId: Number(safe.chainId),
+    safeTransactionData: data,
+  })
+
+  const { EIP712Domain: _, ...types } = typedData.types
```
iamacook, line 22: `ethers` generates `EIP712Domain` itself.

src/services/tx/generateSafeTxHash.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { SafeTransactionData } from '@gnosis.pm/safe-core-sdk-types'
+import type { SafeInfo } from '@gnosis.pm/safe-react-gateway-sdk'
+
+import { isValidSafeVersion } from '@/hooks/coreSDK/safeCoreSDK'
+import ErrorCodes from '../exceptions/ErrorCodes'
+import { logError } from '../exceptions'
+
+export const generateSafeTxHash = (safe: SafeInfo, data: SafeTransactionData): string => {
+  if (!isValidSafeVersion(safe.version)) {
+    return ''
```
schmanu, line 12: Should this not also be logged in some way as an error?

Or should this method probably return `undefined` if the hash creation fails? Currently when we call this we do not check if the returned value is an empty string.

src/services/tx/generateSafeTxHash.ts
```diff
@@ -0,0 +1,40 @@
+import { utils } from 'ethers'
+import { generateTypedData } from '@gnosis.pm/safe-core-sdk-utils'
+import type { SafeTransactionData } from '@gnosis.pm/safe-core-sdk-types'
+import type { SafeInfo } from '@gnosis.pm/safe-react-gateway-sdk'
+
+import { isValidSafeVersion } from '@/hooks/coreSDK/safeCoreSDK'
+import ErrorCodes from '../exceptions/ErrorCodes'
+import { logError } from '../exceptions'
+
+export const generateSafeTxHash = (safe: SafeInfo, data: SafeTransactionData): string => {
```
schmanu, line 10: I think we should have a jest test for this.

schmanu (commented):
It's looking good. I only have one question about error handling / testing.

src/services/tx/generateSafeTxHash.ts
iamacook: > Should this not also be logged in some way as an error?
> 
> Or should this method probably return `undefined` if the hash creation fails? Currently when we call this we do not check if the returned value is an empty string.

`isValidSafeVersion` should perhaps be an "invariant" function but I think we should revisit this in a separate issue.

I have now changed `generateSafeTxHash` to return `undefined` if the Safe version is invalid/generation is unsuccessfuly. `useTxNotifications` also only shows the `safeTxHash` if it is defined.

schmanu (approved):
Good work! Looking good :)

liliya-soroka:
- [ ] 1 of N safe: creation+ execution
- [ ] 1 of N : only creation
- [ ] 1 of N : only execution
- [ ] N of M: creation
- [ ] N of M: signing
- [ ] N of M: execution
- [ ] Hard wallets/Mobile pairing
- [ ] MM wallet

PR #1063, opened 2022-11-04
[Safe Creation] Layout adjustments

usame-algan:
## What it solves

- Adjust button layout for safe creation steps
- Remove padding from Stepper and instead handle row spacing of each view with a single css class and optimize it for smaller screens
- Remove negative margins from `Divider` elements
- Improve `NetworkSelector` clickable area
- Adjust Threshold select size and color

## How to test it

- Click through each create safe step on `/new-safe/create` and compare with Design

## Screenshots

Screenshot 2022-11-04 at 15 43 46

Screenshot 2022-11-04 at 15 44 56
Screenshot 2022-11-04 at 15 45 15
Screenshot 2022-11-04 at 15 45 37

### DIFF
--- a/src/components/common/Header/styles.module.css
+++ b/src/components/common/Header/styles.module.css
@@ -37,7 +37,8 @@
 }
 
 .networkSelector {
-  padding-right: var(--space-1);
+  padding-right: 0;
+  padding-left: 0;
   border-right: none;
 }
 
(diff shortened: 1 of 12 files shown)


iamacook (commented):
I've been through the flow and it looks great. There are a couple things things I noticed.

The threshold field isn't centre aligned:

image

It's not in the design but this dialog content doesn't seem to "match". Maybe we need to move the "Creating a Safe on" into a select like the second step. What do you think? cc @liliiaorlenko 

image

usame-algan:
I discussed the second point with @liliiaorlenko and created a separate task for it: https://github.com/safe-global/web-core/issues/1080

PR #1066, opened 2022-11-04
feat(safe-apps): Add support for receiving analytics from our Safe Apps

yagopv:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/987

## How this PR fixes it
We add support for receiving analytics from Safe Apps

## How to test it
This need to be combined with the corresponding [wallet connect changes](https://github.com/safe-global/safe-react-apps/pull/566/files)

Once you add the wallet connect app from that PR then you can start testing:

1) Open the Wallet Connect Safe App
2) Connect any third party application opened in a new Tab
3) Each time 2) happen and a session starts an Analytic event "WalletConnect: New session" should be sent. The `safeAppName` dimension must be the name of the app (if is an app from the default list) or the URL if is a custom one.
  Only the session start should be sent so if we close the Safe and open it again with the same session nothing is sent
4) When a transaction is initialised and the popup shows in the Web a "WalletConnect: Transaction sent" is tracked


DaniSomoza (approved):

dasanra (approved):

mikhailxyz (approved):

DaniSomoza (approved):

DashaBochkar (commented):
Thanks

PR #1067, opened 2022-11-04
feat: unsigned warning

### LINKED ISSUE #898: Show a security warning on the Single Tx page if the tx is unsigned
## What is the feature about
It's possible to propose unsigned transactions that won't appear in the queue but are still visible when opened via a direct link.

This can be potentially misused so it would be nice to display a warning that the transaction is not signed and is "untrusted".

iamacook:
## What it solves

Resolves #898

## How this PR fixes it

A new warning is shown when a proposed transaction has no signatures.

## How to test it

Open `/gor:0xF3E977e7Eea1A91ce5B2d5077e5A7195Ae18b722/transactions/tx?id=multisig_0xF3E977e7Eea1A91ce5B2d5077e5A7195Ae18b722_0xc0c5380e113ecff6054c69d02fe7d5ee753db29635548c80794bd028cf614bb5` and observe the warning.

## Screenshots

image

### DIFF
--- a/src/components/transactions/TxDetails/index.tsx
+++ b/src/components/transactions/TxDetails/index.tsx
@@ -27,7 +27,7 @@ import SignTxButton from '@/components/transactions/SignTxButton'
 import RejectTxButton from '@/components/transactions/RejectTxButton'
 import useWallet from '@/hooks/wallets/useWallet'
 import useIsWrongChain from '@/hooks/useIsWrongChain'
-import { DelegateCallWarning } from '@/components/transactions/Warning'
+import { DelegateCallWarning, UnsignedWarning } from '@/components/transactions/Warning'
 import Multisend from '@/components/transactions/TxDetails/TxData/DecodedData/Multisend'
 
 export const NOT_AVAILABLE = 'n/a'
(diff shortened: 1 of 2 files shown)


katspaugh:
And text-wise, and I would write something like “This transaction is unsigned. To avoid phishing, only sign it if you trust the source of the link.”
Perhaps too grim?

iamacook:
> IMHO this warning isn't visible enough. I would write it in big red letters above the transaction. Perhaps in an `<Alert>`

This follows the style of our other warnings. I'll ask design and see what they think.

DiogoSoaress (approved):
Code looks good and if this is aligned with designer 👍 

I agree with this comment however
>IMHO this warning isn't visible enough. I would write it in big red letters above the transaction. Perhaps in an <Alert>

though an unsigned transaction can come from a friendly source, I think it would be worth to make it more prominent (as suggested) and thus not closing the issue just with this PR.

iamacook:
> I agree with this comment however
> 
> > IMHO this warning isn't visible enough. I would write it in big red letters above the transaction. Perhaps in an
> 
> though an unsigned transaction can come from a friendly source, I think it would be worth to make it more prominent (as suggested) and thus not closing the issue just with this PR.

After syncing with @liliiaorlenko, we intend to document our transaction list in detail and then reassess all transaction-related notifications thereafter.

katspaugh:
I was talking about the Single Tx page btw. The list is safe, we never load untrusted txns there. But anyone can create and share a deeplink.

katspaugh (approved):

PR #1070, opened 2022-11-07
Fix: show upgrade notification only to owners

katspaugh:
## What it solves
The upgrade notification is only actionable if you're the owner.
Noticed this when opening the demo safe.

## How to test
1)
* Open a 1.1.1 Safe that doesn't belong to you (or simply disconnect the wallet)
* The upgrade notification should NOT pop up

2)
* Open a 1.1.1 Safe as an owner
* Observe the notification


iamacook (approved):

francovenica:
I cannot see the notification even with the owner of the safe, but also I cannot see the "update" button at all, so maybe that's why the notification is not popping up.

image

katspaugh:
@francovenica it's not ready for QA.

francovenica:
I'm still having the same issue as before: no notification and the update button is not there
image

katspaugh:
@francovenica you're probably using the staging CGW. It sadly doesn't detect outdated Safe versions.

cc @safe-global/safe-services

fmrsabino:
> @francovenica you're probably using the staging CGW. It sadly doesn't detect outdated Safe versions.
> 
> cc @safe-global/safe-services

I suspect this is more of a data related issue that should be checked. The production deployment is the same as the staging one (minus some minor commits going into the next release).

So it should detect outdated safes in the same way as the production one. The data used by these checks might be the differing factor.

@francovenica can you share the Safe Address, chain and environment being used? Thanks!

fmrsabino:
> @fmrsabino here's an example:
> 
> * Prod, shows OUTDATED correctly: https://safe-client.safe.global/v1/chains/1/safes/0x85C9f5aA0F82A531087a356a55623Cf05E7Bb895
> * Staging: UP_TO_DATE https://safe-client.staging.5afe.dev/v1/chains/1/safes/0x85C9f5aA0F82A531087a356a55623Cf05E7Bb895

The `recommendedMasterCopyVersion` on Staging is `1.1.1`: https://safe-config.staging.5afe.dev/api/v1/chains/1/

Last time it was changed was on Aug. 3, 2021. I can set this value to be the same as production but maybe it was set to this value for testing purposes.

katspaugh:
Oh I see, yes, pelase @fmrsabino!

fmrsabino:
> Oh I see, yes, please @fmrsabino!

Updated ✅ https://safe-config.safe.global/api/v1/chains/1/

katspaugh:
Thanks Fred! It shows OUTDATED correctly now 👍 

@francovenica please give it another go.

francovenica:
Now it works fine.
The notification is only shown to a connected user who is also owner of that safe.

The Update button is there again as well
Tested with the CGW switch on and off and it works fine for both.

PR #1078, opened 2022-11-07
fix: set default `recipient`

### LINKED ISSUE #1069: [Soft release] Send funds form crashes the page
## Bug description
Just trying to open the send funds form crashes the page
Tryied with a ledget and MM, in ethereum and goerli, so the device and network seems to be irrelevant

## Environment
 - Browser: Chrome
 - Wallet: Any
 - Chain: Any

Main ticket: #1039 

## Steps to reproduce
 1. Go to a safe you own
 2. Click New transactions > Send funds

## Screenshots
crash

iamacook:
## What it solves

Resolves #1069 

## How this PR fixes it

We [set the `initialData` of `TokenTransferModal` and `NftTransferModal`](https://github.com/safe-global/web-core/blob/dev/src/components/tx/modals/NewTxModal/index.tsx#L88-L91) when [sending to a contact from the address book](https://github.com/safe-global/web-core/blob/dev/src/components/address-book/AddressBookTable/index.tsx#L163). 

We weren't setting a default `recipient` value [for our `formData`](https://github.com/safe-global/web-core/blob/dev/src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx#L72). An empty string is now used by default.

## How to test it

Unable to replicate.

### DIFF
--- a/src/components/tx/modals/NewTxModal/index.tsx
+++ b/src/components/tx/modals/NewTxModal/index.tsx
@@ -34,7 +34,7 @@ const useTxBuilderApp = (): { app?: SafeAppData; link: UrlObject } => {
   }
 }
 
-const NewTxModal = ({ onClose, recipient }: { onClose: () => void; recipient?: string }): ReactElement => {
+const NewTxModal = ({ onClose, recipient = '' }: { onClose: () => void; recipient?: string }): ReactElement => {
   const [tokenModalOpen, setTokenModalOpen] = useState<boolean>(false)
   const [nftsModalOpen, setNftModalOpen] = useState<boolean>(false)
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
👍

katspaugh (approved):

iamacook:
> I noticed that in the address book section an entry of undefined address is present event after the fix. But this fix that you pushed now fixed the underlying issue related to 'Send Tokens'.

Can you confirm whether an `undefined` entry is still being migrated? You will need to clear your `localStorage` on `app.safe.global` first. We've [added validation](https://github.com/safe-global/web-core/pull/1081) that should prevent this.

PR #1079, opened 2022-11-07
[Epic] Transaction replacement

### LINKED ISSUE #971: [Epic] Transaction replacement (v1)
## What is the feature about
Users complain that canceling/replacing transactions is currently confusing.

## The list of requirements

* [x] https://github.com/safe-global/web-core/issues/972
* [x] https://github.com/safe-global/web-core/issues/973
* [x] https://github.com/safe-global/web-core/issues/718
* [ ] Add info tips about irrevokable signatures

## Links
Roadmap item: https://www.notion.so/gnosis-safe/Replacing-cancelling-existing-transactions-in-the-queue-is-cumbersome-3e3880dc709f4e25b34d80c57df0983c
Slides: https://docs.google.com/presentation/d/1tvKBh6NbVf3GYc0_IL5aFwup-yWSz19JSpjhzqLpuPI/edit

iamacook:
## What it solves

Resolves #971

## How this PR fixes it

* https://github.com/safe-global/web-core/issues/972
* https://github.com/safe-global/web-core/issues/973
* https://github.com/safe-global/web-core/issues/718

## How to test it

The above issues contain individual testing procedures.


francovenica:
Testing still in progress. The main flow looks fine for now.

Minor remarks:
- The dropdown of nonces has its text blending with whatever text is behind. Maybe a border and a bit of padding to the right would help
image

- The "Already Exist" text is redundant after a while, and also doesn't allow to show fully what is in "( )" which I'd argue is more important to show fully

- I'd suggest changing red X icon for some circular arrows like this one: 
image
image

- For the button "Replace" mabye we can use the same style the "Reject Transaction" button in the new modal: White button with black border and black letters, feels less aggresive than the current red color

image
image

Note that the red X Icon is used for tx on chain rejections. Is fine for those to remain there.
image

usame-algan:
> The dropdown of nonces has its text blending with whatever text is behind. Maybe a border and a bit of padding to the right would help

I would suggest we add the normal backdrop like we do for other selects

iamacook:
>     * The dropdown of nonces has its text blending with whatever text is behind.

I have added a backdrop similar to modals as per @usame-algan's suggestion.

image

image

>     * The "Already Exist" text is redundant after a while, and also doesn't allow to show fully what is in "( )" which I'd argue is more important to show fully

I have removed "already exists". It now shows "n (xyz transaction)". Please see the screenshots above.

>     * I'd suggest changing red X icon for some circular arrows.

That would indicate repeating something to me. Replacing a transaction ≈ cancelling it in my eyes so I think it's currently correct but let's see what @liliiaorlenko thinks.

>     * For the button "Replace" mabye we can use the same style the "Reject Transaction" button in the new modal: White button with black border and black letters, feels less aggresive than the current red color

The button is according to design (the link for which no longer works?). I would argue that the "less agressive" colour is what we want though. Users would more than likely want to exeute a "proper" transaction over a rejection. Let's see what design thinks regarding this as well though.

francovenica:
Issues:

1 - If you choose to use Replace > spending limits, during the review step the system allows you to edit the nonce and other advance parameters. This should not happen.

See in the gif how doing Spending limits normally would not allow you to open the adv options, but doing it through the replacement button it allows it.


2 - There is a flicker effect that happens while you have the dropdown with the different options for nonce open. 
Have a few nonces in the list, open it, wait for a bit and you will see it 
flicker

3 - It seems that the tx nonce=0 cannot be rejected. After opening the "replace" modal, clicking on "Reject transaction" does nothing. There are no errors in the console or network tab
When you click the "Reject transaction" button a " 0 " shows next to the replace button in the tx overview:
image

francovenica:
Issue:

Once you create a rejection and Queue it, the "replace" button is gone, even tho it should be reasonalble to replace even the rejection. (probably the replace button is just following the logic of the old reject button)
What shouldn't be available anymore is the "Reject button" in the "Replace tx modal". The button could be grayed out the tooltip saying "this nonce already has a rejection proposed"

image
image

francovenica:
LGTM now. The issues where fixed

katspaugh (approved):
Great work @iamacook!

PR #1081, opened 2022-11-07
fix: validate migrated address book

iamacook:
## What it solves

Resolves invalid address book entry migration

## How this PR fixes it

The validity of migrated a migrated address/name is checked prior to entry insertion.

## How to test it

Add an invalid address and/or empty name on the old deployment. Clear the migration flag and observe the invalid entries not migrate to the new deployment.

### DIFF
--- a/src/services/ls-migration/addressBook.ts
+++ b/src/services/ls-migration/addressBook.ts
@@ -1,4 +1,5 @@
 import { type AddressBookState } from '@/store/addressBookSlice'
+import { utils } from 'ethers'
 import type { LOCAL_STORAGE_DATA } from './common'
 import { parseLsValue } from './common'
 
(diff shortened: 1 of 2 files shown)


src/services/ls-migration/addressBook.ts
```diff
@@ -12,6 +13,9 @@ export const migrateAddressBook = (lsData: LOCAL_STORAGE_DATA): AddressBookState
     console.log('Migrating address book')
 
     const newAb = legacyAb.reduce<AddressBookState>((acc, { address, name, chainId }) => {
+      if (!name || !utils.isAddress(address)) {
```
katspaugh, line 16: I would add `!name || !address || ` in the beginning because isAddress seems to accept only strings.

katspaugh (approved):

PR #1085, opened 2022-11-07
fix: switching hardware wallet networks

### LINKED ISSUE #1073: Unable to execute transactions on Goerli via Trezor & Ledger
## Bug description
Executing a tx via Ledger and Trezor on Goerli is not possible at the moment.

Note: this also happens on the legacy interface

## Environment
 - Browser: Chrome
 - Wallet: Ledger & Trezor
 - Chain: Goerli

## Steps to reproduce
Try to create & execute any transaction using these devices

## Expected result
Possible to execute transactions from Ledger & Trezor.

## Obtained result
Executing & signing + executing a tx is throwing this error for both devices:
ledger

No issues when only signing.

iamacook:
## What it solves

Resolves #1073

## How this PR fixes it

If a hardware wallet is connected, it is diconnected and reconnected when switching network as it is not possible to change the network of a hardware wallet.

## How to test it

Open the Safe on a network where a Ledger/Trezor is enabled and connect it. Switch to another network that also supports Ledger/Trezor. Observe the address scanner modal open.


src/components/common/ChainSwitcher/index.tsx
```diff
@@ -15,7 +16,17 @@ const ChainSwitcher = ({ fullWidth }: { fullWidth?: boolean }): ReactElement | n
...
     if (!chain) return
     const chainId = hexValue(parseInt(chain.chainId))
-    onboard?.setChain({ chainId })
+
+    const wallets = onboard?.state.get()?.wallets || []
+    const wallet = getConnectedWallet(wallets)
+
+    if (wallet && isHardwareWallet(wallet)) {
+      onboard?.disconnectWallet({ label: wallet.label }).then(() => {
+        connectWallet(onboard, { autoSelect: wallet.label })
```
iamacook, line 25: Note: the address scanner modal opens on Ethereum by default and you must manually select the current network. I will investigate setting this (it would need to be implemented in onboard, unless we reinitialise onboard with 1 chain configuration before connecting a hardware wallet).

image

katspaugh (approved):
👍

katspaugh:
@francovenica please test this with Keystone as well.

francovenica:
> @francovenica please test this with Keystone as well.

I don't have a keystone (those are illegal in Argentina :D ). @liliya-soroka I think you have one

francovenica:
I tried both with trezor one and ledger NanoS and in both cases the connection persisted while changing networks. 
It is expected to disconnect and makes you reconnect manually choosing the new network.

trezor connect

francovenica:
Miss understood the moment were it disconnects. You actually have to click "Change to NETWORK" for the disconnect to happen, which is correct. So disregard my last comment

francovenica:
I've checked that the onboard doesn't allow the user to connect the device to a network where is not compatible. Example: connection to ETH with trezor, then switching to aurora and clicking the button "Switch to Aurora" shows an error message in the onboard, which is the correct behavior
image

francovenica:
LGTM
Tested with a Trezor One and a ledger NanoS. Started in the Eth network and change it to goerli, setting that network in the onboard js.
I was able to create a safe and add a new owner for both devices.

PR #1086, opened 2022-11-07
fix: adjust modals to hug long addresses

### LINKED ISSUE #1046: [Soft release] Minor visual issue on tx pop up due to large chain prefix name
## Bug description
The 'Sending from' section does not fit entire due to the network prefix being on the larger end.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Polygon, Aurora

## Steps to reproduce
I was able to reproduce this on a macbook pro 16 inch & 27 inch 2k monitor screens. 

Create a tx via: send funds, send NFT or tx-builder safe app.

Safe used on Polygon: `matic:0x80C13FCb96A79Aa608e44Fd47C5910D99548e3AE`

Safe used on Aurora:
`aurora:0xCEA454dD3d76Da856E72C3CBaDa8ee6A789aD167`

(make sure that Prepend chain prefix to addresses is checked in the settings)

## Expected result
No visual issues.

## Obtained result
Minor visual issues with the link to the explorer not fitting on screen.
Example 1
Screenshot 2022-11-03 at 11 46 46

Example 2
Screenshot 2022-11-03 at 11 50 14

DiogoSoaress:
## What it solves

Resolves #1046

## How this PR fixes it
Sets min-width in the `TxModal` and reduces the addresses font-size to always have addresses and buttons visible inside the Modal.

## How to test it
2. Open create a new transaction modal.
3. All legit addresses (40 chars) represented in the modal shall not push the buttons out of visible area.

## Analytics changes
N/A

## Screenshots
Screenshot 2022-11-09 at 18 21 12

### DIFF
--- a/src/components/common/EthHashInfo/index.tsx
+++ b/src/components/common/EthHashInfo/index.tsx
@@ -70,10 +70,12 @@ const EthHashInfo = ({
         )}
 
         <Box className={css.addressRow}>
-          <Typography variant="body2" fontWeight="inherit" component="div" className={css.address}>
+          <Typography variant="body2">
             {showPrefix && shouldPrefix && prefix && <b>{prefix}:</b>}
-            <span className={css.mobileAddress}>{shortenAddress(address)}</span>
-            <span className={css.desktopAddress}>{shortAddress ? shortenAddress(address) : address}</span>
+            <Box className={css.address} component="span">
+              <span className={css.mobileAddress}>{shortenAddress(address)}</span>
+              <span className={css.desktopAddress}>{shortAddress ? shortenAddress(address) : address}</span>
+            </Box>
           </Typography>
 
           {showCopyButton && (
--- a/src/components/common/EthHashInfo/index.tsx
+++ b/src/components/common/EthHashInfo/index.tsx
@@ -70,12 +70,10 @@ const EthHashInfo = ({
         )}
 
         <Box className={css.addressRow}>
-          <Typography variant="body2">
+          <Typography variant="body2" fontWeight="inherit">
             {showPrefix && shouldPrefix && prefix && <b>{prefix}:</b>}
-            <Box component="span">
-              <span className={css.mobileAddress}>{shortenAddress(address)}</span>
-              <span className={css.desktopAddress}>{shortAddress ? shortenAddress(address) : address}</span>
-            </Box>
+            <span className={css.mobileAddress}>{shortenAddress(address)}</span>
+            <span className={css.desktopAddress}>{shortAddress ? shortenAddress(address) : address}</span>
           </Typography>
 
           {showCopyButton && (
--- a/src/components/common/EthHashInfo/index.tsx
+++ b/src/components/common/EthHashInfo/index.tsx
@@ -72,7 +72,7 @@ const EthHashInfo = ({
         <Box className={css.addressRow}>
           <Typography variant="body2">
             {showPrefix && shouldPrefix && prefix && <b>{prefix}:</b>}
-            <Box className={css.address} component="span">
+            <Box component="span">
               <span className={css.mobileAddress}>{shortenAddress(address)}</span>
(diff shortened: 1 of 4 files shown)


katspaugh (commented):
IMHO monospace doesn't look that good. Surely there are other solutions, e.g. make the popup bigger, or reduce the letter-spacing, etc.

src/components/common/ModalDialog/index.tsx
```diff
@@ -1,34 +1,17 @@
...
-      content: '""',
-      order: 1,
-      flex: 1,
-    },
-  },
-  '& .MuiDialogTitle-root': {
-    borderBottom: `2px solid ${theme.palette.divider}`,
-  },
-}))
-
```
DiogoSoaress, line 31: I've moved this code to the css.module

src/components/common/EthHashInfo/index.tsx
```diff
@@ -70,10 +70,12 @@ const EthHashInfo = ({
         )}
 
         <Box className={css.addressRow}>
-          <Typography variant="body2" fontWeight="inherit" component="div" className={css.address}>
+          <Typography variant="body2">
```
iamacook, line 73: This doesn't inherit the `font-weight` anymore.

src/components/common/EthHashInfo/styles.module.css
```diff
@@ -2,11 +2,6 @@
   display: flex;
   align-items: center;
   gap: 0.5em;
-}
-
-.container,
-.address {
-  font-size: inherit;
```
iamacook, line 9: This doesn't inherit the `font-size` anymore.

src/components/common/ModalDialog/index.tsx
```diff
@@ -70,15 +53,21 @@ const ModalDialog = ({
   const fullScreen = useMediaQuery(theme.breakpoints.down('sm'))
 
   return (
-    <StyledDialog {...restProps} fullScreen={fullScreen} scroll="body" onClick={(e) => e.stopPropagation()}>
+    <Dialog
+      {...restProps}
+      fullScreen={fullScreen}
+      scroll={fullScreen ? 'paper' : 'body'}
```
iamacook, line 59: What's the reason behind switching the scroll container?

src/components/common/EthHashInfo/index.tsx
```diff
@@ -70,10 +70,12 @@ const EthHashInfo = ({
...
             {showPrefix && shouldPrefix && prefix && <b>{prefix}:</b>}
-            <span className={css.mobileAddress}>{shortenAddress(address)}</span>
-            <span className={css.desktopAddress}>{shortAddress ? shortenAddress(address) : address}</span>
+            <Box component="span">
+              <span className={css.mobileAddress}>{shortenAddress(address)}</span>
+              <span className={css.desktopAddress}>{shortAddress ? shortenAddress(address) : address}</span>
+            </Box>
```
iamacook, line 78: These need not be wrapped in another `span`. `mobileAddress` and `desktopAddress` inversely display based on the resolution.

src/components/common/EthHashInfo/index.tsx
DiogoSoaress: It was repeated in the `.address` css

src/components/common/EthHashInfo/styles.module.css
DiogoSoaress: The approach from this PR reduces the font-size of the address and prefix form 16px to 14px so even the longest `<short_name>:<address>` can be contained inside the Modal's 600px

src/components/common/ModalDialog/index.tsx
DiogoSoaress: It was the way I've found to solve a faulty behavior in responsiveness. I've considered it because we can scroll the whole paper if we are in fullScreen. I think the `scroll="body"` property aims to target scrolling solely in _> 600px_ resolutions.

iamacook (approved):

PR #1087, opened 2022-11-07
[Create Safe] Handle wallet connection

### LINKED ISSUE #1080: [Safe Creation] Handle wallet connection
## What is the feature about

Handle wallet connection during safe creation flow

## The list of requirements

- Navigate the user to the first step if they disconnect their wallet
- Disable the Next button if the wallet network changes
- On the first step, navigate the user to the second step on wallet connection and network change
- Hide Back button in the second step
- Hide Next button in the first step
- Set connected wallet as the first owner

## Designs/sketches
Screenshot 2022-11-07 at 14 14 43

Figma: https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=1964%3A53663

usame-algan:
## What it solves

Resolves #1080 

## How this PR fixes it

- Removes the Cancel and Next button in the first step
- Removes the Back button in the second step
- Navigates the user to the first step if they disconnect their wallet
- Disables the Next button and displays an info box if the connected wallet is on the wrong network
- Always populates the first owner with the connected wallet

### Other changes

- Removed the `useIsConnected` hook again since it isn't being used anymore
- Removed RHF from the third step since it is read-only
- Takes `formState` into account for disabling the Next button

## How to test it

1. Create a new safe
2. Connect a wallet
3. Observe being navigated to step 2
4. No Back button visible
5. Disconnect wallet
6. Observe being navigated to step 1
7. No Cancel or Next button visible
8. Connect wallet that is on a different network
9. Observe being navigated to step 2 with the info box showing
10. Observe the Next button being disabled

## Screenshots

Screenshot 2022-11-07 at 17 01 53

Screenshot 2022-11-07 at 17 01 41


src/components/new-safe/CreateSafe/useSetCreationStep.ts
```diff
@@ -4,20 +4,22 @@ import type { StepRenderProps } from '@/components/new-safe/CardStepper/useCardS
 import type { PendingSafeData } from '@/components/new-safe/steps/Step4'
 import type { NewSafeFormData } from '@/components/new-safe/CreateSafe/index'
 import { SAFE_PENDING_CREATION_STORAGE_KEY } from '@/components/new-safe/steps/Step4'
+import useWallet from '@/hooks/wallets/useWallet'
 
-const useSetCreationStep = (setStep: StepRenderProps<NewSafeFormData>['setStep'], isConnected: boolean) => {
+const useSetCreationStep = (setStep: StepRenderProps<NewSafeFormData>['setStep']) => {
```
iamacook, line 9: Nit: can we rename this to be a bit more declarative, e.g. `useSyncSafeCreationStep`.

src/components/new-safe/steps/Step2/index.tsx
```diff
@@ -48,12 +49,15 @@ const CreateSafeStep2 = ({
...
   })
 
-  const { register, handleSubmit, control, watch } = formMethods
+  const { register, handleSubmit, control, watch, formState } = formMethods
 
   const allFormData = watch()
 
   const { fields: ownerFields, append: appendOwner, remove: removeOwner } = useFieldArray({ control, name: 'owners' })
 
+  const isValid = Object.keys(formState.errors).length === 0
```
iamacook, line 58: Does `formState.isValid` not work? If it's `true` from the start, I'd suggest we call this `hasErrors` or something instead.

usame-algan: I copied this from the old form where it had a comment saying: `// do not use formState.isValid because names can be empty`

iamacook: Can you try `isValid` again as we updated RHF? If it doesn't work, it'd be nice to keep the comment.

src/components/new-safe/CreateSafe/useSyncSafeCreationStep.ts
```diff
@@ -4,20 +4,22 @@ import type { StepRenderProps } from '@/components/new-safe/CardStepper/useCardS
 import type { PendingSafeData } from '@/components/new-safe/steps/Step4'
 import type { NewSafeFormData } from '@/components/new-safe/CreateSafe/index'
 import { SAFE_PENDING_CREATION_STORAGE_KEY } from '@/components/new-safe/steps/Step4'
+import useWallet from '@/hooks/wallets/useWallet'
 
-const useSetCreationStep = (setStep: StepRenderProps<NewSafeFormData>['setStep'], isConnected: boolean) => {
+const useSyncSafeCreationStep = (setStep: StepRenderProps<NewSafeFormData>['setStep']) => {
```
schmanu, line 9: I think even though it's quite trivial we should write a small test for this.

schmanu (approved):
Good changes :) Just a small suggestion if there is time but no blocker otherwise.

src/components/new-safe/steps/Step2/index.tsx
usame-algan: Changed it to `formState.isValid` and it works fine, thanks!

src/components/new-safe/CreateSafe/useSyncSafeCreationStep.ts
usame-algan: done

PR #1092, opened 2022-11-08
fix: sanitize migrated address books

iamacook:
## What it solves

Resolves malformed address books

## How this PR fixes it

Initial address book migration [did not include validation](https://github.com/safe-global/web-core/pull/1081). As we have since seen reports of errors being thrown, this not sanitises address books post-migration.

## How to test it

Add invalid address book entries in the localStorage, i.e. empty names or invalid addresses and delete the `SAFE_v2__sanitizeFinished` key. Refresh the page and observe the invalid entries removed from the `localStorage`.

Note: empty chain-specific address books (post-sanitization) should not be present (the `chainId` should not exist in the address book object) and entirely invalid address books should be an empty object.

### DIFF
--- a/src/services/ls-migration/addressBook.ts
+++ b/src/services/ls-migration/addressBook.ts
@@ -30,24 +30,20 @@ export const migrateAddressBook = (lsData: LOCAL_STORAGE_DATA): AddressBookState
 
 // Temporary post-migration fix for malformed data
 export const sanitizeMigratedAddressBook = (state: AddressBookState): AddressBookState => {
-  const hasValidAb = Object.values(state).every((ab) => {
-    return Object.keys(ab).every(utils.isAddress)
-  })
-
-  if (hasValidAb) {
-    return state
-  }
+  console.log('Sanitizing address book')
 
   const sanitizedAb = Object.keys(state).reduce<AddressBookState>((sanitizedAb, chainId) => {
     const chainAb = state[chainId]
 
     const sanitizedChainAb = Object.keys(chainAb).reduce<AddressBook>((sanitizedChainAb, address) => {
       if (!address || !utils.isAddress(address)) {
+        console.log(`Removing malformed address book entry (invalid address): ${address}`)
         return sanitizedChainAb
       }
 
-      const name = chainAb[address]
+      const name = chainAb[address]?.trim()
       if (!name) {
+        console.log(`Removing malformed address book entry (invalid name): ${address}`)
         return sanitizedChainAb
       }
 
--- a/src/services/ls-migration/addressBook.ts
+++ b/src/services/ls-migration/addressBook.ts
@@ -1,4 +1,5 @@
-import { type AddressBookState } from '@/store/addressBookSlice'
+import { initialState } from '@/store/addressBookSlice'
+import type { AddressBook, AddressBookState } from '@/store/addressBookSlice'
 import { utils } from 'ethers'
 import type { LOCAL_STORAGE_DATA } from './common'
 import { parseLsValue } from './common'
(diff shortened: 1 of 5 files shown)


usame-algan:
I tried both removing names and corrupting the address in local storage on the new app but the page renders without any errors after refresh. How to reproduce an error there?
Screenshot 2022-11-08 at 11 52 43

src/services/ls-migration/addressBook.ts
```diff
@@ -26,3 +27,45 @@ export const migrateAddressBook = (lsData: LOCAL_STORAGE_DATA): AddressBookState
...
+  const sanitizedAb = Object.keys(state).reduce<AddressBookState>((sanitizedAb, chainId) => {
+    const chainAb = state[chainId]
+
+    const sanitizedChainAb = Object.keys(chainAb).reduce<AddressBook>((sanitizedChainAb, address) => {
+      if (!address || !utils.isAddress(address)) {
+        return sanitizedChainAb
+      }
+
+      const name = chainAb[address]
+      if (!name) {
```
schmanu, line 50: What about a name only consisting of whitespace?

src/services/ls-migration/addressBook.ts
```diff
@@ -26,3 +27,45 @@ export const migrateAddressBook = (lsData: LOCAL_STORAGE_DATA): AddressBookState
     }
   }
 }
+
+// Temporary post-migration fix for malformed data
+export const sanitizeMigratedAddressBook = (state: AddressBookState): AddressBookState => {
+  const hasValidAb = Object.values(state).every((ab) => {
+    return Object.keys(ab).every(utils.isAddress)
```
schmanu, line 34: We do not check the `name` in this condition.
This must be why the unit test also fails

src/services/ls-migration/addressBook.ts
```diff
@@ -26,3 +27,45 @@ export const migrateAddressBook = (lsData: LOCAL_STORAGE_DATA): AddressBookState
...
+
+  if (hasValidAb) {
+    return state
+  }
+
+  const sanitizedAb = Object.keys(state).reduce<AddressBookState>((sanitizedAb, chainId) => {
+    const chainAb = state[chainId]
+
+    const sanitizedChainAb = Object.keys(chainAb).reduce<AddressBook>((sanitizedChainAb, address) => {
+      if (!address || !utils.isAddress(address)) {
```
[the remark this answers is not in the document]
iamacook: We checksum addresses addresses on entry in both `safe-react` and `web-core`. Whilst it would be nice to check them again, I don't think we should modify the values.

I've added logs to inform users if an address was removed.

src/services/ls-migration/addressBook.ts
iamacook: Good catch! We now `trim` this.

src/services/ls-migration/addressBook.ts
iamacook: I've remove the early return. This also addresses @usame-algan's testing comment above.

iamacook:
> I tried both removing names and corrupting the address in local storage on the new app but the page renders without any errors after refresh. How to reproduce an error there? Screenshot 2022-11-08 at 11 52 43

The early return that prevented this has now been removed.

usame-algan:
> The early return that prevented this has now been removed.

I don't mean on this branch but I tried reproducing the issue on a different branch without the sanitation and still couldn't get any errors.

iamacook:
> I don't mean on this branch but I tried reproducing the issue on a different branch without the sanitation and still couldn't get any errors.

Prior to our [migration validation](https://github.com/safe-global/web-core/pull/1081), addresses were added as is. This sanitization is for address books that were add migrated before this. If you have an invalid address book and open the send tokens modal, it will crash. It might need to be [pre-default value assignment](https://github.com/safe-global/web-core/pull/1078).

This predominantly covers those which have already migrated.

katspaugh (approved):
The code is quite dense and needs to be QA'ed thoroughly.

iamacook:
After discussing this further with the team, I'm going to close this.

As our migration logic now [validates addresses](https://github.com/safe-global/web-core/pull/1081), this handles a pre-validation edge case. If users experience further issues, the preferred fix would be clearing their `localStorage` and allowing for migration to occur again or alternatively use the forthcoming Safe export/import functionality

PR #1093, opened 2022-11-08
fix: fallback to empty string when shortening address

iamacook:
## What it solves

Resolves [ongoing Sentry errors](https://sentry.io/organizations/gnosis/issues/3725083966/?project=6769196&query=is%3Aunresolved&referrer=issue-stream&sort=date&statsPeriod=14d)

## How this PR fixes it

An empty string fallback has been added to `shortenAddress` if passed a falsy value.

## How to test it

It is covered by tests.

### DIFF
--- a/src/utils/__tests__/formatters.test.ts
+++ b/src/utils/__tests__/formatters.test.ts
@@ -45,5 +45,15 @@ describe('formatters', () => {
     it('should shorten an address with custom length', () => {
       expect(formatters.shortenAddress('0x1234567890123456789012345678901234567890', 5)).toEqual('0x12345...67890')
     })
+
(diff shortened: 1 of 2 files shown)


DiogoSoaress (commented):
It should be preferred to not call `shortenAddress` if we have an undefined or null address. I would leave the util function as it was.
Do we know which scenario is causing these Sentry error?

iamacook:
> It should be preferred to not call `shortenAddress` if we have an undefined or null address. I would leave the util function as it was. Do we know which scenario is causing these Sentry error?

We believe it to be happening in the send assets form. Whilst I agree with you, this is a fallback.

katspaugh (approved):

usame-algan (approved):

PR #1098, opened 2022-11-08
Feat: banner with an AB export link

katspaugh:
## What it solves

Adds a closable banner with a link to the old app.

Screenshot 2022-11-08 at 11 49 57


src/components/common/PsaBanner/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const onClose = () => {
+    setClosed(true)
+  }
+
+  return showBanner ? (
+    <div className={styles.banner}>
+      <div className={styles.wrapper}>
+        <div className={styles.content}>{banner}</div>
+
+        <IconButton className={styles.close} onClick={onClose} aria-label="dismiss transaction queue bar">
```
iamacook, line 60:
```suggestion
        <IconButton className={styles.close} onClick={onClose} aria-label="dismiss announcement banner">
```

DiogoSoaress (approved):
Should we add the `no-redirect=true` as well?

usame-algan (approved):

francovenica:
Looks fine. 

Notes:
It steps a bit over the add button, but since this banner is supposed to be a something that shouldn't be there for a long time I think we can ignore it
image

katspaugh:
Yeah, it's not ideal, but changing the whole page layout at this moment would be disastrous.

PR #1107, opened 2022-11-08
test: spy and wait for slow network requests

DiogoSoaress:
## What it solves
Broken `create_tx` E2E test in the CI

## How this PR fixes it
Spy and wait for the `/estimations` network request
Stub the second `/estimations` request
Spy and wait for the `/propose` network request

### DIFF
--- a/cypress/e2e/smoke/create_tx.cy.js
+++ b/cypress/e2e/smoke/create_tx.cy.js
@@ -35,8 +35,14 @@ describe('Queue a transaction on 1/N', () => {
   })
 
   it('should create a queued transaction', () => {
+    // Spy the /estimations request
+    cy.intercept('POST', '/**/multisig-transactions/estimations').as('estimations')
+
     // Alias for New transaction modal
     cy.contains('h2', 'Review transaction').parents('div').as('modal')
+
+    // Wait for /estimations response
+    cy.wait('@estimations')
 
     // Estimation is loaded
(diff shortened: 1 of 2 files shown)


cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -35,9 +35,15 @@ describe('Queue a transaction on 1/N', () => {
...
 
   it('should create a queued transaction', () => {
+    // Spy the /estimations request
+    cy.intercept('POST', '/**/multisig-transactions/estimations').as('estimations')
+
     // Alias for New transaction modal
     cy.contains('h2', 'Review transaction').parents('div').as('modal')
 
+    // Wait for /estimations response
+    cy.wait('@estimations')
```
DiogoSoaress, line 45: Instead of waiting for the real request, I've also considered stubbing the first request by using `Date.now()` for the recommended nonce. This way we wouldn't have to hit the server with any `/estimations` request at all and we'd guarantee the nonce is never repeated.

It would make the test more robust but it felt like an overkill. WDYT?

katspaugh: IMHO e2e tests should not stub anything.

DiogoSoaress: I've changed it. Only spy on the request now

iamacook (approved):

PR #1114, opened 2022-11-09
fix: "Aborting fetching component for router"

### LINKED ISSUE #1112: Aborting fetching component for router
## Bug description

When opening the app, an error is thrown: "Aborting fetching component for router".

After [some reasearch](https://github.com/vercel/next.js/discussions/39040), it seems as though multipl calls to `router.push`/`router.replace` can interfere with each other. As we are calling `router.replace` in a `useEffect` in `@/pages/index.tsx`, this seems like the likely culprite. We should only be calling this once.

## Steps to reproduce
 1. Open the app at `/home` and observe the console error.

## Expected result

No error is thrown.

## Obtained result

"Aborting fetching component for router" is shown in the console.

## Screenshots

image

iamacook:
## What it solves

Resolves #1112

## How this PR fixes it

As we don't want to interact with the DOM, `useLayoutEffect` removed the error (which was caused by calling `router.replace` too many times).

## How to test it

Open `/` and observe no error in the console. Navigating to the following should work as before:

- Previously used Safe: `?safe=xyz:0x123`
- Currently selected network: `?chain=xyz` (no added Safe)
- Otherwise: `/welcome`

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from 'react'
+import { useLayoutEffect } from 'react'
 import type { NextPage } from 'next'
 import { useRouter } from 'next/router'
 import useLastSafe from '@/hooks/useLastSafe'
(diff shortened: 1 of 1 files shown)


schmanu (approved):
When tested the routing errors do not show up anymore :)

katspaugh (approved):
🧠 👍

Issue #974, opened 2022-10-25
[Governance Visibility] New Governance section


DiogoSoaress:
## What is the feature about

Allow ~resizable~ Safe apps widgets in the Dashboard.

## The list of requirements
- Iframes can be inserted in the Dashboard
- ~Widgets frame and content are resizable~
- New collapsible _Governance_ section (https://www.figma.com/file/8nKOe7h90OS7jmn3nksbq9/Governance-visibility-widgets?node-id=34%3A3627)

## Links
Safe Apps PoC: https://github.com/safe-global/safe-react/pull/4035
Epic: #969

PR #1125, opened 2022-11-10
[Governance Visibility] Dashboard section

DiogoSoaress:
## What it solves
Gives a frame to display the Governance components inside a collapsible section

Part of #974 

## How to test it
1. Go to dashboard and observe the new section

## Screenshots
Screenshot 2022-11-10 at 11 03 32

### DIFF
--- a/src/components/dashboard/GovernanceSection/GovernanceSection.tsx
+++ b/src/components/dashboard/GovernanceSection/GovernanceSection.tsx
@@ -1,37 +1,66 @@
-import { Typography, Grid, Card, Box } from '@mui/material'
+import { Typography, Grid, Card, Box, IconButton } from '@mui/material'
 import { WidgetBody, WidgetContainer } from '@/components/dashboard/styled'
+import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
+import ExpandLessIcon from '@mui/icons-material/ExpandLess'
+import { useState } from 'react'
 
-const GovernanceSection = () => (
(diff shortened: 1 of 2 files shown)


iamacook (commented):
While all looks good, I think we should leverage [MUI's `Accordion` component](https://mui.com/material-ui/react-accordion/) so we get a smooth transition between it being open and closed much like the transaction/Safe Apps lists.

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
```diff
@@ -1,36 +1,50 @@
...
 
 const GovernanceSection = () => (
   <Grid item xs={12} md>
-    <WidgetContainer>
-      <Typography component="h2" variant="subtitle1" fontWeight={700} mb={2}>
-        Governance
-      </Typography>
+    <Accordion className={css.accordion}>
+      <AccordionSummary expandIcon={<ExpandMoreIcon color="border" />}>
+        <div>
```
iamacook, line 13: I don't think this is needed.

iamacook (approved):

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
DiogoSoaress: The `<AccordionSummary />` has display: flex so it was placing the titles after each other
Screenshot 2022-11-10 at 17 24 20

PR #1127, opened 2022-11-10
fix: checksum lowercase addresses + validate checksums

### LINKED ISSUE #1122: [Load Existing Safe] Implement auto-checksum rules during "Add safe" flow
In the safe-react, the next rules were defined for the address checksum validation:

1. If there are any uppercase symbols, check the sum and if it fails, show an error: "Must be a valid address, ENS or Unstoppable domain". The same error is used for all invalid address cases.
2.  Auto-checksum all addresses if they are 100% lower-case

https://github.com/safe-global/safe-react/issues/2620

**Steps to reproduce:**
1. try to add safe=0x62da87ff2e2216f1858603a3db9313e178da3112 ( 100% lower-cases)
Current result: There is no auto-checksum and the lower-cases address is sent to the CGW.
The error is returned by the CGW - https://safe-client.staging.5afe.dev/v1/chains/1/safes/0x62da87ff2e2216f1858603a3db9313e178da3112

iamacook:
## What it solves

Resolves #1122

## How this PR fixes it

Lowercase addresses entered in the `AddressInput` component are now transformed to their checksum counterpart and their checksum validity checked.

## How to test it

Open the load Safe flow an enter a lowercase address into the address field. Observe that the address is checksummed. Altering a capital letter of the checksummed address to be lowercase should show the checksum error message.

## Screenshots

checksum

### DIFF
--- a/src/utils/__tests__/addresses.test.ts
+++ b/src/utils/__tests__/addresses.test.ts
@@ -17,6 +17,11 @@ describe('Addresses', () => {
       expect(value).toBe('0x62Da87ff2E2216F1858603A3Db9313E178da3112')
     })
 
+    it('should return uppercase addresses as is', () => {
+      const value = checksumAddress('0X62DA87FF2E2216F1858603A3DB9313E178DA3112')
+      expect(value).toBe('0X62DA87FF2E2216F1858603A3DB9313E178DA3112')
+    })
+
     it('should return non-addresses as is', () => {
       const value = checksumAddress('sdfgsdfg')
       expect(value).toBe('sdfgsdfg')
@@ -35,8 +40,13 @@ describe('Addresses', () => {
     })
 
     it('should return false for mixed case addresses', () => {
-      const value = checksumAddress('0x62Da87ff2E2216F1858603A3Db9313E178da3112')
-      expect(value).toBe('0x62Da87ff2E2216F1858603A3Db9313E178da3112')
+      const value = isChecksummedAddress('0x62Da87ff2E2216F1858603A3Db9313E178da3112')
+      expect(value).toBe(false)
+    })
+
+    it('should return false for uppercase addresses', () => {
+      const value = isChecksummedAddress('0X62DA87FF2E2216F1858603A3DB9313E178DA3112')
+      expect(value).toBe(false)
     })
 
     it('should return false for non-/invalid addresses', () => {


src/utils/__tests__/addresses.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    it('should return checksummed addresses as is', () => {
+      const value = checksumAddress('0x62Da87FF2E2216F1858603A3Db9313E178da3112')
+      expect(value).toBe('0x62Da87FF2E2216F1858603A3Db9313E178da3112')
+    })
+
+    it('should return non-addresses as is', () => {
+      const value = checksumAddress('sdfgsdfg')
+      expect(value).toBe('sdfgsdfg')
+    })
```
schmanu, line 18: I'd add another interesting testcase which is an incorrect mixed case address(with wrong checksum) should still return the same incorrect mixed case address

src/utils/__tests__/addresses.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    it('should return true for checksummed addresses', () => {
+      const value = isChecksummedAddress('0x62Da87FF2E2216F1858603A3Db9313E178da3112')
+      expect(value).toBe(true)
+    })
+
+    it('should return false for non-checksummed addresses', () => {
+      const value = isChecksummedAddress('0x62da87ff2e2216f1858603a3db9313e178da3112')
+      expect(value).toBe(false)
+    })
+
```
schmanu, line 31: Also here I could see another test case for non-checksummed mixed case address

schmanu (commented):
The AddressInput tests fail.
Otherwise I like the changes around checksummed addresses.
Also very happy with having a test for such crucial address functions

schmanu (approved):
Looking good 🚀

francovenica:
It works fine, but why it works differently with lower and upper case?

If the full address is in lower case, the input will automatically checksum it and convert the characters that are lower case to uppercase. But if you have a character that is uppercase and has to be lowercase the checksum error shows up instead of just correcting it.

Can't we make for both, to automatically checksum the address as you copy/paste it
image
image

iamacook:
> Can't we make for both, to automatically checksum the address as you copy/paste it

The issue stated that we should only checksum lowercase addresses.

We are currently checksumming addresses as follows:

```ts
export const checksumAddress = (address: string): string => {
  return isAddress(address) ? getAddress(address) : address
}
```

The `isAddress` validation function returns `false` for uppercase addresses so we return `address` as is. I'm not sure whether we should first be converting `address` to `address.toLowerCase()` if the validation returns `false` by default. What do you think? cc @safe-global/safe-web

liliya-soroka:
The same discussion we already had in the safe-react, and it was agreed that only "100% lower-case" addresses would be auto-checksummed. 
Check the original ticket for details.
I suggest sticking to the original agreement and don't add additional validation for the upper-case address.

francovenica:
Then is all good.
So the way it will work is:
If everything is lower cased it will be automatically checksummed
If a character that is supposed to be lowerCase is UpperCase an error will show up in the input

PR #1128, opened 2022-11-10
fix: Store newly created safe

### LINKED ISSUE #1121: [Sidebar] - Created safe shows with undefined address
## Bug description
When you create a safe that safe is listed with an undefined address

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Tested in Goerli and Poly, but probalby happens in every network

## Steps to reproduce
 1. Go to the app and create a safe, give it a name you can recognize
 2. Submit and wait for the safe to be created
 3. Check the sidebar

## Expected result
The safe you created is listed and selected

## Obtained result
The safe is listed with the name you set, but with an undefined address.

## Screenshots
Note: In the gif you can see that the safe creation "fails" for a moment and then succeeds. That is a bug already reported, can be ignored for this ticket
undefined address

usame-algan:
## What it solves

Resolves #1121 

## How this PR fixes it

We were using `pendingSafe.address` instead of `pendingSafe.safeAddress`. I opted against refactoring the types to keep this PR small but added a TODO.

## How to test it

1. Create a new safe
2. Observe that it is being added to the added safes list

### DIFF
--- a/src/components/create-safe/status/useSafeCreationEffects.ts
+++ b/src/components/create-safe/status/useSafeCreationEffects.ts
@@ -71,9 +71,15 @@ const useSafeCreationEffects = ({
       trackEvent(CREATE_SAFE_EVENTS.CREATED_SAFE)
 
       // Add the Safe and add names to the address book
-      if (pendingSafe) {
+      if (pendingSafe && pendingSafe.safeAddress) {
         dispatch(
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
👍

iamacook (approved):

PR #1132, opened 2022-11-11
feat: replacement modal

### LINKED ISSUE #718: (feature) Add Stepper on Reject transaction modal
Reject popup: Stepper was proposed after user tests to explain where the user is in the process and when to pay gas.    

Screenshot 2022-09-28 at 13 13 14

### Links
- [Figma mockup](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=1076%3A40781)

iamacook:
## What it solves

Resolves #718 

## How this PR fixes it

The new create/replace transaction modals have been separated, with the latter including a stepper according to [this design](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=2104%3A59002).

## How to test it

- Queue a transaction, then click replace. Observe the new modal, with the buttons working as intended.
- The create modal should function as before, displaying the Transaction Builder.

## Screenshots

image

image

image

replacement stepper

### DIFF
--- a/src/components/tx/modals/NewTxModal/ReplacementModal.tsx
+++ b/src/components/tx/modals/NewTxModal/ReplacementModal.tsx
@@ -1,0 +1,124 @@
+import {
+  Box,
+  Button,
+  DialogContent,
+  SvgIcon,
+  Tooltip,
+  Typography,
+  Stepper,
+  Step,
+  StepLabel,
+  DialogActions,
+  Grid,
+} from '@mui/material'
+
+import ModalDialog from '@/components/common/ModalDialog'
+import InfoIcon from '@/public/images/notifications/info.svg'
+import RocketIcon from '@/public/images/transactions/rocket.svg'
+import CheckIcon from '@mui/icons-material/Check'
+import DeleteIcon from '@/public/images/common/delete.svg'
+import { SendTokensButton, SendNFTsButton } from './TxButton'
+
+import css from './styles.module.css'
+
+const wrapIcon = (icon: React.ReactNode) => <div className={css.circle}>{icon}</div>
+
+const steps = [
+  {
+    label: 'Create new transaction with same nonce',
+    icon: <div className={css.redCircle} />,
+  },
+  {
+    label: 'Collect confirmations from owners',
+    icon: wrapIcon(<CheckIcon fontSize="small" color="border" />),
+  },
+  {
+    label: 'Execute replacement transaction',
+    icon: wrapIcon(<SvgIcon component={RocketIcon} inheritViewBox fontSize="small" color="border" />),
+  },
+  {
+    label: 'The initial transaction is replaced',
+    icon: wrapIcon(<SvgIcon component={DeleteIcon} inheritViewBox fontSize="small" color="border" />),
+  },
+]
+
+const ReplacementModal = ({
+  txNonce,
+  onClose,
+  onTokenModalOpen,
+  onNFTModalOpen,
+  onRejectModalOpen,
+}: {
+  txNonce: number
+  onClose: () => void
+  onTokenModalOpen: () => void
+  onNFTModalOpen: () => void
+  onRejectModalOpen: () => void
+}) => {
+  return (
+    <ModalDialog open dialogTitle={`Replace transaction with nonce ${txNonce}`} onClose={onClose}>
+      <DialogContent>
+        <Box className={css.container}>
+          <Typography variant="h5" gutterBottom>
+            How to replace/change a transaction?
+          </Typography>
+          <Typography variant="body1" textAlign="center">
+            A signed transaction cannot be removed but it can be replaced with a new transaction with the same nonce.
+          </Typography>
+          <Stepper alternativeLabel className={css.stepper}>
+            {steps.map(({ label }) => (
+              <Step key={label}>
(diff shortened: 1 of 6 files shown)


src/components/tx/modals/NewTxModal/ReplacementModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+            <SendNFTsButton onClick={onNFTModalOpen} />
+          </Grid>
+          <Grid xs={1}>
+            <Typography variant="body2" textAlign="center" fontWeight={700}>
+              or
+            </Typography>
+          </Grid>
+          <Grid xs>
+            <Button onClick={onRejectModalOpen} variant="outlined" fullWidth sx={{ mt: 2 }}>
+              Replace transaction
```
katspaugh, line 99: IMHO this wording here makes the whole modal confusing. If this is this button replaces the tx, what do the two others do?

I would write "Empty transaction", "Stub transaction", "Rejection transaction" or something similar.

katspaugh (commented):
Please double-check margins/paddings. Almost all of the elements are farther apart from each other on Figma.

iamacook:
> Please double-check margins/paddings. Almost all of the elements are farther apart from each other on Figma.

I spoke with @liliiaorlenko and we tweaked the designs a bit. I've updated the PR accordingly:

image

(The modal width was a bit off in the design after we tweaked it for long `shortName`s.)

src/components/tx/modals/NewTxModal/ReplacementModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+      </DialogContent>
+      <DialogActions className={css.container}>
+        <Grid container alignItems="center" justifyContent="center" spacing={3}>
+          <Grid item xs={12}>
+            <Typography variant="body2" textAlign="center" fontWeight={700}>
+              Select the type of replacement transaction
+            </Typography>
+          </Grid>
+          <Grid item xs>
+            <SendTokensButton onClick={onTokenModalOpen} sx={{ mb: 1, maxWidth: '320px !important' }} />
```
usame-algan, line 86: This doesn't seem to do anything even when I remove it. Why do we need to limit the width of this button and not the others?

src/components/tx/modals/NewTxModal/styles.module.css
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.stepper {
+  padding-top: var(--space-3);
+  gap: var(--space-2);
+}
+
+.stepper :global .MuiStepConnector-root {
+  left: calc(-50% + 14px);
+  right: calc(50% + 28px);
```
usame-algan, line 33: nit: Its also possible to do this via transform which is a bit clearer regarding the magic numbers.

```
left: -8px;
transform: translateX(-50%);
width: calc(100% - 32px);
```

By changing the second parameter in width its possible to adjust the size of the connector without having to change any other properties. This solution only works if all columns have the same size though.

usame-algan (commented):
Looking good already!

Do we have a design for mobile? I did a quick sketch on how it could look like, reducing the size of the stepper and collapsing the buttons. cc @liliiaorlenko 
Screenshot 2022-11-14 at 11 29 35

src/components/tx/modals/NewTxModal/ReplacementModal.tsx
iamacook: Good catch! It's a leftover from testing that I forgot to remove.

iamacook:
> Do we have a design for mobile?

I really like your suggestion have tweaked the layout to be as follows:

image

Some of the alignment in the design is not quite pixel perfect, e.g. the "How does it work" text so I've followed my instict with the overall alignment:

image

Here's the "standard" view:

image

cc @liliiaorlenko

liliiaorlenko:
> > Do we have a design for mobile?
> 
> I really like your suggestion have tweaked the layout to be as follows:
> 
> image
> 
> Some of the alignment in the design is not quite pixel perfect, e.g. the "How does it work" text so I've followed my instict with the overall alignment:
> 
> image
> 
> Here's the "standard" view:
> 
> image
> 
> cc @liliiaorlenko

I like this version, everything looks good on mobile!

src/components/tx/modals/NewTxModal/ReplacementModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+          <Grid item container justifyContent="center" gap={1} xs={12} sm>
+            <SendTokensButton onClick={onTokenModalOpen} sx={btnWidth} />
+            <SendNFTsButton onClick={onNFTModalOpen} sx={btnWidth} />
+          </Grid>
+          <Grid item>
+            <Typography variant="body2" className={css.or}>
+              or
+            </Typography>
+          </Grid>
+          <Grid item container xs={12} sm sx={{ justifyContent: { xs: 'center', sm: 'flex-start' } }}>
```
usame-algan, line 109: On some widths below 600px the Button and text are not collapsing. We could explicitly set a column based layout in viewports < 600px to fix this.
Screenshot 2022-11-15 at 14 58 23

iamacook: This should now work.

usame-algan (approved):
🚀

katspaugh (approved):

Issue #864, opened 2022-06-09
[Epic] Safe Creation Redesign


usame-algan:
## Overview

Improving the drop-off rate during safe creation by implementing the [new design](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=51%3A5661)

## Tasks

- [x] Build a stepper component to manage the form screens #931
- [x] Build a component for the info widget #927 
- [x] Build a component for the overview widget #928
- [x] Build a component for the status screens #932
- [x] Implement redesign for the status screens #994 
- [x] Implement redesign Network + Name step #961 
- [x] Implement redesign Owner step #962
- [x] Implement redesign Review step #963
- [x] Implement redesign Connect Wallet step #1003 
- [x] Layout #995
- [x] Implement Overview/Info Widget display logic #996 
- [x] Handle wallet connection #1080 
- [ ] A/B testing: #968

## How to test

- Open the new route `/new-safe/create`
- Click through the steps and compare to the Figma designs
- Check that the requirements of each subtask is met

## Links

Problem statement: https://github.com/safe-global/safe-pm/issues/82
Notion: https://www.notion.so/gnosis-safe/Safe-Creation-7ec069306e8c4101bffd7c17299eec53
Figma: https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=51%3A5661
Analytics: https://analytics.google.com/analytics/web/#/analysis/p308247657/edit/h41dtVeqQ0icP3o_O_Cx-g

francovenica:
Question.
The previous iteration of safe creation would put as the first owner the currently connected wallet, and if you changed wallet it would replace that first row with it. Was that feature dropped?

francovenica:
Question/suggestion : In the review step (step 4) the addresses don't have the copy to clipboard/blockexplorer icons. They are not in the design, but we put it everywhere in the app, we should probably put them here too

francovenica:
Question:
In the designs it seems to be a way to complete cancel the safe creation, but in any of the step, not even step one, there is a "Cancel". Was decided to not put a way to exit the safe creation?
Currently, the only way for the user to get out is using the logo on the top left (which is not even that good, since that if the user has a "last used safe" it will take him there first)

usame-algan:
> The previous iteration of safe creation would put as the first owner the currently connected wallet, and if you changed wallet it would replace that first row with it. Was that feature dropped?

[...]

PR #1135, opened 2022-11-14
fix: Show error notifications in safe creation

usame-algan:
## What it solves

Part of #864 

## How this PR fixes it

- Shows a notification with more detailed errors
- Closes the notification when the user presses Retry

## Other changes

- Extends the notification slice to support closing all notifications
- Temporarily links to new safe creation on welcome page to make testing easier

## How to test it

1. Open the Safe
2. Create a new safe
3. Reject the tx
4. Observe a notification being shown
5. Cancel a tx
6. Observe a notification being shown
7. Press Retry
8. Observe the notification disappearing

## Screenshots

Screenshot 2022-11-14 at 11 04 21

### DIFF
--- a/src/components/new-safe/steps/Step4/useSafeCreation.ts
+++ b/src/components/new-safe/steps/Step4/useSafeCreation.ts
@@ -12,10 +12,11 @@ import {
   getSafeCreationTxInfo,
   getSafeDeployProps,
   handleSafeCreationError,
+  SAFE_CREATION_ERROR_KEY,
   showSafeCreationError,
 } from '@/components/new-safe/steps/Step4/logic'
 import { useAppDispatch } from '@/store'
-import { closeAllNotifications } from '@/store/notificationsSlice'
+import { closeByGroupKey } from '@/store/notificationsSlice'
 
 export enum SafeCreationStatus {
   AWAITING,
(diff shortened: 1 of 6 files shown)


src/components/new-safe/steps/Step4/useSafeCreation.ts
```diff
@@ -50,6 +54,7 @@ export const useSafeCreation = (
     if (!pendingSafe || !provider || !chain || !wallet || isCreating) return
 
     setIsCreating(true)
+    dispatch(closeAllNotifications())
```
iamacook, line 57: These notifications share the same `groupKey` and will close each other if they "overflow". I don't think we should programmatically close them like this.

usame-algan: Unfortunately, we only show notifications for errors. This adds a bit of confusion imo when retrying the tx and the error notification is still being displayed. In the past we looked into duplicating the `TxEvents` flows for safe creation i.e. showing notifications for signing etc. but decided against it as we wanted to use inline messages. Might be worth reconsidering but I don't want to blow up this PR.

iamacook: That makes sense. I would adjust `closeAllNotifications` to `closeByGroupKey` to prevent the closing of unrelated notifcations though.

PR #1136, opened 2022-11-14
fix(safe-apps): Safe apps analysis improvements

yagopv:
## What it solves

Add some fixes:

- Duplicated "Open Safe App" event tracking
- 404 when retrieving the apps default list and no chainId in redux store
- Remove querystrings when storing url's from custom apps in analytics (We detect eth addresses there)


DaniSomoza (approved):

src/services/analytics/gtm.ts
```diff
@@ -141,13 +141,15 @@ export const gtmTrackPageview = (pagePath: string): void => {
   gtmSend(gtmEvent)
 }
 
+const stripQueryStringAndHashFromPath = (url: string) => url.split('?')[0].split('#')[0]
```
katspaugh, line 144:
```suggestion
const normalizeAppName = (appName?: string): string => {
  // App name is a URL
  if (appName?.startsWith('http')) {
    // Strip search query and hash
    return appName.split('?')[0].split('#')[0]
  }
  return appName || ''
}
```

Worth adding some tests?

yagopv: done

katspaugh (approved):
👍

liliya-soroka:
Verified

PR #1137, opened 2022-11-14
feat: redesign address book import

### LINKED ISSUE #1118: Redesign import dialog for address book
## What is the feature about
For the global `Import all data` feature we redesigned the file upload / import dialog.
We should also apply this redesign to our 'Import addressbook' dialog.

## The list of requirements
- [x] Redesign dropzone
- [x] Redesign summary of import (outside / instead of dropzone)

## Designs/sketches
https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=2053%3A57776

schmanu:
## What it solves
Our import dialogs have inconsistent designs

Resolves #1118 

## How this PR fixes it
- Extracts `FileInput` from `ImportAllDialog` 
- uses this component in `ImportDialog` of addressbook

## How to test it
Open the import dialog in addressbook
It should work the same but have the same design as our global Import All dialog

## Analytics changes
None

## Screenshots
Screenshot 2022-11-14 at 13 54 26
Screenshot 2022-11-14 at 13 56 44


src/components/common/FileUpload/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+const UploadSummary = ({ fileInfo, onRemove }: { fileInfo: FileInfo; onRemove: (() => void) | MouseEventHandler }) => {
+  return (
+    <Grid container direction="column" gap={1} mt={3}>
+      <Grid container gap={1} display="flex" alignItems="center">
+        <Grid item xs={1}>
+          <ColoredFileIcon color="primary" />
+        </Grid>
+        <Grid item xs={7}>
+          {fileInfo.name} {fileInfo.additionalInfo && `- ${fileInfo.additionalInfo}`}
```
iamacook, line 26:
```suggestion
          {fileInfo.name}{fileInfo.additionalInfo && ` - ${fileInfo.additionalInfo}`}
```

src/components/common/FileUpload/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  isDragActive = false,
+  fileType,
+  fileInfo,
+  onRemove,
+}: {
+  isDragReject?: boolean
+  isDragActive?: boolean
+  fileType: string
+  getInputProps?: () => any
+  getRootProps: () => any
```
iamacook, line 79: I'd suggest we use the types from `react-dropzone` here.

src/components/address-book/ImportDialog/index.tsx
```diff
@@ -114,52 +114,41 @@ const ImportDialog = ({ handleClose }: { handleClose: () => void }): ReactElemen
...
-                      </Typography>
-                    )}
-                  </div>
-                ) : (
-                  'Drop your CSV file here or click to upload.'
-                )}
-              </Box>
+              <FileUpload
+                fileInfo={fileInfo}
+                fileType="CSV"
```
iamacook, line 141: Let's extract this to a constant.

src/components/settings/ImportAllDialog/index.tsx
```diff
@@ -113,102 +105,43 @@ const ImportAllDialog = ({ handleClose }: { handleClose: () => void }): ReactEle
...
-                  <Link href="#" color="secondary">
-                    choose a file
-                  </Link>
-                </Typography>
-              </Box>
-            </Box>
-          )}
-        </div>
+        <FileUpload
+          fileType="JSON"
```
iamacook, line 137: Let's extract this to a constant.

src/components/address-book/ImportDialog/index.tsx
```diff
@@ -114,52 +114,41 @@ const ImportDialog = ({ handleClose }: { handleClose: () => void }): ReactElemen
...
-            const { onClick, ...removeProps } = getRemoveFileProps()
+            const { onClick } = getRemoveFileProps()
 
             const onRemove = (e: MouseEvent<HTMLSpanElement>) => {
               setCsvData(undefined)
               setError(undefined)
               onClick(e)
             }
 
+            const fileInfo = acceptedFile
```
iamacook, line 125: Can we create a type for this and export it from `FileUpload`? It can be used here and in the `ImportAllDialog`.

iamacook (approved):

francovenica:
I think we should remove one of the warning messages when a row is invalid (a row with no address or invalid chainID) because is redundant

Also the "Learn about the address book import and export" is a link but is colored black with no icon so is not obvious. I think there is no green color for links anymore, but We could add the icon that was there in the previous iteration

image

image

francovenica:
We have decided that the link issue should be addressed more general, so the solution is applied to all links in the site and not just this one

PR #1149, opened 2022-11-14
feat: Dashboard governance widgets

### LINKED ISSUE #974: [Governance Visibility] New Governance section

DiogoSoaress:
## What it solves
Mini Safe apps iframes are displayed in the dashboard

Resolves #974

## Screenshots
Screenshot 2022-11-23 at 16 43 47
Screenshot 2022-11-22 at 18 50 58
Screenshot 2022-11-23 at 16 43 11
Screenshot 2022-11-22 at 19 17 18

### DIFF
--- a/src/components/dashboard/GovernanceSection/GovernanceSection.tsx
+++ b/src/components/dashboard/GovernanceSection/GovernanceSection.tsx
@@ -1,4 +1,4 @@
-import { Typography, Grid, Card } from '@mui/material'
+import { Typography, Grid, Card, Box, Paper } from '@mui/material'
 import { WidgetBody } from '@/components/dashboard/styled'
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
 import Accordion from '@mui/material/Accordion'
@@ -9,27 +9,29 @@ import SafeAppsErrorBoundary from '@/components/safe-apps/SafeAppsErrorBoundary'
 import SafeAppsLoadError from '@/components/safe-apps/SafeAppsErrorBoundary/SafeAppsLoadError'
 import AppFrame from '@/components/safe-apps/AppFrame'
 import { useBrowserPermissions } from '@/hooks/safe-apps/permissions'
-
-export type WidgetType = {
-  appId: string
-  appUrl: string
-  widgetType?: 'iframe'
-}
-
-// TODO: use remote safe apps endpoint and the 'dashboard-mini-app' tag
-export const appWidgets: WidgetType[] = [
-  {
-    appId: '1',
-    appUrl: 'http://localhost:3001/safe-claiming-app#snapshot_widget',
-  },
-  {
-    appId: '2',
-    appUrl: 'http://localhost:3001/safe-claiming-app#widget',
-  },
-]
+import { useRemoteSafeApps } from '@/hooks/safe-apps/useRemoteSafeApps'
+import { SafeAppsTag } from '@/config/constants'
 
 const GovernanceSection = () => {
   const { getAllowedFeaturesList } = useBrowserPermissions()
+  const [claimingSafeApp, , loadingclaimingSafeApp] = useRemoteSafeApps(SafeAppsTag.SAFE_CLAIMING_APP)
+  const claimingApp = claimingSafeApp?.[0]
+
+  if (loadingclaimingSafeApp) {
+    return (
+      <Paper>
+        <Box sx={{ height: '300px' }} display="flex" alignItems="center" justifyContent="center">
+          <Typography variant="h1" color="text.secondary">
+            Loading Safe Claiming App...
+          </Typography>
+        </Box>
+      </Paper>
+    )
+  }
+
+  if (!claimingApp) {
+    return null
+  }
 
   return (
     <Grid item xs={12} md>
@@ -52,8 +54,8 @@ const GovernanceSection = () => {
                 <Card sx={{ height: '300px' }}>
                   <SafeAppsErrorBoundary render={() => <SafeAppsLoadError onBackToApps={() => {}} />}>
                     <AppFrame
-                      appUrl={appWidgets[0].appUrl}
-                      allowedFeaturesList={getAllowedFeaturesList(appWidgets[0].appUrl)}
+                      appUrl={`${claimingApp.url}#snapshot-widget`}
+                      allowedFeaturesList={getAllowedFeaturesList(claimingApp.url)}
                       isQueueBarDisabled
                     />
                   </SafeAppsErrorBoundary>
@@ -63,8 +65,8 @@ const GovernanceSection = () => {
                 <Card sx={{ height: '300px' }}>
                   <SafeAppsErrorBoundary render={() => <SafeAppsLoadError onBackToApps={() => {}} />}>
                     <AppFrame
-                      appUrl={appWidgets[1].appUrl}
-                      allowedFeaturesList={getAllowedFeaturesList(appWidgets[1].appUrl)}
+                      appUrl={`${claimingApp.url}#claiming-widget`}
+                      allowedFeaturesList={getAllowedFeaturesList(claimingApp.url)}
                       isQueueBarDisabled
                     />
                   </SafeAppsErrorBoundary>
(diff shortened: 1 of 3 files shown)


src/components/dashboard/GovernanceSection/GovernanceSection.tsx
```diff
@@ -1,51 +1,83 @@
...
-            <Grid item xs={4}>
-              <Card>
-                <Box m={2} sx={{ minHeight: '200px' }}>
-                  <Typography variant="h3">
-                    <strong>Claiming app</strong>
-                  </Typography>
-                </Box>
-              </Card>
+  if (!claimingApp) {
+    return null
```
iamacook, line 33: We should give some feedback instead of showing nothing.

iamacook (commented):
Looking good. I've noticed a few issues though:

- On initial load "Explore Safe App" flashes, then the loading message and finally the section closes.
  - "Explore Safe Apps" shouldn't show.
  - Neither it or the loading message should flash.
  - I think the `Accordion` should then be open by default and we save the expansion preference locally.

flash
---
- The Claiming App widget does not load.
- The truncation of proposal names doesn't work in the `iframe`.

widgets

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
DiogoSoaress: Can we write "There was an error fetching the widgets" ?
Note: this is for an error on the following request https://safe-client.staging.5afe.dev/v1/chains/1/safe-apps?client_url=http://localhost:3000

DiogoSoaress:
> I think the Accordion should then be open by default and we save the expansion preference locally.

This default resulted from https://www.notion.so/gnosis-safe/Kickoff-SafeDAO-27-10-2022-2234309ec4ba47ce87ccf662c867aa8c 

Can you confirm @liliiaorlenko ? Should the Governance section start collapsed or expanded?

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
iamacook: "There was an error fetching the widget" works.

iamacook:
> This default resulted from https://www.notion.so/gnosis-safe/Kickoff-SafeDAO-27-10-2022-2234309ec4ba47ce87ccf662c867aa8c

If that's the intended design then let's stick to it but the loading text shouldn't show unless it's open.

liliiaorlenko:
> > I think the Accordion should then be open by default and we save the expansion preference locally.
> 
> This default resulted from https://www.notion.so/gnosis-safe/Kickoff-SafeDAO-27-10-2022-2234309ec4ba47ce87ccf662c867aa8c
> 
> Can you confirm @liliiaorlenko ? Should the Governance section start collapsed or expanded?

It should start open. Let me know when I can recommend suggestions for implemented visual improvements :)

DiogoSoaress:
@liliiaorlenko now it is a good time for design remarks 🎨  you can access it here https://dashboard-mini-safe-apps.web-core.pages.dev/

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
```diff
@@ -1,14 +1,25 @@
...
+import { SafeAppsTag } from '@/config/constants'
 
-const GovernanceSection = () => (
-  <Grid item xs={12} md>
-    <Accordion className={css.accordion}>
+const GovernanceSection = () => {
+  const { getAllowedFeaturesList } = useBrowserPermissions()
+  const [claimingSafeApp, errorFetchingClaimingSafeApp] = useRemoteSafeApps(SafeAppsTag.SAFE_CLAIMING_APP)
+  const claimingApp = claimingSafeApp?.[0]
+  const fetchingSafeClaimingApp = !claimingApp && !errorFetchingClaimingSafeApp
```
DiogoSoaress, line 19: I've implemented this loading logic because the loading state returned by `useRemoteSafeApps` starts false.

iamacook: Can we not move this logic inside the hook?

iamacook: Can we not move this logic into the hook? This will forever load if an error occurs.

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
```diff
@@ -21,31 +32,49 @@ const GovernanceSection = () => (
...
-                  <Typography variant="h3">
-                    <strong>Claiming app</strong>
-                  </Typography>
-                </Box>
-              </Card>
+        {fetchingSafeClaimingApp ? (
+          <Paper>
+            <Box sx={{ height: '300px' }} display="flex" alignItems="center" justifyContent="center">
+              <Typography variant="h1" color="text.secondary">
+                Loading Safe Claiming App...
```
iamacook, line 39: I know that they both hypothetically the Claiming App but I think it would look visually better if we split it into the same layout as the widgets, e.g. [ Loading Snapshot ] [ Loading Claiming ]. What do you think? cc @liliiaorlenko

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
```diff
@@ -21,31 +32,49 @@ const GovernanceSection = () => (
...
+            </Box>
+          </Paper>
+        ) : claimingApp ? (
+          <WidgetBody>
+            <Grid spacing={3} container>
+              <Grid item xs={12} md={6} lg={8}>
+                <Card sx={{ height: '300px' }}>
+                  <SafeAppsErrorBoundary render={() => <SafeAppsLoadError onBackToApps={() => {}} />}>
+                    <AppFrame
+                      appUrl={`${claimingApp.url}#snapshot-widget`}
```
iamacook, line 50: I suggest we move both hashes to constants.

iamacook (commented):
The truncation looks good now. However, the loading screen still flashes before closing the `Accordion`. As [suggested](https://github.com/safe-global/web-core/pull/1149#issuecomment-1323834653), it should load open. I would suggest we then cache the open preference locally.

widgets3

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
DiogoSoaress: In this case we are just fetching the remote Safe apps endpoint. Each of the widgets loads with its own Skeleton component. So this "loading" component was more for the whole Governance section

DiogoSoaress:
> As https://github.com/safe-global/web-core/pull/1149#issuecomment-1323834653, it should load open.

I added the `defaultExpanded` property to the Accordion but for some reason it works in localhost but not work in the deployment URL. Investigating...

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
iamacook: I would adjust it to reference "governance" instead of the Claiming App specifically then.

DiogoSoaress:
> Can we not move this logic into the hook? This will forever load if an error occurs.

After revisiting it, the hook is returning correctly IMO. This is more of a component state logic so I would leave it in the component. Maybe I can rename the variable to `fetchingRemoteApps` ?

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
iamacook: That's quite generic imo. If you wanted to specify apps, I would say splitting the loading section in two as my suggestion further up would make sense.

iamacook:
> After revisiting it, the hook is returning correctly IMO. This is more of a component state logic so I would leave it in the component. Maybe I can rename the variable to `fetchingRemoteApps` ?

It's not regarding the naming, it's that both the error and loading flags are referenced. If an error occurs, the UI will show them as persistently loading. We shouldn't combine them.

DiogoSoaress:
> If an error occurs, the UI will show them as persistently loading.

If an error occurs, it throws inside of the useAsync and the error is set. The UI will display
Screenshot 2022-11-22 at 18 50 14

iamacook (approved):

PR #1151, opened 2022-11-15
feat: export/import 2.0 versioned data

### LINKED ISSUE #1147: Implement version 2.0 data export/import
## What is the feature about

As `safe-react` can export data, we should also implement it in `web-core`.

## The list of requirements

- Add export functionality with `version: 2.0` JSON. We can leverage `getPersistedState` to export *only* the slices we persist.
- Add `/export` route
- Add `version: 2.0` parsing for address book and added Safes. We should extend the parsing to include all persisted slices.

iamacook:
## What it solves

Resolves #1147

## How this PR fixes it

Safe apps, settings and session data has been added to the export data, as well as import for them. The imported address book is filtered in the same manner that `version: 1.0` is/was.

Tests have been added for `version 2.0`, covering the address book, added Safes, Safe apps, settings and sessions.

Note: the cookies, pending transactions or local flags are not included in the export/import.

## How to test it

Export all data from the Safe and observe the version is set to `2.0` and that there are keys for the address book, added Safes, Safe apps, settings and sessions. Importing the "backup" should set all of these as they were. No sanitisation is done, only the prevention of adding invalid address book entries (empty names, or empty/non-checksummed addresses).

## Screenshots

image

image

### DIFF
--- a/src/components/settings/DataManagement/index.tsx
+++ b/src/components/settings/DataManagement/index.tsx
@@ -5,43 +5,81 @@ import { SETTINGS_EVENTS } from '@/services/analytics'
 import { Paper, Grid, Typography, Tooltip, SvgIcon, Button } from '@mui/material'
 import { useState } from 'react'
 import ImportAllDialog from '../ImportAllDialog'
+import { getPersistedState } from '@/store'
+import { SAFE_EXPORT_VERSION } from '../ImportAllDialog/useGlobalImportFileParser'
+import { pendingTxsSlice } from '@/store/pendingTxsSlice'
+import { cookiesSlice } from '@/store/cookiesSlice'
+
+const handleExport = () => {
+  const filename = `safe-data-${new Date().toISOString().slice(0, 10)}.json`
+
+  // Exclude pending transactions and cookie consent from export
+  const { [pendingTxsSlice.name]: _p, [cookiesSlice.name]: _c, ...exportData } = getPersistedState()
+
+  const data = JSON.stringify({ version: SAFE_EXPORT_VERSION.V2, data: exportData })
+
+  const blob = new Blob([data], { type: 'text/json' })
+  const link = document.createElement('a')
+
+  link.download = filename
+  link.href = window.URL.createObjectURL(blob)
+  link.dataset.downloadurl = ['text/json', link.download, link.href].join(':')
+  link.dispatchEvent(new MouseEvent('click'))
+}
+
+const InfoTooltip = ({ title }: { title: string }) => {
+  return (
+    <Tooltip placement="top" title={title}>
+      <span>
+        <SvgIcon
+          component={InfoIcon}
+          inheritViewBox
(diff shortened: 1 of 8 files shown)


iamacook:
I've set this as a draft as the mobile team need to be included regarding this.

katspaugh:
Suggestion to inline the text from the tooltip next to the title. Please also separate the import/export blocks visually too–add more margins between them.

Screenshot 2022-11-15 at 14 17 39

src/components/settings/DataManagement/index.tsx
```diff
@@ -1,47 +1,75 @@
...
+import { getPersistedState } from '@/store'
+import { SAFE_EXPORT_VERSION } from '../ImportAllDialog/useGlobalImportFileParser'
+import { pendingTxsSlice } from '@/store/pendingTxsSlice'
+import { cookiesSlice } from '@/store/cookiesSlice'
+
+const handleExport = () => {
+  const filename = `safe-data-${new Date().toISOString().slice(0, 10)}.json`
+
+  // Exclude pending transactions and cookie consent from export
+  const { [pendingTxsSlice.name]: _p, [cookiesSlice.name]: _c, ...exportData } = getPersistedState()
```
schmanu, line 15: As we now know what exact data we want to import, should we change the export to also just include that needed data? 
 *  address book
 *  added Safes
 *  safeApps
 *  settings
 *  session

src/components/settings/DataManagement/index.tsx
```diff
@@ -1,47 +1,75 @@
...
-import { Paper, Grid, Typography, Tooltip, SvgIcon, Button } from '@mui/material'
+import { Paper, Grid, Typography, Button } from '@mui/material'
 import { useState } from 'react'
 import ImportAllDialog from '../ImportAllDialog'
+import { getPersistedState } from '@/store'
+import { SAFE_EXPORT_VERSION } from '../ImportAllDialog/useGlobalImportFileParser'
+import { pendingTxsSlice } from '@/store/pendingTxsSlice'
+import { cookiesSlice } from '@/store/cookiesSlice'
+
+const handleExport = () => {
```
schmanu, line 11: In our address book export we have a small dialog which gives a summary of what will be exported and an export button in the dialog.

For consistency we should do it the same way here.

schmanu (commented):
I think in the summary of Import file contents we should include info about settings / safeApps:
i.e: "The file contains **settings**"

I tested it and realized that the import of settings does not work.
The exported settings are not applied.
I guess `merge` causes the old state to preserve somehow.

My testcase:
In chrome I set `darkMode` to `true` and removed the address prefixes.

Then I imported in Firefox and it was still in light mode with address prefixes.

iamacook:
Thank you for the feedback @schmanu! As we've since changed a lot of things since creating this PR, I will close this for the time being and we can use it as a reference for when we eventually tackle this issue.

PR #1160, opened 2022-11-15
fix: handle invalid shortname in url

### LINKED ISSUE #1154: Invalid chain short name in the URL error is not caught
## Bug description

Opening the app with a safe address that contains an invalid shortname i.e. undefined:0x... will result in an error that is not caught.

Sentry: https://sentry.io/organizations/gnosis/issues/3740249359/?project=6769196&query=is%3Aunresolved&referrer=issue-stream

## Steps to reproduce
 1. Go to the Safe app
 2. Open a specific safe
 3. Change the shortname in the url to something that is not a chain
 4. Observe the app crashing and spamming errors in the console

## Expected behaviour

There are multiple ways we could solve this:

- Navigate user to `/welcome`. This is the same behaviour as on safe-react
- Show the 404 page
- Show a "Safe couldn't be loaded" error. (This would still do backend requests which is not ideal imo)
- Show a new error e.g. "Safe couldn't be loaded. Chain xyz does not exist". (This would still do backend requests which is not ideal imo)

Lets go with solution 1

usame-algan:
## What it solves

Resolves #1154 

## How this PR fixes it

- Instead of throwing an error, we navigate the user to the `/welcome` page if there is an invalid shortname in the URL. This is the same behaviour as in safe-react.
- Error is being logged to the console and sentry instead

## How to test it

1. Open a safe
2. Change the URL shortname to an invalid shortname e.g. "test" or "undefined"
3. Observe being navigated to the welcome page
4. Change the URL shortname to a valid shortname for which the safe doesn't exist
5. Observe a Safe couldn't be loaded error

### DIFF
--- a/src/hooks/useChainId.ts
+++ b/src/hooks/useChainId.ts
@@ -6,6 +6,7 @@ import { useAppSelector } from '@/store'
 import { selectSession } from '@/store/sessionSlice'
 import { parsePrefixedAddress } from '@/utils/addresses'
 import { prefixedAddressRe } from '@/utils/url'
+import { AppRoutes } from '@/config/routes'
 
 const defaultChainId = IS_PRODUCTION ? chains.eth : chains.gor
 
@@ -41,7 +42,7 @@ export const useUrlChainId = (): string | undefined => {
   if (shortName) {
     const chainId = Object.entries(chains).find(([key]) => key === shortName)?.[1]
     if (chainId == null) {
-      throw Error('Invalid chain short name in the URL')
+      router.push(AppRoutes.welcome)
     }
     return chainId
   }
--- a/src/hooks/useChainId.ts
+++ b/src/hooks/useChainId.ts
@@ -6,7 +6,6 @@ import { useAppSelector } from '@/store'
 import { selectSession } from '@/store/sessionSlice'
 import { parsePrefixedAddress } from '@/utils/addresses'
 import { prefixedAddressRe } from '@/utils/url'
-import { AppRoutes } from '@/config/routes'
 
 const defaultChainId = IS_PRODUCTION ? chains.eth : chains.gor
 
@@ -39,13 +38,7 @@ export const useUrlChainId = (): string | undefined => {
   const { prefix } = parsePrefixedAddress(safe)
   const shortName = prefix || chain
 
-  if (shortName) {
-    const chainId = Object.entries(chains).find(([key]) => key === shortName)?.[1]
-    if (chainId == null) {
-      router.push(AppRoutes.welcome)
-    }
-    return chainId
-  }
+  return Object.entries(chains).find(([key]) => key === shortName)?.[1]
 }
 
 export const useChainId = (): string => {
--- a/src/hooks/usePathRewrite.ts
+++ b/src/hooks/usePathRewrite.ts
@@ -26,7 +26,7 @@ const usePathRewrite = () => {
 
     const chainId = Object.entries(chains).find(([key]) => key === prefix)?.[1]
     if (!chainId) {
-      logError(ErrorCodes._621, prefix)
+      logError(ErrorCodes._104, prefix)
       router.push(AppRoutes.welcome)
       return
     }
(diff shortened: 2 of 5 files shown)


src/hooks/useChainId.ts
```diff
@@ -41,7 +42,7 @@ export const useUrlChainId = (): string | undefined => {
   if (shortName) {
     const chainId = Object.entries(chains).find(([key]) => key === shortName)?.[1]
     if (chainId == null) {
-      throw Error('Invalid chain short name in the URL')
+      router.push(AppRoutes.welcome)
```
katspaugh, line 45: This hook is called from 150+ components. It's supposed to be a getter-only hook.

I suggest we handle bogus short names outside of this hook. E.g. somewhere in _app.tsx. This hook could be wrapped in a try-catch there.

katspaugh: OK.

src/hooks/useChainId.ts
```diff
@@ -38,13 +38,7 @@ export const useUrlChainId = (): string | undefined => {
...
   const shortName = prefix || chain
 
-  if (shortName) {
-    const chainId = Object.entries(chains).find(([key]) => key === shortName)?.[1]
-    if (chainId == null) {
-      throw Error('Invalid chain short name in the URL')
-    }
-    return chainId
-  }
+  return Object.entries(chains).find(([key]) => key === shortName)?.[1]
```
iamacook, line 41: By removing the error we now need to check that the returned value exists where it's used. We could fallback to an empty string but I think that could cause issues down the line.

I'm surprised TypeScript didn't pick up on this.

katspaugh: Let's return the default chainId as a fallback.

src/hooks/usePathRewrite.ts
```diff
@@ -17,6 +22,15 @@ const usePathRewrite = () => {
     if (Array.isArray(safe)) safe = safe[0]
     if (!safe) return
 
+    const { prefix } = parsePrefixedAddress(safe)
+
+    const chainId = Object.entries(chains).find(([key]) => key === prefix)?.[1]
```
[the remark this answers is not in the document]
katspaugh: Chain id can be also passed as its own query param.

src/hooks/useChainId.ts
usame-algan: This returns undefined in case it can't find the chain in our config. The behaviour should be the same as what we had previously. The fallbacks to the last session chain id and default chain id are already in place in `useChainId`.

src/hooks/usePathRewrite.ts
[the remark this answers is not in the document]
usame-algan: Should we check for that case here? i.e. we try to open `/bogus:0x00...?chain=gor` imo it should still navigate to the welcome page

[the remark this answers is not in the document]
katspaugh: I mean, it can be _just_ the chain param. E.g. `/welcome?chain=bogus`.
`useChainId` handles both.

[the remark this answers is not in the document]
usame-algan: Correct it will now stay on the last opened chain / default chain but still display the `?chain=bogus` query param where previously it would crash the app. Should we remove the chain query param if its not a valid chain? Afaik we are not preserving it when navigating through the app.

src/services/exceptions/ErrorCodes.ts
```diff
@@ -27,13 +27,13 @@ enum ErrorCodes {
...
   _619 = '619: Error fetching data from master-copies',
   _620 = '620: Error loading chains',
-  _621 = '621: Invalid chain short name in the URL',
 
   _700 = '700: Failed to read from local/session storage',
   _701 = '701: Failed to write to local/session storage',
   _702 = '702: Failed to remove from local/session storage',
   _703 = '703: Error importing an address book',
   _704 = '704: Error importing global data',
+  _705 = '705: Invalid chain short name in the URL',
```
katspaugh, line 36: I would use the 1xx range, 7xx are all about local storage.

iamacook (approved):

PR #1165, opened 2022-11-16
fix: E2E tests

iamacook:
## What it solves

Resolves broken E2E tests

## How this PR fixes it

`create_tx.cy.js`
- Add longer timeouts to estimate/propose endpoints

`load_safe.cy.js`
- Force click "+ Add Safe" (to avoid banners)
- Correct expected address inputs in address step of the load Safe flow

## How to test it

Observe all E2E tests pass.


cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -42,7 +42,7 @@ describe('Queue a transaction on 1/N', () => {
     cy.contains('h2', 'Review transaction').parents('div').as('modal')
 
     // Wait for /estimations response
-    cy.wait('@estimations')
+    cy.wait('@estimations', { timeout: 30_000 })
```
iamacook, line 45: We need to wait longer as Cypress intercepts every network request.

usame-algan (approved):

DiogoSoaress (approved):
🤞

cypress/e2e/smoke/load_safe.cy.js
```diff
@@ -19,8 +19,8 @@ describe('Load existing Safe', () => {
     cy.visit('/welcome?chain=matic')
     cy.contains('Accept selection').click()
 
-    // Enters Loading Safe form
-    cy.contains('Add existing Safe').click()
+    // Enters Loading Safe form - force if banner is present
+    cy.contains('Add existing Safe').click({ force: true })
```
katspaugh, line 23: Wouldn't it be better to close the banner in `before()`?

iamacook: We'd need to remove it with the banner then, and dynamic Beamer banners would also break the test.

liliya-soroka (approved):

PR #1167, opened 2022-11-16
Fix: added safes across chains

### LINKED ISSUE #1166: [Analytics] Added Safes count event isn't useful
## Bug description
The current ADDED_SAFES event is per chain and is tracked on add, which is not very useful. What would be more useful is knowing how many Safes a user has across chains.

katspaugh:
## What it solves

Resolves #1166

## How this PR fixes it
Added Safes across chains and queue size are now tracked in `useGtm`.

* Added Safe will be tracked on page load and whenever a Safe is added
* The queue size will be tracked when one opens a safe

## How to test it
Trigger those events and see that the correct value is sent.

## Analytics changes

I also removed the old per-chain Added Safes event, so it should not be tracked anymore.

## Screenshots
Screenshot 2022-11-16 at 12 21 13

### DIFF
--- a/src/services/analytics/events/overview.ts
+++ b/src/services/analytics/events/overview.ts
@@ -31,9 +31,9 @@ export const OVERVIEW_EVENTS = {
     action: 'Sidebar',
     category: OVERVIEW_CATEGORY,
   },
-  ADDED_SAFES_ON_NETWORK: {
+  TOTAL_ADDED_SAFES: {
     event: EventType.META,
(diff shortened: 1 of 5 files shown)


iamacook (approved):

liliya-soroka:
waiting for the confirmation from the prod team to check and approve the PR on the QA side - [slack discussion](https://5afe.slack.com/archives/C03DAGWJCR5/p1668599046525949)

katspaugh:
@liliya-soroka confirmation has been given by Johannes.

PR #1168, opened 2022-11-16
fix: pressing enter doesn't set max

### LINKED ISSUE #1164: [Transfer funds] Pressing enter will set the amount to max
## Bug description
https://github.com/5afe/safe-support/issues/66

When pressing enter in the transfer funds modal the amount is set to max.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to transfer funds
 2. enter an amount
 3. press enter

## Expected result
Either nothing should happen or we should submit the form

## Obtained result
The amount is set to max

iamacook:
## What it solves

Resolves #1164

## How this PR fixes it

The `InputValueHelper` required the `type="button"` attribute to prevent it from submitting the parent form and `shouldValidate` to be `true` in `onMaxAmountClick`. I also wrapped `InputValueHelper` in the correct `InputAdornment` component.

## How to test it

Attempt to send tokens and, after selecting one and entering an amount, press the enter key whilst the field is in focus. Observe no change in value or form submission. Doing so in an empty field should also result in the same.

Clicking "Max" should set the max value, as well as validate the amount, e.g. when no token amount exists, it should show an error regarding a value of `0`.


src/components/common/InputValueHelper/index.tsx
```diff
@@ -3,18 +3,29 @@ import { Link } from '@mui/material'
...
+const InputValueHelper = ({
+  children,
+  // We use onMouseDown instead of onClick to avoid "Enter" from triggering it
+  onMouseDown,
+  disabled = false,
+}: InputValueHelperProps) => {
   const handleClick = (e: SyntheticEvent) => {
     e.preventDefault()
-    onClick()
+    onMouseDown()
```
usame-algan, line 18: Will this still work on mobile devices?

usame-algan:
The examples in the MUI Documentation don't seem to react to an enter press even though there is an onClick handler attached to it: https://mui.com/material-ui/react-text-field/#input-adornments

It might have to do with the missing `InputAdornment` wrapper.

katspaugh (commented):
I think the problem here isn't onClick per se, but the fact that it thinks the button is a submit button.
`type=button` alone should already fix it.
You can also add `e.preventDefault()` in the onClick handler.

src/components/common/InputValueHelper/index.tsx
iamacook: I've reverted to using `onClick`.

iamacook:
After reverting to `onClick`, `type` did the job. I've wrapped it in an `InputAdornment` nonetheless as it's the "correct" implementation. (`e.preventDefault()` was already there. I'll leave it as such.)

schmanu (approved):
Nice!
I tested it again and it works nicely:
Enter submits the form and leads to the next step if the form data is valid.
The next step does not submit on enter (which is correct as there is no text input).

This matches the [HTML spec](https://www.w3.org/MarkUp/html-spec/html-spec_8.html#SEC8.2):
> When there is only one single-line text input field in a form, the user agent should  accept Enter in that field as a request to submit the form.

katspaugh:
Note to the QA: let's start testing forms using keyboard shortcuts and not just with a mouse. Tab + Tab + Tab + Enter

PR #1181, opened 2022-11-16
fix: reduce threshold when removing owner

### LINKED ISSUE #1178: [Safe creation]Threshold number is missed when the owner is deleted from the step 3 if threshold=max number
## Bug description
Threshold number is missed if the owner is deleted from the step 3 but the old max threshold is used on the review screen in safe configuration 

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain:Goerli

## Steps to reproduce
 1. Go to safe creation flow
 2. go to the safe owners step
 3.  add 2 owners
 4. set up threshold as 2
 5. remove one owner
 Current result: The setup threshold is empty.
image
image

## Expected result
the threshold should take into account removed owner if it was set up to the max available number

relates to : https://github.com/safe-global/web-core/issues/864

iamacook:
## What it solves

Resolves #1178

## How this PR fixes it

The `threshold` if the remaining number of owners is less than it.

## How to test it

Add two owners in the Safe creation flow and set the threshold to 2. Delete the second owner and observe the threshold change to 1.

## Screenshots

threshold


src/components/new-safe/steps/Step2/index.tsx
```diff
@@ -49,11 +49,18 @@ const CreateSafeStep2 = ({
...
+  const { handleSubmit, control, watch, formState, getValues, setValue } = formMethods
 
   const threshold = watch(CreateSafeStep2Fields.threshold)
 
-  const { fields: ownerFields, append: appendOwner, remove: removeOwner } = useFieldArray({ control, name: 'owners' })
+  const { fields: ownerFields, append: appendOwner, remove } = useFieldArray({ control, name: 'owners' })
+
+  const removeOwner = (index: number): void => {
+    if (ownerFields.length === threshold) {
+      setValue('threshold', threshold - 1)
```
usame-algan, line 60: Its possible to do `setValue(CreateSafeStep2Fields.threshold, threshold - 1)` if we set the `CreateSafeStep2Form` type differently further up:

```
export type CreateSafeStep2Form = {
  [CreateSafeStep2Fields.owners]: NamedAddress[]
  [CreateSafeStep2Fields.threshold]: number
}
```

src/components/new-safe/steps/Step2/index.tsx
```diff
@@ -49,11 +49,18 @@ const CreateSafeStep2 = ({
...
-  const { register, handleSubmit, control, watch, formState, getValues } = formMethods
+  const { handleSubmit, control, watch, formState, getValues, setValue } = formMethods
 
   const threshold = watch(CreateSafeStep2Fields.threshold)
 
-  const { fields: ownerFields, append: appendOwner, remove: removeOwner } = useFieldArray({ control, name: 'owners' })
+  const { fields: ownerFields, append: appendOwner, remove } = useFieldArray({ control, name: 'owners' })
+
+  const removeOwner = (index: number): void => {
+    if (ownerFields.length === threshold) {
```
schmanu, line 63: I feel like `setValue('threshold', Math.min(threshold, ownerField.length - 1))` is more direct.
But this might be very subjective.

src/components/new-safe/steps/Step2/index.tsx
iamacook: I've changed all field name references to use the `enum`.

schmanu (approved):
Just a very subjective suggestion. Otherwise this looks good :)

usame-algan (approved):

PR #1182, opened 2022-11-16
fix: improve reliability of E2E tests

iamacook:
## What it solves

Unreliable transaction creation test.

## How this PR fixes it

- The timeout for E2E wallet connection has been increased
- Tests relevant to the estimation/proposal/safe-apps endpoint incerceptions have been moved inside [the `intercept` callback](https://docs.cypress.io/api/commands/intercept#Using-the-yielded-object)

## How to test it

Observe the test passes in the E2E job.

### DIFF
--- a/cypress/e2e/smoke/create_tx.cy.js
+++ b/cypress/e2e/smoke/create_tx.cy.js
@@ -16,7 +16,9 @@ describe('Queue a transaction on 1/N', () => {
 
   it('should create and queue a transaction', () => {
     // Assert that "New transaction" button is visible
-    cy.contains('New transaction').should('be.visible', { timeout: 30_000 })
+    cy.contains('New transaction', {
+      timeout: 60_000, // `lastWallet` takes a while initialize in CI
+    }).should('be.visible')
 
     // Open the new transaction modal
     cy.contains('New transaction').click()
--- a/cypress/e2e/smoke/create_tx.cy.js
+++ b/cypress/e2e/smoke/create_tx.cy.js
@@ -15,8 +15,11 @@ describe('Queue a transaction on 1/N', () => {
   })
 
   it('should create and queue a transaction', () => {
+    // Assert that "New transaction" button is visible
+    cy.contains('New transaction').should('be.visible', { timeout: 30_000 })
+
     // Open the new transaction modal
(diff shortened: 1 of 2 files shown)


cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -15,8 +15,13 @@ describe('Queue a transaction on 1/N', () => {
   })
 
   it('should create and queue a transaction', () => {
+    // Assert that "New transaction" button is visible
+    cy.contains('New transaction', {
+      timeout: 60_000, // `lastWallet` takes a while initialize in CI
```
iamacook, line 20: This is horribly long but 30s was sometimes too short. It doesn't seem to need this long but I'd sooner have it longer than necessary. As soon as the button is present, it continues.

katspaugh (approved):
👍

cypress/e2e/smoke/create_tx.cy.js
DiogoSoaress: Why the long timeout in this assertion? There is no request before this modal opens. 
If this is a bottleneck disregard my comment -> I thought the timeout would come only with the `/estimations` request

cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -35,89 +40,83 @@ describe('Queue a transaction on 1/N', () => {
...
-    cy.wait('@estimations', { timeout: 30_000 })
-
-    // Estimation is loaded
-    cy.get('button[type="submit"]').should('not.be.disabled')
-
-    // Gets the recommended nonce
-    cy.contains('Signing the transaction with nonce').should(($div) => {
-      // get the number in the string
-      recommendedNonce = $div.text().match(/\d+$/)[0]
+    cy.intercept('POST', '/**/multisig-transactions/estimations').then(() => {
```
DiogoSoaress, line 47: For how long does the test try to intercept the request? the default timeouts are usually 5sec so it might be worth to keep the 30sec timeout that we had

DiogoSoaress: if it waits for the request, all good. I will resolve then

DiogoSoaress (approved):
🚀

PR #1183, opened 2022-11-16
fix: redirect to `/home` of Safe present in URL

### LINKED ISSUE #1172: [Landing page] fix Try Demo button link
## Bug description
The link has a safe address, but it needs to have a "/home" at the end to actually go to that safe

## Environment
 - Browser: Any
 - Wallet: Any
 - Chain: Any
 - Env: https://gnosis-safe.io/

## Steps to reproduce
 1. Go to https://gnosis-safe.io/
 2. Check the try demo button in the middle

## Screenshots
image

iamacook:
## What it solves

Resolves #1172

## How this PR fixes it

If an address is present in the URL, it will redirect to the `/home` of that Safe.

## How to test it

Open `/{{shortName}}:{{address}}` with no session cache (e.g. in Incognito) and observe that it directs to `/{{shortName}}:{{address}}/home`. When opening the Safe with a `lastSafe` cached, it should open to that one.

All Safe-specific routes should function as before, e.g. opening `/{{shortName}}:{{address}}/settings/setup` should not direct to `/{{shortName}}:{{address}}/home`.

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -6,18 +6,19 @@ import { AppRoutes } from '@/config/routes'
 
 const IndexPage: NextPage = () => {
   const router = useRouter()
-  const { chain } = router.query
+  const { safe, chain } = router.query
   const lastSafe = useLastSafe()
+  const safeAddress = safe || lastSafe
 
   useLayoutEffect(() => {
     router.replace(
-      lastSafe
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
👍

DiogoSoaress (approved):

PR #1193, opened 2022-11-18
Fix: sandbox safe app icons

katspaugh:
## What it solves

Loading Safe App icons directly inadvertently tracks the user on those websites, which is a security/privacy concern.

This PR wraps icons in sandboxed iframes and this effectively prevents 3rd-party cookies from being set.

I've tested this in Safari, Chrome, and Firefox.

## How to test
On dev or prod:
* Open the app, clear the cookies, local storage, etc
* Open the Dashboard and/or the Apps page
* Observe that a lot of 3rd-party cookies (e.g. Zerion) are set on their corresponding domains.

On this branch:
* External cookies will not be set unless you open a Safe App

See [this doc](https://www.notion.so/gnosis-safe/Third-party-cookies-tracking-through-Safe-App-Icons-fd8cc28e3ccb4de7a515eeee25c3b5cb) for details.


iamacook:
We should also use this in the transaction modal. The icon is embedded in the top left.

iamacook (approved):

francovenica:
Looks good. I tested in the new build https://pr1193--webcore.review-react-br.5afe.dev/

A comparison between prod and this PR
Prod
image

PR: (I opened 2 apps to see the cookies being listed in the sidebar)
image


Note:
I tested in brave to cover all major browsers, but it seems that it blocks most cookies, even when its "shield" is down.
This is everything it tracked in current prod, even after entering some apps.
In another time we should check how brave handles cookies in general.
image

PR #1194, opened 2022-11-18
fix: flag Keystone wallets as hardware wallets + adjust modal `z-index`

### LINKED ISSUE #1110: [Goerli & L2s] Keystone not working
> Can not create a new safe wallet on Goerli testnet.  The error message shows that the chain ID does not match the chain ID of common. Please see the first picture for more details below.

> Transaction sent failed on Goerli testnet. The error message shows that the method eth_signTypeData_v4 does not exist/is not available.  Please see the second screenshot for more details below.

> We did not test all the chains. And from what we’ve got now, it only works on the Ethereum mainnet but not on any other chains we tested.

iamacook:
## What it solves

Resolves #1110

## How this PR fixes it

Keystone wallets are now marked as hardware wallets and the `z-index` of the Keystone modal container has been increased.

## How to test it

Connect a Keystone wallet to a certain network then switch network. The wallet should be disconnected and the connection window reopened.

All Keystone modals in the Safe should be "above" other elements.

### DIFF
--- a/src/styles/onboard.css
+++ b/src/styles/onboard.css
@@ -32,6 +32,9 @@
   --onboard-warning-600: var(--color-error-main);
   --onboard-warning-700: var(--color-error-dark);
 
+  /* var(--color-backdrop-main) + opacity */
+  --onboard-modal-backdrop: rgba(99, 102, 105, 0.75);
+
   --account-select-modal-white: var(--color-background-paper);
   --account-select-modal-black: var(--color-primary-main);
   --account-select-modal-primary-100: var(--color-secondary-background);
--- a/src/styles/onboard.css
+++ b/src/styles/onboard.css
@@ -69,8 +69,9 @@
   --onboard-border-radius-3: 8px;
 
   /* Z-index */
-  --onboard-modal-z-index: 1201;
-  --onboard-account-select-modal-z-index: 1201;
+  /* General modals are 998 because Keystone modal is 999 */
+  --onboard-modal-z-index: 997;
+  --onboard-account-select-modal-z-index: 997;
 }
 
 #walletconnect-qrcode-modal {
--- a/src/styles/onboard.css
+++ b/src/styles/onboard.css
@@ -83,5 +83,5 @@
 
 /* Keystone modal */
 #kv_sdk_container + .ReactModalPortal > div {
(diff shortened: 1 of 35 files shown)


katspaugh (approved):
👍

src/styles/theme.ts
```diff
@@ -250,6 +250,12 @@ const initTheme = (darkMode: boolean) => {
         defaultProps: {
           fullWidth: true,
         },
+        styleOverrides: {
+          root: {
+            // Keystone modals have a z-index of 999
+            zIndex: 998,
+          },
+        },
```
usame-algan, line 258: This now hides the mobile pairing QR code behind the onboard modal again.

usame-algan (changes requested):

src/styles/theme.ts
iamacook: Good catch, I've lowered the `z-index` of onboard and it works as expected now:

image

src/styles/onboard.css
```diff
@@ -69,8 +69,9 @@
   --onboard-border-radius-3: 8px;
 
   /* Z-index */
-  --onboard-modal-z-index: 1201;
-  --onboard-account-select-modal-z-index: 1201;
+  /* General modals are 998 because Keystone modal is 999 */
+  --onboard-modal-z-index: 997;
+  --onboard-account-select-modal-z-index: 997;
```
katspaugh, line 74: Not sure this a good idea. The default modal zIndex in MUI is 1200.

iamacook: [I've set the modal `z-index` to `998`.](https://github.com/safe-global/web-core/blob/keystone/src/styles/theme.ts#L256)

katspaugh: What about the header and sidebar?

iamacook: It seems like there are a lot of dependents. I look into whether Keystone can provide us with a CSS variable to alter this ourselves.

usame-algan (approved):
LGTM

iamacook:
@liliya-soroka, this should work as I set the `z-index` to the value we found in our tests. Please confirm in the PR directly though.

liliya-soroka:
Verified for Kestone wallet

Regression done for :
- MM
- Mobile pairing

PR #1200, opened 2022-11-18
fix: QR reader crashes app

### LINKED ISSUE #1171: [Safe creation] The browser hangs if "Upload an image" is used in "scan QR code" on add owners step
1. start safe creation flow
2. add name
3. go to the next step 
4. click add owner->scan QR code
5. Upload an image -> Upload a valid QR code
Current result: The browser hangs
Expected result: It should be possible to add an owner using "Upload an image" on Scan QR code window

ENV: https://creation.web-core.pages.dev/new-safe/create
wallet: paired mobile app
relates to : https://github.com/safe-global/web-core/issues/864

### LINKED ISSUE #1185: [Safe creation] Uploading a non-QR code image for owner crashes the page
## Bug description
Trying a png that is not a QR code and then trying to upload something again makes the page crash after a while (I think there is an infinite loop here)

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Goerli
 - Epic: #864 

## Steps to reproduce
 1. Go to Owner step for safe creation
 2. Upload a non-QR code image
 3. Click "Upload image" again
 4. Try to pick another image from your folder

## Expected result
If picked another image this is uploaded an if it is a valid QR it reads it as the address
If it is not a valid QR shows the error again

## Obtained result
The Tab freezes and after a while the browser ask you to close the page (infinite loop probably)

## Screenshots
Frozen page

usame-algan:
## What it solves

Resolves #1171 
Resolves #1185 

## How this PR fixes it

- Uses `useRef` instead of `createRef` and fixes side effect dependencies to prevent rerenders

## How to test it

1. Open the Safe
2. Create a new safe
3. Navigate to the add owner screen
4. Press the QR Button with camera allowed
5. Press the Upload an Image button
6. Observe the app not crashing

### DIFF
--- a/src/styles/theme.ts
+++ b/src/styles/theme.ts
@@ -57,7 +57,7 @@ const initTheme = (darkMode: boolean) => {
     },
     spacing: base,
     shape: {
-      borderRadius: '6px',
+      borderRadius: 6,
     },
     shadows: [
       'none',
--- a/src/components/common/ScanQRModal/index.tsx
+++ b/src/components/common/ScanQRModal/index.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, createRef, useState } from 'react'
+import { useState, useRef, useEffect } from 'react'
 import { Box, Dialog, DialogTitle, IconButton, Button, Divider } from '@mui/material'
 import QrReader from 'react-qr-reader'
 import CloseIcon from '@mui/icons-material/Close'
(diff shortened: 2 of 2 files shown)


src/styles/theme.ts
```diff
@@ -57,7 +57,7 @@ const initTheme = (darkMode: boolean) => {
     },
     spacing: base,
     shape: {
-      borderRadius: '6px',
```
iamacook, line 60: Nit: I think we should keep this for consistency (I initially thought it would use theme spacing).

iamacook (approved):

src/styles/theme.ts
usame-algan: I only changed the value to be `6` instead of `6px` because there were errors in the console.

katspaugh (approved):

PR #1216, opened 2022-11-22
feat: EIP-1271 event bus, notifications + statuses

### LINKED ISSUE #1205: [EIP-1271] Events, statuses and notifications
## What is the feature about

EIP-1271 event bus and notifications

## The list of requirements

- Create equivalent of `txEvents.ts` with `MsgEvents` defining: proposed (failure), signed (failure) and confirmed
- Create equivalent of `useTxNotifications` for the above events
- Create "pendingMessages" slice and use similar logic to setting/clearing pending status as pending transactions
- Dispatch notification regarding messages awaiting confirmations
- Update `useIsMsgPending` to reference the above slice(s)
- Use the above to extend `useMessageStatus`

## Links

- [Problem statement/data structure](https://www.notion.so/gnosis-safe/EIP-1271-signatures-can-only-be-generated-on-chain-2800b5e516124d959c38ba8f2042f9f7?d=e8ca1d332df54ae587e66240fd40e230#a8489e3724234f75a9c3d389e323a725)

iamacook:
## What it solves

Resolves #1205

## How this PR fixes it

- `msgEvents` and a `MsgEvents` event bus was created covering: message creation/failure, confirmation/failure, successful/full confirmation. (There are comments explasining each one).
- `pendingMsgsSlice` incorporates a hash map of `{ [messageHash]: true }` for messages being created/confirmed.
- `useIsMsgPending` references the `pendingMsgsSlice`.
- `useMsgPendingStatuses` sets messages as pending/clears them.
- `useMessageStatus` returns a string status depending on the remove/local statuses of messages.
- `useMsgNotifications` dispatches notifications according to the above.

## How to test it

n/a as this is a logic framework. Testing will be done in a forthcoming PR. Test coverage for all hooks has, however, been included.


src/hooks/useMessageStatus.ts
```diff
@@ -1,15 +1,34 @@
...
-  return STATUS_LABELS[item.status]
+const useMessageStatus = (msg: Message) => {
+  const isPending = useIsMsgPending(msg.messageHash)
+  const wallet = useWallet()
+
+  if (isPending) {
+    return STATUS_LABELS[ConfirmingStatus]
+  }
+
+  const hasWalletSigned = wallet && msg.confirmations.some(({ owner }) => owner.value === wallet.address)
```
schmanu, line 26: This will return "Awaiting confirmations" instead of "Confirmed" if a message is fully confirmed.
Let's add a test case for that: "Should show the message status if the message is fully confirmed"

src/hooks/useMsgPendingStatuses.ts
```diff
@@ -0,0 +1,44 @@
...
+    const unsubFns = entries.map(([event, isPending]) =>
+      msgSubscribe(event, (detail) => {
+        const messageHash = 'messageHash' in detail && detail.messageHash
+
+        if (!messageHash) {
+          return
+        }
+
+        if (!isPending) {
+          dispatch(clearPendingMsg(messageHash))
```
usame-algan, line 31: In `useTxPendingStatuses` there is an early return for this case that is missing here.

src/hooks/useMsgPendingStatuses.ts
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        dispatch(setPendingMsg(messageHash))
```
usame-algan, line 34: When setting a pending tx we are storing a lot more meta data like `chainId`, `txId`, `status`, `txHash`, `groupKey`, `signerAddress`. Do we need any of them when setting a pending message?

src/hooks/useMsgNotifications.ts
```diff
@@ -0,0 +1,47 @@
...
+const useMsgNotifications = () => {
+  const dispatch = useAppDispatch()
+
+  useEffect(() => {
+    const entries = Object.entries(MsgNotifications) as [keyof typeof MsgNotifications, string][]
+
+    const unsubFns = entries.map(([event, baseMessage]) =>
+      msgSubscribe(event, (detail) => {
+        const isError = 'error' in detail
+        const isSuccess = event === MsgEvent.FULLY_CONFIRMED
```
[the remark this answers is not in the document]
iamacook: As per our discussion (what was `MsgEvent.CREATE, now `MsgEvent.PROPOSE`) is deemed a successful event matching that of `useTxNotifications`.

Regarding the latter (now `MsgEvent.CONFIRM_PROPOSE`), we will keep it as `info` to match that of `useTxNotifications`.

We can revisit adjusting the both of them but we don't want to get out of scope.

src/services/msg/msgEvents.ts
```diff
@@ -0,0 +1,40 @@
+import EventBus from '../EventBus'
+
+export enum MsgEvent {
+  CREATE = 'CREATE',
+  CREATE_FAILED = 'CREATE_FAILED',
+  CONFIRM = 'CONFIRM',
+  CONFIRM_FAILED = 'CONFIRM_FAILED',
+  CONFIRMATION_SAVED = 'CONFIRMATION_SAVED',
```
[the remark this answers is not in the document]
iamacook: As discussed, these have been updated and explanatory comments moved up into the `enum`.

src/hooks/useMessageStatus.ts
iamacook: I've adjusted the logic and added a test accordingly.

src/hooks/useMsgPendingStatuses.ts
iamacook: I've updated the logic and adjusted the tests to check this as well.

src/hooks/useMsgPendingStatuses.ts
iamacook: The `messageHash` should be unique every time.

schmanu (approved):
My issues were all addressed :) All good from my side!

PR #1219, opened 2022-11-22
feat: pass dark mode in the widgets url

DiogoSoaress:
## What it solves
Part of https://github.com/safe-global/web-core/issues/1221
Allows passing a dark mode flag in the url to the governance widgets

## How this PR fixes it
Reads app dark setting and passes it in the `appUrl`

## How to test it
Needs https://github.com/safe-global/safe-react-apps/pull/580 to be merged or it will break parsing the widgets ID

## Analytics changes
N/A

## Screenshots
Nothing changes

### DIFF
--- a/src/components/dashboard/GovernanceSection/GovernanceSection.tsx
+++ b/src/components/dashboard/GovernanceSection/GovernanceSection.tsx
@@ -22,7 +22,7 @@ const GovernanceSection = () => {
   const fetchingSafeClaimingApp = !claimingApp && !errorFetchingClaimingSafeApp
 
   return (
-    <Accordion className={css.accordion} defaultExpanded={true}>
+    <Accordion className={css.accordion} defaultExpanded>
       <AccordionSummary expandIcon={<ExpandMoreIcon color="border" />}>
         <div>
           <Typography component="h2" variant="subtitle1" fontWeight={700}>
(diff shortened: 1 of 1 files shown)


iamacook (changes requested):
We should adjust the `safe-apps-sdk` to support this instead of passing it via the URL.

usame-algan:
I agree that this should be handled in the `safe-apps-sdk` to make it available to all safe apps but for the current scope of displaying the governance widgets on the dashboard I think passing it via the URL is fine. Can we add a TODO to remove it later on?

DiogoSoaress:
> We should adjust the safe-apps-sdk to support this instead of passing it via the URL.

That falls out of scope for this Epic. Let's stick with passing the `+dark` in the URL as per https://github.com/safe-global/web-core/issues/1221 @iamacook

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
```diff
@@ -43,7 +45,7 @@ const GovernanceSection = () => {
                   {claimingApp ? (
                     <SafeAppsErrorBoundary render={() => <SafeAppsLoadError onBackToApps={() => {}} />}>
                       <AppFrame
-                        appUrl={`${claimingApp.url}${SNAPSHOT_WIDGET_ID}`}
+                        appUrl={`${claimingApp.url}${SNAPSHOT_WIDGET_ID}${isDarkMode ? '+dark' : ''}`}
```
iamacook, line 48: I would recommend we sooner pass this as a search parameter, otherwise it is part of the `hash`.

iamacook (approved):

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
katspaugh: I disagree with this. Widget if is passed in the hash, why would dark mode be passed differently?

katspaugh: Besides, a search param invalidates the HTTP cache. So you’d have to download that app again when the theme changes.

But most importantly I just don’t get why we come up with arbitrary requests to change the code like that.

PR #1226, opened 2022-11-23
fix: show warning based on `trusted` flag

iamacook:
## What it solves

Resolves default warning when no signatures exist.

## How this PR fixes it

A warning should no longer shown by default when no signatures are present, e.g. when queuing an approval of a nested Safe. Instead, they are shown when the `trusted` flag is `false`.

Note: no warning will be shown [until the flag is implemented](https://github.com/safe-global/safe-client-gateway/issues/1005).

## How to test it

Open an "untrusted" transaction and observe the warning

### DIFF
--- a/src/components/transactions/TxDetails/index.tsx
+++ b/src/components/transactions/TxDetails/index.tsx
@@ -46,8 +46,9 @@ const TxDetailsBlock = ({ txSummary, txDetails }: TxDetailsProps): ReactElement 
   const awaitingExecution = isAwaitingExecution(txSummary.txStatus)
   const isUnsigned =
     isMultisigExecutionInfo(txSummary.executionInfo) && txSummary.executionInfo.confirmationsSubmitted === 0
-  const isTrusted =
-    isMultisigDetailedExecutionInfo(txDetails.detailedExecutionInfo) && txDetails.detailedExecutionInfo.trusted
+  const isUntrusted =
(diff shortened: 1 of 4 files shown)


src/components/transactions/Warning/index.tsx
```diff
@@ -62,9 +62,9 @@ export const ThresholdWarning = (): ReactElement => (
   />
 )
 
-export const UnsignedWarning = (): ReactElement => (
+export const UntrustedWarning = (): ReactElement => (
   <Warning
-    title="This transaction is unsigned and could have been created by anyone. To avoid phishing, only sign it if you trust the source of the link."
```
katspaugh, line 67: I think this wording was better.

src/components/transactions/TxDetails/index.tsx
```diff
@@ -75,7 +78,7 @@ const TxDetailsBlock = ({ txSummary, txDetails }: TxDetailsProps): ReactElement
         )}
 
         <div className={css.txSummary}>
-          {isUnsigned && <UnsignedWarning />}
+          {isTrusted === false && <UnsignedWarning />}
```
katspaugh, line 81: This is also not correct because if `isMultisigDetailedExecutionInfo(txDetails.detailedExecutionInfo)` returns false, you will show the warning.

katspaugh (approved):

PR #1230, opened 2022-11-23
refactor: Remove unused code that was duplicated

usame-algan:
## What it solves

Part of #864 

## How this PR fixes it

- Removes code that was duplicated at the start of #864 but never used
- Adds the AB test again

### DIFF
--- a/src/components/new-safe/steps/Step4/index.tsx
+++ b/src/components/new-safe/steps/Step4/index.tsx
@@ -26,8 +26,6 @@ export const SAFE_PENDING_CREATION_STORAGE_KEY = 'pendingSafe'
 export type PendingSafeData = NewSafeFormData & {
   txHash?: string
   tx?: PendingSafeTx
-  safeAddress?: string
-  saltNonce: number
 }
 
 export const CreateSafeStatus = ({ setProgressColor }: StepRenderProps<NewSafeFormData>) => {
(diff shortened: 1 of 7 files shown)


src/pages/open.tsx
```diff
@@ -17,10 +17,6 @@ const Open: NextPage = () => {
     }
   }, [router, shouldUseNewCreation])
 
-  if (shouldUseNewCreation) {
-    return <></>
-  }
```
usame-algan, line 22: This was causing hydration errors on my end but now there is a small loading flash which is also not ideal.

src/components/new-safe/steps/Step4/index.tsx
```diff
@@ -26,8 +26,6 @@ export const SAFE_PENDING_CREATION_STORAGE_KEY = 'pendingSafe'
 export type PendingSafeData = NewSafeFormData & {
   txHash?: string
   tx?: PendingSafeTx
-  safeAddress?: string
-  saltNonce: number
```
usame-algan, line 30: These are already part of `NewSafeFormData` so I removed them here

src/pages/open.tsx
schmanu: Could we not render different links on the welcome page instead of replacing the route?

schmanu (approved):
The removed parts / refactoring looks good!

src/pages/open.tsx
usame-algan: Good idea! I will try that

usame-algan: Seems to work good. I also had to update it for the safe app landing pages since they also link to the safe creation but I think its fine cc @iamacook

PR #1232, opened 2022-11-23
Bug: safes 1.3.0 cannot execute transactions with safeTxGas

mikhailxyz:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/710

## How this PR fixes it

The problem was caused by Core SDK overwriting passed safeTxGas to 0 if the safe version was >= 1.3.0, this was fixed https://github.com/safe-global/safe-core-sdk/issues/292 and the PR pulls the updated SDK into the web UI

## How to test it
1. Create a transaction with custom safeTxGas (prob via a script, can be found in linked sdk issue)
2. Execute it from the interface
------------
1. Create a transaction with safe 1.3.0 from the interface
2. It shouldn't have safeTxGas set and execute just fine
------------
1. Create a transaction with safe 1.1.1 from the interface
2. It should have safeTxGas set and execute just fine

## Additional info

My test transaction - http://localhost:3000/gor:0x6Aa244e4eeE4F31d391674BD8cDf6Bc38Ad1D83B/transactions/tx?id=multisig_0x6Aa244e4eeE4F31d391674BD8cDf6Bc38Ad1D83B_0x3c67fd3ebbddcb6d53068053ff6b02d3f0ab8f7977dd301c9d610976bf6c652c


package.json
```diff
@@ -65,7 +65,7 @@
     "classnames": "^2.3.1",
     "date-fns": "^2.29.2",
     "ethereum-blockies-base64": "^1.0.2",
-    "ethers": "^5.6.8",
+    "ethers": "5.7.2",
```
mikhailxyz, line 68: the core sdk imports type from ethers and it has a constant string type on `version` variable equal to 5.7.2, hence the pinning

iamacook (approved):

germartinez (changes requested):
With the new release of the SDK, if a `safeTxGas` is provided to the method `createTransaction`, that same `safeTxGas` will be used. In the previous releases it was replaced with 0 by the SDK for Safes with version 1.3.0.

Knowing this, to keep sending `safeTxGas=0` for v1.3.0 Safes, the estimation that is executed inside `getRecommendedTxParams()` method should be avoided for version 1.3.0, and no `safeTxGas` should be added for that version.

src/services/tx/txSender.ts
```diff
@@ -72,6 +72,9 @@ const getRecommendedTxParams = async (
   const safeSDK = getAndValidateSafeSDK()
   const chainId = await safeSDK.getChainId()
   const safeAddress = safeSDK.getAddress()
+  const contractVersion = await safeSDK.getContractVersion()
+  const safeTxGasRequired = isSafeTxGasRequired(contractVersion)
```
germartinez, line 76: I would rewrite this as follows, but not a big deal
```js
  const isSafeTxGasRequired = isLegacyVersion(contractVersion)
```

germartinez (approved):

src/services/tx/txSender.ts
mikhailxyz: done señor

iamacook (approved):

katspaugh (approved):

francovenica:
Here Mikha helped by creating a tx with a SafeTxGas > 0 in a safe 1.3.0.

The warning is showing no more which is correct, still we have to hide the SafeTxGas for safes 1.3.0. 

https://pr1232--webcore.review-react-br.5afe.dev/gor:0x531BF97b83973E8aAfb7B5E201Ca279e3948A2aE/transactions/tx?id=multisig_0x531BF97b83973E8aAfb7B5E201Ca279e3948A2aE_0x2aa5ca58469edabeca41e9869382c6742778b623472f3c320fadf4ca0aae239f

image

mikhailxyz:
@francovenica The safe transaction gas should be visible because it's a kinda important part of the transaction details. But it should be immutable. If I click `edit`, it is immutable as expected


Screenshot 2022-11-28 at 14 42 51

mikhailxyz:
@francovenica is it good to go now?

PR #1235, opened 2022-11-24
Refactor: paginated NFT select

### LINKED ISSUE #1227: [NFTs] In the Send NFT modal, we load all collections in the autocomplete
In order to list all of a Safe's NFT collections, we load each page of the paginated NFTs in a closed loop. 

This is obviously against the very purpose of pagination, so we need to find a way to not do that.

Screenshot 2022-11-23 at 12 02 42

katspaugh:
## What it solves

Resolves #1227

Screenshot 2022-11-24 at 09 48 20

The select is now using InfiniteScroll to load more NFTs as you scroll the popup.

## How to test it
- Open a Safe with tons of NFTs
- Open the New transaction -> Send NFT modal
- Try to select some NFT beyond the first page


usame-algan (approved):

francovenica:
LGTM
nft list

PR #1243, opened 2022-11-24
feat: connect list, propose and confirm endpoints

### LINKED ISSUE #1159: [EIP-1271] Fetching of messages
## What is the feature about

Implementation of EIP-1271 "Messages" store/polling.

## The list of requirements

- Create "messages" slice
- Create `useInitMessages` poller hook
- Update the stubbed `useMessages` hook that returns mocked data from the accordion PR to use the store

## Links

- [Problem statement/data structure](https://www.notion.so/gnosis-safe/EIP-1271-signatures-can-only-be-generated-on-chain-2800b5e516124d959c38ba8f2042f9f7?d=e8ca1d332df54ae587e66240fd40e230#a8489e3724234f75a9c3d389e323a725)

iamacook:
## What it solves

Resolves #1159

## How this PR fixes it

The `safe-react-gateway-sdk` has been updated and all of the types accordingly. The list, propose and confirm endpoints have been connected to the relevant hooks: `useLoadSafeMessages`, `useMessages` and the dispatchers accordingly.

## How to test it

Programmatically call `dispatchSafeMsgProposal` and observe the new message in the list. Confirmation can be done via `dispatchSafeMsgConfirmation`.

Safe `gor:0xF3E977e7Eea1A91ce5B2d5077e5A7195Ae18b722` already has a submitted message.

Note: the gateway is fixing issues surrounding EIP-712 messages so only EIP-191 messages are working at the moment.


src/hooks/loadables/useLoadSafeMessages.ts
```diff
@@ -1,104 +1,21 @@
...
-      name: 'Alice',
-    },
-    preparedSignature: '0x123',
-  },
-]
 
 export const useLoadMessages = (): AsyncResult<SafeMessageListPage> => {
   const { safe, safeAddress, safeLoaded } = useSafeInfo()
 
+  // Re-fetch when chainId/address
```
usame-algan, line 13: nit: This comment just describes the dependency array and is not necessary imo

src/hooks/useSafeMessages.ts
```diff
@@ -1,31 +1,36 @@
...
+      return getSafeMessages(safe.chainId, safeAddress, pageUrl)
     },
     [safe.chainId, safeAddress, safeLoaded, pageUrl],
     false,
   )
 
+  // The latest page of the messages is always in the store
+  const messagesState = useAppSelector(selectSafeMessages)
+
+  // Return the new page or the stored page
```
usame-algan, line 33: Similar here, this comment just describes the next lines. Alternatively we could make it more obvious and remove the comment e.g.

```
const newPage = {...}
const storedPage = {...}

return pageUrl ? newPage : storedPaged
```

src/utils/safe-message-guards.ts
```diff
@@ -1,12 +1,10 @@
...
 
-export const isSafeMessageListDateLabel = (
-  item: SafeMessageListPage['results'][number],
-): item is SafeMessageDateLabel => {
+export const isSafeMessageListDateLabel = (item: SafeMessageListItem): item is SafeMessageDateLabel => {
   return item.type === SafeMessageListItemType.DATE_LABEL
 }
 
-export const isSafeMessageListItem = (item: SafeMessageListPage['results'][number]): item is SafeMessage => {
+export const isSafeMessageListItem = (item: SafeMessageListItem): item is SafeMessage => {
```
usame-algan, line 8: `isSafeMessage`?

src/components/safeMessages/MsgModal/index.tsx
```diff
@@ -1,29 +1,51 @@
...
+
+type ProposeProps = BaseProps & {
+  safeAppId: number
+  messageHash?: never
+}
+
+type ConfirmProps = BaseProps & {
+  safeAppId?: never
+  messageHash: string
+}
```
usame-algan, line 29: Not sure I understand these types. Why is it possible that the `safeAppId` is missing when confirming an existing message?

src/hooks/loadables/useLoadSafeMessages.ts
iamacook: This mirrors our other loading hooks.

src/hooks/loadables/useLoadSafeMessages.ts
```diff
@@ -1,104 +1,21 @@
...
     if (!safeLoaded) {
       return
     }
-    return Promise.resolve({
-      next: undefined,
-      previous: undefined,
-      results: MOCK_DATA,
-    })
+    return getSafeMessages(safe.chainId, safeAddress)
   }, [safeLoaded, safe.chainId, safeAddress])
```
schmanu, line 19: For the queue and history we have tags which get updated if something new is added. Do or will we have something similar for messages? Otherwise we will not poll them, right?

src/hooks/loadables/useLoadSafeMessages.ts
iamacook: These comments mirror our other polling hooks. However, I agree with you and have adjusted it accordingly.

src/hooks/useSafeMessages.ts
iamacook: I've also adjusted these, but only return the objects when necessary.

src/utils/safe-message-guards.ts
iamacook: I don't want to change these as they mirror that of the transaction list.

src/components/safeMessages/MsgModal/index.tsx
iamacook: I've added a comment here explaining why. A proposed message does not return the `safeAppId` but the `logoUri` and `name` of the Safe App that proposed it.

src/hooks/loadables/useLoadSafeMessages.ts
iamacook: Good catch! There is no `tag` for messages so I've added a polling interval as a dependency.

schmanu (approved):
🔥

PR #1249, opened 2022-11-25
fix: Send current step with tracking event

### LINKED ISSUE #1248: Stepper is not tracking current step
## Bug description

The stepper events that are sent when going to the next or previous step do not include the step number

This is needed for #864 

## Steps to reproduce
 1. Go to Safe creation
 2. Click Next
 3. Observe the `eventLabel` from the event does not contain the current step number

## Expected result

The events should include the current step number

usame-algan:
## What it solves

Resolves #1248 

## How this PR fixes it

- Adds the current step as label to the tracked event

## How to test it

1. Open the safe creation
2. Click Next
3. Observe that the GTM event in the console contains the current step as label


iamacook (approved):

katspaugh (approved):
👍

Issue #1225, opened 2022-11-23
Add network switch button to modals


iamacook:
## What is the feature about

When connected to the wrong chain, we should a button in the header but it would also be helpful to have one within related modals (`SignOrExecuteForm`) in order to not cancel transaction flows.

## The list of requirements

- Add button to `SignOrExecuteForm` when `isWrongChain`

## Designs/sketches

tbd

## Links

- [Feature request thread](https://5afe.slack.com/archives/C03DAGWJCR5/p1669146223597379)

## Screenshots

![image](https://user-images.githubusercontent.com/20442784/203491672-0e89065d-6abe-41f4-a79d-bdd8fed82a86.png)

katspaugh:
This can be trivially solved by replacing the Submit button with a Switch Network button in that scenario.

qqqzhch:
@ katspaugh  
How can I reproduce this problem? When I send a transaction to the preview page, and I switch the network to an unsupported network, the page automatically detects the network and jumps to the home page
I can't find such a dialog box
![image](https://user-images.githubusercontent.com/3917994/224460306-c56ed269-4e0f-4f45-bbe7-183eec7dff79.png)

katspaugh:
Closing this issue because it will be solved differently in https://github.com/safe-global/web-core/pull/1746
Thanks for checking it, @qqqzhch!

PR #1259, opened 2022-11-28
fix: Reset Core SDK when switching safes

### LINKED ISSUE #1253: Nonce of the previously opened safe is used  for a safe on another chain during tx creation
## Bug description
Nonce of the previously opened safe is used for a safe on another chain during tx creation 
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: any network

## Steps to reproduce
 1. MM is connected as a wallet
 2. the same MM address is used as an owner for the safes on Goerli and Ethereum chains
 3. go to the Goerli and create tx ( remember the nonce that was used for the tx creation)
 4. select safe on Ethereum using side menu
 5. go to the safe-apps
 6. use any safe app to create a tx
 Current result: Review tx screen is displayed with the nonce from the Goerli safe,the error about the owner on the wrong chain, and disabled submit button

7. Switch chain for the owner in MM from Goerli to the Ethereum
Current result: Submit button is enabled so the user can submit the tx, but the nonce was not updated to the Ethereum safe nonce.
So the user creates a tx with the wrong nonce on another chain

## Expected result
the correct nonce should be used during the tx creation.

usame-algan:
## What it solves

Resolves #1253 

## How this PR fixes it

There are 2 ways of switching to a safe on a different network:

1. User is on Safe A
2. They switch to Safe B and change their wallet network from A to B
---
1. User is on Safe A
2. They change their wallet network to B from within their wallet
3. They switch to Safe B

In the second case we try to initialize the Core SDK twice, once with stale data from Safe A and the second time with the correct data from Safe B. The first call throws a `SafeProxy contract is not deployed on the current network` error. Depending on which promise resolves first it then resets the Core SDK instance which breaks the UI.

### Other fixes

- Hides the Send button on the assets page and address book if the wallet is on a wrong network. This is the same behaviour as for all other buttons that can result in a transaction

## How to test it

Go the the assets page and address book and switch network in your wallet. The Send button should disappear.

1. Open a Safe on Görli
2. Navigate to a Safe app and create a transaction
4. Observe the correct nonce being shown in the modal
5. Switch to a different network from your wallet
6. Observe the modal showing a `Your wallet is connected to the wrong chain.` error
7. Switch to a Safe on the new network
8. Repeat steps 2-7

---

1. Open a safe app on mainnet while the wallet is set to a different network
2. Reload the page
3. Try to create a transaction so that the modal opens
4. Switch the wallet networ to mainnet
5. Observe that no error is being displayed and the correct nonce is there

---

Additionally, we should do a full regression on all types of transactions (Batch execute, create/remove/update spending limit, add/remove/change owner, sign/sign+execute/execute tx, ...)


src/hooks/coreSDK/useInitSafeCoreSDK.ts
```diff
@@ -23,7 +29,7 @@ export const useInitSafeCoreSDK = (): Error | null => {
         setSafeSDK(undefined)
         setError(e)
       })
-  }, [chain, wallet, safe, safeLoaded])
+  }, [chainId, wallet?.provider, wallet?.chainId, safe, safeLoaded])
```
usame-algan, line 32: Narrowing the `wallet` dependency here also reduced the number of Core SDK calls from 2 to 1 when reloading the page normally.

src/hooks/coreSDK/useInitSafeCoreSDK.ts
```diff
@@ -1,18 +1,24 @@
...
   const [error, setError] = useState<Error | null>(null)
 
   useEffect(() => {
-    if (!safeLoaded || !chain || !wallet || chain.chainId !== wallet.chainId) {
+    if (
+      !safeLoaded ||
+      !wallet?.provider ||
+      !wallet.chainId ||
+      chainId !== wallet.chainId ||
+      safe.chainId !== wallet.chainId
```
usame-algan, line 20: This additional condition prevents the second/first init. Could also look into possibly fixing it in `useSafeInfo`.

src/hooks/coreSDK/useInitSafeCoreSDK.ts
```diff
@@ -23,7 +29,7 @@ export const useInitSafeCoreSDK = (): Error | null => {
         setSafeSDK(undefined)
         setError(e)
       })
-  }, [chain, wallet, safe, safeLoaded])
+  }, [chainId, wallet?.provider, wallet?.chainId, safe, safeLoaded])
 
   return error
```
usame-algan, line 34: It also seems that we are not using this `error` at all thats why we didn't detect it. Might be worth logging the error to sentry or at least the console just in case.

usame-algan:
Another thing I noticed is that when creating a transaction from the sidebar, it will close the modal if the user switches their network and will automatically open it again if they switch to a safe on that network or switch the network back. This does not work for other modals e.g. sending a transaction from the assets table. Imo this is unintended behaviour and should be consistent across the app. We could fix it with #1225

src/hooks/coreSDK/useInitSafeCoreSDK.ts
```diff
@@ -1,18 +1,24 @@
...
   useEffect(() => {
-    if (!safeLoaded || !chain || !wallet || chain.chainId !== wallet.chainId) {
+    if (
+      !safeLoaded ||
+      !wallet?.provider ||
+      !wallet.chainId ||
+      chainId !== wallet.chainId ||
+      safe.chainId !== wallet.chainId
+    ) {
       return
```
katspaugh, line 22: Shouldn't we set the SDK to null or undefined here?

usame-algan: I agree it would make sense to do that. There is one case where it doesn't work in safe apps:

1. Open a safe app on network A e.g. Drain Account
2. Switch wallet network to B
3. Press Transfer everything
4. It shows `Your wallet is connected to the wrong chain.` which is correct
5. Switch wallet network to A
6. It shows `The Safe SDK could not be initialized. Please be aware that we only support v1.1.1 Safes and up.` which is not correct

It seems that the `getSafeSDK` call inside the modal happens sooner than the `initSafeSDK` call that we do in Step 5 so it returns `undefined` and fails.

One possible solution I can think of is to use the `useSafeSDK` hook and pass it in the `useAsync` call inside `ReviewSafeAppsTx` to create the tx but it would probably still flash the error once.

katspaugh: `getSafeSDK` should never be used in React components anyway, we should use a hook.

katspaugh: I guess there's no harm in keeping the SDK instance stale. The submit button is disabled. And it will be re-initialized when the user switches to the right chain.

usame-algan: Thinking a bit more about this we should make sure that it uses the re-initialized SDK in the case I described and not just "masks" the error.

src/hooks/coreSDK/useInitSafeCoreSDK.ts
iamacook, line 20: To summarise our discussion: if you pass the entire `safe` (`SafeInfo`) to `initSafeSDK`, you can get the `chainId`, `address` and `version` without needing to reference the wallet (which may be incorrect). These clauses are then no longer necessary. However, as agreed, I'll leave you to look into this further.

usame-algan: After thinking about this, to initialize the Core SDK in every case we need a wallet provider, safe address and safe version. This means it needs to be dependant on the wallet and safe info. I think we can remove the dependency on `chainId` that comes through the URL and instead just make sure that the wallet and safe info have the same chainId.

In that case we can simplify the method signature of `initSafeSDK` and just pass the `SafeInfo` to get all the values we need apart from the wallet provider.

I will add a few tests for `useInitSafeCoreSDK` first and then attempt the refactor.

katspaugh: I wouldn't use the entire SafeInfo object tho, it's updated every 15 seconds. It's better to use the individual fields you need.

iamacook (approved):

katspaugh (approved):
Looks good!

francovenica:
I follow the steps in the original issue:
Created a tx in goerli (no execution)
Go straight to a ETH safe (no change in network yet)
Open the Tx builder and create a dummy app
The modal doesn't estimate nonce (or gas or anything). An error asking for me to change the network
I changed the network but I still get an error GS031 and no estimation of nonce is produced.

image

francovenica:
Now i got an issue with the SDK not starting at all
image

src/hooks/coreSDK/useInitSafeCoreSDK.ts
usame-algan: > I guess there's no harm in keeping the SDK instance stale. The submit button is disabled. And it will be re-initialized when the user switches to the right chain.

There are 2 issues that @francovenica found:

1. Its possible that the modal displays a wrong nonce because even with a stale SDK instance, the createTx call succeeds if both safes have the same owner.
2. Its possible to open the transaction modal from within a safe app even if the wallet is on the wrong network. This is something we can't control.

I propose that we reset the SDK instance as discussed which would solve the first issue. For the second issue I can think of multiple solutions:

1. Use `useSafeSDK` and pass the instance to its txSender function. We can limit this to the safe app modal because everywhere else the user can't open the modal if they are on the wrong network
2. Add logic to the `AppFrame` to close the modal if there is no SDK instance. This has the downside that the user is still able to "click" the button but nothing would happen in the UI.

katspaugh: > Its possible to open the transaction modal from within a safe app even if the wallet is on the wrong network. This is something we can't control.

But they won't be able to submit it, right? Unless they switch the chain to the right one. At which point it should re-initialize the SDK.

usame-algan: Exactly, this is what currently happens but since the rendered component that is visible inside the modal doesn't have a dependency to the core SDK it will not rerender and re-run the logic inside the `useAsync` where it accesses the getter.

Since it is possible to reset the Core SDK from anywhere inside the app and at any time (via the wallet) we can't reliably use `getSafeSDK` but should instead always rely on `useSafeSDK` to detect changes. Right now I only implemented this for certain calls with a fallback to `getSafeSDK` but imo we should adjust it to always pass the SDK as a dependency to those functions inside `txSender`

francovenica:
Ok, it works now. The nonce is being properly estimated after following the steps descirbed:
Creating a tx in some network
Switching to a safe of a different network
Trying to create a tx from an app (drain app was used for this)
Checking that it shows the error of "wrong network" and doesn't estimate anything
Switching network in the wallet (MM extension wallet was used during this test)
See that the error goes away and the nonce for that safe is properly calculated

There is an issue that this new nonce estimation is taking a solid 10 secs or more, with no change in the UI to indicate that the user should wait. There is this glowing "gray bar" where the nonce is supposed to show, but since that was happening before the network change, for the user it looks like nothing is happening, and they might end up just cancelling everything and starting again.
Idk if it is possible to make this faster or not

app networks

src/components/safe-apps/SafeAppsSignMessageModal/ReviewSafeAppsSignMessage.tsx
```diff
@@ -64,12 +64,15 @@ const ReviewSafeAppsSignMessage = ({
...
-      operation: OperationType.DelegateCall,
-    })
+    return createTx(
+      {
+        to: signMessageAddress,
+        value: '0',
+        data: txData || '0x',
+        operation: OperationType.DelegateCall,
+      },
+      undefined,
```
katspaugh, line 74: Why do you need to pass undefined?

src/components/settings/SafeModules/RemoveModule/steps/ReviewRemoveModule.tsx
```diff
@@ -16,6 +16,7 @@ export const ReviewRemoveModule = ({
   data: RemoveModuleData
   onSubmit: (txId: string) => void
 }) => {
+  const { createRemoveModuleTx } = useTxSender()
   const [safeTx, safeTxError] = useAsync<SafeTransaction>(() => {
```
katspaugh, line 20: Shouldn't this be also `useAsync<SafeTransaction | undefined>`?

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -146,6 +141,7 @@ const SignOrExecuteForm = ({
     const id = txId || (await proposeTx(createdTx))
     const txOptions = getTxOptions(advancedParams, currentChain)
 
+    console.log('ASD')
```
katspaugh, line 144: Don't forget to remove it.

src/services/tx/txSender.ts
```diff
@@ -22,35 +22,34 @@ import type { Web3Provider } from '@ethersproject/providers'
...
+import { getSpendingLimitContract, getSpendingLimitModuleAddress } from '@/services/contracts/spendingLimitContracts'
 import EthersAdapter from '@gnosis.pm/safe-ethers-lib'
 import { Errors, logError } from '@/services/exceptions'
 import { EMPTY_DATA } from '@gnosis.pm/safe-core-sdk/dist/src/utils/constants'
+import { useCallback } from 'react'
+import type { SpendingLimitState } from '@/store/spendingLimitsSlice'
+import { createAddDelegateTx, createResetAllowanceTx, createSetAllowanceTx } from '@/services/tx/spendingLimitParams'
+import { parseUnits } from '@ethersproject/units'
+import { currentMinutes } from '@/utils/date'
+import type { NewSpendingLimitData } from '@/components/settings/SpendingLimits/NewSpendingLimit/steps/ReviewSpendingLimit'
```
katspaugh, line 34: Given that these dependencies are only for the spending limits, I would move the spending limit function into a separate module and re-export it here.

src/services/tx/txSender.ts
```diff
@@ -22,35 +22,34 @@ import type { Web3Provider } from '@ethersproject/providers'
...
 
-const getAndValidateSafeSDK = (): Safe => {
-  const safeSDK = getSafeSDK()
-  if (!safeSDK) {
-    throw new Error('The Safe SDK could not be initialized. Please be aware that we only support v1.1.1 Safes and up.')
-  }
-  return safeSDK
-}
+const SDK_NOT_FOUND_ERROR =
+  'The Safe SDK could not be initialized. Please be aware that we only support v1.1.1 Safes and up.'
```
katspaugh, line 37: Since you're touching this, please remove "Please be aware that we only support v1.1.1 Safes and up.".

src/services/tx/txSender.ts
```diff
@@ -95,358 +94,440 @@ const getRecommendedTxParams = async (
...
-
-  // If the nonce is not provided, we get the recommended one
-  if (nonce === undefined) {
-    const recParams = (await getRecommendedTxParams(txParams)) || {}
-    txParams = { ...txParams, ...recParams }
-  } else {
-    // Otherwise, we use the provided one
-    txParams = { ...txParams, nonce }
-  }
+export const useTxSender = () => {
```
katspaugh, line 97: His hook is _really_ verbose. I'm sure there's a more elegant way to wrap all those functions.

src/services/tx/txSender.ts
usame-algan: Should we already remove it or wait until after it is implemented in the core sdk?

src/services/tx/txSender.ts
usame-algan: I extracted the function into a new file which is also exported by the txSender.

katspaugh (commented):
The code looks good! 🎩 🪄 
Some tests are failing though.

src/hooks/coreSDK/useInitSafeCoreSDK.ts
```diff
@@ -1,29 +1,35 @@
...
-        setSafeSDK(undefined)
-        setError(e)
+        dispatch(
+          showNotification({
+            message: `The Safe SDK could not be initialized.`,
+            groupKey: 'core-sdk-init-error',
+            variant: 'error',
+          }),
+        )
+        trackError(ErrorCodes._105, (e as Error).message)
```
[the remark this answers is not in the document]
usame-algan: We are now always resetting the instance. This means that if the init throws it will just stay undefined so no need to reset it there anymore.

src/hooks/useTxSender.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const useTxSender = (): NullableTxSenderFunctions => {
+  const sdk = useSafeSDK()
+
+  // @ts-ignore
+  return useMemo(
+    () =>
+      Object.keys(txSender).reduce((result, key) => {
+        // @ts-ignore
```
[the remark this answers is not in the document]
usame-algan: I was able to remove one ts-ignore but the other one proves difficult because `txSender` functions expect various number of arguments so it doesn't allow just sprading `...args`. Any ideas?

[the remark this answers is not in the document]
iamacook: Perhaps this may work?

```ts
result[key] = sdk ? txSender[key] : undefined
```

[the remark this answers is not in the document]
usame-algan: Had to adjust it so its a function that returns undefined but this way it works, thanks!

katspaugh (approved):
Looks good!

usame-algan:

usame-algan:
recheck

usame-algan:

katspaugh:

francovenica:
It works fine for me
I've checked again the nonce issue by following the steps described in the previous comments. It works fine

Also I did a regression regarding txs creation and execution With 1/x and 2/x policies.
Send funds
Send NFT's 
Tx builder using a dummy contract
Ower adding/replacing/removing
Change in policies
Spending limit creation, edition, removal
Use of spending limits
Enabling disabling modules

Issue #1261, opened 2022-11-28
Don't propose unsigned transactions


katspaugh:
In a 1/X Safe, it's possible to create and immediately execute a transaction w/o signing it.

We're currently "proposing" such txns to the backend, where they are marked as untrusted until indexed.

@Uxio0 says we shouldn't do this. The ability to "propose" w/o signatures will be removed in the future on the backend.

iamacook:
We need to ensure this works well with smart contract wallet owners. Currently, if a Safe owns a Safe ("parent" vs. "nested"), the "parent" can queue the approval of a transaction of the "nested" Safe. As such, the "nested" Safe currently proposes a transaction without signatures.

Issue #1220, opened 2022-11-22
Reference transaction `trusted` flag when showing the `UnsignedWarning`


iamacook:
## What is the feature about

The `UnsignedWarning` is currently shown when no signatures are present. However, the gateway will soon forward the `trusted` flag of each transaction as well, which should also be referenced.

## The list of requirements

The conditional rendering logic needs to be extend to include the `trusted` flag as follows:

```ts
{!trusted && <UnsignedWarning />}
```

## Links
- [Gateway issue](https://github.com/safe-global/safe-client-gateway/issues/1005)

katspaugh:
Shouldn't we use _only_ the trusted flag?
@Uxio0 it's always `trusted=true` for signed transactions, right?

Uxio0:
Yes, `trusted` flag is all you need. Transactions signed but at least one owner are `trusted`. Even if you remove the owner, transaction will still be `trusted`

katspaugh:
@francovenica the trusted flag is now on staging backend, please test the following cases:

* A delegate-created tx should NOT be marked as untrusted
* MMI txns should NOT be untrusted
* Regular transactions proposed w/o a signature SHOULD be untrusted (ask a dev to make you one)

iamacook:
>     * Regular transactions proposed w/o a signature SHOULD be untrusted – you can create them on this branch https://drafts.web-core.pages.dev – just uncheck the Sign checkbox in the tx modal. But view them via a deep link on dev.

I just created an unsigned transaction on staging here for you to confirm: `gor:0x65c57CC1317a1f728E921Cbf7bC08b3894196D29/transactions/tx?id=multisig_0x65c57CC1317a1f728E921Cbf7bC08b3894196D29_0xec49a9f9eee0446834c1c4ee00ca1b366f604d5e09555c34c9cd9a29c508757f`

![image](https://user-images.githubusercontent.com/20442784/203928791-09e582f2-ee98-4dca-a6db-b3d376470102.png)

JagoFigueroa:
Adding this scenario affecting safe apps:

https://user-images.githubusercontent.com/92332710/204233045-86ac699d-b277-4a82-ba96-05557874f42b.mp4

When the pop up to view the transaction is clicked, txs are showing as untrusted for me until the transaction is indexed. Then it will lose this status.

katspaugh:
@JagoFigueroa fixed in #1261. Immediately executed transactions in 1/x Safes won't be visible until executed.

PR #1262, opened 2022-11-28
Fix: comment out the untrusted badge

katspaugh:
Temporarily comment out the Untrusted tx badge from #1220.

To be restored after #1261 is closed.

### DIFF
--- a/src/components/create-safe/InfoWidget/styles.module.css
+++ b/src/components/create-safe/InfoWidget/styles.module.css
@@ -3,7 +3,7 @@
 }
 
 .title {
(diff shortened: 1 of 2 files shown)


src/components/create-safe/InfoWidget/styles.module.css
```diff
@@ -3,7 +3,7 @@
 }
 
 .title {
-  width:fit-content;
```
katspaugh, line 6: Prettier fixed this.

iamacook (approved):

PR #1265, opened 2022-11-28
Fix format amount review modal

### LINKED ISSUE #705: Add thousand separators to review amount
## Overview
Large amounts on the review modal can be confusing. As such, we should add thousand separators.

## Requirements
Format token amount in review modal header.

## Screenshot
image
image

DiogoSoaress:
## What it solves

Resolves #705

## How this PR fixes it
Implements a new formatter `formatAmountPrecise` to format a number with a passed precision without adhering to our style guide notation.
Uses the app `formatAmountPrecise` formatter for review modal header's amounts.

## How to test it
1. Click to create a transaction with thousands and decimals
2. Observe the review modal displays the amount formatted as per locale

## Analytics changes
N/A

## Screenshots
Screenshot 2022-12-07 at 15 49 12

### DIFF
--- a/src/components/tx/modals/TokenTransferModal/ReviewTokenTx.tsx
+++ b/src/components/tx/modals/TokenTransferModal/ReviewTokenTx.tsx
@@ -8,7 +8,7 @@ import { SendTxType } from '@/components/tx/modals/TokenTransferModal/SendAssets
 import TokenIcon from '@/components/common/TokenIcon'
 import ReviewSpendingLimitTx from '@/components/tx/modals/TokenTransferModal/ReviewSpendingLimitTx'
 import ReviewMultisigTx from '@/components/tx/modals/TokenTransferModal/ReviewMultisigTx'
-import { localeNumberFormatter } from '@/utils/formatNumber'
+import { formatAmount } from '@/utils/formatNumber'
 
 export const TokenTransferReview = ({
   amount,
@@ -19,8 +19,6 @@ export const TokenTransferReview = ({
   tokenInfo: TokenInfo
   children?: ReactNode
 }) => {
-  const localeAmount = localeNumberFormatter(amount)
-
   return (
     <Box className={css.tokenPreview}>
       <Box className={css.tokenIcon}>
@@ -29,7 +27,7 @@ export const TokenTransferReview = ({
 
       <Box mt={1} fontSize={20}>
         {children}
-        {localeAmount} {tokenInfo.symbol}
+        {formatAmount(amount)} {tokenInfo.symbol}
       </Box>
     </Box>
   )
--- a/src/components/tx/modals/TokenTransferModal/ReviewTokenTx.tsx
+++ b/src/components/tx/modals/TokenTransferModal/ReviewTokenTx.tsx
@@ -8,6 +8,7 @@ import { SendTxType } from '@/components/tx/modals/TokenTransferModal/SendAssets
 import TokenIcon from '@/components/common/TokenIcon'
 import ReviewSpendingLimitTx from '@/components/tx/modals/TokenTransferModal/ReviewSpendingLimitTx'
 import ReviewMultisigTx from '@/components/tx/modals/TokenTransferModal/ReviewMultisigTx'
+import { localeNumberFormatter } from '@/utils/formatNumber'
 
 export const TokenTransferReview = ({
   amount,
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/formatNumber.test.ts
```diff
@@ -268,4 +268,24 @@ describe('formatNumber', () => {
       expect(formatCurrency(amount4, 'BHD')).toBe('< -0.001 BHD')
     })
   })
+
+  describe('localeNumberFormatter', () => {
```
iamacook, line 272: We should add international tests.

src/utils/formatNumber.ts
```diff
@@ -193,3 +193,22 @@ export const formatCurrency = (number: string | number, currency: string): strin
 
   return format(number, currencyFormatter, minimum)
 }
+
+export const localeNumberFormatter = (amount: number | string): string => {
+  // get locale decimal separator
+  const number = 123456.789
```
iamacook, line 199: Using the arbitrary value confused me at first. I would suggest using [`Intl.NumberFormat.prototype.formatToParts`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts) instead as you can directly get the separators from `amount` directly.

src/utils/formatNumber.ts
```diff
@@ -193,3 +193,22 @@ export const formatCurrency = (number: string | number, currency: string): strin
...
+  const [thousandSeparator, decimalSeparator] = localeNumber.match(/[^0-9]/g) || []
+
+  const [integerString, decimalString] = amount.toString().split(decimalSeparator)
+
+  const decimalFinal = decimalString ? `${decimalSeparator}${decimalString}` : ''
+
+  // trailing zeros
+  const decimalFinalWithoutTrailingZeros = decimalString ? decimalFinal.replace(/0+$/, '') : ''
+
+  const integerFinal = integerString.replace(/\B(?=(\d{3})+(?!\d))/g, thousandSeparator)
```
iamacook, line 211: Are thousand separators universal/set every three digits in all number formats?

katspaugh: Definitely not: https://learn.microsoft.com/en-us/globalization/locale/number-formatting
Hindi – two digits, Japanese four.

DiogoSoaress: Good point here, the regex was very opinionated. I'm using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString now to format amount's integer part.

katspaugh (approved):

katspaugh:
Looks good!
Screenshot 2022-12-07 at 16 02 33

Noticed a different bug when testing this. The same amount in the select and input below is formatted differently:

Screenshot 2022-12-07 at 16 02 15

DiogoSoaress:
Using our utils formatter does not give a very precise view to the amount of tokens to be transferred so I've created a new ticket for this enhancement https://github.com/safe-global/web-core/issues/1321

src/components/tx/modals/TokenTransferModal/ReviewTokenTx.tsx
```diff
@@ -26,7 +27,7 @@ export const TokenTransferReview = ({
 
       <Box mt={1} fontSize={20}>
         {children}
-        {amount} {tokenInfo.symbol}
+        {formatAmount(amount)} {tokenInfo.symbol}
```
iamacook, line 30: I wouldn't use the current formatter for the reasons you stated [here](https://github.com/safe-global/web-core/pull/1265#issuecomment-1341097212). This is arguably less clear than non-formatted numbers.

DiogoSoaress: I won't merge this PR until we discuss it tomorrow

katspaugh: I tend to agree it might be worse than no formatting in some (rather extreme?) cases.

katspaugh: On the other hand, this is how we format numbers in other places in the app, like it or not. I don’t see why this modal has to be different than tx list, balances etc.

iamacook: This is the one place that I would argue we always display the exact precision though as assets have aribtrary values. If I were sending 100,400, I would much sooner "review" that than 100K. The same with [decimal truncation](https://github.com/5afe/safe/wiki/How-to-format-amounts).

katspaugh (approved):
👍

iamacook (approved):

PR #1280, opened 2022-11-29
fix: prevent scroll adjusting number inputs

### LINKED ISSUE #1234: Prevent scroll from altering `type="number"` fields
## Bug description

Scrolling on a trackpad or mousewheel increases/decreases the value of `type="number"` fields. This should not happen in order to prevent unexpected nonces, etc.

Note: this is relevant for *all* `type="number"` fields across the app.

Mui docs on type number fields: https://mui.com/material-ui/react-text-field/#type-quot-number-quot

## Steps to reproduce
 1. Open the transaction creation modal and edit the nonce value.
 2. Scroll on the trackpad or mousewheel inside the field.

## Expected result

The field values do not change.

## Obtained result

The field values change.

## Screenshots

scrollnumber

iamacook:
## What it solves

Resolves #1234

## How this PR fixes it

A new `NumberField` component has been created (isolating the common `type` and `autoComplete` attributes). It blurs the field `onWheel`.

A `ref` is used as `TextField` doesn't work with `currentTarget`.

## How to test it

Ensure the following fields work as expected, with validation and preventing scroll from editing the value:

- Transaction filter: amount, nonce
- Spending limit amount
- Send asset amount
- Advanced transaction parameters: nonce, safeTxGas, wallet nonce, gas limit, maxPriorityFeePerGas, maxFeePerGas


usame-algan:
How does this compare to the alternative that is suggested in the [MUI docs](https://mui.com/material-ui/react-text-field/#type-quot-number-quot) i.e. not using `type="number"` but instead using a pattern argument?

iamacook:
> How does this compare to the alternative that is suggested in the [MUI docs](https://mui.com/material-ui/react-text-field/#type-quot-number-quot) i.e. not using `type="number"` but instead using a pattern argument?

It blurred the input on scroll.

The GOV.UK Design System team make a very [compelling argument](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/) for the proposed implementation. As it's the suggested implementation by MUI as well, I think we should use it.

usame-algan:
> The GOV.UK Design System team make a very [compelling argument](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/) for the proposed implementation. As it's the suggested implementation by MUI as well, I think we should use it.

One issue I noticed with it is that the validation and pattern is not aligned anymore e.g. inputs like `+12` and `1e2` are allowed by our validation logic but not by the pattern. This leads to the UI not showing an error but the user can't submit the form. Afaiu the inputmode numeric is specifically for digits 0-9 so changing the pattern would probably not be wise.

Also, should we use the new field for the Amount input as well?

iamacook:
> One issue I noticed with it is that the validation and pattern is not aligned anymore e.g. inputs like `+12` and `1e2` are allowed by our validation logic but not by the pattern. This leads to the UI not showing an error but the user can't submit the form. Afaiu the inputmode numeric is specifically for digits 0-9 so changing the pattern would probably not be wise.

As this issue is purely related to scroll prevention, I would suggest we revert to [my initial implementation that blurred the input on scroll input](https://github.com/safe-global/web-core/blob/scroll-number-inputs/src/components/inputs/NumberField/index.tsx) and research then this further. What do you think?

usame-algan (approved):

PR #1289, opened 2022-11-30
[PoC] feat: return signature for 1/n threshold Safes

iamacook:
## What it solves

Proof of concept for immediately returning signatures for https://github.com/safe-global/web-core/issues/1206

## How this PR fixes it

If a signing is requested, off-chain is chosen when the Safe threshold is 1/n and the signature is immediately returned.

This builds upon proposed `safe-apps-sdk` changes [here](https://github.com/safe-global/safe-apps-sdk/pull/411).

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@gnosis.pm/safe-deployments": "^1.15.0",
     "@gnosis.pm/safe-ethers-lib": "^1.6.1",
     "@gnosis.pm/safe-modules-deployments": "^1.0.0",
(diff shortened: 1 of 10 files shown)


iamacook:
After [syncing with Snapshot](https://www.notion.so/gnosis-safe/Snapshot-sync-03a3179fedd64408b233fef5aa5c2be7), we will initially integrate the toggling of off-chain signatures via a custom RPC call. As such, I will close this but we can reopen it/use it as reference if we want to integrate returning the signatures immediately for 1/n Safes.

PR #1285, opened 2022-11-29
refactor: use `isValidTransaction` from SDK

iamacook:
## What it solves

Resolves #1238

## How this PR fixes it

`useIsValidExecution` has been refactored to use `isValidTransaction` from `safe-core-sdk` instead.

Duplicate tests have been removed as the library has full test coverage [here](https://github.com/safe-global/safe-core-sdk/blob/main/packages/safe-core-sdk/tests/execution.test.ts#L37-L101).

## How to test it

Attempt to execute an invalid transaction and observe the "This transaction will most likely fail. To save gas costs, reject this transaction." error.

## Screenshots

image

PR #1293, opened 2022-11-30
fix: Handle timeout error in safe creation

### LINKED ISSUE #1291: There is no "Cancel", "retry" button when Code 800: Safe creation failed appears in console

usame-algan:
## What it solves

Resolves #1291 

## How this PR fixes it

- Timeout errors thrown by ethers do not contain a `receipt` so we should handle it as an optional type

### Other changes

- Removes `receipt` from `TxEvent.PROCESSED` and `TxEvent.REVERTED` as it was not used

## How to test it

1. Create a new safe
2. Set the gas fee too low in your wallet and submit
3. Wait for 6.5 minutes OR reload the page and wait for 6.5 minutes
4. Observe a timeout error
5. Observe that the Cancel and Retry button is visible

## Screenshots
Screenshot 2022-11-30 at 18 19 13

### DIFF
--- a/src/utils/ethers-utils.ts
+++ b/src/utils/ethers-utils.ts
@@ -8,6 +8,7 @@ export enum EthersTxReplacedReason {
   replaced = 'replaced',
 }
 
+// TODO: Replace this with ethers v6 types once released
 export type EthersError = Error & { code: ErrorCode; reason?: EthersTxReplacedReason; receipt?: TransactionReceipt }
 
 export const didRevert = (receipt: EthersError['receipt']): boolean => {


src/hooks/useIsValidExecution.ts
```diff
@@ -11,7 +11,7 @@ import ContractErrorCodes from '@/services/contracts/ContractErrorCodes'
 import { sameAddress } from '@/utils/addresses'
 
 const isContractError = <T extends EthersError>(error: T): error is T & { reason: keyof typeof ContractErrorCodes } => {
-  return Object.keys(ContractErrorCodes).includes(error.reason)
+  return Object.keys(ContractErrorCodes).includes(error.reason!)
```
usame-algan, line 14: I didn't bother with this as it will be removed with #1285

iamacook: I would suggest something along the lines of:

```ts
  const isEthersError = <T extends Error | EthersError>(error: T): error is EthersError
```

iamacook (commented):
Appears to work well but there are failing tests.

src/hooks/useIsValidExecution.ts
usame-algan: Why does the function need to handle `Error` and `EthersError` if it is only used in one place? I think we should rework `EthersError` in general. Ethers v6 will add [proper error types](https://github.com/ethers-io/ethers.js/issues/3565#issuecomment-1332668704) but for the time being I would suggest opting for a more simple solution e.g.

```
const isContractError = (error: EthersError) => {
  if (!error.reason) return false

  return Object.keys(ContractErrorCodes).includes(error.reason)
}
```

iamacook: The reasoning behind the union was because a standard error could be caught. Let's keep your solution and adjust the type as suggested in the future.

iamacook (approved):

francovenica:
The error is there.
image

PR #1298, opened 2022-12-01
Fix: don't propose txs w/o signatures

### LINKED ISSUE #1261: Don't propose unsigned transactions

katspaugh:
## What it solves

Resolves #1261

## How this PR fixes it
If a transaction is created and immediately executed w/o signatures, don't propose it to the backend.

In this case, we don't show a "View transaction" link in the tx notifications, but the notifications are still grouped by the safeTxHash.

When a smart contract wallet creates a tx, it will still propose a transaction w/o initial signatures.

## How to test it
Case 1:
* Create and immediately execute a tx in a 1/X Safe
* Observe that the tx notifications change each other as the tx is being processed, but there's no deep link in the notifications.
* Once indexed, the tx will appear in the history

Case 2:
* Create and sign a tx
* Execute that tx
* Notifications with a "View transaction" link should appear

Case 3:
* Connect a smart contract wallet, e.g. Safe or Argent
* Create a tx and sign it
* The tx will appear in the queue immediately but with 0 sigs
* Signature will be added when it's indexed

## Analytics changes
None


usame-algan (approved):
Looks good 👍

francovenica:
1 - When I sign and execute a tx with a safe as an owner in a 1/x safe, I get only 1 of the 3 notifications that you would get with a regular user
image

2 - SafeA owner of SafeB. Create and execute a send funds tx in SafeB, a popup in safeA shows so the owner of SafeA can sign. Once the tx in SafeA is successfully executed the modal in both safes remain there instead of closing.
I tried this in Prod and there the modal close in both safes after the tx is done.

Steps:
SafeA owns SafeB
Create and execute a send funds tx in SafeB. Lets say the nonce is 10
In SafeA the popUp to sign and execute the tx shows up
Execute the tx in SafeA with its owner.
Wait for the tx to be done executing in SafeA

Result:
SafeA still shows the modal to execute the tx
SafeB still shows the modal to execute the tx and the nonce was increased to 11
Closing the modal in safeA shows a notification in SafeB that the tx was declined

katspaugh:
> When I sign and execute a tx with a safe as an owner in a 1/x safe, I get only 1 of the 3 notifications that you would get with a regular user

That's the correct new behavior.

> the modal in both safes remain there instead of closing

I'll look into this, thanks Franco!

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -122,11 +122,14 @@ const SignOrExecuteForm = ({
...
-  const onSign = async (): Promise<string> => {
+  const onSign = async (): Promise<string | undefined> => {
     const [connectedWallet, createdTx, provider] = assertDependencies()
 
     // Smart contract wallets must sign via an on-chain tx
     if (await isSmartContractWallet(connectedWallet)) {
+      // If the first signature is a smart contract wallet, we have to propose w/o signatures
+      // Otherwise the backend won't pick up the tx
+      // The signature will be added once the on-chain signature is indexed
       const id = txId || (await proposeTx(createdTx))
```
katspaugh, line 133: @Uxio0 FYI, as a temporary solution, we'll keep proposing txs w/o sigs when a smart contract wallet initiates a new transaction.
As per @rmeissner's comment:
Screenshot 2022-12-08 at 12 37 56

If you have a better idea, please let us know!

katspaugh:
@usame-algan, one unit test is still failing (`SignOrExecuteForm -> displays an error if execution submission fails`), but you can review it anyway, if you like.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -155,16 +158,15 @@ const SignOrExecuteForm = ({
   }
 
   // Execute transaction
-  const onExecute = async (): Promise<string> => {
+  const onExecute = async (): Promise<string | undefined> => {
     const [, createdTx, provider] = assertDependencies()
 
     // If no txId was provided, it's an immediate execution of a new tx
-    const id = txId || (await proposeTx(createdTx))
```
usame-algan, line 162: I believe we also still need to propose when executing with a smart contract wallet e.g. a 1/n safe has another safe as owner

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -266,6 +270,8 @@ export const dispatchSpendingLimitTxExecution = async (
   return result?.hash
 }
 
-export const dispatchSafeAppsTx = (txId: string, safeAppRequestId: RequestId) => {
-  txDispatch(TxEvent.SAFE_APPS_REQUEST, { txId, safeAppRequestId })
+export const dispatchSafeAppsTx = async (safeTx: SafeTransaction, safeAppRequestId: RequestId) => {
+  const sdk = getAndValidateSafeSDK()
+  const safeTxHash = await sdk.getTransactionHash(safeTx)
```
iamacook, line 275: These can both throw.

src/hooks/useTxNotifications.ts
```diff
@@ -23,7 +25,10 @@ const TxNotifications = {
   [TxEvent.PROCESSING]: 'Your transaction is being processed.',
   [TxEvent.PROCESSING_MODULE]:
     'Your transaction has been submitted and will appear in the interface only after it has been successfully processed and indexed.',
-  [TxEvent.AWAITING_ON_CHAIN_SIGNATURE]: 'An on-chain signature request was submitted.',
+  [TxEvent.ONCHAIN_SIGNATURE_REQUESTED]:
+    'An on-chain signature is required. Please confirm the transaction in your wallet.',
+  [TxEvent.ONCHAIN_SIGNATURE_SUCCESS]:
+    'On-chain signature was successfully submitted. Please wait for the signature to be indexed.',
```
iamacook, line 31: As the "unchecked" SDK doesn't require wallet approval, there is not necessarily an immediate signature, e.g. it is queued in a Safe.

src/components/tx/SignOrExecuteForm/index.tsx
iamacook: I think this is fine as it's regarding execution. Final confirmations/immediate executions eventually wind up on chain where they will be indexed by our services.

usame-algan: Yes, that still works but its now missing notifications on the safe so it never displays that the transaction was successfully executed. I don't see a reason to change this behaviour for execution but not for signing with a smart contract wallet.

katspaugh: The reason is described in the issue. On-chain signatures just happen not to work at all when creating a new transaction. An alternative that I'm considering would be to disable tx creation for SC wallets.

src/services/tx/tx-sender/dispatch.ts
katspaugh: Good catch (pun not intended).

src/hooks/useTxNotifications.ts
katspaugh: I'm aware, but you would still have to wait for it to be indexed, right?
What message would you suggest?

iamacook: Only if the wallet approves it immediately. I would just remove the second sentence.

katspaugh: I mean, the original Safe (not the owner) needs to index the approveHash transaction. Immediate or after a week doesn't matter.

iamacook: Sure, but the transaction may not be immediately executed.

src/components/tx/SignOrExecuteForm/index.tsx
usame-algan: The difference in behaviour I mean is that when I sign the transaction in Safe A and execute the confirmation immediately in Safe B, it shows notifications in Safe A saying "An on-chain signature is required. Please confirm the transaction in your wallet.". This one is missing when immediately executing the transaction in Safe A even though it also has to wait for the other transaction to be executed in Safe B.

Another case is that if both transactions are immediately executes, previously it would show "Your transaction was successfully executed." in Safe A but now it doesn't which seems like a regression to me.

katspaugh: "Your transaction was successfully executed." isn't shown now because it comes from the backend. Once a tx lands in the history, we show this notification, but only if there's a pending tx with this id. Without proposing, there's no id.
So it's not a regression, it's a logical consequence of not proposing on creation+immediate execution. And it's the same for SC wallets and regular wallets.

src/services/tx/tx-sender/dispatch.ts
katspaugh: Wrapped the parent function in a try-catch.

src/hooks/useTxNotifications.ts
katspaugh: Updated the text as per our discussion.

src/components/tx/SignOrExecuteForm/index.tsx
usame-algan: My argument ist that whether it is an immediate execution or just signing, Safe A will always have to wait for another transaction to be executed in Safe B. This could happen immediately (if Safe B also immediately executes) or could take days/weeks or longer. Why do we propose in one of the cases but not the other?

katspaugh: We propose only for SC wallets because if we don't, the backend cannot match the tx and the sig from approveHash.

src/components/tx/SignOrExecuteForm/index.tsx
Uxio0: No worries, as discussed on the weekly, we can go for now with allowing not signature transactions

usame-algan (approved):

src/components/tx/SignOrExecuteForm/index.tsx
usame-algan: Ok, that makes sense. So in case of an immediate execution, the transaction in Safe B contains the execution of Safe A so its all in one place and doesn't need a reference like the sign tx does.

iamacook (approved):

katspaugh:
@francovenica to recap for you, we left smart contract wallets as they were, so only immediate execution is affected.

francovenica:
I wanted to check safe as an owner signing again, but the WC app is not working in this PR. 
image

katspaugh:
@francovenica it doesn't work on dev either. Please use the prod CGW as a workaround.

francovenica:
It wasn't working for me even with the CGW enabled, but now it does 🤷 

I tested the same scenario that I reported in the previous comment. The modal are closing just fine now and the tx is being executed successfully. Tested with sing&execute as one action

PR #1300, opened 2022-12-01
fix: replace nonce 0, flicker, no spending limit + disable duplite rejections

iamacook:
## What it solves

Resolves isses found [here](https://github.com/safe-global/web-core/pull/1079#issuecomment-1333132981).

## How this PR fixes it

1. Spending limit transactions are not shown creating a transaction via the replacement flow.
2. The backdrop no longer flickers whent the autocomplete dropdown is open.
3. Transactions with nonce 0 now open the replacement modal.
4. Disable duplicate rejection transactions.

## How to test it

1. Attempt to replace a transaction. Observe that spending limit options are not displayed.
2. Open the nonce autocomplete limit and observe no backdrop flicker.
3. Attempt to replace a transaction with nonce 0. Clicking "Replace transaction" in the modal should open the next step.
4. The replacement button is now always shown. Attempting to replace a transaction that already has a rejection transaction will render a disabled rejeection button in the modal.

### DIFF
--- a/src/components/tx/NonceForm/index.tsx
+++ b/src/components/tx/NonceForm/index.tsx
@@ -1,7 +1,7 @@
 import { memo, useMemo, useState } from 'react'
 import type { ReactElement } from 'react'
 import { useController, useFormContext, useWatch } from 'react-hook-form'
-import { Autocomplete, Backdrop, IconButton, InputAdornment, MenuItem, Popper, TextField, Tooltip } from '@mui/material'
+import { Autocomplete, Backdrop, IconButton, InputAdornment, MenuItem, TextField, Tooltip } from '@mui/material'
 import RotateLeftIcon from '@mui/icons-material/RotateLeft'
 import useSafeInfo from '@/hooks/useSafeInfo'
 import useTxQueue, { useQueuedTxByNonce } from '@/hooks/useTxQueue'
@@ -39,10 +39,7 @@ const NonceFormOption = memo(({ nonce, ...props }: { nonce: number } & MenuItemP
 const NonceForm = ({ name, nonce, recommendedNonce, readonly }: NonceFormProps): ReactElement => {
   const { safe } = useSafeInfo()
   const safeNonce = safe.nonce || 0
-  const [popper, setPopper] = useState<{ open: boolean; anchorEl: (EventTarget & Element) | null }>({
-    open: false,
-    anchorEl: null,
-  })
+  const [backdrop, setBackdrop] = useState(false)
 
   // Initialise form field
   const { setValue, control } = useFormContext() || {}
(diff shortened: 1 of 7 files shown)


src/components/tx/NonceForm/index.tsx
```diff
@@ -98,61 +95,59 @@ const NonceForm = ({ name, nonce, recommendedNonce, readonly }: NonceFormProps):
...
-          }}
-          InputLabelProps={{
-            ...params.InputLabelProps,
-            shrink: true,
-          }}
-        />
-      )}
-    />
+    <>
+      <Backdrop open={backdrop} />
```
katspaugh, line 99: An interesting workaround, but have you figured out why the flicker happens specifically with this dropdown?

francovenica:
The three issues fixed in this ticket are correct
Spending limit is not available anymore for replacement tx
Tx with nonce 0 can be rejected/replaced
The flicker effect doesn't happen anymore.

Issue: Now that I can replace tx with nonce 0 I noticed this:

In adv options and having only the tx with nonce 0 available for replacement, if you click on the nonce field to see the list of nonces you can replace it shows no list. It would be expected to show the list with only the nonce 0 tx.
0 nonce

How it is expected to behave:
image

src/components/tx/NonceForm/index.tsx
iamacook: It's related to the gas price polling [here](https://github.com/safe-global/web-core/blob/dev/src/hooks/useGasPrice.ts#L64-L78). The `Backdrop` seems to be very tempermental between renders.

schmanu (approved):
I'm still puzzled by Backdrop rendering but the workaround works :D

iamacook:
> Now that I can replace tx with nonce 0 I noticed this:

This should now work:

image

I've also included your findings from the epic [here](https://github.com/safe-global/web-core/pull/1079#issuecomment-1334841934) as the PR is still open.

> Once you create a rejection and Queue it, the "replace" button is gone, even tho it should be reasonalble to replace even the rejection. (probably the replace button is just following the logic of the old reject button)
What shouldn't be available anymore is the "Reject button" in the "Replace tx modal". The button could be grayed out the tooltip saying "this nonce already has a rejection proposed"

image

src/store/safeInfoSlice.ts
```diff
@@ -16,6 +16,7 @@ export const defaultSafeInfo: SafeInfo = {
   collectiblesTag: '',
   txQueuedTag: '',
   txHistoryTag: '',
+  messagesTag: '',
```
iamacook, line 19: This fixes a type issue from updating our packages.

src/components/tx/NonceForm/index.tsx
katspaugh: Sounds like the polling re-renders more than necessary in that form. I would take another look at that.

katspaugh: I deleted the backdrop entirely. It now looks the same as the AddressBookInput. Just a dropdown.

iamacook: Without the `Backdrop`, the options blur into the background though. It was suggested [here](https://github.com/safe-global/web-core/pull/1079#issuecomment-1327121065):

image

cc @usame-algan

PR #1302, opened 2022-12-01
fix: dont overlay banner over header

### LINKED ISSUE #1202: Banner is displayed over the app header
## Bug description
## Environment
 - Browser: any
 - Wallet: any
 - Chain: any

## Steps to reproduce

## Expected result

## Screenshots
image

usame-algan:
## What it solves

Resolves #1202 

## How this PR fixes it

- Gives the banner a relative position so it doesn't overlay the header
- Changes the header and sidebar to have a sticky position instead of fixed
- As a result the sidebar drawers don't have a fixed offset anymore but are aligned to the top of the window

## How to test it

1. Open the safe with the banner open (clear local storage if needed)
2. Observe that the header is visible
3. Scroll down
4. Observe that the banner disappears but the header and sidebar are fixed
5. Close the banner
6. Observe the that header and sidebar still have a fixed position
7. Open the safe list sidebar
8. Observe that it overlays everything else and is positioned at the top of the window
9. Switch to a smaller screen
10. Open the sidebar and safe list sidebar
11. Observe that they overlay everything else and are positioned at the top of the window

## Screenshots

Screenshot 2022-12-01 at 20 22 26
Screenshot 2022-12-01 at 20 22 36
Screenshot 2022-12-01 at 20 22 47
Screenshot 2022-12-01 at 20 23 15

### DIFF
--- a/src/components/common/PageLayout/SideDrawer.tsx
+++ b/src/components/common/PageLayout/SideDrawer.tsx
@@ -43,12 +43,13 @@ const SideDrawer = ({ isOpen, onToggle }: SideDrawerProps): ReactElement => {
   return (
     <>
       <Drawer
+        className={css.drawer}
         variant={isSmallScreen ? 'temporary' : 'persistent'}
         anchor="left"
         open={isOpen}
         onClose={() => onToggle(false)}
       >
-        <aside>
+        <aside className={css.sidebar}>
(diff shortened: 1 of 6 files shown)


katspaugh:
An alternative solution: make the banner part of the header.

Issue #1303, opened 2022-12-02
Open accordions with a smooth transition


usame-algan:
## What is the feature about

Open accordions with a smooth transition. This was tried in #1064 but resulted in too many network requests on page load. Instead we could change the logic to only fetch when the accordion is expanded.

## The list of requirements

- Open accordions with a smooth animation
- Only fetch their data once they are expanded

francovenica:
@usame-algan The multisig call everytime you open a tx:
![multisig call](https://user-images.githubusercontent.com/4503659/205947131-f06d2c09-52bc-40a0-9231-54ee54ac9295.gif)

PR #1064, opened 2022-11-04
fix: Keep accordion details mounted for transition

usame-algan:
For some reason we had the `unmountOnExit` prop set for the accordions which made the animation look choppy.

Before:

https://user-images.githubusercontent.com/5880855/200006728-112dabed-4663-4dbb-bb0f-1648907e4a18.mov

After:

https://user-images.githubusercontent.com/5880855/200006743-fe18f28c-eeaf-40df-b1cd-afb1e331725e.mov

PR #1304, opened 2022-12-02
fix: Revert accordion transitions

usame-algan:
## What it solves

Reverts #1064 

Also added an enhancement issue to look at this again in the future #1303

## How this PR fixes it

- Reverts changes done in #1064 that resulted in too many network requests on page load

## How to test it

- Open the Transaction history/queue
- Observe that there are no network requests for tx details
- Open a transaction accordion
- Observe that there is one network request for its details


iamacook (approved):

francovenica:
The revert looks fine. Just loading the tx list doesn't make a multisig call for every tx anymore

Note: I've noticed that a multisig call is done every time you open a tx details, even if you opened it before. We are going to fix that in this ticket #1303 
multisig call

PR #1307, opened 2022-12-05
Fix: use a non-tracking version of the AppFrame

katspaugh:
## What it solves
The governance widget on the Dashboard doesn't need to track the Safe App usage. And it doesn't need most of the communicator methods. Only the SafeInfo.

So I've extracted the iframe into a separate component and initialized a separate instance of the communicator that only provides SafeInfo.


src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -36,23 +36,18 @@ import useSignMessageModal from '../SignMessageModal/useSignMessageModal'
...
   allowedFeaturesList: string
-  isWidget?: boolean
 }
 
-// see sandbox mdn docs for more details https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox
-const IFRAME_SANDBOX_ALLOWED_FEATURES =
-  'allow-scripts allow-same-origin allow-popups allow-popups-to-escape-sandbox allow-forms allow-downloads allow-orientation-lock'
-
-const AppFrame = ({ appUrl, allowedFeaturesList, isWidget = false }: AppFrameProps): ReactElement => {
+const AppFrameWrapper = ({ appUrl, allowedFeaturesList }: AppFrameWrapperProps): ReactElement => {
```
yagopv, line 50: I think this component can continue being `AppFrame` as the folder name and the extracted iframe component itself could be `Iframe` with capital i. This components is more than a simple wrapper.

Now we have this `index.tsx` containing `AppFrameWrapper` and the `iframe.tsx` containing `AppIframe`, then this latest one is imported as `AppFrame` or `AppIframe` depending on where so for me is a little confusing.

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
```diff
@@ -30,10 +37,26 @@ const GovernanceSection = () => {
...
+  // Initialize the app communicator
+  useAppCommunicator(iframeRef, claimingApp, chain, {
+    onGetSafeInfo: () => ({
+      safeAddress,
+      chainId: chain ? parseInt(chain.chainId, 10) : -1,
+      owners: safe.owners.map((owner) => owner.value),
+      threshold: safe.threshold,
+      isReadOnly: !granted,
+      network: chain ? getLegacyChainName(chain.chainName || '', chain.chainId).toUpperCase() : '',
+    }),
```
iamacook, line 57: Perhaps we can extract the handlers so we can reuse it here?

iamacook (commented):
The `communicator` will still track calls to `onGetSafeInfo` [here](https://github.com/safe-global/web-core/blob/dev/src/components/safe-apps/AppFrame/useAppCommunicator.ts#L82-L90).

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
yagopv: yes but this is like an instance of the climing app isn't it? so is a sdk method call we have to count

src/components/safe-apps/AppFrame/index.tsx
katspaugh: Good point. The only problem is `AppIframe` and `AppFrame` – maybe I should call `AppIframe` something else? Like `SafeAppIframe`.

src/components/safe-apps/AppFrame/iframe.tsx
```diff
@@ -0,0 +1,31 @@
...
+  title?: string
+  iframeRef?: React.MutableRefObject<HTMLIFrameElement | null>
+  onLoad?: () => void
+}
+
+// see sandbox mdn docs for more details https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox
+const IFRAME_SANDBOX_ALLOWED_FEATURES =
+  'allow-scripts allow-same-origin allow-popups allow-popups-to-escape-sandbox allow-forms allow-downloads allow-orientation-lock'
+
+const AppIframe = ({ appUrl, allowedFeaturesList, iframeRef, onLoad, title }: AppIFrameProps): ReactElement => {
```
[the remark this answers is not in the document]
katspaugh: Which param are you referring to?

src/components/dashboard/GovernanceSection/GovernanceSection.tsx
katspaugh: Agree that tracking the safe info call is OK in this case.

> Perhaps we can extract the handlers so we can reuse it here?

That's a good idea, I'll see if it can be done.

katspaugh:
I'll go ahead and merge this into the epic, and then we can adjust things in a follow-up PR.

yagopv:
`SafeAppIframe` works for me as well @katspaugh !

PR #1312, opened 2022-12-05
fix: Track safe creation events under the same category

usame-algan:
## What it solves

Related to #968 

## How this PR fixes it

- The vertical tx stepper passes the category parameter so that the creation flow uses the same `eventCategory` as the new safe creation flow

## How to test it

1. Open the safe with the AB test set to false
2. Navigate to `/welcome`
3. Create a new safe
4. Observe that the GTM events include `eventCategory: create-safe` instead of `modal`

## Analytics changes

- This will track all safe creation events under the same `eventCategory`


iamacook (approved):

katspaugh:
Tested, the categories are correct. ✅ 

Screenshot 2022-12-09 at 08 07 07

PR #1319, opened 2022-12-07
Fix: reload widget on safe change + catch errors

### LINKED ISSUE #1317: [Governance Visibility] Widgets are not refreshing when changing Safe
## Bug description
The governance widgets do not reload when switching Safe. Likely to have been introduced by https://github.com/safe-global/web-core/pull/1307

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to Dashboard
 2. Open a Safe that has not tokens
 3. Switch to a Safe with voting power
 4. Observe the governance widgets state does not change when you switch Safe.

## Expected result
The governance widgets should refresh upon switching Safe

## Obtained result
The governance widgets state does not change when you switch Safe.

## Screenshots
Screenshot 2022-12-07 at 11 10 42

katspaugh:
## What it solves

Resolves #1317

## How this PR fixes it

I've added a check for `safeLoading`, like in the big AppFrame. It will re-mount the component on Safe change.

Also an error fallback is now rendered if the manifest fails to load. I've adjusted the error fallback to only show the error message once for both widget. It makes little sense to have two of the same.

### DIFF
--- a/src/components/dashboard/GovernanceSection/GovernanceSection.tsx
+++ b/src/components/dashboard/GovernanceSection/GovernanceSection.tsx
@@ -1,3 +1,4 @@
+import { useRef } from 'react'
 import { Typography, Card, Box, Alert, IconButton, Link, SvgIcon } from '@mui/material'
 import { WidgetBody } from '@/components/dashboard/styled'
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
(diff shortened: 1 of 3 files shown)


DiogoSoaress (approved):
It looks good to me and the widgets are reloading nicely when switching Safes.

PR #1323, opened 2022-12-07
fix: Change banner to a notification

### LINKED ISSUE #1202: Banner is displayed over the app header

usame-algan:
## What it solves

Resolves #1202 

## How this PR fixes it

- Changes the `PsaBanner` to be a notification that is dispatched
- Extends `Notification` to allow disabling `autoHide`
- Extends `Notification` to allow a centered position and larger size for notifications through the `isBanner` argument
- Extends `Notification` to allow an `onClose` handler being passed

## How to test it

1. Open the Safe on Firefox in a private window
2. Observe that there is no banner that is overlapping the entire header anymore
3. Observe a notification popping up on the center screen
4. Create a transaction
5. Observe that those notifications still look and behave the same as before

## Screenshots
Screenshot 2022-12-07 at 16 25 46
Screenshot 2022-12-07 at 16 25 38

### DIFF
--- a/src/components/common/Notifications/index.tsx
+++ b/src/components/common/Notifications/index.tsx
@@ -12,6 +12,7 @@ import ChevronRightIcon from '@mui/icons-material/ChevronRight'
 import { OVERVIEW_EVENTS } from '@/services/analytics/events/overview'
 import Track from '../Track'
 import { isRelativeUrl } from '@/utils/url'
+import classNames from 'classnames'
 
 const toastStyle = { position: 'static', margin: 1 }
 
@@ -50,6 +51,7 @@ const Toast = ({
   link,
   onClose,
   id,
+  autoHide = true,
 }: {
   variant: AlertColor
   onClose: () => void
@@ -70,7 +72,7 @@ const Toast = ({
   const autoHideDuration = variant === 'info' || variant === 'success' ? 5000 : undefined
 
   return (
-    <Snackbar open onClose={handleClose} sx={toastStyle} autoHideDuration={autoHideDuration}>
+    <Snackbar open onClose={handleClose} sx={toastStyle} autoHideDuration={autoHide ? autoHideDuration : null}>
       <Alert severity={variant} onClose={handleClose} elevation={3} sx={{ width: '340px' }}>
         {message}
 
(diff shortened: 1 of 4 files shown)


src/components/common/Notifications/index.tsx
```diff
@@ -50,6 +51,7 @@ const Toast = ({
   link,
   onClose,
   id,
+  autoHide = true,
```
iamacook, line 54: Nit: I would flip this to something like `persist` instead as autohiding is a default of ours.

src/store/notificationsSlice.ts
```diff
@@ -1,18 +1,22 @@
...
   detailedMessage?: string
   groupKey: string
   variant: AlertColor
   timestamp: number
   isDismissed?: boolean
   isRead?: boolean
   link?: { href: LinkProps['href']; title: string }
+  autoHide?: boolean
+  isBanner?: boolean
+  onClose?: () => void
```
iamacook, line 19: Did you try using the `Toast` component directly instead of modifying the notification system?

iamacook (commented):
Code looks good to me but the stretching across the middle seems a bit weird to me when compared to our normal notifications. I'm not sure about the layering either. What do you think @liliiaorlenko?

image

I personally prefer the implementation we had on `safe-react` where the UI moved down relative to the presence of a banner.

katspaugh (approved):
I like how it looks

src/components/common/Notifications/index.tsx
usame-algan: Good idea! Will change it to `persist`

src/store/notificationsSlice.ts
usame-algan: Will check if its easy to implement with the fixed positioning, then we can keep the notification system more pure.

liliiaorlenko:
I have a questions, from where user can export data? 
For me personally this notification is still not good enough motivator to switch to new interface, maybe a modal style with backdrop with more prominent CTAs can work better and more persistant. I can draw my proposal in next few hours.

katspaugh:
Since we haven't reached a consensus regarding the banner design, and the banner itself is going to be removed in early February, I suggest we close this.

katspaugh:
That said, I personally like the proposed design and would just merge it.

usame-algan:
Lets revisit this once we have a banner design.

PR #1324, opened 2022-12-07
Adjust safe token widget allocation formula

### LINKED ISSUE #1208: [Safe token widget] Adjust amount after redeem deadline is reached
**Note: Add this to first december sprint**
## What is the feature about
Once the redeem deadline of the airdrop contracts (user & guardian) is reached, we have to adjust the voting power in the safe token widget.

Currently it just adds all allocations from our hosted data.

Then we need to check on-chain if the tokens have been redeemed and only count allocations which have been redeemed.

## The list of requirements
- Adjust formula after 27th December

Note: We will create a new issue to implement these changes in the safe-claiming-app

schmanu:
## What it solves
The token widget did not take token balance and the redeem deadline into account.

## TODOs
- [x] add more tests
- [x] Port the hook + tests into the safe claiming app

Resolves #1208 

## How this PR fixes it
Adjusts the formula to `totalAllocation - tokensClaimed + balance`
When computing the total allocation after the `redeemDeadline` has passed we ignore unredeemed allocations.

## How to test it
- On görli the redeem deadline was on 30th of November, meaning unredeemed allocations should not be counted anymore:
  - Safe with only user airdrop but not ecosystem airdrop redeemed `gor:0x5f310dc66F4ecDE9a1769f1B7D75224dA592201e`
- On mainnet this date is on the 27th of Dec

## Analytics changes
None

### DIFF
--- a/src/hooks/useSafeTokenAllocation.ts
+++ b/src/hooks/useSafeTokenAllocation.ts
@@ -25,7 +25,8 @@ type VestingData = {
 
 type Vesting = VestingData & {
   isExpired: boolean
-  claimedTokens: string
+  isRedeemed: boolean
+  amountClaimed: string
 }
 
 // We currently do not have typechain as dependency so we fallback to human readable ABIs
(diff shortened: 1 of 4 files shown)


src/hooks/useSafeTokenAllocation.ts
```diff
@@ -39,25 +99,63 @@ const fetchAllocation = async (chainId: string, safeAddress: string): Promise<Ve
...
 
-const useSafeTokenAllocation = (): BigNumber | undefined => {
-  const [allocation, setAllocation] = useState<BigNumber>()
+const fetchTokenBalance = async (chainId: string, safeAddress: string): Promise<string> => {
+  try {
+    const web3ReadOnly = getWeb3ReadOnly()
+    const safeTokenAddress = getSafeTokenAddress(chainId)
+    if (!safeTokenAddress || !web3ReadOnly) return '0'
+
+    const tokenBalance = await web3ReadOnly?.call({
```
schmanu, line 108: I decided to fetch this on-chain instead of our redux store, because the remainder of this data (tokens claimed, etc) is also fetched on-chain so the data will always be in sync like this.

src/hooks/useSafeTokenAllocation.ts
```diff
@@ -18,6 +23,61 @@ type VestingData = {
...
+    throw new Error('Cannot fetch vesting without web3 provider')
+  }
+  const onChainVestingData = await web3ReadOnly?.call({
+    to: allocation.contract,
+    data: airdropInterface.encodeFunctionData('vestings', [allocation.vestingId]),
+  })
+
+  if (!onChainVestingData) {
+    throw new Error(`Could not load vesting information for vesting with id ${allocation.vestingId}`)
+  }
```
usame-algan, line 54: Can this ever be undefined? It seems to return empty data even for non-existing vesting ids.

src/hooks/useSafeTokenAllocation.ts
```diff
@@ -18,6 +23,61 @@ type VestingData = {
...
+
+  const decodedVestingData = defaultAbiCoder.decode(
+    ['address', 'uint8', 'bool', 'uint16', 'uint64', 'uint128', 'uint128', 'uint64', 'bool'],
```
usame-algan, line 57: I would suggest adding a comment to identify these a bit better e.g.

`// amount, curveType, managed, ...`

src/hooks/useSafeTokenAllocation.ts
```diff
@@ -18,6 +23,61 @@ type VestingData = {
...
+    onChainVestingData,
+  )
+  if (decodedVestingData[0].toLowerCase() !== ZERO_ADDRESS.toLowerCase()) {
+    // allocation is redeemed, return the amountClaimed from on-chain
+    return { ...allocation, isExpired: false, claimedTokens: decodedVestingData[6] }
+  }
```
usame-algan, line 63: You could extract this condition into a variable instead of adding a comment i.e.

```
const isRedeemed = decodedVestingData[0].toLowerCase() !== ZERO_ADDRESS.toLowerCase()

if (isRedeemed) {
    const amountClaimed = decodedVestingData[6]
    return ...
}
```

usame-algan (changes requested):

src/hooks/useSafeTokenAllocation.ts
usame-algan: I noticed we do 6 additional RPC calls on load now. Its something we should probably keep track of and maybe optimize e.g. batching `eth_call`

src/hooks/useSafeTokenAllocation.ts
usame-algan: With the logic above where we check that `web3ReadOnly` is defined, `onChainVestingData` should always exist.

schmanu: True, I changed it. This makes a lot of checks obsolete :)

usame-algan (approved):
Looks good 👍

liliya-soroka:
Verified
can be included into 1.2.0 release

PR #1329, opened 2022-12-08
fix: allow resetting to recommended nonce of 0

### LINKED ISSUE #1311: [Advanced Parameters] Recomended nonce button doesn't work for nonce = 0
## Bug description
Thanks @usame-algan for pointing it out
If the tx is the very first one the recomended nonce is 0. If you edit that value and try to use the "Recomended nonce" button it won't work

## Environment
 - Browser: Any
 - Wallet: Any
 - Chain: Any

## Steps to reproduce
 1. Create a safe or enter into a safe that never created a tx
 2. Create a new tx and go to Advanced Parameters
 3. The nonce should be 0 by default, change it to any number
 4. Click the "Recomended nonce" button

## Expected result
The nonce is reset to 0

## Obtained result
Nothing happens, the button does nothing in this case

## Screenshots
image

iamacook:
## What it solves

Resolves #1311

## How this PR fixes it

The `typeof  recommendedNonce` is checked that it is a `'number'` instead of whether it is truthy.

## How to test it

1. Create a new Safe and create a new transaction.
2. Edit the nonce to >0 and reset the nonce to the recommended value.
3. Observe that it sets to 0.

## Screenshots

recommended nonce

### DIFF
--- a/src/components/tx/NonceForm/index.tsx
+++ b/src/components/tx/NonceForm/index.tsx
@@ -89,7 +89,7 @@ const NonceForm = ({ name, nonce, recommendedNonce, readonly }: NonceFormProps):
   const label = fieldState.error?.message || nonceWarning || 'Safe transaction nonce'
 
   const onResetNonce = () => {
-    if (recommendedNonce) {
+    if (typeof recommendedNonce === 'number') {
       setValue(name, recommendedNonce, { shouldValidate: true })
     }
   }
--- a/src/components/tx/NonceForm/index.tsx
+++ b/src/components/tx/NonceForm/index.tsx
@@ -89,7 +89,7 @@ const NonceForm = ({ name, nonce, recommendedNonce, readonly }: NonceFormProps):
   const label = fieldState.error?.message || nonceWarning || 'Safe transaction nonce'
 
   const onResetNonce = () => {
-    if (typeof recommendedNonce === 'number') {
+    if (recommendedNonce != null) {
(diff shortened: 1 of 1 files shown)


src/components/tx/NonceForm/index.tsx
```diff
@@ -89,7 +89,7 @@ const NonceForm = ({ name, nonce, recommendedNonce, readonly }: NonceFormProps):
   const label = fieldState.error?.message || nonceWarning || 'Safe transaction nonce'
 
   const onResetNonce = () => {
-    if (recommendedNonce) {
+    if (typeof recommendedNonce === 'number') {
```
katspaugh, line 92:
```suggestion
    if (recommendedNonce >= 0) {
```

Or `!isNaN(recommendedNonce)`.

schmanu: I like `!isNaN(recommendedNonce)`

The validation already checks if the nonce is positive. So I'd say we do not need to check that here.

iamacook: Whilst I agree that both suggestions are cleaner, TypeScript complains that `recommendedNonce` is otherwise `undefined`. I've altered it to `recommendedNonce != null` instead.

katspaugh (approved):

katspaugh:
Verified, it resets to zero. ✅ 

Screenshot 2022-12-09 at 08 17 49

schmanu (approved):

PR #1342, opened 2022-12-09
fix: adjust text of buttons in transfer modal

iamacook:
## What it solves

Mixture of nouns and verbs in the transfer modal.

## How this PR fixes it

The "Contract interaction" text has been changed to "Build transaction" and the title of the replacement button area to "Select how you would like to replace this transaction" in order to match the semantics of the buttons that are all now titled with verbs ("Rejection transaction" is now "Reject transaction").

## How to test it

1. Open the new transaction modal either by the sidebar or transaction replacement.
2. Observe the new text.

## Screenshots

image

image

### DIFF
--- a/src/components/tx/modals/NewTxModal/ReplacementModal.tsx
+++ b/src/components/tx/modals/NewTxModal/ReplacementModal.tsx
@@ -101,7 +101,7 @@ const ReplacementModal = ({
         <Grid container alignItems="center" justifyContent="center" flexDirection={flexDirection}>
           <Grid item xs={12}>
             <Typography variant="body2" textAlign="center" fontWeight={700} mb={3}>
-              Select how would you like to replace this transaction
+              Select how you would like to replace this transaction
             </Typography>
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
In the new tx modal, verbs were better imho, more actionable.  But it’s OK since it’s the same component.

src/components/tx/modals/NewTxModal/ReplacementModal.tsx
```diff
@@ -101,7 +101,7 @@ const ReplacementModal = ({
         <Grid container alignItems="center" justifyContent="center" flexDirection={flexDirection}>
           <Grid item xs={12}>
             <Typography variant="body2" textAlign="center" fontWeight={700} mb={3}>
-              Select the type of replacement transaction
+              Select how would you like to replace this transaction
```
katspaugh, line 104:
```suggestion
              Select how you would like to replace this transaction
```

PR #1346, opened 2022-12-12
fix: Display available spending limit amount

### LINKED ISSUE #1310: UI displays the wrong amount for spending limits
## Bug description

- When creating a spending limit transaction, the UI displays the full amount instead of amount - spent.
- The spacing inside the modal also looks a bit off

## Steps to reproduce
 1. Open a Safe
 2. Create a new spending limit for your wallet
 3. Execute a spending limit transaction
 4. Observe that it displays the correct spent amount on the settings page
 5. Create a new spending limit transaction
 6. Observe that it still displays the original amount

## Expected result

- It should display the full amount - spent amount inside the modal
- The area should have consistent top and bottom spacing

## Obtained result

It displays the full amount inside the modal

## Screenshots
In this case a transaction with 100 SAFE was already executed but it still displays 500 instead of 400. Going to the next step results in an error that is not caught in the UI.
Screenshot 2022-12-05 at 17 28 32

usame-algan:
## What it solves

Resolves #1310 

## How this PR fixes it

- Displays the available spending limit amount (amount-spent) with more precision

## How to test it

1. Open a safe with a spending limit in place
2. Create a new transaction
3. Select the token that has a spending limit
4. Observe the amount is displayed with enough precision
5. Press Max
6. Observe the same amount in the input field

## Screenshots
Screenshot 2022-12-12 at 11 57 37

### DIFF
--- a/src/components/tx/SpendingLimitRow/index.tsx
+++ b/src/components/tx/SpendingLimitRow/index.tsx
@@ -1,25 +1,25 @@
 import { FormControl, FormControlLabel, Radio, RadioGroup, Typography } from '@mui/material'
 import { Controller, useFormContext } from 'react-hook-form'
-import { formatVisualAmount } from '@/utils/formatters'
+import type { BigNumber } from '@ethersproject/bignumber'
+import { safeFormatUnits } from '@/utils/formatters'
 import type { TokenInfo } from '@safe-global/safe-gateway-typescript-sdk'
-import type { SpendingLimitState } from '@/store/spendingLimitsSlice'
 import { SendAssetsField, SendTxType } from '@/components/tx/modals/TokenTransferModal/SendAssetsForm'
 
 const SpendingLimitRow = ({
-  spendingLimit,
+  availableAmount,
   selectedToken,
 }: {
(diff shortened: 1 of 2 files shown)


src/components/tx/SpendingLimitRow/index.tsx
```diff
@@ -1,25 +1,25 @@
...
   selectedToken,
 }: {
-  spendingLimit: SpendingLimitState
+  availableAmount: BigNumber
   selectedToken: TokenInfo | undefined
 }) => {
   const { control } = useFormContext()
 
-  const formattedAmount = formatVisualAmount(spendingLimit.amount, selectedToken?.decimals)
+  const formattedAmount = safeFormatUnits(availableAmount, selectedToken?.decimals)
```
usame-algan, line 17: I decided to change this so that it doesn't cut off decimal places because it is the same as in safe-react.

schmanu (approved):
Works great!

One thing we should consider adding soon here is a cypress test for spending limits. we could setup a safe with a Spending limit and just check that the interface reacts accordingly.

PR #1348, opened 2022-12-12
fix: change favicon per theme

DiogoSoaress:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/1123

## How this PR fixes it
Implements a hook to detect the system's preferred color scheme and subscribes to that media query changes

## How to test it
Go to your OS preferences and change the dark/light theme.
Observe the favicon in the browsers tab changing accordingly.

## Screenshots
Screenshot 2022-12-12 at 15 41 49


src/hooks/__tests__/usePrefersColorScheme.test.ts
```diff
@@ -0,0 +1,13 @@
+import usePrefersColorScheme from '@/hooks/usePrefersColorScheme'
+import { renderHook } from '@/tests/test-utils'
+
+describe('usePrefersColorScheme', () => {
+  it('should return the initial color scheme', () => {
```
schmanu, line 5: Could we add more tests for the listeners of the media query which would set the result to dark / light?

DiogoSoaress: Added. I found it a bit difficult to mock the media query event without adding the dependency. Maybe we can look at it quickly together

src/hooks/usePrefersColorScheme.ts
```diff
@@ -0,0 +1,33 @@
+import { useEffect, useState } from 'react'
+
+type ColorSchemeType = 'dark' | 'light' | 'no-preference'
+
+const usePrefersColorScheme = (): ColorSchemeType => {
```
iamacook, line 5: We should also take the dark theme into account. We have a very similar hook called `useDarkMode`.

DiogoSoaress: You mean considering the app's theme to affect the favicon?

iamacook: Yes. The app theme may not match that of the OS.

DiogoSoaress: Exactly for that reason, we should only make the favicon contrast with the OS theme. The app theme does not  affect the browser tabs color.

iamacook: I see. Oversight on my behalf.

Does this work nicely with the pairing module/WC? I don't have an iPhone so I can't test it.

src/hooks/__tests__/usePrefersColorScheme.test.ts
schmanu: I can have a look later to do it as described here:
https://jestjs.io/docs/26.x/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom

src/components/common/MetaTags/index.tsx
```diff
@@ -41,8 +41,6 @@ const MetaTags = ({ prefetchUrl }: { prefetchUrl: string }) => (
     {/* Favicons */}
     <link rel="shortcut icon" href="/favicons/favicon.ico" />
     <link rel="apple-touch-icon" sizes="180x180" href="/favicons/apple-touch-icon.png" />
-    <link rel="icon" type="image/png" sizes="32x32" href="/favicons/favicon-32x32.png" />
-    <link rel="icon" type="image/png" sizes="16x16" href="/favicons/favicon-16x16.png" />
     <link rel="mask-icon" href="/favicons/safari-pinned-tab.svg" color="#000" />
```
iamacook, line 44: Do we not need to update these other favicons?

src/hooks/usePrefersColorScheme.ts
iamacook: And the icons are clearly visible in light/dark mode on WC/app?

src/hooks/usePrefersColorScheme.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const isDark = window.matchMedia('(prefers-color-scheme: dark)')
+    const isLight = window.matchMedia('(prefers-color-scheme: light)')
+
+    const darkListener = ({ matches }: MediaQueryListEvent) => {
+      matches && setPreferredColorScheme('dark')
+    }
+    const lightListener = ({ matches }: MediaQueryListEvent) => {
+      matches && setPreferredColorScheme('light')
+    }
```
iamacook, line 19: You could curry these to reduce repetition, e.g.

```suggestion
    const listener = (type: ColorSchemeType) = ({ matches }: MediaQueryListEvent) => {
      matches && setPreferredColorScheme(type)
    }
    
    const darkListener = listener('dark')
    const lightListener = listener('light')
```

src/pages/_app.tsx
```diff
@@ -77,11 +78,16 @@ interface WebCoreAppProps extends AppProps {
...
+  const prefersColorScheme = usePrefersColorScheme()
+  const darkPath = prefersColorScheme === 'dark' ? '-dark' : ''
+
   return (
     <StoreHydrator>
       <Head>
         <title key="default-title">Safe</title>
         <MetaTags prefetchUrl={GATEWAY_URL} />
+        <link rel="icon" type="image/png" sizes="32x32" href={`/favicons/favicon${darkPath}-32x32.png`} />
+        <link rel="icon" type="image/png" sizes="16x16" href={`/favicons/favicon${darkPath}-16x16.png`} />
```
iamacook, line 90: Can we not keep these and associated logic inside `MetaTag`?

src/hooks/usePrefersColorScheme.ts
DiogoSoaress: Yes. Those icons have the green circle background (I didn't change it)

src/pages/_app.tsx
DiogoSoaress: I tried to include the logic inside <MetaTag /> component but it was not taking effect. I tried to log from inside that component but unsuccessfully. Happy to look at it together if we want to move these `<link>` back there

src/components/common/MetaTags/index.tsx
DiogoSoaress: I've added dark versions to the ones that were subject to poor contrast in browser's tabs.
If I get it correctly the ones I left in `<MetaTags>` are for other uses.

iamacook: I'm still seeing the old icon in Firefox:
image

I would suggest we test adding an icon to the home screen of an iPhone, as well as pinning a tab in Safari.

src/hooks/usePrefersColorScheme.ts
DiogoSoaress: Good suggestion! Done

src/components/common/MetaTags/index.tsx
iamacook: I opened it in a Private Window though.

src/hooks/usePrefersColorScheme.ts
```diff
@@ -0,0 +1,35 @@
...
+type ColorSchemeType = 'dark' | 'light' | 'no-preference'
+
+const usePrefersColorScheme = (): ColorSchemeType => {
+  const [preferredColorScheme, setPreferredColorScheme] = useState<ColorSchemeType>('no-preference')
+
+  useEffect(() => {
+    if (typeof window.matchMedia !== 'function') return
+
+    const isDark = window.matchMedia('(prefers-color-scheme: dark)')
+    const isLight = window.matchMedia('(prefers-color-scheme: light)')
```
katspaugh, line 12: Aren't they mutually exclusive?

src/pages/_app.tsx
```diff
@@ -77,11 +78,16 @@ interface WebCoreAppProps extends AppProps {
 }
 
 const WebCoreApp = ({ Component, pageProps, emotionCache = clientSideEmotionCache }: WebCoreAppProps): ReactElement => {
+  const prefersColorScheme = usePrefersColorScheme()
+  const darkPath = prefersColorScheme === 'dark' ? '-dark' : ''
```
katspaugh, line 82: `darkPath` is not the best name if it can be either dark or light.

katspaugh (changes requested):
You propose adding 128 lines of JS and an external library to change a favicon.
From the maintainability standpoint, I would much prefer a low-tech solution like an icon with a background.

katspaugh:
There's a CSS-only solution too, have you tried it?
https://stackoverflow.com/a/60158619/352796

src/hooks/usePrefersColorScheme.ts
DiogoSoaress: From my understanding of https://w3c.github.io/csswg-drafts/mediaqueries-5/#prefers-color-scheme they are not mutually exclusive but correct me if I'm wrong. The user can prefer `dark` or `light` or have no preference.

DiogoSoaress:
> You propose adding 128 lines of JS and an external library to change a favicon.

I know. Well, the changes are more numerous given the added hook and the test coverage. The dev dependency is also only used for tests and it was added following the code review.

I will talk with @liliiaorlenko to get a `svg` favicon to explore the CSS-only solution and to get her feedback on changing the favicon to a round background that would have good contrast regardless of the OS preferred theme.

src/hooks/usePrefersColorScheme.ts
katspaugh: Yeah, but you don't have a third icon for `!isDark && !isLight`, do you?
Also from the docs you linked (emphasis mine):
> `light`
> Indicates that user has expressed the preference for a page that has a light theme (dark text on light background), **or has not expressed an active preference (and thus should receive the "web default" of a light theme)**.

DiogoSoaress:
> There's a CSS-only solution too, have you tried it?

I was trying this implementation but I don't think it suits our use case give it builds over the assumption that nowadays you can use an SVG as a favicon for your website, which is hardly true.
https://caniuse.com/link-icon-svg

I also got the favicon with the green background from the designer. I will create a different PR for it so we can compare the 2 solutions

DiogoSoaress:
I've addressed the comments and opened an alternative implementation to this PR using the favicon with the green background https://github.com/safe-global/web-core/pull/1376

katspaugh:
It wasn't necessary to spend time addressing the comments. Closing in favor of the other PR.

PR #1365, opened 2022-12-13
fix: display factory and mastercopy known names in create tx info

### LINKED ISSUE #1347: Known addresses info lost on safe creation tx
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to https://app.safe.global/eth:0x92Db8FF884Fd2872888a61244FAEB1C50bf10cc3/transactions/history
 2. check Safe creation tx

## Expected result
image

## Screenshots
image

Found in 1.2.0 rc , but existing on prod

DiogoSoaress:
## What it solves

Resolves #1347 

## How this PR fixes it
Passes mastercopy and factory known address names to the CreationTxInfo elements

## How to test it
Expand a create tx info
(`eth:0x92Db8FF884Fd2872888a61244FAEB1C50bf10cc3/transactions/history`)

## Screenshots
Screenshot 2022-12-13 at 18 37 47

Screenshot 2022-12-14 at 16 03 20


DiogoSoaress:
> Are you sure that these are "known addresses" on all networks? Afaiu, these need to have been manually named on the infra side.

If the name is not know by the transaction services, it will display the one from AB or none.
We follow the same logic in `safe-react` https://github.com/safe-global/safe-react/blob/abcc2e420423662a632ae6dc33e1616cdc94c132/src/routes/safe/components/Transactions/TxList/TxInfoCreation.tsx#L28-L53

iamacook:
> If the name is not know by the transaction services, it will display the one from AB or none. We follow the same logic in `safe-react` https://github.com/safe-global/safe-react/blob/abcc2e420423662a632ae6dc33e1616cdc94c132/src/routes/safe/components/Transactions/TxList/TxInfoCreation.tsx#L28-L53

It [won't show anything if there is no `factory` or `implementation`](https://github.com/safe-global/web-core/blob/fix-known-addr-safe-creation/src/components/transactions/SafeCreationTx/index.tsx#L27-L42).

DiogoSoaress:
> Are you sure that these are "known addresses" on all networks?

If they are not known (not answered by the services) we simply don't display the Address name. How do you think it should behave? From my tests across networks they are always present.

iamacook:
> If they are not known (not answered by the services) we simply don't display the Address name. How do you think it should behave? From my tests across networks they are always present.

Your [example from `safe-react`](https://github.com/safe-global/safe-react/blob/abcc2e420423662a632ae6dc33e1616cdc94c132/src/routes/safe/components/Transactions/TxList/TxInfoCreation.tsx#L34-L36) shows `'n/a'` if no `factory` or `implementation` is returned. We're hiding them entirely.

Do we want to keep this logic and show the title with a pre-defined "empty state" or just hide them entirely? There is [no design reference](https://www.figma.com/file/47wkjxMZkyEeCRbCrLJBJa/Tx-details?node-id=1358%3A20119&t=yP9L1j7YHKzhIMoo-0) for this. cc @liliiaorlenko

DiogoSoaress:
I've added a case to handle missing factory / mastercopy txInfo
Screenshot 2022-12-14 at 16 03 20

mimics `safe-react` WDYT?

iamacook (approved):

iamacook:
Verified that this works well.

image

PR #1370, opened 2022-12-14
fix: accordion mounting

### LINKED ISSUE #1303: Open accordions with a smooth transition

iamacook:
## What it solves

Resolves #1303

## How this PR fixes it

The `Accordion`s now only mount on enter, not unmounting when closing.

## How to test it

Open/close the transaction accordions/Safe Apps queue bar and observe that there is only one network request. Subsequent opens/closes are then smooth.

## Screenshots

accordions

### DIFF
--- a/src/components/safe-apps/AppFrame/TransactionQueueBar/index.tsx
+++ b/src/components/safe-apps/AppFrame/TransactionQueueBar/index.tsx
@@ -49,7 +49,7 @@ const TransactionQueueBar = ({
                 enter: 0,
                 exit: 500,
               },
-              unmountOnExit: true,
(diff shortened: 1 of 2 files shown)


usame-algan (approved):
It also doesn't do the fetch anymore everytime it is reopened, really nice fix! 🚀

katspaugh:
Verified, works smoothly and no extra requests. ✅

PR #1375, opened 2022-12-14
fix: use break-word

### LINKED ISSUE #1277: Fix the text in the error message when the tx will most likely fail
Steps:
1. go to the safe gno:0x9DF58F04F3e914ED8D972146401C85313A53FecA on prod env
2. try to execute the tx with nonce=44

Current result:
"missing revert data in call exception; Transaction reverted without a reason s
tring" 
Image

Expected result: Could we control the view of the error ?

Prod env.
Chain: Gnosis chain

DiogoSoaress:
## What it solves

Resolves #1277 

## How this PR fixes it
Uses `word-break: break-word` for the error details message

## How to test it
Follow https://github.com/safe-global/web-core/issues/1277 steps

## Screenshots
Screenshot 2022-12-14 at 17 04 57
Screenshot 2022-12-14 at 17 04 39


usame-algan (approved):

katspaugh:
Verified, looks good.

Screenshot 2022-12-19 at 10 16 22

PR #1377, opened 2022-12-15
[Spam tokens] manually hide tokens 

### LINKED ISSUE #1337: [Epic] Manually hiding tokens
## What is the feature about
**Related to Epic #865**
This feature enables users to manually hide unwanted / spam tokens from our interface.
For every asset (except native ones) we will add a checkbox / button to hide the asset. 
Hidden assets will not be displayed / counted:
- in the asset list
- in token auto completions (Send funds, Spending Limits)
- In the total balance

Note: We will not hide transactions in the history.

## The list of requirements
- [x] add new menu to hide / unhide tokens
- [x] ignore hidden tokens in UI
- [x] remember hidden token in local storage

## Designs/sketches
https://www.figma.com/file/ZKWJPH1MjJIyUDTHYD3kk8/Default-token-lists?node-id=2%3A26&t=bKTHJUzmMmfe6jLa-0

## Links
https://www.notion.so/gnosis-safe/Default-token-lists-and-local-hiding-of-tokens-cd353630b55c4625bac5d4f6ae546c4f

schmanu:
## What it solves
Unwanted / spam tokens could not be hidden from the interface.

Resolves #1337 

## How this PR fixes it
- New button in Assets page to manually hide / unhide tokens
- new parameter for `useBalances` to get all or only not hidden tokens

## How to test it
- Hide some tokens in the Assets page of a safe
- These tokens should be hidden for all Safes on that chain
  - Not appear in assets
  - not counted in total balance
  - not appear through safe apps sdk `getBalance`
  - not show in Dashboard Overview

## Analytics changes
- New events for hiding / unhiding tokens

### DIFF
--- a/src/components/balances/AssetsHeader/index.tsx
+++ b/src/components/balances/AssetsHeader/index.tsx
@@ -5,7 +5,7 @@ import NavTabs from '@/components/common/NavTabs'
 import PageHeader from '@/components/common/PageHeader'
 import CurrencySelect from '@/components/balances/CurrencySelect'
 import { balancesNavItems } from '@/components/sidebar/SidebarNavigation/config'
-import { useHiddenAssets } from '@/hooks/useHiddenAssets'
+import useHiddenAssets from '@/hooks/useHiddenAssets'
 import { VisibilityOutlined, VisibilityOffOutlined } from '@mui/icons-material'
 
 import css from './styles.module.css'
--- a/src/components/balances/AssetsHeader/index.tsx
+++ b/src/components/balances/AssetsHeader/index.tsx
@@ -1,19 +1,46 @@
-import { Box } from '@mui/material'
-import type { ReactElement } from 'react'
+import { Box, ButtonBase, Typography } from '@mui/material'
+import { type ReactElement, useContext } from 'react'
 
 import NavTabs from '@/components/common/NavTabs'
 import PageHeader from '@/components/common/PageHeader'
 import CurrencySelect from '@/components/balances/CurrencySelect'
 import { balancesNavItems } from '@/components/sidebar/SidebarNavigation/config'
+import { useHiddenAssets } from '@/hooks/useHiddenAssets'
+import { VisibilityOutlined, VisibilityOffOutlined } from '@mui/icons-material'
 
-const AssetsHeader = ({ currencySelect = false }: { currencySelect?: boolean }): ReactElement => {
+import css from './styles.module.css'
+import { HiddenAssetsContext } from '../HiddenAssetsProvider'
+
+const AssetsHeader = ({
+  hiddenAssets = false,
+  currencySelect = false,
+}: {
+  hiddenAssets?: boolean
+  currencySelect?: boolean
+}): ReactElement => {
+  const currentHiddenAssets = useHiddenAssets()
+  const hiddenAssetCount = currentHiddenAssets ? Object.keys(currentHiddenAssets).length : 0
+  const { showHiddenAssets, toggleShowHiddenAssets } = useContext(HiddenAssetsContext)
+
   return (
     <PageHeader
       title="Assets"
       action={
         <Box display="flex" justifyContent="space-between" alignItems="center">
           <NavTabs tabs={balancesNavItems} />
-          {currencySelect && <CurrencySelect />}
+          <Box display="flex" flexDirection="row" alignItems="center" gap={1}>
+            {hiddenAssets && hiddenAssetCount > 0 && (
+              <ButtonBase className={css.hiddenAssetsButton} onClick={toggleShowHiddenAssets}>
+                {showHiddenAssets ? (
+                  <VisibilityOutlined color="border" fontSize="small" />
+                ) : (
+                  <VisibilityOffOutlined color="border" fontSize="small" />
+                )}{' '}
+                <Typography fontSize="small">{hiddenAssetCount}</Typography>
+              </ButtonBase>
+            )}
+            {currencySelect && <CurrencySelect />}
+          </Box>
         </Box>
       }
     />
--- a/src/components/balances/AssetsTable/index.tsx
+++ b/src/components/balances/AssetsTable/index.tsx
@@ -1,6 +1,7 @@
-import { useState, type ReactElement, useMemo } from 'react'
-import { Button, Tooltip, Typography, SvgIcon } from '@mui/material'
-import { type SafeBalanceResponse, TokenType } from '@safe-global/safe-gateway-typescript-sdk'
+import { useState, type ReactElement, useMemo, useContext, useCallback } from 'react'
+import { Button, Tooltip, Typography, SvgIcon, IconButton, Box } from '@mui/material'
+import type { SafeBalanceResponse, TokenInfo } from '@safe-global/safe-gateway-typescript-sdk'
+import { TokenType } from '@safe-global/safe-gateway-typescript-sdk'
 import css from './styles.module.css'
 import FiatValue from '@/components/common/FiatValue'
 import TokenAmount from '@/components/common/TokenAmount'
@@ -12,14 +13,44 @@ import useIsGranted from '@/hooks/useIsGranted'
 import Track from '@/components/common/Track'
 import { ASSETS_EVENTS } from '@/services/analytics/events/assets'
 import InfoIcon from '@/public/images/notifications/info.svg'
+import { VisibilityOffOutlined, VisibilityOutlined } from '@mui/icons-material'
+import { HiddenAssetsContext } from '../HiddenAssetsProvider'
+import { useHiddenAssets } from '@/hooks/useHiddenAssets'
 
 interface AssetsTableProps {
   items?: SafeBalanceResponse['items']
 }
 
+const isNativeToken = (tokenInfo: TokenInfo) => {
+  return tokenInfo.type === TokenType.NATIVE_TOKEN
+}
+
 const AssetsTable = ({ items }: AssetsTableProps): ReactElement => {
   const [selectedAsset, setSelectedAsset] = useState<string | undefined>()
   const isGranted = useIsGranted()
+  const { showHiddenAssets, toggleAsset, saveChanges, assetsToHide, assetsToUnhide, reset } =
+    useContext(HiddenAssetsContext)
+  const hiddenAssets = useHiddenAssets()
+
+  const visibleItems = useMemo(
+    () =>
+      showHiddenAssets
+        ? items
+        : items?.filter(
+            (item) => isNativeToken(item.tokenInfo) || typeof hiddenAssets?.[item.tokenInfo.address] === 'undefined',
+          ),
+    [hiddenAssets, items, showHiddenAssets],
+  )
+
+  // Assets are selected if they are either hidden or marked for hiding
+  const isAssetSelected = useCallback(
+    (address: string) =>
+      (hiddenAssets && typeof hiddenAssets[address] !== 'undefined' && !assetsToUnhide.includes(address)) ||
+      assetsToHide.includes(address),
+    [assetsToHide, assetsToUnhide, hiddenAssets],
+  )
+
+  const selectedAssetCount = visibleItems?.filter((item) => isAssetSelected(item.tokenInfo.address)).length || 0
 
   const shouldHideActions = !isGranted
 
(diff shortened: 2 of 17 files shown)


src/hooks/useBalances.ts
```diff
@@ -3,22 +3,52 @@ import isEqual from 'lodash/isEqual'
 import { type SafeBalanceResponse } from '@safe-global/safe-gateway-typescript-sdk'
 import { useAppSelector } from '@/store'
 import { selectBalances } from '@/store/balancesSlice'
+import type { HiddenAssetsOnChain } from '@/store/hiddenAssetsSlice'
+import useHiddenAssets from './useHiddenAssets'
+import { safeFormatUnits, safeParseUnits } from '@/utils/formatters'
+import { BigNumber } from 'ethers'
 
-const useBalances = (): {
+const removeHiddenBalances = (
```
katspaugh, line 11: `remove` sounds like a destructive operation. `filterHiddenTokens` would be better.

src/hooks/useBalances.ts
```diff
@@ -3,22 +3,52 @@ import isEqual from 'lodash/isEqual'
...
+  balances: SafeBalanceResponse,
+  hiddenAssets: HiddenAssetsOnChain,
+): SafeBalanceResponse => {
+  const fiatTotalVisible = safeFormatUnits(
```
katspaugh, line 15: I wouldn't mess with the total. Hidden doesn't mean it's not there at all. I would keep the total as is sent from the backend.

src/store/hiddenAssetsSlice.ts
```diff
@@ -0,0 +1,54 @@
+import { createSelector, createSlice, type PayloadAction } from '@reduxjs/toolkit'
+import type { RootState } from '.'
+
+export type HiddenAssetsOnChain = {
+  [tokenAddress: string]: string
```
katspaugh, line 5: Why does it need to be a map instead of just a flat list of token addresses?

src/store/hiddenAssetsSlice.ts
```diff
@@ -0,0 +1,54 @@
+import { createSelector, createSlice, type PayloadAction } from '@reduxjs/toolkit'
+import type { RootState } from '.'
+
+export type HiddenAssetsOnChain = {
```
katspaugh, line 4: What does "onchain" mean here?

src/components/balances/HiddenAssetsProvider/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  showHiddenAssets: false,
+  toggleShowHiddenAssets: () => {},
+  toggleAsset: () => {},
+  saveChanges: () => {},
+  reset: () => {},
+  isAssetSelected: () => true,
+  visibleAssets: [],
+})
+
+const HiddenAssetsProvider = ({ children }: { children: ReactNode }): ReactElement => {
```
katspaugh, line 29: Why do you need a provider when we usually just use selectors and dispatchers directly? Feels like an unnecessary abstraction layer.

katspaugh (commented):
I would suggest rethinking the terminology. E.g. UserTokenBlocklist instead of HiddenAssets.

src/hooks/useBalances.ts
schmanu: During problem shaping we made a list of all places where hiding an asset could have influence and discussed each.

For total balances we thought it would be nice as a Safe could receive scam tokens with a value denoted to them. 
Also using different token lists and manually hiding tokens would have consistent effects like this as the total balance from our tx service would also change if we filter by default list.

katspaugh: When the filter is on the backend, it obviously should change the total. Tokens hidden on the client-side affecting the total makes the client-side logic too complex. We should avoid messing with the financially significant data on the client side. Besides, it creates more divergence between mobile and desktop, both technically and product-wise.
On mobile, the balance would still include the hidden tokens and it might confuse the user.

src/components/balances/AssetsTable/index.tsx
```diff
@@ -12,118 +13,258 @@ import useIsGranted from '@/hooks/useIsGranted'
...
+      (hiddenAssets.includes(address) && !assetsToUnhide.includes(address)) || assetsToHide.includes(address),
+    [assetsToHide, assetsToUnhide, hiddenAssets],
+  )
+
+  const visibleAssets = useMemo(
+    () =>
+      showHiddenAssets
+        ? balances.items
+        : balances.items?.filter(
+            (item) => item.tokenInfo.type === TokenType.NATIVE_TOKEN || !hiddenAssets.includes(item.tokenInfo.address),
```
katspaugh, line 114: Why is the native token filtered out?

katspaugh (commented):
As you TODO says, there's a lot of new logic inside the AssetsTable that would be good to extract into a module dedicated to token hiding.

src/components/balances/AssetsTable/index.tsx
```diff
@@ -12,118 +13,258 @@ import useIsGranted from '@/hooks/useIsGranted'
...
+  const [assetsToHide, setAssetsToHide] = useState<string[]>([])
+  const [assetsToUnhide, setAssetsToUnhide] = useState<string[]>([])
+  const hiddenAssets = useHiddenTokens()
+  const { balances } = useBalances(true)
+  const dispatch = useAppDispatch()
+  const chainId = useChainId()
+  const [hidingAsset, setHidingAsset] = useState<string>()
 
-  const shouldHideActions = !isGranted
+  // TODO: move logic into hook
```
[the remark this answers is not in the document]
schmanu: I just finished this TODO :) Now it is ready.

src/components/balances/AssetsTable/index.tsx
schmanu: Currently because I can't imagine people actually want to hide it on purpose and because the native token has no address. 
So we would need some hacky workaround such as "If the Zero address is in the hidden tokens, we do not show the native token"

src/components/balances/AssetsTable/index.tsx
```diff
@@ -12,118 +13,258 @@ import useIsGranted from '@/hooks/useIsGranted'
...
+                      color="primary"
+                      size="small"
+                      onClick={() => setSelectedAsset(item.tokenInfo.address)}
+                    >
+                      Send
+                    </Button>
+                  </Track>
+                )}
+                {showHiddenAssets ? (
+                  <Track {...ASSETS_EVENTS.HIDE_CHECKBOX}>
```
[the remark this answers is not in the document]
schmanu: I will move the tracking to all dialog buttons. Makes more sense 👍

src/hooks/useBalances.ts
```diff
@@ -3,22 +3,52 @@ import isEqual from 'lodash/isEqual'
...
+    balances.items
+      .reduce((acc, balanceItem) => {
+        if (typeof hiddenAssets?.[balanceItem.tokenInfo.address] !== 'undefined') {
```
katspaugh: I still think we shouldn't touch the total balance. What did backend and mobile say to this?

Sidenote: we use `typeof === 'undefined'` for entirely undeclared variables, for regular null/undefined-values you can simply do `== null` or `=== undefined`.

src/components/balances/AssetsTable/index.tsx
katspaugh: I don't see how it's a reason not to allow the user to do it if they want to. Zero address (0x0000...) is what backend sends us the native token's address IIRC.

src/hooks/useBalances.ts
schmanu: This is the last discussion point where I want product to have the last call. If @johannesmoormann decides to leave it untouched it is a very quick change. 

The decision was made based on the fact that there are spam tokens which are part of liquidity pools leading to price oracles returning a USD value for them. This could cloud the real total USD balance / value of the Safe.

src/components/balances/AssetsTable/index.tsx
schmanu: Okay I'll remove this condition and the disabled flags in the components.
I checked and true: Zero address is returned by backend.

src/hooks/useBalances.ts
katspaugh: I'm talking about the solution with summing up fiat values on the client side.
Moving this logic to `useVisibleBalances` as discussed on Slack makes it slightly better.

src/hooks/useVisibleBalances.ts
```diff
@@ -0,0 +1,21 @@
...
+  const balances = useBalances()
+  const hiddenTokens = useHiddenTokens()
+  const { items, fiatTotal } = balances.balances
+
+  return {
+    ...balances,
+    balances: {
+      items: items.filter((item) => !hiddenTokens.includes(item.tokenInfo.address)),
+      fiatTotal,
+    },
```
katspaugh, line 19: Bear in mind this will filter the list and create a new object on each render. Would be good to memo it.

src/components/balances/AssetsTable/useHideAssets.ts
```diff
@@ -0,0 +1,88 @@
+import useChainId from '@/hooks/useChainId'
+import useHiddenTokens from '@/hooks/useHiddenTokens'
+import { useAppDispatch } from '@/store'
+import { setHiddenTokensForChain } from '@/store/settingsSlice'
+import { useCallback, useState } from 'react'
+
+// This is the default for MUI Collapse
+export const COLLAPSE_TIMEOUT_MS = 300
```
katspaugh, line 8: No need to export it, it's used in the same module.

usame-algan:
There is some overlap/spacing issues on smaller screens. Do we have mobile designs? cc @liliiaorlenko 

Screenshot 2023-01-05 at 11 30 37

katspaugh (approved):
Looks good from my side, apart from the remaining comments.

src/components/balances/AssetsTable/useHideAssets.ts
schmanu: The test is using it.

src/components/balances/AssetsTable/useHideAssets.ts
```diff
@@ -0,0 +1,90 @@
...
+  const [assetsToHide, setAssetsToHide] = useState<string[]>([])
+  const [assetsToUnhide, setAssetsToUnhide] = useState<string[]>([])
+  const [hidingAsset, setHidingAsset] = useState<string>()
+  const hiddenAssets = useHiddenTokens()
+
+  const toggleAsset = useCallback(
+    (address: string) => {
+      if (assetsToHide.includes(address)) {
+        const newAssetsToHide = [...assetsToHide]
+        newAssetsToHide.splice(newAssetsToHide.indexOf(address), 1)
```
katspaugh, line 23: We usually use `.filter` for this.

src/components/balances/AssetsTable/useHideAssets.ts
```diff
@@ -0,0 +1,90 @@
...
+      if (assetsToUnhide.includes(address)) {
+        const newAssetsToUnhide = [...assetsToUnhide]
+        newAssetsToUnhide.splice(newAssetsToUnhide.indexOf(address), 1)
+        setAssetsToUnhide(newAssetsToUnhide)
+        return
+      }
+
+      const assetIsHidden = hiddenAssets.includes(address)
+      if (!assetIsHidden) {
+        setAssetsToHide([...assetsToHide, address])
```
katspaugh, line 37: And for this, IMHO, concat is better. `setAssetsToHide((items) => items.concat(address))`.

francovenica:
1 - The feature works fine for desktop, but for mobile it not usable: you cannot see the checkboxes while the "hide tokens menu" is open so you cannot use the feature.
I'd suggest hiding the button for mobile views until we can design it to be mobile friendly
image

2 - The widget that counts the amount of tokens in the homepage still shows all the tokens you have, hidden or not. I'd argue that if you want to hide tokens, you don't want them to be counted with the ones you didn't hide.
I'd propose not count them at all or, if you have hidden tokens, you show a number in "()" indicating how many of them are hidden ones. 
image

3 - Just refreshing a page with a bunch of tokens throws a ton of GTM calls. No tokens are hidden right now. 
I'd assume the calls should be:
When the menu is opened/closed
When the cancel/deselect button are clicked
When the save button is clicked, sending how many tokens are being hidden by that save action.
When the "eye" icon is clicked to hide a token
image

Questions/Concerns:
- Deselect all seems to clear the storage, so if you use it in a safe with barely no tokens you will end up deselecting a ton of tokens that you might have hidden in another safe without you knowing it. Maybe deselect can just trigger the click action in the checkboxes that are visible instead, so deselecting in one safe won't affect the others (unless they share the token being hidden/shown)

- The total safe balance is not being affected by hiding tokens. I've been told that we are going to ask Johannes first about this, but I personally think that yes, the balances should reflect only the visible tokens

- Is unlikely that somebody will set the "Safe" token as invisible, but if somebody does the token widget in the top bar and in the Home widget is still there. I'm gonna assume that is expected behaviour.

schmanu:
Hey @francovenica 

I addresses most issues:
1. I removed the button in mobile.
2. The widget now does not count hidden assets anymore.
3. I moved the tracking to how we track the totalBalance (useMetaEvents). That should fix these issues. Although I wasn't able to reproduce those issues :/ 

For the questions:
1. _Deselect all_ was implemented wrong. Now it only deselects tokens which are present in the opened Safe
2. After talking with product we agreed that the total balance should be adjusted. I added that now.
3. That is expected. The Safe token widget should not be affected by this.

francovenica:
Yes, all the issues were addressed. The GTM calls are also working fine (not 10 calls just by refreshing the page anymore)

src/components/balances/AssetsHeader/index.tsx
```diff
@@ -1,20 +1,42 @@
...
 
 import NavTabs from '@/components/common/NavTabs'
 import PageHeader from '@/components/common/PageHeader'
 import CurrencySelect from '@/components/balances/CurrencySelect'
 import { balancesNavItems } from '@/components/sidebar/SidebarNavigation/config'
 
-const AssetsHeader = ({ currencySelect = false }: { currencySelect?: boolean }): ReactElement => {
+import HiddenTokenButton from '../HiddenTokenButton'
+
+const AssetsHeader = ({
```
usame-algan, line 11: I think we can simplify this component by only passing `children?: ReactNode` instead of the other props since they are not being used in `AssetsHeader`. This would reduce the amount of nesting we have and we could include the child components directly in the balances page like this:

```
<AssetsHeader>
    <Box display="flex" flexDirection="row" alignItems="center" gap={1}>
      {hiddenAssets && (
        <HiddenTokenButton showHiddenAssets={showHiddenAssets} toggleShowHiddenAssets={toggleShowHiddenAssets} />
      )}
      <CurrencySelect />
    </Box>
</AssetsHeader>
```

usame-algan: As discussed I will push this small refactor

src/hooks/useVisibleBalances.ts
```diff
@@ -0,0 +1,59 @@
...
+      .toString(),
+    PRECISION,
+  )
+
+  const balanceItemsVisible = balances.items.filter(
+    (balanceItem) => !hiddenAssets.includes(balanceItem.tokenInfo.address),
+  )
+
+  return { fiatTotal: fiatTotalVisible, items: balanceItemsVisible }
+}
```
katspaugh, line 42: Could you please split this into two functions, `filterHiddenTokens` and `getFilteredTotal`?

katspaugh (approved):
Looks good!

Btw according to the SafeBalanceResponse type, `items` and `fiatTotal` are the only fields in this object, so you don't need the `...balances`.

schmanu:
@katspaugh 
The names are a bit misleading there the returned object from `const balances = useBalances` will return
```ts
{
 balances: SafeBalanceResponse
  loading: boolean
  error?: string
}
```

So I spread to use the `loading` and `error` props from the `useBalances` hook as the visibleBalances would be loading / erroneous too.

PR #1379, opened 2022-12-15
Fix: move token tracking to meta events

katspaugh:
## What it solves

Tokens were only tracked on the Balances page although we have them on any Safe page.

I moved it to `useMetaEvents` and moved the hook itself to a separate file.

## How to test
Open any Safe and observe the `tokens` event fired immediately w/o going to the Balances page.


iamacook (approved):

iamacook:
Can confirm that the event is dispatched as soon as the Safe is opened.

event

PR #1380, opened 2022-12-16
Chore: create a release when a PR is merged to main

katspaugh:
## What it solves

I've added an action to automatically tag a merged RC and make a release with the same text as the release PR.

I've tested the new action in a private repo.


.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,44 @@
...
+        if: github.event.pull_request.merged == true
+        with:
+          fetch-depth: 0
+
+      - name: Extract version
+        if: github.event.pull_request.merged == true
+        id: version
+        run: |
+          NEW_VERSION=$(node -p 'require("./package.json").version')
+          echo "::set-output name=version::v$NEW_VERSION"
```
usame-algan, line 24: Would this be the same as `echo "version=v$NEW_VERSION" >> $GITHUB_OUTPUT` like described [here](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter)?

.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,44 @@
...
+          echo "${{ github.event.pull_request.body }}" > CHANGELOG.md
+
+      - name: Create a git tag
+        if: github.event.pull_request.merged == true
+        run: git tag $NEW_VERSION && git push --tags
+        env:
+          NEW_VERSION: ${{ steps.version.outputs.version }}
```
usame-algan, line 31: Does `$NEW_VERSION` contain the prefix `v` here even if we set the environment variable after?

katspaugh: Ah, right, good catch. Let me add a `v`.

.github/workflows/tag-release.yml
katspaugh: I'll try this syntax

.github/workflows/tag-release.yml
katspaugh: Checked, it works correctly like it is.

usame-algan: Testing this in a private repo it seems to include the prefix

usame-algan (approved):

PR #1382, opened 2022-12-16
fix: expire/clear pairing session after 24h

### LINKED ISSUE #701: Webapp doesn't expire "Connect to mobile" connection
## Description

WalletConnect connections ("Connect to Web" on Mobile) are not removed after more than 24h of inactivity

## Environment
 - Browser: Chrome
 - Wallet: Gnosis Safe Mobile
 - Chain: Rinkeby

## Steps to reproduce
 1. On the iOS mobile app: Settings -> Connect to Web -> Scan Code 
 2. Scan "Connect to Mobile" on the web app
 3. Wait 24 hours (or adjust the expiry date on mobile) and make sure the mobile app is not able to tell the web app to drop the connection (use flight mode)
 4. Make sure connection is not available on mobile anymore
 5. Open web app 

## Expected result

Expected behavior: The connection should not be available anymore on web. At least it should be terminated, when it cannot be used

## Obtained result
Observed behavior: The webapp tells me to sign the tx, but the mobile phone never gets the message (because the connection has been dropped on mobile.

This can happen weeks after the connection was established. When at least after 24h it should be expired.

## Screenshots
image

iamacook:
## What it solves

Resolves #701

## How this PR fixes it

Previously cached pairing sessions (if last connected >24h) are no longer automatically connected to/are cleared.

## How to test it

Connect to the Safe UI via pairing. Increase system time by 24h (+1m) and observe that reconnection does not automatically happen. The `SAFE_v2__pairingConnector` `localStorage` item should also no longer be present.

When returning to the Safe UI within a 24h window, the pairing session should connect as before.

### DIFF
--- a/src/services/pairing/utils.ts
+++ b/src/services/pairing/utils.ts
@@ -1,6 +1,10 @@
+import { addDays, isBefore } from 'date-fns'
+import type { IWalletConnectSession } from '@walletconnect/types'
 import type WalletConnect from '@walletconnect/client'
 
 import { CGW_NAMES, WALLET_KEYS } from '@/hooks/wallets/wallets'
+import local from '@/services/local-storage/local'
+import { PAIRING_MODULE_STORAGE_ID } from '@/services/pairing/connector'
 
 export const formatPairingUri = (wcUri: string) => {
   const PAIRING_MODULE_URI_PREFIX = 'safe-'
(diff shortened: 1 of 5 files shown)


src/services/pairing/utils.ts
```diff
@@ -28,3 +32,28 @@ export const killPairingSession = (connector: InstanceType<typeof WalletConnect>
 export const isPairingSupported = (disabledWallets?: string[]) => {
   return !!disabledWallets?.length && !disabledWallets.includes(CGW_NAMES[WALLET_KEYS.PAIRING] as string)
 }
+
+export const _hasPairingSessionExpired = (session: IWalletConnectSession): boolean => {
```
usame-algan, line 36: nit: I would name this `_hasExpiredPairingSession` or `_isPairingSessionExpired`

usame-algan (approved):

PR #1386, opened 2022-12-16
fix(safe-apps-e2e): Add timeout for avoid flakiness when transaction modal is showed

yagopv:
## What it solves

Flakiness in transaction modal e2e test


katspaugh (approved):
Thanks!

PR #1388, opened 2022-12-16
Bug: Bookmark button not shown in search results for safe apps

mikhailxyz:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/1387

## How this PR fixes it

It passes pin function and pinned safe app ids

## How to test it
1. Open safe apps page
2. Jot down a search query
3. Search results app cards should have a pin button
4. Click the button
5. Remove the search query
6. Confirm the app is pinned

### DIFF
--- a/src/components/safe-apps/SafeAppList.tsx
+++ b/src/components/safe-apps/SafeAppList.tsx
@@ -74,7 +74,13 @@ const SafeAppList = () => {
       pageBody = <SafeAppsSearchPlaceholder searchQuery={searchQuery} />
     } else {
       pageBody = (
-        <SafeAppsSection title={`Search results (${filteredApps.length})`} apps={filteredApps} allApps={allSafeApps} />
+        <SafeAppsSection
+          title={`Search results (${filteredApps.length})`}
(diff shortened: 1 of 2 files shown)


mikhailxyz:

yagopv (approved):

JagoFigueroa:
Buen señor! A small issue. We should not allow to pin custom apps when searching as it is not possible to do so from the main screen:

Screenshot 2022-12-16 at 16 37 41

Screenshot 2022-12-16 at 16 38 01

mikhailxyz:
now I remember why I didn't do this in the first place 🙄

JagoFigueroa:
I will open a separate ticket related to this as something similar happens with custom apps on the main screen as well. Custom apps are now available available under 'All apps' with possibility to pin them as well but in reality this is not possible as well. We might need to rethink how custom apps are tackled on the interface (a discussion for another time, just as a fyi)

mikhailxyz:
Closing this because it will be tackled with the redesign of the Safe Apps list

PR #1390, opened 2022-12-16
chore: added Dockerfile

DaniSomoza:
## What it solves

Added `Dockerfile` to Resolve safe-global/safe-infrastructure#58

## How this PR fixes it

based in [this template](https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile) recommended in [the nextjs docs](https://nextjs.org/docs/deployment#docker-image)

Added a `Dockerfile` file

## How to test it

To test the image run:

- Create the docker image: `docker build -t web-core .`
- Run the docker image: `docker run -p 3000:3000  web-core`
- Open your browser on `http://127.0.0.1:3000`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,8 +21,6 @@ FROM node:16-alpine AS runner
 WORKDIR /app
 
 ENV NODE_ENV production
-# Uncomment the following line in case you want to disable telemetry during runtime.
-ENV NEXT_TELEMETRY_DISABLED 1
 
 RUN addgroup --system --gid 1001 nodejs
 RUN adduser --system --uid 1001 nextjs
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,39 @@
+# Build the source code
+FROM node:16-alpine AS builder
+RUN apk add --no-cache libc6-compat git python3 py3-pip make g++
+WORKDIR /app
+COPY . .
+
+# install deps
+RUN yarn install
+
+# Next.js collects completely anonymous telemetry data about general usage.
+# Learn more here: https://nextjs.org/telemetry
+# Uncomment the following line in case you want to disable telemetry during the build.
+ENV NEXT_TELEMETRY_DISABLED 1
+
+RUN yarn build
+RUN yarn export
+
+
+# Production image, copy out dir and run serve
+FROM node:16-alpine AS runner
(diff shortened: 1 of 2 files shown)


katspaugh (dismissed):
Nice!

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN yarn export
+
+
+# Production image, copy out dir and run serve
+FROM node:16-alpine AS runner
+WORKDIR /app
+
+ENV NODE_ENV production
+# Uncomment the following line in case you want to disable telemetry during runtime.
+ENV NEXT_TELEMETRY_DISABLED 1
```
katspaugh, line 25: Do we need it a second time?

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+RUN yarn build
+RUN yarn export
+
+
+# Production image, copy out dir and run serve
+FROM node:16-alpine AS runner
+WORKDIR /app
+
+ENV NODE_ENV production
```
katspaugh, line 23: This should probably go before yarn build?
Also what about all the other env variables? Like the Infura key etc

Dockerfile
DaniSomoza: yep it seems that its needed, I used [this repo as a template](https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile) because its recommended in[ the nextjs docs](https://nextjs.org/docs/deployment#docker-image)

Dockerfile
DaniSomoza: variables are loaded from your `.env` file.

I updated my `NEXT_PUBLIC_INFURA_TOKEN` to make sure that it updates properly and its working fine:

Captura de pantalla 2022-12-19 a las 10 24 57
Captura de pantalla 2022-12-19 a las 10 25 35

DaniSomoza: And [the fixed value `ENV NODE_ENV production` in the Dockerfile](https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile#L37) its seems just to make sure that the node runs in production mode

Dockerfile
katspaugh: I mean it's already set on line 13 and we don't call any Next scripts after this.

PR #1391, opened 2022-12-19
Fix: psa banner fixed on top

### LINKED ISSUE #1202: Banner is displayed over the app header

katspaugh:
## What it solves

Resolves #1202

## How this PR fixes it
The banner is now statically positioned on top of the header. When it's opened, the global header-height variable is set to 104px. When it's closed, it's set back to 52px.

Screenshot 2022-12-19 at 16 05 03

### DIFF
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -41,16 +41,20 @@ button {
 }
 
 :root {
-  --header-height: 52px;
+  --header-inner-height: 52px;
+}
+
+:root {
+  --header-height: var(--header-inner-height);
 }
 
 .psaBanner:root {
-  --header-height: 104px;
+  --header-height: calc(var(--header-inner-height) * 2);
 }
 
 @media (max-width: 600px) {
   .psaBanner:root {
-    --header-height: 120px;
+    --header-height: calc(var(--header-inner-height) * 2 + 16px);
   }
 }
 
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -52,6 +52,7 @@ button {
   --header-height: calc(var(--header-inner-height) * 2);
 }
 
+/* On mobile, the banner text is two lines + padding */
 @media (max-width: 600px) {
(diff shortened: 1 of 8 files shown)


src/styles/globals.css
```diff
@@ -41,7 +41,21 @@ button {
...
+  --header-height: var(--header-inner-height);
+}
+
+.psaBanner:root {
+  --header-height: calc(var(--header-inner-height) * 2);
+}
+
+@media (max-width: 600px) {
+  .psaBanner:root {
+    --header-height: calc(var(--header-inner-height) * 2 + 16px);
```
usame-algan, line 57: This seems to be very specific to the current banner that takes up 2 lines on smaller screens but it would not work for longer content. I would suggest adding a comment here to explain that.

We could also keep the same height but make the inner content scrollable to cover smaller screens.

katspaugh: Actually on my phone it takes 3 lines so this magic number won’t work. Although the worst that happens is it’s still partially covering some content.  
I don’t like the scroll idea tbh.

katspaugh: Added a comment.

usame-algan:
There is an issue on the Transaction page when scrolling down, the navbar is only half-visible. I think we need to use the header-height var there too.
Screenshot 2022-12-20 at 09 38 59

katspaugh:
@usame-algan good catch, I've adjusted the PageHeader to use the global var as well.

usame-algan (approved):

francovenica:
How many rows of text is the average for this banners? When there are 3 lines or more it start to "eat" the page 
image

katspaugh:
The text is currently hardcoded, so 1 line.

francovenica:
Ok then. Looks good to me

PR #1401, opened 2022-12-20
Feat: add PWA support (next-pwa)

katspaugh:
## What it solves

Adds a framework to easily configure service workers for offline support.

See https://github.com/shadowwalker/next-pwa

## How to test

* Load the app
* Go to Chrome devtools, Network panel
* Set the network to Offline
* Navigate to different pages within the app – they should still open


iamacook:
Have you noticed that this causes all bundles to download? It crashed Chrome a couple of times whilst attempting to record this.

pwa

katspaugh:
That's no bueno, I'll check if there's an option to disable this.

usame-algan (approved):
Looks good 👍 
I tested opening different views, setting offline and hard-reloading the page. After a short while the service worker loads the cached page with images and content.

If I never visited that page before it will show a network error after reloading. We could add an offline page for that case as described [here](https://github.com/shadowwalker/next-pwa#offline-fallbacks)

Another optimization we can do is adding [head meta tags and a manifest file](https://github.com/shadowwalker/next-pwa#step-2-add-manifest-file-example). This is mostly for mobile phones I think.

katspaugh:
A manifest.json is already in place.

Issue #1404, opened 2022-12-20
Remove old safe creation and AB test


usame-algan:
## Description

The A/B test results for the new safe creation flow were positive so we want to clean up the old components and the A/B test setup.

## Requirements

- Copy existing tests from `src/components/create-safe` into the new directory `src/components/new-safe`
- Remove `src/components/create-safe`
- Remove AB test related logic (this will most likely be replaced at a later time with something like Google Optimize)

katspaugh:
Only `/new-safe/create` and `/new-safe/add` should remain.

Issue #1203, opened 2022-11-21
Add loading icon when the address check is in progress


liliya-soroka:
## Bug description
Add loading icon when the address check is in progress .
It's really useful when the back-end is slow for some reason.
We had green loading icon in safe-react
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to add existing safe flow on staging CGW ( it's slow now)
 2. add valid address
Current result: Continue button stays grey for a long time and it seems that the UI does nothing

## Expected result
add loading icon to the field to show that the address check is in progress
## Screenshots
<img width="852" alt="image" src="https://user-images.githubusercontent.com/338622/203028374-5db1f4ce-6e4a-459f-9e29-1f7d2c200f78.png">

usame-algan:
I tried reproducing it without success. There is a loading icon in the name field when ENS is resolving but the Continue button is only disabled if there is an invalid address so it should not be affected by it. The valid address check is done without anything asynchronous going on so we don't have a loading state there.

francovenica:
I've checked this in the new Add safe workflow and there it works fine, the load icon shows up until the safe is validated.
I'll close this one

PR #1405, opened 2022-12-20
feat: redesign Load Safe

### LINKED ISSUE #964: [Add Safe] Implement redesign
## What is the feature about

Implements the redesign of the Add Safe flow. Blocked by #864 

## The list of requirements

- Should reuse the components from #961, #962, #963
- Should reuse the new stepper from #931 

## Designs/sketches
Screenshot 2022-12-16 at 13 53 33
Screenshot 2022-12-16 at 13 54 52
Screenshot 2022-12-16 at 13 55 13

## Links

Figma: https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=468%3A11746

usame-algan:
## What it solves

Resolves #964 

## How this PR fixes it

- Existing components from `src/components/load-safe` were copied into the new structure within `src/components/new-safe/load` with minor layout adjustments
- The first and second step of the old flow are combined
- Removes the old routes `/open` and `/load` and their pages
- Uses the new routes `/new-safe/create` and `/new-safe/load` everywhere
- Restructures the `new-safe` directory

## ToDos

- [x] Move `OwnerRow` into the parent directory
- [x] Remove old load safe components inside `src/components/load-safe`
- [x] Find a solution for cut-off safe addresses in the new design

## How to test it

1. Open the Safe
2. Navigate to the welcome page
3. Click on "Add existing Safe"
4. Observe the new design
5. Navigate to a specific safe
6. Open the safe list sidebar and click on "Add Safe"
7. Observe being navigated to the load safe flow and the safe address being pre-filled

## Screenshots
Screenshot 2022-12-20 at 14 59 24
Screenshot 2022-12-21 at 11 49 22
Screenshot 2022-12-20 at 15 05 38


src/components/new-safe/load/steps/Step2/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { useAppDispatch } from '@/store'
+import { useRouter } from 'next/router'
+import { addOrUpdateSafe } from '@/store/addedSafesSlice'
+import { defaultSafeInfo } from '@/store/safeInfoSlice'
+import { upsertAddressBookEntry } from '@/store/addressBookSlice'
+import { LOAD_SAFE_EVENTS, trackEvent } from '@/services/analytics'
+import { AppRoutes } from '@/config/routes'
+import ReviewRow from '@/components/new-safe/ReviewRow'
+
+const LoadSafeStep2 = ({ data, onBack }: StepRenderProps<LoadSafeFormData>) => {
```
katspaugh, line 21: Not a fan of numeric names, tbh. The name should reflect what the component does, not the position this component is included at.

katspaugh:
Screenshot 2022-12-20 at 15 53 50

☝️ these inputs look read-only until focused.

katspaugh (approved):
The code looks good, and in my testing, it also works fine.

I don't like calling components Step0, Step1 etc, but it's the same in the creation...

src/components/new-safe/load/steps/Step2/index.tsx
usame-algan: I agree, lets change it for load safe now since the old components are deleted and for create safe with #1404

katspaugh (approved):
👍

katspaugh:
Screenshot 2022-12-22 at 09 00 55

☝️ can we stretch the address field to the full width?

katspaugh:
I've stretched the address field like so:

Screenshot 2022-12-27 at 09 34 05
Screenshot 2022-12-27 at 09 34 14

src/components/new-safe/load/steps/SetAddressStep/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+  // Address book, ENS, mnemonic
+  const fallbackName = name || ens || randomName
+
+  const validateSafeAddress = async (address: string) => {
+    if (addedSafes && Object.keys(addedSafes).includes(address)) {
+      return 'Safe is already added'
+    }
+
+    try {
+      await getSafeInfo(currentChainId, address)
```
schmanu, line 74: async validation is not handled that well here.
If `getSafeInfo` takes a bit longer (which it sometimes does) the Next button is disabled without explanation why. 
We should add a spinner to the `AddressInput` to have an indicator that the validation is in progress. 

This would also solve #1203

schmanu: I fixed it in a separate fix PR. So nevermind this comment :)

francovenica:
Issue:

When using and ENS name, after the name is translated to a valid address, is expected to also be autofilled in the "Name" input.
The current result is that the generic random name is still used to fill the input
Try "goerlisafe.eth" for testing

image

francovenica:
Correction. I've checked and the ENS autofill name is broken in prod as well. It would be great if that can be fixed here, but is not a blocker to pass this ticket. 
Let me know if that can be fixed here or if I should open a new ticket

usame-algan:
> Correction. I've checked and the ENS autofill name is broken in prod as well. It would be great if that can be fixed here, but is not a blocker to pass this ticket. Let me know if that can be fixed here or if I should open a new ticket

For me it resolves the ENS and autofills it as a placeholder to the name input. I think this is the intended behaviour as it was changed to be a placeholder in #353

francovenica:
I didn't noticed that the name wasn't properly set. Now it works fine

PR #1406, opened 2022-12-20
Docs: add a couple more env vars

katspaugh:
## What it solves
Some env vars weren't documented.

Also I removed the fallback value for the WC bridge and added it to the repo secrets instead.


.env.example
```diff
@@ -1,28 +1,28 @@
-### Required variables ###
+## Infura token (required)
 NEXT_PUBLIC_INFURA_TOKEN=
 NEXT_PUBLIC_SAFE_APPS_INFURA_TOKEN=
 
+## CGW
+GATEWAY_URL_PRODUCTION=
+GATEWAY_URL_STAGING=
```
usame-algan, line 7:
```suggestion
NEXT_PUBLIC_GATEWAY_URL_PRODUCTION=
NEXT_PUBLIC_GATEWAY_URL_STAGING=
```

usame-algan (approved):

PR #1411, opened 2022-12-21
Fix: app iframe permissions by domain

### LINKED ISSUE #1407: WalletConnect camera not working due to Iframe permissions overwriting the ones for safe apps
## Bug description
At the moment we are using specific iframe permissions to allow or deny actions like camera usage on WalletConnect safe app. The problem is that iframe permissions are now coming with a general allow for all which overlaps with our own permission system for safe apps and replaces it.

In the case of WC app, it is no longer possible for us to use the camera to scan QR codes because no specific camera permissions are being detected.

## Steps to reproduce
Go to WC safe app and click to open the camera. If camera permissions were already given beforehand (before release 1.2.0 at least I would say) everything should work as usual and reflect iframe permissions as the one seen on my screenshot for the expected result.

1. Go to settings -> permissions and remove the permissions for this app if any.

2. Open WalletConnect app and say yes to camera permissions (plus make sure they are also allowed on the browser side).

3. Click to open the camera.

## Expected result
Camera working as expected.

This is how permissions were handled before:
Screenshot 2022-12-20 at 14 44 38

## Obtained result
Camera not working.

This is how permissions are handled now:
Screenshot 2022-12-20 at 14 38 36

katspaugh:
## What it solves

Resolves #1407

## How this PR fixes it
The permissions were saved for the entire domain but checked for a specific path.

It's now checking the permissions also by domain.

## How to test it
* Open the WC app
* Allow camera
* Check that the iframe has an `allow="camera"` attribute.


yagopv (approved):

yagopv:
Found a [different issue](https://github.com/safe-global/web-core/issues/1416) testing this, let's merge this as it fix the major one and Safe Apps team will take care of the new one

yagopv:
Now is working fine. Tested it

PR #1412, opened 2022-12-21
Fix: restore default WC_BRIDGE

katspaugh:
## What it solves

Restored the default value for the WC bridge.


usame-algan (approved):

usame-algan:
verified. I can see the walletconnect and mobile pairing option if I add the `NEXT_PUBLIC_WC_BRIDGE` env var. Once I remove and restart both options are not visible in the UI anymore.

usame-algan (approved):

PR #1415, opened 2022-12-21
Chore: extract the build into a composite action

katspaugh:
## What it solves
The same build step is done in four workflows. Let's extract it into a reusable action.


usame-algan (approved):
Interesting that we need to parse the secrets as json. Have you tried `secrets: inherit` as described [here](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow)? It says "Workflows that call reusable workflows in the same organization or enterprise can use the inherit keyword to implicitly pass the secrets."

katspaugh:
`inherit` is a feature request, it doesn't support that.

katspaugh:
Inherit somehow didn't work for me. Resetting back to the prev commit.

PR #1417, opened 2022-12-21
fix: check for compatibility handler

### LINKED ISSUE #1413: [EIP-1271] Can't get off chain modal to show up
## Bug description
Tried to create an off-chain type of signature but I always get the on-chain modal to sign

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: goerli
 - Env: https://eip_1271--webcore.review-web-core.5afe.dev/gor:0x8FE062e9b0865dc47552415a745CEEe392eFb6C3
 - WC app used as custom: https://pr608--safereactapps.review-react-hr.5afe.dev/wallet-connect/

## Steps to reproduce
 1. Started the Test app, http://localhost:5173/
 2. Opened the epic branch on web app
 3. Added the Wallet connect app as custom app
 4. Connected both apps
 5. Enabled off-chain type of message
 6. Wrote a message and used the "sign" button
 7. Checked the web app waiting for the offchain type of modal

## Expected result
Get the offchain type of modal

## Obtained result
I always get the on-chain type of modal, and an error in the console (no errors in the network tab)

## Screenshots
image

schmanu:
## What it solves
Does not run into an error if the compatibility handler is not set.

Resolves #1413 

## How this PR fixes it
- Handles the case when the fallback handler is `null`
- Adds jest test for the supportsEip1271 function

## How to test it
- Connect using our test dapp with this PR deployment using the patched wallet connect version
- Using a safe without compatiblity fallback handler
- Observe warning in log when pressing "enable off-chain signing"

### DIFF
--- a/src/components/safe-apps/AppFrame/index.tsx
+++ b/src/components/safe-apps/AppFrame/index.tsx
@@ -121,12 +121,17 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement 
       }
     },
     onSetSafeSettings: (safeSettings: SafeSettings) => {
+      const isEIP1271Supported = supportsEIP1271(safe)
       const newSettings: SafeSettings = {
         ...settings,
-        offChainSigning: supportsEIP1271(safe) && !!safeSettings.offChainSigning,
+        offChainSigning: isEIP1271Supported && !!safeSettings.offChainSigning,
       }
 
       setSettings(newSettings)
+
+      !isEIP1271Supported &&
+        safeSettings.offChainSigning &&
+        console.warn('The connected Safe does not support off-chain signing.')
 
       return newSettings
     },
--- a/src/components/safe-apps/AppFrame/index.tsx
+++ b/src/components/safe-apps/AppFrame/index.tsx
@@ -129,9 +129,9 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement 
 
       setSettings(newSettings)
 
-      !isEIP1271Supported &&
-        safeSettings.offChainSigning &&
+      if (!isEIP1271Supported && safeSettings.offChainSigning) {
         console.warn('The connected Safe does not support off-chain signing.')
+      }
 
       return newSettings
     },
--- a/src/utils/safe-messages.ts
+++ b/src/utils/safe-messages.ts
@@ -3,7 +3,7 @@ import { gte } from 'semver'
 import type { SafeInfo, SafeMessage, EIP712TypedData } from '@safe-global/safe-gateway-typescript-sdk'
 
 import { hashTypedData } from '@/utils/web3'
-import { validateAddress } from './validation'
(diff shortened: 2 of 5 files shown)


src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -121,13 +121,18 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement
...
         ...settings,
-        offChainSigning: supportsEIP1271(safe) && !!safeSettings.offChainSigning,
+        offChainSigning: isEIP1271Supported && !!safeSettings.offChainSigning,
       }
 
       setSettings(newSettings)
 
+      !isEIP1271Supported &&
+        safeSettings.offChainSigning &&
+        console.warn('The connected Safe does not support off-chain signing.')
```
usame-algan, line 134: nit: I would use a normal `if` clause here for readability

src/utils/safe-messages.ts
```diff
@@ -47,8 +47,8 @@ export const supportsEIP1271 = ({ chainId, version, fallbackHandler }: SafeInfo)
 
   const isHandledByFallbackHandler = gte(version, EIP1271_FALLBACK_HANDLER_SUPPORTED_SAFE_VERSION)
   if (isHandledByFallbackHandler) {
-    const fallbackHandlerDeployment = getCompatibilityFallbackHandlerDeployment({ network: chainId, version })
-    return !!fallbackHandler.value && fallbackHandler.value === fallbackHandlerDeployment?.networkAddresses[chainId]
+    // We only check if any fallback Handler is set as we expect / assume that users who overwrite the fallback handler by a custom one know what they are doing
+    return fallbackHandler !== null && validateAddress(fallbackHandler.value) === undefined
```
usame-algan, line 51: Should this not be `validateAddress(fallbackHandler.value) !== undefined` so that `supportsEIP1271` returns true if there is a fallbackhandler address set?

schmanu: `validateAddress` will return `undefined` or a validation error `string` 
So if the return value is `!== undefined` it would mean that some validation errors exist and the address is not valid.
While return  value `=== undefined` means that no validation error exists and the adress is valid.

usame-algan: Ah I just realized `validateAddress` returns undefined if everything is ok with the address so it should be fine as it is.

usame-algan (approved):

src/utils/safe-messages.ts
schmanu: I will add a small util function "isValidAddress" so it reads better. This can be confusing.

francovenica:
Looks good. Used this safe:

https://eip_1271_fix_compatibility_check--webcore.review-web-core.5afe.dev/gor:0xC835087fC5A67d5D4612c582F9C308979001e2E3/apps?appUrl=https%3A%2F%2Fpr608--safereactapps.review-react-hr.5afe.dev%2Fwallet-connect%2F

image

PR #1420, opened 2022-12-21
Fix: cookies-policy "self"

katspaugh:
## What it solves

Beamer sets its cookies on the main domain instead of app.safe.global. This PR adds a CSP to limit cookies to the current subdomain.

### DIFF
--- a/src/config/securityHeaders.ts
+++ b/src/config/securityHeaders.ts
@@ -19,9 +19,10 @@ export const ContentSecurityPolicy = `
  };
  frame-src *;
  style-src 'self' 'unsafe-inline' https://*.getbeamer.com https://*.googleapis.com;
- font-src 'self' data:; 
+ font-src 'self' data:;
  worker-src 'self' blob:;
  img-src * data:;
+ cookies-policy 'self';
 `
   .replace(/\s{2,}/g, ' ')
   .trim()


katspaugh:
Closing, it didn't help with Beamer cookies.

schmanu:
I think there is no CSP directive for cookies (yet).

I investigated how beamer sets those cookies and it seems they set them via some `Beamer.setCookie` call.
Usually when [writing a new cookie]((https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#write_a_new_cookie)) via javascript there is an option to adjust the domain:
i.e.
`document.cookie = "_EXAMPLE_COOKIE=EXAMPLE_VALUE;  domain=app.safe.global"`

I tried that out and cookies I set this way did not appear on the top level domain.

Sadly I found no way to somehow enforce this and I also found no option in the [developer docs of Beamer](https://www.getbeamer.com/docs/#cookies) to set the domain for their cookies.

katspaugh:
I see, thanks for checking, Manu!

Beamer actually told me yesterday:

> Quick news: The tech team just enabled an option for cookies to be saved in your full (sub)domain rather than on your top domain. This change should be already in effect for cookies saved from now on

PR #1427, opened 2022-12-22
Fix: close connect popup on button click

### LINKED ISSUE #1367: [Mobile] Search bar in WalletConnect popup doesn't work
## Bug description
There's a search function in the WC popup on mobile, it's not clickable atm.

## Steps to reproduce
 1. Open the app on mobile
 2. Click on Connect wallet
 3. Try searching for something
 4. Keyboard isn't appearing

## Note
I already fixed this in the past, but it came back. **Has to do with z-indexes.**

katspaugh:
## What it solves

Resolves #1367

## How this PR fixes it

The popover was still open after clicking on Connect, covering the WC popup and blocking input clicks.

### DIFF
--- a/src/components/common/ConnectWallet/WalletDetails.tsx
+++ b/src/components/common/ConnectWallet/WalletDetails.tsx
@@ -6,7 +6,12 @@ import type { ConnectedWallet } from '@/services/onboard'
 import useConnectWallet from '@/components/common/ConnectWallet/useConnectWallet'
 
 const WalletDetails = ({ onConnect }: { onConnect?: (wallet?: ConnectedWallet) => void }): ReactElement => {
-  const handleConnect = useConnectWallet(onConnect)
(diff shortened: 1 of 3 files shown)


src/components/common/ConnectWallet/WalletDetails.tsx
```diff
@@ -6,7 +6,12 @@ import type { ConnectedWallet } from '@/services/onboard'
 import useConnectWallet from '@/components/common/ConnectWallet/useConnectWallet'
 
 const WalletDetails = ({ onConnect }: { onConnect?: (wallet?: ConnectedWallet) => void }): ReactElement => {
```
usame-algan, line 8: Checking the places we use `WalletDetails` it looks like `onConnect` doesn't have any params anymore so it could just be `onConnect?: () => void` now.

usame-algan (approved):

PR #1429, opened 2022-12-22
Fix: rm balance in review modal

### LINKED ISSUE #1428: Remove ETH balance from the Send modal
Showing the ETH balance of the Safe itself is not useful, since it's not the Safe who pays the gas fees.

katspaugh:
## What it solves

Resolves #1428 

## Screenshot

Screenshot 2022-12-22 at 11 35 26


usame-algan (approved):
I think this was something we just copied from safe-react but I agree it doesn't provide any useful info at that step of the modal.

PR #1433, opened 2022-12-22
fix: Adds icon to empty address book

### LINKED ISSUE #1192: Empty state for the addressbook is missed
## Bug description
Empty state for the address book is missed
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to the empty address book
 Current result: 
image

## Expected result
image

usame-algan:
## What it solves

Resolves #1192 

## How this PR fixes it

Displays an illustration if the address book is empty

## How to test it

1. Open a safe without any address book entries
2. Observe an icon in the center

## Screenshots

Screenshot 2022-12-22 at 13 30 31

Screenshot 2022-12-22 at 13 30 40


next.config.mjs
```diff
@@ -44,7 +44,7 @@ const nextConfig = {
           loader: '@svgr/webpack',
           options: {
             prettier: false,
-            svgo: true,
+            svgo: false,
```
usame-algan, line 47: Had to disable svgo because it minifies clip-path ids so that there is only one clip-path allowed on a given page. Thats why no icon was visible on the address book page so far.

PR #1436, opened 2022-12-22
fix: Add owner icon to confirmations

### LINKED ISSUE #1279: Dashboard / Transaction queue: add the owners icon to the signatures element
## Bug description
For the design improvement and mobile team alignment I would like to add a owners icon to the signature element, and fix the colours at the element in both light and dark mode
Figma: https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=1%3A3&t=WbjFLE0kr1XPch41-1
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to

usame-algan:
## What it solves

Resolves #1279 

## How this PR fixes it

- Displays the owner icon for transactions on the dashboard 
- Uses a lighter green for the confirmations background in the dashboard according to the design
- Changes owner icon color in the queue to be consistent with the text color 

## How to test it

1. Open a safe with a queued transaction
2. Navigate to the dashboard
3. Observe the owner icon next to the number of confirmations
4. Navigate to the Queue
5. The owner icon should have the same color as the text next to it

## Screenshots
Screenshot 2022-12-22 at 14 14 08
Screenshot 2022-12-22 at 14 15 48
Screenshot 2022-12-22 at 14 15 55


src/components/dashboard/PendingTxs/PendingTxListItem.tsx
```diff
@@ -35,7 +36,15 @@ const PendingTx = ({ transaction, url }: PendingTxType): ReactElement => {
           <Box gridArea="confirmations">
             {isMultisigExecutionInfo(transaction.executionInfo) ? (
               <Box className={css.confirmationsCount}>
-                {`${transaction.executionInfo.confirmationsSubmitted}/${transaction.executionInfo.confirmationsRequired}`}
+                <SvgIcon
+                  component={OwnersIcon}
+                  inheritViewBox
+                  fontSize="small"
+                  sx={{ color: ({ palette }) => palette.text.primary }}
```
katspaugh, line 43: Can this be inside the icon itself?

usame-algan: I tried but it overrides the other places where we use this icon.

In the queue the icon is either grey or black/green.
In the dashboard its always black/white.

Might be a good idea to make this more consistent but maybe there is a reason the icons + text are not green on the dashboard in dark mode. cc @liliiaorlenko

katspaugh (approved):
LGTM but tests are failing.

francovenica:
1 - Now that the background color is a dimmer green, it is lost when you hover the transaction. In the old version even by hovering you could still see the background greend color

Old:
image

New:
image

2 - In the design, even in dark mode the background color shoulud be a dim color of green:
Currently implemented:
image

Design:
image

usame-algan:
> 1 - Now that the background color is a dimmer green, it is lost when you hover the transaction. In the old version even by hovering you could still see the background greend color

I reverted the color changes and suggest we tackle it as a separate task after aligning with design.

francovenica:
Ok, so only the Owner Icon is being take in account for this ticket, and the icon looks good.
We will check colors in a different ticket then

PR #1440, opened 2022-12-22
Bug: Custom app warning flash

mikhailxyz:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/1326 and https://github.com/safe-global/safe-react-apps/issues/619

## How this PR fixes it

By checking that the remote app list was loaded before checking if the app exists there

## How to test it

Refresh a non-custom app and confirm the warning isn't there

### DIFF
--- a/src/pages/apps.tsx
+++ b/src/pages/apps.tsx
@@ -19,7 +19,7 @@ const Apps: NextPage = () => {
   const router = useRouter()
   const chainId = useChainId()
   const [appUrl, routerReady] = useSafeAppUrl()
-  const { remoteSafeApps } = useSafeApps()
+  const { remoteSafeApps, remoteSafeAppsLoading } = useSafeApps()
   const { isLoading, safeApp } = useSafeAppFromManifest(appUrl || '', chainId)
   const { addPermissions, getPermissions, getAllowedFeaturesList } = useBrowserPermissions()
   const origin = getOrigin(appUrl)
@@ -37,6 +37,7 @@ const Apps: NextPage = () => {
     permissions: safeApp?.safeAppsPermissions || [],
     addPermissions,
     getPermissions,
+    remoteSafeAppsLoading,
   })
 
   if (!routerReady || isLoading) {
--- a/src/pages/apps.tsx
+++ b/src/pages/apps.tsx
@@ -37,7 +37,7 @@ const Apps: NextPage = () => {
     permissions: safeApp?.safeAppsPermissions || [],
     addPermissions,
     getPermissions,
-    remoteSafeAppsLoading: remoteSafeAppsLoading || remoteSafeApps.length === 0,
+    remoteSafeAppsLoading,
   })
 
   if (!routerReady || isLoading) {
--- a/src/pages/apps.tsx
+++ b/src/pages/apps.tsx
@@ -37,7 +37,7 @@ const Apps: NextPage = () => {
     permissions: safeApp?.safeAppsPermissions || [],
     addPermissions,
     getPermissions,
-    remoteSafeAppsLoading,
+    remoteSafeAppsLoading: remoteSafeAppsLoading || remoteSafeApps.length === 0,
   })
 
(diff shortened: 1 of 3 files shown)


yagopv (approved):

mikhailxyz:

mikhailxyz:
looks like the CLA ting isn't working

JagoFigueroa:
Buen señor! The cabronazo still pops up for me 😢 I even double checked with a couple of people and for some is there and for others it is not. Any idea? :pepe-scream:

https://user-images.githubusercontent.com/92332710/209330214-4c02a1ea-ebf3-4269-9a63-98ac8de8b203.mp4

DmitryBespalov:
recheck

mikhailxyz:
@JagoFigueroa señor please recheck :)

JagoFigueroa:
Todo bueno ahora, gracias!

src/pages/apps.tsx
```diff
@@ -37,6 +37,7 @@ const Apps: NextPage = () => {
     permissions: safeApp?.safeAppsPermissions || [],
     addPermissions,
     getPermissions,
+    remoteSafeAppsLoading: !remoteSafeAppsLoading || remoteSafeApps.length === 0,
```
katspaugh, line 40: `remoteSafeAppsLoading: !remoteSafeAppsLoading`? 👀

mikhailxyz: Ooops, don't know what I was thinking. thanks!

src/pages/apps.tsx
```diff
@@ -37,6 +37,7 @@ const Apps: NextPage = () => {
     permissions: safeApp?.safeAppsPermissions || [],
     addPermissions,
     getPermissions,
+    remoteSafeAppsLoading: remoteSafeAppsLoading || remoteSafeApps.length === 0,
```
katspaugh, line 40: What if remoteSafeApps.length is legit 0?

mikhailxyz: I agree it's hacky, but I don't see any other way to detect a state where the apps aren't loaded yet and the request hadn't been triggered. We don't have an environment with 0 safe apps, at least "native" ones are available on any env.

mikhailxyz: The ultimate solution is to refactor from using boolean variables to determine request state to a state machine with clear status ("not loaded", "loading", "loaded", error"), but since it uses an app-wide hook that's too much of a refactor at this time.

katspaugh: Shouldn’t remoteSafeApps be completely undefined prior to the request?

mikhailxyz: I'd would help in this particular case but on a global scale it's also a hacky solution

mikhailxyz: you'd still have to write clumsy stuff like `typeof remoteSafeApps !== 'undefined' && !loading && !error` to determine request's state + have a complexity of checking for undefined everywhere before accessing

katspaugh: You don’t need to use typeof, the variable is always defined. Hacky or not, it’s less wrong than checking for length.

mikhailxyz: What's wrong with checking for length if you factor in that we don't have a single environment with 0 apps and thus an array can't be empty if the list is loaded?

mikhailxyz: Actually, a fair argument for your solution would be interface works. We can't guarantee that forked environments will contain apps. I'm going on vacation for 2 weeks, I'll give a heads-up to the team to iterate on this PR.

katspaugh: I've made a commit to this branch. Please feel free to revert or adjust it (whoever overtakes this PR).

katspaugh:
@JagoFigueroa could you please take another look?

JagoFigueroa:
Sorry for the delay señores. 

Loading looking mucho bueno after the changes, thanks! 😄

PR #1442, opened 2022-12-22
fix: Max amount button when sending tokens

### LINKED ISSUE #1441: [Send Assets] Setting Max amount is not working
## Bug description

Introduced with #1346 

Clicking the "Max" button does nothing

## Steps to reproduce
 1. Go to a safe
 2. Create a new transaction
 3. Select a token
 4. Click the "Max" button

## Expected result

The maximum token amount should be inserted into the input field

## Obtained result

Nothing is inserted into the input field

usame-algan:
## What it solves

Resolves #1441 

## How this PR fixes it

- Removes the spending limit check from the early return statement

## How to test it

1. Open a safe
2. Create a new transaction
3. Select a token that doesn't have a spending limit
4. Press "Max"
5. Observe the max amount in the input field
6. Select a token that has a spending limit
7. Press "Max"
8. Observe the max spending limit amount in the input field

### DIFF
--- a/cypress/e2e/create_tx.cy.js
+++ b/cypress/e2e/create_tx.cy.js
@@ -34,8 +34,8 @@ describe('Queue a transaction on 1/N', () => {
     cy.get('input[name="tokenAddress"]').prev().click()
     cy.get('ul[role="listbox"]').contains('Görli Ether').click()
 
-    // Insert amount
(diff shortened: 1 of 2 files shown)


katspaugh (approved):

katspaugh:
Tested, it works now. ✅

PR #1448, opened 2022-12-23
Fix: disallow editing nonce for existing txs

### LINKED ISSUE #1364: [Replace tx] Edit nonce is available during signing "unsigned tx" in the queued list
## Bug description
Edit nonce is available during signing "unsigned tx" in the queued list
## Environment
 - Browser: Chrome
 - Wallet: MMI extension
 - Chain: Goerli

## Steps to reproduce
 1. Create a tx in the safe using MMI extension => unassigned tx is created in the queued list as Next tx- 

image

2. go to the web-core interface and start the signing process with the owner EOA
3. go to the advanced option
4. Edit nonce is available 
Current result: nonces of all  txs in the queued list is available for selecting 

5. select another nonce and finish signing
Current result: the original ts stays unassigned and a new tx appears under selected nonce

## Expected result
Do we want to allow "edit nonce" for unassigned tx? It's not allowed during signing existed tx in general.

## Screenshots
image

katspaugh:
## What it solves

Resolves #1364

## How this PR fixes it

As per @schmanu's suggestion, we check if a tx is already proposed (txId is there), and if so we don't allow editing the nonce.

## How to test it
* Create a tx either with MMI or a smart contract wallet
* Start signing it from another wallet – the nonce should **not** be editable

Screenshot 2022-12-23 at 09 25 12

### DIFF
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -93,8 +93,8 @@ const SignOrExecuteForm = ({
 
   // Estimating gas
   const isEstimating = willExecute && gasLimitLoading
-  // Nonce cannot be edited if the tx is already signed, or it's a rejection
-  const nonceReadonly = !!tx?.signatures.size || isRejection
+  // Nonce cannot be edited if the tx is already proposed, or signed, or it's a rejection
+  const nonceReadonly = !!txId || !!tx?.signatures.size || isRejection
 
   // Assert that wallet, tx and provider are defined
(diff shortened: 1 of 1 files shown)


usame-algan (approved):

francovenica:
Works fine, I don't have MMI installed, but I used safe-as-owner to test this.

Proposed a tx with a safe-as-owner (SafeA) of another safe (SafeB)
Disconnected with SafeA and connected with another owner of safeB, a regular MM account
Signed the proposed tx
Checked that the nonce value cannot be edited during the signing.

PR #1314, opened 2022-12-06
Chore: try a different CLA config

katspaugh:
## What it solves

Copies the action from https://github.com/safe-global/safe-ios/pull/2806

PR #1452, opened 2022-12-23
Revert "Chore: try a different CLA config (#1314)"

katspaugh:
The CLA action is really finicky and hard to maintain (needs a separate repo, an access token etc).

This solution relies on a PR template that already includes a CLA consent.


DmitryBespalov:
the other PR check is working now.

You can re-trigger CLA check by adding a `recheck`  comment to the PR

PR #1454, opened 2022-12-23
fix: Keyboard support

### LINKED ISSUE #1450: Missing keyboard support for some elements
## Bug description

- Some links/buttons have redundant nested links/buttons which are focused twice when tabbing through the page with a keyboard
- The Tab navigation is not usable with a keyboard

## Steps to reproduce
 1. Go to Safe
 2. Tab through the page
 3. Observe some buttons such as the blockscan button in the header are focused twice
 4. Try to go through the tab navigation with your keyboard
 5. Observe its not possible

## Expected result

All clickable elements should only be focused once

## Obtained result

Some clickable elements are focused more than once

usame-algan:
## What it solves

Resolves #1450 

## How this PR fixes it

- Shows hidden buttons when they are focused
- Fixes the Tab Navigation keyboard support
- Removes redundant links on the dashboard
- Disables keyboard support for safe app images that are embedded via iframe
- Only renders advanced details for transactions if its expanded

## How to test it

1. Open the Safe
2. Tab through the dashboard
3. Observe that each clickable element is only focused once
4. Navigate to the assets page
5. Tab through the page
6. Observe the Send buttons are visible on focus
7. Observe the Tab Navigation at the top of the page is usable via keyboard (arrow keys)
8. Navigate to the NFT Page
9. Observe that the Send Button is visible on focus
10. Navigate to the History
11. Tab through a transaction
12. Observe that advanced details are ignored in the tab order if they are not expanded


src/components/dashboard/SafeAppsDashboardSection/SafeAppsDashboardSection.tsx
```diff
@@ -41,13 +42,15 @@ const ExploreSafeAppsCard = () => {
   const safeAppsLink = `${AppRoutes.apps}?safe=${router.query.safe}`
 
   return (
-    <AppCardContainer url={safeAppsLink}>
```
katspaugh, line 44: Won't this make the card not clickable?

usame-algan: Yes, now only the button itself is clickable. Previously, both the button and the card were clickable which could be confusing especially when using the keyboard to navigate.

katspaugh: I thought it was a feature and a good one.

usame-algan: I can revert it but would suggest we look into a better solution e.g. converting the button into a title that is not focusable/clickable to avoid having the same user action twice.

src/components/sidebar/SidebarHeader/index.tsx
```diff
@@ -29,18 +28,6 @@ import Track from '@/components/common/Track'
 import { OVERVIEW_EVENTS } from '@/services/analytics/events/overview'
 import { SvgIcon } from '@mui/material'
 
-const HeaderIconButton = ({
```
DiogoSoaress, line 32: Why did we need to remove the abstraction?

src/components/notification-center/NotificationCenter/styles.module.css
```diff
@@ -1,7 +1,7 @@
 .bell {
   display: flex;
   justify-content: center;
-  height: 100%;
+  padding: 4px;
```
DiogoSoaress, line 4: Is the padding just to increase the clickable area? maybe we could make the whole square clickable?

src/components/balances/AssetsTable/styles.module.css
```diff
@@ -3,7 +3,8 @@
   transition: opacity 0.2s;
 }
 
-.container tr:hover td:last-of-type button {
+.container tr:hover td:last-of-type button,
+.container td:last-of-type button:focus-visible {
```
DiogoSoaress, line 7: Nice one!

DiogoSoaress (changes requested):

src/components/sidebar/SidebarHeader/index.tsx
usame-algan: After removing the redundant `a` element, the `IconButton` components were different for the 2 places. We could pass those props to the abstraction but imo its not worth the added complexity.

src/components/notification-center/NotificationCenter/styles.module.css
usame-algan: After adding the focus style I noticed that the area looked strange so I removed the full height and added a consistent padding. We could also change it to take up the full size but then I would suggest to do it for all header elements.

Before/After:
Screenshot 2023-01-03 at 11 33 07
Screenshot 2023-01-03 at 11 32 59

DiogoSoaress: Yeah, I think we can leave it per your change. I'm going to resolve this, it is non-blocking.

DiogoSoaress (approved):
LGTM

francovenica:
Navigation itself looks good, I can reach the whole site with it and with that the ticket would be fine to pass. I didn't found places where tabbing would select the same item twice.

PR #1456, opened 2022-12-27
Change license from MIT to LGPL

katspaugh:
## What it solves

Per @rmeissner's comment:

> we do not expect unconditional forking (that is why we have lgpl as a license and not mit)

This PR changes the project's license to [LGPL v3](https://www.gnu.org/licenses/lgpl-3.0.en.html).

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,21 +1,163 @@
-MIT License
+GNU Lesser General Public License
+=================================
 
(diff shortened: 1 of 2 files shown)


rmeissner:
This should be changed in the package.json too

katspaugh:
Good catch, updated there too.

rmeissner (approved):

PR #1460, opened 2022-12-27
refactor: rename ExternalLink prop

DiogoSoaress:
## What it solves
`suppressIcon` -> `noIcon`


katspaugh (approved):
Thanks! 🙇

katspaugh:
Tested, works the same as before ✅

PR #1462, opened 2022-12-27
Perf: extract redirect logic into a static HTML page

katspaugh:
## What it solves

According to the PageSpeed test, an initial redirect for the dynamic Safe address in the path takes about 30% of the page load time:
Screenshot 2022-12-27 at 14 38 38

Without the redirect, i.e. using `?safe=` directly:

Screenshot 2022-12-27 at 14 40 29

## How it solves it

While this PR doesn't eliminate the client-side redirect, it makes it much faster by using a very small HTML file.


public/_redirects
```diff
@@ -1,7 +0,0 @@
-/:safe/balances /balances?safe=:safe
-/:safe/balances/* /balances/:splat?safe=:safe
-/:safe/settings /settings?safe=:safe
-/:safe/settings/* /settings/:splat?safe=:safe
-/:safe/transactions/queue /transactions/queue?safe=:safe
-/:safe/address-book /address-book?safe=:safe
-/:safe/home /home?safe=:safe
```
katspaugh, line 7: This file isn't needed anymore, it was used only for CloudFlare.

katspaugh:
While this does make the page load much faster, Google doesn't like it. Closing.

PR #1464, opened 2022-12-28
fix: async validation loader in AddressInput

### LINKED ISSUE #1203: Add loading icon when the address check is in progress

schmanu:
## What it solves
Fixes async validation for AddressInput

Resolves #1203 

## How this PR fixes it
Shows a loading spinner while the form is validating.

## How to test it
- Find a address input with slow validation or add a way of throttling the validation (Slow resolving promise, network throttling)

### DIFF
--- a/src/components/common/AddressInput/index.test.tsx
+++ b/src/components/common/AddressInput/index.test.tsx
@@ -124,6 +124,31 @@ describe('AddressInput tests', () => {
     await waitFor(() => expect(utils.getByLabelText(`${TEST_ADDRESS_B} is wrong`, { exact: false })).toBeDefined())
   })
 
+  it('should show a spinner when validation is in progress', async () => {
+    const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms))
+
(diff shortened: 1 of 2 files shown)


DiogoSoaress (approved):
Cherry on top 🍒  would be both spinners being sync but not worth the extra effort I guess.

schmanu:
@DiogoSoaress 
The two spinners not in sync should also be the case when resolving ENS names. As it already displayed two spinners then.

I didnt find an easy way to achieve that. There is nothing in the MUI docs / examples.

PR #1465, opened 2022-12-28
[EIP-1271] support Safe's < 1.3.0

### LINKED ISSUE #1432: [EIP-1271] Verify signature works only on Goerli
## Bug description
Verify signature works only on Goerli 
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum , Gnosis Chain, Polygon ( chains available on staging back-end env)

## Steps to reproduce
 1. Create and fully confirm a message on the Ethereum/Gnosis chain/Polygon
 2. click "Verify signature" in the test Dapp
Current result: Unable to fetch message is displayed instead of the confirmation state

## Expected result
The verification should be done and the confirmation state should be displayed

schmanu:
## What it solves
For Safes `<1.3.0` a wrong Safe message hash is computed when signing off-chain messages.

Resolves #1432 

## How this PR fixes it
- Does not add the `chainId` into the `domain` part of the Safe message's typed data if Safe version < 1.3.0

## How to test it
- Testing is blocked by the tx service also computing a wrong Safe message hash when creating off-chain messages.

### DIFF
--- a/src/utils/__tests__/safe-messages.test.ts
+++ b/src/utils/__tests__/safe-messages.test.ts
@@ -8,8 +8,9 @@ const MOCK_ADDRESS = ethers.utils.hexZeroPad('0x123', 20)
 
 describe('safe-messages', () => {
   describe('createSafeMessage', () => {
-    it('should generate the correct types for a EIP-191 message', () => {
+    it('should generate the correct types for a EIP-191 message for >= 1.3.0 Safes', () => {
       const safe = {
+        version: '1.3.0',
         address: {
           value: MOCK_ADDRESS,
         },
@@ -34,8 +35,35 @@ describe('safe-messages', () => {
       })
     })
 
-    it('should generate the correct types for an EIP-712 message', () => {
+    it('should generate the correct types for a EIP-191 message for < 1.3.0 Safes', () => {
       const safe = {
+        version: '1.1.1',
+        address: {
+          value: MOCK_ADDRESS,
+        },
+        chainId: 1,
+      } as unknown as SafeInfo
+
+      const message = 'Hello world!'
+
+      const safeMessage = generateSafeMessageTypedData(safe, message)
+
+      expect(safeMessage).toEqual({
+        domain: {
+          verifyingContract: MOCK_ADDRESS,
+        },
+        types: {
+          SafeMessage: [{ name: 'message', type: 'bytes' }],
+        },
+        message: {
+          message: '0xaa05af77f274774b8bdc7b61d98bc40da523dc2821fdea555f4d6aa413199bcc',
+        },
+      })
+    })
+
+    it('should generate the correct types for an EIP-712 message for >=1.3.0 Safes', () => {
+      const safe = {
+        version: '1.3.0',
         address: {
           value: MOCK_ADDRESS,
         },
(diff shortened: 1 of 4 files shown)


src/utils/safe-messages.ts
```diff
@@ -38,7 +54,6 @@ export const generateSafeMessageHash = (safe: SafeInfo, message: SafeMessage['me
 
 export const supportsEIP1271 = ({ version, fallbackHandler }: SafeInfo): boolean => {
   const EIP1271_SUPPORTED_SAFE_VERSION = '1.0.0'
-  // From v1.3.0, EIP-1271 support was moved to the CompatibilityFallbackHandler
   const EIP1271_FALLBACK_HANDLER_SUPPORTED_SAFE_VERSION = '1.3.0'
```
DiogoSoaress, line 57: `EIP1271_FALLBACK_HANDLER_SUPPORTED_SAFE_VERSION` is now declared twice. The other constant `EIP1271_SUPPORTED_SAFE_VERSION` could be moved to the top of the file.

src/utils/safe-messages.ts
```diff
@@ -16,12 +22,22 @@ export const generateSafeMessageMessage = (message: SafeMessage['message']): str
...
-    domain: {
-      chainId: safe.chainId,
-      verifyingContract: safe.address.value,
-    },
+    domain: isHandledByFallbackHandler
+      ? {
+          chainId: chainId,
+          verifyingContract: address.value,
+        }
+      : { verifyingContract: address.value },
```
DiogoSoaress, line 40: could we pass
```suggestion
   domain: {
      chainId: isHandledByFallbackHandler ? chainId : undefined,
      verifyingContract: address.value,
    },
```

DiogoSoaress, line 40: could we pass
```suggestion
   domain: {
      chainId: isHandledByFallbackHandler ? chainId : undefined,
      verifyingContract: address.value,
    },
```

src/utils/__tests__/safe-messages.test.ts
```diff
@@ -122,6 +150,95 @@ describe('safe-messages', () => {
...
+        version: '1.1.1',
+        address: {
+          value: MOCK_ADDRESS,
+        },
+        chainId: 1,
+      } as unknown as SafeInfo
+
+      const message = {
+        domain: {
+          chainId: 5,
```
DiogoSoaress, line 165: shouldn't the message's `chainId` be the same as the Safe?

DiogoSoaress (commented):
I couldn't test this fix successfully. Maybe we can look at it in a bit?

src/utils/safe-messages.ts
schmanu: I'm not sure if an undefined value will have impact on the resulting hash. I will give it a try.

schmanu: `undefined` values will cause errors when creating the hash with ethers. So we cannot rewrite it.

DiogoSoaress:
We've just tested this fix works if we sign the message with a 1/N Safe and call the verify signature with the prepared signature.
This will only fully work on the PR deployment once addressed by the backend.

DiogoSoaress (approved):

PR #1467, opened 2022-12-28
refactor: replace Stepper for MUI's List

DiogoSoaress:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/1218

## How this PR fixes it
Replaces `TxSigners` logic around MUI's Stepper by MUI's List in alignment with `MsgSigners` component.

## How to test it
The TxSigners stepper should work as before for all types of queued and historic transactions.
Icons colors should remain unaltered.

## Screenshots
Screenshot 2022-12-28 at 17 11 03
Screenshot 2022-12-28 at 17 11 14

## Expected behaviour 
If the connected account can execute the queued tx it will show the checkmark icon. In the case you describe, the connected account can execute because if it signs, it will satisfy the threshold


DiogoSoaress:
@liliiaorlenko In order to completely replace MUI's icons in this component by our custom ones, we will need:
- Cancel icon (actually https://mui.com/material-ui/material-icons/?query=cancel&selected=Cancel)
- CheckCircle (https://mui.com/material-ui/material-icons/?query=CheckCircle&selected=CheckCircle)
- RadioButtonUnchecked (https://mui.com/material-ui/material-icons/?query=RadioButtonUncheckedO&selected=RadioButtonUnchecked)

🙏

src/components/transactions/TxSigners/index.tsx
```diff
@@ -19,18 +22,21 @@ import { isCancellationTxInfo, isExecutable, isMultisigDetailedExecutionInfo } f
 import EthHashInfo from '@/components/common/EthHashInfo'
 
 import css from './styles.module.css'
+import txSignersCss from '@/components/transactions/TxSigners/styles.module.css'
```
iamacook, line 25: This is already imported as `css`.

src/components/transactions/TxSigners/index.tsx
```diff
@@ -19,18 +22,21 @@ import { isCancellationTxInfo, isExecutable, isMultisigDetailedExecutionInfo } f
...
 import useSafeInfo from '@/hooks/useSafeInfo'
+import palette from '@/styles/colors'
```
iamacook, line 27: This does not take dark mode into account:

image

src/components/transactions/TxSigners/index.tsx
iamacook: We should also reuse this in `MsgSigners`.

katspaugh:
Please update the branch.

iamacook (approved):

liliya-soroka:
@DiogoSoaress , the red circle is changed to the green one during the signing process and comes back to the red one after the signing is done. 
Should we display the right circle state only when the signatures are updated?
image

liliya-soroka:
2nd issue :
1. There are 2 owners in the MM ( owner A and owner B) and 2 of N safe
2. Create tx using owner A
3. reject created tx as an owner A
4. switch to the owner B
5. Confirm first tx without execution as owner B
6. open rejection tx 
Current result: both tx has "all confirmations are received" circle . But only the fully confirmed tx should have this icon 
image

DiogoSoaress:
Ty for QAing @liliya-soroka 

1. I am trying to reproduce it. Can you perhaps provide a short video? Is it only in a 3/N setup?

2. That is expected. The logic we have in PROD is maintained in this PR
_If the connected account can execute the queued tx it will show the checkmark icon. In the case you describe, the connected account can execute because if it signs, it will satisfy the threshold_

DiogoSoaress:
@liliya-soroka please recheck. Point 1 should be fixed

PR #1468, opened 2022-12-29
fix: update signers confirmations

### LINKED ISSUE #1374: confirmations list is not updated on tx details without expand/collapse the tx or manual refresh
## Bug description
confirmations list is not updated on tx details without expand/collapse the tx or manual refresh
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Goerli

## Steps to reproduce
 1. Go to the queued tx list
 2. expand tx and click the "Confirm" button on tx details without tx execution
 3. wait till the number of confirmations is updated 

## Expected result
Make sure that the tx details in expanded view is updated too after signature is received without extra step from the user
## Obtained result
the tx header is updated with the correct number of confirmations, but the tx details in expanding a view is not updated. It might be the reason of some support comments about the wrong number of confirmations, because if I change the MM owner Confirm button is available and I can add one extra signature 
## Screenshots
image
image

DiogoSoaress:
## What it solves

Resolves #1374 

## How this PR fixes it
Adds `txQueuedTag` to the dependency array to fetch TransactionDetails

## How to test it
1. Create a queued transaction in a Safe with a threshold greater than 1
2. Expand the transaction in the queue so that TxSigners is visible
2. Add a confirmation and observe the TxSigners displaying the new data

## Analytics changes
N/A

## Screenshots
Screenshot 2022-12-29 at 13 23 06


src/components/transactions/TxSigners/index.tsx
```diff
@@ -102,6 +106,22 @@ export const TxSigners = ({
...
+  const chainId = useChainId()
+  const [confirmations, setConfirmations] = useState<MultisigConfirmation[]>([])
+
+  useEffect(() => {
+    if (isMultisigDetailedExecutionInfo(detailedExecutionInfo)) {
+      setConfirmations(detailedExecutionInfo.confirmations)
+    }
+  }, [detailedExecutionInfo])
+
+  txSubscribe(TxEvent.SIGNATURE_INDEXED, ({ txId }) => {
```
schmanu, line 118: In the TxDetails we already have a bunch of conditions under which we will refresh the details. I would just add the txQueuedTag and txHistoryTag to that array:

```ts
const TxDetails = ({
  txSummary,
  txDetails,
}: {
  txSummary: TransactionSummary
  txDetails?: TransactionDetails // optional
}): ReactElement => {
  const chainId = useChainId()

  const [txDetailsData, error, loading] = useAsync<TransactionDetails>(async () => {
    return txDetails || getTransactionDetails(chainId, txSummary.id)
  }, [txDetails, chainId, txSummary.id, safe.txQueuedTag, safe.txHistoryTag])

// ...
}
```

That is how we also refresh the tx details for a singleTx in the `SingleTx`

schmanu (changes requested):
I think we should align the way we refresh our `TxDetails` with how we do it for `SingleTx`.

Otherwise for `SingleTx`s we would refresh twice for instance.

src/components/transactions/TxSigners/index.tsx
DiogoSoaress: oh I didn't think about the tags! I'm going to give it a try

DiogoSoaress: I've implemented this way. Thank you for the valuable input!

schmanu (approved):
Looks good! At some point we could add a test to the TxDetails I guess :)

francovenica:
I'd prefer it without having it without collapsing and re expanding, but is good and works fine

confirmation refresh

DiogoSoaress:
> I'd prefer it without having it without collapsing and re expanding, but is good and works fine

I also didn't like that it was collapsing but we have a rule to collapse it when:
- the nr of confirmations matches the threshold

meaning that is happening because you're testing a 2/N Safe.

Alternatively we could remove that condition but I'd prefer having that decision coming from design.

francovenica:
Yes. the collapse is not happening anymore, it just reloads the tx description seemsly
no-collapse

PR #1470, opened 2023-01-02
Chore: update all packages to latest

katspaugh:
Updated all packages except Next.js, Jest (problems with spyOn in v29), Cypress, and react-qr-reader (the latest version has an incompatible interface).


usame-algan:
Just tested and the account switching bug still persists in onboard-core with the updated version (https://github.com/blocknative/web3-onboard/issues/1385).

francovenica (approved):
LGTM

PR #1471, opened 2023-01-02
feat: Redesign welcome page

### LINKED ISSUE #1423: Implement Welcome page redesign
## What is the feature about

Implementing the new welcome page design.

## The list of requirements

- The Safe list sidebar is always visible
- A connect wallet screen is shown if no wallet is connected
- Once the wallet is connected there is an option to create a new safe or load an existing safe
- There is a dynamic gradient in the background, slowly moving in a circular way

## Links

Figma: https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=2904%3A66456&t=gr9zVprciNoouLPA-0

usame-algan:
## What it solves

Resolves #1423 

## How this PR fixes it

Implements the new Redesign for the welcome page.

- Adds an animated background
- Adds the safe list sidebar
- Slightly adjusts the safe list drawer/sidebar layout

## How to test it

1. Open the welcome page
2. Compare it to the design (desktop + mobile)

## Screenshots
Screenshot 2023-01-03 at 09 28 07
Screenshot 2023-01-04 at 10 28 23


src/components/welcome/index.tsx
```diff
@@ -1,57 +1,77 @@
...
+      <Grid item xs={12} md={4} lg={3.5}>
+        <div className={css.sidebar}>
+          <SafeList closeDrawer={() => {}} />
+        </div>
+      </Grid>
+      <Grid item xs={12} md={8} lg={8.5}>
+        <div className={css.main}>
+          <div className={css.content}>
+            <Typography variant="h1" fontSize={52} lineHeight={1} letterSpacing={-1.5} mb={1}>
+              Welcome to the Safe.
```
katspaugh, line 27: Typography-wise, should we remove the final dot?
The two blocks below (Create Safe and Load Safe) don't have a dot in the headings.

katspaugh:
This text looks weird, probably because of the grey underline instead of black?
Screenshot 2023-01-03 at 09 09 49

Please remove the dot. Btw in future it would be nice to add an ℹ️ article about what a wallet is.
Screenshot 2023-01-03 at 09 10 03

src/components/welcome/index.tsx
```diff
@@ -1,57 +1,85 @@
...
+                      + Create new Safe
+                    </Button>
+                  </Track>
+                </Paper>
+              </Grid>
+              <Grid item xs={12} sm={6}>
+                <Paper sx={{ padding: 4, height: 1 }}>
+                  <SvgIcon component={LoadSafeIcon} inheritViewBox sx={{ width: '42px', height: '42px' }} />
+                  <Typography variant="h3" fontWeight={700} mb={1} mt={3}>
+                    Add Existing Safe
```
katspaugh, line 57: Lowercase e, please.

src/components/welcome/index.tsx
```diff
@@ -1,57 +1,86 @@
...
+          {wallet ? (
+            <Grid container spacing={3} sx={{ maxWidth: '800px' }}>
+              <Grid item xs={12} sm={6}>
+                <Paper sx={{ padding: 4, height: 1 }}>
+                  <SvgIcon component={NewSafeIcon} inheritViewBox sx={{ width: '42px', height: '42px' }} />
+                  <Typography variant="h3" fontWeight={700} mb={1} mt={3}>
+                    Create Safe
+                  </Typography>
+                  <Typography variant="body2" mb={3}>
+                    New Safe that is controlled by one or multiple owners.{' '}
```
katspaugh, line 45: This copy definitely needs work, but as it stands it lacks an "A" at the beginning.

src/components/welcome/index.tsx
```diff
@@ -1,57 +1,86 @@
...
+                  </Track>
+                </Paper>
+              </Grid>
+            </Grid>
+          ) : (
+            <ButtonBase className={css.button} onClick={connectWallet} disableRipple>
+              <Box display="flex" alignItems="center" gap={2} textAlign="left">
+                <CircularIcon icon={Keyhole} size={40} />
+                <Typography variant="h4" fontWeight={700}>
+                  Connect wallet to start using Safe.
```
katspaugh, line 76: Please remove the final dot, it looks odd on a button.

DiogoSoaress:
Can we hide the overlay sidebar when navigating to `/welcome` ?
sidebar staying

cypress/e2e/smoke/load_safe.cy.js
```diff
@@ -16,14 +16,16 @@ const OWNER_ADDRESS = '0xE297437d6b53890cbf004e401F3acc67c8b39665'
 
 describe('Load existing Safe', () => {
   before(() => {
+    cy.connectE2EWallet()
```
DiogoSoaress, line 19: Does this test need the connected wallet?

src/components/sidebar/SafeList/index.tsx
```diff
@@ -68,119 +71,146 @@ const SafeList = ({ closeDrawer }: { closeDrawer: () => void }): ReactElement =>
     setOpen((prev) => ({ ...prev, [chainId]: open }))
   }
 
+  const isWelcomePage = router.pathname === AppRoutes.welcome
```
DiogoSoaress, line 74: Can we use the function from `SideDrawer` ? It is the same logic

src/components/sidebar/SafeListItem/index.tsx
```diff
@@ -56,6 +56,7 @@ const SafeListItem = ({
     <ListItem
       className={css.container}
       disablePadding
+      sx={{ '& .MuiListItemSecondaryAction-root': { right: '24px' } }}
```
DiogoSoaress, line 59: also works
```suggestion
      sx={{ '& .MuiListItemSecondaryAction-root': { right: 24 } }}
```

src/components/welcome/index.tsx
```diff
@@ -1,57 +1,118 @@
...
+          </AccordionSummary>
+
+          <AccordionDetails sx={{ padding: 0 }}>
+            <div className={css.sidebar}>
+              <SafeList closeDrawer={() => {}} />
+            </div>
+          </AccordionDetails>
+        </Accordion>
+        <Box className={css.sidebar} display={{ xs: 'none', md: 'block' }}>
+          <SafeList closeDrawer={() => {}} />
```
DiogoSoaress, line 50: Maybe better making the prop optional?

src/components/welcome/styles.module.css
```diff
@@ -0,0 +1,69 @@
...
+
+.content {
+  background: linear-gradient(-90deg, #10ff84, #5fddff);
+  background-size: 200% 200%;
+  animation: gradient 15s ease infinite;
+  height: 100%;
+  border-radius: 6px;
+  padding: var(--space-12);
+  padding-top: 184px;
+}
```
DiogoSoaress, line 34: 🚀

usame-algan:
> Because the Sidebar is narrower now, some addresses are cut

We could give it a fixed width per viewport and instead make the content area smaller since there is more unused space there anyway.

DiogoSoaress (changes requested):
Some minor remarks but great job nevertheless!! Trippy gradient 😵

cypress/e2e/smoke/load_safe.cy.js
DiogoSoaress: Oh! Now we do of course!

usame-algan: It doesn't seem to work on the pipeline though. Might need to add a timeout there.

src/components/welcome/index.tsx
usame-algan: Good idea, we can actually use that instead of checking for `isWelcomePage`

src/components/sidebar/SafeList/index.tsx
usame-algan: I removed this as per my other comment.

src/components/welcome/index.tsx
DiogoSoaress: I think we lose a bit of readability that way 😬 . I would just make the prop optional to not have to pass the empty function and use the other route util from `SideDrawer`.

DiogoSoaress:
Can we display the cards vertically if for narrow screens?

Screenshot 2023-01-04 at 14 48 57

DiogoSoaress (commented):
One last thing comparing with the designs: the footer should be always in the viewport and in the branch I have to scroll to see it.

usame-algan:
> Can we display the cards vertically if for narrow screens?

I adjusted the breakpoints to display them vertically < 1200px

francovenica:
There is a flicker of the old sidebar everytime you click in the "Safe" logo in the top left corner
This flicker also happens if you switch networks in the top right corner while being connected with an owner that has safes in those networks
sidebar

usame-algan:
> There is a flicker of the old sidebar everytime you click in the "Safe" logo in the top left corner
This flicker also happens if you switch networks in the top right corner while being connected with an owner that has safes in those networks

This happens because we have some redirection logic going on. I've hidden the sidebar on "/" to fix this.

src/utils/route.ts
```diff
@@ -0,0 +1,26 @@
...
+  return pathname === AppRoutes.newSafe.create
+}
+
+export const isWelcomeRoute = (pathname: string): boolean => {
+  return pathname === AppRoutes.welcome
+}
+
+export const isLoadSafeRoute = (pathname: string): boolean => {
+  return pathname === AppRoutes.newSafe.load
+}
```
katspaugh, line 18: I think this is an overkill. They are not used anywhere other than in the side drawer. `isIndexRoute` isn't used anywhere at all according to my grep fu.

Side drawer could just do `[ AppRoutes.welcome, AppRoutes.newSafe.create, AppRoutes.newSafe.load ].includes(pathname)`

DiogoSoaress (approved):

francovenica:
Issue:

For the mobile view. Using the arrow icon to collapse/display the list of owned safes triggers the collapse of the "My Safe" widgets.
collapsable widget

francovenica:
Nitpick:

I think the message "Create or add an existing safe" is wrong, implies that you can create an existing safe. I'd change it to "Create a new safe or add an existing one" if is not too long.

usame-algan:
> For the mobile view. Using the arrow icon to collapse/display the list of owned safes triggers the collapse of the "My Safe" widgets.

This should work now.

> I think the message "Create or add an existing safe" is wrong, implies that you can create an existing safe. I'd change it to "Create a new safe or add an existing one" if is not too long.

Makes sense, thanks! I adjusted the wording accordingly.

francovenica:
It works fine now. If peer review passes we can put it in QA done

katspaugh (approved):
🚀

PR #1473, opened 2023-01-02
fix: don't show previous Safe notifications

DiogoSoaress:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/1267

## How this PR fixes it
Before checking if the Safe version is outdated, verifies that the loaded Safe is the selected one.

## How to test it
1. Open a 1.1.1 Safe, which should show an update notification
2. Change to a > 1.3.0 Safe and there should be no `Update Safe` notification popping up or in the Notification center

## Screenshots
outdated notifications fix


src/hooks/useSafeNotifications.ts
```diff
@@ -26,12 +27,14 @@ const useSafeNotifications = (): void => {
...
   const { chainId, version, implementationVersionState } = safe
   const isOwner = useIsSafeOwner()
+  const safeAddress = useSafeAddress()
 
   /**
    * Show a notification when the Safe version is out of date
    */
 
   useEffect(() => {
+    if (safe.address.value !== safeAddress) return
```
katspaugh, line 37: I think you can achieve the same like this:
```
const { safe, safeAddress } = useSafeInfo()

if (safeAddress) return
```

`safeAddress` from `useSafeInfo` becomes undefined when a Safe is being (re)loaded.

DiogoSoaress: Checking it! Thanks

DiogoSoaress: It turns undefined indeed when fetching but in the moment we call `useSafeNotifications` it is still returning the old safe address.

I found comparing the URL address with the loaded Safe address more reliable. WDYT?

katspaugh: I see, thanks for checking!

katspaugh (approved):
👍

francovenica:
LGTM

PR #1475, opened 2023-01-02
Chore: rm .html files on S3

### LINKED ISSUE #1029: Deployment job: don't upload unused .html files.
The current deployment script uploads everything in the build output directory, and then additionally uploads HTML files w/o an extension.
Ideally, we should not upload the `.html` files, it will save some space.

image (1)

katspaugh:
## What it solves

Resolves #1029

## How this PR fixes it

Excludes HTML files when syncing.

### DIFF
--- a/scripts/github/s3_upload.sh
+++ b/scripts/github/s3_upload.sh
@@ -4,16 +4,16 @@ set -ev
 
 cd out
 
-# First, upload the new files w/o deleting the old ones
-aws s3 sync . $BUCKET
+# First, upload new files w/o deleting the old ones
+aws s3 sync . $BUCKET --exclude "*.html"
 
 # Second, upload them again but delete the old files this time
 # This allows for a no-downtime deployment
-aws s3 sync . $BUCKET --delete
+aws s3 sync . $BUCKET --delete --exclude "*.html"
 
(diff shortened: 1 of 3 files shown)


scripts/github/s3_upload.sh
```diff
@@ -4,16 +4,19 @@ set -ev
...
 
-# Finally, upload all HTML files again but w/o an extention so that URLs like /welcome open the right page
+# Upload HTML files but w/o an extention so that URLs like /welcome open the right file
 for file in $(find . -name '*.html' | sed 's|^\./||'); do
-    aws s3 cp ${file%} $BUCKET/${file%.*} --content-type 'text/html'
+    aws s3 cp ${file%} $BUCKET/${file%.*} --content-type "text/html"
 done
 
+# Upload index.html as the default file
+aws s3 cp index.html $BUCKET
```
katspaugh, line 20: WFYT @luarx? This avoids uploading unused .html files, but we then have to upload `index.html` explicitly. Do you think it's worth it?

katspaugh:
Closing because the complexity overhead in the upload script is not worth the extra few KBs.

luarx:
Ok

scripts/github/s3_upload.sh
luarx: Maybe not

PR #1476, opened 2023-01-03
Chore: roll back onboard-core

katspaugh:
## What it solves

The latest onboard has a [bug with account switching](https://github.com/blocknative/web3-onboard/issues/1385), so we're rolling it back.


usame-algan (approved):
Might also need to generate a new lockfile.

katspaugh:
Of course, thank you 🤦

PR #1477, opened 2023-01-03
fix: trigger validation upon owner removal

DiogoSoaress:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/1350

## How this PR fixes it
Triggers a form validation after an owners array change

## How to test it
In the creation flow, add a repeated owner.
Remove the repeated owner.
The form should have no errors.

### DIFF
--- a/src/components/new-safe/create/steps/Step2/index.tsx
+++ b/src/components/new-safe/create/steps/Step2/index.tsx
@@ -50,7 +50,7 @@ const CreateSafeStep2 = ({
     },
   })
 
-  const { handleSubmit, control, watch, formState, getValues, setValue } = formMethods
+  const { handleSubmit, control, watch, formState, getValues, setValue, trigger } = formMethods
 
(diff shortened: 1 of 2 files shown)


src/components/new-safe/create/steps/Step2/index.tsx
```diff
@@ -64,6 +64,7 @@ const CreateSafeStep2 = ({
     // Set threshold if it's greater than the number of owners
     setValue(CreateSafeStep2Fields.threshold, Math.min(threshold, ownerFields.length - 1))
     remove(index)
+    trigger()
```
usame-algan, line 67:
```suggestion
    trigger(CreateSafeStep2Fields.owners)
```
Not a big difference but its enough to trigger validation for owners only

usame-algan (approved):
Noticed that we have validation `mode: 'all'` for the owner step and it seems to run the validation twice as much as needed. I believe changing it to `onChange` would suffice.

francovenica:
LGTM

PR #1479, opened 2023-01-03
[Prototype] NFTs table

katspaugh:
## What it solves
A prototype to display NFTs as a table

## Screenshots
Screenshot 2023-01-03 at 16 39 28


usame-algan:
Nice, it looks much more structured than the current view imo! Would be interesting to explore how we could display the images e.g. show them as a "moving" tooltip on hover.

katspaugh:
I was thinking that too, but OpenSea is rate-limited at 4 req/sec...

src/components/nfts/NftGrid/index.tsx
```diff
@@ -1,57 +1,204 @@
...
+                </TableCell>
+
+                {/* Checkbox */}
+                <TableCell align="right">
+                  <Checkbox checked={selectedNfts.includes(item)} />
+                </TableCell>
+              </TableRow>
+            ))}
+
+            {/* Fill up the table up to min rows */}
```
schmanu, line 198: Why do we need these min rows?

Is it related to the InfiniteScrolling?

src/components/nfts/NftCollections/index.tsx
```diff
@@ -1,61 +1,109 @@
...
 
   return (
     <>
-      {/* Aggregated NFTs grouped by collection */}
-      {allNfts.length > 0 && (
-        <NftGrid collectibles={allNfts} onSendClick={isGranted ? (nft) => setSendNft(nft) : undefined} />
+      {allNfts?.length > 0 && (
+        <>
+          {/* Mass send button */}
+          <Box mb={2} display="flex" alignItems="center" gap={1}>
```
schmanu, line 49: Should we make this sticky?
Especially useful with InfiniteScrolling IMO.

src/components/nfts/NftGrid/index.tsx
```diff
@@ -1,57 +1,204 @@
...
+interface NftsTableProps {
+  nfts: SafeCollectibleResponse[]
+  selectedNfts: SafeCollectibleResponse[]
+  onSelect: (item: SafeCollectibleResponse) => void
+  onFilter: (value: string) => void
+}
+
+const minRows = 10
+const iconSize = 20
+const iconStyle = { width: '100%', maxHeight: '100%' }
```
schmanu, line 37: This stretches some images if their original height is bigger than their width.

For instance:
This image
Original

Looks like this
Screenshot 2023-01-19 at 16 22 59

src/components/nfts/NftGrid/index.tsx
katspaugh: Yes. When you filter, and there are fewer than 10 rows, we don't want the page to jump and trigger InfiniteScroll.

katspaugh:
@schmanu fixed everything, thanks!
Regarding the sticky header, let's do it in another PR and also make the send form a shared component.

schmanu (approved):
Nice! 🚀

liliya-soroka:
1. It's strange to see the message when no opensea app 
image

katspaugh:
Fixed. ✅

usame-algan:
Is it possible to align these white boxes?
Screenshot 2023-01-23 at 16 48 49

I noticed that there is no loading state
Screenshot 2023-01-23 at 16 43 07

When not an owner or not connected with a wallet, it is still possible to select/deselect tokens without being able to do any other action. Is this intended or should we hide the filter and select boxes unless an owner is connected?

liliya-soroka:
2. https://nfts--webcore.review-web-core.5afe.dev/gor:0xC642bEc0e4f69815e89723C09caA8DAa524cAc9c/balances/nfts - Goerli
children instead of token ID
image

katspaugh:
Fixed, thanks!

@usame-algan reg. the selection being always enabled, I don't see why not. You can select a few items, and then connect a wallet.

liliya-soroka:
GA tracking for new NFT page ?

liliya-soroka:
3. I want to leave it here as a comment - the table is not resized if there are no safe-apps for the NFT for the selected chain
image

liliya-soroka:
5. Send NFT from the side menu is not opened if the user is located on the NFTs page:
- open NFT page
- click New tx in the side menu
- select send NFT
Current result: Nothing happens

liliya-soroka:
6. Add the "Batched transactions" title before the list of actions as we have for the Batch option on the history 

image

image

katspaugh:
* I've added titles in tx modals + a scrollable area for the batched items in the modal
* Added tracking on the Send button (tracks how many NFTs are sent)
* Removed the Send NFT button from the New Transaction modal
* The Send button in the Address Book now opens the Send Token modal directly

liliya-soroka:
regression after the fix : no send NFTs when "New transaction" is used on transactions, assets(tokens list) or any tab.
I think it make sense to hide only for the NFTs list or don't hide it at all  
image

katspaugh:
Well, it was intentional, but I like the idea of not showing it only on the NFT page.

katspaugh:
Restored the button 👍

src/components/nfts/NftCollections/index.tsx
katspaugh: Made it sticky, thanks for the idea!

liliya-soroka:
Could we keep the same view for the long list of actions for NFTs butch , as we have for usual txs with many actions ?
NFTs:
image

Usual actions list in the modal window 
image

liliya-soroka:
Also question to the table when the list is short:
Tokens list :
image

NFTs list:
image

katspaugh:
@liliya-soroka I've removed the scroll in the review modal. It was better with it, but I agree it should be consistent with other modals.

Also removed the empty rows in the table. They'll be shown only when filtering to avoid having the page jump.

liliya-soroka:
Verified

PR #1494, opened 2023-01-06
fix: Downgrade RHF version to 7.41.1

usame-algan:
## What it solves

The recent update to RHF >= 7.41.2 broke the form validation on the safe creation screen (and potentially other places).

## How this PR fixes it

Downgrades RHF to the last working version 7.41.1

## How to test it

1. Open the Safe
3. Create a new safe and navigate to the owner policy screen
4. Observe that the Next button is not disabled

## References

[Slack thread with more info](https://5afe.slack.com/archives/C03DAGWJCR5/p1672932360224469)

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "react-dom": "18.2.0",
     "react-dropzone": "^14.2.3",
     "react-gtm-module": "^2.0.11",
-    "react-hook-form": "^7.41.3",
+    "react-hook-form": "7.41.1",
     "react-papaparse": "^4.0.2",
     "react-qr-reader": "2.2.1",
     "react-redux": "^8.0.2",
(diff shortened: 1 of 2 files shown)


iamacook (approved):

katspaugh (approved):
Thanks!
It would be nice to catch that in an e2e test...

usame-algan:
> It would be nice to catch that in an e2e test...

I agree! We could split up the create safe e2e test into two. One of them as a smoke test purely going through the form flow (which would catch this) and the other one testing the transaction logic at the end.

francovenica:
Looks good. 
I compared with dev, where you have to pick again the policies to enable the "Next" button, and here the next button is enabled as long all the owners fields are properly filled

PR #1495, opened 2023-01-06
refactor: Remove old safe creation code

### LINKED ISSUE #1404: Remove old safe creation and AB test

usame-algan:
## What it solves

Resolves #1404 

## How this PR fixes it

- Removes old safe creation components under `/src/components/create-safe`
- Moves applicable tests to new directory `/src/components/new-safe/create`
- Removes abTest for safe creation
- Renames create safe steps to be more descriptive

## How to test it

1. Open the Safe
2. Go through the safe creation flow
3. Observe that it still works


katspaugh (changes requested):
We shouldn't remove the A/B test feature from GTM.
It was meant to be reusable.

usame-algan:
> We shouldn't remove the A/B test feature from GTM.

I restored the ab test related code and only removed its initialization from the app

src/pages/_app.tsx
```diff
@@ -51,7 +49,6 @@ const InitApp = (): null => {
   useTxPendingStatuses()
   useTxTracking()
   useBeamer()
-  useABTesting(AbTest.SAFE_CREATION)
```
katspaugh, line 54: Please also remove `SAFE_CREATION` from the enum.

katspaugh (approved):
Looks good, just one last comment.

src/pages/_app.tsx
usame-algan: I've removed it and added a comment so we know where to put future abTest entries

PR #1496, opened 2023-01-06
Feat: allow switching wallet

### LINKED ISSUE #1490: Allow switching to a previously connected wallet
## Problem
As a user, I have Safes with MetaMask and Ledger as owners, and I need to often switch wallets to sign transactions.
I have to disconnect and connect again.

## Improvement
Add an option to switch to a previous wallet if it’s an owner of the current Safe.

Screenshot 2023-01-05 at 15 49 16

We currently persist only the last connected wallet and re-connect to it on page load. We could instead remember the last 10 connected wallets and offer to switch to the most recent one if it's an owner of the current Safe.

katspaugh:
## What it solves

Resolves #1490

## How this PR fixes it

Instead of making the user disconnect to connect to another wallet, allow simply connecting in parallel.

Screenshot 2023-01-06 at 14 39 43

Screenshot 2023-01-06 at 14 40 51

## How to test it
* Connect to MM
* Switch wallet to WC
* Switch back to MM

## Analytics changes
None, but it should track the wallet each time


usame-algan (approved):
Really cool feature and nice that we make use of it now!

francovenica:
I thought this was not possible. Glad to see it working

I was able to switch between MM, WC (MM phone app), Trezor and Ledger with no issues for now.

PR #1497, opened 2023-01-06
[Ecosystem friday] add: lavamoat-allow-scripts

schmanu:
## What it solves
Adds lavamoat-allow-scripts to protect from supply chain attacks during dependency installation.

## Breaking changes (will be added into Readme):
Instead of running `yarn` or `yarn install` we would have to run the new `yarn setup` script.

It will:
1. run `yarn install`  (without running any scripts due to `.yarnrc`)
2. run LavaMoat's `allow-scripts` to
  - link allowed bin files to PATH
  - run preinstall / install / postinstall scripts
3. run what we previously ran in `postinstall`

## How to install new dependencies / update dependencies
1. run `yarn add <dependency>`
2. run `yarn allow-scripts auto --experimental-bins` to update the `package.json` (new scripts might be added)
3. Check the diff of the`package.json`
  - if nothing is changed: there is nothing else to do 🥳 
  - if new lines where added, they get added with `false` flag by default, causing scripts to not be executed on install.
  - check if the scripts of those dependencies are actually needed by running `npx can-i-ignore-scripts`
  - if they are needed: change the flag to `true`

To simplify this, we should add a new git workflow which fails if `yarn allow-scripts auto --experimental-bins` would create a diff - meaning that dependencies were changed without running the command.


katspaugh:
That's really cool! 🔥

PR #1498, opened 2023-01-09
Chore: integrate DevSkim

katspaugh:
It's a security linter from Microsoft.
See https://github.com/microsoft/DevSkim


katspaugh:
If SonarLint is better, I'm gonna close this PR, let's add that instead. 👍

PR #1499, opened 2023-01-09
Fix: add PWA fields in the manifest

katspaugh:
## What it solves
According to MDN, the correct extension is `.webmanifest`, and it needs `start_url` for the PWA to be installable.

Also updated the desktop icons from the brand figma.

Screenshot 2023-01-09 at 10 44 23


DiogoSoaress (approved):
LGTM!
I could test it in the deployment branch but not if running the branch locally. Do you know why?

katspaugh:
Because https.

liliya-soroka:
the corners of the app icon is not rounded
image

katspaugh:
Good catch, I missed that one. Fixed here: https://github.com/safe-global/web-core/pull/1526

PR #1502, opened 2023-01-09
Feat: add a chain switch button in Safe creation

katspaugh:
## What it solves

I was trying to create a Safe and was really confused why the submit button was disabled.

### Before
Screenshot 2023-01-09 at 15 31 59

### After
Screenshot 2023-01-09 at 15 29 58


usame-algan (approved):
Makes sense to have an action there! Should we explore adding it as an inline button? cc @liliiaorlenko

francovenica:
Works nicely. LGTM

PR #1503, opened 2023-01-09
style: mobile Address Book layout

### LINKED ISSUE #1028: Update mobile address book layout
## What is the feature about

The current mobile address book layout does not match that of [Figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=1004%3A49906). We should adjust the layout to match that.

## Designs/sketches
Screenshot 2022-11-08 at 17 57 26

DiogoSoaress:
## What it solves

Resolves #1028 

## How this PR fixes it
Places the buttons under the search bar
Table cells become Cards in mobile viewports
Don't hide the Send button

## How to test it
Open the Address Book and check the mobile layout

## Analytics changes
N/A

## Screenshots
Screenshot 2023-01-10 at 14 36 42


src/components/common/EnhancedTable/index.tsx
```diff
@@ -150,13 +149,15 @@ function EnhancedTable({ rows, headCells, variant }: EnhancedTableProps) {
                   selected={row.selected}
                   className={row.collapsed ? css.collapsedRow : undefined}
                 >
+                  {CellCard && <CellCard row={row} className={css.cellCard} />}
```
usame-algan, line 152: Not sure if we need to render a different component for mobile. It should be possible to break the table cells via `display` into a row based layout.

DiogoSoaress: I wanted to achieve displaying a custom Card that can be passed to the `<EnhancedTable>` for mobile views, as we've already had some issues and designs that suggested going in that direction.
I've considered keeping the table cells as well but the reason I've decided for this approach was the flexibility it gives.
Do you find the other approach more suitable?

usame-algan: I would try to avoid rendering the same content twice if avoidable e.g. if we can achieve layout changes with css

DiogoSoaress: I've removed the Card component and made the table accept a `mobileVariant`

src/components/common/EnhancedTable/styles.module.css
```diff
@@ -21,3 +21,19 @@
...
+    display: none;
+  }
+
+  .mobileColumn thead th:first-of-type {
+    display: block;
+  }
+
+  .mobileColumn tbody td {
+    display: block;
+    padding-left: 16px !important;
```
usame-algan, line 36:
```suggestion
    padding-left: var(--space-2) !important;
```

usame-algan (approved):

liliya-soroka:
Verified.
Not one to one with the figma file, but I would say that it's expected. Looks good in the mobile view

PR #1506, opened 2023-01-09
fix: Show sidebar on mobile when pressing the burger menu icon

usame-algan:
## What it solves

Fixes a regression introduced in #1471 

## How this PR fixes it

- Keeps the sidebar mounted but hidden

## How to test it

1. Open the welcome/create safe/load safe page on mobile
2. Press the Burger menu icon
3. Observe the sidebar opening


katspaugh (approved):
Tested, it works now 👍

PR #1515, opened 2023-01-10
Fix: fiat total skeleton

katspaugh:
## What it solves

A small regression from the spam tokens feature. While balances are loading, we were briefly displaying 0.00 USD instead of a skeleton in the sidebar.

### DIFF
--- a/src/hooks/useVisibleBalances.ts
+++ b/src/hooks/useVisibleBalances.ts
@@ -43,17 +43,17 @@ export const useVisibleBalances = (): {
   loading: boolean
   error?: string
 } => {
-  const balances = useBalances()
+  const data = useBalances()
   const hiddenTokens = useHiddenTokens()
 
   return useMemo(
     () => ({
-      ...balances,
+      ...data,
       balances: {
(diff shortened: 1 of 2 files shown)


src/hooks/useVisibleBalances.ts
```diff
@@ -43,17 +43,17 @@ export const useVisibleBalances = (): {
...
 
   return useMemo(
     () => ({
-      ...balances,
+      ...data,
       balances: {
-        items: filterHiddenTokens(balances.balances.items, hiddenTokens),
-        fiatTotal: getVisibleFiatTotal(balances.balances, hiddenTokens),
+        items: filterHiddenTokens(data.balances.items, hiddenTokens),
+        fiatTotal: data.balances.fiatTotal ? getVisibleFiatTotal(data.balances, hiddenTokens) : '',
```
schmanu, line 54: I'm wondering how this condition is fixing it:
The `initialState` of the balances slice contains: `fiatTotal: '',` 
So I think it can never be `undefined`.

Did you find out where the undefined state is coming from?

schmanu (approved):
👍

PR #1516, opened 2023-01-10
Bug: No placeholder icon when safe app's logo cannot be loaded

mikhailxyz:
## What it solves

Resolves https://github.com/safe-global/safe-react-apps/issues/637 by rendering an img tag inside the sandboxed iframe for fetching safe apps icon (for more context see https://github.com/safe-global/web-core/pull/1193)

thanks to @katspaugh for answering questions and suggesting script tag workaround

## How this PR fixes it
It renders an img tag and attaches error event listener. It also allows scripts inside the iframe to define the error listener. [Shouldn't be a security risk](https://security.stackexchange.com/questions/124840/xss-vectors-in-img-src-and-background-image-url). Also it seems like scripts must be same-origin and cross origin ones have to be explicitly enabled

## How to test it

1. Fetch an app with an erroring image (I don't know such app so i ran https://github.com/safe-global/safe-apps-sdk/tree/master/packages/safe-apps-test-app locally)

## Screenshots
image


src/components/safe-apps/SafeAppIcon/index.tsx
```diff
@@ -1,17 +1,21 @@
...
-      padding: 0;
-      margin: 0;
-      background-color: transparent;
-      height: 100%;
-    }
-    body {
-      background: url('${src}') center center/contain no-repeat;
-    }
-  </style>`
+const APP_LOGO_FALLBACK_IMAGE = `/images/apps/app-placeholder.svg`
```
mikhailxyz, line 3: a new cross-them icon by Lila

src/components/safe-apps/SafeAppIcon/index.tsx
```diff
@@ -1,17 +1,19 @@
...
+
+const getIframeContent = (url: string, alt: string, width: number, height: number): string => {
+  return `
+     <body style="margin: 0;">
+       <img src=${url} alt=${alt} width="${width}" height="${height}" />
+       <script>
+          document.querySelector("img").addEventListener("error", (e) => {
+           e.target.onerror = null
```
katspaugh, line 11: You need to assign the handler like this as well then: `document.querySelector("img").onerror =`

src/components/safe-apps/SafeAppIcon/index.tsx
```diff
@@ -1,17 +1,19 @@
...
-    }
-    body {
-      background: url('${src}') center center/contain no-repeat;
-    }
-  </style>`
+const APP_LOGO_FALLBACK_IMAGE = `/images/apps/app-placeholder.svg`
+
+const getIframeContent = (url: string, alt: string, width: number, height: number): string => {
+  return `
+     <body style="margin: 0;">
```
katspaugh, line 7: Some browsers add margins to <html> too, please test in FF, Safari, etc.

src/components/safe-apps/SafeAppIcon/index.tsx
```diff
@@ -1,17 +1,19 @@
...
+       <img src=${url} alt=${alt} width="${width}" height="${height}" />
```
katspaugh, line 8: I would add `style="display: block"` to avoid a line underneath.

src/components/safe-apps/SafeAppIcon/index.tsx
mikhailxyz: I changed it but just fyi it's unclear why I "need to assign" it that way from your message (I assumed stylistic reasons)

src/components/safe-apps/SafeAppIcon/index.tsx
mikhailxyz: What line? I didn't see anything unwanted in all browsers I tested (chrome, safari desktop, firefox)

mikhailxyz: I remember you mentioned it in the slack convo but when working on it I didn't see a need for it

src/components/safe-apps/SafeAppIcon/index.tsx
mikhailxyz: Tested in Safari desktop/mobile, chrome and FF - all good

src/components/safe-apps/SafeAppIcon/index.tsx
katspaugh: I meant it won't work like this. Try this in the console, it will log `error` twice.
```
const img = new Image()
img.addEventListener('error', () => {
    console.log('error')
    img.onerror = null
})
img.src = '123'
setTimeout(() => {
    img.src = '456'
}, 100)
```

src/components/safe-apps/SafeAppIcon/index.tsx
katspaugh: This line:
Screenshot 2023-01-11 at 08 46 35
Inline elements such as img have a line-height that are bigger than the height of the image itself.

katspaugh: Doesn't really matter in this case tho, because the iframe itself is 40x40.

katspaugh, line 8: I would wrap `url` and `alt` in double-quotes, and escape any double quotes in those strings. Otherwise, we have a classic XSS hole here.

katspaugh: Actually, I would hardcode smth in the alt tag like "app logo". The name of the app is already on the card anyway.

src/components/safe-apps/SafeAppIcon/index.tsx
mikhailxyz: Screenshot 2023-01-11 at 11 08 09

mikhailxyz: Not sure I get it why it wouldn't have worked

src/components/safe-apps/SafeAppIcon/index.tsx
mikhailxyz: Good spot regarding xss. I've checked other platforms with lists to see how they implement alt attributes and seems like they don't lol (checked chrome web store, slack store)

src/components/safe-apps/SafeAppIcon/index.tsx
katspaugh: Screenshot 2023-01-11 at 11 54 53

My understanding is you can't clear an addEventListener handler via an old style DOM1 `onerror = null` assignment.

src/components/safe-apps/SafeAppIcon/index.tsx
```diff
@@ -1,17 +1,19 @@
...
-    body {
-      background: url('${src}') center center/contain no-repeat;
-    }
-  </style>`
+const APP_LOGO_FALLBACK_IMAGE = `/images/apps/app-placeholder.svg`
+
+const getIframeContent = (url: string, width: number, height: number): string => {
+  return `
+     <body style="margin: 0;">
+       <img src="${url}" alt="App logo" width="${width}" height="${height}" />
```
katspaugh, line 8: Like I said, you also might want to escape dangerous symbols here. `encodeURI(url)` should do the trick.

katspaugh:
Have you tried this btw?
https://stackoverflow.com/a/980910/352796

src/components/safe-apps/SafeAppIcon/index.tsx
mikhailxyz: Here it gets a valid URL as a source, so the browser makes a request (hence a failed request to http://0.0.0.1), and after 100ms the src is changed to another URL and the browser makes a second request. I guess the handlers may be cached at the time of the request. If you change the first `img.src` assignment to an invalid URL to guarantee that the handler will be called before the source change, it will log `error` only once.

katspaugh: Yes. My point is, you need to either use `addEventListener` + `removeEventListener`, or `onerror` + `onerror = null`, otherwise the handler won't be removed.

mikhailxyz: > My understanding is you can't clear an addEventListener handler via an old style DOM1 onerror = null assignment.

Yeah you seem to be right, if I change the snippet to `onerror`-style assignment it works as expected

src/components/safe-apps/SafeAppIcon/index.tsx
mikhailxyz: The url is validated at the fetch time and also browsers seem to be smart nowadays: https://security.stackexchange.com/questions/124840/xss-vectors-in-img-src-and-background-image-url

mikhailxyz: Given we have all these safety things in place I think an additional check inside the component is redundant

mikhailxyz:
> Have you tried this btw? https://stackoverflow.com/a/980910/352796

It seems to follow the CSP policy set in the app and doesn't allow loading from cross origin domains. If I understand correctly, an `object` tag can load arbitrary data and not only images, that's why it's enforced
Screenshot 2023-01-11 at 14 22 51

src/components/safe-apps/SafeAppIcon/index.tsx
katspaugh: It's not about fetching, you're inserting a string directly into HTML. It can close the tag and render a butt.

mikhailxyz: Doesn't seem like a big security threat to me given that we validate the URL before and the amount of sandboxing we enforce on the iframe. But to get the pr through i added it

katspaugh (approved):
Dunno if it's worth it just to show a placeholder for broken icons (it would be better to fix such icons quickly) but looks good. 👍

JagoFigueroa:
Muchas gracias gentlemen, todo bueno 👍

PR #1518, opened 2023-01-10
[Eip-1271] normalize EIP-712 messages

### LINKED ISSUE #1517: [EIP-1271] Normalize EIP 712 messages before submitting to services
## Bug description
Currently we submit the EIP 712 messages as we retrieve them to our tx service.
We should first create the payload using`_TypedDataCoder.getPayload` to verify and normalize the message.

From the ethers docs:
>Returns the standard payload used by various JSON-RPC eth_signTypedData* calls.
>All domain values and entries in value are normalized and the types are verified.

## Steps to reproduce
 1. Send a typed data message without EIP712 domain type or without `primaryType` set.
 2. It will be accepted and signed by the interface but then fails when submitted to the tx service.

## Expected result
It shoud either fail before signing, when we verify the message in our interface or go through without errors.

## Obtained result
It fails when submitted to tx services.

schmanu:
## What it solves

Before submitting EIP 712 messages to our tx service we normalize the message.

Normalizing a message does:
- Infers the EIP712Domain type
- Infers the primaryType

Resolves #1517 

## TODOs
- [x] Add unit test for message normalization

## How this PR fixes it
Uses the `TypedDataEncoder.getPayload` function to normalize and verify a message before submitting it.

## How to test it
Sign a EIP 712 message without primaryType or without EIP712Domain type.

## Analytics changes
None

### DIFF
--- a/src/services/safe-messages/__tests__/safeMsgSender.test.ts
+++ b/src/services/safe-messages/__tests__/safeMsgSender.test.ts
@@ -5,6 +5,7 @@ import type { JsonRpcSigner } from '@ethersproject/providers'
 import { dispatchSafeMsgConfirmation, dispatchSafeMsgProposal } from '@/services/safe-messages/safeMsgSender'
 import * as utils from '@/utils/safe-messages'
 import * as events from '@/services/safe-messages/safeMsgEvents'
+import { hexZeroPad } from 'ethers/lib/utils'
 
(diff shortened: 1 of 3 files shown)


schmanu:
@liliya-soroka: I will merge this into the epic branch. You can then QA test it there.

Issue #1519, opened 2023-01-10
[Switching account] Tx signing with trezor and ledger


francovenica:
## Bug description
Switching to physical devices and trying to execute a tx shows an error initializing the SDK and an error in the send funds modal "unknown account"
I only encounter this issue when changing from any wallet to the trezor or ledger wallet

Note: This issue doens't happen if you connect with the ledger/trezor manually (no switching) and try to sign a tx

## Environment
 - Browser: Chrome
 - Wallet: Ledger NanoS, Trezor One
 - Chain: Goerli
 - Env: https://release_1_3_0--webcore.review-web-core.5afe.dev/gor:0x24F8542Da0C5f267c177a224270442C73804EE80/home
 - RC: #1508 

## Steps to reproduce
 1. Go to app and connect with Trezor or ledger
 2. Connect now with a MM account
 3. Switch back to the ledger/Trezor device
 4. Try to execute a send funds tx

## Expected result
Execution of the tx normally

## Obtained result
Sdk initialization error and "unknown account" error in the send funds modal

## Screenshots
![image](https://user-images.githubusercontent.com/4503659/211681834-4d24a0f3-46fc-4902-b4bb-33c2c5d614b9.png)

katspaugh:
Both signing and executing worked for me with Ledger and MM connected simultaneously.

iamacook:
I just tested this with my Ledger and it signed and executed fine. However, I encountered some other issues.

1. I signed a queued transaction first with my Ledger then switched to MM. Attempting to sign or sign and execute it showed this error:

![image](https://user-images.githubusercontent.com/20442784/211769337-31a6a225-3c61-4e66-bc52-8937ff46d7a4.png)

2. If disconnecting the Ledger and trying to reconnect it, it's not possible to select the network to scan for accounts:

![switch accounts back](https://user-images.githubusercontent.com/20442784/211769581-22dadbc8-2fed-45a4-b5c6-b8584e068855.gif)

katspaugh:
I couldn't reproduce the v4 issue.

iamacook:
> I couldn't reproduce the v4 issue.

It seems to have resolved itself.

I've just tested the above on my Trezor and it works as well.

katspaugh:
@francovenica please try again, maybe it was a temporary RPC issue.

francovenica:
I'm still getting the issue. It happens to me when I have both trezor and ledger connect and switch back and forth

francovenica:
[...]

PR #1522, opened 2023-01-11
Fix: hide Switch wallet button

katspaugh:
## What it solves

There seems to be a problem with SDK init, so I'm commenting out the Switch button until it's fixed.


usame-algan (approved):

katspaugh:
Closing because #1519 seems to be working fine.

PR #1525, opened 2023-01-11
Fix: disconnect wallet if SDK fails to init

### LINKED ISSUE #1519: [Switching account] Tx signing with trezor and ledger

katspaugh:
## What it solves

Resolves #1519

The root cause was that if you dismiss the HW account selection popup, the account is seemingly still connected but the Core SDK fails to init.
And the second part of the problem was that the SDK wouldn't even re-init when that popup is dismissed because the provider object doesn't refresh in that case.

## How this PR fixes it
I fixed the problem by
 1. making the entire wallet object a dependency of the SDK hook (as opposed to just the provider)
 2. the wallet will be disconnected if the SDK fails to init.

## How to test it
* Connect to Ledger
* Attempt to connect to Ledger again, but dismiss the account selection modal w/o selecting an account
* Attempt to execute a tx or a batch

### DIFF
--- a/src/components/common/ConnectWallet/AccountCenter.tsx
+++ b/src/components/common/ConnectWallet/AccountCenter.tsx
@@ -5,7 +5,7 @@ import css from '@/components/common/ConnectWallet/styles.module.css'
 import EthHashInfo from '@/components/common/EthHashInfo'
 import ExpandLessIcon from '@mui/icons-material/ExpandLess'
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
-import useOnboard, { closeAccountSelectionModal, connectWallet } from '@/hooks/wallets/useOnboard'
+import useOnboard, { switchWallet } from '@/hooks/wallets/useOnboard'
 import { useAppSelector } from '@/store'
 import { selectChainById } from '@/store/chainsSlice'
 import Identicon from '@/components/common/Identicon'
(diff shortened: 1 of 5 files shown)


francovenica:
LGTM
Now it disconnects from the physical device if the connection modal is ignored (where you pick network and account).
If you pick network and address the connection is succesful and the user can sign&execute tx normally.

iamacook (approved):

PR #1529, opened 2023-01-12
fix: wait until successful SDK call to set pending state

DiogoSoaress:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/1489

## How this PR fixes it
Waits for a successful SDK interaction to dispatch the tx PENDING state

## How to test it
1. Open a fully signed tx
2. Press Execute
3. Do not approve or reject in the wallet
4. Refresh the page
5. There is no "Submitting" transaction yet

## Analytics changes
N/A

## Screenshots
https://user-images.githubusercontent.com/32431609/212056935-3965af77-002d-48cc-95cd-1eea1adea614.mov

### DIFF
--- a/src/services/tx/tx-sender/dispatch.ts
+++ b/src/services/tx/tx-sender/dispatch.ts
@@ -86,12 +86,11 @@ export const dispatchOnChainSigning = async (safeTx: SafeTransaction, provider: 
   const safeTxHash = await sdkUnchecked.getTransactionHash(safeTx)
   const eventParams = { txId }
 
-  txDispatch(TxEvent.ONCHAIN_SIGNATURE_REQUESTED, eventParams)
-
   try {
     // With the unchecked signer, the contract call resolves once the tx
     // has been submitted in the wallet not when it has been executed
     await sdkUnchecked.approveTransactionHash(safeTxHash)
+    txDispatch(TxEvent.ONCHAIN_SIGNATURE_REQUESTED, eventParams)
   } catch (err) {
     txDispatch(TxEvent.FAILED, { ...eventParams, error: err as Error })
     throw err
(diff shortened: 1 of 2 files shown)


src/services/tx/tx-sender/dispatch.ts
```diff
@@ -115,12 +115,11 @@ export const dispatchTxExecution = async (
   const sdkUnchecked = await getUncheckedSafeSDK(provider)
   const eventParams = txId ? { txId } : { groupKey: await sdkUnchecked.getTransactionHash(safeTx) }
 
-  txDispatch(TxEvent.EXECUTING, eventParams)
-
   // Execute the tx
   let result: TransactionResult | undefined
   try {
     result = await sdkUnchecked.executeTransaction(safeTx, txOptions)
+    txDispatch(TxEvent.EXECUTING, eventParams)
```
katspaugh, line 122: So basically if I don't confirm or reject the execution, this won't be called?

DiogoSoaress: Yes. But I think the subscribing problem still persists. After the refresh I don't see the "Submitting" state

katspaugh: Well, at least it will pick it up from the history if it's executed. Or not if it's rejected.

katspaugh (approved):
👍

iamacook (changes requested):
There are other dispatchers that still dispatch events before calling the SDK. We should probably change those as well.

francovenica:
Issue:
The issue still happens if you use the "Execute Batch" button

Create 2 fully singed tx
Click the Execute Batch button
Press "sign" so the Wallet pops up but don't sign
Refresh the page

image

DiogoSoaress:
@francovenica I'm moving it to "In Progress" to address Aaron's request that should fix that case

francovenica:
I've checked again and the issue was fixed. 
If it passes peer review in can be put in the QA done column

iamacook (changes requested):
We are still dispatching `TxEvent.ONCHAIN_SIGNATURE_REQUESTED` before calling the SDK.

If the on-chain signing request is rejected, it would cause the same bug but you won't see it in the UI as we don't set on-chain signatures as pending. Nonetheless, we should move it for the sake of consistency.

iamacook (approved):

PR #1531, opened 2023-01-12
feat: set fallback handler

### LINKED ISSUE #1318: Add option to update `CompatibilityFallbackHandler`
## What is the feature about

Setting the `CompatibilityFallbackHandler` (`CFH`) is not a necessity when deploying a Safe. This means that some "normal" functionality may not be available, e.g. [EIP-1271 signature validation](https://github.com/safe-global/safe-contracts/blob/main/contracts/handler/CompatibilityFallbackHandler.sol#L28). As the address needs to be manually set, it also allows for custom `CFH`s to be set in its place.

We should add a CRUD in the Safe settings to allow for the modification of the `CFH` address/setting it to the "recommended" deployment address. It would also be beneficial to show what it entails and whether one is set or not.

## The list of requirements

- Add a new section to the settings displaying the current  `CFH` deployment address used.
- Display associated information regarding what the  `CFH` is and what address is currently set/whether it is the "recommended" one.
- Add a CRUD modal flow to create a transaction that modifies the set `CFH`.
- The fallback handler section in the settings should only be visible if the current contract version supports fallback handlers

## Designs/sketches

tbd

## Links

- [Discussion thread](https://5afe.slack.com/archives/C03DAGWJCR5/p1670412974409759) regarding missing `CFH`s during early development of `web-core` resulting in this issue.

schmanu:
## What it solves
- Shows the fallback handler address in Settings / Setup page
- If no fallback handler is set, offers a new dialog to set the recommended fallback handler.

Resolves #1318 

## How to test it
- Open settings page and see the current fallback handler contract address
- Remove the fallback handler from a Safe and try the new "Set fallback handler" button / dialog.

## Analytics changes
tbd

## Screenshots
tbd

### DIFF
--- a/src/components/settings/FallbackHandlerVersion/SetFallbackHandlerDialog.tsx
+++ b/src/components/settings/FallbackHandlerVersion/SetFallbackHandlerDialog.tsx
@@ -1,0 +1,76 @@
+import { Box, Button, Typography } from '@mui/material'
+import { useState } from 'react'
+
+import TxModal from '@/components/tx/TxModal'
+
+import useTxSender from '@/hooks/useTxSender'
+import useAsync from '@/hooks/useAsync'
+
+import SignOrExecuteForm from '@/components/tx/SignOrExecuteForm'
+import type { TxStepperProps } from '@/components/tx/TxStepper/useTxStepper'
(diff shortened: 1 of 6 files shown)


katspaugh:
We decided not to include this at this time, although personally I find this a nice feature.

PR #1537, opened 2023-01-13
[WIP] New safe apps Section design

DaniSomoza:
## What it solves

Resolves #


src/tests/pages/apps.test.tsx
```diff
@@ -84,145 +42,291 @@ describe('AppsPage', () => {
...
+
+        // shows custom safe App App Url input
+        const customSafeAppURLInput = screen.getByLabelText(/App URL/)
+        expect(customSafeAppURLInput).toBeInTheDocument()
       })
     })
 
-    it('allows adding custom apps', async () => {
+    it('Adds a Custom Safe App', async () => {
       const APP_URL = 'https://apps.gnosis-safe.io/compound'
```
dasanra, line 209: There are still some elements using the old domain. As this is a mock shouldn't cause issues but just in case maybe it's better to update. Also in terms of marketing

DaniSomoza: done! :3

package.json
```diff
@@ -13,8 +13,9 @@
     "lint:fix": "next lint --fix",
     "prettier": "prettier -w \"{src,cypress,mocks,scripts}/**/*.{ts,tsx,css,js}\"",
     "fix": "yarn lint:fix && ts-prune && yarn prettier",
-    "test": "cross-env TZ=CET DEBUG_PRINT_LIMIT=30000 jest",
+    "test": "cross-env TZ=CET DEBUG_PRINT_LIMIT=30000 jest --watchAll",
```
mikhailxyz, line 16: is the `--watchAll` addition intended?

src/components/common/PageLayout/SideDrawer.tsx
```diff
@@ -17,7 +17,7 @@ type SideDrawerProps = {
 }
 
 const isSafeAppRoute = (pathname: string, query: ParsedUrlQuery): boolean => {
-  return pathname === AppRoutes.apps && !!query.appUrl
+  return pathname === AppRoutes.apps.index && !!query.appUrl
```
mikhailxyz, line 20: now when the function only checks for the index route, should it be renamed too?

src/components/dashboard/FeaturedApps/FeaturedApps.tsx
```diff
@@ -7,7 +7,7 @@ import NextLink from 'next/link'
 import { AppRoutes } from '@/config/routes'
 import { SafeAppsTag } from '@/config/constants'
 import { useRemoteSafeApps } from '@/hooks/safe-apps/useRemoteSafeApps'
-import SafeAppIcon from '@/components/safe-apps/SafeAppIcon'
+import SafeAppIconCard from '@/components/safe-apps/SafeAppIconCard/SafeAppIconCard'
```
mikhailxyz, line 10: why the tautology? `SafeAppIconCard/SafeAppIconCard`

src/components/dashboard/SafeAppsDashboardSection/SafeAppsDashboardSection.tsx
```diff
@@ -6,12 +6,15 @@ import Button from '@mui/material/Button'
 
 import { WidgetContainer } from '../styled'
 import { useSafeApps } from '@/hooks/safe-apps/useSafeApps'
-import { AppCard, AppCardContainer } from '@/components/safe-apps/AppCard'
+import useSafeAppPreviewDrawer from '@/hooks/safe-apps/useSafeAppPreviewDrawer'
+import SafeAppPreviewDrawer from '@/components/safe-apps/SafeAppPreviewDrawer/SafeAppPreviewDrawer'
```
mikhailxyz, line 10: why the tautology? `SafeAppPreviewDrawer/SafeAppPreviewDrawer`

src/components/safe-apps/AddCustomSafeAppCard/AddCustomSafeAppCard.tsx
```diff
@@ -1,41 +1,39 @@
...
+          <AddCustomAppIcon alt="Add Custom Safe App card" />
+
+          {/*  Add Custom Safe App Button */}
           <Button
             variant="contained"
             size="small"
+            onClick={() => setAddCustomAppModalOpen(true)}
             sx={{
-              mt: 1,
+              mt: '24px',
```
mikhailxyz, line 28: why hardcode the value instead of using the scaling factor?

src/components/safe-apps/AddCustomSafeAppCard/AddCustomSafeAppCard.tsx
```diff
@@ -46,4 +44,4 @@ const AddCustomAppCard = ({ onSave, safeAppList }: Props) => {
   )
 }
 
-export { AddCustomAppCard }
+export default AddCustomSafeAppCard
```
mikhailxyz, line 47: not a big fan of named > default export change: https://basarat.gitbook.io/typescript/main-1/defaultisbad

no need to change, just sharing why it was done the way it was done before

src/pages/apps/index.tsx
```diff
@@ -40,6 +51,7 @@ const Apps: NextPage = () => {
     remoteSafeAppsLoading,
   })
 
+  // TODO: add a loading state
```
mikhailxyz, line 54: does this need design? if yes i'd create a github issue instead so we're less likely to forget

src/components/safe-apps/SafeAppsSDKLink/SafeAppsSDKLink.tsx
```diff
@@ -0,0 +1,51 @@
...
+import Fab from '@mui/material/Fab'
+import KeyboardDoubleArrowUpRoundedIcon from '@mui/icons-material/KeyboardDoubleArrowUpRounded'
+
+import CodeIcon from '@/public/images/apps/code-icon.svg'
+import css from './styles.module.css'
+
+const SafeAppsSDKLink = () => {
+  const [openModal, setOpenModal] = useState<boolean>(false)
+
+  const safeAppsSDKUrl = 'https://docs.gnosis-safe.io/learn/safe-tools/sdks/safe-apps'
```
mikhailxyz, line 16: Would it be better to have this as a field in the config service? At this stage, we already will require one deployment in the future to change the link to a new domain + with the addition of Tomiwa docs are expected to improve and therefore change a lot

dasanra: Maybe too much to create a dynamic field in the config service for something that shouldn't change that often.

In this case it's true we will need to update it once the new domain and docs are ready but then we shouldn't

What I would definitely do is to move this to a constant in this file, so it's easier to track
https://github.com/safe-global/web-core/blob/dev/src/config/constants.ts

JagoFigueroa:
Hola! A couple of questions probably for you @liliiaorlenko 😄 

Should we limit the number of tags shown per app to avoid scenarios like this one which could impact the presentation for the whole row? (probably not a lot of apps to be expected with more than 3-4 tags but just in case)

Screenshot 2023-02-01 at 16 04 33


When searching for apps via category name I see that apps will be shown as long as they include 1 of the selected categories. Is this expected or should do a more strong filtering and only show apps including all of the selected categories?

Screenshot 2023-02-01 at 15 50 05

Cheers!

src/components/common/PageLayout/SideDrawer.tsx
DaniSomoza: I think that the name should be the same because this function checks if you are in the Safe App page, and the url is always like: 
```
.../apps?appUrl=https%3A%2F%2Fswap.cow.fi
```

package.json
DaniSomoza: oh! nope, I prefer the `--watchAll` flag to work with test. I will remove it

src/components/dashboard/FeaturedApps/FeaturedApps.tsx
DaniSomoza: it is just because `SafeAppIconCard` component is inside the `SafeAppIconCard` folder.

src/components/dashboard/SafeAppsDashboardSection/SafeAppsDashboardSection.tsx
DaniSomoza: same as `SafeAppIconCard` component

src/components/safe-apps/AddCustomSafeAppCard/AddCustomSafeAppCard.tsx
DaniSomoza: done!  `mt: '24px'`  ->  `mt: 3`

JagoFigueroa:
Mis dieses señores, todo bueno from me 😄

src/pages/apps/index.tsx
DaniSomoza: I removed the comment!

src/components/dashboard/FeaturedApps/FeaturedApps.tsx
mikhailxyz: I get why it happens, My question is related to why not follow how other components are stored `folder name/index.tsx` which allows you to only specify the folder name in the import statement

src/components/common/PageLayout/SideDrawer.tsx
mikhailxyz: But now there are new subroutes:
```
    bookmarked: '/apps/bookmarked',
    custom: '/apps/custom',
```

They are technically Safe App routes, aren't they? But the function would return false for them. Maybe `isSafeAppFrameRoute` is a better name

package.json
```diff
@@ -13,7 +13,7 @@
     "lint:fix": "next lint --fix",
     "prettier": "prettier -w \"{src,cypress,mocks,scripts}/**/*.{ts,tsx,css,js}\"",
     "fix": "yarn lint:fix && ts-prune && yarn prettier",
-    "test": "cross-env TZ=CET DEBUG_PRINT_LIMIT=30000 jest --watchAll",
```
katspaugh: Why remove watchAll?

DaniSomoza: Because in the `dev` & `main` branches the `test` command do not have the ` --watchAll` flag. I added it by mistake because for me its more confortable work with that flag enabled.

see: https://github.com/safe-global/web-core/blob/dev/package.json#L16

src/components/common/PageLayout/SideDrawer.tsx
DaniSomoza: done!

src/components/dashboard/FeaturedApps/FeaturedApps.tsx
DaniSomoza: done!

src/components/dashboard/SafeAppsDashboardSection/SafeAppsDashboardSection.tsx
DaniSomoza: done! renamed from `SafeAppPreviewDrawer.tsx` to `index.tsx`

src/components/safe-apps/SafeAppActionButtons/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+      )}
+    </Box>
+  )
+}
+
+export default SafeAppActionButtons
+
+export const getShareSafeAppUrl = (router: NextRouter, appUrl: string, currentChain?: ChainInfo) => {
+  const shareUrlObj: UrlObject = {
+    protocol: typeof window !== 'undefined' ? window.location.protocol : '',
```
katspaugh, line 99: This might cause hydration issues because you're accessing the window object outside of an effect.

src/components/safe-apps/SafeAppSocialLinksCard/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const hasSocialLinks = socialProfiles?.length > 0
+
+  if (!hasSocialLinks && !developerWebsite) {
+    return null
+  }
+
+  const discordSocialLink = socialProfiles.find(({ platform }) => platform === SafeAppSocialPlatforms.DISCORD)
+  const twitterSocialLink = socialProfiles.find(({ platform }) => platform === SafeAppSocialPlatforms.TWITTER)
+  const githubSocialLink = socialProfiles.find(({ platform }) => platform === SafeAppSocialPlatforms.GITHUB)
```
katspaugh, line 32: Probably worth making a little function to avoid code repetition.

katspaugh (commented):
Can we somehow avoid adding two pairs of each icon? The only difference is the color and you can pass that as `currentColor`.

Thanks for writing comments in the code, btw 👍

src/components/safe-apps/SafeAppActionButtons/index.tsx
DaniSomoza: I am just using the same logic that [its defined in our `AppCard.tsx` component in dev](https://github.com/safe-global/web-core/blob/dev/src/components/safe-apps/AppCard/index.tsx#L194) (I just created a `getShareSafeAppUrl` function to abstract it). Do you think it's worth to update it?

katspaugh: Ah, let's keep it then, I'll make a tech debt issue.

katspaugh (approved):

mikhailxyz (approved):
good job señor

dasanra (approved):

PR #1540, opened 2023-01-13
New safe apps design e2e

yagopv:
## What it solves
Fix e2e with the new changes and adds spec for the drawer


JagoFigueroa:
Tests are passing locally for me without issue so todo bueno from my side 🖖

PR #1500, opened 2023-01-09
WIP: fix: Create safe e2e test

usame-algan:
WIP

PR #1543, opened 2023-01-13
fix: Extend create safe e2e test

usame-algan:
## What it solves

Follow up on #1500

Tests more cases during safe creation flow:

- Changing network
- Removing an owner
- Saving data even when going back


katspaugh (approved):
👍

PR #1547, opened 2023-01-16
fix: GA tracking wrong eventLabels, missing event type, overtracking wallet connect event

usame-algan:
## What it solves

- Fixes overtracking introduced with #1496 
- Sends the meta event type as a custom dimension to GA for making it easier to filter
- Resets `eventLabel` for events that don't use it to prevent pollution

## How to test it

1. Open the Safe
2. Connect with MM
3. Observe only one meta event for connecting the wallet
4. Observe a new field `eventType` in GTM that is only set with meta events
5. Go to Assets and hide a token
6. Observe `eventLabel` with an undefined value in GTM

## Analytics changes

A new optional attribute `eventType` was added to the data layer

### DIFF
--- a/src/services/analytics/gtm.ts
+++ b/src/services/analytics/gtm.ts
@@ -79,6 +79,7 @@ type ActionGtmEvent = GtmEvent & {
   eventCategory: string
   eventAction: string
   eventLabel?: EventLabel
+  eventType?: EventType
 }
 
 type PageviewGtmEvent = GtmEvent & {
@@ -103,8 +104,17 @@ export const gtmTrack = (eventData: AnalyticsEvent): void => {
     eventAction: eventData.action,
   }
 
+  if (eventData.event === EventType.META) {
+    gtmEvent.eventType = EventType.META
+  } else {
+    gtmEvent.eventType = undefined
+  }
+
   if (eventData.label) {
     gtmEvent.eventLabel = eventData.label
+  } else {
+    // Otherwise, whatever was in the datalayer before will be reused
+    gtmEvent.eventLabel = undefined
   }
 
   const abTest = getAbTest()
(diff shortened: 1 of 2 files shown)


src/services/analytics/gtm.ts
```diff
@@ -103,8 +104,17 @@ export const gtmTrack = (eventData: AnalyticsEvent): void => {
     eventAction: eventData.action,
   }
 
+  if (eventData.event === EventType.META) {
```
usame-algan, line 107: We could also track the eventType in general not just if its a meta event

katspaugh (approved):
Good catch, thanks for fixing my bug!

francovenica:
LGTM
Compared current prod build with the PR
in prod everytime you connect or switch wallets 2 events trigger "WalletConnectPeer" and "Connect Wallet" both with an "eventType: undefined"
In this PR only the "Conect wallet" event is triggered and the "eventyType : metadata" is set

Question: 
Tried the hide token event to see the eventLabel as undefined. 
When you hide a token 2 events are triggered, "Customclick" and "Hide token". Custom click always has en eventLabel of undefined, and HideToken seems to have an eventLabel that shows how many tokens are currently hidden (so if no token is hidden it will be undefined, if 3 tokens are hidden after a "save" action then the eventLabel will be "number: 3"
The line in the description is referring to that CustomClick event only?

usame-algan:
> The line in the description is referring to that CustomClick event only?
It should work for any event we send. Whenever there is no `eventLabel` included it will be reset to undefined.

src/services/analytics/gtm.ts
usame-algan: I adjusted this to be more generalized and track the event type in all cases. Don't see a strong reason to only do it for metadata.

francovenica:
Then is fine.
QA passed

PR #1552, opened 2023-01-17
fix: `merge` hydration of persisted state

### LINKED ISSUE #1551: Merge `initialState` and persisted state on hydration
## Bug description

We currently overwrite slice `initialState` on [persisted state hydration](https://github.com/safe-global/web-core/blob/dev/src/store/storeHydrator.tsx#L25). This means [it overwrites everything](https://github.com/safe-global/web-core/blob/dev/src/store/index.ts#L63-L68), e.g. potential new keys.:

```ts
const initialState = {
  test: {
    foo: true,
    bar: false
  }
};
```

```ts
const persistedState = "Hello world!";
```

```ts
const state = "Hello world!";
```

In the above example, accesing `state.test.foo` would crash the app.

## Steps to reproduce

Create a new nested object in the `settingsSlice` `intialState` and attempt to access it via a selector *without* optional chaining.

## Expected result

The app doesn't crash and it shows the value, with the value populated in the state.

## Obtained result

The app crashes and the value is not present in the state.

## Potential solution

Instead of shallow spreading the persisted state over the entire store, it would be better to leverage [`merge` from `lodash`](https://lodash.com/docs/4.17.15#merge) in the `action.payload`.

```ts
this.store.dispatch({
  type: HYDRATE_ACTION,
  payload: merge({}, initialState, getPersistedState()),
})
```

iamacook:
## What it solves

Resolves #1551

## How this PR fixes it

Hydrated state is now merged with the initial state.

## How to test it

1. Run the project locally.
2. Add a new key to a slice's `initialState`.
3. Reload the app and observe the new key present in the state in Redux DevTools.

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -6,6 +6,7 @@ import {
   type AnyAction,
 } from '@reduxjs/toolkit'
 import { useDispatch, useSelector, type TypedUseSelectorHook } from 'react-redux'
+import merge from 'lodash/merge'
 import { IS_PRODUCTION } from '@/config/constants'
 import { createStoreHydrator, HYDRATE_ACTION } from './storeHydrator'
 import { chainsSlice } from './chainsSlice'
@@ -61,10 +62,8 @@ export const getPersistedState = () => {
 
 const hydrationReducer: typeof rootReducer = (state, action) => {
   if (action.type === HYDRATE_ACTION) {
-    return {
-      ...state,
-      ...action.payload,
-    }
+    // `merge` mutates the first argument, so we need to create a new object
+    return merge({}, state, action.payload)
   }
   return rootReducer(state, action)
 }
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -60,7 +60,7 @@ export const getPersistedState = () => {
   return getPreloadedState(persistedSlices)
 }
 
-const hydrationReducer: typeof rootReducer = (state, action) => {
+export const _hydrationReducer: typeof rootReducer = (state, action) => {
   if (action.type === HYDRATE_ACTION) {
     // `merge` mutates the first argument, so we need to create a new object
     return merge({}, state, action.payload)
@@ -70,7 +70,7 @@ const hydrationReducer: typeof rootReducer = (state, action) => {
 
 const makeStore = (initialState?: Record<string, any>) => {
   return configureStore({
-    reducer: hydrationReducer,
+    reducer: _hydrationReducer,
     middleware: (getDefaultMiddleware) => getDefaultMiddleware().concat(middleware),
     devTools: !IS_PRODUCTION,
     preloadedState: initialState,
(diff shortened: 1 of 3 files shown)


iamacook:
Just converted this back to a draft to add a test.

src/store/index.ts
```diff
@@ -61,10 +62,8 @@ export const getPersistedState = () => {
 
 const hydrationReducer: typeof rootReducer = (state, action) => {
   if (action.type === HYDRATE_ACTION) {
-    return {
-      ...state,
-      ...action.payload,
-    }
+    // `merge` mutates the first argument, so we need to create a new object
+    return merge({}, state, action.payload)
```
katspaugh, line 66: Merging is walking on thin ice, what if the stored data is in conflict with the new data?

Can you explain a little bit more about how new data may appear in the state before the hydration and why just avoiding setting undefined is not enough to fix the problem?

iamacook: It's related to [this comment](https://github.com/safe-global/web-core/pull/1527#discussion_r1071025396).

If we extend the intial state of a slice, the default values are currently [overwritten by the persisted state](https://github.com/safe-global/web-core/blob/dev/src/store/index.ts#L63-L68). We will alternatively have to manually assign "extensions" of initial state in dispatched actions if they are not present.

katspaugh: I'm aware, but the problem seems to revolve around "undefined" being retrieved from the LS.

iamacook: After discussing this and confirming that is **only** related to new initial state properties, I've added an explanatory comment and a further test. Lmk what you think.

src/store/index.ts
```diff
@@ -59,19 +60,25 @@ export const getPersistedState = () => {
...
   if (action.type === HYDRATE_ACTION) {
-    return {
-      ...state,
-      ...action.payload,
-    }
+    /**
+     * When changing the schema of a Redux slice, previously stored data in LS might become incompatible.
+     * To avoid this, we should always migrate the data on a case-by-case basis in the corresponding slice.
+     * However, as a catch-all measure, we attempt to merge the stored data with the initial Redux state,
+     * so that any new properties added to the slice are not overwritten as undefined.
```
katspaugh, line 69:
```suggestion
     * so that any newly added properties in the initial state are preserved, and existing properties are taken from the LS.
```

A small edit.

katspaugh (approved):
Looks good as an intermediate solution for `undefined` and added keys.
Will work in most cases.

katspaugh (changes requested):
I guess this [`| undefined` in the settings slice](https://github.com/safe-global/web-core/blob/dev/src/store/settingsSlice.ts#L13) can be removed now?

Did you check what happens when there's nothing persisted yet for a certain key? It shouldn't overwrite anything then.

katspaugh (approved):
Perfect, thanks!

PR #1555, opened 2023-01-17
feat: use EIP1271 feature toggle per chain

### LINKED ISSUE #1532: [EIP-1271] Hide Messages behind a feature toggle
Create a new feature toggle and hide/show the Messages tab based on that. Hide in the prod config for all chains.

schmanu:
## What it solves
Uses new Feature toggle for EIP-1271 off chain messages.

Resolves #1532 

## How to test it
On Staging:
Check a Safe on görli => It shows messages
Check a Safe on mainnet => it shows no messages.

Currently the feature is only available on görli

## Analytics changes
None

### DIFF
--- a/src/pages/transactions/navigation.tsx
+++ b/src/pages/transactions/navigation.tsx
@@ -1,0 +1,15 @@
+import NavTabs from '@/components/common/NavTabs'
+import { transactionNavItems } from '@/components/sidebar/SidebarNavigation/config'
+import { useCurrentChain } from '@/hooks/useChains'
+import { hasFeature } from '@/utils/chains'
+import { FEATURES } from '@safe-global/safe-gateway-typescript-sdk'
+
+export const TransactionNavigation = () => {
+  const chain = useCurrentChain()
+  debugger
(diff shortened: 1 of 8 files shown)


src/pages/transactions/navigation.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export const TransactionNavigation = () => {
+  const chain = useCurrentChain()
+  debugger
+  const isEIP1271 = chain && hasFeature(chain, FEATURES.EIP1271)
+
+  return (
+    <NavTabs tabs={isEIP1271 ? transactionNavItems : transactionNavItems.filter((item) => item.label !== 'Messages')} />
+  )
+}
```
katspaugh, line 15: You shouldn't put non-page components in `src/pages`.
Build fails with
```
> Build optimization failed: found page without a React Component as default export in 
[86](https://github.com/safe-global/web-core/actions/runs/3941723183/jobs/6744492319#step:6:88)
pages/transactions/navigation
``

katspaugh (approved):
Looks good!

I'm starting to think we should move the features enum to web-core and type it as string on the SDK side. Next time.

PR #1556, opened 2023-01-17
feat: Added safes pending actions

DiogoSoaress:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/1534

Notion doc:
https://www.notion.so/gnosis-safe/Pending-actions-for-added-Safes-c03e8e6cb7bc4826a266757d0011831d?d=d4529b4c688c45bcabd9814fd8345f34#e7db9d95e9c94fe9a9fb431050e44f24

## How this PR fixes it
On opening the Safe list sidebar, requests the queued transactions for each added Safe
A badge with the queued transactions displays next to the SafeIcon with a tooltip "`3` pending transaction(s)"
Adds tracking for the 2 newly added buttons
Increases the Safe list drawer width

## How to test it
1. Open the Safes list sidebar
2. If any of your added Safes has pending transactions, an icon displays next to the Safe icon with that number of queued transactions
3. If any of your added Safes is owned by the connected account, a wallet icon with the number of transactions you can sign are displayed next to the Safe list item in the Safe list

## Screenshots
Screenshot 2023-01-25 at 12 52 44

## Figma/ Designs
https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=3800%3A79205&t=h0jessQxMON1eXvO-0


public/images/common/update.svg
```diff
@@ -0,0 +1,4 @@
...
+<path fill-rule="evenodd" clip-rule="evenodd" d="M12.9435 9.27051C12.4485 9.03151 11.8495 9.23951 11.6085 9.73651C10.9445 11.1115 9.52952 11.9995 8.00452 11.9995C6.66052 11.9995 5.45352 11.3405 4.72052 10.2845L6.24752 10.4645C6.79752 10.5345 7.29252 10.1375 7.35752 9.58851C7.42252 9.04151 7.03052 8.54251 6.48152 8.47851L2.84652 8.04851C2.33852 7.98751 1.85152 8.33451 1.74952 8.84551L1.01952 12.4925C0.911521 13.0345 1.26252 13.5605 1.80352 13.6685C1.87052 13.6825 1.93552 13.6885 2.00152 13.6885C2.46752 13.6885 2.88552 13.3605 2.98052 12.8845L3.23252 11.6295C4.34052 13.0935 6.07852 13.9995 8.00452 13.9995C10.2915 13.9995 12.4135 12.6675 13.4095 10.6055C13.6495 10.1085 13.4415 9.51051 12.9435 9.27051Z" fill="currentColor"/>
+</svg>
```
katspaugh, line 4: Can we use the Autorenew icon from MUI instead?

src/components/common/SafeIcon/index.tsx
```diff
@@ -3,27 +3,36 @@ import { Box } from '@mui/material'
...
   owners: number | string
 }
 const Threshold = ({ threshold, owners }: ThresholdProps): ReactElement => (
-  <Box className={css.threshold} sx={{ color: ({ palette }) => palette.static.main }}>
+  <Box className={classnames(css.badge, css.threshold)} sx={{ color: ({ palette }) => palette.static.main }}>
     {threshold}/{owners}
   </Box>
 )
 
+const PendingActionsBadge = ({ queuedTxs }: { queuedTxs: number }): ReactElement => (
```
katspaugh, line 18: The badge isn't super clear IMHO.
I would experiment with adding a line of text under each Safe row, literally saying "N transactions need signatures".

src/components/sidebar/SafeList/index.tsx
```diff
@@ -75,27 +79,51 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
   const hasNoSafes = Object.keys(ownedSafes).length === 0 && Object.keys(addedSafes).length === 0
   const isWelcomePage = router.pathname === AppRoutes.welcome
 
+  const handleFetchQueued = async () => {
+    const fetchedQueuedTxs: Record<string, number> = {}
+
+    Object.entries(addedSafes).forEach(([chainId, safes]) => {
+      Object.keys(safes).forEach(async (safeAddress) => {
+        const result = await getTransactionQueue(chainId, safeAddress)
+        fetchedQueuedTxs[safeAddress] = result.results.filter(isTransactionListItem).length
```
katspaugh, line 88: It's fine for the safe of a prototype, but keep in mind this is paginated, so if there's `result.next`, we should show `> 10`.

public/images/common/update.svg
DiogoSoaress: We are moving away from MUI icons AFAIK but cc @liliiaorlenko

DiogoSoaress:
Changes proposed by Design:

- [ ] colour update button to blue
- [ ] widen the safe list for 20-30px (so buttons and other info would have some space)
- [ ] add a tooltip on a number of pending actions that says: “pending actions at this safe”

public/images/common/update.svg
liliiaorlenko: yes we have our own library so pls align with it https://www.figma.com/file/hsPhJR6ErHx0O4j71E9Dd5/Icon-Library?node-id=1008%3A223&t=7zTIwetwK0nXHQJ3-1

src/components/sidebar/SafeList/index.tsx
DiogoSoaress: I've push a commit that handles it

src/components/common/SafeIcon/index.tsx
```diff
@@ -1,29 +1,40 @@
...
 }
 const Threshold = ({ threshold, owners }: ThresholdProps): ReactElement => (
-  <Box className={css.threshold} sx={{ color: ({ palette }) => palette.static.main }}>
+  <Box className={classnames(css.badge, css.threshold)} sx={{ color: ({ palette }) => palette.static.main }}>
     {threshold}/{owners}
   </Box>
 )
 
+const PendingActionsBadge = ({ queuedTxs }: { queuedTxs: string }): ReactElement => (
+  <Tooltip title="queued transactions at this Safe" placement="top">
```
katspaugh, line 19: `at Safe` is not grammatically correct.

```suggestion
  <Tooltip title={`${queuedTxs} pending transactions`} placement="top">
```

iamacook (commented):
A suggestion for exploration: poll for the queued transactions of all added Safes when the Safe loads or the sidebar opens and then cache it.

src/components/sidebar/SafeList/index.tsx
```diff
@@ -75,27 +79,59 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
   const hasNoSafes = Object.keys(ownedSafes).length === 0 && Object.keys(addedSafes).length === 0
   const isWelcomePage = router.pathname === AppRoutes.welcome
 
+  const handleFetchQueued = async () => {
+    const fetchedQueuedTxs: Record<string, string | undefined> = {}
+
+    for (let [chainId, safes] of Object.entries(addedSafes)) {
+      for (let safeAddress of Object.keys(safes)) {
```
katspaugh, line 86: You could omit the current safe and use `useTxQueue` instead. It's already loaded for the current safe.

src/components/sidebar/SafeList/index.tsx
```diff
@@ -75,27 +79,59 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
...
+          <div className={css.headerButtons}>
+            <Button
+              disableElevation
+              size="small"
+              variant="outlined"
+              onClick={handleFetchQueued}
+              startIcon={<SvgIcon component={UpdateIcon} inheritViewBox fontSize="small" />}
+              sx={{ color: ({ palette }) => palette.info.main }}
+            >
+              Update
```
usame-algan, line 128: Did we explore other namings instead of "Update"? Imo it is not clear what it will do. I've seen other apps call it "Sync addresses" for example.

src/components/sidebar/SafeList/index.tsx
```diff
@@ -75,27 +79,59 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
...
+            My Safes
+          </Typography>
+          <div className={css.headerButtons}>
+            <Button
+              disableElevation
+              size="small"
+              variant="outlined"
+              onClick={handleFetchQueued}
+              startIcon={<SvgIcon component={UpdateIcon} inheritViewBox fontSize="small" />}
+              sx={{ color: ({ palette }) => palette.info.main }}
```
usame-algan, line 126: The blue outlined button is hard to read, especially in light mode. Not sure if its only an issue on my screen but running it through a contrast checker also fails: https://webaim.org/resources/contrastchecker/?fcolor=5FDDFF&bcolor=FFFFFF

DiogoSoaress: cc @liliiaorlenko

src/components/sidebar/SafeList/index.tsx
DiogoSoaress: I will raise this with @johannesmoormann to think of a better name that doesn't blow the button size

iamacook: We could poll when the sidebar opens instead of requiring user input (if the last poll after a defined period to prevent spam).

src/components/sidebar/SafeList/index.tsx
DiogoSoaress: I've skipped the request for the currentSafe and use the data from `useTxQueue`

src/components/sidebar/SafeList/index.tsx
DiogoSoaress: I've removed the button. Now the queues are requested when opening the Safe list sidebar. I've updated the PR description.

public/images/common/update.svg
DiogoSoaress: The button containing the icon was removed

src/components/sidebar/SafeList/index.tsx
```diff
@@ -75,27 +82,69 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
...
+        }
+
+        try {
+          const result = await getTransactionQueue(chainId, safeAddress)
+
+          if (result.results.length === 0) continue
+
+          fetchedQueuedTxs[safeAddress] = `${result.results.filter(isTransactionListItem).length}${
+            result.next ? '+' : ''
+          }`
```
katspaugh, line 110: Can we avoid the duplicate code here?

DiogoSoaress: Yes. Created a small function to avoid the duplication.

src/components/sidebar/SafeList/index.tsx
```diff
@@ -75,27 +81,66 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
...
-                variant="outlined"
-                onClick={closeDrawer}
-                startIcon={<SvgIcon component={AddIcon} inheritViewBox fontSize="small" />}
-              >
-                Add
-              </Button>
-            </Link>
-          </Track>
-        )}
+      <div className={css.headerWrapper}>
```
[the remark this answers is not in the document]
DiogoSoaress: This was to make the Safe List header sticky. But it can be removed as it an unnecessary change.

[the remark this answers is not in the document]
DiogoSoaress: I've removed it

liliiaorlenko:
> > What do you think about the badge on the bottom right?
> 
> cc @liliiaorlenko

Is there some problem with previous position,  but I think on a left its better

iamacook:
> Is there some problem with previous position, but I think on a left its better

It was a preference suggestion.

usame-algan (approved):
👍

francovenica:
Issue:
Several tx with the same nonce are counted as separated tx (for example a tx and the rejection of that tx). I think the number should reflect how many different nonces you have in the queue tab

Create a tx.
Then create the rejection of that tx or a replacement
Check the sidebar

Snapshot:
Those 2 tx are the original tx and its rejection, same nonce
image

src/components/sidebar/SafeList/index.tsx
```diff
@@ -75,12 +88,83 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
...
+      }
+
+      txsActionsByChain[chainId] = {
+        ...(txsActionsByChain[chainId] ?? {}),
+        [address]: {
+          ...txsActionsByChain[chainId]?.[address],
+          queued: `${page.results.filter(isTransactionListItem).length}${page.next ? '+' : ''}`,
+        },
+      }
+    }
```
iamacook, line 127: I'd suggest extracting this and adding a test. It may solve the slow iteration issues by assigning values directly and not spreading.

public/images/common/check.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M8.99901 19C8.73301 19 8.47901 18.895 8.29201 18.707L2.29301 12.704C1.90201 12.313 1.90201 11.68 2.29301 11.29C2.68401 10.899 3.31601 10.899 3.70701 11.29L8.99901 16.586L20.291 5.29301C20.682 4.90201 21.314 4.90201 21.705 5.29301C22.096 5.68401 22.096 6.31601 21.705 6.70701L9.70601 18.707C9.51901 18.895 9.26501 19 8.99901 19Z" fill="currentColor"/>
+<path d="M8.99901 19V19C8.73301 19 8.47901 18.895 8.29201 18.707L2.29301 12.704C1.90201 12.313 1.90201 11.68 2.29301 11.29C2.68401 10.899 3.31601 10.899 3.70701 11.29L8.99901 16.586L20.291 5.29301C20.682 4.90201 21.314 4.90201 21.705 5.29301C22.096 5.68401 22.096 6.31601 21.705 6.70701L9.70601 18.707C9.51901 18.895 9.26501 19 8.99901 19" stroke="currentColor"/>
+</svg>
```
katspaugh, line 4: Why are we not reusing the existing confirm icon?
Screenshot 2023-01-25 at 13 04 02

src/components/sidebar/SafeList/index.tsx
```diff
@@ -59,13 +64,55 @@ export const _shouldExpandSafeList = ({
...
+  txsActionsByChain[chainId] = {
+    ...(txsActionsByChain[chainId] ?? {}),
+    [address]: {
+      ...txsActionsByChain[chainId]?.[address],
+      queued: `${page.results.filter(isTransactionListItem).length}${page.next ? '+' : ''}`,
+    },
+  }
+
+  return txsActionsByChain
+}
```
katspaugh, line 99: Please move this function to a separate file.

src/components/sidebar/SafeList/index.tsx
```diff
@@ -75,12 +122,51 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
...
+    setSafeTxsActions(txsActionsByChain)
+  }, [addedSafes, currentChainId, currentSafeAddress, ownedSafes, page, safeTxsActions, wallet?.address])
+
+  useEffect(() => {
+    handleFetchQueued()
+  }, [handleFetchQueued])
+
+  useEffect(() => {
+    setSafeTxsActions(undefined)
+  }, [wallet?.address])
```
katspaugh, line 161: Move to a hook.

src/services/analytics/events/overview.ts
```diff
@@ -68,4 +68,12 @@ export const OVERVIEW_EVENTS = {
     action: 'Open Safe Claiming App from widget',
     category: OVERVIEW_CATEGORY,
   },
+  MISSING_SIGNATURES: {
```
katspaugh, line 71: Not the best name, you need to name the action, not the button.

src/services/analytics/events/overview.ts
```diff
@@ -68,4 +68,12 @@ export const OVERVIEW_EVENTS = {
...
+    action: 'Go to queue from missing signatures',
+    category: OVERVIEW_CATEGORY,
+  },
+  QUEUED_TRANSACTIONS: {
```
katspaugh, line 75: Same here.

src/components/sidebar/SafeListItem/index.tsx
```diff
@@ -53,54 +68,84 @@ const SafeListItem = ({
...
+                <ButtonBase className={classnames(css.pendingButton, css.queued)} onClick={closeDrawer}>
+                  <SvgIcon component={CheckIcon} inheritViewBox fontSize="small" />
+                  <Typography variant="body2">{queued}</Typography>
+                </ButtonBase>
+              </Tooltip>
+            </Track>
+          </NextLink>
+        )}
+      </Box>
+    </Box>
```
katspaugh, line 148: Please extract this into a component.

src/components/sidebar/SafeList/index.tsx
```diff
@@ -59,13 +64,55 @@ export const _shouldExpandSafeList = ({
 const MAX_EXPANDED_SAFES = 3
 const NO_SAFE_MESSAGE = 'Create a new safe or add'
 
+export const addActionsToTxs = (
```
katspaugh, line 67: I don't understand from the name what the function does and why it manipulates some transactions in the Safe List.
The body of the function doesn't help either. Some comments would be helpful, or better, a refactoring to simplify it.

katspaugh (changes requested):

katspaugh:
It's really hard to review this PR in the current state.
Please refactor it in the follow manner, then we'll do another round:

* A single new custom hook that
  * Fetches the queues for each added safe
  * Returns a `Record<safeAddress, { totalQueued: number, currentWalletQueued: number }>`
* A new component `<PendingActions totalQueued={...} currentWalletQueued={...} />` with two buttons that link to the queue

src/components/sidebar/SafeList/index.tsx
```diff
@@ -59,13 +64,55 @@ export const _shouldExpandSafeList = ({
...
+  chainId: string,
+  address: string,
+  page: TransactionListPage,
+  isSafeOwned: boolean,
+  txsActionsByChain: Record<string, Record<string, SafeActions>>,
+  walletAddress?: string,
```
iamacook, line 73: Nit: can you pass these arguments as an object. It's hard to follow them when called.

src/components/sidebar/SafeList/index.tsx
```diff
@@ -59,13 +64,55 @@ export const _shouldExpandSafeList = ({
...
+) => {
+  if (page.results.length === 0) return txsActionsByChain
```
iamacook, line 75: I'd suggest not mutating the object but returning a modified clone of it instead.

src/components/sidebar/SafeList/index.tsx
```diff
@@ -59,13 +64,55 @@ export const _shouldExpandSafeList = ({
...
+    }, txsActionsByChain)
+  }
+
+  txsActionsByChain[chainId] = {
+    ...(txsActionsByChain[chainId] ?? {}),
+    [address]: {
+      ...txsActionsByChain[chainId]?.[address],
+      queued: `${page.results.filter(isTransactionListItem).length}${page.next ? '+' : ''}`,
+    },
+  }
```
iamacook, line 96: I'd suggest assigning the properties directly here as well. [Spreading a large object is not as performant](https://www.youtube.com/watch?v=tcZbY-Q0TIE).

src/components/sidebar/SafeListItem/index.tsx
```diff
@@ -44,6 +55,10 @@ const SafeListItem = ({
   const isCurrentSafe = chainId === currChainId && sameAddress(safeAddress, address)
   const name = allAddressBooks[chainId]?.[address]
   const shortName = chain?.shortName || ''
+  const { queued = 0, signing = 0 } = queuedTxs || {}
+  const wallet = useWallet()
+  const url = `${AppRoutes.transactions.queue}?safe=${shortName}:${address}`
```
iamacook, line 60: Please create this in a `UrlObject` format instead.

src/components/sidebar/SafeListItem/index.tsx
iamacook, line 148: Accessibility-wise, these should also be within the `ListItem`.

src/services/analytics/events/overview.ts
```diff
@@ -68,4 +68,12 @@ export const OVERVIEW_EVENTS = {
     action: 'Open Safe Claiming App from widget',
     category: OVERVIEW_CATEGORY,
   },
+  MISSING_SIGNATURES: {
+    action: 'Go to queue from missing signatures',
```
iamacook, line 72:
```suggestion
    action: 'Go to queue from missing signatures in sidebar',
```

src/services/analytics/events/overview.ts
```diff
@@ -68,4 +68,12 @@ export const OVERVIEW_EVENTS = {
...
+    category: OVERVIEW_CATEGORY,
+  },
+  QUEUED_TRANSACTIONS: {
+    action: 'Go to queue from number of queued transactions',
```
iamacook, line 76:
```suggestion
    action: 'Go to queue from number of queued transactions in sidebar',
```

src/components/sidebar/SafeListItem/styles.module.css
```diff
@@ -2,19 +2,61 @@
...
+  gap: 4px;
+  padding-left: 0px;
+  margin-right:  var(--space-2);
+}
+
+.pendingButton {
+  width: 52px;
+  height: 22px;
+  border-top-right-radius: 6px;
+  border-bottom-right-radius: 6px;
```
iamacook, line 36: We should use the theme's `shape.borderRadius` here instead. You will have to apply it inside of an `sx` prop.

src/services/analytics/events/overview.ts
katspaugh: Action != description, it should be a one-two word name, like an id, easily filterable in GA and not subject to edits in future.

public/images/common/check.svg
DiogoSoaress: We are pushing the use of our Icon Library icons. I've already given the same response in other PR so I suggest if we are not completely aligned in this regard that we briefly talk about it after tomorrow's standup.

The next iteration would be to replace other uses of MUI icons as in the case of `TxSummary`

katspaugh: It's not what I asked. I got the memo that we're pushing our own icon library, that's fine. I'm asking why the two icons that are supposed to be the same are using a different svg.

katspaugh: In other words, I suggest we replace icons in a more systematic way. E.g. use the existing checkmark in this PR and update all of them in a separate PR. Otherwise we'll have a mix of both sprinkled around the app.

src/components/sidebar/SafeList/index.tsx
DiogoSoaress: The function calculates 2 values:
- the transactions that can be signed
- the size of the transaction queue

having being passed a `TransactionListPage` to it and the object that we are building `txsActionsByChain`.

Then it returns an object
```
{
  [chainId]: {
    [safeAddress]: {
      signing?: <number | string>
      queued?: <number | string>
    }
  }
}
```

to later be passed to each SafeListItem.

I am about to rename the function to `addActionsToSafeTxs` but WDYT? Any suggestion?

public/images/common/check.svg
DiogoSoaress: Alright. I agree with this. Will use the MUI icon in this PR.

src/services/analytics/events/overview.ts
DiogoSoaress: what about 
```suggestion
  OPEN_MISSING_SIGNATURES: {
```
?

src/components/sidebar/SafeListItem/styles.module.css
```diff
@@ -2,19 +2,34 @@
...
+
+.pendingButtons :global .MuiListItemSecondaryAction-root {
+  right: 70px;
+}
+
+@media (max-width: 400px) {
+  .pendingButtons :global .MuiListItemSecondaryAction-root {
+    right: 40px;
+  }
+}
```
DiogoSoaress, line 35: @usame-algan do you foresee incompatibilities with the bug fix you are working on?

src/components/sidebar/SafeListItem/index.tsx
```diff
@@ -54,9 +59,8 @@ const SafeListItem = ({
 
   return (
     <ListItem
-      className={css.container}
+      className={classnames(css.container, { [css.pendingButtons]: queued || signing })}
```
katspaugh, line 62: `pendingButtons` is a confusing name for a container class. It's not the buttons themselves, it's the container. `withPendingActions` would be clearer.

src/hooks/usePendingActions.ts
```diff
@@ -0,0 +1,75 @@
...
+import { selectAllAddedSafes } from '@/store/addedSafesSlice'
+import { sameAddress } from '@/utils/addresses'
+import type { PendingActionsByChain } from '@/utils/queuedTxsActions'
+import { addPendingActionsByChain } from '@/utils/queuedTxsActions'
+import { getTransactionQueue } from '@safe-global/safe-gateway-typescript-sdk'
+import { useEffect, useRef, useState } from 'react'
+
+const usePendingActions = () => {
+  const addedSafes = useAppSelector(selectAllAddedSafes)
+  const ownedSafes = useOwnedSafes()
```
katspaugh, line 17: This hook makes a request. Which results in two requests, one from the safe list and one from here.

I would suggest not to use it, and infer owned safes from the list of owners of each added safe. Or pass it from the safe list.
Screenshot 2023-01-25 at 21 07 19

src/hooks/usePendingActions.ts
```diff
@@ -0,0 +1,75 @@
...
+        }
+      }
+
+      isLoading.current = false
+
+      setSafeTxsActions(pendingActionsByChain)
+    }
+
+    getPendingActionsByChain()
+  }, [addedSafes, currentChainId, currentSafeAddress, ownedSafes, page, wallet?.address, safeTxsActions])
```
katspaugh, line 65: This is _really_ hard to read. I can't comment specifically how to improve it w/o rewriting it myself. I suggest we pair on this.

src/utils/queuedTxsActions.ts
```diff
@@ -0,0 +1,43 @@
...
+export type SafeTxsActions = {
+  queued?: string | number
+  signing?: string | number
+}
+
+export type PendingActionsByChain = {
+  [chainId: string]: { [safeAddress: string]: SafeTxsActions }
+}
+
+export const addPendingActionsByChain = (
```
katspaugh, line 13: I already commented on the naming of this function.

It's just a getter, it shouldn't be called "add".

It's not a util either, it's used only in that hook and should be moved there.

src/utils/queuedTxsActions.ts
```diff
@@ -0,0 +1,43 @@
+import { isSignableBy, isTransactionListItem } from '@/utils/transaction-guards'
+import type { TransactionListPage } from '@safe-global/safe-gateway-typescript-sdk'
+
+export type SafeTxsActions = {
+  queued?: string | number
+  signing?: string | number
```
katspaugh, line 6: Why are they a string or a number? Just stick with a string if you need to say `10+`.

Naming-wise, `totalQueued` or `queuedCount` would be better.

`signing` doesn't work at all. It's the number of txs you need to sign. They are not in the process of being signed by you. I would call it `totalToSign`.

katspaugh (changes requested):

src/hooks/usePendingActions.ts
katspaugh: One idea to simplify it would be to call this hook from the `PendingActions` component itself and request the queue for that one safe in each of them.
Instead of generating one giant nested object and then iterating over it again to render the buttons.

src/hooks/usePendingActions.ts
DiogoSoaress: Good catch! I will adjust it

src/components/sidebar/SafeListItem/styles.module.css
usame-algan: My idea for the bug fix was to just increase the padding since the value is already getting rounded. I think there should be no problems with the changes in this PR.

src/components/sidebar/PendingActions/styles.module.css
```diff
@@ -0,0 +1,38 @@
+.pendingButtons {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  gap: 4px;
+  padding-left: 0px;
+  margin-right:  var(--space-2);
```
iamacook, line 7:
```suggestion
  margin-right: var(--space-2);
```

src/components/sidebar/SafeList/index.tsx
```diff
@@ -118,8 +119,8 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
         configs.map((chain) => {
           const ownedSafesOnChain = ownedSafes[chain.chainId] ?? []
           const addedSafesOnChain = addedSafes[chain.chainId] ?? {}
-          const isCurrentChain = chain.chainId === chainId
           const addedSafeEntriesOnChain = Object.entries(addedSafesOnChain)
+          const isCurrentChain = chain.chainId === chainId
```
iamacook, line 123: There's no need to move this.

src/components/sidebar/SafeListItem/index.tsx
```diff
@@ -74,32 +78,42 @@ const SafeListItem = ({
...
+        </Link>
+      </Box>
+
+      <PendingActionButtons
+        safeAddress={address}
+        closeDrawer={closeDrawer}
+        shortName={shortName}
+        totalQueued={totalQueued}
+        totalToSign={totalToSign}
+      />
```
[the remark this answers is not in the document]
DiogoSoaress: It is already inside the `ListItem` ser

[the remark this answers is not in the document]
iamacook: 🙈 Too many `<ListItem*` components!

iamacook (approved):

katspaugh (approved):
👍

liliya-soroka:
@johannesmoormann  Do we count fully signed but not executed txs ? 
image

johannesmoormann:
@liliya-soroka for general pending actions, we count the whole transaction queue (up to pagination limit) and for wallet actions only transactions that need to be signed by the connected wallet.

liliiaorlenko:
pls increase the size of the whole sidebar together with a content on 20px, now elements feels cramped

liliiaorlenko:
also would we update safe list here https://app.safe.global/welcome ?

liliya-soroka:
@liliiaorlenko , sometimes the generated safe name is very long, as a result the safe blocks are "dancing" .Is it ok, or it's better to cover long names somehow( truncate) instead of changing the safe block size?
image

liliiaorlenko:
> @liliiaorlenko , sometimes the generated safe name is very long, as a result the safe blocks are "dancing" .Is it ok, or it's better to cover long names somehow( truncate) instead of changing the safe block size? image

I think better to make a drawer window wider for 20px at least. I think this is less info costly then also cutting names.

DiogoSoaress:
> @liliiaorlenko , sometimes the generated safe name is very long, as a result the safe blocks are "dancing" .Is it ok, or it's better to cover long names somehow( truncate) instead of changing the safe block size?

I am in favor of truncating the Safe name and perhaps display the whole name only on hover. I incremented the Safe list width 20px but it is just pushing the issue a bit. It does not solve it

liliiaorlenko:
Im ok to truncate it at some point but pls follow my recommendation of increasing the width of the list also! for me its a matter of well organised UI. + this fixes + the sizes of selected safe and other safes is different

Screenshot 2023-01-30 at 15 32 09

src/hooks/useWasVisible.ts
```diff
@@ -0,0 +1,30 @@
+import { useEffect, useState } from 'react'
+import type { MutableRefObject } from 'react'
+
+const useOnceVisible = (element: MutableRefObject<HTMLElement | null>): boolean => {
```
katspaugh, line 4: Please name the file the same as the hook.

liliya-soroka:
- [ ] The state is not updated for the safes when the side menu is opened in FF

src/hooks/useWasVisible.ts
DiogoSoaress: renamed

src/hooks/useWasVisible.ts
```diff
@@ -0,0 +1,30 @@
...
+  useEffect(() => {
+    if (!element.current || onceVisible) {
+      return
+    }
+
+    const observer = new IntersectionObserver((entries: IntersectionObserverEntry[]) => {
+      // Firefox returns multiple entries, of which one will be true when intersecting
+      const isIntersecting = entries.some(({ isIntersecting }) => isIntersecting)
+      if (isIntersecting) {
+        setOnceVisible(true)
```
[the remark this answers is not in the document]
DiogoSoaress: I unobserved here after setting the state and on unmount

liliya-soroka:
Verified

PR #1557, opened 2023-01-18
fix: adjust message display style

### LINKED ISSUE #1393: [EIP-1271] Format the message nicely
## Bug description
Message text is displayed above confirmation block on message detail - EIP-712
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Goerli

## Steps to reproduce
 1. Create EIP-712 sign message 
 2. go to the messages list
 3. open message details 

## Screenshots
image

iamacook:
## What it solves

Resolves #1393

## How this PR fixes it

The message is now display inside a `textarea` according to [the design](1393).

## How to test it

Open the messages list and observe the new formatting of the messages, as well as in the signing modal.

## Screenshots

image

image

image

image

### DIFF
--- a/src/components/safe-messages/Msg/styles.module.css
+++ b/src/components/safe-messages/Msg/styles.module.css
@@ -1,21 +1,6 @@
-.toggle {
-  font-weight: 700;
-  text-decoration: underline;
-  text-align: left;
-}
-
-.truncated {
-  display: -webkit-box;
-  -webkit-line-clamp: 3;
-  -webkit-box-orient: vertical;
-  overflow: hidden;
-  max-height: 9ex;
-}
-
-.overflow {
-  overflow: auto;
-}
-
-.pre {
-  margin: 0;
+.msg :global .MuiInputBase-input {
+  white-space: pre;
+  font-family: monospace;
+  font-size: 0.85rem;
+  -webkit-text-fill-color: var(--color-text-primary);
 }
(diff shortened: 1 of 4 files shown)


src/components/safe-messages/Msg/styles.module.css
```diff
@@ -1,21 +1,6 @@
...
-.overflow {
-  overflow: auto;
-}
-
-.pre {
-  margin: 0;
+.msg :global .MuiInputBase-input {
+  white-space: pre;
+  font-family: monospace;
+  font-size: 0.85rem;
```
usame-algan, line 4: This ends up being 13.6px for me. I would suggest we use an explicit pixel value like 14px.

usame-algan (approved):

src/components/safe-messages/Msg/styles.module.css
iamacook: Whilst I agree, this copies the sizing from the Safe Apps display. Would you suggest we change it in both places? I'd say `12/13px` instead of `14px` in which case.

usame-algan: If its like this in Safe Apps I would leave it like this for now.

liliya-soroka:
Verified. looks much better

iamacook:
I'll merge this as it's confirmed suitable by @liliya-soroka.

PR #1559, opened 2023-01-18
Add increment for failing estimations in Gnosis Chain

dasanra:
## What it solves
Resolves gas limit is not high enough for transactions in Gnosis Chain when the safeTxGas is set.
Due to a problem on Nethermind it's returning a lower gas estimation for those transactions having a `safeTxGas` set. 
`safeTxGas`is used as a mechanism so developers can suggest a gasLimit of the inner transaction. So whenever an app suggests a gasLimit it's set as safeTxGas.

## How this PR fixes it
Add an increment of 30% to the estimated `gasLimit`.
The fix is not ideal, but there is no other way to workaround this until a proper fix is added to Nethermind. Fix will be applied only specifically for Gnosis Chain and when the `safeTxGas` is set to a different value than 0.

## How to test it
Use `CowSwap` with latest version of WalletConnect Safe app and try to execute an `approve`
In this PR it should have enough gas limit, not the same case as production web app.

### DIFF
--- a/src/hooks/useGasLimit.ts
+++ b/src/hooks/useGasLimit.ts
@@ -5,12 +5,15 @@ import { encodeSignatures } from '@/services/tx/encodeSignatures'
 import type { SafeTransaction } from '@safe-global/safe-core-sdk-types'
 import { OperationType } from '@safe-global/safe-core-sdk-types'
 import useAsync from '@/hooks/useAsync'
+import useChainId from '@/hooks/useChainId'
 import { useWeb3ReadOnly } from '@/hooks/wallets/web3'
 import useSafeAddress from './useSafeAddress'
 import useWallet from './wallets/useWallet'
 import { useSafeSDK } from './coreSDK/safeCoreSDK'
 import useIsSafeOwner from './useIsSafeOwner'
 import { Errors, logError } from '@/services/exceptions'
+
+const GNOSIS_CHAIN_ID = '100'
 
 const getEncodedSafeTx = (safeSDK: Safe, safeTx: SafeTransaction, from?: string): string => {
   const EXEC_TX_METHOD = 'execTransaction'
@@ -44,6 +47,7 @@ const useGasLimit = (
   const wallet = useWallet()
   const walletAddress = wallet?.address
   const isOwner = useIsSafeOwner()
+  const currentChainId = useChainId()
 
   const encodedSafeTx = useMemo<string>(() => {
     if (!safeTx || !safeSDK || !walletAddress) {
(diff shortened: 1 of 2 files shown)


src/hooks/useGasLimit.ts
```diff
@@ -5,13 +5,16 @@ import { encodeSignatures } from '@/services/tx/encodeSignatures'
...
 import useAsync from '@/hooks/useAsync'
+import useChainId from '@/hooks/useChainId'
 import { useWeb3ReadOnly } from '@/hooks/wallets/web3'
 import useSafeAddress from './useSafeAddress'
 import useWallet from './wallets/useWallet'
 import { useSafeSDK } from './coreSDK/safeCoreSDK'
 import useIsSafeOwner from './useIsSafeOwner'
 import { Errors, logError } from '@/services/exceptions'
 
+const GNOSIS_CHAIN_ID = '100'
```
katspaugh, line 16: There's a file with all the chain ids in src/config/chains.ts, you can import that instead.

src/hooks/useGasLimit.ts
```diff
@@ -60,12 +64,23 @@ const useGasLimit = (
...
-    })
+    return web3ReadOnly
+      .estimateGas({
+        to: safeAddress,
+        from: walletAddress,
+        data: encodedSafeTx,
+        type: operationType,
+      })
+      .then((gasLimit) => {
+        // Due to a bug in Nethermind estimation we need to increment 30% the gasLimit
```
katspaugh, line 75:
```suggestion
        // Due to a bug in Nethermind estimation on Gnosis Chain, we need to increment the gasLimit by 30%
```

src/hooks/useGasLimit.ts
```diff
@@ -60,12 +64,23 @@ const useGasLimit = (
...
+        // when the safeTxGas is defined and not 0.
+        const safeTxGas = safeTx?.data?.safeTxGas
+        if (safeTxGas && safeTxGas !== 0 && currentChainId === GNOSIS_CHAIN_ID) {
+          return gasLimit.mul(130).div(100)
+        }
```
katspaugh, line 80: Since it's just for GC, I would move the chain id check to the beginning.

```suggestion
        if (currentChainId === chains.gno) {
          const gasLimitFactor = 1.3
          const safeTxGas = safeTx?.data?.safeTxGas
          if (safeTxGas && safeTxGas !== 0) {
            return gasLimit.mul(gasLimitFactor * 100).div(100)
          }
        }
```

Also, does `mul` accept only integers? Otherwise you could do `mul(1.3)`.

src/hooks/useGasLimit.ts
dasanra: Technically the gas estimation issue can happen on any chain that relies on Nethermind. If we would use Nethermind on mainnet the same bug could happen. So the bug is indeed on Nethermind estimation

src/hooks/useGasLimit.ts
dasanra: Yes, mul only accepts integers, I already tested that

src/hooks/useGasLimit.ts
katspaugh: Yeah, but the fix is only for GC, isn't it?

src/hooks/useGasLimit.ts
```diff
@@ -60,12 +63,23 @@ const useGasLimit = (
...
+        from: walletAddress,
+        data: encodedSafeTx,
+        type: operationType,
+      })
+      .then((gasLimit) => {
+        // Due to a bug in Nethermind estimation, we need to increment the gasLimit by 30%
+        // when the safeTxGas is defined and not 0. Currently Nethermind is used only for Gnosis Chain.
+        const safeTxGas = safeTx?.data?.safeTxGas
+        if (currentChainId === chains.gno && safeTxGas && safeTxGas !== 0) {
+          return gasLimit.mul(130).div(100)
```
katspaugh, line 78: Please extract 130 into a constant as per my previous comment.

src/hooks/useGasLimit.ts
```diff
@@ -60,12 +63,23 @@ const useGasLimit = (
...
+        }
+
+        return gasLimit
+      })
   }, [safeAddress, walletAddress, encodedSafeTx, web3ReadOnly, operationType])
```
katspaugh, line 83: Btw, strange that we're not getting lint errors here, since you're adding `safeTx` and `currentChainId` to the hook.

src/hooks/useGasLimit.ts
katspaugh: The difference in computational performance would be infinitesimally small. The difference in readability, IMHO, is significant. A separate if would show clearly that that entire block of code is only a workaround for GC.

katspaugh (approved):
LGTM!

JagoFigueroa:
I have been checking on a few apps via wc plus also a 1.1.1 safe and can confirm the 30% increment when safeTxGas is not 0. 

(Note that transactions on 1.1.1 safes will still fail without further increasing it manually but that it is to be expected)

Cheers!

PR #1562, opened 2023-01-19
fix: NFT selection in SendNFTForm

### LINKED ISSUE #1560: Unable to select certain NFTs when trying to send assets
Crosspost from https://github.com/5afe/safe-support/issues/119

### Issue Category

Safe loading issue, doesnt load on any interface

### What happened?

User trying to transfer NFTs but is unable to select the BoredApeYachtClub (without space), whichever user chooses it automatically selects the bored ape yacht club (with space) which is a malicious NFT, same behavior for the otherdeed NFT. Other NFTs are working fine.

Issue only occurs in the new interface (tested different browsers). Old interface works fine. 

https://user-images.githubusercontent.com/114049668/213335195-3e9870b8-efd2-43e3-bc82-9cb307d12166.mov

safe address: eth:0xd99e9aeC0a12a710420f27Cf13fcD248F8470b00

### Date and time this happened or you first noticed this issue

19-01-2023

### Network

Ethereum Mainnet

### Safe Address

0xd99e9aeC0a12a710420f27Cf13fcD248F8470b00

### Token Address

_No response_

### Token Type

None

### Transactions

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

schmanu:
## What it solves
NFTs were only identified by their ID in the SendNFTForm's Select component.

Resolves #1560 

## How this PR fixes it
NFTs are identified by their tokenAddress + id

## How to test it
In a Safe with multiple NFTs with the same id:
1. NFT modal
- open "Send NFT" modal
- Try to select the NFTs with same id and observe that always the correct one is selected

2. Send NFTs from assets
- Go to assets / nfts
- send NFTs with duplicate ids and observe that always the correct one is selected

## Analytics changes
None

### DIFF
--- a/src/components/tx/modals/NftTransferModal/SendNftForm.tsx
+++ b/src/components/tx/modals/NftTransferModal/SendNftForm.tsx
@@ -26,13 +26,11 @@ import InfiniteScroll from '@/components/common/InfiniteScroll'
 
 enum Field {
   recipient = 'recipient',
-  tokenAddress = 'tokenAddress',
   tokenId = 'tokenId',
 }
 
 type FormData = {
   [Field.recipient]: string
-  [Field.tokenAddress]: string
   [Field.tokenId]: string
 }
 
@@ -72,8 +70,7 @@ const SendNftForm = ({ params, onSubmit }: SendNftFormProps) => {
   const formMethods = useForm<FormData>({
     defaultValues: {
       [Field.recipient]: params?.recipient || '',
-      [Field.tokenAddress]: params?.token?.address || '',
-      [Field.tokenId]: params?.token?.id || '',
+      [Field.tokenId]: params?.token ? `${params.token.address};${params.token.id}` : '',
     },
   })
   const {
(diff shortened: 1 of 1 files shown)


src/components/tx/modals/NftTransferModal/SendNftForm.tsx
```diff
@@ -133,7 +131,7 @@ const SendNftForm = ({ params, onSubmit }: SendNftFormProps) => {
                   error={!!errors.tokenId}
                 >
                   {combinedNfts.map((item) => (
-                    <MenuItem key={item.address + item.id} value={item.id}>
+                    <MenuItem key={item.address + item.id} value={`${item.address};${item.id}`}>
```
iamacook, line 134: Nit: we should write a function to create the value to ensure it's the same in the `defaultValues` and here. Might be good to also have a function for getting the selected token as well for consistencies sake.

iamacook (approved):

katspaugh (commented):
Not sure I'm following the logic. Why was it necessary to remove the token address?

Just fixing `const token = combinedNfts.find((item) => item.id === data.tokenId)` would be enough wouldn't it?

schmanu:
> Not sure I'm following the logic. Why was it necessary to remove the token address?
> 
> Just fixing `const token = combinedNfts.find((item) => item.id === data.tokenId)` would be enough wouldn't it?

No the token address was only set when opening the modal from an specific NFT. 
The Select component only updated the ID form field and did not touch the token address

katspaugh:
Yeah, but when you select an item from the Select, it will have both an address and an id, right?

So you could do
```
const token = combinedNfts.find((item) => item.id === data.tokenId && item.address === data.tokenAddress)
```

katspaugh:
Since the NFT redesign is in QA now, I suggest we close this PR as the modal won't have a select in the redesigned version.
It was a nice fix nonetheless.

PR #1569, opened 2023-01-20
Fix: propose immediately executed txs

### LINKED ISSUE #1563: Always propose Safe Apps transactions
## Bug description

Safe Apps transactions propose [an "origin" that includes metadata about the app](https://github.com/safe-global/web-core/blob/dev/src/utils/transactions.ts#L181-L195). This metadata is used to display details of the app in the transaction list.

A [recent optimisation](https://github.com/safe-global/web-core/pull/1298/files#diff-70addbd231a36637989f9e3732bed07149598277eee9498f69fb69dc12463ffcR161-R169) removed proposal of immediately executed transactions, thus also the proposal of Safe App metadata.

We should *always propose* Safe App transactions, including their metadata.

## Steps to reproduce
 1. Create and immediately execute a Safe App transaction.
 2. After being successfully indexed, observe no Safe App metadata in the transaction list for the transaction.

## Expected result

The transaction is shown with the [associated metadata of the Safe App](https://github.com/safe-global/web-core/blob/dev/src/hooks/useTransactionType.ts#L80).

## Obtained result

The transaction is shown as a standard transaction.

## Screenshots

image

## Links

- [Discussion](https://5afe.slack.com/archives/C03DAGWJCR5/p1674118750610329)

katspaugh:
## What it solves

Resolves #1563

## How this PR fixes it
It restores the previous logic which proposes before an immediate execution.

I've also added a condition not to show the Untrusted Tx warning when executing a tx.

## How to test it
* Create a tx in the Drain app
* Immediately execute
* Click on "View transaction" in the notification
* Observe that the tx doesn't have an Untrusted warning
* Once in the history, the tx will have the Drain app logo

## Screenshots
34 is before the fix, 35 – after.
Screenshot 2023-01-20 at 11 26 27

### DIFF
--- a/src/components/settings/ContractVersion/UpdateSafeDialog.tsx
+++ b/src/components/settings/ContractVersion/UpdateSafeDialog.tsx
@@ -41,7 +41,7 @@ const UpdateSafeDialog = () => {
   )
 }
 
-const ReviewUpdateSafeStep = ({ onSubmit }: { onSubmit: (txId?: string) => void }) => {
+const ReviewUpdateSafeStep = ({ onSubmit }: { onSubmit: () => void }) => {
   const { safe, safeLoaded } = useSafeInfo()
   const chain = useCurrentChain()
   const { createMultiSendCallOnlyTx } = useTxSender()
--- a/src/components/settings/SafeModules/RemoveModule/steps/ReviewRemoveModule.tsx
+++ b/src/components/settings/SafeModules/RemoveModule/steps/ReviewRemoveModule.tsx
@@ -9,13 +9,7 @@ import { useEffect } from 'react'
 import { Errors, logError } from '@/services/exceptions'
 import { trackEvent, SETTINGS_EVENTS } from '@/services/analytics'
 
-export const ReviewRemoveModule = ({
-  data,
-  onSubmit,
-}: {
-  data: RemoveModuleData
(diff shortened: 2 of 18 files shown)


yagopv (approved):

yagopv:
Tested branch in Goerli and Ethereum and it works with the /proposal call

iamacook (commented):
Looks good but the types need to be reverted as well. The changes can be found in the [previous PR](https://github.com/safe-global/web-core/pull/1298/files#diff-4bd368d586160b58af6d413101e33cbd08c1c587de04e711ec50038c03177d48).

iamacook (approved):

liliya-soroka:
the CowSwap info is still missing in the history ( true for both 1/N safe and for N/M safes 
 https://txid--webcore.review-web-core.5afe.dev/gor:0xC642bEc0e4f69815e89723C09caA8DAa524cAc9c/transactions/tx?id=multisig_0xC642bEc0e4f69815e89723C09caA8DAa524cAc9c_0xdb4f4303b5621226594f3865e2975c1d435a135b3cd6647b86daa326cc7a7d8a
 The tx created using old app has Cowswap icon

katspaugh:
As discussed on Slack, this was the only issue, otherwise it's all good.

PR #1570, opened 2023-01-20
fix: "Recipient" -> "Recipient address or ENS"

katspaugh:
## What it solves

This PR makes it clearer that you can enter either an address or an ENS name in the recipient field.

## Screenshots
Screenshot 2023-01-20 at 11 37 54

## How to test
All the places where we have address inputs – the recipient field in tx modals, address book entry modal, owner management etc – will have "or ENS" added in the label.

### DIFF
--- a/src/components/address-book/EntryDialog/index.tsx
+++ b/src/components/address-book/EntryDialog/index.tsx
@@ -59,7 +59,7 @@ const EntryDialog = ({
             <Box>
               <AddressInput
                 name="address"
-                label="Address"
+                label="Contact"
                 variant="outlined"
                 fullWidth
                 required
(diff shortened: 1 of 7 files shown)


iamacook (approved):
What do you think about changing this for all other instances of `AddressBookInput` as well?

katspaugh:
I think it would be good but it takes the label as a prop and it can be any text.

Although it will probably work in most cases since it's usually a noun like "Owner", "Recipient" etc.

iamacook (approved):
Nice touch moving it into the component!

liliya-soroka:
Verified
On both networks : with and without ENS support

PR #1571, opened 2023-01-20
fix: Remove last used wallet when disconnecting

usame-algan:
## What it solves

Removes the last used wallet when disconnecting.

## How to test it

1. Open the Safe
2. Connect with a wallet
3. Reload the page
4. Observe it reconnecting to the wallet
5. Disconnect the wallet
6. Reload the page
7. Observe it not reconnecting to the wallet

### DIFF
--- a/src/hooks/wallets/useOnboard.ts
+++ b/src/hooks/wallets/useOnboard.ts
@@ -19,7 +19,11 @@ export type ConnectedWallet = {
   provider: EIP1193Provider
 }
 
-export const lastWalletStorage = localItem<string>('lastWallet')
+const lastWalletStorage = localItem<string>('lastWallet')
+
+export const forgetLastWallet = () => {
+  lastWalletStorage.remove()
+}
 
 const { getStore, setStore, useStore } = new ExternalStore<OnboardAPI>()
 
--- a/src/hooks/wallets/useOnboard.ts
+++ b/src/hooks/wallets/useOnboard.ts
@@ -19,7 +19,7 @@ export type ConnectedWallet = {
   provider: EIP1193Provider
 }
 
-const lastWalletStorage = localItem<string>('lastWallet')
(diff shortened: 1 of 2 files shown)


src/hooks/wallets/useOnboard.ts
```diff
@@ -19,7 +19,7 @@ export type ConnectedWallet = {
   provider: EIP1193Provider
 }
 
-const lastWalletStorage = localItem<string>('lastWallet')
+export const lastWalletStorage = localItem<string>('lastWallet')
```
katspaugh, line 22: Could you make a function `forgetLastWallet` and export that instead of exporting the storage?

katspaugh (approved):
Thanks!

katspaugh:
I've tested it, it works correctly now, doesn't reconnect to MM if I disconnect explicitly via the popup button.

iamacook (commented):
If you are connected to two wallets and disconnect from one, it shouldn't remove the last used wallet entirely.

last wallet

We should check that onboard has no wallets connected before removing it, e.g.

```ts
const handleDisconnect = async () => {
  const wallets = await onboard.disconnectWallet({
    label: wallet.label,
  })

  if (wallets.length === 0) {
    forgetLastWallet()
  }
}
```

katspaugh:
It's a separate thing. If you want to remember all connected wallets, it would have to be an array.

Just checking `if (wallets.length === 0) {` will result in the bug that you won't be able to forget your connection at all when connected to more than one wallet.

iamacook:
> Just checking `if (wallets.length === 0) {` will result in the bug that you won't be able to forget your connection at all when connected to more than one wallet.

After disconnecting from all wallets it would remove the last used one. It depends on what you define as a "last used wallet" though. I'd argue that a switched-to wallet is last used.

> If you want to remember all connected wallets, it would have to be an array.

This would be a nicer approach. Definitely something to consider.

PR #1572, opened 2023-01-20
fix: allow non-owner spending limit benficiaries to transact

### LINKED ISSUE #1561: [Spending limits] Non-owners with a spending limit allowance cannot interact with the Safe
## Bug description
It's possible to add spending limits to any Ethereum address, even non-owners. However, the UI still shows that Safe as "read-only" for such users.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Add a spending limit to a non-owner
 2. Try to make a transaction (bummer!)

iamacook:
## What it solves

Resolves #1561

## How this PR fixes it

Non-owner spending limit beneficiaries can now create and execute transactions according to their limits.

## How to test it

1. Add a non-owner as a spending limit beneficiary to a Safe.
2. Switch to the spending limit beneficiary
3. Observe that it is possible to create and execute transactions spending limit transactions with the associated token.

## Screenshots

spending limit beneficiary


katspaugh (approved):
A very solid approach! Scoped to the Send token modal.

francovenica:
Looks good to me

Checked that you can only access the send funds functionality. 
Only the token you have set in the spending limits are listed in the token list selection
Only the option to use spending limit (the radiobutton) is selected in the form
I wasn't able to do any other action with the former owner, which is the expected behavior.

Notes:
Some nish functionalities could be considered for this, like the "Send" button that shows in the assets table when you hover the  assets row, or the "Send Again" functionality that you can use from the tx details in the tx history, but I'd assume those require a lot of extra validations. So I'm ok leaving those out and just allow the user to use the "New transaction" button as the only way to send funds

iamacook:
> Notes: Some nish functionalities could be considered for this, like the "Send" button that shows in the assets table when you hover the assets row, or the "Send Again" functionality that you can use from the tx details in the tx history, but I'd assume those require a lot of extra validations. So I'm ok leaving those out and just allow the user to use the "New transaction" button as the only way to send funds

Thank you for the recommendations. I will file issues accordingly.

PR #1573, opened 2023-01-20
[WIP] eth_signTypedData_v3 for HW wallets and mobile pairing

katspaugh:
## What it solves

Testing eth_signTypedData_v3 for Ledger, Trezor and mobile pairing.

I've already successfully signed and executed transactions with Ledger and mobile pairing in a 1.3.0 Safe.

Need to test Trezor and earlier Safe contract versions.

### DIFF
--- a/src/services/tx/tx-sender/dispatch.ts
+++ b/src/services/tx/tx-sender/dispatch.ts
@@ -59,11 +59,11 @@ export const dispatchTxProposal = async ({
  */
 export const dispatchTxSigning = async (
   safeTx: SafeTransaction,
-  shouldEthSign: boolean,
+  isLegacySignMethod: boolean,
   txId?: string,
 ): Promise<SafeTransaction> => {
   const sdk = getAndValidateSafeSDK()
-  const signingMethod = shouldEthSign ? 'eth_sign' : 'eth_signTypedData'
+  const signingMethod = isLegacySignMethod ? ('eth_signTypedData_v3' as 'eth_signTypedData') : 'eth_signTypedData'
 
   let signedTx: SafeTransaction | undefined
   try {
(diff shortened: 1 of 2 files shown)


katspaugh:
@francovenica could you check signing with Trezor on this branch?

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -59,11 +59,11 @@ export const dispatchTxProposal = async ({
  */
 export const dispatchTxSigning = async (
   safeTx: SafeTransaction,
-  shouldEthSign: boolean,
+  isLegacySignMethod: boolean,
   txId?: string,
 ): Promise<SafeTransaction> => {
   const sdk = getAndValidateSafeSDK()
-  const signingMethod = shouldEthSign ? 'eth_sign' : 'eth_signTypedData'
+  const signingMethod = isLegacySignMethod ? ('eth_signTypedData_v3' as 'eth_signTypedData') : 'eth_signTypedData'
```
mikhailxyz, line 66: does it work for other hardware wallets?

mikhailxyz: the way it worked in `safe-react` is that it tried `eth_signTypedData_v4`, then `v3`, then `eth_sign`. If you try an unsupported method it usually throws `METHOD_NOT_SUPPORTED` error

katspaugh: Yeah, we don't have that cascading try-catch now, the if was catching the cases we knew about.
Richard says we shouldn't use v3 because it's not fully implementing the EIP-712.

iamacook, line 66: This is internally using `"eth_sign"`. [The `safe-core-sdk`  uses it if the `signingMethod` isn't `"eth_signTypedData"`](https://github.com/safe-global/safe-core-sdk/blob/main/packages/safe-core-sdk/src/Safe.ts#L459-L464).

iamacook: I was just reviewing the forthcoming v1.0.0 support and [noticed that this will be extended](https://github.com/safe-global/safe-core-sdk/blob/safe-v1.0.0-compatibility/packages/safe-core-sdk/src/Safe.ts#L442-L446).

PR #1574, opened 2023-01-20
fix: position executor in transaction stepper

iamacook:
## What it solves

Resolves incorrect placement of executor in transaction stepper:

image

## How this PR fixes it

The stepper "line" (`::before`) length is relevant to the display of executor/hint about execution.

## How to test it

- Queue a transaction and observe the hint regarding execution positioned correctly.
- Execute a transaction and observe the executor positions correctly.
- Observe no line that spreads beyond the final step.

## Screenshots

image

image

image

image

### DIFF
--- a/src/components/transactions/TxSigners/index.tsx
+++ b/src/components/transactions/TxSigners/index.tsx
@@ -129,18 +129,10 @@ export const TxSigners = ({ txDetails, txSummary }: TxSignersProps): ReactElemen
   const canExecute = wallet?.address ? isExecutable(txSummary, wallet.address, safe) : false
   const confirmationsNeeded = confirmationsRequired - confirmations.length
   const isConfirmed = confirmationsNeeded <= 0 || canExecute
-  const showExecutor = executor || !isConfirmed
 
   return (
     <>
-      <List
-        className={css.signers}
-        sx={{
-          '::before': {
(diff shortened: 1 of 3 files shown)


src/components/transactions/TxSigners/index.tsx
```diff
@@ -127,10 +129,18 @@ export const TxSigners = ({ txDetails, txSummary }: TxSignersProps): ReactElemen
...
+  const showExecutor = executor || !isConfirmed
 
   return (
     <>
-      <List className={css.signers}>
+      <List
+        className={css.signers}
+        sx={{
+          '::before': {
+            height: `calc(100% - ${showExecutor ? '80px' : '40px'})`,
```
katspaugh, line 140: Why so complicated? Can't the executor 🪓 just go directly after everything above?

iamacook: If we don't have conditional `height`, the stepper line bleeds into the executor positioning:

image

vs.

image

iamacook: 🤦‍♂️ I see what you mean now. Let me adjust it.

katspaugh (approved):

katspaugh:
Tested, it's fixed now 👍

PR #1577, opened 2023-01-23
fix: Reduce cummulative layout shift (CLS)

usame-algan:
## What it solves

Reduces CLS caused by elements in the header and sidebar

## How this PR fixes it

- Adds padding to the `NetworkSelector` in the header
- Adds a skeleton for the `ChainIndicator`
- Removes a redundant loading state for the overview widget on the dashboard

## Screenshots

Before:

[before.webm](https://user-images.githubusercontent.com/5880855/214008107-3df65843-3638-4395-9b8e-c01cad9bad88.webm)

After:

[after.webm](https://user-images.githubusercontent.com/5880855/214008217-78b685ec-965b-4e48-ba3c-b59d82d7ba6c.webm)

### DIFF
--- a/public/safe.webmanifest
+++ b/public/safe.webmanifest
@@ -15,7 +15,5 @@
     }
   ],
   "start_url": "/",
-  "theme_color": "#0FDA6D",
-  "background_color": "#121312",
   "display": "standalone"
 }
--- a/public/safe.webmanifest
+++ b/public/safe.webmanifest
@@ -15,5 +15,7 @@
     }
   ],
   "start_url": "/",
+  "theme_color": "#0FDA6D",
+  "background_color": "#121312",
   "display": "standalone"
 }
--- a/src/components/common/ChainIndicator/index.tsx
+++ b/src/components/common/ChainIndicator/index.tsx
@@ -38,7 +38,7 @@ const ChainIndicator = ({
 
   return chainConfig?.chainName ? (
     <span style={style} className={classnames(inline ? css.inlineIndicator : css.indicator, className)}>
-      {chainConfig.chainName || ' '}
+      {chainConfig.chainName}
     </span>
   ) : (
     <Skeleton width="100%" height="22px" variant="rectangular" sx={{ flexShrink: 0 }} />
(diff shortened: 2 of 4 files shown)


public/safe.webmanifest
```diff
@@ -15,5 +15,7 @@
     }
   ],
   "start_url": "/",
+  "theme_color": "#0FDA6D",
+  "background_color": "#121312",
```
katspaugh, line 19: Will that work for both light and dark modes?

src/components/common/ChainIndicator/index.tsx
```diff
@@ -35,10 +36,12 @@ const ChainIndicator = ({
 
   if (!chainConfig?.chainName && !renderWhiteSpaceIfNoChain) return null
 
-  return (
+  return chainConfig?.chainName ? (
     <span style={style} className={classnames(inline ? css.inlineIndicator : css.indicator, className)}>
-      {chainConfig?.chainName || ' '}
+      {chainConfig.chainName || ' '}
```
katspaugh, line 41: You can remove the ` || ' '` then.

src/components/common/NetworkSelector/index.tsx
```diff
@@ -76,7 +76,9 @@ const NetworkSelector = (): ReactElement => {
       })}
     </Select>
   ) : (
-    <Skeleton width={94} height={31} />
+    <Box px={2}>
+      <Skeleton width={94} height={31} />
+    </Box>
```
katspaugh, line 81: A Box just to add a padding? Maybe just use sx on the Skeleton?

katspaugh (approved):
Nice work, looks much smoother!

public/safe.webmanifest
usame-algan: Afaik this is still an experimental spec with not dark mode support. I added them because Lighthouse complained but looking more into it, we already have `<meta name="theme-color" ...>` for light and dark mode. I will revert this change.

src/components/common/NetworkSelector/index.tsx
usame-algan: Nice! Worked without the Box and instead margin on the Skeleton.

src/components/dashboard/Overview/Overview.tsx
```diff
@@ -47,7 +47,7 @@ const SkeletonOverview = (
           <Skeleton variant="circular" width="48px" height="48px" />
         </IdenticonContainer>
 
-        <Box mb={2}>
+        <Box mb={4} mt={4}>
```
iamacook, line 50:
```suggestion
        <Box my={4}>
```

iamacook (approved):

PR #1580, opened 2023-01-23
[EIP-1271] Decode eip712 messages

schmanu:
## What it solves
Instead of showing the raw JSON we show the messages data is a more human readable way.

## How this PR fixes it
Uses the types to show a better component for each message field.
Similar to how transaction data gets decoded.

## How to test it
Sign a EIP 712 message
Observe the new message display.

## Analytics changes
None

## Screenshots
tbd

### DIFF
--- a/src/components/safe-messages/DecodedMsg/index.tsx
+++ b/src/components/safe-messages/DecodedMsg/index.tsx
@@ -1,0 +1,70 @@
+import { generateDataRowValue, TxDataRow } from '@/components/transactions/TxDetails/Summary/TxDataRow'
+import { Value } from '@/components/transactions/TxDetails/TxData/DecodedData/ValueArray'
+import { isByte } from '@/utils/transaction-guards'
+import { normalizeTypedData } from '@/utils/web3'
+import { Box, Typography } from '@mui/material'
+import type { EIP712TypedData, SafeMessage } from '@safe-global/safe-gateway-typescript-sdk'
+import { isAddress } from 'ethers/lib/utils'
+import { isObject } from 'lodash'
+import type { ReactElement } from 'react'
+import Msg from '../Msg'
+
+export const DecodedMsg = ({ message }: { message: SafeMessage['message'] }): ReactElement => {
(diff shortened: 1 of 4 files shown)


src/components/safe-messages/MsgModal/index.tsx
```diff
@@ -151,7 +151,7 @@ const MsgModal = ({
             Message:
```
iamacook, line 151: It should also be possible to copy the entire message here as well, like we do in the transaction list.

iamacook (commented):
Really nice work! Maybe we should consider looking into this for the Safe Apps signing modal as they only display a `textarea` as well.

src/components/safe-messages/DecodedMsg/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+            ) : isNested ? (
+              <Box
+                sx={{
+                  border: ({ palette }) => `1px ${palette.border.light} solid`,
+                  whiteSpace: 'pre',
+                  fontFamily: 'monospace',
+                  fontSize: '0.85rem',
+                  overflow: 'auto',
+                  borderRadius: '6px',
+                  padding: 1,
```
iamacook, line 49: Can we move this to a CSS module? The `borderRadius` can stay here and use the theme's `shape.borderRadius`.

src/components/safe-messages/DecodedMsg/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const normalizedMsg = normalizeTypedData(message)
+
+  return (
+    <Box
+      sx={{
+        maxHeight: '300px',
+        overflow: 'auto',
+        padding: 1,
+        border: ({ palette }) => `1px solid ${palette.border.light}`,
+        borderRadius: '6px',
```
iamacook, line 84: Can we move this to a CSS module? The `borderRadius` can stay here and use the theme's `shape.borderRadius`.

iamacook (approved):

Issue #1369, opened 2022-12-14
Convert RHF components to be controlled


iamacook:
## What is the feature about

[React Hook Form works based on uncontrolled components. However, MUI inputs should be controlled.](https://react-hook-form.com/advanced-usage#ControlledmixedwithUncontrolledComponents).

Whilst we are currently `register`ing our inputs successfully, we are sometimes encountering [trivial errors](https://github.com/safe-global/web-core/blob/dev/src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx#L194) that can be solved [by following the correct implementation](https://github.com/react-hook-form/react-hook-form/issues/220#issuecomment-1289560753).

There are [alternative packages](https://github.com/dohomi/react-hook-form-mui) that can be used instead of creating our own, should we want to do so.

## The list of requirements

As of the date of this ticket, we only have 17 registrations but this will require extensive testing, perhaps even splitting this issue into smaller ones.

## Notes

We need to take into account scroll wheel interaction when using "number" `type` fields and label/value collision.

## Links

- [A previous PoC of reusable, controlled components.](https://github.com/safe-global/web-core/pull/953)

PR #1583, opened 2023-01-23
fix: asset selection validation

### LINKED ISSUE #1501: Update to react-hook-form > v7.41.1 breaks form validation
## Bug description

Updating the react-hook-form package to > v7.41.1 causes issues with form validation in create safe and other forms.

Possibly a misconfiguration on our side. Adjusting validation props in `AddressBookInput` and removing `register` in `NameInput` fixes the issue.

## Steps to reproduce
 1. Update react-hook-form to > v7.41.1
 2. Go to safe creation
 3. Press Next
 4. Observe that the Next button is disabled until the user focuses an input field

## Screenshots
Screenshot 2023-01-09 at 13 24 30

### LINKED ISSUE #1545: The field with valid data is marked with red frame
1. Go to the safe on prod
2. click New transaction ->send funds
3. select safe and select a token
4. put cursor to the amount field
Current result: Select a token field is marked with red frame
Expected result: The red frame should not be used for the field with valid data

Wallet: MM
Env: Prod
Browser: Chrome 
app version : 1.2.1

Screenshot 
image

iamacook:
## What it solves

Resolves #1545

## How this PR fixes it

The `tokenAddress` field in the `SendAssetsForm` has been converted to use RHF's `Controller` component.

Note: this partially resolves #1501 and demonstrates the importance of controlled MUI components with RHF (#1369).

## How to test it

Open the transaction creation modal and observe that the field is no longer erroneous after selecting a token. Selecting a token when a value is set should clear the value field as well.

## Screenshots

asset-validation

### DIFF
--- a/src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx
+++ b/src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx
@@ -95,6 +95,7 @@ const SendAssetsForm = ({ onSubmit, formData, disableSpendingLimit = false }: Se
     register,
     handleSubmit,
     setValue,
+    resetField,
     watch,
     formState: { errors },
   } = formMethods
@@ -157,7 +158,7 @@ const SendAssetsForm = ({ onSubmit, formData, disableSpendingLimit = false }: Se
               error={!!errors.tokenAddress}
               {...register(SendAssetsField.tokenAddress, {
                 required: true,
-                onChange: () => setValue(SendAssetsField.amount, ''),
+                onChange: () => resetField(SendAssetsField.amount),
               })}
             >
               {balances.items.map((item) => (
--- a/src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx
+++ b/src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx
@@ -1,5 +1,5 @@
 import type { ReactElement } from 'react'
-import { useForm, FormProvider } from 'react-hook-form'
+import { useForm, FormProvider, Controller } from 'react-hook-form'
 import {
   Button,
   FormControl,
@@ -98,6 +98,7 @@ const SendAssetsForm = ({ onSubmit, formData, disableSpendingLimit = false }: Se
     resetField,
     watch,
     formState: { errors },
+    control,
   } = formMethods
 
   const recipient = watch(SendAssetsField.recipient)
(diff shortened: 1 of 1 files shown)


src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx
```diff
@@ -146,27 +147,34 @@ const SendAssetsForm = ({ onSubmit, formData, disableSpendingLimit = false }: Se
...
           </FormControl>
 
-          <FormControl fullWidth>
-            <InputLabel id="asset-label" required>
-              Select an asset
-            </InputLabel>
-            <Select
-              labelId="asset-label"
-              label={errors.tokenAddress?.message || 'Select an asset'}
-              defaultValue={formData?.tokenAddress || ''}
```
iamacook, line 156: `defaultValue` is set in `useForm` and is not needed.

usame-algan:
Is there a way to reproduce the red outline issue on the assets input?

iamacook:
> Is there a way to reproduce the red outline issue on the assets input?

Select an asset on `dev`:

asset-validation

katspaugh (approved):
Looketh goode!

usame-algan:
> Select an asset on dev:

I couldn't reproduce it on dev so I updated RHF locally again, saw the red outline after the update. I switched to this Branch and the issue is fixed with a controlled input even with a newer RHF version 👍

usame-algan (approved):

katspaugh:
Found a small bug:
* Select a token with only a small amount owned
* Type a big number, it will show an error as expected
* Switch to another token
* The error remains

Screenshot 2023-01-24 at 15 29 51

src/components/tx/modals/TokenTransferModal/SendAssetsForm.tsx
```diff
@@ -157,7 +158,7 @@ const SendAssetsForm = ({ onSubmit, formData, disableSpendingLimit = false }: Se
               error={!!errors.tokenAddress}
               {...register(SendAssetsField.tokenAddress, {
                 required: true,
-                onChange: () => setValue(SendAssetsField.amount, ''),
+                onChange: () => resetField(SendAssetsField.amount),
```
katspaugh, line 161: This resets both the value and the error state, so it's back to clean.

katspaugh:
Tested, it works correctly now. ✅ 
Selecting a token resets the amount field completely, and no red border on the select anymore.

Sending a specific token from the asset list also works like before.

PR #1585, opened 2023-01-23
fix: DOM nesting error

iamacook:
## What it solves

Resolves console error regarding DOM nesting

image

## How this PR fixes it

The wrapper `Typography` uses specifies the `component` instead.

## How to test it

Open the transaction creation modal and observe no visual change regarding the "Sending from" block, as well as no console errors regarding DOM nesting.

## Screenshots

Before:
image

After:
image

### DIFF
--- a/src/components/tx/SendFromBlock/index.tsx
+++ b/src/components/tx/SendFromBlock/index.tsx
@@ -14,9 +14,9 @@ const SendFromBlock = (): ReactElement => {
         Sending from
       </Typography>
 
-      <Typography variant="body2">
+      <Box sx={({ typography }) => typography.body2}>
         <EthHashInfo address={address} shortAddress={false} hasExplorer showCopyButton />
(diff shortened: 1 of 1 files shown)


src/components/tx/SendFromBlock/index.tsx
```diff
@@ -14,9 +14,9 @@ const SendFromBlock = (): ReactElement => {
         Sending from
       </Typography>
 
-      <Typography variant="body2">
+      <Box sx={({ typography }) => typography.body2}>
```
katspaugh, line 17: Alternatively, it could be `<Typography variant="body2" component="div">`.

What about the Typography right above?

iamacook: I've specified the `component` instead.

The `Typography` above contains text.

katspaugh (approved):

PR #1587, opened 2023-01-23
fix: fallback to `chainId` of `wallet`

### LINKED ISSUE #1554: Use the chain the wallet is connected to as fallback
When the URL doesn't contain a chain id (e.g. when visiting the index page, or the welcome page), and a wallet is connected, use the wallet's network as the app chain.

E.g.
* User opens the `/welcome` page
* Connects their wallet, which is connected to Polygon
* The app should switch to Polygon

The current behavior falls back to a default chain instead (mainnet on prod and goerli on dev).

iamacook:
## What it solves

Resolves #1554

## How this PR fixes it

`useChainId` returns the `chainId` of the connected wallet if no `chainId` is present in the URL or cached from the last session.

## How to test it

1. Open a "non-`chainId`" URL in an Incognito window (e.g. `/welcome`).
2. Ensure the wallet is connected to a chain that is present in the environment chain configuration and connect the wallet.
3. Observe the chain in the top right matches that of the wallet.

---

1. Open a "non-`chainId`" URL in an Incognito window (e.g. `/welcome`).
2. Ensure the wallet is connected to a chain that is NOT present in the environment chain configuration and connect the wallet.
3. Observe the chain in the top right is the default fallback, Mainnet on prod. or Goerli on dev.

### DIFF
--- a/src/hooks/useChainId.ts
+++ b/src/hooks/useChainId.ts
@@ -6,6 +6,8 @@ import { useAppSelector } from '@/store'
 import { selectSession } from '@/store/sessionSlice'
 import { parsePrefixedAddress } from '@/utils/addresses'
 import { prefixedAddressRe } from '@/utils/url'
+import useWallet from './wallets/useWallet'
+import useChains from './useChains'
 
 const defaultChainId = IS_PRODUCTION ? chains.eth : chains.gor
 
@@ -44,7 +46,15 @@ export const useUrlChainId = (): string | undefined => {
 export const useChainId = (): string => {
   const session = useAppSelector(selectSession)
   const urlChainId = useUrlChainId()
-  return urlChainId || session.lastChainId || defaultChainId
+  const wallet = useWallet()
+  const chains = useChains()
+
+  const fallbackChainId =
+    wallet?.chainId && chains?.configs.some(({ chainId }) => chainId === wallet.chainId)
+      ? wallet.chainId
+      : defaultChainId
+
+  return urlChainId || session.lastChainId || fallbackChainId
 }
 
 export default useChainId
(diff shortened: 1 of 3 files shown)


src/hooks/useChainId.ts
```diff
@@ -44,7 +46,15 @@ export const useUrlChainId = (): string | undefined => {
...
-  return urlChainId || session.lastChainId || defaultChainId
+  const wallet = useWallet()
+  const chains = useChains()
+
+  const fallbackChainId =
+    wallet?.chainId && chains?.configs.some(({ chainId }) => chainId === wallet.chainId)
+      ? wallet.chainId
+      : defaultChainId
+
+  return urlChainId || session.lastChainId || fallbackChainId
```
katspaugh, line 57: I would define the wallet chainId as a separate thing and fall back in this order:
```
  return urlChainId || walletChainId || session.lastChainId || fallbackChainId
```

usame-algan: Is `fallbackChainId` and `walletChainId` the same thing?

iamacook: I've inverted the order and added tests to cover it.

iamacook: > Is fallbackChainId and walletChainId the same thing?

In my previous commit, `fallbackChainId` was either that of the `wallet` or from `defaultChainId`. I've changed it in accordance with the first suggestion. Wdyt?

katspaugh (approved):
Thanks! 👍

katspaugh:
Tested, it correctly sets the app network with the current wallet network if you're on a page w/o a Safe address or `?chain=` in the URL. ✅ 
It even changes the app network when you change the wallet network.

PR #1589, opened 2023-01-23
fix: iterate signing method on error

### LINKED ISSUE #1568: Errors when signing long messages
## Bug description

When signing large transaction batches created in the transaction builder I typically run into some sort of signing error. It used to be a different error (something like "message longer than 1024 characters"), but the new error is still caused by a similar action. See screenshot below for details.

The workaround is actually using the old safe app. Raising this issue now because I saw the old safe app was going to be deprecated.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask (and Frame wallet)
 - Chain: Optimism mainnet

## Steps to reproduce
 1. Go to the new safe app
 2. Open up the transaction builder
 3. Try to batch together 10 or more transactions (can require more/less to trigger, really depends how long the signature is)
 4. Sign the message with your wallet
 5. Watch Safe fail

## Expected result

The transaction should be signed successfully and enter the transaction queue.

## Obtained result

See screenshot.

## Screenshots

signal-2023-01-18-172514_002

iamacook:
## What it solves

Resolves #1568 and https://github.com/5afe/safe-support/issues/73

## How this PR fixes it

By default, hardware wallets/pairing use `eth_sign` but it is not possible to determine whether a hardware wallet is connected via MetaMask. If a signing attempt fails and the error throw was not a rejection, the subsequent one is attempted [as in `safe-react`](https://github.com/safe-global/safe-react/blob/dev/src/logic/safe/transactions/offchainSigner/index.ts).

Note: Only Safes >=v1.1.0 support `eth_sign`.

## How to test it

Ensure that signing transactions with hardware wallets and the Safe Mobile app still work as before. (I have tested this myself with a Ledger Nano X, Trezor Model T and the Safe Mobile app successfully.)

---

1. Connect to a Safe with a Trezor via MetaMask.
2. Open the Transaction Builder and create a "Custom data" transaction with the transaction data below.
3. Click "Create batch", sign the message and observe that it fails but attempts the next signing method.
4. The transaction should be added to the transaction queue.

<details>
  <summary>Transaction data</summary>
0x8d80ff0a0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000532e00920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000315892d8b693e3001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000004537fcd9095489ebe38180a382341b962b501d00000000000000000000000000000000000000000000000000315892d8b693e30000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000006d4ff66f29224001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000000339339653c106786180838dc0c0ab6ffc0cc4650000000000000000000000000000000000000000000000000006d4ff66f292240000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000e8b61c2e04d71759001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000000ab3cd94e756b856366194e440f60ea97eca2c32000000000000000000000000000000000000000000000000e8b61c2e04d717590000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000005d61fdade252625001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000000a1eeff6c67398c0878ac4ec012b9a243afa879e00000000000000000000000000000000000000000000000005d61fdade2526250000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000011986652ce7fb9f3001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000000cf603c60a2c058857210cbc539f19e48779b7fe00000000000000000000000000000000000000000000000011986652ce7fb9f30000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000006efb8b6df6568e001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed300000000000000000000000010dbc99c90234e4447f0366e8368d688f622475a000000000000000000000000000000000000000000000000006efb8b6df6568e0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000be3a30f6b78088001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000170706fe7d55bec8059cefa3f8b03e1f73e0332d00000000000000000000000000000000000000000000000000be3a30f6b780880000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000012ca10b5ae9456e001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed300000000000000000000000026ca32b540bdb8080a514f82705c9e9ab53a5207000000000000000000000000000000000000000000000000012ca10b5ae9456e0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000003080df5e05e63a001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000003542b2d6172bf3f39ab7f4cbc37ad0bb40dfe403000000000000000000000000000000000000000000000000003080df5e05e63a0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000dae1af37b66965001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000370a5275f69cb3c2e76cfe575cf7f846f906b34d00000000000000000000000000000000000000000000000000dae1af37b669650000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000eeb9638024691c68001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed300000000000000000000000037dc6358788da0f43fb80a0a5212d0eac43214b2000000000000000000000000000000000000000000000000eeb9638024691c680000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000010005f82afda175001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000410e08b3e64aac18ae18f2c5066d26e848ac8b9c000000000000000000000000000000000000000000000000010005f82afda1750000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000ccbbb6b3be30e3001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000435b7d470767cb121f37dd296b2ac7913fdf542700000000000000000000000000000000000000000000000000ccbbb6b3be30e30000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000916ab0f0b226914001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000004864e9ba3ab4270e338d1d52a49c383294c341340000000000000000000000000000000000000000000000000916ab0f0b2269140000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000a3c9243b988943001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000004cb2b7c5aa7af4176cc59584dea6246abfb1733a00000000000000000000000000000000000000000000000000a3c9243b9889430000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000001157e0c09808202001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000004feffa3cf9f3983e3d5a6a030cb41501500ef7b100000000000000000000000000000000000000000000000001157e0c098082020000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000263f392d1edbc631001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000529656620d914443405a55ec42ad0eaeaf0b4a2c000000000000000000000000000000000000000000000000263f392d1edbc6310000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000005085dcf615394920001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000005950dc5c3d7820937478f9312539b2e953f62e330000000000000000000000000000000000000000000000005085dcf6153949200000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000028375c79ef5fce001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed300000000000000000000000059d2e0a2be0db463c36038a5fd2535f58ab3b38f0000000000000000000000000000000000000000000000000028375c79ef5fce0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000002d5ed74ce938fc7001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000005f66d70b9ee2a793ac340beb647d0ad55765dbb600000000000000000000000000000000000000000000000002d5ed74ce938fc70000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000327798dfb24aecb001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed300000000000000000000000062365e0af7b6a188d063825afa3399ba0b5aa7cd0000000000000000000000000000000000000000000000000327798dfb24aecb0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000002c290026b810e44001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000640619af8b2820e0b1f7eac15d0b769a7822571500000000000000000000000000000000000000000000000002c290026b810e440000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000001b90f8e695f19ad9e001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000006ff60e1cf749ae639fbe124ddbc2178c4c988203000000000000000000000000000000000000000000000001b90f8e695f19ad9e0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000005a3e00f680fc6001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed300000000000000000000000076bcd1cd3244a4dc59211ebadb0462a62cbc08490000000000000000000000000000000000000000000000000005a3e00f680fc60000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000005693725e48a1ac8001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000007a62d6cf2517af31032a947ffdd8a385b45c80ce00000000000000000000000000000000000000000000000005693725e48a1ac80000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000324111b9129e144001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed300000000000000000000000086ed8d965561418ec4fc347721bac919487c75af0000000000000000000000000000000000000000000000000324111b9129e1440000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000015b57917b118d934001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000008bf0083ecea9bbe0b6ca47bdb3cd1c39f10bdf0200000000000000000000000000000000000000000000000015b57917b118d9340000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000028992f6702a5912001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000009e37af1c6d3a14d0da1259726ef6fe774058eba1000000000000000000000000000000000000000000000000028992f6702a59120000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000001af16d6539568b0001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed30000000000000000000000009bb0819842d82e6f4436bbd2e2f7d1910929a88200000000000000000000000000000000000000000000000001af16d6539568b00000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000602c2cd7fdaf59b9001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000a110d783a5b6b0b2a2c867094364e8f9752d5b93000000000000000000000000000000000000000000000000602c2cd7fdaf59b90000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000026448cb76fff801001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000a3e373ad4b28ce9b6a35253d7fddeb875df5fb77000000000000000000000000000000000000000000000000026448cb76fff8010000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000058f15752d778c001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000b6c9e6a2415ab8d26335498beac8eebed4bb225900000000000000000000000000000000000000000000000000058f15752d778c0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000001c25b65f6a38edc7001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000b7babe35ce543e2cf2f615cb1c792a2025feb5720000000000000000000000000000000000000000000000001c25b65f6a38edc70000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000001688df2a05e6e4001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000ba77e0b03b41a980bb4e7bb32ec7a2dfe1b66f8f000000000000000000000000000000000000000000000000001688df2a05e6e40000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000002961f1582df7cf0001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000c54570b8eb8138ace95132b944b7b6bb391976dc00000000000000000000000000000000000000000000000002961f1582df7cf00000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000001d72bbc8acdcea6f001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000cf9832e039c1920a17ee4ab34a23d3593d11a7ca0000000000000000000000000000000000000000000000001d72bbc8acdcea6f0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000007b5e6c9e0e5066001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000d58dcd4594087c2758717a59f44a5ac09db65105000000000000000000000000000000000000000000000000007b5e6c9e0e50660000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000002ca8a08eac930af001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000d77c6bb30cee69c435948709838eb99d60b33bf200000000000000000000000000000000000000000000000002ca8a08eac930af0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000632ade03e94138001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000d82a92991df76f59fd0c3c1de786f03d8a01e7fc00000000000000000000000000000000000000000000000000632ade03e941380000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000229463188901706b001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000d98b0f8a15eab299d0acca03f24ed0abb24dd5ec000000000000000000000000000000000000000000000000229463188901706b0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000fd40cc0397912a001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000dde346eaecec99ff62b8454fbd721593d40e2fbc00000000000000000000000000000000000000000000000000fd40cc0397912a0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000f5871238d72b6b39001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000dfab977372a039e78839687b8c359465f0f17532000000000000000000000000000000000000000000000000f5871238d72b6b390000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000e6b2e633444e125001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000db9918673d782a8116f583937f2f4a32352f11a20000000000000000000000000000000000000000000000000e6b2e633444e1250000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000670aa0d028b5df9001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000f5f61157f3c13e4bbc5c42e75ae80946b72de83f0000000000000000000000000000000000000000000000000670aa0d028b5df90000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000c727083679ccf85001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000f68d2bfcecd7895bba05a7451dd09a17490264540000000000000000000000000000000000000000000000000c727083679ccf850000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000004380ce1bf8fb2460001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000fcf325ccc3cd124efe1aa8cff362e3f74fded64a0000000000000000000000000000000000000000000000004380ce1bf8fb24600000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000002fb9bc43a75e2f001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000fe1a00487dd9eb84a7363a1c827a6f045fb121e4000000000000000000000000000000000000000000000000002fb9bc43a75e2f0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000073da0a4f58bf2c54001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000fee6be6b5cc8cb4ee8189850a69973e774e7614e00000000000000000000000000000000000000000000000073da0a4f58bf2c540000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000014717839d75786a001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000ac9f5a0a1c0f6862f2485b2f1390d31bfc90f4ab000000000000000000000000000000000000000000000000014717839d75786a0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000008ef6fab29eca51e001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000af0fdd39e5d92499b0ed9f68693da99c0ec1e92e00000000000000000000000000000000000000000000000008ef6fab29eca51e0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000003d8b1c44963625b001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000afa306dbb13beb550511ffbd304d97461e71603c00000000000000000000000000000000000000000000000003d8b1c44963625b0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000002bce8b288c75060001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000b436141073efe6c21ac6be9a5bb0d1d74f0ce87c00000000000000000000000000000000000000000000000002bce8b288c750600000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000003bab4e2998f5f6f001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000b726916fb98efe5f3d167c26dfe32c5c4f869d8400000000000000000000000000000000000000000000000003bab4e2998f5f6f0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be2000000000000000000000000000000000000000000000000077ce2b97a5c774e001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000b7e309b2fba55a46d4fcb18e000ef27e6ba04814000000000000000000000000000000000000000000000000077ce2b97a5c774e0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000004cedc5f469c1bf6001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000ba743394ca6c38886c066e45bb9980a09aec7a6500000000000000000000000000000000000000000000000004cedc5f469c1bf60000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000003190d8bd9243361001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000bf6e93fb31a9b2f6818004995d89eee73a4cbeac00000000000000000000000000000000000000000000000003190d8bd92433610000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000004ccff20257c6172001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000bfe8838de25ad36d106fd600f24471baf31e3ff300000000000000000000000000000000000000000000000004ccff20257c61720000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000010251c6c3635e824001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000c6e859b400daa289541c955dc3d4ef70a08aef6800000000000000000000000000000000000000000000000010251c6c3635e8240000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000104808c02df36fb001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000c874901c1aa292781110118ee4d3731c74e05b160000000000000000000000000000000000000000000000000104808c02df36fb0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000017f8d1680d070fac9001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000c9e7d61bbdef3f821da08bfdb19a2bf7e0ec60b70000000000000000000000000000000000000000000000017f8d1680d070fac90000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000182c22268a7cf7b001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000e29f21f462576bb4566f1920eada02a45693c7ea0000000000000000000000000000000000000000000000000182c22268a7cf7b0000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000013512609579f64a1001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000fc8333829a69d922b9d691f4d76185bdf4f4764200000000000000000000000000000000000000000000000013512609579f64a10000000000000000000000000000000000000000000000000000000001dfe20000920cf626a271321c151d027030d5d08af699456b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000001066a8eb3d90af0ad3f89839b974658577e75be20000000000000000000000000000000000000000000000000ba980de53a132e5001066a8eb3d90af0ad3f89839b974658577e75be200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064a0416ed3000000000000000000000000ff5144b6d2aaa53eade50c308c23d5c2151582d20000000000000000000000000000000000000000000000000ba980de53a132e50000000000000000000000000000000000000000000000000000000001dfe200000000000000000000000000000000000000
</details>


src/services/tx/tx-sender/dispatch.ts
```diff
@@ -57,20 +61,47 @@ export const dispatchTxProposal = async ({
...
  * Sign a transaction
  */
+
+type SigningMethods = Parameters<Safe['signTransaction']>[1]
+
+/**
+ * It's not possible to determine if a hardware wallet is connected via MetaMask. We therefore
+ * try to sign with EIP-712 if a wallet supports it, otherwise falling back to `eth_sign`.
+ */
+const getSupportedSigningMethods = (wallet: ConnectedWallet): SigningMethods[] => {
```
germartinez, line 71: Here the Safe version should be taken into account as well.
`eth_sign` is only supported by Safes `>= v1.1.0`
https://github.com/safe-global/safe-react/blob/c413702a0ad5887b356f6bc6f30aabcfcba1882d/src/logic/safe/transactions/offchainSigner/index.ts#L21-L39

germartinez (changes requested):

src/services/tx/tx-sender/dispatch.ts
germartinez: Well, or not needed and when the loop is executed if Safe < v1.1.0 and eth_sign is tried this error will be returned: https://github.com/safe-global/safe-core-sdk/blob/40061dedc748525612cf30f2839ea194c8c3a1d2/packages/safe-core-sdk/src/Safe.ts#L474

germartinez (approved):

src/services/tx/tx-sender/dispatch.ts
iamacook: I've adjusted [`getSupportedSigningMethods`](https://github.com/safe-global/web-core/blob/iterate-signing-methods/src/hooks/wallets/wallets.ts#L105-L122) to take the Safe version into account. Let me know what you think.

src/hooks/wallets/wallets.ts
```diff
@@ -97,3 +99,24 @@ export const isSmartContractWallet = async (wallet: ConnectedWallet) => {
...
   return isHardwareWallet(wallet) || isSafeMobileWallet(wallet)
 }
+
+type SigningMethods = Parameters<Safe['signTransaction']>[1]
+
+export const getSupportedSigningMethods = (
+  safeVersion: SafeInfo['version'],
+  wallet: ConnectedWallet,
+): SigningMethods[] => {
+  const EIP712_METHODS: SigningMethods[] = ['eth_signTypedData_v4', 'eth_signTypedData_v3', 'eth_signTypedData']
```
katspaugh, line 109: Richard said we shouldn't use `eth_signTypedData_v3` because it was a beta version. And `eth_signTypedData_v4` is just a synonym for `eth_signTypedData`, apparently.

So in the end only `eth_signTypedData` and `eth_sign` should remain.

src/hooks/wallets/wallets.ts
```diff
@@ -97,3 +99,24 @@ export const isSmartContractWallet = async (wallet: ConnectedWallet) => {
...
+
+  const safeSupportsEthSign = !!safeVersion && semverSatisfies(safeVersion, '>=1.1.0')
+  if (!safeSupportsEthSign) {
+    return EIP712_METHODS
+  }
+
+  if (shouldUseEthSignMethod(wallet)) {
```
katspaugh, line 116: If we do a try-catch anyway, why do we need this hardcoded check for HW wallets and mobile pairing?

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -57,20 +60,38 @@ export const dispatchTxProposal = async ({
...
+
+      // Try next signing method if the user did not reject the transaction
+      if (!isWalletRejection(error)) {
+        continue
+      }
+
+      txDispatch(TxEvent.SIGN_FAILED, { txId, error })
+
+      throw error
+    }
```
katspaugh, line 94: I would move it to some other module. Maybe `tx-sender/sdk.ts`?

src/hooks/wallets/wallets.ts
iamacook: Will this cause any cross-version compatability issues @germartinez? I see that [v adjustment was added for v1.0.0 support](https://github.com/safe-global/safe-core-sdk/blob/main/packages/safe-core-sdk/src/Safe.ts#L465-L470).

src/hooks/wallets/wallets.ts
iamacook: The MetaMask signing window will popup twice otherwise.

katspaugh: How's MM related to HW and mobile pairing?

src/services/tx/tx-sender/sdk.ts
```diff
@@ -29,3 +35,58 @@ export const getUncheckedSafeSDK = (provider: Web3Provider): Promise<Safe> => {
...
+  if (!safeVersion || !semverSatisfies(safeVersion, SAFE_VERSION_SUPPORTS_ETH_SIGN)) {
+    return ['eth_signTypedData_v4']
+  }
+
+  if (shouldUseEthSignMethod(wallet)) {
+    return ['eth_sign']
+  }
+
+  // Hardware wallet may be connected via MetaMask so we try `eth_sign` after EIP-712
+  return ['eth_signTypedData_v4', 'eth_sign']
```
katspaugh, line 56: An enum for these would be good.

src/hooks/wallets/wallets.ts
katspaugh: We briefly synced, to recap:
* Let's remove the check and keep only the try-catch to reduce the complexity here
* We remove v3 and v4 and keep only the un-postfixed `eth_signTypedData` (because that's the only true one that the EIP defines)

src/services/tx/tx-sender/sdk.ts
iamacook: As discussed, I'll keep this as a string now that I've reduced the returned methods as the signing method argument of `signTransaction` [is a string union](https://github.com/safe-global/safe-core-sdk/blob/main/packages/safe-core-sdk/src/Safe.ts#L442-L446).

src/services/tx/tx-sender/__tests__/sdk.test.ts
```diff
@@ -0,0 +1,21 @@
+import { _getSupportedSigningMethods } from '@/services/tx/tx-sender/sdk'
+
+describe('getSupportedSigningMethods', () => {
+  it('should return `eth_signTypedData` for older Safes', () => {
+    const result = _getSupportedSigningMethods('1.0.0')
+
+    expect(result).toEqual(['eth_signTypedData'])
+  })
+
+  it('should return `eth_signTypedData` for invalid Safes', () => {
```
mikhailxyz, line 10: hmm I'm curious why... does `null` really support `eth_signTypedData`?

mikhailxyz (approved):

src/services/tx/tx-sender/__tests__/sdk.test.ts
iamacook: The backend can return `SafeInfo['version']` as `null` when it's an unsupported contract. Can you think of a better way to handle this, or should I adjust the test description?

mikhailxyz: Ahh, then it makes sense. A better description (what you've just said about the backend and unsupported contract) would be nice 🙏

src/services/tx/tx-sender/sdk.ts
```diff
@@ -29,3 +33,40 @@ export const getUncheckedSafeSDK = (provider: Web3Provider): Promise<Safe> => {
...
+): Promise<SafeTransaction | undefined> => {
+  const sdk = getAndValidateSafeSDK()
+
+  let signedTx: SafeTransaction | undefined
+
+  for await (const signingMethod of _getSupportedSigningMethods(safeVersion)) {
+    try {
+      signedTx = await sdk.signTransaction(safeTx, signingMethod)
+
+      break
```
katspaugh, line 61: You could a return right here.

src/services/tx/tx-sender/sdk.ts
```diff
@@ -29,3 +33,40 @@ export const getUncheckedSafeSDK = (provider: Web3Provider): Promise<Safe> => {
 
   return sdk.connect({ ethAdapter })
 }
+
+type SigningMethods = Parameters<Safe['signTransaction']>[1]
+
+export const _getSupportedSigningMethods = (safeVersion: SafeInfo['version']): SigningMethods[] => {
+  const SAFE_VERSION_SUPPORTS_ETH_SIGN = '>=1.1.0'
+
+  if (!safeVersion || !semverSatisfies(safeVersion, SAFE_VERSION_SUPPORTS_ETH_SIGN)) {
```
katspaugh, line 42: What I don't get tho, why do earlier Safe versions support `eth_signTypedData` (AFAIU a newer EIP), but not `eth_sign`? Or is `eth_sign` newer?

iamacook: [`eth_sign` wasn't added to the contracts until v1.1.0](https://github.com/safe-global/safe-contracts/pull/127) as EIP-712 is/was the preferred signing method.

schmanu: I'm also a bit puzzled by that. How did you find that out?
I cannot really see the related change [here](https://github.com/safe-global/safe-contracts/releases/tag/v1.1.0).

src/services/tx/tx-sender/sdk.ts
```diff
@@ -29,3 +33,34 @@ export const getUncheckedSafeSDK = (provider: Web3Provider): Promise<Safe> => {
...
+    return ['eth_signTypedData']
+  }
+
+  return ['eth_signTypedData', 'eth_sign']
+}
+
+export const tryOffChainSigning = async (
+  safeTx: SafeTransaction,
+  safeVersion: SafeInfo['version'],
+): Promise<SafeTransaction | undefined> => {
```
katspaugh, line 52: I don't think it can return undefined, can it? It either throws or returns.

src/services/tx/tx-sender/sdk.ts
katspaugh: Two constants would be good then, to not repeat the string literal twice.

src/services/tx/tx-sender/sdk.ts
katspaugh: That's a nice tidbit but I still don't understand why eth_sign wasn't supported and eth_signTypedData was.

src/services/tx/tx-sender/sdk.ts
iamacook: I think so, though it's not intelligent enough to pick it up unless we explicitly `throw` again at the end of the function.

src/services/tx/tx-sender/sdk.ts
mikhailxyz: I think it was supported, it's more that nowadays most wallets prepend the EIP-191 prefix to `eth_sign` messages and signing with prefix wasn't supported.

mikhailxyz: Original EIP-191 was supposed to be implemented in a separate `personal_sign` method afaik: https://eips.ethereum.org/EIPS/eip-191

Some history (in metamask's interpretation): https://docs.metamask.io/guide/signing-data.html#a-brief-history

katspaugh: Thanks @mikhailxyz!

Btw, I was told MM are going to turn off eth_sign by default.

katspaugh (approved):

francovenica:
I asked Aaron the json file since the raw data was giving me trouble. 
I imported my trezor One into MM and signed the tx. I was able to create the tx and execute it with no issues.

LGTM

PR #1591, opened 2023-01-24
fix: Handle safe creation errors

### LINKED ISSUE #1558: Creating a safe on Arbitrum instantly times out
## Bug description

The create safe transaction on Arbitrum results in an error that is not handled. The flow defaults to a Timeout error instead. The transaction still gets executed though.

## Steps to reproduce
 1. Go to Safe
 2. Create a new safe on Arbitrum
 3. Observe an instant timeout error

## Screenshots
Screenshot 2023-01-18 at 15 13 23

usame-algan:
## What it solves

Resolves #1558 

Note: I couldn't reproduce ethers returning an error but the transaction still being executed/being in the mempool.

## How this PR fixes it

Uses the generic `ERROR` instead of `TIMEOUT` as a fall-through when creating a safe. This makes more sense since ethers returns a specific error code for timeouts but not for other possible errors like the one reported.

## How to test it

1. Create a new safe
2. Select 21000 gas limit in your wallet
3. Submit transaction
4. Observe a generic error in the UI instead of a timeout error
5. Create a new safe
6. Select extremely low gas and maxPrioFee
7. Reload the page
8. Wait for 6.5 minutes
9. Observe a timeout error

### DIFF
--- a/src/components/new-safe/create/logic/index.test.ts
+++ b/src/components/new-safe/create/logic/index.test.ts
@@ -64,12 +64,10 @@ describe('checkSafeCreationTx', () => {
     expect(result).toBe(SafeCreationStatus.REVERTED)
   })
 
-  it('returns TIMEOUT if transaction couldnt be found within the timeout limit', async () => {
+  it('returns TIMEOUT if transaction could not be found within the timeout limit', async () => {
     const mockEthersError = {
       ...new Error(),
-      receipt: {
-        status: 1,
-      },
+      code: ErrorCode.TIMEOUT,
     }
 
     waitForTxSpy.mockImplementationOnce(() => Promise.reject(mockEthersError))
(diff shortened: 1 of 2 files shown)


iamacook:
Looks good to me.

With the gas set to `21000`, the wallet shows a generic error:

image

Setting a low gas/maxPrioFee and refreshing also resulted in a timeout after 6.5 minutes: (I made sure that the experimental gas UI is turned on in MetaMask.)

image

iamacook (approved):

DiogoSoaress (approved):
The fix makes sure we display only a timeout error if that's what we get from ethers. A generic error otherwise is more correct 👍

francovenica:
It works fine for me as well, but the timeout I had to test it in another network, since in arbitrum what ever number you set in the gas price and fee the tx will fail right away or succeed, there is no in between

PR #1590, opened 2023-01-24
fix: `EthHashInfo` nesting

iamacook:
## What it solves

Resolves invalid DOM nested, on top of https://github.com/safe-global/web-core/pull/1585

image

## How this PR fixes it

All instances of `EthHashInfo` found inside `Typography` have been changed to use `div` elements as they shouldn't be nested inside `p` elements.

## How to test it

1. Open the account centre when connected to a wallet without an ENS and observe no console error.
2. Load a Safe and observe no console errors on the owner overview step.
3. Review a transaction to be created and observe console errors.

PR #1588, opened 2023-01-23
Feat: support 1.0.0 Safes

katspaugh:
## What it solves

Resolves #1188

## How this PR fixes it

- The SDK has been updated, as well as associated assertions.
- Update and error messages have been adjusted to match the newly supported version.
- Update code has been modified.

## How to test it

I've tested the following, confirmed working:

### Transfers
- [x] Sending native token
- [x] Sending ERC-20
- [x] Sending NFT
- [x] MultiSend
- [x] Spending limit

### Wallets
- [x] Smart contract wallets (on-chain signatures)
  - [x] v1.0.0 Safe owned by a Safe
  - [x] v1.0.0 Safe as owner of a Safe
- [x] Hardware wallets
  - [x] Ledger Nano X
  - [x] Trezor Model T
- [x] Safe Mobile app

### Settings
- [x] Upgrading v1.0.0 to v1.3.0
- [x] Owner management
  - [x] Add owner
  - [x] Remove owner
  - [x] Add owner and change threshold
  - [x] Remove owner and change threshold

- Intermediary upgrades, e.g. v1.1.0 to v1.3.0 should also be tested as the update code was modified.
- Post-upgrade functionality should be double-checked, especially that which relies on the fallback handler.

PR #1586, opened 2023-01-23
fix: reset error boundary after navigation

iamacook:
## What it solves

Resolves #1539

## How this PR fixes it

A standard link is used instead of an internally routed one.

## How to test it

Navigate to `/apps?safe=gor:0x4127839cdf4F73d9fC9a2C2861d8d1799e9DF40C&appUrl=appUrl%3Djavascript%3AstartHack%28%29)` and click "Go Home". Observe that the `/welcome` page is opened and error stack is reset.

PR #1581, opened 2023-01-23
feat: add smoke test for spending limit modal

iamacook:
## What it solves

E2E test for spending limit modal

## How this PR fixes it

An E2E test for the basic overview of spending limit transactions in the creation modal has been added, covering the opening, drafting and review of the desired transaction.

## How to test it

Running `/cypress/e2e/spending_limit.cy.js' locally should pass.

PR #1594, opened 2023-01-25
Release 1.4.0

katspaugh:
## Features
* Feat: support 1.0.0 Safes (#1588)

## Improvements
* Switch app network to wallet network on /welcome (#1587)
* "Recipient" -> "Recipient address or ENS" (#1570)
* Explorer link in multisend/spending limit tx notifications (#1542)
* feat(safe-apps): Use SignMessageLib from safe-core-sdk (#1419)

## Major fixes
* Fix: propose immediately executed txs (#1569)
* fix: iterate signing method on error (#1589)
* Add increment for failing estimations in Gnosis Chain (#1559)
* fix: GA tracking wrong eventLabels, missing event type, overtracking wallet connect event (#1547)

## Minor fixes
* fix: asset selection validation (#1583)
* fix: Handle safe creation errors (#1591)
* fix: reset error boundary after navigation (#1586)
* fix: `EthHashInfo` nesting (#1590)
* fix: `merge` hydration of persisted state (#1552)
* fix: transaction info styling (#1576)
* fix: DOM nesting error (#1585)
* fix: Reduce cummulative layout shift (CLS) (#1577)
* fix: ref forwarding (#1578)
* fix: align the ...-menu with EthHashInfo
* fix: position executor in transaction stepper (#1574)
* fix: prevent scrollbars on Safe App icons (#1565)
* fix: Remove last used wallet when disconnecting (#1571)
* refactor: replace Stepper for MUI's List (#1467)
* style: mobile Address Book layout (#1503)
* fix: wait until successful SDK call to set pending state (#1529)
* fix: show input loading state when "Add Safe" from the sidebar (#1528)
* render an img tag in SafeAppIcon iframe (#1516)

## Tests
* Extend create safe e2e test (#1543)
* add smoke test for spending limit modal (#1581)


francovenica:
I've swapped an owner in this safe about 15 mins ago. It took a manual reload to see the tx in the history tab (Safe 1/x, with signing and execution at the same time) and the old owner is still listed in the settings 
https://dev--webcore.review-web-core.5afe.dev/gor:0x41EfabA9C6264BA81f235B7Bc2e3eb2390782790/settings/setup
owner swap

katspaugh:
Looks like a backend hiccup. By the time it lands in the history it should already drop the untrusted flag (because it wasn't signed originally). Give it some time.

francovenica:
Ok, yes, it fixed it on its own. after a while

francovenica:
I don't see this ticket working: https://github.com/safe-global/web-core/pull/1542 (the link to the block explorer in the notification after executing a tx), but also I see that the ticket was created because of the removal of the proposal of tx that are created and executed at the same time, thing that we reverted. So... maybe is ok for this ticket not to be working?

francovenica:
Ticket 1547.
I'm getting some inconsistent calls respect on how it was working on that original ticket. Before reporting it as an issue I'll debug it with Usame directly

francovenica:
I'm not sure how to validate https://github.com/safe-global/web-core/pull/1559. I'll request help from Dani Sanchez when possible

katspaugh:
The block explorer link will now appear only for transactions that will never appear in the queue (e.g. spending limit txs). Regular immediate executions are now proposed like before.

Regarding the Gnosis Chain gas, you need to set a custom safeTxGas (from a Safe App), and compare the resulting gas with production. It should be +30% on this branch.

katspaugh:
Here's a new one: https://github.com/safe-global/web-core/pull/1599

PR #1595, opened 2023-01-25
refactor: helper for checking version features

### LINKED ISSUE #1593: Centralise Safe-specific feature flags
## What is the feature about

We are checking the version of Safes in multiple places to determine if they support certain features, e.g. `safeTxGas`, `eth_sign`, etc.

We should centralise these checks and rely on a helper function instead. We can use [the `safe-core-sdk` as inspiration](https://github.com/safe-global/safe-core-sdk/blob/main/packages/safe-core-sdk/src/utils/safeVersions.ts).

## The list of requirements

- Create feature `enum` and helper function.
- Search for `semver` project-wide and replace with helper function.

iamacook:
## What it solves

Resolves #1593

## How this PR fixes it

A new helper function `hasFeature` has been created to check the enabled features relevant to a Safe version. Instances where it can be used have been replaced.

## How to test it

- Ensure added guards can be seen
- Ensure v1.0.0 Safes upgrade as expected
- Ensure v1.0.0 Safes sign correctly

### DIFF
--- a/src/utils/safe-versions.ts
+++ b/src/utils/safe-versions.ts
@@ -1,0 +1,40 @@
+import semverSatisfies from 'semver/functions/satisfies'
+
+import type { SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
+
+export enum FEATURES {
+  SAFE_TX_GAS_OPTIONAL = 'SAFE_TX_GAS_OPTIONAL',
+  SAFE_TX_GUARDS = 'SAFE_TX_GUARDS',
+  SAFE_FALLBACK_HANDLER = 'SAFE_FALLBACK_HANDLER',
+  ETH_SIGN = 'ETH_SIGN',
+}
+
+const FEATURES_BY_VERSION: Record<FEATURES, string> = {
+  [FEATURES.ETH_SIGN]: '>=1.1.0',
+  [FEATURES.SAFE_FALLBACK_HANDLER]: '>=1.1.1',
+  [FEATURES.SAFE_TX_GAS_OPTIONAL]: '>=1.3.0',
+  [FEATURES.SAFE_TX_GUARDS]: '>=1.3.0',
+}
+
+// Note: backend returns `SafeInfo['version']` as `null` for unsupported contracts
+
+const isEnabledByVersion = (feature: FEATURES, version: SafeInfo['version']): boolean => {
+  if (!version || !(feature in FEATURES_BY_VERSION)) {
+    return false
+  }
+  return semverSatisfies(version, FEATURES_BY_VERSION[feature])
+}
+
+export const enabledFeatures = (version: SafeInfo['version']): FEATURES[] => {
+  if (!version) {
+    return []
+  }
+  return Object.values(FEATURES).filter((feature) => isEnabledByVersion(feature, version))
+}
+
+export const hasFeature = (name: FEATURES, version: SafeInfo['version']): boolean => {
+  if (!version) {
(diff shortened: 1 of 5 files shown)


src/utils/safe-versions.ts
```diff
@@ -0,0 +1,40 @@
...
+  SAFE_TX_GAS_OPTIONAL = 'SAFE_TX_GAS_OPTIONAL',
+  SAFE_TX_GUARDS = 'SAFE_TX_GUARDS',
+  SAFE_FALLBACK_HANDLER = 'SAFE_FALLBACK_HANDLER',
+  ETH_SIGN = 'ETH_SIGN',
+}
+
+const FEATURES_BY_VERSION: Record<VERSION_FEATURES, string> = {
+  [VERSION_FEATURES.ETH_SIGN]: '>=1.1.0',
+  [VERSION_FEATURES.SAFE_FALLBACK_HANDLER]: '>=1.1.1',
+  [VERSION_FEATURES.SAFE_TX_GAS_OPTIONAL]: '>=1.3.0',
```
katspaugh, line 15: This doesn't seem to be used anywhere except tests.
I think it's supposed to be checked in the AdvancedParams?

iamacook: We already have a check in place there [based on the form values](https://github.com/safe-global/web-core/blob/dev/src/components/tx/AdvancedParams/AdvancedParamsForm.tsx#L110).

I don't think we should remove this from the `enum` though as we may need it at some point.

katspaugh: I would remove it, you can leave it in the enum but comment out if you like, but we shouldn't keep code "for future".

katspaugh (commented):
Looks good now, just waiting for a Bescheid about whether we can import this code from Core SDK instead. We can expose new functions if necessary. cc @germartinez

iamacook:
> Looks good now, just waiting for a Bescheid about whether we can import this code from Core SDK instead. We can expose new functions if necessary. cc @germartinez

The function can be imported already but it's:

a) Returning all features enabled for v1.0.0 Safes at the moment (see https://github.com/safe-global/safe-core-sdk/issues/330).
b) Named the same as the other function we have locally for checking chain-specific features.
c) Would be better exported from `safe-core-sdk-utils`, otherwise we need import it from `'@safe-global/safe-core-sdk/dist/src/utils/safeVersions'`.

It's cleanest to import it from the SDK. I'd suggest we keep this PR open and make the required adjustments on the SDK side. What do you think?

katspaugh:
Sounds good, thanks for drilling down!

iamacook:
> a) Returning all features enabled for v1.0.0 Safes at the moment (see https://github.com/safe-global/safe-core-sdk/issues/330).
> b) Named the same as the other function we have locally for checking chain-specific features.
> c) Would be better exported from safe-core-sdk-utils, otherwise we need import it from '@safe-global/safe-core-sdk/dist/src/utils/safeVersions'.

I've created a PR addressing a) and b) [here](https://github.com/safe-global/safe-core-sdk/pull/331). Looking through the structure of the SDK, I don't think moving it to `safe-core-sdk-utils` makes the most sense. It would only cleanup imports which isn't something overly important. 

What do you think? cc @germartinez

germartinez:
It was a good catch!
Moving that logic to `safe-core-sdk-utils` looks good to me

Thank you @iamacook!

katspaugh (approved):
👍

PR #1597, opened 2023-01-26
Fix: default wallet when none are enabled

### LINKED ISSUE #1596: 8 wallets are displayed by default when no selected wallets in the config service
8 wallets are displayed by default when no selected wallets in the config service
1. add new network to the config service 
2. all wallets should be off for the network
3.  switch to the new network on web-core
4. click connect 
Current result: 

Image

katspaugh:
## What it solves

Resolves #1596

## How this PR fixes it

When all wallets are disabled, we pass an empty array to onboard, but it needs at least one wallet. So I made it return the "injected" wallet as a default.

## How to test it
Cello has currently 0 wallets enabled in the staging config.

### DIFF
--- a/src/hooks/wallets/wallets.ts
+++ b/src/hooks/wallets/wallets.ts
@@ -67,9 +67,13 @@ export const getSupportedWallets = (chain: ChainInfo): WalletInit[] => {
   if (window.Cypress && CYPRESS_MNEMONIC) {
     return [e2eWalletModule(chain.rpcUri)]
   }
-  return Object.entries(WALLET_MODULES)
-    .filter(([key]) => isWalletSupported(chain.disabledWallets, key))
-    .map(([, module]) => module())
+  const enabledWallets = Object.entries(WALLET_MODULES).filter(([key]) => isWalletSupported(chain.disabledWallets, key))
+
+  if (enabledWallets.length === 0) {
+    return [WALLET_MODULES.INJECTED()]
+  }
+
+  return enabledWallets.map(([, module]) => module())
 }
 
 export const isHardwareWallet = (wallet: ConnectedWallet): boolean => {


src/hooks/wallets/wallets.ts
```diff
@@ -67,9 +67,13 @@ export const getSupportedWallets = (chain: ChainInfo): WalletInit[] => {
   if (window.Cypress && CYPRESS_MNEMONIC) {
     return [e2eWalletModule(chain.rpcUri)]
   }
-  return Object.entries(WALLET_MODULES)
-    .filter(([key]) => isWalletSupported(chain.disabledWallets, key))
-    .map(([, module]) => module())
+  const enabledWallets = Object.entries(WALLET_MODULES).filter(([key]) => isWalletSupported(chain.disabledWallets, key))
+
+  if (enabledWallets.length === 0) {
+    return [WALLET_MODULES.INJECTED()]
```
iamacook, line 73: MetaMask can be disabled:

image

We should probably return an empty array here, as well as filter the "recommended injected wallets" in `getRecommendedInjectedWallets`. What do you think?

katspaugh: I think you're thinking too much into it. In practice, no chain will have all wallets disabled, and defaulting to the injected wallet is a quick and pragmatic solution w/o adding too many conditions IMHO.

iamacook: We should remove MetaMask from the list of wallets that can be disabled when we eventually [remove the "legacy" wallet names](https://github.com/safe-global/web-core/issues/713) then.

iamacook (approved):

liliya-soroka:
@katspaugh , I have a problem with the Ledger connection , only in the PR. No such problem on staging 
image

liliya-soroka:
verified
can be merged

PR #1602, opened 2023-01-26
feat: Integrate safe react components

yagopv:
## What it solves

Resolves https://github.com/safe-global/safe-react-components/issues/198

## How this PR fixes it
We are removing the theme files, fonts and `EthHashInfo` and `ExplorerButton` components from the web-core project and migrating them to a shared library.

This PR integrates the `safe-react-components` v2 implementing these changes


src/components/common/EthHashInfo/index.tsx
```diff
@@ -1,104 +1,23 @@
...
 
 const PrefixedEthHashInfo = ({
   showName = true,
   ...props
 }: EthHashInfoProps & { showName?: boolean }): ReactElement => {
   const settings = useAppSelector(selectSettings)
-  const currentChainId = useChainId()
-  const chain = useAppSelector((state) => selectChainById(state, props.chainId || currentChainId))
+  const currentChain = useCurrentChain()
+  const chain = useAppSelector((state) => selectChainById(state, props.chainId || currentChain?.chainId || ''))
```
iamacook, line 18: We should fall back to `useChainId` as it returns a default chain.

src/pages/_app.tsx
```diff
@@ -57,14 +59,19 @@ const InitApp = (): null => {
...
+  const themeMode = isDarkMode ? 'dark' : 'light'
 
   return (
-    <ThemeProvider theme={theme}>
-      <Sentry.ErrorBoundary showDialog fallback={ErrorBoundary}>
-        {children}
-      </Sentry.ErrorBoundary>
-    </ThemeProvider>
+    <SafeThemeProvider mode={themeMode}>
+      {(safeTheme: Theme) => (
```
iamacook, line 67: Question: why are we returning a provier and not a theme intitiation hook as we have in `web-core`? The dark mode hooks would likely benefit from existing in SRC as well, no?

iamacook (commented):
Great work! For consistency sake, I would suggest keeping the `lightPalette` name instead of renaming it.

yagopv:
> Great work! For consistency sake, I would suggest keeping the `lightPalette` name instead of renaming it.

You mean the `lightPalette as palette` in several files? Yeah i can do it

src/components/common/EthHashInfo/index.tsx
iamacook: I'd keep the logic as before but call `getBlockExplorerLink` with `chain`.

yagopv: aaah :D

yagopv: make sense no doubt 😂

yagopv: done

src/pages/_app.tsx
yagopv: well, we are [exporting one indeed](https://github.com/safe-global/safe-react-components/blob/7b8929d497003d5905ec732485499b43ce8d80b1/src/hooks/useThemeMode.ts#L1-L36). We export the `createSafeTheme` as well but the `SafeThemeProvider` is convenient because creates it for you using the theme mode and return it in the render prop fn.

I'm using the hook you have instead of the exported one because there are custom work there setting the `data-theme`

katspaugh:
`colors.ts` was used in the `css-vars.ts` script (and the corresponding command in package.json) to generate CSS variables from the theme. Needs to be adjusted accordingly.

src/pages/_app.tsx
yagopv: The provider is convenient due [this bug](https://github.com/mui/material-ui/issues/32806) (I added a new comment with our specific case). So it returns the children wrapping the `ThemeProvider` necessary for the components in the library to have access to the theme.

yagopv:
> `colors.ts` was used in the `css-vars.ts` script (and the corresponding command in package.json) to generate CSS variables from the theme. Needs to be adjusted accordingly.

yep i did it, is working i think

scripts/css-vars.ts
```diff
@@ -1,7 +1,7 @@
-import palette from '../src/styles/colors.js'
-import darkPalette from '../src/styles/colors-dark.js'
+import reactComponents from '@safe-global/safe-react-components'
 import spacings from '../src/styles/spacings.js'
 
+const { lightPalette: palette, darkPalette } = reactComponents
```
iamacook, line 4: Nit: can we change this to be `lightPalette` as well?

iamacook (approved):

scripts/css-vars.ts
yagopv: changed! thanks

dasanra (approved):

katspaugh (changes requested):
Looks like there's a visual regression in the EthHashInfo:

On this branch:
Screenshot 2023-02-03 at 07 30 59

On prod:
Screenshot 2023-02-03 at 07 32 21

yagopv:
> Looks like there's a visual regression in the EthHashInfo:

Ok I see, there is some 2.3em! important! width and heights not being applied in the lib. Will take a look

yagopv:
> Looks like there's a visual regression in the EthHashInfo:

@katspaugh the issue is fixed

katspaugh (approved):

liliya-soroka:
Verified

PR #1604, opened 2023-01-26
fix: migrate/sanitise legacy owner format

### LINKED ISSUE #1603: SAFE GLOBAL UI ALWAYS CRASHES
## Bug description
Whenever I click on the arrow to display different addresses, the app crashes.  

- Red rectangle points the button i am referring to, address blurred for privacy:
Screenshot 2023-01-26 at 16 10 19

- Error screen seen afterwards:
Screenshot 2023-01-26 at 15 55 47

- Console log:
Screenshot 2023-01-26 at 16 14 03

## Environment
 - Browser: Brave
 - Wallet: MetaMask + ledger
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to any url of the app, click on the arrow and crash

## Expected result
A list of addresses I have stored locally 

## Obtained result
Shown above

## Screenshots
Shown above 

# PLEASE DO NOT DISCONTINUE OLD GNOSIS-SAFE APP NEXT WEEK UNTIL THAT WORKS, THANKS.

iamacook:
## What it solves

Resolves #1603 

## How this PR fixes it

Legacy added Safe owner formats (from the original address book) are now migrated correctly and sanitised post-corrupt migration.

The original address book format alteration can be found [here](https://github.com/safe-global/safe-react/pull/2296/files#diff-bd2bbcb55c5dde038dc34315180a7f921a4df36bd925a72a39c8ffd33676392a):

image

## How to test it

1. Modify added Safe `localStorage` on `safe-react` to have `owners` as `Array<{ address: string, name: string }>` and reattempt migration. Observe no crashing when opening the sidebar in `web-core`.
2. Modify added Safe `localStorage` on `web-core` to have `owners` as `Array<{value: { address: string, name: string }}>` and refresh. Observe no crashing when opening the sidebar.

### DIFF
--- a/src/services/ls-migration/addedSafes.ts
+++ b/src/services/ls-migration/addedSafes.ts
@@ -1,6 +1,9 @@
 import { type AddedSafesState, type AddedSafesOnChain } from '@/store/addedSafesSlice'
 import type { LOCAL_STORAGE_DATA } from './common'
 import { parseLsValue } from './common'
+import { isChecksummedAddress } from '@/utils/addresses'
+import { isObject } from 'lodash'
+import type { AddressEx } from '@safe-global/safe-gateway-typescript-sdk'
 
 const IMMORTAL_PREFIX = '_immortal|v2_'
 
@@ -30,16 +33,24 @@ type OldAddedSafes = Record<
 
 export const migrateAddedSafesOwners = (
   owners: OldAddedSafes[string]['owners'],
-): AddedSafesState[string][string]['owners'] => {
-  return owners.map((value) => {
-    if (typeof value === 'string') {
-      return { value }
-    }
-    return {
-      value: value.address,
-      ...(value.name && { name: value.name }),
-    }
-  })
+): AddedSafesState[string][string]['owners'] | undefined => {
+  const migratedOwners = owners
+    .map((value) => {
+      if (typeof value === 'string' && isChecksummedAddress(value)) {
+        return { value }
+      }
+
+      if (isObject(value) && typeof value.address === 'string' && isChecksummedAddress(value.address)) {
+        const owner: AddressEx = {
+          value: value.address,
+          ...(typeof value.name === 'string' && { name: value.name }),
+        }
+        return owner
+      }
+    })
+    .filter((owner): owner is AddressEx => !!owner)
+
+  return migratedOwners.length > 0 ? migratedOwners : undefined
 }
 
 export const migrateAddedSafes = (lsData: LOCAL_STORAGE_DATA): AddedSafesState | void => {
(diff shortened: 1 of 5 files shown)


src/services/ls-migration/addedSafes.ts
```diff
@@ -23,11 +23,25 @@ type OldAddedSafes = Record<
...
+export const migrateAddedSafesOwners = (
+  owners: OldAddedSafes[string]['owners'],
+): AddedSafesState[string][string]['owners'] => {
+  return owners.map((value) => {
+    if (typeof value === 'string') {
+      return { value }
+    }
+    return {
+      value: value.address,
+      ...(value.name && { name: value.name }),
```
katspaugh, line 40: You're assuming here that if it's not a string, it's necessarily `{ name?: string; address: string }`. I would add a type check here as well and not write anything if it's neither.

src/store/addedSafesSlice.ts
```diff
@@ -36,6 +37,16 @@ export const addedSafesSlice = createSlice({
       // Otherwise, migrate
       return action.payload
     },
+    fixOwners: (state) => {
```
katspaugh, line 40: Maybe call it `migrateLegacyOwners` or something.

src/services/ls-migration/addedSafes.ts
iamacook: I've adjusted it accordingly and added checksum verification to match that of the address book migration as well.

katspaugh (approved):
Looks good, but I would (again) suggest renaming `fixLegacyOwners` to `migrateLegacyOwners` to more clearly categorize this action a migration-related.

src/services/ls-migration/addedSafes.ts
```diff
@@ -23,11 +26,33 @@ type OldAddedSafes = Record<
...
+  const migratedOwners = owners
+    .map((value) => {
+      if (typeof value === 'string' && isChecksummedAddress(value)) {
+        return { value }
+      }
+
+      if (isObject(value) && typeof value.address === 'string' && isChecksummedAddress(value.address)) {
+        const owner: AddressEx = {
+          value: value.address,
+          ...(typeof value.name === 'string' && { name: value.name }),
```
schmanu, line 46: Do we use this name of the owners anywhere? 
I only saw places where we lookup the name in the address book.

Probably we should add the name to the address book if the address has no entry yet?

katspaugh: Let's not complicate this, it's probably one out of a million users who will have this old format.

iamacook: > Do we use this name of the owners anywhere?

We seemingly don't, but our Safe creation does [add full `AddressEx` entries](https://github.com/safe-global/web-core/blob/dev/src/components/new-safe/create/logic/address-book.ts#L38) nonetheless.

I see no harm in keeping it. What do you think?

> Probably we should add the name to the address book if the address has no entry yet?

The address book is migrated separately. I don't think they should adjust the entries of one another.

schmanu (approved):

iamacook:
After discussing a bug with @francovenica, I've altered the flags to always call the fix action, regardless of environment. It should now be possible to test this on our PR deployments. (The actual migration will still only happen on production).

francovenica:
Lookd good now. Merge it

PR #1605, opened 2023-01-27
fix: prevent multiple pairing tracking calls

### LINKED ISSUE #1600: [GTM tracking] - Mobile pairing is being tracked multiple times
## Bug description
When a wallet is connected it is tracked in the GTM. That track should happen once. For mobile pairing with the safe mobile app that tracking happens twice, and even more times if other wallets are also currently connected to the app

## Environment
 - Browser: Chrome
 - Wallet: Safe mobile (mobile pairing)
 - Chain: Any
 - Env: RC 1.4.0 #1599

## Steps to reproduce
 1. Go to the app and turn on analytics. Open the dev console 
 2. Connect with mobile pairing (result 1)

 1. Disconnect all wallets
 2. Connect with MM first, then with Trezor (or ledger) so you can swith between both
 3. Connect with mobile pairing (result 2)

## Expected result
The event "Connect wallet" should happen once.

## Obtained result
1 - The event is triggered twice
2 - The event is triggered multiple times, (up to 5 times in some cases)

## Screenshots
safe mobile connect

iamacook:
## What it solves

Resolves #1600

## How this PR fixes it

A re-entrant lock has been added to our `connectWallet` wrapper.

When pairing via connect/switch wallet buttons and then scanning the QR code, `connectWallet` will have been called and then called again by the WalletConnect `connect` event. We cannot remove the `connectWallet` from the `connect` event callback because that is responsible for the reconnection of previously cached pairing sessions.

## How to test it

1. Connect to MetaMask and switch to Safe Mobile. Observe one tracking event for each wallet.
2. Switching back to MetaMask and then back to Safe Mobile should also only track each wallet once.
3. Refresh whilst connected to Safe Mobile. Observe the cached session reconnecting and one event being tracked.

### DIFF
--- a/src/hooks/wallets/useOnboard.ts
+++ b/src/hooks/wallets/useOnboard.ts
@@ -83,8 +83,18 @@ const trackWalletType = (wallet: ConnectedWallet) => {
 // Detect mobile devices
 const isMobile = () => /iPhone|iPad|iPod|Android/i.test(navigator.userAgent)
 
+// `connectWallet` is called when connecting/switching wallets and on pairing `connect` event (when prev. session connects)
+// This re-entrant lock prevents multiple `connectWallet`/tracking calls that would otherwise occur for pairing module
+let isConnecting = false
+
 // Wrapper that tracks/sets the last used wallet
 export const connectWallet = async (onboard: OnboardAPI, options?: Parameters<OnboardAPI['connectWallet']>[0]) => {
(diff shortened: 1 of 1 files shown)


katspaugh (approved):

francovenica:
LGTM
Safe mobile pairing and the other wallets as well only create 1 tracking action.

PR #1606, opened 2023-01-27
fix: shown unknown when current chain not found in config

iamacook:
## What it solves

Resolves missing wallet details in header

## How this PR fixes it

When the wallet is connected to a chance not available in the UI, it now displays "Unknown" instead.

image

## How to test it

Connect wallet to UI and switch the wallet chain to one not available in the UI. Observe the wallet displaying with "Unknown" in the header and account centre. Navigate to Safe creation and observe the widget shown "Unknown" as well.

## Screenshots

image

image

image

### DIFF
--- a/src/components/common/WalletInfo/index.tsx
+++ b/src/components/common/WalletInfo/index.tsx
@@ -13,7 +13,7 @@ import css from './styles.module.css'
 export const UNKNOWN_CHAIN_NAME = 'Unknown'
 
 const WalletInfo = ({ wallet }: { wallet: ConnectedWallet }): ReactElement => {
-  const walletChain = useAppSelector((state) => selectChainById(state, wallet?.chainId || ''))
+  const walletChain = useAppSelector((state) => selectChainById(state, wallet.chainId))
   return (
     <Box className={css.container}>
       <Box className={css.imageContainer}>
@@ -34,7 +34,6 @@ const WalletInfo = ({ wallet }: { wallet: ConnectedWallet }): ReactElement => {
               address={wallet.address}
               showName={false}
               showAvatar
-              showPrefix={!walletChain}
               avatarSize={12}
             />
           )}
--- a/src/components/common/WalletInfo/index.tsx
+++ b/src/components/common/WalletInfo/index.tsx
@@ -5,11 +5,15 @@ import type { ReactElement } from 'react'
 import EthHashInfo from '@/components/common/EthHashInfo'
 import WalletIcon from '@/components/common/WalletIcon'
 import type { ConnectedWallet } from '@/hooks/wallets/useOnboard'
-import type { ChainInfo } from '@safe-global/safe-gateway-typescript-sdk'
+import { useAppSelector } from '@/store'
+import { selectChainById } from '@/store/chainsSlice'
 
 import css from './styles.module.css'
 
-const WalletInfo = ({ wallet, chain }: { wallet: ConnectedWallet; chain?: ChainInfo }): ReactElement => {
+export const UNKNOWN_CHAIN_NAME = 'Unknown'
+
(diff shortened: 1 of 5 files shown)


usame-algan:
For the overview in Safe creation we also use `WalletInfo` but with a different `chain` object. I suggest we also change it to display "Unknown".
Screenshot 2023-01-27 at 15 37 17

katspaugh (approved):
Nice improvement, I was confused by this on staging.

iamacook:
> For the overview in Safe creation we also use `WalletInfo` but with a different `chain` object. I suggest we also change it to display "Unknown".

I've refactored `WalletInfo` to display the chain information of the connected wallet so that fixes the widget issue. I also found that the `AccountCenter` needed the name showing as well so I tweaked that too (see description for screenshots).

src/components/common/WalletInfo/index.tsx
```diff
@@ -19,17 +23,18 @@ const WalletInfo = ({ wallet, chain }: { wallet: ConnectedWallet; chain: ChainIn
...
           {wallet.ens ? (
             <div>{wallet.ens}</div>
           ) : (
             <EthHashInfo
-              prefix={chain.shortName}
+              prefix={walletChain?.shortName || ''}
               address={wallet.address}
               showName={false}
               showAvatar
+              showPrefix={!walletChain}
```
DaniSomoza, line 37: I think that this `!walletChain` flag is removing always the prefix in the address

Captura de pantalla 2023-01-27 a las 16 13 34
Captura de pantalla 2023-01-27 a las 16 13 42

iamacook: Nice catch. Remnants from before the refactor 🙈. I've fixed it.

DaniSomoza (approved):

francovenica:
There is an inconsistency with prefixes. Maybe we should remove the prefix of the address if the user is not in a known network:

Here I'm goerly, but MM is connected to Optimism:
image
image

iamacook:
> There is an inconsistency with prefixes. Maybe we should remove the prefix of the address if the user is not in a known network:
> 
> Here I'm goerly, but MM is connected to Optimism: image image

I originally adjusted the account center and then realised that the prefix would need to be adjusted app-wide. I think it semantically makes the most sense to not show the prefix as the network is inherently "unknown". The prefix is now not shown when connected to an unknown network.

iamacook:
> There is an inconsistency with prefixes.

I've adjusted the prefix in the account center to show that of the unknown network.

image

src/utils/wallets.ts
```diff
@@ -48,3 +51,10 @@ export const isWalletUnlocked = async (walletName: string): Promise<boolean> =>
 
   return false
 }
+
+export const getEthHashInfoPrefix = (
+  shortName?: ChainInfo['shortName'],
+  chainId?: ConnectedWallet['chainId'],
+): string | undefined => {
+  return shortName || (chainId ? Object.entries(chains).find((chain) => chain[1] === chainId)?.[0] : undefined)
+}
```
katspaugh, line 60: I wouldn't place it here, as it's only used in that one place.
This is short enough for an inline call IMHO.
```
shortName || Object.entries(chains).find(([_, val]) => val === chainId)?.[0]
```

iamacook: It's used in two places though.

katspaugh: I would suggest extracting only the find part, as it's used in many more places in the app.
I.e.
```
const getChainPrefix = (chainId) => Object.keys(chains).find(key => chains[key] === chainId
```
And it shouldn't be in `wallets.ts` obviosuly.

katspaugh: Just saw you did exactly that in #1619. :)

francovenica:
Ok, LGTM
image

PR #1608, opened 2023-01-30
fix: check if all tokens are hidden on dashboard

### LINKED ISSUE #1607: [Hidden tokens] The token counter never finish loading if all tokens are hidden
## Bug description
Tested in a safe 1.0.0. 
This safe only has ETH tokens in the list. After hiding the token, in the main dashboard the token counter loads indefinitely. Tried the same in current prod build and there it works fine: the counter shows a " 0 "

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Goerli
 - Env: https://release_1_4_0--webcore.review-web-core.5afe.dev/gor:0x41EfabA9C6264BA81f235B7Bc2e3eb2390782790/home
 - RC: https://github.com/safe-global/web-core/pull/1599

## Steps to reproduce
 1. Go to The safe listed
 2. Hide the token it has in the assets list
 3. Go to the dashboard

## Expected result
The Counter should show a "0" in the counter (this happens like that in current prod build)

## Obtained result
The counter loads indefinitely

## Screenshots
image
image

iamacook:
## What it solves

Resolves #1607

## How this PR fixes it

The hiding of native tokens is no longer possible.

## How to test it

1. Open a Safe with multiple tokens and observe that the "Tokens" amount on the dashboard switches from the loading skeleton to the number of tokens owned.
2. Navigate to the assets page and hide all tokens then back to the dashboard and refresh. Observe that the "Tokens" amount changes from the loading skeleton to 0.
3. Switch to a Safe that has fewer tokens (hidden token amount is higher). Observe that the the loading skeleton switches from loading skeleton to 0.

Varing amounts of hidden tokens should no make the loading skeleton change to 0 to n. It should always change from the skeleton to the amount of tokens not hidden.

## Screenshots

hidden-tokens


katspaugh (changes requested):
A whole global hook just to check if all tokens are hidden?
I would instead show the skeleton only when balances are actually loading.
Or alternatively, disable hiding the default token like it was originally suggested.

iamacook:
> A whole global hook just to check if all tokens are hidden? I would instead show the skeleton only when balances are actually loading. Or alternatively, disable hiding the default token like it was originally suggested.

If we hide the amount based on when the balances are loading it will first render 0 before the true balance. I think the latter is a better suggestion.

iamacook:
> A whole global hook just to check if all tokens are hidden? I would instead show the skeleton only when balances are actually loading. Or alternatively, disable hiding the default token like it was originally suggested.

If we hide the amount based on when the balances are loading it will first render 0 before the true balance. I think the latter is a better suggestion.

src/hooks/useHasHiddenAllTokens.ts
```diff
@@ -0,0 +1,14 @@
+import { useMemo } from 'react'
+import { useVisibleBalances } from './useVisibleBalances'
+import useHiddenTokens from './useHiddenTokens'
+
+const useHasHiddenAllTokens = (): boolean => {
+  const { balances, loading } = useVisibleBalances()
+  const hiddenTokens = useHiddenTokens()
+
+  return useMemo(() => {
+    return !loading && hiddenTokens.length > 0 && hiddenTokens.length >= balances.items.length
```
schmanu, line 10: This condition is technically not correct.
`useHiddenTokens` returns all hidden tokens per chain. It does not check if those are also part of the current Safe's balance. 

`hiddenTokens.length >= balances.items.length` therefore does not really mean anything because there could for instance be 10 hidden spam tokens and 3 legit tokens in the current balance leading to 10 >= 3 but at the same time not all tokens are hidden.

src/components/dashboard/Overview/Overview.tsx
```diff
@@ -139,7 +141,9 @@ const Overview = (): ReactElement => {
                     <Typography color="border.main" variant="body2">
                       Tokens
                     </Typography>
-                    <StyledText fontSize="lg">{tokenCount || <ValueSkeleton />}</StyledText>
+                    <StyledText fontSize="lg">
+                      {tokenCount || hasHiddenAllTokens ? tokenCount : <ValueSkeleton />}
```
schmanu, line 145: I would attach this to 
```ts
const { balances, loading } = useVisibleBalances()
{/* ... */}
<StyledText fontSize="lg">
  {loading ? <ValueSkeleton /> : tokenCount }
</StyledText>
```

schmanu (changes requested):

src/hooks/useHasHiddenAllTokens.ts
iamacook: I've altered the PR to only disable the hiding of native tokens. I think it's a cleaner solution.

src/components/dashboard/Overview/Overview.tsx
iamacook: When the `loading` flag changes, it will initially show `0` tokens before the `tokenCount`. I've changed the logic to not allow for the hiding of native token now. I think it's a cleaner solution.

schmanu (approved):
🚀

francovenica:
I still see the issue even if the main network token is not hidden. This issue I have with this safe also happens in the RC, so maybe the main problem is not just about hiding the tokens. Check this safe
https://dashboard_hidden_tokens--webcore.review-web-core.5afe.dev/oeth:0xe81170d6Db8834CCA22e317089Eff1d06217BC70/home

iamacook:
> I still see the issue even if the main network token is not hidden. This issue I have with this safe also happens in the RC, so maybe the main problem is not just about hiding the tokens.

@schmanu lent a hand on this and it should now be resolved.

src/components/dashboard/Overview/Overview.tsx
```diff
@@ -100,6 +100,8 @@ const Overview = (): ReactElement => {
   // Native token is always returned even when its balance is 0
   const tokenCount = useMemo(() => balances.items.filter((token) => token.balance !== '0').length, [balances])
   const nftsCount = useMemo(() => (nfts ? `${nfts.next ? '>' : ''}${nfts.results.length}` : ''), [nfts])
+  // Only initially the items are completely empty because our gateway always returns the native token balance
+  const isTokenCountLoading = balances.items.length === 0
```
katspaugh, line 104: If this is enough, maybe just keep this? The isNative check is not really needed then.

iamacook:
Closing in favour of https://github.com/safe-global/web-core/pull/1614.

PR #1614, opened 2023-01-30
fix: balances loading flag

### LINKED ISSUE #1607: [Hidden tokens] The token counter never finish loading if all tokens are hidden

iamacook:
## What it solves

Resolves #1607

## How this PR fixes it

The balances `loading` flag now checks the reference equality of the initial state.

## How to test it

1. Open a Safe with multiple tokens and observe that the "Tokens" amount on the dashboard switches from the loading skeleton to the number of tokens owned.
2. Navigate to the assets page and hide all tokens then back to the dashboard and refresh. Observe that the "Tokens" amount changes from the loading skeleton to 0.
3. Switch to a Safe that has fewer tokens (hidden token amount is higher). Observe that the the loading skeleton switches from loading skeleton to 0.

Varing amounts of hidden tokens should no make the loading skeleton change to 0 to n. It should always change from the skeleton to the amount of tokens not hidden.

## Screenshots

hidden-tokens

### DIFF
--- a/src/components/dashboard/Overview/Overview.tsx
+++ b/src/components/dashboard/Overview/Overview.tsx
@@ -83,7 +83,7 @@ const Overview = (): ReactElement => {
   const router = useRouter()
   const safeAddress = useSafeAddress()
   const { safe, safeLoading, safeLoaded } = useSafeInfo()
-  const { balances } = useVisibleBalances()
+  const { balances, loading: balancesLoading } = useVisibleBalances()
   const [nfts] = useCollectibles()
   const chain = useCurrentChain()
   const { chainId } = chain || {}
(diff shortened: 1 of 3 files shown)


katspaugh (approved):
👍

francovenica:
Looks good to me. 
Tried in safes with just 1 token, with the native token with 0 balance (which shows a "0" in the token counter in the dashboard), safes with over 20 tokens, hiding and showing tokens.

PR #1615, opened 2023-01-30
fix: Only use multisend call if its more than 1 transaction

### LINKED ISSUE #1536: Transactions that "will fail" cannot be submitted
See https://github.com/5afe/safe-support/issues/115

usame-algan:
## What it solves

Resolves #1536 

Note: Somehow the Tenderly simulation shows success but I think its a general issue with 1.1.1 Safe contracts that the simulation doesn't propagate if the call was reverted so Tenderly still returns "Success". I think its unrelated to this issue and would fix it seperately.

## How this PR fixes it

Checks the number of transactions that are submitted through safe apps and calls `createMultiSendCallOnlyTx` or `createTx`. Previously it would try to call `createMultiSendCallOnlyTx` but without passing any `safeTxGas` which caused it to fail because the logic inside `safe-core-sdk` falls back to creating a single safe transaction if the array only contains one transaction so we should also handle that case on our side.

## How to test it

1. Open a 1.1.1 Safe
2. Open the TxBuilder safe app
3. Create and batch **one** transaction that would fail/revert e.g. by interacting with 0x22e3f828b3f47dacfacd875d20bd5cc0879c96e7 on mainnet
4. Observe the Submit button being enabled
5. Batch more than one transaction through safe apps in 1.1.1 and 1.3.0 safes and observe being able to queue/execute the tx

## Screenshots
Screenshot 2023-01-30 at 16 44 48

### DIFF
--- a/src/components/safe-apps/SafeAppsTxModal/ReviewSafeAppsTx.tsx
+++ b/src/components/safe-apps/SafeAppsTxModal/ReviewSafeAppsTx.tsx
@@ -27,7 +27,7 @@ type ReviewSafeAppsTxProps = {
 const ReviewSafeAppsTx = ({
   safeAppsTx: { txs, requestId, params, appId, app },
 }: ReviewSafeAppsTxProps): ReactElement => {
-  const { createMultiSendCallOnlyTx, dispatchSafeAppsTx } = useTxSender()
+  const { createMultiSendCallOnlyTx, dispatchSafeAppsTx, createTx } = useTxSender()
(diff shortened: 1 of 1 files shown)


iamacook (approved):

katspaugh (approved):

francovenica:
LGTM

Tried in a safe 1.1.1 in mainnet, using the contract given, with a method "setDelay" just because it has only 1 field.
A comparison between this PR and prod. Same safe, method and values.

image
image

PR #1616, opened 2023-01-31
Release/1.4.5

katspaugh:
## Features
* NFT page redesign

## Bug fixes
* fix: balances loading flag
* fix: Only use multisend call if its more than 1 transaction (#1615)
* fix: old app export link (#1625)


francovenica:
Issue:
The GTM tracks whenver the send button is clicked, even if the button is disabled. Only track if the button is enabled
nft trakcing

PR #1620, opened 2023-01-31
fix: fetch queue when Safe list item is visible

DiogoSoaress:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/1610

## How this PR fixes it
Uses `IntersectionObserver` to assert that the SafeListItem is in the viewport before sending a API request

## How to test it
On adjusting the application viewport, the `transactions/queued` network requests should equal the number of Safes visible in the Safe list sidebar

## Screenshots
Screenshot 2023-01-31 at 11 18 08

### DIFF
--- a/src/components/sidebar/SafeListItem/index.tsx
+++ b/src/components/sidebar/SafeListItem/index.tsx
@@ -21,6 +21,7 @@ import EthHashInfo from '@/components/common/EthHashInfo'
 import { sameAddress } from '@/utils/addresses'
 import PendingActionButtons from '@/components/sidebar/PendingActions'
 import usePendingActions from '@/hooks/usePendingActions'
+import useIntersectionObserver from '@/hooks/useIntersectionObserver'
 
 const SafeListItem = ({
   address,
@@ -48,7 +49,8 @@ const SafeListItem = ({
   const isCurrentSafe = chainId === currChainId && sameAddress(safeAddress, address)
   const name = allAddressBooks[chainId]?.[address]
   const shortName = chain?.shortName || ''
-  const { totalQueued, totalToSign } = usePendingActions(chainId, isAdded ? address : undefined)
+  const isVisible = useIntersectionObserver(safeRef)
+  const { totalQueued, totalToSign } = usePendingActions(chainId, isVisible, isAdded ? address : undefined)
 
   // Scroll to the current Safe
   useEffect(() => {
--- a/src/components/sidebar/SafeListItem/index.tsx
+++ b/src/components/sidebar/SafeListItem/index.tsx
@@ -50,7 +50,7 @@ const SafeListItem = ({
   const name = allAddressBooks[chainId]?.[address]
   const shortName = chain?.shortName || ''
   const isVisible = useIntersectionObserver(safeRef)
-  const { totalQueued, totalToSign } = usePendingActions(chainId, isVisible, isAdded ? address : undefined)
+  const { totalQueued, totalToSign } = usePendingActions(chainId, isAdded && isVisible ? address : undefined)
 
   // Scroll to the current Safe
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/hooks/usePendingActions.ts
```diff
@@ -17,16 +17,16 @@ const getSignableCount = (queue: TransactionListPage, walletAddress: string): nu
...
   const { safeAddress: currentSafeAddress } = useSafeInfo()
   const { page: currentSafeQueue } = useTxQueue()
   const isCurrentSafe = currentSafeAddress === safeAddress
 
   const [loadedQueue] = useAsync<TransactionListPage>(() => {
-    if (isCurrentSafe || !safeAddress) return
+    if (isCurrentSafe || !safeAddress || !isVisible) return
     return getTransactionQueue(chainId, safeAddress)
-  }, [chainId, safeAddress, isCurrentSafe])
+  }, [chainId, safeAddress, isCurrentSafe, isVisible])
```
katspaugh, line 29: `isVisible` is going to be flipped back and forth as you scroll, and it will trigger the effect every time.

src/components/sidebar/SafeListItem/index.tsx
```diff
@@ -48,7 +49,8 @@ const SafeListItem = ({
   const isCurrentSafe = chainId === currChainId && sameAddress(safeAddress, address)
   const name = allAddressBooks[chainId]?.[address]
   const shortName = chain?.shortName || ''
-  const { totalQueued, totalToSign } = usePendingActions(chainId, isAdded ? address : undefined)
+  const isVisible = useIntersectionObserver(safeRef)
+  const { totalQueued, totalToSign } = usePendingActions(chainId, isVisible, isAdded ? address : undefined)
```
katspaugh, line 53:
```suggestion
  const { totalQueued, totalToSign } = usePendingActions(chainId, isAdded && isVisible ? address : undefined)
```
Since address is optional already...

DiogoSoaress: Nice suggestion! This also the previous comment is also addressed 👍

katspaugh: No, it won't. It will change from false to true and back as you scroll, triggering loading every time.

DiogoSoaress: I've cached the loadedQueue to avoid issuing request on scrolling back and forth

src/hooks/usePendingActions.ts
```diff
@@ -22,10 +22,19 @@ const usePendingActions = (chainId: string, safeAddress?: string): PendingAction
...
-    return getTransactionQueue(chainId, safeAddress)
+
+    if (cachedQueue.current) {
+      return cachedQueue.current
+    }
+
+    const fetchedQueue = await getTransactionQueue(chainId, safeAddress)
+    cachedQueue.current = fetchedQueue
+
+    return fetchedQueue
```
katspaugh, line 37: Looks good!

Would this work without async/await as well? So that if you scroll quickly, it wouldn't create simultaneous requests.

```suggestion
    if (!cachedQueue.current) {
      cachedQueue.current = getTransactionQueue(chainId, safeAddress) 
    }
    return cachedQueue.current
```

DiogoSoaress: Nice suggestion! We cache earlier and works like a charm

src/hooks/usePendingActions.ts
```diff
@@ -22,10 +22,16 @@ const usePendingActions = (chainId: string, safeAddress?: string): PendingAction
   const { safeAddress: currentSafeAddress } = useSafeInfo()
   const { page: currentSafeQueue } = useTxQueue()
   const isCurrentSafe = currentSafeAddress === safeAddress
+  const cachedQueue = useRef<Promise<TransactionListPage>>()
 
-  const [loadedQueue] = useAsync<TransactionListPage>(() => {
+  const [loadedQueue] = useAsync<TransactionListPage | undefined>(() => {
```
katspaugh, line 27: You can now remove `undefined` then. useAsync handles it.

katspaugh (approved):

PR #1621, opened 2023-01-31
test: non-owner spending limit beneficiary

iamacook:
## What it solves

Adds a test for non-owner spending limit beneficiaries (https://github.com/safe-global/web-core/pull/1572).

## How this PR fixes it

An E2E test has been added that checks whether the "New transaction" button is enabled and the transaction creation modal flow is accessible/submittable for non-owner spending limit beneficiaries.

## How to test it

Run Cypress locally and execute the `non_owner_spending_limit.cy.js` test.

## Screenshots

image


DiogoSoaress (approved):
Covers the case  👍

PR #1622, opened 2023-01-31
Refactor: move safe address from path to query

katspaugh:
Use query params for Safe addresses in the URL or stick with the path params.

### Visual difference
Current style (pathname):
https://app.safe.global/eth:0xA77DE01e157f9f57C7c4A326eeE9C4874D0598b6/balances
Query param style:
https://app.safe.global/balances?safe=eth:0xA77DE01e157f9f57C7c4A326eeE9C4874D0598b6

### Performance implications
The query param would allow our web server to load a pre-compiled static page (e.g. balances.html) with the page layout already in place and skeletons in place of actual data.
This would drastically increase the initial page load speed as well as caching.
The current approach is we first load a 404 page that redirects to the right static page.
The query param also makes hosting our app much simpler, no need for server-side redirects.

### Bookmarks and redirects
Nothing would change for users who already have the pathname URLs bookmarked, as they’d be simply redirected to the query URLs.

### Implementation notes
I had to add a hook to replace `%3A` with `:` in the `safe` query param. All params are URI fragment-encoded, but this is not strictly needed for this param, and `:` looks cleaner.

## How to test
Everything should work as before, only the address in the browser URL bar will be different.
* Test navigation links and reloading the page on different routes


iamacook (commented):
Do we still need `/public/_redirects`?

katspaugh:
@iamacook nope, deleted 👍

iamacook (approved):

francovenica:
LGTM

Navigated through the site and reload everywhere, i had no issues
I've checked that old URL would redirect to this new one
I've checked that GA tracking still tracks the the url in the old format to not create new entries

PR #1624, opened 2023-02-01
Fix: nft page mobile layout

katspaugh:
A few small fixes for the NFT page layout on mobile and tablet.

This is how it was:

Now:
Screenshot 2023-02-01 at 08 22 35

### DIFF
--- a/src/components/nfts/NftGrid/index.tsx
+++ b/src/components/nfts/NftGrid/index.tsx
@@ -115,7 +115,7 @@ const NftGrid = ({ nfts, selectedNfts, isLoading, children, onSelect }: NftsTabl
                   align="left"
                   padding="normal"
                   sx={{
-                    display: headCell.xsHidden ? { xs: 'none', md: 'table-cell' } : undefined,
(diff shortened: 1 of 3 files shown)


iamacook (commented):
Looks great!

I noticed that the placeholder is not centred. (It's slightly offset at higher resolutions too.)

image

image

Can we make the table width 100% at the `md` breakpoint as well?

image

francovenica:
The tooltip about the links in Mobile is not well placed, I'd say even that it shouldn't show for mobile since there are no links to click 
image

francovenica:
If the banner at the top is enabled, the Token/NFT tabs overlap the Select All/Send buttons.
mobile NFT

katspaugh:
Fixed Aaron's finding.
@francovenica the banner won't exist after Feb 7th, so let's ignore this for now.
Reg. the tooltip on mobile, it's hard to hide it w/o using JS, so I suggest we ignore that too.

francovenica:
Ok, then looks good

PR #1625, opened 2023-02-01
Fix: old app export link

katspaugh:
Fixes the link in the blue banner. The old app will redirect to the new app, except for the export path, so we shouldn't link to settings.


francovenica:
LGTM. 
Comparison old / new:
image
image

francovenica:
Ok, it was here as well. looks good 
image

PR #1626, opened 2023-02-01
Fix: track nft send only when enabled

katspaugh:
## What it solves
The Send button was tracked even if disabled.
See [this comment](https://github.com/safe-global/web-core/pull/1616#issuecomment-1411965500).


DiogoSoaress (approved):
It looks good to me

francovenica:
LGTM:
nft trakcing fix

PR #1629, opened 2023-02-01
Fix: avoid flashing the sidebar on mobile and desktop

katspaugh:
## What it solves

The sidebar is initially rendered in an uncollapsed state because that's how it will stay on the desktop.
Then we detect if it's a small screen via JS and collapse the sidebar. This causes the sidebar to flash on the initial page load on mobile.

This PR hides the sidebar via CSS on mobile until it's collapsed.

## How to test
* The sidebar should not be initially visible on mobile (compare to prod)
* On desktop, when going from Safe Creation to an individual Safe, the sidebar should appear instantly (it's animated on on prod)
* In individual Safe Apps, the sidebar should still be animated when collapsing/expanding it.

### DIFF
--- a/src/components/common/ConnectWallet/AccountCenter.tsx
+++ b/src/components/common/ConnectWallet/AccountCenter.tsx
@@ -113,7 +113,13 @@ const AccountCenter = ({ wallet }: { wallet: ConnectedWallet }) => {
 
           <ChainSwitcher fullWidth />
 
-          <Button variant="contained" size="small" onClick={handleSwitchWallet} fullWidth>
+          <Button
+            variant="contained"
+            size="small"
+            onClick={handleSwitchWallet}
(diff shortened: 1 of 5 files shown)


usame-algan (approved):
Nice! 💯

katspaugh:
@usame-algan I noticed another similar annoyance: when navigating from Welcome to a Safe page, it expands the sidebar with animation. I'll try to remove the animation in this case.

francovenica:
The button on the top left doesn't work in mobile view. If the idea of that button is that is redundant since the list of added safes is in the "My Safes" menu in the middle then remove the hamburger icon

image

katspaugh:
Fixed, I made the menu icon link to the start page if the sidebar is disabled.

PR #1630, opened 2023-02-01
[EIP-1271 Sync] Synchronous signing flow for arbitrary thresholds

schmanu:
Part of #1886 
## What it solves
Wait until the message is fully signed and return the signed message directly in the Safe App's RPC response. 
This allows immediate off-chain signing on OpenSea w/o going through our API.
Further explore how we can enable this for m/n safes that provide the required signatures while the session is open. 

Depends on https://github.com/safe-global/safe-apps-sdk/pull/458
WC app: https://github.com/safe-global/safe-react-apps/pull/657


francovenica:
Once you sign the button to switch wallets shows up so you can sign with another of the owners. The issue is that, if you change to something that is not metamask (or another browser extension) and you messed up (picked an account that is not owner) you are locked waiting for somebody else to sign in another browser/computer, because the modal will just say that the current account is not owner and won't give you the option to switch again (and the modal overlaps the page, so you cannot access the top right corner of the topbar to change accounts either).
Maybe we should give the users the option to change accounts in the modal if they are not currently the owner of the safe, if not now, in a future implementation

image
image

francovenica:
aside my previous comment the feature looks good

I used the WC app since is the one dapp that can sign these types of off-chain tx, connecting to the snapshot website and using a safe test group to do proposals, also tried connecting to opensea just fine (didn't buy anything tho)

I tried with MM, Phone MM, safe mobile pairing, trezor and leger and tried with several treshold limits and it worked fine.
The checkbox button in the settings works fine as well, it reverts to the old way of signing

The other dapps should work as they did before. I don't have much funds in any network to try other apps, but I made sure that trying to use them still works (propsed tx, but didn't execute them). I only tried with just a handful of apps. we will have to wait for feedback if there are any issues

I think is good to add to the RC. I'll give it another try there

katspaugh (approved):
🚀

PR #1637, opened 2023-02-02
fix: clarify SDK initialisation error notifications

### LINKED ISSUE #1632: Change error message: "The Safe SDK could not be initialised..."
## Bug description

Currently we display the error message "The Safe SDK could not be initialized. Please try connecting the wallet again."

## Environment

Occurred on
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Polygon
 

## Steps to reproduce
not clear

## Expected result

Provide better user guidance. If not clearer instructions can be found, use the following message:

"Please try connecting your wallet again."

## Screenshots
image (6)

SandeepGumaste:
## What it solves

Resolves #1632 

## How this PR fixes it
Changed error message from `The Safe SDK could not be initialized. Please try connecting the wallet again.` to `Please try connecting your wallet again.` in useInitSafeCoreSDK hook


iamacook (commented):
We should provide more details by leveraging the [`detailedMessage`](https://github.com/safe-global/web-core/blob/dev/src/store/notificationsSlice.ts#L9) property. It is [conditionally rendered inside notifications](https://github.com/safe-global/web-core/blob/dev/src/components/common/Notifications/index.tsx#L78-L83). I'd suggest using the `e.message`.

SandeepGumaste:
@iamacook Updated it to show e.message in detailed message

src/hooks/coreSDK/useInitSafeCoreSDK.ts
```diff
@@ -26,9 +26,10 @@ export const useInitSafeCoreSDK = () => {
       .catch((e) => {
         dispatch(
           showNotification({
-            message: `The Safe SDK could not be initialized. Please try connecting the wallet again.`,
+            message: `Please try connecting your wallet again.`,
```
katspaugh, line 29: Thanks for the fix, @SandeepGumaste!

I would mention RPC in the message because that's what usually makes the SDK fail, e.g. `Problem connecting to the RPC provider. Please try re-connecting your wallet.`

katspaugh, line 29: Thanks for the fix, @SandeepGumaste!

I would mention RPC in the message because that's what usually makes the SDK fail, e.g. 

```suggestion
            message: 'Problem connecting to the RPC provider. Please try connecting your wallet again.',
```

iamacook (approved):

katspaugh (approved):

PR #1646, opened 2023-02-07
fix: spending limits for tokens without balance

schmanu:
## What it solves
If a spending limit exists for a token without a balance we cannot infer the token info from our balance items.

## How this PR fixes it
When fetching spending limits we also get the token info and store it in the spending limits slice.
We first try to get the token info from our balances and then to get them from on-chain.

## How to test it
1. Create a spending limit for a token
2. Transfer the complete balance of that token out of the safe
3. Open the spending limits settings page

OR

Test it with this safe:
`gor:0xAecDFD3A19f777F0c03e6bf99AAfB59937d6467b`

It has a spending limit for COW tokens but holds not COW tokens.

## Analytics changes
None


src/hooks/loadables/useLoadSpendingLimits.ts
```diff
@@ -19,18 +24,48 @@ const isModuleEnabled = (modules: string[], moduleAddress: string): boolean => {
...
-
   return {
     beneficiary: delegate,
-    token,
+    token: getTokenInfoFromBalances(tokenInfoFromBalances, token) ||
+      (await getTokenInfoOnChain(token)) || {
+        address: token,
+        decimals: 18,
+        symbol: '',
+      },
```
iamacook, line 68: I'd suggest moving the fallback to a constant.

src/hooks/loadables/useLoadSpendingLimits.ts
```diff
@@ -1,16 +1,21 @@
...
 import { getSpendingLimitContract, getSpendingLimitModuleAddress } from '@/services/contracts/spendingLimitContracts'
 import type { AddressEx } from '@safe-global/safe-gateway-typescript-sdk'
 import { sameAddress } from '@/utils/addresses'
+import { ERC20__factory } from '@/types/contracts'
 import type { AllowanceModule } from '@/types/contracts'
+
 import { sameString } from '@safe-global/safe-core-sdk/dist/src/utils'
+import useBalances from '../useBalances'
+
+type TokenInfoFromBalances = { [address: string]: { symbol: string; decimals: number; logoUri?: string } }
```
iamacook, line 18: Let's reuse the type to be safe.

```suggestion
type TokenInfoFromBalances = { [address: string]: Omit<SpendingLimitState['token'], 'address'>
```

src/hooks/loadables/useLoadSpendingLimits.ts
```diff
@@ -73,12 +120,27 @@ export const useLoadSpendingLimits = (): AsyncResult<SpendingLimitState[]> => {
...
+    }
+    return balances.items.reduce((dictionary, item) => {
+      dictionary[item.tokenInfo.address] = {
+        symbol: item.tokenInfo.symbol,
+        decimals: item.tokenInfo.decimals,
+        logoUri: item.tokenInfo.logoUri,
+      }
+      return dictionary
+    }, {} as TokenInfoFromBalances)
+  }, [balances.items])
```
iamacook, line 137:
```suggestion
  const tokenInfoFromBalances = useMemo(() => {
    if (balances.items.length === 0) {
      return undefined
    }
    return balances.items.reduce<TokenInfoFromBalances>((dictionary, item) => {
      dictionary[item.tokenInfo.address] = {
        symbol: item.tokenInfo.symbol,
        decimals: item.tokenInfo.decimals,
        logoUri: item.tokenInfo.logoUri,
      }
      return dictionary
    }, {})
  }, [balances.items])
```

iamacook: We could alternatively create a selector that does this for us. What do you think?

iamacook (approved):

src/hooks/loadables/useLoadSpendingLimits.ts
```diff
@@ -19,18 +28,36 @@ const isModuleEnabled = (modules: string[], moduleAddress: string): boolean => {
 const discardZeroAllowance = (spendingLimit: SpendingLimitState): boolean =>
   !(sameString(spendingLimit.amount, '0') && sameString(spendingLimit.resetTimeMin, '0'))
 
+const getTokenInfoFromBalances = (tokenInfoFromBalances: TokenInfo[], address: string) =>
+  tokenInfoFromBalances.find((token) => token.address === address)
+
+const getTokenInfoOnChain = async (address: string) => {
```
katspaugh, line 34: Please add a return type to this and other functions you've added in this PR.

src/hooks/loadables/useLoadSpendingLimits.ts
```diff
@@ -19,18 +28,36 @@ const isModuleEnabled = (modules: string[], moduleAddress: string): boolean => {
...
+  const web3 = getWeb3()
+  if (!web3) {
+    return undefined
+  }
```
katspaugh, line 38: A bit shorter. :)
```suggestion
  if (!web3) return
```

src/hooks/loadables/useLoadSpendingLimits.ts
```diff
@@ -19,18 +28,36 @@ const isModuleEnabled = (modules: string[], moduleAddress: string): boolean => {
...
+
+  const erc20 = ERC20__factory.connect(address, web3)
+  return {
+    address,
+    symbol: await erc20.symbol(),
+    decimals: await erc20.decimals(),
```
katspaugh, line 44: This would be slightly faster (in parallel).
```
const [symbol, decimals] = await Promise.all([
  erc20.symbol(),
  erc20.decimals()
])
```

katspaugh (approved):

liliya-soroka:
1. Could we display a token symbol ?
2. spending limits page reloads each N sec

schmanu:
@liliya-soroka: I fixed both issues.

liliya-soroka:
Verified

PR #1652, opened 2023-02-09
Fix: limit spending limit tokens in menu

katspaugh:
## What it solves

Resolves https://github.com/safe-global/web-core/pull/1648#issuecomment-1423715767

Filtered balance `items` were defined but not used for rendering the menu.


iamacook (approved):

liliya-soroka:
Verified

PR #1657, opened 2023-02-09
Fix: spending limit max amount

### LINKED ISSUE #1656: Spending limits: no validation error for the amount when the available token balance < the allowed amount according to the spending limit
## Bug description
Spending limits: no validation error for the amount when the available token balance < the allowed amount according to the spending limit

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain:Goerli

## Steps to reproduce
 1. create a spending limit with the amount > the current token balance in the safe
 2. Click send funds and select spending limits 
 3. the current token amount <entered amount = the allowed according to spending limit amount 
 Current result: no error and the user can go to the next page
image

4. check if its possible to execute tx
Current result: Estimation can not be done 
image
image

## Expected result
the amount  validation should be done on the first step

katspaugh:
## What it solves

Resolves #1656

## How this PR fixes it

The validation function only checked if the amount is <= than the allowance, but didn't check it against the total amount of the token.


iamacook (approved):

liliya-soroka:
Verified, can be merged to RC

PR #1668, opened 2023-02-14
Feat: NFT preview

### LINKED ISSUE #1633: [NFT page] Show image previews
The OpenSea metadata format is now [supported by our backend](https://github.com/safe-global/safe-transaction-service/pull/1305/files), so many more image previews will be available.

However, the images are quite large so we shouldn't render them directly in the table.

We can show an image preview on hover or click, tho. Design needed.

katspaugh:
## What it solves

Resolves #1633

## How this PR fixes it
Adds a preview on click.

## Analytics changes
A new `Preview NFT` event should be triggered when clicking on an NFT row.

## Screenshots
Screenshot 2023-02-14 at 09 07 59


src/components/nfts/NftGrid/index.tsx
```diff
@@ -103,6 +95,19 @@ const NftGrid = ({ nfts, selectedNfts, isLoading, children, onSelect }: NftsTabl
...
+  const inactiveIcon = useMemo(
+    () => (
+      <Tooltip title="There's no preview for this NFT" placement="top" arrow>
+        <span>
+          <NftIndicator color="border" />
+        </span>
+      </Tooltip>
+    ),
+    [],
+  )
```
iamacook, line 109: We can move these outside the component and then not need memoise them.

src/components/nfts/NftPreviewModal/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <ModalDialog
+      open={!!nft?.imageUri}
+      onClose={onClose}
+      dialogTitle={title}
+      fullScreen
+      sx={{ margin: [0, 2], '.MuiPaper-root': { borderRadius: [0, '6px'] } }}
+    >
+      {nft && (
+        <div className={css.wrapper}>
+          <a href={nft.imageUri} target="_blank" rel="noreferrer" className={css.imageLink}>
```
iamacook, line 24:
```suggestion
          <a href={nft.imageUri} target="_blank" rel="noopener noreferrer" className={css.imageLink}>
```

DiogoSoaress (approved):
Looks great ✅ 
I would only like to see https://github.com/safe-global/web-core/pull/1668/files#r1105466829 addressed

francovenica:
Not a big fan of the full screen modal. Can't it be addapted to the size of the image?
Also is the link to the PNG something that should be there? 
image

katspaugh:
@liliiaorlenko what do you think about the questions Franco raised?

IMHO a fullscreen modal makes sense because them images are pretty big. I would maybe make it more informative tho, like Castle and Etherscan.
Reg. the direct link to the image file, I think it's useful if you want to download the jpeg you own. Alternatively, we could close the modal on image click.

katspaugh:
I've removed the image link and made it close the modal on click.

PR #1670, opened 2023-02-14
fix: Sticky z-index in assets and NFTs

DiogoSoaress:
## What it solves

Resolves https://github.com/safe-global/web-core/issues/1636

## How this PR fixes it
Reuse `<Sticky/>` in `TokenMenu`

## How to test it

1. Go to This safe that has a bunch of tokens in ETH eth:0x8675B754342754A30A2AeF474D114d8460bca19b
2. Open the menu to hide tokens
3. Scroll down a bit so the save button overlaps a checkbox
4. Click the Save button where the checkbox in the row behind it would be

### DIFF
--- a/src/components/balances/AssetsTable/index.tsx
+++ b/src/components/balances/AssetsTable/index.tsx
@@ -148,12 +148,7 @@ const AssetsTable = ({
                   </Track>
                 )}
                 {showHiddenAssets ? (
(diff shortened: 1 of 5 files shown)


katspaugh (approved):

katspaugh (approved):
👍

PR #1672, opened 2023-02-16
Fix: sort apps alphabetically

### LINKED ISSUE #1669: [Safe apps list] Order safe apps alphabetically
## Bug description
On the previous design for the safe apps main page order was performed alphabetically but it is not the case anymore for the new design. We should go back to that ordering logic.

## Steps to reproduce
Go to /apps screen and check app ordering

## Expected result
Safe apps available on each section (all apps, bookmarked apps & my custom apps) ordered alphabetically plus the same when filtering them by tags.

## Obtained result
Order not alphabetical.

katspaugh:
## What it solves

Resolves #1669

## How this PR fixes it

The apps are sorted on the client-side now (because they come unsorted from the backend).

I've also optimized the search a bit.

## How to test it
Load the safe apps page and make sure it's alphabetically ordered.
Filter by category, it should be still sorted.

## Analytics changes
None.


iamacook (approved):

francovenica:
By default is properly sorted
Using categories is also sorted fine

When I try to use the filter by text the sorting is not correct:
Try in mainnet https://fix_apps_sort--webcore.review-web-core.5afe.dev/apps?safe=eth:0x8675B754342754A30A2AeF474D114d8460bca19b
Search for " a "
image

katspaugh:
@francovenica that's expected, the search function sorts the result according to its own scoring algorithm.

PR #1674, opened 2023-02-16
Fix: Safe Apps loading state

### LINKED ISSUE #1671: Skeleton loaders around Safe
## Bug description
Pages that don't have a skeleton:
- Assets (Tokens and NFTs)
- Address Book (Not sure)
- Apps
- Settings (Not sure)

@liliya-soroka if you can help with further investigation of states thats needed skeletons
If design is needed let me know, but the skeletons thats already exist is great for the reference.

katspaugh:
## What it solves

Resolves #1671

## How this PR fixes it
I've created a separate route for individual Safe Apps: `/apps/open?appUrl=...`.
The Safe Apps list will remain on `/apps`.

This makes it easier to pre-render the route in an efficient way, and show placeholder skeletons.

## How to test it
Go to apps, open an app, reload the page.
Test custom apps.


src/pages/balances/nfts.tsx
```diff
@@ -12,18 +12,18 @@ import { useRemoteSafeApps } from '@/hooks/safe-apps/useRemoteSafeApps'
...
 
-  if (!nftApps?.length) {
+  if (nftApps?.length === 0) {
     return null
   }
 
   return (
     <Grid item sm={12} lg={3} order={{ lg: 1 }}>
-      <Typography component="h2" variant="subtitle1" fontWeight={700} mb={2} mt={0.6}>
+      <Typography component="h2" variant="subtitle1" fontWeight={700} mb={2} mt={0.75}>
```
DiogoSoaress, line 21: Rather put pixels here no?

DiogoSoaress (commented):
Tested and looks good. Only thing I 'd suggest to adjust is this skeleton in the `NftGrid` to match the checkbox width

src/pages/balances/nfts.tsx
```diff
@@ -23,11 +23,17 @@ const NftApps = memo(function NftApps(): ReactElement | null {
...
-            <SafeAppCard safeApp={nftSafeApp} />
+        {nftApps ? (
+          nftApps.map((nftSafeApp) => (
+            <Grid item lg={12} md={4} xs={6} key={nftSafeApp.id}>
+              <SafeAppCard safeApp={nftSafeApp} />
+            </Grid>
+          ))
+        ) : (
+          <Grid item lg={12} md={4} xs={6}>
+            <Skeleton variant="rounded" height="245px" />
```
DiogoSoaress, line 34: Knowing that we have 2 `NftApp` in mainnet I'd place 2 skeletons. Your call though

src/pages/balances/nfts.tsx
katspaugh: It will scale with the rest of the grid if change the base values.

src/pages/balances/nfts.tsx
katspaugh: I went with 1 because it aligns perfectly with 3 skeleton rows in the NFTs table. Also it's actually 3 apps that we have on prod.

src/pages/apps/index.tsx
```diff
@@ -50,11 +49,8 @@ const SafeApps: NextPage = () => {
     remoteSafeAppsLoading,
   })
 
-  if (!routerReady || isLoading) {
-    return null
-  }
-
-  if (appUrl) {
+  // Indidual Safe Apps live on the same URL but with a query param
```
iamacook, line 52:
```suggestion
  // Individual Safe Apps live on the same URL but with a query param
```

src/pages/apps/index.tsx
```diff
@@ -21,8 +21,8 @@ import { getOrigin } from '@/components/safe-apps/utils'
 const SafeApps: NextPage = () => {
   const chainId = useChainId()
   const router = useRouter()
-
-  const [appUrl, routerReady] = useSafeAppUrl()
```
iamacook, line 25: It seems like [we shouldn't have been using `router.isReady` for conditional rendering](https://nextjs.org/docs/api-reference/next/router#router-object) at all.

iamacook (commented):
Looks good to me but I think we should consider adjusting the placeholder of the NFT page. There's no table when the Safe has no NFTs.

nft skeleton

We keep the table when all tokens are hidden:

asset skeleton

katspaugh:
> Looks good to me but I think we should consider adjusting the placeholder of the NFT page. There's no table when the Safe has no NFTs.

That's why I originally had a circular loader there but now we optimize for the happy path, i.e. when you do have NFTs. One can argue that if a person deliberately visits that page, they likely have NFTs.

katspaugh:
Reg. showing an empty assets table, it probably was a bad idea to allow hiding the native token (I insisted on it), but it's an edge case.

katspaugh:
I accidentally merged this branch to dev prior to the last two commits.

PR #1676, opened 2023-02-16
Fix: avoid multiple pending txs with the same nonce

### LINKED ISSUE #1667: UI displaying wrong transaction nonce when quickly executing transactions
## Bug description

When executing a transaction (tested with a 1/n Safe) and quickly creating a new transaction, the UI will show the nonce of the previously executed transaction but submit the transaction with the correct nonce (or at least a nonce higher then is being displayed). 

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet and Görli

## Steps to reproduce
 
With a 1/n Safe create and execute multiple transactions and observe the nonce being displayed in the UI. 

## Expected result

The submitted and signed nonce should always match the displayed nonce in the UI.  

## Obtained result

see video

https://user-images.githubusercontent.com/93598450/218501568-b099bce8-6b67-48bf-8921-343a3563bebf.mp4
image (7)

katspaugh:
## What it solves

Resolves #1667

## How this PR fixes it

Disallows immediately executing a newly created tx w/o signatures if there are already other pending transactions.

## How to test
In a 1/X Safe, create and immediately execute a tx. Remember its nonce.
Quickly create another tx. Notice that it will have the same nonce as the previous tx, but the Execute option will be disabled.


src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -62,11 +63,12 @@ const SignOrExecuteForm = ({
   const isOwner = useIsSafeOwner()
   const provider = useWeb3()
   const currentChain = useCurrentChain()
+  const hasPending = useHasPendingTxs()
```
iamacook, line 66: We should also check that the pending transactions are on the current chain.

iamacook (approved):

francovenica:
Didn't worked for me. 
Created the a tx and executed it right away
Quickly created a new tx. This one had the same nonce as the one I was going to execute but with no "Execute now" checkbox in the modal. Queued it anyways
Waited for the first tx to finish to execute
The result is 2 tx with the same nonce, one in history and one in queue. The one in history is labeled as "untrusted" and theone in queue can still be executed (or at least you can attempt to)

Note: I had to speed up the 1st tx cuz it was taking a long time to execute, but that just the network being slow.
Note2: I had to refresh the history tab manually for the executed tx to show up

The safe:
https://fix_nonce--webcore.review-web-core.5afe.dev/transactions/queue?safe=gor:0x24F8542Da0C5f267c177a224270442C73804EE80
tx nonces

katspaugh:
That's expected. You can replace that tx.

francovenica:
After a while (1h +) the tx with nonce 150 that was in queue dissapeared on its own

katspaugh:
It disappeared because the other tx with the same nonce got indexed and replaced the queued one. So it's correct.

I've added a feature to display untrusted pending transactions on top of the regular queue. This should make things more transparent.

Screenshot 2023-02-22 at 09 34 08

Screenshot 2023-02-22 at 09 34 17

francovenica:
Well, I tried again in Poly ( goerli is pretty slow these days) and the I create 2 tx quickly. The first one is executing/indexing, and the 2nd one, with the same nonce as the 1st one, was cancelled right away

I didn't had the issue of having 2 tx with the same nonce in the history and queue again.

PR #1678, opened 2023-02-16
feat: add tracking to default token list menu

DiogoSoaress:
## What it solves
Adds tracking to selecting between `trusted` and `all` tokens list

Resolves #

## How this PR fixes it
Wraps the MenuItem with `Track` component

## How to test it
In the PR deployment you should see an event logged when switching between "TRUSTED" and "ALL" token lists

## Analytics changes
Adds tracking for:
- SHOW_DEFAULT_TOKENS
- SHOW_ALL_TOKENS


iamacook (approved):

Austindgk2:
> ## What it solves
> Adds tracking to selecting between `trusted` and `all` tokens list
> 
> Resolves #1679
> 
> ## How this PR fixes it
> Tracks opening/closing the token list menu Tracks the `TokenListSelect` menu items
> 
> ## How to test it
> In the PR deployment you should see an event logged when switching between "TRUSTED" and "ALL" token lists
> 
> ## Analytics changes
> Adds tracking for:
> 
> * TOKEN_LIST_MENU
> * SHOW_DEFAULT_TOKENS
> * SHOW_ALL_TOKENS

src/components/balances/TokenListSelect/index.tsx
```diff
@@ -26,6 +28,13 @@ const TokenListSelect = () => {
     dispatch(setTokenList(selectedString))
   }
 
+  const handleTrack = (label: 'Open' | 'Close') => {
```
usame-algan, line 31: Could we make these labels more specific? Otherwise there might be some overlaps in GA. Maybe something like "Open token list", "Close token list"

DiogoSoaress: I've changed it.

usame-algan (approved):
Thanks!

src/components/balances/TokenListSelect/index.tsx
```diff
@@ -15,6 +17,11 @@ const TokenListLabel = {
   [TOKEN_LISTS.ALL]: 'All tokens',
 }
 
+const enum TokenListActions {
+  OPEN = 'Open token list',
+  CLOSE = 'Close token list',
```
katspaugh, line 22: I would suggest making two separate objects in the analytics/events enum that would differ only by their label.
E.g.
```
OPEN_TOKEN_LIST_MENU: {
  ...
},
CLOSE_TOKEN_LIST_MENU: {
  ...
}
```

DiogoSoaress: Nice suggestion! I've addressed it.

src/components/balances/TokenListSelect/index.tsx
```diff
@@ -58,8 +54,8 @@ const TokenListSelect = () => {
           label="Tokenlist"
           onChange={handleSelectTokenList}
           renderValue={(value) => TokenListLabel[value]}
-          onOpen={() => handleTrack(TokenListActions.OPEN)}
-          onClose={() => handleTrack(TokenListActions.CLOSE)}
+          onOpen={() => handleTrack(ASSETS_EVENTS.OPEN_TOKEN_LIST_MENU)}
+          onClose={() => handleTrack(ASSETS_EVENTS.CLOSE_TOKEN_LIST_MENU)}
```
katspaugh, line 58: On a second thought, do we really want to track when it's closed? What insight can you get from this?

francovenica:
It works fine, but I agree with Ivan's comment: I don't think we need to track the closing of the menu. Everytime you choose an option you will track 2 events, the element you chose and the closing of the menu

The opening of the menu I think is fine, it will track if people are at least interested in the feature.

src/components/balances/TokenListSelect/index.tsx
DiogoSoaress: Valid point. I saw the `CURRENCY_MENU` was tracking both events so I applied the same criteria but it brings little value indeed.

katspaugh (approved):

PR #1680, opened 2023-02-16
Fix: remove claiming app from dashboard

katspaugh:
## What it solves

The dashboard has a whole section for governance, no need to push our app also in the lower Safe Apps section.


usame-algan (approved):
I didn't even know that we always showed the claiming app on position 1 🤔 Makes sense to remove it!

PR #1681, opened 2023-02-17
fix: spending limit validation

### LINKED ISSUE #1677: Owner with spending limit cannot spend more than the limit set even for multi-sig transactions
What happened?
An owner wallet who has a spending limit set to 10,000 USDC can only spend 10,000 USDC tokens even if set to multi-sig.
Users should still be able to spend more than the spending limit set if multisig is selected prior to initiating transaction.

image
Steps:
1. there is a token B in the same with the amount 100
2. set up spending limits for the owner with token B and amount =50
3.  click send funds
4. select multisig option
5. enter 51 as an amount for send funds
Current result: the error is displayed for the multisig option
Expected result: the owner should be able to spend more then spending limit if the multisig option is selected

Chrome/prod 1.5.0

iamacook:
## What it solves

Resolves #1677

## How this PR fixes it

The multisig option no longer validates according to a spending limit. When changing the transaction type, the amount is now again validated as well.

## How to test it

Create a transaction and select "Multisig Transaction" when a spending limit is set. Observe no validation error when entering an amount above the spending limit. Switching to a "Spending Limit Transaction" should validate and show an error.

## Screenshots

spending limit


katspaugh (approved):
My bug! Thanks for rectifying it.

usame-algan (approved):
Looks good 👍

francovenica:
Looks good to me

The limit now responds to the choice the user makes in the modal
The max button works fine depending on your choice
Changing your choice after the amout is set is being validated as well 
I tried changing the spending limit, and even disabling the module. it works fine

PR #1688, opened 2023-02-21
test: add pending actions E2E

DiogoSoaress:
## What it solves
Adds E2E coverage to the pending actions feature.

## How this PR fixes it
Adds a E2E test suite for the pending actions happy paths.

## How to test it
Run the `pending_actions` test from the cypress panel

### DIFF
--- a/cypress/e2e/smoke/pending_actions.cy.js
+++ b/cypress/e2e/smoke/pending_actions.cy.js
@@ -59,7 +59,7 @@ describe('Pending actions', () => {
     cy.contains('h3', 'Transactions').should('be.visible')
 
     // contains 3 queued transactions
-    cy.get('span:contains("out of 2")').should('have.length', 3)
+    cy.get('span:contains("1 out of 2")').should('have.length', 3)
 
     // Ensure wallet is connected
     cy.contains('E2E Wallet @ Görli')
--- a/cypress/e2e/smoke/pending_actions.cy.js
+++ b/cypress/e2e/smoke/pending_actions.cy.js
@@ -10,7 +10,7 @@ describe('Pending actions', () => {
     cy.contains('button', 'Accept all').click()
 
     // Ensure wallet is connected to correct chain via header
-    cy.contains('E2E Wallet @ Görli')
+    cy.contains('MetaMask @ Görli')
   })
 
   it('should add the Safe with the pending actions', () => {
--- a/cypress/e2e/smoke/pending_actions.cy.js
+++ b/cypress/e2e/smoke/pending_actions.cy.js
@@ -1,0 +1,50 @@
+const SAFE = 'gor:0xCD4FddB8FfA90012DFE11eD4bf258861204FeEAE'
+
+describe('Pending actions', () => {
+  it('should navigate to the form', () => {
+    cy.connectE2EWallet()
+
+    cy.visit('/welcome')
+
+    // Close cookie banner
+    cy.contains('button', 'Accept all').click()
+
+    // Ensure wallet is connected to correct chain via header
+    cy.contains('E2E Wallet @ Görli')
+  })
+
+  it('should add the Safe with the pending actions', () => {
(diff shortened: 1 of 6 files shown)


DiogoSoaress:
I've opened this PR so I could verify if the job runs successfully with the CI account.

cypress/e2e/smoke/pending_actions.cy.js
```diff
@@ -0,0 +1,63 @@
...
+  })
+
+  it('should have the right number of queued and signable transactions', () => {
+    // Navigates to the tx queue
+    cy.contains('h3', 'Transactions').should('be.visible')
+
+    // contains 3 queued transactions
+    cy.get('span:contains("out of 2")').should('have.length', 3)
+
+    // TODO: This is a workaround to ensure the wallet is connected after page navigation
```
DiogoSoaress, line 54: The command we have (`connectE2EWallet`) was not taking effect after the page navigation and I didn't find a better way than setting the localStorage manually.
Any hints or help is very welcome.

cypress/e2e/smoke/pending_actions.cy.js
```diff
@@ -0,0 +1,63 @@
...
+    window.localStorage.setItem('SAFE_v2__lastWallet', JSON.stringify('E2E Wallet'))
```
DiogoSoaress, line 55: The command we have (`connectE2EWallet`) was not taking effect after the page navigation and I didn't find a better way than setting the localStorage manually.
Any hints or help is very welcome.

cypress/e2e/smoke/pending_actions.cy.js
```diff
@@ -0,0 +1,68 @@
...
+    cy.contains('E2E Wallet @ Görli')
+  })
+
+  beforeEach(() => {
+    cy.restoreLocalStorageCache()
+  })
+
+  afterEach(() => {
+    cy.saveLocalStorageCache()
+  })
```
iamacook, line 22: Can you add a comment why this is necessary?

cypress/e2e/smoke/pending_actions.cy.js
```diff
@@ -0,0 +1,68 @@
...
+      cy.get('[data-testid=CheckIcon]').next().click()
+    })
+  })
+
+  it('should have the right number of queued and signable transactions', () => {
+    // Navigates to the tx queue
+    cy.contains('h3', 'Transactions').should('be.visible')
+
+    // contains 3 queued transactions
+    cy.get('span:contains("out of 2")').should('have.length', 3)
```
iamacook, line 60: I think we should bee explicit here.

```suggestion
    cy.get('span:contains("1 out of 2")').should('have.length', 3)
```

DiogoSoaress: I left that part out of the assertion because it brings complexity to assert the queue size. For example if one transaction has `1 out of 2` and the next one has `2 out of 2`.

cypress/e2e/smoke/pending_actions.cy.js
DiogoSoaress: WDYT?
```suggestion
  beforeEach(() => {
    // Uses the previously saved local storage
    // to preserve the wallet connection between tests
    cy.restoreLocalStorageCache()
  })

  afterEach(() => {
    cy.saveLocalStorageCache()
  })
```

iamacook: Looks good to me.

cypress/e2e/smoke/pending_actions.cy.js
iamacook: All [queued transactions in that Safe](https://safe-web-core.staging.5afe.dev/transactions/queue?safe=gor:0xCD4FddB8FfA90012DFE11eD4bf258861204FeEAE) only have one signature each though.

katspaugh (approved):

PR #1692, opened 2023-02-23
Refactor: minimize the usage of the static chains config

katspaugh:
## What it solves

New chains are popping up like mushrooms and the static chainId<->shortName config becomes a hurdle when adding new chain support via the config service.

I removed all the places where looking up a shortName in the static config was blocking the UI if the chain isn't there.

`useChainId` still relies on the static config to resolve a shortName from the URL faster, but it will now also look it up in the remote config. This means that loading a page on a chain that isn't in the config will redirect to the default chain (because the remote chains wouldn't be loaded instantly), but navigating to that chain via the menu will be still possible.

## How to test
Test opening different chains, and also reloading the page on a given chain.


src/components/common/AddressInput/index.test.tsx
```diff
@@ -82,13 +82,6 @@ describe('AddressInput tests', () => {
   it('should validate the address on input', async () => {
     const { input, utils } = setup('')
 
-    act(() => {
-      fireEvent.change(input, { target: { value: `xyz:${TEST_ADDRESS_A}` } })
-      jest.advanceTimersByTime(1000)
-    })
-
-    await waitFor(() => expect(utils.getByLabelText('Invalid chain prefix "xyz"', { exact: false })).toBeDefined())
```
iamacook, line 90: Allowing invalid prefixes somewhat negates ERC-3770. Can we validate against the prefixes currently in the store?

src/utils/chains.ts
```diff
@@ -26,11 +25,3 @@ export const getBlockExplorerLink = (
...
   }
 }
-
-export const getShortName = (chainId: string): string | undefined => {
-  return Object.keys(chains).find((prefix) => chains[prefix] === chainId)
-}
-
-export const getChainId = (shortName: string): string | undefined => {
-  return chains[shortName]
-}
```
iamacook, line 36: This logic is still used in a number of places. I think we should keep it/its tests.

src/components/common/AddressInput/index.test.tsx
katspaugh: It still checks that the prefix matches the current chain, that's more than enough. No need to check it against all possible chains.

src/utils/chains.ts
katspaugh: It's only used in one place (useChainId), inline makes more sense to me.

sgilflores3 (approved):

katspaugh:
@iamacook I've also removed our config and use that from safe-core-sdk-utils instead. It's better because it's now consistent with the Core SDK's chain support which we depend on already.

src/components/common/NetworkSelector/index.tsx
```diff
@@ -33,38 +34,45 @@ const NetworkSelector = (): ReactElement => {
...
-      },
-    }
+    const newChainId = event.target.value
+    const shortName = configs.find((item) => item.chainId === newChainId)?.shortName
 
-    if (router.query?.safeViewRedirectURL) {
-      newRoute.query.safeViewRedirectURL = router.query?.safeViewRedirectURL.toString()
+    if (shortName) {
+      trackEvent({ ...OVERVIEW_EVENTS.SWITCH_NETWORK, label: newChainId })
+      router.push(getNetworkLink(shortName))
```
iamacook, line 68: Now that each `MenuItem` has a `Link`, you don't need to `push` here as well.

src/config/chains.ts
```diff
@@ -1,343 +1,20 @@
...
+ *   gor: '5',
+ *   ...
+ * }
+ */
+type Chains = Record<string, string>
+
+const chains: Chains = networks.reduce((result, { shortName, chainId }) => {
+  result[shortName] = chainId.toString()
+  return result
+}, {} as Chains)
```
iamacook, line 18: Nit, but if you specify the type as a generic, you need not cast.

```suggestion
const chains = networks.reduce<Chains>((result, { shortName, chainId }) => {
  result[shortName] = chainId.toString()
  return result
}, {})
```

src/components/common/WalletInfo/index.tsx
```diff
@@ -1,24 +1,20 @@
...
 
 export const UNKNOWN_CHAIN_NAME = 'Unknown'
 
 const WalletInfo = ({ wallet }: { wallet: ConnectedWallet }): ReactElement => {
   const walletChain = useAppSelector((state) => selectChainById(state, wallet.chainId))
-
-  const prefix = useMemo(() => {
-    return walletChain?.shortName || getShortName(wallet.chainId)
-  }, [walletChain?.shortName, wallet.chainId])
+  const prefix = walletChain?.shortName
```
iamacook, line 17: Here's another instance related to showing prefixes of chains not in the configs.

src/components/common/ConnectWallet/AccountCenter.tsx
```diff
@@ -14,17 +13,13 @@ import ChainSwitcher from '../ChainSwitcher'
...
 const AccountCenter = ({ wallet }: { wallet: ConnectedWallet }) => {
   const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>(null)
   const onboard = useOnboard()
   const chainInfo = useAppSelector((state) => selectChainById(state, wallet.chainId))
   const addressBook = useAddressBook()
-
-  const prefix = useMemo(() => {
-    return chainInfo?.shortName || (wallet?.chainId && getShortName(wallet?.chainId))
-  }, [chainInfo?.shortName, wallet?.chainId])
+  const prefix = chainInfo?.shortName
```
iamacook, line 22: Do we not want to show the prefixes of networks not in the chain configs but in `chains`? See [here](https://github.com/safe-global/web-core/pull/1606#issuecomment-1410229573).

src/components/common/NetworkSelector/index.tsx
katspaugh: Sadly it doesn't work if you use the keyboard. So the link will only work if you open it in a new tab.

src/config/chains.ts
katspaugh: Neat!

src/components/common/WalletInfo/index.tsx
katspaugh: Not sure what you mean. If the walletChain isn't found, it won't show any prefix, and that's fine.

src/components/common/ConnectWallet/AccountCenter.tsx
katspaugh: It will just an address w/o any prefix, it's fine.

src/components/common/NetworkSelector/index.tsx
iamacook: Intriguing. Just double-checked and it doesn't add two entries to the history 👍

src/components/common/WalletInfo/index.tsx
iamacook: Previously we were always showing them if they were present in `chains` but admittedly that doesn't make sense if it is an "unknown" network.

iamacook (approved):

francovenica:
LGTM

Being switching networks around, switching between safes, checking the switching the network in the safe creation, checking the "switch to x network" button when the wallet is in the wrong network.

PR #1693, opened 2023-02-23
Fix: remove unnecessary safe contract fallback

katspaugh:
## What it solves
Falling back to default contract addresses was a leftover from safe-react.
Safe Core SDK will fail to initialize if a fallback handler is missing in safe-deployments for a particular chain, so the app won't work anyway. We can thus remove the fallbacks and assume that all supported chains are present in safe-deployments.

## How to test
This should affect only Safe creation.
Test it on a couple different chains.

### DIFF
--- a/src/services/contracts/safeContracts.ts
+++ b/src/services/contracts/safeContracts.ts
@@ -136,15 +136,10 @@ export const getMultiSendCallOnlyContractInstance = (
 // GnosisSafeProxyFactory
 
 const getProxyFactoryContractDeployment = (chainId: string) => {
-  return (
-    getProxyFactoryDeployment({
-      version: LATEST_SAFE_VERSION,
-      network: chainId,
-    }) ||
-    getProxyFactoryDeployment({
-      version: LATEST_SAFE_VERSION,
-    })
-  )
+  return getProxyFactoryDeployment({
+    version: LATEST_SAFE_VERSION,
+    network: chainId,
+  })
 }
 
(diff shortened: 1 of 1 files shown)


iamacook (approved):

usame-algan (approved):

francovenica:
I had an issue that happened twice, once in Goerli and one in Eth. ( I didn't had this issue in Polygon tho)
I created a safe and during the creation it got stucked in this state
image

I still used the safe but it won't load ever. I've waited for over an hour and it still won't load. 
The safes load fine in prod, so the safe were created successfully, but they won't work in this PR.

Links:
https://fallback--webcore.review-web-core.5afe.dev/home?safe=gor:0x77f9Bb99b8324Cff0d83da73d1B65fE4E0E8eE50
Block explorer:
https://goerli.etherscan.io/tx/0xdfe14f80a0664d102145cea6d5c03c2bd9e940c01cd17ede16b5f554994ce9fe

https://fallback--webcore.review-web-core.5afe.dev/home?safe=eth:0xb3f964e79EE3b4f8b1dFE69683F50B2426460032
Block explorer:
https://etherscan.io/tx/0xfec5f41edd5709ae5417cf75835280eeaf0c2449601314b39d77f66c35ee4149

katspaugh:
It works for me with the prod cgw toggle on.

katspaugh:
I'm going to close this since it's just a very minor optimization and I'm not 100% certain it doesn't break any edge cases.

PR #1694, opened 2023-02-23
Fix: safe-deployments update

katspaugh:
A fix for Base.


liliya-soroka:
An error in the console during safe creation with MM : "EthersAdapter.js:112 Uncaught (in promise) Error: Invalid CompatibilityFallbackHandler contract address"

JagoFigueroa (approved):

PR #1699, opened 2023-02-24
Add Metis Goerli testnet

t0mcr8se:
## What it solves:
Error 104: Invalid chain short name in the URL
This happens when opening safes prefixed with metis-goerli

https://chainlist.org/chain/599
https://github.com/safe-global/safe-deployments/pull/161


t0mcr8se:

katspaugh:
This config will be gone in #1692. Please see the reasoning there.
If the fix in this PR helps, your stack should probably also work on the branch from #1692.

PR #1709, opened 2023-02-28
Fix: single transaction skeleton

### LINKED ISSUE #1671: Skeleton loaders around Safe

katspaugh:
## What it solves

Resolves #1671

The final piece of #1671 – a skeleton for the single tx page.

Screenshot 2023-02-28 at 12 21 47

## How this PR fixes it
I've replaced the circular loader with a skeleton in an expanded accordion.
Also changed the links from the dashboard queue widget to open individual txs instead of the entire queue. "View all" still links to the queue.

### DIFF
--- a/src/components/transactions/TxListItem/ExpandableTransactionItem.tsx
+++ b/src/components/transactions/TxListItem/ExpandableTransactionItem.tsx
@@ -1,5 +1,5 @@
 import { type Transaction, type TransactionDetails } from '@safe-global/safe-gateway-typescript-sdk'
-import { Accordion, AccordionDetails, AccordionSummary } from '@mui/material'
+import { Accordion, AccordionDetails, AccordionSummary, Box, Skeleton } from '@mui/material'
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
 import TxSummary from '@/components/transactions/TxSummary'
 import TxDetails from '@/components/transactions/TxDetails'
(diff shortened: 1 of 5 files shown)


src/components/transactions/TxListItem/ExpandableTransactionItem.tsx
```diff
@@ -53,4 +53,22 @@ export const ExpandableTransactionItem = ({
   )
 }
 
+export const TransactionSkeleton = () => (
+  <>
+    <Box pt="20px" pb="4px">
+      <Skeleton variant="text" width="35px" />
+    </Box>
```
DiogoSoaress, line 60: This Skeleton is unnecessary in the Single Tx because there is no label

DiogoSoaress (approved):
Works well!
Left one comment though

src/components/transactions/TxListItem/ExpandableTransactionItem.tsx
katspaugh: For queued transactions, there's always a label. I guess it's arbitrary if we want to have it or not.

katspaugh:
Tested again, works as expected. ✅

PR #1710, opened 2023-02-28
Chore: remove an unnecessary deployment step

katspaugh:
## What it solves

The build is uploaded to S3 to a new folder every time we deploy a release, so it makes no sense trying to keep old files while uploading the new ones.

Only the dev and staging buckets use the same folder across deployments, but we don't care about downtime there.


usame-algan (approved):
Makes sense!

PR #1711, opened 2023-02-28
fix: Display manual gasLimit in GasParams overview

### LINKED ISSUE #1661: Can't set up gaslimit manually in case of "This transaction will most likely fail"
## Bug description
Can't set up gaslimit manually in case of "This transaction will most likely fail"
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Goerli

## Steps to reproduce 
 1. Try to execute the tx that will fail with GS013 ( I have tried to send NFT from my test safe to 1.0.0 safe )
 2. open advanced estimation parameter
 Current result: Gas limit - Cannot estimate. 

3. edit gas limit and save edited data 
4. reopen estimation parameters
Current result: Gas limit - Cannot estimate. 

## Expected result
It should be possible to enter gaslimit manually 
## Screenshots
image

usame-algan:
## What it solves

Resolves #1661 

## How this PR fixes it

- Displays the manual `gasLimit` in the `GasParams` overview if it exists

## How to test it

1. Open the Safe
2. Create a transaction where gas limit estimation would fail (e.g. send an NFT to a 1.0.0 safe)
3. Observe Estimated Fee showing "Cannot estimate"
4. Edit the gas limit and save
5. Observe seeing an Estimated Fee and the manual gas limit

## Screenshots
Screenshot 2023-02-28 at 14 08 50
Screenshot 2023-02-28 at 14 09 42

### DIFF
--- a/src/components/tx/GasParams/index.tsx
+++ b/src/components/tx/GasParams/index.tsx
@@ -34,6 +34,7 @@ const GasParams = ({ params, isExecution, isEIP1559, onEdit, gasLimitError }: Ga
 
   const chain = useCurrentChain()
   const isLoading = !gasLimit || !maxFeePerGas
+  const isError = gasLimitError && !gasLimit
 
   // Total gas cost
   const totalFee = !isLoading
@@ -56,7 +57,7 @@ const GasParams = ({ params, isExecution, isEIP1559, onEdit, gasLimitError }: Ga
         {isExecution ? (
           <Typography display="flex" alignItems="center" justifyContent="space-between" width={1}>
             <span>Estimated fee </span>
-            {gasLimitError ? null : isLoading ? (
+            {isError ? null : isLoading ? (
               <Skeleton variant="text" sx={{ display: 'inline-block', minWidth: '7em' }} />
             ) : (
               <span>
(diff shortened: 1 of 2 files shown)


src/components/tx/GasParams/index.tsx
```diff
@@ -34,6 +34,7 @@ const GasParams = ({ params, isExecution, isEIP1559, onEdit, gasLimitError }: Ga
 
   const chain = useCurrentChain()
   const isLoading = !gasLimit || !maxFeePerGas
+  const isError = gasLimitError && !gasLimit
```
usame-algan, line 37: Not sure if we need `gasLimitError` or could pass the loading flag instead and do `const isError = !gasLimitLoading && !gasLimit`

katspaugh (commented):
Could you explain a little bit what the root cause of the bug was and how you fixed it?

usame-algan:
> Could you explain a little bit what the root cause of the bug was and how you fixed it?

Changing the `gasLimit` is possible in any case but it wasn't reflected in the UI because the condition was only checking if there is a gas limit error but not if the `gasLimit` was manually changed.

katspaugh (approved):
I get it now, thanks!

iamacook (approved):

francovenica:
Looks good. The value now reflects in the UI. 
I also checked that the circular arrow that reset the estimated gas limit is not present (since there was no estimation duh)

PR #1712, opened 2023-02-28
Feat: better UX for disabld buttons

### LINKED ISSUE #1708: Better indication of why certain actions are disabled
In some situations we disable/hide transaction-related buttons:

* when the wallet isn't connected
* when the wallet is on the wrong chain
* when the wallet isn't an owner/spending limit beneficiary

## Problems
1. In some places in the UI, we just disable transaction buttons, in some we hide them completely. This makes the UI inconsistent.
2. There's no explanation why those buttons are disabled.
3. There's no way for the user to know how to fix that.

## The list of affected buttons
* The New transaction button in the sidebar
* Add owner button (currently hidden completely!)
* The final Submit button in the tx review modal
* The Send button in the Assets
* Possibly many more buttons throughout the app

## Solution
1. Always show all actionable buttons but disable them if the action cannot be performed.
2. Show a tooltip near the button explaining why
3. ~In the tooltip, show a button "Connect wallet" or "Switch chain"~

Screenshot 2023-02-28 at 08 52 51

katspaugh:
## What it solves

Resolves #1708

## How this PR fixes it

Creates a new HoC called `<CheckWallet>` that wraps all the buttons that depend on the wallet connection/being an owner.

## Todo
* [ ] Unit tests


src/components/tx/SpendingLimitRow/index.tsx
```diff
@@ -28,25 +28,28 @@ const SpendingLimitRow = ({
           name={SendAssetsField.type}
           render={({ field: { onChange, ...field } }) => (
             <RadioGroup
-              onChange={async (e) => {
+              onChange={(e) => {
+                onChange(e)
+
```
iamacook, line 33: Did `Promise.resolve()` not work?

src/components/settings/SpendingLimits/SpendingLimitsTable.tsx
```diff
@@ -88,9 +85,9 @@ export const SpendingLimitsTable = ({
       { id: 'beneficiary', label: 'Beneficiary' },
       { id: 'spent', label: 'Spent' },
       { id: 'resetTime', label: 'Reset time' },
-      { id: 'actions', label: 'Actions', sticky: true, hide: shouldHideactions },
```
iamacook, line 91: `hide` can now be removed from `EnhancedCell['hide']` as well as the associated CSS.

src/components/common/CheckWallet/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+type CheckWalletProps = {
+  children: (ok: boolean) => ReactElement
+  allowSpendingLimit?: boolean
+  allowNonOwner?: boolean
+}
+
+enum Message {
+  WalletNotConnected = 'Please connect your wallet',
+  NotSafeOwner = "Your connected wallet isn't a Safe owner",
+  OnlySpendingLimitBeneficiary = 'You can only create ERC20 transactions within the spending limit',
```
iamacook, line 18:
```suggestion
  OnlySpendingLimitBeneficiary = 'You can only create ERC-20 transactions within your spending limit',
```

src/components/tx/SpendingLimitRow/index.tsx
katspaugh: I replaced it with a timeout because it's more explicit/idiomatic. Waiting for a promise that does nothing is not self-explanatory.

katspaugh:
@iamacook I've just realized switching the chain isn't required for most actions. Signing a transaction can be done independent on the RPC your wallet is connected to. Only execution needs to be on the same chain.
So I removed that condition, and now the only button that is disabled when on the wrong chain is Submit in the execute modal.

schmanu:
If I am connected to the wrong chain the create Transaction modals are all stuck in the Review step. It doesn't break anything but if we want to allow signing we would need changes there too.
Screenshot 2023-03-01 at 14 05 11

katspaugh:
@schmanu thanks, I'll look into this. 👍

katspaugh:
@schmanu turns out we always init the SDK with the wallet provider, instead of the app RPC, so this signing on a wrong chain won't work out of the box. I'll make another PR to enable that.

src/components/settings/SpendingLimits/NewSpendingLimit/index.tsx
```diff
@@ -30,11 +31,16 @@ export const NewSpendingLimit = () => {
...
+      <CheckWallet>
+        {(isOk) => (
+          <Track {...SETTINGS_EVENTS.SPENDING_LIMIT.NEW_LIMIT}>
+            <Button onClick={() => setOpen(true)} sx={{ marginTop: 2 }} variant="contained" disabled={!isOk}>
+              New spending limit
+            </Button>
+          </Track>
+        )}
+      </CheckWallet>
+
```
usame-algan, line 43: The tooltip position looks off because the `Track` wrapper takes up the full width
Screenshot 2023-03-02 at 10 42 55

src/components/settings/owner/AddOwnerDialog/index.tsx
```diff
@@ -39,17 +40,21 @@ export const AddOwnerDialog = () => {
...
-            startIcon={<SvgIcon component={AddIcon} inheritViewBox fontSize="small" />}
-          >
-            Add new owner
-          </Button>
-        </Track>
-      </div>
+      <CheckWallet>
+        {(isOk) => (
+          <Track {...SETTINGS_EVENTS.SETUP.ADD_OWNER}>
+            <Button
```
usame-algan, line 46: Same here with the Tooltip position. Maybe we should give the `Track` wrapper `display: inline-block`
Screenshot 2023-03-02 at 10 44 18

katspaugh: I made it a span by default.

francovenica:
Checked:
New Tx button on the sidebar
Send button in Assets 
Send button in Address Book
Add/replace/remove owner buttons
Change policies
Delete a module
Create a spending limit
Remove a spending limit
Updating the safe
"Send again" option in the tx history

All these with the 4 states: not connected, connected with non owner, connected with owner in the wrong chain, connected with non-owner with an allowance.
Looks good

If you want to be really picky. This happens when your asked to sign the safe creation with your wallet and Cancel instead, and then disconnect. The retry button disables itself, but the message it gives is that you are in the wrong chain instead of not being connected
image

katspaugh:
Oh, I didn’t touch Safe creation at all in this PR, it’s using its own logic.
Should’ve mentioned that.

francovenica:
Ok, it is a really nish scenario anyways.
LGTM

Issue #1714, opened 2023-02-28
Mixed themes when loading the page in dark mode


katspaugh:
If you disable JS and load the balances page, you can see how the page looks before the client-side hydration kicks in.

<img width="1246" alt="Screenshot 2023-02-28 at 20 43 34" src="https://user-images.githubusercontent.com/381895/221961849-8030654f-b0ab-4463-8422-5392f30cea81.png">

Some elements (table, icons) are in the light theme, but others (background, logo) are in the dark theme.

PR #1715, opened 2023-02-28
Fix: pre-render the nav sidebar

katspaugh:
## What it solves

Partly #1714

## How this PR fixes it

The sidebar was previously rendered conditionally in two variants: for `?safe` routes and for non-safe routes.
Since we now don't show the sidebar for pages like Welcome, Safe Creation etc, there's no need for this variance, and we can always render the nav sidebar for a better static site performance.

## How to test it
Disable JS and load the Balances page of any Safe. The sidebar should appear much faster.

### DIFF
--- a/src/components/sidebar/Sidebar/index.tsx
+++ b/src/components/sidebar/Sidebar/index.tsx
@@ -1,7 +1,6 @@
-import { useState, type ReactElement } from 'react'
-import { Divider, Drawer, IconButton } from '@mui/material'
+import { useCallback, useState, type ReactElement } from 'react'
+import { Box, Divider, Drawer, IconButton } from '@mui/material'
 import ChevronRight from '@mui/icons-material/ChevronRight'
-import { useRouter } from 'next/router'
 
 import ChainIndicator from '@/components/common/ChainIndicator'
 import SidebarHeader from '@/components/sidebar/SidebarHeader'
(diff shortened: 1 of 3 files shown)


src/components/sidebar/Sidebar/index.tsx
```diff
@@ -11,54 +10,49 @@ import SidebarFooter from '@/components/sidebar/SidebarFooter'
...
-  const onDrawerToggle = () => {
-    trackEvent({ ...OVERVIEW_EVENTS.SIDEBAR, label: isDrawerOpen ? 'Close' : 'Open' })
-    setIsDrawerOpen((prev) => !prev)
-  }
+  const onDrawerToggle = useCallback(() => {
+    setIsDrawerOpen((isOpen) => {
+      trackEvent({ ...OVERVIEW_EVENTS.SIDEBAR, label: isOpen ? 'Close' : 'Open' })
+
+      return !isOpen
+    })
```
iamacook, line 22: Why was this change necessary?

iamacook, line 22: Why was this change necessary?

iamacook (approved):

src/components/sidebar/Sidebar/index.tsx
katspaugh: To not create a new callback reference on each render.

usame-algan (approved):

katspaugh:
I've tested it and since it doesn't touch any logic, going to merge it directly.

PR #1716, opened 2023-03-01
Fix: apps sdk widget design

### LINKED ISSUE #1665: [Safe apps list] Design changes to how to build safe app component
## Bug description
We need to update the current design for this component so that it matches the design specification as seen in [figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=2231%3A60797). This change will affect both desktop and mobile views and we might not want to have the expanded view on mobile mode by default.

Note: there is currently an issue on mobile view mode with this component, making it hard to click we might want to fis this along this design change.

## Steps to reproduce
Go to the /apps section on the web interface

## Expected result
How to build safe app? text always visible until scrolling is done and switches to the 'reduced' view we currently have by default.
Screenshot 2023-02-13 at 13 31 29

## Obtained result
'reduced' view set as default when it should only be shown when scrolling.
Screenshot 2023-02-13 at 13 31 40

katspaugh:
## What it solves

Resolves #1665

## How this PR fixes it

* I made it open on hover as per design
* Initially, it will show "How to build on Safe"
* If scrolled down a bit, it will minimize itself to not overlap with the page content

Screenshot 2023-03-01 at 09 32 26

### DIFF
--- a/src/components/safe-apps/SafeAppsSDKLink/index.tsx
+++ b/src/components/safe-apps/SafeAppsSDKLink/index.tsx
@@ -1,49 +1,43 @@
-import { useState } from 'react'
-import Link from 'next/link'
-import Box from '@mui/material/Box'
-import Collapse from '@mui/material/Collapse'
-import Typography from '@mui/material/Typography'
-import MUILink from '@mui/material/Link'
-import Fab from '@mui/material/Fab'
+import { useEffect, useState } from 'react'
+import { Fab, Typography } from '@mui/material'
 import KeyboardDoubleArrowUpRoundedIcon from '@mui/icons-material/KeyboardDoubleArrowUpRounded'
-
+import classnames from 'classnames'
 import CodeIcon from '@/public/images/apps/code-icon.svg'
 import { SAFE_APPS_SDK_DOCS_URL } from '@/config/constants'
 import css from './styles.module.css'
+import ExternalLink from '@/components/common/ExternalLink'
 
 const SafeAppsSDKLink = () => {
-  const [openModal, setOpenModal] = useState<boolean>(false)
+  const [mini, setMini] = useState(false)
+
+  useEffect(() => {
+    const minScroll = 130
+    const onScroll = () => {
+      const isScrolled = document.documentElement.scrollTop > minScroll
+      setMini(isScrolled)
+    }
+
+    document.addEventListener('scroll', onScroll)
+
+    return () => document.removeEventListener('scroll', onScroll)
+  }, [])
 
   return (
-    <Box className={css.position} onMouseEnter={() => setOpenModal(true)} onMouseLeave={() => setOpenModal(false)}>
-      <Collapse in={openModal}>
-        {
-          <Box className={css.container}>
-            <CodeIcon />
+    <div className={classnames(css.container, { [css.mini]: mini })}>
+      <CodeIcon />
 
-            <Typography variant="h6" className={css.title}>
-              How to build on Safe?
-            </Typography>
+      <Typography variant="h6" className={css.title}>
+        How to build on Safe?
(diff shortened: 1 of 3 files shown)


src/components/safe-apps/SafeAppsSDKLink/index.tsx
```diff
@@ -1,49 +1,43 @@
...
 import { SAFE_APPS_SDK_DOCS_URL } from '@/config/constants'
 import css from './styles.module.css'
+import ExternalLink from '@/components/common/ExternalLink'
 
 const SafeAppsSDKLink = () => {
-  const [openModal, setOpenModal] = useState<boolean>(false)
+  const [mini, setMini] = useState(false)
+
+  useEffect(() => {
+    const minScroll = 130
```
iamacook, line 14: Nit: can we align this with constant styles?

```suggestion
    const MIN_SCROLL = 130
```

src/components/safe-apps/SafeAppsSDKLink/index.tsx
```diff
@@ -1,49 +1,43 @@
...
-
+import classnames from 'classnames'
 import CodeIcon from '@/public/images/apps/code-icon.svg'
 import { SAFE_APPS_SDK_DOCS_URL } from '@/config/constants'
 import css from './styles.module.css'
+import ExternalLink from '@/components/common/ExternalLink'
 
 const SafeAppsSDKLink = () => {
-  const [openModal, setOpenModal] = useState<boolean>(false)
+  const [mini, setMini] = useState(false)
```
iamacook, line 11: This threw me off initially. Can we name it `isMini` instead?

iamacook (commented):
Looks good but it overlaps the title on smaller resolutions and it's not possible to fully close it.

image

katspaugh:
@iamacook fixed, thanks!

iamacook (approved):

src/components/safe-apps/SafeAppsSDKLink/index.tsx
```diff
@@ -1,49 +1,45 @@
...
+
+      <Typography variant="h6" className={css.title}>
+        How to build on Safe?
+      </Typography>
+
+      <ExternalLink href={SAFE_APPS_SDK_DOCS_URL} className={css.link} noIcon variant="body2">
+        <span>Learn more about Safe Apps SDK</span>
+      </ExternalLink>
+
+      <Fab className={css.openButton} variant="extended" size="small" color="secondary">
```
usame-algan, line 39: Should we change this to something thats not a button? Currently its still clickable but nothing happens. Could be an issue for keyboard navigation and screen reader users.

katspaugh: Good point, I've actually made the button clickable now.

katspaugh:
Fixed, it will now close on click anywhere outside of the widget (not on the repeated click on the arrow button though).

You can also select the widget with the keyboard (<kbd>Tab</kbd>), in which case it becomes focused and expanded as well.

francovenica:
Not sure about the "clicking outside" part. In this snpashot I marked the only place where you can "click outside" without clicking something else:
image

Regarding the TAB button it works and is focused, but there is a problem. That display has 2 parts, the display and the link inside. Focusing the display does nothing if you click the "Enter" key, you have to TAB a 2nd time to focus the link that actually can be interacted with, and whe you focus that link the display shrinks hiding the link. (To be fair this is for blind people so if the link is not visible probably is not a big issue, but still)
tab key

katspaugh:
Tbh I think we're going too deep into nuances. It works pretty well for the majority of cases, and is sufficiently accessible.
Going to merge it.

PR #1721, opened 2023-03-01
fix: show new name in review modal

### LINKED ISSUE #1613: [Add owner] - New owner name is replaced by the entry in the address book
## Bug description
If an entry is in the address book and you add that entry as owner of the safe, the name you set during the "add owner" form is overwritten for whatever is in the address book. It should be the opposite 

## Environment
 - Browser: Chrome
 - Wallet: any
 - Chain: any
 - Prod: https://app.safe.global/

## Steps to reproduce
 1. Go to a safe you own
 2. Add in the address book an entry of another wallet you control and set it a recongizable name
 3. Go to add owner 
 4. Add the entry you have in the AB with a different name than the one in the AB
 5. Set any policy
 6. Check the name of the new owner in the review step

## Expected result
The owner set during the form should be the one for that new owner, should overwrite the entry in the AB once is proccess of adding the owner is done

## Obtained result
The name is the one set in the AB, ignoring what you set during the form

## Screenshots
owner name

iamacook:
## What it solves

Resolves #1613

## How this PR fixes it

The new name of the owner-to-be-added is shown in the review modal instead of that from the address book.

## How to test it

1. Add an address that is NOT AN OWNER of the current Safe to the address book.
2. Attempt to add the aformentioned address as an owner of the Safe, specifying a new name.
3. Observe the new name shown in the review modal under which the address will be stored under

## Screenshots

new name

## Checklist

### DIFF
--- a/src/components/settings/owner/AddOwnerDialog/DialogSteps/ReviewOwnerTxStep.tsx
+++ b/src/components/settings/owner/AddOwnerDialog/DialogSteps/ReviewOwnerTxStep.tsx
@@ -120,7 +120,13 @@ export const ReviewOwnerTxStep = ({ data, onSubmit }: { data: ChangeOwnerData; o
           </div>
           <Divider />
           <Box padding={2}>
-            <EthHashInfo address={newOwner.address} shortAddress={false} showCopyButton hasExplorer />
+            <EthHashInfo
+              name={newOwner.name}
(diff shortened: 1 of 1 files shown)


schmanu (approved):

francovenica:
LGTM. The name of the form is the one shown in the review step

PR #1729, opened 2023-03-02
[Experiment] sign on any chain

katspaugh:
## What it solves

This PR allows signing Safe transaction while the signer wallet is connected to any chain.

We use a hybrid provider based on the wallet provider with the "read only" JSON RCP provider mixed in for any calls other than signing and executing.

Screenshot 2023-03-02 at 17 14 32

### DIFF
--- a/src/hooks/coreSDK/safeCoreSDK.ts
+++ b/src/hooks/coreSDK/safeCoreSDK.ts
@@ -1,12 +1,12 @@
 import chains from '@/config/chains'
-import { createWeb3ReadOnly, getWeb3 } from '@/hooks/wallets/web3'
+import { getWeb3 } from '@/hooks/wallets/web3'
 import ExternalStore from '@/services/ExternalStore'
 import { invariant } from '@/utils/helpers'
-import { Web3Provider } from '@ethersproject/providers'
+import { Web3Provider, type JsonRpcProvider } from '@ethersproject/providers'
 import Safe from '@safe-global/safe-core-sdk'
 import type { SafeVersion } from '@safe-global/safe-core-sdk-types'
 import EthersAdapter from '@safe-global/safe-ethers-lib'
-import type { ChainInfo, SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
+import type { SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import { type EIP1193Provider } from '@web3-onboard/core'
 import { ethers } from 'ethers'
 import semverSatisfies from 'semver/functions/satisfies'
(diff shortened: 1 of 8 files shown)


src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -44,16 +44,37 @@ export const initSafeSDK = async (
...
   if (isLegacyVersion(safeVersion)) {
     isL1SafeMasterCopy = true
   }
 
-  const ethersProvider = new Web3Provider(provider)
+  const signerProvider = new Web3Provider(provider)
+
+  // Monkey patch the signerProvider to proxy requests to the "readonly" provider if on the wrong chain
+  if (walletChainId !== chainId) {
+    const signerMethods = ['eth_sign', 'eth_signTypedData', 'eth_accounts', 'personal_sign']
```
schmanu, line 60: I think this is missing one [signing method](https://docs.metamask.io/guide/signing-data.html#signtypeddata-v4)

```suggestion
    const signerMethods = ['eth_sign', 'eth_signTypedData', 'eth_signTypedData_v4', 'eth_accounts', 'personal_sign']
```

src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -44,16 +44,37 @@ export const initSafeSDK = async (
   chainId: string,
   safeAddress: string,
   safeVersion: string,
+  chainInfo: ChainInfo,
+  walletChainId: string,
```
schmanu, line 48: I would prefix them so it is clearer which `chainId` is from which source (opened Safe or connected wallet)
```suggestion
  safeChainId: string,
  safeAddress: string,
  safeVersion: string,
  safeChainInfo: ChainInfo,
  walletChainId: string,
```

schmanu (commented):
Very creative solution to combine a different RPC signer and provider!

I still think we should display an additional warning `Your Safe and connected wallet are on different networks. [Switch network]` because this could be a indicator that a user might send funds on a wrong chain by mistake. 
Especially considering that when using relaying these transactions would also get executed.

src/hooks/coreSDK/safeCoreSDK.ts
iamacook: Nit: can we create an enum of these? These are used in `_tryOffChainSigning` as well. Maybe it could be a "global" enum.

src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -44,16 +44,37 @@ export const initSafeSDK = async (
...
+  const signerProvider = new Web3Provider(provider)
+
+  // Monkey patch the signerProvider to proxy requests to the "readonly" provider if on the wrong chain
+  if (walletChainId !== chainId) {
+    const signerMethods = ['eth_sign', 'eth_signTypedData', 'eth_accounts', 'personal_sign']
+    const readonlyProvider = createWeb3ReadOnly(chainInfo.rpcUri)
+    const originalSend = signerProvider.send
+
+    signerProvider.send = (request, ...args) => {
+      console.log(request, ...args)
```
iamacook, line 65: Do we want to keep this on production?

iamacook (commented):
This is great! How does it fare with hardware wallets?

src/hooks/coreSDK/safeCoreSDK.ts
katspaugh: I'm not sure yet this entire hack can go to prod.

src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -44,16 +44,37 @@ export const initSafeSDK = async (
...
   if (isLegacyVersion(safeVersion)) {
     isL1SafeMasterCopy = true
   }
 
-  const ethersProvider = new Web3Provider(provider)
+  const signerProvider = new Web3Provider(provider)
+
+  // Monkey patch the signerProvider to proxy requests to the "readonly" provider if on the wrong chain
+  if (walletChainId !== chainId) {
+    const signerMethods = ['eth_sign', 'eth_signTypedData', 'eth_signTypedData_v4', 'eth_accounts', 'personal_sign']
```
rmeissner, line 60: this kind of removes support for EIP-712 (eth_signTypedData) signatures. Do we really want that? Also the flow "should" not work with eth_signTypedData (in case it gets there), as by definition in the EIP it should enforce the chainid.

From https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator:
> `uint256 chainId` the [EIP-155](https://eips.ethereum.org/EIPS/eip-155) chain id. The user-agent should refuse signing if it does not match the currently active chain.

katspaugh:
Closing because, as @rmeissner pointed out, `eth_signTypedData` actually _is_ chain-specific, and this solution falls back to `eth_sign` which is suboptimal.

PR #1730, opened 2023-03-03
feat: relay a transaction trough the `sponsoredCall` endpoint

DiogoSoaress:
## What it solves

Partially resolves https://github.com/safe-global/web-core/issues/1700

## How this PR fixes it
Adds a function to call the tx relayer if the relaying feature is enabled in the current chain and if the connected wallet can relay the tx. SC wallets can only relay fully signed transactions.
Adds a Relaying event to the observable events.
Monitors the relay transaction. The monitor timeout after 3 minutes.

## How to test it

_1/N Safe_
1. Create a transaction.
2. Check the "Execute transaction".
3. Opt for "Via relayer" execution method.
4. Submit

_m/N Safe_
1. Click to sign a transaction that can you can execute.
2. Check the "Execute transaction".
3. Opt for "Via relayer" execution method.
4. Submit

5. Repeat previous steps (1 - 4) with a SC wallet connected.
6. The execution method option shall not appear.

_Fully signed transaction_
1. Click to execute a fully signed transaction.
2. Opt for "Via relayer" execution method.
4. Submit

You shall see the notifications for the Relay transaction as well as the TxStatus changing from "Relaying" -> "indexing" -> "Success"

## Screenshots
Screenshot 2023-03-06 at 14 38 16
Screenshot 2023-03-06 at 14 38 33

## Checklist


DiogoSoaress:
@reviewers the commit history is a bit polluted. Sorry for that, this branch was based on a previous one that was squashed into `epic-relaying-v1` and because of that I can't really rebase `sponsored-call`

src/components/tx/ExecutionMethod/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (createdTx.signatures.size < safe.threshold) {
+    const signedTransaction = await dispatchTxSigning(createdTx, safe.version)
+
+    if (!signedTransaction) {
+      throw new Error('Could not sign transaction')
+    }
+
+    txId = await proposeCallback(signedTransaction)
+    transactionToRelay = signedTransaction
+  }
```
iamacook, line 24: We should keep this logic in `onRelay` to remain consistent with `onSign`/`onExecute`.

src/components/tx/ExecutionMethod/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+    transactionToRelay.data.value,
+    transactionToRelay.data.data,
+    transactionToRelay.data.operation,
+    transactionToRelay.data.safeTxGas,
+    transactionToRelay.data.baseGas,
+    transactionToRelay.data.gasPrice,
+    transactionToRelay.data.gasToken,
+    transactionToRelay.data.refundReceiver,
+    transactionToRelay.encodedSignatures(),
+  ])
```
iamacook, line 39: I'd suggest moving this to the dispatcher.

src/components/tx/SignOrExecuteForm/SignOrExecuteForm.test.tsx
```diff
@@ -484,6 +490,12 @@ describe('SignOrExecuteForm', () => {
       <SignOrExecuteForm isExecutable={true} onSubmit={jest.fn} safeTx={mockTx} onlyExecute={true} />,
     )
 
+    const executionMethod = result.getByText('With connected wallet')
```
iamacook, line 493: Can we add a test for relaying as well?

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -175,14 +184,36 @@ const SignOrExecuteForm = ({
     return id
   }
 
+  const onRelay = async () => {
+    const { gasLimit } = getTxOptions(advancedParams, currentChain)
```
iamacook, line 188: Nit: let's move this below the assertion in case there is no `createdTx`.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -175,14 +184,36 @@ const SignOrExecuteForm = ({
...
+      return
+    }
+
+    try {
+      const relayResponse = await sponsoredCall({ chainId: safe.chainId, to: safeAddress, data: txData.data, gasLimit })
+      await dispatchTxRelay(relayResponse.taskId, txData.txId)
+    } catch (err) {
+      logError(Errors._631, (err as Error).message)
+      return
+    }
```
iamacook, line 206: As mentioned further up, let's keep the conditional signing here but move the rest to `dispatchTxRelay`.

src/hooks/useTxPendingStatuses.ts
```diff
@@ -32,17 +33,24 @@ const useTxMonitor = (): void => {
...
-      if (!txHash || !isProcessing || isMonitored) {
+      if (!(isProcessing || isRelaying) || isMonitored) {
         continue
       }
 
       monitoredTxs.current[txId] = true
 
-      waitForTx(provider, txId, txHash)
+      if (isProcessing) {
+        waitForTx(provider, txId, txHash)
```
iamacook, line 48: A transaction cannot be processing and relaying at the same time.

```suggestion
        waitForTx(provider, txId, txHash)
        continue
```

src/services/tx/txMonitor.ts
```diff
@@ -33,3 +36,52 @@ export const waitForTx = async (provider: JsonRpcProvider, txId: string, txHash:
...
+export const waitForRelayedTx = (taskId: string, txId: string): void => {
+  const gelato = new GelatoRelay()
+
+  const checkTxStatus = async () => {
+    const taskStatus = await gelato.getTaskStatus(taskId)
+
+    switch (taskStatus?.taskState) {
+      case 'CheckPending':
+      case 'ExecPending':
+      case 'WaitingForConfirmation':
```
iamacook, line 49: Can we create an enum for the `taskState`s?

src/services/tx/txMonitor.ts
```diff
@@ -33,3 +36,52 @@ export const waitForTx = async (provider: JsonRpcProvider, txId: string, txHash:
...
+      default:
+        txDispatch(TxEvent.FAILED, {
+          txId,
+          error: new Error(`Relayed transaction was not found.`),
+        })
+        return
+    }
+  }
+
+  setTimeout(checkTxStatus, 2000)
```
iamacook, line 86:
```suggestion
  setTimeout(checkTxStatus, POLLING_INTERVAL)
```

src/services/tx/txMonitor.ts
```diff
@@ -33,3 +36,52 @@ export const waitForTx = async (provider: JsonRpcProvider, txId: string, txHash:
...
+      case 'ExecReverted':
+        txDispatch(TxEvent.REVERTED, {
+          txId,
+          error: new Error(`Relayed transaction reverted by EVM.`),
+        })
+        return
+      case 'Blacklisted':
+        txDispatch(TxEvent.FAILED, {
+          txId,
+          error: new Error(`Relayed transaction was blacklisted by Relay provider`),
```
iamacook, line 67:
```suggestion
          error: new Error(`Relayed transaction was blacklisted by relay provider.`),
```

src/services/tx/txMonitor.ts
```diff
@@ -33,3 +36,52 @@ export const waitForTx = async (provider: JsonRpcProvider, txId: string, txHash:
...
+        })
+        return
+      case 'Cancelled':
+        txDispatch(TxEvent.FAILED, {
+          txId,
+          error: new Error(`Relayed transaction was cancelled by Relay provider`),
```
iamacook, line 73:
```suggestion
          error: new Error(`Relayed transaction was cancelled by relay provider.`),
```

src/services/tx/txMonitor.ts
```diff
@@ -33,3 +36,52 @@ export const waitForTx = async (provider: JsonRpcProvider, txId: string, txHash:
     })
   }
 }
+
+export const waitForRelayedTx = (taskId: string, txId: string): void => {
```
iamacook, line 40: I think we should add a limit on this like we do in `waitForTx`. If it's taking longer than usual, we remove the status. Wdyt?

iamacook (changes requested):

src/services/tx/txMonitor.ts
schmanu: I checked the Gelato docs and did not find much information about expected times for exectution. 
The only comment I found is about the `WaitingForConfirmation` state:
> the relay request has been received by Gelato Relay and it is waiting to be confirmed as a task in the relay task queue. This state usually lasts only a few seconds.

So I guess for that one we could have a ~30s timeout.  But for execution the timeout should rather be too big than too small.

schmanu (commented):
I'm not sure about scopes but I found these issues  while testing:
- We currently offer relaying on all networks. But currently it is only available on görli. 
   We should use introduce a new Feature for this on our config service
- If a Safe is connected as owner, it currently always askes twice for a Signature first `signTypedData` and then `eth_sign`. In general I was not able to relay with a Safe connected.
- If a transaction is fully signed, the pending status does not work. It is always shown as "Awaiting Execution"

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -175,14 +184,36 @@ const SignOrExecuteForm = ({
...
+    try {
+      txData = await prepareRelayTxData(createdTx, safe, proposeTx)
+    } catch (err) {
+      console.error('Error signing relay tx data', err)
+      return
+    }
+
+    try {
+      const relayResponse = await sponsoredCall({ chainId: safe.chainId, to: safeAddress, data: txData.data, gasLimit })
+      await dispatchTxRelay(relayResponse.taskId, txData.txId)
```
schmanu, line 202: If the transaction is already signed, it will not get proposed again in `prepareRelayTxData` and thus `txData.txId` will be undefined. You should use `txData.txId || txId` instead

yarn.lock
```diff
@@ -4647,6 +4655,13 @@ axe-core@^4.4.3:
   resolved "https://registry.yarnpkg.com/axe-core/-/axe-core-4.6.1.tgz#79cccdee3e3ab61a8f42c458d4123a6768e6fbce"
   integrity sha512-lCZN5XRuOnpG4bpMq8v0khrWtUOn+i8lZSb6wHZH56ZfbIEv6XwJV84AAueh9/zi7qPVJ/E4yz6fmsiyOmXR4w==
 
+axios@0.24.0:
```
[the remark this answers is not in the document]
DiogoSoaress: The downside is that we will have to hardcode the URL. Alternatively, it could be exposed by the relay-service. Wdyt?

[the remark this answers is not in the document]
iamacook: This dependency is only added because we are installing the service directly to get the types. It won't be there when we create a types package (or equivalent).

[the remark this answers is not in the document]
schmanu: No it is a dependency of the `gelato-relay-sdk` not from our service.
We use the sdk to track the status of `taskId`s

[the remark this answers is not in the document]
iamacook: Ah, I didn't realise that. We should definitely discuss this.

src/components/tx/ExecutionMethod/utils.ts
DiogoSoaress: Moved

src/components/tx/ExecutionMethod/utils.ts
DiogoSoaress: Moved it to `onRelay`

src/services/tx/txMonitor.ts
DiogoSoaress: I will have to use
```
switch (taskStatus.taskState as string) {
```

as we get `taskStatues.taskState` typed from the Gelato request and they don't export that type. Should I create the enum anyway?

src/services/tx/txMonitor.ts
DiogoSoaress: Here we just need to delay a bit fetching the task status as it is not immediately available if we fetch it right away (though the tx is successfully relayed).
Screenshot 2023-03-07 at 16 03 44

src/services/tx/txMonitor.ts
DiogoSoaress: _Question:_ as you can't cancel the relaying transaction, should we display something along the lines of the `waitForTx` timeout msg?

```
throw new Error(
  `Transaction not relayed in ${TIMEOUT_MINUTES} minutes. Be aware that it might still be relayed.`,
)
```

src/components/tx/SignOrExecuteForm/index.tsx
DiogoSoaress: I've changed this logic. If the tx is already signed we use that `txId`

src/services/tx/sponsoredCall.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// TODO: import type from relay-service
+export type SponsoredCallPayload = {
+  chainId: string
+  to: string
+  data: SafeTransactionData['data'] | undefined
+  gasLimit?: string | number
+}
+
+const SPONSORED_CALL_URL = 'http://localhost:3001/v1/relay'
```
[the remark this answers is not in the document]
DiogoSoaress: Moved it to a env var so we can pass it different values.
However, this change also requires the var to be set in the CI `Build` job, right?

src/services/tx/txMonitor.ts
iamacook: Can you move the value to a constant with a comment explaining this?

DiogoSoaress:
> We currently offer relaying on all networks. But currently it is only available on görli.
We should use introduce a new Feature for this on our config service

We've added the `RELAYING` feature in the config service and we now render the UI conditionally

src/services/tx/txMonitor.ts
DiogoSoaress: That sounds good. In general everything that concerns the Gelato SDK would be nice to be consumed from our `relaying-service` if possible

src/services/tx/txMonitor.ts
DiogoSoaress: Thanks! I've moved the value to a constant and also added a comment.

DiogoSoaress:
> If a transaction is fully signed, the pending status does not work. It is always shown as "Awaiting Execution"

@schmanu it worked well for me. Can you please check again?

src/services/tx/txMonitor.ts
DiogoSoaress: I think I misread this comment the other day as I was trying to type `taskState`. I've created the Enum for the TaskStates.

src/services/tx/sponsoredCall.ts
```diff
@@ -0,0 +1,35 @@
...
+// TODO: import type from relay-service
+export type SponsoredCallPayload = {
+  chainId: string
+  to: string
+  data: SafeTransactionData['data'] | undefined
+  gasLimit?: string | number
+}
+
+export const getSafeGelatoRelayServiceUrl = () => {
+  return `${SAFE_GELATO_RELAY_SERVICE_URL}/v1/relay`
```
iamacook, line 13: There is no dynamic value in this so it can be imported from the constant directly.

src/components/tx/ExecutionMethod/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+export const prepareRelayTxData = async (
+  createdTx: SafeTransaction,
+  safe: SafeInfo,
+  proposeCallback: (tx: SafeTransaction) => Promise<string>,
+) => {
+  let transactionToRelay = createdTx
+  let txId: string | undefined
+
+  // Add missing signature
+  if (createdTx.signatures.size < safe.threshold) {
```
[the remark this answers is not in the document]
DiogoSoaress: I verify now if the connected wallet can relay the tx: SC wallets can only relay fully signed txs.

src/services/tx/txMonitor.ts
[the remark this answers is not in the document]
DiogoSoaress: Added.

yarn.lock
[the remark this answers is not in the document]
DiogoSoaress: I've removed this dependency

src/services/tx/txMonitor.ts
DiogoSoaress: I've added a timeout of 3 min and added a test for it

DiogoSoaress:
I've addressed all the comments and covered the implementation with unit tests.

src/services/tx/__tests__/txMonitor.test.ts
```diff
@@ -94,4 +104,161 @@ describe('txMonitor', () => {
...
+      const mockData = {
+        task: {
+          taskState: 'CheckPending',
+        },
+      }
+      global.fetch = jest.fn().mockImplementation(setupFetchStub(mockData))
+
+      const mockFetch = jest.spyOn(global, 'fetch')
+      const mockSetTimeout = jest.spyOn(global, 'setTimeout')
+      // const mockCheckTxStatus = jest.spyOn(global, 'checkTxStatus')
```
iamacook, line 119:
```suggestion
```

src/services/tx/sponsoredCall.ts
```diff
@@ -0,0 +1,31 @@
+import { SAFE_GELATO_RELAY_SERVICE_URL } from '@/config/constants'
+import { type SafeTransactionData } from '@safe-global/safe-core-sdk-types'
+
+// TODO: import type from relay-service
+export type SponsoredCallPayload = {
+  chainId: string
+  to: string
+  data: SafeTransactionData['data'] | undefined
```
iamacook, line 8: `data` cannot be `undefined`.

```suggestion
  data: SafeTransactionData['data']
```

src/services/tx/txMonitor.ts
```diff
@@ -33,3 +36,109 @@ export const waitForTx = async (provider: JsonRpcProvider, txId: string, txHash:
...
+  // A small delay is necessary before the initial polling as the task status
+  // is not immediately available after the sponsoredCall request
+  const INITIAL_POLLING_DELAY = 2_000
+
+  const TIMEOUT_MINUTES = 3
+  let timeoutId: NodeJS.Timeout
+
+  const checkTxStatus = async () => {
```
iamacook, line 74: Please add tests for this.

src/hooks/useWalletCanRelay.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const useWalletCanRelay = (tx: SafeTransaction | undefined) => {
+  const [isSCWallet, , loadingWalletType] = useIsSmartContractWallet()
+  const { safe } = useSafeInfo()
+
+  if (!tx || loadingWalletType) return false
+
+  const hasEnoughSignatures = tx.signatures.size >= safe.threshold
+
+  return !isSCWallet || (isSCWallet && hasEnoughSignatures)
```
iamacook, line 13: This will return `true` is `useIsSmartContractWallet` throws.

src/components/tx/SignOrExecuteForm/SignOrExecuteForm.test.tsx
```diff
@@ -368,6 +376,114 @@ describe('SignOrExecuteForm', () => {
...
+    jest.spyOn(txSender, 'default').mockImplementation(
+      () =>
+        ({
+          dispatchTxProposal: jest.fn(() => Promise.resolve({})),
+          dispatchTxRelay: relaySpy,
+        } as unknown as NullableTxSenderFunctions),
+    )
+
+    // SC wallet connected
+    jest.spyOn(useIsSmartContractWallet, 'default').mockReturnValue([true, , false])
```
iamacook, line 458: This is only testing the relay of fully signed transactions *from smart contract wallets, right? If so, can we adjust the description and add a test for normal wallets?

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -175,14 +190,40 @@ const SignOrExecuteForm = ({
...
 
+  const onRelay = async () => {
+    const [, createdTx] = assertDependencies()
+
+    const { gasLimit } = getTxOptions(advancedParams, currentChain)
+
+    let id = txId
+    let safeTx = createdTx
+    // Add missing signature
+    if (createdTx.signatures.size < safe.threshold) {
```
iamacook, line 201: Can you extract this logic to a function along the lines of `isRelayable` as the same logic is use in `useWalletCanRelay`. It could live alongside `isSignableBy`.

schmanu (approved):
This looks very good! 
I only have some minor suggestions.

Also one comment on the smart contract owner UX:
Now if the tx is not fully signed Relaying will not show up as an  disabled option. So users might not know that if they fully sign the tx relaying would be possible. 
I don't know if we want to show this differently.
But it is also a niche case for now so it's not too important.

src/services/exceptions/ErrorCodes.ts
```diff
@@ -30,6 +30,8 @@ enum ErrorCodes {
   _619 = '619: Error fetching data from master-copies',
   _620 = '620: Error loading chains',
   _630 = '630: Error fetching remaining hourly relays',
+  _631 = '631: Transaction failed to be relayed',
+  _632 = '632: Error fetching task status',
```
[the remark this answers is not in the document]
iamacook: A task can reference anything though. If you want to be specific then I'd suggest combining the both: "Error fetching relay task status".

src/components/tx/SignOrExecuteForm/index.tsx
DiogoSoaress: Sure, it can be abstracted. However I'd call this condition something like `hasEnoughSignatures` like I've done in `useWalletCanRelay` WDYT?
I also think it does not quite fit the `transaction-guards` file, I'd rather move it to `utils/transactions` no?

iamacook: `hasEnoughSignatures` sounds fine by me. I'm impartial where you situate it.

DiogoSoaress: Done

iamacook (approved):

PR #1731, opened 2023-03-03
fix: Add policy pages

### LINKED ISSUE #1722: Update policy pages
## The list of requirements

- Adds a privacy policy, cookie policy and imprint page to the app

## Links

https://www.notion.so/gnosis-safe/Terms-policy-update-f3827a6f724340deae9554d24afe8c29

usame-algan:
## What it solves

Resolves #1722 

## How this PR fixes it

- Adds a new page for Privacy policy, Cookie policy, Terms and Imprint
- Links those pages in the footer
- Displays the footer on those pages
- Hides the sidebar on those pages
- Shows a one time terms popup for users who have previously dismissed the cookie banner (existing users)

## How to test it

1. Open the Safe
2. Observe a Terms banner showing up in the bottom corner
3. Open the Safe in a private tab and observe only the cookie banner showing
4. Accept the cookie banner and observe the Terms banner not showing up
5. Reload the page and observe the Terms banner not showing up
6. Go to Settings to see the Footer
7. Click on any of the three links to see the page
8. Compare to https://www.notion.so/gnosis-safe/Terms-policy-update-f3827a6f724340deae9554d24afe8c29
9. Create or Load a new safe and click on terms of use or privacy policy and observe seeing the new pages

## Screenshots

Screenshot 2023-03-03 at 11 06 23

### DIFF
--- a/src/components/common/TermsBanner/index.tsx
+++ b/src/components/common/TermsBanner/index.tsx
@@ -1,0 +1,44 @@
+import { AppRoutes } from '@/config/routes'
+import css from './styles.module.css'
+import { Button, Typography } from '@mui/material'
+import Link from 'next/link'
+import MUILink from '@mui/material/Link'
+import useLocalStorage from '@/services/local-storage/useLocalStorage'
+
+const TERMS_KEY = 'terms_dismissed'
+
+const TermsBanner = () => {
+  const [isDismissed = false, setIsDismissed] = useLocalStorage<boolean>(TERMS_KEY)
+
+  const dismissBanner = () => {
+    setIsDismissed(true)
+  }
+
+  if (isDismissed) return null
+
+  return (
+    <div className={css.wrapper}>
+      <Typography variant="h4" fontWeight="bold" mb={1}>
+        Terms
+      </Typography>
+      <Typography>
+        We updated our Terms and Conditions. Review the new terms{' '}
+        <Link href={AppRoutes.terms} passHref>
+          <MUILink color="success.main"> here</MUILink>
+        </Link>
+      </Typography>
+      <Button
+        variant="contained"
+        color="success"
+        size="small"
+        className={css.button}
+        onClick={dismissBanner}
+        disableElevation
+      >
+        Ok
+      </Button>
+    </div>
+  )
+}
+
+export default TermsBanner
--- a/src/components/common/TermsBanner/index.tsx
+++ b/src/components/common/TermsBanner/index.tsx
@@ -28,10 +28,11 @@ const TermsBanner = () => {
         Terms
       </Typography>
       <Typography>
-        We updated our Terms and Conditions. Review the new terms{' '}
+        We&apos;ve updated our Terms and Conditions. You can review them{' '}
         <Link href={AppRoutes.terms} passHref>
-          <MUILink color="success.main"> here</MUILink>
+          <MUILink color="success.main">here</MUILink>
         </Link>
+        .
       </Typography>
       <Button
         variant="contained"
--- a/src/components/common/TermsBanner/index.tsx
+++ b/src/components/common/TermsBanner/index.tsx
@@ -4,7 +4,7 @@ import { Button, Typography } from '@mui/material'
 import Link from 'next/link'
 import MUILink from '@mui/material/Link'
 import useLocalStorage from '@/services/local-storage/useLocalStorage'
-import { useEffect, useMemo, useState } from 'react'
+import { useMemo } from 'react'
 import { type CookiesState, CookieType } from '@/store/cookiesSlice'
 import local from '@/services/local-storage/local'
 
@@ -15,22 +15,15 @@ const isExistingUser = () => {
 }
 
 const TermsBanner = () => {
-  const [isMounted, setIsMounted] = useState<boolean>(false)
   const [isDismissed = false, setIsDismissed] = useLocalStorage<boolean>(TERMS_KEY)
 
   // Check on page load if "necessary" cookies have been accepted
   const existingUser = useMemo(() => isExistingUser(), [])
   const shouldOpen = existingUser && !isDismissed
 
-  useEffect(() => {
-    setIsMounted(true)
-  }, [])
-
   const dismissBanner = () => {
     setIsDismissed(true)
   }
-
-  if (!isMounted) return <></>
 
   return shouldOpen ? (
     <div className={css.wrapper}>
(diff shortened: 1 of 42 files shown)


src/components/common/TermsBanner/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <div className={css.wrapper}>
+      <Typography variant="h4" fontWeight="bold" mb={1}>
+        Terms
+      </Typography>
+      <Typography>
+        We updated our Terms and Conditions. Review the new terms{' '}
+        <Link href={AppRoutes.terms} passHref>
+          <MUILink color="success.main"> here</MUILink>
```
iamacook, line 27:
```suggestion
        We've updated our Terms and Conditions. You can review them{' '}
        <Link href={AppRoutes.terms} passHref>
          <MUILink color="success.main">here</MUILink>.
```

src/components/cookie-policy/index.tsx
```diff
@@ -0,0 +1,572 @@
...
+  return (
+    <div>
+      <h1>Cookie Policy</h1>
+      <p>Last updated on March 2023</p>
+      <p>
+        As described in our{' '}
+        <Link href={AppRoutes.privacy} passHref>
+          <MUILink>Privacy Policy</MUILink>
+        </Link>
+        , For general web-browsing of this website, your personal data is not revealed to us, although certain
```
iamacook, line 16:
```suggestion
        , for general web-browsing of this website, your personal data is not revealed to us, although certain
```

katspaugh (approved):

katspaugh:
Please also bump the minor version in package.json.

francovenica:
LGTM

tschubotz:
When opening for the first time, the cookie banner comes up. The link to the cookie policy is https://safe.global/cookie instead of https://app.safe.global/cookie 

image

tschubotz:
image

will new users (i.e. users who have never been on the page before) also get this?

(Ideally it would only be displayed to existing users. If we don't have a nice way of determining existing users, I'd say users that have at least 1 safe added.)

tschubotz:
I know this wasn't specified anywhere, but I think eventually we should move the licenses also to the `app` subdomain since it affects the apps only and doesn't make sense on the landing page. Where is the best spot to create an issue for this? 


image

src/components/common/TermsBanner/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+import css from './styles.module.css'
+import { Button, Typography } from '@mui/material'
+import Link from 'next/link'
+import MUILink from '@mui/material/Link'
+import useLocalStorage from '@/services/local-storage/useLocalStorage'
+import { useEffect, useState } from 'react'
+
+const TERMS_KEY = 'terms_dismissed'
+
+const TermsBanner = () => {
```
katspaugh, line 11: I would suggest we use the OnboardingTooltip for this. It implements all the same logic and looks more consistent with the rest of the theme.

usame-algan: Currently the component only displays a small Tooltip when wrapped around a child but maybe we can extract some of it into a hook and reuse that.

tschubotz:
> > will new users (i.e. users who have never been on the page before) also get this?
> 
> Yes, the Banner would be shown to all users until they click it away.

@usame-algan Is there a way to limit the terms update banner to only existing users? 

(If not, am I right to assume that we would leave this up for e.g. 1 month and then disable it again? Would be weird to have this banner up until the end of time.)

iamacook:
> @usame-algan Is there a way to limit the terms update banner to only existing users?

We could check the presence of the cookie preferences in `localStorage`. If the user previously (dis-)agreed to cookies, the key will be there.

usame-algan:
> @usame-algan Is there a way to limit the terms update banner to only existing users?

> We could check the presence of the cookie preferences in localStorage. If the user previously (dis-)agreed to cookies, the key will be there.

I've changed it so that the Terms banner is only shown if a user previously dismissed the cookie banner. This should cover most existing users.

> (If not, am I right to assume that we would leave this up for e.g. 1 month and then disable it again? Would be weird to have this banner up until the end of time.)

We can remove it with a future release.

src/components/common/TermsBanner/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const [isMounted, setIsMounted] = useState<boolean>(false)
+  const [isDismissed = false, setIsDismissed] = useLocalStorage<boolean>(TERMS_KEY)
+
+  // Check on page load if "necessary" cookies have been accepted
+  const existingUser = useMemo(() => isExistingUser(), [])
+  const shouldOpen = existingUser && !isDismissed
+
+  useEffect(() => {
+    setIsMounted(true)
+  }, [])
```
iamacook, line 27: Can we not instantiate the state `true`?

usame-algan: It didn't work. We need to wait until the component is mounted before we set the value.

usame-algan: I removed the state and import `TermsBanner` via `next/dynamic` with ssr turned off for this component to solve the hydration error.

src/pages/_app.tsx
```diff
@@ -34,6 +35,8 @@ import createEmotionCache from '@/utils/createEmotionCache'
 import MetaTags from '@/components/common/MetaTags'
 import useAdjustUrl from '@/hooks/useAdjustUrl'
 
+const TermsBanner = dynamic(() => import('@/components/common/TermsBanner'), { ssr: false })
```
iamacook, line 38: We should add a comment why disabling SSR is required.

iamacook (approved):

francovenica:
> I've changed it so that the Terms banner is only shown if a user previously dismissed the cookie banner. This should cover most existing users.

This works, the new terms and conditions message only shows up if you cleared the cookie banner. NOTE: It shows up only if you do a full refresh of the page, just going to another section of the app (like settings or assets) won't be enough. 

> When opening for the first time, the cookie banner comes up. The link to the cookie policy is https://safe.global/cookie instead of https://app.safe.global/cookie

This one was fixed as well

usame-algan:
> This works, the new terms and conditions message only shows up if you cleared the cookie banner by accepting some or all the preferences. NOTE: It shows up only if you do a full refresh of the page, just going to another section of the app (like settings or assets) won't be enough.

I pushed another update to this so that the Terms banner should never show up for new users now, even if they reload the page after accepting cookies.

francovenica:
TLDR; works fine

So testing this makes me a "new user" every time, since clearing the local storage puts the cookie banner in "not cleared" therefore the LS key for the Terms banner will also be set in "not show".
To be an "old user" I have to remove the new key that shows/hide the Terms banner (as old users should not have that key, so still a realistic case). Reloading the page creates and sets the key again in true, showing me the Terms banner.

PR #1733, opened 2023-03-03
Fix: revert regression in useSafeAppFromManifest

katspaugh:
## What it solves

Reverts a change from #1674

## How this PR fixes it

A regression was introduced in #1674 which prevented Safe Apps with an unreachable manifest.json from being loaded.

## How to test it
Load the InsurAce app, it should load w/o problem, but on prod it won't.

### DIFF
--- a/src/hooks/safe-apps/useSafeAppFromManifest.ts
+++ b/src/hooks/safe-apps/useSafeAppFromManifest.ts
@@ -11,7 +11,7 @@ type UseSafeAppFromManifestReturnType = {
 }
 
 const useSafeAppFromManifest = (appUrl: string, chainId: string): UseSafeAppFromManifestReturnType => {
-  const [data, error, loading] = useAsync<SafeAppDataWithPermissions>(() => {
+  const [data, error, isLoading] = useAsync<SafeAppDataWithPermissions>(() => {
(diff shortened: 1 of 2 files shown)


JagoFigueroa:
Hola señores, I am comparing this PR with the PR for 1.5.1 release and I've noticed this case for xdai bridge app on mainnet (might be more cases like this one, still checking).

On [this PR](https://fix_manifest--webcore.review-web-core.5afe.dev/apps/open?safe=eth:0xBEFA89D90c112A416139F42046Cc2d4A8a308815&appUrl=https%3A%2F%2Fbridge.xdaichain.com) it won't connect to the safe automatically (making it not usable)

But on [1.5.1 release PR](https://hotfix_1_5_1--webcore.review-web-core.5afe.dev/eth:0xBEFA89D90c112A416139F42046Cc2d4A8a308815/apps?appUrl=https%3A%2F%2Fbridge.xdaichain.com) I see that the connection is established as expected.

Any ideas what could be the change?

katspaugh:
Should be good now. Looks like there’s a pretty fragile dependency on the timing of the router.

JagoFigueroa (approved):

JagoFigueroa:
Looks good to me. Thank you for the quick action!

katspaugh:
Thank you caballero!

PR #1738, opened 2023-03-06
feat: allow initialisation of SDK with unsupported mastercopies

iamacook:
## What it solves

SDK initialisation on unsupported networks.

## How this PR fixes it

If the mastercopy is an official deployment that is not supported on the current network, e.g. L1 contract on L2 network, we now initialise the SDK.

We now fetch the Safe version directly from the contract (as it is `null` from backend) and detect whether the address matches an official Safe deployment.

## How to test it

1. Open an unsupported Safe and observe that it is now possible to execute a transaction.
2. Observe the warning in the transaction modal.

It is possible to create an "unsupported" either via the CLI or by using the Copycat app and deploying a mainnet address to an L2 network, e.g. Polygon.

## Screenshots

image

## Checklist


JagoFigueroa:
Tested and verified with the help of señor @iamacook 😉

src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -39,19 +41,38 @@ export const createEthersAdapter = (provider = getWeb3()) => {
...
+    const safeL2Deployment = getSafeL2SingletonDeployment({ network: chainId, version: safeVersion })
+
+    isL1SafeMasterCopy = masterCopy === safeL1Deployment?.defaultAddress
+    const isL2SafeMasterCopy = masterCopy === safeL2Deployment?.defaultAddress
+
+    if (!isL1SafeMasterCopy && !isL2SafeMasterCopy) {
+      // Unknown masterCopy, which we do not want to support
+      return Promise.resolve(undefined)
+    }
+  } else {
```
usame-algan, line 69: Afaiu the main fix in this PR is that we now initialize the Core SDK even if a `safe.version` doesn't exist but I don't understand why we need this part.

To test, I deployed a Safe on Polygon with the L1 master copy contract instead of L2 (`matic:0x2B7a8C8afaEc010140970488190275A26b1436DD`) which should fall under the "unsupported contract" category. For some reason, the backend still returns a version for this Safe so it is not caught by this condition but it is still possible to create and execute transactions.

Should we instead use the same condition we use for showing the notification `isValidMasterCopy`?

usame-algan: `isValidMasterCopy` seems to catch Safes with an unsupported contract both when there is a version and when there is no version.

usame-algan (approved):

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -261,6 +267,16 @@ const SignOrExecuteForm = ({
...
         ) : submitError ? (
           <ErrorMessage error={submitError}>Error submitting the transaction. Please try again.</ErrorMessage>
+        ) : willExecute && isUnknown ? (
+          <ErrorMessage>
+            This Safe was created with an unsupported base contract. It should <b>ONLY</b> be used for fund recovery.
+            Transactions will execute but the transaction list may not immediately update. Transaction success can be
+            verified on the{' '}
+            <ExternalLink href={getExplorerLink(safeAddress, currentChain!.blockExplorerUriTemplate).href}>
+              {currentChain?.chainName} explorer
+            </ExternalLink>
```
usame-algan, line 277: I would suggest not using `!` here but instead checking that `currentChain` exists otherwise this could lead to runtime errors.

PR #1746, opened 2023-03-07
refactor: delay switching networks until execution

### LINKED ISSUE #1758: Connect the wallet on any network without UI limitations
## What is the feature about

Currently, the user wallet has to be on the same network as the safe that the user is connected to in order to interact with many parts of the UI. This is an unnecessary limitation because a signer is only needed when executing or signing a transaction.

## The list of requirements

- Only prompt the user to switch to the correct network when signing or executing a transaction
- Remove all `Switch to` buttons in the UI
- Enable all buttons that were previously disabled because of this limitation
- Initialise the Core SDK without a signer
- Connect to the Core SDK with a signer when needed

## Designs/sketches

223422490-379e6d82-2b83-47f6-9d7f-04c0b6fd3ad8

usame-algan:
## What it solves

Resolves #1758 

So far the user wallet always had to be on the same chain as the Safe in order to interact with many parts of the UI. This removes that limitation and instead shows a warning message, requesting chain switch when submitting.

- Globally initialises the Core SDK in read-only mode instead of with a signer
- Prompts the wallet to switch to the correct network before connecting to the Core SDK instance
- Removes `useTxSender` as it is not needed anymore

Note: Safe creation has not been touched.

## Other changes

- Use `safe.chainId` instead of `useChainId` when executing Safe App/spending limit txs

## How to test it

The key areas that need testing are transaction related but this likely needs a full regression. This needs to be tested with hardware wallets, WC, Mobile pairing and MM.

1. Open the Safe
2. Connect your wallet on a different chain than the safe
3. Observe there are no limitations in the UI (apart from the "Switch to" buttons in the header/account centre)
4. Create a transaction
5. Observe the new warning
6. Clicking Submit first requests you to switch chain

Note: this has been tested with WC, pairing, MM, Trezor Model T, Ledger Nano X successfully.

## Screenshots

image

## Checklist


src/services/tx/tx-sender/sdk.ts
```diff
@@ -17,15 +18,30 @@ export const getAndValidateSafeSDK = (): Safe => {
   return safeSDK
 }
 
+export const switchWalletChain = (wallet: ConnectedWallet, chainId: SafeInfo['chainId']) => {
+  const chainIdInHex = ethers.utils.hexValue(BigNumber.from(chainId))
+
+  return wallet.provider.request({
```
katspaugh, line 24: I wonder if this is in some way superior or inferior to `onboard.setChain`. Or perhaps it's the same thing.

iamacook: `setChain` indeed [makes this request](https://github.com/blocknative/web3-onboard/blob/develop/packages/core/src/chain.ts#L61) [internally](https://github.com/blocknative/web3-onboard/blob/develop/packages/core/src/provider.ts#L436-L444). Looks like there's a lot of validation beforehand which we may benefit from.

katspaugh: As long as it's not affecting any internal state in onboard, I guess it's fine to bypass it. Passing a wallet arg is probably simpler than onboard.

usame-algan: There is also some logic in place to handle hardware wallets (`handleChainSwitch`) where we have to first disconnect and reconnect again. It probably makes sense to reuse that function and rely on `onboard` directly.

src/hooks/useIsValidExecution.ts
```diff
@@ -4,14 +4,45 @@ import type { EthersError } from '@/utils/ethers-utils'
...
+    const signerMethods = ['eth_accounts']
+    const originalSend = signerProvider.send
+
+    signerProvider.send = (request, ...args) => {
+      if (!signerMethods.includes(request)) {
+        return readOnlyProvider.send.call(readOnlyProvider, request, ...args)
+      }
+      return originalSend.call(signerProvider, request, ...args)
+    }
+  }
```
usame-algan, line 41: The hook seems to over-request the RPC a lot now. I get around 70 requests when trying to execute. I suspect it has to do with how we overwrite the `.send` call. Any ideas @katspaugh?

iamacook: It is creating a new provider on every re-render. I'd look into memoisation and/or cleanup of the previous provider.

usame-algan: I checked and the hook only runs once. In the network requests I can see that its the `.send.call` we do probably because we actually call the function instead of just assigning I assume

katspaugh: As Aaron suggested, it seems you might be creating a recursion by repeatedly overwriting the send method. The first time you patch it, the original send is the actual original send, but on subsequent calls it will be the patched version.

usame-algan: I just pushed a fix for #1774 by switching from using `sdk.isValidTransaction` to using the contract directly. This also fixed the over-fetching issue.

src/hooks/useIsValidExecution.ts
```diff
@@ -20,14 +51,38 @@ const useIsValidExecution = (
...
-      return await safeSdk.isValidTransaction(safeTx, { gasLimit: gasLimit.toString() })
+      const provider = getPatchedSignerProvider(wallet, safe.chainId, readOnlyProvider)
+      const { contract } = getSpecificGnosisSafeContractInstance(safe, provider)
+
+      /**
+       * We need to call the contract directly instead of using sdk.isValidTransaction
+       * because gasLimit errors are not propagated otherwise. It also fixes the over-fetching
+       * issue that the monkey patched provider does
+       */
+      return contract.callStatic.execTransaction(
```
usame-algan, line 73: I believe the root cause is that we don't throw an error / swallow errors in the core-sdk: https://github.com/safe-global/safe-core-sdk/blob/main/packages/safe-ethers-lib/src/contracts/GnosisSafe/GnosisSafeContractEthers.ts#L126

We could fix it there and use the sdk call again I think.

src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -79,6 +90,20 @@ export const initSafeSDK = async (provider: EIP1193Provider, safe: SafeInfo): Pr
   })
 }
 
+export const initReadOnlySafeSDK = async (chainId: string, safeAddress: string, safeVersion: string): Promise<Safe> => {
```
iamacook, line 93: `SafeInfo['version']` can be `null`.

iamacook: We should also initialise this "readonly" SDK as we do with the standard one so that it supports unsupported contracts.

src/hooks/useIsValidExecution.ts
```diff
@@ -4,14 +4,45 @@ import type { EthersError } from '@/utils/ethers-utils'
...
+// Monkey patch the signerProvider to proxy requests to the "readonly" provider if on the wrong chain
+const getPatchedSignerProvider = (
+  wallet: ConnectedWallet,
+  chainId: SafeInfo['chainId'],
+  readOnlyProvider: JsonRpcProvider,
+) => {
+  const signerProvider = createWeb3(wallet.provider)
+
+  if (wallet.chainId !== chainId) {
+    const readOnlyMethods = ['eth_getCode', 'eth_call']
```
iamacook, line 32: Can you add a comment why we specifically need to patch these method as it is not immediately clear.

```suggestion
    const READ_ONLY_METHODS = ['eth_getCode', 'eth_call']
```

src/hooks/useIsWrongChain.ts
```diff
@@ -1,10 +1,11 @@
 import useChainId from '@/hooks/useChainId'
 import useWallet from '@/hooks/wallets/useWallet'
 
+// TODO: Remove this later
```
iamacook: 👀

src/hooks/wallets/useInitWeb3.ts
```diff
@@ -12,19 +12,16 @@ export const useInitWeb3 = () => {
...
   useEffect(() => {
-    if (!chain) {
-      return
-    }
+    if (!chain) return
+
     const web3ReadOnly = createWeb3ReadOnly(chain.rpcUri, customRpc?.[chain.chainId])
     setWeb3ReadOnly(web3ReadOnly)
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [chain?.chainId, customRpc])
```
iamacook, line 29: We should keep this dependency. The readonly provider will always initialise, regardless of wallet connection. We should therefore keep it in line with the current `chainId`.

schmanu:
Do we have plans to re-add a warning about the wrong chain? Currently the only way to spot if the connected wallet and Safe are on the same network is by checking the small `Wallet @ Network` label on top of the connected wallet address:
Screenshot 2023-03-15 at 16 31 37

We've seen in the past multiple users accidentally sending funds to addresses on the wrong network. 
While this new chain switching flow saves clicks, it also increases the probability of people accidentally using a Safe on a wrong chain without realizing it and possibly sending funds to addresses on the wrong chain.

src/services/tx/tx-sender/sdk.ts
```diff
@@ -17,15 +22,44 @@ export const getAndValidateSafeSDK = (): Safe => {
...
 }
 
+export const switchWalletChain = async (onboard: OnboardAPI, chainId: string) => {
+  const chainIdInHex = hexValue(parseInt(chainId))
+
+  const wallet = getConnectedWallet(onboard.state.get().wallets)
+
+  if (wallet && isHardwareWallet(wallet)) {
+    await onboard.disconnectWallet({ label: wallet.label })
+    await connectWallet(onboard, { autoSelect: wallet.label })
```
iamacook, line 32: The user may connect to a wrong network here. We could recursively connect again if the newly connect wallet is on the wrong chain. What do you think?

src/hooks/useIsWrongChain.ts
iamacook: I saw that this was now removed. We can also remove the `ChainSwitcher` as well.

iamacook:
If a hardware wallet is connected, clicking "Submit" causes it to disconnect then connect again. In that short period, an error message is shown in the transaction modal:

> You are currently not an owner of this Safe and won't be able to submit this transaction.

After successful connection, this disappears but the submit button is disabled for a while until the transaction appears on the hardware wallet. It is a bit confusing. It would be nice to show a loading indicator, perhaps instead of "Submit" - the state of which can be used to now show the error, e.g.

```tsx
{!isOwner && !isSubmitting && "Error message"}
```

src/services/tx/tx-sender/sdk.ts
usame-algan: True, this is an issue with HW wallets specifically because it is possible to select any network when being triggered. Selecting a wrong network will result in a Core SDK error. I am not sure if it has to be done through recursion but we should definitely make sure that once we leave `switchWalletChain` that the wallet network really switched to the correct one.

usame-algan:
> It would be nice to show a loading indicator, perhaps instead of "Submit" - the state of which can be used to now show the error

Good idea! This would be pretty cool in general because it is easy to miss if the transaction window pops up somewhere else

usame-algan:
> We've seen in the past multiple users accidentally sending funds to addresses on the wrong network.
> While this new chain switching flow saves clicks, it also increases the probability of people accidentally using a Safe on a wrong chain without realizing it and possibly sending funds to addresses on the wrong chain.

Thats a valid concern. While the new flow should prevent this (because it always makes sure the wallet is connected on the correct network before executing), it might still be useful to have an indicator somewhere in the UI. cc @liliiaorlenko @katspaugh

src/services/tx/tx-sender/sdk.ts
katspaugh: We could use the approach I proposed in #1729 as a fallback in that case. Basically allow to sign on the wrong chain – the only down downside is that typed signatures won't work, but HW wallets don't support them anyway, AFAIR.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -248,9 +235,7 @@ const SignOrExecuteForm = ({
         />
 
         {/* Error messages */}
-        {isWrongChain ? (
-          <ErrorMessage>Please connect your wallet to {currentChain?.chainName}</ErrorMessage>
-        ) : cannotPropose ? (
+        {isSubmittable && cannotPropose ? (
```
iamacook, line 238: This prevents the error shown between hardware wallet network switching explained [here](https://github.com/safe-global/web-core/pull/1746#issuecomment-1470284593).

src/services/tx/tx-sender/sdk.ts
iamacook: We were previously only throwing an error if the wallet disconnected. I've now added one for if the wallet connects to the wrong chain:

wrong chain hw wallet

katspaugh: Can't say this flow is the most intuitive. The wallet is already connected and you're suddenly popping up the connection again w/o any explanation.
It would confuse me.
Since an automatic switch cannot be performed for HW wallets, I would rather show a warning and a switch button instead re-connecting them silently.

src/services/tx/tx-sender/sdk.ts
```diff
@@ -17,15 +22,58 @@ export const getAndValidateSafeSDK = (): Safe => {
...
+  } else {
+    await onboard.setChain({ chainId: hexValue(parseInt(chainId)) })
+  }
+
+  /**
+   * Onboard doesn't update immediately and returns a wallet that's on the
+   * previous chain, so we add a slight timeout before accessing its state again
+   */
+  await sleep(500)
```
katspaugh, line 51: Wow. Para nada bueno.

katspaugh: Does the raw RPC call have the same problem? Maybe we should pivot back to that.

iamacook: > Does the raw RPC call have the same problem?

Sadly, yes.

I'd also like to avoid this but the current wallet is needed to [connect the SDK](https://github.com/safe-global/web-core/blob/poc-chain-switch/src/services/tx/tx-sender/sdk.ts#L73-L97)/[call the contract](https://github.com/safe-global/web-core/blob/poc-chain-switch/src/services/tx/tx-sender/dispatch.ts#L247-L251).

src/services/tx/tx-sender/sdk.ts
iamacook: As per our discussion, we've added a warning regarding the reconnection attempt (only for hardware wallets):

image

src/services/tx/tx-sender/sdk.ts
katspaugh: Can we replace this with a polling function or something? I won't be able to sleep if we commit this.

katspaugh: Or some onboard event.

iamacook: I've made some adjustments and it now leverages the store subscription instead.

katspaugh (approved):
Looks good from my side! (Please rm the console.log if not intentional)

src/services/tx/tx-sender/sdk.ts
liliiaorlenko: > As per our discussion, we've added a warning regarding the reconnection attempt (only for hardware wallets):
> 
> image

is this a last version of implementation? I would propose a bit of design improvements for it :)

schmanu (commented):
When playing around a bit more I realized that two things do not work currently:
- Nested safes
- off-chain messages

While off-chain messages if EOAs /hw wallets are connected should be fixable, the nested Safes are quite problematic.

I know its not the average use-case but I think this whole workaround can not work for owner Safes connected to the wrong chain. Probably we should just disable the buttons if the connected wallet is a smart contract and connected to the wrong wallet.

src/components/safe-messages/MsgModal/index.tsx
```diff
@@ -92,7 +90,7 @@ const MsgModal = ({
 
   const hasSigned = !!alreadyProposedMessage?.confirmations.some(({ owner }) => owner.value === wallet?.address)
 
-  const isDisabled = isWrongChain || !isOwner || hasSigned || !web3
+  const isDisabled = !isOwner || hasSigned || !web3
```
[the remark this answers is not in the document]
iamacook: I've adjusted this to now request chain switch before signing.

src/hooks/useIsValidExecution.ts
```diff
@@ -4,14 +4,46 @@ import type { EthersError } from '@/utils/ethers-utils'
...
+const getPatchedSignerProvider = (
+  wallet: ConnectedWallet,
+  chainId: SafeInfo['chainId'],
+  readOnlyProvider: JsonRpcProvider,
+) => {
+  const signerProvider = createWeb3(wallet.provider)
+
+  if (wallet.chainId !== chainId) {
+    // The RPC methods that are used when we call contract.callStatic.execTransaction
+    const READ_ONLY_METHODS = ['eth_chainId', 'eth_call']
```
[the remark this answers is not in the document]
iamacook: The patched signer is only used here to check the validity of the transaction. I've added an extra comments and made the code more declarative to help.

liliya-soroka:
image
1. wallet has Goerli selected
2.  safe on Ethereum is opened on UI
3. click add safe 
Current result: the UI is switched automatically to the wallet network on the "add safe" screen. But the error appears in a few sec after screen opening

iamacook:
> 1. wallet has Goerli selected 2. safe on Ethereum is opened on UI 3. click add safe Current result: the UI is switched automatically to the wallet network on the "add safe" screen. But the error appears in a few sec after screen opening

This should now be fixed.

iamacook:
For future reference: I saw @usame-algan tried stripping wrong chain blockers from Safe creation but it wasn't working as desired so I reverted those changes for now. This PR *only* affects the transaction modals. We can revisit creation if desired after this.

katspaugh (approved):
🚀

iamacook (approved):

francovenica:
Questions:
In the main ticket it says in the description says that we should remove all buttons "Switch to" in the UI, but the ones in the topbar are still present. Was decided to keep those?

This ticket doesn't take in account safe creation, since it won't let you go more than the 1st step. Is this correct?

francovenica:
I'm getting this error very often. Is not consistent, but mostly happens when you switch your safe to the wrong network and try to create a tx right away (without navigating anywhere in the app)
Currenlty I'm testing in a Goerli Safe 1/x, With the MM wallet in chrome, connected to the ETH network. The error message pops up after I click "submit" and I wait for the MM popUp to change my network
image

francovenica:
For Trezor I'm having this issue with Tx already queued, where when tried to be executed the error message shown is the warning that the tx might fail instead of the "wrong network". 
It only happens with trezor ( I have a Trezor One).

Create a tx and queue it. Then with the treozr try to execute it or replace it while being in the wrong network 
image

iamacook:
> Questions: In the main ticket it says in the description says that we should remove all buttons "Switch to" in the UI, but the ones in the topbar are still present. Was decided to keep those?

This is redudant - sorry. The "Switch to" remains in the header (including the account center). I'll update the description.

> This ticket doesn't take in account safe creation, since it won't let you go more than the 1st step. Is this correct?

Safe creation was not touched.

francovenica:
> [Others were experiencing this yesterday](https://5afe.slack.com/archives/C049H57FS1E/p1679937059348629) and I noticed that Goerli was excessively slow. I am not sure whether that's related to the PR. If you experience it again, can you let us know.

I'll check other networks besides goerli to see if this keeps happening

> Edit: after investigation, this seems to be working. Which "incorrect" network were you originally on? I have the suspicion that this may be RPC related (if you were previously connected to Polygon, https://github.com/safe-global/web-core/issues/1748).

I set the device in ETH using it in a goerli safe. Same as before, I'll try to use other network than goerli.

francovenica:
Not necessarely a bug:
The warning message sometimes is red and sometimes is oragange. any reason for it?
Here we have the creation of a spending limit (red warning) and then when I'm trying to use the spending limit (orange warning)

image
image

francovenica:
Besides my previous comment about the colors the feature looks fine.

I didn't had the issue with trezor anymore 
I tried with MM, WC with MM phone app, Mobile pairing, Trezor one and Ledger nano S. 
I've checked Tx with immediate execution and queuing tx as well. 
I tried send funds, editing owners, using the tx builder, executing a spending limit....

LGTM

PR #1756, opened 2023-03-10
fix: Stay on the same route when switching safes

### LINKED ISSUE #1734: Stay on the same route when switching Safes
E.g. if you're on the Transaction Queue page and you switch the Safe via the Safe list, the new Safe should also be opened on the Transaction Queue page. Currently it goes to the Dashboard.

Beware that some Safe Apps don't exist on all chains, so staying on the same route might not work in that case.

usame-algan:
## What it solves

Resolves #1734 

Note: Some Safe apps are only available on certain networks. Switching to a safe on a network without that safe app can result in unexpected behaviour.

## How this PR fixes it

- Keeps the same `pathname` when switching safes except when the user is on a safe unspecific route like `/welcome`

## How to test it

1. Navigate to a Safe
2. Switch to a different Safe
3. Observe still being on the same route
4. Navigate to the welcome page
5. Open a Safe
6. Observe being on the Dashboard
7. Navigate to a single transaction
8. Switch to a different Safe
9. Observe the transaction being replaced with a loading screen

## Checklist


src/components/sidebar/SafeListItem/index.tsx
```diff
@@ -80,7 +82,13 @@ const SafeListItem = ({
         )
       }
     >
-      <Link href={{ pathname: AppRoutes.home, query: { safe: `${shortName}:${address}` } }} passHref>
+      <Link
+        href={{
+          pathname: !safeAddress ? AppRoutes.home : router.pathname,
```
iamacook, line 87: Why do we need to check for the presence of `safeAddress`? The Safe address (`address`) is always defined.

src/components/transactions/SingleTx/index.tsx
```diff
@@ -40,21 +40,17 @@ const SingleTx = () => {
...
-      return getTransactionDetails(safe.chainId, transactionId).then((details) => {
-        // If the transaction is not related to the current safe, throw an error
-        if (!sameAddress(details.safeAddress, safeAddress)) {
-          return Promise.reject(new Error('Transaction with this id was not found in this Safe'))
-        }
-        return details
-      })
-    },
-    [transactionId, safe.chainId, safe.txQueuedTag, safe.txHistoryTag, safeAddress],
-    false,
```
iamacook, line 56: Instead of clearing the data (causing a rerender if the tags change), I think we should create a Safe-specific flag along the lines of:

```ts
  const { prefix } = parsePrefixedAddress(router.query.safe)
  const shouldClearData = safe.address.value !== txDetails?.safeAddress || prefix !== chains[safe.chainId]
```

src/components/sidebar/SafeListItem/index.tsx
usame-algan: Thats how we make sure that we are not on a safe-unspecific route like `/welcome` since `safeAddress` is the one from the URL. Otherwise switching the Safe does nothing and the user is stuck on that page.

usame-algan: I just realized something like `/welcome?safe=xyz` would fall through this and get the user stuck as well. Maybe an array of routes makes more sense then.

src/components/transactions/SingleTx/index.tsx
usame-algan: After testing this again it seems to clear the transaction fine when switching safes so I reverted it.

src/components/sidebar/SafeListItem/index.tsx
usame-algan: Its only the `welcome` page where its necessary because its the only place where we render the `SafeList` without the `Sidebar`. I moved this logic up into the `SafeList` so we don't have to repeat it.

usame-algan: We would have to repeat the `isWelcomePage` condition which might change for `SafeList` in the future. We could only extract this condition and reuse it in both components but this way we also don't need the `router` dependency in every `SafeListItem`.

iamacook (approved):

francovenica:
Issues:
Maybe is a technical limitation, but I don't see a reason trying to load a single tx of a safe when switching to another, since it will never be successful and you will always see the "failed to load the tx" error message. Can we just switch it to the history tab by default?

There are apps that don't exist in all networks, but it still opens fine even if the safe you are switching it is from a network that doesn't support it
Example:
There is an app called Otoco that is in ETH but not in Poly. Try to open the app in a safe in eth and the switch to a safe in poly, the app will open no issue (and it should not)
Safe ETH: eth:0xE3706C25CAC381741c4A103755c2568e7a2Bde50
Safe Poly: matic:0x4E5a1650abFD4F9CaBa67f0BE91845541975C860
image

src/components/sidebar/SafeListItem/index.tsx
```diff
@@ -28,13 +28,15 @@ const SafeListItem = ({
   chainId,
   closeDrawer,
   shouldScrollToSafe,
+  destination,
```
katspaugh, line 31: `href` maybe?

src/components/sidebar/SafeList/index.tsx
```diff
@@ -75,6 +76,16 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
   const hasNoSafes = Object.keys(ownedSafes).length === 0 && Object.keys(addedSafes).length === 0
   const isWelcomePage = router.pathname === AppRoutes.welcome
 
+  const getDestination = useCallback(
```
katspaugh, line 79: Please add a comment along the lines of
```
On the Welcome page, link to the Dashboard; otherwise, stay on the same route
```

src/components/sidebar/SafeList/index.tsx
```diff
@@ -180,6 +195,7 @@ const SafeList = ({ closeDrawer }: { closeDrawer?: () => void }): ReactElement =
                       owners={safe.owners.length}
                       chainId={safe.chainId}
                       closeDrawer={closeDrawer}
+                      destination={{ pathname: router.pathname, query: router.query }}
```
katspaugh, line 198: Is it intended that this destination isn't doing the welcome vs non-welcome pages?

usame-algan: Yes, we render safes that are neither owned nor added separately. They are only visible in the sidebar when already open so clicking on it should essentially do nothing but close the sidebar.

usame-algan:
> Maybe is a technical limitation, but I don't see a reason trying to load a single tx of a safe when switching to another, since it will never be successful and you will always see the "failed to load the tx" error message. Can we just switch it to the history tab by default?

We can add an exception for this but I am not sure which route would make more sense, either History or Queue or maybe even just the Dashboard as the generic fallback. cc @katspaugh 

> There are apps that don't exist in all networks, but it still opens fine even if the safe you are switching it is from a network that doesn't support it

I noticed this too which made me think of not adding any rerouting for those cases since it might still show something. Some Safe apps even display a warning message when opening them on an unsupported network.

usame-algan:
> Or we could try to open the app in any case but redirect to the app's share URL if the app isn't supported on the chain.

I looked into this but with the current logic thats in place we can't show the share url on a network that is not supported. Imo this is a limitation that can be removed but I would suggest we open a separate issue for it as it requires quite a few changes.

francovenica:
LGTM. The redirection to the history tab is fine and we will take care of the apps in that other PR.

PR #1760, opened 2023-03-10
fix: Only store owned safes for their given chainId

### LINKED ISSUE #1759: Owned safes list displays wrong safes
## Bug description

Sometimes the safe list sidebar displays safes under the wrong network.

## Environment
 - Browser: any
 - Wallet: any
 - Chain: any

## Steps to reproduce
 1. Go to Safe
 2. Clear the `ownedSafes` local storage
 3. Connect a wallet that owns safes on different networks
 4. Open the safe list sidebar
 5. Observe owned safes are fetched and saved in the local storage
 6. Switch to a safe on a different network
 7. Observe the safe list sidebar closes
 8. Observe the local storage contains the safes from the previous network under the current network key
 9. Open the safe list sidebar again
 10. Observe that the local storage gets overriden with the correct entries

If the fetch request in Step 10 fails for whatever reason we end up with a corrupted state for `ownedSafes`

## Expected result

Owned safes are only set for their given network

## Obtained result

Owned safes are set for other networks

## Screenshots
Polygon showing 23 Safes that actually belong to Görli:

Screenshot 2023-03-10 at 14 25 34
Screenshot 2023-03-10 at 14 23 32

usame-algan:
## What it solves

Resolves #1759 

## How this PR fixes it

- Awaits the fetch of owned safes before setting the cache to avoid setting it for a stale state

## How to test it

1. Go to Safe
2. Clear the ownedSafes local storage
3. Connect a wallet that owns safes on different networks
4. Open the safe list sidebar
5. Observe owned safes are fetched and saved in the local storage
6. Switch to a safe on a different network
7. Observe the safe list sidebar closes
8. Observe the `ownedSafes` local storage doesn't update
9. Open the safe list sidebar again
10. Observe that the `ownedSafes` local storage gets overriden with the correct entries

## Checklist

### DIFF
--- a/src/hooks/useOwnedSafes.ts
+++ b/src/hooks/useOwnedSafes.ts
@@ -1,11 +1,11 @@
 import { useEffect } from 'react'
 import { getOwnedSafes, type OwnedSafes } from '@safe-global/safe-gateway-typescript-sdk'
 
-import useChainId from '@/hooks/useChainId'
 import useLocalStorage from '@/services/local-storage/useLocalStorage'
 import useWallet from '@/hooks/wallets/useWallet'
 import useAsync from './useAsync'
 import { Errors, logError } from '@/services/exceptions'
+import useSafeInfo from '@/hooks/useSafeInfo'
 
 const CACHE_KEY = 'ownedSafes'
 
@@ -16,26 +16,26 @@ type OwnedSafesCache = {
 }
 
 const useOwnedSafes = (): OwnedSafesCache['walletAddress'] => {
-  const chainId = useChainId()
+  const { safe } = useSafeInfo()
   const { address: walletAddress } = useWallet() || {}
   const [ownedSafesCache, setOwnedSafesCache] = useLocalStorage<OwnedSafesCache>(CACHE_KEY)
 
   const [ownedSafes, error] = useAsync<OwnedSafes>(() => {
-    if (!chainId || !walletAddress) return
-    return getOwnedSafes(chainId, walletAddress)
-  }, [chainId, walletAddress])
+    if (!safe.chainId || !walletAddress) return
+    return getOwnedSafes(safe.chainId, walletAddress)
+  }, [safe.chainId, walletAddress])
 
   useEffect(() => {
-    if (!ownedSafes || !walletAddress || !chainId) return
+    if (!ownedSafes || !walletAddress || !safe.chainId) return
 
     setOwnedSafesCache((prev) => ({
       ...prev,
       [walletAddress]: {
         ...(prev?.[walletAddress] || {}),
-        [chainId]: ownedSafes.safes,
+        [safe.chainId]: ownedSafes.safes,
       },
     }))
-  }, [ownedSafes, setOwnedSafesCache, walletAddress, chainId])
+  }, [ownedSafes, setOwnedSafesCache, walletAddress, safe.chainId])
 
   useEffect(() => {
     if (error) {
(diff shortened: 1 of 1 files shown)


iamacook (approved):

francovenica:
The "Owned safes" are not being fetched on the sidebar after connecting with an owner. This also happens after switching networks, where the sidebar should populate with the safes owned on that network

Try to connect with a safe that owns safes. See the sidebar remains empty:
image

usame-algan:
> Try to connect with a safe that owns safes. See the sidebar remains empty:

Nice catch, thanks for testing! It was because I was relying on the `SafeInfo` which doesn't exist on the welcome page. I reverted that change and fixed it another way.

src/hooks/useOwnedSafes.ts
```diff
@@ -16,26 +16,23 @@ type OwnedSafesCache = {
...
-    return getOwnedSafes(safe.chainId, walletAddress)
-  }, [safe.chainId, walletAddress])
+  const [_, error] = useAsync<void>(async () => {
+    if (!chainId || !walletAddress) return
 
-  useEffect(() => {
-    if (!ownedSafes || !walletAddress || !safe.chainId) return
+    const ownedSafes = await getOwnedSafes(chainId, walletAddress)
 
     setOwnedSafesCache((prev) => ({
```
katspaugh, line 28: Setting the state inside useAsync is IMHO an anti-pattern.
Is there another way to fix it?

src/hooks/useOwnedSafes.ts
```diff
@@ -16,26 +16,23 @@ type OwnedSafesCache = {
...
-  const { safe } = useSafeInfo()
+  const chainId = useChainId()
   const { address: walletAddress } = useWallet() || {}
   const [ownedSafesCache, setOwnedSafesCache] = useLocalStorage<OwnedSafesCache>(CACHE_KEY)
 
-  const [ownedSafes, error] = useAsync<OwnedSafes>(() => {
-    if (!safe.chainId || !walletAddress) return
-    return getOwnedSafes(safe.chainId, walletAddress)
-  }, [safe.chainId, walletAddress])
+  const [_, error] = useAsync<void>(async () => {
```
katspaugh, line 23: Syntax tip:
```suggestion
  const [, error] = useAsync<void>(async () => {
```
(Irrelevant tho, see my other comment).

src/hooks/useOwnedSafes.ts
katspaugh: What's the root cause of the bug? The chainId changes over time?

katspaugh: If you want to keep it in this imperative style, I would suggest just using an effect like so:
```
useEffect() => {
  getOwnedSafes(chainId).then(data => setOwnedSafesCache(prev => {
    ...
  }))
}, [chainId, etc])
```

usame-algan: The root cause is that the `Drawer` unmounts when switching to a safe on a different network but there is a brief period before it unmounts where the `chainId` already changed but the `ownedSafes` data is stale. It runs the `useEffect` hook which updates the local storage for the new `chainId` with the data from the previous network.

Alternatively, we could keep the `Drawer` mounted through the `keepMounted` prop that MUI offers which solves the issue but @iamacook raised a valid point that this would not update a newly created safe unless the user either reconnects their wallet or switches to a different chain.

Another thing we could do is to not use `useAsync` for this case to avoid this anti-pattern.

iamacook (approved):

katspaugh (approved):

PR #1772, opened 2023-03-14
Fix: tx-summary word-break only on mobile

katspaugh:
## What it solves
Randomly noticed this on prod:
Screenshot 2023-03-14 at 12 41 10

Breaking words like this seems to be unnecessary on desktop (and even on mobile, because we break the summary into several lines).

After the fix:
Screenshot 2023-03-14 at 12 39 01
Screenshot 2023-03-14 at 12 39 08
Screenshot 2023-03-14 at 12 39 34

### DIFF
--- a/src/components/transactions/TxSummary/styles.module.css
+++ b/src/components/transactions/TxSummary/styles.module.css
@@ -32,10 +32,13 @@
 
 .columnWrap {
   white-space: normal;
-  word-break: break-word;
 }
 
 @media (max-width: 600px) {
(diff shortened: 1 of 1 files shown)


iamacook (approved):

francovenica:
LGTM
For mobile I had to really shrink the screen, but the wordbreak is there
For desktop the words won't break anymore.
A few snapshots for comparison:
How it looks previous to the fix
image

How it looks now
image

How it looks now in a mobile view
image

PR #1773, opened 2023-03-14
refactor: simplify `isValidMasterCopy`

iamacook:
## What it solves

Simplifes `isValidMasterCopy`

## How this PR fixes it

We now check the `implementationVersionState` for the validity of a Safe's master copy. The gateway [internally checks whether the implementation address is that of a "supported" master copy](https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/safes/converters.rs#L77-L79) and [returns `UNKNOWN` as the `implementationVersionState` if it is not](https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/safes/handlers/safes.rs#L28-L31).

## How to test it

Open a Safe with an unsupported base contract (e.g. `matic:0x2B7a8C8afaEc010140970488190275A26b1436DD`) and observe that a notification is still shown.

## Screenshots

image

## Checklist

### DIFF
--- a/src/hooks/__tests__/useSafeNotifications.test.ts
+++ b/src/hooks/__tests__/useSafeNotifications.test.ts
@@ -1,9 +1,7 @@
-import { act } from '@testing-library/react'
 import { renderHook } from '@/tests//test-utils'
 import useSafeNotifications from '../../hooks/useSafeNotifications'
 import useSafeInfo from '../../hooks/useSafeInfo'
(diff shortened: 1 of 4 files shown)


katspaugh (approved):
Nice!

usame-algan (approved):
Nice find! LGTM 🚀

francovenica:
I've checked the the safe given in the description and a safe that Jago created also with an unsupported base contract and the error message shows for both.
LGTM

PR #1778, opened 2023-03-16
Fix: rm automatic safe-react data migration

### LINKED ISSUE #1728: Remove the automatic data migration from gnosis-safe.io
When we migrated to the new domain (app.safe.global), we had to migrate all the localStorage data from the old domain (see #165).

We should eventually remove that migration script. Provisional deadline: June 2023.

* [ ] Remove everything related to that migration (iframe loading, postMessage etc)
* [ ] Align with devops to remove migrate.html around the same date

katspaugh:
## What it solves
Resolves #1728

## How this PR fixes it
`safe-react` was fully deprecated and shut-down in the beginning of February, and web-core has been live since November last year.
At this point, I think the automatic migration rather hurts than helps anyone. If you decide to clear your localStorage (to start from scratch), it will populate it again with old data.

Once this is merged, I'll ask @luarx to remove the migrate.html file from the old domain.

## How to test
The automatic migration shouldn't happen anymore, but the manual Import All Data should still work.


iamacook (approved):

francovenica:
The data export still works fine.

I can't check the automatic migration since this PR is not set to pupulate with what there is in other envs.

PR #1784, opened 2023-03-20
Fix: strip query params from Sentry events

katspaugh:
## What it solves

Removes sensitive data such as the Safe address and tx ids from Sentry events.
So far they haven't been needed for debugging anyway.

### DIFF
--- a/src/services/sentry.ts
+++ b/src/services/sentry.ts
@@ -12,9 +12,15 @@ Sentry.init({
   ignoreErrors: ['Internal JSON-RPC error', 'JsonRpcEngine', 'Non-Error promise rejection captured with keys: code'],
 
   beforeSend: (event) => {
-    // Remove URL query params
-    if (event.request?.query_string) {
-      delete event.request?.query_string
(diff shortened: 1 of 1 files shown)


src/services/sentry.ts
```diff
@@ -10,6 +10,14 @@ Sentry.init({
   sampleRate: 0.1,
   // ignore MetaMask errors we don't control
   ignoreErrors: ['Internal JSON-RPC error', 'JsonRpcEngine', 'Non-Error promise rejection captured with keys: code'],
+
+  beforeSend: (event) => {
+    // Remove URL query params
+    if (event.request?.query_string) {
+      delete event.request?.query_string
```
iamacook, line 17:
```suggestion
      delete event.request.query_string
```

iamacook (commented):
What do you think about keeping the `appUrl`? Perhaps we should define our "standard" search params like we do with `AppRoutes` if so.

katspaugh:
Not easy to test w/o modifying the code, so I'll merge this directly.

PR #1793, opened 2023-03-22
Fix: sort Safe App tags

### LINKED ISSUE #1788: Safe App tags aren't sorted
The select needs to be sorted alphabetically.
An alternative could be that we display the number of apps in each category and sort by that. (@schmanu)

Screenshot 2023-03-20 at 14 47 22

katspaugh:
## What it solves

Resolves #1788

## How this PR fixes it

Sorts the tags alphabetically.

Screenshot 2023-03-22 at 08 19 37

## How to test it
Check that the categories select is sorted.

## Other changes
Unrelated to tags, I've also added tracking on the grid/list view toggle. There should be a new event when you click on one of these: 
Screenshot 2023-03-22 at 09 38 30
These buttons will now also remember the last view choice.

### DIFF
--- a/src/components/safe-apps/SafeAppCard/index.tsx
+++ b/src/components/safe-apps/SafeAppCard/index.tsx
@@ -159,7 +159,7 @@ const SafeAppCardListView = ({
 }: SafeAppCardViewProps) => {
   return (
     <SafeAppCardContainer safeAppUrl={safeAppUrl} onClickSafeApp={onClickSafeApp}>
-      <CardContent>
+      <CardContent sx={{ pb: '16px !important' }}>
         <Box display="flex" flexDirection="row" alignItems="center" gap={2}>
           <div className={css.safeAppIconContainer}>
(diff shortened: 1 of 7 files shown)


src/components/safe-apps/SafeAppCard/index.tsx
```diff
@@ -159,7 +159,7 @@ const SafeAppCardListView = ({
 }: SafeAppCardViewProps) => {
   return (
     <SafeAppCardContainer safeAppUrl={safeAppUrl} onClickSafeApp={onClickSafeApp}>
-      <CardContent>
+      <CardContent sx={{ pb: '16px !important' }}>
```
katspaugh, line 162: The default was 24px and it wasn't symmetrical.

Before:
Screenshot 2023-03-22 at 09 44 51

After:
Screenshot 2023-03-22 at 09 44 35

iamacook (approved):

PR #1797, opened 2023-03-23
test: e2e relayer

francovenica:
## What it solves
Adds relayer e2e test

## How this PR fixes it
Checks relayer for executable transactions
Checks grayed out inputs in adv parameters when the relayer is selected

## Checklist


cypress/e2e/create_tx.cy.js
```diff
@@ -77,6 +79,41 @@ describe('Queue a transaction on 1/N', () => {
     })
     cy.contains('Estimated fee').should('exist')
 
+    // Asserting the relayer option is present
+    cy.contains('Via relayer').find('[type="radio"]').should('be.checked')
+    cy.contains('Sponsored by Gnosis Chain')
+    cy.get('span').contains('Estimated fee').next().should('have.css', 'text-decoration-line', 'line-through')
+    cy.contains('Via relayer').contains(/[0-5] of 5/)
```
DiogoSoaress, line 86: Here i think we should assert the exact string "5 of 5" instead of a regex

cypress/e2e/create_tx.cy.js
```diff
@@ -77,6 +79,41 @@ describe('Queue a transaction on 1/N', () => {
...
+
+    cy.contains('Estimated fee').click()
+    cy.contains('Edit').click()
+    cy.contains('Owner transaction (Execution)').parents('form').as('Paramsform')
+
+    // Only gaslimit should be editable when the relayer is selected
+    const arrayNames = ['userNonce', 'maxPriorityFeePerGas', 'maxFeePerGas']
```
DiogoSoaress, line 93: Can we select the inputs from their label? I.e, "Wallet nonce", "Max priority fee (Gwei)" and "Max fee (Gwei)" ?

cypress/e2e/create_tx.cy.js
```diff
@@ -77,6 +79,41 @@ describe('Queue a transaction on 1/N', () => {
...
+    arrayNames.forEach((element) => {
+      cy.get('@Paramsform').find(`[name="${element}"]`).should('be.disabled')
+    })
+    cy.get('@Paramsform')
+      .find('[name="gasLimit"]')
+      .clear()
+      .type('300000')
+      .invoke('prop', 'value')
+      .should('equal', '300000')
+    cy.get('@Paramsform').find('[name="gasLimit"]').parent('div').find('[data-testid="RotateLeftIcon"]').click()
```
DiogoSoaress, line 103: Same as comment above. Would be preferred to select the input based of visible text: either the label or the placeholder.

cypress/e2e/create_tx.cy.js
```diff
@@ -103,10 +144,11 @@ describe('Queue a transaction on 1/N', () => {
       cy.get('input[type="checkbox"]').should('not.exist')
     })
 
-    cy.contains('Submit').click()
+    // We don't want to keep queing tx for now
+    //cy.contains('Submit').click()
```
DiogoSoaress, line 148: Maybe I'm shortsighted here but what is the problem of keep queuing?

cypress/e2e/create_tx.cy.js
```diff
@@ -103,10 +144,11 @@ describe('Queue a transaction on 1/N', () => {
...
   })
 
-  it('should click the notification and see the transaction queued', () => {
+  it.skip('should click the notification and see the transaction queued', () => {
```
DiogoSoaress, line 151: We should undo this change if are queueing the created transaction as this test validates the notification.

DiogoSoaress (changes requested):
Very nice improvement to the test in general!
I've left few comments and I also think we can keep signing the tx and testing the notification.
If this test is reliable let's move it to the `/smoke` folder :)

cypress/e2e/create_tx.cy.js
iamacook: This will otherwise drain our 1Balance.

cypress/e2e/create_tx.cy.js
francovenica: The "name" attribute is already in the inputs, so I don't have to navigate from the label to the input. Is easier to leave it as it is

cypress/e2e/create_tx.cy.js
francovenica: same as before

cypress/e2e/create_tx.cy.js
francovenica: Ok, I'll make it 5 of 5 strict

cypress/e2e/create_tx.cy.js
francovenica: I don't know if a safe has a ton of queued tx is stable (long to load or something like that). As long we keep this test out of the CI I can enable it again

cypress/e2e/create_tx.cy.js
francovenica: Since I disable the tx proposal I skipped the notification test as well. If we all agree to have tx keep being queued I will remove the skip

cypress/e2e/create_tx.cy.js
DiogoSoaress: @iamacook The test ends up just queueing a transaction, not relaying it.

cypress/e2e/create_tx.cy.js
DiogoSoaress: We've already had this discussion in other PRs and the approach we are taking is to target visible text whenever possible. The attribute is not something that is read in the interface.

DiogoSoaress: Please see https://github.com/safe-global/web-core/blob/568ea0db252647ecf1524daa8f03e82b0a16883c/cypress/e2e/add_owner.cy.js#L71-L91

cypress/e2e/create_tx.cy.js
iamacook: > @iamacook The test ends up just queueing a transaction, not relaying it.

Disregard my comment 🙈

cypress/e2e/create_tx.cy.js
francovenica: Ok, I used the labels

francovenica:
For some reason Prettier allows me to format the first "get" but it asks me to put the this second "get" all in one line 🤷.
Is there a way to tell prettier "dude, you are wrong"?
image

cypress/e2e/create_tx.cy.js
```diff
@@ -32,7 +34,7 @@ describe('Queue a transaction on 1/N', () => {
     cy.get('input[name="recipient"]').type(EOA)
     // Click on the Token selector
     cy.get('input[name="tokenAddress"]').prev().click()
-    cy.get('ul[role="listbox"]').contains('Görli Ether').click()
+    cy.get('ul[role="listbox"]').contains('Goerli Ether').click()
```
DiogoSoaress, line 37:
```suggestion
    cy.get('ul[role="listbox"]').contains(/G(ö|oe)rli Ether/).click()
```

DiogoSoaress (approved):
Nice! Please just apply the suggestion because Goerli is written differently in staging and prod configs

francovenica:
The tx proposal was being validated with that "sendValue" but it was always failing since we always used the "max" button during the test.
I fixed so now we are validating that the max button actually sets the max value asset the safe has, but later it sets the "sendValue" variable to validate the tx proposal.
Also I have to click in the "queue" tab since we change the behavior that clicking on "Transactions" in the sidebar now takes to the history tab instead

cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -3,8 +3,8 @@ const EOA = '0xE297437d6b53890cbf004e401F3acc67c8b39665'
 
 // generate number between 0.00001 and 0.00020
 const sendValue = Math.floor(Math.random() * 20 + 1) / 100000
-let recommendedNonce
 const currentNonce = 3
+let recommendedNonce
```
DiogoSoaress, line 7: Avoid unnecessary changes

cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -41,6 +41,25 @@ describe('Queue a transaction on 1/N', () => {
...
+          .parents('form')
+          .find('label')
+          .contains('Amount')
+          .next()
+          .find('input')
+          .should('have.value', maxBalance)
+          .clear()
+          .type(sendValue)
+      })
+
```
DiogoSoaress, line 62: I suggest to split this chunk in:
- Get max value to a variable
- Assert the input is correctly filled when clicking "Max"
- Clear the "Amount" input and type `sendValue`

DiogoSoaress (commented):
It seems to have fixed the CI job 👍 
I left some code suggestions

cypress/e2e/smoke/create_tx.cy.js
francovenica: The problem here is that any variable set inside the `then()` will lose its value outside of that scope, even if you declare it outside of the `then()` like this:
```
let variable = 0
cy.get('something').then((element)=>{
variable = element.text()})
cy.get('somethingElse).should('cointain', variable)
```
That "variable" will be 0 no matter what you assigned inside the then, so the navigation and assertion must be done inside the brackets.
I could clear and type outside of the then, but i'd have to navigate to the input again for no reason

katspaugh:
@DiogoSoaress should we merge this PR?

DiogoSoaress:
Yes, I've updated Franco's test to reflect the latest UI implementation. Needs a last review before merging but can definitely go into the epic branch

katspaugh (approved):

PR #1801, opened 2023-03-24
[e2e] Collapse expand all actions

francovenica:
## What it solves
Adds a test to check expand and collapse of all actions in a multiaction tx

Resolves #

## How this PR fixes it
Adds the test in the existing "tx_history" test. Nothing in the original test was touch, it is purely an addition

## How to test it
Run the test as it is. In the safe given for that test I added manually a tx with 3 actions

## Checklist


DiogoSoaress:
@francovenica the changed test is failing in the CI

cypress/e2e/smoke/tx_history.cy.js
```diff
@@ -131,4 +131,25 @@ describe('Transaction history', () => {
         cy.contains('span', 'Success').should('exist')
       })
   })
+
+  it('should expand/collapse all actions', () => {
+    // Open the tx details
+    cy.contains('div', 'Mar 23, 2023')
+      .next()
+      .click()
```
DiogoSoaress, line 139: Can you verify that before the click the elements are not visible? (Same as you do after clicking "Collapse all")

DiogoSoaress (approved):
🚀

PR #1811, opened 2023-03-27
[WIP] Feat: sign or batch

### LINKED ISSUE #1808: [Epic] Batch transactions before signing
**tl;dr:** allow the user combine multiple transactions into one Safe transaction before proposing it for others to sign.

[Epic doc:](https://www.notion.so/safe-global/Batched-transactions-43c5bab5a9d14a9d949832bac5b2fd9a) https://www.notion.so/safe-global/Batched-transactions-43c5bab5a9d14a9d949832bac5b2fd9a

**Designs:** tbd

## What problem are we solving?

In the Safe UI, it's possible to batch-execute Safe transactions, but it's not possible to **create** a multi-send Safe transaction consisting of Token Transfers, Settings changes, Safe App transactions etc. (It's only possible to batch-send NFTs atm).

Note: Transaction Builder allows creating complex multi-send transactions, but it's not part of the main UI.

## Why do we need batch transactions?
* They save time: you don't have to sign each individual transaction, instead you just sign one multi-send tx.
* They save money: no signatures in each tx means paying less gas
* They keep the queue clean: the batch serves as a personal tx basket, so that you can remove unwanted txs before submitting them to the queue
 * They showcase Safe's programmability as per the AA concept

## How will it look in the UI?
There will one and only one locally-stored batch queue. When creating any tx, the user will have the choice to either propose the tx immediately, or add it to his batch.

Designs and UI flows tbd, but as a quick draft:
Screenshot 2023-03-27 at 13 43 42

Screenshot 2023-03-27 at 13 43 56

### Can I delete transactions for a batch?
Yes.

### What if I create a batch that has mutually exclusive transactions?
Yes, the whole batch will fail. See https://github.com/safe-global/web-core/issues/1130

katspaugh:
## What it solves

Resolves #1808

## How this PR fixes it

Basically, each Safe can now have a single batch tx associated with it.
It's an actual multisend tx saved on the backend, but not signed. We store its id in a persisted Redux slice.

When the user adds another tx to the batch, the multi-send transaction is decoded, the new transaction is added to it, and the batch tx is "proposed" again. The batch's txId is replaced with the new txId in the store.

On the Queue page, we fetch the untrusted transactions list, find the batch tx there, and display it on top of the normal queue.

* [x] A Redux slice that holds batch tx ids per chain and safe address
* [x] A hook that allows to get an existing batch and add transactions to it
* [x] Display the batch in the queue
* [x] Show a radio group in `<SignOrExecuteForm>` (should be now renamed to `BatchOrSignOrExecuteForm` lol)
* [x] Make `<DecodedTx>` display multi-send txs with decoded sub-transactions
* [ ] Allow removing transactions from the batch
* [ ] Align with the designer

## Screenshots
Sign or batch:
Screenshot 2023-03-27 at 18 41 17

The batch on top of the queue:
Screenshot 2023-03-27 at 18 32 40

### DIFF
--- a/cypress/e2e/smoke/create_tx.cy.js
+++ b/cypress/e2e/smoke/create_tx.cy.js
@@ -102,7 +102,6 @@ describe('Queue a transaction on 1/N', () => {
 
     // Asserting the sponsored info is present
     cy.contains('Sponsored by').should('be.visible')
-    cy.contains('Gnosis Chain').should('be.visible')
 
     cy.get('span').contains('Estimated fee').next().should('have.css', 'text-decoration-line', 'line-through')
     cy.contains('Transactions per hour')
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "homepage": "https://github.com/safe-global/web-core",
   "license": "GPL-3.0",
   "type": "module",
-  "version": "1.8.1",
+  "version": "1.9.0",
   "scripts": {
     "dev": "next dev",
     "start": "next dev",
--- a/src/components/dashboard/Relaying/index.tsx
+++ b/src/components/dashboard/Relaying/index.tsx
@@ -8,10 +8,13 @@ import GasStationIcon from '@/public/images/common/gas-station.svg'
 import ExternalLink from '@/components/common/ExternalLink'
 import classnames from 'classnames'
 import css from './styles.module.css'
+import { useCurrentChain } from '@/hooks/useChains'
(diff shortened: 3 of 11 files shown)


schmanu:
When adding a `multiSend` to the existing batch, we have to "unwrap" it and add the inner transactions to the batch.
Otherwise we will have a `multiSend` which tries to do a `delegateCall` to the `MultiSendContract` as one of the actions which is not possible.

Screenshot 2023-05-12 at 08 53 38

katspaugh:
Good catch, Manu!

PR #1812, opened 2023-03-28
fix: Dismiss update safe notification for 90 days

### LINKED ISSUE #1807: [Safe upgrade] Snooze upgrade notification for 90 days
If the user closed the upgrade notification, remember it and don't show it next time for that Safe unless 90 days passed.

usame-algan:
## What it solves

Resolves #1807 

## How this PR fixes it

- Stores a flag in localStorage with an expiry date of 90 days when the user closes the update safe notification

## How to test it

1. Open a Safe with an older version
2. Observe the update safe notification
3. Dismiss the notification by either pressing "x" in the corner or the link
4. Observe a new localStorage entry in the format of `SAFE_v2__dismissUpdateSafe_<chainId>:<safeAddress>` that contains a timestamp for 90 days in the future
5. Manually change the timestamp to one in the past e.g. by using `Date.now()`
6. Reload the page
7. Observe the localStorage entry is deleted
8. Observe the update safe notification showing up

## Checklist

### DIFF
--- a/src/components/common/Notifications/index.tsx
+++ b/src/components/common/Notifications/index.tsx
@@ -6,12 +6,16 @@ import type { Notification } from '@/store/notificationsSlice'
 import { closeNotification, readNotification, selectNotifications } from '@/store/notificationsSlice'
 import type { AlertColor, SnackbarCloseReason } from '@mui/material'
 import { Alert, Link, Snackbar } from '@mui/material'
+import { type SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import css from './styles.module.css'
 import NextLink from 'next/link'
 import ChevronRightIcon from '@mui/icons-material/ChevronRight'
 import { OVERVIEW_EVENTS } from '@/services/analytics/events/overview'
 import Track from '../Track'
 import { isRelativeUrl } from '@/utils/url'
+import local from '@/services/local-storage/local'
+import useSafeInfo from '@/hooks/useSafeInfo'
+import { DISMISS_NOTIFICATION_KEY, isUpdateSafeNotification } from '@/hooks/useSafeNotifications'
 
 const toastStyle = { position: 'static', margin: 1 }
 
@@ -91,17 +95,45 @@ const getVisibleNotifications = (notifications: Notification[]) => {
   return notifications.filter((notification) => !notification.isDismissed)
 }
 
+export type DismissedUpdateNotifications = {
+  [chainId: string]: {
+    [address: string]: number
+  }
+}
+
+const handleDismissUpdateSafeNotification = (notification: Notification, safe: SafeInfo) => {
+  const EXPIRY_DAYS = 90
+
+  if (!isUpdateSafeNotification(notification.groupKey)) return
+
+  const dismissedNotifications = local.getItem<DismissedUpdateNotifications>(DISMISS_NOTIFICATION_KEY)
+
+  const expiryDate = Date.now() + EXPIRY_DAYS * 24 * 60 * 60 * 1000
+
+  const newState = {
+    ...dismissedNotifications,
+    [safe.chainId]: {
+      ...dismissedNotifications?.[safe.chainId],
+      [safe.address.value]: expiryDate,
+    },
+  }
+
+  local.setItem<DismissedUpdateNotifications>(DISMISS_NOTIFICATION_KEY, newState)
+}
+
 const Notifications = (): ReactElement | null => {
   const notifications = useAppSelector(selectNotifications)
   const dispatch = useAppDispatch()
+  const { safe } = useSafeInfo()
 
   const visible = getVisibleNotifications(notifications)
 
   const handleClose = useCallback(
     (item: Notification) => {
       dispatch(closeNotification(item))
+      handleDismissUpdateSafeNotification(item, safe)
     },
-    [dispatch],
+    [dispatch, safe],
   )
 
   // Close previous notifications in the same group
--- a/src/components/common/Notifications/index.tsx
+++ b/src/components/common/Notifications/index.tsx
@@ -91,12 +91,6 @@ const getVisibleNotifications = (notifications: Notification[]) => {
   return notifications.filter((notification) => !notification.isDismissed)
 }
 
-export type DismissedUpdateNotifications = {
-  [chainId: string]: {
-    [address: string]: number
-  }
-}
-
 const Notifications = (): ReactElement | null => {
   const notifications = useAppSelector(selectNotifications)
   const dispatch = useAppDispatch()
--- a/src/components/common/Notifications/index.tsx
+++ b/src/components/common/Notifications/index.tsx
@@ -13,9 +13,9 @@ import ChevronRightIcon from '@mui/icons-material/ChevronRight'
 import { OVERVIEW_EVENTS } from '@/services/analytics/events/overview'
 import Track from '../Track'
 import { isRelativeUrl } from '@/utils/url'
-import local from '@/services/local-storage/local'
 import useSafeInfo from '@/hooks/useSafeInfo'
 import { DISMISS_NOTIFICATION_KEY, isUpdateSafeNotification } from '@/hooks/useSafeNotifications'
+import useLocalStorage from '@/services/local-storage/useLocalStorage'
 
 const toastStyle = { position: 'static', margin: 1 }
 
@@ -101,12 +101,15 @@ export type DismissedUpdateNotifications = {
   }
 }
 
-const handleDismissUpdateSafeNotification = (notification: Notification, safe: SafeInfo) => {
+const handleDismissUpdateSafeNotification = (
+  notification: Notification,
+  safe: SafeInfo,
+  dismissedNotifications: ReturnType<typeof useLocalStorage<DismissedUpdateNotifications>>[0],
+  setDismissedNotifications: ReturnType<typeof useLocalStorage<DismissedUpdateNotifications>>[1],
+) => {
   const EXPIRY_DAYS = 90
 
   if (!isUpdateSafeNotification(notification.groupKey)) return
-
-  const dismissedNotifications = local.getItem<DismissedUpdateNotifications>(DISMISS_NOTIFICATION_KEY) || {}
 
   const expiryDate = Date.now() + EXPIRY_DAYS * 24 * 60 * 60 * 1000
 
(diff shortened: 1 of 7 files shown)


src/hooks/useSafeNotifications.ts
```diff
@@ -35,16 +36,18 @@ const useSafeNotifications = (): void => {
     if (!isOwner) return
     if (implementationVersionState !== ImplementationVersionState.OUTDATED) return
 
+    const isDismissed = local.getWithExpiry<boolean>(DISMISS_NOTIFICATION_KEY + '_' + chainId + ':' + safeAddress)
```
iamacook, line 39: This key is very specific and is also used in the middleware. I'd suggest creating a function for getting it.

I also don't think you should dispatch it if it's dismissed. It might be worth separating these two into independent effects.

iamacook: Nit: can you rename this to `isUpdatedDismissed` as this shouldn't relate to the unsupported notifications.

src/store/notificationsSlice.ts
```diff
@@ -72,3 +74,34 @@ export const showNotification = (payload: Omit<Notification, 'id' | 'timestamp'>
...
+export const notificationsMiddleware: Middleware<{}, RootState> = (store) => (next) => (action) => {
+  const result = next(action)
+
+  switch (action.type) {
+    case notificationsSlice.actions.closeNotification.type: {
+      const state = store.getState()
+      const safe = selectSafeInfo(state)
+
+      const { payload } = action as ReturnType<typeof notificationsSlice.actions.closeNotification>
+      const isUpdateSafeNotification = payload.groupKey === 'safe-outdated-version'
```
iamacook, line 89: I'd suggest you export this from the notification file.

src/store/notificationsSlice.ts
```diff
@@ -72,3 +74,34 @@ export const showNotification = (payload: Omit<Notification, 'id' | 'timestamp'>
...
+
+      if (!isUpdateSafeNotification || !safe.data) return
+
+      const dismissedKeyForSafe = DISMISS_NOTIFICATION_KEY + '_' + safe.data.chainId + ':' + safe.data.address.value
+
+      const isAlreadyDismissed = local.getWithExpiry<boolean>(dismissedKeyForSafe)
+
+      if (isAlreadyDismissed) return
+
+      const EXPIRY_DAYS = 90
```
iamacook, line 99: Nit: can we move this to the top of the `case`?

iamacook (commented):
A middleware seems a bit excessive and the notification is still shown in the notification centre. I wouldn't dispatch it during the "snooze" period.

I'd suggest keeping everything in the hook as the `id`, etc. is already there/it centralises the dispatch logic. I know it's unlikely to have multiple Safes requiring update but you could alternatively save a hashmap of timestamps to prevent polluting the `localStorage` as well.

What do you think?

src/hooks/useSafeNotifications.ts
usame-algan: > This key is very specific and is also used in the middleware. I'd suggest creating a function for getting it.
Good idea, I will extract it into a function.

> I also don't think you should dispatch it if it's dismissed.
It should be easy to change that however we decide. I don't think having the notification inside the center does any harm but we could also not dispatch it at all.

usame-algan:
> I know it's unlikely to have multiple Safes requiring update but you could alternatively save a hashmap of timestamps to prevent polluting the localStorage as well.

I tried this solution first but decided against it because it requires us to reimplement the logic that `set/getWithExpiry` already does. In addition, we would have to manage the data in a structure like `[chainId]: Array<{address, expiry}>`, find the correct element, manually check the expiry, update the whole object and save it again which seems too much considering the number of users who own multiple safes and how many of those are outdated versions.

Alternatively, we could extend the logic of `get/setWithExpiry` to support nested data structures but this would also add more code we have to manage which I think is not necessary for this use-case.

iamacook:
> I tried this solution first but decided against it because it requires us to reimplement the logic that `set/getWithExpiry` already does.

True, but we currently aren't using `set/getWithExpiry` in the app at all. The hashmap could just remain and we update it when necessary.

> In addition, we would have to manage the data in a structure like `[chainId]: Array<{address, expiry}>`, find the correct element, manually check the expiry, update the whole object and save it again

This is what the current middleware is doing. The data structure needn't use an array, e.g.

```ts
type UpdateNotifications = {
  [chainId: string]: {
    [address: string]: number
  }
}
```

> which seems too much considering the number of users who own multiple safes and how many of those are outdated versions.

True that users currently won't have outdated Safes but consider we will soon be releasing new contracts.

src/store/notificationsSlice.ts
```diff
@@ -26,7 +26,7 @@ export const notificationsSlice = createSlice({
     enqueueNotification: (state, { payload }: PayloadAction<Notification>): NotificationState => {
       return [...state, payload]
     },
-    closeNotification: (state, { payload }: PayloadAction<{ id: string }>): NotificationState => {
+    closeNotification: (state, { payload }: PayloadAction<Notification>): NotificationState => {
```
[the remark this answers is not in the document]
usame-algan: Good catch! I will revert it

src/components/common/Notifications/index.tsx
```diff
@@ -91,17 +95,45 @@ const getVisibleNotifications = (notifications: Notification[]) => {
...
   const notifications = useAppSelector(selectNotifications)
   const dispatch = useAppDispatch()
+  const { safe } = useSafeInfo()
 
   const visible = getVisibleNotifications(notifications)
 
   const handleClose = useCallback(
     (item: Notification) => {
       dispatch(closeNotification(item))
+      handleDismissUpdateSafeNotification(item, safe)
```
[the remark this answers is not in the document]
usame-algan: I haven't found a way how we can distinguish between a notification that was closed automatically (via the cleanup hook) or explicitely by the user. We would have to pass additional metadata in the `onClose` call to determine that which is also not ideal. Imo having it explicitely in the `onClose` vs. a side effect is clearer even if it adds to the component.

src/components/common/Notifications/index.tsx
```diff
@@ -91,17 +95,45 @@ const getVisibleNotifications = (notifications: Notification[]) => {
...
+    [address: string]: number
+  }
+}
+
+const handleDismissUpdateSafeNotification = (notification: Notification, safe: SafeInfo) => {
+  const EXPIRY_DAYS = 90
+
+  if (!isUpdateSafeNotification(notification.groupKey)) return
+
+  const dismissedNotifications = local.getItem<DismissedUpdateNotifications>(DISMISS_NOTIFICATION_KEY)
```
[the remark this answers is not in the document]
usame-algan: I just tested it and its not throwing when trying to spread `undefined` or `null` but I agree lets add it as a fallback so its clearer.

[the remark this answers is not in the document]
iamacook: Wow, you learn something new every day.

image

src/components/common/Notifications/index.tsx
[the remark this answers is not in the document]
iamacook: What about extending the `Notification` type to have an `onClose` callback we conditionally call here instead? It's more versatile.

It's not best practice to put non-serialisable code in the Redux store but as we don't persist the notifications, this shouldn't be an issue. We can suppress the warning by setting the `serializableCheck` to `false`: `getDefaultMiddleware({ serializableCheck: false })`.

src/store/notificationsSlice.ts
usame-algan: I moved it into the hook since the constant is being used there as well.

src/components/common/Notifications/index.tsx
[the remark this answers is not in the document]
usame-algan: > What about extending the Notification type to have an onClose callback we conditionally call here instead? It's more versatile.

Seems cleaner to me and keeps all the logic in the appropriate hook lets go with it.

src/hooks/useSafeNotifications.ts
```diff
@@ -35,12 +44,24 @@ const useSafeNotifications = (): void => {
     if (!isOwner) return
     if (implementationVersionState !== ImplementationVersionState.OUTDATED) return
 
+    const dismissedNotifications = local.getItem<DismissedUpdateNotifications>(DISMISS_NOTIFICATION_KEY)
+    const dismissedNotificationTimestamp = dismissedNotifications?.[chainId]?.[safeAddress]
+
+    if (dismissedNotificationTimestamp) {
+      if (Date.now() >= dismissedNotificationTimestamp) {
+        delete dismissedNotifications?.[chainId]?.[safeAddress]
+        local.setItem<DismissedUpdateNotifications>(DISMISS_NOTIFICATION_KEY, dismissedNotifications)
```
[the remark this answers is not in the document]
usame-algan: done

src/components/common/Notifications/index.tsx
```diff
@@ -91,6 +91,12 @@ const getVisibleNotifications = (notifications: Notification[]) => {
   return notifications.filter((notification) => !notification.isDismissed)
 }
 
+export type DismissedUpdateNotifications = {
```
iamacook, line 94: Let's move this to where it's used.

src/hooks/useSafeNotifications.ts
```diff
@@ -9,23 +9,55 @@ import { useRouter } from 'next/router'
...
 
 const CLI_LINK = {
   href: 'https://github.com/5afe/safe-cli',
   title: 'Get CLI',
 }
 
+export const DISMISS_NOTIFICATION_KEY = 'dismissUpdateSafe'
+export const OUTDATED_VERSION_KEY = 'safe-outdated-version'
+
+export const isUpdateSafeNotification = (groupKey: string) => {
```
iamacook, line 23: These need not be exported.

src/hooks/useSafeNotifications.ts
```diff
@@ -9,23 +9,55 @@ import { useRouter } from 'next/router'
...
+  return groupKey === OUTDATED_VERSION_KEY
+}
+
 /**
  * General-purpose notifications relating to the entire Safe
  */
 const useSafeNotifications = (): void => {
+  const [dismissedNotifications, setDismissedNotifications] =
```
iamacook, line 31: Nit:

```suggestion
  const [dismissedUpdateNotifications, setDismissedUpdateNotifications] =
```

iamacook (approved):

francovenica:
LGTM
The date of the value in the LS is 3 months for now
The notification doesn't show up anymore after clearing it
Editing the value to a past date or removing the value show the notification again

Question:
I don't have that many safes to update so I don't wanna update mine. The question is, if I update the safe is the key removed from the LS?

usame-algan:
> I don't have that many safes to update so I don't wanna update mine. The question is, if I update the safe is the key removed from the LS?

Good catch! I adjusted the hook so that it always checks the dismissUpdate flag first before checking if a safe version is outdated. That should clean up the value even after a safe has been updated.

francovenica:
Nice, thanks for the fix. 

LGTM

PR #1815, opened 2023-03-28
feat: Relay safe creation

### LINKED ISSUE #1802: [Relaying V1.1] Safe creation flow
As a user
I want to create Safes without paying network fees
because I want to enjoy a holistic gasless experience

Background
In the V1 execution flow of relaying the Safe creation was not included due to ease of implementation and unclear technical feasibility on the backend side to limit the Safe creation to prevent users from exploiting the system. The research has been done and to make the Safe{Wallet} experience more easily accessible for the users we agreed on incl. Safe creation into the gasless experience as well.
Safe creation is supported by the BE service and is documented in this issue https://github.com/safe-global/safe-gelato-relay-service/issues/34

What needs to be done
Implement/adjust the screens for Safe creation according to the designs in Figma https://www.figma.com/file/M3J3IJ5YFxRKaT95F2uBAt/Relaying?node-id=1559-35730&t=3sfF37KCJZJMDZGO-4

Acceptance criteria
Screens implemented according to the designs

DiogoSoaress:
## What it solves

Resolves #1802

## How this PR fixes it
Implements relaying for Safe creation

## Todo:

- [x] Improve test coverage

## How to test it
1. Create a Safe in Goerli
2. In the review step, if all the Safe owners have available relaying transaction, a component with that info is diplayed
3. On clicking "Next", the Safe creation will be relayed.
4. The transaction will be monitored until being indexed.

## Screenshots
Screenshot 2023-03-28 at 23 05 45

## Checklist


katspaugh:
Design feedback:
* I suggest we put the estimated fee _before_ the "sponsored by" block, and cross out the price
Screenshot 2023-03-29 at 19 58 04
* Remove the "You will have to confirm a transaction with your connected wallet." text. You don't have to confirm anything with the relayer.

katspaugh:
Or even like this:
Screenshot 2023-03-29 at 20 01 37
Since we don't offer the choice to execute with the wallet anyway.
@liliiaorlenko wdyt?

src/services/tx/txMonitor.ts
```diff
@@ -106,38 +111,45 @@ export const waitForRelayedTx = (taskId: string, txId: string): void => {
...
+            error: new Error(`Relayed transaction was blacklisted by relay provider.`),
+          })
         clearTimeout(failAfterTimeoutId)
         return
       case TaskState.Cancelled:
-        txDispatch(TxEvent.FAILED, {
-          txId,
-          error: new Error(`Relayed transaction was cancelled by relay provider.`),
-        })
+        txId &&
```
iamacook, line 138: Need we even call `waitForRelayedTx` if no `txId` exists in the first place?

Nit: I don't think `this && notation` is very clear, especially when it spreads multiple lines (personal taste though so feel free to ignore).

src/components/new-safe/create/logic/index.ts
```diff
@@ -263,3 +265,52 @@ export const getRedirect = (
...
+    threshold,
+    to: ZERO_ADDRESS,
+    data: EMPTY_DATA,
+    fallbackHandler: fallbackHandlerAddress,
+    paymentToken: ZERO_ADDRESS,
+    payment: 0,
+    paymentReceiver: ZERO_ADDRESS,
+  }
+
+  const initializer = Gnosis_safe__factory.createInterface().encodeFunctionData('setup', [
```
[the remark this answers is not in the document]
iamacook: > it seems we will need to call the factory class to call createInterface().

`createInterface` is a method from `typechain`. You can access the `ethers`' interface directly via `.contract.interface` (see line 303).

> We are only supporting L1 at the moment, is it worth creating an util to conditionally return the L1 or L2 factory contract?

Yes, I think we best adhere to a SSOT. Although Gnosis Chain supports both (afaiu), it is [marked as an L2](https://safe-client.safe.global/v1/chains/100) by the gateway.

src/components/new-safe/create/steps/StatusStep/useSafeCreation.ts
```diff
@@ -61,34 +64,65 @@ export const useSafeCreation = (
...
-      setStatus(status)
-
-      if (status !== SafeCreationStatus.SUCCESS) {
-        dispatch(showSafeCreationError(_err))
+    const { owners, threshold, saltNonce } = pendingSafe
+    const ownersAddresses = owners.map((owner) => owner.address)
+
+    if (willRelay) {
+      try {
+        const taskId = await createNewSafeViaRelayer(chain, ownersAddresses, threshold, saltNonce)
```
usame-algan, line 72: Is it possible to move the relayer logic into a separate function and doing the check inside the `useEffect` since they don't share most of the dependencies (relaying doesn't need a wallet or a provider) e.g.

```
const createSafeViaRelayer = () => {}
const createSafe = () => {}

useEffect(() => {
    if (willRelay) {
        void createSafeViaRelayer()
    } else {
        void createSafe()
    }
}
```

src/components/new-safe/create/steps/StatusStep/StatusStepper.tsx
```diff
@@ -28,7 +28,7 @@ const StatusStepper = ({ pendingSafe, status }: { pendingSafe: PendingSafeData;
         </StatusStep>
       </Step>
       <Step>
-        <StatusStep isLoading={!pendingSafe.txHash} safeAddress={pendingSafe.safeAddress}>
+        <StatusStep isLoading={!pendingSafe.txHash && !pendingSafe.taskId} safeAddress={pendingSafe.safeAddress}>
```
usame-algan, line 31: Wouldn't this break normal safe creations since there is never a `taskId`? Should it be `||` instead?

src/components/new-safe/create/logic/index.ts
```diff
@@ -263,3 +265,52 @@ export const getRedirect = (
   const appendChar = hasQueryParams ? '&' : '?'
   return redirectUrl + `${appendChar}safe=${address}`
 }
+
+export const createNewSafeViaRelayer = async (
```
usame-algan, line 269: Is it possible to use the core-sdk for part of this function i.e. to get the call data?

usame-algan (commented):
Very smooth UX, well done! I have a few comments

src/hooks/useRemainingRelays.ts
```diff
@@ -27,4 +28,16 @@ const useRemainingRelays = () => {
   }, [chainId, safeAddress])
 }
 
-export default useRemainingRelays
+export const useLeastRemainingRelays = (ownerAddresses: string[]) => {
```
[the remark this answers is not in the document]
DiogoSoaress: I've done as you suggested and updated the tests. Works nicely and saves a lot of network requests indeed

src/services/tx/txMonitor.ts
DiogoSoaress: > Need we even call waitForRelayedTx if no txId exists in the first place?

Yes. For when you're relaying the Safe creation.

I agree the pattern is not very elegant. The reason why I has to use it was to reuse the `waitForRelayedTx` to monitor both Safe creation and the modal execution. Any suggestion to improve it here?

src/hooks/useRemainingRelays.ts
```diff
@@ -27,4 +28,16 @@ const useRemainingRelays = () => {
...
+  const chainId = useChainId()
+
+  const getMinimum = (result: number[]) => Math.min(...result)
```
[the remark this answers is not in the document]
DiogoSoaress: Done

src/hooks/useRemainingRelays.ts
```diff
@@ -27,4 +28,16 @@ const useRemainingRelays = () => {
...
+
+  return useAsync(
+    () =>
+      Promise.all(ownerAddresses.map((address) => fetchRemainingRelays(chainId, address))).then((result) => {
+        return getMinimum(result)
+      }),
+    [chainId, ownerAddresses],
+  )
```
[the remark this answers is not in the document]
DiogoSoaress: I've changed it to `async/await` syntax ser

src/services/tx/txMonitor.ts
iamacook: > The reason why I has to use it was to reuse the waitForRelayedTx to monitor both Safe creation and the modal execution. Any suggestion to improve it here?

As the dispatchers play no role and you are only setting the status on `TaskState.ExecSuccess` or `TaskState.Cancelled`, I would not try to reuse this function. You can then avoid the conditional logic.

Not everything needs to be duplicated, e.g. you can extract `checkTxStatus`.

src/components/new-safe/create/steps/StatusStep/useSafeCreation.ts
```diff
@@ -61,34 +64,65 @@ export const useSafeCreation = (
...
+
+        if (!taskId) {
+          throw new Error('Transaction could not be relayed')
+        }
```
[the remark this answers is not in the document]
DiogoSoaress: You're right, if `sponsoredCall` throws, this condition is never run. Will remove it

src/components/new-safe/create/steps/StatusStep/StatusStepper.tsx
DiogoSoaress: Good catch. It was working well for relaying but would break in the normal Safe creations. Fixed it.

src/components/new-safe/create/logic/index.ts
DiogoSoaress: I've checked it but the calls to `encodeSetupCallData` are private that's why I had to call it this way.

src/hooks/useRemainingRelays.ts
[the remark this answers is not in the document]
usame-algan: There is a dangling console.log and the type of result says `any[]`. It can be fixed by adding a return type to `fetchRemainingRelays`

src/services/tx/txMonitor.ts
iamacook: Resolving as we refactored this together.

src/components/new-safe/create/logic/index.ts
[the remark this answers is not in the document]
iamacook: Resolving as we refactored this together.

iamacook (approved):
sans `console.log`, LGTM

PR #1816, opened 2023-03-28
feat: import/export data

### LINKED ISSUE #1147: Implement version 2.0 data export/import

iamacook:
## What it solves

Resolves #1147

## How this PR fixes it

The data settings have been extended to include data export from the new interface, as well as a new "widget" in the sidebar of the welcome page.

### Data export/import via Safe
1. Open the data settings of a Safe and export.
2. Clear the `localStorage` and refresh.
3. Import the exported data and observe that the interface is as it was prior to clear..

### Data export/import via welcome page
1. Open the welcome page and export.
2. Clear the `localStorage` and refresh.
3. Import the exported data and observe that the interface is as it was prior to clear..

### Data import via welcome page
Note: ensure you have a backup!

1. Clear the `localStorage` and refresh.
2. Open the welcome page and import.
3. Observe that the interface is as it was prior to clear.

All of the above should work with the "legacy" v1 export JSON files as well.

## Screenshots

export

import

import2(1)

image

image

image

image

image

image

## Checklist


src/components/settings/DataManagement/FileListCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+            ),
+            secondary: showPreview ? getItemSecondaryText(chains, addedSafes, 'Safe', 'Safes') : undefined,
+          },
+        ]
+      : []),
+    ...(addressBookChainAmount > 0
+      ? [
+          {
+            primary: (
+              <>
```
usame-algan, line 97: This is pretty hard to read imo. I would suggest changing it to something like:

```
const items = []

if (addedSafeChainAmount > 0) {
    const addedSafes = {primary: ..., secondary: ...}
    items.push(addedSafes)
}
...
```

src/components/settings/DataManagement/ImportDialog.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  return (
+    <ModalDialog open onClose={handleClose} dialogTitle="Data import" hideChainIndicator>
+      <DialogContent>
+        {!jsonData || !fileName ? (
+          <Box mt={2}>
+            <ImportFileUpload setFileName={setFileName} setJsonData={setJsonData} />
+          </Box>
+        ) : (
```
usame-algan, line 88: What are the steps to see this component inside the Modal? From my tests so far I only ever get into the other case e.g. when dropping a .json into the field.

src/components/settings/DataManagement/FileListCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+  settings,
+  safeApps,
+  error,
+  showPreview = false,
+  ...cardHeaderProps
+}: Props): ReactElement => {
+  const chains = useChains()
+  const items = useMemo<Array<ListItemTextProps>>(() => {
+    return getItems({ addedSafes, addressBook, settings, safeApps, error, chains: chains.configs, showPreview })
+  }, [addedSafes, addressBook, settings, safeApps, error, chains.configs, showPreview])
```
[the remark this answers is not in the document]
iamacook: Theoretically, each shouldn't change because you cannot alter them from the settings pane and `chains` is only loaded when the app opens.

src/components/settings/DataManagement/ImportDialog.tsx
iamacook: If you click "Import" from the welcome page, it should open the modal with the dropzone. (See recording in the description.)

src/components/settings/DataManagement/FileListCard.tsx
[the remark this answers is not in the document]
usame-algan: In that case I think `useMemo` is not needed.

src/components/settings/DataManagement/index.tsx
```diff
@@ -1,40 +1,124 @@
...
+import { FileListCard } from '@/components/settings/DataManagement/FileListCard'
+
+import css from './styles.module.css'
+
+const getExportFileName = () => {
+  const today = new Date().toISOString().slice(0, 10)
+  return `safe-${today}.json`
+}
+
+export const handleExport = () => {
```
usame-algan, line 25: nit: I would give this a more specific name since its being exported. Maybe something like `exportAppData`?

src/components/settings/DataManagement/useGlobalImportFileParser.ts
```diff
@@ -0,0 +1,132 @@
...
+  Object.values(data).reduce<number>((count, entry) => count + Object.keys(entry).length, 0)
+
+export const _filterValidAbEntries = (ab?: AddressBookState): AddressBookState | undefined => {
+  if (!ab) {
+    return undefined
+  }
+
+  return Object.entries(ab).reduce<AddressBookState>((acc, [chainId, chainAb]) => {
+    const sanitizedChainAb = Object.entries(chainAb).reduce<AddressBook>((acc, [address, name]) => {
+      if (name?.trim() && address && isChecksummedAddress(address)) {
```
usame-algan, line 34: Is it possible for `name` or `address` to be undefined? If so we should leave a comment why it is not reflected in the type.

src/services/analytics/events/settings.ts
```diff
@@ -79,8 +79,8 @@ export const SETTINGS_EVENTS = {
     },
   },
   DATA: {
-    IMPORT_ALL_BUTTON: {
-      action: 'Import all data button clicked',
+    EXPORT_ALL_BUTTON: {
```
usame-algan, line 82: This doesn't seem to be used yet

src/components/settings/DataManagement/FileListCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+    <Card className={css.card}>
+      <CardHeader {...cardHeaderProps} className={css.header} />
+      <CardContent className={css.content}>
+        <List sx={{ p: 0 }}>
+          {items.map((item, i) => (
+            <ListItem key={i} sx={{ p: 0 }}>
+              <ListItemIcon className={css.listIcon}>
+                <SvgIcon component={FileIcon} inheritViewBox fontSize="small" sx={{ fill: 'none' }} />
+              </ListItemIcon>
+              <ListItemText {...item} />
```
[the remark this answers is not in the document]
iamacook: You can manually set the `primary` and `secondary` `Typography` props. I've changed the `secondary` element to be a div:

```ts
secondaryTypographyProps={{ component: 'div' }}
```

src/components/settings/DataManagement/ImportDialog.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const isDisabled = (!addedSafes && !addressBook && !settings && !safeApps) || !!error
+
+  const handleClose = () => {
+    setFileName(undefined)
+    setJsonData(undefined)
+    onClose?.()
+  }
+
+  const handleImport = () => {
+    trackEvent(SETTINGS_EVENTS.DATA.EXPORT_ALL_BUTTON)
```
usame-algan, line 45: Shouldn't this be in the export handler?

src/components/welcome/styles.module.css
```diff
@@ -19,8 +19,12 @@
   margin: 0;
 }
 
+.sidebar {
+  gap: 20px;
+}
+
 .sidebar :global .MuiPaper-root {
-  height: calc(100vh - var(--header-height) - var(--footer-height) - 24px);
+  max-height: calc(100vh - var(--header-height) - var(--footer-height) - 24px);
```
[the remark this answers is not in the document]
usame-algan: Now if the right area is bigger than the sidebar area there is this big space between the widget and navigation:

Screenshot 2023-04-03 at 13 45 54

I think if we change the navigation grid to `flex-grow: 0` it improves it
Screenshot 2023-04-03 at 13 47 06

usame-algan (approved):
Thanks for adding the changes. LGTM! 🚀

francovenica:
Things that work fine:
non json file cannot be uploaded
non properly formatted json files are not uploaded (the error is shown)
json files with no data or with data that cannot be imported (files broken or empty) cannot be uploaded and the error is shown
Uploading old files (v1)
Importing and exporting new files work in the welcome page
Importing and exporting new files work in the Data tab
Imports the data mentioned: Safes added per chain, Address book per chain, settings in appearances, fiat currency selected, pinned apps per chain

Issues:
1 - Importing is not setting the fiat currency selection (the one you set from the assets tab), settings preferences (dark mode, prefix chain...) and the pinned apps for the chain
image

image

2 - There are error that pop up in the console when refreshing the page in the Data tab. Going to the data tab is not enough, you have to manually refresh the page while being in the data tab. This issues happen even with a clear app (Local Storage cleared) and they don't happen in prod currently, so I think they were introduced here
image

src/store/safeAppsSlice.ts
```diff
@@ -22,15 +22,23 @@ export const safeAppsSlice = createSlice({
       state[chainId] ??= { pinned: [] }
       state[chainId].pinned = pinned
     },
+    setSafeApps: (state, { payload }: PayloadAction<SafeAppsState>) => {
+      // We much return as we are overwriting the entire state
```
usame-algan, line 26:
```suggestion
      // We must return as we are overwriting the entire state
```

src/store/settingsSlice.ts
```diff
@@ -90,6 +91,11 @@ export const settingsSlice = createSlice({
     setEnv: (state, { payload }: PayloadAction<EnvState>) => {
       state.env = payload
     },
+    setSettings: (_, { payload }: PayloadAction<SettingsState>) => {
+      // We much return as we are overwriting the entire state
```
usame-algan, line 95:
```suggestion
      // We must return as we are overwriting the entire state
```

src/store/safeAppsSlice.ts
```diff
@@ -22,15 +22,23 @@ export const safeAppsSlice = createSlice({
...
+      return payload
```
usame-algan, line 27: I would think `state = payload` would also overwrite the entire state. What is the difference to returning?

usame-algan (approved):
Tested the reported issues which are resolved now 👍

src/store/safeAppsSlice.ts
iamacook: Assigning uses a draft state via [Immer](https://immerjs.github.io/immer/) wheras returning uses the `payload` as is. Assigning _should_ work in theory but my debugging says otherwise 🤷‍♂️

francovenica:
- I still see the same errors in the console I reported in the previous comment. To see them navigate to the Data tab and refresh the page

- In the Data tab the "Bookmarked safe apps" are listed there even if there are no pinned apps at all to export. 

- In the Data tab the "address book" is listed to be exported even if every entry was deleted. I assume is because you still have the chain value even if there are no entries.
Maybe this is an issue about the LocalStorage and not the export, so I'd agree to make this issue its own ticket if you want
image

This one is a nish:
- The export button in the sidebar in the welcome page doesn't show up until you actually add a safe or some entry in the address book, but in the Data tab the export button shows right away, allowing you to download the settings (prefix adding, dark mode, fiat currency). I think we show both, or hide both.
Note: I said yhis one is a nish because since in order to see the Data tab without having a safe in the sidebar you have to copy/paste a safe in the URL or have it in your bookmarks or something. I assume 99% of people will have a safe in the side before seeing the Data tab, so they will see both, the list in the Data tab and the Export in the sidebar welcome page.

iamacook:
> * I still see the same errors in the console I reported in the previous comment. To see them navigate to the Data tab and refresh the page

This has been fixed and I can confirm that it is no longer appearing on the PR deployment.

>     * In the Data tab the "Bookmarked safe apps" are listed there even if there are no pinned apps at all to export.

This now only shows if pinned Apps are present on a chain.

>     * In the Data tab the "address book" is listed to be exported even if every entry was deleted. I assume is because you still have the chain value even if there are no entries.
>       Maybe this is an issue about the LocalStorage and not the export, so I'd agree to make this issue its own ticket if you want

As agreed, I fixed the root cause of this issue. Any users that previously added then manually removed all entries (before this fix) will see the same as your screenshot, but this is an edge case we decided not to cater for.

> This one is a nish:
> 
>     * The export button in the sidebar in the welcome page doesn't show up until you actually add a safe or some entry in the address book, but in the Data tab the export button shows right away, allowing you to download the settings (prefix adding, dark mode, fiat currency). I think we show both, or hide both.
>       Note: I said yhis one is a nish because since in order to see the Data tab without having a safe in the sidebar you have to copy/paste a safe in the URL or have it in your bookmarks or something. I assume 99% of people will have a safe in the side before seeing the Data tab, so they will see both, the list in the Data tab and the Export in the sidebar welcome page.

As discussed, this won't be adjusted as a user does not inherently _have_ data unless they have an added Safe.

usame-algan (approved):

francovenica:
The fixes look good, no more errors and the list doesn't show address book or bookmarked apps if there are no entries and pinned apps respectively

PR #1820, opened 2023-03-29
feat: relay batch modal

### LINKED ISSUE #1818: [Relaying V1] Relay in batch transactions modal
## What is the feature about
Add the relay transaction in the batch transactions modal

## The list of requirements
- The Multisend call should be equally eligible to be relayed
- Include a tooltip in `<SponsoredBy />` as per the linked design

## Designs/sketches
Screenshot 2023-03-28 at 23 11 38

## Links
https://www.figma.com/file/M3J3IJ5YFxRKaT95F2uBAt/Relaying?node-id=1193-31921&t=Cgf3iUNDcyDitout-0

DiogoSoaress:
## What it solves

Resolves #1818 

## How this PR fixes it
Implements relaying for batch executions

## How to test it
1. In Goeli, fully sign 2+ transactions so that you can batch them
2. The "How to pay" info should be present in the Modal
3. Click "Send" will relay the Multisend transaction

## Screenshots
Screenshot 2023-03-29 at 20 08 14

## Checklist


src/hooks/useRemainingRelays.ts
```diff
@@ -4,7 +4,7 @@ import useSafeAddress from '@/hooks/useSafeAddress'
 import { Errors, logError } from '@/services/exceptions'
 import { SAFE_GELATO_RELAY_SERVICE_URL } from '@/config/constants'
 
-const fetchRemainingRelays = async (chainId: string, safeAddress: string): Promise<number | undefined> => {
+const fetchRemainingRelays = async (chainId: string, safeAddress: string) => {
```
iamacook, line 7: Nit: I would not remove the return type entirely.

```suggestion
const fetchRemainingRelays = async (chainId: string, safeAddress: string): Promise<number> => {
```

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -333,9 +333,53 @@ export const dispatchTxRelay = async (
...
+    txs.forEach(({ txId }) => {
+      txDispatch(TxEvent.RELAYING, { taskId, txId, groupKey })
+    })
+
+    // Monitor relay tx
+    waitForRelayedTx(
+      taskId,
+      txs.map((tx) => tx.txId),
+      groupKey,
+    )
```
iamacook, line 374: Let's keep consistent and move these out of the `try` (into their own `try` when necessary) as we do with all dispatchers.

src/components/tx/modals/BatchExecuteModal/ReviewBatchExecute.tsx
```diff
@@ -91,6 +112,19 @@ const ReviewBatchExecute = ({ data, onSubmit }: { data: BatchExecuteData; onSubm
         </Typography>
         <DecodedTxs txs={txsWithDetails} />
 
+        {canRelay ? (
+          <>
+            <Typography mt={2} color="primary.light">
+              How to pay:
+            </Typography>
+            <SponsoredBy
```
usame-algan, line 120: Can we add a slight margin bottom to "How to pay:" like with the other caption?
Screenshot 2023-03-30 at 12 31 40

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -333,9 +333,53 @@ export const dispatchTxRelay = async (
...
+    const relayResponse = await sponsoredCall({
+      chainId,
+      to,
+      data,
+    })
+    const taskId = relayResponse.taskId
+
+    if (!taskId) {
+      throw new Error('Transaction could not be relayed')
+    }
```
usame-algan, line 363: I believe `sponsoredCall` can already throw so we would never get to line 361. Do we still need that check?

src/components/tx/modals/BatchExecuteModal/ReviewBatchExecute.tsx
```diff
@@ -91,6 +112,19 @@ const ReviewBatchExecute = ({ data, onSubmit }: { data: BatchExecuteData; onSubm
         </Typography>
         <DecodedTxs txs={txsWithDetails} />
 
+        {canRelay ? (
```
usame-algan, line 115: In the safe creation relay PR, this is called `willRelay`. I would suggest keeping the naming consistent.

src/components/tx/modals/BatchExecuteModal/ReviewBatchExecute.tsx
```diff
@@ -91,6 +112,19 @@ const ReviewBatchExecute = ({ data, onSubmit }: { data: BatchExecuteData; onSubm
...
+          <>
+            <Typography mt={2} color="primary.light">
+              How to pay:
+            </Typography>
+            <SponsoredBy
+              remainingRelays={remainingRelays}
+              tooltip="You can only relay multisend transactions containing
+executions from the same Safe."
```
usame-algan, line 123: What does this mean? Aren't all transactions part of the same safe?

usame-algan (commented):
Is it intended that each relayed batch execution only removes one from the limit?

DiogoSoaress:
> Is it intended that each relayed batch execution only removes one from the limit?

Yes

PR #1821, opened 2023-03-30
fix: rejection transaction description + link text

iamacook:
## What it solves

Resolves wrong text

## How this PR fixes it

The description of rejection transactions and the link to the relative help article have been adjusted.

## How to test it

Queue a rejection transaction and observe the following, with an updated description after executing it.

The description should read (relative to queue and history):

"This is an on-chain rejection that won't send any funds. Executing this on-chain rejection will replace all currently awaiting transactions with nonce n."
"This is an on-chain rejection that did't send any funds. This on-chain rejection replaced all transactions with nonce n."

The help article link text should read:

"Why do I need to pay to reject a transaction?"

## Screenshots

image

image

## Checklist


francovenica:
LGTM

PR #1823, opened 2023-03-30
fix: Remove overlap of address and balance in safe list sidebar

### LINKED ISSUE #1492: [Sidebar, Mobile view] Safe address overlapped by current balance
## Bug description
In the mobile view, for added safes in the sidebar, if the balance is too long or has too many decimals it overlaps the safe address

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Gnosis chain
 - Env: https://app.safe.global/gno:0xB8d760a90a5ed54D3c2b3EFC231277e99188642A/home

## Steps to reproduce
 1. Go to the app and add the safe given above (it has a balance with a lot of decimals)
 2. After adding the safe open the sidebar

## Expected result
The address should't overlapp with the balance

## Obtained result
The address is overlapped by the balance

## Screenshots
image

usame-algan:
## What it solves

Resolves #1492 

## How this PR fixes it

- Increases padding-right of address element to cover long balance values

## How to test it

1. Open the Safe
2. Add a safe that has a balance of native tokens < 0.00001
3. Observe no overlap in the sidebar on mobile screens

## Screenshots
Screenshot 2023-03-30 at 16 30 30

## Checklist

### DIFF
--- a/src/components/sidebar/SafeListItem/index.tsx
+++ b/src/components/sidebar/SafeListItem/index.tsx
@@ -94,7 +94,7 @@ const SafeListItem = ({
             <SafeIcon address={address} {...rest} />
           </ListItemIcon>
           <ListItemText
-            sx={noActions ? undefined : { pr: 10 }}
+            sx={noActions ? undefined : { pr: 13 }}
             primaryTypographyProps={{
               variant: 'body2',
               component: 'div',
--- a/src/components/sidebar/SafeListItem/styles.module.css
+++ b/src/components/sidebar/SafeListItem/styles.module.css
@@ -28,9 +28,3 @@
 .withPendingButtons :global .MuiListItemSecondaryAction-root {
   right: 74px;
 }
-
-@media (max-width: 400px) {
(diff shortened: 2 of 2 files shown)


katspaugh:
This doesn't look like an improvement to me:
Screenshot 2023-03-30 at 18 40 00
Lots of unused white space but cut-off text. 
IMHO we need to rethink the number formatting there, not paddings.

katspaugh:
I would go with option 2. Two lines, although they look OK, are a design change, IMHO we can fix the bug without changing the design.

I like the fiat value idea, just keep in mind it's very volatile and you'd have to update it every time you open the sidebar. I think it should be fine with 1-5 Safes.
It's defo more representative of your Safe total balance than just ETH.
But you'd still have to format the number in a concise way.

liliiaorlenko:
as an option we can also increase width of safe list on whole mobile screen
further updates with design I can introduce when working on redesign of safe navigation :)

usame-algan:
Closing this PR to wait for the new safe navigation design.

PR #1824, opened 2023-03-30
refactor: split create safe callbacks

DiogoSoaress:
## What it solves

Resolves https://github.com/safe-global/web-core/pull/1815#discussion_r1153032221

## How this PR fixes it
Splits the logic inside `useSafeCreation`

## How to test it
Create safe should work as expected with relay and without


DiogoSoaress:
@usame-algan I didn't verify the commit had failed before merging https://github.com/safe-global/web-core/pull/1815

We've done this changes together and it should change nothing in the create flow behavior but have another look nevertheless.

usame-algan (approved):

PR #1827, opened 2023-03-31
fix: Track most searched safe apps

### LINKED ISSUE #1814: [Analytics] Most searched apps events not being tracked
## Description

As found by señor Yago, we have an issue with the analytics for the most searched apps as we are not receiving any new data since the 1st of March.

## Steps to reproduce
 1. Go to analytics page -> Most searched apps
 2. Filter starting from 1st of march and modify the end date a couple of times to a different later date to see that indeed there is no change on the number of apps tracked.

## Expected result
More events should be tracked when the end date is changed.

## Obtained result
No new events are being tracked when selecting a later date since the 1st of March.

usame-algan:
## What it solves

Resolves #1814 

## How this PR fixes it

- Adds back tracking for safe apps search which was accidently removed in #1537 (https://github.com/safe-global/web-core/pull/1537/files#diff-3b305e872dfa1d5d247bcb20a563b4a5f388c1cb0615d97d1cd911c450af4be8L30-L36)

## How to test it

1. Open the Safe
2. Navigate to Safe apps
3. Search for a safe app
4. Observe a tracking event after a small delay

## Screenshots

Screenshot 2023-03-31 at 10 43 56

## Checklist

### DIFF
--- a/src/hooks/safe-apps/useSafeAppsFilters.ts
+++ b/src/hooks/safe-apps/useSafeAppsFilters.ts
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 import type { Dispatch, SetStateAction } from 'react'
 import type { SafeAppData } from '@safe-global/safe-gateway-typescript-sdk'
 
@@ -6,6 +6,7 @@ import { useAppsFilterByCategory } from './useAppsFilterByCategory'
 import { useAppsSearch } from './useAppsSearch'
 import { useAppsFilterByOptimizedForBatch } from './useAppsFilterByOptimizedForBatch'
 import useDebounce from '../useDebounce'
+import { SAFE_APPS_EVENTS, trackSafeAppEvent } from '@/services/analytics'
 
 type ReturnType = {
   query: string
@@ -28,6 +29,14 @@ const useSafeAppsFilters = (safeAppsList: SafeAppData[]): ReturnType => {
   const filteredAppsByQueryAndCategories = useAppsFilterByCategory(filteredAppsByQuery, selectedCategories)
 
   const filteredApps = useAppsFilterByOptimizedForBatch(filteredAppsByQueryAndCategories, optimizedWithBatchFilter)
+
+  const debouncedSearchQuery = useDebounce(query, 500)
+
+  useEffect(() => {
+    if (debouncedSearchQuery) {
+      trackSafeAppEvent({ ...SAFE_APPS_EVENTS.SEARCH, label: debouncedSearchQuery })
+    }
+  }, [debouncedSearchQuery])
 
   return {
     query,


src/hooks/safe-apps/useSafeAppsFilters.ts
```diff
@@ -29,6 +30,14 @@ const useSafeAppsFilters = (safeAppsList: SafeAppData[]): ReturnType => {
 
   const filteredApps = useAppsFilterByOptimizedForBatch(filteredAppsByQueryAndCategories, optimizedWithBatchFilter)
 
+  const debouncedSearchQuery = useDebounce(query, 500)
```
iamacook, line 33: Just a thought: what about tracking this on unmount instead, e.g. when the chosen app is open? This is highly dependent on the speed of the typing.

usame-algan: We might miss some data points then i.e. users who search for an app but don't open it. Also I think we already track whenever a safe app is open via the `SAFE_APPS_EVENTS.OPEN_APP` event.

katspaugh: Maybe 2 seconds then?

katspaugh: Or better, store the list of distinct queries, and track on unmount.
If a query starts with the same symbols as an existing query, overwrite it instead of adding to the set.

If we end up with stuff like `gov`, `govern`, `safe gover` all tracked in the same session, that would be too much.

katspaugh: A bigger debounce should be OK too, though. Maybe 5 seconds.
The search is also pretty slow and blocks typing.

iamacook: > We might miss some data points then i.e. users who search for an app but don't open it.

The only time that would happen is if the user closes the browser, otherwise unmount always occurs when the user navigates within the app. It's certainly something to consider though.

> Also I think we already track whenever a safe app is opened via the `SAFE_APPS_EVENTS.OPEN_APP` event.

I'm thinking about over-tracking of the search term. Perhaps `onBlur` would be a better reference point as the `useEffect` may run >1 times for slow typers. What do you think?

usame-algan: I noticed that there already was a debounce of 400ms in place so I removed my change and adjusted the other hook instead. So currently, the safe apps list is updated after 400ms of not typing and thats where we also dispatch the analytics event. We could increase that number but imo it would make the UX worse because it would take longer to see the results of the search.

Was this call removed because of over-tracking? If not I suggest we restore the previous state and create a separate ticket for future improvements on the mechanism.

usame-algan: I can imagine that if the list updates too fast it could "interrupt" the user. I assume that it could also lead to users not typing out longer search terms anymore because they already see a result. There is [some interesting info](https://ux.stackexchange.com/questions/95336/how-long-should-the-debounce-timeout-be) on this topic which I think we should explore before making such a change.

As for the analytics, lets add a second debounce (2s) as suggested.

katspaugh: I think there's some inefficiency in the re-rendering too, it shouldn't be so slow. In Safari, the search bar is super sluggish.

katspaugh (approved):

francovenica:
Looks good, like the snapshot in the description

Question, Is the idea to track the most searched apps? or should we track anything the user writes? I think they are technically different things. 
So the question would be if you want to track what the user writes even if he's not finding anything.
Also the event triggers after 2 seconds, so if you are a slow writer or you stop mid search you might end up tracking several times what is actually 1 search. 

Or maybe I'm just over thinking. Anyways it looks fine and I'm okay shipping it how it is right now

usame-algan:
> Question, Is the idea to track the most searched apps? or should we track anything the user writes? I think they are technically different things.
So the question would be if you want to track what the user writes even if he's not finding anything.
Also the event triggers after 2 seconds, so if you are a slow writer or you stop mid search you might end up tracking several times what is actually 1 search.

I think the idea is to track anything the user searches for. On the analytics end we can then write queries to find out things like most searched terms.

Issue #1828, opened 2023-03-31
[Epic] Refactor and redesign the tx modals


katspaugh:
## What is the feature about

We have plans to add more information and knobs in the Safe transaction flow (e.g. better decoding, security warnings, batching).

The existing tx flow will need a refactoring and a redesign to accommodate all that new content, and to unify but also allow custom representations for different transaction types.

## Goals

* Separate logic from UI
* Modularize the code so that it can be reused in other apps (maybe be even make separate packages)
* Define common components that are present in all transaction types
* Refactor the tx stepper so that it's more transparent and fully typed
* Add more steps in the tx flow, so that we can break down the info into several screens

A proposed new tx flow: https://kinopio.club/tx-flow-6x3PhJ26UOGjjhmYo-hDw

## Designs
Tbd by @liliiaorlenko

PR #1830, opened 2023-03-31
Refactor: extract SignOrExecute logic into hooks

katspaugh:
## What it solves
Related to #1828

A small, almost mechanical, refactoring of SignOrExecuteForm.

I've extracted some parts of the component into a few hooks.

## How to test
* Create and submit transactions
* Test immediate execution
* Test signing


src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -0,0 +1,87 @@
...
+        await dispatchTxExecution(safeTx, txOptions, id, onboard, chainId)
+        return id
+      },
+    }
+  }, [chainId, safeAddress, version, onboard, wallet])
+}
+
+export const useValidateNonce = (safeTx?: SafeTransaction): boolean => {
+  const { safe } = useSafeInfo()
+  return !!safe && !!safeTx && safeTx?.data.nonce === safe.nonce
```
usame-algan, line 74: Similarly here we don't need to check for the existance of `safe` and could keep the shorthand syntax from before since the default fallback nonce is -1 which no tx should ever have.

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -0,0 +1,88 @@
...
+      }
+
+      // Otherwise, sign off-chain
+      const signedTx = await dispatchTxSigning(safeTx, version, onboard, chainId, txId)
+      return await proposeTx(wallet.address, signedTx, txId, origin)
+    }
+
+    const executeTx: TxActions['executeTx'] = async (safeTx, txId, txOptions) => {
+      if (!onboard) throw 'Onboard not ready'
+      if (!wallet) throw 'Wallet not connected'
```
usame-algan, line 61: I would suggest to throw with an `Error` object to stay consistent with the othe throws.

src/components/tx/SignOrExecuteForm/UnknownContractError.tsx
```diff
@@ -0,0 +1,31 @@
...
+import useSafeInfo from '@/hooks/useSafeInfo'
+import { getExplorerLink } from '@/utils/gateway'
+import ErrorMessage from '../ErrorMessage'
+
+const UnknownContractError = (): ReactElement | null => {
+  const { safe, safeAddress } = useSafeInfo()
+  const currentChain = useCurrentChain()
+
+  // Unsupported base contract
+  const isUnknown = safe.implementationVersionState === ImplementationVersionState.UNKNOWN
```
iamacook, line 14: Whilst working on a separate refactor, I found a function we can reuse here: `isValidMasterCopy` in `@/services/contracts/safeContracts.ts`

PR #1831, opened 2023-03-31
Relay service deployment toggle

DiogoSoaress:
## What it solves

Adds a toggle in non-PRODUCTION environments to toggle between deployments (STG vs PROD) of the Safe Gelato relay service

## How to test it
1. Switching on the "Use prod Relay Service" should make the app use the PROD deployment of the  Safe Gelato relay service

## Screenshots
Screenshot 2023-03-31 at 16 59 59

### DIFF
--- a/src/components/sidebar/DebugToggle/index.tsx
+++ b/src/components/sidebar/DebugToggle/index.tsx
@@ -1,23 +1,26 @@
 import { type ChangeEvent, type ReactElement } from 'react'
 import { Box, FormControlLabel, Switch } from '@mui/material'
 import { localItem } from '@/services/local-storage/local'
-import useLocalStorage from '@/services/local-storage/useLocalStorage'
+import useLocalStorage, { type Setter } from '@/services/local-storage/useLocalStorage'
 import { setDarkMode } from '@/store/settingsSlice'
 import { useDarkMode } from '@/hooks/useDarkMode'
 import { useAppDispatch } from '@/store'
 
 const LS_KEY = 'debugProdCgw'
+const RELAY_SERVICE_STORAGE_KEY = 'relayService'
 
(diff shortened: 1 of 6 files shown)


katspaugh (commented):
Can you explain why you think it's required?
I would rather expect that QA should test either with a complete prod CGW, or with a complete staging CGW and not mix environments.
Suggest closing this.

DiogoSoaress:
> I would rather expect that QA should test either with a complete prod CGW, or with a complete staging CGW and not mix environments.

So one single toggle to switch both services?

liliya-soroka:
Can we have the case when we need to test new changes for the staging relay service on the Gnosis chain? This is the case when it will be helpful to have the toggle because we don't have the Gnosis chain on staging CGW.
What do you think?

schmanu:
On service side there is currently no implementation difference between gnosis chain and Görli.
Nonetheless I think in staging and branch preview environments it should be sufficient to test the staging relayer service. Then we can test new changes by using the prod CGW which will enable gnosis chain. 
If we would make that toggle also switch the relay service to prod it would become impossible to test new changes on service side on gnosis chain.

iamacook:
> So one single toggle to switch both services?

I would suggest we modify `SAFE_GELATO_RELAY_SERVICE_URL` as mentioned in [the epic description](https://github.com/safe-global/web-core/pull/1776).

```ts
const SAFE_GELATO_RELAY_SERVICE_URL = IS_PRODUCTION ? : SAFE_GELATO_RELAY_SERVICE_URL_PRODUCTION : SAFE_GELATO_RELAY_SERVICE_URL_STAGING
```

iamacook:
After discussion, I am closing this in favour of https://github.com/safe-global/web-core/pull/1832.

PR #1833, opened 2023-04-03
fix: cleanup relay env. vars

iamacook:
## What it solves

Add env. var. to readme/example and remove unnecessary one from workflow.

## How to test it

Relaying should still work as expected.

## Checklist


.github/workflows/build/action.yml
```diff
@@ -35,4 +35,3 @@ runs:
         NEXT_PUBLIC_TENDERLY_SIMULATE_ENDPOINT_URL: ${{ fromJSON(inputs.secrets).NEXT_PUBLIC_TENDERLY_SIMULATE_ENDPOINT_URL }}
         NEXT_PUBLIC_WC_BRIDGE: ${{ fromJSON(inputs.secrets).NEXT_PUBLIC_WC_BRIDGE }}
         NEXT_PUBLIC_CYPRESS_MNEMONIC: ${{ inputs.e2e_mnemonic }}
-        NEXT_PUBLIC_SAFE_GELATO_RELAY_SERVICE_URL: ${{ fromJSON(inputs.secrets).NEXT_PUBLIC_SAFE_GELATO_RELAY_SERVICE_URL }}
```
iamacook, line 38: This is not needed as the constant has a fallback.

katspaugh: Fallback is a fallback, but we need to pass both URLs from the secrets to the env.

iamacook: I've added both to the workflow.

katspaugh (approved):

PR #1836, opened 2023-04-03
Fix: filter pending txs by safe address

### LINKED ISSUE #1817: Directly executing txs is blocked while txs are pending
## Bug description
When a transaction in any Safe is pending, it is impossible to directly execute transactions in other 1/n Safes.
It only works from the confirmation modal in the queue.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Görli

## Steps to reproduce
 1. Execute a transaction in Safe A
 2. While the transaction is indexing, open a new 1/n Safe B with an empty queue
 3. In Safe B try to create a tx and witness that there is no Execute transaction option given. You can only sign and propose the tx.

## Expected result
We can only sign transactions in other 1/n Safes.

## Obtained result
In other Safes without pending txs, we should be able to directly execute transactions if it's the next nonce.

## Screenshots
Check that the next nonce is the same as in the created tx and the Safe is 1/3
Screenshot 2023-03-28 at 18 50 05
Screenshot 2023-03-28 at 18 50 22

katspaugh:
## What it solves

Resolves #1817

## How this PR fixes it

When we check if there are pending transactions, we need to filter all pending transactions by chainId and safeAddress.
So I've added `safeAddress` to the pending transaction type.
This wasn't required before because we were always selecting pending txs by id (taken from the queue or history).

## How to test it
* In a 1/X Safe, create and immediately execute a tx
* Attempt to create and execute another tx in the same Safe – it won't let you execute it while the 1st one is still pending.
* In another 1/X Safe, attempt to create and execute a tx – you should be able to do so regardless of the other Safe's state


src/store/pendingTxsSlice.ts
```diff
@@ -11,6 +12,7 @@ export enum PendingStatus {
 
 export type PendingTx = {
   chainId: string
+  safeAddress: string
```
schmanu, line 15: Safes that have transactions pending before we update could run into unexpected errors as the existing pending txs (before we release) will not have the `safeAddress` set.

schmanu (commented):
The jest tests are failing because they do not set the `safeAddress` in their mock initial state.

src/store/pendingTxsSlice.ts
katspaugh: Yeah, I've tested this locally. It will basically not display these txs in the pending queue on top of the normal queue and it won't block immediate execution. IMHO that's fine, they disappear after a while.

schmanu (approved):
🔥 🚒 👨‍🚒  thanks for fixing the issue :)

francovenica:
Issue: After executing the tx in the 1st safe, When you go to the 2nd safe to execute another tx, the "Execute now" checkbox disappears when the notification that the first tx is being process shows up.

In the 1st safe create the tx with low gas fee so it takes slightly longer to execute the tx
Switch the 2nd safe
Prepare a tx to be executed and wait before hitting submit
Wait for the notification for the 1st tx to show up

Result:
The checkbox to execute now is removed in the modal of the 2nd safe. Even going back and forth for a new stimation won't make the checkbox to show again, you have to close the modal and start over

gif:
execute now

francovenica:
The issue I reported is still there. If it is decided that the issue won't be solved please say so in a comment

src/hooks/useTxPendingStatuses.ts
```diff
@@ -85,6 +87,7 @@ const useTxPendingStatuses = (): void => {
           dispatch(
             setPendingTx({
               chainId,
+              safeAddress,
```
schmanu, line 90: This leads to errors if the event is emitted from a different Safe.

E.g.
Safe A executes a tx and it is processing
Then Safe B is opened.
Not Safe A's tx is executed and moves to "indexing". We would not set Safe B to pending.

I'll address this issue.

iamacook (approved):
One small suggestion, though not critical. Otherwise LGTM!

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -175,7 +175,7 @@ const SignOrExecuteForm = ({
     const id = txId || (await proposeTx(createdTx))
     const txOptions = getTxOptions(advancedParams, currentChain)
     trackEvent({ ...MODALS_EVENTS.PROPOSE_TX, label: ExecutionType.CONNECTED_WALLET })
-    await dispatchTxExecution(createdTx, txOptions, id, onboard, safe.chainId)
+    await dispatchTxExecution(createdTx, txOptions, id, onboard, safe.chainId, safeAddress)
```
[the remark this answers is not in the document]
katspaugh: Please don't, it has fields this function doesn't need.

francovenica:
It works fine now. 
LGTM

katspaugh:
Thank you, Manu and Franco!

Issue #1829, opened 2023-03-31
Unhandled error when executing transaction in queue


usame-algan:
## Bug description

There is an unhandled error when executing a transaction from the queue sometimes.

Sentry: https://gnosis.sentry.io/issues/3979496225/events/edb6e22d75db48009523c3ad6880c6cd/?project=6769196

## Environment
 - Browser: any
 - Wallet: any
 - Chain: any

## Steps to reproduce
 1. Go to a safe
 2. Queue up a transaction
 3. Navigate to the queue
 4. Execute the transaction

## Expected result

- The transaction disappears after being processed

## Obtained result

- An unhandled error occurs

## Screenshots

<img width="989" alt="Screenshot 2023-03-31 at 12 00 33" src="https://user-images.githubusercontent.com/5880855/229148407-b8a1927a-31ca-4fcf-8fe3-c588a100b0bd.png">

JagoFigueroa:
Looks good to me, I've been signing and executing a bunch of transactions from the queue on different L2s and the error did not pop up for me once.

Cheers!

PR #1842, opened 2023-04-04
refactor: `SignOrExecute` elements

katspaugh:
## What it solves

Part of #1828

An umbrella PR for code refactors in preparation for a tx flow redesign.

## How to test
Make all kinds of transactions, immediate executions/signatures/on-chain signatures/relayed executions. Everything should work like before.

## Analytics changes
Relayed executions now have their own event instead of "Propose".

PR #1843, opened 2023-04-04
fix: Over-initialization of the safe core sdk

usame-algan:
## What it solves

While working on #1829, I noticed that the core-sdk is re-initialized every time the queue or history updates e.g. when a user executes a transaction. This is not necessary so I narrowed down the dependencies instead of using the whole `safe` object.

## How to test it

1. Open the Safe
2. Create and execute a transaction
3. Observe that the `useInitSafeCoreSDK` hook is not running anymore

## Checklist

### DIFF
--- a/src/hooks/coreSDK/__tests__/safeCoreSDK.test.ts
+++ b/src/hooks/coreSDK/__tests__/safeCoreSDK.test.ts
@@ -4,10 +4,10 @@ import {
   getProxyFactoryContract,
   getSafeContract,
 } from '@safe-global/safe-core-sdk/dist/src/contracts/safeDeploymentContracts'
-import type { SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
+import { ImplementationVersionState } from '@safe-global/safe-gateway-typescript-sdk'
 import { Web3Provider } from '@ethersproject/providers'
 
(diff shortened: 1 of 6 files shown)


iamacook (commented):
You need not modify the argument of `initSafeSDK`. You can simply make the dependency array more grandular.

Although it arguably the same, if you don't want to break the linter rule, you can add an equality function to the Safe info selector (requiring direct use of `selectSafeInfo`), e.g.

```ts
const { data: safe } = useAppSelector(selectSafeInfo, (prev, next) => {
  return prev.address.value !== curr.address.value || next.chainId !== next.chainId
})
```

katspaugh:
Should we merge this into #1842 instead of dev?

katspaugh (approved):

usame-algan:
Just to summarize what @iamacook and I discussed regarding `SafeInfo` and how we pass it here. We looked at 2 options, keep passing the whole object and adjust the dependency array vs. only passing the values that are needed (this was merged). While Option 1 requires fewer changes and makes the hook shorter/more readable it also requires us to ignore missing elements in the dependency array which is unconventional.

Through the discussion, we identified the problem that most of the information in `SafeInfo` never changes like `safeVersion` and `safeAddress` while a few values change a lot like `queueTag` and `historyTag`. This could be the basis for a future refactor e.g. where we split up the data accordingly.

PR #1845, opened 2023-04-04
[POC] feat: validate batch and exclude txs that revert

### LINKED ISSUE #1130: Improve transaction batching
## Bug description

Transactions that modify owners, thresholds or upgrade the Safe should not be able to be batched. The batched transaction will otherwise become stuck.

Edit: we may be able to use the decoding service (for security) for this.

## Steps to reproduce

Queue two transactions that modify the owners and execute them in batch.

## Expected result

Only "standard" transactions should be able to be batched.

## Obtained result

Batched transactions can get stuck.

## Links

- [Slack discussion](https://5afe.slack.com/archives/C03BHKS5K6D/p1668086873310059)
- [Proposed solutions](https://www.notion.so/gnosis-safe/Improving-transaction-batching-88222641f452497589abe10d9dd52701)

iamacook:
## What it solves

Resolves #1130

## How this PR fixes it

The initial batched transaction is validated and, if it fails, each nested transaction is sequentially included and validated inside a new batch. The valid batch is then returned, excluding the transaction that would otherwise fail and those thereafter.

## How to test it

Open a m/n Safe and queue the following transactions in order:

1. Remove owner and set threshold to 1.
2. Set threshold to m.

Transaction 2 is invalid as it attempts to set the threshold higher than that of the Safe is transaction 1 succeeds, which should fail validation.

## Screenshots

image

## Checklist

### DIFF
--- a/src/components/tx/modals/BatchExecuteModal/ReviewBatchExecute.tsx
+++ b/src/components/tx/modals/BatchExecuteModal/ReviewBatchExecute.tsx
@@ -19,6 +19,8 @@ import SponsoredBy from '@/components/tx/SponsoredBy'
 import { dispatchBatchExecution, dispatchBatchExecutionRelay } from '@/services/tx/tx-sender'
 import useOnboard from '@/hooks/wallets/useOnboard'
 import { WrongChainWarning } from '@/components/tx/WrongChainWarning'
+import { getValidBatch } from '@/utils/transactions'
+import type { MetaTransactionData } from '@safe-global/safe-core-sdk-types'
 
 const ReviewBatchExecute = ({ data, onSubmit }: { data: BatchExecuteData; onSubmit: (data: null) => void }) => {
   const [isSubmittable, setIsSubmittable] = useState<boolean>(true)
@@ -31,7 +33,7 @@ const ReviewBatchExecute = ({ data, onSubmit }: { data: BatchExecuteData; onSubm
   const willRelay = !!remainingRelays
   const onboard = useOnboard()
 
-  const [txsWithDetails, error, loading] = useAsync<TransactionDetails[]>(() => {
+  const [txsWithDetails, txWithDetailsError, txWithDetailsLoading] = useAsync<TransactionDetails[]>(() => {
     if (!chain?.chainId) return
 
     return getTxsWithDetails(data.txs, chain.chainId)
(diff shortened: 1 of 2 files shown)


katspaugh:
Awesome idea!
An alternative UI could be: highlight the failing/mutually exclusive sub-transactions in red and add a trash icon to them and suggest to the user to exclude them manually.

iamacook:
I've just pushed a change that validates the transactions syncronously and it is noticeably faster. However, it also validates transactions after invalid one(s) in the proposed batch, resulting in unnecessary RPC calls.

---

I investigated using Tenderly simulations and it is possible using this (unoptimised) snippet:

```ts
const callErrors = simulation.transaction.call_trace.filter((call) => {
  return call.error
})

const firstRevertedTx = allMultiSendTxs.find((tx) => {
  return callErrors.some((error) => tx.data === error.input)
})
```

However, this is not type-correct (we are using unofficial types). (@usame-algan mentioned that Tenderly are currently working on a TypeScript SDK though.)

---

To summarise: the former increases RPC calls, the latter needs more testing (and we'd need to confirm the extra simulation calls don't negatively affect our API limit).

I will leave this as a draft for the time being as we still need to discuss this with design.

src/utils/transactions.ts
```diff
@@ -130,6 +132,40 @@ export const getMultiSendTxs = (
...
+
+  const multiSendTxs = getMultiSendTxs(txs, chain, safeAddress, safeVersion)
+
+  const isValidBatch = await _isValidMultiSend(safe, multiSendTxs)
+
+  if (isValidBatch) {
+    return multiSendTxs
+  }
+
+  const validity = await Promise.all(multiSendTxs.map((_, i) => _isValidMultiSend(safe, multiSendTxs.slice(0, i))))
```
schmanu, line 163: This will always test all possible tx combinations.
I think the most request number efficient implemention would be through binary search:

```ts
  let idx = multiSendTxs.length
  let bestBatch: MetaTransactionData[] = []
  while (idx > 0 && idx <= multiSendTxs.length) {
    const isValid = await _isValidMultiSend(safe, multiSendTxs.slice(0, idx))
    if (isValid) {
      bestBatch = multiSendTxs.slice(0, idx)
      idx += Math.floor(idx / 2)
    } else {
      idx = Math.floor(idx / 2)
    }
  }
  return bestBatch
```

schmanu (commented):
Killer feature! 🔪 

After reading your comment I think the tenderly solution would be the cleanest because it has a constant runtime. As both our RPC endpoints and Tenderly have rate limits this should make it quite cost efficient too.
The only issue with using tenderly is that it could be feature-toggled off via our config service. So we would need a fallback for when tenderly is not available. Although the fallback could just be to batch all txs.

iamacook:
> After reading your comment I think the tenderly solution would be the cleanest because it has a constant runtime. As both our RPC endpoints and Tenderly have rate limits this should make it quite cost efficient too.

Valid point. I think this needs to be discussed with the wider team as it would have an impact on our Tenderly rate limit. I think we should look into the usage of batching first though.

> The only issue with using tenderly is that it could be feature-toggled off via our config service. So we would need a fallback for when tenderly is not available. Although the fallback could just be to batch all txs.

I totally agree as well.

I'll come back to this shortly.

iamacook:
@schmanu, I opted to use "our" simulation endpoint as the [Tenderly SDK](https://github.com/Tenderly/tenderly-sdk) requires an access key, which we use a proxy for. I confirmed the designs with @liliiaorlenko as well (see the description for an updated screenshot).

src/components/tx/modals/BatchExecuteModal/ReviewBatchExecute.tsx
```diff
@@ -44,28 +48,65 @@ const ReviewBatchExecute = ({ data, onSubmit }: { data: BatchExecuteData; onSubm
...
+      executionOwner: wallet.address,
+      safe,
+      transactions: allMultiSendTxs,
+    })
+
+    // Use default Tenderly API
+    return getSimulation(simulationPayload, undefined)
+  }, [wallet?.address, safe.address.value, allMultiSendTxs])
+
+  const validMultiSendTxs = useMemo(() => {
```
schmanu, line 70: Let's move this logic outside of the component and test it.

src/components/tx/modals/BatchExecuteModal/ReviewBatchExecute.tsx
```diff
@@ -126,7 +181,7 @@ executions from the same Safe."
           </>
         ) : null}
 
-        {multiSendTxs && <TxSimulation canExecute transactions={multiSendTxs} disabled={submitDisabled} />}
+        {allMultiSendTxs && <TxSimulation canExecute transactions={allMultiSendTxs} disabled={submitDisabled} />}
```
schmanu, line 184: The simulation should use the `validMultiSendTxs` instead of `allMultiSendTxs`

schmanu (commented):
I really like the solution using tenderly.

schmanu (approved):
🔥

usame-algan (commented):
Really cool feature!

I agree with @schmanu suggestion to add a fallback in case Tenderly is disabled i.e. when `simulation` stays undefined. Currently, when I block the simulation request via network panel we don't show the user any info about the batch anymore:

Screenshot 2023-04-14 at 13 27 48

iamacook:
> Really cool feature!
> 
> I agree with @schmanu suggestion to add a fallback in case Tenderly is disabled i.e. when `simulation` stays undefined. Currently, when I block the simulation request via network panel we don't show the user any info about the batch anymore:

I missed that suggestion 🙈 I've adjusted it to now return all transactions if the simulation fails:

image

> As another iteration it would also be nice to have a loading state

I've also added a loader:

loading-batch

src/components/tx/modals/BatchExecuteModal/ReviewBatchExecute.tsx
```diff
@@ -44,28 +66,59 @@ const ReviewBatchExecute = ({ data, onSubmit }: { data: BatchExecuteData; onSubm
...
+    })
+
+    // Use default Tenderly API
+    return getSimulation(simulationPayload, undefined)
+  }, [wallet?.address, safe.address.value, allMultiSendTxs])
+
+  const validMultiSendTxs = useMemo(() => {
+    // Simulation failed, return all txs
+    if (simulationError) {
+      console.log('error')
```
usame-algan, line 91:
```suggestion

```

usame-algan (approved):
Thanks for adding the loader! Looks amazing 💯

francovenica:
Issue:
If the very 1st tx will fail, there is basically no batch, but the send button is still available. Clicking the send button will show a 422 error undefined

The Send button should be disabled if the batch tx list is empty. 
image

iamacook:
After [discussion](https://5afe.slack.com/archives/C03DAGWJCR5/p1681814556445119), we will close this for now and assess the approach again when the modal redesign is implemented.

PR #1847, opened 2023-04-05
feat: add relay debug toggle

iamacook:
## What it solves

Aiding in testing of non-relay relay related features.

## How this PR fixes it

A new debug toggle allows enabling/disabling relaying has been added to the sidebar.

## How to test it

Open the Safe and observe the new debug switch which, when toggles, enables/disables relaying in the UI.

## Screenshots

image

relaying-toggle

## Checklist


katspaugh (commented):
Not a fan of adding more dev toggles tbh but not super against it.
As an alternative you could add a copy of Goerli config on staging but with the feature disabled, so it would be like another chain.

usame-algan (approved):
I think this is good as a temporary solution but I would suggest toremove the toggle (and turn off relaying on Görli) after the feature has been released with v1.8.0.

PR #1849, opened 2023-04-05
Fix: a better cookie banner layout

katspaugh:
## What it solves

A taller but more informative and transparent cookie banner.

Changes:
* Removed Necessary cookies (we don't use any)
* A shorter text at the top
* Checkboxes are explained
* "Accept selected" is not outlined

**Note:**
As @usame-algan [found out](https://curia.europa.eu/juris/document/document.jsf;jsessionid=49316CA6F4DC8BF2709229B488B77E04?text=&docid=218467&pageIndex=0&doclang=de&mode=req&dir=&occ=first&part=1&cid=5583083), it's not possible to pre-select the checkboxes.

Screenshot 2023-04-13 at 16 08 02
Screenshot 2023-04-13 at 16 08 22
Screenshot 2023-04-13 at 16 08 34


usame-algan:
What about moving it to the corner and inverting the colors so it is better separated from the main UI?

Screenshot 2023-04-05 at 15 19 03

tschubotz (approved):
Wording wise this change looks good to me. Important from my perspective:

- Clear messaging (This update is better than what we had before)
- Links to cookie policy
- Nothing preselected.
- Accept all button available.

src/components/common/CookieBanner/styles.module.css
```diff
@@ -3,7 +3,7 @@
   position: fixed;
   z-index: 1300;
   bottom: var(--space-2);
-  left: var(--space-2);
+  right: var(--space-2);
```
usame-algan, line 6: I suggest to also change the responsive style from left to right further down in this file

usame-algan (approved):
Looks good 👍

PR #1855, opened 2023-04-12
Add Dcent support

hyunjung222:
## What it solves
This PR adds support for D'cent. 

**Changes**:
  - Add '@web3-onboard/dcent' module
  - Add some style variables to support '@web3-onboard/common' version 2.3.1 or later.

## Screenshots
**light mode**
image

image

**dark mode**
image

image

Since version 2.3.1 of '@web3-onboard/common', the way you modify the UI style is changed, so the button/character color may be slightly different.

## Checklist


johannesmoormann:
thank you for making the PR. we are currently not adding new wallet options to our interfaces but will let you once we are open for new wallet integrations again.

PR #1861, opened 2023-04-13
fix: check simulation call trace for reversion

iamacook:
## What it solves

Resolves successful simulations that should fail.

## How this PR fixes it

The simulation `call_trace` is also checked for errors if the Tenderly simulation is successful. Example simulation can be found [here](https://dashboard.tenderly.co/public/safe/safe-apps/simulator/3076a15b-a3f7-4ffe-b768-2ecf603b0da8/logs).

## How to test it

Open a 1.1.1 Safe on Gnosis Chain and create the following batch in the Transaction Builder:

1. `to` of current Safe, `changeMasterCopy` to `0x3E5c63644E683549055b9Be8653de26E0B4CD36E`
2. `to` of Current Safe, `setMasterCopy` to `0xf48f2B2d2a534e402487b3ee7C18c33Aec0Fe5e4`
3. `to` of `0x7a48Dac683DA91e4faa5aB13D91AB5fd170875bd`, `swapOwner`
4. `to` of `0x7a48Dac683DA91e4faa5aB13D91AB5fd170875bd`, `swapOwner` from to-be-added owner added in 3.

Simulate transaction (in the creation modal) and observe the simulation failing with:

>The transaction failed during the simulation. Full simulation report is available <b>on Tenderly</b>.

## Screenshots

image

## Checklist

### DIFF
--- a/src/components/tx/TxSimulation/SimulationResult.tsx
+++ b/src/components/tx/TxSimulation/SimulationResult.tsx
@@ -38,11 +38,10 @@ export const SimulationResult = ({
     return null
   }
 
-  const isSuccess = !simulation?.simulation.status
+  const isSuccess = simulation?.simulation.status
 
-  const callTraceErrors = getCallTraceErrors(simulation)
-  // Safe can emit `ExecutionFailure` even though Tenderly simulation succeeds
-  const isCallTraceError = !isSuccess && callTraceErrors.length > 0
+  // Safe can emit failure event even though Tenderly simulation succeeds
+  const isCallTraceError = isSuccess && getCallTraceErrors(simulation).length > 0
 
   // Error
   if (requestError || !isSuccess || isCallTraceError) {
@@ -58,19 +57,16 @@ export const SimulationResult = ({
           </Typography>
         ) : (
           <Typography>
-            The transaction failed during the simulation{' '}
             {isCallTraceError ? (
-              <>
-                with error <b>{callTraceErrors[0].error}</b>.
-              </>
+              <>The transaction failed during the simulation.</>
             ) : (
               <>
-                throwing error <b>{simulation?.transaction.error_message}</b> in the contract at{' '}
+                The transaction failed during the simulation throwing error{' '}
+                <b>{simulation?.transaction.error_message}</b> in the contract at{' '}
                 <b>{simulation?.transaction.error_info?.address}</b>.
               </>
             )}{' '}
-            Full simulation report is available
-            <ExternalLink href={simulationLink}>on Tenderly</ExternalLink>.
+            Full simulation report is available <ExternalLink href={simulationLink}>on Tenderly</ExternalLink>.
           </Typography>
         )}
       </Alert>


src/components/tx/TxSimulation/SimulationResult.tsx
```diff
@@ -15,6 +15,14 @@ type SimulationResultProps = {
   onClose: () => void
 }
 
+const getCallTraceErrors = (simulation?: TenderlySimulation) => {
+  if (!simulation) {
+    return []
+  }
+
+  return simulation.transaction.call_trace.filter((call) => call.error)
```
schmanu, line 23: Why do we go for call errors instead of the `ExecutionFailure` event as written in the comment?

We could iterate over `simulation.transaction.logs` and filter by `log.name === 'ExecutionFailure' && log.raw.address === safeAddress` emitted from `safeAddress`.

src/components/tx/TxSimulation/SimulationResult.tsx
```diff
@@ -44,8 +58,18 @@ export const SimulationResult = ({
           </Typography>
         ) : (
           <Typography>
-            The transaction failed during the simulation throwing error <b>{simulation?.transaction.error_message}</b>{' '}
-            in the contract at <b>{simulation?.transaction.error_info?.address}</b>. Full simulation report is available{' '}
+            The transaction failed during the simulation{' '}
+            {isCallTraceError ? (
+              <>
+                with error <b>{callTraceErrors[0].error}</b>.
```
schmanu, line 64: I think this leads to too technical errors. The one in the PRs screenshot for instance reads "value: execution error" which does not really help users IMO.

I would just scan for the event log and give a generic error here OR if we want to give more details on what failed, we could start collecting common `call_trace` errors and map them to human readable error messages. 

For the out of gas error for instance this error is present in the call traces:
`"error": "out of gas"` which we could rephrase to "The transaction failed with an out of gas error. Try increasing the  safeTxGas and try again."

schmanu:
In general I think this is a great addition to our simulations as it is quite confusing that they show up as "Successful" although they revert in the assembly block of our contracts. We could also reach out to tenderly about this issue.

src/components/tx/TxSimulation/SimulationResult.tsx
iamacook: For future reference: after discussing this, differing Safe versions emit different events (<1.1.0 emits `ExecutionFailed`) so the logs aren't consistent.

As such, we've adjusted the code to only extract the call traces if the transaction simulation "succeeds".

src/components/tx/TxSimulation/SimulationResult.tsx
iamacook: For future reference: we've generalised the error messages in accordance with the above clarification.

schmanu (approved):
Nice! This should improve the simulations for 1.1.0 Safes!

DiogoSoaress (approved):
Nice improvement!

JagoFigueroa:
Hola team! The improvement is working nicely. 

Only issue I see is that it will conflict with the current tenderly simulation that it is performed on the app itself as the behaviour there will still be the old one. I don't know if this change should be integrated on the app side as well to avoid cases like this:

https://user-images.githubusercontent.com/92332710/232790831-27f2ab78-1181-45dd-9d10-7fe40258b56c.mp4

Cheers!

iamacook:
> Only issue I see is that it will conflict with the current tenderly simulation that it is performed on the app itself as the behaviour there will still be the old one. I don't know if this change should be integrated on the app side as well

The Transaction Builder needs to be updated to include the changes in this PR. I'll create an issue for it to be tackled now.

JagoFigueroa:
Gracias señor! All good from me to be merged 😎

PR #1868, opened 2023-04-14
fix: Unhandled exception in useOnboard

### LINKED ISSUE #1783: Unhandled exception in useOnboard
`getAddress` needs to be wrapped in a try-catch and the error properly handled.

Screenshot 2023-03-20 at 08 06 50

https://gnosis.sentry.io/issues/3897265821/?project=6769196&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d

deepaksing:
## What it solves
This PR fixes #1783. Unhandled exception in useOnboard

## How this PR fixes it
In this PR i have added a try catch block that handles any error thrown by the getAddress() function inside the try block. If getAddress() function throws an error, the catch block will be executed and the error will be caught and logged to the console. Then, null will be returned as the value of getConnectedWallet() function.

## Checklist

### DIFF
--- a/src/hooks/wallets/useOnboard.test.ts
+++ b/src/hooks/wallets/useOnboard.test.ts
@@ -46,7 +46,7 @@ describe('getConnectedWallet', () => {
     })
   })
 
-  it('Failed to get connected wallet', () => {
+  it('should return null if the address is invalid', () => {
     const wallets = [
       {
         label: 'Wallet 1',
--- a/src/hooks/wallets/useOnboard.test.ts
+++ b/src/hooks/wallets/useOnboard.test.ts
@@ -45,4 +45,24 @@ describe('getConnectedWallet', () => {
       chainId: '4',
     })
   })
+
+  it('Failed to get connected wallet', () => {
+    const wallets = [
+      {
+        label: 'Wallet 1',
+        icon: '',
+        provider: null as unknown as EIP1193Provider,
+        chains: [{ id: '0x4' }],
+        accounts: [
+          {
+            address: '0xinvalid',
+            ens: null,
+            balance: null,
+          },
+        ],
+      },
+    ] as WalletState[]
+
+    expect(getConnectedWallet(wallets)).toBeNull()
+  })
 })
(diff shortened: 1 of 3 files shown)


src/hooks/wallets/useOnboard.ts
```diff
@@ -46,12 +46,18 @@ export const getConnectedWallet = (wallets: WalletState[]): ConnectedWallet | nu
...
+    const address = getAddress(account.address)
+    return {
+      label: primaryWallet.label,
+      address,
+      ens: account.ens?.name,
+      chainId: Number(primaryWallet.chains[0].id).toString(10),
+      provider: primaryWallet.provider,
+    }
+  } catch (error) {
+    console.error(error)
```
iamacook, line 59: We track our errors via Sentry so it would be good to use our logging function here instead. In order to do so, you need add an error message to our [`ErrorCodes`](https://github.com/safe-global/web-core/blob/dev/src/services/exceptions/ErrorCodes.ts) and use `logError` ([example](https://github.com/safe-global/web-core/blob/dev/src/utils/transactions.ts#L186-L188)).

I'd suggest 106, e.g. "106: Failed to get connected wallet".

iamacook (changes requested):
Thank you for putting your time into this, @deepaksing!

We have a test for [`getConnectedWallet`](https://github.com/safe-global/web-core/blob/dev/src/hooks/wallets/useOnboard.test.ts). Could you extend it to cover your changes?

If you need any help, don't hestitate to ask!

src/hooks/wallets/useOnboard.ts
deepaksing: Ok sure, I will work on that

src/hooks/wallets/useOnboard.test.ts
```diff
@@ -45,4 +45,24 @@ describe('getConnectedWallet', () => {
       chainId: '4',
     })
   })
+
+  it('Failed to get connected wallet', () => {
```
iamacook, line 49: We tend to describe the expectation of a test in it's name.

```suggestion
  it('should return null if the address is invalid', () => {
```

iamacook (commented):
I have small suggestion, but otherwise LGTM!

Thanks once again, @deepaksing!

iamacook (approved):

katspaugh (approved):
Another great PR, thanks much!

PR #1870, opened 2023-04-16
Fix: redirect old tx id URLs

### LINKED ISSUE #1869: Old transaction id links don't work
We used to have transaction deep links in this format:
```
{chain prefix}:{safe_address}/transactions/{client_gateway_transaction_id}
```
And this format is still used by the mobile app.

The new format is
```
/transactions/tx?safe={chain prefix}:{safe_address}&id={client_gateway_transaction_id}
```

The old format should redirect to the new format. As of now it's giving a 404.

Also the share link in the tx details should be updated to use the Safe address as a query param.

See https://github.com/safe-global/safe-ios/issues/2480#issuecomment-1509376303

katspaugh:
## What it solves

Resolves #1869

## How this PR fixes it
I've added a condition to rewrite paths containing `/transactions/(something-that-is not-in-the-static-routes)` to `/transactions/tx?id=...`.

## How to test it

Open a link like this:

https://fix_tx_url--webcore.review-web-core.5afe.dev/eth:0xA77DE01e157f9f57C7c4A326eeE9C4874D0598b6/transactions/multisig_0xA77DE01e157f9f57C7c4A326eeE9C4874D0598b6_0x102f72ce18d977a144ddef78c1f35a3102d04e94cc39e3e59d874874f22a7ec2

It should redirect to the new format.

### DIFF
--- a/src/pages/404.tsx
+++ b/src/pages/404.tsx
@@ -18,7 +18,7 @@ export const _getRedirectUrl = (location: Location): string | undefined => {
       const isStaticPath = Object.values(AppRoutes.transactions).some((route) => route === newPath)
       if (!isStaticPath) {
         const txId = newPath.match(/\/transactions\/([^/]+)/)?.[1]
-        newPath = '/transactions/tx'
+        newPath = AppRoutes.transactions.tx
         newSearch = `${newSearch}&id=${txId}`
       }
     }
--- a/src/pages/404.tsx
+++ b/src/pages/404.tsx
@@ -1,20 +1,30 @@
 import { useEffect, useState } from 'react'
 import type { NextPage } from 'next'
 import { useRouter } from 'next/router'
+import { AppRoutes } from '@/config/routes'
 
 // Rewrite the URL to put the Safe address into the query.
-const getRedirectUrl = (): string | undefined => {
-  if (typeof location === 'undefined') return
-
+export const _getRedirectUrl = (location: Location): string | undefined => {
   const { pathname, search } = location
   const re = /^\/([^/]+?:0x[0-9a-f]{40})/i
   const [, pathSafe] = pathname.match(re) || []
 
   if (pathSafe) {
-    const newPath = pathname.replace(re, '') || '/'
+    let newPath = pathname.replace(re, '') || '/'
+    let newSearch = search ? '&' + search.slice(1) : ''
+
+    // TxId used to be in the path, rewrite it to the query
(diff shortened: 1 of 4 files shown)


src/pages/404.tsx
```diff
@@ -1,20 +1,30 @@
...
-    const newPath = pathname.replace(re, '') || '/'
+    let newPath = pathname.replace(re, '') || '/'
+    let newSearch = search ? '&' + search.slice(1) : ''
+
+    // TxId used to be in the path, rewrite it to the query
+    if (newPath.startsWith(AppRoutes.transactions.index)) {
+      const isStaticPath = Object.values(AppRoutes.transactions).some((route) => route === newPath)
+      if (!isStaticPath) {
+        const txId = newPath.match(/\/transactions\/([^/]+)/)?.[1]
+        newPath = '/transactions/tx'
```
iamacook, line 21:
```suggestion
        newPath = AppRoutes.transactions.tx
```

iamacook (approved):
Small suggestion, otherwise LGTM

src/pages/404.tsx
katspaugh: 🏋️🧠

src/pages/404.tsx
```diff
@@ -1,20 +1,30 @@
...
 import { useRouter } from 'next/router'
+import { AppRoutes } from '@/config/routes'
 
 // Rewrite the URL to put the Safe address into the query.
-const getRedirectUrl = (): string | undefined => {
-  if (typeof location === 'undefined') return
-
+export const _getRedirectUrl = (location: Location): string | undefined => {
   const { pathname, search } = location
   const re = /^\/([^/]+?:0x[0-9a-f]{40})/i
```
usame-algan, line 9: Nit: We have a simliar expression in our url utils file called `prefixedAddressRe`. I would suggest moving this there or at least giving it a more descriptive name. I assume it also describes prefixed addresses but less restrictive when it comes to the shortname prefix?

usame-algan (approved):
👍

src/pages/404.tsx
katspaugh: I wouldn't touch this at this time, so as not to break anything that isn't related to the tx id accidentally.

katspaugh: A good mini-refactoring idea tho, feel free to make a tech debt issue.

liliya-soroka:
Verified

PR #1874, opened 2023-04-17
feat: Msg signers list

DiogoSoaress:
## What it solves
- Adds a list with the message signers in the Message modal
- Adds a close dialog to ask the user if he wants to discard the ongoing signatures

## Checklist

### DIFF
--- a/src/components/safe-messages/MsgModal/CloseDialog.tsx
+++ b/src/components/safe-messages/MsgModal/CloseDialog.tsx
@@ -1,0 +1,31 @@
+import { type SetStateAction } from 'react'
+import { Button, Dialog, DialogActions, DialogContent, DialogContentText, DialogTitle } from '@mui/material'
+
+const CloseDialog = ({
+  isOpen,
+  onClose,
+  onConfirm,
+}: {
+  isOpen: boolean
+  onClose: (value: SetStateAction<boolean>) => void
+  onConfirm: () => void
+}) => {
+  return (
+    <Dialog open={isOpen}>
+      <DialogTitle>Are you sure you want to close the modal without a fully signed message?</DialogTitle>
+      <DialogContent>
+        <DialogContentText>Messages that are not fully signed by the multisig NGMI.</DialogContentText>
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={() => onClose(false)} autoFocus>
+          Please no!
+        </Button>
+        <Button onClick={onConfirm} autoFocus>
+          Agree
+        </Button>
+      </DialogActions>
+    </Dialog>
+  )
+}
+
+export default CloseDialog
--- a/src/components/safe-messages/MsgModal/Confirmations.tsx
+++ b/src/components/safe-messages/MsgModal/Confirmations.tsx
@@ -1,0 +1,31 @@
+import { Box, Stack, Typography } from '@mui/material'
+import { SafeMessageStatus, type SafeMessage } from '@safe-global/safe-gateway-typescript-sdk'
+import EthHashInfo from '@/components/common/EthHashInfo'
+
+const Confirmations = ({ message, threshold }: { message?: Omit<SafeMessage, 'type'>; threshold: number }) => {
+  if (!message) return null
+
+  return (
+    <Box
+      mt={2}
+      borderRadius={1}
+      padding={1}
+      border={'1px solid'}
+      borderColor={message?.status === SafeMessageStatus.NEEDS_CONFIRMATION ? 'red' : 'green'}
+    >
+      <Typography fontWeight={700} mb={1}>
+        Confirmations{' '}
+        <Typography component="span" sx={{ color: ({ palette }) => palette.border.main }}>
+          ({`${message?.confirmationsSubmitted || 0} of ${threshold}`})
+        </Typography>
+      </Typography>
+      <Stack direction="column" spacing={2}>
+        {message?.confirmations?.map(({ owner }) => (
+          <EthHashInfo address={owner.value} name={owner.name} key={owner.value} />
+        ))}
+      </Stack>
+    </Box>
+  )
+}
+
+export default Confirmations
--- a/src/components/safe-messages/MsgModal/index.tsx
+++ b/src/components/safe-messages/MsgModal/index.tsx
@@ -1,6 +1,6 @@
 import { Grid, DialogActions, Button, Box, Typography, DialogContent, SvgIcon } from '@mui/material'
-import { useCallback, useMemo, useState } from 'react'
-import { getSafeMessage } from '@safe-global/safe-gateway-typescript-sdk'
+import { useCallback, useEffect, useMemo, useState } from 'react'
+import { getSafeMessage, SafeMessageStatus } from '@safe-global/safe-gateway-typescript-sdk'
 import type { ReactElement } from 'react'
 import type { SafeMessage } from '@safe-global/safe-gateway-typescript-sdk'
 import type { RequestId } from '@safe-global/safe-apps-sdk'
@@ -26,6 +26,7 @@ import { DecodedMsg } from '../DecodedMsg'
 import CopyButton from '@/components/common/CopyButton'
 import { WrongChainWarning } from '@/components/tx/WrongChainWarning'
 import CloseDialog from '@/components/safe-messages/MsgModal/CloseDialog'
+import Confirmations from '@/components/safe-messages/MsgModal/Confirmations'
 
 const APP_LOGO_FALLBACK_IMAGE = '/images/apps/apps-icon.svg'
 const APP_NAME_FALLBACK = 'Sign message off-chain'
@@ -96,6 +97,40 @@ const MsgModal = ({
 
   const isDisabled = !isOwner || hasSigned || !onboard
 
+  const [isPolling, setIsPolling] = useState(false)
+
+  // TODO: move to hook
+  const [ongoingMessage, setOngoingMessage] = useState<Omit<SafeMessage, 'type'>>()
+  const CONFIRMATIONS_POLLING_INTERVAL = 3000
+  useEffect(() => {
+    if (!isPolling) return
+
+    const fetchSafeMessage = async () => {
+      try {
+        const message = await getSafeMessage(safe.chainId, safeMessageHash)
+        setOngoingMessage(message)
+
+        if (message.confirmationsSubmitted === message.confirmationsRequired) {
+          setIsPolling(false)
+          clearInterval(intervalId)
+        }
+
+        console.log('message', message)
+      } catch (e) {
+        // TODO: write a logError
+        console.error(e)
+      }
+    }
+
+    fetchSafeMessage()
+
+    const intervalId = setInterval(() => {
+      fetchSafeMessage()
+    }, CONFIRMATIONS_POLLING_INTERVAL)
+
+    return () => clearInterval(intervalId)
+  }, [isPolling, safe.chainId, safeMessageHash])
+
   const onSign = useCallback(async () => {
     // Error is shown when no wallet is connected, this appeases TypeScript
     if (!onboard) {
@@ -105,18 +140,22 @@ const MsgModal = ({
     setSubmitError(undefined)
 
     try {
+      // If you are connected through WC
       if (requestId && !alreadyProposedMessage) {
         await dispatchSafeMsgProposal({ onboard, safe, message: decodedMessage, requestId, safeAppId })
+        setIsPolling(true)
       } else {
         await dispatchSafeMsgConfirmation({ onboard, safe, message: decodedMessage, requestId })
       }
 
-      onClose()
+      if (ongoingMessage && ongoingMessage?.confirmationsSubmitted === ongoingMessage?.confirmationsRequired - 1) {
+        // TODO: send the fully signed message
+        onClose()
+      }
     } catch (e) {
       setSubmitError(e as Error)
     }
-  }, [alreadyProposedMessage, decodedMessage, onClose, requestId, safe, safeAppId, onboard])
-
+  }, [alreadyProposedMessage, decodedMessage, onClose, requestId, safe, safeAppId, onboard, ongoingMessage])
 
   const handleClose = () => {
     if (!alreadyProposedMessage || alreadyProposedMessage?.status === SafeMessageStatus.NEEDS_CONFIRMATION) {
@@ -164,18 +203,7 @@ const MsgModal = ({
               />
             </Typography>
             <DecodedMsg message={decodedMessage} isInModal />
-            <Box
-              mt={2}
-              sx={{
-                padding: '8px',
-                border: '1px solid',
-                borderColor: alreadyProposedMessage?.status === SafeMessageStatus.NEEDS_CONFIRMATION ? 'red' : 'green',
-              }}
-            >
-              {alreadyProposedMessage?.confirmations.map(({ owner }) => (
-                <EthHashInfo address={owner.value} key={owner.value} />
-              ))}
-            </Box>
+            <Confirmations message={ongoingMessage || alreadyProposedMessage} threshold={safe.threshold} />
             <Typography fontWeight={700} mt={2}>
               SafeMessage:
             </Typography>
--- a/src/components/safe-messages/MsgModal/index.tsx
+++ b/src/components/safe-messages/MsgModal/index.tsx
@@ -25,6 +25,7 @@ import txStepperCss from '@/components/tx/TxStepper/styles.module.css'
 import { DecodedMsg } from '../DecodedMsg'
 import CopyButton from '@/components/common/CopyButton'
 import { WrongChainWarning } from '@/components/tx/WrongChainWarning'
+import CloseDialog from '@/components/safe-messages/MsgModal/CloseDialog'
 
 const APP_LOGO_FALLBACK_IMAGE = '/images/apps/apps-icon.svg'
 const APP_NAME_FALLBACK = 'Sign message off-chain'
@@ -58,6 +59,8 @@ const MsgModal = ({
 }: ProposeProps | ConfirmProps): ReactElement => {
   // Hooks & variables
   const [submitError, setSubmitError] = useState<Error | undefined>()
+
+  const [closeDialogOpen, setCloseDialogOpen] = useState(false)
 
   const onboard = useOnboard()
   const { safe } = useSafeInfo()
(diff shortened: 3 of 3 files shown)


src/components/safe-messages/MsgModal/CloseDialog.tsx
```diff
@@ -0,0 +1,31 @@
...
+}) => {
+  return (
+    <Dialog open={isOpen}>
+      <DialogTitle>Are you sure you want to close the modal without a fully signed message?</DialogTitle>
+      <DialogContent>
+        <DialogContentText>Messages that are not fully signed by the multisig NGMI.</DialogContentText>
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={() => onClose(false)} autoFocus>
+          Please no!
```
katspaugh, line 21: "Continue signing"

src/components/safe-messages/MsgModal/CloseDialog.tsx
```diff
@@ -0,0 +1,31 @@
...
+        </Button>
+        <Button onClick={onConfirm} autoFocus>
+          Agree
```
katspaugh, line 24: "Discard signatures"

src/components/safe-messages/MsgModal/Confirmations.tsx
```diff
@@ -0,0 +1,31 @@
...
+      borderColor={message?.status === SafeMessageStatus.NEEDS_CONFIRMATION ? 'red' : 'green'}
+    >
+      <Typography fontWeight={700} mb={1}>
+        Confirmations{' '}
+        <Typography component="span" sx={{ color: ({ palette }) => palette.border.main }}>
+          ({`${message?.confirmationsSubmitted || 0} of ${threshold}`})
+        </Typography>
+      </Typography>
+      <Stack direction="column" spacing={2}>
+        {message?.confirmations?.map(({ owner }) => (
```
katspaugh, line 23: Please don't reinvent the wheel and use the existing signatures component from tx details. It looks much nicer.

src/components/safe-messages/MsgModal/CloseDialog.tsx
```diff
@@ -0,0 +1,31 @@
+import { type SetStateAction } from 'react'
+import { Button, Dialog, DialogActions, DialogContent, DialogContentText, DialogTitle } from '@mui/material'
+
+const CloseDialog = ({
```
katspaugh, line 4: Can we avoid a popup on top of another popup?
I would show a tooltip next to the submit/close button instead.
Screenshot 2023-04-18 at 09 45 56

src/components/safe-messages/MsgModal/index.tsx
```diff
@@ -93,6 +97,40 @@ const MsgModal = ({
 
   const isDisabled = !isOwner || hasSigned || !onboard
 
+  const [isPolling, setIsPolling] = useState(false)
+
+  // TODO: move to hook
+  const [ongoingMessage, setOngoingMessage] = useState<Omit<SafeMessage, 'type'>>()
```
schmanu, line 103:
```suggestion
  const [ongoingMessage, setOngoingMessage] = useState<Omit<SafeMessage, 'type'>>(alreadyProposedMessage)
```

Currently if the message was already proposed we show an error "Your connected wallet has already signed this message."
I think we should then also start polling for that message.

src/components/safe-messages/MsgModal/index.tsx
```diff
@@ -93,6 +97,40 @@ const MsgModal = ({
...
+      try {
+        const message = await getSafeMessage(safe.chainId, safeMessageHash)
+        setOngoingMessage(message)
+
+        if (message.confirmationsSubmitted === message.confirmationsRequired) {
+          setIsPolling(false)
+          clearInterval(intervalId)
+        }
+
+        console.log('message', message)
```
schmanu, line 118: Here we should emit the `SafeMsgEvent.SIGNATURE_PREPARED` event with the `requestID` and `signature` and listen to that in the `AppFrame`

src/components/safe-messages/MsgModal/index.tsx
```diff
@@ -102,94 +140,109 @@ const MsgModal = ({
...
+        onClose()
+      }
     } catch (e) {
       setSubmitError(e as Error)
     }
-  }, [alreadyProposedMessage, decodedMessage, onClose, requestId, safe, safeAppId, onboard])
+  }, [alreadyProposedMessage, decodedMessage, onClose, requestId, safe, safeAppId, onboard, ongoingMessage])
+
+  const handleClose = () => {
+    if (!alreadyProposedMessage || alreadyProposedMessage?.status === SafeMessageStatus.NEEDS_CONFIRMATION) {
```
schmanu, line 161: I think he have to check the ongoingMessage here
```suggestion
    if (!ongoingMessage || ongoingMessage?.status === SafeMessageStatus.NEEDS_CONFIRMATION) {
```

schmanu (commented):
Looking good 🔥

Can't wait to test this with m/n Safes 🚀

src/components/safe-messages/MsgModal/Confirmations.tsx
DiogoSoaress: I've tried with that component but it felt unnecessary big with the "Created" and Hide/Show options. it is a direct substitution tho. I can introduce it again

katspaugh: Those could be toggled via a new prop.

src/components/safe-messages/MsgModal/CloseDialog.tsx
DiogoSoaress: What is the case we want to warn in the "Submit" button? this modal only has "Sign"

src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -204,7 +204,7 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement
   }, [appName, chainId, closeSignMessageModal, closeTxModal, communicator, signMessageModalState, txModalState])
 
   useEffect(() => {
-    const unsubFns = [SafeMsgEvent.PROPOSE, SafeMsgEvent.CONFIRM_PROPOSE].map((event) => {
+    const unsubFns = [SafeMsgEvent.SIGNATURE_PREPARED].map((event) => {
```
iamacook, line 207: You no longer need loop over this. It should also preserve type as well meaning you can remove some of the logic within the callback.

src/components/tx/SignOrExecuteForm/SignOrExecuteForm.test.tsx
```diff
@@ -572,7 +572,7 @@ describe('SignOrExecuteForm', () => {
     expect(proposeSpy).toHaveBeenCalledTimes(1)
   })
 
-  it('smart contract wallets have to propose when creating a tx with an on-chain signature', async () => {
+  xit('smart contract wallets have to propose when creating a tx with an on-chain signature', async () => {
```
iamacook, line 575: Why are we skipping this now?

src/components/tx/SignOrExecuteForm/SignOrExecuteForm.test.tsx
```diff
@@ -594,7 +594,7 @@ describe('SignOrExecuteForm', () => {
     expect(proposeSpy).toHaveBeenCalled()
   })
 
-  it('smart contract wallets should not propose when on-chain signing an existing transactions', async () => {
+  xit('smart contract wallets should not propose when on-chain signing an existing transactions', async () => {
```
iamacook, line 597: Same as above.

src/hooks/__tests__/useSafeMessageNotifications.test.ts
```diff
@@ -161,7 +161,7 @@ describe('useSafeMessageNotifications', () => {
   it('should show a notification when a message fully is confirmed', () => {
     renderHook(() => useSafeMessageNotifications())
 
-    safeMsgDispatch(SafeMsgEvent.SIGNATURE_PREPARED, { messageHash: '0x012' })
+    safeMsgDispatch(SafeMsgEvent.SIGNATURE_PREPARED, { messageHash: '0x012', requestId: 'test-id', signature: '0x456' })
```
iamacook, line 164: Question: are we no longer allowing on-chain signers at all?

src/hooks/usePollOffchainMessage.ts
```diff
@@ -0,0 +1,52 @@
...
+import { safeMsgDispatch, SafeMsgEvent } from '@/services/safe-messages/safeMsgEvents'
+import { getSafeMessage, type SafeMessage } from '@safe-global/safe-gateway-typescript-sdk'
+import { Errors, logError } from '@/services/exceptions'
+
+const usePollOffchainMessage = (chainId: string, safeMessageHash: string, onClose: () => void, requestId?: string) => {
+  const [ongoingMessage, setOngoingMessage] = useState<SafeMessage>()
+  const [isPolling, setIsPolling] = useState(false)
+  const CONFIRMATIONS_POLLING_INTERVAL = 3000
+
+  useEffect(() => {
```
iamacook, line 11: We could use `useAsync` here and poll as we do for Safe info (with an interval counter).

src/hooks/usePollOffchainMessage.ts
```diff
@@ -0,0 +1,52 @@
...
+    const fetchSafeMessage = async () => {
+      try {
+        // the response has to be a SafeMessage as it is the only type with safeMessageHash
+        const message = (await getSafeMessage(chainId, safeMessageHash)) as SafeMessage
+        setOngoingMessage(message)
+        setIsPolling(true)
+
+        if (message.confirmationsSubmitted === message.confirmationsRequired && message.preparedSignature) {
```
iamacook, line 19: Nit: can we extract this to a helper function?

src/components/safe-messages/MsgModal/index.test.tsx
```diff
@@ -305,7 +306,7 @@ describe('MsgModal', () => {
       Promise.resolve()
     })
 
-    const confirmationSpy = jest.spyOn(sender, 'dispatchSafeMsgConfirmation')
+    const confirmationSpy = jest.spyOn(sender, 'dispatchSafeMsgProposal')
```
schmanu, line 309: This changes what the test wants to test.
The test wants to test the case, that the message already exists and we just add a signature to the existing message.

So instead of changing the expected dispatch function we should mock usePollOffchainMessage to return something similar to how you changed the later test cases.

src/hooks/usePollOffchainMessage.ts
```diff
@@ -0,0 +1,52 @@
...
+const usePollOffchainMessage = (chainId: string, safeMessageHash: string, onClose: () => void, requestId?: string) => {
+  const [ongoingMessage, setOngoingMessage] = useState<SafeMessage>()
+  const [isPolling, setIsPolling] = useState(false)
+  const CONFIRMATIONS_POLLING_INTERVAL = 3000
+
+  useEffect(() => {
+    const fetchSafeMessage = async () => {
+      try {
+        // the response has to be a SafeMessage as it is the only type with safeMessageHash
+        const message = (await getSafeMessage(chainId, safeMessageHash)) as SafeMessage
```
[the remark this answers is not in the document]
iamacook: Are the types incorrect on `safe-gateway-typescript-sdk`? I would suggest we update the package and add a TODO here (not urgent but it needs be done if necessary), or calculate the hash if the types are correct.

src/hooks/usePollOffchainMessage.ts
iamacook: After our discussion: we need not poll here, we already have the messages being polled for. We can reference the store instead.

DiogoSoaress: I removed the single message polling

src/components/tx/SignOrExecuteForm/SignOrExecuteForm.test.tsx
DiogoSoaress: SC wallets are not signing on chain in this branch but i was unsure if should remove the test. Can we discuss this with @schmanu ?

schmanu: I don't think this PR touches the on-chain signing modal. So this test should still work.

In general I would not remove all ability for on-chain signing. It is not decided yet how we go forward with deciding which signing method to use.

src/components/safe-messages/MsgModal/index.tsx
```diff
@@ -80,16 +86,13 @@ const MsgModal = ({
     return messageHash ?? generateSafeMessageHash(safe, decodedMessage)
   }, [messageHash, safe, decodedMessage])
 
-  // Get already proposed message
-  const [alreadyProposedMessage] = useAsync<SafeMessage | Omit<SafeMessage, 'type'>>(() => {
-    const localMessage = messages.page?.results
-      .filter(isSafeMessageListItem)
-      .find((message) => message.messageHash === messageHash)
+  const existingMessage = messages.page?.results
```
schmanu, line 89: If we want to use these messages instead of polling I would suggest reducing the polling interval.
Currently we only poll every 15 seconds. That's a possibly long time to wait. Especially if it would be a 1/n Safe.

src/components/safe-messages/MsgModal/index.tsx
```diff
@@ -102,22 +105,45 @@ const MsgModal = ({
...
     try {
-      if (requestId && !alreadyProposedMessage) {
+      // When collecting the first signature
+      if (requestId && !ongoingMessage) {
         await dispatchSafeMsgProposal({ onboard, safe, message: decodedMessage, requestId, safeAppId })
+        const message = await dispatchPreparedSignature(safe.chainId, safeMessageHash, onClose, requestId)
+        setOngoingMessage(message)
       } else {
         await dispatchSafeMsgConfirmation({ onboard, safe, message: decodedMessage, requestId })
+        dispatchPreparedSignature(safe.chainId, safeMessageHash, onClose, requestId)
```
schmanu, line 115: Only dispatching when we sign will keep users stuck if they already signed  the message.

For instance if I login, logout and login again fast on opensea I will get the same login message to sign. Our previous solution fixed that.

This will also not submit the signature to the dapp if the last signer is a different owner in a different browser.

schmanu (changes requested):
This new implementation breaks the polling.
Only closing the modal if the user signs is not enough.

src/components/safe-messages/MsgModal/index.tsx
schmanu: Actually the polling does not work at all for me when I sign in different browsers:
Screenshot 2023-04-19 at 14 54 45
This screenshot shows the same message in 2 browsers. On the left the initial signer (with WalletConnect app open).
After adding the second signature (see window on the right) the message left still only has 1/2 signatures.

schmanu: I suggest moving it into a useEffect like this for instance
```ts
  useEffect(() => {
    if (ongoingMessage?.preparedSignature) {
      dispatchPreparedSignature(safe.chainId, safeMessageHash, onClose, requestId)
    }
  }, [ongoingMessage, safe.chainId, safeMessageHash, onClose, requestId])
```

src/services/safe-messages/safeMsgSender.ts
```diff
@@ -97,3 +97,30 @@ export const dispatchSafeMsgConfirmation = async ({
     requestId,
   })
 }
+
+const isMessageFullySigned = (message: SafeMessage): message is SafeMessage & { preparedSignature: string } => {
+  return message.confirmationsSubmitted >= message.confirmationsRequired && !!message.preparedSignature
+}
+
+export const dispatchPreparedSignature = async (
```
schmanu, line 105: Why don't we pass in the `message` as it is already present in the modal?

schmanu: Nvm. We want to load the message if we were the last signer to avoid waiting for the polling

src/components/safe-messages/MsgModal/CloseDialog.tsx
schmanu: Let's leave the UX details out of scope for now.
The main purpose of this PR should be to make the flow functional for m/n Safe Accounts.
 
I will create an issue about necessary design / UX changes needed for the synch flow.

schmanu (approved):
It works smoothly now! 
Let's tackle the UX details in a separate PR / issue.

src/components/safe-messages/MsgModal/index.tsx
```diff
@@ -102,22 +110,59 @@ const MsgModal = ({
...
-
-      onClose()
     } catch (e) {
       setSubmitError(e as Error)
     }
-  }, [alreadyProposedMessage, decodedMessage, onClose, requestId, safe, safeAppId, onboard])
+  }, [onboard, requestId, ongoingMessage, safe, decodedMessage, safeAppId, safeMessageHash, onClose])
+
+  const handleClose = useCallback(() => {
+    if (requestId && (!ongoingMessage || ongoingMessage?.status === SafeMessageStatus.NEEDS_CONFIRMATION)) {
```
iamacook, line 141:
```suggestion
    if (requestId && (!ongoingMessage || ongoingMessage.status === SafeMessageStatus.NEEDS_CONFIRMATION)) {
```

src/components/safe-messages/MsgModal/index.tsx
```diff
@@ -102,22 +110,59 @@ const MsgModal = ({
...
+        await dispatchSafeMsgConfirmation({ onboard, safe, message: decodedMessage })
+
+        // No requestID => we are in the confirm message dialog and do not need to leave the window open
+        if (!requestId) {
+          onClose()
+          return
+        }
+
+        // If the last signature was added to the message, we can immediately dispatch the signature
+        if (ongoingMessage?.confirmationsRequired === ongoingMessage.confirmationsSubmitted + 1) {
```
iamacook, line 131:
```suggestion
        if (ongoingMessage?.confirmationsRequired <= ongoingMessage.confirmationsSubmitted + 1) {
```

src/services/safe-messages/safeMsgSender.ts
```diff
@@ -94,6 +88,32 @@ export const dispatchSafeMsgConfirmation = async ({
...
+    const message = (await getSafeMessage(chainId, safeMessageHash)) as SafeMessage
+
+    if (!isMessageFullySigned(message)) return message
+
+    safeMsgDispatch(SafeMsgEvent.SIGNATURE_PREPARED, {
+      messageHash: safeMessageHash,
+      requestId,
+      signature: message.preparedSignature,
+    })
+    onClose()
```
iamacook, line 115: Nit: let's only fetch the message in the `try` and then handle the logic outside of this block, as we do in other dispatchers. Also, why do we need return the `message`?

iamacook (approved):

PR #1880, opened 2023-04-19
fix: allow using the injected wallet on mobile

### LINKED ISSUE #1878: [Mobile] Enable injected wallet if present
We currently skip the first wallet selection screen on mobile and show the WalletConnect modal right away.
This makes sense in iOS Safari, where there's no and cannot be an injected wallet, but is not desired in a webview of mobile wallet apps. They inject their wallet into the webview.

We shouldn't skip to WalletConnect if window.ethereum is defined.

iamacook:
Note: I do not have access to an iOS device and have only tested this on Android. Faking the UA on desktop shows it working on iOS, but it should still be tested on a physical device.

## What it solves

Resolves #1878

## How this PR fixes it

The presence of `window.ethereum` is also checked before automatically connecting to WalletConnect on mobile. If it _is_ present (an inject wallet exists), no autoconnection occurs.

### _Without_ injected wallet
1. Open the Safe on mobile _without_ an injected wallet installed: on Safari (on iPhone) or on Chrome/Firefox with no extensions (on Android).
2. Click connect and observe the WalletConnect QR immediately shown.

### _With_ injected wallet
1. Open the Safe on mobile _with_ an injected wallet installed: on Chrome/Firefox with the Coinbase/MetaMask extension installed (on Android).
2. Click connect and observe the standard wallet modal shown.

### _Without_ injected wallet

without-injected

### _With_ injected wallet

with-injected

## Checklist
* [x] I've tested the branch on *Android* 📱

### DIFF
--- a/src/hooks/wallets/useOnboard.ts
+++ b/src/hooks/wallets/useOnboard.ts
@@ -92,12 +92,7 @@ const trackWalletType = (wallet: ConnectedWallet) => {
 const isMobile = () => /iPhone|iPad|iPod|Android/i.test(navigator.userAgent)
 
 // Detect injected wallet
-const hasInjectedWallet = () => {
-  if (typeof window === 'undefined') {
-    return false
-  }
-  return !!window?.ethereum
-}
+const hasInjectedWallet = () => typeof window !== 'undefined' && !!window?.ethereum
 
 // `connectWallet` is called when connecting/switching wallets and on pairing `connect` event (when prev. session connects)
 // This re-entrant lock prevents multiple `connectWallet`/tracking calls that would otherwise occur for pairing module
--- a/src/hooks/wallets/useOnboard.ts
+++ b/src/hooks/wallets/useOnboard.ts
@@ -91,6 +91,14 @@ const trackWalletType = (wallet: ConnectedWallet) => {
 // Detect mobile devices
 const isMobile = () => /iPhone|iPad|iPod|Android/i.test(navigator.userAgent)
 
+// Detect injected wallet
+const hasInjectedWallet = () => {
+  if (typeof window === 'undefined') {
+    return false
+  }
+  return !!window?.ethereum
+}
+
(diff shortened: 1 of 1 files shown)


usame-algan:
Tested it on iOS with the MM app and it gives me options to choose from 👍 

One small issue I noticed is that there seem to be 2 scrollbars present and sometimes it doesn't let me scroll to the bottom of the list.

src/hooks/wallets/useOnboard.ts
```diff
@@ -91,6 +91,14 @@ const trackWalletType = (wallet: ConnectedWallet) => {
 // Detect mobile devices
 const isMobile = () => /iPhone|iPad|iPod|Android/i.test(navigator.userAgent)
 
+// Detect injected wallet
+const hasInjectedWallet = () => {
+  if (typeof window === 'undefined') {
+    return false
+  }
+  return !!window?.ethereum
```
katspaugh, line 99: Can you please make this one line?

src/hooks/wallets/useOnboard.ts
```diff
@@ -106,8 +114,8 @@ export const connectWallet = async (
 
   isConnecting = true
 
-  // On mobile, automatically choose WalletConnect
-  if (!options && isMobile()) {
+  // On mobile, automatically choose WalletConnect if there is no injected wallet
+  if (!options && isMobile() && !hasInjectedWallet()) {
```
katspaugh, line 118: LGTM! Did you manage to test it with the coinbase wallet?

iamacook: Yes, it works nicely.

katspaugh (approved):
🚀

JagoFigueroa:
Hola que tal. Working as expected for me for both Android (checked Chrome and Firefox) and iOS (Safari). Only different thing is that the WC pop up itself looks a little bit different for me but the behaviour is the expected one so all bueno from me.

Screenshot 2023-04-20 at 16 18 57

PR #1883, opened 2023-04-19
fix: Multiple transactions when creating a safe

### LINKED ISSUE #1881: [safe creation]2 txs instead of 1 when the safe creation executor is changed on retry
## Bug description
[safe creation]2 txs instead of 1 when the safe creation executor is changed on retry.
If the first "safe creation" executor doesn't have enough funds and the executor is changed on Retry, the second executor pays twice for the safe creation. The second executor is asked to confirm 2 txs but one of them will fail.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: any 

## Steps to reproduce
 1. Start safe creation with the executor that doesn't have enough funds for the safe creation 
 2. submit Safe creation on the web and reject tx in MM, as there is not enough funds
 3. change the executor in MM to the one with enough funds
 4. click retry
 5. Confirm both "create proxy" txs in the MM ( 2 requests appear instead of 1 on sage creation in the MM for the second executor)
 Current result: Finally, one of the tx will succeed, and one will fail. The process of safe creation on UI will jump from retry to success, but with the grey "Validating transaction" step. 

## Expected result
Only one "create safe" tx should be proposed for the second executor.

## Screenshots
https://drive.google.com/file/d/1DDTAa2Z1H9xAaV-pLGrV2ois9l0G_iHl/view?usp=share_link

usame-algan:
## What it solves

Resolves #1881 

In #1815, we split up the code of `createSafe` into `createSafeViaRelayer` and `handleCreateSafe` for better readability and moved some of the conditions into the inner function calls as well. Unfortunately, this doesn't work well with how React states update and thus results in multiple inner function calls.

## How this PR fixes it

Reverts the abstraction we built as part of #1815 and moves all conditions back into the main function call.

## How to test it

1. Open the Safe app
2. Create a new safe
3. Reject in Wallet
4. Retry
5. Observe there is only one transaction in the wallet
6. Reject
7. Reload the page
8. Observe the wallet prompting only one transaction

Creating a safe through relaying should still work

## Checklist

### DIFF
--- a/src/components/new-safe/create/steps/StatusStep/useSafeCreation.ts
+++ b/src/components/new-safe/create/steps/StatusStep/useSafeCreation.ts
@@ -5,7 +5,7 @@ import { useCurrentChain } from '@/hooks/useChains'
 import useWallet from '@/hooks/wallets/useWallet'
 import type { EthersError } from '@/utils/ethers-utils'
 import type { PendingSafeData } from '@/components/new-safe/create/steps/StatusStep/index'
-import type { NamedAddress, PendingSafeTx } from '@/components/new-safe/create/types'
+import type { PendingSafeTx } from '@/components/new-safe/create/types'
 import {
   createNewSafe,
   getSafeDeployProps,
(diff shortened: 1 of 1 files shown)


src/components/new-safe/create/steps/StatusStep/useSafeCreation.ts
```diff
@@ -58,29 +58,23 @@ export const useSafeCreation = (
     [setStatus, setPendingSafe],
   )
 
-  const createSafeViaRelayer = useCallback(
-    async (ownersAddresses: string[], threshold: number, saltNonce: number) => {
-      if (!chain) return
+  const handleCreateSafe = useCallback(async () => {
+    if (!pendingSafe || !provider || !chain || !wallet || isCreating) return
```
DiogoSoaress, line 62: `pendingSafe` always get me. I think we could come up with a better naming

DiogoSoaress (approved):
Tested and code looks good to me. This is closer to the previous logic before we abstracted both creation flows.

liliya-soroka:
Looks good for the safe creation with EOA. I want to test safe creation with relay before the ticket is approved.
I am planning to do it , as soon as relay is back ( the issue on Gelato side)

usame-algan:
> I want to test safe creation with relay before the ticket is approved.

Relay is activated on Görli staging so this could be tested there.

PR #1897, opened 2023-04-23
feat: Make it easier to paste eth addresses

### LINKED ISSUE #1631: Make it easier to paste eth addresses
Make it easier to copy&paste addresses in our inputs by allowing pasting any text with an address and cleaning it up.

E.g.
* the user pastes `"Here's my address: 0x1234..."`
* find the `/0x[a-f0-9]{40}/` part
* remove the rest
* If the extracted address doesn't have a prefix, assume eth:. The prefix validation must remain.
* if nothing that looks like an address is found, leave it as it is as it can be an incomplete input or an ENS name

kritarthAviate:
## What it solves

Resolves #1631 

## How this PR fixes it

- Adds a cleanup function `cleanInputValue` which finds a valid address in the input and removes the rest

## How to test it

1. Open the Safe
2. Click on the `New transaction` button on the sidebar
3. Click on the `Send tokens` button
4. Paste a string including a valid Ethereum address
(Eg. - `Here's my address: 0x0000000000000000000000000000000000000000`)
5. It should resolve this string to just the address.

## Checklist

### DIFF
--- a/src/components/common/AddressInput/index.test.tsx
+++ b/src/components/common/AddressInput/index.test.tsx
@@ -83,6 +83,15 @@ describe('AddressInput tests', () => {
     const { input, utils } = setup('')
 
     act(() => {
+      fireEvent.change(input, { target: { value: `eth:${TEST_ADDRESS_A}` } })
+      jest.advanceTimersByTime(1000)
+    })
+
+    await waitFor(() =>
+      expect(utils.getByLabelText(`"eth" doesn't match the current chain`, { exact: false })).toBeDefined(),
+    )
+
+    act(() => {
       fireEvent.change(input, { target: { value: 'gor:0x123' } })
       jest.advanceTimersByTime(1000)
     })
--- a/src/components/common/AddressInput/index.test.tsx
+++ b/src/components/common/AddressInput/index.test.tsx
@@ -83,15 +83,6 @@ describe('AddressInput tests', () => {
     const { input, utils } = setup('')
 
     act(() => {
-      fireEvent.change(input, { target: { value: `eth:${TEST_ADDRESS_A}` } })
-      jest.advanceTimersByTime(1000)
-    })
-
-    await waitFor(() =>
-      expect(utils.getByLabelText(`"eth" doesn't match the current chain`, { exact: false })).toBeDefined(),
-    )
-
-    act(() => {
       fireEvent.change(input, { target: { value: 'gor:0x123' } })
       jest.advanceTimersByTime(1000)
     })
@@ -242,5 +233,22 @@ describe('AddressInput tests', () => {
     })
 
     expect(methods.getValues().recipient).toBe(TEST_ADDRESS_A)
+  })
+
+  it('should clean up the input value if it contains a valid address', async () => {
+    ;(useCurrentChain as jest.Mock).mockImplementation(() => ({
+      shortName: 'gor',
+      chainId: '5',
+      chainName: 'Goerli',
+      features: [],
+    }))
+
+    const { input } = setup(``)
+
+    act(() => {
+      fireEvent.change(input, { target: { value: `Here's my address: ${TEST_ADDRESS_A}` } })
+    })
+
+    await waitFor(() => expect(input.value).toBe(TEST_ADDRESS_A))
   })
 })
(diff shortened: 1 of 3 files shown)


src/components/common/AddressInput/index.test.tsx
```diff
@@ -82,15 +82,6 @@ describe('AddressInput tests', () => {
...
     const { input, utils } = setup('')
 
-    act(() => {
-      fireEvent.change(input, { target: { value: `eth:${TEST_ADDRESS_A}` } })
-      jest.advanceTimersByTime(1000)
-    })
-
-    await waitFor(() =>
-      expect(utils.getByLabelText(`"eth" doesn't match the current chain`, { exact: false })).toBeDefined(),
-    )
```
katspaugh, line 92: As I mentioned, the prefix and its validation shouldn’t be removed.

If the user enters “here’s my address: matic:0x1234…”, it should extract “matic:0x1234”. If it’s not a Polygon safe, an error should be shown.

katspaugh (changes requested):

kritarthAviate:

kritarthAviate:
recheck

src/components/common/AddressInput/index.test.tsx
kritarthAviate: Added back the test case, refactored util function to support any prefix and show error.

src/utils/addresses.ts
```diff
@@ -52,3 +52,19 @@ export const parsePrefixedAddress = (value: string): PrefixedAddress => {
 export const formatPrefixedAddress = (address: string, prefix?: string): string => {
   return prefix ? `${prefix}:${address}` : address
 }
+
+export const cleanInputValue = (value: string): string => {
+  const regex = /(0x[a-fA-F0-9]{40})\b/
+  const regexWithPrefix = /([a-zA-Z0-9]+):(0x[a-fA-F0-9]{40})\b/
```
katspaugh, line 58: You could do it with one regex like so:
```
/(?:([a-z0-9]+):)?(0x[a-f0-9]{40})\b/i
```

Screenshot 2023-05-09 at 06 44 03

Screenshot 2023-05-09 at 06 46 16

Also you could combine test and match in one call:
```
const match = addressRe.match(value)
if (match) return match[0] // address with optional prefix
```

katspaugh (changes requested):
Thanks for the update! Please see my comment.

src/utils/addresses.ts
kritarthAviate: updated logic to use single regex

src/utils/addresses.ts
```diff
@@ -52,3 +52,14 @@ export const parsePrefixedAddress = (value: string): PrefixedAddress => {
 export const formatPrefixedAddress = (address: string, prefix?: string): string => {
   return prefix ? `${prefix}:${address}` : address
 }
+
+export const cleanInputValue = (value: string): string => {
+  const regex = new RegExp('(?:([a-z0-9]+):)?(0x[a-f0-9]{40})\\b', 'i')
+
+  const match = value.match(regex)
+  // if match, return the address with optional prefix
+  if (match?.length) return match[0]
```
katspaugh, line 61: `match` will be null if no match, so no need to check the length.
```suggestion
  if (match?.length) return match[0]
```

src/utils/addresses.ts
```diff
@@ -52,3 +52,14 @@ export const parsePrefixedAddress = (value: string): PrefixedAddress => {
 export const formatPrefixedAddress = (address: string, prefix?: string): string => {
   return prefix ? `${prefix}:${address}` : address
 }
+
+export const cleanInputValue = (value: string): string => {
+  const regex = new RegExp('(?:([a-z0-9]+):)?(0x[a-f0-9]{40})\\b', 'i')
```
katspaugh, line 57: Why this syntax and not just
```suggestion
  const regex = /(?:([a-z0-9]+):)?(0x[a-f0-9]{40})\b/i
```
?

katspaugh (commented):
Looks good! Just two minor suggestions.

katspaugh (approved):
Thank you!

Issue #1809, opened 2023-03-27
[Epic] Transaction risk mitigation 


johannesmoormann:
## What is the feature about

https://www.notion.so/safe-global/Proposal-2-Transaction-risk-mitigation-163d075886684c6da0bed6d9102e198c

PR #1902, opened 2023-04-24
feat: transaction security warnings

schmanu:
## What it solves

#1809

## How this PR fixes it

A module security service has been added, to which payloads can be passed and a severity overview returned. The service consists of indivicual modules and a dispatcher to access them.

## How to test it

tbd

## Screenshots

tbd

## Checklist


src/components/tx/SecurityWarnings/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    case SecuritySeverity.HIGH:
+      return 'Critical risk'
+
+    case SecuritySeverity.LOW:
+      return 'Low risk'
+
+    case SecuritySeverity.NONE:
+      return 'No issues found'
+  }
+}
```
katspaugh, line 33: Can you please make enums or records out of this? Too much syntax for such a simple mapping.

src/components/tx/TxSecurityWarnings/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+      request: {
+        chainId: Number(safe.chainId),
+        safeTransaction,
+        safeAddress,
+        walletAddress: wallet.address,
+        threshold: safe.threshold,
+      },
+    })
+  }, [safe.chainId, safe.threshold, safeAddress, safeTransaction, wallet?.address])
+}
```
katspaugh, line 55: Are we going to wrap and dump every single module here? Can't we generalize this?

src/components/tx/TxSecurityWarnings/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+      {/* Redefine:
+      {redefineScanResult.payload && (
+        <SecurityHint
+        severity={redefineScanResult.severity}
+        text={redefineScanResult.payload?.issues.map((issue) => issue.description).join('\n')}
+        />
+        )}
+        <SecurityWarning severity={redefineScanResult.severity} />
+      */}
+      Unknown Address:
```
katspaugh, line 76: Unknown address:
```suggestion
      Unknown address:
```

src/services/security/modules/RedefineModule/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    params: [string, string]
+  }
+}
+
+type RedefineSeverity = {
+  code: number
+  label: 'CRITICAL' | 'LOW' | 'NO_ISSUES'
+}
+
+type RedefineResponse = {
```
katspaugh, line 31: They don't provide official types?

src/services/security/modules/RedefineModule/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (label === 'CRITICAL') {
+    return SecuritySeverity.HIGH
+  }
+
+  if (label === 'LOW') {
+    return SecuritySeverity.LOW
+  }
+
+  return SecuritySeverity.NONE
+}
```
katspaugh, line 57: Make it a record please.

src/services/security/service.ts
```diff
@@ -0,0 +1,31 @@
...
+  [SecurityModuleNames.REDEFINE]: new RedefineModule(),
+  [SecurityModuleNames.UNKNOWN_ADDRESS]: new UnknownAddressModule(),
+} as const
+
+type Modules = typeof SecurityModules
+
+type Req<T extends SecurityModuleNames> = Modules[T] extends SecurityModule<infer R, unknown> ? R : never
+type Res<T extends SecurityModuleNames> = Modules[T] extends SecurityModule<unknown, infer R> ? R : never
+
+export async function dispatchTxScan<
```
katspaugh, line 23: Why is it called dispatch? It makes me think it's a store action. It just makes a request.

src/services/security/service.ts
```diff
@@ -0,0 +1,31 @@
...
+  Type extends SecurityModuleNames,
+  Request extends Req<Type>,
+  Response extends Res<Type>,
+>({ type, request }: { type: Type; request: Request }) {
+  const InsightModule = SecurityModules[type] as unknown as SecurityModule<Request, Response>
+
+  return InsightModule.scanTransaction(request)
```
katspaugh, line 30: I don't see the point of this dispatcher. Why does a hook need to call a module by a string (`type`) instead of directly importing and calling the module?

src/services/security/modules/RedefineModule/index.ts
schmanu: No. Not yet at least. I asked for a sdk / typescript types ~2 weeks ago.

src/components/tx/TxSecurityWarnings/useRedefine.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!isAnalyzing) {
+      return
+    }
+
+    let timeoutId = setTimeout(() => setRetryCounter((prev) => prev + 1), REDEFINE_RETRY_TIMEOUT)
+    return () => clearTimeout(timeoutId)
+  }, [redefinePayload])
+
+  return redefineResponse
+}
```
katspaugh, line 54: You moved it to a hook but it doesn't change the wrapping and dumping much. I didn't mean that the problem was that they are in the same file.
The problem is that you have:
* A redefine module
* A redefine hook
* Redefine components

All dependent on each other but all in different places.

francovenica:
Small issues/questions:

- How do you login into Redefine? I tried to find a way to create an account by I can't. Doesn't Redefine have a public report like Tenderly?
I know this is not on our end, but pretty sure that most people will try to find how to see the reports. Maybe is there a place we can link them so they can create the account?

- It seems that if you try to send a ton of NFT they are all listed individually, so they could make a really long modal
image

francovenica:
So far so good

Tested in ETH with the tx given seeing they trigger the warnings (the tx still work tho)
I tried regular tx that should not trigger the warnings (Send funds, NFT's, Owner edition, spending limit edition, safe upgrade, modules...)
I tried a few apps, but a lot of them don't work on Stg CGW. Still
Tested in Polygon as well

I'll try to find more tx that might trigger more warning messages.

francovenica:
Sometimes the simulation fails, this is usually when the tx cannot be executed (like if you try to add the same owner twice in a Tx builder batch) and in those cases it would not return any warning as expected since the full report cannot be done. 
Still I found a case when the simulation failed but still returned some reports, so I wanted to discuss if, even if a simulation failed we should call it "simulation failed" when some reports were able to be done:

In this case I'm in ethereum, sending spam tokens I have in a safe to a dubious random contract address:
image

This is how usually tx with the "simulation failed" status look like:
image

usame-algan:
Pushed a small UI tweak (discussed with @liliiaorlenko) for the balance changes since they were overflowing for larger values
Screenshot 2023-06-20 at 17 37 36

src/components/balances/TokenListSelect/index.tsx
```diff
@@ -10,6 +9,7 @@ import { OnboardingTooltip } from '@/components/common/OnboardingTooltip'
 import Track from '@/components/common/Track'
 import { ASSETS_EVENTS, trackEvent } from '@/services/analytics'
 import { HelpCenterArticle } from '@/config/constants'
+import { useHasFeature } from '@/hooks/useChains'
```
iamacook, line 12: There are still a few other instances where we have not replaced `hasFeature` from what I can tell.

Note: this is not related to this line.

src/components/tx/security/shared/LoadingLabel.tsx
```diff
@@ -0,0 +1,16 @@
...
+  return (
+    <Typography variant="body2" color="text.secondary" display="flex" alignItems="center" gap={1} p={2}>
+      <CircularProgress
+        thickness={2}
+        size={24}
+        sx={{
+          color: ({ palette }) => palette.text.secondary,
+        }}
+      />
+      Calculating
```
iamacook, line 13:
```suggestion
      Calculating...
```

src/components/tx/security/shared/SecurityWarnings/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+          />
+        }
+      >
+        {severity !== SecuritySeverity.NONE && <Typography variant="h5">{pluralizedLabel}</Typography>}
+        <Box display="flex" flexDirection="column" gap={2}>
+          {warnings.map((warning) => (
+            <Typography key={warning} variant="body2">
+              {warning}
+            </Typography>
+          ))}
```
iamacook, line 93: If there is more than one warning, shall we add them to a list? Currently there is no clear definition between warnings. What do you think? cc @liliiaorlenko

src/components/tx/security/shared/SecurityWarnings/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+          </Grid>
+        </AccordionSummary>
+        <AccordionDetails>{children}</AccordionDetails>
+      </Accordion>
+      {needsConfirmation && (
+        <Box pl={2}>
+          <Track {...MODALS_EVENTS.ACCEPT_RISK}>
+            <FormControlLabel
+              label="I understand the risks and would like to continue this transaction"
+              control={<Checkbox checked={isConfirmed} onChange={() => setIsConfirmed(!isConfirmed)} />}
```
iamacook, line 166:
```suggestion
              control={<Checkbox checked={isConfirmed} onChange={() => setIsConfirmed((prev) => !prev)} />}
```

src/components/tx/security/shared/SecurityWarnings/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+            />
+          </Track>
+        </Box>
+      )}
+    </Box>
```
iamacook, line 171: Nit: can you separate some of these components. It's not easy to follow.

src/services/security/modules/RedefineModule/index.ts
```diff
@@ -0,0 +1,154 @@
...
+  }[]
+}
+
+export class RedefineModule implements SecurityModule<RedefineModuleRequest, RedefineModuleResponse> {
+  async scanTransaction(request: RedefineModuleRequest): Promise<SecurityResponse<RedefineModuleResponse>> {
+    const { chainId, safeAddress } = request
+
+    const txTypedData = generateTypedData({
+      safeAddress,
+      safeVersion: '1.3.0', // TODO: pass to module, taking into account that lower Safe versions don't have chainId in payload
```
iamacook, line 114: Should this be tackled here?

src/components/tx/security/shared/SecurityWarnings/index.tsx
usame-algan: Good point. Here is an example of how 2 warnings look like. Semantically we should make them list elements but I would also agree to make it more visible as separate elements.
Screenshot 2023-06-21 at 12 20 34

src/services/security/modules/RedefineModule/index.ts
usame-algan: Good question. I remember there being an issue with passing a different safeVersion and it worked even with 1.1.1 safes so we decided to postpone it but this is a good time to revisit this.

src/components/balances/TokenListSelect/index.tsx
usame-algan: I noticed that too but afaik the places where we still use `hasFeature` either need the currentChain dependency for other things or its somewhere where we can't use hooks.

src/components/tx/security/shared/SecurityWarnings/index.tsx
usame-algan: Something like this @liliiaorlenko ?
Screenshot 2023-06-21 at 12 36 30

src/components/tx/security/shared/SecurityWarnings/index.tsx
katspaugh: This only makes sense if you also memoize the function…

src/services/security/modules/RedefineModule/index.ts
iamacook: I guess the `TODO` is a good marker for us to revisit this. We can look at it when porting it to the new flow.

I'll resolve this for the time being

iamacook (approved):

src/components/tx/security/shared/SecurityWarnings/index.tsx
iamacook: I just noticed that it's always pluralised. It should be: "1 more issue" in the "Low issue" box, for example.

usame-algan: Fixed it

PR #1905, opened 2023-04-25
Chore: repace http-server with serve

### LINKED ISSUE #1900: Refreshing on Assets page of a safe shows 404
## Bug description

On Assets page of a Safe, reloading the page adds a trailing slash to `/balances`, and shows up 404.

This does not show up on the Next.js dev server, but only on a static build.

## Environment
 - Browser: Chrome
 - Wallet: Any
 - Chain: Any chain

## Steps to reproduce
 1. Select any safe
 2. Go to Assets page. Notice the page works correctly.
 3. Reload the page. Now it shows 404.

## Expected result

Reloading the page should show up the Assets page as usual, with no added trailing slash to pathname

## Obtained result

404 page comes up.

katspaugh:
## What it solves

Resolves #1900

## How this PR fixes it

Replaces http-server with the serve package from Vercel. It's much faster and doesn't add trailing slashes to directory URLs.

## How to test it
Run `year static-serve`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,8 @@
     "cypress:open": "cross-env TZ=UTC cypress open",
     "cypress:run": "cypress run",
     "cypress:ci": "yarn cypress:run --config baseUrl=http://localhost:8080 --spec cypress/e2e/smoke/*.cy.js",
-    "static-serve": "yarn build && yarn export && yarn serve",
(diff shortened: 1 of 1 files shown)


usame-algan (approved):
Tried it locally and it works 👍

PR #1918, opened 2023-04-26
fix: warn if `safeTxGas` >= block gas limit

### LINKED ISSUE #1795: Incorrect `safeTxGas` used in Safe App transaction simulation
## Bug description

https://github.com/5afe/safe-support/issues/160

When simulating a Thales transaction via WC, our simulation is not using the correct `safeTxGas`.

This is likely related to the fact that the App [can propose the `safeTxGas` via WC](https://github.com/safe-global/safe-apps-sdk/blob/main/packages/safe-apps-provider/src/provider.ts#L81-L98) and it is proposing a high value. We should check that the standard gas limit is higher than that of the `safeTxGas`, e.g. get the gas limit of the last block and check that it is higher than the proposed `safeTxGas`.

`safeTxGas` values are also not good. If possible, we should look into subtracting ~10% of it here (see discussion thread).

## Environment
 - Browser: Chrome
 - Wallet: MetaMask controlling nested Safe via WC
 - Chain: Optimism

## Steps to reproduce

Simulate a Thales transaction on Optimism and observe it failing.

## Expected result

Simulation succeeds.

## Obtained result

Simulation fails because of high `safeTxGas`.

## Links

- [Discussion](https://5afe.slack.com/archives/C03DAGWJCR5/p1679487311386139)
- [Safe simulation](https://dashboard.tenderly.co/public/safe/safe-apps/simulator/ee587701-1e38-4751-a74d-2e6c685a64c8)
- [Manual simulation](https://dashboard.tenderly.co/KirilA/project/simulator/883d4c95-2439-4f9e-ac39-063fb252a878)

DiogoSoaress:
## What it solves

Resolves #1795

## How this PR fixes it
Places a warning in the create transaction modal if `safeTxGas` is greater than or equal to the last block gas limit

## How to test it
1. Create a staking Thales transaction on Optimism and observe the warning in the modal.
2. The Simulation should fail with error code GS010

## Screenshots
Screenshot 2023-05-03 at 21 04 56

## Checklist


DiogoSoaress:
The solution implemented in this PR comes from https://5afe.slack.com/archives/C03DAGWJCR5/p1679487311386139 inputs

mmv08 (approved):

src/components/tx/TxSimulation/utils.ts
```diff
@@ -99,14 +100,20 @@ export const _getSingleTransactionPayload = (
...
+
   const readOnlySafeContract = getReadOnlyCurrentGnosisSafeContract(params.safe)
 
   const input = readOnlySafeContract.encode('execTransaction', [
     transaction.data.to,
     transaction.data.value,
     transaction.data.data,
     transaction.data.operation,
-    transaction.data.safeTxGas,
+    adjustedSafeTxGas || transaction.data.safeTxGas,
```
iamacook, line 116: Shouldn't we also use this value for the transaction? The simulation otherwise differs to what is executed.

iamacook, line 116: Shouldn't we also use this value for the transaction? The simulation otherwise differs to what is executed.

DiogoSoaress: I think we don't want to tamper with the tx parameters passed by the dApp so I'm not confident taking this decision.

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -135,3 +136,20 @@ export const useIsExecutionLoop = (): boolean => {
...
+  const [hasSafeTxGasError, setHasSafeTxGasError] = useState<boolean>()
+
+  useEffect(() => {
+    if (!safeTx) return
+    ;(async function getBlockGasLimit() {
+      const blockGasLimit = await getLatestBlockGasLimit()
+
+      const hasError = blockGasLimit <= safeTx.data.safeTxGas
+      setHasSafeTxGasError(hasError)
+    })()
```
katspaugh, line 151:
```suggestion
    if (!safeTx) return
    const getBlockGasLimit  = async () => {
      const blockGasLimit = await getLatestBlockGasLimit()

      const hasError = blockGasLimit <= safeTx.data.safeTxGas
      setHasSafeTxGasError(hasError)
    }
    getBlockGasLimit()
```

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -209,6 +218,14 @@ const SignOrExecuteForm = ({
           <ErrorMessage>
             Cannot execute a transaction from the Safe Account itself, please connect a different account.
           </ErrorMessage>
+        ) : safeTxGasError ? (
+          <ErrorMessage>
+            You cannot execute a transaction with a <b>safeTxGas</b> equal or higher than the block gas limit. Please
```
katspaugh, line 223:
```suggestion
            You cannot execute a transaction with a <b>safeTxGas</b> equal to or higher than the block gas limit. Please
```

katspaugh (commented):
Seeing how edge-case this is, and that it requires an RPC call, I'd rather not add a custom warning for this. Suggest closing.

DiogoSoaress:
@katspaugh it is indeed an edge case and in this reported issue it reveals the dApp sending an absurd `safeTxGas` value. Besides informing the dApp devs, I think this is an improvement in our app as we make visible an execution error ahead of execution. It comes with _one_ extra RPC call but we could briefly discuss if it is worth fixing it.

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -135,3 +136,22 @@ export const useIsExecutionLoop = (): boolean => {
...
+    if (!safeTx) return
+
+    const getBlockGasLimit = async () => {
+      const blockGasLimit = await getLatestBlockGasLimit()
+      const hasError = blockGasLimit <= safeTx.data.safeTxGas
+      setHasSafeTxGasError(hasError)
+    }
+
+    getBlockGasLimit()
+  }, [safeTx])
```
iamacook, line 154: I'd suggest using `useAsync` for this. We are not checking whether the hook is mounted or if `getLatestBlockGasLimit` throws otherwise.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -209,6 +218,14 @@ const SignOrExecuteForm = ({
           <ErrorMessage>
             Cannot execute a transaction from the Safe Account itself, please connect a different account.
           </ErrorMessage>
+        ) : safeTxGasError ? (
+          <ErrorMessage>
+            You cannot execute a transaction with a <b>safeTxGas</b> equal to or higher than the current block&apos;s
+            gas limit. Please change it on your risk or set it to 0.{' '}
```
iamacook, line 224:
```suggestion
            You cannot execute a transaction with <b>safeTxGas</b> equal to or higher than the current block
            gas limit. Please either set it to 0 or increase the gasLimit at your own risk.{' '}
```

src/components/tx/SignOrExecuteForm/SignOrExecuteForm.test.tsx
```diff
@@ -104,6 +105,7 @@ describe('SignOrExecuteForm', () => {
     } as unknown as ChainInfo)
     jest.spyOn(walletUtils, 'isSmartContractWallet').mockResolvedValue(false)
     jest.spyOn(useRelaysBySafe, 'useRelaysBySafe').mockReturnValue([{ remaining: 5, limit: 5 }, undefined, false])
+    jest.spyOn(signOrExecutionFormUtils, 'getLatestBlockGasLimit').mockImplementation(() => Promise.resolve(15_000_000))
```
iamacook, line 108: Let's add a test to verify that this shows the error as well.

DiogoSoaress:
Closing after responding to support with our agreed action https://github.com/5afe/safe-support/issues/160#issuecomment-1534286291

PR #1919, opened 2023-04-27
feat: execution method selector

### LINKED ISSUE #1894: [Relaying] Allow execution via connected wallet as an alternative when relaying is enabled
## What is the feature about
For networks with relaying enabled, the user should have the choice to execute via their connected wallet as an alternative. 

## The list of requirements

Relay as default
Execution via connected wallet as a visible alternative 

- [x] design @liliiaorlenko 
- [x] frontend implementation 
  - [x] Transaction modal
  - [x] Batch execution
  - [x] Safe creation

iamacook:
## What it solves

Resolves #1894

## How this PR fixes it

The option to choose the execution method is now always present, regardless of remaining relays.

## How to test it

- Ensure relaying is possible.
- Ensure it's possible to execute transactions via the connected wallet (when there are remaining relays).
- Ensure it's possible to execute transactions via the connected wallet (when there are _no_ remaining relays). There should be no option to select the execution method.

## Screenshots

image
Note: the duplicate "Goerli" is there because we have no icon on Goerli.

image

image

## Checklist


src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -175,6 +182,13 @@ const SignOrExecuteForm = ({
           willRelay={willRelay}
         />
 
+        <ExecutionMethodSelector
+          executionMethod={executionMethod}
+          setExecutionMethod={setExecutionMethod}
+          canRelay={canRelay}
```
usame-algan, line 188: This prop is only being used for the rendering logic of `ExecutionMethodSelector`. I would suggest we instead check it here and not pass it further down e.g.

```
{canRelay && <ExecutionMethodSelector ...>}

File: src/services/tx/relaying.ts [hunk 2]
```diff
@@ -19,7 +19,6 @@ type SponsoredCallResponse = {
 export type RelayResponse = {
   limit: number
   remaining: number
-  expiresAt?: number
```
usame-algan, line 22: I assume this was removed on the relay service and we are updating it here?

usame-algan (commented):
Looks good 👍

Just one small thing

usame-algan (approved):

src/services/tx/relaying.ts
iamacook: Yeah, saw it and thought to update it.

src/components/tx/GasParams/index.tsx
```diff
@@ -26,6 +26,7 @@ type GasParamsProps = {
   isEIP1559: boolean
   onEdit: () => void
   gasLimitError?: Error
+  willExecute?: boolean
```
DiogoSoaress, line 29:
```suggestion
  isExecution?: boolean
```

DiogoSoaress: I think it is a better naming + coherent with `<AdvancedParamsForm />`

DiogoSoaress (commented):
Can we also unit test
> when there are no remaining relays, there should be no option to select the execution method.

src/components/tx/GasParams/index.tsx
iamacook: I just realised that it can be totally removed as we have `isExecution` here as well.

francovenica:
So I can fix re-add my e2e test I wrote back then?  😄 

Questions:

I see the "Sponsored by" and it shows the current network. I think before it was always by "gnosis safe" even in goerli. So is ok to show the current network? 
image

Issue:
During the safe proposal, there is a moment where the modal flickers and shows the modal without the relayer for a second:
modal flikr


Besides this the PR looks good

iamacook:
> I see the "Sponsored by" and it shows the current network. I think before it was always by "gnosis safe" even in goerli. So is ok to show the current network?

It should show the current network.

Side note: I know we don't have a logo for Goerli but I don't think is an issue as it's only for development/is disabled on prod.

> During the safe proposal, there is a moment where the modal flickers and shows the modal without the relayer for a second:

The flicker looks like more of a general issue. It looks like it remounts the modal, hence having to load everything again (we don't show the sponsor option until we verify the remaining relays available). I'll look into this and fix/file a new issue if necessary.

PR #1921, opened 2023-04-27
fix: improve relay error

iamacook:
## What it solves

Partially resolves https://github.com/safe-global/safe-gelato-relay-service/issues/74

## How this PR fixes it

The relay error messages are now correctly displayed.

## How to test it

- Try to relay to yourself and observe the "Validation failed" error now shown.

### Transaction modal
image

### Snackbar
image

## Checklist


usame-algan (approved):
Looks good 👍 

While this improves the error message, we still propose and sign a transaction which ends up in the queue. I suggest we create an issue to fix that e.g. by preventing self transactions on the frontend side if relaying is selected in the UI.

iamacook:
> Looks good 👍
> 
> While this improves the error message, we still propose and sign a transaction which ends up in the queue. I suggest we create an issue to fix that e.g. by preventing self transactions on the frontend side if relaying is selected in the UI.

It would be good to block them but I'm not sure if we should extend that beyond the service. Considering we will soon be able to execute via the connected wallet at all times, if we improve the error message returned from the backend, that may be the better option.

What do you think?

usame-algan:
> It would be good to block them but I'm not sure if we should extend that beyond the service.

The problem with the current approach is that the relay call is the last one we do and only then we know if it works or fails but at that point we already propose + sign the transaction which ends up in the queue even though the user sees an error. In addition, the user has to wait quite a while for all of this to happen just so they see an error. Since we have all the necessary data in the frontend to prevent that I think its worth blocking it as soon as possible. To me it seems very similar to the concept of sanitizing form input which ideally is also done on both frontend and backend.

francovenica:
I was going to comment about the tx being proposed and being in the queue, and then I read the discussion. Ignored for now then

The message looks fine, like in the description of the ticket. 

image

PR #1926, opened 2023-05-01
Fix: useTxActions should accept origin

### LINKED ISSUE #1925: SignOrExecute is not passing origin to txActions
Origin isn't passed to proposeTx.

katspaugh:
## What it solves

Resolves #1925

## How this PR fixes it
We didn't notice the bug, because there's a global variable `window.origin` which was masking the undefined `origin`.

`executeTx` and `signTx` now take an `origin` parameter.

### DIFF
--- a/src/components/tx/SignOrExecuteForm/hooks.test.ts
+++ b/src/components/tx/SignOrExecuteForm/hooks.test.ts
@@ -370,7 +370,7 @@ describe('SignOrExecute hooks', () => {
         dynamicPart: () => '',
       })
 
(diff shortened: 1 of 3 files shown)


usame-algan:
How does the bug affect user behaviour and what are the steps to reproduce it? I am not aware of how we used `origin` in the tx flow so far.

katspaugh:
It’s used by Safe Apps for server-side tracking and to show the Safe App’s icon in the queue/history.

iamacook (approved):

katspaugh:
Tested for both signing and executing, the app icon is now passed correctly. ✅

PR #1928, opened 2023-05-02
fix: fetch all balances if chain does not support default tokens

### LINKED ISSUE #1927: Safe apps getBalance uses trusted tokens if default tokens feature is deactivated
## Bug description
If the feature toggle for the default token list is inactive, on the current Sage's chain, the `getBalance` call from safe apps in the `AppFrame` component should not query trusted tokens.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Set the default token setting to true in the local storage for a chain without token lists
 2. Use a safe app that queries getBalances through the safe apps sdk.

## Expected result
All assets of the opened Safe should be returned

## Obtained result
No assets will be returned because no trusted assets exist

schmanu:
## What it solves

Resolves #1927 

## How this PR fixes it
`getBalance` additionally checks if the feature toggle is active for the current chain.

## How to test it
- On a supported chain: Set the token list to default.
- Open the Safe on Gnosis Chain and use the Safe Test app 'getBalance' button.

## Checklist

### DIFF
--- a/src/components/safe-apps/AppFrame/index.tsx
+++ b/src/components/safe-apps/AppFrame/index.tsx
@@ -106,11 +106,13 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement 
       origin: document.location.origin,
     }),
     onGetSafeInfo: useGetSafeInfo(),
-    onGetSafeBalances: (currency) =>
-      getBalances(chainId, safeAddress, currency, {
+    onGetSafeBalances: (currency) => {
+      const isDefaultTokenlistSupported = chain && hasFeature(chain, FEATURES.DEFAULT_TOKENLIST)
+      return getBalances(chainId, safeAddress, currency, {
         exclude_spam: true,
-        trusted: TOKEN_LISTS.TRUSTED === tokenlist,
-      }),
+        trusted: isDefaultTokenlistSupported && TOKEN_LISTS.TRUSTED === tokenlist,
+      })
+    },
     onGetChainInfo: () => {
       if (!chain) return
 


katspaugh (approved):

iamacook (approved):

src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -106,11 +106,13 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement
...
-      getBalances(chainId, safeAddress, currency, {
+    onGetSafeBalances: (currency) => {
+      const isDefaultTokenlistSupported = chain && hasFeature(chain, FEATURES.DEFAULT_TOKENLIST)
+      return getBalances(chainId, safeAddress, currency, {
         exclude_spam: true,
-        trusted: TOKEN_LISTS.TRUSTED === tokenlist,
-      }),
+        trusted: isDefaultTokenlistSupported && TOKEN_LISTS.TRUSTED === tokenlist,
+      })
+    },
```
iamacook, line 115: Just looked back over the codebase and we're using the same logic in `useLoadBalances`.

What do you think about using its return value or abstracting that logic it into a simpler hook to use in both places?

schmanu: I also thought about simply returning the balances from our `balanceSlice`.
That would also be much faster and they get updated / polled regularly.

But I wasn't sure if there is a specific reason we fetch the balances fresh from our service here.

katspaugh:
@schmanu the deployment link doesn't work for some reason. 404 not found.

katspaugh:
I've tested it with the Drain Safe app on Gnosis Chain, and it works correctly now. The tokens that were previously not shown are shown now. ✅

PR #1935, opened 2023-05-03
Fix: track QR button in address inputs

katspaugh:
## What it solves

I'd like to re-evaluate the usefulness of a QR scanner on desktop. Adding a tracker to see how many users click on it.


iamacook (approved):

francovenica:
I use the QR quite a bit for testing, makes finding a recipient easier 😄

PR #1937, opened 2023-05-03
Fix: improve "wrong chain" switching UX

katspaugh:
## What it solves

I see two problems with the current wrong chain UX:

* The chain switch button in the header is unnecessary – the user doesn't need to switch their chain until they perform a transaction action
* The wrong chain warning in the tx modal is not actionable. It's red and scary but doesn't tell the user what to do to fix the problem.

## What changed
* There was a "Switch to X" button in the header – not anymore
* There was a red warning in the modal – it's now ℹ️ and blue

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Deploy PR branch
         if: github.event.number
         env:
-          BUCKET: s3://${{ secrets.AWS_REVIEW_BUCKET_NAME }}/webcore/${{ steps.extract_branch.outputs.branch }}
+          BUCKET: s3://${{ secrets.AWS_REVIEW_BUCKET_NAME }}/walletweb/${{ steps.extract_branch.outputs.branch }}
         run: bash ./scripts/github/s3_upload.sh
 
       # Comnment
@@ -90,7 +90,7 @@ jobs:
             ## Branch preview
             ✅  Deploy successful!
 
-            https://${{ steps.extract_branch.outputs.branch }}--webcore.review-web-core.5afe.dev
+            https://${{ steps.extract_branch.outputs.branch }}--walletweb.review-wallet-web.5afe.dev
           message-failure: |
             ## Branch preview
             ❌  Deploy failed!
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
-# <img src="https://user-images.githubusercontent.com/381895/186411381-e05075ff-7565-4b4e-925e-bb1e85cb165b.png" height="60" width="60" valign="middle" /> Safe Web Core
+# <img src="/public/images/logo.svg" height="60" valign="middle" alt="Safe{Wallet}" />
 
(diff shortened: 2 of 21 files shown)


iamacook (commented):
LGTM but the placement feels a bit off inside the warning.

As we switch the chain, informing the user makes sense to me. Perhaps we could change the colour/icon to be that of the `info` variant? I'd even consider removing the `ChainSwitcher` entirely:

image

@liliiaorlenko, do you have any thoughts?

liliiaorlenko:
> LGTM but the placement feels a bit off inside the warning.
> 
> As we switch the chain, informing the user makes sense to me. Perhaps we could change the colour/icon to be that of the `info` variant? I'd even consider removing the `ChainSwitcher` entirely:
> 
> image
> 
> @liliiaorlenko, do you have any thoughts?

I think this version is clear and not intrusive. If it comes to switching networks our goal is to make experience as seamless as possible. The colour of the warning in yellow perhaps is scary, yes, in blue is more calm and rather informative. If we do automatic chain switch, then we inform user of the interface change and thats it. I can customise the warning so it resemble network switch closer. Maybe text can be improved too: "Currently you wallet is at the wrong chain. When pressing submit we automatically change the network and you can continue on your transaction". I think warning should be more of what is happening right now and how its gonna change after you press SUBMIT. Ill prepare the designs shortly.

liliiaorlenko:
WDYT
https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=5361%3A143633&t=5XEwoU3ALBpHHmAW-1
I designed banner a bit more pleasant and less error like. We gonna use our current connected wallet logo, a bit simpler text with fast recognisable title.

katspaugh:
I think that's a bit... blown out of proportion. I raised this whole topic again because it was already too in-your-face.

Screenshot 2023-05-04 at 14 24 45

katspaugh:
I agree, and this is a good solution. Thanks.

iamacook (commented):
LGTM but there is a failing test.

iamacook (approved):

francovenica:
LGTM

The button is gone and the message is present in the modal
I assume nothing about the switching of the network was changed, but still I checked that it still works fine with MM, Mobile pairing, MM and Trezor.

francovenica:
Note:
The button in the welcome page is still actually useful to change to a network you might have safes in (the sidebar only shows Owned networks of the network you are currently connected to). For MM is easy to change networks, but for mobile pairing or the MM phone app is much easier just to click the "Switch to NETWORK" button and see a prompt in your phone that actually enter into the apps and change the network.
TLDR; can we leave the button for the welcome page?

katspaugh:
Good point, Franco!

katspaugh:
@francovenica to address your concern, I've made the network selector stay on the welcome page. So you can change the app's network to load owned safes on that network.

Also we've added a Connect Wallet button in the sidebar.

francovenica:
Works really nice. Thanks

PR #1938, opened 2023-05-04
fix: decoding txs without parameters

schmanu:
## What it solves
Transactions that call functions without parameters, for instance wrapping a native token. 
Lead to uncaught errors in our DecodeTx component.

## How this PR fixes it
Checks if the first parameter is `undefined`.

## How to test it
Try to wrap xdai to wxdai on gnosis chain by doing a `deposit()` call (e.g. by using cowswap or tx builder)

## Checklist

### DIFF
--- a/src/components/tx/DecodedTx/index.test.tsx
+++ b/src/components/tx/DecodedTx/index.test.tsx
@@ -82,8 +82,6 @@ describe('DecodedTx', () => {
       return Promise.resolve()
     })
 
-    result.debug()
-
     expect(result.queryByText('transfer')).toBeInTheDocument()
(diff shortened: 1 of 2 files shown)


src/components/tx/DecodedTx/index.test.tsx
```diff
@@ -82,8 +82,6 @@ describe('DecodedTx', () => {
       return Promise.resolve()
     })
 
-    result.debug()
```
katspaugh, line 85: Oops, I left it there.

katspaugh (approved):
Thanks, nice catch!

PR #1942, opened 2023-05-04
Fix: sandbox safe app icons in tx list

### LINKED ISSUE #1940: Safe app icons visible in the history set cookies
## Bug description

App Icons in the history/queue need to be sandboxed the same way as was done in https://github.com/safe-global/safe-wallet-web/pull/1193 since they might be loaded from external domains that might set cookies

## Steps to reproduce
 1. Go to the transaction history
 2. Find a transaction that was done on Uniswap where the Uniswap icon is also displayed
 3. Observe cookies being set from uniswap

## Expected result

No cookies should be set from external domains

## Obtained result

Cookies are set from external domains

## Screenshots
Screenshot 2023-05-04 at 13 45 59

katspaugh:
## What it solves

Resolves #1940

## How this PR fixes it

I've replaced the icon with a SafeAppCardIcon (a bad name in this case).

### DIFF
--- a/src/components/safe-apps/SafeAppIconCard/index.tsx
+++ b/src/components/safe-apps/SafeAppIconCard/index.tsx
@@ -2,14 +2,14 @@ import { type ReactElement, memo } from 'react'
 
 const APP_LOGO_FALLBACK_IMAGE = `/images/apps/app-placeholder.svg`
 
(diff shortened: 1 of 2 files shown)


usame-algan (approved):
Looks good 👍

PR #1947, opened 2023-05-04
Eip 1271 backward compatibility

### LINKED ISSUE #1943: [Sync EIP-1271] Ensure backward compatibility
Part of #1886 

## Overview
Some Safe apps might still expect on-chain signatures. For those we want to ensure that it is still possible to sign on-chain as a fallback.

## Requirements
- [x] Use on-chain signatures for safe apps that use  safe-apps-sdk version  `<7.11.0`, which does not include [this change](https://github.com/safe-global/safe-apps-sdk/pull/458#issuecomment-1534829055) yet
- [x] Offer a setting to always use on-chain signatures. We should add a info text which tells the users possible implications of this.

schmanu:
## What it solves

Resolves #1943 

## How this PR fixes it
- [x] Uses on-chain signing for safe-apps-sdk version `<7.11.0`
- [x] Adds a Setting to always sign on-chain for backward compatibility

New analytics event:
- Change "Use on-chain signatures" setting

## How to test it
- Add the new Wallet Connect [PR deployment](https://pr657--safereactapps.review-react-hr.5afe.dev/wallet-connect/) as custom app
- Open a dapp that requires signatures (e.g. opensea login)
- Witness that by default it uses off-chain signing
- Change the setting to on-chain signing
- Witness that it now uses on-chain signing

## Screenshots
Screenshot 2023-05-05 at 11 19 56

## Checklist

### DIFF
--- a/src/components/safe-apps/AppFrame/index.tsx
+++ b/src/components/safe-apps/AppFrame/index.tsx
@@ -37,7 +37,7 @@ import { safeMsgSubscribe, SafeMsgEvent } from '@/services/safe-messages/safeMsg
 import { useAppSelector } from '@/store'
 import { selectSafeMessages } from '@/store/safeMessagesSlice'
 import { isSafeMessageListItem } from '@/utils/safe-message-guards'
-import { supportsEIP1271 } from '@/utils/safe-messages'
+import { isOffchainEIP1271Supported } from '@/utils/safe-messages'
 import PermissionsPrompt from '@/components/safe-apps/PermissionsPrompt'
 import { PermissionStatus } from '@/components/safe-apps/types'
 
@@ -92,8 +92,10 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement 
       message: string | EIP712TypedData,
       requestId: string,
       method: Methods.signMessage | Methods.signTypedMessage,
+      sdkVersion: string,
     ) => {
-      openSignMessageModal(message, requestId, method, !!settings.offChainSigning)
+      const isOffChainSigningSupported = isOffchainEIP1271Supported(safe, chain, sdkVersion)
+      openSignMessageModal(message, requestId, method, isOffChainSigningSupported && !!settings.offChainSigning)
     },
     onGetPermissions: getPermissions,
     onSetPermissions: setPermissionsRequest,
@@ -130,17 +132,12 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement 
       }
     },
     onSetSafeSettings: (safeSettings: SafeSettings) => {
-      const isEIP1271Supported = supportsEIP1271(safe) && chain && hasFeature(chain, FEATURES.EIP1271)
       const newSettings: SafeSettings = {
         ...settings,
-        offChainSigning: isEIP1271Supported && !!safeSettings.offChainSigning,
+        offChainSigning: !!safeSettings.offChainSigning,
       }
 
       setSettings(newSettings)
-
-      if (!isEIP1271Supported && safeSettings.offChainSigning) {
-        console.warn('The connected Safe Account does not support off-chain signing.')
-      }
 
       return newSettings
     },
--- a/src/components/safe-apps/AppFrame/index.tsx
+++ b/src/components/safe-apps/AppFrame/index.tsx
@@ -65,7 +65,7 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement 
   const [signMessageModalState, openSignMessageModal, closeSignMessageModal] = useSignMessageModal()
   const { safe, safeLoaded, safeAddress } = useSafeInfo()
   const tokenlist = useAppSelector(selectTokenList)
-  const useOnChainSigning = useAppSelector(selectOnChainSigning)
+  const onChainSigning = useAppSelector(selectOnChainSigning)
 
   const addressBook = useAddressBook()
   const chain = useCurrentChain()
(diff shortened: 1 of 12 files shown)


src/components/settings/SafeAppsSigningMethod/index.tsx
```diff
@@ -0,0 +1,52 @@
+import ExternalLink from '@/components/common/ExternalLink'
+import { SETTINGS_EVENTS, trackEvent } from '@/services/analytics'
+import { useAppDispatch, useAppSelector } from '@/store'
+import { selectUseOnChainSigning, setUseOnChainSigning } from '@/store/settingsSlice'
+import { FormControlLabel, Checkbox, Paper, Typography, FormGroup, Grid } from '@mui/material'
+
+export const SafeAppsSigningMethod = () => {
+  const useOnChainSigning = useAppSelector(selectUseOnChainSigning)
```
DiogoSoaress, line 8: I would also suggest to rename the variables that read from redux to not cause confusion with hooks

DiogoSoaress (commented):
Screenshot 2023-05-08 at 11 04 41

an UI detail

src/components/settings/SafeAppsSigningMethod/index.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+        } as any,
+      },
+    })
+
+    const checkbox = result.getByRole('checkbox')
+    expect(checkbox).not.toBeChecked()
+
+    act(() => fireEvent.click(checkbox))
+
+    expect(checkbox).toBeChecked()
```
DiogoSoaress, line 21: can we also checked the Redux state is correct?

DiogoSoaress: Or maybe this test could start with `useOnChainSigning: true`

DiogoSoaress (changes requested):

src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -92,8 +93,11 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement
       message: string | EIP712TypedData,
       requestId: string,
       method: Methods.signMessage | Methods.signTypedMessage,
+      sdkVersion: string,
     ) => {
-      openSignMessageModal(message, requestId, method, !!settings.offChainSigning)
+      const isOffChainSigningSupported = isOffchainEIP1271Supported(safe, chain, sdkVersion)
+      const signOffChain = isOffChainSigningSupported && !useOnChainSigning
+      openSignMessageModal(message, requestId, method, signOffChain && !!settings.offChainSigning)
```
DiogoSoaress, line 100:
```suggestion
      openSignMessageModal(message, requestId, method, signOffChain)
```

Do we still need to keep the state variable https://github.com/safe-global/safe-wallet-web/blob/56c74af38ef6a33306a0e19a31752979ccc81105/src/components/safe-apps/AppFrame/index.tsx#L60-L63 ?

src/components/settings/SafeAppsSigningMethod/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <Typography variant="h4" fontWeight="bold" mb={1}>
+            Signing method
+          </Typography>
+        </Grid>
+
+        <Grid item xs>
+          <Typography mb={2}>
+            This setting determines how your Safe Account will sign message requests from Safe Apps. By default it is
+            preferred to use gasless off-chain signatures. In case some Dapps only support on-chain signatures the
+            default behavior can be changed with this setting. Learn more about message signing{' '}
```
iamacook, line 30: I'm not sure about this description as it is a wallet-wide change for _all_ Safe Apps. I would simplify a bit (and explain that it is {Wallet}-based)

```suggestion
            This setting determines how the Safe{Wallet} will sign message requests from Safe Apps. Gasless, off-chain signing is used by default. Learn more about message signing{' '}
```

src/store/settingsSlice.ts
```diff
@@ -38,6 +38,9 @@ export type SettingsState = {
     darkMode?: boolean
   }
   env: EnvState
+  signing: {
+    useOnChainSigning: boolean
```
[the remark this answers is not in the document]
iamacook: I'd suggest `onChingSigning` as it matches the semantics of the `safeSettings`. (Remember to change other instances to match this, including the action/selectors.)

src/components/safe-apps/AppFrame/index.tsx
iamacook: These are needed. The `settings` (preferences) are [set by the Safe App](https://docs.safe.global/learn/safe-core/safe-core-protocol/signatures/eip-1271#enabling-off-chain-signing).

src/store/settingsSlice.ts
```diff
@@ -60,6 +63,9 @@ const initialState: SettingsState = {
       accessToken: '',
     },
   },
+  signing: {
+    useOnChainSigning: false,
+  },
```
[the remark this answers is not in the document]
schmanu: I like the idea. But it would be a bigger change. I'd say lets see if users will run into the problem that specific Safe apps only support on-chain signing and then add it if needed.

src/components/settings/SafeAppsSigningMethod/index.test.tsx
schmanu: I'm not sure if it is possible to access the redux state which we pass in to the `customRender`
I did not find any test which does that.

src/store/settingsSlice.ts
```diff
@@ -91,6 +97,9 @@ export const settingsSlice = createSlice({
     setEnv: (state, { payload }: PayloadAction<EnvState>) => {
       state.env = payload
     },
+    setUseOnChainSigning: (state, { payload }: PayloadAction<boolean>) => {
```
iamacook, line 100: Nit: this should also be renamed to `setOnChangeSigning`.

src/store/settingsSlice.ts
```diff
@@ -129,3 +139,5 @@ export const selectRpc = createSelector(selectSettings, (settings) => settings.e
 export const selectTenderly = createSelector(selectSettings, (settings) => settings.env.tenderly)
 
 export const isEnvInitialState = createSelector(selectSettings, (settings) => isEqual(settings.env, initialState.env))
+
+export const selectUseOnChainSigning = createSelector(selectSettings, (settings) => settings.signing.onChainSigning)
```
iamacook, line 143: Nit: this should also be renamed to `selectOnChainSigning`.

iamacook (approved):

PR #1949, opened 2023-05-05
Refactor: executeTx hook

katspaugh:
## What it solves

A small refactor/fix for the executeTx function.
Before, it was always first proposing the tx if there's no txId. Now, it first does a signature for relayed txs which might create an id, then proposes.

Basically, this avoids two proposals in a row for "immediately relayed" transactions.

## How to test
Case 1:
* Relay a transaction in a 1/1 Safe
* The "Transaction proposed successfully" notification should pop-up only once, after signing the tx (before it twice – was before and after signing)

Case 2:
* Execute a transaction in a 1/1 Safe
* The "Transaction proposed successfully" notification should pop-up before executing in the wallet (like before)

### DIFF
--- a/src/components/tx/SignOrExecuteForm/hooks.ts
+++ b/src/components/tx/SignOrExecuteForm/hooks.ts
@@ -96,22 +96,25 @@ export const useTxActions = (): TxActions => {
       assertWallet(wallet)
       assertOnboard(onboard)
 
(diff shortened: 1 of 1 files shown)


iamacook (approved):

francovenica:
LGTM

Tested with the prod CGW enabled in GNO

For relayer the "Tx was proposed" notification now pops up only after signing with your wallet and not before
For regular wallet execution still works like before (which is expected)

PR #1953, opened 2023-05-05
Fix: support injected wallet icons

katspaugh:
## What it solves
Injected wallets can have a custom icon, which we were ignoring. I've added it as a prop, so it's now displayed.

Before:
Screenshot 2023-05-05 at 14 09 34

After:
Screenshot 2023-05-05 at 14 06 01

## How to test

* Connect with a custom wallet like Rabby or Zerion
* Check the wallet info in the header
* Check the Pending Actions in the Safe List
* Check the icon in the Relaying vs Wallet select

### DIFF
--- a/src/components/common/WalletIcon/index.tsx
+++ b/src/components/common/WalletIcon/index.tsx
@@ -1,53 +1,4 @@
 import { Skeleton } from '@mui/material'
-import metamaskIcon from '@web3-onboard/injected-wallets/dist/icons/metamask'
-import coinbaseIcon from '@web3-onboard/coinbase/dist/icon'
-import keystoneIcon from '@web3-onboard/keystone/dist/icon'
-import walletConnectIcon from '@web3-onboard/walletconnect/dist/icon'
-import trezorIcon from '@web3-onboard/trezor/dist/icon'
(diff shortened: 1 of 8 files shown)


src/components/common/WalletIcon/index.tsx
```diff
@@ -49,14 +49,24 @@ const getWalletIcon = (provider: string) => {
...
+  width = 30,
+  height = 30,
+  icon,
+}: {
+  provider: string
+  width?: number
+  height?: number
+  icon?: string
+}) => {
+  const svg = getWalletIcon(provider) || icon
```
iamacook, line 63: Maybe out of scope but I just tried inverting this and it works nicely (although there's a hydration issue).

Do you think it's still worth keeping `getWalletIcon`?

iamacook (approved):

src/components/common/WalletIcon/index.tsx
katspaugh: Yeah, good point, we could kill it and not import all those icons if onboard handles that.

katspaugh: Tested with different wallets, icons still work. ✅

PR #1968, opened 2023-05-09
fix: clear next pages when switching Safe

iamacook:
## What it solves

Resolves stale transaction and messages from previous Safe.

## How this PR fixes it

When switching Safe, every page of transactions/messages after the initial is cleared.

## How to test it

1. Load a subsequent page of transactions on the history/queue or messages and switch Safe.
2. Observe no stale transactions/messages from the previous Safe.

## Checklist


katspaugh:
Any other places in the app where this could be a problem? Dashboard? Maybe we should solve it for all pages by adding the  safeAddress + chainId as a key prop on the <App> component?

src/components/common/PaginatedTxns/index.tsx
```diff
@@ -69,7 +69,7 @@ const PaginatedTxns = ({ useTxns }: { useTxns: typeof useTxHistory | typeof useT
   const [pages, setPages] = useState<string[]>([''])
   const [filter] = useTxFilter()
 
-  // Reset the pages when the filter changes
+  // Reset the pages when the filter/Safe changes
```
katspaugh, line 72:
```suggestion
  // Reset the pages when the filter changes
```

src/components/safe-messages/PaginatedMsgs/index.tsx
```diff
@@ -69,7 +69,7 @@ const PaginatedMsgs = (): ReactElement => {
   }
 
   return (
-    <Box mb={4} position="relative">
```
katspaugh, line 72: Necessary?

src/pages/_app.tsx
```diff
@@ -108,7 +110,7 @@ const WebCoreApp = ({
           <InitApp />
 
           <PageLayout pathname={router.pathname}>
-            <Component {...pageProps} />
+            <Component {...pageProps} key={safe} />
```
katspaugh, line 113:
```suggestion
            <Component {...pageProps} key={router.query.safe.toString()} />
```

katspaugh (approved):
Thanks!

src/components/safe-messages/PaginatedMsgs/index.tsx
iamacook: Leftover from before using the `key`. I've reverted it.

katspaugh (approved):
👍

iamacook:
> A nish regarding the "Messages" tab. Some safes don't have the tab (whatever is cuz messages are not supported or there are no messages). If you are in a safe that has the message tab and switch to another that doesn't have it, you re "still" put in the messages tab. I'd suggest kicking the user to the history tab, which is the default tab when you click on "transactions" in the sidebar
> 
> image

I've created [an issue](https://github.com/safe-global/safe-wallet-web/issues/1972) for it as it is unrelated.

francovenica:
Then it looks good. you can merge it into the RC

PR #1969, opened 2023-05-09
fix: improve disable condition and tooltip tile on `TxSummary` buttons

### LINKED ISSUE #1932: Transaction error when trying to execute too fast
## Bug description

If the user tries to enter the transaction flow before that it fails because the Safe Core SDK has not been initialized yet.

We could disable transaction interactions until the Safe Core SDK instance is ready.

## Environment
 - Browser: any
 - Wallet: any
 - Chain: any

## Steps to reproduce
 1. Go to the Queue
 2. Reload the page
 3. Immediately try to Execute a queue transaction

## Expected result

There should be no error

## Obtained result

There is an error saying the safe core SDK has not been initialized

## Screenshots

Screenshot 2023-05-03 at 12 25 45

DiogoSoaress:
## What it solves

Resolves #1932

## How this PR fixes it
Includes the `coreSDK` initialization in the condition to disable the `<TxSummary>` buttons (`ExecuteTxButton`, `RejectTxButton` and `SignTxbutton`)
Improve tooltip description as per the disable condition

### Not next tx
1. Go to a Tx queue where a queued tx is fully signed but not the "Next" in the queue
2. Observe the execution icons is disabled and show a tooltip "Transaction {nonce} must be executed first"

### ~Transaction is pending execution~
~1. Go to a Tx queue and have 2 transactions ready to be executed
2. Execute the first one
3. Observe the execution icons in the second one is disabled and show a tooltip "Pending transaction must first succeed"~

### Safe SDK not initialized
1. Go to a Tx queue where there is a tx queued in the "next" nonce
2. Refresh the page
3. Try to hover the Execute/Reject/Sign buttons immediately
4. Observe they are disabled and show a tooltip 'Waiting for the SDK to initialize'

## Screenshots
Screenshot 2023-05-09 at 18 13 36

## Checklist

### DIFF
--- a/src/components/transactions/ExecuteTxButton/index.tsx
+++ b/src/components/transactions/ExecuteTxButton/index.tsx
@@ -33,7 +33,7 @@ const ExecuteTxButton = ({
   const isNext = txNonce !== undefined && txNonce === safe.nonce
   const isDisabled = !isNext || isPending || !safeSDK
 
-  const tooltipTitle = getTxButtonTooltip('Execute', { isNext, nonce: safe.nonce, isPending, hasSafeSDK: !!safeSDK })
+  const tooltipTitle = getTxButtonTooltip('Execute', { isNext, nonce: safe.nonce, hasSafeSDK: !!safeSDK })
 
   const onClick = (e: SyntheticEvent) => {
     e.stopPropagation()
--- a/src/components/transactions/ExecuteTxButton/index.tsx
+++ b/src/components/transactions/ExecuteTxButton/index.tsx
@@ -14,6 +14,7 @@ import { TX_LIST_EVENTS } from '@/services/analytics/events/txList'
 import { ReplaceTxHoverContext } from '../GroupedTxListItems/ReplaceTxHoverProvider'
 import CheckWallet from '@/components/common/CheckWallet'
 import { useSafeSDK } from '@/hooks/coreSDK/safeCoreSDK'
+import { getTxButtonTooltip } from '@/components/transactions/utils'
 
 const ExecuteTxButton = ({
   txSummary,
@@ -32,14 +33,7 @@ const ExecuteTxButton = ({
   const isNext = txNonce !== undefined && txNonce === safe.nonce
   const isDisabled = !isNext || isPending || !safeSDK
 
-  const tooltipTitle =
-    isDisabled && !isNext
-      ? `Transaction ${safe.nonce} must be executed first`
-      : isPending
-      ? 'Pending transaction must first succeed'
-      : !safeSDK
-      ? 'Waiting for the SDK to initialize'
-      : 'Execute'
+  const tooltipTitle = getTxButtonTooltip('Execute', { isNext, nonce: safe.nonce, isPending, hasSafeSDK: !!safeSDK })
 
   const onClick = (e: SyntheticEvent) => {
     e.stopPropagation()
--- a/src/components/transactions/ExecuteTxButton/index.tsx
+++ b/src/components/transactions/ExecuteTxButton/index.tsx
@@ -13,7 +13,7 @@ import Track from '@/components/common/Track'
 import { TX_LIST_EVENTS } from '@/services/analytics/events/txList'
 import { ReplaceTxHoverContext } from '../GroupedTxListItems/ReplaceTxHoverProvider'
 import CheckWallet from '@/components/common/CheckWallet'
-import { getSafeSDK } from '@/hooks/coreSDK/safeCoreSDK'
+import { useSafeSDK } from '@/hooks/coreSDK/safeCoreSDK'
 
 const ExecuteTxButton = ({
   txSummary,
@@ -27,7 +27,7 @@ const ExecuteTxButton = ({
   const txNonce = isMultisigExecutionInfo(txSummary.executionInfo) ? txSummary.executionInfo.nonce : undefined
   const isPending = useIsPending(txSummary.id)
   const { setSelectedTxId } = useContext(ReplaceTxHoverContext)
-  const safeSDK = getSafeSDK()
+  const safeSDK = useSafeSDK()
 
   const isNext = txNonce !== undefined && txNonce === safe.nonce
   const isDisabled = !isNext || isPending || !safeSDK
--- a/src/components/transactions/ExecuteTxButton/index.tsx
+++ b/src/components/transactions/ExecuteTxButton/index.tsx
@@ -13,6 +13,7 @@ import Track from '@/components/common/Track'
 import { TX_LIST_EVENTS } from '@/services/analytics/events/txList'
 import { ReplaceTxHoverContext } from '../GroupedTxListItems/ReplaceTxHoverProvider'
 import CheckWallet from '@/components/common/CheckWallet'
+import { getSafeSDK } from '@/hooks/coreSDK/safeCoreSDK'
 
 const ExecuteTxButton = ({
   txSummary,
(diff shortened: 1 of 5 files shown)


src/components/transactions/ExecuteTxButton/index.tsx
```diff
@@ -26,9 +27,19 @@ const ExecuteTxButton = ({
   const txNonce = isMultisigExecutionInfo(txSummary.executionInfo) ? txSummary.executionInfo.nonce : undefined
   const isPending = useIsPending(txSummary.id)
   const { setSelectedTxId } = useContext(ReplaceTxHoverContext)
+  const safeSDK = getSafeSDK()
```
iamacook, line 30: `getSafeSDK` will not react to changes in `ExternalStore` state. We should use the hook when accessing the value.

src/components/transactions/ExecuteTxButton/index.tsx
```diff
@@ -26,9 +27,19 @@ const ExecuteTxButton = ({
...
 
   const isNext = txNonce !== undefined && txNonce === safe.nonce
-  const isDisabled = !isNext || isPending
+  const isDisabled = !isNext || isPending || !safeSDK
+
+  const tooltipTitle =
+    isDisabled && !isNext
+      ? `Transaction ${safe.nonce} must be executed first`
```
iamacook, line 37: I don't think we need to show this, maybe even the button at all.

src/components/transactions/ExecuteTxButton/index.tsx
```diff
@@ -26,9 +27,19 @@ const ExecuteTxButton = ({
...
+      : isPending
+      ? `There's a pending transaction`
```
iamacook, line 39: I would be more descriptive here, e.g.

```suggestion
      ? `You cannot execute while while a transaction is pending.`
```

src/components/transactions/ExecuteTxButton/index.tsx
```diff
@@ -26,9 +27,19 @@ const ExecuteTxButton = ({
...
+      : !safeSDK
+      ? 'Waiting for the SDK to initialize'
```
iamacook, line 41: I'm not sure about this - it's somewhat technical for the "standard" user. It shouldn't happen to too long either.

src/components/transactions/ExecuteTxButton/index.tsx
usame-algan, line 39: I like these new messages but would suggest to document them in the test plan of this PR. If possible, I would also suggest to extract this into a helper function that can be reused in all three buttons and write a few unit tests for it.

iamacook (commented):
I've added some comments to the `ExecuteTxButton` but they stand for each. I'd suggest consulting with design about this.

A simpler solution would be disabling the submit button in the transaction `SignOrExecute` modal when the SDK is initialising.

src/components/transactions/ExecuteTxButton/index.tsx
DiogoSoaress: I was trying to juggle between a more complete description and a more succinct one given it's a tooltip. Said that are you still in favor of the longer one?

src/components/transactions/ExecuteTxButton/index.tsx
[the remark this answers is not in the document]
DiogoSoaress: Nice catch! I was thinking of using the hook but for reason I used the getter and it worked so I've propagated it. It was a distraction

src/components/transactions/ExecuteTxButton/index.tsx
DiogoSoaress: I've answered in https://github.com/safe-global/safe-wallet-web/pull/1969/files#r1189586069

src/components/transactions/ExecuteTxButton/index.tsx
iamacook: Compared to the `title` on line 37, I think it needs more information. An alternative: "Pending transaction must first succeed".

DiogoSoaress:
> A simpler solution would be disabling the submit button in the transaction SignOrExecute modal when the SDK is initialising.

That is already too late, the error appears when loading the values in that modal. That's why I think we should prevent opening it.

src/components/transactions/ExecuteTxButton/index.tsx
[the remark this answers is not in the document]
iamacook: I would drop the SDK altogether, e.g. "Initialising Safe".

iamacook:
> That is already too late, the error appears when loading the values in that modal. That's why I think we should prevent opening it.

You could suppress the error until the SDK has initialised.

DiogoSoaress:
> You could suppress the error until the SDK has initialised/failed to initialise.

The way we render `ErrorMessage` would delay any error until the SDK initialized/failed to initialize. But what I meant was that in order to display the tx data in the modal we rely on the SDK being initialized. See https://github.com/safe-global/safe-wallet-web/blob/3bc22a0425adf1e7c0ddc0a648b744051f95e48d/src/components/tx/modals/ConfirmTxModal/ConfirmProposedTx.tsx#L32-L46

src/components/transactions/ExecuteTxButton/index.tsx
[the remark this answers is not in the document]
DiogoSoaress: But that's not quite what's happening... do you have any suggestion other than "Loading" @usame-algan ?

src/components/transactions/ExecuteTxButton/index.tsx
DiogoSoaress: Extracted the util and added the tests

src/components/transactions/ExecuteTxButton/index.tsx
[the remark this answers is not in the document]
iamacook: The SDK is an inherent part of the Safe though.

src/components/transactions/utils.ts
```diff
@@ -0,0 +1,27 @@
+type DisabledPropsType =
+  | {
+      isNext: false
+      nonce: number
+      isPending: boolean
+      hasSafeSDK: boolean
+    }
+  | {
+      isNext?: boolean
+      nonce?: number
```
iamacook, line 10:
```suggestion
      isNext?: never
      nonce?: never
```

src/components/transactions/__tests__/utils.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      expect(getTxButtonTooltip(enabledTitle, disabledProps)).toBe('Transaction 2 must be executed first')
+    })
+
+    it('should return the "is tx pending" message', () => {
+      const enabledTitle = 'Execute'
+      const disabledProps = { ...disabledPropsBase, isPending: true }
+
+      expect(getTxButtonTooltip(enabledTitle, disabledProps)).toBe('Pending transaction must first succeed')
+      expect(true).toBe(true)
```
iamacook, line 27:
```suggestion
```

src/components/transactions/ExecuteTxButton/index.tsx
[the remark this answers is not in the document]
usame-algan: Imo Loading is generic enough and still better than what we had before. We could also add no Tooltip in that case and just disable the button but I wouldn't mention any technical things there like the SDK.

francovenica:
Issue:
The tx, while is being executed, shows the "Pending tx must succeed first", giving the impression that the button will be available at some point, but actually that button should never be since the tx will go away after being executed.
Or we don't show nothing at all in the tx that is being currently executed
Or we show a "Tx in execution"

image

DiogoSoaress:
@usame-algan @francovenica What do you think of changing that tooltip to "Transaction in execution" ? dropping the "must execute first" that can be misleading.
Or, should we simply not show a tooltip when we have a pending tx?

katspaugh:
Can we please wrap this up? The amount of discussion and attention this fix receives starts to worry me.

francovenica:
I think not showing the tooltip in a executing tx is fine

francovenica:
I still see tooltips while the tx is executing, maybe because the status is not "pending" but "processing" or "relaying".

Still I don't think is that big of a deal. I think we can leave it this way until we get feedback about it.
image
image

DiogoSoaress:
> I still see tooltips while the tx is executing

These tooltips ("Execute", "Replace", "Confirm") were already in place and were not affected by this bug fix/improvement.

francovenica:
Yes, I think we are done with this ticket. I'll pass it

Issue #1974, opened 2023-05-10
Switching between Safes on high resolutions loads pages of previous Safe


iamacook:
## Bug description

On high resolutions (capable of showing >1 page of transactions on screen), switching _from_ a Safe that has more than a few pages _to_ a different one, the previous infinite scroll is loaded.

Note: this was tested _including_ https://github.com/safe-global/safe-wallet-web/pull/1968.

## Steps to reproduce

 1. Open Safe A (which has 2+ pages of transactions) on a high resolution.
 2. Ensure only the first 2 pages are loaded.
 3. Switch to Safe B (which has no transactions).
 4. Observe page 3 of Safe A loading.

## Expected result

Only transactions of Safe B are loaded.

## Obtained result

Safe A infinite scroll is persisted.

## Screenshots

![large-res](https://github.com/safe-global/safe-wallet-web/assets/20442784/696ad803-88c1-4749-b33f-5a7459805cb9)

PR #1976, opened 2023-05-11
Fix: useAsync should clear data if no promise

katspaugh:
## What it solves

#1974 

The `useAsync` hook with a third param set to false currently caches data even if the callback does an early exit.
It should only cache data between reloads, not when the callback returns (doesn't matter a promise or not).
If the callback returns no promise, it should be treated as if it returns a promise that resolves to undefined, just w/o the async part.

## How to test it
`useAsync` with `false` as a third param is used in places where we poll info but don't want to clear the data between polls:
* Single tx page (because we poll the tx details when the queue/history tags update)
* Expanded tx details in the tx list (for the same reason)
* Approval editor (polls balances)
* Gas price
* Pending txs queue
* Regular queue
* Messages queue
* History
* Assets

### DIFF
--- a/src/hooks/loadables/useLoadBalances.ts
+++ b/src/hooks/loadables/useLoadBalances.ts
@@ -14,15 +14,17 @@ export const useLoadBalances = (): AsyncResult<SafeBalanceResponse> => {
   const currency = useAppSelector(selectCurrency)
   const settings = useAppSelector(selectSettings)
   const chain = useCurrentChain()
+
   const isTrustedTokenList = useMemo(() => {
-    const hasTrustedList = chain !== undefined && hasFeature(chain, FEATURES.DEFAULT_TOKENLIST)
+    if (chain === undefined) return
+    const hasTrustedList = hasFeature(chain, FEATURES.DEFAULT_TOKENLIST)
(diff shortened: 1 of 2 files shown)


src/hooks/loadables/useLoadBalances.ts
```diff
@@ -14,15 +14,17 @@ export const useLoadBalances = (): AsyncResult<SafeBalanceResponse> => {
...
   const isTrustedTokenList = useMemo(() => {
-    const hasTrustedList = chain !== undefined && hasFeature(chain, FEATURES.DEFAULT_TOKENLIST)
+    if (chain === undefined) return
+    const hasTrustedList = hasFeature(chain, FEATURES.DEFAULT_TOKENLIST)
     return hasTrustedList && (!settings.tokenList || settings.tokenList === TOKEN_LISTS.TRUSTED)
   }, [chain, settings.tokenList])
 
   // Re-fetch assets when the entire SafeInfo updates
   const [data, error, loading] = useAsync<SafeBalanceResponse | undefined>(
-    async () => {
```
katspaugh, line 24: This async was here as a workaround for the problem this PR solves.

iamacook (approved):

katspaugh:
I'll merge it into another opened PR.

PR #1981, opened 2023-05-11
refactor: upgrade + optimise Redux

iamacook:
## What it solves

Upgrades Redux to the latest version and optimises/improves type safety of middlewares.

## How this PR fixes it

- `react-redux` and `@reduxjs/toolkit` have been upgraded to the latest versions.
- All midleware (apart from persistence-related ones) have been migrated to [`createListenerMiddleware`](https://redux-toolkit.js.org/api/createListenerMiddleware)

## How to test it

No visible changes should be noticed:
- Added Safes in the sidebar should remain up-to-date (balance).
- Pending transaction signatures/executions should clear successfully.
- Pending messages should clear when signed successfully.

## Checklist


src/store/index.ts
```diff
@@ -90,7 +88,10 @@ export const _hydrationReducer: typeof rootReducer = (state, action) => {
 const makeStore = (initialState?: Record<string, any>) => {
   return configureStore({
     reducer: _hydrationReducer,
-    middleware: (getDefaultMiddleware) => getDefaultMiddleware({ serializableCheck: false }).concat(middleware),
+    middleware: (getDefaultMiddleware) => {
+      listeners.forEach((listener) => listener(listenerMiddlewareInstance))
```
usame-algan, line 92: Is there a difference in registering the listeners in this function or on the top level? In the examples from the documentation they are registered on the top level right after the middleware is created.

usame-algan (approved):
Looks good 👍 
I like the better type support and reduced nesting with the new middleware.

src/store/index.ts
iamacook: I don't think so.

I moved it here as this is just before store creation/allowed for better testing. What are your thoughts?

francovenica:
LGTM

Checked that sidebar updates with the safes, new safes created, added safes, owned safes in new networks
Checked the lists queue history and message. Tx update just fine after creating/executing them
Tried with messages on and off chain, they update fine as well
Tried pagination, pagination with filter, tx updating after switching safes

katspaugh:
@iamacook please don't merge just yet, let's create an RC w/o it first.

PR #1982, opened 2023-05-12
Fix: env-dependent legal pages

katspaugh:
## What it solves

As pointed out by @fmrsabino, legal pages like Terms and Imprint should be customized by Safe forks according to their own policies, and CC GmbH's name and policies should not be used by default.

With this PR, forks will have to modify the code to insert their own legal content. Otherwise, the footer and the legal pages will be empty.

Screenshot 2023-05-12 at 09 20 08


iamacook (approved):

.env.example
```diff
@@ -30,3 +30,6 @@ NEXT_PUBLIC_CYPRESS_MNEMONIC=
 # Safe Gelato relay service
 NEXT_PUBLIC_SAFE_GELATO_RELAY_SERVICE_URL_PRODUCTION=
 NEXT_PUBLIC_SAFE_GELATO_RELAY_SERVICE_URL_STAGING=
+
+# Company name
+NEXT_PUBLIC_SAFE_WALLET_HOST = 'app.safe.global'
```
usame-algan, line 35: Should this be `NEXT_PUBLIC_APP_HOST`?

src/components/common/Footer/index.tsx
```diff
@@ -37,38 +38,45 @@ const Footer = (): ReactElement | null => {
...
-          </Link>
-          &nbsp;&mdash;&nbsp;
-          <Link href={`${router.pathname}#`} passHref>
-            <MUILink onClick={onCookieClick}>Preferences</MUILink>
-          </Link>
-        </li>
+        {IS_OFFICIAL_HOST ? (
+          <>
+            <li>
+              <Typography variant="caption">&copy;2022–{new Date().getFullYear()} Core Contributors GmbH</Typography>
```
katspaugh, line 44: We could put this into an env var as well, then forks won't have to modify this code.

.env.example
katspaugh: Removed, it was a leftover.

katspaugh:
Going to merge directly, as there's no visual change. Tested on the branch preview and it works like before. ✅

PR #1985, opened 2023-05-12
Fix: don't refresh main component on first page load

### LINKED ISSUE #1980: Two balances requests
On the Assets page, when Balances are polled, it makes two almost simultaneous requests instead of just one.
Both are initiated by `useLoadBalances()`.

Screenshot 2023-05-11 at 14 48 15

## How to reproduce it
* Open the Assets page
* Open the Devtools -> Network
* Wait for 15 seconds – observe two xhr requests

katspaugh:
## What it solves

Resolves #1980

This was a regression from #1968.

## How this PR fixes it

The `router.query` is initially empty. This is to avoid hydration issues with the static build which doesn't have any query params by definition.

On safe-specific routes, `router.query.safe` then gets populated with the value from the URL, thus going from undefined to a string. This causes the main page component to re-render due to the `key` prop.

I've added a hook that only changes the key if the previous value was set.

### DIFF
--- a/src/hooks/loadables/useLoadBalances.ts
+++ b/src/hooks/loadables/useLoadBalances.ts
@@ -14,15 +14,17 @@ export const useLoadBalances = (): AsyncResult<SafeBalanceResponse> => {
   const currency = useAppSelector(selectCurrency)
   const settings = useAppSelector(selectSettings)
   const chain = useCurrentChain()
+
   const isTrustedTokenList = useMemo(() => {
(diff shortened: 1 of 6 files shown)


iamacook (approved):

francovenica:
I got 1 call after refreshing normally and then 1 call in the first 15 secs polling, but after that every 15 secs I always got 2 calls for balances:

In the gif I reload manually, clear the storage and wait 15 secs twice, clearing the storage every time
calls

katspaugh:
@francovenica fixed 👍

PR #1988, opened 2023-05-15
feat: use new CGW URL

DiogoSoaress:
The aim of this PR is to test the new CGW endpoint in the web app


iamacook:
I've confirmed with QA that we can close this as we're now using the NestJS gateway on dev.

PR #1989, opened 2023-05-15
fix: merge custom env vars per chain

### LINKED ISSUE #1978: Setting a custom RPC overwrites existing RPC settings on other networks
## Bug description

Setting a custom RPC on one network and then on another will overwrite the setting on the first network.

The warning in the sidebar is also shown across networks even if that network still has the default RPC.

## Steps to reproduce
 1. Go to `/settings/environment-variables` on Görli
 2. Set a custom RPC Provider 
 3. Switch to a safe on Mainnet
 4. Observe a warning is still showing in the sidebar
 5. Set a custom RPC Provider
 6. Switch back to Görli
 7. Observe that there is no custom RPC provider anymore

## Expected result

Custom RPC providers should be persisted for each network and the warning should only be visible per network

## Obtained result

Only the last set provider is persisted

iamacook:
## What it solves

Resolves #1978

## How this PR fixes it

Setting custom env. vars. no longer fully overwrite the `EnvState`. RPC/Tenderly settings now have their own actions:
- `setRpc` only sets the current chain
- `setTenderly` only sets the Tenderly settings

The sidebar hint is now only shown when there is either a custom RPC for the current chain or custom Tenderly settings.

### RPCs
1. Set a custom RPC on Goerli and observe the warning in the sidebar.
2. Switch to another network and observe no warning.

### Tenderly
1. Set custom Tenderlys ettings and observe the warning in the sidebar.
2. Switch to another network and observe the warning.

## Checklist

### DIFF
--- a/src/components/settings/EnvironmentVariables/EnvHintButton/index.tsx
+++ b/src/components/settings/EnvironmentVariables/EnvHintButton/index.tsx
@@ -6,10 +6,12 @@ import { useAppSelector } from '@/store'
 import { isEnvInitialState } from '@/store/settingsSlice'
 import css from './styles.module.css'
 import AlertIcon from '@/public/images/common/alert.svg'
+import useChainId from '@/hooks/useChainId'
 
 const EnvHintButton = () => {
   const router = useRouter()
-  const isInitialState = useAppSelector(isEnvInitialState)
+  const chainId = useChainId()
+  const isInitialState = useAppSelector((state) => isEnvInitialState(state, chainId))
 
   if (isInitialState) {
     return null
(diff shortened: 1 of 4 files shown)


usame-algan (approved):
Looks good 👍

usame-algan (approved):
Looks good 👍

francovenica:
It looks good for the RPC, now every chain can have its own without overwriting the other networks.

Question: for tenderly it seems that all networks share the same URL. Is this expected? 
If it is then we can pass the ticket

iamacook:
> Question: for tenderly it seems that all networks share the same URL. Is this expected? If it is then we can pass the ticket

Yes, we use the same API key for simulations normally as well.

PR #1997, opened 2023-05-17
test: E2E remove NFT tooltip assertion

DiogoSoaress:
## What it solves
Removes a leftover E2E test after removing the NFT links tooltip

## How this PR fixes it
- Removes a the tooltip test
- ~Comments out a fiat currency selector test relying on a failing CGW endpoint.~


katspaugh (approved):
👍

DiogoSoaress:
The https://safe-client.staging.5afe.dev/v1/balances/supported-fiat-codes endpoint is operational again. I've reverted the skipped test

PR #1999, opened 2023-05-17
Feat: general app settings in the footer

### LINKED ISSUE #1994: No way to change settings on non-Safe-specific URLs like /welcome
We have a number of settings pages that are not Account-specific:
* Env vars
* Appearance
* Data import/export

However, they can be currently only accessed via Account-specific settings.
It makes it impossible to change them before opening or creating a Safe.

## Proposed solution

Create a general Settings page, `/settings` and link it in the footer.

To save space in the footer, remove the Preferences from the Cookie Policy link and move the Cookie preferences to a tab on the new Settings page.

Screenshot 2023-05-17 at 08 33 26

katspaugh:
## What it solves

Resolves #1994

## How this PR fixes it

Makes the Settings page work for not Account-specific URLs.

## How to test it
Case 1:
* Open Welcome
* Go to Settings via the footer

Case 2:
* Open a Safe
* Go to Safe settings via the sidebar menu

Case 3:
* Check that the Cookie banner still works and looks like before

## Screenshots
Screenshot 2023-05-17 at 12 48 59


src/components/common/Footer/index.tsx
```diff
@@ -65,11 +57,12 @@ const Footer = (): ReactElement | null => {
             </li>
             <li>
               <Link href={AppRoutes.cookie} passHref>
-                <MUILink>Cookie Policy</MUILink>
+                <MUILink>Cookie policy</MUILink>
               </Link>
-              &nbsp;&mdash;&nbsp;
-              <Link href={`${router.pathname}#cookie-preferences`} passHref>
-                <MUILink onClick={onCookieClick}>Preferences</MUILink>
```
usame-algan, line 72: With this removed, its not possible to update the cookie preferences anymore unless going to the non-safe settings page. Is there a reason to move the cookie preferences there or could we still keep it as a popup like it was?

src/pages/settings/index.tsx
```diff
@@ -1,3 +1,26 @@
...
+import Head from 'next/head'
+import { useRouter } from 'next/router'
+import { generalSettingsNavItems, settingsNavItems } from '@/components/sidebar/SidebarNavigation/config'
 
-export default SetupPage
+const Settings: NextPage = () => {
+  const router = useRouter()
+
+  useEffect(() => {
+    const redirectPath = router.query.safe ? settingsNavItems[0].href : generalSettingsNavItems[0].href
```
usame-algan, line 11: When I am on the safe settings page and I click on "Settings" in the Footer, the general settings page opens and takes me out of my safe. I would suggest redirecting to `settingsNavItems[0].href` in that case as well so it doesn't navigate away from the current safe.

src/components/common/PageLayout/index.tsx
```diff
@@ -27,11 +26,17 @@ const isNoSidebarRoute = (pathname: string): boolean => {
...
-  const noSidebar = isNoSidebarRoute(pathname)
+  const router = useRouter()
+  const [noSidebar, setNoSidebar] = useState<boolean>(isNoSidebarRoute(pathname))
   const [isSidebarOpen, setSidebarOpen] = useState<boolean>(true)
   let isAnimated = useDebounce(!noSidebar, 300)
   if (noSidebar) isAnimated = false
 
+  useEffect(() => {
+    const noSafeAddress = router.isReady && !router.query.safe
+    setNoSidebar(isNoSidebarRoute(pathname) || noSafeAddress)
```
usame-algan, line 37: Afaik all the paths in `isNoSidebarRoute` are non-safe routes. Could we instead add the new settings path there as well or remove that check entirely and always depend on `!router.query.safe` to hide the sidebar?

src/components/common/Footer/index.tsx
katspaugh: You can still click on Settings in the footer from any page where the footer is displayed, and it will open cookie prefs.

src/pages/settings/index.tsx
katspaugh: If you want to go to the Safe settings, you must click on "Settings" in the Safe nav sidebar.

I could rename Settings in the footer to Preferences to avoid confusion?

src/components/common/PageLayout/index.tsx
katspaugh: The settings path is the same – `/settigns`. Could be a new path tho, e.g. `/preferences`.

usame-algan: Ah makes sense then if its the same page.

francovenica:
Looks good. I would not noticed the change in the title of the page if I hadn't checked that last commits, good early catch.

I've checked that the settings opens fine from the welcome page or the settings page (the 2 places where you can see the footer)
Tested that the cookies policies still work fine in this new page. I like that the cookie banner that opens for first users closes automatically if you accept selection/accept all in this new preferences/settings page
The settings work fine (prefix, dark mode, data import/export, RPC change)

PR #2004, opened 2023-05-19
fix: Do not rerender approval editor in confirmation modal

### LINKED ISSUE #2003: ApprovalEditor rerenders too often in queue
## Bug description
The ApprovalEditor re-renders on every poll. 

We pass` { ...decodedData, to: txDetails?.txData?.to.value }` as prop to the ApprovalEditor in the queue. 
The value needs to be memoized to avoid unnecessary rerenders.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Queue a transaction containing approval(s)
 2. Wait ~10-15s
 3. Witness the approval editor rerendering

## Expected result
In the queue the tx cannot be modified so the ApprovalEditor should only render once.

## Obtained result
It rerenders on every poll

schmanu:
## What it solves
Avoids rerenders of approval editor, when the queue gets polled

Resolves #2003 

## How this PR fixes it
Memoize prop for approval editor

## How to test it
1.    Queue a transaction containing approval(s)
2.    Wait ~10-15s
3.    Witness the approval editor rerendering

## Checklist

### DIFF
--- a/src/components/tx/DecodedTx/index.tsx
+++ b/src/components/tx/DecodedTx/index.tsx
@@ -1,4 +1,4 @@
-import type { SyntheticEvent, ReactElement } from 'react'
+import { type SyntheticEvent, type ReactElement, useMemo } from 'react'
 import { Accordion, AccordionDetails, AccordionSummary, Box, Skeleton, Typography } from '@mui/material'
 import { OperationType, type SafeTransaction } from '@safe-global/safe-core-sdk-types'
 import {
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
👍

DiogoSoaress (approved):

francovenica:
Made a comment in the original issue that i wasn't able to reproduce the issue but I wasn't looking in the right place. I found it now, so I was able to reproduce it.

LGTM
I'll post how it looks before and I can confirm that now that flicker doesn't happen anymore with the fix
appEditFlikr

PR #2005, opened 2023-05-21
Changed a selector in balances.cy.js to fix the block "pagination should work"

bashiulAlam:
## What it solves
Test failure in smoke-apps

## How this PR fixes it
Changed one selector

---


cypress/e2e/smoke/balances.cy.js
```diff
@@ -204,7 +204,7 @@ describe('Assets > Coins', () => {
       cy.contains('button', 'Accept selection').click()
 
       // Find button with the text OK (terms banner), and if it exists, click it
-      cy.get('button').contains('Ok').click({ force: true })
+      cy.get('.styles_wrapper__8YyOG button').click({ force: true })
```
katspaugh, line 207: That’s a generated class, it will change on each build.

PR #2032, opened 2023-05-24
Feat: blink title if walletconnect needs attention

katspaugh:
Screenshot 2023-05-24 at 10 17 25

If a SafeAppReviewModal or a MsgModal appear in a hidden tab, the title will start blinking so that the user knows they need to switch to that tab and confirm the tx/message signature.

## How to test

* Open https://blink--walletweb.review-wallet-web.5afe.dev/
* Connect your wallet and open a Safe account
* Go to the WalletConnect safe app
* Open Snapshot.org in another tab
* Connect to Snapshot via the WalletConnect safe app
* Edit your Snapshot profile name
* Observe that the Safe tab starts blinking

<video src="https://github.com/safe-global/safe-wallet-web/assets/381895/94d599bd-c795-4b16-b450-5647c24896e9" />


DiogoSoaress (commented):
Code looks good to me and the implementation works smoothly. 👍 
I would however not skip a review from @johannesmoormann or @liliiaorlenko as I am unsure  the red exclamation mark is the way to flag some required action in the hidden tab.

DiogoSoaress (approved):
Looks good to me!
One suggestion: I'd rather document the hook (with the `/**` syntax) than leaving the same comment in three different files

PR #2035, opened 2023-05-24
test: e2e Import Export data

### LINKED ISSUE #2008: Add an e2e test for data import and added safes
* Go to the Welcome page
* Import a pre-exported file with a few added safes on different chains and an address book
* Click on one of the added safes in the safe list
* Check that the name and address match
* Go to the address book and check that it's populated

francovenica:
## What it solves

Resolves #2008

## How this PR fixes it
Test Import export of Data files in web browser
Checks imported data
Access to a safe
Checks the address book, settings, pinned apps and Data for export

## Checklist

### DIFF
--- a/cypress/e2e/smoke/import_export_data.cy.js
+++ b/cypress/e2e/smoke/import_export_data.cy.js
@@ -1,0 +1,71 @@
+import 'cypress-file-upload'
+const path = require('path')
+import { format } from 'date-fns'
+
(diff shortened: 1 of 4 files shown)


francovenica:
Note: in one file that is not a test we changed a text from "Safes" to "Safe" because that is the correct term for that section.

katspaugh (approved):
Nice one, and thanks for the text fix!

PR #2036, opened 2023-05-24
Modals 2

Seroxdesign:
Made modals for load and create safe

TODO:

add buttons to toggle those somewhere in chat


katspaugh:
Bruh, it's already the third time. Please pay attention.

PR #2038, opened 2023-05-25
chore: fix super-linear runtime due to backtracking

suraneti:
## What it solves

Regex used here, which is vulnerable to super-linear runtime due to backtracking

Resolves #

Update regex

## How this PR fixes it

Resolved potential backtracking and denial of service (DoS) vulnerabilities

## How to test it

Run unit-test

## Checklist

### DIFF
--- a/src/utils/addresses.ts
+++ b/src/utils/addresses.ts
@@ -54,7 +54,7 @@ export const formatPrefixedAddress = (address: string, prefix?: string): string 
 }
 
 export const cleanInputValue = (value: string): string => {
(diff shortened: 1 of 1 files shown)


katspaugh (changes requested):
This breaks the regex though, doesn't it? It needs to capture that substring anywhere in the input string, not just at the beginning.
Also I don't think there's a backtracking problem as the `0x[a-f0-9]{40}` acts as a lookahead and limits the `+` operator.

suraneti:
@katspaugh My bad, just test it again. It works properly.

PR #2039, opened 2023-05-25
refactor: improve type safety of errors

iamacook:
## What it solves

Removes needs to cast `Error`s

## How this PR fixes it

Caught errors are typed as `unknown` because they can have any value. As such, we were blindly casting them to `Error`.

A new `asError` function is now used to coerce the `unknown` errors into an `Error` instance accordingly.

## How to test it

Observe all unit tests passing.

## Checklist


katspaugh (changes requested):
Although neat, I'm apprehensive of adding an import every time I need to handle an error just to cast it. We already import `logError` in most of those cases, so maybe we could modify `logError` so that it accepts the entire error object (whatever that might be) and uses `asError` internally.

iamacook:
> We already import `logError` in most of those cases, so maybe we could modify `logError` so that it accepts the entire error object (whatever that might be) and uses `asError` internally.

I've refactored this accordingly, leaving a few as is when the error is needed beyond `logError`.

src/components/tx/modals/BatchExecuteModal/ReviewBatchExecute.tsx
```diff
@@ -92,10 +93,12 @@ const ReviewBatchExecute = ({ data, onSubmit }: { data: BatchExecuteData; onSubm
 
     try {
       await (willRelay ? onRelay() : onExecute())
-    } catch (err) {
-      logError(Errors._804, (err as Error).message)
+    } catch (_err) {
+      const err = asError(_err)
+
+      logError(Errors._804, err.message)
```
katspaugh, line 99: Should it be `logError(Errors._804, err)`?

katspaugh (commented):
There's a few places where you're still passing a message to logError instead of the thrown object.

katspaugh (approved):
Thanks! 👍

iamacook:
Closing in favour of https://github.com/safe-global/safe-wallet-web/pull/2127.

Issue #2021, opened 2023-05-22
[TX flow] layout


johannesmoormann:
## Requirements

- Implement Stepper
- Open full screen modal or display inside a drawer (Safe apps)
- Implement main area and sidebar area
- New blocks are added to the bottom of the main area when progressing through the stepper
- Mobile responsiveness

PR #2040, opened 2023-05-25
feat: Add StepperFactory and new tx layout

usame-algan:
## What it solves

Part of #2021 

## How this PR fixes it

- Adds new global context `ModalProvider` for orchestrating modals in the app
- Adds new folder structure `TxFlow`
- Adds `NewTxMenu`, `ReplaceTxMenu`, `RejectTx`, `TokenTransfer` to the provider
- Adds `TxLayout` component
- Adds `StepperFactory` and uses it for `TokenTransferFlow`

## ToDos

- [x] Close Modal when navigating in the sidebar
- [ ] Resolve TODOs

## How to test it

1. Open the app
2. Press on `New Transaction` in the sidebar
3. Observe the `NewTxMenu` opening as a fullscreen modal
4. Press on `Send Tokens`
5. Observe the new `TokenTransferFlow` opening as a fullscreen modal
6. Navigate to `/queue`
7. Replace an existing transaction
8. Observe the `ReplaceTxMenu` opening as a fullscreen modal
9. Press on either `Send Tokens` or `Reject Transaction`
10. Observe `TokenTransferFlow` or `RejectTx` opening as a fullscreen modal

## Screenshots
Screenshot 2023-05-26 at 11 51 16
Screenshot 2023-05-26 at 11 51 24
Screenshot 2023-05-26 at 11 51 35
Screenshot 2023-05-26 at 11 51 46

## Checklist


src/components/TxFlow/ModalProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+  type SetStateAction,
+  useState,
+  type ComponentProps,
+  useEffect,
+} from 'react'
+import ModalDialog from '@/components/common/ModalDialog'
+import TokenTransferFlow from '@/components/TxFlow/TokenTransfer/TokenTransferFlow'
+import RejectTx from '@/components/TxFlow/RejectTx'
+import NewTxMenu from '@/components/TxFlow/NewTx'
+import ReplaceTxMenu from '@/components/TxFlow/ReplaceTx'
```
katspaugh, line 15: This part can be extracted into another module. Which could be imported and inserted here.

src/components/TxFlow/ModalProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+  ReplaceTx = 'replaceTx',
+  NewTx = 'newTx',
+}
+
+const ModalTypes = {
+  [ModalType.SendTokens]: TokenTransferFlow,
+  [ModalType.RejectTx]: RejectTx,
+  [ModalType.ReplaceTx]: ReplaceTxMenu,
+  [ModalType.NewTx]: NewTxMenu,
+}
```
katspaugh, line 30: I've just realized it's probably easier to just let modal-opening buttons pass `TokenTransferFlow` directly instead of a string. That would make the dependency explicit.

src/components/TxFlow/TokenTransfer/TokenTransferFlow.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { ProgressBar } from '@/components/common/ProgressBar'
+
+const initialData = {
+  recipient: '',
+  tokenAddress: '',
+  amount: '',
+  type: SendTxType.multiSig,
+}
+
+const TokenTransferFlow = ({ txNonce }: { txNonce?: number }) => {
```
katspaugh, line 15: The naming is a bit inconsistent, this is the only transaction type named with "Flow" at the end.

usame-algan: We were also thinking of naming `RejectTx` to `RejectTxFlow` but it only has one step and we didn't use the Stepper there so we left it as `RejectTx`.

src/components/TxFlow/ModalProvider.tsx
DiogoSoaress: The implementation we opted for aimed to narrow the possible components we want to show in the modal. It comes at the cost of maintaining the map but setting a component in the dispatcher would make harder to reinforce the types.
We are also not sold on storing the component as state.

katspaugh: You can still limit the type of components setVisibleModal accepts.

katspaugh (approved):

PR #2041, opened 2023-05-25
Fix: terms link + rm terms banner

### LINKED ISSUE #1936: Remove the terms banner
We've updated our terms of service on March 7 2023, and added a banner for existing users to accept the new terms. See #1731.

We should remove this banner after the 7th of June, as per @johannesmoormann's suggestion:

> we accept that we do this on a device basis for e.g. 3 months. after that we deactivate the message.

### LINKED ISSUE #1991: Update Terms link on the safe-apps Disclaimer
## Bug description
Update Terms link on the safe-apps Disclaimer 
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: any

## Steps to reproduce
 1. Go to the safe-apps
 2. click open app
 3. click Terms link on the Desclaimer
 Current result: https://safe.global/terms is opened

## Expected result: the link to the https://app.safe.global/terms  should be used

image

katspaugh:
## What it solves

Resolves #1936.
Resolves #1991.

## How this PR fixes it

Replaces the terms link with a local route and removes the Terms Banner.


katspaugh:
I'll merge this directly, as it's a very straight-forward change. I've tested that the new Terms link opens the right page.

PR #2047, opened 2023-05-26
Refactor: move wallet utils to utils

katspaugh:
The `wallets.ts` file contains heavy wallet module imports which should be done dynamically.
Onboard.ts already imports them dynamically, but there were a bunch of places where utils were imported from that module unnecessarily.


usame-algan (approved):
Makes sense to extract consts and utils from that file but I am not sure I understand how it helps with dynamic imports.

usame-algan (approved):
Makes sense to extract consts and utils from that file but I am not sure I understand how it helps with dynamic imports.

katspaugh:
@usame-algan the wallets.ts module was imported statically in many places, now it's only imported dynamically in useOnboard.

PR #2050, opened 2023-05-26
Refactor: reactive recommended nonce

### LINKED ISSUE #2034: "Proposing" transactions one after another can result in the same nonce
## Problem
If you quickly create two back-to-back transactions, the recommended nonce for them (from the backend) may end up the same resulting in the second tx cancelling the first one.

## Example
An example of how to reproduce: https://www.loom.com/share/89a490a44d454d9bba67c15d9e602ff1

What happens under the hood:

- User makes a tx on CoW Swap
- The WC app picks it up via the RPC call
- WC safe app calls the wallet app and tells it to popup a modal
- User confirms the modal and signs the tx
- Tx is added to the queue (aka "proposed")
- RPC request returns a response to the CoW app

Although adding to the queue should increment the "recommended" nonce on the backend, it doesn't. Apparently, there's an internal lag/asynchronous state involved.

## Solution
One potential solution would be to wait/poll until the recommended nonce increments after proposing a tx and make any subsequent txs wait for it.

katspaugh:
## What it solves

Resolves #2034

## What this PR does

The "recommended" nonce was previously fetched and set only once, when generating a SafeTransaction.

If the queue updates (and thus the recommended nonce) while the user is still looking at the Sign/Execute modal, neither the Advanced Params form, not the SafeTransaction would pick up an updated recommended nonce.

With this PR, the recommended nonce is fetched every time the queue is updated.

It still doesn't guarantee that the user doesn't submit the modal with an outdated nonce, but it makes it much more robust.

### DIFF
--- a/src/components/tx/SignOrExecuteForm/hooks.ts
+++ b/src/components/tx/SignOrExecuteForm/hooks.ts
@@ -15,7 +15,8 @@ import { useHasPendingTxs } from '@/hooks/usePendingTxs'
 import { sameString } from '@safe-global/safe-core-sdk/dist/src/utils'
 import type { ConnectedWallet } from '@/services/onboard'
 import type { OnboardAPI } from '@web3-onboard/core'
-import { hasEnoughSignatures } from '@/utils/transactions'
+import { getRecommendedTxParams } from '@/services/tx/tx-sender/recommendedNonce'
+import useAsync from '@/hooks/useAsync'
 
 type TxActions = {
   signTx: (safeTx?: SafeTransaction, txId?: string, origin?: string) => Promise<string>
@@ -97,7 +98,7 @@ export const useTxActions = (): TxActions => {
       assertOnboard(onboard)
 
       // Relayed transactions must be fully signed, so request a final signature if needed
-      if (isRelayed && !hasEnoughSignatures(safeTx, safe)) {
+      if (isRelayed && safeTx.signatures.size < safe.threshold) {
         safeTx = await signRelayedTx(safeTx)
         txId = await proposeTx(wallet.address, safeTx, txId, origin)
       }
@@ -137,4 +138,37 @@ export const useIsExecutionLoop = (): boolean => {
   const wallet = useWallet()
   const { safeAddress } = useSafeInfo()
   return wallet ? sameString(wallet.address, safeAddress) : false
+}
+
+export const useRecommendedNonce = (
+  safeTx?: SafeTransaction,
+):
+  | {
+      nonce: number
+      safeTxGas: number
+    }
+  | undefined => {
+  const { safe } = useSafeInfo()
+  const safeTxData = safeTx?.data
+
+  // Memoize only the necessary params, so that it doesn't change every time safeTx is changed
+  const safeTxParams = useMemo(
+    () =>
+      safeTxData?.to
+        ? {
+            to: safeTxData.to,
+            value: safeTxData.value,
+            data: safeTxData.data,
+            operation: safeTxData.operation,
+          }
+        : undefined,
+    [safeTxData?.to, safeTxData?.value, safeTxData?.data, safeTxData?.operation],
+  )
+
+  const [recommendedParams] = useAsync(() => {
+    if (!safeTxParams) return
+    return getRecommendedTxParams(safeTxParams)
+  }, [safeTxParams, safe?.txQueuedTag])
+
+  return recommendedParams
 }
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -76,6 +76,9 @@ const SignOrExecuteForm = ({
   const isExecutionLoop = useIsExecutionLoop()
   const canExecute = isCorrectNonce && (isExecutable || isNewExecutableTx)
 
+  // Nonce cannot be edited if the tx is already proposed, or signed, or it's a rejection
+  const nonceReadonly = !isCreation || !!tx?.signatures.size || isRejection
+
   // If checkbox is checked and the transaction is executable, execute it, otherwise sign it
   const willExecute = (onlyExecute || shouldExecute) && canExecute
 
@@ -99,7 +102,7 @@ const SignOrExecuteForm = ({
   const [advancedParams, setAdvancedParams] = useAdvancedParams(
     gasLimit,
     // Initial nonce or a recommended one
-    (safeTx && safeTx.data.nonce !== -1) || !recommendedParams ? safeTx?.data.nonce : recommendedParams?.nonce,
+    nonceReadonly ? safeTx?.data.nonce : recommendedParams?.nonce,
     // Initial safeTxGas or a recommended one
     safeTx?.data.safeTxGas ?? recommendedParams?.safeTxGas,
   )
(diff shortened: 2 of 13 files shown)


src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -138,3 +140,29 @@ export const useIsExecutionLoop = (): boolean => {
...
+      nonce: number
+      safeTxGas: number
+    }
+  | undefined => {
+  const { safe } = useSafeInfo()
+
+  const [data] = useAsync(() => {
+    if (!safeTx) return
+    return getRecommendedTxParams(safeTx)
+  }, [safeTx, safe?.txQueuedTag])
```
schmanu, line 158: Should we also react to the `txHistoryTag`?

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -138,3 +140,29 @@ export const useIsExecutionLoop = (): boolean => {
...
+
+  return useMemo(() => {
+    return data
+      ? {
+          nonce: initialNonce !== undefined && initialNonce !== -1 ? initialNonce : data.nonce,
```
schmanu, line 163: Nit: For readability I would let `useRecommendedNonce` always return the recommended one and react to an already initially set nonce in a separate step, which overrides the txs nonce if no `initialNonce` was set.

src/components/tx/SignOrExecuteForm/hooks.ts
katspaugh: No, the nonce is incremented only when a tx is added to the queue.

src/components/tx/SignOrExecuteForm/hooks.ts
katspaugh: I initially had it like this, but this introduced the same not-so-nice condition in SignOrExecute...

src/hooks/useWalletCanRelay.ts
```diff
@@ -3,12 +3,12 @@ import useSafeInfo from '@/hooks/useSafeInfo'
 import useWallet from '@/hooks/wallets/useWallet'
 import { isSmartContractWallet } from '@/hooks/wallets/wallets'
 import { Errors, logError } from '@/services/exceptions'
-import { hasEnoughSignatures } from '@/utils/transactions'
 import { type SafeTransaction } from '@safe-global/safe-core-sdk-types'
 
 const useWalletCanRelay = (tx: SafeTransaction | undefined) => {
   const { safe } = useSafeInfo()
   const wallet = useWallet()
+  const hasEnoughSignatures = tx && tx.signatures.size >= safe.threshold
```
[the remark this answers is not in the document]
katspaugh: Tx changes whenever the nonce changes, so that would be bad. That's the whole point why I moved it out here and in gas estimation.

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -138,3 +139,36 @@ export const useIsExecutionLoop = (): boolean => {
...
+      safeTxGas: number
+    }
+  | undefined => {
+  const { safe } = useSafeInfo()
+  const safeTxData = safeTx?.data
+
+  // Memoize only the necessary params, so that it doesn't change every time safeTx is changed
+  const safeTxParams = useMemo(
+    () =>
+      safeTxData?.to
```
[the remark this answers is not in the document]
katspaugh: To no depend on the entire `safeTxData`

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -138,3 +139,36 @@ export const useIsExecutionLoop = (): boolean => {
...
+            data: safeTxData.data,
+            operation: safeTxData.operation,
+          }
+        : undefined,
+    [safeTxData?.to, safeTxData?.value, safeTxData?.data, safeTxData?.operation],
+  )
+
+  const [recommendedParams] = useAsync(() => {
+    if (!safeTxParams) return
+    return getRecommendedTxParams(safeTxParams)
```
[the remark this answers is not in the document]
katspaugh: Have you read the comment for `safeTxParams`? 

```
  // Memoize only the necessary params, so that it doesn't change every time safeTx is changed
```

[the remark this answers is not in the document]
usame-algan: Ah, that makes sense, thanks!

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -138,3 +139,36 @@ export const useIsExecutionLoop = (): boolean => {
...
+            to: safeTxData.to,
+            value: safeTxData.value,
+            data: safeTxData.data,
+            operation: safeTxData.operation,
+          }
+        : undefined,
+    [safeTxData?.to, safeTxData?.value, safeTxData?.data, safeTxData?.operation],
+  )
+
+  const [recommendedParams] = useAsync(() => {
```
usame-algan, line 168: I would suggest also exporting the loading flag here since there is a small window when creating transactions now where it will show the execute option in any case until the nonce is estimated.

src/hooks/useWalletCanRelay.ts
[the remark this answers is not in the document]
katspaugh: No, it will update only if the number of signatures change.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -36,218 +27,85 @@ type SignOrExecuteProps = {
...
-    (willExecute && isExecutionLoop)
-
-  const error = props.error || (willExecute ? gasLimitError || executionValidationError : undefined)
+    createTx({ ...tx.data, safeTxGas: advancedParams.safeTxGas }, advancedParams.nonce)
+      .then(setTx)
+      .catch((err) => logError(Errors._103, (err as Error).message))
+  }, [nonceReadonly, tx?.data, advancedParams.nonce, advancedParams.safeTxGas])
 
   return (
-    <form onSubmit={handleSubmit}>
```
usame-algan, line 165: By removing the `form` element here we lost some padding. We could add a div instead or adjust the css for DialogContent

src/services/tx/tx-sender/create.ts
```diff
@@ -1,121 +1,56 @@
...
-
-  // If the nonce is not provided, we get the recommended one
-  if (nonce === undefined) {
-    const recParams = (await getRecommendedTxParams(txParams)) || {}
-    txParams = { ...txParams, ...recParams }
-  } else {
-    // Otherwise, we use the provided one
-    txParams = { ...txParams, nonce }
-  }
-
```
usame-algan, line 71: I just realized that we don't estimate the nonce anymore for the initial tx that is passed to `SignOrExecuteForm`. Thats probably the reason why the execute checkbox is showing initially. I suggest we restore this block.

usame-algan: It also applies to the other `create` calls. We might need the `withRecommendedNonce` helper after all.

src/components/tx/SignOrExecuteForm/index.tsx
katspaugh: Why does it matter? There'll be a new design for all of these things.

usame-algan (approved):

PR #2072, opened 2023-06-02
Fix: pass nonce correctly

katspaugh:
The nonce wasn't passed to the context in the right place.


usame-algan (approved):
Nice catch! Noticed earlier too that rejection didn't work anymore but now it uses the correct nonce

PR #2075, opened 2023-06-07
feat: Remember the state of Execute transaction checkbox in settings

plusminushalf:
## What it solves

This pull request addresses the need to save the last checkbox state of the "Execute Transaction" checkbox in the settings. With this enhancement, the system will remember the user's preference regarding immediate transaction execution. If a user chooses not to execute the transaction immediately, this setting will be preserved, eliminating the need to uncheck the box repeatedly. However, users can still recheck the box, and the checked state will also be remembered.

## How this PR fixes it

This PR adds a `transactionExecution` flag in the settings, enabling users to enable or disable transaction execution based on their preferences easily.

## How to test it

Steps to test:

1. Create a new Transaction
2. Uncheck the `Execute transaction` checkbox (Assuming the transaction is ready for execution)
3. Refresh the page and create another transaction, the last state of the `Execute transaction` checkbox should be remembered.

## Screenshots

NA

## Checklist

### DIFF
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -22,6 +22,8 @@ import { useRelaysBySafe } from '@/hooks/useRemainingRelays'
 import useWalletCanRelay from '@/hooks/useWalletCanRelay'
 import { ExecutionMethod, ExecutionMethodSelector } from '../ExecutionMethodSelector'
 import { hasRemainingRelays } from '@/utils/relaying'
+import { useAppDispatch, useAppSelector } from '@/store'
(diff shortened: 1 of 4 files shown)


src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -149,6 +154,11 @@ const SignOrExecuteForm = ({
     setAdvancedParams(data)
   }
 
+  const hangleExecuteCheckboxChange = (checked: boolean) => {
```
schmanu, line 157: There is a small typo here

```suggestion
  const handleExecuteCheckboxChange = (checked: boolean) => {
```

PR #2081, opened 2023-06-12
Safe Apps tx flow

katspaugh:
The Safe Apps tx flow is a bit complicated because it needs to call a function when the tx modal is closed.
I've added another state var to the TxModalContext for this.

This PR covers only the tx modal, still need to port the two MsgModals.


src/components/safe-apps/SafeAppsTxModal/ReviewSafeAppsTx.tsx
```diff
@@ -30,41 +30,43 @@ const ReviewSafeAppsTx = ({
...
   }
 
   const origin = useMemo(() => getTxOrigin(app), [app])
 
-  // TODO: Need to use the SafeTxProvider here
   return (
-    <SignOrExecuteForm onSubmit={handleSubmit} origin={origin}>
-      <>
+    <DialogContent>
+      <SignOrExecuteForm onSubmit={handleSubmit} origin={origin}>
```
usame-algan, line 69: I think this component can be self-closing now.

src/components/tx-flow/index.tsx
```diff
@@ -6,34 +6,57 @@ import {
...
+    return () => router.events.off('routeChangeComplete', handleModalClose)
+  }, [router, handleModalClose])
+
+  // On close callback
+  useEffect(() => {
+    if (!txFlow) {
+      setOnClose((prevOnClose) => {
+        prevOnClose?.()
+        return noop
+      })
```
usame-algan, line 49: I don't understand why we can't simply call `onClose()` here.

usame-algan (approved):
I've noticed 2 issues:

1. The modal doesn't close when executing a safe apps tx
2. For Safe Apps transactions we don't want to show them inside the modal but a drawer that pops up from the bottom of the screen. Should we implement this in a separate PR?

src/components/tx-flow/index.tsx
katspaugh: It’s just to avoid having it as a dependency

katspaugh:
@usame-algan thanks for the review. I've removed `onClose`/`setOnClose` for now, as it was violating React state rules.
We need to figure out how to handle an early modal close.

> The modal doesn't close when executing a safe apps tx

And this as well. I don't see a good way to do this yet, but we'll figure it out.

PR #2083, opened 2023-06-12
refactor: save entire data structure on submission

iamacook:
## What it solves

This updates the `nextStep` function to require the entire data structure to be set on submission.

## Other small changes

- Only renders current step
- Adds back button (yet to be styled)
- Closes modal when signing

## How to test it

Open the transaction creation flow and click next, observing the data present in the second step. Clicking back should also retain the values on the first step.

## Checklist


katspaugh (approved):
👍

PR #2084, opened 2023-06-13
Refactor: single type for tx flows

katspaugh:
## What it solves

Reuse the same type for both steps in TokenTransfer and in NftTransfer.


src/components/tx-flow/flows/TokenTransfer/index.tsx
```diff
@@ -35,23 +33,21 @@ const defaultData: TokenTransferParams = {
...
-const TokenTransferFlow = ({ txNonce, ...params }: TokenTransferFlowProps) => {
-  const initialData = useMemo(() => merge({}, defaultData, params), [params])
-
-  const { data, step, nextStep, prevStep } = useTxStepper<[TokenTransferParams, TokenTransferParams | undefined]>([
-    initialData,
-    undefined,
-  ])
+const TokenTransferFlow = (props: TokenTransferFlowProps) => {
+  const { data, step, nextStep, prevStep } = useTxStepper<TokenTransferParams>({
+    ...props,
```
usame-algan, line 38: Would this also push the `txNonce` to the stepper as initial data?

src/components/tx-flow/flows/NftTransfer/index.tsx
```diff
@@ -4,31 +4,32 @@ import useTxStepper from '../../useTxStepper'
...
-    undefined,
-  ])
+  const { data, step, nextStep, prevStep } = useTxStepper<NftTransferParams>({
+    ...defaultParams,
+    ...params,
+  })
 
   const steps = [
-    <SendNftBatch key={0} params={data[0]} onSubmit={(formData) => nextStep([formData, formData])} />,
+    <SendNftBatch key={0} params={data} onSubmit={(formData) => nextStep({ ...data, ...formData })} />,
```
iamacook, line 28: If we no longer want to separate the data of each step, I would suggest we instead merge inside the hook.

I'd be cautious of this approach though as the types are less strict per step and we run the risk of overwriting same-named fields.

iamacook: A separate improvement would be passing the type as a generic to the component and then using that inside, e.g.

```ts
const SendNftBatch = <T extends Object>() => {
  const formMethods = useForm<T>({ ... })

  return (
    ...
  )
}

<SendNftBatch<NftTransferParams> {...props} />
```

src/components/tx-flow/flows/NftTransfer/SendNftBatch.tsx
```diff
@@ -51,11 +51,11 @@ const NftItem = ({ image, name, description }: { image: string; name: string; de
 
 const SendNftBatch = ({ params, onSubmit }: SendNftBatchProps) => {
   const addressBook = useAddressBook()
-  const { tokens = [] } = params
+  const { tokens } = params
 
   const formMethods = useForm<FormData>({
```
iamacook, line 56: We could select the fields from the type here.

src/components/tx-flow/flows/NftTransfer/index.tsx
katspaugh: The stepper hook should just manage steps and hold whatever data is passed from outside. It shouldn't have heuristics about how the data flows between steps.

src/components/tx-flow/flows/NftTransfer/SendNftBatch.tsx
iamacook: Instead of creating the new type, derive it from the "main" one:

```ts
type FormData = Pick<NftTransferParams, 'recipient'>
```

PR #2067, opened 2023-06-02
[Epic] Tx flow redesign

katspaugh:
Resolves #1828.

Designs: [on Figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?type=design&node-id=6093-176670&t=0M7eGvpbDvNs6LbA-0)

See [related issues](https://github.com/safe-global/safe-wallet-web/issues?q=is%3Aissue+is%3Aopen+%22%5Btx+flow%5D%22)

Screenshot 2023-06-02 at 08 27 24

PR #2085, opened 2023-06-13
feat: Change threshold flow

usame-algan:
## What it solves

Part of #2067 

## How this PR fixes it

- Moves the Change Threshold to the new tx flow
- Fixes a bug regarding gasLimit estimation in the SignOrExecuteForm

## How to test it

1. Open the Settings page on a 2/n safe
2. Change the threshold
3. Observe the new tx flow

## Screenshots
Screenshot 2023-06-13 at 11 21 17

## Checklist


src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -34,7 +34,7 @@ const SignOrExecuteForm = (props: SignOrExecuteProps): ReactElement => {
   const willExecute = (props.onlyExecute || shouldExecute) && canExecute
 
   // Estimate gas limit
-  const { gasLimit, gasLimitError } = useGasLimit(safeTx)
+  const { gasLimit, gasLimitError } = useGasLimit(willExecute ? safeTx : undefined)
```
katspaugh, line 37: Why this change? Gas limit is needed for simulation. If it's only estimated for executions, it should be moved to the ExecuteForm.

katspaugh (approved):

src/components/tx/SignOrExecuteForm/index.tsx
usame-algan: This was a regression when we refactored this component. Moving it into the `ExecuteForm` would make the Simulation always fetch the latest block gas limit which is less accurate so I would suggest keeping it here for now.

PR #2087, opened 2023-06-13
feat: replace owner flow

iamacook:
## What it solves

Migrating the replace owner flow to the new transaction flow framework.

## How this PR fixes it

Each step of the replace owner flow has been ported, as well as removing the old one.

## How to test it

Replace an owner via the settings and observe each step working correctly as well as a successful transaction.

## Screenshots

replace owner flow

## Checklist


src/components/tx-flow/flows/AddOwner/SetThreshold.tsx
usame-algan, line 1: This component overlaps quite a bit with the same one from `RemoveOwner`. Do you think its possible to combine them by e.g. extracting `newNumberOfOwners`?

usame-algan: After looking through it together we decided to tackle this when implementing the new redesign

usame-algan (approved):

src/components/tx-flow/flows/AddOwner/SetThreshold.tsx
iamacook: For future reference: we discussed this and will keep improving this in mind when implementing the new designs.

PR #2093, opened 2023-06-14
feat: new spending limit flow

iamacook:
## What it solves

Migrating the new/update spending limit flow  to the new transaction flow framework.

## How this PR fixes it

The new/update spending limit flow has been ported, as well as removing the old one.

## How to test it

Create a new/update a spending limit via the settings and observe a successful transaction.

## Screenshots

new spending limit flow

## Checklist


src/components/tx-flow/flows/UpdateSafe/UpdateSafeReview.tsx
```diff
@@ -26,29 +26,29 @@ export const UpdateSafeReview = () => {
...
+        confirm the update in case more than one confirmation is required for this Safe Account.
+      </Typography>
+
+      <Typography mb={2}>
+        <b>Warning:</b> this upgrade will invalidate all unexecuted transactions. This means you will be unable to
+        access or execute them after the upgrade. Please make sure to execute any remaining transactions before
+        upgrading.
+      </Typography>
+
+      <SignOrExecuteForm onSubmit={() => null} />
```
iamacook, line 51: I realised this should not wrap the above text so included it here.

src/components/tx-flow/flows/NewSpendingLimit/CreateSpendingLimit.tsx
```diff
@@ -46,15 +36,22 @@ export const _validateSpendingLimit = (val: string, decimals?: number) => {
...
   const [showResetTime, setShowResetTime] = useState<boolean>(false)
   const { balances } = useVisibleBalances()
 
   const resetTimeOptions = useMemo(() => getResetTimeOptions(chainId), [chainId])
+  const defaultResetTime = resetTimeOptions[0].value
 
-  const formMethods = useForm<NewSpendingLimitData>({
-    defaultValues: { ...data, resetTime: '0' },
+  const formMethods = useForm<NewSpendingLimitFlowProps>({
+    defaultValues: params,
```
usame-algan, line 54: Somehow when going back to this step the tokenAddress resets to undefined and I am getting warnings in the console

src/components/tx-flow/flows/NewSpendingLimit/CreateSpendingLimit.tsx
```diff
@@ -46,15 +36,22 @@ export const _validateSpendingLimit = (val: string, decimals?: number) => {
...
+}: {
+  params: NewSpendingLimitFlowProps
+  onSubmit: (data: NewSpendingLimitFlowProps) => void
+}) => {
   const chainId = useChainId()
   const [showResetTime, setShowResetTime] = useState<boolean>(false)
   const { balances } = useVisibleBalances()
 
   const resetTimeOptions = useMemo(() => getResetTimeOptions(chainId), [chainId])
+  const defaultResetTime = resetTimeOptions[0].value
```
usame-algan, line 51: An idea for when we move on to refactoring this component. If the user selects a reset time and goes back to this step it should be active and show the previously selected option instead of being closed and resetting to the default value.

src/components/tx-flow/flows/NewSpendingLimit/CreateSpendingLimit.tsx
iamacook: Seems like there's a bug with the `register` function. I've fixed it and added a comment to look into this further when we update react-hook-form.

src/components/tx-flow/flows/NewSpendingLimit/CreateSpendingLimit.tsx
iamacook: Good point. I've adjusted this to preserve the value.

PR #2094, opened 2023-06-14
Onclose callback

katspaugh:
If the user closes the tx modal by clicking on the X button, or navigating to another route, the current Safe Apps transaction should be cancelled.

### DIFF
--- a/src/components/common/NewModalDialog/index.tsx
+++ b/src/components/common/NewModalDialog/index.tsx
@@ -1,11 +1,10 @@
-import { type ReactElement, type ReactNode, useContext } from 'react'
+import { type ReactElement, type ReactNode } from 'react'
 import { IconButton, type ModalProps } from '@mui/material'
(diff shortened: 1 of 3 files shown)


src/components/common/NewModalDialog/index.tsx
```diff
@@ -85,7 +84,7 @@ const NewModalDialog = ({
         <span style={{ flex: 1 }} />
         <IconButton
           aria-label="close"
-          onClick={() => setTxFlow(undefined)}
+          onClick={(e) => onClose?.(e, 'backdropClick')}
```
usame-algan, line 87: Do we have to call `onClose` with e and backdropClick here even if its not being used?

usame-algan (approved):

src/components/common/NewModalDialog/index.tsx
katspaugh: Yeah, that's the signature of that prop inherited from the MUI Dialog.

PR #2095, opened 2023-06-14
feat: add WalletConnect v2 wallet module

### LINKED ISSUE #2017: Migrate to WalletConnect v2
## What is the feature about
On the 28th of June WalletConnect v1 will be shutdown.
We need to migrate `web3-onboard` to use WalletConnect v2.

## The list of requirements
Follow the [migration guide](https://docs.walletconnect.com/2.0/advanced/migration-from-v1.x/dapps#web3-onboard)

iamacook:
## What it solves

Partially resolves #2017

## How this PR fixes it

A second (v2) WalletConnect module has been added to onboard. It can be toggled via the config service with wallet name `walletConnect_v2` and has been enabled on all non-prod. chains.

Note: onboard dedupes wallet modules and the label of the old module has therefore been modified from "WalletConnect" to "WalletConnect v1".

## How to test it

Connect a wallet and observe the "WalletConnect v1" and "WalletConnect" modules are present in the modal.

## Screenshots

image

## Checklist


src/hooks/wallets/wallets.ts
```diff
@@ -14,10 +14,33 @@ import pairingModule from '@/services/pairing/module'
 import e2eWalletModule from '@/tests/e2e-wallet'
 import { CGW_NAMES, WALLET_KEYS } from './consts'
 
+// We need to modify the module name as onboard dedupes modules with the same label and the WC v1 and v2 modules have the same
+// @see https://github.com/blocknative/web3-onboard/blob/d399e0b76daf7b363d6a74b100b2c96ccb14536c/packages/core/src/store/actions.ts#L419
+const walletConnectV2 = (): WalletInit => {
+  return (helpers) => {
+    const MODULE_LABEL = 'WalletConnect v2'
```
katspaugh, line 21: The plan was to disable v1 via the config-service, at which point this module should be called just `WalletConnect`. Can you use the name from the config-service instead of a hardcoded one?

iamacook: As discussed, due to the current limitations of wallet naming on the config service we will keep this as "WalletConnect" but alter the v1 module to be "WalletConnect v1" instead:

image

We can/should cleanup the config service naming convention at some stage, for which we already have [an issue](https://github.com/safe-global/safe-wallet-web/issues/713).

usame-algan (approved):

katspaugh (approved):
👍

francovenica:
TL;DR: I have a lot of issues with WC v2, but since v1 will be deprecated soon I think we have no choice but to have it. So I'd be fine publishing it along with v1, but I'd also start to check Discord about the feedback we are going to have.

The MM phone app cannot connect to it. Probably the app is still not compatible for WC v2 and for what I read in a forum is their goal to make it compatible around this time.

Ledger: I tried ledger live as it gives it as a default option, but no matter what network I use in the web app, it will always tell me that "I'm trying to connect to a network they do not support"
image

For safe-as-owner it works the same as v1, but I see a bunch of erros when I connect the safe:
image

iamacook:
> The MM phone app cannot connect to it. Probably the app is still not compatible for WC v2 and for what I read in a forum is their goal to make it compatible around this time.

This sounds out of our control, though the follow changes may influence the. I'd suggest testing this again but if it's not working then we need to keep an eye on it.

> Ledger: I tried ledger live as it gives it as a default option, but no matter what network I use in the web app, it will always tell me that "I'm trying to connect to a network they do not support"

This has been resolved. We are now setting the `requiredChains` to that of the current one.

As [previously mentioned](https://github.com/safe-global/safe-wallet-web/pull/2018#issuecomment-1594236669) this will only work with chains supported by Ledger Live. To connect to other chains, MetaMask will need to be used in combination with the Ledger device.

> For safe-as-owner it works the same as v1, but I see a bunch of erros when I connect the safe:

We isolated the cause of this to the `@web3-onboard/walletconnect` module. We've informed WC and [they are actively looking into it](https://5afe.slack.com/archives/C0462F65CLD/p1687173093612339) but [we've patched it](https://github.com/safe-global/safe-wallet-web/pull/2095/commits/2032aeab6656b4895d61964dbb98418974234563) for now so all should work as expected.

francovenica:
We will leave MM phone app alone then. Hopefully they update the app soon
Ledger now works and recognize the networks:
image

The safe-as-owner still shows a "key.pairing" error in the console, but the rest do not show anymore. Still creating and executing tx works so I think we can file it as a different ticket for now
image

PR #2102, opened 2023-06-14
fix: Move sign msg modal to new flow

usame-algan:
## What it solves

Part of #2067 

## How this PR fixes it

- Ports the SignMsgModal (Off-chain signing) to the new tx flow
- Closes the modal when creating a safe apps tx

## How to test it

1. Open a Safe
2. Create a sign message request (e.g. walletconnect safe app using the demo eip1271 app)
3. Observe the new sign message flow
4. Press cancel
5. Observe seeing the confirmation tooltip
6. Confirm the cancellation
7. Observe the tx modal closing
8. Navigate to the Messages tab
9. Sign an existing message
10. Observe the same tx flow
11. Press cancel
12. Observe the tx modal closing
13. Press sign
14. Observe being able to sign the message

## Screenshots
Screenshot 2023-06-14 at 15 47 29

## Checklist


src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -229,11 +248,10 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement
 
   useEffect(() => {
     const unsubscribe = txSubscribe(TxEvent.SAFE_APPS_REQUEST, async ({ safeAppRequestId, safeTxHash }) => {
-      const currentSafeAppRequestId = currentRequestId
-
-      if (currentSafeAppRequestId === safeAppRequestId) {
+      if (safeAppRequestId && currentRequestId === safeAppRequestId) {
```
usame-algan, line 251: I've refactored this to be consistent with the same logic that we use for signing messages in line 263

src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -229,11 +248,10 @@ const AppFrame = ({ appUrl, allowedFeaturesList }: AppFrameProps): ReactElement
...
         trackSafeAppEvent(SAFE_APPS_EVENTS.PROPOSE_TRANSACTION, appName)
         communicator?.send({ safeTxHash }, safeAppRequestId)
+        setTxFlow(undefined)
```
usame-algan, line 254: Not related to this PR but this ensures that the tx modal closes when creating a normal safe apps tx.

katspaugh (approved):

src/components/tx-flow/common/TxLayout.tsx
```diff
@@ -32,7 +33,8 @@ const TxLayout = ({ title, children, step = 0, txSummary, onBack }: TxLayoutProp
             <Grid item xs={7} component={Paper}>
               <ProgressBar value={progress} />
 
-              <Box display="flex" justifyContent="flex-end" py={2} px={3}>
+              <Box display="flex" justifyContent="space-between" py={2} px={3}>
+                <span>{subtitle}</span>
```
iamacook, line 37:
```suggestion
                {subtitle && <span>{subtitle}</span>}
```

iamacook: Not sure if this even needs be wrapped in a `span`, e.g.

```suggestion
                {subtitle}
```

src/components/tx-flow/flows/SignMessage/SignMessage.tsx
```diff
@@ -212,54 +211,48 @@ const MsgModal = ({
...
+          />
+        </InfoBox>
+
+        <WrongChainWarning />
+
+        <MessageDialogError isOwner={isOwner} submitError={submitError} />
+      </DialogContent>
+
+      <DialogActions>
+        <Button onClick={handleClose}>Cancel</Button>
```
iamacook, line 248: Should we move this logic to the cross in the top right?

src/components/tx-flow/common/TxLayout.tsx
usame-algan: I left the span in so the txNonce is moved to the right side. We could also conditionally change the flex property. What do you prefer?

src/components/tx-flow/flows/SignMessage/SignMessage.tsx
usame-algan: Good idea! I will check to see if its possible to move this to the onClose call of the modal provider.

usame-algan: After discussing this together we decided to solve this in a separate PR because it requires some work done on the modal provider. Essentially, we need to find a way so that child components can hook into the onClose logic of the provider to prevent it from closing. Currently, its possible to pass an onClose handler to the provider but it has to be defined outside the child components. cc @katspaugh

PR #2104, opened 2023-06-15
Refactor: move AdvancedParams to ExecuteForm

katspaugh:
AdvancedParams are only used for execution, so they are now in the ExecuteForm.


src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -33,48 +32,27 @@ const SignOrExecuteForm = (props: SignOrExecuteProps): ReactElement => {
...
-      <TxSimulation
-        canExecute
-        gasLimit={advancedParams.gasLimit?.toNumber()}
-        transactions={safeTx}
-        disabled={!!gasLimitError}
-      />
-
       <DecodedTx tx={safeTx} txId={props.txId} />
 
+      <TxSimulation canExecute={canExecute} disabled={false} transactions={safeTx} />
```
usame-algan, line 41: Not passing the gasLimit to the simulation will mean that it will fall back to the latest block gasLimit and won't catch failures related to it anymore. Imo not an issue since we already show an error from the estimation if gasLimit is too low.

PR #2112, opened 2023-06-15
[Tx flow] feat: address input redesign

DiogoSoaress:
## What it solves

Resolves https://github.com/safe-global/safe-wallet-web/issues/2099

## How this PR fixes it
- Implements the Address component according to the new design
- Uses the modified components in the `CreateTokenTransfer`

## How to test:
1. Navigate to the Send tokens flow
2. Observe the new component design in the address inputs
  - The caret only shows if there is any option in the suggested list
  - If you are inputing an address that is not contained in your AB, it is suggested to add it

## Screenshots
Screenshot 2023-06-15 at 15 32 46

Screenshot 2023-06-15 at 15 33 09

Screenshot 2023-06-15 at 17 30 11

Screenshot 2023-06-15 at 17 30 16

## Checklist


src/components/common/AddressBookInput/index.tsx
```diff
@@ -25,6 +25,7 @@ const AddressBookInput = ({ name, ...props }: AddressInputProps): ReactElement =
 
   return (
     <Autocomplete
+      disableClearable
```
DiogoSoaress, line 28: I've used this prop to collapse the padding in the end of the input

katspaugh:
Another thing that I noticed is that the input jumps when selecting an address. Presumably because of the avatar. I would make the avatar smaller to fit into the input.

src/components/common/AddressBookInput/index.tsx
```diff
@@ -13,37 +15,67 @@ const abFilterOptions = createFilterOptions({
...
-        },
-      }}
-      renderOption={(props, option) => (
-        <Typography component="li" variant="body2" {...props}>
-          <EthHashInfo address={option.label} name={option.name} shortAddress={false} />
+    <>
+      <Autocomplete
+        open={open}
+        onOpen={() => setOpen(true)}
+        onClose={() => setOpen(false)}
```
iamacook, line 43: Why does this now need to be controlled?

src/components/common/AddressInput/index.tsx
```diff
@@ -55,73 +76,86 @@ const AddressInput = ({ name, validate, required = true, deps, ...props }: Addre
...
-        <Grid item>
+  const endAdornment = (
+    <InputAdornment position="end">
+      {resolving || isValidating ? (
+        <CircularProgress size={20} />
+      ) : !props.disabled ? (
+        <>
+          <IconButton onClick={() => setOpen(true)}>
+            <SvgIcon component={SaveAddressIcon} inheritViewBox fontSize="small" color="primary" />
+          </IconButton>
```
iamacook, line 87: Nit: I would not show this if the address is already present in the address book.

src/components/common/AddressInputReadOnly/index.tsx
```diff
@@ -0,0 +1,44 @@
+import { useState, type ReactElement } from 'react'
+import { IconButton, InputAdornment, InputLabel, OutlinedInput, SvgIcon, Typography } from '@mui/material'
+import EthHashInfo from '@/components/common/EthHashInfo'
+import SaveAddressIcon from '@/public/images/common/save-address.svg'
+import css from './styles.module.css'
+import EntryDialog from '@/components/address-book/EntryDialog'
+import useAddressBook from '@/hooks/useAddressBook'
+
+const AddressInputReadOnly = ({ label, address }: { label: string; address: string }): ReactElement => {
```
iamacook, line 9: A lot of this code is similar to the `AddressBookInput` where we have a read only flag. Can we use that instead?

src/components/common/AddressBookInput/index.tsx
```diff
@@ -13,37 +15,67 @@ const abFilterOptions = createFilterOptions({
...
+      />
+      {canAdd ? (
+        <Typography variant="body2" className={css.unknownAddress}>
+          <SvgIcon component={InfoIcon} fontSize="small" />
+          <span>You are sending tokens to an unknown address. Add it to address book.</span>
         </Typography>
-      )}
-      renderInput={(params) => <AddressInput {...params} {...props} name={name} />}
-    />
+      ) : null}
```
iamacook, line 77: Nit: this component is not related to sending assets. I would suggest we move this to where it is relevant. What do you think?

iamacook (commented):
The design does not truncate the sender address:

image

src/components/common/AddressBookInput/index.tsx
DiogoSoaress: So I needed to make this controlled to pass the handler to the `AddressInput` to control the Autocomplete opening from the endAdornment icon (on top of just clicking the input).

src/components/common/AddressInput/index.tsx
DiogoSoaress: If the address is present in the AB, it renders the `AddressInputReadOnly` and we only show the button there if the address is not contained in the AB.

iamacook: I see. It probably makes more sense if we use the read only flag and don't keep the `AddressInputReadOnly` component then.

src/components/common/AddressBookInput/index.tsx
DiogoSoaress: Given this is the AddressBookInput I think it makes sense to give this information. I would phrase it differently to be more generic wdyt? cc @liliiaorlenko 
> You've input an unknown address. Add it to your address book

src/components/common/AddressInputReadOnly/index.tsx
DiogoSoaress: They seem too different to me to abstract them. While `AddressBookInput` contains the Autocomplete component and deals mostly with the shown options, the `AddressInputReadOnly` is more dumb and it only displays the Address in the redesign style.

katspaugh (approved):
👍

src/components/common/AddressBookInput/index.tsx
```diff
@@ -13,37 +15,67 @@ const abFilterOptions = createFilterOptions({
...
+            name={name}
+            onOpenListClick={hasVisibleOptions ? handleOpenAutocomplete : undefined}
+            isAutocompleteOpen={open}
+          />
+        )}
+      />
+      {canAdd ? (
+        <Typography variant="body2" className={css.unknownAddress}>
+          <SvgIcon component={InfoIcon} fontSize="small" />
+          <span>You&apos;ve input an unknown address. Add it to your address book.</span>
```
iamacook, line 75: Do you think we should only show this when it's a valid address?

image

I would also suggest we tweak the wording a bit and have it open the modal when clicking on "adding it to your address book".

```suggestion
          <span>This is an unknown address. Consider adding it to your address book.</span>
```

iamacook (changes requested):

iamacook (approved):

PR #2118, opened 2023-06-16
Chore: fix e2e tests

katspaugh:
The Balances test broke because of a spam token that was airdropped on goerli. And because the backend on Goerli suddenly started filtering out all the tokens other than ETH as spam.

All the tests that depended on a wallet connect didn't work because Onboard initially has a different set of wallets enabled, and the E2E wallet appears only after the reconnection to the last wallet has already run.

### DIFF
--- a/src/hooks/wallets/useOnboard.ts
+++ b/src/hooks/wallets/useOnboard.ts
@@ -197,12 +197,8 @@ export const useInitOnboard = () => {
       onboard.state.actions.setWalletModules(supportedWallets)
     }
 
-    enableWallets()
-  }, [chain, onboard])
-
-  // Connect to the last connected wallet
-  useEffect(() => {
-    if (onboard && onboard.state.get().wallets.length === 0) {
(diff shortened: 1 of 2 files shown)


src/hooks/wallets/useOnboard.ts
```diff
@@ -197,12 +197,8 @@ export const useInitOnboard = () => {
       onboard.state.actions.setWalletModules(supportedWallets)
     }
 
-    enableWallets()
-  }, [chain, onboard])
-
-  // Connect to the last connected wallet
-  useEffect(() => {
-    if (onboard && onboard.state.get().wallets.length === 0) {
```
iamacook, line 205: We still need to check that there is no active wallet (`onboard.state.get().wallets.length === 0`) before connecting again otherwise we get this warning:

image

katspaugh: Restored it. 👍

PR #2124, opened 2023-06-19
fix: update amount field style

iamacook:
## What it solves

Implementation of the updated amount field style.

## How this PR fixes it

The amount input field has been updated according to [the designs](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?type=design&node-id=5230-264187&t=xZ8Pv94K8PF5BdpI-0).

## How to test it

Create a transaction and observe the new design of the input field.

## Screenshots

image

## Checklist


usame-algan:
I noticed that we don't show the Token amount anymore (only when the user presses the MAX button). I left a comment on the design to discuss this further.

iamacook:
> I noticed that we don't show the Token amount anymore (only when the user presses the MAX button). I left a comment on the design to discuss this further.

I agree that this would be a good improvement. Can we however merge this and make further adjustments once design have implemented them?

usame-algan (approved):

PR #2125, opened 2023-06-19
fix: execution method style

iamacook:
## What it solves

Tweaks execution method component style.

## How this PR fixes it

This slightly adjusts the style of the execution method component to match that of the [new design](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?type=design&node-id=6903-162691&t=CQl06jnWgxedwquU-0).

## How to test it

Create a transaction on a chain that supports relaying and observe the updated design for relaying execution.

Note: the component is also used in Safe creation and on the dashboard. No differences should be seen there.

## Screenshots

image

## Checklist

### DIFF
--- a/src/components/tx/SponsoredBy/index.tsx
+++ b/src/components/tx/SponsoredBy/index.tsx
@@ -17,7 +17,7 @@ const SponsoredBy = ({ relays, tooltip }: { relays: RelayResponse; tooltip?: str
   return (
     <Box className={css.sponsoredBy}>
       <SvgIcon component={GasStationIcon} inheritViewBox className={css.icon} />
-      <Stack direction="column">
+      <div>
         <Stack direction="row" spacing={0.5} alignItems="center" mb={0.2}>
           <Typography variant="body2" fontWeight={700} letterSpacing="0.1px">
             Sponsored by
@@ -47,7 +47,7 @@ const SponsoredBy = ({ relays, tooltip }: { relays: RelayResponse; tooltip?: str
             {relays.remaining} of {relays.limit}
           </Box>
         </Typography>
-      </Stack>
+      </div>
     </Box>
   )
 }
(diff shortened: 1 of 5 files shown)


src/components/tx/SponsoredBy/index.tsx
```diff
@@ -18,7 +18,7 @@ const SponsoredBy = ({ relays, tooltip }: { relays: RelayResponse; tooltip?: str
     <Box className={css.sponsoredBy}>
       <SvgIcon component={GasStationIcon} inheritViewBox className={css.icon} />
       <Stack direction="column">
```
DiogoSoaress, line 20:
```suggestion
      <div>
```

src/components/tx/SponsoredBy/index.tsx
```diff
@@ -18,7 +18,7 @@ const SponsoredBy = ({ relays, tooltip }: { relays: RelayResponse; tooltip?: str
...
-        <Stack direction="row" spacing={0.5} alignItems="center" mb={1}>
+        <Stack direction="row" spacing={0.5} alignItems="center" mb={0.2}>
```
DiogoSoaress, line 21: Nit: I'd rather put the pixel value, 0.2 always require a bit of mental effort to follow.

DiogoSoaress (approved):
Approved the scope of this PR ✅ 

I left some nit suggestions though and in general there is some more cleaning we could do in `ExecutionMethodSelector`. 
Namely some redundant uses MUI's `Box` component.
```tsx
<Box className={css.container} sx={{ borderRadius: ({ shape }) => `${shape.borderRadius}px` }}>
      <Box className={css.method}>
 ```

src/components/tx/SponsoredBy/index.tsx
iamacook: I think `0.2` is better than `1.6px`. It is reactive to any base spacing changes and the latter is also not super "clear" either. As the `spacing` prop on this component is also relevant to the base, I'd sooner leave this as is for now.

iamacook:
> Namely some redundant uses MUI's `Box` component.
> 
> ```tsx
> <Box className={css.container} sx={{ borderRadius: ({ shape }) => `${shape.borderRadius}px` }}>
>       <Box className={css.method}>
> ```

Does this quoted code relate to a requested change?

PR #2127, opened 2023-06-19
fix: remove error casting

iamacook:
## What it solves

Removes needs to cast `Error`s

## How this PR fixes it

Caught errors are typed as `unknown` because they can have any value. As such, we were blindly casting them to `Error`.

A new `asError` function is now used to coerce the `unknown` errors into an `Error` instance accordingly.

## How to test it

Observe all unit tests passing.

## Checklist


katspaugh (approved):
👍

PR #2136, opened 2023-06-20
[TX flow] Send tokens overview

### LINKED ISSUE #2027: [TX flow] tx overview confirm block (send token)
Screenshot 2023-05-22 at 15 36 40

- including Redefine (just design) but as async needs to be optional -> needs further discussion 
-

usame-algan:
## What it solves

Resolves #2027 

## How this PR fixes it

Implements the Figma design

- Also adjusts the TxLayout header design and spacing of TxCard

## How to test it

1. Create a new token transfer
2. Observe the overview screen

## Screenshots

Screenshot 2023-06-20 at 15 49 01

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/TokenTransfer/SendAmountBlock.tsx
+++ b/src/components/tx-flow/flows/TokenTransfer/SendAmountBlock.tsx
@@ -1,35 +1,19 @@
 import { type TokenInfo } from '@safe-global/safe-gateway-typescript-sdk'
-import { Grid, Skeleton, Typography } from '@mui/material'
+import { Grid, Typography } from '@mui/material'
 import css from './styles.module.css'
 import TokenIcon from '@/components/common/TokenIcon'
 import { formatAmountPrecise } from '@/utils/formatNumber'
 
-const Loading = () => {
-  return (
-    <>
-      <Skeleton variant="circular" width={26} height={26} />
-      <Skeleton variant="rounded" width={30} />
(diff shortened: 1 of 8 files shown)


src/components/tx-flow/flows/TokenTransfer/SendAmountBlock.tsx
```diff
@@ -0,0 +1,38 @@
...
+      </Grid>
+      <Grid item className={css.token}>
+        {tokenInfo ? (
+          <>
+            <TokenIcon logoUri={tokenInfo.logoUri} tokenSymbol={tokenInfo.symbol} />
+            <Typography fontWeight="bold">{tokenInfo.symbol}</Typography>
+            <Typography ml={1}>{formatAmountPrecise(amount, tokenInfo.decimals)}</Typography>
+          </>
+        ) : (
+          <Loading />
```
katspaugh, line 31: Why does it take time to load tokenInfo? Aren't balances already loaded by then?

katspaugh (approved):

src/components/tx-flow/flows/TokenTransfer/SendAmountBlock.tsx
usame-algan: Good point. Looking at it again its not really a loading flag. While the `useBalances` hook has a loading flag, as you say they are already loaded at that point so I will remove this.

PR #2140, opened 2023-06-21
fix: hide `safeTxGas` when `0`

iamacook:
## What it solves

Resolves showing `safeTxGas` on Safe versions that don't need it.

## How this PR fixes it

If the Safe version is "legacy" (<1.3.0), the `safeTxGas` field is shown. This is preferable to checking whether the value is `0` as the user may set it to `0` on legacy versions, removing the field without the options of altering it again after seeing a failing simulation.

## How to test it

- Create a transaction on a Safe <1.3.0 and observe the `safeTxGas` field present.
- Create a transaction on a Safe >=1.3.0 and observe the `safeTxGas` field _not_ present.

## Checklist

### DIFF
--- a/src/components/transactions/TxDetails/Summary/index.tsx
+++ b/src/components/transactions/TxDetails/Summary/index.tsx
@@ -76,9 +76,11 @@ export const PartialSummary = ({ safeTx }: { safeTx: SafeTransaction }) => {
   const txData = safeTx.data
   return (
     <>
-      <TxDataRow title="safeTxGas:">
-        <SafeTxGasForm />
-      </TxDataRow>
+      {!!txData.safeTxGas && (
+        <TxDataRow title="safeTxGas:">
+          <SafeTxGasForm />
+        </TxDataRow>
+      )}
       <TxDataRow title="Raw data:">{generateDataRowValue(txData.data, 'rawData')}</TxDataRow>
     </>
   )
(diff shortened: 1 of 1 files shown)


src/components/transactions/TxDetails/Summary/index.tsx
```diff
@@ -73,12 +75,15 @@ const Summary = ({ txDetails, defaultExpanded = false }: Props): ReactElement =>
...
 
 export const PartialSummary = ({ safeTx }: { safeTx: SafeTransaction }) => {
+  const { safe } = useSafeInfo()
   const txData = safeTx.data
   return (
     <>
-      <TxDataRow title="safeTxGas:">
-        <SafeTxGasForm />
-      </TxDataRow>
+      {safe.version && isLegacyVersion(safe.version) && (
```
katspaugh, line 82: Don’t you also need to check if it’s 0 for 1.3.0+ Safes? For when a safe app sends a custom safeTxGas

iamacook: Good point, then we should show it.

However, if the user was then to set it to `0`, the field would disappear if it's rendered relative to `0`. Should we just _always_ show it?

katspaugh: Well, bro, that's how it was :D

katspaugh: Seriously though, I think this whole Advanced details section is quite obscure and it's probably OK to just always show it there. In fact, that's how it has always been for _existing_ transactions. Now it's just also editable for new txs.
Maybe we should add some info tooltip discouraging folks from touching it if not necessary.

iamacook:
After [our discussion](https://github.com/safe-global/safe-wallet-web/pull/2140#discussion_r1236794128) I will close this for now and discuss with design improvements regarding providing extra information.

PR #2092, opened 2023-06-13
fix: use `baseGas` in estimation

iamacook:
## What it solves

Resolves gas etimations failing when `baseGas` is set.

## How this PR fixes it

The set `baseGas` is used instead of a default `0` when estimating gas.

## How to test it

- Create a transaction with a `refundReceiver`, `baseGas` (and potentially `gasPrice`) set and observe no failing estimation.
- [This](https://app.safe.global/transactions/tx?safe=matic:0x7FAE0c531719A2ae946f098a177F90040910B2aa&id=multisig_0x7FAE0c531719A2ae946f098a177F90040910B2aa_0xd0a128883e01dd78ab6dc1bbc90527ae00170a248cef07c5b4fad6757aabc94e) transaction should successfully estimated if impersonating an owner and attempting to execute it.

## Checklist
* [ ] I've tested the branch on mobile 📱
* [ ] I've documented how it affects the analytics (if at all) 📊
* [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #2142, opened 2023-06-21
Release 1.13.0

katspaugh:
### New features
  - Transaction security warnings using Redefine (#1902)
  - Added WalletConnect v2 wallet module (#2095)
  - Remember the state of "Execute transaction" checkbox in settings (#2075)

### Bug fixes
  - Update claiming app data URL (#2133)
  - Show messages tab if chain supports EIP-1271 (#2129)
  - Use `baseGas` in estimation (#2092)
  - Sign messages with hardware wallets through Metamask (#2079)
  - Used REVERSE_PROXY_UI_PORT env var in the serve command (#2082)


francovenica:
Only 1 small UI issue with the list of token/NFT's in the "Balance change" block:
image

The rest is fine. I've checked the tickets are fine as we checked in their PR's (We already discussed how WC v2 right now is not readable for MM phone app and is kinda wonky with ledger)
The regression was done shorter than usual since we had a release recently already. Still checked the main flows:
Tx creation/addition, owner edition, send funds, Tx builder, spending limits, address book, a few Dapps...

Looks good to me

PR #2144, opened 2023-06-21
refactor: Remove promise workaround in tests

usame-algan:
## What it solves

Removes the workarounds we have in some of our tests `await act(() => Promise.resolve())`. These only work because the state updat happens on the next tick of the event loop. Either using `await waitFor` or RTL built-in selectors like `await screen.findByText(...)` will achieve the same.

## How to test it

All tests should still be green


katspaugh (approved):
🧠

PR #2155, opened 2023-06-22
[TX flow] Display Token balance

usame-algan:
## How this PR fixes it

- Displays the token balance in the amount input field
- Slightly decreases the spacing of the amount input as discussed [here](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF?node-id=5230:264187#476048952)

## How to test it

1. Open the Safe
2. Create a new transaction
3. Observe the token balance is visible

## Screenshots
Screenshot 2023-06-22 at 15 32 36
Screenshot 2023-06-22 at 15 32 43

## Checklist


iamacook (commented):
Looks good apart from the alignment of the max button and the token selector. I would suggest making the max button a little bit taller so they are the same height.

usame-algan:
> Looks good apart from the alignment of the max button and the token selector. I would suggest making the max button a little bit taller so they are the same height.

I slightly decreased the spacing and line-height to get it back to the same size instead since its closer to the design (50px height button)

iamacook (approved):

PR #2160, opened 2023-06-23
fix: Enable spending limit txs in new flow

usame-algan:
## What it solves

Part of #2067 

## How this PR fixes it

- Moves the existing spending limit flow into the new tx flow
- Updates "Send again" for the new tx flow
- Updates "Send" via address book for the new tx flow

## How to test it

1. Open a safe with a spending limit
2. Create a new transaction
3. Select the token with a spending limit
4. Select Spending limit from the checkbox
5. Press Next
6. Observe the spending limit review screen
7. Press Submit
8. Observe a transaction without nonce
9. Go to history
10. Open a send transaction
11. Press the three dots and select Send again
12. Observe the new tx flow
13. Go to address book
14. Press send next to one of the entries
15. Observe the new tx flow

## Screenshots
Screenshot 2023-06-23 at 11 31 36
 
## Checklist


usame-algan:
For now I only copied the existing flow but I suggest we split them up in a separate PR and also style the inner components.

katspaugh (approved):

PR #2163, opened 2023-06-23
Fix: hide Edit for safeTxGas when 0 in 1.3.0+

katspaugh:
Another take on #2140

The safeTxGas Edit button will be shown only of safeTxGas is > 0 or it's a Safe below v1.3.0.

Also added baseGas and refundReceiver.

Screenshot 2023-06-23 at 16 26 40

### DIFF
--- a/src/components/transactions/TxDetails/SafeTxGasForm.tsx
+++ b/src/components/transactions/TxDetails/SafeTxGasForm.tsx
@@ -3,15 +3,15 @@ import { Link, Box, Paper, Button } from '@mui/material'
 import { useForm } from 'react-hook-form'
 import { SafeTxContext } from '@/components/tx-flow/SafeTxProvider'
 import NumberField from '@/components/common/NumberField'
+import useSafeInfo from '@/hooks/useSafeInfo'
+import { isLegacyVersion } from '@/hooks/coreSDK/safeCoreSDK'
 
 type FormFields = {
   safeTxGas: number
 }
 
-const SafeTxGasForm = () => {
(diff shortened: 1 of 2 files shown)


src/components/transactions/TxDetails/SafeTxGasForm.tsx
```diff
@@ -20,52 +20,60 @@ const SafeTxGasForm = () => {
...
+    <Box display="flex" alignItems="center" gap={1} position="relative">
+      {safeTxGas}
+
+      {isEditable && (
+        <Link component="button" onClick={() => setEditing(true)} fontSize="small">
+          Edit
+        </Link>
       )}
+
+      {editing && <Form onSubmit={() => setEditing(false)} />}
```
katspaugh, line 76: Moving the form into a separate component fixes the bug that the default value wasn't updated when safeTxGas is recalculated.

src/components/transactions/TxDetails/SafeTxGasForm.tsx
```diff
@@ -20,52 +20,58 @@ const SafeTxGasForm = () => {
...
-              })}
-            />
-            <Button type="submit" size="small" variant="contained" sx={{ ml: 1 }}>
-              Save
-            </Button>
-          </form>
-        </Paper>
+  return (
+    <Box display="flex" alignItems="center" gap={1} position="relative">
+      {safeTxGas}
```
usame-algan, line 66: I also just realized that we need to adjust the prio of `finalSafeTxGas` in the provider in case a safe app sends it it should take prio over the recommended value. Right now all Safe App txs send safeTxGas regardless of safe version but we display 0.

usame-algan (approved):

PR #2171, opened 2023-06-26
fix: onboard warning colours

iamacook:
## What it solves

Resolves unclear warning perceived as an error

## How this PR fixes it

The warning palette of onboard has been aligned with those from the Safe.

Note: previously the rejection message was red but with the palette adjustment, it is now yellow.

image

## How to test it

Open the wallet connection modal and observe the "Why don't I see my wallet?" information in a yellow wrapper.

No error messages should have chaged, e.g. not able to connect to hardware wallet:

image

## Screenshots

image

## Checklist

### DIFF
--- a/src/styles/onboard.css
+++ b/src/styles/onboard.css
@@ -40,11 +40,11 @@
   --onboard-danger-600: var(--color-error-main);
   --onboard-danger-700: var(--color-error-dark);
 
-  --onboard-warning-100: var(--color-error-background);
-  --onboard-warning-400: var(--color-error-light);
-  --onboard-warning-500: var(--color-error-light);
-  --onboard-warning-600: var(--color-error-main);
-  --onboard-warning-700: var(--color-error-dark);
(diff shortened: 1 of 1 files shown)


katspaugh (approved):

francovenica:
The message is not present anymore.

Still, I wasn't able to trigger this "if not auto-selecting a wallet", to see if the message would popUp again. I tried connecting to ledger in one network and switch to another where is not present and refresh the page, but that doesn't seem to be the way to trigger the message.

iamacook:
> Still, I wasn't able to trigger this "if not auto-selecting a wallet", to see if the message would popUp again. 

The [initial modal is not shown](https://github.com/safe-global/safe-wallet-web/blob/onboard-palette/src/hooks/wallets/useOnboard.ts#L205) when we reconnect to a previously connected wallet.

usame-algan:
Opened a PR against bnc-onboard repo to [add an option to hide this notice](https://github.com/blocknative/web3-onboard/pull/1786)

PR #2175, opened 2023-06-26
[TX flow] Adds nonce validation

### LINKED ISSUE #2110: [TX flow] Nonce form
## What is the feature about

The new nonce input needs to be implemented according to the design

## The list of requirements

- Should have a button to reset to recommended nonce
- Should display a dropdown list of all queued transactions

## Designs/sketches
Screenshot 2023-06-15 at 15 29 55

## Links

[Figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?type=design&node-id=5188-171636&t=GWGq12B5tpydDFmy-0)

usame-algan:
## What it solves

Resolves #2110 

## How this PR fixes it

- Displays an error message in a tooltip around the nonce form
- Always displays the reset button but in a disabled state to prevent the input from jumping

## How to test it

1. Open the Safe
2. Create a new transaction
3. Choose a nonce lower than the current safe nonce
4. Observe an error inside a tooltip

## Screenshots
Screenshot 2023-06-26 at 17 08 10

## Checklist


src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
```diff
@@ -224,7 +224,7 @@ const CreateTokenTransfer = ({
           </FormControl>
 
           {isDisabled && (
-            <Box mt={1} display="flex" alignItems="center">
+            <Box display="flex" alignItems="center" mt={-2} mb={3}>
```
usame-algan, line 227: This fixes the issue that the safe error is too close to the divider

katspaugh (approved):

katspaugh:
> We should also think about disabling the "Next" and "Submit" button if the nonce is invalid. Previously this was covered since the nonce was edited in a separate modal and the user couldn't leave without entering a valid input there. I suggest we add another state nonceError to the context but it would have to be added to every component that has Next or Submit buttons.

How about just not letting the user choose an incorrect nonce?

usame-algan:
After our discussion:

1. We don't set the nonce if there is a validation error. Once the user blurs the input it will reset back to the previous value that was valid
2. Add a max number input validation so it doesn't flip to "Infinity"

src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -51,34 +51,56 @@ const NonceFormOption = memo(function NonceFormOption({
...
+      if (isNaN(nonce)) {
+        return 'Nonce must be a number'
+      }
+
+      if (nonce < safe.nonce) {
+        return `Nonce can't be lower than ${safe.nonce}`
+      }
+
+      if (nonce >= Number.MAX_SAFE_INTEGER) {
+        return 'Nonce is too big'
```
iamacook, line 75:
```suggestion
        return 'Nonce is too high'
```

src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -51,34 +51,56 @@ const NonceFormOption = memo(function NonceFormOption({
...
     },
     [safe.nonce],
   )
 
   const handleChange = useCallback(
     (_e: SyntheticEvent, value: string | AutocompleteValue<unknown, false, false, false>) => {
+      const isNotValid = validateInput(value)
+
+      if (isNotValid) {
+        setError(validateInput(value))
```
iamacook, line 86:
```suggestion
      const error = validateInput(value)

      if (error) {
        setError(error)
```

src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -51,34 +51,56 @@ const NonceFormOption = memo(function NonceFormOption({
...
-      setError(!isValidInput(value))
-      setNonce(nonce)
+      setNonce(Number(value))
     },
-    [isValidInput, setNonce],
+    [validateInput, setNonce],
   )
 
+  const handleBlur = useCallback(() => {
+    setError(undefined)
```
iamacook, line 99: What if the current `nonce` is erroneous?

iamacook:
For visibility (as per our discussion) two thoughts:

1. A `Tooltip` inside a `Tooltip` seemed initially wrong. However, [MUI mounts a `Popper` outside of the `children`](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Tooltip/Tooltip.js#L678) so that's ok.
2. When replacing a transaction, we should disable nonce editing:
image

src/components/tx-flow/common/TxNonce/index.tsx
usame-algan: Since we don't save the nonce if its incorrect the current nonce must be correct so we can safely remove the error.

usame-algan:
> I'm not sure about preventing the deletion of the value. I was initially confused why I couldn't backspace. (Here I'm trying to delete the value:)

I agree its not ideal but since we use the same value for controlling the input we can't easily let the user remove and not update it.

iamacook:
> I agree its not ideal but since we use the same value for controlling the input we can't easily let the user remove and not update it.

We could save the new nonce if validation fails and not return early.

iamacook (approved):

PR #2178, opened 2023-06-26
fix: don't check WC v1 session for v2 reconnection

iamacook:
## What it solves

Resolves WC v1 session caching interfering with v2.

## How this PR fixes it

The local session of WC v1 is no longer checked when trying to auto-connect to v2.

## How to test it

Create a session with WC v1 and refresh the Safe. Observe that WC v2 does not try to auto-connect.

## Checklist


katspaugh (approved):

francovenica:
LGTM

Connecting with WC v1 and refreshing doesn't show WC v2 trying to auto connect (it just reconnects to WC v1)

PR #2179, opened 2023-06-27
[e2e] - fix dashboard - change token counter

francovenica:
## What it solves

The safe has received a spam token, increasing the token counter in the dashboard. For now I just updated the number so the test passes since there isn't much we can do about spam tokens


katspaugh (approved):
Thanks!

PR #2185, opened 2023-06-27
Fix: restore tracking of send modals

katspaugh:
I've restored the tracking of the following tx modal events:

 * `Send tokens`
 * `Send NFTs`
 * `Contract interaction`

They are now tracked again when clicking on the corresponding buttons.


usame-algan (approved):
The buttons look a bit off but I assume we will tackle this with the redesign

Screenshot 2023-06-27 at 13 13 41

katspaugh:
I'll merge this as an ugly reminder to design that modal :)

PR #2187, opened 2023-06-27
feat: show alert before closing transaction flow

iamacook:
## What it solves

Resolves direct exit of new transaction flow

## How this PR fixes it

A confirmation dialog is now shown when exiting or navigating away from the new transaction flow which needs to first be accepted.

## How to test it

Create a transaction, attempt to close the flow/navigate away with back button and observe the new confirmation dialog.

## Screenshots

image

alert

## Checklist

### DIFF
--- a/src/components/tx-flow/common/TxFlowExitWarning/index.tsx
+++ b/src/components/tx-flow/common/TxFlowExitWarning/index.tsx
@@ -1,0 +1,46 @@
+import { DialogContent, DialogContentText, DialogActions, Button, SvgIcon } from '@mui/material'
+import type { ReactElement } from 'react'
+
+import ModalDialog from '@/components/common/ModalDialog'
+import AlertIcon from '@/public/images/notifications/alert.svg'
+
+export const TxFlowExitWarning = ({
+  open,
+  onCancel,
+  onClose,
+}: {
+  open: boolean
+  onCancel: () => void
+  onClose: () => void
+}): ReactElement => {
+  return (
+    <ModalDialog
+      open={open}
+      onClose={onCancel}
+      dialogTitle={
+        <>
+          <SvgIcon
+            component={AlertIcon}
+            inheritViewBox
+            fontSize="inherit"
+            color="error"
+            sx={{ verticalAlign: 'middle', mr: '12px' }}
+          />{' '}
+          Are you sure?
+        </>
+      }
+    >
+      <DialogContent sx={{ p: 'var(--space-3) !important' }}>
+        <DialogContentText color="primary">
+          If you leave this page, the current progress will be lost.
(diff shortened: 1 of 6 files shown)


src/components/tx-flow/common/TxFlowExitWarning/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+            color="error"
+            sx={{ verticalAlign: 'middle', mr: '12px' }}
+          />{' '}
+          Are you sure?
+        </>
+      }
+    >
+      <DialogContent sx={{ p: 'var(--space-3) !important' }}>
+        <DialogContentText color={isDarkMode ? undefined : 'primary'}>
+          If you leave this page, the current progress will be lost.
```
katspaugh, line 37: This sounds like a threat.

src/components/tx-flow/flows/SignMessage/SignMessage.tsx
```diff
@@ -175,9 +130,9 @@ export type ConfirmProps = BaseProps & {
   requestId?: RequestId
 }
 
-const SignMessage = ({ onClose, message, safeAppId, requestId }: ProposeProps | ConfirmProps): ReactElement => {
+const SignMessage = ({ message, safeAppId, requestId }: ProposeProps | ConfirmProps): ReactElement => {
   // Hooks & variables
-  const [showCloseTooltip, setShowCloseTooltip] = useState<boolean>(false)
+  const { onClose } = useContext(TxModalContext)
```
katspaugh, line 135: How is this used? It was supposed to be an internal callback that outside components should only assign, but not call. Only TxModalProvider should call it. It shouldn't even be exported (my mistake that it is).

iamacook: When the user submits the [last sync signature](https://github.com/safe-global/safe-wallet-web/blob/dev/src/hooks/messages/useSyncSafeMessageSigner.ts#L37-L57) or it is received [via poll](https://github.com/safe-global/safe-wallet-web/blob/dev/src/hooks/messages/useSyncSafeMessageSigner.ts#L21-L25), it [closes the modal](https://github.com/safe-global/safe-wallet-web/blob/dev/src/services/safe-messages/safeMsgNotifications.ts#L34-L40).

katspaugh: In this scenario you’re supposed to call setTxFlow(undefined).

iamacook: Good point. I've updated it accordingly, as well as removing `onClose` from the context.

katspaugh (approved):

PR #2189, opened 2023-06-27
fix: disable transition + backdrop of modals

iamacook:
## What it solves

Resolves workaround to hide modal backdrop

## How this PR fixes it

This replaces the `TransitionComponent` and `slot.backdrop` with a `Fragment`, removing the open transition and backdrop from `TxModalDialog`.

## How to test it

Create a transaction and observe no transition when opening the modal, as well as no `MuiDialog-backdrop` in the DOM:

Previous error:

image

Now:

image

## Checklist


katspaugh (approved):
Thank you!

PR #2196, opened 2023-06-29
Fix: autofocus recipient if not set

katspaugh:
## What it solves

Send Tokens: auto-focus on the recipient field unless a recipient address is passed as a prop.


src/components/common/TxModalDialog/index.tsx
```diff
@@ -25,8 +25,7 @@ const TxModalDialog = ({
       scroll={fullScreen ? 'paper' : 'body'}
       className={classnames(css.dialog, { [css.fullWidth]: fullWidth })}
       onClick={(e) => e.stopPropagation()}
-      TransitionComponent={Fragment}
```
iamacook, line 28: If we don't set the `TransitionComponent`, the modal does not immediately mount. It's not an issue but thought I'd make you aware.

iamacook (approved):

PR #2211, opened 2023-06-30
[PoC] feat: use WalletConnect outside of Safe Apps

iamacook:
## What it solves

Makes WalletConnect a first-class citizen

## How this PR fixes it

A new WC icon in the header mounts an `AppFrame`, loading the WC Safe App in the background. This allows for interaction with the Safe, no longer needing to keep the WC Safe App open.

## How to test it

1. Open the Safe and connect to an EOA.
2. Open the new WC modal by clicking on the icon on the header.
3. Connect to another Safe or DApp via WC.
4. You can exit the modal and navigate around the Safe, but still transact via WC.

## Screenshots

The bottom Safe is connected to an EOA, whilst the top on connects via WC:

wc header

## Checklist

### DIFF
--- a/src/components/common/Header/index.tsx
+++ b/src/components/common/Header/index.tsx
@@ -14,6 +14,7 @@ import useChainId from '@/hooks/useChainId'
 import SafeLogo from '@/public/images/logo.svg'
 import Link from 'next/link'
 import useSafeAddress from '@/hooks/useSafeAddress'
+import { WalletConnect } from '@/components/common/WalletConnect'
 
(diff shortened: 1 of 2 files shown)


katspaugh:
Awesome work! And so simple!

Now let's wrap it into a browser extension, and we got ourselves a Safe Connect. 😎

iamacook:
As per discussion, we will close this for now but use it as reference when we tackle the roadmap item.

katspaugh:
For the record: I loved it

PR #2218, opened 2023-07-03
Fix: pass Redefine API URL from GitHub secrets

katspaugh:
I've added an env var but forgot to pass it in the build YML.


usame-algan (approved):

katspaugh:
I've tested on the PR preview, Redefine works now.

PR #2227, opened 2023-07-04
fix: TxFlow submit button alignment, error message spacing

usame-algan:
## What it solves

Part of #2067 

## How this PR fixes it

- Aligns the submit button to the right at all times
- Adds spacing above error messages

## How to test it

1. Open a Safe
2. Confirm an existing transaction
3. Switch to a wallet thats not an owner
4. Observe the Submit button still aligned to the right
5. Create a transaction that will fail
6. Observe enough space above the error message

## Screenshots

Screenshot 2023-07-04 at 16 49 38

Screenshot 2023-07-04 at 16 49 55

## Checklist

### DIFF
--- a/src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
+++ b/src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
@@ -96,7 +96,6 @@ const ExecuteForm = ({
   }
 
   const submitDisabled = !safeTx || !isSubmittable || disableSubmit || isValidExecutionLoading || isExecutionLoop
-  const hasError = isExecutionLoop || executionValidationError || gasLimitError || submitError
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
```diff
@@ -122,21 +123,25 @@ const ExecuteForm = ({
...
-            {isNewExecutableTx
-              ? 'To save gas costs, avoid creating the transaction.'
-              : 'To save gas costs, reject this transaction.'}
-          </ErrorMessage>
-        ) : (
-          submitError && (
-            <ErrorMessage error={submitError}>Error submitting the transaction. Please try again.</ErrorMessage>
-          )
+        {hasError && (
+          <div className={css.errorWrapper}>
```
iamacook, line 127: Not sure if it's better or worse but if you pass the `className` to the `ErrorMessage` components, you need not create the `hasError` flag. What do you think?

iamacook (changes requested):
We need to do the same for the [`SignForm`](https://github.com/safe-global/safe-wallet-web/blob/epic-tx-flow/src/components/tx/SignOrExecuteForm/SignForm.tsx#L67-L75) as well.

src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
usame-algan: Nice, I didn't see that we could already pass a `className` to `ErrorMessage`. Lets do it that way so we can remove `hasError`.

iamacook (approved):

PR #2229, opened 2023-07-05
<fix: whitespace in pending tx dashboard>

### LINKED ISSUE #2060: CSS bug in pending txs Dashboard widget
Screenshot 2023-05-30 at 12 34 11

There's too much white space on the left, and the right part is cramped. Items should be evenly spread instead.

ElizabethOgbee:
## What it solves
This PR solves the excess whitespace on the pending trasaction dashboard

Resolves #2060

## How this PR fixes it
This PR fixes it by changing the grid to flex and then added a justify-content of space between which will spread out the items evenly and finally, flex-wrap was added to wrap up the item as the screen gets smaller so that their will no be an overflow as the screen gets smaller.

After styling the flex, the was another excess whitespace between the SvgIcon and ChevronRight icons and so i had to reduce the 12px marginleft on the ChevronRight tag at line 64 of the pendingTxListItem component to 0px to eliminate the excess whitespace 

The only concern i have is the container name which is .grid-container. I'm thinking that since the style has been changed from grid to flex, I thought that the container name should also be changed

## Screenshots
Picture

## Checklist
* [ ✔] I've tested the branch on mobile 📱
Picture1

### DIFF
--- a/src/components/dashboard/PendingTxs/PendingTxListItem.tsx
+++ b/src/components/dashboard/PendingTxs/PendingTxListItem.tsx
@@ -61,7 +61,7 @@ const PendingTx = ({ transaction }: PendingTxType): ReactElement => {
             )}
           </Box>
 
-          <Box gridArea="icon" marginLeft="12px">
+          <Box gridArea="icon" marginLeft="0px">
             <ChevronRight color="border" />
           </Box>
         </Box>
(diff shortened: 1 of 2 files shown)


katspaugh:
On your screenshot, it looks like the green box is now too far from the arrow.
Screenshot 2023-07-03 at 09 42 32

katspaugh:
Thanks for redoing the PR btw, looks good overall!

PR #2231, opened 2023-07-05
fix: Apply input styles to all input fields

usame-algan:
## What it solves

Part of #2067 

## How this PR fixes it

- Extracts inputs styles into a common css file `inputs.module.css`
- Applies the new input styles (66px height, border-color) to all input fields

## How to test it

1. Open a Safe
2. Create/Add a safe and observe all input fields have the same border color and height
3. Do the same for all other flows
4. Check the transaction history filter

## Screenshots
Screenshot 2023-07-05 at 11 45 27
Screenshot 2023-07-05 at 11 45 58

## Checklist


usame-algan:
I suggest we do this in the theme file once we copy it over to this repository instead of having a custom css that we need to remember to always include.

iamacook (approved):

PR #2233, opened 2023-07-05
fix: Hide confirmation dialog if set via setTxFlow

usame-algan:
## What it solves

Part of #2067 

## How this PR fixes it

- Uses the browser native confirmation dialog
- Prevents navigation if user cancels
- Adds another option to `setTxFlow` to bypass confirmation dialog

## How to test it

1. Open a Safe
2. Create new transaction
3. Navigate away
4. Observe no confirmation dialog
5. Click on Send tokens
6. Try to navigate away or close the window
7. Observe a confirmation dialog

## Screenshots
Screenshot 2023-07-05 at 15 21 33

## Checklist

### DIFF
--- a/src/components/tx-flow/common/TxNonce/index.tsx
+++ b/src/components/tx-flow/common/TxNonce/index.tsx
@@ -149,7 +149,6 @@ const TxNonceForm = ({ nonce, recommendedNonce }: { nonce: number; recommendedNo
                       readOnly,
                     }}
                     className={css.input}
-                    title={field.value}
                     sx={{
                       minWidth: `clamp(1ch, ${field.value.length}ch, 200px)`,
                     }}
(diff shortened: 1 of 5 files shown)


usame-algan:
We might be able to use our own confirmation dialog but we need to figure out how to do it all in one callback that we can await and throw if necessary. The previous approach of rendering another component depending on a local state didn't work with blocking the router event.

src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -149,7 +149,6 @@ const TxNonceForm = ({ nonce, recommendedNonce }: { nonce: number; recommendedNo
                       readOnly,
                     }}
                     className={css.input}
-                    title={field.value}
```
usame-algan, line 152: This caused MUI errors because Tooltips can't also contain a title.

katspaugh (approved):
Neat.

PR #2236, opened 2023-07-05
Chore: escape quotes in release action

katspaugh:
## What it solves

The release action pastes the contents of the RC pull request into a bash variable using an echo command, but if the PR description contains a quote it breaks the bash syntax.

E.g.
```
Remove "What's new"
```

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
         if: github.event.pull_request.merged == true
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-release.yml
```diff
@@ -3,7 +3,7 @@ name: Tag & GitHub release
 on:
   pull_request_target:
     branches:
-      - main
+      - dev
```
usame-algan, line 6: Why the change from main to dev?

katspaugh:
After trying a few different things, the best solution was to pass the body as a variable instead of a file. For that, I've replaced our release action which also fixed the obsolete set_output warnings.

PR #2239, opened 2023-07-06
fix: reduce space between name and address

iamacook:
## What it solves

Unclear spacing

## How this PR fixes it

This reduces the margin between the address and name (if it exists) within the transaction details for clarity.

## How to test it

Ensure an address is in the address book and create a transaction to that address. Observe the reduced margin between the name/address.

## Screenshots

image

## Checklist

### DIFF
--- a/src/components/transactions/TxDetails/Summary/TxDataRow/styles.module.css
+++ b/src/components/transactions/TxDetails/Summary/TxDataRow/styles.module.css
@@ -26,6 +26,11 @@
   align-items: center;
 }
 
+/* 2nd only exists if there is address book entry */
+.address > div > div > div:nth-child(2) {
+  margin-top: -4px;
+}
+
 @media (max-width: 599.95px) {
   .gridRow {
     grid-template-columns: 1fr;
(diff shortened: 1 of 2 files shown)


src/components/transactions/TxDetails/Summary/TxDataRow/styles.module.css
```diff
@@ -26,6 +26,11 @@
   align-items: center;
 }
 
+/* 2nd only exists if there is address book entry */
+.address > div > div > div:nth-child(2) {
+  margin-top: -4px;
+}
```
usame-algan, line 32: This works but we need to keep in mind that it highly depends on the structure of EthHashInfo. As an alternative we could also decrease the line-height of the first child although its not as effective but assumes less of the structure e.g.

```
.address > div {
  line-height: 1;
}
```

@liliiaorlenko did we consider also increasing the spacing between each row like this (4px)?
Screenshot 2023-07-06 at 15 43 00

usame-algan (approved):

katspaugh (commented):
I don't get this PR. You're moving the address up if it has a name? I thought the idea was to space them out, i.e. push away from each other, not closer.

Also not a fan of the > div > div thing.

iamacook:
> I don't get this PR. You're moving the address up if it has a name? I thought the idea was to space them out, i.e. push items in the ABI arguments away from each other, not closer.
> 
> Also not a fan of the > div > div thing.

As per discussion, I've opted for the `line-height` suggestion above.

katspaugh (approved):

PR #2243, opened 2023-07-06
fix: Write tests for ApprovalEditor

usame-algan:
## What it solves

Part of #2067

## How this PR fixes it

- Write tests for `ApprovalEditor`, `ApprovalEditorForm` and `useApprovalInfos`


schmanu (approved):
Great tests 🥇

PR #2244, opened 2023-07-06
Persist tx flow on reload

katspaugh:
When you open a new tx modal and reload the page, the popup will reappear.

It doesn't save the internal state, however, only the external props.

It also changes the URL to have a `tx=true` query param which makes the Back button close the modal.

### DIFF
--- a/src/components/tx-flow/index.tsx
+++ b/src/components/tx-flow/index.tsx
@@ -76,7 +76,7 @@ export const TxModalProvider = ({ children }: { children: ReactNode }): ReactEle
     })
     dispatch(clearNewTx(pathname))
     setQueryParam(undefined)
-  }, [dispatch, pathname, setQueryParam, setOnClose, setQueryParam])
+  }, [dispatch, pathname, setOnClose, setQueryParam])
 
   const handleShowWarning = useCallback(() => {
     if (!shouldWarn) {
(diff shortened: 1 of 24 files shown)


src/components/tx-flow/index.tsx
```diff
@@ -1,11 +1,41 @@
...
+  ConfirmTxFlow: require('@/components/tx-flow/flows/ConfirmTx').default,
+  ReplaceTxMenu: require('@/components/tx-flow/flows/ReplaceTx').default,
+  NftTransferFlow: require('@/components/tx-flow/flows/NftTransfer').default,
+}
 
 const noop = () => {}
 
+export type TxFlow = {
+  component: keyof typeof Flows
+  props?: any
```
katspaugh, line 34: @iamacook can you remind me how you guys made this type safe originally? Before we pivoted to defining the current flow as a JSX element.

iamacook: You'll need to cast every import in `Flows` to their respective component as `require` imports are `any`, then you can extract their props with the helper type of React, `ComponentProps`, something like:

```ts
type TxFlow<T extends keyof typeof Flows> = {
  component: T
  props: ComponentProps<typeof Flows[T]>
}
```

PR #2246, opened 2023-07-06
fix: use the deployment by `chainId`

iamacook:
## What it solves

Fallback handler/unsupported mastercopy warnings on non-default addressed chains

## How this PR fixes it

The deployment address of the current chain is no longer assumed to be ["default"](https://github.com/safe-global/safe-deployments/blob/main/src/assets/v1.3.0/gnosis_safe.json#L2), but is instead [referenced by `chainId`](https://github.com/safe-global/safe-deployments/blob/main/src/assets/v1.3.0/gnosis_safe.json#L7).

All instances of `defaultAddress` have been chanced to `networkAddresses[chainId]`.

## How to test it

- Open a non-supported mastercopy and observe the warning, as well as no warning on official deployments.
- Open a Safe with a fallback handler (of non-default address) set and observe no warning in the settings that it is unofficial. When an unofficial on is set, however, it should still warn.

## Checklist

### DIFF
--- a/src/components/settings/FallbackHandler/__tests__/index.test.tsx
+++ b/src/components/settings/FallbackHandler/__tests__/index.test.tsx
@@ -45,6 +45,43 @@ describe('FallbackHandler', () => {
     })
   })
 
+  it('should render the Fallback Handler without warning when one that is not a default address is set', async () => {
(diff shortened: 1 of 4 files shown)


schmanu (approved):
Good catch 🎣 
Let's add a small test case that verifies this.

liliya-soroka:
Verified
Tested:
1. official deployed fallback handlers for the networks
2. enabled failed fallback handler to check the warning

PR #2247, opened 2023-07-06
fix: readonly nonce input

schmanu:
## What it solves
Improves readability of nonce input in readonly mode + reduces the padding of the input field.
- change nonce input for signed txs to readonly.
- only add `padding-right` to input if recommendedNonce button is shown.

## Screenshots
- Edit mode with reset to recommended nonce button
Screenshot 2023-07-07 at 11 17 57

- Edit mode without reset to recommended nonce button
Screenshot 2023-07-07 at 11 21 41

- Readonly mode for queued & signed tx
Screenshot 2023-07-07 at 11 27 35

## Checklist


src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -116,7 +119,7 @@ const TxNonceForm = ({ nonce, recommendedNonce }: { nonce: number; recommendedNo
               }
             }}
             options={previousNonces}
-            disabled={readOnly}
+            readOnly={readOnly}
```
katspaugh, line 122: I liked your initial idea to just show a span in this case.

usame-algan (commented):
I would suggest we remove the input border when its readOnly or change it so its more "static". Right now the border changes width when clicking into the input. We could also render a span instead like it was suggested

src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -136,21 +144,26 @@ const TxNonceForm = ({ nonce, recommendedNonce }: { nonce: number; recommendedNo
...
+                    className={classNames([
+                      css.input,
+                      {
+                        [css.withAdornment]: showRecommendedNonceButton,
+                      },
+                    ])}
                     sx={{
-                      minWidth: `clamp(1ch, ${field.value.length}ch, 200px)`,
+                      minWidth: `clamp(1ch, calc(${field.value.toString().length}ch${
+                        showRecommendedNonceButton ? ' + 28px' : ' + 4px'
```
schmanu, line 165: I couldn't figure out a more elegant way to do handle the dynamic widths :/

src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -102,6 +103,14 @@ const TxNonceForm = ({ nonce, recommendedNonce }: { nonce: number; recommendedNo
         },
       }}
       render={({ field, fieldState }) => {
+        if (readOnly) {
+          return <>{nonce}</>
+        }
+
+        console.log(field, field.value)
```
usame-algan, line 110:
```suggestion

```

usame-algan (approved):

src/components/tx-flow/common/TxNonce/index.tsx
usame-algan: If we want to have the input as small as possible we need to take this into account so I think its fine.

src/components/tx-flow/common/TxNonce/index.tsx
schmanu: whoops. Return of the classic console log leftover 🤦

katspaugh:
It's almost as per the design at this point, the only thing lacking is bold.
Screenshot 2023-07-07 at 12 21 28

PR #2262, opened 2023-07-10
fix: Hide approval editor for non-approval transactions

usame-algan:
## What it solves

Part of #2067 

## How this PR fixes it

- Hides the approval editor when a tx doesn't have any approvals

## How to test it

1. Open a Safe with a non-approval transaction in the queue
2. Execute/Sign the transaction
3. Observe the ApprovalEditor is not flashing

## Checklist

### DIFF
--- a/src/components/tx/ApprovalEditor/index.tsx
+++ b/src/components/tx/ApprovalEditor/index.tsx
@@ -1,4 +1,4 @@
-import { Box, Divider, SvgIcon, Typography } from '@mui/material'
+import { Alert, Box, Divider, Skeleton, SvgIcon, Typography } from '@mui/material'
 import { type MetaTransactionData, type SafeTransaction } from '@safe-global/safe-core-sdk-types'
 import css from './styles.module.css'
 import { ApprovalEditorForm } from './ApprovalEditorForm'
@@ -32,9 +32,9 @@ export const ApprovalEditor = ({
   safeTransaction: SafeTransaction | undefined
   updateTransaction?: (txs: MetaTransactionData[]) => void
 }) => {
-  const [readableApprovals] = useApprovalInfos(safeTransaction)
+  const [readableApprovals, error, loading] = useApprovalInfos(safeTransaction)
 
-  if (!readableApprovals || readableApprovals?.length === 0 || !safeTransaction) {
+  if (readableApprovals?.length === 0 || !safeTransaction) {
     return null
   }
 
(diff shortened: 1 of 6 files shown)


src/components/tx/ApprovalEditor/index.tsx
```diff
@@ -32,9 +32,9 @@ export const ApprovalEditor = ({
   safeTransaction: SafeTransaction | undefined
   updateTransaction?: (txs: MetaTransactionData[]) => void
 }) => {
-  const [readableApprovals, error, loading] = useApprovalInfos(safeTransaction)
+  const [readableApprovals] = useApprovalInfos(safeTransaction)
 
-  if (readableApprovals?.length === 0 || !safeTransaction) {
+  if (!readableApprovals || readableApprovals?.length === 0 || !safeTransaction) {
```
usame-algan, line 37: @schmanu this is a regression from when I added the tests but having this early return here also means there can't be an error/loading state. Ideally, we would check if there are any approvals before going into this component or before calling the hook but we would have to "duplicate" the logic from the module. What do you think?

schmanu: If I understand the problem correctly this could help:

Looking at the `useApprovalInfos` hook, the first step is to scan if there are any approvals present, which does not need to be async.
We could consider modifying the `useAsync` result a bit before returning so that it better represents the loading states

i.e.
```ts
export const useApprovalInfos = (safeTransaction: SafeTransaction | undefined) => {
  const [approvals] = useApprovalData(safeTransaction)

  const { balances } = useBalances()

  const hasApprovalSignatures = approvals && approvals.payload && approvals.payload.length > 0

  const [approvalInfos, error, loading] = useAsync<ApprovalInfo[]>(
    async () => {
      if (!hasApprovalSignatures) return Promise.resolve([])

      return Promise.all(
        approvals.payload.map(async (approval) => {
          let tokenInfo: Omit<TokenInfo, 'name' | 'logoUri'> | undefined = balances.items.find(
            (item) => item.tokenInfo.address === approval.tokenAddress,
          )?.tokenInfo
          if (!tokenInfo) {
            tokenInfo = await getERC20TokenInfoOnChain(approval.tokenAddress)
          }

          const amountFormatted = UNLIMITED_APPROVAL_AMOUNT.eq(approval.amount)
            ? PSEUDO_APPROVAL_VALUES.UNLIMITED
            : ethers.utils.formatUnits(approval.amount, tokenInfo?.decimals)

          return { ...approval, tokenInfo: tokenInfo, amountFormatted }
        }),
      )
    },
    [balances.items.length, approvals],
    false, // Do not clear data on balance updates
  )

  return [approvalInfos ?? [], error, hasApprovalSignatures && loading]
}
```

This would only switch to `loading = true` if there are approval signatures present to avoid a flickering approval editor skeleton.

usame-algan: Good idea! I've changed the hook so that it always returns `[]` if there are no approvals.

schmanu (approved):

PR #2268, opened 2023-07-11
[TX flow] Batch execute fixes

usame-algan:
## What it solves

Part of #2067 

## How this PR fixes it

- Hide the status widget for batch executions
- More visible warning message
- Close the modal when submitting the transaction

## How to test it

1. Open a Safe with at least 2 fully signed transactions
2. Go to the Queue
3. Press Batch Execute
4. Observe an orange warning box at the bottom
5. Observe no Status widget
6. Switch to mobile view and observe no status widget toggle button
7. Execute transaction
8. Observe the modal closes

## Screenshots

Screenshot 2023-07-11 at 11 51 57

Screenshot 2023-07-11 at 11 52 03

## Checklist

### DIFF
--- a/src/components/tx-flow/common/TxLayout/index.tsx
+++ b/src/components/tx-flow/common/TxLayout/index.tsx
@@ -24,6 +24,7 @@ type TxLayoutProps = {
   txSummary?: TransactionSummary
   onBack?: () => void
   hideNonce?: boolean
+  hideStatus?: boolean
   isReplacement?: boolean
 }
 
@@ -36,6 +37,7 @@ const TxLayout = ({
   txSummary,
   onBack,
   hideNonce = false,
+  hideStatus = false,
   isReplacement = false,
 }: TxLayoutProps): ReactElement => {
   const [statusVisible, setStatusVisible] = useState<boolean>(true)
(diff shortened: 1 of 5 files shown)


usame-algan:
> All looks good but I'm not sure about entirely hiding the status widget. Perhaps we should just hide the irrelevant steps. What do you think?

I was also thinking about that but the only relevant step is "Execute" which we already tell the user in the subtitle by saying "Execute Batch".

iamacook:
> I was also thinking about that but the only relevant step is "Execute" which we already tell the user in the subtitle by saying "Execute Batch".

It seems odd without the widget as it doesn't "match" the flow. We could still show the "Execute" step. cc @liliiaorlenko (see screenshot in description)

liliiaorlenko:
> > I was also thinking about that but the only relevant step is "Execute" which we already tell the user in the subtitle by saying "Execute Batch".
> 
> It seems odd without the widget as it doesn't "match" the flow. We could still show the "Execute" step. cc @liliiaorlenko (see screenshot in description)

yes I would still keep the execute step for sure, it makes sense in overall structure of transaction flow
maybe instead of "Execute batch" I would just use Batch

usame-algan:
> yes I would still keep the execute step for sure, it makes sense in overall structure of transaction flow maybe instead of "Execute batch" I would just use Batch

I've updated the status widget messages (see PR description).

iamacook (approved):

PR #2270, opened 2023-07-11
fix: increase `min-width` of nonce field

iamacook:
## What it solves

Increase `min-width` of nonce field

## How this PR fixes it

The "base" `min-width` of the nonce field has been increased to 4 characters as per design request.

## How to test it

Create a transaction and observe the increased `min-width` of the nonce field.

## Screenshots

nonce

## Checklist


usame-algan (approved):
I don't mind if the input is a bit wider but since we have the space now, is it possible to not make it jump when the reset button appears? I believe its already absolutely positioned but somehow there is still a jump.

PR #2271, opened 2023-07-11
fix: remove unnecessary hook

iamacook:
## What it solves

Resolves unnecessary hook

## How this PR fixes it

The `TxStatusWidget` no longer requires `useDarkMode` to set the colour of the badge. Instead it now uses the `data-theme` attribute.

## How to test it

Create a transaction and switch between the light and dark mode. Observe the badge colour changing correctly.

## Checklist


usame-algan (approved):
💡 Makes a lot of sense!

PR #2275, opened 2023-07-11
fix: Align decoded tx ui for batch execute

usame-algan:
## What it solves

Part of #2067 

## How this PR fixes it

- Adds Expand/Collapse all buttons to batch execute
- Renames Batch execute submit button to "Submit"
- Doesn't show the confirm dialog when batch executing

## Other fixes

- Updates the spending limit components to use `SendAmountBlock` instead of `AmountBlock` since it supports a custom title now and is more in line with the rest of the layout

## How to test it

1. Open a Safe
2. Queue and confirm at least 2 transactions
3. Press Execute Batch
4. Observe the decoded tx ui with buttons

## Screenshots

Screenshot 2023-07-11 at 17 04 50

## Checklist


src/components/tx-flow/flows/ExecuteBatch/DecodedTxs.tsx
```diff
@@ -4,51 +4,69 @@ import useSafeInfo from '@/hooks/useSafeInfo'
...
-            txData={transaction.txData}
-            actionTitle={`${idx + 1}`}
-            showDelegateCallWarning={false}
-          />
-        )
-      })}
-    </Box>
+    <>
+      <MultisendActionsHeader title="Batched transactions" setOpen={setOpenMap} amount={txs.length} compact />
+
```
usame-algan, line 21: There is some duplication between this and the `Multisend` component because of the data type difference but I suggest we refactor it as a future task.

PR #2276, opened 2023-07-12
refactor: use wallet RPC when available

### LINKED ISSUE #2000: Investigate always using wallet provider when connected
## What is the feature about

We are currently using a mixture of our "read-only" provider and that of the wallet (when it is connected). This can result in two sources of truth if they have different RPCs.

We should investigate how other DApps approach this and whether we should _only_ use that of the wallet if it is connected.

## The list of requirements

- Investigate how other DApps approach this
- Look into where we use which provider for what calls
- Perhaps adjust [provider "creation"](https://github.com/safe-global/safe-wallet-web/blob/dev/src/hooks/wallets/web3.ts#L17-L30) and usage
- Update SDK usage accordingly

## Hypthesised provider

1. Is no wallet connected or UI is on different chain from wallet? -> Use "read-only" provider
2. Is wallet connected and UI on the chain of the wallet? -> Use waller provider

Note: If a wallet provider RPC wasn't working, we would need to fallback to the "read-only" one/notify the user of this otherwise the UI would not work.

## Links

- [Comment](https://github.com/5afe/safe-support/issues/189#issuecomment-1543434882) which raised discussion
- [Discussion](https://5afe.slack.com/archives/C03DAGWJCR5/p1684329486507859)

iamacook:
## What it solves

Resolves #2000 

## How this PR fixes it

Whenever a wallet is connected to the current chain, the provider of the wallet is used. Otherwise a read-only HTTP-provider (or the Safe Apps read-only provider) is.

## How to test it

Note: this will likely need a full regression.

Observe no noticeable differences apart from the same RPC endpoint being used in accordance with the above.

## Checklist


src/hooks/wallets/useProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!isReadOnly) {
+      return createWeb3(wallet.provider)
+    }
+
+    if (!isSafeApp) {
+      return createWeb3ReadOnly(chain.rpcUri, customRpc)
+    } else {
+      return createSafeAppsWeb3Provider(chain.safeAppsRpcUri, customRpc)
+    }
+  }, [chain, wallet, isSafeApp, customRpc])
```
iamacook, line 33: This is the main logic.

src/hooks/useIsValidExecution.ts
```diff
@@ -22,14 +24,11 @@ const isContractError = (error: EthersError) => {
 
 // Monkey patch the signerProvider to proxy requests to the "readonly" provider if on the wrong chain
 // This is ONLY used to check the validity of a transaction in `useIsValidExecution`
-const getPatchedSignerProvider = (
-  wallet: ConnectedWallet,
-  chainId: SafeInfo['chainId'],
-  readOnlyProvider: JsonRpcProvider,
-) => {
+const getPatchedSignerProvider = (wallet: ConnectedWallet, chain: ChainInfo, customRpc?: Record<string, string>) => {
```
iamacook, line 27: We might be able to remove this. What do you think @usame-algan?

src/hooks/wallets/useProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  return useMemo(() => {
+    if (!chain) {
+      return
+    }
+
+    const isReadOnly = !wallet || wallet.chainId !== chain.chainId
+
+    if (!isReadOnly) {
+      return createWeb3(wallet.provider)
```
schmanu, line 25: As most users probably have a wallet connected I would assume that then most of the time the wallet provider will be used.
I would like to understand what was the user story for our custom RPC provider in the Safe Settings as this PR would circumvent this customRpc most of the time now. If privacy reasons / broken rpcs in some wallets were the reason we should consider giving the customRPC higher priority if readonly methods are needed.

src/hooks/wallets/useProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+import { createSafeAppsWeb3Provider, createWeb3, createWeb3ReadOnly } from '@/hooks/wallets/web3'
+import { useAppSelector } from '@/store'
+import { selectRpc } from '@/store/settingsSlice'
+
+export const useProvider = (isSafeApp = false): JsonRpcProvider | Web3Provider | undefined => {
+  const chain = useCurrentChain()
+  const wallet = useWallet()
+  const rpcMap = useAppSelector(selectRpc)
+
+  const customRpc = chain ? rpcMap[chain.chainId] : undefined
```
usame-algan, line 15: This can be moved inside the `useMemo` and doesn't need to check for the existance of `chain` anymore.

src/hooks/useIsValidExecution.ts
usame-algan: True, we should test if the validity check still works when in the tx flow while connected to a different network if we use `useWeb3` instead of this function.

src/hooks/wallets/useProvider.ts
usame-algan: As far as I understand, with this refactor we now only have one provider at all times be it read-only or not so there shouldn't be any possible desync issues anymore. I agree that for the custom rpc setting it is not clear how it will affect the usage so maybe we should clarify that as you suggested i.e. if the wallet is not connected or if its not connected to the current network.

I think having the custom rpc option is still useful because safe apps can potentially have a lot of rpc requests which would go through the default rpc otherwise.

src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -77,7 +77,7 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
   }
 
   const createSafe = async () => {
-    if (!wallet || !provider || !chain) return
+    if (!wallet || !provider || isWeb3ReadOnly(provider) || !chain) return
```
usame-algan, line 80: I saw that we only use this condition here. How do we handle this case for normal tx executions i.e. how do we make sure that its not a read-only provider before pressing submit? Maybe we should try to apply the same logic to keep it consistent.

src/hooks/useIsValidExecution.ts
iamacook: It looks like we can't, sadly. The contract instance requires a non-"read only" provider but it's possible that it can be one.

src/components/new-safe/create/steps/ReviewStep/index.tsx
iamacook: > I saw that we only use this condition here

It's used in two other places: Safe creation retry and getting the MultiSend contract to submit a batch.

> How do we handle this case for normal tx executions i.e. how do we make sure that its not a read-only provider before pressing submit? Maybe we should try to apply the same logic to keep it consistent.

It's not necessary to include everywhere as you cannot submit when the wallet isn't connected. [Our `CheckWallet` component handles that](https://github.com/safe-global/safe-wallet-web/blob/consistent-provider/src/components/common/CheckWallet/index.tsx#L26-L27).

src/hooks/wallets/useProvider.ts
iamacook: > I think having the custom rpc option is still useful because safe apps can potentially have a lot of rpc requests which would go through the default rpc otherwise.

This would not be the case. If the wallet is connected to the current chain, the wallet RPC would be used.

usame-algan: > This would not be the case. If the wallet is connected to the current chain, the wallet RPC would be used.

Is this change intended? Previously, we always used a read-only provider inside the app communicator with its own infura token I assume to keep it separate from rpc requests of the main app.

src/components/new-safe/create/steps/ReviewStep/index.tsx
usame-algan: Should we add the CheckWallet component to the submit buttons then instead of having this check?

src/hooks/wallets/useProvider.ts
iamacook: > Is this change intended? Previously, we always used a read-only provider inside the app communicator with its own infura token I assume to keep it separate from rpc requests of the main app.

Yes, otherwise there will different RPCs being used.

src/components/new-safe/create/steps/ReviewStep/index.tsx
iamacook: > Should we add the CheckWallet component to the submit buttons then instead of having this check?

A close equivalent is being used. The creation cannot happen if the wallet [is on the wrong chain](https://github.com/safe-global/safe-wallet-web/blob/trim-token-amount/src/components/new-safe/create/steps/ReviewStep/index.tsx#L200).

In this context, this would always be needed as it is a type guard for [`computeNewSafeAddress`](https://github.com/safe-global/safe-wallet-web/blob/dev/src/components/new-safe/create/logic/index.ts#L75).

katspaugh:
> Whenever a wallet is connected to the current chain,

What if the wallet is connected to the wrong chain? We currently allow that and switch only when signing/executing. How does it work in this PR?

> Otherwise a read only (Safe Apps) provider is.

Why Safe Apps?

> This means that the same RPC endpoint will always be used.

What do you mean by "the same RCP" if it's still using a read-only one?

katspaugh:
Will this PR allow us to get rid of the required INFURA_KEY in the env vars?

iamacook:
> Will this PR allow us to get rid of the required INFURA_KEY in the env vars?

No, the "read only" provider relies on Infura RPCs on [certain chains](https://safe-client.safe.global/v1/chains) (`rpcUri`/`safeAppsRpcUri`).

katspaugh:
What does this PR solve then?

src/hooks/wallets/useProvider.ts
usame-algan: As per our discussion, a side-effect of this change is that we will most likely see decreased usage of the safe apps infura token which makes sense.

usame-algan (approved):

francovenica:
I've changed the RPC in the site to "cloudflare", and I'm connected with MM which has the "infura", so I was expecting not see any cloudflare calls but reloading the page the onboard modal shows up a lot of cloudflare calls are made. 
Could it be because after refreshing the page technically there is no wallet connected? 
image

francovenica:
Another instance, here while going through the send funds form. The app has "Cloudflare" changed in its settings, the connected wallet has "Infura"
RPC_2

iamacook:
> I've changed the RPC in the site to "cloudflare", and I'm connected with MM which has the "infura", so I was expecting not see any cloudflare calls but reloading the page the onboard modal shows up a lot of cloudflare calls are made. Could it be because after refreshing the page technically there is no wallet connected for a second until the onboard modal shows up?

Need to discuss this with you. From what I can tell, this is because onboard is initialised with the chain configuration, e.g. _before_ the wallet is connected.

I think we'd need to reinitialise onboard with the RPC of the wallet but then trust that it reconnects to the wallet. Ideally, this should be done in a separate PR.

francovenica:
@iamacook This is the RPC I'm using in ETH mainnet: https://cloudflare-eth.com/

iamacook:
@francovenica, I found the root of the issue and fixed it in https://github.com/safe-global/safe-wallet-web/pull/2276/commits/6c2a33a31a8e37334a4d777884459cfd34c67095. We were still using read only instances of contracts which have now been removed.

It is another sizeable change so I will request review of this again.

src/components/settings/EnvironmentVariables/EnvHintButton/index.tsx
```diff
@@ -7,13 +7,15 @@ import { isEnvInitialState } from '@/store/settingsSlice'
...
+import useWallet from '@/hooks/wallets/useWallet'
 
 const EnvHintButton = () => {
   const router = useRouter()
   const chainId = useChainId()
+  const wallet = useWallet()
   const isInitialState = useAppSelector((state) => isEnvInitialState(state, chainId))
 
-  if (isInitialState) {
+  if (isInitialState || wallet) {
```
schmanu, line 18: Why do we not show any hints if a wallet is connected?
If I now for instance change my tenderly simulation URL the hint will disappear as soon as I connect a wallet.

src/components/tx-flow/flows/SignMessageOnChain/ReviewSignMessageOnChain.tsx
```diff
@@ -38,15 +39,19 @@ const ReviewSignMessageOnChain = ({ message, method, requestId }: SignMessageOnC
...
   useHighlightHiddenTab()
 
   const isTextMessage = method === Methods.signMessage && typeof message === 'string'
   const isTypedMessage = method === Methods.signTypedMessage && isObjectEIP712TypedData(message)
 
-  const readOnlySignMessageLibContract = useMemo(() => getReadOnlySignMessageLibContract(chainId), [chainId])
-  const signMessageAddress = readOnlySignMessageLibContract.getAddress()
+  const signMessageLibContract = useMemo(() => {
+    if (web3) {
+      return getSignMessageLibContract(chainId, web3)
```
schmanu, line 52: We should think about adding a getInterface / getDeployment function which only uses (abi + address) as this would be enough here and in many other components. Then we would not require `web3` at all. Only the `chainId` and Safe version.

schmanu (commented):
I tested it and it works well.

I really think we should add a way to just get a Contract's interface / address to encode function calls and get the contract address. In 95% where we call getXYZContract we only need to encode some data but all components that need to encode data now need a web3 provider + undefined checks.
I am fine with doing that in a future iteration though as it requires changes on the core-sdk side.

src/components/tx-flow/flows/SignMessageOnChain/ReviewSignMessageOnChain.tsx
iamacook: I saw this pattern emerge whilst removing the read only instances and totally agree. We could use the deployment getters from `safe-deployments` for this.

I would suggest this is done in a separate PR though so as to not increase the scope of this: the getters return contract instances or `undefined`. We'd need to `return` early or `throw` to ensure type safety. What do you think?

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -315,12 +316,13 @@ export const dispatchTxRelay = async (
   safeTx: SafeTransaction,
   safe: SafeInfo,
   txId: string,
+  provider: JsonRpcProvider | Web3Provider,
```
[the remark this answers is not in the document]
iamacook: I agree, see https://github.com/safe-global/safe-wallet-web/pull/2276#discussion_r1288060030.

src/components/settings/EnvironmentVariables/EnvHintButton/index.tsx
iamacook: Good point. I've adjusted the logic to preserve the Tenderly warning.

iamacook: After discussion, I've reverted this. We can revisit it later if necessary.

schmanu (approved):

src/components/tx-flow/flows/SignMessageOnChain/ReviewSignMessageOnChain.tsx
usame-algan: I think that was the reason we added those readOnly instances of the contracts to not have a dependency to the provider. Are those contracts responsible for the read only RPC requests that @francovenica found?

iamacook: > I think that was the reason we added those readOnly instances of the contracts to not have a dependency to the provider.

This I'm not sure about, but we should certainly pivot to using the deployments repo imo.

> Are those contracts responsible for the read only RPC requests that @francovenica found?

They were indeed.

francovenica:
I'm still getting Read-only RPC calls while connected

When I enter into a tx that is ready to be executed the RPC call is done. So in order to reproduce this you must first queue a tx, have it ready to be executed and enter into the form to executed
If you try to create a tx that has not been proposed yet (like in a 1/1 safe proposal + execution at the same time) the call IS NOT made.

In this gif, the wallet has an "infura" RPC  and the safe has set the "llamarpc", and this one is called there. Is called as soon as I open the modal and every 15 secs if you stay there
RPC calls

katspaugh:
I’m closing this pull request because it’s not and never was in any sprint. Let’s not waste time.

PR #2278, opened 2023-07-12
fix: Hide nonce for spending limit txs

usame-algan:
## What it solves

Part of #2067 

## How this PR fixes it

- Adds an additional field `nonceNeeded` to the `SafeTxContext` so the nonce can be hidden from inside flow components
- Closes the flow when submitting a spending limit tx

## How to test it

1. Open a Safe with a spending limit
2. Create a new transaction
3. Switch between spending limit and normal transaction
4. Observe the nonce field is disappearing
5. Observe the status widget is changing text
6. Submit the transaction
7. Observe the modal is closing

## Screenshots

Screenshot 2023-07-12 at 12 24 10
Screenshot 2023-07-12 at 12 24 18

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
+++ b/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
@@ -1,5 +1,5 @@
 import type { ReactElement, SyntheticEvent } from 'react'
-import { useMemo, useState } from 'react'
+import { useContext, useMemo, useState } from 'react'
 import type { BigNumberish, BytesLike } from 'ethers'
 import { Button, CardActions, Typography } from '@mui/material'
 import SendToBlock from '@/components/tx-flow/flows/TokenTransfer/SendToBlock'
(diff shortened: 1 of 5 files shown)


usame-algan:
Imo its not ideal that we need another flag `nonceNeeded` in the `SafeTxContext` now. Ideally we should decide for a given flow if a nonce is needed or not. This is an edge case because we essentially have 2 flows combined (normal transactions and spending limit transactions). Lets revisit the discussion around splitting up the flows for the next iteration cc @katspaugh

iamacook (approved):

src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
```diff
@@ -88,7 +90,7 @@ const ReviewSpendingLimitTx = ({
 
     try {
       await dispatchSpendingLimitTxExecution(txParams, txOptions, onboard, safe.chainId, safeAddress)
-
+      setTxFlow(undefined)
       onSubmit()
```
iamacook, line 94: We should call `onSubmit` before unmount.

```suggestion
      onSubmit()
      setTxFlow(undefined)
```

PR #2284, opened 2023-07-13
fix: trim/remove leading zeros from values

### LINKED ISSUE #2281: Spending limits - Error show when blank spaces are in the amount input
## Bug description
When creating a spending limit, if there are blank spaces in front of the number an error "Amount too small" is shown. In other amount inputs (like send funds) blank characters are ignored, the same should happen for spending limit creation

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Any
 - Env: Current prod v1.14 and new RC v1.15

## Steps to reproduce
 1. Go to the Spending limit and start the creation of one
 2. Pick any token and enter any amount
 3. Add after the amount at least 1 blank space

## Expected result
No errors should pop up

## Obtained result
An error saying amount too small

## Screenshots
image

iamacook:
## What it solves

Resolves #2281

## How this PR fixes it

The value of the `NumberField` component now trims and removes leading zeros.

## How to test it

The aforementioned component is used in:

- Token amount field
- `safeTxGas` field
- Transaction filter: amount, nonce fields
- Advanced params: wallet nonce, max prio fee, max fee/gas price fields
- Approval editor: value field
- Transaction nonce field

## Checklist

### DIFF
--- a/src/components/common/TokenAmountInput/index.tsx
+++ b/src/components/common/TokenAmountInput/index.tsx
@@ -67,6 +67,10 @@ const TokenAmountInput = ({
           {...register(TokenAmountFields.amount, {
             required: true,
             validate: validate ?? validateAmount,
+            onChange: (event) => {
+              event.target.value = event.target.value.trim()
+              return event
+            },
           })}
         />
         <Divider orientation="vertical" flexItem />
--- a/src/components/common/TokenAmountInput/index.tsx
+++ b/src/components/common/TokenAmountInput/index.tsx
@@ -67,10 +67,6 @@ const TokenAmountInput = ({
           {...register(TokenAmountFields.amount, {
             required: true,
             validate: validate ?? validateAmount,
-            onChange: (event) => {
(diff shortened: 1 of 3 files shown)


src/components/common/TokenAmountInput/index.tsx
```diff
@@ -67,6 +67,10 @@ const TokenAmountInput = ({
           {...register(TokenAmountFields.amount, {
             required: true,
             validate: validate ?? validateAmount,
+            onChange: (event) => {
+              event.target.value = event.target.value.trim()
+              return event
+            },
```
usame-algan, line 73: As an additional thing, should we also remove leading zeroes?

usame-algan (approved):

src/components/common/TokenAmountInput/index.tsx
iamacook: Great idea! I've added the logic to the `NumberField` directly.

francovenica:
LGTM

PR #2290, opened 2023-07-13
fix: Show Tenderly fail if there is a call trace error

usame-algan:
## What it solves

Fixes for #2067 

## How this PR fixes it

- Checks if there is a call trace error for the tenderly message
- Adds back the padding for decoded txs if its not compact

## How to test it

1. Open a Safe and simulate a transaction that has a call trace error
2. Observe the button showing Fail instead of Success

## Screenshots
Screenshot 2023-07-13 at 16 49 29
Screenshot 2023-07-13 at 16 49 53

## Checklist

### DIFF
--- a/src/components/tx/security/tenderly/index.tsx
+++ b/src/components/tx/security/tenderly/index.tsx
@@ -43,12 +43,14 @@ const TxSimulationBlock = ({ transactions, disabled, gasLimit }: TxSimulationPro
   const isDarkMode = useDarkMode()
   const { safeTx } = useContext(SafeTxContext)
   const {
-    simulation: { simulateTransaction, simulationRequestStatus, resetSimulation },
+    simulation: { simulation, simulateTransaction, simulationRequestStatus, resetSimulation },
   } = useContext(TxInfoContext)
 
   const isLoading = simulationRequestStatus === FETCH_STATUS.LOADING
   const isSuccess = simulationRequestStatus === FETCH_STATUS.SUCCESS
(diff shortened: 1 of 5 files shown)


src/components/tx/security/tenderly/index.tsx
```diff
@@ -43,13 +43,15 @@ const TxSimulationBlock = ({ transactions, disabled, gasLimit }: TxSimulationPro
...
   const {
-    simulation: { simulateTransaction, simulationRequestStatus, resetSimulation },
+    simulation: { simulation, simulateTransaction, simulationRequestStatus, resetSimulation },
   } = useContext(TxInfoContext)
 
   const isLoading = simulationRequestStatus === FETCH_STATUS.LOADING
   const isSuccess = simulationRequestStatus === FETCH_STATUS.SUCCESS
   const isError = simulationRequestStatus === FETCH_STATUS.ERROR
 
+  const isCallTraceError = isSuccess && getCallTraceErrors(simulation).length > 0
```
iamacook, line 53: The `simulation.status` [also needs to be taken into account](https://github.com/safe-global/safe-wallet-web/pull/2206/files).

For some reason merging the epic didn't include this fix.

iamacook (changes requested):

src/components/tx/security/tenderly/index.tsx
usame-algan: Good find! We also extracted the status logic into the context

iamacook (approved):

PR #2291, opened 2023-07-13
fix: `useAsync` dependency arrays

### LINKED ISSUE #2061: Add exhaustive dependency check for `useAsync` hook
## What is this about

We don't have any eslint rules in place to check dependencies when the `useAsync` hook is being used. This can be fixed by extending the eslint rules with:

```
"react-hooks/exhaustive-deps": ["warn", {
    "additionalHooks": "useAsync"
}]
```

Extending this rule will result in a few warnings that need to be addressed.

iamacook:
## What it solves

Resolves #2061

## How this PR fixes it

This adds the exhaustive dependency array linter to our `useAsync` hook. All relative findings have been adjusted accordingly.

Dependency arrays that needed changing were found in:

1. `DecodedTx`
2. `useRedefine`
3. `useGasPrice`
4. `useIsValidExecution`
5. `useWalletCanRelay`

## How to test it

According to the above list:

1. Observe the transaction info displaying in the sign/execute form (above the balance change).
2. Redefine works.
3. Gas estimation works in Safe creation.
4. Invalid executions are warned about in the execution form ("This transaction will most likely fail...").
5. It is _not_ possible to relay 6 times in an hour.

## Checklist


schmanu (approved):
A very necessary change!
Good work :)

This could have some side effects though, which may be hard to find. Our current release is already quite big. I'd therefore want to discuss if we should include it in the following release. WDYT?

iamacook:
> This could have some side effects though, which may be hard to find. Our current release is already quite big. I'd therefore want to discuss if we should include it in the following release. WDYT?

I only adjusted the dependencies of five hooks which are listed in the description but I totally agree. Let's leave it until the next release and see if QA find any issues until then.

src/hooks/loadables/useLoadSpendingLimits.ts
```diff
@@ -105,7 +105,7 @@ export const useLoadSpendingLimits = (): AsyncResult<SpendingLimitState[]> => {
     if (!provider || !safeLoaded || !safe.modules || !tokenInfoFromBalances) return
 
     return getSpendingLimits(provider, safe.modules, safeAddress, chainId, tokenInfoFromBalances)
-  }, [provider, safeLoaded, safe.modules?.length, safeAddress, chainId, safe.txHistoryTag, tokenInfoFromBalances])
```
usame-algan, line 108: Are we sure the `txHistoryTag` needs to be removed here? I remember we wanted to automatically update the value on the settings page without reloading

francovenica:
LGTM 

Tried few tx checking the details
Created several safes, the initial estimation of gas worked fine every time
Redefine works fine
Trying to execute tx that will fail shows the warning properly
Relaying only works 5 times per hour

src/hooks/loadables/useLoadSpendingLimits.ts
iamacook: True, I understand the requirement. I've added it back.

usame-algan (approved):

PR #2280, opened 2023-07-12
Release/1.15.0

usame-algan:
## Features

- #2067
- #2307 

## Fixes

- #2228
- #2238
- #2232
- #2255
- #2267
- #2273

PR #2294, opened 2023-07-14
fix: Reopen QR modal in safe creation

usame-algan:
## What it solves

Fix for #2280 

## How this PR fixes it

- Only runs the validation if the value has changed

## How to test it

1. Open Safe creation
2. Add an owner
3. Press the QR button and select an address
4. Press it again and observe it opens again

## Checklist

### DIFF
--- a/src/components/common/AddressInput/index.tsx
+++ b/src/components/common/AddressInput/index.tsx
@@ -50,6 +50,7 @@ const AddressInput = ({
     formState: { errors },
     trigger,
   } = useFormContext()
+
   const currentChain = useCurrentChain()
   const rawValueRef = useRef<string>('')
   const watchedValue = useWatch({ name, control })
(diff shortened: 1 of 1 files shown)


iamacook (approved):
There's a test that needs to be fixed but otherwise looks good to me!

usame-algan:
Looks like the validation now doesn't run anymore in certain cases. Will have to take a look at this again. Ideally we can find a way to not trigger the validation when pressing the QR Code button.

francovenica:
> Looks like the validation now doesn't run anymore in certain cases. Will have to take a look at this again. Ideally we can find a way to not trigger the validation when pressing the QR Code button.

Yeah, In the address book test it seems that is failing becaus the "Save" button becomes clickable as soon as you write anything, and not when a valid address is in the input.
In the load safe is failing because a random string of text is not triggering the error in the input

usame-algan:
I've switched to using RHF native `isValidating` flag to solve this issue.

iamacook (approved):

PR #2295, opened 2023-07-14
feat: pending safes per chainId

### LINKED ISSUE #2292: [Safe creation] Store pending state per chain and let users abort a stuck creation
## What is the feature about
Currently we store the pending safe creation state globally.
If a safe creation is stuck on some chain the user is blocked to create a Safe on another chain.

Additionally sometimes the safe creation is pending for a long time or the pending state does not get cleared properly. Then users are blocked from Safe creation until they empty their local storage.
There should be a way to abandon a currently pending Safe creation.

## The list of requirements
- [ ] Store pending Safes per chainId
- [ ] Offer a way to abandon a stuck Safe creation (without having to wait for the timeout)

## Designs/sketches
tbd

schmanu:
## What it solves

Resolves #2292 

## How this PR fixes it
- Stores pending Safe per chainId

## How to test it
- Create a Safe on Görli but do not wait for the tx to process
- Switch to Polygon 
- Create a Safe on polygon in parallel

## TODO
- [x] fix and write unit-test
- [x] refresh state when switching chain on Status screen

## Checklist

### DIFF
--- a/src/components/new-safe/create/__tests__/useSyncSafeCreationStep.test.ts
+++ b/src/components/new-safe/create/__tests__/useSyncSafeCreationStep.test.ts
@@ -3,14 +3,27 @@ import useSyncSafeCreationStep from '@/components/new-safe/create/useSyncSafeCre
 import * as wallet from '@/hooks/wallets/useWallet'
 import * as localStorage from '@/services/local-storage/useLocalStorage'
 import type { ConnectedWallet } from '@/services/onboard'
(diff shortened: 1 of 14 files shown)


katspaugh:
@schmanu is this ready for review?

schmanu:
@katspaugh No not yet.
I need to fix up the tests and write a new test for this. Switching chains while being in Safe creation also ends up in a very weird state.

iamacook (approved):

francovenica:
LGTM

PR #2297, opened 2023-07-21
Update links pointing to the Safe documentation

germartinez:
## What it solves
Fixes some broken links pointing to the Safe docs

## How this PR fixes it
Replaces the links with new ones

## How to test it
Check the new are links are the right ones and they work properly


katspaugh (approved):
Thank you sir

katspaugh:
I've tested that the links work.

PR #2298, opened 2023-07-21
Static page for redirects regarding web3auth

biafra23:
## What it solves

More often than not universal links meant for the mobile app end up in requests to the universal link itself. We want to show a button that, when clicked, redirect the oauth token to the mobile app, 

Resolves #

## How this PR fixes it

creating a page at web3auth

## Checklist


biafra23:
Not needed with url schemes

PR #2299, opened 2023-07-21
Feat: import/export in the Safe List sidebar

katspaugh:
## What it solves

We received user feedback that the Import/Export feature isn't discoverable. It's only shown on the Welcome page and deep in the Settings atm.

The sidebar listing your Safe Accounts seems like an intuitive other place for this feature.

## Todo
* [x] Adjust design as per [Figma](https://www.figma.com/file/HKW5iarMST9rL7Kxkf59qY/MUI-for-Safe?type=design&node-id=11426%3A67673&mode=design&t=qkoZAcOtQA0PP6hV-1)

## Screenshots

Screenshot 2023-07-21 at 18 38 22


francovenica:
The import button is not accesible if the list of owned safes is too long. 
Import sidebar

francovenica:
Looks Good.

The buttons are visible now. 
They work the same as the ones from the homepage
Tracking is also working (same events as  the ones from the homepage)

Note: in the chrome mobile emulator the buttons cannot be seen (the scroll doesn't reach) but in my iPhone it works fine so my assumption is that the emulator is not accurate simulating the mobile device for this case, so I'll skip it.

PR #2317, opened 2023-07-27
Fix: minimize queue drawer when tx modal is opened

katspaugh:
Setting a high z-index on the modal was a mistake because ot was covering the Safe List.

Instead, it will now collapse the bottom drawer when a tx modal is opened.


iamacook (approved):

francovenica:
LGTM

PR #2322, opened 2023-07-28
Fix: warn if tx was already signed by connected wallet

### LINKED ISSUE #2321: [Tx flow] The same owner can sign tx a few times
## Bug description
The same owner can sign tx a few times
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Goerli

## Steps to reproduce
 1. Create tx as an owner 1
 2. switch MM account from the owner 1 to the owner 2 
 3. click Sign created tx
 4. switch MM back to the owner 1
 Current result: The submit button is available, and the same owner can click the sign again 

## Expected result
The sign button should be disabled on the Tx modal if the owner already signed the tx
## Screenshots
https://drive.google.com/file/d/1tbTvTBJ2ILOllDQIhIdyIa4Bo3sLmoNf/view?usp=sharing

katspaugh:
## What it solves

Resolves #2321

Technically, nothing bad will happen if you sign a tx with the same owner twice. It will just overwrite the first signature.
But still good to warn about that.

However, I would keep the Sign button enabled because there can be a situation when a tx is signed locally, but sending it to the backend failed. In this case signing again is helpful.

In addition to that, some small stylistic adjustments in the tx flow.

## How this PR fixes it

Shows a warning is the tx has already been signed by the connect wallet.
Screenshot 2023-07-28 at 10 49 06

### DIFF
--- a/src/components/tx-flow/common/TxLayout/styles.module.css
+++ b/src/components/tx-flow/common/TxLayout/styles.module.css
@@ -1,5 +1,5 @@
 .container {
-  margin-top: 42px;
+  margin-top: 10px;
 }
 
 .header {
--- a/src/components/tx-flow/flows/SuccessScreen/index.tsx
+++ b/src/components/tx-flow/flows/SuccessScreen/index.tsx
@@ -1,26 +1,24 @@
-import { useRouter } from 'next/router'
 import StatusMessage from './StatusMessage'
 import StatusStepper from './StatusStepper'
-import { AppRoutes } from '@/config/routes'
 import { Button, Container, Divider, Paper } from '@mui/material'
 import classnames from 'classnames'
-import Link from 'next/link'
-import { type UrlObject } from 'url'
 import css from './styles.module.css'
 import { useAppSelector } from '@/store'
 import { selectPendingTxById } from '@/store/pendingTxsSlice'
-import { useEffect, useState } from 'react'
+import { useCallback, useContext, useEffect, useState } from 'react'
 import { getBlockExplorerLink } from '@/utils/chains'
 import { useCurrentChain } from '@/hooks/useChains'
 import { TxEvent, txSubscribe } from '@/services/tx/txEvents'
+import { TxModalContext } from '../..'
 
 export const SuccessScreen = ({ txId }: { txId: string }) => {
   const [localTxHash, setLocalTxHash] = useState<string>()
   const [error, setError] = useState<Error>()
-  const router = useRouter()
-  const chain = useCurrentChain()
+  const { setTxFlow } = useContext(TxModalContext)
   const pendingTx = useAppSelector((state) => selectPendingTxById(state, txId))
   const { txHash = '', status } = pendingTx || {}
+  const chain = useCurrentChain()
+  const txLink = chain && localTxHash ? getBlockExplorerLink(chain, localTxHash) : undefined
 
   useEffect(() => {
     if (!txHash) return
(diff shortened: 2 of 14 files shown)


src/components/tx/SignOrExecuteForm/SignForm.tsx
```diff
@@ -64,6 +65,8 @@ const SignForm = ({
 
   return (
     <form onSubmit={handleSubmit}>
+      {hasSigned && <ErrorMessage level="warning">You have already signed this transaction.</ErrorMessage>}
```
schmanu, line 68: Should we not move it into the `CheckWallet` component and add a new message there?

schmanu (commented):
I think there is no use-case in letting users double sign a tx. AFAIK we do not change the safe transaction object locally when signing. So it should only pickup the new signature when successfully proposed and updated in the queue.

katspaugh:
> AFAIK we do not change the safe transaction object locally when signing

We do, it's a destructive operation in the SDK. It mutates the signatures map.

src/components/tx/SignOrExecuteForm/SignForm.tsx
katspaugh: CheckWallet doesn't know anything about transactions, it only checks the wallet itself.

schmanu:
> We do, it's a destructive operation in the SDK. It mutates the signatures map.

Oh nevermind my comment then. I'll approve :)

schmanu (approved):

src/components/tx-flow/common/TxLayout/styles.module.css
```diff
@@ -1,5 +1,5 @@
 .container {
-  margin-top: 42px;
+  margin-top: 10px;
```
iamacook, line 2: The modal alignment no longer matches:

tx-modal-jump

src/components/tx-flow/flows/SuccessScreen/index.tsx
```diff
@@ -71,8 +71,8 @@ export const SuccessScreen = ({ txId }: { txId: string }) => {
       <Divider />
       <div className={classnames(css.row, css.buttons)}>
         <Link href={homeLink} passHref>
-          <Button variant="outlined" size="small">
-            Back to dashboard
```
iamacook, line 75: I'd sooner leave this as is. We lose information changing this.

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -179,3 +179,10 @@ export const useSafeTxGas = (safeTx: SafeTransaction | undefined): number | unde
 
   return safeTxGas
 }
+
+export const useAlreadySigned = (safeTx: SafeTransaction | undefined): boolean => {
+  const wallet = useWallet()
+  const hasSigned =
+    safeTx && wallet && (safeTx.signatures.has(wallet.address.toLowerCase()) || safeTx.signatures.has(wallet.address))
```
iamacook, line 186: What do you think about using `sameAddress` here, e.g.

```suggestion
    safeTx && wallet && Array.from(safeTx.signatures.values()).some((address) => sameAddress(address, wallet.address))
```

src/components/tx-flow/flows/SuccessScreen/index.tsx
katspaugh: What information? "Back to dashboard" is plain incorrect in most cases. It doesn't go to dashboard, it just closes the modal.

iamacook: >What information? "Back to dashboard" is plain incorrect in most cases. It doesn't go to dashboard, it just closes the modal.

We lose the context of where it navigates to - [the dashboard](https://github.com/safe-global/safe-wallet-web/blob/can-sign/src/components/tx-flow/flows/SuccessScreen/index.tsx#L41-L44).

katspaugh: Ah, I see. I'll remove the link and will make it close the modal.

francovenica:
Not a big fan of allowing the user to sign again, but if is not harmful and we all agree is fine, then is ok for me.
LGTM

PR #2323, opened 2023-07-28
Fix: always keep the nonce of existing txs

katspaugh:
## What it solves

Solves a bug with delegated txs:
* A delegate address creates a tx with nonce N
* An actual owner address signs it
* Instead of signing the tx N, it creates another tx with a nonce N + 1

## How this PR fixes it

The tx confirmation flow will now always use the original transaction nonce and won't update it to the recommended nonce.

## How to test it

* Create an unsigned tx using MetaMask institutional or a delegate address.
* Sign it with an owner key
* Observe that it signed the original tx and not created a new tx


liliya-soroka:
The fix for the unsigned txs is verified. The same nonce is suggested during signing as was used for the tx creation

liliya-soroka:
Quick check was done for :
1. reject tx 
2. replace tx
3. create new tx ( send token, send NFT, safe apps tx)
4. spending limit tx( module tx)
5. recommended nonce option

PR #2325, opened 2023-07-28
Release 1.15.2

katspaugh:
* Fix: delegate-created txs should have a fixed nonce (#2323)


francovenica:
Lily checked the unsigned tx in the original 2323 ticket.
I checked the rest of regular tx:

App tx
Send funds
Replace tx /Reject
Spending limit creation
Spending limit use (no nonce, but still)
Owner edition.

Looks good

PR #2341, opened 2023-08-01
Fix: re-connecting a locked wallet

### LINKED ISSUE #2306: [MM] Reconnecting to last wallet fails and onboard will not show up
## Environment
* Chrome
* MetaMask extension

## Bug
When you have previously connected with MM, sometimes it would not reconnect automatically, but also won't show the onboard popup.
So you cannot connect basically, the Connect button does nothing.
Goes away after reopening the browser.

katspaugh:
## What it solves

Resolves #2306

## How this PR fixes it

The cause of the bug was that
* `window.ethereum.isConnected()` returns true even if MetaMask is locked
* We didn't await the promise from `_metamask.isUnlocked` and it always returning false

I've removed `isConnected` and made it wait for the promise. Also added custom logic for WC v2 and tests for it.

### DIFF
--- a/src/utils/wallets.ts
+++ b/src/utils/wallets.ts
@@ -37,8 +37,14 @@ export const isWalletUnlocked = async (walletName: string): Promise<boolean> => 
     return (await window.ethereum?._metamask?.isUnlocked?.()) || false
   }
 
-  // Wallet connect creates a localStorage entry when connected and removes it when disconnected
-  if (walletName === WalletNames.WALLET_CONNECT || walletName === WalletNames.WALLET_CONNECT_V2) {
+  // WalletConnect v2
+  if (walletName === WalletNames.WALLET_CONNECT_V2) {
+    const wcSession = window.localStorage.getItem('wc@2:client:0.3//session')
+    return !!wcSession && wcSession !== '[]'
+  }
+
+  // WalletConnect v1 creates a localStorage entry when connected and removes it when disconnected
+  if (walletName === WalletNames.WALLET_CONNECT) {
     return window.localStorage.getItem('walletconnect') !== null
   }
 
--- a/src/utils/wallets.ts
+++ b/src/utils/wallets.ts
@@ -32,17 +32,13 @@ export const WalletNames = {
 export const isWalletUnlocked = async (walletName: string): Promise<boolean> => {
   if (typeof window === 'undefined') return false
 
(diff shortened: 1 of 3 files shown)


src/utils/wallets.ts
```diff
@@ -32,16 +32,18 @@ export const WalletNames = {
...
-
   // Only MetaMask exposes a method to check if the wallet is unlocked
   if (walletName === WalletNames.METAMASK) {
-    return window.ethereum?._metamask?.isUnlocked?.() || false
+    return (await window.ethereum?._metamask?.isUnlocked?.()) || false
+  }
+
+  // WalletConnect v2
+  if (walletName === WalletNames.WALLET_CONNECT_V2) {
+    const wcSession = window.localStorage.getItem('wc@2:client:0.3//session')
```
iamacook, line 42: It looks like the `0.3` is [versioning](https://github.com/WalletConnect/walletconnect-monorepo/blob/06ca6bbf5aa26e869a4e1bfd83b5cca38eae7960/packages/core/src/controllers/store.ts#L74C1-L74C1).

You can call [`storageKey`](https://github.com/WalletConnect/walletconnect-monorepo/blob/06ca6bbf5aa26e869a4e1bfd83b5cca38eae7960/packages/core/src/controllers/store.ts#L73) (and maybe even [`get`](https://github.com/WalletConnect/walletconnect-monorepo/blob/06ca6bbf5aa26e869a4e1bfd83b5cca38eae7960/packages/core/src/controllers/store.ts#L143)) from WC [here](https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/sign-client/src/controllers/session.ts#L9) to avoid outdated session data.

katspaugh: I'm aware of the trade-offs here and would sooner check local storage directly rather than call some internal WC methods. They can be made more robust by searching for a key named `wc@2:client:*//session` perhaps.

iamacook: > I'm aware of the trade-offs here and would sooner check local storage directly rather than call some internal WC method.

`storageKey` would return the key with which you could check the `localStorage` directly.

> This can be made more robust by searching for a key named `wc@2:client:*//session` perhaps.

I'd personally prefer this.

katspaugh:
I'll close this PR to focus on the onboard's auto-connect solution.

Issue #2342, opened 2023-08-01
WalletConnect v2 tracking


katspaugh:
WalletConnect-connected wallet apps are tracked only for WC v1 atm, but not for v2.

<img width="831" alt="Screenshot 2023-08-01 at 13 37 32" src="https://github.com/safe-global/safe-wallet-web/assets/381895/9e2bec40-0c23-4acf-9148-6a7245356484">

katspaugh:
Like this:
<img src="https://media1.giphy.com/media/xT5LMqUQBL44YD02Jy/giphy.gif" />

PR #2343, opened 2023-08-01
Refactor: onboard auto connect

### LINKED ISSUE #2306: [MM] Reconnecting to last wallet fails and onboard will not show up

katspaugh:
## What it solves

Resolves #2306 and #2342.

Onboard has its own auto-connect-to-last-wallet feature. This PR removes our old custom logic and uses that.

## How to test this
 * Make sure the same wallet re-connects when reloading the page
 * Try locking MetaMask manually and it shouldn't re-connect or even popup MetaMask after a page reload
Screenshot 2023-08-03 at 13 46 47

 * Check that we still track wallet connects (both manual and automatic reconnections)
 * Check that we now properly track which wallet is used with WalletConnect


iamacook (changes requested):
Works nicely but the automatically connected wallet is not being tracked anymore. I'd suggest we move `trackWalletType` somewhere else.

katspaugh:
@iamacook fixed, plus added WC v2 peer tracking.

src/components/common/NavTabs/index.tsx
```diff
@@ -32,7 +32,7 @@ const NextLinkComposed = forwardRef<HTMLAnchorElement, Props>(function NextCompo
 
 const NavTabs = ({ tabs }: { tabs: NavItem[] }) => {
   const router = useRouter()
-  const activeTab = tabs.map((tab) => tab.href).indexOf(router.pathname)
+  const activeTab = Math.max(0, tabs.map((tab) => tab.href).indexOf(router.pathname))
```
iamacook, line 35: Why did we need to change this? If the user is not on one of the `tabs`, I don't think we should default to one of them.

src/hooks/wallets/useOnboard.ts
```diff
@@ -67,19 +60,29 @@ export const getConnectedWallet = (wallets: WalletState[]): ConnectedWallet | nu
...
+  const peerWalletV1 = connector.peerMeta?.name
 
-  return peerWallet ?? UNKNOWN_PEER
+  return peerWalletV2 || peerWalletV1 || UNKNOWN_PEER
 }
 
+let lastTracked = ''
 const trackWalletType = (wallet: ConnectedWallet) => {
+  if (wallet.label === lastTracked) return
+  lastTracked = wallet.label
```
iamacook, line 84: If the user disconnects and reconnects the same wallet, this will not track.

We could reset `lastTracked` on line 193 in an `else`. What do you think?

src/components/common/NavTabs/index.tsx
katspaugh: Because it was throwing lots of errors about a -1 tab.

iamacook: Ok, makes sense then.

src/hooks/wallets/useOnboard.ts
katspaugh: Done.

iamacook (approved):
There's a lint issue.

francovenica:
What works:
The app reconects fine after refreshing the page, the popUp doesn't show up anymore (except for physical devices, but that's expected)
The wallet that reconects does track with GA, both manually and automatically
Disconnecting the wallet (or locking MM) and refreshing the page won't try to reconect it.

Issue:
Regarding WC, I've seen that only the new version of WC shows the wallet, but the older one just shows WallectConectv1.
If this is on purpose because v1 will be deprecated soon let me know  
WC v2:

image

WC v1:
image

katspaugh:
> Regarding WC, I've seen that only the new version of WC shows the wallet, but the older one just shows WallectConectv1.

Fixed, thank you. ✅

francovenica:
Works fine now for WC v1:
image
image

PR #2353, opened 2023-08-06
Refactor: fetch pending queue only in 1/N Safes

katspaugh:
## What it solves

There's no need to fetch the untrusted queue in non-1/X Safes. In higher threshold Safes, all pending txs will be signed or delegated, so they'll be displayed in the main queue.

## How to test

1.
* Go to a 2+/X Safe
* Make and execute a tx
* Open the Network panel in the devtools
* Quickly go to the queue and observe the pending tx
* Observe that the title above the queue says "Next" (correct)
* Observe that NO `/queued?trusted=false` request is made

2.
* Go to a 1/X Safe
* Make and execute a tx
* Open the Network panel in the devtools
* Quickly go to the queue and observe the pending tx
* Observe that the title above the queue says "Pending"
* Observe that there IS a `/queued?trusted=false` request made


iamacook (commented):
With this change we lose the "Pending" label:

pending

katspaugh:
> With this change we lose the "Pending" label:

This label has never been, and should not be, displayed for non-1/X Safes.
Due to a bug in CGW, the pending queue has been erroneously displayed for all Safes since the new CGW release, along with this label, but it's not correct.

iamacook (approved):

francovenica:
I was going to comment about the "Next" label for tx in safes 2/N, but given the previous comment I won't

I still see the queue?trusted=false
In safes 1/x:
queue call

In safes 2/x they also show, but they do when the tx dissapears from the queue tab after being successfully executed (Not during signing, or executing, just as it dissapears from the queue tab)
image

katspaugh:
I've tested again and haven't seen untrusted requests in 2/X Safes. I suggest we merge this and test in the RC again.

PR #2354, opened 2023-08-07
Fix: tx can have been created w/o signatures

katspaugh:
## What it solves

Delegate txs can be unsigned but still be in the queue, so the definition of `isCreation` should be the absence of a tx id.

## How to test

* Create a transaction as a delegate
* Confirm the transaction as an owner
* Observe that it no longer says "Create" anywhere in the confirmation tx flow like it did before

### DIFF
--- a/src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
+++ b/src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
@@ -8,7 +8,7 @@ import { useCurrentChain } from '@/hooks/useChains'
 import { getTxOptions } from '@/utils/transactions'
 import useIsValidExecution from '@/hooks/useIsValidExecution'
 import CheckWallet from '@/components/common/CheckWallet'
-import { useImmediatelyExecutable, useIsExecutionLoop, useTxActions } from './hooks'
+import { useIsExecutionLoop, useTxActions } from './hooks'
 import { useRelaysBySafe } from '@/hooks/useRemainingRelays'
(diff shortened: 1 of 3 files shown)


usame-algan (commented):
We do the same check in `ExecuteForm`. I suggest to change it there as well.

src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
```diff
@@ -50,8 +51,6 @@ const ExecuteForm = ({
   const { needsRiskConfirmation, isRiskConfirmed, setIsRiskIgnored } = useContext(TxSecurityContext)
 
   // Check that the transaction is executable
-  const isCreation = safeTx?.signatures.size === 0
-  const isNewExecutableTx = useImmediatelyExecutable() && isCreation
```
usame-algan, line 54: I think it makes sense to not check for immediate execution here because if its a creation and its possible to execute it has to be an immediate execution so only checking the `isCreation` flag should be enough.

It would be nice to extract the error displaying logic into its own "dumb" component so we could test these cases though. I suggest to add a TODO.

usame-algan (approved):

src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
katspaugh: Feel free to create a tech debt ticket, I'd like to avoid polluting the code with TODO comments.

francovenica:
We can merge this one. it will be tested during regression

PR #2357, opened 2023-08-07
Fix: sidebar links to queue if there are queued txs

### LINKED ISSUE #2352: Transactions link: open History if Queue is empty
## What is the feature about
Currently the Transactions link in the sidebar always opens the Tx History.

It would be more useful to open the Queue if it’s not empty, because it’s more actionable for the user.

(A request from the Yearn team)

katspaugh:
## What it solves

Resolves #2352

## How this PR fixes it

The `/transactions` route will now redirect to either `/transactions/history` or `/transactions/queue` if the queue is not empty.


usame-algan:
Have we considered how we can measure the success of this change? I think it could be confusing if the same button has 2 different behaviours without the user knowing i.e. expecting to go to the history but ending up in the queue and vice versa especially if both pages look very similar it could slow the user down more.

src/pages/transactions/index.tsx
```diff
@@ -1,3 +1,25 @@
...
+  const hasQueuedTxs = !!page?.results.length
+
+  useEffect(() => {
+    const pathname = hasQueuedTxs ? AppRoutes.transactions.queue : AppRoutes.transactions.history
+
+    router.push({
+      pathname,
+      query: { safe: router.query.safe },
+    })
+  }, [router, hasQueuedTxs])
```
iamacook, line 20: Can this be achieved in the `return`? Currently, if the user is on the history page and a transaction is added to the queue, they will be redirected without user input.

katspaugh: They will be redirected only once to queue or history, and this effect won't run on those pages.

schmanu (approved):
Should we add a small unit test, that checks that this rerouting will trigger?
Otherwise it's one of the small features that might get lost in a refactor.

Other than that this looks good and I'll already approve :)

francovenica:
The main feature works fine, the redirection to queue when there is a tx there works fine, to the history tab otherwise. 

There is an issue with GA tho. Now visiting the tx history or queue triggers the /transaction/history or /queue event, but also there is an event for just "/transaction".
In Stg env only the /transaction/history or /queue shows:

transaction tracking

katspaugh:
@francovenica I've made the sidebar link itself change to either queue or history, so there won't be a `/transaction` event anymore.

francovenica:
Yeah, only 1 call for /history or /queue. 

Looks good to me.

PR #2358, opened 2023-08-07
Fix: prevent an early redirect on the Safe Apps share page

### LINKED ISSUE #2320: [Safe Apps] Shared URL redirects to dashboard
Sometimes a shared app URL like https://app.safe.global/share/safe-app?appUrl=https%3A%2F%2Fswap.cow.fi&chain=eth redirects to the dashboard.

I'm consistently seeing this with Safari, but not in Chrome. Looks like some race condition because Chrome seems to load the page faster.

katspaugh:
## What it solves

Resolves #2320

## How this PR fixes it

It should wait for the client-side router to kick in before attempting a redirect.

## How to test

Open this link:

https://fix_app_share--walletweb.review-wallet-web.5afe.dev/share/safe-app?appUrl=https%3A%2F%2Fsafe-apps.dev.5afe.dev%2Fdrain-safe&chain=gor

It should not redirect to the Dashboard.

### DIFF
--- a/src/pages/share/safe-app.tsx
+++ b/src/pages/share/safe-app.tsx
@@ -19,7 +19,7 @@ const ShareSafeApp = () => {
     )
   }
 
-  if (!appUrl || !validChain || !chain) {
+  if (router.isReady && (!appUrl || !validChain || !chain)) {
     router.push(AppRoutes.index)
     return null
   }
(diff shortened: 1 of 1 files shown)


iamacook (commented):
`!appUrl || !validChain || !chain` can be moved to a variable as it's used twice.

iamacook (approved):

PR #2362, opened 2023-08-08
fix: Use PNG instead of SVG for Redefine logo

### LINKED ISSUE #2314: [Bundle size] Redefine logos are 300kb
They aren't real SVGs. They are bitmaps inserted into an SVG. The should be PNGs. And we shouldn't bundle them with the app.

Screenshot 2023-07-26 at 21 18 45

usame-algan:
## What it solves

Resolves #2314 

## How this PR fixes it

- Uses PNG instead of SVG

## How to test it

1. Create a transaction on mainnet
2. Observe the Redefine logo looking the same as before in both dark and light mode

## Screenshots
Screenshot 2023-08-08 at 10 17 32
Screenshot 2023-08-08 at 10 17 39

## Checklist


katspaugh (commented):
You can make those pngs 50% smaller if you run them thru tinypng.com

src/components/tx/security/redefine/index.tsx
```diff
@@ -11,14 +11,15 @@ import { ErrorBoundary } from '@sentry/react'
 import { REDEFINE_ARTICLE, REDEFINE_SIMULATION_URL } from '@/config/constants'
 import css from 'src/components/tx/security/redefine/styles.module.css'
 import sharedCss from '@/components/tx/security/shared/styles.module.css'
-import RedefineLogoDark from '@/public/images/transactions/redefine-dark-mode.svg'
-import RedefineLogo from '@/public/images/transactions/redefine.svg'
+import RedefineLogoDark from '@/public/images/transactions/redefine-dark-mode.png'
+import RedefineLogo from '@/public/images/transactions/redefine.png'
```
katspaugh, line 15: Are they imported as URLs in this case?

src/components/tx/security/redefine/index.tsx
```diff
@@ -81,11 +82,9 @@ const RedefineBlock = () => {
 
           <Typography variant="caption" className={sharedCss.poweredBy} position="relative">
             Powered by{' '}
-            <SvgIcon
-              inheritViewBox
-              sx={{ height: '40px', width: '52px', position: 'absolute', right: '-58px' }}
-              component={isDarkMode ? RedefineLogoDark : RedefineLogo}
-            />
+            <div className={css.redefine}>
+              <Image src={isDarkMode ? RedefineLogoDark : RedefineLogo} alt="Redefine Logo" />
```
katspaugh, line 86:
```suggestion
              <Image src={isDarkMode ? RedefineLogoDark : RedefineLogo} alt="Redefine logo" />
```

src/components/tx/security/redefine/index.tsx
usame-algan: Good question, I think they might be part of the js bundle in this case but will double check. Otherwise we can import them like ``

usame-algan: Looks like they are not part of any js chunk when using `next/image`. The URLs are added to the sw.js though I assume for offline support

katspaugh (approved):

PR #2364, opened 2023-08-08
Tests: run e2e tests only on staging

katspaugh:
## What it solves

Some of the smoke tests were running with a production gateway URL because staging was not indexed properly at some point. I've adjust them to only run on staging.


francovenica (approved):
LGTM

francovenica:
I think the safe app test that are failing were fixed here https://github.com/safe-global/safe-wallet-web/pull/2367

PR #2365, opened 2023-08-08
fix: minimum fee of 0.1 gwei on optimism

### LINKED ISSUE #2222: EIP-1559 is not enabled on Optimism and gas price estimation is leading to struck txs
## Bug description
Optimism Team is reporting that user transactions are stuck due to low gas prices. While checking it, we have noticed that EIP-1559 is not enabled on Optimism yet. 

Doc:
https://community.optimism.io/docs/developers/bedrock/wallet-upgrade/#transaction-fees

Context of the issue:
We are seeing increased reports of users having transactions stuck in mempool from the base fee being set too low.
It is currently the case that the OP Mainnet base fee can spike significantly faster than the L1 Ethereum Base fee.
Key Recommendation:
Please set a lower bound default of 0.1 gwei for the max base fee set for users’s transactions
This will ensure that 98% of the time, a user’s transaction will be guaranteed to be included in the next block

FAQ:
Will this mean users pay more than they should?
This means users can pay up to .1 gwei base fee, but currently, 98% of the time the base fee will be lower(95% of the time, users will pay below .01 gwei)
this is in the worst case still <$0.01 paid for a token transfer
In addition, please increase the multiplier added on top of the max base fee higher than what you would set on L1 Ethereum
Because the rate of base fee increase can be so much higher on OP Mainnet, we recommend having much higher multiples set for the max base fee to adjust for fee spikes:
Some initial recommendations are: slow 1.0x / medium, 2.0x  / fast 10.0x, but we would recommend monitoring for user complaints and consider increase these multipliers further if users are still anticipating txs stuck in the mempool.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to

schmanu:
## What it solves

Resolves #2222 

## How this PR fixes it
Uses a minimum gas price of 0.1 gwei.

## How to test it
- Create different txs on Optimism:
  - Safe creation
  - Safe transactions
  - Batch txs
  - Spending limit txs

Ideally we should also enable EIP-1559 after this PR gets merged. I could only test it locally by overwriting my local state.

## Checklist

### DIFF
--- a/src/hooks/useGasPrice.ts
+++ b/src/hooks/useGasPrice.ts
@@ -8,6 +8,12 @@ import { useWeb3ReadOnly } from '../hooks/wallets/web3'
 import { Errors, logError } from '@/services/exceptions'
 import { FEATURES, hasFeature } from '@/utils/chains'
 import { asError } from '@/services/exceptions/utils'
+import { adjustGasEstimationForChain } from '@/utils/gas'
+
+export type EstimatedGasPrice = {
+  maxFeePerGas: BigNumber | undefined
+  maxPriorityFeePerGas: BigNumber | undefined
+}
 
 // Update gas fees every 20 seconds
 const REFRESH_DELAY = 20e3
@@ -53,10 +59,7 @@ const getGasPrice = async (gasPriceConfigs: GasPrice): Promise<BigNumber | undef
   }
 }
 
-const useGasPrice = (): AsyncResult<{
-  maxFeePerGas: BigNumber | undefined
-  maxPriorityFeePerGas: BigNumber | undefined
-}> => {
+const useGasPrice = (): AsyncResult<EstimatedGasPrice> => {
   const chain = useCurrentChain()
   const gasPriceConfigs = chain?.gasPrice
   const [counter] = useIntervalCounter(REFRESH_DELAY)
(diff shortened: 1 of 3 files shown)


src/utils/gas.ts
```diff
@@ -0,0 +1,19 @@
...
+export const adjustGasEstimationForChain = (gas: EstimatedGasPrice, chainId: string | undefined) => {
+  if (chainId === chains.oeth) {
+    // On Optimism we lower bound the base fee to 0.1 gwei because the base fee can spike quite fast
+    return {
+      ...gas,
+      maxFeePerGas: gas.maxFeePerGas?.lt(OPTIMISM_MINIMUM_BASE_FEE_WEI)
+        ? OPTIMISM_MINIMUM_BASE_FEE_WEI
+        : gas.maxFeePerGas,
+    }
+  }
```
katspaugh, line 16: That's hard to read a bit, can you make maxFeePerGas a separate variable above?

src/hooks/useGasPrice.ts
```diff
@@ -77,10 +80,12 @@ const useGasPrice = (): AsyncResult<{
       const maxFee = gasPrice || (isEIP1559 ? feeData?.maxFeePerGas : feeData?.gasPrice) || undefined
       const maxPrioFee = (isEIP1559 && feeData?.maxPriorityFeePerGas) || undefined
 
-      return {
+      const gasEstimation = {
         maxFeePerGas: maxFee,
         maxPriorityFeePerGas: maxPrioFee,
       }
+
+      return adjustGasEstimationForChain(gasEstimation, chain?.chainId)
```
katspaugh, line 88: Naming-wise, I would call this `gasPriceEstimation` to not confuse it with the gas limit estimation.

katspaugh (approved):

francovenica:
Looks good to me

Made sure that tx have a Gas price of 0.1 every time.
I've checked Safe creation, Send funds, Owner adding, Tx rejection, tx batching (the old Excecute batch and the new batch feature as well) 
The user can still edit the amount and that one is passed correctly to the wallet

I tested with MM and Trezor, since sometimes physical devices ignore the estimation we give and use their own, but it works fine this time, the 0.1 Gwei is passed correctly

Note: there is no spending limits on Optimism, is not enabled there

katspaugh:
Apparently we should be using the same logic on Base as well.

schmanu:
Settings a fixed gas price for EIP1559 is currently not possible.
We should add this feature to our gas settings in the config service.
Closing this issue in favor of that.

PR #2367, opened 2023-08-08
Tests: fix tx modal e2e test

usame-algan:
## What it solves

Fixes the tx-modal e2e test

## How this PR fixes it

- The safe app fixture immediately sent a transaction to the communicator which sometimes made the test fail because the manifest wasn't fully fetched yet. This adds a small delay there
- Also removes the `useSafeAppFromManifest` hook inside `AppFrame` since we already use the hook in the parent so we can pass it as a prop instead

## How to test it

- Safe apps should still work
- The tx-modal e2e test should not fail anymore

## Checklist

### DIFF
--- a/cypress/fixtures/safe-app.html
+++ b/cypress/fixtures/safe-app.html
@@ -32,10 +32,13 @@
         const path = window.location.pathname
         switch (path.split('/')[1]) {
           case 'dummy':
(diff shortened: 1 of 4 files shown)


iamacook (approved):

francovenica:
The Safe app test are not failing in the PR anymore, so it is ok.

PR #2372, opened 2023-08-08
fix: better UX for off-chain message signing

### LINKED ISSUE #2019: [EIP1271] Offchain signing UI tweaks
* Signatures from the same browser should be registered immedialy (currently there's a big lag even if it's from the connected wallet)
Screenshot 2023-05-22 at 13 01 42

* `To sign this message, you need to collect 2 owner signatures of your Safe.` -> `To sign this message, collect signatures from 2 owners of your Safe`
* Remove this message if the threshold is 1 (`To sign this message, you need to collect 1 owner signatures of your Safe.`)
* Before closing the modal after the final signature comes in, add a 0.5-1s delay so that the Confirmations list is updated and the user has a chance to see it.
* Hide this in a Message details accordion (not so useful info to an average person):
Screenshot 2023-05-22 at 13 04 36

schmanu:
## What it solves

Resolves #2019 

## How this PR fixes it
- Changes text
- Adds a Success Card
- Shows Success Card for 1s when message is fully signed
- immediately re-fetches message after signing

## How to test it
- Open a 2/3 Safe
- Create a message signing request through Wallet connect
- Add the first signature
  - Observe that it gets added immediately
- Change the owner
- Add a second signature
  - Observe new success card with Continue button
  - It should disappear after 1s

## Screenshots
Screenshot 2023-08-08 at 16 48 51

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/SignMessage/SignMessage.test.tsx
+++ b/src/components/tx-flow/flows/SignMessage/SignMessage.test.tsx
@@ -1,5 +1,5 @@
 import { hexlify, hexZeroPad, toUtf8Bytes } from 'ethers/lib/utils'
-import type { ChainInfo, SafeInfo, SafeMessage } from '@safe-global/safe-gateway-typescript-sdk'
+import type { ChainInfo, SafeInfo, SafeMessage, SafeMessageListPage } from '@safe-global/safe-gateway-typescript-sdk'
 import { SafeMessageListItemType } from '@safe-global/safe-gateway-typescript-sdk'
 
 import SignMessage from './SignMessage'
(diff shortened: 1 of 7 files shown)


iamacook (approved):

iamacook (approved):

francovenica:
I've used this app to test this https://github.com/5afe/eip-1271-dapp

Have a safe with 2 owners at least and created a message. I open both owners in different browsers
Made sure that after the 1st owner signs in the form the confirmation shows shortly after (2 or 3 seconds takes for the confirmation to show up)
Signing with the other person in the other broswer takes a bit longer but it is expected
The form shows the success message for about a second and then redirects to the WC app again
The message of "Collect sigantures" was changed with the wording suggested in the main ticket and it's not visible for safes 1/x
The message was put into an accordeon block

With this the ticket is fine to be closed but Some suggestion

I think 1 second is too short, especially if other owners are signing in different browsers, if you blink you will miss it. Make it 3 sconds
Since we remove the message of "Collect signatures" for 1/x safes, I think this message should also change from
"Please keep this modal open until all signers confirm this message" to "Please keep this modal open until the message is signed"
image

Also I've noticed that for messages the right side "Tx status" is missplaced. I'll create a new ticket to deal with it
08-10-2023_1247x578(6338)

francovenica:
Here's the issue https://github.com/safe-global/safe-wallet-web/issues/2384

PR #2375, opened 2023-08-09
Feat: load analytics in cookie-less mode

katspaugh:
## What it solves

We want to make GA more useful by tracking 100% of users. GA4 has a so-called "consent mode" which doesn't set any cookies and doesn't require a cookie consent. This PR enables this mode, and enables GA cookies only if the user opts in.

See https://www.notion.so/safe-global/Client-side-analytics-improvements-152dad5585d6410b8e6cfb20334d6c05

## How to test it

* Open the branch preview site
* Observe that GTM events are fired w/o accepting the Cookie Banner
* Check that the GA script is loaded with `gcs: G100` parameter (which means cookies are disabled).
* Check that no _ga cookies are set
* Press "Accept all" in the cookie banner
* Observe that GA cookies are added
* Observe that GA requests now contain `gcs: G101` param (meaning cookies are enabled)

### DIFF
--- a/src/services/analytics/TagManager.ts
+++ b/src/services/analytics/TagManager.ts
@@ -2,8 +2,6 @@ import Cookies from 'js-cookie'
 
 import { IS_PRODUCTION } from '@/config/constants'
 
-type DataLayer = Record<string, unknown>
-
 export type TagManagerArgs = {
   // GTM id, e.g. GTM-000000
   gtmId: string
@@ -11,8 +9,6 @@ export type TagManagerArgs = {
   auth: string
   // GTM environment, e.g. env-00.
   preview: string
-  // Object that contains all of the information that you want to pass to GTM
-  dataLayer?: DataLayer
 }
 
 const DATA_LAYER_NAME = 'dataLayer'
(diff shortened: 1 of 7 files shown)


src/services/analytics/TagManager.ts
```diff
@@ -38,40 +34,58 @@ const TagManager = {
...
+      console.info('[GTM] -', data)
+    }
   },
+
   initialize: (args: TagManagerArgs) => {
-    if (TagManager.isInitialized()) {
-      return
+    window[DATA_LAYER_NAME] = window[DATA_LAYER_NAME] || []
+    // This function MUST be in `window`, otherwise GTM Consent Mode just doesn't work
+    ;(window as any).gtag = function () {
```
usame-algan, line 50: We can add gtag types: `yarn add -D @types/gtag.js` and add gtag in the `definitions.d.ts` so we don't have to cast here. Also noticed that we have a typo and missing import in the definitions file.

src/services/analytics/TagManager.ts
```diff
@@ -38,40 +34,58 @@ const TagManager = {
...
+      window[DATA_LAYER_NAME].push(arguments)
     }
```
usame-algan, line 52: I don't understand why we have to overwrite the whole `gtag` object with a custom function instead of checking that window.gtag exists and if it doesn't we return early.

src/services/analytics/useGtm.ts
```diff
@@ -34,13 +47,14 @@ const useGtm = () => {
   // Track page views – anononimized by default.
   // Sensitive info, like the safe address or tx id, is always in the query string, which we DO NOT track.
```
usame-algan, line 48: Should we also update this comment or are we still anonymizing these?

usame-algan (commented):
We can use gtag types by adding the package `yarn add -D @types/gtag.js` and updating the `definitions.d.ts` so we don't have to cast the window object to any. I Also noticed that we have a typo and missing import in the definitions file which we could fix.

The `@types/gtag.js` doesn't have all the consent attributes so I've opened a PR to update it accordingly: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66327

src/services/analytics/TagManager.ts
katspaugh: Why would it exist if nobody defines it except us?

src/services/analytics/useGtm.ts
katspaugh: We're still anonymizing these.

src/services/analytics/TagManager.ts
usame-algan: I would assume the gtag script we load attaches it to the window.

katspaugh: It's usually part of the code snippet because apparently it needs to exist before you load the gtm script. In our case it wasn't there. I don't know if we customized it or it was like this to begin with. Anyway, the gtag function wasn't there.

usame-algan: According to [this](https://developers.google.com/tag-platform/gtagjs/install) we need to define that function ourselves indeed. Can we add a reference link in our code in case this comes up again?

usame-algan (approved):
Works nicely for me. I tried with and without cookies set and the change is reflected in tagmanager. Have we checked how cookieless-mode will affect the data in GA i.e. creating funnels might have limited options?

`@types/gtag.js` has been released with the changes so we could also use that if we want to have better type support.

katspaugh:
I don't want to include a dependency for these types, as the type of the gtag function is pretty loose anyway, it basically takes any number of any arguments.

katspaugh:
@francovenica we'd like to release this as a separate release sometime next week. Are you OK with that timeline?

francovenica:
Note: I've checked that the track is there with no cookies, but I haven't checked every action that triggers a tracking event in the entire app, I think that's out of the scope. I'll check during the next regression
Note2: I can only see the calls, but I'd like to see if we can see actually the tracking being saved in GA or wherever it goes when the Analytics checkbox is not enabled.


I've checke the GTM calls in the console when no cookies have been enabled.

I've checked the G100 and G101 params Depending on if the tracker is enabled or not by cookies:
image
image

I've checked that the _ga cookies show up/go away dedpending on if the tracker is enabled or not. I noticed some _ga[NUMBERS] cookies there, but those also show up the stg env when the analytics are disabled, so I assume is expected for those to be there.
image
image

katspaugh:
No, no cookies should be there when analytics is not enabled.

francovenica:
Then I'll kick it back

katspaugh:
@francovenica _ga_XXX cookies are now removed as well ✅

PR #2382, opened 2023-08-10
Fix: clear batch on confirm

katspaugh:
## What it solves

When a batch is confirmed, it should be removed. It didn't work because I was passing the wrong id.

## How to test it

* Create a batch
* Confirm it
* It should be proposed to the queue (or executed)
* It should disappear from the batch sidebar


schmanu (approved):
Approve to get the fix out

But let's add a test for the BatchSidebar logic soon.

PR #2383, opened 2023-08-10
fix: reset nonce if outdated

schmanu:
## What it solves
If the recommended nonce and safe nonce updates while the transaction modal is open, the current nonce is not updated and transactions with too low nonces are created which lead to validation errors when proposed.

## How this PR fixes it
Resets nonce to recommended nonce if the currently selected nonce is lower than the Safe nonce.

## How to test it
1. Open a 1/n Safe with nothing in the queue
2. Initiate any transaction, note the used nonce and immediately execute it (without signing it first)
3. While the transaction is executing, initiate another transaction
4. Observe that this new transaction can initially not be executed and the nonce being the same as the pending transaction
5. Do nothing and wait in the transaction modal until the first transaction got executed and indexed.
6. Observe that the new transaction's nonce gets updated to the next nonce and that the execute button is now available for this transaction.

## Checklist

### DIFF
--- a/src/components/tx-flow/SafeTxProvider.tsx
+++ b/src/components/tx-flow/SafeTxProvider.tsx
@@ -4,6 +4,7 @@ import type { SafeTransaction } from '@safe-global/safe-core-sdk-types'
 import { createTx } from '@/services/tx/tx-sender'
 import { useRecommendedNonce, useSafeTxGas } from '../tx/SignOrExecuteForm/hooks'
 import { Errors, logError } from '@/services/exceptions'
+import useSafeInfo from '@/hooks/useSafeInfo'
 
 export const SafeTxContext = createContext<{
   safeTx?: SafeTransaction
@@ -36,6 +37,8 @@ const SafeTxProvider = ({ children }: { children: ReactNode }): ReactElement => 
   const [nonceNeeded, setNonceNeeded] = useState<boolean>(true)
   const [safeTxGas, setSafeTxGas] = useState<number>()
 
+  const { safe } = useSafeInfo()
+
   // Signed txs cannot be updated
   const isSigned = safeTx && safeTx.signatures.size > 0
 
@@ -43,8 +46,9 @@ const SafeTxProvider = ({ children }: { children: ReactNode }): ReactElement => 
   const recommendedNonce = useRecommendedNonce()
   const recommendedSafeTxGas = useSafeTxGas(safeTx)
 
-  // Priority to external nonce, then to the recommended one
-  const finalNonce = isSigned ? safeTx?.data.nonce : nonce ?? recommendedNonce ?? safeTx?.data.nonce
+  // Priority to external nonce if valid, then to the recommended one
+  const isNonceValid = nonce !== undefined && nonce >= safe.nonce
+  const finalNonce = isSigned ? safeTx?.data.nonce : isNonceValid ? nonce : recommendedNonce ?? safeTx?.data.nonce
   const finalSafeTxGas = isSigned ? safeTx?.data.safeTxGas : safeTxGas ?? recommendedSafeTxGas ?? safeTx?.data.safeTxGas
 
   // Update the tx when the nonce or safeTxGas change
--- a/src/components/tx-flow/SafeTxProvider.tsx
+++ b/src/components/tx-flow/SafeTxProvider.tsx
@@ -4,6 +4,7 @@ import type { SafeTransaction } from '@safe-global/safe-core-sdk-types'
 import { createTx } from '@/services/tx/tx-sender'
 import { useRecommendedNonce, useSafeTxGas } from '../tx/SignOrExecuteForm/hooks'
 import { Errors, logError } from '@/services/exceptions'
+import useSafeInfo from '@/hooks/useSafeInfo'
 
 export const SafeTxContext = createContext<{
   safeTx?: SafeTransaction
@@ -35,15 +36,16 @@ const SafeTxProvider = ({ children }: { children: ReactNode }): ReactElement => 
   const [nonce, setNonce] = useState<number>()
   const [nonceNeeded, setNonceNeeded] = useState<boolean>(true)
   const [safeTxGas, setSafeTxGas] = useState<number>()
+  const { safe } = useSafeInfo()
 
   // Signed txs cannot be updated
   const isSigned = safeTx && safeTx.signatures.size > 0
 
   // Recommended nonce and safeTxGas
-  const recommendedNonce = useRecommendedNonce()
+  const recommendedNonce = Math.max(safe.nonce, useRecommendedNonce() ?? 0)
   const recommendedSafeTxGas = useSafeTxGas(safeTx)
 
-  // Priority to external nonce then to the recommended one
+  // Priority to external nonce, then to the recommended one
   const finalNonce = isSigned ? safeTx?.data.nonce : nonce ?? recommendedNonce ?? safeTx?.data.nonce
   const finalSafeTxGas = isSigned ? safeTx?.data.safeTxGas : safeTxGas ?? recommendedSafeTxGas ?? safeTx?.data.safeTxGas
 
--- a/src/components/tx-flow/SafeTxProvider.tsx
+++ b/src/components/tx-flow/SafeTxProvider.tsx
@@ -47,8 +47,7 @@ const SafeTxProvider = ({ children }: { children: ReactNode }): ReactElement => 
   const recommendedSafeTxGas = useSafeTxGas(safeTx)
 
   // Priority to external nonce if valid, then to the recommended one
-  const isNonceValid = nonce !== undefined && nonce >= safe.nonce
-  const finalNonce = isSigned ? safeTx?.data.nonce : isNonceValid ? nonce : recommendedNonce ?? safeTx?.data.nonce
+  const finalNonce = isSigned ? safeTx?.data.nonce : nonce ?? recommendedNonce ?? safeTx?.data.nonce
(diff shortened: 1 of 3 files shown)


src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -109,6 +109,16 @@ const TxNonceForm = ({ nonce, recommendedNonce }: { nonce: string; recommendedNo
     formMethods.setValue(TxNonceFormFieldNames.NONCE, recommendedNonce)
   }
 
+  const currentFormNonce = formMethods.watch(TxNonceFormFieldNames.NONCE)
+  const isFormChanged = formMethods.formState.isDirty
+
+  // Update to recommended nonce if the current nonce is unchanged and invalid after update
+  useEffect(() => {
+    if (!isFormChanged && Number(currentFormNonce) < safe.nonce) {
```
katspaugh, line 117: Why are we checking the safe.nonce here and not the recommended nonce?

schmanu: This fix is mainly for the case when a user immediately executes, prepares a second tx while the first is indexing. Now with this change we will update the nonce to the next one once the indexer catches up. Previously we left an invalid nonce and did not show any error.

katspaugh: Yeah, but you're setting the input value to recommendedNonce, so you expect it to be updated. Then why not compare against that?

schmanu: True. I will change it to reset to the `safe.nonce` as that is updated / correct for sure. I think comparing to the `recommendedNonce` could sometimes overwrite a lower nonce for instance when replacing a transaction.

katspaugh: OK. Another question:

When you call setValue like this, what happens in the SafeTxProvider? Will it trigger a setNonce call? Which would be an equivalent of manually setting the nonce? Will this effectively stop it from being further updated should a new recommendedNonce arrive in the meanwhile?

src/components/tx-flow/SafeTxProvider.tsx
```diff
@@ -36,15 +37,18 @@ const SafeTxProvider = ({ children }: { children: ReactNode }): ReactElement =>
...
 
   // Recommended nonce and safeTxGas
   const recommendedNonce = useRecommendedNonce()
   const recommendedSafeTxGas = useSafeTxGas(safeTx)
 
-  // Priority to external nonce, then to the recommended one
-  const finalNonce = isSigned ? safeTx?.data.nonce : nonce ?? recommendedNonce ?? safeTx?.data.nonce
+  // Priority to external nonce if valid, then to the recommended one
+  const isNonceValid = nonce !== undefined && nonce >= safe.nonce
+  const finalNonce = isSigned ? safeTx?.data.nonce : isNonceValid ? nonce : recommendedNonce ?? safeTx?.data.nonce
```
katspaugh, line 51: `nonce` is the manually set nonce. It will likely not be there in the use case you're trying to fix.
You should be checking `recommendedNonce` for validity instead.

I would do something like

```js
const finalNonce = isSigned ? safeTx?.data.nonce : nonce : recommendedNonce >= safe.nonce ? recommendedNonce : safe.nonce
```

usame-algan: I think `nonce` is automatically set in the beginning to what `recommendedNonce` is. I think its a bit confusing to check against `safe.nonce` but then return `recommendedNonce` or vice versa. From my tests, `safe.nonce` updates a bit sooner than `recommendedNonce` but I am not sure why. We should try to understand how these values connect and only use one of them.

katspaugh: > I think nonce is automatically set in the beginning to what recommendedNonce is

It's not.

usame-algan: Sorry, I should've clarified. The Provider returns the `recommendedNonce` as the `finalNonce` and the form uses that as the `defaultValue`. This runs the form validation where we call `setNonce` which then updates `nonce` to what the  `recommendedNonce` is at the time.

katspaugh: I see. That's bad. `setNonce` should only be called if the user deliberately edited the nonce.

That's why the solution with `setValue` in this PR is no bueno.

schmanu: Not changing the form's field value but only triggering validation leads to weird states:
- Indeed no wrong / too low nonce will be used in the `SafeTxProvider` `finalNonce` (which is good)
- But: Our UI still shows that you use a wrong nonce.

Here is a screenshot showing a nonce error while using the actual updated `recommendedNonce` in the `SafeTxProvider` and thus enabling the sign button.
When signing a transaction the updated higher nonce is being used. 
Screenshot 2023-08-14 at 16 07 21

I think we form's nonce value should always be in sync with the `finalNonce` from the `SafeTxProvider` which will be used for the transaction.

I now changed the behavior slightly:
- Don't trigger form validation and `setNonce` for the `defaultValues` (so `nonce` is undefined in the SafeTxProvider when computing the `finalNonce`)
- When the `finalNonce` updates, we propagate the change to the TxNonce field through the `values` property of `useForm`:
```ts
  const formMethods = useForm({
    defaultValues: {
      [TxNonceFormFieldNames.NONCE]: nonce,
    },
    mode: 'onChange',
    values: {
      [TxNonceFormFieldNames.NONCE]: nonce,
    },
  })
```

Like this the validation and `finalNonce` computations happens in `SafeTxProvider` and we will never store an invalid nonce as finalNonce. Plus the valid `finalNonce` will always be displayed in the UI if it is updated.

katspaugh: The form code looks good now but you still need to fix the provider code. I would restore it to how it was and move the safe.nonce thing to useRecommendedNonce.

katspaugh:
@schmanu are you planning to finalize this PR?

schmanu:
@katspaugh 
I had to adjust it a little to get rid of the SafeTxProvider adjustment.
The mode onTouch makes it work. Then it setNonce only gets called once the user once touched / blurred the field and not initially.

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -164,7 +164,7 @@ export const useRecommendedNonce = (): number | undefined => {
       return getRecommendedNonce(safe.chainId, safeAddress)
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [safeAddress, safe.chainId, safe.txQueuedTag], // update when tx queue changes
+    [safeAddress, safe.chainId, safe.txQueuedTag, safe.nonce], // update when tx queue or safe nonce changes
```
katspaugh, line 167: This can be slightly optimized if we just return `Math.max(safe.nonce, recommendedNonce)` from the outer hook, no need to refetch the recommended nonce.

src/components/tx-flow/SafeTxProvider.tsx
```diff
@@ -43,7 +43,7 @@ const SafeTxProvider = ({ children }: { children: ReactNode }): ReactElement =>
   const recommendedNonce = useRecommendedNonce()
   const recommendedSafeTxGas = useSafeTxGas(safeTx)
 
-  // Priority to external nonce, then to the recommended one
```
katspaugh, line 46: A comma is actually needed here, isn't it?

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -164,7 +164,7 @@ export const useRecommendedNonce = (): number | undefined => {
       return getRecommendedNonce(safe.chainId, safeAddress)
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [safeAddress, safe.chainId, safe.txQueuedTag], // update when tx queue changes
+    [safeAddress, safe.chainId, safe.txQueuedTag],
```
katspaugh, line 167:
```suggestion
    [safeAddress, safe.chainId, safe.txQueuedTag], // update when tx queue changes
```

katspaugh (approved):

francovenica:
UPDATE:

It seems that is working fine, it just that sometimes it takes just a few seconds to updates and other 15 secs or so, but it updates fine so I'm ok with it
nonce update 2

katspaugh:
> The result is that, the input "understood" that the nonce had to be updated (the little round arrow shows up, indicating that you can update to the suggested nonce) but the nonce itself never changed

Sounds like an unwanted effect. The field probably got “updated” to the same value and became “dirty” which triggered the reset button.

francovenica:
The best solution for the user would be that, as soon as you propose a tx the nonce in the safe updates right away, so the next tx already has blocked the nonce of the tx being processed, and if for some reason the tx fails the nonce is kicked back to the value it was before, but I'd assume that is not a feasible solution

katspaugh:
I returned it back to todo to check why the reset button appears.

schmanu:
It appears because franco clicked inside the field.
Currently we trigger validation "onTouch". So as soon as a user touches the nonce field it counts as changed by the user.

If we do not want this we have to revert to the additional previous fix in the SafeTxProvider.
OR we could also fetch the nonce on-chain additionally when the modal is open.
Here is another user with this issue. This time with batched transactions and a slow indexer:
https://github.com/5afe/safe-support/issues/280

katspaugh:
Ah, OK. Hauptsache it shouldn't show a reset button if the field was updated programmatically.

PR #2395, opened 2023-08-14
Fix: keep prev spending limits when polling

### LINKED ISSUE #2389: Spending limit table disappears/reappears with the 15 secs pooling
## Bug description

Spending limit table disappears/reappears with the 15 secs pooling. I assume is to update any changes to the balances

## Environment
 - Browser: Chrome
 - Wallet: any
 - Chain: Ethereum mainnet
 - Env: current prod 1.16 build

## Steps to reproduce
 1. Go to a safe with spending limits like this one eth:0x8675B754342754A30A2AeF474D114d8460bca19b
 2. Go to the spending limit table and wait

## Expected result
The table don't disappears. The values are updated with every pooling

## Obtained result
The table goes away and reappears a few secs later

## Screenshots
spending limit flickr

katspaugh:
## What it solves

Resolves #2389

## How this PR fixes it

Don't reset spending limits on each poll.

### DIFF
--- a/src/components/safe-apps/types.ts
+++ b/src/components/safe-apps/types.ts
@@ -40,7 +40,9 @@ const FEATURES = [
   'speaker-selection',
 ]
 
-export type AllowedFeatures = typeof FEATURES[number]
+type FeaturesType = typeof FEATURES
+
+export type AllowedFeatures = FeaturesType[number]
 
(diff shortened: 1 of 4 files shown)


usame-algan (approved):
Deployment fails for lint reasons I think but I've tested the change locally. The table doesn't disappear anymore.

PR #2398, opened 2023-08-15
feat: eip-1559 fixed gas prices

### LINKED ISSUE #2222: EIP-1559 is not enabled on Optimism and gas price estimation is leading to struck txs

schmanu:
## What it solves
This PR fixes the usage of chain gas configs in general and adds a new fixed gas config.
Resolves #2222 

## How this PR fixes it
Enabled fixed gas prices for EIP-1559

## How to test it
**Fixed 1559 gas prices**
We added Optimism with the desired gas price config to staging and enabled 1559 there.
1. Create any tx on optimism staging
2. Go to execute modal
3. Expand the gas price details
4. Observe 0.1 for maxFeePerGas and 0.0001 for maxPriorityFeePerGas
5. Execute
6. Observe same values in i.e. connected metamask

**Bug fix: Use gas price configs**
This PR also fixed our gas estimations in general which never considered the gas price configs from our config service before.
To test this try to create transactions on networks with specific gas price configs:
- Gas price oracles
  - Mainnet
  - Polygon
- Fixed
  - Görli

## Screenshots
The screenshot shows the desired fixed gas price of 0.1 maxFee and 0.0001 maxPrio on Optimism.
Screenshot 2023-08-15 at 10 35 36

## Checklist

### DIFF
--- a/src/hooks/__tests__/useGasPrice.test.ts
+++ b/src/hooks/__tests__/useGasPrice.test.ts
@@ -1,6 +1,7 @@
 import { BigNumber } from 'ethers'
 import { act, renderHook } from '@/tests/test-utils'
 import useGasPrice from '@/hooks/useGasPrice'
+import { useCurrentChain } from '../useChains'
 
 // mock useWeb3Readonly
 jest.mock('../wallets/web3', () => {
@@ -8,8 +9,8 @@ jest.mock('../wallets/web3', () => {
     getFeeData: jest.fn(() =>
       Promise.resolve({
         gasPrice: undefined,
-        maxFeePerGas: BigNumber.from('0x956e'),
-        maxPriorityFeePerGas: BigNumber.from('0x136f'),
+        maxFeePerGas: BigNumber.from('0x956e'), //38254
+        maxPriorityFeePerGas: BigNumber.from('0x136f'), //4975
       }),
     ),
   }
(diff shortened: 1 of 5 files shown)


src/hooks/__tests__/useGasPrice.test.ts
```diff
@@ -170,6 +170,56 @@ describe('useGasPrice', () => {
...
+      await Promise.resolve()
+    })
+    // assert the hook is not loading
+    expect(result.current[2]).toBe(false)
+
+    // assert fixed gas price as minimum of 0.1 gwei
+    expect(result.current[0]?.maxFeePerGas?.toString()).toBe('100000000')
+
+    // assert fixed priority fee
+    expect(result.current[0]?.maxPriorityFeePerGas?.toString()).toBe('100000')
```
usame-algan, line 198: nit: You can also wrap all the `expects` with an `await waitFor()` call instead of having an empty `Promise.resolve()` to achieve the same

src/hooks/useGasPrice.ts
```diff
@@ -53,10 +96,35 @@ const getGasPrice = async (gasPriceConfigs: GasPrice): Promise<BigNumber | undef
...
+      maxFeePerGas: estimation.gasPrice,
+      maxPriorityFeePerGas: isEIP1559 ? feeData?.maxPriorityFeePerGas : undefined,
+    }
+  }
+
+  return {
+    maxFeePerGas: undefined,
+    maxPriorityFeePerGas: undefined,
+  }
+}
```
usame-algan, line 126: This looks more readable now but I would suggest to write a few tests to cover some of the edge-cases like if there is no estimation and it is EIP1559 but there is also no feeData etc.

usame-algan (approved):

francovenica:
For goerli I think 200 is waaay to high. Not a big deal since goerli is a test network, but I think we can reduce it to 50 or so:
here in MM estimates that "Aggressive" should be " 2 ".
gas fee

francovenica:
Tried send funds for ETH and Poly and Add owner for Optimism

In optimism the values are set to 0.1 for maxFeePerGas and 0.0001 for maxPriorityFeePerGas and the tx works fine. These values were suggested by the people on optimism just fine
image

For ETH the tx worked fine, the values passed to MM are correct
image

For Poly we is also working fine, takes about 2 or so mins for the tx to go through tho. We are estimating a value that is slower than the "Slow" MM estimation, so we should monitor the user experience after releasing this.
image

NOTE: we seen with Manu and Usame that in Poly MM is using the same number of Prio Fee and uses it into Max base fee as well. It seems is the way it works for poly and it is how it works in Prod currently, so not an issue, but still leaving it here for the 
record

Since executing the tx works, and the only "issue" that I had was goerli way overestimating the gasprice I'm fine passing this ticket.

schmanu:
Hey @francovenica 
To me this is fine.
If we get user complaints about too slow transactions on polygon, we can just remove the gas oracle config which will fallback to the behavior as it was before.
Same for Görli. We should just remove the too high fixed value which will lead to the previous behavior.

francovenica:
Passing it then. thanks for the comment

PR #2399, opened 2023-08-15
fix: Add addOwner route and open tx flow

### LINKED ISSUE #2381: Linking to the 'add owner' screen
As a new safe member,
I want to be added to the safe,
So that I could start using it.

As a safe owner,
I want to add new member to the owners list,
So that they could start using it.

Overview:
* Mobile app provides a way for new owners to join existing safe. In short, it works like this:
  * To-be-added owner shares a link with the safe owner
  * Safe owner opens link and _adds a new owner or replaces existing owner_ with the address from the link
* Those links used to be supported before rewrite of the web ui
* This functionality is used by ~10% of mobile users

URI pattern:

     {web app}/{chain prefix}:{safe address}/addOwner?address={new owner address}

Example of a link

    https://app.safe.global/gor:0xabcdEEeEeEeEeEeEeEeEabCdEeeEEEEEEEeeEEeE/addOwner?address=0xaABCdeeEeeeeeeeeEeEEEABcdeEEeEeeEEEEeeEe

At present, such links lead to nowhere, and there's no way to link to the existing "add owner" or "replace owner" transaction modal.

The link is generated by the mobile app.

Original issue: https://github.com/safe-global/safe-ios/issues/3096

usame-algan:
## What it solves

Resolves #2381 

## How this PR fixes it

- Adds a new route `/addOwner` that redirects to the `/settings/setup` page
- Opens the txFlow with pre-filled address

## How to test it

1. Open `/addOwner?safe=<your safe address>&address=<the address you want to add as an owner>`
2. Add owner txFlow opens with the address filled in
3. Open `/addOwner?safe=<your safe address>`
4. Redirected to the settings page but no txFlow opening
5. Open `/addOwner?safe=<your safe address>&address=<an invalid address>`
6. Redirected to the settings page and txFlow opens with the address value but an error for invalid value is displayed

## Screenshots

Screenshot 2023-08-15 at 12 17 06

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/AddOwner/index.tsx
+++ b/src/components/tx-flow/flows/AddOwner/index.tsx
@@ -16,17 +16,7 @@ export type AddOwnerFlowProps = {
   threshold: number
 }
 
-const AddOwnerFlow = () => {
-  const { safe } = useSafeInfo()
-
-  const defaultValues: AddOwnerFlowProps = {
-    newOwner: {
-      address: '',
-      name: '',
-    },
-    threshold: safe.threshold,
-  }
-
+const FlowInner = ({ defaultValues }: { defaultValues: AddOwnerFlowProps }) => {
   const { data, step, nextStep, prevStep } = useTxStepper<AddOwnerFlowProps>(defaultValues)
 
   const steps = [
@@ -50,6 +40,22 @@ const AddOwnerFlow = () => {
       {steps}
     </TxLayout>
   )
+}
+
+const AddOwnerFlow = ({ address }: { address?: string }) => {
+  const { safe, safeLoading, safeLoaded } = useSafeInfo()
+
+  const defaultValues: AddOwnerFlowProps = {
+    newOwner: {
+      address: address || '',
+      name: '',
+    },
+    threshold: safe.threshold,
+  }
+
+  if (!safeLoaded || safeLoading) return null
+
+  return <FlowInner defaultValues={defaultValues} />
 }
 
 export default AddOwnerFlow
--- a/src/config/routes.ts
+++ b/src/config/routes.ts
@@ -9,6 +9,7 @@ export const AppRoutes = {
   home: '/home',
   cookie: '/cookie',
   addressBook: '/address-book',
+  addOwner: '/addOwner',
   _offline: '/_offline',
   apps: {
     open: '/apps/open',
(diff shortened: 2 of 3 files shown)


usame-algan:
One potential issue I found is that if the user presses Next too fast, the core-sdk might not be loaded yet and it errors out. To fix this we would have to add the `useSafeSDK` hook in those components and make sure to include it in the `setSafeTx` useEffect block.

katspaugh (approved):
Looks good. Please run `yarn routes` to update the routing table.

katspaugh:
Tested, it works. ✅

francovenica:
The cases described in the description are fine
Safe address and owner address takes to the tx flow to add that owner, with the address field already filled. This even work with ENS names.

If the address of the owner is invalid it will prefill it still, but validate the error. Tested this for invalid addresses, addreses that are already owners, the safe itself, wrongly checksummed addresses, addresses with the wrong prefix (addresses with no prefix will adopt the prefix of the current network)

If no owner address is in the URL it takes just to the settings of that safe. This is for both; if the address parameter is not in the URL or if the address parameter has no value

francovenica:
Issues/question:

Since this is mobile, probably there isn't much room for human error when creating the URL, but still, if you add a wrong address in the safe, instead of an error it will take to this wrong safe:
image

The URL can be used when not an owner or when disconnected. In These two scenarios the form should not be possible to be opened at all.
Note: there is a safeguard in the form, where an error that says "you are not the owner, you cannot execute the tx" is shown, so this issue could be left like it is as long we all agree that that error in the form is enough.
image

usame-algan:
Thanks for testing @francovenica!

> Since this is mobile, probably there isn't much room for human error when creating the URL, but still, if you add a wrong address in the safe, instead of an error it will take to this wrong safe:

I think this is also how it is on prod. For me it shows an error: https://add_owner--walletweb.review-wallet-web.5afe.dev/settings/setup?safe=nothing
Screenshot 2023-08-15 at 17 01 32
The `nothin...hing` looks weird though. I suggest we create a ticket for it to be fixed. Probably better to not display anything there if the safe can't be loaded.

> The URL can be used when not an owner or when disconnected. In These two scenarios the form should not be possible to be opened at all.

It is also possible to get into this state when you disconnect your wallet after going into a tx-flow so I suggest to leave it as it is.

francovenica:
Ah, ok, so these are existing issues.

I'll report the wrong safe address not being rejected by the app in another ticket. 

We can pass this one

PR #2409, opened 2023-08-16
fix: do not fetch safe version on-chain

schmanu:
## What it solves

Reduces RPC calls during tx creation

## How this PR fixes it
Pass the safe version from the SafeInfo instead of fetching it on-chain

## How to test it
1. Create any transaction. 
2. Witness no more Infura calls coming from `recommendedNonce.ts`

## Checklist


usame-algan:
Should we test that `safeTxGas` is correctly fetched/not fetched still?

katspaugh (approved):
👍

usame-algan (approved):
I've tested creating transactions in 1.3.0 and 1.1.1 safes and can confirm that the safeTxGas is still being fetched / not fetch.

PR #2410, opened 2023-08-16
Revert onboard autoconnect

katspaugh:
Revert #2343 because it increased RPC requests from 26 to 42 on an initial page load.


katspaugh:
I was wrong, autoconnect was innocent.

PR #2411, opened 2023-08-16
fix: memoize deadline + smart contract check calls

iamacook:
## What it solves

Reduces `eth_call` requests

## How this PR fixes it

The request for `redeemDeadline` in `useSafeTokenAllocation` is now memoized according to the `chainId` and `address` meaning it will only request once per specific address on a chain. `isSmartContractWallet` has been memoized in a similar fashion.

## How to test it

- Switch Mainnet/Goerli and observe one less request for getting the redemption deadline (from `useSafeTokenAllocation` in the call stack). The values should match that as on prod.
- Switch between the option of executing via relayer/wallet by going back/forward in Safe creation and observe one less request for checking the contract code (from `useWalletCanRelay` -> `isSmartContractWallet` in the call stack).

## Checklist

### DIFF
--- a/src/hooks/useSafeTokenAllocation.ts
+++ b/src/hooks/useSafeTokenAllocation.ts
@@ -9,13 +9,15 @@ import { useMemo } from 'react'
 import useAsync from './useAsync'
 import useSafeInfo from './useSafeInfo'
 import { getWeb3ReadOnly } from './wallets/web3'
+import { memoize } from 'lodash'
+import type { JsonRpcProvider } from '@ethersproject/providers'
 
 export const VESTING_URL =
   IS_PRODUCTION || cgwDebugStorage.get()
     ? 'https://safe-claiming-app-data.safe.global/allocations/'
     : 'https://safe-claiming-app-data.staging.5afe.dev/allocations/'
 
-type VestingData = {
+export type VestingData = {
   tag: 'user' | 'ecosystem' | 'investor' | 'user_v2' // SEP #5
   account: string
   chainId: number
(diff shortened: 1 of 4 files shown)


src/hooks/useSafeTokenAllocation.ts
```diff
@@ -66,10 +78,7 @@ const completeAllocation = async (allocation: VestingData): Promise<Vesting> =>
   }
 
   // Allocation is not yet redeemed => check the redeemDeadline
-  const redeemDeadline = await web3ReadOnly.call({
-    to: allocation.contract,
-    data: airdropInterface.encodeFunctionData('redeemDeadline'),
-  })
+  const redeemDeadline = await _getRedeemDeadline(allocation, web3ReadOnly)
```
schmanu, line 81: We could also cache `isRedeemed` if it is true as it cannot change once the user redeemed to improve it a bit further.

katspaugh (approved):

src/hooks/useSafeTokenAllocation.ts
schmanu: Nevermind. We need to fetch again as the `amountClaimed` could be changed!

francovenica:
The calls are hard to count since there are calls on every polling so is difficult to compare between prod and this PR, so I'm just making sure nothing has broken with these changes. I trust the peer review.
Also, by Aaron request, I've checked the relayer in the safe creation/tx signature and the safe tokens in the top bar are the same as they are in prod

LGTM

PR #2415, opened 2023-08-17
Fix: status widget for 1271 messages

### LINKED ISSUE #2384: [Tx Flow] Messages and the Transaction Status block
## Bug description

For messages, the tx status block on the right side is misplaced: is not a transaction, the confirmation there never changes, and it does not executes.

## Environment
 - Browser: Chrome
 - Wallet: any
 - Chain: any (but tested in goerli)
 - Env: Prod 1.16

## Steps to reproduce
 1. Clone this repo and open the application https://github.com/5afe/eip-1271-dapp
 2. Open a safe you own. Open the wallet connect Dapp in the safe
 3. Use the "Connect" in the eip1271 app and copy/paste the WC code from the modal in the safe WC Dapp
 4. In the eip1271 enable off-chain sigature, write any message and send
 5. In the safe, it should now show the form to sign the message

## Expected result
I'd suggest making it work with the same counter that is in the bottom of the form or remove the block completely

## Obtained result
The snapshot

## Screenshots
08-10-2023_1247x578(6338)

katspaugh:
## What it solves

Resolves #2384

When signing a message, it will say "Message status" instead of "Transaction status".

Screenshot 2023-08-17 at 12 04 09

### DIFF
--- a/src/components/tx-flow/common/TxLayout/index.tsx
+++ b/src/components/tx-flow/common/TxLayout/index.tsx
@@ -57,6 +57,7 @@ type TxLayoutProps = {
   hideNonce?: boolean
   isBatch?: boolean
   isReplacement?: boolean
+  isMessage?: boolean
 }
 
 const TxLayout = ({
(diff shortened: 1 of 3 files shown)


iamacook (approved):

francovenica:
LGTM

It looks like the snapshot in the tickets description.

francovenica:
Sorry, missclicked when closing it. I reopened it. Moving it to QA done

PR #2416, opened 2023-08-17
fix: get Safe address on-chain for ZK creations

### LINKED ISSUE #2370: [Safe creation] Predicted Safe address is wrong on zkSync
During a Safe creation on zkSync, we show the future Safe address and it's different from the actual resulting address.

See https://5afe.slack.com/files/U03DA8MS83Z/F05MHA2V52L/zksync_safe_creation.mp4

iamacook:
## What it solves

Resolves #2370

## How this PR fixes it

This updates the `@safe-global/safe-core-sdk` to `3.3.5` [as per requested](https://5afe.slack.com/archives/C03DAGWJCR5/p1692284450682389) which fixes Safe address prediction on ZK chains.

## How to test it

Create a Safe on a ZK chain and observe the correct Safe address in the creation stepper/navigation link after indexed.

## Checklist


src/components/new-safe/create/steps/StatusStep/useIsZkEvmChain.ts
```diff
@@ -0,0 +1,13 @@
...
+  ZKSYNC_ERA_MAINNET = '324',
+  POLYGON_ZKEVM = '1101',
+}
+
+// TODO: Remove when we update SDK
+export const useIsZkEvmChain = (): boolean => {
+  const chainId = useChainId()
+
+  return Object.values(ZkChains).includes(chainId as ZkChains)
+}
```
katspaugh, line 13: I would suggest using a config-service feature toggle for this instead.

iamacook: I've updated it accordingly, as well as adding it to the staging config service. I also added a note to the rollout plans of these networks.

src/components/new-safe/create/steps/StatusStep/StatusStepper.tsx
```diff
@@ -1,40 +1,91 @@
...
+      return
+    }
+
+    let isCurrent = true
+
+    web3ReadOnly
+      .getTransactionReceipt(pendingSafe.txHash)
+      .then((receipt) => {
+        if (isCurrent) {
+          const safeAddress = receipt.logs[0].address
```
schmanu, line 36: We should check the log's topic and search for the first log with the correct one (ProxyCreation)

If we for instance enable relaying on one of those networks there will also be Events from the Gelato Relayer.
Or if we execute the tx from another Smart Contract wallet there might be events from that wallet as well.

src/components/new-safe/create/steps/StatusStep/StatusStepper.tsx
```diff
@@ -1,40 +1,91 @@
...
+  const shouldFetchAddress = useHasFeature(FEATURES.SAFE_CREATION_FETCH_ADDRESS)
+  const [pendingSafe, setPendingSafe] = usePendingSafe()
+  const web3ReadOnly = useWeb3ReadOnly()
+
+  useEffect(() => {
+    if (
+      !shouldFetchAddress ||
+      status !== SafeCreationStatus.SUCCESS ||
+      !pendingSafe ||
+      !pendingSafe.txHash ||
```
schmanu, line 23: If we relay this will not work as we will never have a txHash.

It's okay as long as we first update the SDK before we enable relaying on any of the zkevm chains ;)

src/components/new-safe/create/steps/StatusStep/StatusStepper.tsx
```diff
@@ -1,40 +1,91 @@
...
 import EthHashInfo from '@/components/common/EthHashInfo'
 import { SafeCreationStatus } from '@/components/new-safe/create/steps/StatusStep/useSafeCreation'
 import StatusStep from '@/components/new-safe/create/steps/StatusStep/StatusStep'
 import { usePendingSafe } from './usePendingSafe'
+import { useHasFeature } from '@/hooks/useChains'
+import { useWeb3ReadOnly } from '@/hooks/wallets/web3'
+import { FEATURES } from '@/utils/chains'
+
+// TODO: Remove when we update SDK
+const useSyncZkEvmSafeAddress = (status: SafeCreationStatus) => {
```
schmanu, line 13: We need a test for this.

schmanu (changes requested):

src/components/new-safe/create/steps/StatusStep/StatusStepper.tsx
iamacook: I thought about this but I don't forsee up allowing relaying from the beginning. We should prioritise upgrading the SDK first.

src/components/new-safe/create/steps/StatusStep/StatusStepper.tsx
iamacook: I've updated it to parse the logs against the ProxyFactory interface.

src/components/new-safe/create/steps/StatusStep/StatusStepper.tsx
iamacook: I've added full test coverage.

usame-algan (approved):

francovenica:
Looks good.
The address shown during the safe creation process is the same as the safe after being created:
image
image
image

francovenica:
@iamacook Since @schmanu requested a change I'll not pass the ticket to QA done yet. Let me know if that change requieres a re-test.

PR #2417, opened 2023-08-17
Refactor: Safe creation hooks + rm unused files

### LINKED ISSUE #2313: Created Safes not added to `addedSafes`
## Bug description

When creating a Safe, it is not always added to the store.

## Steps to reproduce
 1. Create a Safe and observe that it is not present in the sidebar.

## Expected result

Safe is present in the sidebar after creation.

## Obtained result

Safe is not always present in the sidebar after creation.

katspaugh:
## What it solves

Resolves #2313

## How this PR fixes it

Adds a new created Safe to the Added Safes list as soon as a creation tx is executed.

+ also removed some unused files.

## How to test it

* Create a Safe.
* Right after executing the transaction, close the tab
* Safe should not be added

* Create another Safe
* wait till the tx executes
* check that your new Safe is added


src/components/new-safe/create/steps/StatusStep/useSafeCreationEffects.ts
```diff
@@ -19,36 +19,32 @@ const useSafeCreationEffects = ({
   const chainId = useChainId()
   const [pendingSafe, setPendingSafe] = usePendingSafe()
 
+  // Asynchronously wait for Safe creation
   useEffect(() => {
```
katspaugh, line 23: I've split this effect into 4 smaller effects, each with their own set of dependencies and actions.

src/components/new-safe/create/steps/StatusStep/useSafeCreationEffects.ts
```diff
@@ -19,36 +19,32 @@ const useSafeCreationEffects = ({
...
-      if (pendingSafe?.safeAddress) {
-        pollSafeInfo(chainId, pendingSafe.safeAddress)
-          .then(() => setStatus(SafeCreationStatus.INDEXED))
-          .catch(() => setStatus(SafeCreationStatus.INDEX_FAILED))
-      }
-      return
+    if (pendingSafe?.safeAddress) {
+      pollSafeInfo(chainId, pendingSafe.safeAddress)
+        .then(() => setStatus(SafeCreationStatus.INDEXED))
+        .catch(() => setStatus(SafeCreationStatus.INDEX_FAILED))
```
usame-algan, line 27: We should wait until `SUCCESS` before we start polling for the safe info otherwise it starts to poll even before the user has submitted the transaction in their wallet.

src/components/new-safe/create/steps/StatusStep/useSafeCreationEffects.ts
```diff
@@ -19,36 +19,32 @@ const useSafeCreationEffects = ({
...
     }
+  }, [chainId, pendingSafe?.safeAddress, setStatus])
 
-    if (status === SafeCreationStatus.WALLET_REJECTED) {
-      trackEvent(CREATE_SAFE_EVENTS.REJECT_CREATE_SAFE)
+  // Add Safe to Added Safes and add owner and safe names to Address Book
+  useEffect(() => {
+    if (pendingSafe?.safeAddress && status === SafeCreationStatus.PROCESSING) {
```
usame-algan, line 33: We are setting `PROCESSING` as soon as the user submits the transaction in their wallet but the transaction could still fail or revert. Should we still add those entries to the address book even if the safe hasn't been deployed?

katspaugh: I guess it's a trade-off, I also tend to think it's better to keep SUCCESS like before.

katspaugh: Restored SUCCESS in both hooks.

usame-algan (approved):
👍

francovenica:
So how it works is, that if you close the tab right after you click sign in your wallet the safe will NOT be added to the sidebar. In order for that to happen you have to stay in the "processing safe creation page" (where the bullet points show) and wait for those to reach to the end for the safe to be added to the sidebar.

Is this the expected behavior?

image

katspaugh:
It’s a trade-off:
* if you add it before the tx executes, and it fails, you’re left with a broken safe added to your list
* If you add it after it executes, some users might get impatient and close the tab

So I don’t know what’s best. Keeping it as it was seemed like the safest choice.

francovenica:
Personally I never had the issue of safes not showing in the sidebar after creation. also most people will have like 2, 3, maybe 6 safes, not 100 so is not gonna be hard to look for them in the owned safes list
Maybe we can just put a warning during the safe creation process to ask people to wait for a bit or if it is taking too long comeback to safe creation to see their safe process to be done so we asure the safe will be listed in the "Added safes" list in the sidebar.

katspaugh:
Good idea, I’ll add a prompt before closing a tab.

francovenica:
I see a warning, but is an alert in the browser and it doesn't say the same message I see in the file you changed. Is triggering the right warning?
Also it only triggers if you close the tab or open a new URL in that tab, but if you click the Safe logo for example the alert doesn't show

image

katspaugh:
Yeah, that's correct. That message is standard, in some browsers it will take the custom message you see in the code.

PR #2420, opened 2023-08-18
Refactor: load Spending limits on demand instead of global store

### LINKED ISSUE #2414: Refactor use/loading of spending limit data
## What is the feature about

`useLoadSpendingLimits` breaks the rules of hooks, as well as makes a large number of requests. We should look into refactoring this to not require the disabling of exhaustive dependencies and reduce the requests.

Note: the `txHistoryTag` is a dependency to update the settings.

## The list of requirements

- [ ] Improve dependencies (move `isSpendingLimitEnabled` check outside of `getSpendingLimits`).
- [ ] Reduce number of requests and/or only use the data where it is needed, e.g. in settings, transaction flow, convert `CheckWallet` to only check if there's a spending limit (check where selectors are used)

katspaugh:
## What it solves

Resolves #2414

## How it solves it

* Spending limits are now not stored in a global store
* They are fetched on demand from three places:
   1. "New transaction" button – to check if the wallet is a beneficiary, if wallet isn't an owner and spending limits are enabled
   2. Spending limits settings – to display the list of existing allowances
   3. Send tokens flow – to allow sending tokens from an allowance

## How to test it

1.
* Open a Safe with spending limits
* Go to Settings and check that all the allowances are listed
* Create a new allowance
* Remove an allowance

2.
* Open a Safe with spending limits
* Connect a wallet that isn't an owner but is a beneficiary
* New Transaction should be active
* Send a token from your allowance


src/components/balances/AssetsTable/index.tsx
```diff
@@ -93,10 +92,8 @@ const SendButton = ({
   tokenInfo: TokenInfo
   onClick: (tokenAddress: string) => void
 }): ReactElement => {
-  const spendingLimit = useSpendingLimit(tokenInfo)
-
   return (
-    <CheckWallet allowSpendingLimit={!!spendingLimit}>
```
katspaugh, line 99: It's OK to just not enable this button for non-owners.

iamacook: It's possible to add spending limits for non-owners though.

src/hooks/useSpendingLimits.ts
```diff
@@ -70,56 +82,71 @@ export const getTokensForDelegate = async (
...
-    ),
+    delegates.map(async (delegate) => getTokensForDelegate(contract, safeAddress, delegate, tokenInfoFromBalances)),
   )
+
   return spendingLimits.flat().filter(discardZeroAllowance)
 }
 
-export const useLoadSpendingLimits = (): AsyncResult<SpendingLimitState[]> => {
-  const { safeAddress, safe, safeLoaded } = useSafeInfo()
+export const useSpendingLimitDelegates = (isEnabled: boolean): AsyncResult<string[]> => {
```
iamacook, line 99: We should add test coverage for this.

src/hooks/useSpendingLimits.ts
```diff
@@ -70,56 +82,71 @@ export const getTokensForDelegate = async (
...
+  const isEnabled = useHasFeature(FEATURES.SPENDING_LIMIT)
+  if (!safe || !safe.modules || !isEnabled) return false
+
+  const moduleAddress = getSpendingLimitModuleAddress(safe.chainId)
+  if (!moduleAddress) return false
+
+  return safe.modules.some((module) => sameAddress(module.value, moduleAddress))
+}
+
+export const useAllSpendingLimits = (): AsyncResult<SpendingLimitState[]> => {
```
iamacook, line 124: We should also add test coverage for this.

src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
```diff
@@ -56,7 +55,7 @@ const CreateTokenTransfer = ({
   const safeTokenAddress = getSafeTokenAddress(chainId)
   const isSafeTokenPaused = useIsSafeTokenPaused()
   const isOnlySpendingLimitBeneficiary = useIsOnlySpendingLimitBeneficiary()
```
iamacook, line 57: I think we need to `allow` this for non-owners beneficiaries.

iamacook (commented):
This reduces the initial requests, but results in more overall as we aren't caching the responses anymore. It might be worthwhile keeping the store. What do you think?

I also noticed that the spending limits overview won't update in the settings if a spending limit transaction executes, or the Safe info updates (e.g. the module is removed) as we are no longer reacting to the tag.

src/components/balances/AssetsTable/index.tsx
katspaugh: I'm aware but this button is not worth the extra requests. They can still transact via the New tx button.

src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
katspaugh: It is allowed.

katspaugh:
> This reduces the initial requests, but results in more overall as we aren't caching the responses anymore. It might be worthwhile keeping the store. What do you think?

This depends on your usage and is now on demand. The previous code would fetch spending limits and potentially token info from the blockchain unconditionally, and refresh based on the history. New code will only fetch the beneficiary list if you're not an owner and spending limits are activated (for the New transaction button). It will also fetch this info when you make a Send tokens tx or open the Spending limit settings.

> I also noticed that the spending limits overview won't update in the settings if a spending limit transaction executes, or the Safe info updates (e.g. the module is removed) as we are no longer reacting to the tag.

I'll restore it.

src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
iamacook: `allow` is `false` [by default](https://github.com/safe-global/safe-wallet-web/blob/refactor-spending/src/hooks/useIsOnlySpendingLimitBeneficiary.tsx#L5).

katspaugh: Ah, gotcha, thanks!

PR #2421, opened 2023-08-18
Fix: do not sandbox safe icon URLs

### LINKED ISSUE #2419: Self-hosted icons are rendered inside an iframe and not cached
## Bug description

All icons in the history/queue are rendered inside an iframe to prevent external domains to set cookies. Some of those icons like `outgoing.svg` are self-hosted and the requests are not cached so it can create unnecessary network overhead.

## Steps to reproduce
 1. Go to history
 2. Open the network tab
 3. Filter by `outgoing.svg`
 4. Observe the same file being fetched more than once

## Expected result

Self-hosted icons should be cached (only fetched once). No need to render them in an iframe.

## Screenshots
Screenshot 2023-08-18 at 10 27 59

katspaugh:
## What it solves

Resolves #2419

## How this PR fixes it

This PR checks if an icon src is a relative path or comes from one of our own domains, and inserts it as a regular image in that case. Otherwise, it embeds it in a sandboxed iframe like before.

### DIFF
--- a/cypress/e2e/smoke/tx_history.cy.js
+++ b/cypress/e2e/smoke/tx_history.cy.js
@@ -31,7 +31,7 @@ describe('Transaction history', () => {
       .last()
       .within(() => {
         // Type
-        cy.get('iframe').should('have.attr', 'title', INCOMING)
+        cy.get('img').should('have.attr', 'alt', INCOMING)
         cy.contains('div', 'Received').should('exist')
 
         // Info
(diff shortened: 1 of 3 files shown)


usame-algan (approved):
Looks good! I think there is one e2e test failing because it expects to find an iframe though.

francovenica:
LGTM

A comparison between dev and this PR

PR #2426, opened 2023-08-22
Fix: hide Batch icon if not on a Safe route

### LINKED ISSUE #2425: Batch tx icon in the top bar present when no safe is loaded
## Bug description

The batch tx button is shown in the top bar when no safe is loaded (in the welcome page or safe creation for example). 
This button opens a sidebar that allows to start a new tx with no safe.

## Environment
 - Browser: Chrome
 - Wallet: any
 - Chain: any
 - Env: Current prod 1.17

## Steps to reproduce
 1. Go to the welcome page
 2. Click on the batch tx button in the topbar

## Expected result
The button shouldn't show at all if you are not in a safe

## Obtained result
The button is present
You can start a tx outside of a safe (you cannot create it tho, the form will fail since there are no tokens to select, so no risks so far)

## Screenshots
batch with no safe

katspaugh:
## What it solves

Resolves #2425

## How this PR fixes it

Hides the Batch icon if there's no Safe on that URL.

## How to test it
* Go to a page like Welcome and see that there's no Batch button in the header

### DIFF
--- a/src/components/batch/BatchIndicator/index.tsx
+++ b/src/components/batch/BatchIndicator/index.tsx
@@ -11,7 +11,7 @@ const BatchIndicator = ({ onClick }: { onClick?: () => void }) => {
   return (
     <BatchTooltip>
       <Track {...BATCH_EVENTS.BATCH_SIDEBAR_OPEN} label={length}>
-        <ButtonBase onClick={onClick} sx={{ p: 0.5 }}>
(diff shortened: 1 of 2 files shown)


francovenica:
LGTM

The button doesn't show anymore if a safe is not currently loaded in the UI

Checked other pages like create/load safe, preferences, the pages listed in the footer... It looks good

PR #2430, opened 2023-08-23
Feat: send deviceType to GTM

katspaugh:
## What it solves

Sending a device type manually would allow us to disable the "[Granular location & device data](https://support.google.com/analytics/answer/12002752?hl=en&utm_id=ad)" on GA.

## How to test

* Open the app on desktop – every GTM event should have `deviceType: 'desktop'`
* Open the app on a tablet (or just resize the window to < 900px) – every GTM event should have `deviceType: 'tablet'`
* Open the app on mobile – – every GTM event should have `deviceType: 'mobile'`

### DIFF
--- a/src/services/analytics/gtm.ts
+++ b/src/services/analytics/gtm.ts
@@ -17,7 +17,7 @@ import {
   GOOGLE_TAG_MANAGER_DEVELOPMENT_AUTH,
 } from '@/config/constants'
 import type { AnalyticsEvent, EventLabel, SafeAppSDKEvent } from './types'
-import { EventType } from './types'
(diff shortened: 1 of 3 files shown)


usame-algan (approved):
Looks good 👍  I've tested navigating with different viewport sizes and saw the values end up in GTM. We still need to add the mapping in GTM/GA once we go live with this.

francovenica:
Looks good

Tried in the different resolutions ( >900px, 900 - 600px, <600px ) navigating through the site,  checking that the "deviceType" is logged with the GTM event, correctly having the value depending on the screen resolution.
image

PR #2436, opened 2023-08-24
Refactor: update to Next 13

katspaugh:
## What it solves

Upgrades the app to Next 13.

Plus `yarn upgrade` + a few small changes in the readme and configs.

## How to test it

It touches a lot of places where internal links are used. It might potentially break some link-related CSS/layout.
So I would suggest testing it by opening all routes and checking that they look the same.


README.md
```diff
@@ -1,4 +1,4 @@
-# <img src="/public/images/logo.svg" height="60" valign="middle" alt="Safe{Wallet}" />
+# <img src="/public/images/logo.svg" height="60" valign="middle" alt="Safe{Wallet}" style="background: #fff; padding: 20px; margin: 0 -20px" />
```
katspaugh, line 1: It blends with the background in dark mode otherwise.

next.config.mjs
```diff
@@ -22,28 +22,20 @@ const nextConfig = {
...
+    },
+    '@mui/icons-material/?(((\\w*)?/?)*)': {
+      transform: '@mui/icons-material/{{ matches.[1] }}/{{member}}',
+    },
+    lodash: {
+      transform: 'lodash/{{member}}',
+    },
+    'date-fns': {
+      transform: 'date-fns/{{member}}',
+    },
```
usame-algan, line 38: Moving this out of the experimental property somehow makes it load a lot more modules again (17k instead of 6k) and slowing down local development. Do we also have to upgrade nextjs for it to work?

katspaugh: I'm seeing an increase in chunk sizes due to the upgraded packages (from 900 KB to 1.01 MB) regardless of this change.

katspaugh: Without this, the build says:
> warn  - You have defined experimental feature (modularizeImports) in next.config.mjs that does not exist in this version of Next.js.

katspaugh: Anyway, let's not touch it for now then.

next.config.mjs
```diff
@@ -37,14 +37,6 @@ const nextConfig = {
...
     },
   },
-  async rewrites() {
-    return [
-      {
-        source: '/:safe([a-z0-9-]+\\:0x[a-fA-F0-9]{40})/:path*',
-        destination: '/:path*?safe=:safe',
-      },
-    ]
-  },
```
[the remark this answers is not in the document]
katspaugh: Because it hits a 404 page that does a redirect. This rewrite was used only for the dev server.

jest.setup.js
```diff
@@ -47,3 +47,11 @@ const englishTestLocale = 'en'
...
 
 jest.spyOn(Intl, 'NumberFormat').mockImplementation((locale, ...rest) => new NumberFormat([englishTestLocale], ...rest))
+
+const defineProperty = Object.defineProperty
+Object.defineProperty = (obj, prop, desc) => {
+  if (prop !== 'prototype') {
+    desc.configurable = true
+  }
+  return defineProperty(obj, prop, desc)
+}
```
katspaugh, line 57: This is needed for jest.spyOn to work.

iamacook: Nit: can we add a comment regarding this?

iamacook (approved):

jest.setup.js
katspaugh: Added.

francovenica:
Went around the site, checking that it looks fine

Safe creation/adding
owner edition
Safe tx creation/execution
Settings (and internal pages)
Spending limits
Address book
Dapps
Transactions tabs queue/history

Is anything else I should cover let me know and I'll do it during the regression of the next RC.

PR #2439, opened 2023-08-24
Fix: multisend verification only worked for the latest contract version

katspaugh:
## What it solves

safe-deployments 1.2.6 made Safe 1.4.1 the default. See https://github.com/safe-global/safe-deployments/pull/248

This breaks our code because we were always retrieving multisend contract addresses w/o a version. It worked for 1.3.0 but not for any other version.

## How to test

1.
* Open a single transaction view for a multisend tx on Safe 1.3.0 on mainnet
* Open a single transaction view for a multisend tx on Safe 1.1.1 on mainnet
* Open a single transaction view for a multisend tx on Safe 1.3.0 on Polygon
* Multisend details should be displayed for all of them

2.
* Create and execute a multisend tx in a 1.3.0 Safe
* Create and execute a multisend tx in a 1.1.1 Safe

### DIFF
--- a/src/services/contracts/safeContracts.ts
+++ b/src/services/contracts/safeContracts.ts
@@ -102,26 +102,16 @@ export const getReadOnlyGnosisSafeContract = (chain: ChainInfo, safeVersion: str
 
 // MultiSend
 
-const getMultiSendContractDeployment = (chainId: string) => {
-  return getMultiSendDeployment({ network: chainId }) || getMultiSendDeployment()
-}
-
-export const getMultiSendContractAddress = (chainId: string): string | undefined => {
-  const deployment = getMultiSendContractDeployment(chainId)
-
-  return deployment?.networkAddresses[chainId]
+export const getMultiSendContractAddress = (network: string, version: string): string | undefined => {
+  const deployment = getMultiSendDeployment({ network, version })
+  return deployment?.networkAddresses[network]
 }
 
 // MultiSendCallOnly
 
-const getMultiSendCallOnlyContractDeployment = (chainId: string) => {
-  return getMultiSendCallOnlyDeployment({ network: chainId }) || getMultiSendCallOnlyDeployment()
-}
-
-export const getMultiSendCallOnlyContractAddress = (chainId: string): string | undefined => {
-  const deployment = getMultiSendCallOnlyContractDeployment(chainId)
-
-  return deployment?.networkAddresses[chainId]
+export const getMultiSendCallOnlyContractAddress = (network: string, version: string): string | undefined => {
+  const deployment = getMultiSendCallOnlyDeployment({ network, version })
+  return deployment?.networkAddresses[network]
 }
 
 export const getMultiSendCallOnlyContract = (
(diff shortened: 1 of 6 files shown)


src/utils/transaction-guards.ts
```diff
@@ -78,10 +78,17 @@ export const isCustomTxInfo = (value: TransactionInfo): value is Custom => {
   return value.type === TransactionInfoType.CUSTOM
 }
 
-export const isSupportedMultiSendAddress = (txInfo: TransactionInfo, chainId: string): boolean => {
+export const isSupportedMultiSendAddress = (
+  txInfo: TransactionInfo,
+  chainId: string,
+  version: string | null,
+): boolean => {
+  // Safe version can be null when a Safe is being loaded
```
iamacook, line 86: The version is also `null` for unsupport base contracts.

iamacook (approved):

katspaugh:
I've realized this is not a good solution either because if a Safe Account is upgraded, existing transactions from a previous version won't be detected as multisend correctly.

Perhaps we should check the `isTrusted` flag instead of comparing contract addresses.

iamacook:
> I've realized this is not a good solution either because if a Safe Account is upgraded, existing transactions from a previous version won't be detected as multisend correctly.
> 
> Perhaps we should check the `isTrusted` flag instead of comparing contract addresses.

We could check whether it's supported by looping over an array of contract versions. We could get the array from either the [`SafeVersion`s in the SDK](https://github.com/safe-global/safe-wallet-web/blob/firebase/src/hooks/coreSDK/safeCoreSDK.ts#L22), or the `released` assets of the deployments.

katspaugh:
@iamacook yeah, I was thinking the same but after talking with infra, we should probably do it on CGW, see https://github.com/safe-global/safe-client-gateway-nest/issues/657.

So I suggest we just check the current Safe version at this time, just for safe-deployments 1.2.6 to still work, and do a more robust solution on CGW.

iamacook (approved):

katspaugh:
Looks like this verification was added not for security reasons but just to not break the UI when some other contract has a `multiSend` method. See https://github.com/5afe/safe-react/pull/4023

I've removed the `isSupportedMultisendContract` function and enhanced `isMultiSendTxInfo` instead. The visual component that renders multisend actions also has guards inside that would just return early if no nested actions are found.

src/services/contracts/safeContracts.ts
```diff
@@ -132,7 +109,7 @@ export const getMultiSendCallOnlyContract = (
   const ethAdapter = createEthersAdapter(provider)
 
   return ethAdapter.getMultiSendCallOnlyContract({
-    singletonDeployment: getMultiSendCallOnlyContractDeployment(chainId),
+    singletonDeployment: getMultiSendCallOnlyDeployment({ network: chainId, version: safeVersion || undefined }),
```
katspaugh, line 112: This is also an important fix, because it was getting 1.3.0 multisend by default which would break any time.

iamacook (approved):

francovenica:
To do a multisend I just batched 2 send funds tx in a single tx, made sure that interacts with `0x40A2aCCbd92BCA938b02010E17A5b8929b49130D` since that's the multisig contract

I compared prod with this PR and the tx looks the same, but still nothing out of order there so is ok
I was able to successfully create and execute tx in safes 1.1.1 and 1.3.0 in ETH and 1.3.0 in Poly

Looks good to me

PR #2423, opened 2023-08-21
Fix: show known addresses in decoded tx data

katspaugh:
## What it solves

Resolves #2422

## How this PR fixes it
Displays name and icon for known addresses in decoded tx details and in multisend details.

## How to test it

* Open a 1.1.1 Safe
* Create a tx to upgrade it to 1.3.0
* See that the mastercopy and fallback handler addresses have a name and icon

## Screenshots
Screenshot 2023-08-21 at 12 49 53

PR #2443, opened 2023-08-25
Feat: show contract names in multisend summary

katspaugh:
## What it solves

A follow-up on #2423

In multisend actions, in addition to displaying the contract method, show the contract name if it comes from the backend.

Before:
Screenshot 2023-08-25 at 12 15 36

After:
Screenshot 2023-08-25 at 12 14 07


usame-algan (approved):
Makes sense to display the name if we have it nice!

Imo we need to adjust the design at some point for displaying name + method maybe having one of them not bold would already help but I am sure @liliiaorlenko has more ideas.

PR #2450, opened 2023-08-28
fix: pin deployment version

### LINKED ISSUE #2207: Ensure that new safes created with the latest version supported by the services
## What is the feature about
https://github.com/safe-global/safe-deployments/pull/248 is soon to be released, and that will switch the default Safe version returned by the deployments package to 1.4.1. 1.4.1 is not supported by the services yet, so the interface needs to ensure it uses the latest supported version.

## The list of requirements
- Pin the deployed Safes version to the latest supported one by the services (1.3.0 at the moment of writing)

iamacook:
## What it solves

Partially resolves #2207

## How this PR fixes it

All deployments and contract instantiation versions are now pinned to `LATEST_SAFE_VERSION` if the Safe version is `<1.0.0 || `>LATEST_SAFE_VERSION`. If the version is pinned, the default mastercopy is used instead of the current one.

### General
- The SDK initialises correctly for versions `1.0.0-1.3.0` on L1/L2.

### Transactions
- Queue batching works.
- Transaction batching works.
- Transaction validity checks work (e.g. "This transaction will most likely fail...").
- (MultiSend) simulation works.
- Relaying works.
- Off-chain message signing works.

### Safe creation
- Estimation works.
- (Relayed) Safe creation works (creating `1.3.0` Safes with `1.3.0` Fallback Handlers).

### Other
- (Un-)official fallback handlers are rendered correctly in the settings.

## Checklist


iamacook:
Closing in favour of #2467.

PR #2451, opened 2023-08-28
fix: populate address from URL in load Safe flow

### LINKED ISSUE #2438: /new-safe/load does not support address parameter
## Bug description

When clicking this link: https://safe-wallet-web.dev.5afe.dev/new-safe/load?chain=eth&address=0x308686553a1EAC2fE721Ac8B814De638975a276e 

The upcoming page is supposed to prefill chain and address. But it only selects the chain

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to https://app.safe.global/new-safe/load?chain=eth&address=0x308686553a1EAC2fE721Ac8B814De638975a276e 

## Expected result

chain and address are prefilled (only the name has to be added by the user)

## Obtained result

Only chain is selected.

## Screenshots
image

iamacook:
## What it solves

Resolves #2438

## How this PR fixes it

The `CardStepper` now has the `address` assigned as a `key` to trigger a re-render if an `address` parameter is in the URL as the `query` is not initially available.

## How to test it

Open the load Safe flow with an `address` search parameter and observe the field populated. Changing the address and navigating forwards/backwards should *not* result in the URL `address` repopulating the field.

## Checklist

### DIFF
--- a/src/components/new-safe/load/index.tsx
+++ b/src/components/new-safe/load/index.tsx
@@ -20,8 +20,8 @@ export const LoadSafeSteps: TxStepperProps<LoadSafeFormData>['steps'] = [
   {
     title: 'Name, address & network',
     subtitle: 'Paste the address of the Safe Account you want to add, select the network and choose a name.',
-    render: (_, onSubmit, onBack, setStep) => (
-      <SetAddressStep onSubmit={onSubmit} onBack={onBack} data={_} setStep={setStep} />
+    render: (data, onSubmit, onBack, setStep) => (
+      <SetAddressStep onSubmit={onSubmit} onBack={onBack} data={data} setStep={setStep} />
     ),
   },
   {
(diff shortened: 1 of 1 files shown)


katspaugh (approved):

francovenica:
I noticed that, if you change the safe during the Add safe process, the URL remains the same, so it still shows the safe address you loaded with the URL. Talked with Aaron about it to know if might be an issue and it seems is not a problem at all. 

So is all good. 
The address load the safe fine. It can be tested with the URL Dirk provided in the main ticket 
https://populate_address--walletweb.review-wallet-web.5afe.dev/new-safe/load?chain=eth&address=0x308686553a1EAC2fE721Ac8B814De638975a276e

PR #2453, opened 2023-08-28
fix: close batch flow when removing last draft

### LINKED ISSUE #2428: [Batching] Remove tx while in the form
## Bug description

When confirming a tx batch, you can open the batch tx sidebar to reorder or delete tx. 
Is possible for the user to delete them all, end with a bogus tx that probably will fail if executed.

## Environment
 - Browser: Chrome
 - Wallet: any
 - Chain: any
 - Env: 1.17

## Steps to reproduce
 1. Go to a safe
 2. Add 2 tx to the batch
 3. Use the "Confirm transaction" button in the Batch tx sidebar so they are put in a single transaction
 4. While in the form, reopen the batch tx sidebar
 5. Remove the tx from the sidebar one by one

## Expected result

Tx shouldn't be executable after all transactions are gone

## Obtained result

A bogus tx remains, with a warning that the tx might fail.

## Screenshots
batch tx removal

iamacook:
## What it solves

Resolves #2428

## How this PR fixes it

If the batch confirmation flow is open, it is now closed when the last draft is removed via the sidebar.

## How to test it

Batch transaction(s) and enter the confirmation flow. Whilst in the flow, remove the last draft from the sidebar and observe the confirmation flow close. It should _not_ close other flows that may be open under the sidebar.

## Screenshots

close-batch

## Checklist

### DIFF
--- a/src/components/batch/BatchSidebar/index.tsx
+++ b/src/components/batch/BatchSidebar/index.tsx
@@ -27,23 +27,20 @@ const BatchSidebar = ({ isOpen, onToggle }: { isOpen: boolean; onToggle: (open: 
     batchTxs.forEach((item) => deleteTx(item.id))
   }, [deleteTx, batchTxs])
 
+  // Close confirmation flow when batch is empty
+  const shouldExitFlow = !!txFlow && batchTxs.length === 0
+  useEffect(() => {
+    if (shouldExitFlow) {
+      setTxFlow(undefined)
+    }
+  }, [setTxFlow, shouldExitFlow])
+
   const onAddClick = useCallback(
     (e: SyntheticEvent) => {
       e.preventDefault()
       setTxFlow(<NewTxMenu />, undefined, false)
     },
     [setTxFlow],
-  )
-
-  const onDelete = useCallback(
-    (id: string) => {
-      const shouldCloseFlow = batchTxs.length === 1
-      deleteTx(id)
-      if (shouldCloseFlow) {
-        setTxFlow(undefined)
-      }
-    },
-    [deleteTx, batchTxs, setTxFlow],
   )
 
   const onConfirmClick = useCallback(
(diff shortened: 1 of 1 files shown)


src/components/batch/BatchSidebar/index.tsx
```diff
@@ -27,6 +27,14 @@ const BatchSidebar = ({ isOpen, onToggle }: { isOpen: boolean; onToggle: (open:
     batchTxs.forEach((item) => deleteTx(item.id))
   }, [deleteTx, batchTxs])
 
+  // Close confirmation flow when batch is empty
+  const shouldExitFlow = !!txFlow && batchTxs.length === 0
+  useEffect(() => {
+    if (shouldExitFlow) {
+      setTxFlow(undefined)
```
usame-algan, line 34: Can we somehow couple this to the user action i.e. add it to the `deleteTx` call instead of having it in a `useEffect`?

iamacook: See https://github.com/safe-global/safe-wallet-web/pull/2453#discussion_r1307455063.

katspaugh: Btw can there be a case when there’s some other tx flow open, not a batch modal?

src/components/batch/BatchSidebar/index.tsx
```diff
@@ -35,6 +35,17 @@ const BatchSidebar = ({ isOpen, onToggle }: { isOpen: boolean; onToggle: (open:
...
 
+  const onDelete = useCallback(
+    (id: string) => {
+      const shouldCloseFlow = batchTxs.length === 1
+      deleteTx(id)
+      if (shouldCloseFlow) {
+        setTxFlow(undefined)
+      }
+    },
+    [deleteTx, batchTxs, setTxFlow],
```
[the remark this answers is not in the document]
usame-algan: Is there a benefit of doing this in an effect over adding it to the onDelete handler?

src/components/batch/BatchSidebar/index.tsx
iamacook: > Btw can there be a case when there’s some other tx flow open, not a batch modal?

I've updated this accordingly.

src/components/batch/BatchSidebar/index.tsx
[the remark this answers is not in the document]
katspaugh: On delete is probably more robust. But at the same time deleting a tx and closing a mods are two completely different actions. Closing the modal is a side effect of having nothing to display. A matter of taste in the end, imperative vs declarative.

PR #2454, opened 2023-08-28
fix: disabled batch sidebar buttons when not owner

### LINKED ISSUE #2404: [Add to batch] non-owners and disconnected users issues
## Bug description

Opening the batch tx feature in the top bar, the "New tx button", "add new transaction" and "confirm batch" are enabled for non-owners and disconnected users

These users cannot execute the tx since the form will stop them, but still, if we gray out the "New transaction" button in the sidebar, we should do the same with the buttons in "Add to batch" feature

## Environment
 - Browser: Chrome
 - Wallet: Any
 - Chain: Any
 - Env: Current prod 1.16

## Steps to reproduce
 1. Go to a safe while being disconnected or an non-owner
 2. Open the Add to batch in the top bar
 3. Try to use the New tx button

 1. Add at least 1 tx to the batch with an owner
 2. Switch owner or disconnect
 3. Open the feature in the top bar
 4. Try to use the buttons Add a new tx or confirm batch

## Expected result
These buttons should be grayed out and a tooltip saying that you should be owner to do this, as we already do with the "New tx" button in the sidebar

## Obtained result
Buttons are available and you can start a tx as a non-owner or disconnected. The user cannot execute the tx since an error is shown in the review step

## Screenshots
image
image
image

iamacook:
## What it solves

Resolves #2404

## How this PR fixes it

The batch sidebar buttons are now disabled if the user is not connected or not an owner.

### Disabled
As a non-owner open the sidebar and observe the disabled buttons with tooltip when a batch _doesn't_ exists and when one _does_.

### Enabled
As an owner or non-owner open the sidebar and observe the enabled buttons tooltip when a batch _doesn't_ exists and when one _does_.

## Screenshots

batch sidebar buttons

## Checklist

### DIFF
--- a/src/components/batch/BatchSidebar/index.tsx
+++ b/src/components/batch/BatchSidebar/index.tsx
@@ -66,7 +66,7 @@ const BatchSidebar = ({ isOpen, onToggle }: { isOpen: boolean; onToggle: (open: 
               <BatchReorder txItems={batchTxs} onDelete={deleteTx} onReorder={onReorder} />
             </div>
 
-            <CheckWallet allowSpendingLimit>
+            <CheckWallet>
               {(isOk) => (
                 <Track {...BATCH_EVENTS.BATCH_NEW_TX}>
                   <Button onClick={onAddClick} disabled={!isOk}>
@@ -79,7 +79,7 @@ const BatchSidebar = ({ isOpen, onToggle }: { isOpen: boolean; onToggle: (open: 
 
             <Divider />
 
-            <CheckWallet allowSpendingLimit>
+            <CheckWallet>
               {(isOk) => (
                 <Track {...BATCH_EVENTS.BATCH_CONFIRM} label={batchTxs.length}>
                   <Button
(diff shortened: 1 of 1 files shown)


katspaugh (approved):

francovenica:
It looks good for non-owners and disconnected users, they see the buttons grayed out.

I dissagree with the spending limit owners being able to use the feature.

A person with an allowance cannot do any of this things: Put a transaction that use the allowance in a batch, nor confirm a batch since none of those listed tx will ever be a spending limit. So both actions "Add tx to batch" and "Confirm batch" cannot be used by a user with a spending limit

iamacook:
> I dissagree with the spending limit owners being able to use the feature.
> 
> A person with an allowance cannot do any of this things: Put a transaction that use the allowance in a batch, nor confirm a batch since none of those listed tx will ever be a spending limit. So both actions "Add tx to batch" and "Confirm batch" cannot be used by a user with a spending limit

Thanks for the thorough testing! I removed access to non-owner spending limit beneficaries in 4c9f0836585ea29af90dc4e580ddd09ff9f82542.

PR #2464, opened 2023-08-29
Fix: Execute batch -> Bulk execute

katspaugh:
Just a small wording change on the bulk execute button in the Queue. This will help disambiguate bulk execution and transaction batching.

Screenshot 2023-08-29 at 10 47 35


src/components/transactions/BatchExecuteButton/index.tsx
```diff
@@ -60,7 +60,7 @@ const BatchExecuteButton = () => {
             disabled={isDisabled}
             onClick={handleOpenModal}
           >
-            Execute batch{isBatchable && ` (${batchableTransactions.length})`}
+            Bulk execute{isBatchable && ` ${batchableTransactions.length} txs`}
```
usame-algan, line 63: Not sure if the user knows what `txs` are. Maybe something like `Bulk execute (2)`?

katspaugh: With `(2) txs` there's at least something to not know. With just `(2)` there's nothing to not know, so even less information.

iamacook: I would not use the term "txs" in the UI.

katspaugh: OK, how about the full form?
Screenshot 2023-09-03 at 13 01 23

iamacook (approved):

PR #2467, opened 2023-08-29
fix: use current/latest Safe deployments

### LINKED ISSUE #2207: Ensure that new safes created with the latest version supported by the services

iamacook:
## What it solves

Resolves #2207

## How this PR fixes it

All deployments now fallback to `LATEST_SAFE_VERSION` if `SafeInfo['version']` is either `null` or an unsupported version.

Instances of contract instantiations via the SDK fallback to `LATEST_SAFE_VERSION` if `SafeInfo['version']` is `null`. It internally handles unsupported versions.

### General
- The SDK initialises correctly for versions `1.0.0-1.3.0` on L1/L2.

### Transactions
- Queue batching works.
- Transaction batching works.
- Transaction validity checks work (e.g. "This transaction will most likely fail...").
- (MultiSend) simulation works.
- Relaying works.
- Off-chain message signing works.

### Safe creation
- Estimation works.
- (Relayed) Safe creation works (creating `1.3.0` Safes with `1.3.0` Fallback Handlers).

### Other
- (Un-)official fallback handlers are rendered correctly in the settings.
- Legacy versions `<1.0.0` load in the UI (not a deciding factor).

## Checklist


src/services/contracts/deployments.ts
```diff
@@ -0,0 +1,87 @@
...
+      version: version || LATEST_SAFE_VERSION,
+    }) ||
+    getDeployment({
+      version: LATEST_SAFE_VERSION,
+      network,
+    }) ||
+    getDeployment({
+      version: LATEST_SAFE_VERSION,
+    })
+  )
```
katspaugh, line 37: It's a little hard to to follow with ORs on two levels. I would try to avoid `version || LATEST_SAFE_VERSION` and make it more explicit. E.g.

```js
if (version !== null) {
  return  getDeployment({
       version,
       network,
     }) ||
     getDeployment({
       version
     })
} else {
  return  getDeployment({
       version: LATEST_SAFE_VERSION,
       network,
     }) ||
     getDeployment({
       version: LATEST_SAFE_VERSION
    })
  )
```

src/services/contracts/deployments.ts
```diff
@@ -0,0 +1,87 @@
...
+  const L2_VERSIONS = '>=1.3.0'
+
+  // We had L1 contracts on xDai, EWC and Volta so we also need to check version is after 1.3.0
+  return (
+    chain.l2 &&
+    semverSatisfies(
+      // Support unsupported versions
+      safeVersion || LATEST_SAFE_VERSION,
+      L2_VERSIONS,
+    )
```
katspaugh, line 55: `|| LATEST_SAFE_VERSION` – again, some inline logic that seems innocuous but might shoot you in the foot.

If would instead do an early return like this:

```js
// Unsupported safe version
if (safeVersion === null) {
  return chain.l2
}
```

katspaugh (changes requested):

katspaugh (approved):
Nice work! Looks like we had quite a few places that would have blown up after the safe-deployments update.

src/services/contracts/deployments.ts
katspaugh: Do I get it right that the Core SDK will fail to init then? It won't crash the app tho, will it?

iamacook: > Do I get it right that the Core SDK will fail to init then? It won't crash the app tho, will it?

Yes. It's [caught during initialisation](https://github.com/safe-global/safe-wallet-web/blob/dev/src/hooks/coreSDK/useInitSafeCoreSDK.ts#L30-L50).

francovenica:
Safes 0.1.0:
Error message by just connecting with a user (not even the owner of the safe)
image

0x6952bb94F2B6bA725868B44E2c654DA67C43Fcc6

francovenica:
For safes 1.3.0 everything is fine

I currently don't own safes 1.0.0. I used impersonator to start a transaction and see if the estimation works in the form and it does, but I cannot go further than that.

There is an error for safes 1.1.1, but this issue also happens in the staging environment. Should we let it through and fix it in another ticket?

The issue: The page crashes when you try to execute 2 tx's with the "Execute batch" action
This is the safe:
0x0993e9D2f35F46d8423C3FdFb26385BCb688815f
image

If it is decided to fix this issue in another ticket we can pass this one.

iamacook:
> There is an error for safes 1.1.1, but this issue also happens in the staging environment. Should we let it through and fix it in another ticket?
> 
> The issue: The page crashes when you try to execute 2 tx's with the "Execute batch" action This is the safe: 0x0993e9D2f35F46d8423C3FdFb26385BCb688815f image
> 
> If it is decided to fix this issue in another ticket we can pass this one.

Considering this is present in staging, I have filed [another issues](https://github.com/safe-global/safe-wallet-web/issues/2498) and will merge this now.

PR #2468, opened 2023-08-29
fix: don't track clicks on disabled elements

### LINKED ISSUE #2405: [GA tracking] Actions tracked when using grayed out buttons as non-owner
## Bug description

When you are a non-owner some buttons are grayed out in the UI, but we still track them in GA as if they were used and it doesn't seem to be a difference between clicking the button enabled or disabled
The buttons I found where:

Adding an owner
Removing an owner
Replacing an owner
Change policies
Adding a spending limit
Removing a spending limit
"Send" button in the assets table
"Send" from the address book

Also recently a ticket was created to grayout buttons for the "Add to batch" tx that, if that ticket is set to be fixed, those buttons should also not trigger GA tracking:
"Add to batch" from a tx from
"New transaction" from the Add to batch section that opens from the topbar
"Add a new tx" that shows when there are already batched tx in the menu displayed from the top bar
"Confirm batch" from that same menu

## Environment
 - Browser: Chrome
 - Wallet: any
 - Chain: any
 - Env: Current Prod build 1.16

## Steps to reproduce
 1. Connect with a wallet and go to a safe you don't own
 2. Visit the Setting > Setup page for owner related buttons
 3. Visit the Settings > Spending limits for the spending limit related buttons (here is a safe with spending limits created: eth:0x8675B754342754A30A2AeF474D114d8460bca19b)
 4. Visit the assets tab and hover a row to show the Send button
 5. Visit the address book page and add any entry to see the Send button
 6. Open the "Add to batch" feature in the top bar"
 7. Add a tx to the batch to see other buttons in that same menu

## Expected result
None of the grayed out buttons should track GA. 
If possible they should show a tooltip sayin that you are not an owner as we do in the "New transaction" button in the sidebar (if can be done in this ticket fine, if not I can open a new ticket for it)

## Obtained result
Every clicked button is tracked as if it was available

## Screenshots
GA tracking

iamacook:
## What it solves

Resolves #2405

## How this PR fixes it

Elements wrapped in `Track` now check whether the `children` is targetted. If not, it means that the clicked `target` is disabled and, as such, not tracked.

## How to test it

Click the following elements whilst disabled and observe no tracking event for them:

- Adding an owner
- Removing an owner
- Replacing an owner
- Change policies
- Adding a spending limit
- Removing a spending limit
- "Send" button in the assets table
- "Send" from the address book
- "Add to batch" from a tx from
- "New transaction" from the Add to batch section that opens from the topbar
- "Add a new tx" that shows when there are already batched tx in the menu displayed from the top bar
- "Confirm batch" from that same menu

## Checklist

### DIFF
--- a/src/components/common/Track/index.tsx
+++ b/src/components/common/Track/index.tsx
@@ -10,6 +10,11 @@ type Props = {
   label?: EventLabel
 }
 
+const shouldTrack = (el: HTMLDivElement) => {
+  const disabledChildren = el.querySelectorAll('*[disabled]')
+  return disabledChildren.length === 0
+}
+
 const Track = ({ children, as: Wrapper = 'span', ...trackData }: Props): typeof children => {
   const el = useRef<HTMLDivElement>(null)
 
@@ -20,9 +25,8 @@ const Track = ({ children, as: Wrapper = 'span', ...trackData }: Props): typeof 
 
     const trackEl = el.current
 
-    const handleClick = (e: MouseEvent) => {
-      // If clicked target is disabled, event.target will be trackEl
-      if (e.target !== trackEl) {
+    const handleClick = () => {
+      if (shouldTrack(trackEl)) {
         trackEvent(trackData)
       }
     }
--- a/src/components/common/Track/index.tsx
+++ b/src/components/common/Track/index.tsx
@@ -32,7 +32,7 @@ const Track = ({ children, as: Wrapper = 'span', ...trackData }: Props): typeof 
     return () => {
       trackEl.removeEventListener('click', handleClick)
     }
-  }, [children.type, el, trackData])
+  }, [el, trackData])
 
   if (children.type === Fragment) {
     throw new Error('Fragments cannot be tracked.')
--- a/src/components/common/Track/index.tsx
+++ b/src/components/common/Track/index.tsx
@@ -20,8 +20,11 @@ const Track = ({ children, as: Wrapper = 'span', ...trackData }: Props): typeof 
 
     const trackEl = el.current
 
-    const handleClick = () => {
-      trackEvent(trackData)
+    const handleClick = (e: MouseEvent) => {
+      // If clicked target is disabled, event.target will be trackEl
+      if (e.target !== trackEl) {
+        trackEvent(trackData)
+      }
     }
 
     // We cannot use onClick as events in children do not always bubble up
@@ -29,7 +32,7 @@ const Track = ({ children, as: Wrapper = 'span', ...trackData }: Props): typeof 
     return () => {
       trackEl.removeEventListener('click', handleClick)
     }
-  }, [el, trackData])
+  }, [children.type, el, trackData])
 
   if (children.type === Fragment) {
     throw new Error('Fragments cannot be tracked.')


src/components/common/Track/index.tsx
```diff
@@ -20,8 +20,11 @@ const Track = ({ children, as: Wrapper = 'span', ...trackData }: Props): typeof
 
     const trackEl = el.current
 
-    const handleClick = () => {
-      trackEvent(trackData)
+    const handleClick = (e: MouseEvent) => {
+      // If clicked target is disabled, event.target will be trackEl
+      if (e.target !== trackEl) {
```
katspaugh, line 25: What if a nested element intercepts the click? Will it always bubble up to `trackEl`?

iamacook: Yes. It seems to work nicely in my tests, even with `stopPropagation` on nested elements.

katspaugh (approved):

usame-algan:
I am still seeing events when I click on a disabled button. Depending on where I click the target could be the icon inside the button or another child element so the condition would still be true I think.
Screenshot 2023-08-29 at 15 41 40

iamacook:
> I am still seeing events when I click on a disabled button.

I'm converting this to a draft whilst I investigate this further.

iamacook:
@katspaugh, @usame-algan, I've adjusted the approach to [check whether child elements are disabled](https://github.com/safe-global/safe-wallet-web/blob/disabled-tracking/src/components/common/Track/index.tsx#L13-L16). Wdyt?

src/components/common/Track/index.tsx
```diff
@@ -10,6 +10,11 @@ type Props = {
   label?: EventLabel
 }
 
+const shouldTrack = (el: HTMLDivElement) => {
+  const disabledChildren = el.querySelectorAll('*[disabled]')
+  return disabledChildren.length === 0
```
katspaugh, line 15: That's gonna check all nested elements which is not ideal.
Maybe [`closest()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)?

iamacook:
> What about the previous idea we discussed of passing a `disabled` flag to the `Track` component? There are not that many buttons that can be disabled and if they are wrapped in `Track` we would have access to their disabled flag anyway so it would be a simple change. Calling `querySelectorAll` seems like a workaround / unconventional way of doing this in React.

Whilst I tend to agree, it is hard to enforce. As we have the input ref, I don't see it as a workaround.

I like [the idea of using `closest`](https://github.com/safe-global/safe-wallet-web/pull/2468#discussion_r1308952487) - will push it shortly.

src/components/common/Track/index.tsx
iamacook: `closest` traverses upwards. We want to check the `children`.

katspaugh (approved):

francovenica:
LGTM

I've checked the buttons in the list. 
Most of them now open the modal to connect if you are disconnected, so the "Open modal" event is tracked, which is expected.
If you are connected with no owner, nothing is tracked when you click the buttons/links as expected
If you are the owner the tracking works as normal, each button triggering the corresponding event as expected

PR #2470, opened 2023-08-29
fix: Remove ENS lookup from address book e2e test

usame-algan:
## What it solves

The address-book e2e test often fails if the provider call to resolve ENS takes too long. The ENS lookup is already covered by tests for the `AddressInput` and it is not possible to intercept the call via cypress so we can remove it there.

## How this PR fixes it

- Removes the ENS lookup from the address book e2e test


katspaugh (approved):
Thanks!

PR #2472, opened 2023-08-30
feat: add claim button to header

iamacook:
## What it solves

Implements claim button in header

## How this PR fixes it

The token allocation element in the header now includes a "New allocation" button according to [the designs](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?type=design&node-id=7969-133988&mode=design&t=6FeK1GEhF8jSaoo7-0). Clicking it will take you to the claiming app and, providing there is nothing to claim, it should not appear.

## How to test it

Open a Safe with an allocation and observe the new button. Claiming should make it dissappear and opening a Safe _without_ an allocation should not show it at all.

Clicking on the button should the following event: "Click on SEP5 allocation button".

## Screenshots

image

image

## Checklist

### DIFF
--- a/src/components/common/SafeTokenWidget/index.tsx
+++ b/src/components/common/SafeTokenWidget/index.tsx
@@ -2,7 +2,7 @@ import { SafeAppsTag, SAFE_TOKEN_ADDRESSES } from '@/config/constants'
 import { AppRoutes } from '@/config/routes'
 import { useRemoteSafeApps } from '@/hooks/safe-apps/useRemoteSafeApps'
 import useChainId from '@/hooks/useChainId'
-import useSafeTokenAllocation, { useSafeTokenBalance, type Vesting } from '@/hooks/useSafeTokenAllocation'
+import useSafeTokenAllocation, { useSafeVotingPower, type Vesting } from '@/hooks/useSafeTokenAllocation'
 import { OVERVIEW_EVENTS } from '@/services/analytics'
 import { formatVisualAmount } from '@/utils/formatters'
 import { Box, Button, ButtonBase, Skeleton, Tooltip, Typography } from '@mui/material'
@@ -28,7 +28,7 @@ const canRedeemSep5Airdrop = (allocation?: Vesting[]): boolean => {
     return false
   }
 
-  return !sep5Allocation.isRedeemed
+  return !sep5Allocation.isRedeemed && !sep5Allocation.isExpired
 }
 
 const SafeTokenWidget = () => {
@@ -38,7 +38,7 @@ const SafeTokenWidget = () => {
   const governanceApp = apps?.[0]
 
   const [allocationData, , allocationDataLoading] = useSafeTokenAllocation()
-  const [allocation, , allocationLoading] = useSafeTokenBalance(allocationData)
+  const [allocation, , allocationLoading] = useSafeVotingPower(allocationData)
 
   const tokenAddress = getSafeTokenAddress(chainId)
   if (!tokenAddress) {
--- a/src/components/common/SafeTokenWidget/index.tsx
+++ b/src/components/common/SafeTokenWidget/index.tsx
@@ -32,6 +32,8 @@ const canRedeemSep5Airdrop = (allocation?: Vesting[]): boolean => {
   return !sep5Allocation.isRedeemed && !sep5Allocation.isExpired
 }
 
+const SEP5_DEADLINE = '27.10.2023 10:00 UTC'
+
 const SafeTokenWidget = () => {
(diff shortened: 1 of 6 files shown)


src/hooks/useSafeTokenAllocation.ts
```diff
@@ -126,30 +142,26 @@ const fetchTokenBalance = async (chainId: string, safeAddress: string): Promise<
  * The Safe token allocation is equal to the voting power.
  * It is computed by adding all vested tokens - claimed tokens + token balance
  */
-const useSafeTokenAllocation = (): [BigNumber | undefined, boolean] => {
+export const useSafeTokenBalance = (allocationData?: Vesting[]): AsyncResult<BigNumber> => {
```
schmanu, line 145:
```suggestion
export const useSafeVotingPower = (allocationData?: Vesting[]): AsyncResult<BigNumber> => {
```

Should we rename it to `useSafeVotingPower`?
`useSafeTokenBalance` sounds to me as if it would just fetch the current balance in a hook.

src/components/common/SafeTokenWidget/index.tsx
```diff
@@ -2,31 +2,43 @@ import { SafeAppsTag, SAFE_TOKEN_ADDRESSES } from '@/config/constants'
...
 }
 
+const canRedeemSep5Airdrop = (allocation?: Vesting[]): boolean => {
+  const sep5Allocation = allocation?.find(({ tag }) => tag === 'user_v2')
+
+  if (!sep5Allocation) {
+    return false
+  }
+
+  return !sep5Allocation.isRedeemed
```
schmanu, line 31: Like this we would need to make a change once the deadline is reached.
I think also checking if the allocation is expired would fix that.
```suggestion
  return !sep5Allocation.isRedeemed && !sep5Allocation.isExpired
```

iamacook: I've added the extra check in ba34f38bd794ba7d4de458ff6aca8551b5d101d4.

src/hooks/useSafeTokenAllocation.ts
iamacook: I've renamed it accordingly in ba34f38bd794ba7d4de458ff6aca8551b5d101d4.

src/components/common/SafeTokenWidget/index.tsx
```diff
@@ -50,14 +63,37 @@ const SafeTokenWidget = () => {
             <Link href={url || ''} passHref legacyBehavior>
               <ButtonBase
                 aria-describedby="safe-token-widget"
-                sx={{ alignSelf: 'stretch' }}
+                sx={canRedeemSep5 ? { height: '42px !important' } : undefined}
                 className={css.tokenButton}
```
usame-algan, line 67: nit: You can also add this via css and avoid the !important e.g. `className={classnames(css.tokenButton, { [css.sep5]: canRedeemSep5 })`

usame-algan (approved):
Looks good 👍 I've tested this with my görli safes.

- The one that is elligible for sep5 had the new button in the header. After redeeming, the button disappears
- The one that is not elligible for sep5 has no button in the header

src/components/common/SafeTokenWidget/index.tsx
```diff
@@ -40,24 +55,51 @@ const SafeTokenWidget = () => {
...
     : undefined
 
+  const canRedeemSep5 = canRedeemSep5Airdrop(allocationData)
   const flooredSafeBalance = formatVisualAmount(allocation || BigNumber.from(0), TOKEN_DECIMALS, 2)
 
   return (
     <Box className={css.buttonContainer}>
-      <Tooltip title={url ? `Open ${governanceApp?.name}` : ''}>
+      <Tooltip
+        title={url ? (canRedeemSep5 ? `New airdrop until ${SEP5_DEADLINE}` : `Open ${governanceApp?.name}`) : ''}
```
katspaugh, line 64: "Claim any amount until 27.10 to be eligible!"

katspaugh (approved):

francovenica:
My only issue is the position of the green dot. I think it should be on the right bottom corner of the widget instead of being in the middle:
image

maybe something like this:
image


The rest is fine:
The button won't show if you don't have tokens to claim or if you already claimed the tokens
The tracking works fine as well
image

iamacook:
> My only issue is the position of the green dot. I think it should be on the right bottom corner of the widget instead of being in the middle

> If the position of the green dot is not a problem for the rest of the team we can pass the ticket to done

We discussed this and deemed it fine as is.

PR #2477, opened 2023-08-31
Fix: revert using wallet RPC for readonly RPC

katspaugh:
## What it solves

Fixes https://github.com/safe-global/safe-wallet-web/pull/2462#issuecomment-1700332666

Reverts a part of #2418.

### DIFF
--- a/src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
+++ b/src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
@@ -1,5 +1,5 @@
 import { type ReactElement, type SyntheticEvent, useContext, useState } from 'react'
-import { Button, CardActions, Divider } from '@mui/material'
+import { Box, Button, CardActions, Divider } from '@mui/material'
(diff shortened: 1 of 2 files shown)


iamacook (approved):

src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
```diff
@@ -137,17 +137,21 @@ const ExecuteForm = ({
...
+              {` To save gas costs, ${isCreation ? 'avoid creating' : 'reject'} this transaction.`}
+            </ErrorMessage>
           )
         )}
 
+        {submitError && (
+          <Box mt={1}>
+            <ErrorMessage error={submitError}>Error submitting the transaction. Please try again.</ErrorMessage>
+          </Box>
+        )}
```
katspaugh, line 153: This error must be shown separate of validation errors because the user can still submit a tx and this error has been only logged in the console.

PR #2478, opened 2023-08-31
Fix: Next 13 - useRouter() -> usePathName(), useParams()

katspaugh:
## What it solves
Fixes https://github.com/safe-global/safe-wallet-web/pull/2462#issuecomment-1700761074

In Next 13, we cannot use useRouter from the App component.


schmanu (commented):
When I navigate to any of my Safes in the sidebar from the welcome page and then click "New Transaction" the modal does not open.
I need to click the button twice now.
Edit: Unrelated to this change. But also present on the release deployment

PR #2493, opened 2023-09-06
Fix: use only current chain with multichain WalletConnect wallets

katspaugh:
## What it solves

Resolves "SafeProxy contract not deployed to the current network" error when trying to sign with some mobile wallets connected via WalletConnect.

IMG_3344

## How this PR fixes it

The root cause of the issue was that some wallets supported multiple chains at once, and onboard was confused as to which one to pick. I'm now initializing onboard only with the current Safe's chain, so it will always pick the right one when making provider calls.

When you open a Safe on another chain, a new onboard instance will be created with that chain's config.

## How to test it
1.
1. Install the Zerion or 1inch iOS apps
2. Connect to a Safe via WalletConnect
3. Sign a transaction, execute a transaction

2.
1. With different wallets, connect initially to the wrong chain
2. Sign a tx and it should switch the wallet's chain in the process

3.
1. Connect with a wallet to the right chain
2. Change the chain in the wallet to the wrong one
3. Sign a tx and it should switch the wallet's chain in the process

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "@web3-onboard/keystone": "^2.3.7",
     "@web3-onboard/ledger": "2.3.2",
     "@web3-onboard/trezor": "^2.4.2",
-    "@web3-onboard/walletconnect": "^2.4.5",
+    "@web3-onboard/walletconnect": "^2.4.6",
     "classnames": "^2.3.1",
     "date-fns": "^2.29.2",
     "ethereum-blockies-base64": "^1.0.2",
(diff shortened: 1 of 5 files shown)


schmanu (approved):
I could not verify that the error is fixed as I couldn't reproduce it on my end still.
But the code looks good.

schmanu (changes requested):
I now tested it with metamask mobile connected via WC2.
My metamask mobile app was connected to Polygon. And I switched the Safe interface to Görli. When I tried to sign on a Görli Safe I got this error when trying to switch chains.

Screenshot 2023-09-06 at 09 45 47

katspaugh:
It looks like this fix is breaking some other things, so I'll close it and revert the onboard modules instead as their update introduced this problem.

katspaugh:
Actually, downgrading onboard modules doesn't help.

PR #2505, opened 2023-09-12
add manta pacific network

samzkback:
Hi Team, 

We are manta pacific network, and we are trying to apply safe as the multi sig wallet in our mainnet.

Is there any guide on doing this, happy to talk with you!

Regards,
sam@manta.network

## What it solves

Resolves #

## Checklist


katspaugh:
We rely on https://github.com/safe-global/safe-modules-deployments for supported chains.
And then this frontend heavily relies on a few backend services which need to be deployed for each chain too.

Adding your chain in a mock file used only for unit tests won't make the app support this chain.

PR #2465, opened 2023-08-29
feat: Human-readable v2

usame-algan:
Depends on https://github.com/safe-global/safe-client-gateway/pull/671

## What it solves

Resolves #2449 

## ToDos

- [x] Add `HumanDescription` component to parse structured data response from CGW once https://github.com/safe-global/safe-client-gateway/pull/671 is merged
- [x] Check usage of `useTransactionType`, `TxType` and `TxInfo` and make sure they are still needed

## How this PR fixes it

- Adjusts the history and queue layout to make room for longer descriptions by combining the type and info column
- Removes now unused components `TxType` and `TxInfo` as well as the `useTransactionType` hook
- Adjusts fallback descriptions as outlined [here](https://www.notion.so/safe-global/UI-integration-9f437dd1d21744518fe79715607c7be7)

## How to test

The changes in this PR affect the history, queue, pending txs in the dashboard and the confirm tx screen

1. Open `/history` of a safe
2. Observe that there are only 4 columns now (nonce, description, date, status)
3. Transfers are now in the format of `Send/Receive <amount> to/from <address/name>`
4. Human Descriptions are more styled now e.g. ERC20 approvals or withdraws will highlighted amounts
5. Multisends are now in the format of `Call multiSend <on contract name><via safe app name> with <x amount> actions` while the existance of contract name and safe app name is optional so it could also be `Call multiSend with <x amount> actions`
6. If the method name of a transaction is known the description has the following format: `Call <method name> on <name/address>`
7. If no method name is known the description has the following format: `Interact with <contract address>`

## Screenshots
Screenshot 2023-09-08 at 13 24 57
Screenshot 2023-09-08 at 13 25 13
Screenshot 2023-09-08 at 13 25 28
Screenshot 2023-09-08 at 13 25 59

## Checklist
* [ ] I've tested the branch on mobile 📱
* [ ] I've documented how it affects the analytics (if at all) 📊
* [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #2518, opened 2023-09-18
feat: Add human readable transfers to history and queue

### LINKED ISSUE #2449: Human-readable descriptions v2
## What is the feature about

This feature is part of the Epic #2011
It aims to improve readability of human descriptions.

### CGW
- Return structured data instead of a simple string

### Web UI
- Implement components to parse each message type (address, number, word, tokenValue)
- Implement a `HumanDescription` component that returns a human description in JSX

## Links

https://www.notion.so/safe-global/Human-readable-tx-descriptions-8ede96b840584c96a8c58b546a59d921

usame-algan:
This is a smaller iteration on #2465 that only adds human-readable transfers to the history/queue but keeps everything else the same.

## What it solves

Resolves #2449 

## How this PR fixes it

- Show Human-readable ERC20, Native and NFT Transfers (Send + Receive), Safe Settings transactions in the transaction history and queue

## How to test it

1. Open the transaction history / queue / single tx view / dashboard pending txs
2. Observe Transfers are in a readable format (`Send <amount> to <address>`, `Receive <amount> from <address>`)
3. Observe other transactions are still in the same layout format

## Screenshots

Screenshot 2023-09-18 at 13 38 38
Screenshot 2023-09-18 at 13 38 50

## Checklist

### DIFF
--- a/src/components/transactions/TxSummary/index.tsx
+++ b/src/components/transactions/TxSummary/index.tsx
@@ -1,7 +1,7 @@
 import type { Palette } from '@mui/material'
 import { Box, CircularProgress, Typography } from '@mui/material'
 import type { ReactElement } from 'react'
-import { type Transaction, TransactionStatus } from '@safe-global/safe-gateway-typescript-sdk'
+import { type Transaction, TransactionInfoType, TransactionStatus } from '@safe-global/safe-gateway-typescript-sdk'
 
 import DateTime from '@/components/common/DateTime'
 import TxInfo from '@/components/transactions/TxInfo'
@@ -52,15 +52,18 @@ const TxSummary = ({ item, isGrouped }: TxSummaryProps): ReactElement => {
     : undefined
 
   const displayConfirmations = isQueue && !!submittedConfirmations && !!requiredConfirmations
+  const displayInfo = !tx.txInfo.richDecodedInfo && tx.txInfo.type !== TransactionInfoType.TRANSFER
 
   return (
     <Box
       className={`${css.gridContainer} ${
         isQueue
-          ? nonce && !isGrouped
+          ? displayInfo
             ? css.columnTemplate
-            : css.columnTemplateWithoutNonce
-          : css.columnTemplateTxHistory
+            : css.columnTemplateShort
+          : displayInfo
+          ? css.columnTemplateTxHistory
+          : css.columnTemplateTxHistoryShort
       }`}
       id={tx.id}
     >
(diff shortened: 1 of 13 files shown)


usame-algan:
I will close #2465 in case we decide to move forward with this PR.

src/components/transactions/TxSummary/index.tsx
```diff
@@ -52,15 +52,18 @@ const TxSummary = ({ item, isGrouped }: TxSummaryProps): ReactElement => {
...
+  const displayInfo = !tx.txInfo.richDecodedInfo && tx.txInfo.type !== TransactionInfoType.TRANSFER
 
   return (
     <Box
       className={`${css.gridContainer} ${
         isQueue
-          ? nonce && !isGrouped
+          ? displayInfo
             ? css.columnTemplate
-            : css.columnTemplateWithoutNonce
```
usame-algan, line 62: These 2 classes were identical so I removed the latter one. Maybe they were needed at some point?

katspaugh (approved):

francovenica:
- Suggestion: Probably we should put the variables in the tx in bold or something like that, since they get lost in the wording.
Since here is all in the same font, the "threshold of 1" part is easily missible: 
image

- For disable module I think we can also do a re-wording, like the one for Enable Module:
image

- Major issue:
Safe: https://human_readable_v2_small--walletweb.review-wallet-web.5afe.dev/transactions/history?safe=eth:0x8675B754342754A30A2AeF474D114d8460bca19b

For the mobile viewport, there are major differences between "human-readable" tx and the ones that remain as they were before (tx that don't have the "human-readable" flag) . Check these 2 types of tx and how the elements order themselves.
Snapshot: The last 2 tx elements where ordereded like that in a previous PR that attempted to fix mobile viewport issues
image

usame-algan:
Thanks for testing!

> Probably we should put the variables in the tx in bold or something like that, since they get lost in the wording.

This is something we would have to do in the CGW since those numbers are currently interpreted as normal text. Something to consider for the next iteration since it would affect all number types.

> For disable module I think we can also do a re-wording, like the one for Enable Module:

Nice find! This is also something we have to add in the CGW which we can do separately.

> For the mobile viewport, there are major differences between "human-readable" tx and the ones that remain as they were before

I've adjusted this so the human-readable rows break in the same way on mobile.

francovenica:
The formatting of the tx was fixed, it looks standarized

I've checked several safes with different tx and I haven't seen anything broken or that goes out of boundaries.
The tx that don't fall under the "Human-readable" flag look the same as they always did.

I've checked the History tab and single view tab. I was able to check some in the queue tab and the list of pending tx in the dashboard since I can only test tx I can create (in some safes there are tx that I cannot replicate myself), but it looks good.

PR #2520, opened 2023-09-18
fix: Update tx flow form when selecting a replacement nonce

usame-algan:
## What it solves

Part of #2383 

The form mode for the nonce form changed to `onTouched` with #2383 but selecting a nonce now doesn't update the form unless the user blurs the nonce input. This change calls `onBlur()` whenever the user selects something from the nonce replacement dropdown.

## How to test it

1. Open a safe with at least one transaction in queue
2. Create a new transaction
3. Go to the review screen
4. Observe seeing the Sign flow
5. Select the existing nonce from the nonce form
6. Observe seeing the Execute flow

## Checklist

### DIFF
--- a/src/components/tx-flow/SafeTxProvider.tsx
+++ b/src/components/tx-flow/SafeTxProvider.tsx
@@ -42,11 +42,12 @@ const SafeTxProvider = ({ children }: { children: ReactNode }): ReactElement => 
   const isSigned = safeTx && safeTx.signatures.size > 0
 
   // Recommended nonce and safeTxGas
-  const recommendedNonce = Math.max(safe.nonce, useRecommendedNonce() ?? 0)
+  const recommendedNonce = useRecommendedNonce()
+  const safeRecommendedNonce = recommendedNonce ? Math.max(safe.nonce, recommendedNonce) : undefined
   const recommendedSafeTxGas = useSafeTxGas(safeTx)
 
   // Priority to external nonce, then to the recommended one
-  const finalNonce = isSigned ? safeTx?.data.nonce : nonce ?? recommendedNonce ?? safeTx?.data.nonce
+  const finalNonce = isSigned ? safeTx?.data.nonce : nonce ?? safeRecommendedNonce ?? safeTx?.data.nonce
   const finalSafeTxGas = isSigned ? safeTx?.data.safeTxGas : safeTxGas ?? recommendedSafeTxGas ?? safeTx?.data.safeTxGas
 
   // Update the tx when the nonce or safeTxGas change
@@ -77,7 +78,7 @@ const SafeTxProvider = ({ children }: { children: ReactNode }): ReactElement => 
         setNonceNeeded,
         safeTxGas: finalSafeTxGas,
         setSafeTxGas,
-        recommendedNonce,
+        recommendedNonce: safeRecommendedNonce,
       }}
     >
       {children}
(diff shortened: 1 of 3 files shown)


katspaugh (approved):
Good catch!

schmanu (approved):
💯 Nice one!

francovenica:
So the main flow works fine, if the nonce is the next one to be executed the form allows execution, if is not only signing is possible.

Small issue:
There used to be a skeleton in the input that allow the load of the nonce without flickering. In this PR that skeleton is missing, so you can see the lowest nonces for a couple of seconds until it changes to the next nonce in the queue:
nonce input

usame-algan:
I've reverted part of #2383 to have the previous validation UX (checked without having to blur the input) and instead added an effect to update the form value when the recommended nonce changes and IF the user has not touched the field yet. I've also added back the loading state. @schmanu @katspaugh can you check if this is the desired behaviour?

src/components/tx-flow/SafeTxProvider.tsx
```diff
@@ -42,11 +42,12 @@ const SafeTxProvider = ({ children }: { children: ReactNode }): ReactElement =>
   const isSigned = safeTx && safeTx.signatures.size > 0
 
   // Recommended nonce and safeTxGas
-  const recommendedNonce = Math.max(safe.nonce, useRecommendedNonce() ?? 0)
+  const recommendedNonce = useRecommendedNonce()
+  const safeRecommendedNonce = recommendedNonce ? Math.max(safe.nonce, recommendedNonce) : undefined
```
katspaugh, line 46: IMHO this should be part of useRecommendedNonce

src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -104,12 +104,20 @@ const TxNonceForm = ({ nonce, recommendedNonce }: { nonce: string; recommendedNo
...
       [TxNonceFormFieldNames.NONCE]: nonce,
     },
   })
 
+  // Update the nonce in the form if the user hasn't changed the value manually
+  // and the recommendedNonce updated in the background
+  useEffect(() => {
+    if (formMethods.formState.touchedFields[TxNonceFormFieldNames.NONCE]) return
+
+    formMethods.setValue(TxNonceFormFieldNames.NONCE, recommendedNonce)
```
katspaugh, line 118: Why is needed? Can’t the input just be a controlled input?

usame-algan: Good point, I think the issue is that the `validate` calls `setNonce` automatically without user input so that the controlled input value `nonce` will never update to the recommended nonce. I've added an early return inside the `validate` function to skip validation if the input value is the same as `nonce` to avoid initial validation. This fixes the issue too and since we know that `nonce` is always valid it is ok to skip validation if the input value is the same.

src/components/tx-flow/SafeTxProvider.tsx
usame-algan: True, I've moved it into the hook.

schmanu (approved):
I tried it out and it works fine.

francovenica:
Looks good to me.
The skeleton to hide the nonce loading is working fine
Changing the nonce (by writing, by selecting from the dropdown or using the circular arrow) changes the form to execution or only-siging just fine.

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -165,10 +165,12 @@ export const useRecommendedNonce = (): number | undefined => {
...
 
   const [recommendedNonce] = useAsync(
-    () => {
+    async () => {
       if (!safe.chainId || !safeAddress) return
 
-      return getRecommendedNonce(safe.chainId, safeAddress)
+      const recommendedNonce = await getRecommendedNonce(safe.chainId, safeAddress)
+
+      return recommendedNonce ? Math.max(safe.nonce, recommendedNonce) : undefined
```
katspaugh, line 173: Please move this to `useRecommendedNonce`.

katspaugh (approved):

Issue #2510, opened 2023-09-14
Highlight queued transaction pending a signature


katspaugh:
Based on user feedback from Karpatkey.

It's hard to see in the queue which transactions you must sign. Their suggestion was to highlight them somehow. Or maybe even push them to the top.

Also, on the Dashboard, in the pending txs widget, we show top 5 queued transaction regardless of whether the connected wallet has already signed them. It should show only signable txs.

cc @liliiaorlenko

PR #2523, opened 2023-09-18
Feat: actionable pending txs on the Dashboard

katspaugh:
## What it solves

Part of #2510

## How this PR fixes it

Shows only signable and executable queued txs, not just any queued txs. Also displays a ✔️ or 🚀 button if a tx can be executed or signed.

Screenshot 2023-09-18 at 17 25 00

### DIFF
--- a/src/components/dashboard/PendingTxs/PendingTxListItem.tsx
+++ b/src/components/dashboard/PendingTxs/PendingTxListItem.tsx
@@ -5,12 +5,16 @@ import { useMemo } from 'react'
 import ChevronRight from '@mui/icons-material/ChevronRight'
 import type { TransactionSummary } from '@safe-global/safe-gateway-typescript-sdk'
 import { Box, SvgIcon, Typography } from '@mui/material'
-import { isMultisigExecutionInfo } from '@/utils/transaction-guards'
+import { isExecutable, isMultisigExecutionInfo, isSignableBy } from '@/utils/transaction-guards'
 import TxInfo from '@/components/transactions/TxInfo'
 import TxType from '@/components/transactions/TxType'
 import css from './styles.module.css'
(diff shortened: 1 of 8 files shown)


src/components/dashboard/PendingTxs/PendingTxListItem.tsx
```diff
@@ -19,6 +23,10 @@ type PendingTxType = {
 const PendingTx = ({ transaction }: PendingTxType): ReactElement => {
   const router = useRouter()
   const { id } = transaction
+  const { safe } = useSafeInfo()
+  const wallet = useWallet()
+  const signable = wallet ? isSignableBy(transaction, wallet.address) : false
+  const executable = wallet ? isExecutable(transaction, wallet?.address, safe) : false
```
usame-algan, line 29: I would suggest to keep the naming convention of the other places we use these functions: `canSign`, `canExecute`

usame-algan (approved):
Nice one 💯

francovenica:
Looks goood to me:

I tested in a safe 3/3:
The users can only see transactions that they can confirm. 
They can see tx that are 1 signature away from execution and the icon shown is the "rocket" to execute right away.
A fully signed tx shows up again with the "Rocket" icon to execute

francovenica:
Reopening this one because of an issue found:

I created a tx with an owner and he can see the tx in the list. Is expected that he shouldn't see the tx since he already signed during creation

The safe:
https://pending--walletweb.review-wallet-web.5afe.dev/home?safe=gor:0xA601D7b958A7Cdd82B5Cf5230D6C40876E9e27C4

The owner: (see if impersonator works)
gor:0x7724b234c9099C205F03b458944942bcEBA13408

Gif:
tx list

katspaugh:
When there are no signable/executable txs, it will display any queued txs, so it's expected.

PR #2524, opened 2023-09-18
Feat: attach safe address to analytics events

usame-algan:
## What it solves

Sends `safeAddress` to GTM

## How to test

- Safe address should be set when opening the app
- Safe address should be updated when switching safes within the same network
- Safe address should be updated when switching safes to a different network


katspaugh (approved):

francovenica:
Looks good to me.

The safe address is being tracked
Made sure it tracks only in URL that have the safe address in it. Neutral pages (safe creation, add safe, terms, privacy, etc...) have an empty value ( like this "" )
Switching safes in the same network or different networks tracks the correct safe you are switching into.

image

PR #2532, opened 2023-09-20
fix: boolean env vars

iamacook:
## What it solves

Resolves wrong env var values

## How this PR fixes it

`IS_PRODUCTION` and `IS_OFFICIAL_HOST` not check whether their respective env var is not `boolean` but `'true'` as env vars are strings.

### `IS_PRODUCTION`

The debug switches should display in the sidebar if `process.env.NEXT_PUBLIC_IS_PRODUCTION` is `false`, when `true`, not.

### `IS_OFFICIAL_HOST`

The `/imprint` page should return nothing if `process.env.NEXT_PUBLIC_IS_OFFICIAL_HOST` is `false`, when `true`, the imprint.

## Checklist


katspaugh (approved):

francovenica:
Aaron walked through this one since is only testeable in local environments. 
We played with the env variables and see the correspondand results:

If the env is set as true for prod, the debug switches don't show
If the env is set as true for official host, the imprint page is visible.

Looks good to me

PR #2535, opened 2023-09-21
Feat: native WalletConnect

katspaugh:
WalletConnect is now a first-class citizen in the app.

* most dapps will now work thanks to dynamic chain switching
* always available from the top toolbar
* supports multiple connections at the same time
* shows security warnings for risky dapps and bridges that don't play well with smart contract wallets
* supports `?wc=` query param in the URL

Screenshot 2023-10-23 at 16 24 03

Screenshot 2023-10-23 at 16 24 37

Specs: https://www.notion.so/safe-global/WalletConnect-as-first-class-citizen-c8578e20f9ac464e9c6f8cb3a973193b?pvs=4


src/components/walletconnect/SessionManager/index.tsx
```diff
@@ -61,6 +64,13 @@ const SessionManager = ({ sessions }: { sessions: SessionTypes.Struct[] }) => {
     return walletConnect.onSessionPropose(setProposal)
   }, [walletConnect])
 
+  // Connect to session present in URL
+  useEffect(() => {
+    if (!walletConnect || !wcUri) return
+
+    walletConnect.connect(wcUri).catch(setError)
```
katspaugh, line 71: I would put stuff like that in the context, not in the session manager component.

liliya-soroka:
1. UI issue on the macOS 13"
image

liliya-soroka:
could we check an issue https://github.com/safe-global/safe-wallet-web/issues/2689 as part of rc?

francovenica:
minor issue: we should put a more friendly error message if the user adds a random string of characters in the input.
image

francovenica:
According to one of the test cases, the connection to "risky" dapps should be shown as an error, but I only see the error in the console when I paste the URL. Right now there is no reaction from the app, like if the url pasted is ignored:

This is the "risky bridge" I'm using: https://cbridge.celer.network/1/56/USDC
image

francovenica:
This ticket will serve as the RC, so issues found during regression will be listed here:

1 - When signing a tx, the alert that " you will leave the tx form " shows. Even tho signing will take you out of the tx form as expected is also expected that the alert doesn't show up:
- Create a tx
- Sign with your wallet
- See the Alert popping up

- Queue a tx
- Then reject it
- When clicking the Reject tx button the alert shows up

regression 1

katspaugh:
Fixed, @francovenica! ✅

francovenica:
It is fixed for the regular send funds modal.

It still happens when you have the "Replace transaction" modal open. clicking any of the buttons "Send token" or "Reject transaction" also triggers the alert
image

francovenica:
The issues with the alert were fixed

I tried again 1inch, the issue I had with the tx not popping up and now it does pop up.

The regression went fine:
Connection with different wallet, with tx creation an execution,
Signing with policies 1/x and 2/x
Adding/removing/replacing tx
Safe creations
Send funds tx
Reject tx
App tx (tx builder with dummy contracts)
Spending limits creation and usage
Bulk tx execution
Batch tx list
Address book
Import/export
.....

I tried a few more dapps (cowswap, Yearn, balance) with a Eth safe 1/x with with those work fine

Still have the issue of some dapps remaining in a pending status after the tx was fully executed in the safe. I still not sure if those are from the dapps side or ours.

I think is impossible to know if every dapp outhere will work fine or not, so let's make sure after the deployment that the feature toggle works fine.

PR #2537, opened 2023-09-22
PoC: recovery via Delay Modifier

iamacook:
## What it solves

Adds ability to enable the Delay Modifier on the Safe via the settings with custom modules (recoverers). It is possible to later modify these recoverers from the same settings.

## Checklist


iamacook:
I'm going to close this but we will still use it as reference.

PR #2542, opened 2023-09-25
Fix: update privacy policy

katspaugh:
Update the Privacy policy page wrt tracking and anonymized tracking.


tschubotz (approved):
Changes look good to me and are as discssued via Slack/Gdocs

iamacook (approved):

Issue #2011, opened 2023-05-22
[EPIC] Human-readable transactions


usame-algan:
## What is the feature about

Ethereum transactions are complex and there’s no easy way to summarize what a single transaction does in one sentence. We propose to create an API to decode contract calls into a short and simple piece of text.

## ToDos

- [x] #2448
- [x] #2449

## Designs/sketches

TBD

## Links

https://www.notion.so/safe-global/Human-readable-tx-descriptions-8ede96b840584c96a8c58b546a59d921

mgarciap:
The Notion link is not working: `You do not have access to this page.`
https://www.notion.so/safe-global/Human-readable-tx-descriptions-8ede96b840584c96a8c58b546a59d921

PR #2543, opened 2023-09-25
fix: Add AB test for human-readable transactions

usame-algan:
## What it solves

Part of #2011 

## How this PR fixes it

- Adds A/B testing for human-readable descriptions in the history/queue
- Adds a new event for `Fetching transaction details`
- Fixes slight (4px difference) column alignment issue on the date between human-readable transactions and other ones

## How to test it

1. Open a Safe
2. Observe there is a new local storage key called `SAFE_v2__AB_human-readable`
3. Navigate to the history/queue
4. Switch the key from true to false or vice versa
5. Observe the history/queue changes layout
6. Remove the local storage key and reload the page until the value changes, if its false observe no abTest key in GTM otherwise observe the abTest key in GTM saying `human-readable`
7. Observe there is a new event `Fetch transaction details` when opening a transaction in the history for the first time

## Screenshots
Screenshot 2023-09-25 at 16 33 03
Screenshot 2023-09-25 at 16 33 18

## Checklist


src/services/tracking/abTesting.ts
```diff
@@ -1,7 +1,9 @@
 /**
  * Holds current A/B test identifiers.
  */
-export const enum AbTest {}
+export const enum AbTest {
+  HUMAN_DESCRIPTION = 'human_description',
```
katspaugh, line 5:
```suggestion
  HUMAN_DESCRIPTION = 'human-readable',
```
I would look for this in GA if I were looking for it in a month.

katspaugh (approved):
Thanks!

francovenica:
Looks good to me

Changing the AB_human_readable key in the LS shows/hides the human readable
I tried deleting the key and refreshing the page to see that sometimes it creates it with a value of true and false

I've checked in the GTM that the key "abTest = undefined" if the "AB_human_readable" key is false and as "abTest = human-readable" if the "AB_human_readable" key is true
image
image

PR #2548, opened 2023-09-26
Fix: prevent Safe Apps tracking on non-Safe App pages

katspaugh:
Safe Apps track all RPC requests a lot of which are triggered from the Governance App on the Dashboard. Since this doesn't reflect actual Safe App usage, it should be disabled.

## How to test

* Safe Apps tracking events should not be triggered on the dashboard anymore
* Events on the /apps page, as well as individual Safe App pages should be still emitted like before.


usame-algan (approved):
🚀

katspaugh:
@dasanra good with you?

francovenica:
Looks good to me

Visiting the dashboard doesn't make any GTM calls with for safe-apps.
Visiting the apps themselves makes the GTM calls for safe-apps just fine.

PR #2556, opened 2023-09-27
fix: only delete all preferences per chain

iamacook:
## What it solves

Resolves https://github.com/safe-global/safe-wallet-web/pull/2546#issuecomment-1736499323

## How this PR fixes it

`deleteAllPreferences` is has been replaced with `deleteAllChainPreferences`, which clears preferences according to a provided `chainId` according to the device unregistration (also per `chainId`).

A loading indicator has also been added to the global preferences "Save" button.

## How to test it

During the following registrations, a loading indicator should appear in the "Save" button:

1. Ensure multiple Safes are registered across differing chains.
2. Navigate to the global notification preferences.
3. Unregister a Safe.
4. Observe _only_ the preferences from that Safe removed.
---

1. Ensure multiple Safes are registered across differing chains.
2. Navigate to the global notification preferences.
3. Unregister _all_ Safes on one network.
6. Observe the Safes on other chain(s) retain their preferences.

## Checklist

### DIFF
--- a/src/services/exceptions/ErrorCodes.ts
+++ b/src/services/exceptions/ErrorCodes.ts
@@ -45,8 +45,8 @@ enum ErrorCodes {
   _702 = '702: Failed to remove from local/session storage',
   _703 = '703: Error importing an address book',
   _704 = '704: Error importing global data',
-  _705 = '704: Failed to read from IndexedDB',
-  _706 = '704: Failed to write to IndexedDB',
+  _705 = '705: Failed to read from IndexedDB',
+  _706 = '706: Failed to write to IndexedDB',
 
   _800 = '800: Safe creation tx failed',
   _801 = '801: Failed to send a tx with a spending limit',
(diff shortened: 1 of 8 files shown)


src/services/exceptions/ErrorCodes.ts
```diff
@@ -45,8 +45,8 @@ enum ErrorCodes {
   _702 = '702: Failed to remove from local/session storage',
   _703 = '703: Error importing an address book',
   _704 = '704: Error importing global data',
-  _705 = '704: Failed to read from IndexedDB',
-  _706 = '704: Failed to write to IndexedDB',
+  _705 = '705: Failed to read from IndexedDB',
+  _706 = '706: Failed to write to IndexedDB',
```
iamacook, line 49: I noticed that these were labelled incorrectly.

usame-algan (approved):

francovenica:
Looks good.

Checked that only the checkboxes you check or unchecked are saved
Checked that last fix regarding the signature counter, now it counts properly how many signatures per network you need.

PR #2563, opened 2023-09-28
fix: nonce field styles

### LINKED ISSUE #2560: Show a warning/prompt when nonce is above the latest nonce
Show a warning/confirmation request when the entered nonce is too high.

* If the user has entered a nonce that's higher than the recommended nonce, show a warning.
* Also, if the recommended nonce from the backend is too far ahead of the Safe nonce due to an unexpected bug, we should also warn the user
* ~In both cases, do not let the tx be submitted w/o a confirmation.~

iamacook:
## What it solves

Partially resolves #2560

## How this PR fixes it

The recommended nonce is always shown and the padding applied for the reset nonce button is included in the field width.

## How to test it

Open the transaction flow and observe the recommended nonce always shown when a non-existent recommended nonce exists, as well as the value always visible.

## Screenshots

recommended-nonce

## Checklist

### DIFF
--- a/src/components/tx-flow/common/TxNonce/index.tsx
+++ b/src/components/tx-flow/common/TxNonce/index.tsx
@@ -14,6 +14,7 @@ import {
   ListSubheader,
   type ListSubheaderProps,
 } from '@mui/material'
+import { createFilterOptions } from '@mui/material/Autocomplete'
 import { Controller, useForm } from 'react-hook-form'
 
 import { SafeTxContext } from '@/components/tx-flow/SafeTxProvider'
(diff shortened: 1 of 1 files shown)


katspaugh (approved):

francovenica:
Issue:

If the nonce is lower than the exepected one the dropdown will show you what is the nonce you must select, which is fine, the problem is that if that nonce you are writing matches the exepected  nonce, it will show it twice in the dropdown
In this snapshot, the safe has no tx in the queue and the next nonce to execute is 158, so if you write "1" or "15" or "8" it will show the recomended nonce twice in the dropdown
image

francovenica:
The issue was fixed. Looks good now:
image

PR #2564, opened 2023-09-29
Refactor: map props to hooks

katspaugh:
To make our components more testable, we need to rely less on external hooks and more on props.

To this end, we're introducing a function that maps props to hooks. E.g.:

```js
// "Dumb" component that takes props
type AddressBookTableProps = {
  chain?: ChainInfo
  setTxFlow: TxModalContextType['setTxFlow']
}

export const AddressBookTable = ({ chain, setTxFlow }: AddressBookTableProps) => {
  //...
}

// A "smart" wrapper for that component that maps additional props to hooks
export default madProps(AddressBookTable, {
  chain: useCurrentChain,
  setTxFlow: () => useContext(TxModalContext).setTxFlow,
})
```

With this approach, the "dumb" `AddressBookTable` component can be more easily tested w/o having to mock those hooks.


src/utils/__tests__/mad-props.test.tsx
usame-algan, line 1: I would suggest adding an additional test case for when only part of the props is passed via `madProps` and the rest through `MappedComponent` directly to make sure `Omit` works correctly.

src/utils/mad-props.tsx
```diff
@@ -0,0 +1,32 @@
+import type { ComponentType, FC } from 'react'
+import React, { memo } from 'react'
+
+type HookMap<P> = {
+  [K in keyof P]?: () => P[K]
+}
+
+const madProps = <P extends Record<string, unknown>, H extends keyof P>(
```
usame-algan, line 8: Can we have a comment explaining what this function is for?

usame-algan (commented):
I tried testing one of the wrapped components but get a type error when trying to `render(<ExportDialog />)`

src/utils/__tests__/mad-props.test.tsx
katspaugh: The second test case in the test file I added tests that, unless I misunderstand your comment. Please note that it's not possible to pass fewer or more props to a mapped component. You'd have to use the original component in that case.

usame-algan: Thanks!

usame-algan (approved):

PR #2565, opened 2023-09-29
Chore: fix CI by forcing postinstall

katspaugh:
`postinstall` isn't being triggered for some reason. This started happening w/o us changing any code. Something related to node-gyp.

This is a temporary workaround for #2566. Root cause still under investigation.


iamacook (approved):
Let's also create an issue to revisit this when we get the chance.

PR #2574, opened 2023-10-02
fix: disabled "Enable all" button height

iamacook:
## What it solves

Resolves style inconsistency for "Enable all" button height when disabled.

## How this PR fixes it

The `CheckWallet` component now wraps the button container element instead of just the "Enable all" button.

## How to test it

Note: this seems to happen only in Chrome.

1. Ensure no wallet is connect.
2. Ensure the Safe has not previously dismissed the push notification banner.
3. Open the Safe.
4. Observe the "Enable all" button (disabled) having the same height as the "Customize" button in the notification banner.

## Screenshots

image

## Checklist


katspaugh (approved):

francovenica:
Is barely noticeable but ok:

This is how it looks in chrome in firefox:
image
image

The button enabled still looks good:
image

PR #2575, opened 2023-10-02
fix: only request new notification signatures

iamacook:
## What it solves

Resolves signature requests for Safes that are already registered for notifications.

## How this PR fixes it

The logic within the banner has been fixed, as well as more unit test coverage added.

## How to test it

1. Add a Safe on chain A and register for notifications with it.
2. Add a Safe on chain B and register for notifications with it.
3. Observe only one signature request.

## Checklist


francovenica:
Looks good.

The Enable all is now asking to sign just by the amount of networks that have yet not signed: so if you have a safe added with notifications enabled, and add a new safe for a different network it will only ask to sign once more for that new added safe.

Also noticed that in dev "enable all" wasn't working if several safes were already added and had the notifications not enabled yet. In this PR that part works fine now

PR #2576, opened 2023-10-02
Fix: notifications popup: don't hide just one item

katspaugh:
Avoid hiding a single notification behind a collapsible box. It occupies exactly the same space.
Plus, the wording "1 other notification<b>s</b>" is wrong.

## How to test
1.
* Trigger 5 notitications
* Before: the 5th one was hidden, and it said "1 other notifications"
* After: all 5 are shown

2.
* Trigger 6 or more notitications
* Only 4 are shown and it says "2 other notifications"

### DIFF
--- a/src/components/transactions/TxType/styles.module.css
+++ b/src/components/transactions/TxType/styles.module.css
@@ -1,7 +1,7 @@
 .txType {
   display: flex;
   align-items: center;
-  flex-wrap: wrap;
+  flex-wrap: nowrap;
   gap: var(--space-1);
   color: var(--color-text-primary);
 }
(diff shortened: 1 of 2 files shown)


iamacook (approved):

src/components/transactions/TxType/styles.module.css
```diff
@@ -1,7 +1,7 @@
 .txType {
   display: flex;
   align-items: center;
-  flex-wrap: wrap;
+  flex-wrap: nowrap;
```
katspaugh, line 4: Unrelated to notification, it fixes this:
Screenshot 2023-10-02 at 15 18 13

francovenica:
So it seems that "1 other notification" will not show up anymore, it shows 5 notifications and if there are 6 it will show 4 and "2 more notifications" to display.

Looks good to me

PR #2584, opened 2023-10-04
fix: enable notifications per chain

iamacook:
## What it solves

Multiple signing windows and unsupported added Safes breaking registration.

## How this PR fixes it

Enabling "all" notifications via the banner is now per chain. Only one signing window will appear and Safes on that chain will be registered for notifications.

On the global notifications settings page, the notifiable Safes are filtered according to the chains of the config service.

### Enabling per chain

1. Ensure there are added Safes on differing networks and that no notification banner has been previously dismissed.
2. Open a network with added Safes and enable the notifications via the banner.
3. Observe only one signing window appear.
4. Switch to another network with added Safes and repeat.

### Enabling globally

1. Add an unsupported chain with Safes to the `SAFE_v2__addedSafes` key in the `localStorage`.
2. Navigate to the global notification settings.
3. Select all Safes or the supported Safes.
4. Click "Save" and observe a successful registration.

## Screenshots

image

## Checklist

### DIFF
--- a/src/services/analytics/events/push-notifications.ts
+++ b/src/services/analytics/events/push-notifications.ts
@@ -38,7 +38,7 @@ export const PUSH_NOTIFICATION_EVENTS = {
   },
   // User enabled all notifications from banner
   ENABLE_ALL: {
-    action: 'Enable all notifications',
+    action: 'Enable all notifications on chain',
     category,
   },
   // User opened Safe notification settings from banner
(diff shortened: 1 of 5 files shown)


src/services/analytics/events/push-notifications.ts
```diff
@@ -38,7 +38,7 @@ export const PUSH_NOTIFICATION_EVENTS = {
   },
   // User enabled all notifications from banner
   ENABLE_ALL: {
-    action: 'Enable all notifications',
+    action: 'Enable all notifications on chain',
```
katspaugh, line 41: Let's not change it now, otherwise we'll end up with orphan events. Also the action should be short.

katspaugh (approved):

PR #2597, opened 2023-10-06
Fix: collapse notifications if more than max + 1

katspaugh:
## What it solves

A fix for #2576, the +1 logic wasn't checked in enough places.


iamacook (approved):

francovenica:
Looks good

I get to see the 5 notifications in a single page
The "x other notifications" shows fine after the 6th notification and on.

PR #2598, opened 2023-10-06
fix: track banner once

iamacook:
## What it solves

Resolves over-tracking of notification banner display

## How this PR fixes it

Tracking of the notification banner now occurs whenever it is shown. It is done within a component to prevent complex caching of tracking. If the `showBanner` flag is true, the component renders and tracks.

## How to test it

Open a Safe that has not been registered for notifications and observe only one "Display notification banner" event. If the banner is not dismissed, navigating back to the Safe should dispatch the tracking again. However, dismissing the banner should prevent any future tracking.

## Checklist

### DIFF
--- a/src/components/settings/PushNotifications/PushNotificationsBanner/index.tsx
+++ b/src/components/settings/PushNotifications/PushNotificationsBanner/index.tsx
@@ -1,7 +1,7 @@
 import { Button, Chip, Grid, SvgIcon, Typography, IconButton } from '@mui/material'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
-import { useCallback, useEffect } from 'react'
+import { useCallback, useEffect, useRef } from 'react'
 import type { ReactElement } from 'react'
 
 import { CustomTooltip } from '@/components/common/CustomTooltip'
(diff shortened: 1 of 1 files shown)


src/components/settings/PushNotifications/PushNotificationsBanner/index.tsx
```diff
@@ -86,6 +86,21 @@ export const _getSafesToRegister = (
...
 
+const TrackBanner = (): null => {
+  const hasTracked = useRef(false)
+
+  useEffect(() => {
+    if (hasTracked.current) {
+      return
+    }
+
+    trackEvent(PUSH_NOTIFICATION_EVENTS.DISPLAY_BANNER)
```
katspaugh, line 97: I suggest we rename the event action for this event to differentiate it from the events we've erroneously tracked so far.

src/components/settings/PushNotifications/PushNotificationsBanner/index.tsx
```diff
@@ -86,6 +86,21 @@ export const _getSafesToRegister = (
...
+    hasTracked.current = true
+  }, [])
```
katspaugh, line 99: Dont' you need to pass hasTracked here?

katspaugh (approved):

francovenica:
Looks good

The notification pop up is tracked once when a safe is added
Is tracked everytime that safe is opened if the pop up wasn't dissmissed before
After being dismissed there is no more tracking of the popup

PR #2599, opened 2023-10-06
Fix: rm forced Beamer NPS survey

katspaugh:
## What it solves

As per @kirkkonen's request, the NPS survey shouldn't be force-triggered on tx completion anymore. It will be controlled from the Beamer admin instead.


iamacook (approved):

katspaugh:
No need to test this one, so I'll merge it directly.

PR #2602, opened 2023-10-09
fix: Track transaction item open events

usame-algan:
## What it solves

Fix for #2543 

This will significantly reduce the number of events that are tracked due to users opening a single transaction view (which is already tracked by a page view event).

## How this PR fixes it

- Removes `Fetch transaction details` event
- Tracks `Expand transaction item` event whenever the user expands a transaction item

## How to test it

1. Open a transaction history
2. Click on a transaction to expand
3. Observe that the event is tracked
4. Close the transaction and open it again
5. Observe that the event is tracked again
6. Open a single transaction view
7. Observe that no event for fetch tx details is tracked

### DIFF
--- a/src/components/transactions/TxDetails/index.tsx
+++ b/src/components/transactions/TxDetails/index.tsx
@@ -29,7 +29,6 @@ import { DelegateCallWarning, UnsignedWarning } from '@/components/transactions/
 import Multisend from '@/components/transactions/TxDetails/TxData/DecodedData/Multisend'
 import useSafeInfo from '@/hooks/useSafeInfo'
 import useIsPending from '@/hooks/useIsPending'
-import { trackEvent, TX_LIST_EVENTS } from '@/services/analytics'
 
 export const NOT_AVAILABLE = 'n/a'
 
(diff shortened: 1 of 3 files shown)


src/components/transactions/TxDetails/index.tsx
```diff
@@ -126,7 +126,9 @@ const TxDetails = ({
 
   const [txDetailsData, error, loading] = useAsync<TransactionDetails>(
     async () => {
-      trackEvent(TX_LIST_EVENTS.FETCH_DETAILS)
+      if (!txDetails) {
```
katspaugh, line 129: That's a negation, isn't it? Why do we even track this event? We should track clicks on the accordeon, not the fetch.

usame-algan: I also considered tracking this event on the accordion click but with the current structure it doesn't seem possible since the fetch is in a nested component and with the accordion click we can only know if it was opened or closed but not if we already fetched the data before. Do you think we should track the first accordion open in general regardless of the fetch status?

katspaugh: Yes. I think we should track the user intention, not how our app interpreted it.

usame-algan: I changed it so we track the accordion open once. I also renamed the event to better fit the new purpose.

katspaugh (approved):

francovenica:
Ok so I did't understand at first since I don't remember at heart what events we track with those actions, so I'll leave this comment here for the sake of documentation on "how it was" vs "how it is right now"

This event used to track whenever you expand a tx item and it would only track once. Also the same event tracks whenever you open a tx in the individual view with its deep link.
image


I've checked now that a different event is tracked when the tx is expanded and no "Fetch" event is tracked when expanding a tx item nor when it is opened with a deep  link
image
image

Looks good to me

PR #2604, opened 2023-10-09
Fix: refactor + don't show push notifications banner if wallet not conencted

katspaugh:
Also don't track banner dismissal on enable all/customize.

## What it solves

The banner was previously shown regarless of whether you're connected which clashed with the onboard popup.

### DIFF
--- a/src/components/settings/PushNotifications/PushNotificationsBanner/index.tsx
+++ b/src/components/settings/PushNotifications/PushNotificationsBanner/index.tsx
@@ -210,7 +210,7 @@ export const PushNotificationsBanner = ({ children }: { children: ReactElement }
   // On dismiss, track the event and dismiss the banner
   const onDismiss = () => {
     trackEvent(PUSH_NOTIFICATION_EVENTS.DISMISS_BANNER)
-    dismissBanner
+    dismissBanner()
   }
 
   // On enable all, track the event and enable all notifications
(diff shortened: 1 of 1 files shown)


usame-algan (approved):
I like the approach of extracting some of the logic into local hooks. It would be nice to have tests for them, especially `useShowBanner` to make sure the change with the connected wallet works without manually testing it but I will approve the PR.

I've checked that the dismiss event only happens when the user closes the banner but not when enabling all or customizing. The other 2 events are still being sent correctly.

katspaugh:
I won't be merging this because we need to release these and other fixes today, so I'll make a more focused PR with just the fixes.
@iamacook when you're back, please take a look and tell me what you think. I'd like us to follow this approach as a general rule.

src/components/settings/PushNotifications/PushNotificationsBanner/index.tsx
```diff
@@ -86,69 +87,146 @@ export const _getSafesToRegister = (
...
+                  </Button>
+
+                  <Link passHref href={{ pathname: AppRoutes.settings.notifications, query }} onClick={onCustomize}>
+                    <Button variant="outlined" size="small" className={css.button}>
+                      Customize
+                    </Button>
+                  </Link>
+                </div>
+              )}
+            </CheckWallet>
```
iamacook, line 146: As we are no longer showing the banner when no wallet is connected, we don't need to wrap this in `CheckWallet`.

iamacook (commented):
I like the approach.

We should fix/extend the current test to cover the added flag, as well as test the hooks in isolation.

PR #2618, opened 2023-10-12
Takehome assignment: wrap and unwrap ETH

jmealy:
## What it solves
This PR is to complete the functionality of the WrappedEth component as part of a take home interview.

Caveats:
- It only works with Goerli testnet
- I did not add tests because I had already spent more than the recommended time on it.

## How this PR fixes it
My changes do the following
- Display the balance of ETH and WETH a user has in their safe. This is done using the `balanceOf` function of the WETH smart contract.
- Allow a user to wrap a specified amount of ETH. This uses the `deposit` function of the WETH contract.
- Allows a user to unwrap a specified amount of WETH using the `withdraw` function of the WETH contract.

The smart contract used is the Goerli version of the WETH contract:
https://goerli.etherscan.io/address/0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6

## How to test it
- Set up a safe using Goerli.
- Get some GOR from a faucet.
- Navigate to `/home`
- You should see the amount of GOR in your safe displayed [1]
- Enter an amount of ETH to wrap and click the `Wrap`  button. Go through the safe transaction flow.
- When you return to the home screen the specified amount of ETH should be converted to WETH [2].
- Do the same using the unwrap box to convert WETH back to ETH.

## Screenshots
[1] Screenshot from 2023-10-12 09-10-43 (1)
[2] Screenshot from 2023-10-12 09-25-30

### DIFF
--- a/src/components/dashboard/WrappedEth/index.test.tsx
+++ b/src/components/dashboard/WrappedEth/index.test.tsx
@@ -1,0 +1,7 @@
+import React from 'react'
+import { render } from '@/tests/test-utils'
+import WrappedEth from '.'
+
+it('renders without errors', () => {
+  render(<WrappedEth />)
+})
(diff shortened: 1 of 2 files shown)


katspaugh:
Thanks James! We've reviewed it and I will close the PR now to not spoil it for other candidates. Cheers!

PR #2620, opened 2023-10-12
fix: Enable 1-click safe create for social login

### LINKED ISSUE #2458: [Seedless-Onboarding]: Add feature toggle
## What is the feature about
We want to be able to enable this feature on a per chain basis.
Therefore we need to add a new feature toggle `MPC_SIGNER` which we will initially only enable on Gnosis Chain / Staging Görli.

Alternatively we could limit this feature by enabling / disabling the Social signer wallet type for the chain.

## Requirements
- [ ] Add feature toggle on staging
- [ ] Only offer social onboarding if feature toggle is on
- [ ] Make network selector read-only if connected via Social Login
- [ ] Show Login button disabled on other networks
- [ ] Skip to review screen on safe creation when social login is used

usame-algan:
## What it solves

Resolves #2458 

## How this PR fixes it

- Disables Social Login when not on Gnosis Chain and shows an info text
- Skips to the review screen on safe creation when using Social Login
- Displays simplified relay sponsor view in review screen
- Disables network selector when using Social Login

## ToDos

- [x] Disable safe creation if no relays left

## How to test it

1. Open the welcome page
2. Switch to an unsupported network
3. Observe the Social Login is disabled with an info text
4. Switch to Gnosis chain and log in via Google
5. Observe the Network selector is disabled
6. Enter the safe creation flow
7. Observe jumping right to the review screen
8. No execution method selector is shown and instead a simple message saying Free
9. Disconnect from Google
10. Connect via MM
11. Network Selector should work
12. Enter the safe creation
13. First step of the safe creation should open

## Screenshots

Screenshot 2023-10-12 at 10 47 18
Screenshot 2023-10-12 at 10 55 47

## Checklist

### DIFF
--- a/src/components/new-safe/create/steps/ReviewStep/index.test.tsx
+++ b/src/components/new-safe/create/steps/ReviewStep/index.test.tsx
@@ -7,7 +7,7 @@ import { type ConnectedWallet } from '@/services/onboard'
 import { ONBOARD_MPC_MODULE_LABEL } from '@/services/mpc/module'
 
 const mockChainInfo = {
-  chainId: '5',
+  chainId: '100',
   l2: false,
   nativeCurrency: {
     symbol: 'ETH',
@@ -15,7 +15,7 @@ const mockChainInfo = {
 } as ChainInfo
 
 describe('NetworkFee', () => {
-  it('displays the total fee if not social login', () => {
+  it('should display the total fee if not social login', () => {
     jest.spyOn(useWallet, 'default').mockReturnValue({ label: 'MetaMask' } as unknown as ConnectedWallet)
     const mockTotalFee = '0.0123'
     const result = render(<NetworkFee totalFee={mockTotalFee} chain={mockChainInfo} willRelay={true} />)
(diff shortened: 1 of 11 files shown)


schmanu (commented):
You need to adjust the `MPCLogin.test.tsx `. You need to mock the currentChain to be on Gnosis Chain such that the buttons are enabled.
You could also add a test case for non supported networks to check that the button is disabled and the hint is shown.

schmanu (approved):

src/components/new-safe/create/steps/ReviewStep/index.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+const mockChainInfo = {
+  chainId: '5',
+  l2: false,
+  nativeCurrency: {
+    symbol: 'ETH',
+  },
+} as ChainInfo
+
+describe('NetworkFee', () => {
+  it('displays the total fee if not social login', () => {
```
schmanu, line 18: Nit: I think usually we chose this form for our test expectations.

```suggestion
  it('should display the total fee if not social login', () => {
```

schmanu (approved):
Just 2 small nits in the test. Otherwise LGTM!

PR #2621, opened 2023-10-12
fix: address WC style feedback

iamacook:
## What it solves

Addresses initial design implementation feedback

## How this PR fixes it

- Only one help accordion can be open at once
- Hiding the accordions is indefinite and then shown only via the info button
- URI input field is no longer disabled after URI entry
- Small text adjustments

## How to test it

Interact with the WC modal and observe the above addressed.

## Checklist


katspaugh (approved):
Thanks!

PR #2622, opened 2023-10-12
feat: wrong network warning + high risk checkbox

iamacook:
## What it solves

Improved warnings

## How this PR fixes it

- High risk sessions now require user confirmation via a checkbox. (Note: scams are outright disabled.)
- Mismatching required chain is warned of.

### High risk session

Please note: this warning is falsified and the warning message is therefore incorrect:

image

### Mismatching required chain

image

## Checklist


katspaugh:
As per our discussions, we shoudn't warn if an app isn't verified. Only show a green message if it _is_ verified, or a red warning if it's a scam/MiTM attack.

As for the "wrong network" warning, the message should be milder and more concrete. E.g.
```
The dApp requires the wallet to support Ethereum mainnet, but the current Safe Account is on Polygon. Please be aware that the Safe Account doesn't exist on all chains on the same address.
```

src/components/walletconnect/ProposalForm/WrongRequiredChain.tsx
```diff
@@ -0,0 +1,13 @@
+import { Alert } from '@mui/material'
+import type { ReactElement } from 'react'
+
+import css from './styles.module.css'
+
+export const WrongRequiredChain = (): ReactElement => {
+  return (
+    <Alert severity="warning" className={css.alert}>
+      The dApp is trying to connect with the wrong network. Approving the session should update it but we advise
+      checking in the dApp before transacting.
```
katspaugh, line 10: Let's not create a separate component for each alert but instead have an enum of each warning and their severity.

katspaugh (approved):
👍

PR #2625, opened 2023-10-12
Fix: show nonce warning when above current nonce

### LINKED ISSUE #2560: Show a warning/prompt when nonce is above the latest nonce

katspaugh:
## What it solves

Resolves #2560

## How this PR fixes it
Shows a warning if the nonce is above the recommended nonce, or if the nonce is more than 100 higher than the safe nonce.

## How to test it
Enter a nonce that's higher than the recommended one.

## Screenshots
Screenshot 2023-10-12 at 17 48 59


src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -82,17 +82,10 @@ const NonceFormOption = memo(function NonceFormOption({
...
+  const MIN_CHARS = 7
   const MAX_WIDTH = '200px'
-  const ADORNMENT_PADDING = '24px'
-
   const clamped = `clamp(calc(${MIN_CHARS}ch + 6px), calc(${Math.max(MIN_CHARS, value.length)}ch + 6px), ${MAX_WIDTH})`
-
-  if (showRecommendedNonceButton) {
-    return `calc(${clamped} + ${ADORNMENT_PADDING})`
-  }
-
```
katspaugh, line 95: Removed this to avoid the jumping when the reset button appears, it was really annoying. I made it wider by default.

iamacook (approved):

src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -102,11 +95,23 @@ enum TxNonceFormFieldNames {
   NONCE = 'nonce',
 }
 
+enum ErrorMessages {
+  NONCE_MUST_BE_NUMBER = 'Nonce must be a number',
+  NONCE_TOO_LOW = "Nonce can't be lower than %%nonce%%",
+  NONCE_TOO_HIGH = 'Nonce is too high',
+  NONCE_TOO_FAR = 'Nonce is very far from the current nonce',
```
usame-algan, line 102: Maybe something like "Nonce is much higher than the current nonce"? "very far" sounds ambiguous imo

src/components/tx-flow/common/TxNonce/index.tsx
```diff
@@ -124,6 +129,20 @@ const TxNonceForm = ({ nonce, recommendedNonce }: { nonce: string; recommendedNo
...
+    if (Number(nonce) >= safe.nonce + MAX_NONCE_DIFFERENCE) {
+      message = ErrorMessages.NONCE_TOO_FAR
+    }
+
+    if (Number(nonce) > Number(recommendedNonce)) {
+      message = ErrorMessages.NONCE_GT_RECOMMENDED
+    }
+
+    setWarning(message)
+  }, [nonce, recommendedNonce, safe.nonce])
```
usame-algan, line 144: Why can't this be part of the `validate` call where we also have the rest of the messages?

src/components/tx-flow/common/TxNonce/index.tsx
katspaugh: Maybe! What does the BritGPT say, @iamacook?

src/components/tx-flow/common/TxNonce/index.tsx
katspaugh: Because in some cases this might be legit what the user wants to do. At least this was like this before the tx flow redesign, IIRC.

src/components/tx-flow/common/TxNonce/index.tsx
iamacook: The suggestion sounds good to me.

liliya-soroka:
Verified.
Checked cases:
1. new nonce<the current safe nonce
2. new nonce< the recommended nonce

PR #2630, opened 2023-10-13
feat: enter pairing URI from clipboard on focus

iamacook:
## What it solves

Automatically connects to the pairing URI present in the clipboard (if the permissions were previously granted).

## How to test it

Copy a pairing URI and open the WC modal.

## Screenshots

focus uri

## Checklist


katspaugh:
When I suggested this, I actually meant a step further, that the popup should appear automatically as soon as we detect `wc:` in the clipboard.

src/components/walletconnect/WcInput/index.tsx
```diff
@@ -51,8 +51,51 @@ const WcInput = (): ReactElement => {
...
+      }
+
+      const clipboard = await getClipboard()
+
+      if (clipboard?.startsWith('wc:')) {
+        onInput(clipboard)
+      }
+    }
+
+    window.addEventListener('focus', autofillUri)
```
katspaugh, line 93: Isn't WcInput rendered only when the WC button in the header is clicked?

iamacook: The `SessionManager` [is always mounted](https://github.com/safe-global/safe-wallet-web/blob/focus-wc-uri/src/components/walletconnect/HeaderWidget/index.tsx#L63) (`keepMounted`) in order to [listen for proposals](https://github.com/safe-global/safe-wallet-web/blob/focus-wc-uri/src/components/walletconnect/SessionManager/index.tsx#L62).

We could hoist [the state](https://github.com/safe-global/safe-wallet-web/blob/focus-wc-uri/src/components/walletconnect/SessionManager/index.tsx#L18) but I think it makes the most sense to keep it in `SessionManager`. What do you think?

src/services/walletconnect/WalletConnectContext.tsx
```diff
@@ -25,7 +28,8 @@ export const WalletConnectProvider = ({ children }: { children: ReactNode }) =>
     safeAddress,
   } = useSafeInfo()
   const [walletConnect, setWalletConnect] = useState<WalletConnectWallet | null>(null)
-  const [wcUri, setWcUri] = useWalletConnectSearchParamUri()
+  const [searchParamWcUri, setSearchParamWcUri] = useWalletConnectSearchParamUri()
+  const [clipboardWcUri, setClipboardWcUri] = useWalletConnectClipboardUri()
```
katspaugh, line 32: I wouldn't overload the context with this.

I would listen to the clipboard/URL from the header widget, and open the poup with the input if a wc-URI is detected. Then populate the input with that, and trigger a propose as usual from the input.

src/components/walletconnect/WcInput/index.tsx
iamacook: By drilling, in accordance with https://github.com/safe-global/safe-wallet-web/pull/2630#discussion_r1360392904, we no longer need to keep this mounted.

src/services/walletconnect/useWalletConnectClipboardUri.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    // Errors handled in the clipboard utils
+    const setClipboard = () => {
+      isClipboardGranted().then((isGranted) => {
+        if (isGranted) {
+          getClipboard().then(setState)
+        }
+      })
+    }
```
katspaugh, line 21: This needs to check if it starts with `wc:`, otherwise it will open the popup for any copied text.

Also please make it an async function.

src/services/walletconnect/WalletConnectContext.tsx
```diff
@@ -7,10 +7,11 @@ import useSafeWalletProvider from '@/services/safe-wallet-provider/useSafeWallet
 import WalletConnectWallet from './WalletConnectWallet'
 import { asError } from '../exceptions/utils'
 import { stripEip155Prefix } from './utils'
-import { useWalletConnectSearchParamUri } from './useWalletConnectSearchParamUri'
 
 const walletConnectSingleton = new WalletConnectWallet()
 
+const isWcUri = (uri: string) => uri.startsWith('wc:')
```
katspaugh, line 13: This doesn't seem to be used.

src/services/walletconnect/useWalletConnectClipboardUri.ts
iamacook: There [is a check in `WcInput`](https://github.com/safe-global/safe-wallet-web/blob/focus-wc-uri/src/components/walletconnect/WcInput/index.tsx#L46) but I added one to the hook as well.

src/services/walletconnect/useWalletConnectClipboardUri.ts
```diff
@@ -0,0 +1,49 @@
...
+    // Update clipboard when returning to Safe
+    window.addEventListener('focus', setClipboard)
+    return () => {
+      window.removeEventListener('focus', setClipboard)
+    }
+  }, [])
+
+  const setClipboard = async (data: string) => {
+    await navigator.clipboard
+      .writeText(data)
```
katspaugh, line 39: Don’t think it’s a good idea to mess with peoples’ clipboards.

src/components/walletconnect/WcInput/index.tsx
```diff
@@ -41,17 +42,19 @@ const WcInput = (): ReactElement => {
     [walletConnect],
   )
 
+  useEffect(() => {
+    if (uri.startsWith('wc:')) {
```
katspaugh, line 46: This code is now repeated. Is it needed?

iamacook: It prevents invalid URIs from the being entered. I've moved it to the search param hook instead.

katspaugh (approved):

PR #2637, opened 2023-10-16
Chore: enable unit test coverage comment

katspaugh:
## What it solves

I've switched our CI to a different Jest action that has a better coverage report and the same annotations as the previous one.

Also, I've deleted the pre-commit hook because it was really slow.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,10 +21,8 @@ jobs:
       - name: Run tests
         run: 'yarn test:ci'
 
-      - name: Comment in failing tests
-        uses: mattallty/jest-github-action@v1.0.3
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Annotations and coverage report
+        uses: ArtiomTr/jest-coverage-report-action@v2
         with:
-          test-command: 'echo unit tests already executed'
-          coverage-comment: true
+          coverage-file: jest.results.json
+          test-script: 'echo coverage generated'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,4 +28,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           test-command: 'echo unit tests already executed'
-          coverage-comment: false
+          coverage-comment: true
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -18,14 +18,10 @@ jobs:
...
-      - name: Run tests
-        run: 'yarn test:ci'
-
-      - name: Comment in failing tests
-        uses: mattallty/jest-github-action@v1.0.3
-        if: failure()
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Annotations and coverage report
+        uses: ArtiomTr/jest-coverage-report-action@v2
```
usame-algan, line 22: Is there a reason we are not using Coveralls? It seems that we already have some projects on there: https://coveralls.io/github/safe-global

katspaugh: I don't see the point of depending on a 3rd-party service for something available locally.

package.json
```diff
@@ -32,9 +32,6 @@
   "engines": {
     "node": ">=16"
   },
-  "pre-commit": [
```
schmanu, line 35: Should we discuss this? It feels out of scope of the coverage report.

schmanu (commented):
One thing that this setup is missing is seeing the diff to the base branch.
Or will this happen once it's merged to `dev`?

.github/workflows/test.yml
```diff
@@ -18,14 +18,10 @@ jobs:
...
         with:
-          test-command: 'echo unit tests already executed'
-          coverage-comment: false
+          skip-step: install
+          package-manager: yarn
+          test-script: yarn test:ci
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
katspaugh: It keeps a single comment already, those two other comments in this PR are due to a different configuration while I was testing it.

package.json
katspaugh: Let's dicuss it, yes. I'll start a thread on Slack.

katspaugh:
@schmanu this action runs Jest on both the target branch and the PR branch and is supposed to give you a diff. I'm not sure why it's not reflected in this PR. Maybe it needs to be merged to dev indeed.

PR #2640, opened 2023-10-16
fix: session management when switching chains

iamacook:
## What it solves

Resolves unsupported sessions remaining connected

## How this PR fixes it

1. When switching chain, DApps that don't support the chain being opened are disconnected.
2. If navigating out of a "Safe route", e.g. the welcome page, all sessions are disconnected.

## How to test it

1. Open a mainnet Safe and connect to AAVE and mint.fun. Observe mint.fun disconnecting when switching to a Polygon Safe.
2. Navigate to the welcome page and observe AAVE disconnecting, as well as the WC button disabled.

## Screenshots

connection

## Checklist

### DIFF
--- a/src/services/walletconnect/WalletConnectContext.tsx
+++ b/src/services/walletconnect/WalletConnectContext.tsx
@@ -8,6 +8,7 @@ import WalletConnectWallet from './WalletConnectWallet'
 import { asError } from '../exceptions/utils'
 import { stripEip155Prefix } from './utils'
 import { useWalletConnectSearchParamUri } from './useWalletConnectSearchParamUri'
+import { useRouter } from 'next/router'
 
 const walletConnectSingleton = new WalletConnectWallet()
 
@@ -28,6 +29,8 @@ export const WalletConnectProvider = ({ children }: { children: ReactNode }) => 
   const [wcUri, setWcUri] = useWalletConnectSearchParamUri()
   const [error, setError] = useState<Error | null>(null)
   const safeWalletProvider = useSafeWalletProvider()
+  const router = useRouter()
+  const isSafeConnected = router.isReady && router.query.safe
 
   // Init WalletConnect
   useEffect(() => {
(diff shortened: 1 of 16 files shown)


src/services/walletconnect/WalletConnectContext.tsx
```diff
@@ -42,6 +45,13 @@ export const WalletConnectProvider = ({ children }: { children: ReactNode }) =>
     walletConnect.updateSessions(chainId, safeAddress).catch(setError)
   }, [walletConnect, chainId, safeAddress])
 
+  // Disconnect all sessions if no Safe is loaded
+  useEffect(() => {
+    if (!walletConnect || isSafeConnected) return
+
+    walletConnect.disconnectAllSessions().catch(setError)
```
katspaugh, line 52: What if a Safe is just being loaded? WC sessions live across page reloads unless disconnected.

src/services/walletconnect/utils.ts
```diff
@@ -28,3 +30,38 @@ export const getSupportedChainIds = (configs: Array<ChainInfo>, params: Proposal
...
+    supportedNamespaces: {
+      [EIP155]: {
+        chains: eip155ChainIds,
+        accounts: eip155Accounts,
+        methods,
+        events,
+      },
+    },
+  })
+}
```
katspaugh, line 67: Stylistically, I don't see this function as a util. I would keep it as a private method in the WalletConnectWallet class.

src/services/walletconnect/WalletConnectContext.tsx
iamacook: We [reference the `safe` search param](https://github.com/safe-global/safe-wallet-web/pull/2640/files#diff-d8b8f8e4abf39b33900def2ffb040c4e238c797d2b4281f97a1d7992d7f3fa6bR31) for this reason.

katspaugh (approved):

PR #2643, opened 2023-10-17
fix: clear pairing URI on rejection

iamacook:
## What it solves

Resolves stale pairing URIs in input

## How this PR fixes it

When rejecting a session proposal, the pairing URI is cleared from the URL/clipboard (state) meaning that the input is not populated with the value.

## How to test it

Copy a pairing URI from a DApp and reject the proposal, observing no value prepopulating the input field.

## Screenshots

rejection-clear

## Checklist

### DIFF
--- a/src/components/walletconnect/WcInput/index.tsx
+++ b/src/components/walletconnect/WcInput/index.tsx
@@ -4,7 +4,7 @@ import type { ReactElement } from 'react'
 
 import { WalletConnectContext } from '@/services/walletconnect/WalletConnectContext'
 import { asError } from '@/services/exceptions/utils'
-import { getClipboard } from '@/utils/clipboard'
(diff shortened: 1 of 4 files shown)


src/components/walletconnect/WcInput/index.tsx
```diff
@@ -66,7 +62,7 @@ const WcInput = ({ uri }: { uri: string }): ReactElement => {
       label={error ? error.message : 'Pairing UI'}
       placeholder="wc:"
       InputProps={{
-        endAdornment: isFirefox ? undefined : (
+        endAdornment: isFirefox() ? undefined : (
```
katspaugh, line 65:
```suggestion
        endAdornment: isPastingSupported() ? undefined : (
```

katspaugh (approved):

PR #2647, opened 2023-10-17
Refactor: track tx origin

katspaugh:
## What it solves

To measure how many txs are made through WalletConnect, we'll track the Safe App origin field in onchain/offchain interaction events.

I also refactored the emitted tx event bus to not pass around humanDescription and instead fetch it from CGW.

## How to test it

Non-Safe App txs:
1. Make a send token tx
2. Observe an "Onchain interaction" event tracked in the console
3. The "label" property of this event should be empty (indicating a tx initiated from our own UI)

Safe App txs:
1. Make a tx with any Safe App
2. Observe an "Onchain interaction" event tracked in the console
3. The label property should be populated with that Safe App's URL


src/hooks/useTxTracking.ts
```diff
@@ -1,23 +1,43 @@
...
+      txSubscribe(txEvent as TxEvent, async (detail) => {
+        const txId = 'txId' in detail ? detail.txId : undefined
+        const txHash = 'txHash' in detail ? detail.txHash : undefined
+        const id = txId || txHash
+
+        let origin = ''
+        if (id) {
+          try {
+            const txDetails = await getTxDetails(chainId, id)
+            origin = txDetails.safeAppInfo?.url || ''
```
usame-algan, line 28: Should we track the wallet type as a fallback here i.e. `wallet.label`?

iamacook: The (peer) label may be a better option as we have to [use the WC Safe App URL as the `appUrl`](https://github.com/safe-global/safe-wallet-web/pull/2641/files#diff-63ad7ff9d4750580543f5094dab28fde1143f63658e694c67eaa680a6cc2f998R91) to get the WC icon in the transaction list.

usame-algan (approved):
Looks good 👍 I was able to test both cases.

One thing: Is there a reason we are tracking the origin url instead of the safe app name? Could be an issue to map certain urls like ipfs to the correct safe app afterwards.

katspaugh:
Hmm, good point. My thinking was URLs change less often, and I think that's what the tx-service also uses as an id (because it maps them to txs by origin). But maybe for tracking purposes a name would be better. Easier to find/type.

PR #2648, opened 2023-10-18
fix: track safeAddress on Safe creation

usame-algan:
## What it solves

Adds safeAddress to safe creation event

## How to test it

1. Create a new safe
2. Observe that the safe address exists in GTM on the Created Safe event
3. Navigate to that safe
4. Observe that the safe address is still there


katspaugh:
Safe address is sent in a separate custom dimension for all events. A label is no bueno. It's usually taken from the URL but you can add the option to take it from the pending Safe creation.

katspaugh (approved):

francovenica:
Looks good. 
The safeAddress value is set in the tracker after the safe is successfully created. 
Visiting the safe dashbard and other areas are being tracked normally

A comparison between current Dev and this PR  to see  the change:
image
image

PR #2652, opened 2023-10-18
feat: block/warn about bridges accordingly

iamacook:
## What it solves

Resolves blocks/warns of risky bridge use

## How this PR fixes it

The usage of certain bridges are either blocked or warned about, requiring user acceptance according to [our research](https://www.notion.so/Native-WalletConnect-Dangerous-bridge-research-22298ea428934846814e5b6599437dd8).

## How to test it

Open one of the bridges specified and observe either a blocked connection window, or a warning that needs to first be accepted.

## Screenshots

image

image

## Checklist

### DIFF
--- a/src/components/walletconnect/ProposalForm/bridges.ts
+++ b/src/components/walletconnect/ProposalForm/bridges.ts
@@ -1,43 +1,11 @@
-const DangerousBridges = [
-  'bridge.arbitrum.io',
-  'bridge.base.org',
-  'cbridge.celer.network',
-  'www.orbiter.finance',
-  'zksync-era.l2scan.co',
-  'app.optimism.io',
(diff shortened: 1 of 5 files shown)


src/components/walletconnect/ProposalForm/bridges.ts
```diff
@@ -0,0 +1,43 @@
+const DangerousBridges = [
```
katspaugh, line 1: Naming-wise, there's nothing dangerous/risky about these bridges per se, it's just that they aren't smart contract account-friendly. Please find more specific names for the two arrays. And please move these lists (not the two functions below) to `src/components/walletconnect/constants.ts`.

katspaugh (approved):
Looks good apart from my comment 👍

PR #2654, opened 2023-10-18
feat: on-chain WC signing

iamacook:
## What it solves

Resolves on-chain signing capibility

## How this PR fixes it

Toggling between on-/off-chain signing is now possible via WC, handling the `safe_setSettings` RPC call.

## How to test it

Open the EIP-1271 example dApp and connect via WC (on the `wallet-connect-v2` branch). Observe signing (typed) messages works as expected after toggling between on-/off-chain signing.

## Screenshots

off-chain signing

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/SignMessage/index.tsx
+++ b/src/components/tx-flow/flows/SignMessage/index.tsx
@@ -4,7 +4,7 @@ import { Box, Typography } from '@mui/material'
 import SafeAppIconCard from '@/components/safe-apps/SafeAppIconCard'
 
 const APP_LOGO_FALLBACK_IMAGE = '/images/apps/apps-icon.svg'
-const APP_NAME_FALLBACK = 'Sign message off-chain'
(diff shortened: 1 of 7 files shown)


src/components/tx-flow/flows/SignMessage/index.tsx
```diff
@@ -4,7 +4,7 @@ import { Box, Typography } from '@mui/material'
 import SafeAppIconCard from '@/components/safe-apps/SafeAppIconCard'
 
 const APP_LOGO_FALLBACK_IMAGE = '/images/apps/apps-icon.svg'
-const APP_NAME_FALLBACK = 'Sign message off-chain'
+const APP_NAME_FALLBACK = 'Sign message'
```
iamacook, line 7: "off-chain" was being shown by default, regardless of whether it was on-/off-chain signing.

katspaugh (approved):
👍

PR #2655, opened 2023-10-18
fix: multipl session disconnects throwing

iamacook:
## What it solves

Resolves multiple session disconnects throwing

## How this PR fixes it

Session updated are now handled one after each other.

## How to test it

Connect two DApps that support chain A but not chain B. Switch from chain A to chain B and observe no WC-related error.

## Checklist

### DIFF
--- a/src/services/walletconnect/WalletConnectWallet.ts
+++ b/src/services/walletconnect/WalletConnectWallet.ts
@@ -141,16 +141,16 @@ class WalletConnectWallet {
     const newEip155ChainId = getEip155ChainId(chainId)
     const newEip155Account = `${newEip155ChainId}:${safeAddress}`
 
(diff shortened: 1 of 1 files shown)


src/services/walletconnect/WalletConnectWallet.ts
```diff
@@ -141,16 +141,16 @@ class WalletConnectWallet {
...
     // Switching to unsupported chain
-    if (hasNewChainId) {
+    if (isUnsupportedChain) {
       return this.disconnectSession(session)
     }
 
-    // Add new account to the session namespace
-    if (hasNewAccount) {
+    // Add new Safe to the session namespace
+    if (isNewSessionSafe) {
```
iamacook, line 153: These renames somehow didn't merge correctly before.

katspaugh (approved):

PR #2661, opened 2023-10-19
fix: emit `chainChanged` first

iamacook:
## What it solves

Resolves dApp chain alignment

## How this PR fixes it

The `chainChanged` event is not emitted before the `accountsChanged event when updating a session.

## How to test it

Connect to [Aura Finance](https://app.aura.finance) and observe the initial chain matching that of the Safe, as well as changing with Safe switches across chains.

## Checklist

### DIFF
--- a/src/services/walletconnect/__tests__/WalletConnectWallet.test.ts
+++ b/src/services/walletconnect/__tests__/WalletConnectWallet.test.ts
@@ -344,36 +344,36 @@ describe('WalletConnectWallet', () => {
       expect(emitSessionEventSpy).toHaveBeenCalledTimes(2)
     })
 
-    it('should call emitSessionEvent with the correct parameters', async () => {
+    it.only('should call emitSessionEvent with the correct parameters', async () => {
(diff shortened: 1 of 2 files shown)


katspaugh (approved):

src/services/walletconnect/__tests__/WalletConnectWallet.test.ts
```diff
@@ -344,36 +344,36 @@ describe('WalletConnectWallet', () => {
       expect(emitSessionEventSpy).toHaveBeenCalledTimes(2)
     })
 
-    it('should call emitSessionEvent with the correct parameters', async () => {
+    it.only('should call emitSessionEvent with the correct parameters', async () => {
```
katspaugh, line 347: Don't forget the only.

iamacook: Addressed in 09db2d41623fff256ef808f743ae3ee7525091b0.

PR #2662, opened 2023-10-19
Feat: open WC popup on dashboard widget click

katspaugh:
## What it solves

The WC dashboard widget will no longer go to the app, but will open the native WC popup instead.


iamacook (approved):
There are a few lint issues but otherwise LGTM.

PR #2664, opened 2023-10-19
feat: unsupported chain disconnection feedback

iamacook:
## What it solves

Resolves no feedback when automatically disconnected from a dApp when switching to an unsupported chain

## How this PR fixes it

When a session is deleted, if it is because of an unsupported chain, a new notification popup is shown.

## How to test it

Open a dApp on a chain that is supported and switch to one that is not. Observe the new popup.

## Screenshots

disconnection-feedback

## Checklist


src/components/walletconnect/HeaderWidget/index.tsx
```diff
@@ -25,12 +28,92 @@ const usePrepopulatedUri = (): [string, () => void] => {
...
+
+    walletConnect.onSessionDelete((session) => {
+      timeout = onDelete(session)
+    })
+
+    return () => clearTimeout(timeout)
+  }, [onDelete, walletConnect])
+
+  return [metadata, setMetadata]
+}
```
katspaugh, line 110: There's a lot going on here. Please revisit this code. I'm not going to point out what exactly is not ideal in terms of code structure, I hope you can see it yourself.

src/services/walletconnect/WalletConnectWallet.ts
```diff
@@ -234,10 +234,12 @@ class WalletConnectWallet {
   /**
    * Subscribe to session delete
    */
-  public onSessionDelete(handler: () => void) {
+  public onSessionDelete = (handler: (session: SessionTypes.Struct) => void) => {
```
iamacook, line 237: There were changed to arrow functions because there were otherwise complaints that `this` was `undefined`.

src/components/walletconnect/HeaderWidget/index.tsx
iamacook: I've refactored the above and created a `useDeferredListener` hook in 231ac0c90f6dd4c8c1e650b19d55932e667df6f1. Lmk what you think.

src/components/walletconnect/HeaderWidget/index.tsx
```diff
@@ -9,7 +8,10 @@ import { useWalletConnectSearchParamUri } from '@/services/walletconnect/useWall
 import Icon from './Icon'
 import SessionManager from '../SessionManager'
 import Popup from '../Popup'
-import { SuccessBanner } from '../SuccessBanner'
+import { ConnectionBanner } from '../ConnectionBanner'
+import useSafeInfo from '@/hooks/useSafeInfo'
+import { isUnsupportedChain } from '@/services/walletconnect/utils'
+import { useDeferredListener } from '@/hooks/useDefferedListener'
```
katspaugh, line 14: GitHub might be hiding some changes but I don't see useDeferredListener used anywhere.

iamacook: It's used in `useSuccessSession` (L35) and `useDeleteSession` (L45).

katspaugh: I see, thanks. IMHO this hook is wrapping some trivial logic in an untrivial way with cryptic paramas, so I would not add it. Let's merge it but I will refactor this later.

katspaugh (approved):

PR #2672, opened 2023-10-20
[Seedless-Onboarding] Add signer account to address book on login

schmanu:
## What it solves
Adds the mpc signer address into the addressbook after login.

## How to test it
1. If the google signer address is in the addressbook, delete it.
2. Re-Login with google
3. Observe that the resolved address is added to the addressbook.

## Checklist

### DIFF
--- a/src/hooks/wallets/mpc/useMPCWallet.ts
+++ b/src/hooks/wallets/mpc/useMPCWallet.ts
@@ -9,6 +9,11 @@ import { SecurityQuestionRecovery } from './recovery/SecurityQuestionRecovery'
 import { DeviceShareRecovery } from './recovery/DeviceShareRecovery'
 import { trackEvent } from '@/services/analytics'
 import { MPC_WALLET_EVENTS } from '@/services/analytics/events/mpcWallet'
+import useAddressBook from '@/hooks/useAddressBook'
+import { useCurrentChain } from '@/hooks/useChains'
+import { upsertAddressBookEntry } from '@/store/addressBookSlice'
+import { useAppDispatch } from '@/store'
+import { ethers } from 'ethers'
 
 export enum MPCWalletState {
   NOT_INITIALIZED,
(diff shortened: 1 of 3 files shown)


src/hooks/wallets/mpc/useMPCWallet.ts
```diff
@@ -30,6 +35,9 @@ export const useMPCWallet = (): MPCWalletHook => {
   const [walletState, setWalletState] = useState(MPCWalletState.NOT_INITIALIZED)
   const mpcCoreKit = useMPC()
   const onboard = useOnboard()
+  const addressBook = useAddressBook()
+  const currentChainId = useCurrentChain()
```
usame-algan, line 39: We can use `useChainId` directly here as we only need the chain id

src/hooks/wallets/mpc/useMPCWallet.ts
```diff
@@ -99,12 +107,21 @@ export const useMPCWallet = (): MPCWalletHook => {
...
+      const wallets = await connectWallet(onboard, {
         autoSelect: {
           label: ONBOARD_MPC_MODULE_LABEL,
           disableModals: true,
         },
       }).catch((reason) => console.error('Error connecting to MPC module:', reason))
+
+      // If the signer is not in the address book => add it as the user's first name
+      if (wallets && currentChainId && wallets.length > 0) {
+        const signerAddress = ethers.utils.getAddress(wallets[0].accounts[0]?.address)
```
usame-algan, line 119: `getAddress` throws an error if we pass undefined. Not sure why the types don't reflect it but I think we need to split up the call. We could also reuse our `checksumAddress` wdyt?

```
const address = wallets[0].accounts[0]?.address
if (!address) return
const checksummedAddress = checksumAddress(address)
```

usame-algan (approved):
Works nicely 👍 

I was also exploring moving the logic into `useWallet` or a similar place where we listen to connected wallets but in that case we are missing the email address and would have to add that dependency. Also `useWallet` has a lot of instances which would mess with inserting an address book entry. We might want to convert it into a global context at some point.

PR #2674, opened 2023-10-20
Fix: uncaught exception in ApprovalEditor

katspaugh:
## What it solves

ABI decoded was throwing in some rare cases.

Screenshot 2023-10-20 at 15 05 02


schmanu (approved):
Good catch 🥁

katspaugh:
Yeah, literally :D

PR #2678, opened 2023-10-23
feat: Show wallet balance

usame-algan:
## What it solves

Displays the connected wallet balance in the header and in the tx-flow

[Figma Designs](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?type=design&node-id=8647-11732&mode=design&t=25lotpzWMdjN7X2q-0) kudos to @TanyaEfremova !

## How this PR fixes it

- extracts the `useWallet` hook into a global context

## How to test it

1. Observe the wallet balance showing in the header for any connected wallet
2. Observe that it always reflects the connected wallet network not necessarily the network the safe is on
3. Create a transaction and observe seeing the wallet balance in the Execute block
4. Observe the new design according to the linked Figma

## Screenshots
Screenshot 2023-10-23 at 09 20 39
Screenshot 2023-10-23 at 09 20 58
Screenshot 2023-12-04 at 09 48 42
Screenshot 2023-12-04 at 09 49 25
Screenshot 2023-12-07 at 14 09 40

## Checklist


src/components/common/WalletBalance/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const currentChain = useCurrentChain()
+
+  if (!balance) {
+    return <Skeleton width={30} />
+  }
+
+  return (
+    <>
+      {formatVisualAmount(balance, currentChain?.nativeCurrency.decimals ?? 18)}{' '}
+      {currentChain?.nativeCurrency.symbol ?? 'ETH'}
```
katspaugh, line 16: Could you extrace 18 and ETH into a `const NATIVE_TOKEN_PLACEHLDER`?

src/hooks/wallets/useWalletBalance.ts
```diff
@@ -0,0 +1,19 @@
+import useAsync from '../useAsync'
+import useWallet from './useWallet'
+import { useWeb3ReadOnly } from '@/hooks/wallets/web3'
+import { type BigNumber } from 'ethers'
+
+const useWalletBalance = () => {
```
katspaugh, line 6: Could you add a return type here?

src/components/common/WalletInfo/styles.module.css
```diff
@@ -8,6 +8,37 @@
...
   justify-content: center;
+  position: relative;
+}
+
+.walletIcon {
+  position: absolute;
+  display: flex;
+  -webkit-box-pack: center;
+  justify-content: center;
+  -webkit-box-align: center;
```
katspaugh, line 19: What are these -webpack rules? Possible to stick to standard CSS?

katspaugh (commented):
Nice little feature!

src/components/common/WalletInfo/styles.module.css
usame-algan: Oh looks like a copy-paste error. Will remove them

src/hooks/wallets/useWalletBalance.ts
```diff
@@ -0,0 +1,19 @@
...
+  const web3ReadOnly = useWeb3ReadOnly()
+  const wallet = useWallet()
+
+  return useAsync<BigNumber | undefined>(() => {
+    if (!wallet || !web3ReadOnly) {
+      return undefined
+    }
+
+    return web3ReadOnly.getBalance(wallet.address, 'latest')
```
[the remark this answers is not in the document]
usame-algan: Unfortunately, the balance that onboard returns is already formatted as a string. We could use this in the header even though it might have a different format than our number formatting rules but @schmanu made a valid point that if we want to warn the user that their balance is not enough to execute a transaction the amount in WEI or similar raw value would be better.

usame-algan:
Note: I suggest we implement the network selector logos in a separate PR since this PR is already quite big

schmanu (approved):
LGTM!

francovenica:
So this balance showing is fine, but if the network doesn't have relayer it will never be displayed.

Is expected to show this only for networks with relayer?
image

francovenica:
The rest is fine.

I tried with MM chrome extension, MM phone app, Trezor, Ledger, Safe-as-owner and in all of them it looks good
The balance updates everytime you open the display in the top right or refresh the form in the tx

I wasn't able to test in social login since it's blocked in this PR

katspaugh:
> So this balance showing is fine, but if the network doesn't have relayer it will never be displayed.

IMO, we should always show this block. Obviously, without the radio buttons if no relayer.

src/components/tx/BalanceInfo/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+const BalanceInfo = () => {
+  const [balance] = useWalletBalance()
+
+  return (
+    <Box className={css.sponsoredBy}>
+      <SvgIcon component={GasStationIcon} inheritViewBox className={css.icon} />
+      <div>
+        <Stack direction="row" spacing={0.5} alignItems="center">
+          <Typography variant="body2" fontWeight={700} letterSpacing="0.1px">
+            Wallet balance
```
katspaugh, line 16: `Connected wallet balance` perhaps? To not confuse it with the Safe's own ETH balance.

usame-algan:
I changed it to always show the `Connected wallet balance` block now either with or without relayer. Added a screenshot in the PR description. @francovenica can you test this new case? Thanks!

usame-algan:
> > On the top right corner, is confusing that the balance shown is from the network the safe is on, but the "Network" label shows the network the wallet is currently connected to.
> 
> @francovenica The reasoning for this is that we only switch the network to the Safe's network when the user submits / executes a transaction. If we would show the balance in the wallet's network it could seem like you have enough balance to execute the transaction although it's just a balance on an unrelated network. I agree that it is a bit confusing though. Probably we could add a warning Icon or something.

I remember we talked about prompting the user to switch network when they connect their wallet instead of delaying it to when they try to sign/execute a transaction. Thats how other dapps do it and even go as far as forcing the user to switch before enabling the UI (e.g. CowSwap).

francovenica:
Ok, thanks for the response.

I get now that is more important to show the user if he has funds to do tx in the current network than to show how much he has in whatever network he's connected to. Still a bit confusing but it makes sense.

The fix to show the balance even without the relayer looks good
image

LGTM

PR #2688, opened 2023-10-24
feat: track user's wallet type through user properties

schmanu:
## What it solves
Track the user's used wallet type as user property.

## Checklist

### DIFF
--- a/src/services/analytics/useGtm.ts
+++ b/src/services/analytics/useGtm.ts
@@ -83,7 +83,9 @@ const useGtm = () => {
   }, [router.pathname])
 
   useEffect(() => {
-    gtmSetUserProperty(AnalyticsUserProperties.WALLET_LABEL, walletLabel ?? WALLET_LABEL_NONE)
+    if (walletLabel) {
+      gtmSetUserProperty(AnalyticsUserProperties.WALLET_LABEL, walletLabel)
+    }
   }, [walletLabel])
 
   // Track meta events on app load
--- a/src/services/analytics/useGtm.ts
+++ b/src/services/analytics/useGtm.ts
@@ -22,15 +22,9 @@ import { useRouter } from 'next/router'
 import { AppRoutes } from '@/config/routes'
 import useMetaEvents from './useMetaEvents'
 import { useMediaQuery } from '@mui/material'
-import { DeviceType } from './types'
+import { AnalyticsUserProperties, DeviceType, WALLET_LABEL_NONE } from './types'
 import useSafeAddress from '@/hooks/useSafeAddress'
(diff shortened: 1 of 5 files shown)


src/services/analytics/useGtm.ts
```diff
@@ -79,6 +82,10 @@ const useGtm = () => {
     gtmTrackPageview(router.pathname)
   }, [router.pathname])
 
+  useEffect(() => {
+    gtmSetUserProperty('walletLabel', walletLabel ?? 'NONE')
```
katspaugh, line 86: Please create an enum for possible user properties, similar to how to define custom events.

src/services/analytics/useGtm.ts
```diff
@@ -26,6 +26,12 @@ import { DeviceType } from './types'
 import useSafeAddress from '@/hooks/useSafeAddress'
 import useWallet from '@/hooks/wallets/useWallet'
 
+enum GA_USER_PROPERTIES {
```
katspaugh, line 29: I wouldn’t put it here. This hook is just a hook to bind GTM to React. The actual analytics service is where it should be, similar to our custom event definitions.

katspaugh (approved):
👍

francovenica:
Manu and I checked together how the new property is tracked in ever event. Looks good.
There is a new exploration "instance" that tracks the events with this property, but it updates every 24hs, so I'll check tomorrow

katspaugh:
Seems to work.

Screenshot 2023-10-30 at 07 35 05

I'm not sure about NONE though, maybe we should just keep it unset?

src/services/analytics/useGtm.ts
```diff
@@ -79,6 +82,10 @@ const useGtm = () => {
     gtmTrackPageview(router.pathname)
   }, [router.pathname])
 
+  useEffect(() => {
+    gtmSetUserProperty(AnalyticsUserProperties.WALLET_LABEL, walletLabel ?? WALLET_LABEL_NONE)
+  }, [walletLabel])
```
katspaugh, line 87: Since the wallet is asynchronous, it will inevitably log NONE for the first few events before the wallet (re)connects. I would suggest to not set it at all until the user connects a wallet.

Optionally, we could also unset it when the user disconnects. You migth have to listen to a specific onboard event for that. But to keep things simple I would just keep the last wallet and never unset it.

schmanu: Makes sense. Especially because the interesting interactions such as signing a message / executing a tx will require a wallet anyways.

katspaugh (approved):

PR #2696, opened 2023-10-26
Fix: on close callback for WalletConnect

katspaugh:
The on-close callback wasn't correctly handled because of how we were setting state.

This wasn't setting the state to a new function, but was calling this function right away instead as it it was a setter callback.
```
setOnClose(onClose ?? noop)
```

I've changed onClose and shouldWarn to useRef to fix this.

Also removed the global USER_QUIT tx event and replaced with with direct onClose callbacks.

### DIFF
--- a/src/components/tx-flow/index.tsx
+++ b/src/components/tx-flow/index.tsx
@@ -1,8 +1,6 @@
-import { createContext, type ReactElement, type ReactNode, useState, useEffect, useCallback } from 'react'
+import { createContext, type ReactElement, type ReactNode, useState, useEffect, useCallback, useRef } from 'react'
 import TxModalDialog from '@/components/common/TxModalDialog'
-import { usePathname } from 'next/navigation'
-import useSafeInfo from '@/hooks/useSafeInfo'
-import { txDispatch, TxEvent } from '@/services/tx/txEvents'
+import { useSearchParams } from 'next/navigation'
 import { SuccessScreen } from './flows/SuccessScreen'
 
 const noop = () => {}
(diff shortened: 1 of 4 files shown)


src/components/tx-flow/index.tsx
```diff
@@ -19,68 +17,64 @@ export const TxModalContext = createContext<TxModalContextType>({
...
-  const [shouldWarn, setShouldWarn] = useState<boolean>(false)
-  const [, setOnClose] = useState<Parameters<TxModalContextType['setTxFlow']>[1]>(noop)
   const [fullWidth, setFullWidth] = useState<boolean>(false)
-  const pathname = usePathname()
-  const [, setLastPath] = useState<string>(pathname)
-  const { safeAddress, safe } = useSafeInfo()
+  const shouldWarn = useRef<boolean>(true)
+  const onClose = useRef<() => void>(noop)
+  const searchParams = useSearchParams()
+  const safeQuery = searchParams.get('safe')
```
usame-algan, line 30: Why not keep using `useSafeInfo` or `useSafeAddress`?

src/components/tx-flow/index.tsx
```diff
@@ -19,68 +17,64 @@ export const TxModalContext = createContext<TxModalContextType>({
...
   useEffect(() => {
-    setLastPath((prev) => {
-      if (prev !== pathname && txFlow) {
-        handleShowWarning()
-      }
-      return pathname
-    })
-  }, [txFlow, handleShowWarning, pathname])
+    if (safeQuery === prevQuery.current) return
+    prevQuery.current = safeQuery ?? ''
```
usame-algan, line 71: This is not taking into account anymore if the user navigates to a different route within the same safe. The tx-flow now stays open instead of showing the user a warning.

katspaugh: As long as they are in the same Safe, navigating shouldn't matter, should it?

katspaugh: Good point, I'll restore it.

PR #2708, opened 2023-10-31
Allow safe deployments to be higher than the version

JABirchall:
## What it solves
When there is a new version of safe deployments it is not downloaded as the package is set to a specific version

## How this PR fixes it
Allow the package to download a higher version if one is released.
I dont know if there is a very specific reason why the deployments is locked to a version, but it does mean that anyone who is trying to use safe-web on a chain that was recently added to deployments cant. 

## How to test it
`npm install` - should see the current release of safe-deployments downloaded instead of locked to 1.25.0
eg. current release is 1.28.0

## Screenshots
n/a

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "@safe-global/safe-apps-sdk": "^8.1.0",
     "@safe-global/safe-core-sdk": "^3.3.5",
     "@safe-global/safe-core-sdk-utils": "^1.7.4",
-    "@safe-global/safe-deployments": "1.25.0",
+    "@safe-global/safe-deployments": "^1.25.0",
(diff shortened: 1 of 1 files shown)


JABirchall:

katspaugh:
Thanks for the PR. I see your point, but I’ll have to close it.

We pin some packages to avoid accidentally upgrading to a breaking version. Specifically with that package, the maintainer mistakenly released a patch version that actually should’ve been a minor version as it had breaking changes.

Besides that, package versions are locked in the yarn.lock file for reproducible builds.

Our approach to package updates is = upgrade consciously, test, lock the new version.

PR #2709, opened 2023-10-31
fix: ignore empty names in addressbook

schmanu:
## What it solves
Creating a Safe with social signers lead to empty names in the addressbook.

## How this PR fixes it
We ignore addressBook entries with empty names and do not insert those anymore.

## How to test it
1. Create a Safe with social signer
2. Witness that no empty entry is created for that Safe anymore

## Checklist

### DIFF
--- a/src/store/__tests__/addressBookSlice.test.ts
+++ b/src/store/__tests__/addressBookSlice.test.ts
@@ -32,6 +32,21 @@ describe('addressBookSlice', () => {
     })
   })
 
+  it('should ignore empty names in the address book', () => {
+    const state = addressBookSlice.reducer(
+      initialState,
+      upsertAddressBookEntry({
+        chainId: '1',
+        address: '0x2',
+        name: '',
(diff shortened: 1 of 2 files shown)


src/store/__tests__/addressBookSlice.test.ts
```diff
@@ -32,6 +32,21 @@ describe('addressBookSlice', () => {
...
+      upsertAddressBookEntry({
+        chainId: '1',
+        address: '0x2',
+        name: '',
+      }),
+    )
+    expect(state).toEqual({
+      '1': { '0x0': 'Alice', '0x1': 'Bob' },
+      '4': { '0x0': 'Charlie', '0x1': 'Dave' },
+    })
```
usame-algan, line 47: nit: Could also check against the initialState here i.e. `expect(state).toEqual(initialState)`

usame-algan (approved):

PR #2712, opened 2023-10-31
fix: disconnect previous wallet

schmanu:
## What it solves
When switching the wallet the previously connected wallet stays connected.
This can lead to many edge cases for the social signer.

## How this PR fixes it
If the user successfully switches the wallet, we disconnect the old wallet label.

## How to test it
1. Login with google
2. Click switch wallet and connect to Metamask
3. Disconnect Metamask
4. you are no longer connected to google automatically

## Checklist


katspaugh:
That was actually a feature. If we're throwing it away, you can just remove the Switch wallet button altogether.

usame-algan:
I would also be in favour of limiting this to social login only and adding tracking for the Switch Wallet button to decide later if we should remove it or not.

src/hooks/wallets/useOnboard.ts
```diff
@@ -130,8 +131,19 @@ export const connectWallet = async (
...
+  const newWallets = await connectWallet(onboard)
+  const newWalletLabel = newWallets ? getConnectedWallet(newWallets)?.label : undefined
+
+  // If the wallet actually changed we disconnect the old connected wallet.
+  if (!newWalletLabel || oldWalletLabel !== ONBOARD_MPC_MODULE_LABEL) {
+    return
+  }
+
+  if (newWalletLabel !== oldWalletLabel) {
+    await onboard.disconnectWallet({ label: oldWalletLabel })
```
katspaugh, line 145: So we dusconnect if the old wallet is social login? What if the new wallet is social login and the old one is MM? Will it work fine?

schmanu: Yes exactly. 
Only the Social signer will be disconnected if switched to another wallet.
We could also remove this behavior once we support every chain. But this is currently causing issues QA found.

usame-algan (approved):

PR #2715, opened 2023-10-31
Fix: rm redirect from root to last safe

katspaugh:
The index page should now always redirect to the Welcome page as the starting page of the app.

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -1,30 +1,17 @@
-import { useEffect, useLayoutEffect } from 'react'
+import { useEffect } from 'react'
 import type { NextPage } from 'next'
 import { useRouter } from 'next/router'
-import useLastSafe from '@/hooks/useLastSafe'
 import { AppRoutes } from '@/config/routes'
-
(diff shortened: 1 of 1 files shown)


src/pages/index.tsx
```diff
@@ -1,30 +1,17 @@
...
-  const lastSafe = useLastSafe()
-  const safeAddress = safe || lastSafe
+  const { chain } = router.query
 
-  useIsomorphicEffect(() => {
-    if (router.pathname !== AppRoutes.index) {
-      return
+  useEffect(() => {
+    if (router.pathname === AppRoutes.index) {
+      router.replace(chain ? `${AppRoutes.welcome}?chain=${chain}` : AppRoutes.welcome)
```
usame-algan, line 12: nit:
```suggestion
      router.replace({ pathname: AppRoutes.welcome, query: router.query })
```

usame-algan (dismissed):
Nice change but wasn't it already working like this "by accident" since the last release?

katspaugh:
@usame-algan it was sometimes redirecting, sometimes not.

katspaugh:
Let's put this on hold for now.

Issue #1781, opened 2023-03-17
Display currently set `CompatibilityFallbackHandler` in the UI


iamacook:
## What is the feature about

We currently don't display the currently set `CompatibilityFallbackHandler` in the UI. As it can be changed, we should show it. We should also consider displaying a message whether the official one is set.

## The list of requirements

- [x] Confirm with design the location of display (likely settings)
- [x] Add (settings) block that shows the current `CFH`

## Designs/sketches

tbd

## Links

- [Discussion thread](https://5afe.slack.com/archives/C03BHKS5K6D/p1679074247966339)
- https://github.com/safe-global/web-core/issues/1318

PR #2716, opened 2023-10-31
Fix: soften fallback handler warning

katspaugh:
## What it solves

The original custom fallback handler warning was added in #1781.

We got feedback from CowSwap that the current warning is too scary to the users when they enable a custom handler for TWAP orders. So we're changing the icon and message to be less alerting.

Official:
Screenshot 2023-11-01 at 13 43 06

Custom:
Screenshot 2023-11-01 at 13 43 28

Unset:
Screenshot 2023-11-01 at 13 48 14


schmanu (commented):
If the CowSwap TWAP fallback handler is also static we could allowlist some individual addresses. If it is a completely unknown fallback handler I would prefer a warning.
What do you think?

katspaugh:
@schmanu allowlist where?

schmanu:
> @schmanu allowlist where?

For now we could have a list in our repo of known fallback handler addresses where we add the [CoW Extensible Fallback Handler addresses](https://github.com/cowprotocol/composable-cow/tree/ab3addad9bc05acdbf5fb040eb5c336209b58e31#deployed-contracts) 
But ideally that would live on the config service / gateway. 

I'm just thinking in general having an unexpected / inofficial fallbackhandler hints at a potentially risky setup. Or do we know many other dApps / protocols that use custom fallback handlers?

katspaugh:
If you're implying hardcoding it, I wouldn't do it.

liliya-soroka:
All custom fallbackhandlers will be green with the fix . I would check with the prod team if it's expected and safe
image

katspaugh:
Fixed. ✅ 

Official:
Screenshot 2023-11-01 at 13 43 06

Custom:
Screenshot 2023-11-01 at 13 43 28

Unset:
Screenshot 2023-11-01 at 13 48 14

liliya-soroka:
Verified

PR #2728, opened 2023-11-01
refactor: environment config for social wallet

schmanu:
## What it solves
Uses JSON in config for Social wallet to avoid having too many variables.

## Checklist


src/services/mpc/config.ts
```diff
@@ -0,0 +1,45 @@
...
+    keys.includes(SocialWalletOptionsKeys.web3AuthAggregateVerifierId) &&
+    keys.includes(SocialWalletOptionsKeys.web3AuthSubverifierId) &&
+    keys.includes(SocialWalletOptionsKeys.googleClientId)
+
+  const hasValues = Object.values(options).every(Boolean)
+  return hasRequiredKeys && hasValues
+}
+
+/** env variables */
+const SOCIAL_WALLET_IS_PRODUCTION = process.env.NEXT_PUBLIC_IS_PRODUCTION === 'true'
```
katspaugh, line 33: Theirs IS_PRODUCTION in the constants.

src/services/mpc/config.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const isSocialWalletOptions = (options: unknown): options is SocialWalletOptions => {
+  if (typeof options !== 'object' || options === null) {
+    return false
+  }
+  const keys = Object.keys(options)
+
+  const hasRequiredKeys =
+    keys.length === 4 &&
+    keys.includes(SocialWalletOptionsKeys.web3AuthClientId) &&
```
katspaugh, line 23: This validation is a bit excessive/boilerplatey

src/services/mpc/config.ts
usame-algan: I removed this const and use `IS_PRODUCTION` directly now.

src/services/mpc/config.ts
usame-algan: Changed it to use the `in` operator to check if all keys exist in the object.

src/services/mpc/config.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/** env variables */
+export const SOCIAL_WALLET_OPTIONS: any = (() => {
+  const SOCIAL_WALLET_OPTIONS_PRODUCTION = process.env.NEXT_PUBLIC_SOCIAL_WALLET_OPTIONS_PRODUCTION || ''
+  const SOCIAL_WALLET_OPTIONS_STAGING = process.env.NEXT_PUBLIC_SOCIAL_WALLET_OPTIONS_STAGING || ''
+
+  try {
+    return JSON.parse(IS_PRODUCTION ? SOCIAL_WALLET_OPTIONS_PRODUCTION : SOCIAL_WALLET_OPTIONS_STAGING)
+  } catch (error) {
+    console.error(error)
```
katspaugh, line 37:
```suggestion
    console.error('Error parsing SOCIAL_WALLET_OPTIONS', error)
```

src/hooks/wallets/mpc/useMPC.ts
```diff
@@ -18,9 +18,10 @@ export const useInitMPC = () => {
   useInitSocialWallet()
 
   useEffect(() => {
-    if (!chain || !onboard) {
+    if (!chain || !onboard || !isSocialWalletOptions(SOCIAL_WALLET_OPTIONS)) {
```
katspaugh, line 21: Can we do this check at the entry point to this flow instead?
I.e. not even show the Google button if these env vars aren't set. This is important for forks.

usame-algan: I think the config service toggle only changes if the button is enabled or not. We could use the env vars to decide if the feature should be enabled or not though.

PR #2731, opened 2023-11-02
Fix: new Safe logo for onboard/walletconnect

katspaugh:
A logo w/o a size gets streched in the onboard modal (see https://github.com/safe-global/safe-wallet-web/pull/2727#issuecomment-1789211248).


liliya-soroka:
Verified

PR #2717, opened 2023-10-31
Fix: adjust WalletConnect events

katspaugh:
## What it solves

* Change the "WC show hints" event to trigger on i-icon click + add a "WC hide hints" event - verified
* Add an event for all RPC requests made through WalletConnect – "WC request"
* Don't ignore chain switch errors
* Display errors in two lines: summary and details - changed , the error is displayed in a few lines. The connection code is too long for one line
* Limit the displayed dapp name to 22 characters in the session list - verified
* Reduce the time to show a connect/disconnect message to 1 second (was 2) - verified
* Fix Safe logo for WalletConnect (resolves #2720) - verified
* Show the new WC popup when the old WC Safe App is opened - verified

PR #2732, opened 2023-11-02
Fix: WalletConnect - track only tx and signature requests

katspaugh:
## What it solves

WalletConnect RPC request tracking was added in #2717.
This PR limits the types of requests being tracked to only transaction and signature requests. Otherwise we would track all kinds of "read-only" requests like `eth_getChainid` etc.

Also, track errors only once when they are thrown, not every time they're shown.

### DIFF
--- a/src/components/common/FileUpload/index.tsx
+++ b/src/components/common/FileUpload/index.tsx
@@ -113,10 +113,7 @@ const FileUpload = ({
           sx={{ fill: 'none', color: ({ palette }) => palette.primary.light }}
         />
         <Typography>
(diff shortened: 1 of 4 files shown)


liliya-soroka:
Confirmed:

This PR limits the types of requests being tracked to only transaction and signature requests. Otherwise we would track all kinds of "read-only" requests like eth_getChainid etc. - "WC request" appears only on tx triggering from the connect dapp on web UI 

Also, track errors only once when they are thrown, not every time they're shown. - confirmed

PR #2737, opened 2023-11-02
fix: Update to new recommended nonce endpoint

usame-algan:
## What it solves

Use the new recommended nonce endpoint on the CGW

## How this PR fixes it

- The recommended nonce request on tx creation uses the new `/nonces` endpoint

## How to test it

1. Open a Safe
2. Switch to staging CGW
3. Create a new transaction
4. Observe the recommended nonce is loading and correct

CGW ticket - https://github.com/safe-global/safe-client-gateway/issues/681

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "@safe-global/safe-core-sdk-utils": "^1.7.4",
     "@safe-global/safe-deployments": "1.25.0",
     "@safe-global/safe-ethers-lib": "^1.9.4",
-    "@safe-global/safe-gateway-typescript-sdk": "^3.12.0",
+    "@safe-global/safe-gateway-typescript-sdk": "^3.13.2",
     "@safe-global/safe-modules-deployments": "^1.0.0",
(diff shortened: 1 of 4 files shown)


francovenica:
So the ticket is labeled as "in progress" but since is in ready for QA I still checked it.

I tried in a sepolia safe play with the nonces.

Create tx
Create rejections
Check the recomended nonce when creating a tx
Checking that out of order tx cannot be executed and the safe indicates what nonce needs to be executed first
Check the order in the queue tab
Tx executed in bulk
Batched Tx
The recomended nonce in the settings

Looks good. 

I'll keep the ticket here until the Work in progress tag is removed. If anything changes let me know and I'll re check

usame-algan:
We can merge it as soon as its ready on prod CGW

francovenica:
Did a quick check again, same cases in safes 1.1.1. 
Looks good as well

katspaugh (commented):
Why so many changes? The only necessary one was `fetchRecommendedParams` -> `getNonces` on one single line.

PR #2747, opened 2023-11-07
Fix: redirect /wc URL

katspaugh:
## What it solves

Resolves https://github.com/WalletConnect/web3modal/issues/1455

## How this PR fixes it
We will be redirecting the `/wc?uri=<WC_URI>` URL to `/home?safe=<LAST_OPENED_SAFE>&wc=<WC_URI>`.
If no Safe has been previously opened, it will redirect to `/welcome`.

## How to test
* Go to http://lab.web3modal.com
* Choose Connect -> All wallets -> Safe -> Web app
* Copy the `/wc?uri=<WC_URI>` part from the opened page
* Open this branch's URL and add `/wc?uri=<WC_URI>` to the URL


src/pages/wc.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { parse } from 'querystring'
+import useLastSafe from '@/hooks/useLastSafe'
+import { AppRoutes } from '@/config/routes'
+
+const WcPage: NextPage = () => {
+  const router = useRouter()
+  const lastSafe = useLastSafe()
+
+  useEffect(() => {
+    if (!router.isReady || router.pathname !== AppRoutes.wc) {
```
usame-algan, line 13: I think since this is always on the `/wc` page we don't need to check if the pathname matches that.

katspaugh: I copied that from the index.tsx redirect. It's probably there to avoid the effect kicking in again after the redirect but before the page is unmounted.

katspaugh:
@francovenica this should ideally go live together with seedless.

usame-algan (approved):

liliya-soroka:
image

liliya-soroka:
If no Safe has been previously opened, it will redirect to /welcome. - confirmed

liliya-soroka (approved):

Issue #2452, opened 2023-08-28
Epic: [Seedless-Onboarding]


schmanu:
## What is the feature about
We want to integrate a solution for seedless onboarding by logging in through an social account.

This is the epic issue. Separate issues will be created for each requirement / user story.

## The list of requirements
- [ ] Integrate authenticators
  - [x] Google
  - [ ] (optional) #2457 v2
- [x] Device share
  - [x] Store device share in local storage
- [ ] Recovery
  - [x] #2455 
  - [ ] (optional) #2486 v2
  - [x] Build Single factor recovery
- [x] Wallet integration
  - [x] #2456
- [x] Blockchain interactions (are done via Provider, should work once the Web3Provider can be created)
  - The returned provider does not correctly implement `eth_estimateGas`, we had to patch it's `eth_estimateGas` method to call the readonly provider's `eth_estimateGas` method. 
  - [x] Sign messages
  - [x] Sign transactions
- [ ] Implement designs / user flows
  - [x] #2459
  - [x] #2569
  - [x] New quick Safe creation flow
  - [x] Update Account center
  - [ ] (optional) Add global MFA prompt
  - [x] #2589
  - [x] #2519
- [x] #2458

## Designs/sketches
tbd

## Links
https://www.notion.so/safe-global/Feature-spec-d4b73908e0804f22b9c92565ec8f98c0

francovenica:
This epic has been tested and the tickets with issues were cleared from the board in notion. So this feature is ready to be put into an RC. I'll pass it to done.

The task marked as optional will be tackled in the next iteration

PR #2770, opened 2023-11-08
fix: Add feature flag for social login

usame-algan:
## What it solves

Part of #2452

## How this PR fixes it

Adds a new feature flag `SOCIAL_LOGIN`. If the feature flag is not enabled on a certain chain the Google Login button in the Header and Welcome page will be hidden and the Social Login option in Onboard will be hidden.

Also removes the redirect to safe creation when logging in on the welcome page.

## How to test it

1. Open Safe on staging CGW
2. Switch to Goerli
3. Observe no Google button visible in the header and welcome page
4. Observe no Social Login button visible in the onboard modal

## Screenshots

Screenshot 2023-11-08 at 10 25 15
Screenshot 2023-11-08 at 10 25 23
Screenshot 2023-11-08 at 10 25 30

## Checklist


usame-algan:
We could also show the "old" Connect button in the header then that immediately opens the onboard modal. Wdyt @schmanu @katspaugh ?

katspaugh:
@usame-algan yes, I think that would be good if not too difficult.

src/components/welcome/WelcomeLogin/index.tsx
```diff
@@ -29,13 +32,17 @@ const WelcomeLogin = () => {
         </Typography>
```
schmanu, line 32: Should we also remove or change the text above?
It still says "Choose how you would like to create your Safe Account" or is it okay as you can choose one of the onboard wallets?

src/services/mpc/SocialLoginModule.ts
```diff
@@ -2,6 +2,8 @@ import { _getMPCCoreKitInstance } from '@/hooks/wallets/mpc/useMPC'
 import { getSocialWalletService } from '@/hooks/wallets/mpc/useSocialWallet'
 import { getWeb3ReadOnly } from '@/hooks/wallets/web3'
 import * as PasswordRecoveryModal from '@/services/mpc/PasswordRecoveryModal'
+import { FEATURES, hasFeature } from '@/utils/chains'
+import { ChainInfo } from '@safe-global/safe-gateway-typescript-sdk'
```
schmanu, line 6: `import type` is required here.

src/services/mpc/__tests__/module.test.ts
```diff
@@ -1,13 +1,20 @@
...
 
 import * as web3 from '@/hooks/wallets/web3'
 import * as useMPC from '@/hooks/wallets/mpc/useMPC'
 import { hexZeroPad } from 'ethers/lib/utils'
 
+const mockChain = chainBuilder().build()
+
 describe('MPC Onboard module', () => {
+  beforeAll(() => {
+    jest.spyOn(feature, 'hasFeature').mockReturnValue(true)
```
schmanu, line 14: Could you add the feature to the `mockChain` using the builder instead?

src/components/welcome/WelcomeLogin/index.tsx
usame-algan: I think its fine because there are multiple options in the onboard modal so the user is technically still choosing.

schmanu (approved):

PR #2771, opened 2023-11-08
Optimize docker image

Uxio0:
- Reduce size by using multistage builds
- Run `yarn build` on the build process and not when the docker image starts (it can take a few minutes to complete)


katspaugh (approved):
Looks good, thank you! I assume you have tested it locally?

PR #2778, opened 2023-11-09
fix: Add ramp view to assets page if there are no assets

usame-algan:
## What it solves

Adds a new screen to the assets page with the Safe QR Code and a link for on-ramping.

- [ ] Add correct on ramp safe app link

## Screenshots
Screenshot 2023-11-09 at 09 40 18

## Checklist

### DIFF
--- a/src/components/balances/AssetsTable/index.tsx
+++ b/src/components/balances/AssetsTable/index.tsx
@@ -1,5 +1,14 @@
+import EthHashInfo from '@/components/common/EthHashInfo'
+import QRCode from '@/components/common/QRCode'
+import { AppRoutes } from '@/config/routes'
+import { useCurrentChain } from '@/hooks/useChains'
+import useSafeAddress from '@/hooks/useSafeAddress'
+import { useAppSelector } from '@/store'
+import { selectSettings } from '@/store/settingsSlice'
+import Link from 'next/link'
+import { useRouter } from 'next/router'
 import { type ReactElement, useMemo, useContext } from 'react'
-import { Button, Tooltip, Typography, SvgIcon, IconButton, Box, Checkbox, Skeleton } from '@mui/material'
+import { Button, Tooltip, Typography, SvgIcon, IconButton, Box, Checkbox, Skeleton, Paper, Grid } from '@mui/material'
 import type { TokenInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import { TokenType } from '@safe-global/safe-gateway-typescript-sdk'
 import css from './styles.module.css'
@@ -20,6 +29,7 @@ import CheckWallet from '@/components/common/CheckWallet'
 import useSpendingLimit from '@/hooks/useSpendingLimit'
 import { TxModalContext } from '@/components/tx-flow'
 import TokenTransferFlow from '@/components/tx-flow/flows/TokenTransfer'
+import AddIcon from '@mui/icons-material/Add'
 
 const skeletonCells: EnhancedTableProps['rows'][0]['cells'] = {
   asset: {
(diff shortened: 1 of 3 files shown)


src/components/balances/AssetsTable/index.tsx
```diff
@@ -247,11 +259,55 @@ const AssetsTable = ({
...
+            Add funds directly from your bank account or copy your address to send tokens from a different account.
+          </Typography>
+          <Box bgcolor="background.main" p={2} borderRadius="6px" alignSelf="flex-start" fontSize="14px">
+            <EthHashInfo address={safeAddress} shortAddress={false} showCopyButton hasExplorer avatarSize={24} />
+          </Box>
+          <Box alignSelf="flex-start">
+            {/* TODO: Insert link for Ramp app */}
+            <Link href={{ pathname: AppRoutes.apps.index, query: router.query }} passHref>
+              <Button variant="contained" size="small" startIcon={<AddIcon />}>
+                Buy Crypto
```
katspaugh, line 303:
```suggestion
                Buy crypto
```

🦸🏻‍♂️🐕

src/components/balances/AssetsTable/index.tsx
```diff
@@ -137,6 +147,8 @@ const AssetsTable = ({
     [hiddenAssets, balances.items, showHiddenAssets],
   )
 
+  const hasNoAssets = balances.items.length === 1 && balances.items[0].balance === '0'
```
usame-algan, line 150: Not sure if there is a better way to find this out. Any ideas?

schmanu: I think this is the only way as our endpoint always returns the native token balance. Even if it is 0.

src/components/balances/AssetsTable/index.tsx
```diff
@@ -247,11 +259,55 @@ const AssetsTable = ({
...
+            Add funds to get started
+          </Typography>
+          <Typography>
+            Add funds directly from your bank account or copy your address to send tokens from a different account.
+          </Typography>
+          <Box bgcolor="background.main" p={2} borderRadius="6px" alignSelf="flex-start" fontSize="14px">
+            <EthHashInfo address={safeAddress} shortAddress={false} showCopyButton hasExplorer avatarSize={24} />
+          </Box>
+          <Box alignSelf="flex-start">
+            {/* TODO: Insert link for Ramp app */}
```
schmanu, line 300: In order to find the app we should probably add a tag "on-ramping" to the Safe App so we can search for the appUrl by tag similar to the Safe Governance App links in the header.

schmanu (commented):
Looks good. Before merging we have to adjust it with the link to the Ramp Safe app or if it doesn't exist on the current chain not display the Button.

francovenica:
It works.

The message looks fine, 
the QR code works, 
the address is fine with icon and buttons to copy / block explorer
The "Buy" button only shows if the app is enabled in the network 
The message is gone if you receive tokens and shows up again if you empty your safe of tokens
The message shows if you have only "non-default" tokens and dissapears if you select "All tokens" from the "Token list" dropdown, which I consider the proper behaviour

Question:
Would it be possible to remove the "hide tokens" menu while that message shows up. It makes no sense to show if if there is no table to filter. 
I don't consider this an issue,  is a suggestion

Note:
I enabled Ramp app in sepolia for a minute in django to see that the "Buy" button shows up, which it did, but after disabling the app the button kept showing up. Maybe cache? I'll check tomorrow if it dissapears

usame-algan:
Thanks for testing!

> I enabled Ramp app in sepolia for a minute in django to see that the "Buy" button shows up, which it did, but after disabling the app the button kept showing up. Maybe cache? I'll check tomorrow if it dissapears

I had the same issue when we enabled it on goerli but it disappeared after a while. Definitely a caching issue.

PR #2751, opened 2023-11-07
Tests: Add safe apps tests

mike10ca:
## What it solves

Resolves #2748 

## How this PR fixes it

- Add new Safe Apps
- Adjust Cypress settings

PR #2534, opened 2023-09-21
Epic Seedless-onboarding

schmanu:
## What it solves
Implements seedless onboarding via MPC Core SDK

Resolves #2452 

## How to test

- Login with Google via Header
- Continue with Google via Welcome page
- Set password
- Extract pk
- Recover with password
- Recover with password and remember

## Checklist
* [ ] I've tested the branch on mobile 📱
* [ ] I've documented how it affects the analytics (if at all) 📊
* [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #2782, opened 2023-11-09
fix: Adjust overview widget on dashboard for new design

usame-algan:
## What it solves

Updates the dashboard overview widget to support easy on-ramping

Figma: https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=8786%3A68251&mode=dev

## How to test it

1. Open a Safe
2. Go to the dashboard
3. Observe the updated design
4. Switch to a network where the Ramp Network app is activated
5. Observe a "Buy crypto" button

## Screenshots
Screenshot 2023-11-09 at 13 57 10
Screenshot 2023-11-09 at 13 57 27

## Checklist
* [ ] I've tested the branch on mobile 📱
* [ ] I've documented how it affects the analytics (if at all) 📊
* [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #2783, opened 2023-11-09
Release 1.22.0

schmanu:
# Features
- Seedless-onboarding (#2534)

# Bug Fixes
- redirect /wc URL (#2747)

# Chores
- Optimize docker image (#2771) 

# Tests
- Add safe apps tests (#2751) 
- Update workflow to enable tests run in parallel (#2739) 
- Make tests independent


schmanu:
We cannot merge this yet.
We want to include the following PRs to this release which are currently in QA:
#2782
#2778

francovenica:
Issue:

If you are connected with WC (using a MM phone app) and then, with the social login you change your password, after the page refreshes, the WC new feature shows that the safe is connected to itself

- Login with WC using the MM phone app
- Connect with social login
- Change your password
- See how the page refreshes itself
- Check the WC feature in the topbar
- Change wallet from social login back to WC and check the address

RC1

usame-algan:
> If you are connected with WC (using a MM phone app) and then, with the social login you change your password, after the page refreshes, the WC new feature shows that the safe is connected to itself

This is strange, I can't reproduce it. Can you show an example of the whole flow starting from when you connect WC to MM app?

liliya-soroka:
The error in case of paying from the social login address when not enough balance. The error text is unexpected and says nothing to the user
image

usame-algan:
> > If you are connected with WC (using a MM phone app) and then, with the social login you change your password, after the page refreshes, the WC new feature shows that the safe is connected to itself
> 
> This is strange, I can't reproduce it. Can you show an example of the whole flow starting from when you connect WC to MM app?

After doing it together I can also reproduce this issue on prod. The steps are as following:

1. Open a Safe where your MM mobile app is an owner
2. Connect via WC to that safe
3. Observe the correct address in the header
4. Reload the page
5. Observe WC shows the safe address

Since this also happens on prod I suggest to create a separate issue

francovenica:
So the issue with WC had nothing to do with seedless or the password, it was just the refreshing of the page. Also the issue is already in prod.
Still I'd like to see if we can fix it now so it can go live soon
Here's a ticket for it

https://github.com/safe-global/safe-wallet-web/issues/2790
cc @usame-algan

francovenica:
I've checked the mergin of the tickets regarding assets widget in the dashboard and banner in the assets tab. Also checked the fix on the dashboard assets widget for mobile viewport. 
Those look good

francovenica:
The RC looks good.

Notes:

I think is agreed that the issue of WC will be tackled in a different ticket and released later. I'm ok with that. Hopefully we can release it as a stand alone "hot fix" during the week.
We got the RC about the warning of running out of relayer usage. If that one is merged we should take a quick look after
It was mentioned that the "Buy token" button in the dashboard widget for mobile has a small style issue. Usame said he has to fix that one

katspaugh:
@usame-algan will adjust the staging config.

liliya-soroka:
I have an issue in the case of switching between the Goerli and Gnosis chains when social login is used.

https://github.com/safe-global/safe-wallet-web/assets/338622/b67d3c40-c96c-444e-9a6c-fb1442e6b97a

liliya-soroka:
QR code should be changed if the prefix is off
image

PR #2789, opened 2023-11-10
Fix: show relays even when 0 relays left

katspaugh:
## What it solves

The relay function disappearing when there are 0 relays is not intutive. It's better to show that the quota is over and that it will reset in an hour.

Plus, I've added a warning that the connected wallet should have enough funds (steps on the toes of #2678, but it's just a quick fix until that PR is merged).

Relay:
Screenshot 2023-11-10 at 13 50 41

Wallet:
Screenshot 2023-11-10 at 13 50 46

### DIFF
--- a/src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
+++ b/src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
@@ -55,9 +55,7 @@ const ExecuteForm = ({
   // SC wallets can relay fully signed transactions
   const [canWalletRelay] = useWalletCanRelay(safeTx)
   // We default to relay
-  const [executionMethod, setExecutionMethod] = useState(
-    canWalletRelay ? ExecutionMethod.RELAY : ExecutionMethod.WALLET,
-  )
+  const [executionMethod, setExecutionMethod] = useState(ExecutionMethod.RELAY)
   // The transaction can/will be relayed
   const willRelay = executionMethod === ExecutionMethod.RELAY
   const hasRelays = !!relays?.remaining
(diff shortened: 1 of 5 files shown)


src/components/tx/SponsoredBy/index.tsx
```diff
@@ -11,43 +11,72 @@ export const SPONSOR_LOGOS = {
...
+          <Typography variant="body2" color="primary.light">
+            Transactions per hour:{' '}
+            <Box component="span" sx={{ fontWeight: '700', color: 'text.primary' }}>
+              {relays?.remaining ?? 0} of {relays?.limit ?? 0}
+            </Box>
+            {relays && !relays.remaining && (
+              <Box component="span" sx={{ color: 'error.main' }}>
+                {' '}
+                &mdash; will reset in the next hour
+              </Box>
```
usame-algan, line 65: I think this is slightly redundant since the label already says "Transactions per hour". Maybe we could instead make the "0 of 5" a red color?

src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
```diff
@@ -57,11 +56,10 @@ const ExecuteForm = ({
   const [executionMethod, setExecutionMethod] = useState(ExecutionMethod.RELAY)
 
   // SC wallets can relay fully signed transactions
-  const [walletCanRelay] = useWalletCanRelay(safeTx)
-
+  const [canRelay] = useWalletCanRelay(safeTx)
```
usame-algan, line 59: I think this naming is more confusing now since its not about if the user can relay but if their wallet supports relaying.

katspaugh: Renamed it to `canWalletRelay`.

francovenica:
I noticed that now executing with the connected wallet is the default option, it used to be the relayer. Is this an intentional change?

src/components/tx/SponsoredBy/index.tsx
katspaugh: I added this based on Liliya's feedback. It might be redundant, but it's definitely clearer. IMO we should err on the safe side.

PR #2794, opened 2023-11-13
Fix: remove deprecated mobile pairing

katspaugh:
## What it solves

Mobile pairing will be deprecated on Nov 15, so we can remove it from the code.


src/hooks/wallets/useOnboard.ts
```diff
@@ -43,7 +42,7 @@ export const getConnectedWallet = (wallets: WalletState[]): ConnectedWallet | nu
   const primaryWallet = wallets[0]
   if (!primaryWallet) return null
 
-  const account = primaryWallet.accounts[0]
+  const account = primaryWallet.accounts[primaryWallet.accounts.length - 1]
```
schmanu, line 45: Why do we need the last account instead of the first one now?

src/components/common/WalletInfo/index.tsx
```diff
@@ -108,10 +109,10 @@ export const WalletInfo = ({
 
       {!IS_PRODUCTION && isSocialLogin && (
         <Button onClick={resetAccount} variant="danger" size="small" fullWidth disableElevation>
-          Delete Account
+          Delete account
```
schmanu, line 112: There is a unit test expecting this label that needs to be adjusted.

src/hooks/wallets/useOnboard.ts
katspaugh: Good catch, it’s an unrelated leftover from another branch. I’ll remove it.

schmanu (approved):

liliya-soroka:
do we know why https://verify.walletconnect.com/... 404 appears during dapp connection?
image

liliya-soroka:
I have the issues with wallet WC -
1. can not sign
2. errors in console on disconnect 
https://github.com/safe-global/safe-wallet-web/assets/338622/219dc251-3c04-414a-9727-00e1ae02f213

katspaugh:
Those 404 errors have been there since the beginning.

Regarding signing, what signer wallet are you using? Please make sure it supports Goerli, or try some other chain.

liliya-soroka:
Verified, mobile pairing info removed
connection via keystone, wc, MM works as before

PR #2795, opened 2023-11-13
Fix: prefix switch for receive QR

katspaugh:
## What it solves

Adds a prefix switch for the QR code.

Screenshot 2023-11-13 at 13 22 57


usame-algan (approved):

liliya-soroka:
Verified

PR #2798, opened 2023-11-13
Fix: WC verification wording

katspaugh:
## What it solves

Adjust the warning as per Derek's request.

> :pray:for implementing Verify API - caveat
your language for “mismatch” is a bit strong and also wrong - we do not flag this scenario as “high risk”. Can you pls make the messaging a bit more clear?

### DIFF
--- a/src/components/walletconnect/WcProposalForm/ProposalVerification.tsx
+++ b/src/components/walletconnect/WcProposalForm/ProposalVerification.tsx
@@ -29,7 +29,7 @@ const Validation: {
   },
   INVALID: {
     color: 'error',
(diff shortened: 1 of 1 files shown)


schmanu (approved):

liliya-soroka:
We tried to go with a more general message because we have only an invalid state from the wallet connect without the original problem.
 The case "has a domain that does not match the sender of this request" might be one of the reasons for unverified apps. Would it not be confusing in case of other reasons for "unverified" by Wallet Connect?

liliya-soroka:
Ok, agreed, verified

PR #2803, opened 2023-11-13
chore: Analyze NextJS bundle size action

### LINKED ISSUE #2802: Add bundle analyzer action
## What is the feature about

New dependencies can increase the bundle size of the app. To keep better track of this and possibly be able to prevent large changes e.g. by dynamically importing modules we can add a github action to monitor changes to the bundle size in each PR.

⚠️ Since v19.1.0 the bundle size and first load JS roughly doubled (900kb -> 1.9mb)

## The list of requirements

- Runs the new github action on every PR (whenever the app is built)

## Links

Github Action: https://github.com/hashicorp/nextjs-bundle-analysis

usame-algan:
## What it solves

Resolves #2802

## How this PR fixes it

Adds a new action to analyze bundle size changes of the nextjs app

### DIFF
--- a/.github/workflows/nextjs_bundle_analysis.yml
+++ b/.github/workflows/nextjs_bundle_analysis.yml
@@ -9,11 +9,6 @@ on:
     branches:
       - dev
 
-defaults:
-  run:
-    # change this if your nextjs app does not live at the root of the repo
-    working-directory: ./
-
 permissions:
   contents: read # for checkout repository
   actions: read # for fetching base branch bundle stats
--- a/.github/workflows/nextjs_bundle_analysis.yml
+++ b/.github/workflows/nextjs_bundle_analysis.yml
@@ -1,0 +1,115 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+name: 'Next.js Bundle Analysis'
+
+on:
+  pull_request:
+  push:
+    branches:
+      - dev
+
+defaults:
+  run:
+    # change this if your nextjs app does not live at the root of the repo
+    working-directory: ./
+
+permissions:
+  contents: read # for checkout repository
+  actions: read # for fetching base branch bundle stats
+  pull-requests: write # for comments
+
+jobs:
+  analyze:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install dependencies
+        uses: ./.github/workflows/yarn
+
+      - name: Restore next build
+        uses: actions/cache@v3
+        id: restore-build-cache
+        env:
+          cache-name: cache-next-build
+        with:
+          # if you use a custom build directory, replace all instances of `.next` in this file with your build directory
+          # ex: if your app builds to `dist`, replace `.next` with `dist`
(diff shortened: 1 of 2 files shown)


.github/workflows/nextjs_bundle_analysis.yml
```diff
@@ -0,0 +1,115 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+name: 'Next.js Bundle Analysis'
+
+on:
+  pull_request:
+  push:
+    branches:
+      - dev
```
katspaugh, line 10: Do we need this?

.github/workflows/nextjs_bundle_analysis.yml
```diff
@@ -0,0 +1,115 @@
...
+
+defaults:
+  run:
+    # change this if your nextjs app does not live at the root of the repo
+    working-directory: ./
```
katspaugh, line 15: This is the root, right?

.github/workflows/nextjs_bundle_analysis.yml
```diff
@@ -0,0 +1,115 @@
...
+        uses: actions/cache@v3
+        id: restore-build-cache
+        env:
+          cache-name: cache-next-build
+        with:
+          # if you use a custom build directory, replace all instances of `.next` in this file with your build directory
+          # ex: if your app builds to `dist`, replace `.next` with `dist`
+          path: .next/cache
+          # change this if you prefer a more strict cache
+          key: ${{ runner.os }}-build-${{ env.cache-name }}
```
katspaugh, line 41: This should probably be in the build job in workflows/build?

.github/workflows/nextjs_bundle_analysis.yml
usame-algan: According to the template this should be the default branch which is dev for us so it can compare bundle sizes with dev later on.

.github/workflows/nextjs_bundle_analysis.yml
usame-algan: Yes, and I think its correct since the generated artifact includes all our routes and the global chunk size: https://github.com/safe-global/safe-wallet-web/actions/runs/6853583646

.github/workflows/nextjs_bundle_analysis.yml
usame-algan: I think we can remove this part since our `./.github/workflows/yarn` already caches the result.

.github/workflows/nextjs_bundle_analysis.yml
katspaugh: I mean, you don't need to specify it as per the comment above.

katspaugh (approved):
Lovely jobly

.github/workflows/nextjs_bundle_analysis.yml
usame-algan: Ah ok, I will try to remove it completely and see if it works.

usame-algan: Still works as expected 👍

PR #2805, opened 2023-11-13
Fix: add loading icon to Sign/Execute buttons

francovenica:
## What it solves
When signing/executing for social login, since no popup shows up the user stays in the form for a few seconds not knowing if something is happening. I added a loading icon in the sign/execute button when clicked. The loading icon replaces the text to not messup with the buttons width

## How this PR fixes it
Adds the icon when the button is grayed out (a tx is executing or being signed)

## How to test it
The feature affects signing, execution and execution in bulk
Go to a form as an owner and click the sign/execute/submit, you sould see the loading icon replacing the text
Go to a form an disconnect. The button should NOT show the loading icon
Go to a form and switch to a non-owner. The button should NOT show the loading icon
Go to a fully sign tx with a non-owner. You SHOULD see the loading icon

## Screenshots
Should look like this:
image

## Checklist


src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
```diff
@@ -159,8 +159,8 @@ const ExecuteForm = ({
           {/* Submit button */}
           <CheckWallet allowNonOwner={onlyExecute}>
             {(isOk) => (
-              <Button variant="contained" type="submit" disabled={!isOk || submitDisabled}>
-                Execute
+              <Button variant="contained" type="submit" disabled={!isOk || submitDisabled} sx={{ width: '112px' }}>
+                {submitDisabled ? <CircularProgress size={20} /> : 'Execute'}
```
katspaugh, line 163: There are cases when the button is legitimately disabled, e.g. when you’re not an owner. It shouldn’t be loading in those cases.

I’d recommend checking !isSubmittable instead of submitDisabled. And it might be better to rename it to isExecuting.

src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
```diff
@@ -159,8 +159,8 @@ const ExecuteForm = ({
           {/* Submit button */}
           <CheckWallet allowNonOwner={onlyExecute}>
             {(isOk) => (
-              <Button variant="contained" type="submit" disabled={!isOk || submitDisabled}>
-                Execute
+              <Button variant="contained" type="submit" disabled={!isOk || submitDisabled} sx={{ width: '112px' }}>
```
katspaugh, line 162: min-width is better

katspaugh (approved):
LGTM

PR #2807, opened 2023-11-14
fix: fix and optimise fetching the recovery state

schmanu:
## What it solves
Querying the recovery state does not work.

## How this PR fixes it
- Uses `blockNumber` instead of `blockHash`
- Fixes the endpoint of our tx service
- filters the queueNonces while querying and not afterwards

## How to test it
- Load a Safe that has a recovery tx queued
- Observe the redux state

## Checklist

### DIFF
--- a/src/services/recovery/recovery-state.ts
+++ b/src/services/recovery/recovery-state.ts
@@ -11,6 +11,8 @@ import { hexZeroPad } from 'ethers/lib/utils'
 
 const MAX_PAGE_SIZE = 100
 
+export const normalizeTxServiceUrl = (url: string): string => (url.endsWith('/') ? url : `${url}/`)
+
 export const _getRecoveryQueueItem = async (
   transactionAdded: TransactionAddedEvent,
   txCooldown: BigNumber,
@@ -41,7 +43,7 @@ export const _getSafeCreationReceipt = memoize(
     safeAddress: string
     provider: JsonRpcProvider
   }): Promise<TransactionReceipt> => {
-    const url = `${transactionService}api/v1/safes/${safeAddress}/creation/`
+    const url = `${normalizeTxServiceUrl(transactionService)}api/v1/safes/${safeAddress}/creation/`
 
     const { transactionHash } = await fetch(url).then((res) => {
       if (res.ok && res.status === 200) {
@@ -60,7 +62,9 @@ const queryAddedTransactions = async (
   delayModifier: Delay,
   queueNonce: BigNumber,
   txNonce: BigNumber,
-  startBlockNumber: number,
+  transactionService: string,
+  provider: JsonRpcProvider,
+  safeAddress: string,
 ) => {
   if (queueNonce.eq(txNonce)) {
     // There are no queued txs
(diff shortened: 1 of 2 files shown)


src/services/recovery/recovery-state.ts
```diff
@@ -7,17 +7,10 @@ import type { JsonRpcProvider } from '@ethersproject/providers'
 import type { TransactionReceipt } from '@ethersproject/abstract-provider'
 
 import type { RecoveryQueueItem, RecoveryState } from '@/store/recoverySlice'
+import { hexZeroPad } from 'ethers/lib/utils'
 
 const MAX_PAGE_SIZE = 100
 
-export const _getQueuedTransactionsAdded = (
```
iamacook, line 13: Let's remove the unit test for this.

src/services/recovery/recovery-state.ts
```diff
@@ -48,7 +41,7 @@ export const _getSafeCreationReceipt = memoize(
     safeAddress: string
     provider: JsonRpcProvider
   }): Promise<TransactionReceipt> => {
-    const url = `${transactionService}/v1/${safeAddress}/creation/`
+    const url = `${transactionService}api/v1/safes/${safeAddress}/creation/`
```
iamacook, line 44: It seems as though [there can be a forward slash on some networks](https://safe-client.safe.global/v1/chains). Perhaps we should check if one is required and/or fix the configs.

src/services/recovery/recovery-state.ts
```diff
@@ -63,6 +56,28 @@ export const _getSafeCreationReceipt = memoize(
   ({ transactionService, safeAddress }) => transactionService + safeAddress,
 )
 
+const queryAddedTransactions = async (
```
iamacook, line 59: Can we add a unit test for this?

src/services/recovery/recovery-state.ts
```diff
@@ -63,6 +56,28 @@ export const _getSafeCreationReceipt = memoize(
...
+  startBlockNumber: number,
+) => {
+  if (queueNonce.eq(txNonce)) {
+    // There are no queued txs
+    return []
+  }
+
+  const transactionAddedFilter = delayModifier.filters.TransactionAdded()
+
+  const diff = queueNonce.sub(txNonce).toNumber()
```
iamacook, line 72: We can move this into the `if` below.

src/services/recovery/recovery-state.ts
```diff
@@ -63,6 +56,28 @@ export const _getSafeCreationReceipt = memoize(
...
+  if (transactionAddedFilter.topics) {
+    const queueNonceFilter = Array.from({ length: diff }, (_, idx) => hexZeroPad(txNonce.add(idx).toHexString(), 32))
```
iamacook, line 74: Can we add a comment explaining this?

src/services/recovery/recovery-state.ts
```diff
@@ -72,21 +87,17 @@ export const getRecoveryState = async ({
   safeAddress: string
   provider: JsonRpcProvider
 }): Promise<RecoveryState[number]> => {
-  const { blockHash } = await _getSafeCreationReceipt(rest)
-
-  const transactionAddedFilter = delayModifier.filters.TransactionAdded()
+  const { blockNumber } = await _getSafeCreationReceipt(rest)
```
iamacook, line 90: Let's move this into `queryAddedTransactions` so it doesn't call `_getSafeCreationReceipt` when no queued transactions exist.

iamacook (commented):
Looks great so far! I'm sure you're aware but there are some breaking tests.

src/services/recovery/recovery-state.ts
schmanu: I added more tests for it's outside function effectively also testing this logic.

src/services/recovery/recovery-state.ts
```diff
@@ -7,16 +7,11 @@ import type { JsonRpcProvider } from '@ethersproject/providers'
...
 const MAX_PAGE_SIZE = 100
 
-export const _getQueuedTransactionsAdded = (
-  transactionsAdded: Array<TransactionAddedEvent>,
-  txNonce: BigNumber,
-): Array<TransactionAddedEvent> => {
-  // Only queued transactions with queueNonce >= current txNonce
-  return transactionsAdded.filter(({ args }) => args.queueNonce.gte(txNonce))
-}
+export const normalizeTxServiceUrl = (url: string): string => (url.endsWith('/') ? url : `${url}/`)
```
iamacook, line 14: Nit: there is a `trimTrailingSlash` function in `@/utils/url.ts` that we could export and use instead of this. We would then have to append the `/` within `_getSafeCreationReceipt`.

iamacook (approved):

PR #2809, opened 2023-11-14
Chore: prod build on staging (main)

katspaugh:
## What it solves

In order to be able to test our app with production settings before a release, we'll be building and deploying a production version on staging from the main branch.

With this setup, https://safe-wallet-web.staging.5afe.dev can be used as a close-to-prod environment for pre-release testing.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -40,6 +40,7 @@ jobs:
       - uses: ./.github/workflows/build
         with:
           secrets: ${{ toJSON(secrets) }}
+          prod: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1


usame-algan (approved):

francovenica:
The only issue I had in this environment is that I cannot use social login because of an authorization error. I don't have this issue in the current staging env.

I'm not sure if this is because of this being a PR environment, or if if has to do with the setup done in this PR. 
If it is the former, then I'm ok with approving this ticket
image

katspaugh:
@schmanu we need to whitelist the domain right?

usame-algan:
Yes this PR deployment is not whitelisted but since the branch name contains a `-` Google doesn't allow it to be whitelisted.

katspaugh:
https://safe-wallet-web.staging.5afe.dev/ needs to be white-listed, not this branch.

PR #2820, opened 2023-11-20
[E2E] Add Safe - qr code + ENS name + name validation

francovenica:
## What it solves

Resolves #
Adds name validation for negative case
Adds ENS name validation
Adds QR code with valid and invalid file upload

## How this PR fixes it
Name validation: inputs a name with 51 characters and checks the error
ENS name: Enters an ENS name and checks that it is translated to a valid safe address
QR code: Uploads a valid QR code with a sepolia safe and checks it is translated correctly
QR code invalid file: It uploads a blank square PNG and checks the error "The QR could not be read"

## Notes
I had to add a data test id for the QR code button, but I wasn't able to find how to do it only in the add safe form, so it is in all address inputs accross the site. If it  is a way to make it so only is in the Add safe form I'd like a suggestion on how to do it.

## Checklist


cypress/e2e/smoke/load_safe.cy.js
```diff
@@ -42,26 +42,35 @@ describe('Load Safe tests', () => {
...
+  })
+
+  it('Verify ENS name is translated to a valid address', () => {
+    safe.inputAddress(constants.ENS_TEST_ETH)
+    safe.verifyAddressInputValue(constants.ETH_ENS_SAFE_ADDRESS_7)
+    safe.verifyNextButtonStatus('be.enabled')
+    safe.clickOnNextBtn()
+  })
+
+  it('Verify a valid QR code is accepted', () => {
```
francovenica, line 62: I need some help here. 
If I leave it commented this way then the ScanQRCode function runs just fine. But for some reason if I uncomment all those lines, the function ScanQRCode is ignored, the VerifyAddressInputValue function (that checks if the address in the input is valid) kicks in first and of course fails since the QR code was not uploaded. I don't know why this happens

mike10ca: The reason for it is that ETH_ENS_SAFE_ADDRESS_6 does not exist it throws the error as undefined. To fix it add ETH_ENS_SAFE_ADDRESS_6 to constants.js. After this fix everything should work as expected.

cypress/e2e/pages/load_safe.pages.js
```diff
@@ -52,13 +55,23 @@ export function verifyIncorrectAddressErrorMessage() {
   cy.get(addressInput).parent().prev('label').contains(invalidAddressFormatErrorMsg)
 }
 
+export function verifyNameLengthErrorMessage() {
+  cy.get(nameInput).parent().prev('label').contains(invalidAddressNameLengthErrorMsg)
+}
+
+export function ScanQRCode(image) {
```
mike10ca, line 62: Please move locators to top of the document to follow established practise and reuse them in the code.

cypress/e2e/smoke/load_safe.cy.js
francovenica: I wrote the wrong name of the constant. Fixed. Thanks

mike10ca (approved):

PR #2821, opened 2023-11-20
Chore: run e2e tests conditionally on a comment

### LINKED ISSUE #2800: Manually trigger E2E action in PRs
## What is the feature about

The E2E test job runs automatically on every commit. Every run takes between 10 and 15 minutes with parallelized config. This can be a bit too excessive in case there are multiple commits in a short span or if a commit only changes a specific place inside the app and doesn't affect other views that are tested through E2E.

⚠️ Github offers a way to manually trigger actions through the `workflow_dispatch` option but without further adjustment it has to be triggered outside of the PR view for a target branch which is not exactly what we want. Instead it should be possible to do this inside a PR e.g. through adding a comment to the PR or assigning a label. ⚠️ 

## The list of requirements

- Change the E2E job to be manually triggered in PRs
- Only allow PR merges if the job ran successfully at least once
- Keep it automated on release branches / when merging into main
- Run it after the lint and unit test job since these are faster

katspaugh:
## What it solves

Resolves #2800

## How this PR fixes it

This PR makes e2e tests run only if you comment `Cypress run` in a PR.


katspaugh:
Cypress run

katspaugh:
Cypress run

PR #2822, opened 2023-11-20
Fix: add help link in footer

katspaugh:
## What it solves

Otherwise, there's no way to access the Help page form the welcome page.

Screenshot 2023-11-20 at 11 09 10


src/components/common/Footer/index.tsx
```diff
@@ -67,6 +67,9 @@ const Footer = (): ReactElement | null => {
             <li>
               <FooterLink href={getHref(AppRoutes.settings.index)}>Preferences</FooterLink>
             </li>
+            <li>
+              <FooterLink href={HELP_CENTER_URL}>Help</FooterLink>
```
usame-algan, line 71: I would suggest to make it an external link and open in a new tab. Should we also track its usage?

usame-algan (approved):

liliya-soroka:
Verified

PR #2826, opened 2023-11-20
Tests: Add Chrome version to smoke tests workflow

### LINKED ISSUE #2825: Specify Chrome version in automation tests
Recently it was noticed that some machines started to fail when running Cypress tests in parallel. This was due to different version of Chrome passed by random machines, where Cypress requires all machines pass identical environment. The fix was to use specific version of Chrome and add this step to our test workflows. 

Part of this task also included increasing number of test retries to reduce flaky tests. This will be observed and adjusted as needed.

mike10ca:
## What it solves

Resolves #2825 

## How this PR fixes it

Recently it was noticed that some machines started to fail when running Cypress tests in parallel. This was due to different version of Chrome passed by random machines, where Cypress requires all machines pass identical environment. The fix was to use specific version of Chrome and add this step to our test workflows. 

Part of this task also included increasing number of test retries to reduce flaky tests. This will be observed and adjusted as needed.

### DIFF
--- a/.github/workflows/ondemand.yml
+++ b/.github/workflows/ondemand.yml
@@ -18,6 +18,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - uses: browser-actions/setup-chrome@v1
+
       - uses: ./.github/workflows/yarn
 
       - name: Install Cypress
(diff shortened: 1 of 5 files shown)


katspaugh:
More context on this for posterity:
GitHub regularly update their images, and there's no way to pin a specific version. 2 days ago was such an update that included a new Chrome version: https://github.com/actions/runner-images/pull/8819

katspaugh (approved):

PR #2829, opened 2023-11-21
Fix: double scroll bar in safe apps

### LINKED ISSUE #2827: Hide iframe scrollbar by default
## Bug description

If the Safe{WALLET} dapp integration contains an in-dapp scrollbar, then the outer iframe scrollbar is really irritating.

The latter should — by default — be switched off. Perhaps there should be a corresponding flag in the `manifest.json` configuration to show/hide the outer iframe scrollbar. 

## Environment
 - Browser: Brave
 - Wallet: MetaMask
 - Chain: Avalanche mainnet

## Steps to reproduce
 1. Go to https://app.safe.global/share/safe-app?appUrl=https%3A%2F%2Fsafe.xpowermine.com&chain=avax

2. Accept dapp and observe 

3. If only a single (outer) scrollbar of the iframe is visible, switch to e.g. the *about* tab.

## Expected result
Single *inner* scrollbar of the dapp (if it doesn't fit into the window height).

## Obtained result
Both the *inner* scrollbar of the dapp and the *outer* scrollbar of the iframe are visible. 

## Screenshots

Observed: 

Expected:

katspaugh:
## What it solves

Resolves #2827

## How this PR fixes it
The iframe was inline-block which created an extra 6px space below it, causing an unnecessary scrollbar on the body.

### DIFF
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -35,7 +35,6 @@ button {
 :root {
   --header-height: 52px;
   --footer-height: 67px;
-  --import-export-widget-height: 166px;
 }
 
 input::-webkit-outer-spin-button,
--- a/src/components/safe-apps/AppFrame/styles.module.css
+++ b/src/components/safe-apps/AppFrame/styles.module.css
@@ -4,6 +4,7 @@
 }
 
 .iframe {
+  display: block;
   height: 100%;
   width: 100%;
   overflow: auto;


src/styles/globals.css
```diff
@@ -35,7 +35,6 @@ button {
 :root {
   --header-height: 52px;
   --footer-height: 67px;
-  --import-export-widget-height: 166px;
```
katspaugh, line 38: This wasn't used anywhere.

PR #2830, opened 2023-11-21
Fix: Safe App share URL not always working

katspaugh:
## What it solves

Safe Apps share URL sometimes redirected to the home page due to a "race condition" when trying to read the chain from the URL.

I've removed the unnecessary chain-detecting logic and the redirect, relying on our existing `useCurrentChain` hook instead.

**Caveat**: if the wrong/non-supported chain is pasted in the URL (e.g. avax while on dev), it will result in an infinite loader. On prod, it would result in a redirect, which isn't much user-friendlier anyway.

## How to test

The original bug happened randomly, so it's hard to reproduce. It's sufficient to test that share links still generally work.


francovenica:
I opened a bunch of apps using the share link in different networks and I didn't see the issue and worked fine.

I'll recheck in prod once deployed to check that the infinite loader doesn't happen

katspaugh:
To clarify, I meant that the current behavior on prod for unknown/bogus chains is to redirect to the index page.
After this PR, it will be an infinite loader.

PR #2305, opened 2023-07-26
Feat: emoji identicons

katspaugh:
Experimental feature: deterministic emojis for Ethereum addresses.

## What it solves

For each displayed address, an emoji is derived based on the address. This helps identify addresses more easily and helps memorize which is which.

## How this PR fixes it

This PR adds an experimental toggle in the Appearance Settings that enables address emojis (off by default).
Screenshot 2023-07-26 at 11 47 57

## Screenshots
Screenshot 2023-07-26 at 11 58 04
Screenshot 2023-07-26 at 11 57 45
Screenshot 2023-07-26 at 11 57 24
Screenshot 2023-07-26 at 11 56 46

PR #2834, opened 2023-11-21
Fix: remove experimental address emoji

katspaugh:
## What it solves

Removing the emoji Safe avatars due to very low usage and no overwhelmingly positive feedback. Originally introduced in #2305.

Stats from Sep 1 to Nov 20:
Screenshot 2023-11-21 at 10 43 40


francovenica:
Looks good. The setting is gone

kinda sad about this one, not gonna lie

PR #2838, opened 2023-11-21
Feat: copy addresses on click

katspaugh:
## What it solves

In all places where we display Ethereum addresses, make them clickable to copy to the clipboard.

Screenshot 2023-11-21 at 12 52 31

Screenshot 2023-11-21 at 12 55 12

This is congruent with how other dapps do it, and makes it much easier for the user to copy addresses.

### DIFF
--- a/src/components/common/CopyAddressButton/index.tsx
+++ b/src/components/common/CopyAddressButton/index.tsx
@@ -1,19 +1,20 @@
-import { type ReactElement } from 'react'
-
+import type { ReactNode, ReactElement } from 'react'
 import CopyButton from '../CopyButton'
 
 const CopyAddressButton = ({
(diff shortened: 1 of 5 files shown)


katspaugh:
Good point, I’ll disable it in those instances.

francovenica (approved):

liliya-soroka:
It is hard to select an address from the address book , because of the copy function for the enter address fields.
I suggest to disable the option for the enter fields
https://github.com/safe-global/safe-wallet-web/assets/338622/0b8e1e72-8bdd-485e-9d2b-615cf48cdd16

katspaugh:
Definitely, @liliya-soroka, will do!

PR #2840, opened 2023-11-21
Analytics: update Privacy policy for Spindl + refactor legal pages

katspaugh:
## What it solves

I've cleaned up the components folder a little bit and remove unnecessary directories.

Also updated the Privacy Policy to include information about Spindl and the tracking of signer addresses.

## How to test it

Just click through pages like Terms, Imprint, Cookie policy etc and check that nothing has changed visually.


francovenica:
I don't consider this issues, but different behaviour from how it is now in prod. 
I don't mind deploying them as they are. Still for the record:

In "Privacy", these links work as anchors: they scroll you to the point you want to see. In this PR they refresh the page completely before scrolling you down
image

Some sub-bulletpoints have a larger font size than what it is in Prod. Here in privacy again:
image
image

katspaugh:
Fixed the links, the font changes are intended.

francovenica:
The links are still reloading the page. The URL that loads for the links it seems they are being tracked I guess? 
Let me know if it is intentional
image

katspaugh:
It's not intentional, some weird stuff is going on with these links.

katspaugh:
Fixed by making them JS links.

PR #2842, opened 2023-11-21
Analytics: add Spindl

katspaugh:
## What it solves

Add a new tracking script from Spindl in lite mode (cookie-less). It will be initialized on page load along with GTM.


francovenica:
a bit lost in this ticket. Is a draft so I assume there is more work to be done.
I cannot find anything in the dev console about spindl to see if the script is being loaded or not

katspaugh:
It's now ready for QA.

liliya-soroka:
A few comments:
1. page_view - I did not find a way to see the whole URL to check it . Maybe we can add pagePath?
2. my macOS version is detected incorrectly for some reason ( 10.15.7 instead of 14.1.1)
3. Are we planning to add a connected wallet label in the future? 

Otherwise
1. Page_view is triggered on each page opening 
2. Wallet_connected is triggered on wallet connection ( MM, Ledger, WC)

katspaugh:
The URL is just truncated in the logs visually, the path is there.

> Are we planning to add a connected wallet label in the future?

Yes.

PR #2849, opened 2023-11-22
Analytics: add a Confirm transaction event

katspaugh:
## What it solves

In #2787, we added a new event `Create transaction`. Since a signature is required for both tx creation and confirmation, we cannot use the existing `Off-chain signature` event to track signatures for existing transactions.
So I've added another event, `Confirm transaction` to track specifically those kind of signatures.


katspaugh:
Verified:
Screenshot 2023-11-23 at 07 38 37

PR #2850, opened 2023-11-22
refactor: recovery loading + trigger

iamacook:
## What it solves

Refetching recovery logic after interaction with the module.

## How this PR fixes it

The data loading has been refactored into a provider that exposes a `refetch` function that is passed to the relevant transaction dispatchers for direct module interaction and called after transaction success.

For Safe transactions, the `to` address is dispatched alongside the `TxEvent.PROCESSED` event and, providing it matches a recovery module, it refetches the recovery data.

Note: this does not work for relayed transactions as we do not readily have the `to` address available in the logic flow, which may require a larger refactor. This has been marked with a comment for now to revisit.

## How to test it

Observe data refetching when editing a recovery module, as well as skipping a recovery attempt.

## Checklist


schmanu (approved):

katspaugh (commented):
Looks good.
Only two suggestions:
* Rename the context to just RecoveryContext
* Add a helper hook to wrap `useContext(RecoveryLoaderContext).state`

iamacook:
@katspaugh, as requested f982e6804315266b205f0cd3fde2c469f574774a removes the recovery slice in favour of context-located state and instead of passing `to` with each `TxEvent.PROCESSED` event, the details are manually fetched.

iamacook:
> Looks good. Only two suggestions:
> 
>     * Rename the context to just RecoveryContext
> 
>     * Add a helper hook to wrap `useContext(RecoveryLoaderContext).state`

I've included both adjustments in fbccae45bb0bb870f2920fa541160fb1d88ebbb1.

PR #2866, opened 2023-11-23
fix: filter Zodiac-deployed Delay Modifiers

iamacook:
## What it solves

Resolves Zodiac-deployed Delay Modifiers from showing as recovery guardians

## How this PR fixes it

The modules of a Delay Modifier are iterated over checked whether they are other Zodiac contracts. (Zodiac only either deploy the Delay Modifier _without_ modules and/or with [Zodiac contracts as modules](https://github.com/gnosis/zodiac-safe-app/blob/e5d6d3d251d128245104ddc638e26d290689bb14/packages/app/src/views/AddModule/wizards/RealityModule/service/moduleDeployment.ts#L88).) In contrasts, we deploy the Delay Modifier with non-Zodiac contracts added enables as modules.

## How to test it

Enable a Delay Modifier via the Zodiac Safe App directly, or alternatively by enabling the Reality module. Observe that neither appears in the recovery settings of the Safe.

## Checklist

### DIFF
--- a/src/components/settings/Recovery/index.tsx
+++ b/src/components/settings/Recovery/index.tsx
@@ -69,6 +69,8 @@ export function Recovery(): ReactElement {
   const { setTxFlow } = useContext(TxModalContext)
   const [recovery] = useRecovery()
 
+  console.log(recovery)
+
   const rows = useMemo(() => {
     return recovery?.flatMap((delayModifier) => {
       const { guardians, txCooldown, txExpiration } = delayModifier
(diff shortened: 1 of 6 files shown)


src/components/settings/Recovery/index.tsx
```diff
@@ -69,6 +69,8 @@ export function Recovery(): ReactElement {
   const { setTxFlow } = useContext(TxModalContext)
   const [recovery] = useRecovery()
 
+  console.log(recovery)
```
schmanu, line 72: 👀

schmanu (approved):
This seems to be a good approach for this.
Alternatively we could check that every moduleAddress is a `Safe` or `getCode({address: moduleAddress})` is `undefined`.

PR #2874, opened 2023-11-24
Fix: wallet chain in wallet info

katspaugh:
The chain indicator in the WalletInfo popup should show the wallet's network, not Safe network.

Screenshot 2023-11-24 at 09 45 19

### DIFF
--- a/src/components/common/WalletInfo/index.tsx
+++ b/src/components/common/WalletInfo/index.tsx
@@ -61,7 +61,7 @@ export const WalletInfo = ({
   return (
     <Box className={css.container}>
       <Box className={css.accountContainer}>
-        <ChainIndicator />
+        <ChainIndicator chainId={wallet.chainId} />
 
         <Box className={css.addressContainer}>
           {isSocialLogin ? (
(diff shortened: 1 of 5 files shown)


src/components/common/WalletInfo/index.tsx
```diff
@@ -61,7 +61,7 @@ export const WalletInfo = ({
   return (
     <Box className={css.container}>
       <Box className={css.accountContainer}>
-        <ChainIndicator />
```
schmanu, line 64: Should we add a small testcase for this fix to the WalletInfo unit test?

schmanu (approved):
Nice!

src/components/common/WalletInfo/index.tsx
katspaugh: It will be redesigned in #2678, so I wouldn't at this time.

schmanu: Makes sense. Vamos then :)

francovenica:
LGTM:
image

PR #2876, opened 2023-11-24
Fix: enable spending limits on Sepolia

katspaugh:
## What it solves

Upgrade to [the latest safe-modules-deployments](https://github.com/safe-global/safe-modules-deployments/releases/tag/v1.2.0) to enable Spending Limits on Sepolia.


francovenica:
Spending limit looks fine.

Created spending limit/edited/removed
For onwner and non-owners
Use of the allowance
Tested teset time
...

PR #2877, opened 2023-11-24
Fix: use wallet RPC for readonly requests

### LINKED ISSUE #2000: Investigate always using wallet provider when connected

katspaugh:
This reverts commit 9bfa05f1fdec3fffc0f5263d705a46beffe5b7d2.
Originally submitted in #2418, then reverted to derisk a release.

## What it solves

We're running out of RPC limits on some chains (Gnosis Chain), and this change would fix that, along with addressing user privacy concerns for those who don't want to use providers like Infura.

Resolves #2000.


src/hooks/wallets/useInitWeb3.ts
```diff
@@ -6,29 +6,33 @@ import { createWeb3, createWeb3ReadOnly, setWeb3, setWeb3ReadOnly } from '@/hook
...
     }
-    const web3ReadOnly = createWeb3ReadOnly(rpcUri, customRpcUrl)
-    setWeb3ReadOnly(web3ReadOnly)
-  }, [rpcUri, customRpcUrl])
+
+    // Wait for wallet to be connected
+    const timeout = setTimeout(() => {
+      const web3ReadOnly = createWeb3ReadOnly(chain.rpcUri, customRpcUrl)
+      setWeb3ReadOnly(web3ReadOnly)
+    }, READONLY_WAIT)
```
iamacook, line 34: Why do we need to wait if there is no wallet connected?

katspaugh: Becuase the wallet takes some time to connect, we don't want to initialize a read-only provider (e.g. Infura) and make the first few RPC calls before a connected wallet kicks in.

iamacook (approved):

katspaugh:
As soon as you connect a wallet, it will use the wallet’s RPC. Without a wallet connection, the same behavior as before.

francovenica:
I set RPC1 in the safe settings and my Wallet has RPC2
If I'm connected I don't see any of the RPC1 calls and when I disconnect or change network in my wallet (so it becomes read only) then I start to see RPC1 calls. All this is expected

Issue:
I see mainnet infura calls in polygon along with polygon-infura calls. Also I get errors regarding the wallet connection often.
I've been switching back and forth changing the wallets network, also I've been connecting/disconnecting the wallet
image

image

image

katspaugh:
@francovenica I've increased the delay before initializing Infura to 3 seconds. If a wallet is (re)connected within the first 3 seconds, it will use the wallet RPC instead.

Please note that you will still see some initial Infura requests anyway, because web3-onboard does them. AFAIK there's no way to disable that.

francovenica:
I tried with poligon, switching to that network, waiting for 3 secs and see the polygon-mainnet.infura calls. If I change to the polygon network and quickly connect to my MM wallet, I don't see those calls anymore, assuming is using the wallet's RPC.
So that's good.

Note: 
I still have the issues I reported a while back in my previous comment, the app crashing or error messages like this one. 
This happens when you switch from polygon to eth and vice versa constantly.
image
image

PR #2881, opened 2023-11-27
fix: don't filter expired recovery attempts from queue

iamacook:
## What it solves

Resolves [malformed queue](https://www.notion.so/safe-global/Wrong-message-about-expired-recovery-in-the-next-recovery-tx-7a61aef7aa1949f98cfcedad587b6588)

## How this PR fixes it

Expired recovery attempts are no longer filtered from the queue.

The Delay Modifier requires the "next" transaction to be executed/cancelled before the [`txNonce` + n transaction can be executed](https://github.com/gnosis/zodiac-modifier-delay/blob/main/contracts/Delay.sol#L195-L198).

Note: missed instances of "skipped" have been changed to "cancelled" as per previous requirements.

## How to test it

1. Ensuring the recovery module is enabled with a low expiration period, queue two recovery attempts.
2. After expiration has passed, observe both attempts remaining in the queue and the first cancellable, the second not.
3. After cancelling the first, the second is then executable.

## Checklist

### DIFF
--- a/src/components/recovery/ExecuteRecoveryButton/index.tsx
+++ b/src/components/recovery/ExecuteRecoveryButton/index.tsx
@@ -52,7 +52,7 @@ export function ExecuteRecoveryButton({
         const isDisabled = !isOk || !isExecutable
 
         return (
-          <Tooltip title={isDisabled ? 'Previous recovery attempts must be executed or skipped first' : null}>
+          <Tooltip title={isDisabled ? 'Previous recovery attempts must be executed or cancelled first' : null}>
             <span>
               {compact ? (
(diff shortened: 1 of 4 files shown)


schmanu (approved):
Looks good :)
But I think we should also offer to skip transactions for non-owners.

It could happen that the recoverers forget to execute in time and then the Safe would be stuck without skipping transactions.
Currently from our interface only the Safe owner can skip the expired recoveries.
There is a public function [skipExpired](https://github.com/gnosis/zodiac-modifier-delay/blob/main/contracts/Delay.sol#L203) that we could use.

PR #2884, opened 2023-11-27
fix: show "none"/"never" for no cooldown/expiration

iamacook:
## What it solves

Resolves [missing label for cooldown/expiration](https://www.notion.so/safe-global/Never-is-not-displayed-as-Expiry-6e1056ce860d48f4be043025afb19fc6)

## How this PR fixes it

If the cooldown or expiration is set to `0`, the periods are labelled as "none"/"never" respectively.

## How to test it

1. Ensure recovery is enabled with a cooldown of `0` and expiration of `0`.
2. Observe a cooldown period of "none" and expiration of "never" in the recovery settings.

## Screenshots

image

## Checklist

### DIFF
--- a/src/components/settings/Recovery/index.tsx
+++ b/src/components/settings/Recovery/index.tsx
@@ -86,7 +86,7 @@ export function Recovery(): ReactElement {
             },
             [HeadCells.TxCooldown]: {
               rawValue: txCooldownSeconds,
-              content: <Typography>{txCooldownSeconds === 0 ? 'never' : getPeriod(txCooldownSeconds)}</Typography>,
+              content: <Typography>{txCooldownSeconds === 0 ? 'none' : getPeriod(txCooldownSeconds)}</Typography>,
             },
             [HeadCells.TxExpiration]: {
               rawValue: txExpirationSeconds,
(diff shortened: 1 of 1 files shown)


src/components/settings/Recovery/index.tsx
```diff
@@ -86,11 +86,11 @@ export function Recovery(): ReactElement {
             },
             [HeadCells.TxCooldown]: {
               rawValue: txCooldownSeconds,
-              content: <Typography>{getPeriod(txCooldownSeconds)}</Typography>,
+              content: <Typography>{txCooldownSeconds === 0 ? 'none' : getPeriod(txCooldownSeconds)}</Typography>,
```
usame-algan, line 89: The minimum delay value is 60 afaik. How can this case happen?

iamacook: On a contractual level, the `txCooldown` [can be any value](https://github.com/gnosis/zodiac-modifier-delay/blob/main/contracts/Delay.sol#L87-L93). (`txExpiration` is [limited to `0` or `>=60` seconds](https://github.com/gnosis/zodiac-modifier-delay/blob/main/contracts/Delay.sol#L100-L103).)

usame-algan (approved):

PR #2886, opened 2023-11-27
fix: refetch recovery state via state changes

iamacook:
## What it solves

Resolves stale state when [cancelling recovery](https://www.notion.so/Delay-in-updating-Account-recovery-tx-after-it-was-cancelled-6edc1878f22d404e9830f9af7239cce7) or [updating the settings](https://www.notion.so/safe-global/A-manual-page-refresh-is-required-to-update-Account-recovery-settings-after-Edit-recovery-tx-executi-4b38fcfcfa37454bb5a375f17fac911c)

## How this PR fixes it

If the most recent transaction interacted with either a Delay Modifier or the MultiSend contract, the state is refetched. This has the benefit of not needing to fetch the transaction details from the gateway, which was sometimes throwing a 404 due to missing proposal and/or index.

## How to test it

1. Cancel a recovery and observe the transaction be removed from the pending recovery list.
2. Edit one/multiple settings of the recovery module and observe the settings update.

## Checklist


src/components/recovery/RecoveryContext/index.tsx
```diff
@@ -34,31 +37,41 @@ export function RecoveryProvider({ children }: { children: ReactNode }): ReactEl
       return
     }
 
-    return txSubscribe(TxEvent.PROCESSED, async (detail) => {
-      if (!detail.txId) {
-        return
-      }
+    // Listen to history polls (only occuring when the txHistoryTag changes)
+    const listener = dispatch(
```
usame-algan, line 41: Is there a way we can move this logic inside the `useRecoveryState` hook without a listener e.g. by using the `useTxHistory` hook as a dependency to the `useAsync` call?

Having this manual `refetch` toggle seems like a workaround to me.

iamacook: I've migrated the logic and added test coverage for it in 3e63cbe13722ee56c7fed9f0fd5004d433d3356d.

iamacook: After investigating this with @usame-algan, we will revert this to a listener for now as there are a lot of race conditions involved with initially loading the recovery state: delay modifiers, chain config, Safe info and provider. All of which need to be loaded beforehand but merging the refetching logic with this causes issues with loading the initial state. We intend to revisit this when we have the capacity.

usame-algan (approved):

PR #2887, opened 2023-11-27
fix: expose `skipExpired` for Guardians

### LINKED ISSUE #2882: [Recovery] Enable `skipExpired` for non-owners
## What is the feature about

We currently allow Safe owners to cancel recovery attempts via `setTxNonce`. However, non-owners may attempt recovery and their attempt expire. We should expose the option to `skipExpired` for non-owners.

## The list of requirements

- [ ] Create new dispatcher to call `skipExpired` that e.g. retrieves the `txNonce` of the contract and compares it to the attempt being cancelled.
- [ ] Call the aforementioned dispatcher if there are >1 expired attempts, of which the the second or up is being cancelled.

iamacook:
## What it solves

Resolves #2882

## How this PR fixes it

This allows Guardians to cancel recovery attempts if they have expired. A new `dispatchRecoverySkipExpired` dispatcher is now called when connected to the Safe as a Guardian and clicking "Cancel" on an expired recovery attempt.

## How to test it

1. Ensure recovery is enabled and an expired proposal exists.
2. Observe "Cancel" enabled which, when clicked, skips all expired recovery attempts.

## Checklist

### DIFF
--- a/src/components/recovery/CancelRecoveryButton/index.tsx
+++ b/src/components/recovery/CancelRecoveryButton/index.tsx
@@ -14,6 +14,7 @@ import useOnboard from '@/hooks/wallets/useOnboard'
 import { logError, Errors } from '@/services/exceptions'
 import { RecoveryContext } from '../RecoveryContext'
 import { useIsGuardian } from '@/hooks/useIsGuardian'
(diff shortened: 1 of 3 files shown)


schmanu (approved):
💯

PR #2888, opened 2023-11-27
fix: only display proposal modal to Guardians

iamacook:
## What it solves

Resolves proposal modal opening for non-Guardians

## How this PR fixes it

The recovery proposal now _only_ opens if a Guardian is connected.

Note: the `RecoveryModal` was also refactored be not take `children` for future safety.

## How to test it

1. Ensuring recovery is enabled, open a Safe as a Guardian and observe the proposal modal.
2. Switch to a non-Guardian and observe no proposal modal.

## Checklist

### DIFF
--- a/src/components/recovery/RecoveryModal/index.tsx
+++ b/src/components/recovery/RecoveryModal/index.tsx
@@ -44,12 +44,12 @@ export function _RecoveryModal({
 
   // Trigger modal
   useEffect(() => {
-    if (!isSidebarRoute && (!isOwner || !isGuardian)) {
+    if (!isSidebarRoute) {
       return
     }
 
     setModal(() => {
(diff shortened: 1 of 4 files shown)


src/components/recovery/RecoveryModal/index.tsx
```diff
@@ -46,7 +44,7 @@ export function _RecoveryModal({
 
   // Trigger modal
   useEffect(() => {
-    if (!isSidebarRoute) {
+    if (!isSidebarRoute && (!isOwner || !isGuardian)) {
```
usame-algan, line 47: I think we need to split this condition or use an `||` because the InProgressCard is shown to non-owners and non-guardians. We have a test-case to cover that but it doesn't fail because we are not mocking `isSidebarRoute`.

usame-algan (approved):

PR #2894, opened 2023-11-28
Feat: custom ENS lookup for Sepolia

### LINKED ISSUE #2687: [Migration to Sepolia]ENS doesn't work on Sepolia
Could we check why the ENS domain doesn't work on Sepolia?

katspaugh:
## What it solves

Resolves #2687.

Ethers v5.7 unfortunately doesn't support Sepolia out of the box, and upgrading to the latest version is complicated.
So as a temporary fix, I'm adding a custom resolver for Sepolia.

### DIFF
--- a/src/services/ens/config.ts
+++ b/src/services/ens/config.ts
@@ -1,0 +1,6 @@
+/**
+ * @see https://docs.ens.domains/ens-deployments
+ */
+export const CUSTOM_REGISTRIES: Record<string, string> = {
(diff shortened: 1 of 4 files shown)


usame-algan (approved):
Works fine when using in our input fields 👍 

Are we also able to support it via our `wallet` object e.g. for displaying the name in the header for a connected wallet?

katspaugh:
No, that would require hacking the provider which I don't think is worth it. Also this PR only address name resolution, not a reverse lookup.

francovenica:
Works fine. 

ENS sepolia

PR #2898, opened 2023-11-28
Fix: close tx flow on "View transaction"

katspaugh:
## What it solves

Bug: if you're on an individual transaction page and execute it, the View transaction button on the Success screen doesn't do anything (it navigates to the same page and doesn't close the modal).

Solution: it will now close the modal.


usame-algan (approved):

francovenica:
LGTM

The modal closes just fine
Tested with app execution as well ( I assumed is the same scenario, but just to be sure)

PR #2902, opened 2023-11-28
fix: allow cancellation after expiration

iamacook:
## What it solves

Resolves ability to [cancel recovery attempts after expiration](https://www.notion.so/safe-global/Cancel-recovery-tx-after-Recovery-delay-f8af1f46751f41a2bb735aec783a388c)

## How this PR fixes it

The disable logic has been altered to only take the expiration into account when not connected as an owner, as well as adjusting the information about cancellation throughout the app.

## How to test it

Observe cancellation is _always_ possible as an owner, wheras if not an owner only after expiration.

## Checklist

### DIFF
--- a/src/components/recovery/CancelRecoveryButton/index.tsx
+++ b/src/components/recovery/CancelRecoveryButton/index.tsx
@@ -13,7 +13,6 @@ import useSafeInfo from '@/hooks/useSafeInfo'
 import useOnboard from '@/hooks/wallets/useOnboard'
 import { logError, Errors } from '@/services/exceptions'
 import { RecoveryContext } from '../RecoveryContext'
-import { useIsGuardian } from '@/hooks/useIsGuardian'
 import { useRecoveryTxState } from '@/hooks/useRecoveryTxState'
 import type { RecoveryQueueItem } from '@/services/recovery/recovery-state'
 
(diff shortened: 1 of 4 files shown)


src/components/settings/Recovery/index.tsx
```diff
@@ -32,7 +32,7 @@ const headCells = [
     label: (
       <>
         Recovery delay{' '}
-        <Tooltip title="You can cancel any recovery attempt when it is not needed or wanted within the delay period.">
+        <Tooltip title="You can cancel any recovery attempt when it is not needed or wanted.">
```
iamacook, line 35: We should extract these to a constant as they are the same throughout the app but I'd suggest so in a separate PR to avoid inflating the scope of this.

src/components/recovery/CancelRecoveryButton/index.tsx
```diff
@@ -55,7 +55,7 @@ export function CancelRecoveryButton({
   return (
     <CheckWallet allowNonOwner>
       {(isOk) => {
-        const isDisabled = !isOk || (isGuardian && !isExpired)
+        const isDisabled = isOwner ? !isOk : !isOk && !isExpired
```
usame-algan, line 58: This makes it possible to cancel a recovery tx as a non-owner before it has expired. I assume that tx would fail so I suggest to still disable it.

iamacook: I've tweaked the logic in 6c7fca83eb530143c7bee0be26e0ced1ff360022 to only disable if not expired as [non-owners are able to skip expired transactions](https://github.com/gnosis/zodiac-modifier-delay/blob/main/contracts/Delay.sol#L203)

usame-algan (approved):

PR #2904, opened 2023-11-28
feat: recovery transaction processing state

iamacook:
## What it solves

Resolves [missing "Processing..." state for recovery transactions](https://www.notion.so/safe-global/Do-we-have-a-processing-relaying-state-instead-of-awaiting-confirmation-for-the-Account-recovery-5a07f9ed7cde429ba976fe7ad16e5f59)

## How this PR fixes it

A new recovery-focused event bus has been added that matches that of the one for "standard" transactions and messages. Using that, a new pending map is updated according to the dispatched events and parsed for the current status of a recovery attempt.

## How to test it

1. Queue a recovery attempt.
2. Execute/cancel the attempt and observe the new processing status.

Note: data loading should also be tested - recovery state updates when a proposal is executed, cancelled and/or settings are changed.

## Screenshots

image

## Checklist

### DIFF
--- a/src/components/recovery/ExecuteRecoveryButton/index.tsx
+++ b/src/components/recovery/ExecuteRecoveryButton/index.tsx
@@ -5,13 +5,12 @@ import type { SyntheticEvent, ReactElement } from 'react'
 import RocketIcon from '@/public/images/transactions/rocket.svg'
 import IconButton from '@mui/material/IconButton'
 import CheckWallet from '@/components/common/CheckWallet'
-import { dispatchRecoveryExecution } from '@/services/tx/tx-sender'
+import { dispatchRecoveryExecution } from '@/services/recovery/recovery-sender'
 import useOnboard from '@/hooks/wallets/useOnboard'
 import useSafeInfo from '@/hooks/useSafeInfo'
 import { useRecoveryTxState } from '@/hooks/useRecoveryTxState'
 import { Errors, trackError } from '@/services/exceptions'
 import { asError } from '@/services/exceptions/utils'
-import { RecoveryContext } from '../RecoveryContext'
 import { RecoveryListItemContext } from '../RecoveryListItem/RecoveryListItemContext'
 import type { RecoveryQueueItem } from '@/services/recovery/recovery-state'
 
@@ -24,10 +23,8 @@ export function ExecuteRecoveryButton({
 }): ReactElement {
   const { setSubmitError } = useContext(RecoveryListItemContext)
   const { isExecutable, isPending } = useRecoveryTxState(recovery)
-  const { setPending } = useContext(RecoveryContext)
   const onboard = useOnboard()
   const { safe } = useSafeInfo()
-  const { refetch } = useContext(RecoveryContext)
 
   const onClick = async (e: SyntheticEvent) => {
     e.stopPropagation()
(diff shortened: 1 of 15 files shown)


src/components/recovery/ExecuteRecoveryButton/index.tsx
```diff
@@ -49,13 +52,18 @@ export function ExecuteRecoveryButton({
 
       trackError(Errors._812, e)
       setSubmitError(err)
+    } finally {
+      setPending((prev) => {
+        const { [recovery.transactionHash]: _, ...rest } = prev
+        return rest
+      })
```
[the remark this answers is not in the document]
iamacook: I've taken the opportunity to refactor the transactional logic of recovery into an event bus that matches that of our "standard" transactions and messages. This helped fix this issue, as well as cleanup a lot of logic.

src/components/recovery/RecoveryStatus/index.tsx
```diff
@@ -1,12 +1,34 @@
...
+  [PendingStatus.SUBMITTING]: 'Submitting',
+  [PendingStatus.PROCESSING]: 'Processing',
+}
 
 export const RecoveryStatus = ({ recovery }: { recovery: RecoveryQueueItem }): ReactElement => {
   const { isExecutable, isExpired } = useRecoveryTxState(recovery)
+  const { pending } = useContext(RecoveryContext)
+
+  const status = pending?.[recovery.transactionHash] ? (
+    STATUS_LABELS[pending[recovery.transactionHash]]
```
usame-algan, line 21: As discussed I think we need to use `recovery.args.txHash` here instead.

src/components/recovery/RecoveryStatus/index.tsx
```diff
@@ -1,12 +1,34 @@
...
 import type { ReactElement } from 'react'
 
 import ClockIcon from '@/public/images/common/clock.svg'
 import { useRecoveryTxState } from '@/hooks/useRecoveryTxState'
 import type { RecoveryQueueItem } from '@/services/recovery/recovery-state'
+import { PendingStatus } from '@/store/pendingTxsSlice'
+import { RecoveryContext } from '../RecoveryContext'
+
+const STATUS_LABELS: Partial<Record<PendingStatus, string>> = {
+  [PendingStatus.SUBMITTING]: 'Submitting',
```
usame-algan, line 12: Should we use the `RecoveryEvent` instead? It doesn't have the Submitting event though (not sure if there is an equivalent)

src/components/recovery/RecoveryContext/useRecoveryState.ts
```diff
@@ -35,34 +33,10 @@ export function useRecoveryState(delayModifiers?: Array<Delay>): {
...
-      safeAddress,
-      safe.chainId,
-      safe.version,
-    ],
-    false,
-  )
+  // Reload recovery data when a Guardian transaction occurs
+  useEffect(() => {
+    return recoverySubscribe(RecoveryEvent.PROCESSED, refetch)
+  }, [refetch])
```
usame-algan, line 39: With this in place, do we still need the tx history listener?

src/hooks/useRecoveryTxState.ts
```diff
@@ -24,8 +30,9 @@ export function useRecoveryTxState({ validFrom, expiresAt, transactionHash, args
   const isNext =
     !delayModifier || (sameAddress(delayModifier.address, address) && args.queueNonce.eq(delayModifier.txNonce))
   const isExecutable = isNext && isValid && !isExpired
+  const isPending = !!pending?.[transactionHash]
```
usame-algan, line 33: I think we also have to change it here to `args.txHash`

src/components/recovery/RecoveryStatus/index.tsx
iamacook: The has was changed in c8a056f7ef0229bcfd4dc12036622ca21ef5ebbf and the spinner was added in b2fca640c6d6d7aea467fda126a3e0ffb97b26af.

src/components/recovery/RecoveryContext/useRecoveryState.ts
iamacook: Yes, the history listener is required for cancellations as we do not know what kind of transaction is executed.

src/components/recovery/RecoveryStatus/index.tsx
iamacook: As discussed: this will be left as is as it is not event but pending status related.

usame-algan (approved):

PR #2905, opened 2023-11-28
Fix: don't refresh ExecuteForm on validation polling

### LINKED ISSUE #2896: Execute form is refreshing every 15 seconds
On the Execution step of the tx flow, the Execute button along with the block above it visually refreshes and jumps every 15 seconds. The Execute button becomes temporarily disabled.

We shouldn't make the UI jump like that, it should cache the previous gas estimation during the polling.

https://github.com/safe-global/safe-wallet-web/assets/381895/24aabb75-75da-42df-b7dd-621ccc268625

katspaugh:
## What it solves

Resolves #2896

## How this PR fixes it

I've removed the flag that disabled the Execute button whilst polling the validity of the tx execution. The validation error itself doesn't block the Execute button, so why should its loading state do that?

### DIFF
--- a/src/components/tx-flow/flows/ExecuteBatch/ReviewBatch.tsx
+++ b/src/components/tx-flow/flows/ExecuteBatch/ReviewBatch.tsx
@@ -41,7 +41,7 @@ export const ReviewBatch = ({ params }: { params: ExecuteBatchFlowProps }) => {
   const { safe } = useSafeInfo()
   const [relays] = useRelaysBySafe()
   const { setTxFlow } = useContext(TxModalContext)
-  const [gasPrice, , gasPriceLoading] = useGasPrice()
+  const [gasPrice] = useGasPrice()
 
   const maxFeePerGas = gasPrice?.maxFeePerGas
   const maxPriorityFeePerGas = gasPrice?.maxPriorityFeePerGas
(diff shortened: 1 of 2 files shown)


usame-algan (commented):
Good point 👍 

I think the hook is reloading because one of the tags inside `safe` is changing e.g. when the queue is empty the `queueTag` changes constantly (There is an [open issue](https://github.com/safe-global/safe-wallet-web/issues/2490) for this). We could look into reducing the dependency of `safe` in that hook as only the address, chainId and version are required.

usame-algan (approved):

katspaugh:
The safe info is actually memoized, as it's selected from the store with an isEqual comparator.

In this case it's the `safeTx` object that is being re-created every time, I think.

usame-algan:
> The safe info is actually memoized, as it's selected from the store with an isEqual comparator.
> 
> 
> 
> In this case it's the `safeTx` object that is being re-created every time, I think.

It is memoized but if one of the tags changes and we depend on the whole `safe` object it will cause an update.

francovenica:
The fix is fine for the "execute" button in the Send tx form. But the issue is also present in the "Submit" button in the Bulk tx execution:
disabled submit

katspaugh:
Fixed that button as well. 👍

PR #2910, opened 2023-11-28
fix: prevent recovery with same setup

iamacook:
## What it solves

Prevents [recovery attempts with the same setup](https://www.notion.so/safe-global/Error-when-creating-editing-recovery-setup-without-changes-f82ad44386e547c99d2a9444854e4344)

## How this PR fixes it

This prevents recovery attempts from being created with the same owner/threshold setup. All owner/threshold fields are dependent on each other and validate together if the setup is the same as the current one.

## How to test it

Create a recovery attempt, proposing the same owners and threshold. Observe the validation error preventing futher progression.

## Screenshots

image

## Checklist


src/services/recovery/transaction.ts
```diff
@@ -93,10 +92,6 @@ export function getRecoveryProposalTransactions({
     txData.push(safeInterface.encodeFunctionData('changeThreshold', [newThreshold]))
   }
 
-  if (newThreshold > _owners.length) {
-    throw new Error('New threshold is higher than desired owners')
-  }
```
iamacook, line 98: Simulation should cover this.

src/services/recovery/transaction.ts
```diff
@@ -105,45 +100,6 @@ export function getRecoveryProposalTransactions({
...
-  const multiSendData = encodeMultiSendData(transactions)
-
-  return {
-    to: multiSendDeployment.networkAddresses[safe.chainId] ?? multiSendDeployment.defaultAddress,
-    value: '0',
-    operation: OperationType.Call,
-    data: multiSendInterface.encodeFunctionData('multiSend', [multiSendData]),
-  }
-}
-
```
iamacook, line 146: This was no longer being used.

usame-algan (commented):
I think we need to call `formMethods.trigger(RecoverAccountFlowFields.owners)` whenever an owner is removed to rerun the validation for the rest of the fields.

usame-algan:
Can we also add validation logic to editing recovery setup in case the user doesn't change anything?

iamacook:
> I think we need to call `formMethods.trigger(RecoverAccountFlowFields.owners)` whenever an owner is removed to rerun the validation for the rest of the fields.

> Can we also add validation logic to editing recovery setup in case the user doesn't change anything?

I've amended/implemented the above in 6a13d288e7696dd65f83f763a0c47d8109922048.

src/components/tx-flow/flows/UpsertRecovery/UpsertRecoveryFlowSettings.tsx
```diff
@@ -27,16 +28,37 @@ import { sameAddress } from '@/utils/addresses'
 import useSafeInfo from '@/hooks/useSafeInfo'
 import InfoIcon from '@/public/images/notifications/info.svg'
 import type { UpsertRecoveryFlowProps } from '.'
+import type { RecoveryState } from '@/services/recovery/recovery-state'
 
 import commonCss from '@/components/tx-flow/common/styles.module.css'
 import css from './styles.module.css'
 
+export function _isSameSetup({
```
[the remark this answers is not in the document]
iamacook: This will work on the initial load but if the user then navigates to the review screen and back, the dirty flag will reference the "new" values.

usame-algan (approved):

PR #2911, opened 2023-11-29
Tests: e2e - select connected wallet (patch)

francovenica:
## What it solves

There is an issue with the Create-tx test. Now that in Sepolia the Relayer is enabled you need to select the connected wallet option in the test. 
This is just a quick patch while I finish with the full refactor of the e2e-test, so the test doesn't fail in other PR's

## Checklist

### DIFF
--- a/cypress/e2e/smoke/create_tx.cy.js
+++ b/cypress/e2e/smoke/create_tx.cy.js
@@ -29,6 +29,7 @@ describe('[SMOKE] Create transactions tests', () => {
     createtx.verifyNativeTokenTransfer()
     createtx.changeNonce(currentNonce)
     createtx.verifyConfirmTransactionData()
+    cy.contains('Connected wallet').click()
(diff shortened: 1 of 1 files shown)


katspaugh (approved):

katspaugh (changes requested):
You should follow POM like the other calls in that test.

mike10ca (changes requested):
@francovenica 

1. Please add data-testid to 'Connected wallet' and use it in the test
2. Please create a method in support pages file to adhere to POM

francovenica:
@katspaugh @mike10ca Sorry, I messed up in the stand up. This patch was a quick fix so the PR's would stop failing because of the relayer in Sepolia.
Now that the [main ticket](https://github.com/safe-global/safe-wallet-web/pull/2948) I was working on is done this ticket is not useful anymore. 

I'll close it

PR #2913, opened 2023-11-29
fix: hide recovery widget if feature enabled

iamacook:
## What it solves

Resolves [recovery feature widget not hiding after being enabled](https://www.notion.so/safe-global/Remove-the-Edit-recovery-banner-from-the-Dashboard-after-the-recovery-is-set-up-cc6336c7391f4faf8cf587f5680ca70f)

## How this PR fixes it

The dashboard widget featuring recovery is now not only hidden if the chain doesn't support it but also if the feature is enabled on the current Safe.

## How to test it

Switch between Safes that have recovery enabled/disabled and observe the dashboard widget hidden accordingly.

## Checklist


usame-algan (approved):
Works nicely.

PR #2914, opened 2023-11-29
Analytics: track tx type on sign/execute

katspaugh:
## What it solves

We'll be now tracking the type of transaction for all tx events (Create/Confirm/Execute).
I've centralized the tracking of these events in the SignOrExecute component.

## How to test it

Create/sign/execute different types of transactions from this list:

```
  // Settings
  owner_add = 'owner_add',
  owner_remove = 'owner_remove',
  owner_swap = 'owner_swap',
  owner_threshold_change = 'owner_threshold_change',

  // Module txs
  guard_remove = 'guard_remove',
  module_remove = 'module_remove',

  // Transfers
  transfer_token = 'transfer_token',
  transfer_nft = 'transfer_nft',

  // Other
  batch = 'batch',
  rejection = 'rejection',
  typed_message = 'typed_message',
  safeapps = 'safeapps',
  walletconnect = 'walletconnect',
  custom = 'custom',
```

It should emit GTM events with the corresponding type in the `eventLabel`.

## Extra changes: merged Safe Open events
@alicecha requested that `Open Safe` and `Sidebar open Safe` events be merged.
Both will now emit `Open Safe` with a label depending on where it happened – `after_create`, `after_add` and `sidebar`.


src/services/analytics/tx-tracking.ts
```diff
@@ -0,0 +1,68 @@
...
+  } catch {
+    return TX_TYPES.custom
+  }
+  const { txInfo } = details
+
+  if (isTransferTxInfo(txInfo)) {
+    if (isERC721Transfer(txInfo.transferInfo)) {
+      return TX_TYPES.transfer_nft
+    }
+    return TX_TYPES.transfer_token
```
schmanu, line 27: Would it be interesting for us to differentiate between ERC20 and native transfers?

src/services/analytics/tx-tracking.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+
+  if (isCustomTxInfo(txInfo)) {
+    if (isCancellationTxInfo(txInfo)) {
+      return TX_TYPES.rejection
+    }
+
+    if (details.safeAppInfo) {
+      return isWalletConnectSafeApp(details.safeAppInfo.url) ? TX_TYPES.walletconnect : TX_TYPES.safeapps
```
schmanu, line 59: I feel like we are mixing source of transaction (SafeApp, WalletConnect) and transaction type here.
Should we add the source as label?

src/services/analytics/tx-tracking.ts
katspaugh: @alicecha wdyt?

src/services/analytics/tx-tracking.ts
katspaugh: I’ll double check.

katspaugh: Actually, it's only custom if it's a multisend or some other custom contract interaction. For single ERC20 transfers, it's categorized as a transfer. So you're right it's mixing apples and oranges a bit, but it will happen only within the custom category. I think it's an OK solution w/o introducing a new custom dimension.

src/services/analytics/tx-tracking.ts
```diff
@@ -0,0 +1,68 @@
...
+  isSettingsChangeTxInfo,
+  isTransferTxInfo,
+  isCustomTxInfo,
+  isCancellationTxInfo,
+} from '@/utils/transaction-guards'
+
+export const getTransactionTrackingType = async (txId: string, chainId: string): Promise<string> => {
+  let details: TransactionDetails
+  try {
+    details = await getTxDetails(txId, chainId)
```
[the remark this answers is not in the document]
usame-algan: The props are also named incorrectly in the function definition. I suggest to update it there as well

src/components/tx-flow/flows/SignMessageOnChain/ReviewSignMessageOnChain.tsx
```diff
@@ -101,11 +99,6 @@ const ReviewSignMessageOnChain = ({ message, method, requestId }: SignMessageOnC
   const handleSubmit = async () => {
     if (!safeTx || !onboard) return
 
-    // Track the creation of a typed message
-    if (isTypedMessage && safeTx.signatures.size === 0) {
-      trackEvent({ ...TX_EVENTS.CREATE, label: TX_TYPES.typed_message })
```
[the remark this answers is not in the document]
katspaugh: No, these ones will be tracked as custom. I think it's an edge case.

usame-algan (approved):

francovenica:
Question regarding typed messages.

It seems the event triggers if the message is created (basically it shows in the "Message" tab for the first time), but it doesn't trigger if the message is already in the message tab and you sign it. 
Is this an expected behavior? that the track is only done when the message hits the "message" tab but not when the message is already there and is signed afterwards?

I tried first connecting to OpenSea, signing the message with a safe 1/1 (so only 1 message needed). After it was done signing the event triggered just fine
image

But then I tried in a safe where I already have a bunch of messages ready to be signed, The event didn't track.
Note, these messages had 0 signatures, because they were created with the dummy app that makes them and the owner was probably kicked from the safe at some point
https://track_tx--walletweb.review-wallet-web.5afe.dev/transactions/messages?safe=gor:0x0EDc6697427E9e9d62A1eE1A2951AfA7bE9e75B7
image

francovenica:
Besides what I commented before the ticket looks good

I was able to check all the new calls has shown in the description

I'll leave a before/after snapshots just for the sake of documentation. These are the same type of tx execution in Dev and this PR:
Before:
image

After:
image

katspaugh:
Fixed ✅ 
Screenshot 2023-12-12 at 13 04 05

katspaugh:
I'll go ahead and merge it, and we can test the message signatures on the RC.

francovenica:
I checked the message fix, it works fine. 
Also checked the Open Safe events after creation and after adding, they work fine as well

LGTM

PR #2915, opened 2023-11-29
feat: add recovery transaction validation

iamacook:
## What it solves

Resolves [recovery attempt validation](https://www.notion.so/safe-global/Help-is-required-with-the-next-safe-gor-0x6Eb821dA05DB250F78945F9a2a25A6CBc436010f-887942b5f46240aeb6fe1e9810496bfb)

## How this PR fixes it

The description of a proposal now checks the owner length of the proposed setup and if it is empty then an error is displayed. On top of this, static calls of all recovery interactions have been added accordingly and are displayed before execution of proposal, execution and cancellation.

## How to test it

- Propose a recovery attempt and then modify the owner setup before execution and observe the new error.
- The validation errors are otherwise hard to test without removing safety checks. The request could potentially be blocked.

## Screenshots

image

Please note, the following errors are "forced" by removing safety checks:

image

image

## Checklist


src/components/recovery/RecoveryDescription/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useIsGuardian } from '@/hooks/useIsGuardian'
+import useSafeInfo from '@/hooks/useSafeInfo'
+import { logError, Errors } from '@/services/exceptions'
+import { getRecoveredSafeInfo } from '@/services/recovery/transaction-list'
+import type { RecoveryQueueItem } from '@/services/recovery/recovery-state'
+
+export function RecoveryDescription({ item }: { item: RecoveryQueueItem }): ReactElement {
+  const { args, isMalicious } = item
+  const { safe } = useSafeInfo()
+  const isGuardian = useIsGuardian()
```
schmanu, line 17: I thought Guardians were renamed to Recoverers across the app and code. Was that reverted?

src/components/recovery/RecoveryDescription/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const newSetup = useMemo(() => {
+    try {
+      return getRecoveredSafeInfo(safe, {
```
schmanu, line 21: This function does not consider some cases
- Swapping an owner, that does not exist anymore
- removing an owner, that does not exist anymore
- adding an owner, that's already added

But: Those calls would all revert anyways.
In general I feel like this is very detailed. Isn't just checking for transaction validity through `isValidExecution` enough?

src/components/recovery/RecoveryDescription/index.tsx
iamacook: It was. This was a merge issue. I've fixed it in 0851e3ee0ea6d138580a1fd0437fe7c7b59a6e64.

src/components/recovery/RecoveryDescription/index.tsx
iamacook: I've added validation for these cases via a static call in 72bcd714ed82c8c2eacf8347d0108135d5a1c9fc.

src/hooks/useIsValidRecoveryExecution.ts
```diff
@@ -0,0 +1,192 @@
...
+  const { safe } = useSafeInfo()
+  const web3ReadOnly = useWeb3ReadOnly()
+  const { isExecutable } = useRecoveryTxState(recovery)
+
+  return useAsync(() => {
+    if (!isExecutable || !wallet?.address || !web3ReadOnly) {
+      return
+    }
+
+    return (async () => {
```
usame-algan, line 132: Why do we need this self executing block? `useAsync` should already catch possible errors and return them.

src/components/recovery/RecoveryDescription/index.tsx
iamacook: As per discussion: this is an edge case that we won't fix as the current implementation is sufficient for a prediciton combined with the validation checks for execution/cancellation.

schmanu (approved):

usame-algan (approved):

PR #2916, opened 2023-11-29
fix: adjust description of expired transactions

iamacook:
## What it solves

Resolves [incorrect description of failed attempts](https://www.notion.so/safe-global/Expired-recovery-process-blocks-start-recovery-from-the-dashboard-c9a79955b4eb4592a70d367445ca958b)

## How this PR fixes it

The description of a failed transaction has been added under the recovery signers.

## How to test it

Queue a recovery transaction with a delay and expiration enabled and observe the following lifecycle description under the signers.

1. "The recovery transaction can be executed after the delay period has passed:"
2. "The recovery transaction can be executed until {{date}}." (without expiration: "The recovery transaction can be executed now.")
3. "The recovery transaction has expired and needs to be cancelled before a new one can be created."

## Checklist


schmanu (approved):
👍

PR #2922, opened 2023-11-29
fix: upsertion submission disabled on wrong chain

iamacook:
## What it solves

Resolves [submit button disabled when trying to upsert recovery and connected to the wrong chain](https://www.notion.so/safe-global/Submit-disabled-when-on-wrong-chain-2f58852d7aad48c59314ec416d58c9ff)

## How this PR fixes it

When connected to the wrong chain, the `web3` provider is not instantiated. As such, the encoding of the recovery-specific transactions is now done with the read-only provider instead.

## How to test it

Attempt to upsert recovery whilst connected to the wrong chain, observing that the submit button is enabled and when clicked, switches first to the correct chain.

Note: all other recovery transactions should work as expected.

## Checklist


src/components/tx-flow/flows/UpsertRecovery/UpsertRecoveryFlowReview.tsx
```diff
@@ -35,22 +35,24 @@ export function UpsertRecoveryFlowReview({
...
-
-      promise.then(setSafeTx).catch(setSafeTxError)
     })
-  }, [guardian, moduleAddress, params, safe.chainId, safeAddress, setSafeTx, setSafeTxError, web3])
+      .then((transactions) => {
+        const promise = transactions.length > 1 ? createMultiSendCallOnlyTx(transactions) : createTx(transactions[0])
+
+        promise.then(setSafeTx).catch(setSafeTxError)
+      })
+      .catch(setSafeTxError)
```
katspaugh, line 54:
```suggestion
      .then((transactions) => {
        return transactions.length > 1 ? createMultiSendCallOnlyTx(transactions) : createTx(transactions[0])
      })
      .then(setSafeTx)
      .catch(setSafeTxError)
```

iamacook: Implemented in e6c3ea18cc749f858c4d410d0481849a02b8dc50.

usame-algan (approved):

liliya-soroka:
Verified, can be merged

PR #2929, opened 2023-11-29
feat: recovery dispatch notifications

iamacook:
## What it solves

This adds recovery notifications and in doing so [hides the proposal widget when a proposal is in progress](https://www.notion.so/safe-global/Show-Account-recovery-in-progress-immediately-57206a130c004fa58f5425c91e6605a0).

## How this PR fixes it

This builds upon the recovery event bus, adding notifications for Recoverer recovery transactions as we do for "standard" transactions and messages:

- Recovery proposal
- Recovery execution
- Skipping of expired recovery attempts

Each of the above have notifications as such throughout their lifecycle:

1. "Confirm the execution in your wallet..."
2. "Validating..."
3. "Successfully validated. Querying..."

If the transaction reverts or fails, the following is displayed accordingly:

1. "Reverted. Please check your gas settings."
2. "Failed."

With the implementation of the above, it is now possible to know when a proposal is "in progress", e.g. being submitted to or validated on-chain. As such, the dashboard widget for Recoverers that allows for starting recovery is hidden when a proposal is "in progress" (see recording below).

## How to test it

1. Propose a recovery and observe the new notifications, but also the "Start recovery" header hidden during the proposal lifecycle.
2. Execute a recovery and observe the new notifications.
3. Skip expired recovery attempts and observe the new notifications.

## Screenshots

image

recovery notifications

## Checklist


src/components/dashboard/RecoveryHeader/index.test.tsx
```diff
@@ -46,22 +41,80 @@ describe('RecoveryHeader', () => {
...
+        recoveryTxHash: faker.string.hexadecimal(),
+        eventType: faker.helpers.enumValue(RecoveryEventType),
+      })
+
+      await waitFor(() => {
+        expect(result.current).toBe(true)
+      })
+    })
+  })
+  ;[RecoveryEvent.REVERTED, RecoveryEvent.PROCESSED, RecoveryEvent.FAILED].forEach((event) => {
```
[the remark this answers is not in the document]
schmanu: I figured it out. It is because the recovery did not hit the queue yet.
Could you add that to the description?

src/components/dashboard/RecoveryHeader/index.test.tsx
```diff
@@ -46,22 +41,80 @@ describe('RecoveryHeader', () => {
...
+
+      const { result } = renderHook(() => _useIsProposalInProgress())
+
+      expect(result.current).toBe(false)
+
+      recoveryDispatch(event, {
+        moduleAddress: faker.finance.ethereumAddress(),
+        txHash: faker.string.hexadecimal(),
+        recoveryTxHash: faker.string.hexadecimal(),
+        eventType: faker.helpers.enumValue(RecoveryEventType),
```
schmanu, line 83: Shouldn't the eventType always be RecoveryEventType.Proposal?

src/components/dashboard/RecoveryHeader/index.test.tsx
```diff
@@ -46,22 +41,80 @@ describe('RecoveryHeader', () => {
...
+      })
+    })
+  })
+  ;[RecoveryEvent.REVERTED, RecoveryEvent.PROCESSED, RecoveryEvent.FAILED].forEach((event) => {
+    it('should return false if there is not a proposal in progress', async () => {
+      const payload = {
+        moduleAddress: faker.finance.ethereumAddress(),
+        txHash: faker.string.hexadecimal(),
+        recoveryTxHash: faker.string.hexadecimal(),
+        eventType: faker.helpers.enumValue(RecoveryEventType),
```
schmanu, line 97: Same here. Shouldn't it always be RecoveryEventType.Proposal?

src/components/dashboard/RecoveryHeader/index.tsx
```diff
@@ -46,11 +48,33 @@ export function _RecoveryHeader({
...
 }
 
+export function _useIsProposalInProgress(): boolean {
+  const [isProposalSubmitting, setIsProposalSubmitting] = useState(false)
+  const queue = useRecoveryQueue()
+
+  useEffect(() => {
+    const unsubFns = Object.values(RecoveryEvent).map((event) =>
+      recoverySubscribe(event, (detail) => {
+        const isValidating = event === RecoveryEvent.EXECUTING || event === RecoveryEvent.PROCESSING
```
schmanu, line 58: Don't we have to ignore all events with `detail.eventType !== RecoveryEventType.PROPOSAL`?

src/hooks/useRecoveryTxNotification.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (!chain?.blockExplorerUriTemplate) {
+      return
+    }
+
+    const unsubFns = Object.entries(RecoveryTxNotifications).map(([event, notification]) =>
+      recoverySubscribe(event as RecoveryEvent, async (detail) => {
+        const isSuccess = event === RecoveryEvent.PROCESSED
+        const isError = 'error' in detail
+
+        const txHash = 'txHash' in detail ? detail.txHash : undefined
```
schmanu, line 48: Not sure where to put this comment.
But there is a edge case of having more than one expired txs.
Canceling one will actually cancel all of them. 
Not highest prio as it's an edge case but I think currently it would only mark the first expired recovery proposal as "processing" or being cancelled.

src/components/dashboard/RecoveryHeader/index.test.tsx
iamacook: Adjusted in 03398c6976df9d148a727b5e15e8ba09feb7309f.

src/components/dashboard/RecoveryHeader/index.test.tsx
iamacook: Adjusted in 03398c6976df9d148a727b5e15e8ba09feb7309f.

src/components/dashboard/RecoveryHeader/index.tsx
iamacook: Adjusted in 03398c6976df9d148a727b5e15e8ba09feb7309f.

src/hooks/useRecoveryTxNotification.ts
iamacook: It's not possible to propose more than one recovery transaction anymore so I would not concern ourselves about this. Users need to cancel any expired recovery attempts before they can make a new one.

PR #2932, opened 2023-11-30
changing order of evaluations

francovenica:
## What it solves

Resolves #
In the Nonce input, if you set a nonce that is 100 numbers higher than the recomended nonce it should display a specific message. That message wasn't showing up because of the order of the "if". The 2nd "if" was always overwriting the first one so the message that should show for 100+ values would never show up

## How this PR fixes it
Change the order of "if"

## How to test it
Go to a send funds tx
Change the nonce to a value slightly higher than the recomended nonce and check the tooltip message
Change the nonce to a value 100+ higher than the recomended nonce and check the tooltip message

## Screenshots
image
image

## Checklist


usame-algan (approved):
Nice fix 👍 Makes sense to go from small to large number validation in this case.

francovenica:
Thanks. I checked the tooltip messages myself.

PR #2948, opened 2023-12-02
Tests: Create-tx - Separating into indiviual tests

francovenica:
## What it solves
Create-tx had an issue with the relayer added to Sepolia making the test fail every time
Also it had test too broad they were splitted as much as possible

## How this PR fixes it
Adds a "BeforeEach" so every test starts clean

Split the first test into individual ones
Adds some other validations in step one to: 
address field, 
nonce field, 
amount field 

Step 2 was split between 
checking the adv parameters and 
checking relayer counter and addtobatch button for non-execution tx

Step 3 has to be modified to create a sign all in one

### Things I wasn't able to make it work:
- Check prefix validations in step 1: Currently the input only validates the prefix if you copy/paste the URL. That in cypress looks to be pretty difficult. There is the option to install a pluggin but I don't think is worth just for this test.
Also I think is the input the one that should validate the prefix no matter how you enter the address (copy/paste, writing or QR code)

- Reaching step 2 right away:
Talked with Usame, it seems that the inputs in step one are stored in a react component that you cannot mess with from the UI or with cypress like you would do with the localStorage. 
So it seems that going through step one a few times with a happy path is the most reliable solution

## Checklist


cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -5,39 +5,80 @@ import * as createtx from '../../e2e/pages/create_tx.pages'
...
+    createtx.verifyRandomStringAddress('Lorem Ipsum')
+    createtx.verifyWrongChecksum(constants.WRONGLY_CHECKSUMMED_ADDRESS)
+  })
+
+  it('[SMOKE] Verify address input validates a valid ENS name', () => {
+    createtx.typeRecipientAddress(constants.ENS_TEST_SEPOLIA)
+    createtx.verifyENSResolves()
+  })
+
+  it('[SMOKE] Verify amount input validates negative numbers', () => {
```
mike10ca, line 34: The test has other validations except negative number. Please rename this test to reflect specific goal of the test.

cypress/e2e/smoke/create_tx.cy.js
```diff
@@ -5,39 +5,80 @@ import * as createtx from '../../e2e/pages/create_tx.pages'
...
     createtx.clickOnSendTokensBtn()
-    createtx.typeRecipientAddress(constants.EOA)
+  })
+
+  it('[SMOKE] Verify address input error values', () => {
+    createtx.verifyRandomStringAddress('Lorem Ipsum')
+    createtx.verifyWrongChecksum(constants.WRONGLY_CHECKSUMMED_ADDRESS)
+  })
+
+  it('[SMOKE] Verify address input validates a valid ENS name', () => {
```
mike10ca, line 29: The test does not seem to test what it is expected to. Please review it and finalise accordingly.

mike10ca (changes requested):

cypress/e2e/smoke/create_tx.cy.js
francovenica: Renamed

mike10ca (approved):

PR #2949, opened 2023-12-03
fix: Update card mobile view, set recovery method correctly

usame-algan:
## What it solves

Resolves [Mobile issues](https://www.notion.so/safe-global/0fa7da35cc494c9ab0ec643634cacc24?v=6746502232bc4464b7ce8ced9fa3873d&p=521136a0e375498db1e8fa1d08a4daab&pm=s) and [Tracking issue](https://www.notion.so/safe-global/0fa7da35cc494c9ab0ec643634cacc24?v=6746502232bc4464b7ce8ced9fa3873d&p=521136a0e375498db1e8fa1d08a4daab&pm=s)

## How this PR fixes it

- Sets the RHF value when selecting a card
- Updates css to align icons on mobile

## Screenshots
Screenshot 2023-12-03 at 21 08 52

## Checklist

### DIFF
--- a/src/components/settings/Recovery/ChooseRecoveryMethodModal.tsx
+++ b/src/components/settings/Recovery/ChooseRecoveryMethodModal.tsx
@@ -3,7 +3,7 @@ import { RECOVERY_FEEDBACK_FORM, HelpCenterArticle } from '@/config/constants'
 import { trackEvent } from '@/services/analytics'
 import { RECOVERY_EVENTS } from '@/services/analytics/events/recovery'
 import { type ChangeEvent, type ReactElement, useContext } from 'react'
-import { useForm } from 'react-hook-form'
+import { Controller, useForm } from 'react-hook-form'
 import Link from 'next/link'
 import {
   Box,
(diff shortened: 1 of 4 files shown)


src/components/settings/Recovery/ChooseRecoveryMethodModal.tsx
```diff
@@ -57,11 +61,12 @@ export function ChooseRecoveryMethodModal({ open, onClose }: { open: boolean; on
     },
     mode: 'onChange',
   })
-  const { register, watch } = methods
+  const { register, watch, setValue } = methods
 
   const currentType = watch(FieldNames.recoveryMethod)
 
   const trackOptionChoice = (e: ChangeEvent<HTMLInputElement>) => {
+    isRecoveryMethod(e.target.value) && setValue(FieldNames.recoveryMethod, e.target.value)
```
iamacook, line 69: `setValue` should not need to be called as this is inherently called within a registration. Is it otherwise not working?

usame-algan: Works with using a `Controller` instead 👍

PR #2950, opened 2023-12-04
Fix: hide outgoing txs w/o nonce

katspaugh:
## What it solves

Fixes cases like https://github.com/5afe/safe-support/issues/342 where scam transaction events get indexed as outgoing transfers. They don't have an `executionInfo` and therefore no nonce.

This is a temporary frontend workaround until it's fixed on the backend.


schmanu (commented):
I think we should write a test for the `useTxHistory` slice.

We need to be aware that this will also hide a lot of legit transfer events. For instance when swapping on many DEXes such as CowSwap. It will only show the incoming token and not the outgoing token anymore as that swap has no nonce.

schmanu (approved):

PR #2952, opened 2023-12-04
refactor: improve `isRecoveryQueueItem` strictness

iamacook:
## What it solves

Improves the strictness of the `isRecoveryQueueItem` type guard.

## How this PR fixes it

The `eventSignature` uniqueness of the `RecoveryQueueItem` is checked that it is `"TransactionAdded(uint256,bytes32,address,uint256,bytes,uint8)"`.

## How to test it

Queue a recovery proposal and observe that the dashboard and transaction list queues work as expected.

## Checklist

### DIFF
--- a/src/utils/transaction-guards.ts
+++ b/src/utils/transaction-guards.ts
@@ -112,7 +112,8 @@ export const isTransactionListItem = (value: TransactionListItem): value is Tran
 }
 
 export function isRecoveryQueueItem(value: TransactionListItem | RecoveryQueueItem): value is RecoveryQueueItem {
-  return 'args' in value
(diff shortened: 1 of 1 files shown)


schmanu (approved):

francovenica:
Looks good to me

Recover tx can be created just fine
The dashboard works accordinly:
Shows the button for the recoverer or
Tells when the tx is already created and can be executed or 
Shows when the tx has expired and must be cancelled

The Queue tab works fine, shows the recoverer tx, its cancellation if it has it, new tx that might be queued
The Tx in the dashboard widget also work fine, they are listed there properly

PR #2954, opened 2023-12-04
fix: wallet connect namespace without requiredNamespace

### LINKED ISSUE #2946: WalletConnect - Missing or invalid emit() event
### Problem

Our dApp only sends `optionalNamespaces` and no (required) `namespaces` as no specific chain is required, only one of 23 we support.

When connecting to our dApp with a Safe Wallet via Wallet Connect, we get the following error:
image

When building the approved namespace for the session, Safe only takes the events from `requiredNamespaces`  on the proposal (in `WalletConnectWallet::getNamespaces`), which is empty in the above case, as such establishing a session which supports no events.

In theory this is within spec so far, however Safe then immediately tries to emit a `chainChanged` event (in `WalletConnectWallet::approveSession`) right after approving the session leading to the above error.

#### Sources

getNamespaces
https://github.com/safe-global/safe-wallet-web/blame/v1.23.0/src/services/walletconnect/WalletConnectWallet.ts#L84

approveSession
https://github.com/safe-global/safe-wallet-web/blame/v1.23.0/src/services/walletconnect/WalletConnectWallet.ts#L115

The error originates in @wallet-connect/sign-client when isValidEmit fails
https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/sign-client/src/controllers/engine.ts#L1429

### Suggested fix

Similar to how all `SAFE_COMPATIBLE_METHODS` are added to the approved namespace `methods`, also add `chainChanged` and `accountsChanged` (and any other that Safe emits) to the list of `events`.

schmanu:
## What it solves

Resolves #2946 

## How this PR fixes it
Adds required events to namespace on approveSession.

## How to test it
- Find a dapp without requiredNamesapce
- Observe no error about a unavailble emit event.

## Checklist

### DIFF
--- a/src/services/walletconnect/WalletConnectWallet.ts
+++ b/src/services/walletconnect/WalletConnectWallet.ts
@@ -8,7 +8,7 @@ import { type JsonRpcResponse } from '@walletconnect/jsonrpc-utils'
 import uniq from 'lodash/uniq'
 
 import { IS_PRODUCTION, LS_NAMESPACE, WC_PROJECT_ID } from '@/config/constants'
-import { EIP155, SAFE_COMPATIBLE_METHODS, SAFE_WALLET_METADATA } from './constants'
+import { EIP155, SAFE_COMPATIBLE_EVENTS, SAFE_COMPATIBLE_METHODS, SAFE_WALLET_METADATA } from './constants'
 import { invariant } from '@/utils/helpers'
 import { getEip155ChainId, stripEip155Prefix } from './utils'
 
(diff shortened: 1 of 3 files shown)


katspaugh (approved):
Thanks!

francovenica:
LGTM

I tried with the app of the people who reported the issue: https://app.beefy.com/

I checked first the issue in prod (I get the same error message that the user reported) 
Then I tried in the PR, I don't get the issue anymore and I can connect just fine
I didn't execute a tx, but I was able to start one, with that I can say that the connection with the app was succesful
image

PR #2957, opened 2023-12-04
fix: do not remove invalid address prefix

### LINKED ISSUE #2939: Prefix validation not working on first try
## Bug description
When copy/pasting an address with a prefix in an address input, the input is supposed to check that the prefix matches to the current network.
There are 2 inputs where this doesn't happen at first try: The owner adding during safe creation and the Send NFT form

## Environment
 - Browser: Any
 - Wallet: Any
 - Chain: Any
 - Env: Current dev v1.23

## Steps to reproduce
 1. Go to Safe creation in ETH
 2. During the step of adding more owners, add a new row and copy this address sep:0x66bE167c36B3b75D1130BBbDec69f9f04E7DA4fC
 3. Select the address and paste again. Do NOT clear the input, paste on top of the existing address.

Repeat these steps in the Send NFT form

## Expected result
Copying the address with a wrong prefix should fail

## Obtained result
In step 2, no errors are shown, like if the prefix was trimmed out
In step 3 now the input "realizes" that the prefix is wrong, which it was expected to happen in step 2

## Screenshots
prefix

schmanu:
## What it solves

Resolves #2939 

## How this PR fixes it
Does not remove invalid address prefixes when setting the value.

## How to test it
- Input an address with an invalid chain prefix
- Blur the address input
- Observe that the error persists

## Checklist

### DIFF
--- a/src/components/common/AddressInput/index.test.tsx
+++ b/src/components/common/AddressInput/index.test.tsx
@@ -35,7 +35,7 @@ const TestForm = ({ address, validate }: { address: string; validate?: AddressIn
     defaultValues: {
       [name]: address,
     },
-    mode: 'onChange',
+    mode: 'all',
   })
 
   return (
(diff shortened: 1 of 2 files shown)


katspaugh (approved):

francovenica:
Looks good to me

Checked the inputs I reported (Safe creation owner addition and Send NFT address input)

PR #2959, opened 2023-12-04
Fix: Hide nonce input if its a spending limit transaction

### LINKED ISSUE #2880: Spending limits - The nonce field is visible when non-owners use the allowance
## Bug description
Non-owners can start a send funds tx if they have an allowance set in the spending limit feature. The issue is that the Nonce field is present in the form.
There are no issues with creating/executing the tx itself. Is just the nonce field that should not be visible

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Any
 - Env: Prod v1.23

## Steps to reproduce
 1. Go to a safe you own
 2. Create a spending limit of any token for a non-owner wallet you control
 3. Connect with that wallet and go into that safe
 4. Start a Send funds tx
 5. The spending limit option should be selected by default in the form
 6. Check the nonce input

## Expected result
The nonce input should not be visible

## Obtained result
The nonce input is showing. I'd assume that the form hides it when you select "use spending limit" in the form, but since is the default option you never click it, therefore the form never checks

## Screenshots
image

usame-algan:
## What it solves

Resolves #2880 

## How this PR fixes it

- Hides the nonce input field if a non-owner tries to create a spending limit transaction
- Pre-selects the first available token when a non-owner tries to create a spending limit transaction
- Displays the nonce input again if an owner switches from a spending limit token to a non-spending limit token

## How to test it

1. Open a safe with a spending limit as a non-owner
2. Create a transaction and observe the nonce input is hidden
3. Observe a token is pre-selected in the form

---

1. Open a safe as an owner that also has a spending limit
2. Create a transaction
3. Observe the nonce input being visible
5. Select the spending limit token and select spending limit in the radio group
6. Observe the nonce input is hidden
7. Switch to a different token
8. Observe the nonce input is visible

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
+++ b/src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
@@ -1,22 +1,15 @@
+import { useTokenAmount, useVisibleTokens } from '@/components/tx-flow/flows/TokenTransfer/utils'
 import madProps from '@/utils/mad-props'
-import { type ReactElement, useState, useContext, useEffect } from 'react'
-import { type SafeBalanceResponse, type TokenInfo } from '@safe-global/safe-gateway-typescript-sdk'
-import { useVisibleBalances } from '@/hooks/useVisibleBalances'
-import useAddressBook from '@/hooks/useAddressBook'
+import { type ReactElement, useContext, useEffect } from 'react'
+import { type TokenInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import { useSafeTokenAddress } from '@/components/common/SafeTokenWidget'
 import useIsSafeTokenPaused from '@/hooks/useIsSafeTokenPaused'
 import useIsOnlySpendingLimitBeneficiary from '@/hooks/useIsOnlySpendingLimitBeneficiary'
-import { useAppSelector } from '@/store'
-import { selectSpendingLimits } from '@/store/spendingLimitsSlice'
-import useWallet from '@/hooks/wallets/useWallet'
 import { FormProvider, useForm } from 'react-hook-form'
-import useSpendingLimit from '@/hooks/useSpendingLimit'
-import { BigNumber } from '@ethersproject/bignumber'
 import { sameAddress } from '@/utils/addresses'
 import { Box, Button, CardActions, Divider, FormControl, Grid, SvgIcon, Typography } from '@mui/material'
 import TokenIcon from '@/components/common/TokenIcon'
 import AddressBookInput from '@/components/common/AddressBookInput'
-import AddressInputReadOnly from '@/components/common/AddressInputReadOnly'
 import InfoIcon from '@/public/images/notifications/info.svg'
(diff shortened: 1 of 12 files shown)


usame-algan:
I tried to keep the changes as minimal as possible. The third case is another bug I found that wasn't reported yet. I want to write tests for this component but would also refactor it in the process because its overly complex at the moment. Wdyt about doing that in this PR?

src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
```diff
@@ -67,6 +67,16 @@ const CreateTokenTransfer = ({
...
 
+  const balancesItems = useMemo(() => {
+    return isOnlySpendingLimitBeneficiary
+      ? balances.items.filter(({ tokenInfo }) => {
+          return spendingLimits?.some(({ beneficiary, token }) => {
+            return sameAddress(beneficiary, wallet?.address || '') && sameAddress(tokenInfo.address, token.address)
+          })
+        })
+      : balances.items
+  }, [balances.items, isOnlySpendingLimitBeneficiary, spendingLimits, wallet?.address])
```
usame-algan, line 78: This is the same as it was before just pushed further up so we can use it to set the default form value.

src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
```diff
@@ -133,6 +134,12 @@ const CreateTokenTransfer = ({
   const isDisabled = isSafeTokenSelected && isSafeTokenPaused
   const isAddressValid = !!recipient && !errors[TokenTransferFields.recipient]
 
+  useEffect(() => {
+    if (!spendingLimit) {
+      setNonceNeeded(true)
+    }
```
katspaugh, line 140: Will it flip back if they switch to a spending limit?

schmanu: Yes the `SpendingLimitRow` `onChange` handler also sets setNonceNeeded.

schmanu (approved):
At some point we should write test for the CreateTokenTransfer component.
We should create a tech debt issue.

francovenica:
Issue:

Gif:
This is an owner I'm connected with
He has spending limit allowance for tokenA, but not for tokenB
He selects tokenA, selects spending limit and the nonce goes away (expected)
When switching to tokenB the nonce shows up again, since he can only do a regular tx, (expected as well)
When switching back to tokenA the spending limit option is marked by default and the nonce input keeps showing (not expected)

Is expected that the nonce field is hidden once again if the "Spending limit" option is selected
spending limit

usame-algan:
> Is expected that the nonce field is hidden once again if the "Spending limit" option is selected

Thanks for checking! I updated it so the nonce input is hidden again.

francovenica:
Issue fixed. 

The rest also looks good to me.

schmanu (approved):

PR #2963, opened 2023-12-05
fix: fetch multisend contract with readonly provider

### LINKED ISSUE #2937: Wallet network switch not working with bulk execute
## Bug description

Bulk execute while the connected wallet is on the wrong chain doesn't prompt the user to switch to the correct chain. Instead it just closes the bulk execute flow.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to queue
 2. Connect to the safe with the wallet on a different chain
 3. Bulk execute multiple transactions
 4. Observe nothing happening

## Expected result

The user should be prompted to switch network

## Obtained result

Nothing happens

schmanu:
## What it solves

Resolves #2937 

## How this PR fixes it
Uses the readonly provider to fetch the contract instance.

## How to test it
Batch execute transactions with a wallet connected to a wrong chain

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/ExecuteBatch/ReviewBatch.tsx
+++ b/src/components/tx-flow/flows/ExecuteBatch/ReviewBatch.tsx
@@ -16,7 +16,6 @@ import { TxSimulation } from '@/components/tx/security/tenderly'
 import { WrongChainWarning } from '@/components/tx/WrongChainWarning'
 import { useRelaysBySafe } from '@/hooks/useRemainingRelays'
 import useOnboard from '@/hooks/wallets/useOnboard'
-import { useWeb3 } from '@/hooks/wallets/web3'
 import { logError, Errors } from '@/services/exceptions'
 import { dispatchBatchExecution, dispatchBatchExecutionRelay } from '@/services/tx/tx-sender'
 import { hasRemainingRelays } from '@/utils/relaying'
@@ -32,6 +31,7 @@ import { TxModalContext } from '@/components/tx-flow'
 import useGasPrice from '@/hooks/useGasPrice'
 import { hasFeature } from '@/utils/chains'
 import type { PayableOverrides } from 'ethers'
+import { useWeb3ReadOnly } from '@/hooks/wallets/web3'
 
 export const ReviewBatch = ({ params }: { params: ExecuteBatchFlowProps }) => {
   const [isSubmittable, setIsSubmittable] = useState<boolean>(true)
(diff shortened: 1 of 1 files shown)


src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -28,7 +28,7 @@ export function assertValidSafeVersion<T extends SafeInfo['version']>(safeVersio
   return invariant(isValidSafeVersion(safeVersion), `${safeVersion} is not a valid Safe Account version`)
 }
 
-export const createEthersAdapter = (provider: Web3Provider) => {
+export const createEthersAdapter = (provider: Web3Provider | JsonRpcProvider) => {
   const signer = provider.getSigner(0)
```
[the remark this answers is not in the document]
schmanu: Good point. I will use that instead.

src/services/contracts/safeContracts.ts
```diff
@@ -80,12 +80,8 @@ export const _getMinimumMultiSendCallOnlyVersion = (safeVersion: SafeInfo['versi
...
 }
 
-export const getMultiSendCallOnlyContract = (
-  chainId: string,
-  safeVersion: SafeInfo['version'],
-  provider: Web3Provider,
-) => {
-  const ethAdapter = createEthersAdapter(provider)
+export const getMultiSendCallOnlyContract = (chainId: string, safeVersion: SafeInfo['version']) => {
+  const ethAdapter = createReadOnlyEthersAdapter()
```
usame-algan, line 84: We already have a function called `getReadOnlyMultiSendCallOnlyContract` which is identical to this one that we can use instead

schmanu: Reverted everything and changed it to use that function instead

src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -28,7 +28,7 @@ export function assertValidSafeVersion<T extends SafeInfo['version']>(safeVersio
   return invariant(isValidSafeVersion(safeVersion), `${safeVersion} is not a valid Safe Account version`)
 }
 
-export const createEthersAdapter = (provider: Web3Provider) => {
+export const createEthersAdapter = (provider: Web3Provider | JsonRpcProvider) => {
```
usame-algan, line 31: Can this also be reverted?

usame-algan (approved):
Looks good 👍 
I tested executing Batch transactions on the correct chain and on the wrong chain
In case of a wrong chain it prompts me to switch chains otherwise it lets me execute normally.

PR #2965, opened 2023-12-05
refactor: Use custom recovery events instead of create tx events

usame-algan:
## What it solves

Part of Recovery EPIC

## How this PR fixes it

- Emits custom recovery events instead of using `TX_EVENTS.CREATE` since with #2914 they would be tracked twice in some cases

## How to test it

Create, edit, remove a recovery setup and observe the new events
Create and cancel a recovery attempt and observe the new events

## Checklist


katspaugh (approved):
Thanks, looks good!

usame-algan:
@alicecha fyi we had to change the events to custom ones. I will update the table in Notion accordingly.

PR #2900, opened 2023-11-28
Fix: remove human description A/B test

katspaugh:
## What it solves

Resolves #2879

## How this PR fixes it

Remove the A/B test until the CSS bugs are fixed (overlapping text in the queue, broken text on mobile).

PR #2967, opened 2023-12-05
Fix: rm human-readable leftovers

katspaugh:
## What it solves

The human-readable A/B test was removed in #2900 but some leftovers remained. This PR cleans that up which fixes the missing tx summary.


liliya-soroka:
incoming txs are not fixed

usame-algan (approved):

usame-algan:
> incoming txs are not fixed Current result: image
> 
> From prod image

This is intended as the state on prod shows a human readable description which we are reverting.

PR #2971, opened 2023-12-06
Tests: History tests - part 1

### LINKED ISSUE #2970: Write Cypress tests for Tx History
We need to provide more coverage for tx history. 
This ticket will cover some of tx types: account creation, sending, receiving, setting / removing spending limits, contract interaction.

mike10ca:
## What it solves

Resolves #2970 

## How this PR fixes it

- We need to provide more coverage for tx history.
- This ticket will cover some of tx types: account creation, sending, receiving, setting / removing spending limits, contract interaction.


cypress/e2e/regression/tx_history.cy.js
```diff
@@ -1,21 +1,190 @@
...
+  it('Verify summary for token receipt', () => {
+    createTx.verifySummaryByIndex(
+      txItemIndex,
+      [typeReceive.title, typeReceive.summaryTxInfo, typeReceive.summaryTime, typeGeneral.statusOk],
+      typeReceive.altImage,
+      typeReceive.altToken,
+    )
+  })
+
+  it('Verify exapnded details for token receipt', () => {
```
katspaugh, line 70:
```suggestion
  it('Verify expanded details for token receipt', () => {
```

PR #2972, opened 2023-12-06
Feat: queue counter in the sidebar menu

katspaugh:
## What it solves

A badge with a queued transaction count in the sidebar menu.

## How to test it
Open a Safe with queued txs, look at the sidebar

## Screenshots
Screenshot 2023-12-06 at 14 31 42

## Checklist

### DIFF
--- a/src/components/sidebar/SidebarNavigation/index.tsx
+++ b/src/components/sidebar/SidebarNavigation/index.tsx
@@ -6,10 +6,11 @@ import { ImplementationVersionState } from '@safe-global/safe-gateway-typescript
 import {
   SidebarList,
   SidebarListItemButton,
+  SidebarListItemCounter,
   SidebarListItemIcon,
   SidebarListItemText,
 } from '@/components/sidebar/SidebarList'
-import { navItems } from './config'
+import { type NavItem, navItems } from './config'
 import useSafeInfo from '@/hooks/useSafeInfo'
 import { AppRoutes } from '@/config/routes'
 import useTxQueue from '@/hooks/useTxQueue'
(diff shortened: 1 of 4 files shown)


src/components/sidebar/SidebarNavigation/index.tsx
```diff
@@ -23,25 +24,29 @@ const Navigation = (): ReactElement => {
   const router = useRouter()
   const { safe } = useSafeInfo()
   const currentSubdirectory = getSubdirectory(router.pathname)
-  const hasQueuedTxs = Boolean(useTxQueue().page?.results.length)
-  const hasRecoveryTxs = Boolean(useRecoveryQueue().length)
+  const queueSize = (useTxQueue().page?.results.length || 0) + useRecoveryQueue().length
```
usame-algan, line 27: This seems to end up as double the number of actual items in the queue

katspaugh: Good catch, fixed!

usame-algan:
Two things I noticed:

- We include transactions with the same nonce in the count, not an issue just want to make sure it is intended
- The count doesn't reflect potential queued transactions on the second page, should we say `19+` in that case?

katspaugh:
I'll add a + if there's a next page.

katspaugh:
That hook became quite big, I'll add tests for it tomorrow.

usame-algan (approved):
Looks like with 2-3 characters the badge has not as much space next to the list item but looks good otherwise 👍 

Screenshot 2023-12-07 at 12 00 47

katspaugh:
> Looks like with 2-3 characters the badge has not as much space next to the list item but looks good otherwise 👍

It's a badge, so it's OK to overlap a tiny bit with the content it refers to.

francovenica:
I noticed that it counts rows and not different nonces, but that's how the current counter in safe list on the sidebar works, so I'm ok with it.

It looks fine, but counter at max value behaves weirdly. It goes 19 > 20 > 19+ . 
Why not 19 > 19+ or 
19 > 20 > 20+?

Beyond this, the ticket is fine and can be passed to Done

katspaugh:
Because backend returns 19 txs, not 20 for some reason. I counted them in the JSON response.

PR #2976, opened 2023-12-07
feat: add export button to owner list

iamacook:
## What it solves

Resolves inability to export owner list

## How this PR fixes it

This adds an export button below the owner list that, when clicked, creates as CSV file listing the current owners of the Safe Account and their name prioritised by address book then "known address".

## How to test it

Open the owner list and click the new export button. Observe that the downloaded CSV lists all the owners.

## Screenshots

image

## Checklist

### DIFF
--- a/src/components/common/EnhancedTable/index.tsx
+++ b/src/components/common/EnhancedTable/index.tsx
@@ -35,7 +35,6 @@ type EnhancedHeadCell = {
   label: ReactNode
   width?: string
   sticky?: boolean
-  hideSort?: boolean
 }
 
 function descendingComparator(a: EnhancedRow, b: EnhancedRow, orderBy: string) {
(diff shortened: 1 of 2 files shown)


iamacook:
> Nice!
> 
> Design-wise, I would make it a button next to the Add owner button, possibliy with the same download icon.
> Screenshot 2023-12-07 at 08 21 40
> 
> Also, if an owner has a name, we might want to export that in the second column.

I've adjusted it in 3320008513c01b10448271ac4977090242b20d22 accordingly. I think it looked much better without the icon so it's as suggested.

iamacook:
> Looks good, but I'm not fully sold in the name of the file having the full address of the safe: 11155111-0x8f4A19C85b39032A37f7a6dCc65234f966F72551-owners Maybe the "short version": Like "0x8f4A_2551"?
> 
> Opinions? @katspaugh @iamacook
> 
> Note: if this was already discussed and decided that way then it's fine and we can pass the ticket

The file name has not been discussed. I don't think it's a good idea to _not_ have the full address though. There is the possibility that shortened addresses are the same.

PR #2978, opened 2023-12-07
Analytics: include query params in the pageview event

katspaugh:
## What it solves

We need to track the query params to pass UTM-params to GA.


usame-algan (approved):
Looks good 👍 I can see additional query params in the pageview event. After navigating within the app these parameters are removed though so they will only be tracked when the app initially opens.

francovenica:
The param is being tracked.

A before after snapshot for documentation purposes:

Before: 
image

After:
image

PR #2981, opened 2023-12-07
fix: Add default values for `value` and `data` in AppCommunicator

dasanra:
## What it solves

Resolves https://github.com/safe-global/safe-apps-sdk/issues/480

This will be a fast hotfix for the issue above. Final fix for the bug was detected in https://github.com/safe-global/safe-apps-sdk/pull/530

Because it may take time to reach all the dependencies where the package is used will publish this as an emergency fix while developers update to the recent version with the fix.

## How this PR fixes it

Set the normalized empty value when undefined is set for `value` or  `data` for a Safe App `sendTransaction` event.

## Checklist

### DIFF
--- a/src/components/safe-apps/AppFrame/useAppCommunicator.ts
+++ b/src/components/safe-apps/AppFrame/useAppCommunicator.ts
@@ -172,7 +172,7 @@ const useAppCommunicator = (
         return {
           to: getAddress(to),
           value: value ? BigNumber.from(value).toString() : '0',
-          data,
+          data: data || '0x',
         }
       })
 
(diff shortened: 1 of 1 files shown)


francovenica:
Thanks @dasanra helping me out to test this.

We checked with an wagmi example fix to check how it looks when it breaks and when it works fine. https://github.com/safe-global/safe-apps-sdk/tree/development

So first we run the application un current dev, tried to execute a example of a tx and it failed
Then in this PR we run the same scenario and the tx started normally. I didn't execute the tx, but the fact that it was started already proves that the issue was fixed.

LGTM

PR #2983, opened 2023-12-08
Fix: reconnect MetaMask only if unlocked

katspaugh:
## What it solves

After #2343, the reconnect behavior with MetaMask became extremely annoying. Every time you open the browser and go to the app, it pops up a locked MetaMask window even if you don't need the wallet at that moment.

This PR limits the auto-connect to only when MetaMask is already password-unlocked. The behavior for other wallets is unchanged.

## How to test it

Unlocked MetaMask:

1. Go to the app
2. Connect MetaMask (unlock it if password-locked)
3. Refresh the page
4. MetaMask should seamlessly reconnect w/o showing the onboard modal

Locked MetaMask:
1. Disconnect MetaMask if connected
2. Completely close the browser
3. Open the browser again and go to the app
4. Make sure MetaMask is password-locked
5. Observe that the app didn't auto-connect to MetaMask

Other wallets:
They should reconnect like before.

### DIFF
--- a/src/utils/wallets.ts
+++ b/src/utils/wallets.ts
@@ -39,3 +39,18 @@ export const isSmartContractWallet = memoize(
   },
   (chainId, address) => chainId + address,
 )
+
+/* Check if the wallet is unlocked. */
+export const isWalletUnlocked = async (walletName: string): Promise<boolean | undefined> => {
+  const METAMASK = 'MetaMask'
+
+  // Only MetaMask exposes a method to check if the wallet is unlocked
+  if (walletName === METAMASK) {
+    if (typeof window === 'undefined' || !window.ethereum?._metamask) return false
+    try {
+      return await window.ethereum?._metamask.isUnlocked()
+    } catch {
+      return false
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


src/utils/wallets.ts
```diff
@@ -39,3 +39,18 @@ export const isSmartContractWallet = memoize(
...
+
+  // Only MetaMask exposes a method to check if the wallet is unlocked
+  if (walletName === METAMASK) {
+    if (typeof window === 'undefined' || !window.ethereum?._metamask) return false
+    try {
+      return await window.ethereum?._metamask.isUnlocked()
+    } catch {
+      return false
+    }
+  }
```
schmanu, line 55: Nit: 
IMO this makes the condition a bit more readable.

```suggestion
  if (walletName === METAMASK) {
    const metaMask = window?.ethereum?._metamask
    if (typeof metaMask === 'undefined') return false
    try {
      return metaMask.isUnlocked()
    } catch {
      return false
    }
  }
```

schmanu (approved):
LGTM

src/utils/wallets.ts
katspaugh: This will throw when window isn't declared tho (on Node).

francovenica:
LGTM.

MM connects fine on refresh. It doesn't show the locked screen if entering the app after closing the browser and reopening.

PR #2985, opened 2023-12-08
Fix: speed up popups

katspaugh:
## What it solves

MUI has a default animation when a dropdown or a popup opens. It makes them feel slow.
This PR sets the animation duration to 0 for the dropdowns in the header. It makes the UI much snappier.


usame-algan (approved):
The initial connect wallet popup still animates a bit slower

katspaugh:
You mean with a delay? There's an initital delay for all of them for some reason. Like it loads something.

usame-algan:
I think the `ConnectionCenter` component could also have a `transitionDuration={0}` to keep it consistent.

PR #2986, opened 2023-12-08
feat: Show error on tx and safe creation if wallet balance is insufficient

usame-algan:
## What it solves

A follow up to #2678 to show an error to the user if their connected wallet balance is too low to execute a transaction or create a safe

## How this PR fixes it

- Shows an error in case relaying is not selected and wallet balance is too low

## ToDos

- [x] Implement a design
- [x] Decide if we want to show the wallet balance on the transaction screen -> yes
- [x] Decide if the Submit button should be disabled in that case -> not disabled
- [x] Write tests for `useWalletCanPay`
- [x] Extend `ExecuteForm` tests
- [x] Check why safe creation throws without the wallet popping up -> Fixed in #3013 

## How to test it

1. Try to execute a transaction with a wallet that doesn't have enough funds
2. Observe an error in the transaction review screen
3. Check that the error is only shown if relaying is not selected
4. Try to create a safe with a wallet that doesn't have enough funds
5. Observe an error in the review screen

## Screenshots
Screenshot 2023-12-13 at 11 06 44
Screenshot 2023-12-13 at 12 25 41

## Checklist

### DIFF
--- a/src/components/new-safe/create/steps/ReviewStep/index.tsx
+++ b/src/components/new-safe/create/steps/ReviewStep/index.tsx
@@ -1,3 +1,5 @@
+import ErrorMessage from '@/components/tx/ErrorMessage'
+import useWalletCanPay from '@/hooks/useWalletCanPay'
 import { useMemo, useState } from 'react'
 import { Button, Grid, Typography, Divider, Box, Alert } from '@mui/material'
 import { lightPalette } from '@safe-global/safe-react-components'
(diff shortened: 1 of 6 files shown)


src/components/tx/ExecutionMethodSelector/index.tsx
```diff
@@ -73,7 +72,7 @@ export const ExecutionMethodSelector = ({
         </FormControl>
       </div>
 
-      {shouldRelay && relays ? <SponsoredBy relays={relays} tooltip={tooltip} /> : wallet ? <BalanceInfo /> : null}
+      {shouldRelay && relays ? <SponsoredBy relays={relays} tooltip={tooltip} /> : null}
```
schmanu, line 75: I feel like we should still show the wallet's balance. Then it is easier to understand why the balance is too low or by how much the balance is too low.

usame-algan: I've rebased this branch to the relabeling branch and kept the wallet balance.

schmanu (approved):
LGTM! Great addition!

francovenica:
The message is fine, but I'd also disable the "Execute button" if that message pops up:

image

usame-algan:
> The message is fine, but I'd also disable the "Next" button if that message pops up:

The problem is that on some chains (like gno) our estimation is not always accurate so blocking the user would be too limiting in those cases thats why we decided not to disable the button.

francovenica:
Ok, make sense. 
Then is good to go. Passed

PR #2987, opened 2023-12-08
fix: Dont reload balance when popup is closed

usame-algan:
## What it solves

The balance doesn't reload anymore whenever the header popup is opened and closed

## How this PR fixes it

Pass the balance from the header to the child component instead of using the same hook twice

## How to test it

1. Connect a wallet
2. Open the account center through the header
3. Observe the balance being visible
4. Close the popup and open it again
5. Observe no loading state

## Checklist


katspaugh (approved):
👍

PR #2988, opened 2023-12-08
Fix: better labels for relaying

katspaugh:
## What it solves

We received user feedback that the relaying feature is unclear. 
This PR changes some of the texts in the execution method selector.

Screenshot 2023-12-08 at 16 33 04
Screenshot 2023-12-08 at 16 32 53


usame-algan (approved):
Looks good apart from a few unit tests failing 👍 
The `create_tx` smoke e2e test also needs to be adjusted for the new wording.

PR #2992, opened 2023-12-11
feat: Display network logos in network selector

usame-algan:
## What it solves

- Shows network logos for the network selector
- Splits up the network selector into testnets and non-testnets
- Shows network logos for the chain indicator except in the sidebar and safe list sidebar

## How to test it

1. Open Safe and observe network logos in the network selector
2. Open the sidebar and observe no network logo there
3. Open the safe list sidebar and observe no network logo there

## Screenshots

Screenshot 2023-12-11 at 12 10 29
Screenshot 2023-12-11 at 15 34 11
Screenshot 2023-12-11 at 15 34 23
Screenshot 2023-12-11 at 15 34 40
Screenshot 2023-12-11 at 15 34 58

## Checklist


public/images/networks/gno.png
katspaugh, line 1: Do we need both an svg and a png?

src/components/common/NetworkSelector/index.tsx
```diff
@@ -27,6 +27,30 @@ const MenuWithTooltip = forwardRef<HTMLUListElement>(function MenuWithTooltip(pr
...
+  aurora: '/images/networks/aurora.svg',
+  base: '/images/networks/base.svg',
+  ['base-gor']: '/images/networks/base.svg',
+  zkevm: '/images/networks/polygon.svg',
+  zksync: '/images/networks/zksync.svg',
+  celo: '/images/networks/celo.svg',
+  gor: '/images/networks/gor.svg',
+  arb1: '/images/networks/arb.svg',
+  avax: '/images/networks/avax.svg',
+  sep: '/images/networks/sep.png',
```
katspaugh, line 47: Could you use the `chains.ts` module for the keys?
Also it would make sense to keep this config in a global module because it's going to be used in all network indictors.

src/components/common/NetworkSelector/index.tsx
```diff
@@ -90,20 +114,43 @@ const NetworkSelector = (props: { onChainSelect?: () => void }): ReactElement =>
...
-        '& svg path': {
-          fill: ({ palette }) => palette.border.main,
-        },
       }}
     >
-      {configs.map((chain) => {
-        return (
-          <MenuItem key={chain.chainId} value={chain.chainId} disabled={isSocialLogin && !isSocialWalletEnabled(chain)}>
-            <Link href={getNetworkLink(chain.shortName)} onClick={props.onChainSelect} passHref>
-              <ChainIndicator chainId={chain.chainId} inline />
```
katspaugh, line 102: Wouldn't it make more sense to include a logo into the ChainIndicator component?

katspaugh (changes requested):

src/components/common/NetworkSelector/index.tsx
usame-algan: I started adding the changes to the `ChainIndicator` but it looks a bit out of place in the current safe list sidebar. Should we still push the changes since the safe list sidebar is getting reworked soon? Otherwise I could also only add the changes for inline chain indicators and leave the safe list out that way

katspaugh: It could be an optional prop.

public/images/networks/gno.png
usame-algan: Nice catch! I removed the `.svg` version since its just an image anyway

src/components/common/NetworkSelector/index.tsx
usame-algan: I've updated the component and display the logo by default now. I've disabled it for the sidebar and safe list sidebar.

For the safe apps preview and share page I show the logos but increased the gap between each item slightly.

src/components/common/NetworkSelector/index.tsx
usame-algan: I extracted them to the `@/config/chains` file since thats where we access the `chains.ts` module as well.

katspaugh:
@TanyaEfremova @liliiaorlenko should we add a rounded border around each chain or something like that on these screen?

src/components/common/ChainIndicator/index.tsx
```diff
@@ -1,3 +1,5 @@
+import { getChainLogo } from '@/config/chains'
+import Image from 'next/image'
```
katspaugh, line 2: I thought we weren't using this component. Does it add some size optimizations with SSG?

katspaugh (approved):
👍

src/components/common/ChainIndicator/index.tsx
katspaugh: Nice.

usame-algan: Looks like its causing our jest tests to fail though so I will change it to ``

francovenica:
LGTM

I think I looked everywhere where the logo might be (like the top right corner of a form like send funds) but I didn't find anything broken or a missing logo.
Checked the network selector and apps
Checked in mobile viewport as well
Checked that there are no logos in the sidebar

PR #2994, opened 2023-12-11
chore: Download latest bundle analysis artifact from base branch

usame-algan:
## What it solves

Follow up for #2803 

The download-artifact action we use downloads artifacts in chronological order so it always compares with the oldest dev bundle instead of the newest. There is an open issue on the board but no fix so far ([see discussion](https://github.com/dawidd6/action-download-artifact/issues/240)). We switch to a custom bash script that is also linked in the discussion to download and unzip the bundle from dev.

## Screenshots
Screenshot 2023-12-11 at 15 55 05

## Checklist

### DIFF
--- a/scripts/github/download_bundle_analyser_artifact.sh
+++ b/scripts/github/download_bundle_analyser_artifact.sh
@@ -1,0 +1,31 @@
+# We use this instead of action-download-artifact. See discussion on
+# https://github.com/dawidd6/action-download-artifact/issues/240
+set -xe
+ORG="safe-global"
+REPO="safe-wallet-web"
+WORKFLOW="nextjs-bundle-analysis.yml"
+ARTIFACT_NAME="bundle"
+DESTINATION=".next/analyze/base"
+BASE_BRANCH="dev"
+
+ARTIFACTS_URL=$(
+  gh api \
+    -H "Accept: application/vnd.github+json" \
+    -H "X-GitHub-Api-Version: 2022-11-28" \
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
If it works, it works, but not ideal that we have to do it "manually"...

scripts/github/download_bundle_analyser_artifact.sh
```diff
@@ -0,0 +1,31 @@
...
+    -H "Accept: application/vnd.github+json" \
+    -H "X-GitHub-Api-Version: 2022-11-28" \
+    "${ARTIFACTS_URL}" \
+    --jq '.artifacts[] | select(.name == "'"${ARTIFACT_NAME}"'") | .archive_download_url'
+)
+
+set +x
+curl -H "Accept: application/vnd.github+json" -H "Authorization: token $GH_TOKEN" -L -o "${DESTINATION}.zip" "$DOWNLOAD_URL"
+set -x
+unzip "${DESTINATION}.zip" -d "${DESTINATION}" && mkdir -p "${DESTINATION}/bundle" && mv "${DESTINATION}/__bundle_analysis.json" "${DESTINATION}/bundle/"
```
usame-algan, line 30: We need to unzip the contents of `${DESTINATION}.zip` into the directory `${DESTINATION}/bundle`. I tried using wildcards but without success so I hard-code the filename here.

usame-algan:
> If it works, it works, but not ideal that we have to do it "manually"...

Agreed, I've left a message in a related issue on https://github.com/hashicorp/nextjs-bundle-analysis/issues/66

PR #3006, opened 2023-12-13
Tests: Use latest stable Chrome version in tests

mike10ca:
## What it solves

Resolves # N/A

## How this PR fixes it

- Install latest stable Chrome version to be used in all jobs

### DIFF
--- a/.github/workflows/cypress/action.yml
+++ b/.github/workflows/cypress/action.yml
@@ -49,6 +49,7 @@ runs:
         browser: chrome
         record: true
         config: baseUrl=http://localhost:8080
+        wait-on: 'http://localhost:8080'
       env:
         CYPRESS_RECORD_KEY: ${{ inputs.record_key || fromJSON(inputs.secrets).CYPRESS_RECORD_KEY }}
         GITHUB_TOKEN: ${{ fromJSON(inputs.secrets).GITHUB_TOKEN }}
--- a/.github/workflows/cypress/action.yml
+++ b/.github/workflows/cypress/action.yml
@@ -26,18 +26,6 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - uses: browser-actions/setup-chrome@v1
-      with:
-        chrome-version: 119
-
-    - name: List Items in Chrome Folder
-      shell: bash
-      run: ls -al /opt/hostedtoolcache/chromium/119.0.6045.105/x64
-
-    - name: Start Chrome
-      shell: bash
-      run: /opt/hostedtoolcache/chromium/119.0.6045.105/x64/chrome --headless --disable-gpu &
-
     - uses: ./.github/workflows/yarn
 
     - name: Install Cypress
@@ -58,7 +46,7 @@ runs:
         spec: ${{ inputs.spec }}
         group: ${{ inputs.group }}
         parallel: true
-        browser: /opt/hostedtoolcache/chromium/119.0.6045.105/x64/chrome
+        browser: chrome
         record: true
         config: baseUrl=http://localhost:8080
       env:
--- a/.github/workflows/cypress/action.yml
+++ b/.github/workflows/cypress/action.yml
@@ -32,6 +32,7 @@ runs:
 
     - run: |
         ${{ steps.setup-chrome.outputs.chrome-path }} --version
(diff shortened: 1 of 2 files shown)


.github/workflows/cypress/action.yml
```diff
@@ -26,13 +26,14 @@ inputs:
...
-
     - uses: ./.github/workflows/yarn
 
-    - name: Install Cypress
+    - name: Install Specific Chrome Version
       shell: bash
-      run: ./node_modules/.bin/cypress install
+      run: |
+        sudo apt-get install -y wget
+        sudo wget -q https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
```
katspaugh, line 35: But stable current is also not a specific version

katspaugh:
Why is the browser version such an issue? The app should work in any modern browser.

.github/workflows/cypress/action.yml
mike10ca: I would stick to latest Chrome version so the tests run on latest version

mike10ca:
> Why is the browser version such an issue? The app should work in any modern browser.

Sometimes there are issues with different browser versions in runners, this causes Cypress failures when we run them in parallel. I hope downloading latest stable version of Chrome and installing it will solve this issue. I checked the logs it seems to be working as expected.

katspaugh:
I sincerely don't get the problem we're trying to solve and how this solution is different to the previous setup. There was an action that was supposed to download and install chrome, and it even supports a version parameter. Now we download the latest stable version, which changes according to Chrome's defintion of latest stable.

It feels like there's not enough "scientific" justification for working on something for half a day. Can you prove the Chrome version is the cause of flaky tests? The app must and does work in all modern version of Chrome, Firefox and other browsers. I _highly_ doubt that minor version variations happening _within_ minutes that these containers are running can cause any problems.

mike10ca:
> I sincerely don't get the problem we're trying to solve and how this solution is different to the previous setup. There was an action that was supposed to download and install chrome, and it even supports a version parameter. Now we download the latest stable version, which changes according to Chrome's defintion of latest stable.
> 
> It feels like there's not enough "scientific" justification for working on something for half a day. Can you prove the Chrome version is the cause of flaky tests? The app must and does work in all modern version of Chrome, Firefox and other browsers. I _highly_ doubt that minor version variations happening _within minutes_ during which these containers are running can cause any problems.

As I mentioned earlier, the problem is with Github actions changing Chrome version in their runners, but not all runners have latest Chrome version at the same time. We run tests in parallel. So some of parallel jobs get v119 and some v120. Cypress cloud only accepts one version at a time, meaning it is unable to accept two versions, 119 and 120 within test run. This causes failures in some jobs. There is nothing we can do about it except enforcing a specific version. We can enforce v 119 for example, but it will require maintenance and we will need to update it manually in workflow file so the app runs on latest / latest-1 versions. This is not ideal. So the best try for us would be to stick to latest stable version. I would assume the tests will work fine on this version because it assumes to be stable enough. The first option we had earlier did not work as expected unfortunately, and even when I tried enforcing it there were problems with browser connections. Using a Docker contain requires more adjustment work. Given the circumstances, I would stick to current fix unless we want to invest more time in it.

katspaugh:
Could show some logs/screenshots where it's evident jobs fail because of this?

.github/workflows/cypress/action.yml
```diff
@@ -26,13 +26,14 @@ inputs:
...
-
     - uses: ./.github/workflows/yarn
 
-    - name: Install Cypress
+    - name: Install Latest stable Chrome Version
       shell: bash
-      run: ./node_modules/.bin/cypress install
+      run: |
+        sudo apt-get install -y wget
+        sudo wget -q https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
```
katspaugh, line 35: `sudo` is not needed for wget

katspaugh (approved):

mike10ca:
I looked at the logs and can confirm the fix is doing what it is required to do by setting specific version over another one as shown below. 
Screenshot 2023-12-15 at 09 26 45
Screenshot 2023-12-15 at 09 25 22

PR #3008, opened 2023-12-13
Fix: safe apps search performance

katspaugh:
## What it solves

The search input in the Safe Apps list was painfully slow to type. Instead of debouncing using a hook, I've added a debounce directly to the onchange listener and it's much faster. Still far from ideal but much faster.

## How to test it

Type something in the Safe Apps search filter.


usame-algan (approved):
Nice, it definitely improved performance for me 👍 

I don't fully understand why `useDebounce` didn't work well but we could rework `useSafeAppsFilters` separately.

francovenica:
LGTM

The search works fine

PR #3010, opened 2023-12-14
Fix: don't auto-reconnect social login

katspaugh:
Untitled

As per @schmanu's comment:

> After rehydrating the Web3Auth session we automatically call connectWallet we have to avoid calling connectWallet on page load as the MPC connection is not ready yet then and the connectWallet call on rehydration get’s blocked somehow.

I've disable auto-reconnect for the social login.


schmanu (approved):
Tested it locally. Works like a charm 💯

katspaugh:
@schmanu merging to dev, since it cannot be tested on this branch.

PR #3013, opened 2023-12-14
fix: Estimate gasLimit for safe creation

### LINKED ISSUE #3012: Gas estimation error when creating a safe
## Bug description

When trying to create a safe with a wallet that doesn't have funds the UI immediately shows an error without first opening the wallet.

## Steps to reproduce
 1. Go to Safe
 2. Connect a wallet without funds
 3. Create a new safe
 4. An error immediately pops up

## Expected result

The wallet (e.g. Metamask) opens and shows the user that they don't have enough funds

## Obtained result

The Safe UI shows an error

## Screenshots
Screenshot 2023-12-14 at 11 09 24

usame-algan:
## What it solves

Resolves #3012 

## How this PR fixes it

Calls the `estimateSafeCreationGas` function during safe creation and passes the `gasLimit` with the rest of the parameters

## How to test it

1. Open Safe
2. Connect a wallet without funds
3. Create a new safe
4. Observe that the wallet (e.g. MM) pops up

## Screenshots

Screenshot 2023-12-14 at 11 11 55

## Checklist

### DIFF
--- a/src/components/new-safe/create/steps/StatusStep/useSafeCreation.ts
+++ b/src/components/new-safe/create/steps/StatusStep/useSafeCreation.ts
@@ -15,6 +15,7 @@ import {
   SAFE_CREATION_ERROR_KEY,
   showSafeCreationError,
   relaySafeCreation,
+  estimateSafeCreationGas,
 } from '@/components/new-safe/create/logic'
 import { useAppDispatch } from '@/store'
 import { closeByGroupKey } from '@/store/notificationsSlice'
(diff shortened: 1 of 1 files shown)


schmanu (approved):
Works great! Tested safe creation with too low funds, with sufficient funds and with relaying.

francovenica:
So the fix is fine, it looks the same as the screenshot in the description, but I think the ticket https://github.com/safe-global/safe-wallet-web/pull/2986 will make this fix redundant, since you will never be able to execute a safe creation tx w/o funds.

I don't mind if we merge both tho.

PR #3014, opened 2023-12-14
Tests: Fix test index

mike10ca:
## What it solves

Resolves #N/A

## How this PR fixes it

- Update test indexes


liliya-soroka (approved):
approved. Improvement card is added to the automation board to use tx's nonce if possible instead of the position in the table

PR #3018, opened 2023-12-15
Analytics: track successfull QR scans

katspaugh:
## What it solves

Adds a new analytics event emitted after a successfull QR code scan.

## How this PR fixes it
I've made a [GA exploration](https://analytics.google.com/analytics/web/#/analysis/p308247657/edit/QUA2dWwnRmyWKbKZDz1H8w
) for the "Scan QR" event. It shows an average of 100-150 events a day. I'd like to test the hypothesis that many of those clicks don't result in an actual scan.

Screenshot 2023-12-15 at 09 37 50

### DIFF
--- a/src/components/common/ScanQRModal/ScanQRButton.tsx
+++ b/src/components/common/ScanQRModal/ScanQRButton.tsx
@@ -1,8 +1,7 @@
 import React, { lazy, useState, Suspense, type ReactElement } from 'react'
 import QrCodeIcon from '@/public/images/common/qr.svg'
 import { IconButton, SvgIcon } from '@mui/material'
-import Track from '../Track'
-import { MODALS_EVENTS } from '@/services/analytics'
+import { MODALS_EVENTS, trackEvent } from '@/services/analytics'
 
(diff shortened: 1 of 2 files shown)


usame-algan (approved):
nit: The event names could be a bit more specific now since it says `Scan QR` when opening and `Scan QR finished` but the first one could be `Scan QR started`.

katspaugh:
That would break compatibility with historical data.

francovenica:
Looks fine.

I've checked the event triggers when reading a "resolvable" Qr code. Uploading a random png won't trigger it.

Note:
It doesn't take in account the validation of QR code, this means, if you use a QR code that resolves into "RandomString" or has an invalid prefix (like using your Receive QR code that the MM phone app provides) those are still counting as "QR Scan Finished" events

If there is any way to validate that what the QR code resolves is a valid address before triggering the event that would be nice, but I don't consider it a blocker for this PR

image

katspaugh:
I think it’s OK, the intention is to measure if people actually try to scan anything (or just click on that button by mistake).

PR #3011, opened 2023-12-14
fix: poll wallet balance

schmanu:
## What it solves
Polls wallet balance every 15s.

Resolves [Notion issue](https://www.notion.so/safe-global/04fafddb03ce4ec79611081c819f45a7?v=8b8e36baa6484c8cb1163334d656a89c&p=57d297e101de4401bc312cce59a36ec1&pm=s)

## How this PR fixes it
Polls the connected wallet's balance every 15s

## How to test it
- Connect a wallet
- Send native tokens out of that wallet or in to the wallet
- observe that it updates automatically after a while

## Checklist
* [x] I've tested the branch on mobile 📱
* [x] I've documented how it affects the analytics (if at all) 📊
* [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #3024, opened 2023-12-15
fix: Revert wallet balance polling

usame-algan:
## What it solves

Reverts #3011 as it was causing too many RPC requests (An additional `eth_chainId` every 15s)
Reverts #2983 as it was causing too many RPC requests (From 8 -> 14 `eth_chainId` requests on initial load)

## How to test it

- Check that the last connected wallet still automatically reconnects
- Check that the wallet balance is visible on initial load

1. Open Safe
2. Open the network tab
3. Observe that there are no additional `eth_chainId` requests over time


francovenica:
The wallet still autoconnects fine
The balance is not updated anymore every few secs, it only updates on the first load of the page

francovenica (approved):

katspaugh (approved):

PR #3027, opened 2023-12-18
Fix: use wallet balance from onboard

katspaugh:
## What it solves

For the header widget, use the balance string returned by web3-onboard as it fetches it from the wallet RPC anyway.
The returned balance always matches the wallet chain, not the Safe chain, so I made it clearer in the header widget:

Screenshot 2023-12-18 at 13 58 14

---

In the tx-flow, it will still get an updated wallet balance directly from the readonly RCP like before:

Screenshot 2023-12-18 at 13 57 53

### DIFF
--- a/src/components/common/ConnectWallet/AccountCenter.tsx
+++ b/src/components/common/ConnectWallet/AccountCenter.tsx
@@ -10,6 +10,7 @@ import WalletInfo from '@/components/common/WalletInfo'
 
 export const AccountCenter = ({ wallet }: { wallet: ConnectedWallet }) => {
   const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>(null)
+  const { balance } = wallet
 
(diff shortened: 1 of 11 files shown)


katspaugh:
A couple more screenshots:

Screenshot 2023-12-18 at 14 37 18
Screenshot 2023-12-18 at 14 37 49

francovenica:
A tx can be started even if the wallet is in a different network. This causes the wallet balance to not load if it is in the wrong network.
My suggestion is not showing the "Wallet balance" tag at all if wallet network is different from the current safe network. (Or add a "switch to [SAFE_NETWORK]" message where the skeleton loads)
image

katspaugh:
Good catch, reverted that part, so it will still use the readonly RPC.

francovenica:
Works now. 

For the record, in the widget at the top we show the balance of the wallet in the wallet's curren network and in forms we show the balance in the safe's network

PR #3032, opened 2023-12-19
Chore: update dependencies

katspaugh:
* Updated a few minor versions
* Removed a duplicate date-fns package
* Made an alias for ethereumjs which was imported by several packages under different versions


package.json
```diff
@@ -36,16 +36,15 @@
     "@web3-onboard/trezor/**/protobufjs": "^7.2.4"
   },
   "dependencies": {
-    "@date-io/date-fns": "^2.15.0",
```
katspaugh, line 39: Apparently this package was a duplicate.

PR #3040, opened 2023-12-20
Fix: [Safe Apps] align the share button

katspaugh:
## What it solves

The Share button now has a hover effect like the other two buttons and is perfectly aligned to the center.

Screenshot 2023-12-20 at 09 57 12


liliya-soroka:
Verified

PR #3041, opened 2023-12-20
Fix: don't cut address in tx flow

### LINKED ISSUE #2975: Send tokens flow: address can be clipped
When a recipient address doesn't fit in the modal, it gets clipped despite ample space on the left.
We should make the recipient field wider.

Screenshot 2023-12-06 at 17 33 37

katspaugh:
## What it solves

Resolves #2975

## How this PR fixes it

I've made the grid flexible, so it wraps on smaller screens.

Large screen:
Screenshot 2023-12-20 at 11 11 43

Medium-small:
Screenshot 2023-12-20 at 11 12 16


francovenica:
Looks good.

There is a nish case in the Bulk execution, where the "Interact with" label gets cropped, but when the address falls to the next row the label remains cropped:

PR #3044, opened 2023-12-20
Refactor: [Recovery] lazy-load based on feature toggle

### LINKED ISSUE #3035: Lazy-load Recovery
The Recovery feature isn't something that you use every day but it gets bundled with the main app chunk and comes with a 1-megabyte dependency.

It would be good to isolate and lazy-load Recovery components, hooks and services. Ideally also put them all in one folder. Right now they are all over the codebase.

## Requirements
* The Recovery feature should have a single import point (an index.ts that re-exports individual modules)
* If the feature is disabled via the `RECOVERY` feature toggle, no related modules should be imported in runtime
* All Recovery-related components, hooks and other modules should be in the same folder

## Current bundle
Screenshot 2023-12-19 at 14 13 29

cc @iamacook

katspaugh:
## What it solves

Resolves #3035

## How this PR fixes it

* I've changed the RecoveryContext to an external store to be able to lazy-load it – lazy-loading a React context provider re-renders the entire app
* It will be not loaded if the feature is disabled
* It's now code-split into its own chunk

Screenshot 2023-12-20 at 15 09 57


src/features/recovery/components/RecoveryContext/RecoveryContextHooks.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useRecoveryPendingTxs } from './useRecoveryPendingTxs'
+import { useRecoverySuccessEvents } from './useRecoverySuccessEvents'
+import store from '.'
+
+function RecoveryContextHooks(): null {
+  const [delayModifiers, delayModifiersError, delayModifiersLoading] = useRecoveryDelayModifiers()
+  const [recoveryState, recoveryStateError, recoveryStateLoading] = useRecoveryState(delayModifiers)
+  const pending = useRecoveryPendingTxs()
+
+  useRecoverySuccessEvents(pending, recoveryState)
```
katspaugh, line 13: This hook is apparently used for side effects and it shouldn't be in a store intializer...

katspaugh:
@francovenica FYI I've ran the [regression suite](https://github.com/safe-global/safe-wallet-web/actions/runs/7286560561) on this branch, it's all green ✅

iamacook (approved):

iamacook:
All looks good to me but as it touches every area of recovery, I think it would be wise to do a full regression test of the feature.

francovenica:
LGTM

Checked in sepolia that I can enable/disable the feature in the django admin

Did a smoke on the feature on a sepolia safe and it works fine.

liliya-soroka:
the full regression for the Recovery feature is required according to @iamacook .
I will move it back to QA for the deep testing

PR #3048, opened 2023-12-21
Refactor: [WalletConnect] lazy-load based on feature toggle

katspaugh:
## What it solves

WalletConnect code is now dynamically imported only if the feature is enabled.

Unfortunately, the large `@walletconnect/ethereum-provider` package still gets bundled with the main chunk because of the web3-onboard module that imports it statically. Fixed in their repo.

## How to test it

* Test that WalletConnect can still connect to dapps
* That transactions via WalletConnect work
* That clicking on the WalletConnect "app" on the dashboard opens the native WalletConnect
* That opening the old WalletConnect app redirects to the list and opens the native WalletConnect

(I already did all of that and it still works.)


katspaugh:
Some page chunks increased by ~90Kb, this is due to Viem from safe-apps-SDK. Fixed in #3049.

katspaugh:
Regression suite: https://github.com/safe-global/safe-wallet-web/actions/runs/7290307284

francovenica:
LGTM. 

The redirection from the dashboard WC widget opens the WC in the header. Same when using the item in the Safe Apps section

I tried connecting to open sea and star a NFT purchase. I had no issues signing the message after login in and the tx started normally when trying to purchase the NFT (I didn't complete the purchase tho, too expensive)

PR #3049, opened 2023-12-21
Refactor: update safe-apps-sdk to v9 with ESM support

katspaugh:
## What it solves

The new safe-apps-sdk exports ESM modules which allows tree-shaking. It reduces the 500-kb Viem dependency significantly.

## Todo

* [x] Remove the copied RPC_CALLS (they are exported only as a type from the SDK atm)
* [ ] Update the package version once it's fully released


katspaugh:
@francovenica I've run regression on this branch ✅

PR #3052, opened 2023-12-23
Refactor: remove Sentry performance tracing

katspaugh:
## What it solves

We've never been using [browser performance tracing](https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/browsertracing/) on Sentry, so I've removed it. And the `@sentry/tracing` package isn't needed either way, as the same thing is also available from `@sentry/react`.


src/components/safe-apps/AppFrame/useAppCommunicator.ts
```diff
@@ -103,13 +103,8 @@ const useAppCommunicator = (
             },
           )
         },
-        onError: (error, data) => {
-          logError(Errors._901, error.message, {
-            contexts: {
-              safeApp: app || {},
-              request: data,
-            },
```
katspaugh, line 111: This made absolutely no difference, as logError doesn't send anything to Sentry. Only trackError does.

PR #3060, opened 2024-01-02
fix: show fieldErrors in readonly input

### LINKED ISSUE #3058: Add owner - The owner field is not validating the safe itself as owner
## Bug description

During add owner or replace owner, the field where you enter the address to be the new owner, the safe itself needs to be validated as a non-valid address
Currently when you add the safe itself no error is shown. The "Next" button is disabled, so there is no risk of adding the safe as owner, but still the field should show the error right away

## Environment
 - Browser: Any
 - Wallet: Any
 - Chain: Any (tested in sepolia originally)
 - Env: Dev 1.25.x

## Steps to reproduce
 1. Go to a safe as owner
 2. Start an Add owner from
 3. Copy the Safe's address
 4. Paste it as a new owner

## Expected result
The input should right away show an error indicating the safe cannot be owner of itself

## Obtained result
The error doesn't show up right away. If you click in the input again, clearing it, the error sudendly shows.

Note:The "Next" button remains disabled so right now there is no risk of an user accidentaly adding the safe as owner

schmanu:
## What it solves

Resolves #3058 

## How this PR fixes it
- [x] Shows `fieldErrors` in readonly component
- [ ] Validates when selecting from the dropdown

## How to test it
1. Paste the Safe's address into the address input when adding an owner
2. Observe that the error is correctly shown

1. Select the Safe from the Address book input
2. Observe that the error is correctly shown

## Checklist

### DIFF
--- a/src/components/common/AddressBookInput/index.tsx
+++ b/src/components/common/AddressBookInput/index.tsx
@@ -1,6 +1,6 @@
 import AddressInputReadOnly from '@/components/common/AddressInputReadOnly'
 import { type ReactElement, useState, useMemo } from 'react'
-import { useFormContext, useWatch } from 'react-hook-form'
+import { get, useFormContext, useWatch } from 'react-hook-form'
 import { Box, SvgIcon, Typography } from '@mui/material'
 import Autocomplete, { createFilterOptions } from '@mui/material/Autocomplete'
 import useAddressBook from '@/hooks/useAddressBook'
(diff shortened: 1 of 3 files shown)


katspaugh (approved):
Thanks for adding all the tests! Refactoring the input to a controlled RHF input is a wise move 👍

francovenica:
LGTM.

I've tested that if the user being added is the safe in a form where that address is invalid the input shows the error right away
I've checked in Safe adding/replacing and the recovery feature (both, the recoverer and the change of settings the recoverer can do)

PR #3061, opened 2024-01-03
Fix: [WalletConnect] rm app.optimism.io from blocked bridges

katspaugh:
## What it solves

From the Optimism team:

> we recently implemented a code change to block smart contract wallets from using the bridge on [app.optimism.io](http://app.optimism.io/). Can you remove the block on your end for [app.optimism.io](http://app.optimism.io/) that prevents safe wallets from connecting? this will allow safe wallets to be used for the non-bridge features in our app.

I've removed the blocked host accordingly.

### DIFF
--- a/src/features/walletconnect/constants.ts
+++ b/src/features/walletconnect/constants.ts
@@ -41,7 +41,6 @@ export const BlockedBridges = [
   'cbridge.celer.network',
   'www.orbiter.finance',
   'zksync-era.l2scan.co',
-  'app.optimism.io',
   'www.portalbridge.com',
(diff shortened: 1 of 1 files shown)


usame-algan (approved):
Can confirm that connecting to app.optimism.io works now. Their UI shows a warning when the bridge view is open as expected. I couldn't find any "non-bridge" features to test the connection.

Screenshot 2024-01-03 at 09 37 23

francovenica:
LGTM too. I was able to connect just fine. It shows the error Usame showcased in his comment tho.

PR #3062, opened 2024-01-03
Feat: Scan EIP-712 typed data signature requests

### LINKED ISSUE #2113: [Risk mitigation] Redefine integration extension
## What is the feature about
In our initial risk mitigation module / integration we scan the transaction data.

We should extend this integration by
- Scanning EIP-712 message requests: With off-chain integrations growing, these messages can contain harmful payloads which can potentially steal funds / nfts.

## The list of requirements
- [x] scan off-chain messages: Our new off-chain message modal should also send a scan request with the message as payload.

## Designs/sketches
tbd

schmanu:
## What it solves

Partially resolves #2113 

## How this PR fixes it
- Adds risk analysis to EIP-712 signature requests via Redefine
- Reuses the same components to display results / warnings

## How to test it
- Sign any EIP-712 message from a Safe
- Observe "Scan for risks"-block
- Ideally also try some malicious signature requests

## Screenshots
Screenshot 2024-01-03 at 11 14 30

## Checklist


src/services/security/modules/RedefineModule/index.ts
```diff
@@ -1,6 +1,9 @@
 import { REDEFINE_API } from '@/config/constants'
+import { normalizeTypedData } from '@/utils/web3'
+import { isObjectEIP712TypedData } from '@safe-global/safe-apps-sdk'
```
katspaugh, line 3: Is the implementation of `isObjectEIP712TypedData` complex?
I would like to avoid depending on the apps SDK outside of the Safe Apps functionality.

src/components/tx-flow/flows/SignMessage/SignMessage.tsx
```diff
@@ -167,6 +171,8 @@ export type ConfirmProps = BaseProps & {
 const SignMessage = ({ message, safeAppId, requestId }: ProposeProps | ConfirmProps): ReactElement => {
   // Hooks & variables
   const { setTxFlow } = useContext(TxModalContext)
+  const { setSafeMessage: setContextSafeMessage } = useContext(SafeTxContext)
+  const { needsRiskConfirmation, isRiskConfirmed, setIsRiskIgnored } = useContext(TxSecurityContext)
```
usame-algan, line 175: Should we also disable the Sign button if `needsRiskConfirmation` like we do in the other flows?

src/components/tx-flow/flows/SignMessage/SignMessage.tsx
```diff
@@ -220,13 +238,15 @@ const SignMessage = ({ message, safeAppId, requestId }: ProposeProps | ConfirmPr
           </Typography>
           <DecodedMsg message={decodedMessage} isInModal />
 
-          <Accordion sx={{ mt: 2, '&.Mui-expanded': { mt: 2 } }}>
+          <Accordion sx={{ mb: 2, mt: 2, '&.Mui-expanded': { mt: 2 } }}>
```
usame-algan, line 241:
```suggestion
          <Accordion sx={{ my: 2, '&.Mui-expanded': { mt: 2 } }}>
```

nit

src/components/tx-flow/flows/SignMessage/SignMessage.tsx
```diff
@@ -220,13 +238,15 @@ const SignMessage = ({ message, safeAppId, requestId }: ProposeProps | ConfirmPr
...
             <AccordionSummary expandIcon={<ExpandMoreIcon />}>SafeMessage details</AccordionSummary>
             <AccordionDetails>
               <MessageHashField label="SafeMessage" hashValue={safeMessageMessage} />
               <MessageHashField label="SafeMessage hash" hashValue={safeMessageHash} />
             </AccordionDetails>
           </Accordion>
+
+          <Redefine />
```
usame-algan, line 249: The info icon says on hover `This transaction has been automatically scanned for risks to help prevent scams. ...`. In this case `message` instead of `transaction` would be more fitting. Wdyt?

src/components/tx-flow/flows/SignMessage/SignMessage.tsx
```diff
@@ -256,6 +276,8 @@ const SignMessage = ({ message, safeAppId, requestId }: ProposeProps | ConfirmPr
             <WrongChainWarning />
 
             <MessageDialogError isOwner={isOwner} submitError={submitError} />
+
+            <RiskConfirmationError />
```
usame-algan, line 280: Similar to the above comment, this currently says `I understand the risks and would like to continue this transaction` but `message` or something similar would make more sense imo.

src/services/security/modules/RedefineModule/index.ts
```diff
@@ -106,25 +109,36 @@ export type RedefineResponse = {
...
+    if (isObjectEIP712TypedData(data)) {
+      const normalizedMsg = normalizeTypedData(data)
+      return JSON.stringify(normalizedMsg)
+    } else {
+      return JSON.stringify(
+        generateTypedData({
+          safeAddress,
+          safeVersion: '1.3.0', // TODO: pass to module, taking into account that lower Safe versions don't have chainId in payload
+          chainId,
+          safeTransactionData: (data as SafeTransaction).data,
```
usame-algan, line 123: Strange that we have to cast here even though `isObjectEIP712TypedData` is a type guard. I would expect that in the else case it infers to `SafeTransaction`

src/services/security/modules/RedefineModule/index.ts
```diff
@@ -150,7 +164,7 @@ export class RedefineModule implements SecurityModule<RedefineModuleRequest, Red
           severity: redefineSeverityMap[issue.severity.label],
         })),
         balanceChange: result.data?.balanceChange,
-        simulation: result.data?.simulation,
+        simulation: isObjectEIP712TypedData(data) ? undefined : result.data?.simulation,
```
usame-algan, line 167: Is there no `result.data.simulation` for EIP712 scans?

schmanu: There is, but we only use the simulation ID to render a link to Redefine's dashboard. when those simulation IDs are opened on the Redefine dashboard for EIP-712 scans they only show "No transaction found". 
As long as that is not fixed on their end we should hide the link to their dashboard.

src/services/security/modules/RedefineModule/index.ts
schmanu: I was also disappointed by this. We could have a second typeguard here of course but both is not ideal.

src/services/security/modules/RedefineModule/index.ts
schmanu: It's a one liner. I'll copy it into our utils.

src/components/tx-flow/flows/SignMessage/SignMessage.tsx
schmanu: AFAIK we do not disable it in our other flows. Instead it will highlight the checkbox when trying to submit and abort the submission. That happens here as well.

I think the idea behind is that the user should understand why the submit button is disabled.

src/services/security/modules/RedefineModule/index.ts
schmanu: I realized why... `isObjectEIP712TypedData` returns a `boolean` and not a typeguard. I rewrote it in our project as a proper typeguard.

src/utils/safe-messages.ts
```diff
@@ -29,6 +29,14 @@ const EIP1271_SUPPORTED_SAFE_VERSION = '1.0.0'
 
 const EIP1271_OFFCHAIN_SUPPORTED_SAFE_APPS_SDK_VERSION = '7.11.0'
 
+/**
+ * Typeguard for EIP712TypedData
+ *
+ */
+export const isObjectEIP712TypedData = (obj: any): obj is EIP712TypedData => {
```
katspaugh, line 36:
```suggestion
export const isEIP712TypedData = (obj: any): obj is EIP712TypedData => {
```

IMHO a better name + automatic imports in the IDE won't confuse it with the apps SDK's one.

francovenica:
Looks good on my side.

Used the dummy app https://github.com/5afe/eip-1271-dapp (in the drain branch) to test this malicious type of message

I tried in this PR and current prod to compare.
In current prod the critical risk is not being catched. In this PR it is. 
The user can still sign if he prefers.

Prod:
image

PR:
image

katspaugh (approved):
LGTM 👍

PR #3064, opened 2024-01-04
fix: avoid crashing when selecting a 4-letter currency

### LINKED ISSUE #3045: Assets page doesn't load when changing currency to SATS
Issue Category
Safe loading issue, doesnt load on any interface

What happened?
Encountering an error when changing the currency to SATS, the Assets page fails to load afterward. This issue seems to affect most networks.
image

Date and time this happened or you first noticed this issue
18-12-2023 4:22pm

Network
Ethereum Mainnet

Safe Address
No response

Token Address
No response

Token Type
None

Transactions
No response

compojoom:
## What it solves
On the assets page if the user was to change the currency to SATS, LINK or BITS the app would crash. It turns out that we rely on Intl.NumberFormat for currency, but it only works with ISO 4217 - which requires the currency length to be exactly 3 characters. With anything else the call will throw. For now we decided to ignore any non ISO 4217 currencies.

Resolves #3045

## How this PR fixes it
@katspaugh and I agreed to remove Currencies that are not 3 characters long (as expected by ISO 4217). The currencies feature is not used much and if it is used it is mainly USD.

## How to test it
The list of currencies will no longer show LINK, BITS or SATS - so running into the crash is no longer possible

## Checklist

### DIFF
--- a/src/components/balances/CurrencySelect/__tests__/useCurrencies.test.ts
+++ b/src/components/balances/CurrencySelect/__tests__/useCurrencies.test.ts
@@ -1,0 +1,56 @@
+import { act, renderHook } from '@testing-library/react'
+import { getFiatCurrencies } from '@safe-global/safe-gateway-typescript-sdk'
+import useCurrencies from '../useCurrencies'
+import { Errors, logError } from '@/services/exceptions'
+
+jest.mock('@safe-global/safe-gateway-typescript-sdk', () => ({
+  getFiatCurrencies: jest.fn(),
+}))
+
+jest.mock('@/services/exceptions', () => {
+  const originalModule = jest.requireActual('@/services/exceptions')
+
+  return {
+    ...originalModule, // this will keep the original exports
+    logError: jest.fn(), // this will override logError with a mock function
+  }
+})
+
+describe('useCurrencies', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('should fetch and return the list of ISO 4217 currencies', async () => {
+    const iso4217Currencies = ['USD', 'EUR', 'GBP']
+    const mockCurrencies = [...iso4217Currencies, 'SATS']
+    const getFiatCurrenciesMock = jest.fn(async () => Promise.resolve(mockCurrencies))
+    const myMock = getFiatCurrencies as jest.Mock
+    myMock.mockImplementation(getFiatCurrenciesMock)
+
+    const { result, rerender } = renderHook(() => useCurrencies())
+
+    expect(result.current).toBeUndefined()
+
+    await act(async () => {
+      rerender()
+    })
+
+    expect(result.current).toEqual(iso4217Currencies)
+    expect(getFiatCurrenciesMock).toHaveBeenCalledTimes(1)
+  })
+
+  it('should log an error if fetching currencies fails', async () => {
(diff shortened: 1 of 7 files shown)


katspaugh (approved):
Nice work, congrats on the first PR! 🎉

katspaugh:
CLA will fail until this is merged to dev, which is fine. We can merge it after QA gives a thumbs up.

src/components/balances/CurrencySelect/__tests__/useCurrencies.test.ts
```diff
@@ -0,0 +1,56 @@
...
+jest.mock('@/services/exceptions', () => {
+  const originalModule = jest.requireActual('@/services/exceptions')
+
+  return {
+    ...originalModule, // this will keep the original exports
+    logError: jest.fn(), // this will override logError with a mock function
+  }
+})
+
+describe.only('useCurrencies', () => {
```
schmanu, line 19:
```suggestion
describe('useCurrencies', () => {
```

I think you left this here by mistake

francovenica:
I wasn't sure about removing tokens, but given that nobody every asked for those specifically and it was always broken I think is ok removing those from the list

Looks good to me

src/components/balances/CurrencySelect/__tests__/useCurrencies.test.ts
compojoom: ooops. Thanks for looking.

compojoom: Ok, added a commit with eslint rule that would help us catch those in the future:
https://github.com/safe-global/safe-wallet-web/pull/3064/commits/66d078650861c11609985cd0ee967d9c43e80f1b

schmanu (approved):
Looking good! Gz on your first PR :) 
The lint rule is a nice addition, too!

src/components/balances/CurrencySelect/__tests__/useCurrencies.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    const { result, rerender } = renderHook(() => useCurrencies())
+
+    expect(result.current).toBeUndefined()
+
+    await act(async () => {
+      await rerender()
+    })
+
+    expect(result.current).toEqual(iso4217Currencies)
+    expect(getFiatCurrenciesMock).toHaveBeenCalledTimes(1)
```
usame-algan, line 40: nit: I am getting an info that await is not necessary here. I think it "promisifies" the expression which makes it work?
Screenshot 2024-01-05 at 11 51 30

An alternative is to use `await waitFor` on the expect expression e.g.

```
act(() => {
  rerender()
}

await waitFor(() => {
  expect(...)
}
```

compojoom: with
```
act(() => {
  rerender()
}
```
the test fails

```
async act(() => {
  rerender()
}
```
works, but the IDEA complains as well:
grafik

and then:
```
    await act(async () => {
      rerender()
    })
```

this works and the IDE doesn't complain.

usame-algan: It even works entirely without `act` as long as we wrap the `expect` call(s) in an `await waitFor`.

PR #3070, opened 2024-01-05
fix: Persist timeout for pending transactions

### LINKED ISSUE #2878: Can not cancel "Processing" state
## Bug description
In case when processing tx takes time on the chain side , there is no way to replace or cancel the tx 
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Gnosis chain

## Steps to reproduce
 1. Create tx on gnosis chain
 2. while the tx in the processing state try to replace or cancel it 
 3. wait for 20-30 min in case of issues with processing on the chain side
 Current result: the processing state is blocking the replacement tx for a long time
Seems that the user has to wait around 30 mins
image

## Expected result
Check the time out for the processing state to reset it and allow the TX replacement 
We might need a notification that the tx still can be executed ( please check such a case as well)

usame-algan:
## What it solves

Resolves #2878 

## How this PR fixes it

Stores a `submittedAt` time for pending transactions and checks for that time in `waitForTx` instead of always falling back to the default timeout of 6.5 minutes.

Additionally, we can call `waitForTx` when dispatching an execution as the `.wait()` call doesn't accept a timeout and can be stuck for > 30 minutes if the user doesn't reload the app or reconnect their wallet (to start the `txMonitor`).

## ToDos

- [ ] Add a `Promise.race` and call `waitForTx` when dispatching an execution

## How to test it

1. Execute a transaction and cancel in your wallet
2. Close the app or reload after a few minutes (< 6.5 minutes)
3. Reopen the app after 6.5 minutes
4. Observe a timeout error for the pending transaction
5. Observe the transaction can be executed again

## Checklist

### DIFF
--- a/src/hooks/useTxPendingStatuses.ts
+++ b/src/hooks/useTxPendingStatuses.ts
@@ -36,7 +36,7 @@ const useTxMonitor = (): void => {
       return
     }
 
-    for (const [txId, { txHash, status, taskId, safeAddress }] of pendingTxEntriesOnChain) {
+    for (const [txId, { txHash, status, taskId, safeAddress, submittedAt }] of pendingTxEntriesOnChain) {
       const isProcessing = status === PendingStatus.PROCESSING && txHash !== undefined
       const isMonitored = monitoredTxs.current[txId]
       const isRelaying = status === PendingStatus.RELAYING && taskId !== undefined
@@ -48,7 +48,7 @@ const useTxMonitor = (): void => {
       monitoredTxs.current[txId] = true
 
       if (isProcessing) {
-        waitForTx(provider, txId, txHash)
+        waitForTx(provider, txId, txHash, submittedAt)
         continue
       }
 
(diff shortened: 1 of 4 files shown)


katspaugh (approved):
Muy bien 👍

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -152,25 +152,30 @@ export const dispatchTxExecution = async (
...
+  const provider = getWeb3ReadOnly()
 
-      if (didReprice(error)) {
-        txDispatch(TxEvent.PROCESSED, { ...eventParams, safeAddress })
-      } else {
-        txDispatch(TxEvent.FAILED, { ...eventParams, error: asError(error) })
-      }
-    })
+  // Asynchronously watch the tx to be mined/validated
+  Promise.race([
```
katspaugh, line 158: Nice one 🐎

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -187,68 +192,74 @@ export const dispatchBatchExecution = async (
...
-    .then((receipt) => {
-      if (didRevert(receipt)) {
-        txs.forEach(({ txId }) => {
-          txDispatch(TxEvent.REVERTED, {
-            txId,
-            error: new Error('Transaction reverted by EVM'),
-            groupKey,
+  const provider = getWeb3ReadOnly()
+
+  Promise.race([
```
katspaugh, line 219: Seems to be almost the same code as above?

usame-algan:
> seems that only replace the tx with the same wallet nonce will actually help. It would be nice re-submit with the same wallet nonce

I've added a new issue to address this: https://github.com/safe-global/safe-wallet-web/issues/3107

PR #3071, opened 2024-01-05
Experimental chore: add lavamoat allow-scripts

schmanu:
## What it does
Adds lavamoat allow-scripts

## How to test it
run `yarn setup` 

## Todo
- [ ] document how to add / update dependencies

## Checklist

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,1 @@
+enableScripts: false
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+ignore-scripts true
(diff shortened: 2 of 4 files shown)


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+enableScripts: false
```
katspaugh, line 1: Are both yarnrc's required?

schmanu: I think the .yarnrc.yml is for yarn 2.0 and can be removed. 
The lavamoat cli auto-created these files during setup.

katspaugh:
What shall we do with this PR? Could you update the package files and merge it?

PR #3076, opened 2024-01-06
Tests: E2e - Remove owner - add signing action

francovenica:
## What it solves
Adds the action of signing to a RemoveOnwer test
It checks the tx is created.

The test is a single thing (signing + verification) because checking the tx was created is the assertion.

Note:
There is still an issue of going to step1 to step2 too fast that the SDK fails to load. I wasn't able to found a call or anything that tells me is done loading so I had to just put a cy.wait which I know is not optimal.

## Checklist


cypress/e2e/pages/owners.pages.js
```diff
@@ -109,6 +111,21 @@ export function openRemoveOwnerWindow(btn) {
   cy.get('div').contains(removeOwnerStr).should('exist')
 }
 
+export function getAddressToBeRemoved() {
```
mike10ca, line 114: I would rather avoid using this method by few reasons: a) it takes always first value, and b) it does not use controlled value. All test values must be controlled and predictable. This is because if there is a bug and the value is incorrect, but the method is it is now just takes the first value no matter what, we will get poor assertion. 
Please add data-testid to the address value field, put address value in constants (if it is not there already), and assert that the address for deletion is what we expect.

cypress/e2e/smoke/remove_owner.cy.js
```diff
@@ -2,21 +2,31 @@ import * as constants from '../../support/constants'
...
     owner.openRemoveOwnerWindow(1)
+    cy.wait(3000)
     createwallet.clickOnNextBtn()
+    //This method creates the @removedAddress alias
+    owner.getAddressToBeRemoved()
     owner.verifyOwnerDeletionWindowDisplayed()
+    createTx.clickOnSignTransactionBtn()
+    createTx.waitForProposeRequest()
+    createTx.clickViewTransaction()
+    createTx.clickOnTransactionItemByName('removeOwner')
```
mike10ca, line 29: Please remove this line as it is not needed here

cypress/e2e/smoke/remove_owner.cy.js
```diff
@@ -2,21 +2,31 @@ import * as constants from '../../support/constants'
...
+    createTx.verifyTxDestinationAddress('@removedAddress')
```
mike10ca, line 30: Rewrite this method as per new assertions requirement

mike10ca (changes requested):

mike10ca (approved):

PR #3077, opened 2024-01-08
Tests: E2e Assets - Adding the token test file

francovenica:
## What it solves

- Removed Balances.cy.js because the Assets.cy.js file has the same tests and more, so it was redundant. (so it wasn't a merge, but a removal of the file)
- Changed the name of assets.cy.js file to tokens.cy.js as we discussed
- Changed the name of the page from balances.pages.js to assets.pages.js .

Other changes:

- Deleted the test: 'Verify that a tool tip is shown pointing to "Token list" dropdown' because it was doing just a reload of the page
- Changed the method verifyAssetNameHasExplorerLink() to remove the "sepolia" parameter since we are now fully using sepolia safes
- Added the method main.verifyOwnerConnected() so it waits for the user to fully connect before doing anything else. The test where it was being used was failing otherwise
- Changed the test 'Verify that the Send button shows when hovering a row' with the new title as the comments asked (didn't add it to the smoke yet tho)
- Added the test 'Verify that when connected user is not owner, Send button is disabled' since it was asked in a comment of the test above

## Checklist


katspaugh:
@francovenica please prefix the PR title with `Tests: `. Same goes for the initial commit.

mike10ca (changes requested):
Test in balances_pagination.cy.js fails. Please update the code accordingly.

francovenica:
Fixed the 2 test to check the button as disabled instead of the tooltip.
Also run the test locally and they all passed

mike10ca:
> Test in balances_pagination.cy.js fails. Please update the code accordingly.

@francovenica there is issue with balances_pagination.cy.js test file, it won't run tests, please try running this specific file on your local machine and fix please.

francovenica:
> > Test in balances_pagination.cy.js fails. Please update the code accordingly.
> 
> @francovenica there is issue with balances_pagination.cy.js test file, it won't run tests, please try running this specific file on your local machine and fix please.

I forgot to add the proper "import" at the top of the file when I removed the "balances" file and replaced it with the "assets" file

mike10ca (approved):

PR #3082, opened 2024-01-08
Fix: restore Ledger ethereumjs dependency

katspaugh:
## What it solves

Resolves https://www.notion.so/safe-global/Can-not-execute-using-hardwallets-04a665fe2ca74e7c8e03ed8154449067

Tx execution with Ledger wasn't working because of an alias for the `@ethereumjs/common` package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
     "node": ">=16"
   },
   "resolutions": {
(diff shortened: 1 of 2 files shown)


liliya-soroka:
Verified, works with Ledger nano X and Trezor. can be merged to the RC 1.26.0

PR #3084, opened 2024-01-08
fix: allow localhost URLs with subdomains when adding custom safe apps

### LINKED ISSUE #1108: Cannot add localhost IPFS safe app.
## Bug description
You cannot add safe apps served from localhost via the UI.

## Environment
 - Browser: Brave
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Run IPFS Desktop.
 2. Click apps tab on the left.
 3. Click Add custom app button.
 4. Paste a http://bafybeiasatviqudql65qxx66ois3dr7vxmhkk7zawn2vhdeax6q6pn4lsa.ipfs.localhost:8080/ as the App URL
 5. Notice the error "The url is invalid".
 6. Navigate to https://app.safe.global/eth:<safe_address>/apps?appUrl=http://bafybeiasatviqudql65qxx66ois3dr7vxmhkk7zawn2vhdeax6q6pn4lsa.ipfs.localhost:8080/
 7. Notice that the app loads just fine.

## Expected result
Can use a locally served dapp with UI.

## Obtained result
Cannot use a locally served dapp with UI.

## Screenshots
image

## Additional Details
I believe `localhost` is now considered a secure context in all major browsers (see https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#when_is_a_context_considered_secure), as well as `*.localhost`.  It appears the UI is rejecting anything that doesn't start with `https`, which only applies to non-localhost names.  If the hostname is `*.localhost` then it should be treated the same as `https` with regards to security.

----

It is worth noting, that serving a site locally or using IPFS is by far the most secure way to interact with a dapp, and it is quite unfortunate that the UI currently doesn't allow users to actually operate things securely.

jmealy:
## What it solves

When adding a custom safe app, if a localhost URL includes subdomains, it will fail the validation. For example `http://localhost:3001/` will be allowed, but `http://foobar.localhost:3000/` will fail.

Resolves #1108 

## How this PR fixes it
- allows localhost domains which include subdomains when adding a custom safe app.
- both `http://localhost:3000` and `http://foobar.localhost:3000` will be allowed.

## How to test it
- Open the `Add a custom safe app` modal.
- Enter a localhost URL with subdomains. The example from the ticket is http://bafybeiasatviqudql65qxx66ois3dr7vxmhkk7zawn2vhdeax6q6pn4lsa.ipfs.localhost:8080/  
- To be able to add the url you will need to run the IPFS desktop app: https://docs.ipfs.tech/install/ipfs-desktop/ 
Although this fix is only to do with the form validation so it isn't necessary.
- There should be no `The url is invalid` error shown. If you are running the safe app locally you will be able to add it to your custom safe apps. If you are not running it locally you will see the error: `The app doesn't support Safe App functionality`

## Screenshots
Before:

After:
Screenshot 2024-01-08 at 13 41 59

## Checklist

### DIFF
--- a/src/utils/__tests__/validation.test.ts
+++ b/src/utils/__tests__/validation.test.ts
@@ -5,6 +5,7 @@ import {
   validatePrefixedAddress,
   validateDecimalLength,
   isValidAddress,
+  isValidURL,
 } from '@/utils/validation'
 
 describe('validation', () => {
@@ -127,6 +128,24 @@ describe('validation', () => {
 
       const result3 = validateDecimalLength('1', 0)
       expect(result3).toBeUndefined()
+    })
+  })
+
+  describe('URL validation', () => {
+    it.only('returns true for localhost URLs', () => {
+      const result1 = isValidURL('http://localhost:3000')
+      expect(result1).toBe(true)
+
+      const result2 = isValidURL('http://subdomain.localhost:3000')
+      expect(result2).toBe(true)
+    })
+
+    it.only('returns false for non https domains ', () => {
+      const result1 = isValidURL('https://example.com')
+      expect(result1).toBe(true)
+
+      const result2 = isValidURL('http://example.com')
+      expect(result2).toBe(false)
     })
   })
 })
--- a/src/utils/__tests__/validation.test.ts
+++ b/src/utils/__tests__/validation.test.ts
@@ -132,7 +132,7 @@ describe('validation', () => {
   })
 
   describe('URL validation', () => {
-    it.only('returns true for localhost URLs', () => {
+    it('returns true for localhost URLs', () => {
       const result1 = isValidURL('http://localhost:3000')
       expect(result1).toBe(true)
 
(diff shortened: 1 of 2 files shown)


iamacook (approved):

usame-algan (approved):
Nice solution 💯 

I noticed some `AddressInput` tests randomly failing but it is unrelated to this PR. I suspect some rogue state during ENS resolution where the value contains the shortName of the current chain that we mock with faker. Might be worth exploring as a separate task.

jmealy:
> Nice solution 💯
> 
> I noticed some `AddressInput` tests randomly failing but it is unrelated to this PR. I suspect some rogue state during ENS resolution where the value contains the shortName of the current chain that we mock with faker. Might be worth exploring as a separate task.

Thanks Usame! Just fyi, the merge was blocked because the first few commits weren't signed so I had to rebase and force push to make them all signed.

PR #3088, opened 2024-01-09
Analytics: key event types

katspaugh:
* I've formalized the key event types in an enum
* Renamed tx events to past tense to match the existing events
* Added SAFE_CREATED, SAFE_OPENED and WALLET_CONNECTED. They were previously tracked via GTM by eventAction:

Screenshot 2024-01-09 at 11 45 37

When this change is released, we'll need to remove that config on GTM.

## How to test it

* Go to the /welcome page
* Open the console
* Open one of the Safes in the sidebar
* Observe a `[GTM]` event in the console with `eventName: 'safe_opened'`

The same for wallet_connected, safe_created, tx_created, tx_confirmed and tx_executed.

### DIFF
--- a/src/services/analytics/events/createLoadSafe.ts
+++ b/src/services/analytics/events/createLoadSafe.ts
@@ -52,7 +52,7 @@ export const CREATE_SAFE_EVENTS = {
     category: CREATE_SAFE_CATEGORY,
   },
   CREATED_SAFE: {
-    event: EventType.META,
+    event: EventType.SAFE_CREATED,
     action: 'Created Safe',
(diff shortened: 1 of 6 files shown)


usame-algan (approved):

liliya-soroka:
image
failed tx is tracked as executed - try to execute tx that will fail using Relay

katspaugh:
That's correct, we track the intention, not the result.
`wallet_connected` is correct too.

liliya-soroka:
ok, can be merged to the RC 1.26.0

PR #3090, opened 2024-01-10
Fix: don't treat user rejections as errors

### LINKED ISSUE #2701: [Tx flow] Don't treat user rejection as an error
1) Don't log "user rejected" as an error to Sentry
Screenshot 2023-10-30 at 07 29 10
2) Don't show a notification for that
Screenshot 2023-10-30 at 07 30 42

jmealy:
## What it solves

Resolves #2701

## How this PR fixes it
When a user rejects a transaction through their connected wallet:
- do not log it as an error to sentry.
- do not send the user an error notification.
- Show `The transaction was rejected by the user.` in the tx execution box instead of a generic error message.

## How to test it
- Create a new transaction.
- When you are prompted to accept the transaction in Metamask, click reject.
- There should not be an error logged to sentry.
- There should be no error notification.
- The Execution box should say `The transaction was rejected by the user.`
- This will effect anywhere the user is prompted to accept or reject a transaction, including:
    - signing a tx
    - executing a tx
    - signing a message
    - batch txs
    - tx with spending limits
    - relayed txs
    - bulk txs
    - account recovery

## Screenshots
Before: 
Screenshot 2024-01-10 at 09 51 20

After:
Screenshot 2024-01-10 at 15 36 39

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
+++ b/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
@@ -102,13 +102,12 @@ const ReviewSpendingLimitTx = ({
     } catch (_err) {
       const err = asError(_err)
       if (isWalletRejection(err)) {
-        setIsSubmittable(true)
         setIsRejectedByUser(true)
       } else {
         logError(Errors._801, err)
-        setIsSubmittable(true)
         setSubmitError(err)
       }
+      setIsSubmittable(true)
     }
 
     trackEvent({ ...TX_EVENTS.CREATE, label: TX_TYPES.transfer_token })
--- a/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
+++ b/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
@@ -26,7 +26,6 @@ import TxCard from '@/components/tx-flow/common/TxCard'
 import { TxModalContext } from '@/components/tx-flow'
 import { type SubmitCallback } from '@/components/tx/SignOrExecuteForm'
 import { TX_EVENTS, TX_TYPES } from '@/services/analytics/events/transactions'
-import { ErrorCode } from '@ethersproject/logger'
 import { isWalletRejection } from '@/utils/wallets'
 
 export type SpendingLimitTxParams = {
--- a/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
+++ b/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
@@ -20,7 +20,6 @@ import { getTxOptions } from '@/utils/transactions'
 import { MODALS_EVENTS, trackEvent } from '@/services/analytics'
 import useOnboard from '@/hooks/wallets/useOnboard'
 import { WrongChainWarning } from '@/components/tx/WrongChainWarning'
-import { asError } from '@/services/exceptions/utils'
 import TxCard from '@/components/tx-flow/common/TxCard'
 import { TxModalContext } from '@/components/tx-flow'
 import { type SubmitCallback } from '@/components/tx/SignOrExecuteForm'
@@ -137,9 +136,7 @@ const ReviewSpendingLimitTx = ({
           <ErrorMessage error={submitError}>Error submitting the transaction. Please try again.</ErrorMessage>
         )}
 
-        {isRejectedByUser && (
-          <ErrorMessage>You've rejected the transaction.</ErrorMessage>
-        )}
+        {isRejectedByUser && <ErrorMessage>{`You've rejected the transaction.`}</ErrorMessage>}
 
         <Typography variant="body2" color="primary.light" textAlign="center">
           You&apos;re about to create a transaction and will need to confirm it with your currently connected wallet.
--- a/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
+++ b/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
@@ -25,6 +25,8 @@ import TxCard from '@/components/tx-flow/common/TxCard'
 import { TxModalContext } from '@/components/tx-flow'
 import { type SubmitCallback } from '@/components/tx/SignOrExecuteForm'
 import { TX_EVENTS, TX_TYPES } from '@/services/analytics/events/transactions'
+import { type EthersError } from '@/utils/ethers-utils'
+import { ErrorCode } from '@ethersproject/logger'
 
 export type SpendingLimitTxParams = {
   safeAddress: string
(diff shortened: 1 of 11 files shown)


src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
```diff
@@ -170,6 +178,12 @@ export const ExecuteForm = ({
           </Box>
         )}
 
+        {isRejectedByUser && (
+          <Box mt={1}>
+            <ErrorMessage>The transaction was rejected by the user.</ErrorMessage>
```
katspaugh, line 183:
```suggestion
            <ErrorMessage>You've rejected the transaction.</ErrorMessage>
```
perhaps?

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -85,10 +86,12 @@ export const dispatchTxSigning = async (
   try {
     signedTx = await tryOffChainTxSigning(safeTx, safeVersion, sdk)
   } catch (error) {
-    txDispatch(TxEvent.SIGN_FAILED, {
-      txId,
-      error: asError(error),
-    })
+    if ((error as EthersError).code !== ErrorCode.ACTION_REJECTED) {
```
katspaugh, line 89: This covers signatures but there's also executions and relays and other types of actions (basically all dispatch functions in this file).

src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
jmealy: Yeah make more sense 👍

src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
```diff
@@ -128,6 +136,8 @@ const ReviewSpendingLimitTx = ({
           <ErrorMessage error={submitError}>Error submitting the transaction. Please try again.</ErrorMessage>
         )}
 
+        {isRejectedByUser && <ErrorMessage>{`You've rejected the transaction.`}</ErrorMessage>}
```
usame-algan, line 139:
```suggestion
        {isRejectedByUser && <ErrorMessage>You&apos;ve rejected the transaction.</ErrorMessage>}
```

src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
```diff
@@ -96,9 +99,14 @@ const ReviewSpendingLimitTx = ({
...
-      setIsSubmittable(true)
-      setSubmitError(err)
+      if ('code' in err && err.code === ErrorCode.ACTION_REJECTED) {
+        setIsSubmittable(true)
+        setIsRejectedByUser(true)
+      } else {
+        logError(Errors._801, err)
+        setIsSubmittable(true)
+        setSubmitError(err)
+      }
```
usame-algan, line 109: We can reuse `isUserRejection` here and pull out `setIsSubmittable` from the condition since its set to true in both cases. This can also be done for the same block in `SignForm` and `ExecuteForm`.

src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
```diff
@@ -170,6 +178,12 @@ export const ExecuteForm = ({
           </Box>
         )}
 
+        {isRejectedByUser && (
+          <Box mt={1}>
+            <ErrorMessage>{`You've rejected the transaction.`}</ErrorMessage>
```
usame-algan, line 183:
```suggestion
            <ErrorMessage>You&apos;ve rejected the transaction.</ErrorMessage>
```

src/components/tx/SignOrExecuteForm/SignForm.tsx
```diff
@@ -97,6 +105,12 @@ export const SignForm = ({
         )
       )}
 
+      {isRejectedByUser && (
+        <Box mt={1}>
+          <ErrorMessage>{`You've rejected the transaction.`}</ErrorMessage>
```
usame-algan, line 110:
```suggestion
          <ErrorMessage>You&apos;ve rejected the transaction.</ErrorMessage>
```

src/hooks/messages/useSafeMessageNotifications.ts
```diff
@@ -24,6 +25,10 @@ const SafeMessageNotifications: Partial<Record<SafeMsgEvent, string>> = {
   [SafeMsgEvent.SIGNATURE_PREPARED]: 'The message was successfully confirmed.',
 }
 
+const isUserRejection = (error: Error) => {
```
usame-algan, line 28: Maybe we can move this function into one of the utils files like `ethers-utils.ts` or create a new one for error utils so we don't have to define it twice and can reuse the logic in other places.

src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
usame-algan: This error message is defined in 4 places. I would suggest to extract it into a separate component like we do for some of the other errors like `NonOwnerError` to avoid duplication.

usame-algan (commented):
Nice work on this! 💯 

Only got a few small refactoring suggestions.

src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
```diff
@@ -96,9 +99,14 @@ const ReviewSpendingLimitTx = ({
       setTxFlow(undefined)
     } catch (_err) {
       const err = asError(_err)
-      logError(Errors._801, err)
-      setIsSubmittable(true)
-      setSubmitError(err)
+      if ('code' in err && err.code === ErrorCode.ACTION_REJECTED) {
```
katspaugh, line 102: Might be good to create a utility for this, perhaps in `src/utils/wallets.ts`.

Actually, I've just checked and that file already has sWalletRejection :)

jmealy: Perfect, thanks!

katspaugh:
@jmealy we don't have a pre-commit hook for lint, because it takes such a long time, but please try to run `yarn lint --fix` locally before a commit.

katspaugh (approved):
👍

usame-algan (approved):
Thanks! 🚀

liliya-soroka:
Opinion about the message: the user can mess up "reject tx as an owner in the connected walley" with reject original tx in the safe. "User rejected signing" was closer to the reality

liliya-soroka:
1. Reject bulk execution : 
Current result: no notification, but the error on tx deatls is the same as "before"
image

jmealy:
> User rejected signing

This makes sense, I'll change the message to `User rejected signing`

liliya-soroka:
Verified

PR #3091, opened 2024-01-10
Fix: send history timezone_offset to gateway

### LINKED ISSUE #3001: Tx list: wrong date labels due to client-side timezone conversion
## Problem
There's a mismatch between DATE_LABEL dates and dates displayed in tx summaries/details.

This is due to the fact that the DATE_LABEL dates come in 00:00 format, while tx dates have an actual hour/minute.
So it can happen that the midnight one gets assigned to a different date due to a timezone shift.

## Solution
Pass a timezone_offset parameter to the backend to group by date in a consistent manner on the backend. This param should be sent to all endpoints returning a list of txs (i.e. history, filtered history, queue).

katspaugh:
## What it solves

Resolves #3001.

Adds a timezone_offset parameter to tx history requests as per https://github.com/safe-global/safe-client-gateway/issues/924

Types were added here: https://github.com/safe-global/safe-gateway-typescript-sdk/pull/142

## How this PR fixes it

History and filtered history requests now send a local timezone shift in milliseconds so that the gateway can group them by date more correctly.

This fixes the bug described in https://github.com/safe-global/safe-client-gateway/issues/924


src/hooks/useTxHistory.ts
```diff
@@ -29,7 +30,7 @@ const useTxHistory = (
 
       return filter
         ? fetchFilteredTxHistory(chainId, safeAddress, filter, pageUrl)
-        : getTransactionHistory(chainId, safeAddress, pageUrl)
```
usame-algan, line 32: Wouldn't this call only fetch trusted transactions but now with the change would also fetch untrusted ones?

src/services/transactions/index.ts
```diff
@@ -0,0 +1,30 @@
+import memoize from 'lodash/memoize'
+import { getTransactionDetails, getTransactionHistory } from '@safe-global/safe-gateway-typescript-sdk'
+
+const tzOffset = new Date().getTimezoneOffset() * 60 * 1000
```
usame-algan, line 4: nit: I would write the word out like `timezoneOffset` for better readability

usame-algan (approved):

src/hooks/useTxHistory.ts
katspaugh: Yes, and this is correct. That it wasn't passed was incorrect. We want to always load untrusted txs. This bug didn't manifest itself only because gateway apparently ignores the query when a pageUrl is passed.

usame-algan: Wow, nice catch then! I was already wondering how we currently display untrusted transactions in the history even though we don't pass false here. 💯

usame-algan:
The steps I used for testing:

1. Open this safes history: eth:0xA77DE01e157f9f57C7c4A326eeE9C4874D0598b6
2. Scroll down to the April 12, 2022 section
3. Change the system timezone to something large like UTC-10
4. On staging CGW it should still display April 12, 2022 but on prod it should switch to April 11, 2022 even though the transaction below the date label still says April 12

francovenica:
I tried in my timezone (-3 GTM). I have this difference in Prod and this PR
Safe: https://timezone--walletweb.review-wallet-web.5afe.dev/transactions/history?safe=eth:0x8675B754342754A30A2AeF474D114d8460bca19b

Prod:
image

PR:
image

francovenica:
Note: I wanted to check more tx at the bottom, but the tooltips don't work in newly loaded elements on the list:

https://github.com/safe-global/safe-wallet-web/issues/3115

katspaugh:
@fmrsabino any idea why the bug on Franco’s screenshot is happening?

fmrsabino:
> @fmrsabino any idea why the bug on Franco’s screenshot is happening?

I cannot reproduce this issue. If the timezone is -3 hours compared to UTC then we would have a timezone offset of `-10800000` (3 hours * 60 minutes * 60 seconds * 1000 milliseconds).

https://safe-client.staging.5afe.dev/v1/chains/1/safes/0x8675B754342754A30A2AeF474D114d8460bca19b/transactions/history?timezone_offset=-10800000

```
{
      "type": "DATE_LABEL",
      "timestamp": 1702354487000
    },
    {
      "type": "TRANSACTION",
      "transaction": {
        "id": "multisig_0x8675B754342754A30A2AeF474D114d8460bca19b_0x99e3df4a6b711b837b843b1146a7f849ca0ff50b0afe085f0b063ec53c3c8202",
        ...
          "nonce": 343,
        ...
    },
    {
      "type": "DATE_LABEL",
      "timestamp": 1702335155000
    },
    {
      "type": "TRANSACTION",
      "transaction": {
        "id": "multisig_0x8675B754342754A30A2AeF474D114d8460bca19b_0x68ebe3a904a9bf5f12aa9e55c845c0a16bed9d7aa2e2effd15307fe7a156918d",
        ...
          "nonce": 342,
        ...
    },
```

From the response, we have two different sections.
Nonce 343 – Dec 12 2023 04:14:47 UTC
Nonce 342 – Dec 11 2023 22:52:35 UTC

If we apply a timezone offset of -3 hours:
Nonce 343 – Dec 12 2023 01:14:47
Nonce 342 – Dec 11 2023 19:52:35

There should still be two groups (as they are represented in the response that I shared).

fmrsabino:
> THis is another one, tx created near midnight. It shows the same way in this PR and in Prod as well

That one also appears in a single group in the CGW – `GET https://safe-client.staging.5afe.dev/v1/chains/1/safes/0x8675B754342754A30A2AeF474D114d8460bca19b/transactions/history?timezone_offset=-10800000`

```javascript
{
  "type": "DATE_LABEL",
  "timestamp": 1699843631000
},
{
  "type": "TRANSACTION",
  "transaction": {
    "id": "multisig_0x8675B754342754A30A2AeF474D114d8460bca19b_0x42359a788436f6e9745fb70502c500ee671ba2b16dd3458f375351e7791432a0",
    "timestamp": 1699843631000,
    "txStatus": "SUCCESS",
    "txInfo": {
      "type": "Custom",
      "humanDescription": null,
      "to": {
        "value": "0x40A2aCCbd92BCA938b02010E17A5b8929b49130D",
        "name": "Safe: MultiSendCallOnly 1.3.0",
        "logoUri": "https://safe-transaction-assets.staging.5afe.dev/contracts/logos/0x40A2aCCbd92BCA938b02010E17A5b8929b49130D.png"
      },
      "dataSize": "324",
      "value": "0",
      "methodName": "multiSend",
      "actionCount": 3,
      "isCancellation": false
    },
    "executionInfo": {
      "type": "MULTISIG",
      "nonce": 327,
      "confirmationsRequired": 1,
      "confirmationsSubmitted": 1,
      "missingSigners": null
    },
    "safeAppInfo": null
  },
  "conflictType": "None"
},
{
  "type": "DATE_LABEL",
  "timestamp": 1698894155000
},
```

francovenica:
I've checked on our test app in mobile (Stg environment), and it works fine there. All the cases I've seen so far that have issues in desktop work fine in mobile:

image

katspaugh:
@francovenica fixed ✅

francovenica:
The issues I mentioned before in my coments were solved, now I can see the tx under the correct labels

There is these cases I found last minute, where I'm suspicious is being caused by the pagination. If it is, I think we can pass this ticket and deal with pagination in another ticket:

image
image

PR #3096, opened 2024-01-10
Feat: toggle ghost transactions

katspaugh:
## What does it solve

Allow hiding "ghost" transactions.

Screenshot 2024-01-10 at 17 03 15

## Implementation details

I've added a new persistent user setting called `showOnlyTrustedTransactions` which controls the `trusted` parameter sent to the backend.

Unfortunately, the backend doesn't handle pagination very well when `trusted` is set to true. It filters individual JSON pages instead of the entire DB query, so in very spammed Safes like Vitalik's, the UI can make several simultaneous requests to fetch enough JSON pages to fill the web page (due to an inifinite scroll pagination).


src/services/analytics/events/txList.ts
```diff
@@ -54,6 +54,12 @@ export const TX_LIST_EVENTS = {
     category: TX_LIST_CATEGORY,
     event: EventType.CLICK,
   },
+  TOGGLE_UNTRUSTED: {
+    action: 'Toggle untrusted transactions',
+    category: TX_LIST_CATEGORY,
+    event: EventType.CLICK,
+    // label: 'hide' | 'show',
```
schmanu, line 61: Can we remove this?

schmanu (approved):
Looks good!

src/services/analytics/events/txList.ts
katspaugh: Why? It's supposed to be a piece of documentation :)
We have the same line in a few other events.

schmanu: ahh okay. Now I understand the comment 💡 
Nevermind, it's late 😪

TanyaEfremova:
The button styles on Assets and Transaction history seems to be different. On the Asset page we already have the same button to hide tokens. Have you tried this button's style, or it doesn't work well?

katspaugh:
Yes, I tried that button but it doesn't play well with the Filter dropdown on the history page.

Btw I think it's time to remove the manual token hiding and that button along with it, because this feature was superceded with the Token list dropdown. I'll double-check the usage in our analytics to validate this.

liliya-soroka:
Verified

PR #3099, opened 2024-01-11
fix: [Spending Limit] wrong decoded data shown

### LINKED ISSUE #2397: Resetting spending-limit shows wrong transaction for a second
Reported here: https://github.com/5afe/safe-support/issues/268

## Bug description

When resetting a one-time spending limit, the UI shortly displays an `addDelegate` transaction in the decoded tx view.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Prod

## Steps to reproduce
 1. Go to Settings > Spending Limits
 2. Create a new one-time spending limit
 3. Execute a spending limit transaction with that
 4. Go to Settings > Spending Limits
 5. Reset the spending limit to another one-time spending limit
 6. Observe the UI showing wrong transaction in the decoded tx view for a second while loading

## Expected result

The decoded tx view should not show wrong transactions for spending limits

usame-algan:
## What it solves

Resolves #2397

## How this PR fixes it

`existingSpendingLimit` was set inside a `useEffect` which caused an additional call to `createNewSpendingLimitTx` with an undefined value for `existingSpendingLimit`. This results in a fallback and displays a wrong transaction to the user inside the decoded data field. Depending on which useEffect finishes first this wrong result can stay visible.

This PR moves `existingSpendingLimit` into a `useMemo` to prevent the additional call.

## How to test it

1. Go to Settings > Spending Limits
1. Create a new one-time spending limit
1. Execute a spending limit transaction with that
1. Go to Settings > Spending Limits
1. Reset the spending limit to another one-time spending limit
1. Observe the UI not showing a wrong transaction in the decoded tx view

## Screenshots
Before: This is the wrong transaction for resetting an already used spending limit
Screenshot 2024-01-11 at 12 20 34

After:
Screenshot 2024-01-11 at 12 19 49

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/NewSpendingLimit/ReviewSpendingLimit.tsx
+++ b/src/components/tx-flow/flows/NewSpendingLimit/ReviewSpendingLimit.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useMemo, useContext } from 'react'
+import { useEffect, useMemo, useContext } from 'react'
 import { useSelector } from 'react-redux'
 import { BigNumber } from 'ethers'
 import { Typography, Grid, Alert } from '@mui/material'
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
👍

liliya-soroka:
Same on the first page of a new spending limits creation with the same token and beneficiary 
image

usame-algan:
> Same on the first page of a new spending limits creation with the same token and beneficiary , when the reset time was changed.

Did you spend part of that spending limit before trying to reset? In case the spending limit wasn't used by the beneficiary it is expected that only `setAllowance` is called. If part of it is already used we have to additionally call `resetAllowance`.

liliya-soroka:
thank you @usame-algan , missed this case.
We will try to add these cases to the spending limits cypress regression run

mike10ca:
> thank you @usame-algan , missed this case. We will try to add these cases to the spending limits cypress regression run Verified

@liliya-soroka [tests](https://github.com/safe-global/safe-wallet-web/pull/4038) were added

PR #3105, opened 2024-01-12
feat: show approval editor for permit2 messages

schmanu:
## What it solves
Permit2 signature are not highlighted in our UI.

## How this PR fixes it
Shows the readonly approval editor for `PermitSingle` and `PermitBatch` Permit2 messages in SignOnchain and Offchain modals

## How to test it
- Request a signature for a PermitSingle or PermitBatch message (I extended the [example drainer app](https://github.com/5afe/eip-1271-dapp/tree/drain-wallet-example-dapp))
- Observe the new Approval Editor

## Screenshots
Screenshot 2024-01-12 at 16 48 10

Screenshot 2024-01-12 at 18 03 37

## Checklist

### DIFF
--- a/src/components/tx/ApprovalEditor/hooks/useApprovalInfos.test.ts
+++ b/src/components/tx/ApprovalEditor/hooks/useApprovalInfos.test.ts
@@ -6,9 +6,11 @@ import { createMockSafeTransaction } from '@/tests/transactions'
 import { OperationType } from '@safe-global/safe-core-sdk-types'
 import { ERC20__factory } from '@/types/contracts'
 import * as balances from '@/hooks/useBalances'
-import { TokenType } from '@safe-global/safe-gateway-typescript-sdk'
+import { type EIP712TypedData, TokenType } from '@safe-global/safe-gateway-typescript-sdk'
 import { BigNumber } from '@ethersproject/bignumber'
 import * as getTokenInfo from '@/utils/tokens'
+import { faker } from '@faker-js/faker'
+import { PSEUDO_APPROVAL_VALUES } from '../utils/approvals'
 
 const ERC20_INTERFACE = ERC20__factory.createInterface()
 
@@ -22,7 +24,7 @@ describe('useApprovalInfos', () => {
   })
 
   it('returns an empty array if no Safe Transaction exists', async () => {
-    const { result } = renderHook(() => useApprovalInfos(undefined))
+    const { result } = renderHook(() => useApprovalInfos({}))
 
     expect(result.current).toStrictEqual([[], undefined, true])
 
@@ -38,7 +40,7 @@ describe('useApprovalInfos', () => {
       operation: OperationType.DelegateCall,
     })
 
-    const { result } = renderHook(() => useApprovalInfos(mockSafeTx))
+    const { result } = renderHook(() => useApprovalInfos({ safeTransaction: mockSafeTx }))
 
     await waitFor(() => {
       expect(result.current).toStrictEqual([[], undefined, false])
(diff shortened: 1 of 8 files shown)


src/components/tx/ApprovalEditor/index.tsx
```diff
@@ -29,17 +30,26 @@ const Title = () => {
...
 }
 
-export const ApprovalEditor = ({ safeTransaction }: { safeTransaction: SafeTransaction | undefined }) => {
+export const ApprovalEditor = ({
+  safeTransaction,
+  safeMessage,
+}: {
+  safeTransaction?: SafeTransaction
+  safeMessage?: EIP712TypedData
+}) => {
```
usame-algan, line 39: With `safeMessage` added here we have to extend a lot of the conditions and it propagates down through `useApprovalInfos` to the `ApprovalModule` and adds more complexity.

Afaiu if there is a `safeMessage` we are always in readOnly so we could split this into two components, one for transactions one for safe messages and add a new method `ApprovalModule.scanMessage` to handle this case and check inside `useApprovalInfos` in one place if `safeMessage ? ApprovalModule.scanMessage : ApprovalModule.scanTransaction`. Wdyt?

schmanu: I think adding a scanMessage function helps readability of the `ApprovalModule` but I do not think we need to create a separate component as most of the components logic is shared / we use the same type for Permits as for Approvals.

I will refactor it tomorrow a bit more :)

usame-algan (approved):
I would suggest to extend the tests for `ApprovalEditor` to cover the new conditions

- Should return null if there is no safeMessage
- Should be read-only if there is a safeMessage

src/services/security/modules/ApprovalModule/index.ts
```diff
@@ -52,11 +60,59 @@ export class ApprovalModule implements SecurityModule<ApprovalModuleRequest, App
     return []
   }
 
-  scanTransaction(request: ApprovalModuleRequest): SecurityResponse<ApprovalModuleResponse> {
-    const { safeTransaction } = request
-    const safeTxData = safeTransaction.data.data
+  private static scanPermitDetails(details: PermitDetails): Pick<Approval, 'amount' | 'tokenAddress'> {
```
usame-algan, line 63: Would it make more sense to name this `getPermitDetails`? Its not really scanning something.

src/components/tx/ApprovalEditor/hooks/useApprovalInfos.test.ts
```diff
@@ -95,6 +97,219 @@ describe('useApprovalInfos', () => {
...
+          },
+        ],
+      },
+    }
+
+    const { result } = renderHook(() => useApprovalInfos({ safeMessage: mockMessage }))
+
+    const expectedApprovals: ApprovalInfo[] = [
+      {
+        amount: getTokenInfo.UNLIMITED_PERMIT2_AMOUNT.toBigInt(),
```
compojoom, line 291: everywhere else i see Bignumber used, but here we do toBigInt(). Why?

francovenica:
So I checked with the dummy app and the approval editor shows up just fine in this PR. For comparison I've checked in prod and there it doesn't show up at all,  so with this I'd be inclined to pass the ticket.

Before passing it a question:
Currently in this scenario, the approval editor is read only and fixed in unlimited. Is this because of the dummy app? does every app send a specific amount that it needs to be allowed to use? or at least leave the field editable for the user to change?

This PR:
image

Prod:
image

src/components/tx/ApprovalEditor/hooks/useApprovalInfos.test.ts
schmanu: Good point. I think I had too many context switched to ethers v6. Replaced it with BigNumber!

schmanu:
> Currently in this scenario, the approval editor is read only and fixed in unlimited. Is this because of the dummy app? 

@francovenica yes this is the case. In the test dApp I fixed the amounts to Unlimited. Lower amounts work too though. I once edited them in the dummy App.

> does every app send a specific amount that it needs to be allowed to use? or at least leave the field editable for the user to change?
It would not work to edit these values as this could not be communicated to the dApp. The dApp will only receive a signature in return. The dApp is responsible for the amount that is being sent. It could also send a lower amount than unlimited.

PR #3110, opened 2024-01-15
chore: only allow imports from own feature or exposed field

schmanu:
## What it solves
Currently we can import components / services / hooks from other features that are supposed to only be used withing the feature's package.

## How this PR fixes it
Introduces new lint rules that forbid importing from different features except for components / services / hooks that are re-exported on the top-level of the feature.

## Checklist


.eslintrc.json
```diff
@@ -14,7 +14,18 @@
...
     ],
-    "no-only-tests/no-only-tests": "error"
+    "no-only-tests/no-only-tests": "error",
+    "no-restricted-imports": [
+      "error",
+      {
+        "patterns": [
+          {
+            "group": ["@/features/**/*", "!@/features/*"],
+            "message": "You can not import from other features!"
```
katspaugh, line 24:
```suggestion
            "message": "You cannot import from other features"
```

katspaugh (commented):
Really cool!
Is there a way to have only one place where this rule is defined?

schmanu:
> Is there a way to have only one place where this rule is defined?

I did not find a way quickly but I also did not invest that much time so far. 

I think currently it is likely also possible to circumvent the rules by importing with a relative path e.g. if in `/features/walletconnect/test` we could probably do `import * from ../../recovery/internalRecoveryPackage`

katspaugh:
I mean, it's OK to copy this file around as long as we don't forget to do it for new features.

katspaugh:
Shall we close or update this PR?

PR #3121, opened 2024-01-18
fix: only rehydrate if social login is enabled

### LINKED ISSUE #3109: Social login reconnects on the wrong chain
## How to reproduce
* Connect via the Social login on Gnosis Chain
* Switch to mainnet
* Refresh the page
* It will try to reconnect even tho Social login isn't enabled on mainnet

Screenshot 2024-01-15 at 14 34 23

schmanu:
## What it solves

Resolves #3109 

## How to test it
- Open a unsupported Social login chain and observe that it does not try to rehydrate the connection

## Checklist

### DIFF
--- a/src/hooks/wallets/mpc/useRehydrateSocialWallet.ts
+++ b/src/hooks/wallets/mpc/useRehydrateSocialWallet.ts
@@ -65,7 +65,6 @@ const useRehydrateSocialWallet = () => {
     }
     const isOnboardModuleEnabled = isSocialWalletEnabled(chain)
     if (isSocialLoginEnabled && isOnboardModuleEnabled) {
-      console.log('REHYDRATIN')
       void rehydrate()
     }
   }, [chain, onboard, updateAddressBook, isSocialLoginEnabled])
--- a/src/hooks/wallets/mpc/useRehydrateSocialWallet.ts
+++ b/src/hooks/wallets/mpc/useRehydrateSocialWallet.ts
@@ -1,6 +1,6 @@
 import useAddressBook from '@/hooks/useAddressBook'
 import useChainId from '@/hooks/useChainId'
-import { useCurrentChain } from '@/hooks/useChains'
+import { useCurrentChain, useHasFeature } from '@/hooks/useChains'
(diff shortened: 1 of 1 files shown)


src/hooks/wallets/mpc/useRehydrateSocialWallet.ts
```diff
@@ -60,9 +63,12 @@ const useRehydrateSocialWallet = () => {
 
       socialWalletService.setOnConnect(onConnect)
     }
-
-    void rehydrate()
-  }, [chain, onboard, updateAddressBook])
+    const isOnboardModuleEnabled = isSocialWalletEnabled(chain)
+    if (isSocialLoginEnabled && isOnboardModuleEnabled) {
+      console.log('REHYDRATIN')
```
usame-algan, line 68: 🥷

usame-algan (approved):

src/hooks/wallets/mpc/useRehydrateSocialWallet.ts
schmanu: Whoops :D

francovenica:
Social login doesn't work in this environment. Can you enable it?

image

schmanu:
Hmm sadly that is not possible.
You can test this regardless though. This PR is about the app sending requests to `tor.us` when being connected on a chain that doesnt support social login. Those requests should not appear anymore.

francovenica:
"does not try to rehydrate the connection"

I'd assume a reconnection would be tried if I was connected in the first place. If a re connection is taking place when I'm not even connected then I think that would be an issue even in networks that can use social login.

Also I'd like to see that it DOES retry the connection in networks where the social login works.

francovenica:
Previous comment aside. The error shown in the original ticket doesn't show anymore when chainging networks.

I'll give it a second look in dev where the social login is enabled

schmanu:
> I'd assume a reconnection would be tried if I was connected in the first place.

The reconnection is tried if there is a session key in the local storage. So yeah you must have been connected once.. You could run it locally and check with localhost or you check it in dev. 
I checked it when before I pushed it in localhost.

PR #3123, opened 2024-01-18
Fix: remove recommended injected wallet in web3-onboard

### LINKED ISSUE #3116: Unsupported wallet error even though it's not true.
## Bug description

I've been getting the error `Safe{Wallet} only supports [MetaMask](https://metamask.io/) on this platform. Please use or install one of the supported wallets to continue` quite regularly even though I'm on e.g. Eth mainnet and other networks where we clearly support more than 1 wallet.

When afterwards clicking "connect wallet" I see the full suite of options.

## Environment
 - Browser: Chrome
 - Wallet: Rabby, Zeal
 - Chain: Ethereum mainnet

## Steps to reproduce

I was struggling to reproduce this. I think there's more to it, but here's how I can reproduce this on my end.

 1. Have Rabby and Zeal wallet installed and unlocked.
 2. Open a Safe , e.g. https://app.safe.global/home?safe=eth:0xA063Cb7CFd8E57c30c788A0572CBbf2129ae56B6

## Expected result

No error. Either a connected wallet or no wallet connection.

## Obtained result

Error popup `Safe{Wallet} only supports [MetaMask](https://metamask.io/) on this platform. Please use or install one of the supported wallets to continue` 

## Screenshots

image

katspaugh:
## What it solves

Resolves #3116

## How this PR fixes it

Removes the `recommendedInjectedWallets` option in the onboard config. Not sure why it was there in the first place. It seems to work the same sans the annoying warning.

The docs say:

>  if your app only supports injected wallets and when no injected wallets detected, recommend the user to install some

This is clearly not our case.

## Mobile wallets

I've also removed auto-connection ot WalletConnect on mobile because it was briefly flashing the same warning.

### DIFF
--- a/src/hooks/wallets/__tests__/useOnboard.test.ts
+++ b/src/hooks/wallets/__tests__/useOnboard.test.ts
@@ -6,7 +6,6 @@ import { getConnectedWallet, switchWallet } from '../useOnboard'
 // mock wallets
 jest.mock('@/hooks/wallets/wallets', () => ({
   getDefaultWallets: jest.fn(() => []),
-  getRecommendedInjectedWallets: jest.fn(() => []),
 }))
 
 describe('useOnboard', () => {
--- a/src/hooks/wallets/useOnboard.ts
+++ b/src/hooks/wallets/useOnboard.ts
@@ -105,12 +105,6 @@ const trackWalletType = (wallet: ConnectedWallet) => {
     .catch(() => null)
 }
 
-// Detect mobile devices
-const isMobile = () => /iPhone|iPad|iPod|Android/i.test(navigator.userAgent)
-
-// Detect injected wallet
-const hasInjectedWallet = () => typeof window !== 'undefined' && !!window?.ethereum
-
 let isConnecting = false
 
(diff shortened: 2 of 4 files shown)


usame-algan (approved):
I was able to verify on mobile that it doesn't open wallet connect anymore but instead shows wallet connect and other wallet options. For some reason it doesn't detect MM even though I have the app installed so as a user I have to press wallet connect now and select MM from there.

I was not able to see the error message even on prod but I only have Rabby wallet not Zeal that I can activate.

katspaugh:
@usame-algan on desktop it flashes very very briefly.

francovenica:
@katspaugh I still see the modal flashing for a bit if you are already connected to WC until it connects itself. That's the flashing you are referring to?
This is in the PR
autoconnect

katspaugh:
@francovenica no, when you first connect on mobile. It used to show the warning from your screencast even on first connection before showing the WC wallet selector.

francovenica:
Ok, so is not the same scenario.
Ok, then the ticket is fine, we can merge it into the RC and I'll give it a second look there

PR #3128, opened 2024-01-18
fix: Show more accurate total fee during safe creation

usame-algan:
## What it solves

The correct formula is `(baseFee + maxPrioFee) * gasLimit` but with the switch to BigInt we did `baseFee + (maxPrioFee * gasLimit)` instead which resulted in the total fee being too low

## How to test it

1. Create a safe on sepolia
2. Observe the total fee not showing something like < 0,000001 ETH

## Screenshots
Screenshot 2024-01-18 at 16 20 23

## Checklist

### DIFF
--- a/src/components/new-safe/create/steps/ReviewStep/index.tsx
+++ b/src/components/new-safe/create/steps/ReviewStep/index.tsx
@@ -6,7 +6,7 @@ import lightPalette from '@/components/theme/lightPalette'
 import ChainIndicator from '@/components/common/ChainIndicator'
 import EthHashInfo from '@/components/common/EthHashInfo'
 import { useCurrentChain } from '@/hooks/useChains'
-import useGasPrice from '@/hooks/useGasPrice'
+import useGasPrice, { getTotalFee } from '@/hooks/useGasPrice'
 import { useEstimateSafeCreationGas } from '@/components/new-safe/create/useEstimateSafeCreationGas'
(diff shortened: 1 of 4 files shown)


src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -131,9 +131,8 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
   const totalFee =
     gasLimit && maxFeePerGas
       ? formatVisualAmount(
-          maxFeePerGas +
-            // maxPriorityFeePerGas is undefined if EIP-1559 disabled
-            (maxPriorityFeePerGas || 0n) * gasLimit,
+          // maxPriorityFeePerGas is undefined if EIP-1559 disabled
+          (maxFeePerGas + (maxPriorityFeePerGas || 0n)) * gasLimit,
```
schmanu, line 135: Let's add a small unit test that would have picked this up.

usame-algan: Extracted it into a helper function and wrote some tests. Also notice we do the same calculation for tx creation but don't take the maxPrioFee into account there. I added a TODO because the expected type is still BigNumber which wouldn't work with the new helper function and the usage of `Partial` somehow returns `any` as a type which is another issue.

PR #3131, opened 2024-01-19
Refactor: animated sidebar only on a Safe App page

katspaugh:
## What it solves

The current sidebar logic evolved over time and became messy, so I've cleaned it up and also fixed that the sidebar was animating between routes like Welcome and Dashboard (which was visually annoying).

* We don't show a navigation  sidebar on pages like Welcome, Create Safe, Terms, Privacy etc
* We show the navigation sidebar only on routes that have a `?safe` query param
* On small screen, and on the individual Safe App page, we collapse the sidebar, and allow expanding it. And it's animated.


jmealy (approved):

francovenica:
LGTM

The sidebar still works fine. 
Looks good in a mobile viewport as well.

cypress/e2e/safe-apps/permissions_settings.cy.js
```diff
@@ -45,7 +45,10 @@ describe('Permissions settings tests', () => {
       )
     })
 
-    cy.visit(constants.SEPOLIA_TEST_SAFE_5 + constants.appSettingsUrl, { failOnStatusCode: false })
+    cy.visit(`${constants.appsUrl}?safe=${constants.SEPOLIA_TEST_SAFE_5}`, {
```
mike10ca, line 48: There should be` constants.appSettingsUrl` used instead of `constants.appsUrl `to navigate to the correct url. Please update it, this will fix the tests @katspaugh

PR #3132, opened 2024-01-21
Fixed the Spacing issue between image and NFT

Deepak2030:
## What it solves
(UI)NFT image and info are too close on the New Transaction page

## How this PR fixes it
Added margin-left property to the parent container to provide spacing between the image and data

## How to test it
Just a UI change, you can run the code on your device and test

## Checklist

### DIFF
--- a/src/components/nfts/NftGrid/index.tsx
+++ b/src/components/nfts/NftGrid/index.tsx
@@ -205,7 +205,7 @@ const NftGrid = ({
                     <Box display="flex" alignItems="center" gap={2}>
                       {item.imageUri ? activeNftIcon : inactiveNftIcon}
 
-                      <div>
+                      <div style={{ marginLeft: '20px' }}>
                         <Typography>{item.tokenName || item.tokenSymbol}</Typography>
 
                         <Typography fontSize="small" color="text.secondary" component="div">


katspaugh:
That's not the fix I expected.

In the happy case, previews already have the correct spacing:
Screenshot 2024-01-22 at 10 16 13

However, if an image is rectangular, i.e. wider than it's height, it will overflow:


Screenshot 2024-01-22 at 10 17 43

☝️ the same ENS logo but I set its width to 100px.

So it's a problem with the flexbox/grid in that list, not with the margins.

Deepak2030:
Hey @katspaugh sorry for that, I was having trouble while running this project on my local device, the connect wallet button was not working on the localhost and also the NFTs stored on my metamask were not displayed in the deployed app. Are there any solutions to that so that I can make my contribution valuable?

katspaugh:
You might need to set an `NEXT_PUBLIC_INFURA_TOKEN` variable in your `.env` file. You can get the token from Infura's site.

Deepak2030:
image

Hey, I still can't see my NFTs on the Assets Page.

katspaugh:
Why would you see your NFTs from another account (loaded in MetaMask) in your Safe Wallet account?

PR #3135, opened 2024-01-22
fix: UI not updating if signing from walletconnect

compojoom:
If another safe is owner of a safe and we use walletconnect to sign transaction the UI was not updating. It turns out we need to port the  UncheckedJsonRpcSigner from ethers v5.

The UncheckedJsonRpcSigner fakes a transaction receipt without actually having one, and this is needed for transactions that don’t actually end on-chain.

## What it solves
Signing with a safe should close the transaction popup in the child's safe.

Resolves #

## How this PR fixes it
Test cases:
1. Create a transaction in C and sign in P
2. Create a transaction in C and immediately execute in P

In both cases the UI flows close and the transaction in C only appears in the queue after the transaction in P has been executed. 

3. Create and immediately execute transaction in C and sign in P
4. Create and immediately execute transaction in C and immediately execute in P

In case 4 there will be a timeout error if P takes longer than a minute to execute which results in the processing transaction of C being deleted from the queue. It later appears in the history once the transaction in P has executed.

## Checklist

### DIFF
--- a/src/features/recovery/services/recovery-sender.ts
+++ b/src/features/recovery/services/recovery-sender.ts
@@ -11,6 +11,7 @@ import { recoveryDispatch, RecoveryEvent, RecoveryTxType } from './recoveryEvent
 import { asError } from '@/services/exceptions/utils'
 import { assertWalletChain } from '../../../services/tx/tx-sender/sdk'
 import { isSmartContractWallet } from '@/utils/wallets'
+import { UncheckedJsonRpcSigner } from '@/utils/providers/UncheckedJsonRpcSigner'
 
 async function getDelayModifierContract({
   onboard,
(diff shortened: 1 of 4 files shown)


usame-algan:
There are some lint issues but apart from that the code looks good.

I tested two cases so far and can confirm that they work as expected:

1. Create a transaction in C and sign in P
2. Create a transaction in C and immediately execute in P

In both cases the UI flows close and the transaction in C only appears in the queue after the transaction in P has been executed. There are (at least) two more cases that need testing:

1. Create and immediately execute transaction in C and sign in P
2. Create and immediately execute transaction in C and immediately execute in P

usame-algan:
There was also a `provider.getUncheckedSigner` call in `getDelayModifierContract` that needs to be restored.

src/features/recovery/services/recovery-sender.ts
```diff
@@ -28,7 +29,9 @@ async function getDelayModifierContract({
   const isSmartContract = await isSmartContractWallet(wallet.chainId, wallet.address)
 
   // Use unchecked signer for smart contract wallets as transactions do not necessarily immediately execute
-  const signer = await provider.getSigner()
+  const signer = isSmartContract
+    ? new UncheckedJsonRpcSigner(provider, await (await provider.getSigner()).getAddress())
+    : await provider.getSigner()
```
usame-algan, line 34: If we need to call `await provider.getSigner()` in both cases I suggest to extract it from the shorthand condition.

usame-algan (approved):
I've checked the remaining cases:

1. Execute immediately in C, execute immediately in P
2. Execute immediately in C, sign in P

Both cases work as expected. In the second case there will be a timeout error if P takes longer than a minute to execute which results in the processing transaction of C being deleted from the queue. It later appears in the history once the transaction in P has executed.

Please document these cases in the PR description for QA to double check.

PR #3138, opened 2024-01-22
fix: show zero wallet balance

schmanu:
## What it solves

Resolves https://www.notion.so/safe-global/Wallet-balance-never-resolves-when-trying-to-submit-bulk-transactions-with-walletconnect-6110e4bcd28444b083315bd21a10100d?d=71a60b3cbfd64773ab148328894330cf

## How this PR fixes it
Checks that the balance is undefined instead of coalescing.

## How to test it
Connect a wallet with 0 balance and look at the wallet balance in the send transaction screen.

## Checklist


compojoom:
Good catch. Tested it and it works!

PR #3142, opened 2024-01-23
Fix: onboard initialization with ethers v6

schmanu:
## What it solves
Switching to some chains threw an error that onboard does not know the chainId.

Resolves https://www.notion.so/safe-global/b2bf1616be274eba8e02dedcfd272d1e?v=a922ede2ebd14ba3b629d2e36d43e3f6&p=ac04ce127afd4987923caa3fb3b483bd&pm=s

## How this PR fixes it
When initializing onboard with all chainIds, we do not pad the chainId to bytes.

## How to test it
- Connect a wallet to Sepolia
- Open a Safe on görli
- Try to sign a tx where a chain switch is required first

## Checklist

### DIFF
--- a/src/services/onboard.ts
+++ b/src/services/onboard.ts
@@ -1,6 +1,5 @@
 import Onboard, { type OnboardAPI } from '@web3-onboard/core'
 import type { ChainInfo } from '@safe-global/safe-gateway-typescript-sdk'
-import { toBeHex } from 'ethers'
 import { getAllWallets, getRecommendedInjectedWallets } from '@/hooks/wallets/wallets'
 import { getRpcServiceUrl } from '@/hooks/wallets/web3'
 import type { EnvState } from '@/store/settingsSlice'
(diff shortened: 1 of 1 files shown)


src/services/onboard.ts
```diff
@@ -17,7 +16,8 @@ export const createOnboard = (
   const wallets = getAllWallets(currentChain)
 
   const chains = chainConfigs.map((cfg) => ({
-    id: toBeHex(parseInt(cfg.chainId)),
+    // We cannot use ethers' toBeHex here as we do not want to pad it to an even number of characters.
+    id: `0x${parseInt(cfg.chainId).toString(16)}`,
```
katspaugh, line 20: There's a `numberToHex` ustil in `src/utils/hex.ts`

compojoom (approved):
just a note: apparently toQuantity from ethers v6 makes the same thing as numberToHex.

katspaugh (approved):

PR #3143, opened 2024-01-23
fix: message signing

compojoom:
## What it solves
resolves: https://www.notion.so/safe-global/Can-not-sign-message-work-on-prod-70a71bf06be54c3fbf2a0e05cf19d657

Resolves #

## How this PR fixes it

ethers v6 converts the chainId to a hex value: https://github.com/ethers-io/ethers.js/blob/50b74b8806ef2064f2764b09f89c7ac75fda3a3c/src.ts/hash/typed-data.ts#L75
Our SDK expects a number, that's why we convert the hex value from ethers v6 to number.

## Checklist


schmanu (approved):
LGTM!

Issue #3153, opened 2024-01-25
Black Screen Issue When Using WalletConnect on Safe.Lightlink.io


sleepyqadir:
## Bug description

When interacting with the Safe Wallet App via https://safe.lightlink.io/, specifically using WalletConnect to connect an external application, the screen turns black after a glitch instead of displaying the WalletConnect modal for entering the pairing code. Additionally, a console error "[Spindl SDK] request failed with http code: 400" is observed.

## Environment

 - URL: https://safe.lightlink.io/
 - Browser: Chrome
 - Wallet: WalletConnect
 - Chain: Lightlink Pegasus

## Steps to reproduce

  1. Go to https://safe.lightlink.io/.
  2. Navigate to 'Apps' and select 'Use Custom WalletConnect'.
  3. Attempt to use WalletConnect to connect an external application to the Safe Wallet.
  4. Observe the screen turning black and the console error.

## Expected result

A modal should appear, prompting to "Paste the pairing code below to connect to your Safe{Wallet} via WalletConnect".

## Obtained result

  - The screen turns black after a momentary glitch.
  - A console error is displayed: [Spindl SDK] request failed with http code: 400.
  - Network request fails with:
      - Request URL: https://spindl.link/events
      - Request Method: POST
      - Status Code: 400 Bad Request
      - Remote Address: 146.70.46.34:443
      - Referrer Policy: strict-origin-when-cross-origin
  
## Screenshots

![Screenshot 2024-01-25 at 2 17 17 PM](https://github.com/BlockApex/Elektrik-V2-Contracts/assets/38910854/1454a878-8466-4027-a12c-4aeaffead6f0)

![Screenshot 2024-01-25 at 2 16 26 PM](https://github.com/BlockApex/Elektrik-V2-Contracts/assets/38910854/3ad13ce7-82c3-4bb9-82b7-4228203b42b4)

![Screenshot 2024-01-25 at 2 18 34 PM](https://github.com/BlockApex/Elektrik-V2-Contracts/assets/38910854/b313f0ea-e65b-4680-b574-91c67dd27a6a)

katspaugh:
Thanks! I've fixed the spindl 404, although it's unrelated to the back screen.

The WalletConnect app gets a special treatment  because of the native WalletConnect feature we now have (in the header). The simplest solution would be to rename your custom Safe App URL to somethign not containing `wallet-connect` (because that's how we're matching it).

PR #3155, opened 2024-01-25
Fix: a more accurate WC app detection

katspaugh:
## What it solves

Reported in #3153

The WalletConnect Safe app was detected by a URL substring which might give false positives for custom apps.

## How this PR fixes it
Makes it check the URL 1-to-1 with the WC Safe App's URL (it's the same on prod and staging unlike the app id).

### DIFF
--- a/src/utils/gateway.ts
+++ b/src/utils/gateway.ts
@@ -28,6 +28,6 @@ export const getExplorerLink = (
 }
 
 export const isWalletConnectSafeApp = (url: string): boolean => {
-  const WALLET_CONNECT = /wallet-connect/
-  return WALLET_CONNECT.test(url)
+  const WC_APP_URL = 'https://apps-portal.safe.global/wallet-connect'
+  return url === WC_APP_URL
 }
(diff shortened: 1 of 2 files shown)


src/utils/gateway.ts
```diff
@@ -28,6 +28,6 @@ export const getExplorerLink = (
 }
 
 export const isWalletConnectSafeApp = (url: string): boolean => {
-  const WALLET_CONNECT = /wallet-connect/
-  return WALLET_CONNECT.test(url)
+  const WC_APP_URL = 'https://apps-portal.safe.global/wallet-connect'
```
usame-algan, line 31: Do we have to account for trailing slashes here?

katspaugh: No, that's the precise URL of this safe app.

usame-algan (approved):

francovenica:
I was able to reproduce the issue in https://safe.lightlink.io/ (i created a safe there), but since I cannot reproduce the issue in our dev or prod env I cannot be sure that this fix will work on their side.
I just checked that WC still works fine connecting to 3rd party apps just fine, so at least I can say that the fix is not braking anything else, but the test will have to be the people of lightlink to check on their safe app once they can update it with this fix.

PR #3158, opened 2024-01-26
Chore: update WalletConnect SDK

katspaugh:
## What it solves

This should fix some of the "no matching keys" issues we've been seeing.

I've tested it with 1inch and Uniswap on mainnet and Polygon. ✅


schmanu (approved):
🚀

PR #3165, opened 2024-01-29
Refactor: remove QR scanner in address inputs

katspaugh:
## What it solves

I've added tracking for successful QR code scans in the AddressInput in #3018.

After 3 weeks of tracking this event, we can see that only 0.2% of users complete a scan after clicking on that button (and only 1.4% click on it in the first place).

Screenshot 2024-01-29 at 19 05 11

With this in mind, it makes sense to remove this feature along with the NPM dependency it comes with.


katspaugh:
Todo: remove the package from the Licenses page.

usame-algan (approved):

usame-algan:
@mike10ca The Scan QR is removed with this PR and we also have to adjust affected e2e tests.

liliya-soroka:
Verified

PR #3176, opened 2024-01-31
feat: Add `/welcome/login` route for displaying safes list

jmealy:
Redesign the welcome page according to the designs: https://www.figma.com/file/WSpulDIM1EoSCI6Ph9QdWq/Seedless-onboarding?type=design&node-id=1047-12423&mode=design&t=MSQvs5eVjdMOLVzr-0

The new design is added to a new route at /welcome/login. Currently it doesn't show acual owned safes. Implementation of getting owned safes will be done later using the new endpoint getAllOwnedSafes: https://github.com/safe-global/safe-gateway-typescript-sdk/pull/144

## Checklist


katspaugh:
There are CI jobs failing.

As we talked about a few times, this PR should not contain any changes to the Redux stores. Only the HTML/CSS for the new page.

src/components/common/OwnedSafeList/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+      {isLoading && (
+        <Box py={4} textAlign="center">
+          <CircularProgress size={20} />
+        </Box>
+      )}
+
+      {!safes.length && (
+        <Box display="flex" flexDirection="column" py={4} sx={{ maxWidth: '250px', margin: 'auto' }}>
+          <Typography variant="body2" color="primary.light" textAlign="center" mt={2} mb={2}>
+            No safe accounts yet
```
katspaugh, line 64:
```suggestion
            No Safe Accounts yet
```

katspaugh (commented):
There's a sidebar flashing when opening the login page, please add it to the NO_SIDEBAR_ROUTES list.

src/hooks/useSafes.ts
```diff
@@ -0,0 +1,59 @@
...
+        const watchedSafesOnChain = watchedSafes[chain.chainId] ?? {}
+        const watchedSafesAdressesOnChain = Object.keys(watchedSafesOnChain)
+        const watchedSafesWithChain = watchedSafesAdressesOnChain.map((safeAddress) => {
+          const { threshold, owners } = watchedSafesOnChain[safeAddress]
+          return { safeAddress, chain, threshold, owners }
+        })
+        watchedSafesOnAllChains = [...watchedSafesOnAllChains, ...watchedSafesWithChain]
+      }
+      const promises = watchedSafesOnAllChains.map(async ({ safeAddress, chain, threshold, owners }) => {
+        const fiatBalance = await getBalances(chain.chainId, safeAddress, 'USD').then((result) => result.fiatTotal)
```
katspaugh, line 44: USD shouldn't be hardcoded. There's a redux selector for the user-preferred currency.

katspaugh: Overall, I wouldn't fetch balances in this hook. They essentially make the entire list wait. I would instead fetch the balances separately and then pass them separately to the corresponding components.

katspaugh:
This seems to be not centered vertically, plus didn't the design have a placeholder element in this block?

Screenshot 2024-02-06 at 12 46 03

src/hooks/useSafes.ts
jmealy: yeah that makes sense. I moved the balance request to SafeListItem

PR #3180, opened 2024-01-31
[Counterfactual] Safe creation

### LINKED ISSUE #3179: Adjust creation flow for 1/1 safes and add fallbacks to CGW calls

usame-algan:
## What it solves

Resolves #3179 

## How this PR fixes it

- Adjust safe creation flow to create counterfactual safes for 1/1 setups
- Adds fallbacks for CGW calls so the UI is unblocked
- Fetches native balance
- Adds feature flag
- Uses an incremental saltNonce for all safe deployments

## How to test it

1. Connect your wallet and create a 1/1 safe
2. Observe no cost on the review screen
3. Press Create
4. The dashboard should be immediately visible
5. There should be no errors
6. Send ETH to the safe address
7. Observe it being visible on the assets page
8. Create another 1/1 safe
9. Observe the same safe address

## Checklist


src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -151,6 +166,41 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
       saltNonce: saltNonce.toString(),
     }
 
+    if (isCounterfactual) {
+      const counterfactualNonce = await getCounterfactualNonce(provider, { ...props, saltNonce: '0' })
```
katspaugh, line 170: If we're switching to an incremental nonce, I would do it for all deployments, not just counterfactual. For the benefitsts that Micah outlined and for consistency.

src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -133,6 +145,9 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
       ? formatVisualAmount(getTotalFee(maxFeePerGas, maxPriorityFeePerGas, gasLimit), chain?.nativeCurrency.decimals)
       : '> 0.001'
 
+  // Only 1 out of 1 safe setups are supported for now
+  const isCounterfactual = data.threshold === 1 && data.owners.length === 1 && isCounterfactualEnabled
```
katspaugh, line 149: Please try to extract all the logic into a service and only do UI rendering in components.

src/components/new-safe/create/steps/ReviewStep/index.tsx
usame-algan: It would require us to check on-chain if a contract is already deployed to that address for every safe that is created but I think it would simplify our code so I will look into refactoring this part of safe creation a bit more.

MissLov3ly:
L5gS82oQh3rDHJug23XYxufELBUx5Ga8F9Aat8qZJfXEKdWQgtTC

src/components/new-safe/create/logic/utils.test.ts
```diff
@@ -0,0 +1,44 @@
...
+  jest.spyOn(creationUtils, 'computeNewSafeAddress').mockReturnValue(Promise.resolve(faker.finance.ethereumAddress()))
+
+  const mockProvider = new BrowserProvider(jest.fn() as unknown as Eip1193Provider)
+  const mockDeployProps = {
+    safeAccountConfig: {
+      threshold: 1,
+      owners: [faker.finance.ethereumAddress()],
+      fallbackHandler: faker.finance.ethereumAddress(),
+    },
+  }
```
schmanu, line 17: Should we put this in a `beforeAll()` block?

src/components/new-safe/create/logic/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { computeNewSafeAddress } from '@/components/new-safe/create/logic/index'
+import { isSmartContract } from '@/hooks/wallets/web3'
+import type { DeploySafeProps } from '@safe-global/protocol-kit'
+import type { BrowserProvider } from 'ethers'
+
+export const getAvailableSaltNonce = async (provider: BrowserProvider, props: DeploySafeProps): Promise<string> => {
+  const safeAddress = await computeNewSafeAddress(provider, props)
+  const isContractDeployed = await isSmartContract(provider, safeAddress)
```
schmanu, line 8: Could we not run into Rate Limit issues here? If someone deployed e.g. 10-20 Safes already?

src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -203,46 +218,52 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
...
-                    You will have to confirm a transaction with your connected wallet.
-                  </Typography>
-                )}
-              </>
-            }
-          />
-        </Grid>
+            <Grid data-testid="network-fee-section" container spacing={3}>
+              <ReviewRow
+                name="Est. network fee"
```
schmanu, line 242: Don't we want to also display the network fee if the users deploys later?
Otherwise users might transfer funds into their Safe and only then learn how much the deployment actually costs.

src/features/counterfactual/store/undeployedSafeSlice.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { PredictedSafeProps } from '@safe-global/protocol-kit'
+
+type UndeployedSafeSlice = { [address: string]: PredictedSafeProps }
+
+type UndeployedSafeState = { [chainId: string]: UndeployedSafeSlice }
+
+const initialState: UndeployedSafeState = {}
+
+export const undeployedSafeSlice = createSlice({
+  name: 'undeployedSafe',
```
schmanu, line 12:
```suggestion
  name: 'undeployedSafes',
```

Our other slices are named in plural

src/components/new-safe/create/logic/utils.test.ts
usame-algan: From what I found online, if the data is synchronous its ok to leave it in the `describe` block but otherwise it should be in `beforeAll`. Only downside of using `beforeAll` is that we have to initialize the variables beforehand and add explicit types i.e. `let mockProvider: BrowserProvider` etc.

src/components/new-safe/create/steps/ReviewStep/index.tsx
usame-algan: As part of https://github.com/safe-global/safe-wallet-web/issues/3210 we will add a Pay now / Pay later option with some educational text.

src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -117,9 +124,9 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
     return {
       owners: data.owners.map((owner) => owner.address),
       threshold: data.threshold,
-      saltNonce,
+      saltNonce: Date.now(), // Doesn't matter for the gas estimation
```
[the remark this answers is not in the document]
schmanu: I double checked:
> A byte of calldata costs either 4 gas (if it is zero) or 16 gas (if it is any other value).

src/components/new-safe/create/logic/utils.test.ts
schmanu: Okay. Let's leave it as is then :)

src/components/new-safe/create/steps/ReviewStep/index.tsx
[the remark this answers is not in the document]
usame-algan: Thanks for checking! In that case I will move the saltNonce calculation into a hook so we can reuse it here.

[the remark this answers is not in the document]
schmanu: I think with Date.now() we would get slightly too high gas estimations as the current Date likely contains more zeros then the number incremented by one each time.

So I guess we could also leave it for now. And change the comment slightly

src/components/new-safe/create/logic/utils.ts
usame-algan: I haven't had any issues so far with around 10 safes created but we should handle possible error responses from the RPC call. In case this becomes an issue with users we could still optimize this e.g. by not incrementally checking saltNonces but using something like binary search.

src/features/counterfactual/store/undeployedSafeSlice.ts
usame-algan: done

src/components/new-safe/create/steps/ReviewStep/index.tsx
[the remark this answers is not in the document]
usame-algan: done

src/components/new-safe/create/logic/utils.ts
usame-algan: I don't think this will realistically happen like what is the use-case for a wallet to create the exact same safe setup 30 times? I agree though that we should try to limit test it cc. @francovenica

usame-algan: We could also add a fallback to the readonly provider in case the wallet provider fails to increase the chances of success.

schmanu (approved):

francovenica:
I've checked that the safe is created properly
Transferring tokens to the safe are shown in the assets tab (not in the sidebar, but that's expected for now)
Re-creating the safe gives you the same address. Even clearing the storage creates the same address, which is expected

LGTM

PR #3182, opened 2024-01-31
Point to development

anxolin:
## What it solves
The current version of CoW Swap in prod doesn't work with the PoC in #feat-cow-swap branch. 

This PR changes the environment to `dev` which means, the iframe use https://dev.swap.cow.fi/ instead of https://swap.cow.fi

We just applied a fix in DEV to disable bundling of transactions if CoW Swap doesn't manage to get an instance of the SWAP, this will solve the issue.

We should bring back bundling of transactions, so this will be a temporal fix.


anxolin:

PR #3185, opened 2024-02-01
[Counterfactual] Deploy safe with first transaction

### LINKED ISSUE #3184: [Counterfactual] Adjust first transaction flow
## Scope

- Deploy the safe with the first transaction
- Show gas estimation
- Clear counterfactual state after deployment

## Out of scope

- Design changes

usame-algan:
## What it solves

Resolves #3184 

## How this PR fixes it

- Deploys safe with first transaction
- Shows gas estimation for the bundle
- Clear counterfactual state after deployment
- Hide nonce in all transaction flows

## ToDos

- [x] Adjust AppFrame to handle incoming requests for counterfactual safes

## How to test it

1. Create a counterfactual 1/1 safe
2. Send some funds to it
3. Create a transaction
4. Gas estimation should be displayed
5. Execute the transaction
6. Observe you have to sign first
7. Observe the safe is deployed after execution
8. Observe the tx history is visible after execution

## Checklist


src/features/counterfactual/hooks/useDeployGasLimit.ts
```diff
@@ -0,0 +1,27 @@
...
+  const chainId = useChainId()
+
+  const [gasLimit, gasLimitError, gasLimitLoading] = useAsync<bigint | undefined>(async () => {
+    if (!safeTx || !onboard) return
+
+    const sdk = await getSafeSDKWithSigner(onboard, chainId)
+
+    const gas = await estimateTxBaseGas(sdk, safeTx)
+    const safeTxGas = await estimateSafeTxGas(sdk, safeTx)
+    const safeDeploymentGas = await estimateSafeDeploymentGas(sdk)
```
katspaugh, line 19: These things don't depend on each other so they could be done in parallel.

src/features/counterfactual/utils.ts
```diff
@@ -35,3 +41,46 @@ export const getUndeployedSafeInfo = (undeployedSafe: PredictedSafeProps, addres
...
+    const wallet = await assertWalletChain(onboard, chainId)
+
+    const signedTx = await sdkUnchecked.signTransaction(safeTx)
+    const deploymentTx = await sdkUnchecked.wrapSafeTransactionIntoDeploymentBatch(signedTx, txOptions)
+
+    const provider = createWeb3(wallet.provider)
+    const signer = await provider.getSigner()
+
+    // @ts-ignore TODO: Check if the other type also works
+    result = await signer.sendTransaction(deploymentTx)
```
katspaugh, line 64: Some of these calls could be parallelized as well, not sure how fast it is already.

src/features/counterfactual/hooks/useDeployGasLimit.ts
usame-algan: Good idea!

src/features/counterfactual/utils.ts
usame-algan: We could run the `wallet` and `deploymentTx` call in parallel but I think its already fast currently

katspaugh (approved):
👍

francovenica:
LGTM

I tried creating a CF safe, transfer funds to it and then sending them.
After the execution, in the history tab, you can see the safe creation tx and the send funds tx as well.

PR #3189, opened 2024-02-01
[Analytics] origin as label for safe app txs

katspaugh:
## What it solves

Safe Apps tx events will now have the Safe App URL as their label.

## How to test it

1. Make a transaction via any Safe App
2. See in the console that the GTM events tx_created, tx_confirmed and tx_executed have the Safe App's URL as the `eventLabel` property.

### DIFF
--- a/src/services/analytics/__tests__/tx-tracking.test.ts
+++ b/src/services/analytics/__tests__/tx-tracking.test.ts
@@ -161,7 +161,7 @@ describe('getTransactionTrackingType', () => {
       },
     } as unknown)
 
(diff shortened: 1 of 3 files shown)


alicecha (approved):

francovenica:
Looks good:
The event tx_created and tx_executed are triggered for safes with a policy of 1. To see the tx_confirmed the tx must be in queue and it has to be only signed, so you won't see that event in safes 1/x.
This is not an issue, just documenting when the events are triggered.

image
image

PR #3199, opened 2024-02-05
[Counterfactual] Add deploy safe flow

### LINKED ISSUE #3195: [Counterfactual] Add deploy safe flow
## What is this about

Add a new flow to deploy a safe after safe creation

## Requirements

- [ ] New transaction flow to deploy a safe
- [ ] Make it accessible via the settings page

usame-algan:
## What it solves

Resolves #3195 

## How this PR fixes it

- Adds a new flow for deploying a safe
- Adds a new section on the settings page that is only visible for counterfactual safes

## How to test it

2. Navigate to the settings page
3. Observe a new section for deploying the safe
4. Press on Activate now
5. Observe a new safe deployment flow
6. Execute the transaction
7. Observe the settings section is not visible anymore
8. Observe the safe is deployed on-chain

## Screenshots
Screenshot 2024-02-05 at 10 00 11
Screenshot 2024-02-05 at 10 00 24

## Checklist


src/features/counterfactual/ActivateAccount.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  if (safe.deployed) return null
+
+  const activateAccount = () => {
+    setTxFlow(<ActivateAccountFlow />)
+  }
+
+  return (
+    <Alert severity="info" sx={{ mb: 3 }}>
+      <Typography fontWeight="bold">Want to skip onboarding?</Typography>
```
katspaugh, line 175: This copy makes little sense. The onboarding has already ended if the user is on this page.
```suggestion
      <Typography fontWeight="bold">Activate your account?</Typography>
```

src/features/counterfactual/ActivateAccount.tsx
```diff
@@ -0,0 +1,186 @@
...
+      dispatch(removeUndeployedSafe({ chainId, address: safeAddress }))
+    } catch (_err) {
+      const err = asError(_err)
+      setIsSubmittable(true)
+      setSubmitError(err)
+      return
+    }
+
+    setTxFlow(undefined)
+  }
```
katspaugh, line 83: Please move some of this into a hook.

katspaugh: It doesn't have to be reusable, just literally copy and paste everything into a use-function in the same or a separate file.

usame-algan: I extracted the logic for getting the total formatted fee and moved the rest of the logic into a local hook.

katspaugh (approved):

francovenica:
Issue with Relayer. (Let me know if relayer is out of the scope of the ticket or not)

Signing with the relayer shows an error message. The tx still goes through and works fine (the safe is activated), so the error is just the error popping up

image

Try on sepolia where relayer is enabled

src/pages/settings/setup.tsx
```diff
@@ -24,6 +25,8 @@ const Setup: NextPage = () => {
       <SettingsHeader />
 
       <main>
+        <ActivateAccount />
```
schmanu, line 28: Should we only render this if the Safe is undeployed?
It always triggers stuff like gasEstimation etc otherwise

usame-algan:
> Signing with the relayer shows an error message. The tx still goes through and works fine (the safe is activated), so the issue is just the error popping up

Looks like we cleared the the counterfactual state before the transaction was executed by the relayer. I just pushed a fix for this. I think it would be nice to have some kind of status screen when the user executes a transaction (either via relayer or EOA).

src/pages/settings/setup.tsx
usame-algan: I have an early return in that component so it only renders if the safe is not deployed yet.

francovenica:
The issue I reported was fixed.

I have another issue that was not happening when I tested the ticket the first time. 
When you click "create" in the last step of the safe creation an error message shows up. You have to click a second time for the safe to be created
CF safe creation

usame-algan:
> When you click "create" in the last step of the safe creation an error message shows up. You have to click a second time for the safe to be created

I fixed it in https://github.com/safe-global/safe-wallet-web/pull/3180 since it belongs to the safe creation flow

francovenica:
Ok, given that the error is fixed in another ticket we can pass it here.

The rest of the ticket is fine. Once the conflicts are fixed we can merge the ticket

PR #3202, opened 2024-02-05
[Counterfactual] Add backup option

### LINKED ISSUE #3203: Add backup and recovery option

usame-algan:
## What it solves

Resolves #3203

## How this PR fixes it

- Shows a backup option when creating a counterfactual safe
- Falls back to read-only provider to fetch balance in case no wallet is connected and caches the response

## How to test it

2. Observe the creation modal with a new section to download a backup
3. Download the backup
4. Disconnect your wallet
5. Reload the app
6. Observe balances still load once
7. Observe no polling for balances

## Screenshots

Screenshot 2024-02-05 at 12 25 23

## Checklist


src/components/common/SafeLoadingError/index.tsx
```diff
@@ -1,20 +1,72 @@
...
+
   return (
     <PagePlaceholder
       img={<img src="/images/common/error.png" alt="A vault with a red icon in the bottom right corner" />}
       text="This Safe Account couldn't be loaded"
     >
+      {isCounterfactualEnabled && (
+        <>
+          <Button component="label" variant="outlined" size="large" sx={{ mt: 2, mb: 1 }}>
+            Recover your safe setup
```
katspaugh, line 64: How would the user even land on this page if they lost their safe address?

src/components/dashboard/CreationDialog/index.tsx
```diff
@@ -73,6 +102,17 @@ const CreationDialog = () => {
             description="Have any questions? Check out our collection of articles."
           />
         </Grid>
+
+        {undeployedSafe && (
+          <Alert severity="info" sx={{ mb: 2 }}>
+            We recommend to{' '}
```
katspaugh, line 108: `We recommend backing up`

src/components/common/SafeLoadingError/index.tsx
usame-algan: This would only cover the case where the user knows their safe address and has a backup and they tried to open that safe on a new browser or different device.

If they lost their safe address but have a backup there is currently not an option to recover that backup from outside a safe route -> We could add this e.g. in the load safe flow

If they lost their safe address and don't have a backup the only thing they can do is go through the safe creation flow again to create "another" 1/1 safe and if nothing about the proxy version has changed it would result in the same address which they could back up then.

katspaugh: If they have a Safe address isn't it trivial to infer the nonce/Safe version?

src/components/common/SafeLoadingError/index.tsx
```diff
@@ -1,20 +1,72 @@
...
+              variant: 'error',
+              groupKey: 'safe-import-error',
+              message: 'Could not recover safe account',
+            }),
+          )
+        }
+      }
+      reader.readAsText(file)
+    }
+  }
```
katspaugh, line 54: This should be in its own component as it has nothing to do with the Safe Loading Error.

src/components/common/SafeLoadingError/index.tsx
usame-algan: When the user opens a counterfactual safe thats not in local storage and they are connected with a wallet we could predict the safe address at the same time and see if it matches i.e. owner is the connected EOA, threshold is 1 out of 1, safe version is latest version, salt nonce is 0 and if it matches we could load it into localStorage. If it doesn't match we could increase the salt nonce x times and check each time so it would essentially be the same thing we do in the safe creation flow just whenever we show a load safe error.

Only issue I see is that if they just open some safe, chances are low that their wallet will be connected so we wouldn't be able to predict an address there.

If they connect their wallet but don't have a safe open we could predict addresses but there wouldn't be anything to match so that would also not be helpful.

src/components/common/SafeLoadingError/index.tsx
usame-algan: Good idea, I moved it to a new component inside the counterfactual feature directory.

usame-algan:
As discussed, I removed the restore option so we only offer to download a backup for now.

katspaugh (approved):

francovenica:
Question: 
When I click the link to download the backup the page refreshes, it happens every time. Is that unavoidable? I'm using Chrome.

usame-algan:
> When I click the link to download the backup the page refreshes, it happens every time. Is that unavoidable? I'm using Chrome.

I pushed a fix for this so the page should not refresh anymore.

francovenica:
Ok, looks good, the page doesnt' refresh anymore

The backup downloads just fine
I've checked that after disconnecting and refreshing the page the balance is loaded just once and not anymore (until the safe is deployed)

Note: At the point of this ticket there is no way to import the safe back

PR #3204, opened 2024-02-05
Refactor: remove custom ENS resolver for Sepolia

katspaugh:
## What it solves

Ethers v6 has Sepolia ENS support out of the box, so we don't need to provide a custom resolver anymore.

Plus, I've fixed that the Address Input would say "address **or ENS**" even on chains where ENS isn't enabled.

## How to test it

Case 1:
* Enter an ENS name into an address field on Sepolia
* It should be resolved correctly

Case 2:
* Open a Safe on a chain where ENS isn't enabled in the config
* Address inputs' labels should say "Recipient address" and not "Recipient address **or ENS**"


src/services/ens/index.test.ts
```diff
@@ -4,7 +4,7 @@ import { logError } from '../exceptions'
 
 // mock rpcProvider
 const rpcProvider = {
-  resolveName: jest.fn(() => Promise.resolve('0x0000000000000000000000000000000000000000')),
+  resolveName: jest.fn(() => Promise.resolve('0x0000000000000000000000000000000000000001')),
```
compojoom, line 7: why are you changing the address here?

compojoom (approved):
lgtm

francovenica:
Looks good to me

Tested ENS names I have for sepolia and ETH and they resolve just fine

I've checked the removal of ENS in the input name for networks that don't have it.

PR #3208, opened 2024-02-06
fix: do not forward eth_chainId

schmanu:
## What it solves
Every page load and whenever we switch the Safe, the core-sdk triggers 10 eth_chainId requests.

## How this PR fixes it
The ReadOnlyProvider caches the chainId as it cannot change anyways as the readonly provider gets recreated if the chain changes.

## How to test it
- Load a Safe and witness fewer eth_chainId requests.

## Checklist


src/utils/providers/ReadonlyRpcProvider.ts
```diff
@@ -0,0 +1,23 @@
+import { JsonRpcProvider } from 'ethers'
+import { numberToHex } from '../hex'
+
+class ReadonlyRpcProvider extends JsonRpcProvider {
+  #chainId: string
```
katspaugh, line 5: Let's not use the `#` syntax and just mark it as private. It's too far off from regular JS.

compojoom (approved):

PR #3211, opened 2024-02-06
Tests: add sidebar tests -p3

### LINKED ISSUE #3187: Add Cypress tests for sidebar
To support our continuous dedication to test coverage, the following automation tests were added:

1. Added some new tests for side bar
2. Added data-tested to support tests
3. Refactoring:
- Moved out test with signing action to regression test suite
- Added wait time before clicking on view tx notification to handle correct nonce redirection
- Other minor refactoring - wip

mike10ca:
## What it solves

Resolves #3187 

## How this PR fixes it

- Added new tests for sidebar
- Added data-testids to support tests

**New tests:**

- Verify Fiat currency changes when edited in the assets tab
- Verify "wallet" tag counter if the safe has tx ready for execution
- Verify "Wallet" tag counter only shows for owners
- Verify New Transaction button enabled for users with Spending limits allowed
- Verify tag counting queue tx show for owners and non-owners

## How to test it

- Run Cypress tests and observe outcome.

### DIFF
--- a/cypress/e2e/pages/assets.pages.js
+++ b/cypress/e2e/pages/assets.pages.js
@@ -17,6 +17,8 @@ const hiddenTokenSaveBtn = 'span[data-track="assets: Save hide dialog"]'
 const hiddenTokenCancelBtn = 'span[data-track="assets: Cancel hide dialog"]'
 const hiddenTokenDeselectAllBtn = 'span[data-track="assets: Deselect all hide dialog"]'
 const hiddenTokenIcon = 'svg[data-testid="VisibilityOffOutlinedIcon"]'
+const currencySelector = '[data-testid="currency-selector"]'
+const currencyItem = '[data-testid="currency-item"]'
 
 const hideTokenDefaultString = 'Hide tokens'
 const assetNameSortBtnStr = 'Asset'
@@ -90,6 +92,15 @@ export const currentcyGnosisFormat = '< 0.00001 GNO'
 
 export const currencyOx = /^0x$/
 export const currentcyOxFormat = '1.003 ZRX'
+
+function clickOnCurrencySelector() {
+  cy.get(currencySelector).click()
+}
+
+export function changeCurrency(currency) {
+  clickOnCurrencySelector()
+  cy.get(currencyItem).contains(currency).click()
+}
 
 export function clickOnSendBtn(index) {
   cy.get('button')
--- a/cypress/e2e/regression/sidebar_2.cy.js
+++ b/cypress/e2e/regression/sidebar_2.cy.js
@@ -2,14 +2,16 @@ import * as constants from '../../support/constants'
 import * as main from '../pages/main.page'
 import * as sideBar from '../pages/sidebar.pages'
 import * as ls from '../../support/localstorage_data.js'
+import * as assets from '../pages/assets.pages.js'
 
 const newSafeName = 'Added safe 3'
-const oldSafeName = 'Added safe 2'
+const oldSafeName = 'Added safe 900'
 const staticSafe100 = 'Added safe 100'
+const staticSafe200 = 'Added safe 200'
 
 describe('Sidebar added sidebar tests', () => {
   beforeEach(() => {
-    cy.visit(constants.homeUrl + constants.SEPOLIA_TEST_SAFE_13)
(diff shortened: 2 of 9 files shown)


cypress/e2e/pages/assets.pages.js
```diff
@@ -17,6 +17,8 @@ const hiddenTokenSaveBtn = 'span[data-track="assets: Save hide dialog"]'
 const hiddenTokenCancelBtn = 'span[data-track="assets: Cancel hide dialog"]'
 const hiddenTokenDeselectAllBtn = 'span[data-track="assets: Deselect all hide dialog"]'
 const hiddenTokenIcon = 'svg[data-testid="VisibilityOffOutlinedIcon"]'
+const currencySelector = '[data-testid="currency-selector"]'
+const currencyItem = '[data-testid="currency-item"]'
```
francovenica, line 21: Not a big fan of having a data test id that should be unique in several elements. Is there a way to have it being "currency-item-1", "currency-item-2", ... ?

cypress/e2e/regression/sidebar_2.cy.js
```diff
@@ -45,4 +47,19 @@ describe('Sidebar added sidebar tests', () => {
     sideBar.openSidebar()
     sideBar.verifySafeReadOnlyState(staticSafe100)
   })
+
+  it('Verify Fiat currency changes when edited in the assets tab', () => {
+    assets.changeCurrency(constants.currencies.cad)
+    sideBar.checkCurrencyInHeader(constants.currencies.cad)
+  })
+
+  it('Verify "wallet" tag counter if the safe has tx ready for execution', () => {
```
francovenica, line 56: This test description is not correct. That counter counts how many tx in queue the current connected owner can sign, not necesarely they can be executed.
Change it to something like "Verify "wallet" tag counter of tx that need signature from current owner"

cypress/e2e/regression/sidebar_2.cy.js
```diff
@@ -45,4 +47,19 @@ describe('Sidebar added sidebar tests', () => {
...
+    sideBar.openSidebar()
+    sideBar.verifyMissingSignature(staticSafe200)
+  })
+
+  it('Verify "Wallet" tag counter only shows for owners', () => {
```
francovenica, line 61: Same with this description, is not correct. The gray tag with the "check" icon shows how many tx are in queue for everyone (owners, non-owners, disconnected users)

src/components/balances/CurrencySelect/index.tsx
```diff
@@ -43,7 +44,7 @@ const CurrencySelect = (): ReactElement => {
         onClose={() => handleTrack('Close')}
       >
         {fiatCurrencies.map((item) => (
-          <MenuItem key={item} value={item} sx={{ overflow: 'hidden' }}>
+          <MenuItem data-testid="currency-item" key={item} value={item} sx={{ overflow: 'hidden' }}>
```
francovenica, line 47: Not a big fan of having a data test id that should be unique in several elements. Is there a way to have it being "currency-item-1", "currency-item-2", ... ?
Same with other elements like "missing-signature-info" and "queued-tx-info"

Not a blocker if not possible/too cumbersome to do

francovenica (changes requested):
Only the ticket description are a request to change.
The comments about the data-testid are a suggestion

src/components/balances/CurrencySelect/index.tsx
mike10ca: I suggest to keep these data-testeds as is, this is a common practise to assign data-testid for the same element types which can be accessed through Cypress in more precise way.

cypress/e2e/regression/sidebar_2.cy.js
mike10ca: This is original test title taken from our hub of tests and scheduled for automation, the implementation is according to the title. If there are changes needed in updating test title, we can do this later, outside of the scope of this task.

cypress/e2e/regression/sidebar_2.cy.js
mike10ca: This is original test title taken from our hub of tests and scheduled for automation, the implementation is according to the title. If there are changes needed in updating test title, we can do this later, outside of the scope of this task.

francovenica (approved):

PR #3219, opened 2024-02-07
fix: ENS resolution

usame-algan:
## What it solves

The ENS resolution stopped working after #3208.

## How this PR fixes it

This PR partially reverts the changes and instantiates the `JsonRpcProvider` without a network object.

## How to test it

1. Open a safe on mainnet or sepolia
2. Create a new transaction
3. Type an ENS as the recipient
4. Observe the name resolves

## Checklist


schmanu (approved):
Good find!

PR #3226, opened 2024-02-08
[Counterfactual] Show notification when submitting counterfactual tx

usame-algan:
## What it solves

Resolves #

## How this PR fixes it

- Closes the deploy safe tx-flow when submitting the transaction
- Shows a notification when submitting the deploy safe transaction
- Shows an info screen on the first-tx + deploy flow breaking down the cost of the transaction
- Fixes an issue where the counterfactual state was cleared even if the transaction failed
- Estimates the actual gas costs of the signed transaction and uses that instead of what `useDeployGasLimit` returns

## How to test it

2. Choose the Activate now option from settings
3. Submit the transaction
4. Observe the modal closes
5. Observe there is a notification
6. Deploy the safe with the first transaction
7. Observe there is an info screen visible
8. Cancel the transaction in your wallet
9. Observe an error notification shown
10. The counterfactual state should not be cleared in that case

## Screenshots

Screenshot 2024-02-08 at 17 54 18
Screenshot 2024-02-09 at 12 29 55

## Checklist


src/features/counterfactual/CounterfactualForm.tsx
```diff
@@ -113,6 +114,29 @@ export const CounterfactualForm = ({
...
+          <strong>
+            {getTotalFeeFormatted(
+              advancedParams.maxFeePerGas,
+              advancedParams.maxPriorityFeePerGas,
+              BigInt(gasLimit?.safeDeploymentGas || '0'),
+              chain,
+            )}{' '}
+            {chain?.nativeCurrency.symbol}
+          </strong>
+          . This is an estimation and the final cost can be slightly higher.
```
katspaugh, line 138:
```suggestion
          . This is an estimation and the final cost can be higher.
```

src/features/counterfactual/CounterfactualForm.tsx
```diff
@@ -141,15 +164,12 @@ export const CounterfactualForm = ({
...
         )}
-
         {submitError && (
           <Box mt={1}>
             <ErrorMessage error={submitError}>Error submitting the transaction. Please try again.</ErrorMessage>
           </Box>
         )}
-
         <Divider className={commonCss.nestedDivider} sx={{ pt: 3 }} />
-
```
katspaugh, line 152: Waren gute, diese.

src/features/counterfactual/utils.ts
```diff
@@ -178,3 +182,16 @@ export const createCounterfactualSafe = (
...
 }
+
+export const showSubmitNotification = (dispatch: AppDispatch, chain?: ChainInfo, txHash?: string) => {
+  const link = chain && txHash ? getBlockExplorerLink(chain, txHash) : undefined
+  dispatch(
+    showNotification({
+      variant: 'info',
+      groupKey: 'cf-activate-account',
+      message: 'Transaction submitted',
+      detailedMessage: 'Your Safe Account will be deployed on-chain after the transaction has been executed.',
```
katspaugh, line 193:
```suggestion
      detailedMessage: 'Your Safe Account will be deployed onchain after the transaction is executed.',
```

katspaugh (approved):

src/features/counterfactual/CounterfactualForm.tsx
usame-algan: This also happened another time but I can't figure out how they are deleted. Afaik we don't have any lint or prettier rules to remove these...

src/features/counterfactual/CounterfactualForm.tsx
```diff
@@ -113,6 +114,30 @@ export const CounterfactualForm = ({
...
+          <strong>
+            {getTotalFeeFormatted(
+              advancedParams.maxFeePerGas,
+              advancedParams.maxPriorityFeePerGas,
+              BigInt(gasLimit?.baseGas || '0') + BigInt(gasLimit?.safeTxGas || '0'),
+              chain,
+            )}{' '}
+            {chain?.nativeCurrency.symbol}
+          </strong>
+          , activation fee is{' '}
```
schmanu, line 128: Should we add here that this fee will only have to be paid once or that future transactions will not have to pay this?

src/features/counterfactual/utils.ts
```diff
@@ -65,8 +67,11 @@ export const dispatchTxExecutionAndDeploySafe = async (
 
     const deploymentTx = await sdkUnchecked.wrapSafeTransactionIntoDeploymentBatch(signedTx, txOptions)
 
+    // We need to estimate the actual gasLimit after the user has signed since it is more accurate than what useDeployGasLimit returns
+    const gas = await signer.estimateGas({ data: deploymentTx.data, value: deploymentTx.value, to: deploymentTx.to })
```
schmanu, line 71: Should we compare this to the estimation and give a prompt / warning if the difference is more than e.g. 50k gas?

src/features/counterfactual/hooks/useDeployGasLimit.ts
```diff
@@ -21,24 +21,33 @@ const getExtraGasForSafety = (safeTx?: SafeTransaction): bigint => {
...
 
-    const [gas, safeTxGas, safeDeploymentGas] = await Promise.all([
+    const [baseGas, safeTxGas, safeDeploymentGas] = await Promise.all([
       safeTx ? estimateTxBaseGas(sdk, safeTx) : '0',
       safeTx ? estimateSafeTxGas(sdk, safeTx) : '0',
       estimateSafeDeploymentGas(sdk),
     ])
 
-    return BigInt(gas) + BigInt(safeTxGas) + BigInt(safeDeploymentGas) + extraGasForSafety
+    const totalGas = BigInt(baseGas) + BigInt(safeTxGas) + BigInt(safeDeploymentGas) + extraGasForSafety
```
schmanu, line 48: Let's add a test for this hook

src/features/counterfactual/CounterfactualForm.tsx
usame-algan: Good idea, I reworded it to say "one time activation fee is ..."

src/features/counterfactual/utils.ts
usame-algan: I like the idea but this would only happen after the user has pressed Sign in their wallet. I am not sure how helpful it would be to see a warning somewhere in our UI at that point so I would leave it out for now and hope there are no substantial differences between those first transactions. Worst case we have the info text to warn the user about higher cost on execution.

francovenica:
Issue/Question:
1 - In this PR I cannot create a tx. I tried send funds and adding owners, but they always fail. I tried increasing the gas fee, but is to no avail.
Are regular tx working? I always get the same error. I want to verify the kind of notifications I see when I execute a send funds tx but currently I can't.
image

usame-algan:
> In this PR I cannot create a tx. I tried send funds and adding owners, but they always fail. I tried increasing the gas fee, but is to no avail.

There was a small issue with our last commit but I just pushed a small fix. It should work again now.

francovenica:
I can execute tx like adding/swapping owners.

Some feedback (we can discuss what is possible to change or not)

- When you execute the "Activate safe" tx, you get this notification, which is fine, but you have to click in the details to see the text and also in the list of notifications you cannot see the messege, since there is no "details" link to click, you only see a "transaction submitted" which it can be anything.
(This one is fine, but you got to click on details, a minor feedback)
image

This is the same tx, but saying just "tx submitted" doesn't describe that this was the tx that deployed the safe
image

- Once the tx that deploys the safe is finally processed and fully executed, I'd expecte a green card notification saying "success" or something. Right now no notification shows up once the safe is finally deployed

- When I execute a tx like swap owner. I only get only the notificaton of "check in blockexplorer" which is the one deploying the safe, but there is no message implying this for the user. I'd expect here for the notification I showed earlier to show up (The "your safe account will be deployed...." notification)
image

usame-algan:
> When you execute the "Activate safe" tx, you get this notification, which is fine, but you have to click in the details to see the text and also in the list of notifications you cannot see the messege

I adjusted the wording to be more specific
Screenshot 2024-02-12 at 09 34 45

> Once the tx that deploys the safe is finally processed and fully executed, I'd expecte a green card notification saying "success" or something. Right now no notification shows up once the safe is finally deployed

This didn't work because we usually show the green "Success" notification when we detect the transaction in the history which we can't for these transactions because they don't have a txId. I changed it to show the success event when the transaction is successfully executed and when we clear the counterfactual state.

liliya-soroka:
@usame-algan , I don't see success notification after the safe was activated :
image

usame-algan:
> I don't see success notification after the safe was activated :

Thanks, I added a success notification there as well.

liliya-soroka:
Verified

PR #3231, opened 2024-02-09
Feat: Native Swaps

### LINKED ISSUE #3585: Native Swaps
## What is this feature about

https://www.notion.so/safe-global/07996126a3c041db90e36e01eb4ce46c?v=13eec337284943818b2c05dacbec157e

## Requirements

- [ ] Swaps are accessible from the dashboard of a Safe
- [ ] Swaps are accessible from the Assets table
- [ ] Swaps are accessible from the Sidebar
- [ ] Swaps are decoded when creating them
- [ ] Swaps are decoded in the transaction queue
- [ ] Swaps are decoded in the transaction history

compojoom:
## What it solves

Resolves #3585 

## How this PR fixes it

Adds a native swap widget to a new page and shows decoded info for swap orders.

## How to test it

1. Open a Safe
2. Go to the dashboard
3. Observe a Swap button in the top right
4. Observe a Swap button in the Sidebar
5. Go to the Assets page
6. Observe a Swap button next to each asset

## Checklist


src/features/swap/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+          }
+        },
+      },
+    ]
+  }, [dispatch])
+
+  const [params, setParams] = useState<CowSwapWidgetParams | null>(null)
+  useEffect(() => {
+    setParams({
+      appCode: 'Safe Wallet Swaps', // Name of your app (max 50 characters)
```
anxolin, line 122: nit: I would add here the name of the app: `Safe Wallet Web` or maybe app.safe.global if you prefer

src/features/swap/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+      width: '100%', // Width in pixels (or 100% to use all available space)
+      height: '860px',
+      // provider: safeAppWeb3Provider, // Ethereum EIP-1193 provider. For a quick test, you can pass `window.ethereum`, but consider using something like https://web3modal.com
+      chainId, // 1 (Mainnet), 5 (Goerli), 100 (Gnosis)
```
anxolin, line 126: I think you don't need to provide this (if the provider is detected, it will ask the provider)

src/features/swap/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+      // standaloneMode: false,
+      disableToastMessages: true,
+      disablePostedOrderConfirmationModal: true,
+      hideLogo: true,
+      hideNetworkSelector: true,
+      sounds: {
+        orderError: null,
+        orderExecuted: null,
+        postOrder: null,
```
anxolin, line 135: 🥲 Why not adding something not too annoying (like a laser gun shoot)
The mooooo  sound is the most praised feedback. Every single week we get people comments on it on the feedback button

src/features/swap/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+        text: palette.text.primary,
+        danger: palette.error.dark,
+        info: palette.info.main,
+        success: palette.success.main,
+        warning: palette.warning.main,
+        alert: palette.warning.main,
+      },
+      partnerFee: {
+        bps: 50,
+        recipient: '0x0B00b3227A5F3df3484f03990A87e02EbaD2F888',
```
anxolin, line 166: this is a smart contract... shouldn't you change this depending on the network? I can see is not deployed in Gnosis Chain... https://gnosisscan.io/address/0x0B00b3227A5F3df3484f03990A87e02EbaD2F888

src/features/swap/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const [params, setParams] = useState<CowSwapWidgetParams | null>(null)
+  useEffect(() => {
+    setParams({
+      appCode: 'Safe Wallet Swaps', // Name of your app (max 50 characters)
+      width: '100%', // Width in pixels (or 100% to use all available space)
+      height: '860px',
+      // provider: safeAppWeb3Provider, // Ethereum EIP-1193 provider. For a quick test, you can pass `window.ethereum`, but consider using something like https://web3modal.com
+      chainId, // 1 (Mainnet), 5 (Goerli), 100 (Gnosis)
+      // standaloneMode: false,
```
anxolin, line 127: I think you should set this `standaloneMode: false` right? 

We use Dapp mode, so the widget doesn't offer to connect / disconnect. We don't need to pass the provider to the widget either, because it detect the Safe SDK provider automatically

compojoom: Isn't the default value false? If I set it to true and not connected I can see the EOAs address, If it is false or not set at all I don't see that.

src/features/swap/index.tsx
compojoom: doesn't seem to be the case. If I remove the chainId and let's say I'm on sepolia the page asks me to switch network to mainnet?

anxolin: My point is we should not hardcode this, as it will depend on the selected safe you use 

In the example, we didn't add it as param 
https://github.com/cowprotocol/safe-app-example/blob/main/src/components/App.tsx

src/features/swap/index.tsx
anxolin: Yep, right, just to not rely on defaults and be explicit. Probably default won't change

src/components/common/BuyCryptoButton/index.tsx
```diff
@@ -30,7 +31,7 @@ const useBuyCryptoHref = (): LinkProps['href'] | undefined => {
 }
 
 const buttonStyles = {
-  minHeight: '40px',
+  minHeight: '37.5px',
```
katspaugh, line 34: What is this magic `37.5px` and why is it necessary in several places?

src/hooks/useTransactionType.ts
```diff
@@ -61,6 +62,14 @@ export const getTransactionType = (tx: TransactionSummary, addressBook: AddressB
         text: isDeleteGuard ? 'deleteGuard' : tx.txInfo.dataDecoded.method,
       }
     }
+    case TransactionInfoType.SWAP_ORDER: {
+      const orderClass = getOrderClass(tx.txInfo)
+
+      return {
+        icon: '/images/transactions/cow-logo.png',
```
katspaugh, line 69: Please minimize the PNGs using tinypng if you haven't already.

src/services/safe-wallet-provider/rpc.ts
```diff
@@ -0,0 +1,35 @@
...
+      iconUrl: 'icon',
+    })
+
+    if ('result' in response) {
+      return response.result
+    }
+  }
+
+  async on() {
+    console.log('swap on =============================')
```
katspaugh, line 29: Useful logs are OK to keep, but please make them look like the other logs, e.g. the GA ones.

src/services/safe-wallet-provider/rpc.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  async request(request: { method: string; params?: Array<any> | Record<string, any> }): Promise<any> {
+    if (request.method === 'eth_requestAccounts') {
+      request.method = 'eth_accounts'
+    }
+    const response = await this.walletProvider.request(0, request, {
+      id: -1,
+      url: '',
+      name: 'Built-inswap',
```
katspaugh, line 18:
```suggestion
      name: 'Built-in swaps',
```

src/services/safe-wallet-provider/rpc.ts
```diff
@@ -0,0 +1,35 @@
...
+      return response.result
+    }
+  }
+
+  async on() {
+    console.log('swap on =============================')
+  }
+
+  async enable() {
+    await this.walletProvider.eth_accounts()
```
katspaugh, line 33: Do you need to return this promise?

src/services/tx/extractTxInfo.ts
```diff
@@ -74,12 +78,21 @@ const extractTxInfo = (
         } else {
           return txDetails.txInfo.transferInfo.tokenAddress
         }
+      case 'SwapOrder':
+        const swapOrderTo = txDetails.txData?.to.value
+        // TODO: Remove assertion after type is corrected
```
katspaugh, line 83: Is it corrected now?

src/store/swapOrderSlice.ts
```diff
@@ -0,0 +1,206 @@
...
+          dispatch(
+            showNotification({
+              title: 'Order cancelled',
+              message: 'Your order has been cancelled',
+              groupKey,
+              variant: 'warning',
+            }),
+          )
+          break
+      }
```
katspaugh, line 163: This can be made much more declaritive. E.g. a config where keys are order statuses and values are messages.

src/styles/globals.css
```diff
@@ -14,6 +14,10 @@ main {
   width: 100%;
 }
 
+main.swapWrapper {
+  height: calc(100vh - 52px);
```
katspaugh, line 18: Can this be not put as as an inline style of the swapWrapper?

src/utils/transaction-guards.ts
```diff
@@ -88,6 +92,32 @@ export const isMultiSendTxInfo = (value: TransactionInfo): value is MultiSend =>
...
+): decodedData is CowSwapConfirmationView => {
+  if (!decodedData) {
+    return false
+  }
+
+  if ('type' in decodedData) {
+    return decodedData.type === 'COW_SWAP_ORDER'
+  }
+
+  return false
```
katspaugh, line 110:
```suggestion
  if (decodedData && ('type' in decodedData)) {
    return decodedData.type === 'COW_SWAP_ORDER'
  }

  return false
```

src/services/safe-wallet-provider/rpc.ts
compojoom: This class is not used at all. We forgot to remove it.

src/services/safe-wallet-provider/rpc.ts
usame-algan, line 1: This can be removed as its not used anywhere

src/features/swap/index.tsx
compojoom: sadly, marketing is against it for the moment.

src/components/common/BuyCryptoButton/index.tsx
compojoom: @usame-algan ^^ :D ?

src/store/swapOrderSlice.ts
compojoom: It's a bit more complicated, because we have some states that we are not reacting too and also in some of the cases we are dispatching multiple events, so I would leave it as it is for now.

src/components/common/BuyCryptoButton/index.tsx
usame-algan: If there is an outlined button and a contained button next to each other they have different heights because of the border. Also depending on the icon inside the button they can have different heights so setting the height this way was a quick solution. Imo we should fix this in our mui theme and align it with our style guide in Figma but I would do that separate from this PR.

src/services/tx/extractTxInfo.ts
compojoom: I don't know. According to the types txData might not exist. If it exist, then it definetely has a to prop. I would leave the typeguard check for now. (I've removed locally and the app has not crashed, but would prefer to have it there a bit longer)

src/features/swap/index.tsx
```diff
@@ -0,0 +1,246 @@
...
+    }),
+    [darkMode],
+  )
+
+  const listeners = useMemo<CowEventListeners>(() => {
+    return [
+      {
+        event: CowEvents.ON_TOAST_MESSAGE,
+        handler: (event) => {
+          console.info('🍞 New toast message:', event)
```
katspaugh, line 86:
```suggestion
          console.info('[Swaps] message:', event)
```

src/services/analytics/events/swaps.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const SWAP_EVENTS = {
+  SWAP_ASSETS: {
+    action: 'Swap asset',
+    category: SWAP_CATEGORY,
+  },
+  SWAP_TOKENS: {
+    action: 'Swap total asset',
+    category: SWAP_CATEGORY,
+  },
```
katspaugh, line 11: @alicecha have you already reviewed the analytics events on this branch?

src/features/swap/hooks/useIsExpiredSwap.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Checks the expiry time of a swap every 10s
+ * and returns true if the swap expired
+ * @param txInfo
+ */
+const useIsExpiredSwap = (txInfo: TransactionInfo) => {
+  const [isExpired, setIsExpired] = useState<boolean>(false)
+
+  const isExpiredSwap = () => {
```
katspaugh, line 16: The name of this function sounds like it returns a boolen but it actually sets the state.

src/features/swap/hooks/useIsExpiredSwap.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (!isSwapTxInfo(txInfo)) return
+
+    setIsExpired(Date.now() > txInfo.validUntil * 1000)
+  }
+
+  useInterval(isExpiredSwap, INTERVAL_IN_MS)
```
katspaugh, line 22: This could be easily replaced with useIntervalCounter + an effect. No need for a new hook then.

src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
```diff
@@ -15,6 +16,7 @@ export type EthHashInfoProps = {
   chainId?: string
   name?: string | null
   showAvatar?: boolean
+  onlyName?: boolean
```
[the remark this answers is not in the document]
jmealy: I think this is more for displaying the names of contracts. Eg.
```
<NamedAddress address={settlementContract} onlyName hasExplorer shortAddress={false} avatarSize={24} />
```

src/features/swap/hooks/useIsExpiredSwap.ts
```diff
@@ -1,7 +1,7 @@
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 import type { TransactionInfo } from '@safe-global/safe-gateway-typescript-sdk'
-import useInterval from '@/hooks/useInterval'
```
katspaugh: I guess this hook itself can be removed now.

katspaugh (approved):
🎉

PR #3234, opened 2024-02-09
feat: [Counterfactual] Update UI elements

### LINKED ISSUE #3207: [Counterfactual] Implement Dashboard design
## What is this about

With counterfactual safe deployments we also want to improve the onboarding experience for the user

## Requirements

- [ ] Update the overview widget
- [ ] Add a new "First steps" widget
- [ ] Add a new "Add funds" widget

## Screenshots

- [Figma](https://www.figma.com/file/fUOaMHVOhsIylBjxiL5uEn/Counterfactual-deployment?node-id=3205%3A2409&mode=dev)

usame-algan:
## What it solves

Resolves #3207

## How this PR fixes it

- Updates the overview widget design
- Adds a "First steps" widget that is only visible for undeployed safes
- Hide every other widget for undeployed safes
- Hide the overview buttons for undeployed safes
- Shows an info message on the assets page with the option to activate the safe immediately
- Displays a new First transaction dialog
- Shows an indicator in the sidebar about the safe account status

## How to test it

1. Only the First steps widget should be visible
2. There should be no buttons visible in the overview widget
3. There should be an info message on the assets page with an "Activate now" button
4. Pressing "Create transaction" should display the new dialog with multiple options
5. There should be a warning indicator in the sidebar as long as the safe is not deploying
6. Once a deployment is underway the indicator should turn blue and spin

## Screenshots
Screenshot 2024-02-15 at 15 15 15
Screenshot 2024-02-14 at 16 44 21

Screenshot 2024-02-14 at 16 42 26
Screenshot 2024-02-14 at 16 55 53

## Checklist


francovenica:
I wasn't able to reproduce the issue Andrei had with the popup of confirmation after the safe is deployed.

I'll keep the ticket up since some give was still being given in a slack thread, but at least from my side the dashboard looks fine

liliya-soroka:
From the UI the PR looks good. 
After passing checks, the PR can be merged 
@usame-algan

src/components/common/BuyCryptoButton/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+import madProps from '@/utils/mad-props'
+import { type ReactNode, useMemo } from 'react'
+import Track from '../Track'
+import { OVERVIEW_EVENTS } from '@/services/analytics'
+import RampLogo from '@/public/images/common/ramp_logo.svg'
+import css from './styles.module.css'
+import ChevronRightRoundedIcon from '@mui/icons-material/ChevronRightRounded'
+
+const useOnrampAppUrl = (): string | undefined => {
+  const [onrampApps] = useRemoteSafeApps(SafeAppsTag.ONRAMP)
```
katspaugh, line 17: This will cause a fetch of the safe app list, so please make sure it's not triggered unnecessarily.
On the Dashboard, we already fetch that list anyway, so it should be fine, but on other pages it will cause an extra request.

src/components/dashboard/FirstSteps/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+      </div>
+      <Typography variant="h4" fontWeight="bold" mb={2} mt={2}>
+        Safe Account is ready!
+      </Typography>
+      <Typography>Continue to improve your account security and unlock more features</Typography>
+    </Card>
+  )
+}
+
+const FirstSteps = () => {
```
katspaugh, line 162: FirstSteps is a very broad name. Perhaps OnboardingSafeActions or something like that?

src/features/counterfactual/utils.ts
```diff
@@ -221,6 +193,30 @@ export const showSubmitNotification = (dispatch: AppDispatch, chain?: ChainInfo,
...
+    const txResponse = await provider.getTransaction(txHash)
+    if (txResponse !== null) {
+      return txResponse
+    }
+    if (attempt < maxAttempts - 1) {
+      await delay(delayMs)
+    }
+  }
+  throw new Error('Transaction not found')
+}
```
katspaugh, line 218: We have the exponential-backoff package for that. Possible to use that?

katspaugh (approved):

src/features/counterfactual/utils.ts
usame-algan: I have a local stash to replace that package with our own implementation so I would leave it as is for now since we will remove that dependency anyway.

Issue #3156, opened 2024-01-25
Counterfactual account creation v1


usame-algan:
## What is the feature about

Users need to have funds on their signer wallets in order to create safes. This feature aims to separate the safe creation and safe deployment by making use of CREATE2 to unlock the wallet UI even for undeployed safes.

## What is in the scope?

The initial version of this feature focuses on 1/1 safe setups to minimize the entropy of predicted safe addresses and to not require additional services.

## What is out of scope?

- Counterfactual safes for Multisigs

## The list of requirements

- [x] #3179
- [x] #3184
- [x] #3191
- [x] #3195
- [x] #3203
- [x] #3210
- [x] #3207
- [x] Implement design for other pages (assets, settings)
- [x] #3270

## Designs/sketches

- [Figma](https://www.figma.com/file/fUOaMHVOhsIylBjxiL5uEn/Counterfactual-deployment?type=design&node-id=0-1&mode=design&t=OXDd4zlX6JnOx7ma-0) (WIP)

## Links

Tech spec: [Notion](https://www.notion.so/safe-global/MVP-Deterministic-1-1-safes-1918b2d500ae4a34a31346eb067ab8ae)

PR #3238, opened 2024-02-12
fix: [Counterfactual] Deploy safe with specific version

usame-algan:
## What it solves

Part of #3156 

## How this PR fixes it

- Adds a new optional parameter `safeVersion` to `createNewSafe` and `relaySafeCreation`
- Passes the `safeVersion` to those functions for undeployed safes

## How to test it

2. Deploy the safe either via first transaction or separate deployment
3. Observe the safe address is deployed
4. Create a safe normally
5. Observe the predicted safe address is the one that is also deployed

## Checklist


francovenica:
I tried to execute a tx in this PR but It always fails. Probably the error has nothint to do with the save version field, but it stops me from checking the deployment via "first tx"
image

francovenica:
Note: 
I just checked and the issue is that I was using Trezor, so is an issue, but it should be tackled in its own ticket.

This one can be passed to Done

PR #3239, opened 2024-02-12
Fix: sync Redux and localStorage across tabs

### LINKED ISSUE #1986: Address book automatically deleted from Local Storage
## Environment
 - Browser: Brave (unable to replicate in FF)
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to Safe app --> Settings --> Manage owners
 2. Add an owner alias using the pencil icon
 3. See the value get inserted into Address Book local storage
 4. Close the tab, and navigate back to the Safe
 5. See the local storage Address Book is cleared of all data to `{}`

## Expected result
Address book data should be persisted across sessions and not deleted.

## Obtained result
Address book is entirely deleted.

## Screenshots
image

katspaugh:
## What it solves

Resolves #1986

## How this PR fixes it

Adds a new Redux middleware that broadcasts whitelisted Redux actions to all open tabs. Receiver tabs will then dispatch that action.

I also refactored our store setup a bit:
* moved all the slice imports into an index
* got rid of the class-based Hydrator component in favor of a hook
* made the localStorage middleware persist individual action LS items, not the entire state as it was before.

## How to test it

* Open the app in two tabs
* Go to the Address book in both
* Add a new entry in one tab
* The other tab should immediately display the new entry
* Remove the entry in the second tab
* The entry should be immediately gone also in the other tab


compojoom:
What is the rule for "whitelisted Redux actions" that should be synced across tabs? Is it only actions that are local and are not created in response to a network request?

katspaugh:
The persisted ones. The aim of this PR is to prevent tabs overwriting each other’s local storage.

src/store/broadcast.ts
```diff
@@ -0,0 +1,33 @@
...
+const tabId = Math.random().toString(32).slice(2)
+
+export const broadcastState = <K extends keyof PreloadedRootState>(sliceNames: K[]): Middleware<{}, RootState> => {
+  return (_) => (next) => (action) => {
+    const result = next(action)
+
+    // Broadcast actions that aren't being already broadcasted
+    if (!action._isBroadcasted) {
+      const sliceType = action.type.split('/')[0]
+      if (sliceNames.some((slice) => slice === sliceType)) {
```
schmanu, line 18:
```suggestion
      if (sliceNames.includes(sliceType)) {
```

src/store/broadcast.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { Middleware, PreloadedState } from '@reduxjs/toolkit'
+import type { RootState } from '@/store'
+
+type PreloadedRootState = PreloadedState<RootState>
+
+const BC_NAME = 'SAFE__store-updates'
+const broadcast = typeof BroadcastChannel !== 'undefined' ? new BroadcastChannel(BC_NAME) : undefined
+const tabId = Math.random().toString(32).slice(2)
+
+export const broadcastState = <K extends keyof PreloadedRootState>(sliceNames: K[]): Middleware<{}, RootState> => {
```
schmanu, line 11: We should try to test this.

src/store/storeHydrator.ts
```diff
@@ -0,0 +1,13 @@
+import { useEffect } from 'react'
+import { getPersistedState, type makeStore } from '@/store'
+
+export const HYDRATE_ACTION = '@@HYDRATE'
+
+export const useHydrateStore = (store: ReturnType<typeof makeStore>) => {
```
schmanu, line 6: Let's rename the file to `useHydrateStore`

schmanu (commented):
This solves a really big pain point!

src/store/broadcast.ts
katspaugh: Added a few tests.

schmanu (approved):
I tried it with multiple Tabs and e.g adding Safes and changing addressbook entries worked like a charm across tabs!

francovenica:
Looks really nice.

I checked that in different tabs the address book are in sync, also using owner name editting in the settings as the original user had issues with:
address

PR #3240, opened 2024-02-12
[Counterfactual] Disable notifications and message signing

usame-algan:
## What it solves

Part of #3156

## How this PR fixes it

- Disables the Notifications Banner for newly created counterfactual safes
- Disables the Button on the Notifications settings page
- Hides counterfactual safes from the Global notifications page
- Disables the sign message flow for counterfactual safes

## How to test it

2. Observe no Popup saying "Enable all"
3. Go to the Notification settings
4. Observe a disabled button
5. Go to the global notification preferences
6. Observe the safe is not showing up in the list
7. Connect your safe with an app to create an off-chain message (e.g. opensea)
8. Observe an error message in the modal and the Sign button is disabled

## Screenshots
Screenshot 2024-02-12 at 17 57 40
Screenshot 2024-02-12 at 17 57 55
Screenshot 2024-02-13 at 17 25 59

## Checklist


francovenica:
Looks good to me

Notifications pop up when creating a CF doesn't show up.
Notifications are locked for CF safes in the settings
The CF is not listed in the global settings for notifications
The error message is in the Message Form to sign (and the sign button is disabled)

Also checked that:
Enabling all notifications in all the safes in the network, then deployed the CF safe, then checking the global settings for notifications the CF safe is not checked (confirming that the enabling of notifcations was not affecting the CF safe)
The notification pop up still works for deployed safes (adding deployed safes, or paying for the creation of one)
Notifications can be enabled for a CF safe that was deployed.

src/components/settings/PushNotifications/GlobalPushNotifications.tsx
```diff
@@ -37,6 +41,16 @@ import css from './styles.module.css'
 
 // UI logic
 
+const _filterUndeployedSafes = (safes: NotifiableSafes, undeployedSafes: UndeployedSafesState) => {
```
compojoom, line 44: safes here is a `{[chain]: [address1, address2]}` but since the variable name is `safes` I thought it was a safe object and later when we do difference(safes, undeployedAddresses) I was astonished that it works :)
Maybe at least in mapValues we could call it safeAddresses?

src/components/settings/PushNotifications/PushNotificationsBanner/index.tsx
```diff
@@ -121,7 +121,12 @@ export const PushNotificationsBanner = ({ children }: { children: ReactElement }
   const isSafeAdded = !!addedSafesOnChain?.[safeAddress]
   const isSafeRegistered = getPreferences(safe.chainId, safeAddress)
   const shouldShowBanner = useDebounce(
```
compojoom, line 123: just curious - why do we have to debounce here?

usame-algan: I think the issue was that there were too many popups opening up at the same time which looked messy. This way there is a small delay before this popup opens. Not ideal but a bit better on the UX side.

src/components/settings/PushNotifications/index.tsx
```diff
@@ -140,7 +140,7 @@ export const PushNotifications = (): ReactElement => {
                         <FormControlLabel
                           control={<Switch checked={!!preferences} onChange={handleOnChange} />}
                           label={preferences ? 'On' : 'Off'}
-                          disabled={!isOk || isRegistering}
+                          disabled={!isOk || isRegistering || !safe.deployed}
```
compojoom, line 143: in SignMessage we do `const isDisabled= ` which I prefer, but ok :)

compojoom (approved):

src/components/settings/PushNotifications/index.tsx
usame-algan: Agree, with that many variables it makes sense to extract it into a new one for better readability.

usame-algan: One issue is that here we have the `isOk` variable which comes from the HoC so we can't extract that easily.

PR #3242, opened 2024-02-12
fix: [Counterfactual] Add event bus and pending state

usame-algan:
## What it solves

Part of #3156

## How this PR fixes it

- Adds an Event Bus for `SafeCreationEvents`
- Dispatches `SafeCreationEvents` when deploying a counterfactual safe
- Persists pending state status in `undeployedSafesSlice`
- Shows notifications for `SafeCreationEvents`
- Monitors undeployedSafes if they are relaying or processing
- Checks for an undeployedSafe if its already deployed on-chain and cleans up local storage in case it is
- Supports cancellation and speed-up safe creations
- Disables the create button in the safe creation flow and adds a loading spinner while submitting

## How to test it

Happy path:

2. Deploy safe via "Activate now" or with first tx
4. Observe Processing notification showing
5. Observe the "Activate now" button is disabled and has a loading state on it
8. Observe a Success notification
9. The undeployed state should be cleaned up

Pending state is persisted:

2. Deploye safe via "Activate now" or with first tx
3. Observe Processing notification showing
4. Reload the app
5. Observe the "Activate now" button is disabled and has a loading state on it
6. Observe a Success notification

Sad path:

2. Deploy safe via "Activate now" or with first tx with too little gas
3. Observe Processing notification
4. Observe Timeout error after a while
5. Cancel transaction in wallet and execute again
6. This time speed up the transaction
7. Observe a success notification

Cleaning state:

2. Deploy the safe via "Activate now" and immediately close the app
3. Wait until the transaction is successfully executed via wallet
4. Open the app again
5. Observe the undeployed state is cleaned up

Creation Loading state:

1. Go through the safe creation flow
2. Press Create
3. Observe that the Create button is disabled and has a loading spinner

## Checklist


src/features/counterfactual/ActivateAccount.tsx
```diff
@@ -81,13 +78,19 @@ const ActivateAccountFlow = () => {
 
   if (!undeployedSafe) return null
 
+  const owners = undeployedSafe.props.safeAccountConfig.owners
+  const threshold = undeployedSafe.props.safeAccountConfig.threshold
+  const saltNonce = undeployedSafe.props.safeDeploymentConfig?.saltNonce
+  const safeVersion = undeployedSafe.props.safeDeploymentConfig?.safeVersion
```
katspaugh, line 84:
```suggestion
  const { owners, threshold } = undeployedSafe.props.safeAccountConfig
  const { saltNonce, safeVersion } = undeployedSafe.props.safeDeploymentConfig || {}
```

src/features/counterfactual/services/safeCreationEvents.ts
```diff
@@ -0,0 +1,45 @@
+import EventBus from '../../../services/EventBus'
```
katspaugh, line 1: Please use an absolute path.

src/features/counterfactual/services/safeCreationEvents.ts
```diff
@@ -0,0 +1,45 @@
...
+const SafeCreationEventBus = new EventBus<SafeCreationEvents>()
+
+export const safeCreationDispatch = SafeCreationEventBus.dispatch.bind(SafeCreationEventBus)
+
+export const safeCreationSubscribe = SafeCreationEventBus.subscribe.bind(SafeCreationEventBus)
+
+// Log all events
+Object.values(SafeCreationEvent).forEach((event: SafeCreationEvent) => {
+  safeCreationSubscribe<SafeCreationEvent>(event, (detail) => {
+    console.info(`Safe creation ${event} event received`, detail)
```
katspaugh, line 43:
```suggestion
    console.info(`[Safe creation]: ${event}`, detail)
```

src/features/counterfactual/utils.ts
```diff
@@ -220,3 +220,90 @@ export const showSubmitNotification = (dispatch: AppDispatch, chain?: ChainInfo,
     }),
   )
 }
+
+// TODO: Reuse this for safe creation flow instead of checkSafeCreationTx
```
katspaugh, line 224: There's a lot of duplicate code indeed.

src/hooks/usePendingSafeStatuses.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    const monitorPendingSafe = async () => {
+      const {
+        status: { status, txHash, taskId },
+      } = undeployedSafe
+
+      const isProcessing = status === PendingSafeStatus.PROCESSING && txHash !== undefined
+      const isRelaying = status === PendingSafeStatus.RELAYING && taskId !== undefined
+      const isMonitored = monitoredSafes.current[safeAddress]
+      console.log({ isMonitored })
```
katspaugh:
```suggestion
```

src/hooks/usePendingSafeStatuses.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      monitoredSafes.current[safeAddress] = true
+
+      if (isProcessing) {
+        console.log(isProcessing)
+        checkSafeActivation(provider, txHash, safeAddress)
+      }
+
+      if (isRelaying) {
+        console.log(isRelaying)
```
katspaugh:
```suggestion
```

src/hooks/usePendingSafeStatuses.ts
```diff
@@ -0,0 +1,134 @@
...
+
+const pendingStatuses: Partial<Record<SafeCreationEvent, PendingSafeStatus | null>> = {
+  [SafeCreationEvent.PROCESSING]: PendingSafeStatus.PROCESSING,
+  [SafeCreationEvent.RELAYING]: PendingSafeStatus.RELAYING,
+  [SafeCreationEvent.SUCCESS]: null,
+  [SafeCreationEvent.FAILED]: null,
+  [SafeCreationEvent.REVERTED]: null,
+}
+
+const usePendingSafeMonitor = (): void => {
```
katspaugh: Should this file be in features/counterfactual?

src/hooks/usePendingSafeNotifications.ts
```diff
@@ -0,0 +1,71 @@
...
+
+enum Variant {
+  INFO = 'info',
+  SUCCESS = 'success',
+  ERROR = 'error',
+}
+
+const successEvents = [SafeCreationEvent.SUCCESS]
+
+const usePendingSafeNotifications = (): void => {
```
katspaugh: Move to features/counterfactual? Or is it for normal safe creation as well?

katspaugh (commented):
We'll have to refactor all those event buses/associated notifications eventually. There are now regular txs, messages, safe creations. Lots of copypasta.

src/pages/_app.tsx
```diff
@@ -1,3 +1,5 @@
+import usePendingSafeNotifications from '@/features/counterfactual/hooks/usePendingSafeNotifications'
+import usePendingSafeStatus from '@/features/counterfactual/hooks/usePendingSafeStatuses'
```
katspaugh, line 2: Ideally feature-specific imports should be lazy.

src/hooks/usePendingSafeStatuses.ts
usame-algan: Yes, eventually it makes sense to move it out when it will be reused for the safe creation flow but I would keep it in the features folder until then.

src/features/counterfactual/utils.ts
usame-algan: Yeah, I tried reusing the existing code but it has very specific types that are too coupled to the safe creation flow. I plan to reuse this new EventBus to refactor the existing safe creation flow eventually.

src/pages/_app.tsx
usame-algan: Agree, for now we can lazy-load them only if we detect that there is a counterfactual state similar to how we do it with Recovery hooks.

src/features/counterfactual/utils.ts
katspaugh: I meant that in this very PR there are two copies of the same routines, down to exact same error messages.

usame-algan: The only difference is that instead of returning the value or calling a setter we call `safeCreationDispatch`. I would prefer we copy it for now and refactor it later as I don't want to change the safe creation flow logic in the scope of this PR.

src/features/counterfactual/hooks/useIsCounterfactualSafe.ts
```diff
@@ -0,0 +1,12 @@
+import { selectIsUndeployedSafe } from '@/features/counterfactual/store/undeployedSafesSlice'
+import useChainId from '@/hooks/useChainId'
+import useSafeInfo from '@/hooks/useSafeInfo'
+import { useAppSelector } from '@/store'
+
+const useIsCounterfactualSafe = () => {
+  const chainId = useChainId()
+  const { safeAddress } = useSafeInfo()
```
katspaugh, line 8:
```suggestion
  const { safeAddress, safe: { chainId } } = useSafeInfo()
```

src/pages/_app.tsx
```diff
@@ -132,6 +127,8 @@ const WebCoreApp = ({
           <PasswordRecoveryModal />
 
           <Recovery />
+
+          <Counterfactual />
```
katspaugh, line 131: This name is a big vague, maybe CounterFactualHooks?

katspaugh (approved):

francovenica:
The happy path and cleaning state work fine

Question:
Given the statement:
"We are not detecting cancellation and speed-up transactions. This is a general issue that we have since the ethers v6 upgrade so I suggest to ignore it when testing"

I can't test the other 2 scenarios:
The cancel one is not removing the pending state, but given the statement it would be expected

The speeding up one is working in the sense that the safe is being deployed, but the pending state also never goes away, I don't get a "deployed" notification from the safe and I have to hard refresh to have the "undeployed safes" key from the local storage cleared

So the question is. Should I ignore those 2 and test them in the RC?

usame-algan:
> So the question is. Should I ignore those 2 and test them in the RC?

I found a fix for this that wasn't documented in ethers v6 but now we should detect cancellations and speed-ups correctly. I also fixed it for the normal safe creation flow since it was broken there as well. Additionally, I increased the timeout from 1 minute to 2 minutes since I think 1 minute is too fast for this kind of case i.e. transaction is not executing so the user first checks why and then chooses speed-up or cancel this would most likely take more than one minute.

francovenica:
The loading status in the create button looks good
The cancelation of the tx works fine. The pending status stops and a notification saying "cancelled" is shown.

Issue:
With speed up I still have issues. 
I tested 2 times with different outcomes:
First time: I waited for 5 mins and the pending status never went away. I reloaded the page and the pending status kept going for 5 seconds, stopped and I got this error 
image

Second time: This time the pending status went away quickly, but I still got a different error:
image

In both cases the tx worked, so the safe was deployed just fine, but my guess is that the system is still waiting for the original tx instead of the new one that is replacing it

usame-algan:
@francovenica I fixed the issue that an error was shown even if there was a speed-up.

There are still two other issues:

1. Sometimes there is an error shown immediately after Submitting ("Transaction not found"). I suspect this happens more on Sepolia than on other networks because there might be fewer nodes that pick up the transaction from the pool. I don't think we can do much here
2. Sometimes cancellation/speed-up is not picked up and its stuck in processing until it times out. I am still looking into the cause of this but I would suggest we move on and fix it separately as this looks like an ethers v6 issue to me

francovenica:
I tried twice and it works really well both times. Didn't have more issues

LGTM

PR #3246, opened 2024-02-13
fix: [Counterfactual] try signing methods when signing

usame-algan:
## What it solves

Part of #3156

## How this PR fixes it

- Uses the existing `tryOffChainSigning` method when signing the bundled transaction

## How to test it

1. Create a counterfactual safe with a hardware wallet
2. Deploy the safe with the first transaction
3. Observe no error when signing

## Checklist


schmanu (approved):

francovenica:
LGTM
I tried with a Trezor One and a Ledger X 
I was able to sign and then execute with no problems

PR #3250, opened 2024-02-13
Fix: don't display 0 nonce for grouped txs

### LINKED ISSUE #3249: UI: Conflicting transactions with nonce 0 contain O in the row , as result the row is too big
## Bug description
UI: Conflicting transactions with nonce 0 contain O in the row , as result the row is too big
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to https://safe-wallet-web.dev.5afe.dev/transactions/queue?safe=sep:0x7eDa350559afE723a01311E5BA3adD9572396f42
 2. check Conflicting transactions with nonce 0
## Expected result
Conflicting transactions with nonce 0 should not contain nonce in the row 
## Screenshots
image

katspaugh:
## What it solves

Resolves #3249

## How this PR fixes it

Nonce is a number, so 0 wasn't passing the check.

After the fix:
Screenshot 2024-02-13 at 17 08 30

### DIFF
--- a/src/components/transactions/TxSummary/index.tsx
+++ b/src/components/transactions/TxSummary/index.tsx
@@ -42,7 +42,7 @@ const TxSummary = ({ item, isGrouped }: TxSummaryProps): ReactElement => {
       })}
       id={tx.id}
     >
-      {nonce && !isGrouped && (
+      {nonce != null && !isGrouped && (
         <Box gridArea="nonce" data-testid="nonce" className={css.nonce}>
           {nonce}
         </Box>
--- a/src/components/transactions/TxSummary/index.tsx
+++ b/src/components/transactions/TxSummary/index.tsx
@@ -42,7 +42,7 @@ const TxSummary = ({ item, isGrouped }: TxSummaryProps): ReactElement => {
       })}
       id={tx.id}
     >
-      {nonce != null && !isGrouped && (
+      {nonce !== undefined && !isGrouped && (
         <Box gridArea="nonce" data-testid="nonce" className={css.nonce}>
           {nonce}
         </Box>


liliya-soroka:
Verified

src/components/transactions/TxSummary/index.tsx
```diff
@@ -42,7 +42,7 @@ const TxSummary = ({ item, isGrouped }: TxSummaryProps): ReactElement => {
       })}
       id={tx.id}
     >
-      {nonce && !isGrouped && (
+      {nonce != null && !isGrouped && (
```
schmanu, line 45: We really have to stop doing implicit boolean coercion with numbers and bignumber. This is now the 6th(?) bug I see around this.

I looked into it and found this eslint rule:

```
    "@typescript-eslint/strict-boolean-expressions": [
      "warn",
      {
        "allowString": true,
        "allowNumber": false,
        "allowNullableObject": true,
        "allowNullableBoolean": true,
        "allowNullableString": true,
        "allowNullableNumber": false,
        "allowNullableEnum": true
      }
    ],
```


It would have caught this error. But it requires [type checked linting](https://typescript-eslint.io/linting/typed-linting) so it is slower.

schmanu (approved):

src/components/transactions/TxSummary/index.tsx
schmanu: I think this is more accurate as `nonce` is of type `number | undefined`
```suggestion
      {nonce !== undefined && !isGrouped && (
```

PR #3254, opened 2024-02-14
Fix: avoid fetching Safe Apps on all pages

### LINKED ISSUE #3253: Safe Apps are fetched on initial page load
For some reason, when you open any Safe route, it fetches the list of Safe Apps.

We should figure out from where it's being fetched and see if we can avoid doing that. Ideally Safe Apps should be fetched only on the Safe Apps page and the Dashboard.

Furthermore, on the Safe Apps page, this request is done twice.

katspaugh:
## What it solves

Resolves #3253

## How this PR fixes it

The "native" WalletConnect feature has to annotate its transactions as if they were coming from the legacy WalletConnect Safe App. This is to ensure that the transaction service adds the correct icon and name to these transactions in the history and queue.

To avoid making an extra request just to fetch the WC Safe App id and URL, I suggest we hardcode it for now. We should solve the root cause of this dependancy on a legacy Safe App on the backend.


schmanu (commented):
The same is required for the Safe Governance app.
We fetch the Safe Apps for the widget in the header to link to the Safe app.

schmanu (approved):
Vamos!

liliya-soroka:
@katspaugh , have we returned WalletConnect safe app ? 
image

katspaugh:
@liliya-soroka this can happen when you use the prod CGW switch. Let me fix it.

francovenica:
I've checked that the apps still work fine.

Tried connecting with opensea and singing the message
Tried the tx builder in sepolia with a dummy contract.
Tried Cow swap both, as the integrated dapp in the safe and connecting with WC from their webpage (all in sepolia).

I didn't have any issues.

PR #3257, opened 2024-02-14
fix: gasEstimation for counterfactual deployments

schmanu:
## What it solves
- We compute the expected native costs wrong by double accounting the `maxPriorityFee`
- Gas estimation for counterfactual transactions sometimes reverts too early for more advanced txs, that require the `msg.sender` to be the undeployed safe proxy

## How this PR fixes it
- fixes `totalFee` calculation by only multiplying `maxFeePerGas * gasLimit`
- Estimates the counterfactual batch together using the `simulate` function of the fallback handler. This way the inner transaction is a correct `delegateCall`

## How to test it
- Create a new counterfactual Safe
- Add an owner as the first transaction
- Enable Recovery as a first transaction
- Observe the expected `gasLimit` costs to be much closer to the actual `gasLimit`

## Screenshots
Screenshot 2024-02-14 at 16 09 37

## Checklist


usame-algan (approved):

francovenica:
LGTM

I've checked in tx that the gas limit value we show in the tx form is the same (or almost the same) as the one that the wallet shows at the time of execution
Before the gas limit could vary a lot between those 2 instances

usame-algan:
@francovenica I disabled the gasLimit input for undeployed safes as discussed

francovenica:
Checked that is disabled. Looks good

PR #3265, opened 2024-02-15
Refactor: stricter boolean checks with numbers involved

schmanu:
## What it solves

A lot of boolean checks involving numbers could quickly lead to errors if the value is 0.

## How this PR fixes it
- Checked all boolean checks involving numbers
- Fixes some bugs
- Uses `??` instead of `||` where applicable

## Checklist


src/components/safe-apps/SafeAppLandingPage/index.tsx
```diff
@@ -57,10 +57,6 @@ const SafeAppLanding = ({ appUrl, chain }: Props) => {
     )
   }
 
-  if (!safeApp) {
-    return <div>No Safe App found</div>
-  }
```
katspaugh, line 62: Porque?

schmanu: `safeApp` can not be `null` / `undefined` because we always return `getEmptySafeApp` inside of `useSafeAppFromManifest`

katspaugh (approved):
Nice clean up. You seem to have correctly replaced the places that actually did a numerical comparison as well as the ones that did a boolean check.

I think we can merge this after running regression tests.

schmanu:
Triggered the Regression tests manually: https://github.com/safe-global/safe-wallet-web/actions/runs/7919519093

francovenica:
I'd be ok with this ticket being merged in dev if the automation regression passed. I'll test it during the regression in the next RC

francovenica:
We decided to leave this ticket for the next release. So we will test it in this PR first and then pass it to be merged to dev. Keep it in  QA for now

katspaugh:
Launched the regression suite: https://github.com/safe-global/safe-wallet-web/actions/runs/8020554489

katspaugh:
@francovenica I suggest we test just one thing manually: replacing a transaction with nonce 0 with another transaction.

francovenica:
Checked the replacement of a tx with nonce 0.

Also I've checked the files changed and tried several things:
Send funds signing/execution
Tx proposal in general
Assets
Pagination of queue and history
Notifications badges (the counters)
Amount inputs in different forms
Nonce inputs
Owner revomal (reseting threshold)
Safe apps opening, signing and execution (just tx builder)
Message signing
Adv parameters
Relayer
Tenderly simulation
Recovery
....

I had no issues.

PR #3267, opened 2024-02-15
Fix: don't load social login if no env var is set

compojoom:
## What it solves
grafik

Resolves #
We should be able to start the app by just providing the infura token and nothing else. 

## How this PR fixes it
If we haven't specified any social_wallet_options, the social login is not enabled. 

## How to test it
1. Clone the PR
2. Create an .env file with just 
```
NEXT_PUBLIC_INFURA_TOKEN=YOUR_KEY
NEXT_PUBLIC_SAFE_APPS_INFURA_TOKEN=YOUR_KEY
``` 

You shouldn't get an error now when you run `yarn dev`

## Checklist


katspaugh (approved):
🙏

PR #3268, opened 2024-02-15
fix: don't show address suggestions after a valid address has been entered

### LINKED ISSUE #3252: Unrelated address book entries are shown when there's already a valid address in the input
## Bug description
I was transacting yesterday and noticed somewhat (in my opinion) suboptimal behaviour. I have already entered an address in the input field. However, when I click on the input field, it displays unrelated addresses from the address book. This increases the possibility of an error, as a potential misclick could result in transferring all the funds to an address from the address book.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Create a transaction that transfers any asset
 2. Enter destination address
 3. Click on the address input

## Expected result

I do not know what the expected result should be. Maybe not display address book entries at all if there's a valid address, perhaps display similar ones, or maybe add a separate toggle button for the address book; I will leave this to the team to decide.

## Obtained result
Clicking on the input displays unrelated address book entries, increasing the surface of a devastating mistake.

## Screenshots
SCR-20240213-qaxj

jmealy:
## What it solves

Resolves #3252 

## How this PR fixes it
Check if the current address input is a valid address. If it is then do not show any suggestions from the address book.

## How to test it
- Create a transaction that transfers any asset
- Enter destination address (something not in the address book)
- Click on the address input
- The autocomplete should not show any suggestions.

### DIFF
--- a/src/components/common/AddressBookInput/index.tsx
+++ b/src/components/common/AddressBookInput/index.tsx
@@ -10,6 +10,7 @@ import InfoIcon from '@/public/images/notifications/info.svg'
 import EntryDialog from '@/components/address-book/EntryDialog'
 import css from './styles.module.css'
 import inputCss from '@/styles/inputs.module.css'
+import { isValidAddress } from '@/utils/validation'
 
 const abFilterOptions = createFilterOptions({
   stringify: (option: { label: string; name: string }) => option.name + ' ' + option.label,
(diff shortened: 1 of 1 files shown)


mmv08 (approved):
looks sound mate

liliya-soroka:
Current result:
1. The enter field is cleaned after the click 
2. no drop-down with address book suggestions 
https://github.com/safe-global/safe-wallet-web/assets/338622/754c13f6-ee41-4fec-89c5-df03d25b0e38

jmealy:
> Screen.Recording.2024-02-16.at.13.00.25.mov 
> Current result:
> 
> 1. The enter field is cleaned after the click
> 2. no drop-down with address book suggestions
> 
> Is the enter field cleaning expected?

It is maintaining current behaviour so it's like this in prod already. I suggest handling this as a separate issue (but personally I think that this behaviour is fine)

francovenica:
LGTM

Is notisable when you copy/paste an address that is not in the address book already

Also I checked that, if the address is wrong or invalid, it shows the list of elments back to pick another address (which I think is the correct behavior)
image

PR #3269, opened 2024-02-15
Refactor: don't poll data in background tabs

### LINKED ISSUE #3125: Don't poll safe info in a backgroudn tab
Safe Info is fetched every 15 seconds even if the app is idle in a background tab.
This can be easily fixed by wrapping the call in a requestAnimationFrame which automatically pauses in a background window.

katspaugh:
## What it solves

Resolves #3125

## How this PR fixes it

I've wrapped all `useIntervalCounter` callbacks into a requestAnimationFrame which [isn't triggered in background/minimized tabs](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame).

`useIntervalCounter` is used for polling, so this change will optimize the polling of Safe Info, Balaces, Gas Price etc.


compojoom (approved):
that's awesome!

katspaugh:
I've tested it myself as it was quite trivial, so I'll merge it directly.

PR #3274, opened 2024-02-16
Add connect button to owned safe list when no wallet is connected

jmealy:
## What it solves

https://www.notion.so/My-accounts-Empty-state-b92fff61cd454471bb9af8cc18c589bd?pvs=4

## How this PR fixes it
Adds a state to the owned safes list for when a wallet is not connected. Includes a button for connecting a wallet

### DIFF
--- a/src/components/welcome/MyAccounts/PaginatedSafeList.tsx
+++ b/src/components/welcome/MyAccounts/PaginatedSafeList.tsx
@@ -7,11 +7,11 @@ import Track from '@/components/common/Track'
 import { OVERVIEW_EVENTS } from '@/services/analytics'
 import css from './styles.module.css'
 import useWallet from '@/hooks/wallets/useWallet'
-import useConnectWallet from '@/components/common/ConnectWallet/useConnectWallet'
 
 type PaginatedSafeListProps = {
   safes: SafeItems
   title: ReactNode
+  noSafesMessage?: ReactNode
   action?: ReactElement
   onLinkClick?: () => void
 }
@@ -20,10 +20,9 @@ const DEFAULT_SHOWN = 5
 const MAX_DEFAULT_SHOWN = 7
 const PAGE_SIZE = 5
 
-const PaginatedSafeList = ({ safes, title, action, onLinkClick }: PaginatedSafeListProps) => {
+const PaginatedSafeList = ({ safes, title, action, noSafesMessage, onLinkClick }: PaginatedSafeListProps) => {
   const [maxShownSafes, setMaxShownSafes] = useState<number>(DEFAULT_SHOWN)
   const wallet = useWallet()
-  const handleConnect = useConnectWallet()
 
   const shownSafes = useMemo(() => {
     if (safes.length <= MAX_DEFAULT_SHOWN) {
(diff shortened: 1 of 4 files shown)


src/components/welcome/MyAccounts/PaginatedSafeList.tsx
```diff
@@ -46,8 +50,15 @@ const PaginatedSafeList = ({ safes, title, action, onLinkClick }: PaginatedSafeL
       </div>
       {shownSafes.length ? (
         shownSafes.map((item) => <AccountItem onLinkClick={onLinkClick} {...item} key={item.chainId + item.address} />)
+      ) : !wallet ? (
+        <Typography variant="body2" color="text.secondary" textAlign="center" my={3} mx="auto" width={250}>
+          Connect a wallet to view your Safe Accounts or to create a new one
+          <Button onClick={handleConnect} disableElevation size="small" sx={{ mt: 2 }}>
+            Connect a wallet
+          </Button>
+        </Typography>
```
katspaugh, line 59: Should it be here or perhaps in the parent component?

jmealy: you mean replace the 'create account' button with 'connect wallet' button when the wallet is not connected?
True that it doesn't make sense to have a create account option if you're not connected. All it does is redirect to the welcome page

katspaugh: I meant just architecturally that it would be better to keep only the list iself in the PaginatedSafeList. But your comment makes even more sense.

jmealy: ok I changed it to pass the empty message as a prop from the parent

katspaugh (approved):

src/components/welcome/MyAccounts/PaginatedSafeList.tsx
```diff
@@ -18,8 +20,9 @@ const DEFAULT_SHOWN = 5
 const MAX_DEFAULT_SHOWN = 7
 const PAGE_SIZE = 5
 
-const PaginatedSafeList = ({ safes, title, action, onLinkClick }: PaginatedSafeListProps) => {
+const PaginatedSafeList = ({ safes, title, action, noSafesMessage, onLinkClick }: PaginatedSafeListProps) => {
   const [maxShownSafes, setMaxShownSafes] = useState<number>(DEFAULT_SHOWN)
+  const wallet = useWallet()
```
katspaugh, line 25: This can be deleted now, I think.

src/components/welcome/MyAccounts/index.tsx
```diff
@@ -30,7 +32,19 @@ const AccountsList = ({ safes, onLinkClick }: AccountsListProps) => {
...
+        <PaginatedSafeList
+          title="My accounts"
+          safes={ownedSafes}
+          onLinkClick={onLinkClick}
+          noSafesMessage={
+            <>
+              Connect a wallet to view your Safe Accounts or to create a new one
+              <Button onClick={handleConnect} disableElevation size="small" sx={{ mt: 2 }}>
+                Connect a wallet
+              </Button>
```
katspaugh, line 44: There's a `ConnectWallet` button component, it can be extended to allow custom text.

katspaugh (approved):

PR #3276, opened 2024-02-16
fix: Set Safe apps categories via URL query

usame-algan:
## What it solves

Resolves #

This is an enhancement for the safe apps list and allows to set categories via the URL. It will be for linking to certain categories like Stake, Bridge, Swap e.g. from the Dashboard.

## How this PR fixes it

- Adds a new hook `useCategoryFilter` that checks the URL for `categories` and syncs it with the local state

## ToDo

- [x] Write tests for `useCategoryFilter`

## How to test it

1. Create a counterfactual Safe
2. Press Create Transaction on the dashboard
3. Select Swap tokens
4. Observe being navigated to the safe apps list with `DEX` apps pre-selected
5. Clear the category
6. Observe the URL and list updating

## Screenshots
Screenshot 2024-02-16 at 15 58 41

## Checklist


src/features/counterfactual/FirstTxFlow.tsx
```diff
@@ -74,8 +74,7 @@ const FirstTxFlow = ({ open, onClose }: { open: boolean; onClose: () => void })
   }
 
   const onSwap = () => {
-    // TODO: Pre-filter DeFi category apps
-    router.push({ pathname: AppRoutes.apps.index, query: router.query })
+    router.push({ pathname: AppRoutes.apps.index, query: { ...router.query, categories: 'DEX' } })
```
katspaugh, line 77: Could you add DEX to the SafeAppsTag in constants.ts?

src/hooks/safe-apps/useQueryFilter.ts
```diff
@@ -0,0 +1,57 @@
+import { type Dispatch, type SetStateAction, useEffect } from 'react'
+import { useRouter } from 'next/router'
```
katspaugh, line 2:
```suggestion
import { useSearchParams } from 'next/navigation'
```

src/hooks/safe-apps/useQueryFilter.ts
```diff
@@ -0,0 +1,57 @@
...
+import { getCategoryOptions } from '@/components/safe-apps/SafeAppsFilters'
+import type { SafeAppData } from '@safe-global/safe-gateway-typescript-sdk'
+
+// TODO: Add search and batch filter as well
```
katspaugh, line 6: Please don't add todos that will never be done.

src/hooks/safe-apps/useQueryFilter.ts
```diff
@@ -0,0 +1,57 @@
...
+      setSelectedCategories(categoryQueryOptions)
+    }
+  }, [router.isReady, router.query.categories, safeAppsList, selectedCategories.length, setSelectedCategories])
+
+  const onSelectCategories = (categories: string[]) => {
+    setSelectedCategories(categories)
+
+    // Clean up the URL
+    if (categories.length === 0) {
+      delete router.query.categories
```
katspaugh, line 39: No bueno to use destructive operations like this.

src/hooks/safe-apps/useQueryFilter.ts
usame-algan: In that case I would suggest we rename this hook to something like `useCategoryFilter`

I could see it in the future that we want to link to a specific search query or with apps pre-selected that support batching.

src/hooks/safe-apps/useQueryFilter.ts
usame-algan: How do we navigate without the router though?

katspaugh: Ah, you also use it to change the route. Nvm then :)

src/hooks/safe-apps/useQueryFilter.ts
katspaugh: `useCategoryFilter` sounds much clearer to me.

To clarify what I meant: code comments are not the right place for feature ideas and it's not a backlog.

src/features/counterfactual/FirstTxFlow.tsx
usame-algan: I think DEX is not the right tag (it only exists on staging). We should probably use DeFi or Aggregator. Afaiu adding it to the SafeAppsTags filters it out in `getCategoryOptions` which we need to check if its a valid tag and not some random string.

src/features/counterfactual/FirstTxFlow.tsx
```diff
@@ -74,8 +74,7 @@ const FirstTxFlow = ({ open, onClose }: { open: boolean; onClose: () => void })
   }
 
   const onSwap = () => {
-    // TODO: Pre-filter DeFi category apps
-    router.push({ pathname: AppRoutes.apps.index, query: router.query })
+    router.push({ pathname: AppRoutes.apps.index, query: { ...router.query, categories: 'Aggregator' } })
```
katspaugh, line 77: Add `Aggregator` to the enum?

katspaugh (approved):

liliya-soroka:
1. Need to delete category twice 

https://github.com/safe-global/safe-wallet-web/assets/338622/7ed6aa55-b2ec-4cf4-8a0a-4de180a15093

src/hooks/safe-apps/useQueryFilter.ts
usame-algan: I had to revert back to this because just omitting the key from router.query caused it to still be there for the next rerender which updated the state again in the useEffect.

usame-algan:
> Need to delete category twice

Can you check again? I pushed a fix

src/hooks/safe-apps/useQueryFilter.ts
usame-algan: Actually got it working by awaiting the router.push before updating the local state.

liliya-soroka:
Verified

PR #3277, opened 2024-02-16
Refactor: remove Show Prefix setting

katspaugh:
## What it solves

Removes the "Show chain prefix" setting. There's no practical difference in whether prefixes are shown or not. It's beneficial to always display them.

The usage was very low, too.

Screenshot 2024-02-16 at 16 48 28

## Screenshots
Screenshot 2024-02-16 at 16 45 57

## How to test it

* Go to the Settings -> Appearance
* Observe that the setting is gone


src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
```diff
@@ -35,8 +35,8 @@ const SrcEthHashInfo = ({
   address,
   customAvatar,
   prefix = '',
-  copyPrefix,
-  showPrefix,
+  copyPrefix = true,
+  showPrefix = true,
```
jmealy, line 39: Why not remove this entirely?

jmealy (changes requested):
Looks like tests need to be updated

src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
katspaugh: I thought there might be some use cases where you want to display just the hash, maybe the prefix is rendered externally or not at all necessary (e.g. a transaction hash).

jmealy (approved):

PR #3278, opened 2024-02-16
Show undeployed safes under owned safes

jmealy:
Before, undeployed safes were being shown under the watchlist. Show them under owned safes instead.

## How to test it
- create a safe using the option to pay later.
- after safe creation, the safe should be shown in the owned safes section of the sidebar and the /accounts page


katspaugh (approved):
Neat!

PR #3282, opened 2024-02-19
Fix: move Linea bridge to warned bridges

### LINKED ISSUE #3078: Unsupported chain bridge: change error from red to yellow
Currently when a user tries to bridge funds to a Safe on an unsupported network we block access to the bridge dApp. Suggesiton is to change the warning from red to yellow and allow with precaution.

Reason: there are unofficial Safes like on Linea, the issue was risen by Consensys team. 

More context: https://5afe.slack.com/archives/C060KN6CMJB/p1701874982283519

katspaugh:
## What it solves

Resolves #3078

## How this PR fixes it

I've moved the Linea bridge from the red unsupported bridges to the yellow warned bridges list.

### DIFF
--- a/src/features/walletconnect/constants.ts
+++ b/src/features/walletconnect/constants.ts
@@ -51,7 +51,6 @@ export const BlockedBridges = [
   // Unsupported chain bridges
   'bridge.zora.energy',
   'bridge.mantle.xyz',
-  'bridge.linea.build',
   'bridge.metis.io',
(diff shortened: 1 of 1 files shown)


usame-algan (approved):

usame-algan:
Can confirm that I see an orange warning now and it lets me connect my safe to the bridge. I couldn't test bridging because of insufficient funds.

PR #3285, opened 2024-02-19
Redirect form /welcome page when wallet is connected

jmealy:
Redirect from the welcome page depending on wether the user has any safes:
- If the user has either owned or added safes, redirect to the accounts page `/welcome/accounts`
- If the user has no safes to show, redirect them to create a new safe page `/new-safe/create`

## How to test it
On the `/welcome` page:
- Connect an account with owned safes, but with no added safes in the browser (private window)
  - should redirect to `/welcome/accounts`
- Connect an account with no owned safes (fresh MM account), but with added safes in the browser
  - should redirect to `/welcome/accounts`
- Connect an account with no owned safes, and no added safes
  - should redirect to `/new-safe/create`

### DIFF
--- a/src/components/welcome/MyAccounts/useAllSafes.ts
+++ b/src/components/welcome/MyAccounts/useAllSafes.ts
@@ -25,9 +25,14 @@ const useAddedSafes = () => {
 export const useHasSafes = () => {
   const { address = '' } = useWallet() || {}
   const allAdded = useAddedSafes()
-  const hasAdded = isEmpty(allAdded)
+  const hasAdded = !isEmpty(allAdded)
   const [allOwned] = useAllOwnedSafes(!hasAdded ? address : '') // pass an empty string to not fetch owned safes
-  return hasAdded || !isEmpty(allOwned)
+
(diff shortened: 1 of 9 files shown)


katspaugh (approved):

src/components/welcome/MyAccounts/useAllSafes.ts
```diff
@@ -25,9 +25,14 @@ const useAddedSafes = () => {
 export const useHasSafes = () => {
   const { address = '' } = useWallet() || {}
   const allAdded = useAddedSafes()
-  const hasAdded = isEmpty(allAdded)
+  const hasAdded = !isEmpty(allAdded)
```
katspaugh, line 28: Good catch!

katspaugh:
I'm connected but it says Connect wallet.

Screenshot 2024-02-20 at 07 41 33

katspaugh:
After testing it a bit, I think we shouldn't redirect from Welcome on page load. It can be surprising/confusing.

We should instead redirect after:

* The user explicitely connects
* The user clicks on the "Continue with Metamask" button

katspaugh:
Tested, looks good!

PR #3292, opened 2024-02-20
Feat: New transaction -> Add to watchlist for readonly Safes

katspaugh:
## What it solves

If an opened Safe is not owned, suggest to add it to the watchlist instead of creating a transaction.

Screenshot 2024-02-20 at 12 25 16

Also, I've renamed "Rename" to "Give name" for unnamed Safes.


src/components/sidebar/WatchlistAddButton/index.tsx
```diff
@@ -0,0 +1,40 @@
+import Button from '@mui/material/Button'
+import { OVERVIEW_EVENTS, trackEvent } from '@/services/analytics'
+import { useRouter } from 'next/router'
+import { AppRoutes } from '@/config/routes'
+import { useCurrentChain } from '@/hooks/useChains'
+import useSafeAddress from '@/hooks/useSafeAddress'
+
+const WatchlistAddButton = () => {
```
jmealy, line 8: Could reuse this in the MyAccounts component

jmealy (changes requested):
The button doesn't seem to go away after adding the safe

jmealy (approved):

PR #3295, opened 2024-02-20
fix: [RC] issues

usame-algan:
## What it solves

When switching from an undeployed safe we run a `useEffect` to check `checkDeploymentStatus` which contains the updated `provider` but can still contain the previous safeInfo. In that case it can happen that the undeployed state is wrongfully removed.

## How this PR fixes it

Make sure that the `provider` and the `safeInfo` have the same chainId before checking if a contract is deployed on the safeAddress

## How to test it

1. Create a counterfactual safe with address X on network A
2. Deploy that safe
3. Create a counterfactual safe on network B and make sure that the safe ends up with the same address X
4. Switch back to network A
5. Switch back to the network B
6. Observe the counterfactual safe on network B is still accessible

---

1. Create a counterfactual safe on Celo
2. Press on "Create transaction" on the dashboard
3. Observe no "Set up recovery" option
4. Create a counterfactual safe on Mainnet
5. Observe there is a "Set up recovery" option

---

1. Open any safe that is already deployed in Firefox
2. Observe the overview buttons on the dashboard are aligned in a row

## Screenshots
Screenshot 2024-02-20 at 16 08 44
Screenshot 2024-02-20 at 16 14 41

## Checklist

### DIFF
--- a/src/features/counterfactual/hooks/usePendingSafeStatuses.ts
+++ b/src/features/counterfactual/hooks/usePendingSafeStatuses.ts
@@ -86,7 +86,8 @@ const usePendingSafeStatus = (): void => {
 
     const checkDeploymentStatus = async () => {
       // In case the safe info hasn't been updated yet when switching safes
-      if ((await provider.getNetwork()).chainId !== BigInt(safe.chainId)) return
+      const { chainId } = await provider.getNetwork()
+      if (chainId !== BigInt(safe.chainId)) return
 
       const isContractDeployed = await isSmartContract(provider, safeAddress)
 
--- a/src/features/counterfactual/hooks/usePendingSafeStatuses.ts
+++ b/src/features/counterfactual/hooks/usePendingSafeStatuses.ts
@@ -85,6 +85,9 @@ const usePendingSafeStatus = (): void => {
     if (!provider || !safeAddress) return
 
     const checkDeploymentStatus = async () => {
+      // In case the safe info hasn't been updated yet when switching safes
+      if ((await provider.getNetwork()).chainId !== BigInt(safe.chainId)) return
+
       const isContractDeployed = await isSmartContract(provider, safeAddress)
 
       if (isContractDeployed) {
--- a/src/components/common/BuyCryptoButton/index.tsx
+++ b/src/components/common/BuyCryptoButton/index.tsx
@@ -80,6 +80,7 @@ const _BuyCryptoButton = ({ href, pagePath }: { href?: LinkProps['href']; pagePa
             sx={buttonStyles}
             startIcon={<AddIcon />}
             className={css.buyCryptoButton}
+            fullWidth
           >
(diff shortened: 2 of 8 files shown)


src/features/counterfactual/hooks/usePendingSafeStatuses.ts
```diff
@@ -85,6 +85,9 @@ const usePendingSafeStatus = (): void => {
     if (!provider || !safeAddress) return
 
     const checkDeploymentStatus = async () => {
+      // In case the safe info hasn't been updated yet when switching safes
+      if ((await provider.getNetwork()).chainId !== BigInt(safe.chainId)) return
```
katspaugh, line 89: Nit pick: this is a bit hard to read.

```suggestion
      const { chainId } = await provider.getNetwork()
      if (chainId !== BigInt(safe.chainId)) return
```

katspaugh (approved):

francovenica:
I tried the first part.

Create a CF in networkA with AddressX. Deploy it. Create a CF in networkB and make sure it has the same addressX.

This already didn't happened to me. Both safes have different address. 
I tried switching between sepolia and ETH (let me know if should be other specific networks to test this)
I tried not deploying the first safe, but same result: both safe have different addresses.

@liliya-soroka Can you give it a try to this?

usame-algan:
> I tried the first part.
> 
> 
> 
> Create a CF in networkA with AddressX. Deploy it. Create a CF in networkB and make sure it has the same addressX.
> 
> 
> 
> This already didn't happened to me. Both safes have different address. 
> 
> I tried switching between sepolia and ETH (let me know if should be other specific networks to test this)
> 
> I tried not deploying the first safe, but same result: both safe have different addresses.
> 
> 
> 
> @liliya-soroka Can you give it a try to this?
> 
> 

You most likely already have that safe deployed either on network A or B so the saltNonces are different. I suggest to create a new EOA in MM and try with that since the saltNonces should be 0 on both networks then.

francovenica:
Yeah, I had to create a brand new account in MM. Tried again and now I was able to create 2 safes with the same address and switch between them just fine:
same address safe

francovenica:
The buttons in Firefox:
If the button "Buy token" is there, then it looks fine, but if that button is not there (because that network doesn't support buying tokens) then it doesn't look good:
image
image

francovenica:
The "recovery" option looks good: shows when it supposed to, is hidden where is not

usame-algan:
> If the button "Buy token" is there, then it looks fine, but if that button is not there (because that network doesn't support buying tokens) then it doesn't look good:

I pushed another fix for this.

> Switching safe in the sidebar shows an notification error:

I had this happen to me too but now I can't reproduce it anymore.

francovenica:
The buttons in firefox was fixed

I don't get the error notification anymore. 

LGTM

PR #3296, opened 2024-02-20
fix: Check all address books to find name for an address

usame-algan:
## What it solves

Showing a name for an address via the `EthHashInfo` component uses the `useAddressBook` hook. This works well when on a specific safe view because there is a unique chain to check. However, on non-safe routes like the global notifications this doesn't work and can result in showing the wrong name for an address in case it has the same address on multiple chains.

## How this PR fixes it

- Adds a new hook `useAllAddressBooks`
- Uses that hook instead of `useAddressBook` to find a name in `EthHashInfo`

## How to test it

1. Create 2 safes with identical addresses on 2 different chains e.g. with counterfactual
2. Deploy both safes
3. Make sure both safe addresses are in the address book with different names
4. Go to the global notifications page
5. Observe both addresses show up with their correct name

## Screenshots
Screenshot 2024-02-20 at 17 23 32

## Checklist

### DIFF
--- a/src/tests/builders/chains.ts
+++ b/src/tests/builders/chains.ts
@@ -27,8 +27,8 @@ const rpcUriBuilder = (): IBuilder<RpcUri> => {
 
 const blockExplorerUriTemplateBuilder = (): IBuilder<BlockExplorerUriTemplate> => {
   return Builder.new<BlockExplorerUriTemplate>().with({
-    address: faker.finance.ethereumAddress(),
-    txHash: faker.string.hexadecimal(),
+    address: faker.internet.url({ appendSlash: false }),
+    txHash: faker.internet.url({ appendSlash: false }),
     api: faker.internet.url({ appendSlash: false }),
   })
 }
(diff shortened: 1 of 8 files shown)


src/tests/builders/chains.ts
```diff
@@ -27,8 +27,8 @@ const rpcUriBuilder = (): IBuilder<RpcUri> => {
 
 const blockExplorerUriTemplateBuilder = (): IBuilder<BlockExplorerUriTemplate> => {
   return Builder.new<BlockExplorerUriTemplate>().with({
-    address: faker.finance.ethereumAddress(),
-    txHash: faker.string.hexadecimal(),
+    address: faker.internet.url({ appendSlash: false }),
+    txHash: faker.internet.url({ appendSlash: false }),
     api: faker.internet.url({ appendSlash: false }),
```
usame-algan, line 32: All of these are urls so I fixed it here. Seems to be a coincidence that this didn't break any tests yet.

francovenica:
LGTM:
image

PR #3305, opened 2024-02-21
fix: [RC] Add pay later hint to safe creation step

usame-algan:
## What it solves

- Adds a hint to the owner step during safe creation informing the user that pay later is only supported for single owner setups
- Changes the 1/1 warning to an info box to make it less scary
- Adds a toggle option to the QR code in the first steps widget
- Removes the wrong chain warning on the owner step because the user is redirected away if their wallet has a different network so they will never see that warning

## How to test it

1. Go through the safe creation flow
2. Observe in the owner step that the 1/1 warning is now blue
3. Observe there is no pay later hint
4. Add another owner
5. Observe there is a pay later hint
6. Create the safe counterfactually
7. Press Add funds on the dashboard
8. Observe a toggle option under the QR code

## Screenshots
Screenshot 2024-02-21 at 15 15 02
Screenshot 2024-02-21 at 15 17 18

## Checklist

### DIFF
--- a/src/features/counterfactual/CounterfactualHint.tsx
+++ b/src/features/counterfactual/CounterfactualHint.tsx
@@ -1,0 +1,12 @@
+import { Alert } from '@mui/material'
+
+const CounterfactualHint = () => {
+  return (
+    <Alert severity="info" sx={{ mt: 2 }}>
+      Please note that Safe Accounts with more than one owner will have to be activated with a network fee right away.
+      You can add more owners after Safe creation.
+    </Alert>
+  )
+}
+
+export default CounterfactualHint
--- a/src/features/counterfactual/CounterfactualHint.tsx
+++ b/src/features/counterfactual/CounterfactualHint.tsx
@@ -1,10 +1,13 @@
-import { Alert } from '@mui/material'
+import { Alert, Typography } from '@mui/material'
 
 const CounterfactualHint = () => {
   return (
     <Alert severity="info" sx={{ mt: 2 }}>
-      Please note that Safe Accounts with more than one owner will have to be activated with a network fee right away.
-      You can add more owners after Safe creation.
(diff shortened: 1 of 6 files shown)


francovenica:
The things in the description of the PR look good.

We have to check why automation fails
I would suggest that the message about not paying fees in a 1/1 safe should sound like an invitation to try the feature and not a warning:
"If set your safe with a policy of 1, you will be able to try the safe experience without having to pay a gas fee. You can add more owners later"

The prefix thing is only for the QR code? I can still copy the address below and it's going to have the prefix all the time. If that is out of the scope of the change let me know and I'll ignore it

usame-algan:
> I would suggest that the message about not paying fees in a 1/1 safe should sound like an invitation to try the feature and not a warning:

Thanks, I adjusted the message like discussed. Also adjusted the 1/1 and 1/n hint wording

katspaugh (approved):

src/features/counterfactual/CounterfactualHint.tsx
```diff
@@ -0,0 +1,15 @@
+import { Alert, Typography } from '@mui/material'
+
+const CounterfactualHint = () => {
+  return (
+    <Alert severity="info" sx={{ mt: 2 }}>
+      <Typography fontWeight="bold" mb={1}>
+        Create now, pay later! Only for 1/1 Accounts
+      </Typography>
+      Explore Safe{'{Wallet}'} without additional hassle. Pay for the creation later at any time, and don&apos;t forget
```
katspaugh, line 9: The only thing I would change:
`Pay for the creation` -> `Pay the gas fees`
Otherwise it may seem as if you have to pay us.

francovenica:
For me the ticket is fine

We gotta see why the test are failing

PR #3307, opened 2024-02-21
Show remove from watchlist option when viewing a watchlist safe

jmealy:
## What it solves

Resolves https://www.notion.so/safe-global/570f3459753047f49022d5b63ee1e57e?v=ed82ae2a7e28401086a736f6af1c31ae&p=0fcb3a7e98424536ad271f4512112594&pm=s

## How this PR fixes it
When viewing a safe that is in the watchlist, show an option to remove it, in place of the add to watchlist button

Also changes the appearance of the add to watchlist button in the sidebar

## Screenshots
image

image

### DIFF
--- a/src/components/sidebar/WatchlistAddButton/index.tsx
+++ b/src/components/sidebar/WatchlistAddButton/index.tsx
@@ -1,14 +1,24 @@
-import Button from '@mui/material/Button'
 import { OVERVIEW_EVENTS, trackEvent } from '@/services/analytics'
 import { useRouter } from 'next/router'
 import { AppRoutes } from '@/config/routes'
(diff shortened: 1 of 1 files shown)


francovenica:
Looks good to me

The Add to watchlist and remove from watchlist buttons change depending on the current status of the safe
For owners and non-owner with an allowance set they still see the "New transaction" button just fine.

katspaugh (approved):

usame-algan:
One thing I noticed is that when loading a safe while auto-connecting a wallet the button now goes through all 3 variants first showing "Add to watchlist" then "Remove from watchlist" and lastly "New transaction". It would be good to add a loading skeleton there. I will file an issue

PR #3308, opened 2024-02-21
fix: show newly depoyed safes under owned safes

jmealy:
## What it solves

Resolves https://www.notion.so/safe-global/570f3459753047f49022d5b63ee1e57e?v=ed82ae2a7e28401086a736f6af1c31ae&p=bc70bcc7233d488e84e2d4dd96c9bcf9&pm=s

## How this PR fixes it
Newly created safes take around 5 minutes to show up in responses from the CGW. As a workaround, display newly created safes in the owned safes list by referencing the owners array in localstorage.

### DIFF
--- a/src/components/welcome/MyAccounts/useAllSafes.ts
+++ b/src/components/welcome/MyAccounts/useAllSafes.ts
@@ -36,8 +36,8 @@ export const useHasSafes = () => {
 }
 
 const useAllSafes = (): SafeItems => {
-  const { address = '' } = useWallet() || {}
-  const [allOwned = {}] = useAllOwnedSafes(address)
+  const { address: walletAddress = '' } = useWallet() || {}
+  const [allOwned = {}] = useAllOwnedSafes(walletAddress)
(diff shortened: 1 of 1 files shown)


src/components/welcome/MyAccounts/useAllSafes.ts
```diff
@@ -54,8 +54,10 @@ const useAllSafes = (): SafeItems => {
       const uniqueAddresses = uniq(addedOnChain.concat(ownedOnChain)).filter(Boolean)
 
       return uniqueAddresses.map((address) => {
+        const owners = allAdded?.[chainId]?.[address]?.owners
+        const isOwner = owners?.map((address) => address.value).includes(walletAddress)
```
katspaugh, line 58:
```suggestion
        const isOwner = owners?.some(({ value }) => sameAddress(walletAddress))
```

There can be a mismatch in lowercase vs checksummed address format otherwise.

liliya-soroka:
Verified and can be added to the RC 1.30.0

PR #3316, opened 2024-02-22
fix: wc "no matching key" errors

### LINKED ISSUE #3284: WalletConnect: no matching session key
## Bug description

* Connect a Safe to another Safe that the first one owns via WC
* Wait... (it can be slow – investigate)
* Once Approve/Rejects appears, press Approve
* Instead of connecting, it shows a "no matching session key" error

This is somehow related to timing and doesn't happen if you wait a bit before pressing Approve.

compojoom:
## What it solves

Resolves #3284

This is an attempt to fix 2 WC bugs that are plaguing some of our users:

No matching key: session
No matching key: keychain

For the keychain error I haven't figured a way how to reproduce it, but it appears that WC has fix in their latest release. For the session error I could reproduce it when connecting to 1inch, but also very edge case and I doubt that users are encountering it that way. We have a lot of errors in sentry and I can't imagine that users trigger it like this.

## How to test it
For the session error you could try the PR and follow the same steps as in the gif. The problem is that the fix is flaky - sometimes it works, sometimes it doesn't, where without it, it always fails. 

For the keychain issue, no idea. We'll have to monitor sentry for that error.

### DIFF
--- a/src/features/walletconnect/components/WcSessionMananger/index.tsx
+++ b/src/features/walletconnect/components/WcSessionMananger/index.tsx
@@ -108,6 +108,7 @@ const WcSessionManager = ({ sessions, uri }: WcSessionManagerProps) => {
     return walletConnect.onSessionPropose((proposalData) => {
       setError(null)
       setProposal(proposalData)
+      isResponded.current = false
     })
   }, [walletConnect, setError])
 
--- a/src/features/walletconnect/components/WcSessionMananger/index.tsx
+++ b/src/features/walletconnect/components/WcSessionMananger/index.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useContext, useEffect, useState } from 'react'
+import { useCallback, useContext, useEffect, useRef, useState } from 'react'
 import type { Web3WalletTypes } from '@walletconnect/web3wallet'
 import type { SessionTypes } from '@walletconnect/types'
 import useSafeInfo from '@/hooks/useSafeInfo'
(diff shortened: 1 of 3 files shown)


src/features/walletconnect/components/WcSessionMananger/index.tsx
```diff
@@ -25,11 +25,14 @@ const WcSessionManager = ({ sessions, uri }: WcSessionManagerProps) => {
   const { walletConnect, error, setError, open, setOpen } = useContext(WalletConnectContext)
   const [proposal, setProposal] = useState<Web3WalletTypes.SessionProposal>()
   const [changedSession, setChangedSession] = useState<SessionTypes.Struct>()
+  const isResponded = useRef<boolean>(false)
```
katspaugh, line 28: Do we need to reset this back to false when the proposal changs? Or is the entire component re-rendering then?

src/features/walletconnect/components/WcSessionMananger/index.tsx
```diff
@@ -62,6 +66,21 @@ const WcSessionManager = ({ sessions, uri }: WcSessionManagerProps) => {
...
+  // but for whatever reason refreshes the browser without rejecting/accepting it
+  // if we don't do this we often get "No matching key: session"
+  useEffect(() => {
+    const handleBeforeUnload = async () => {
+      if (proposal && isResponded.current === false) {
+        await onReject()
+      }
+    }
+
+    window.addEventListener('beforeunload', handleBeforeUnload)
```
katspaugh, line 79: AFAIR beforeunload is supposed to return a boolean to prevent the browser from closing? Maybe we should use onunload?

katspaugh (commented):
Please adjust the title of the PR as per our convention.

src/features/walletconnect/components/WcSessionMananger/index.tsx
katspaugh: Thanks!

src/features/walletconnect/components/WcSessionMananger/index.tsx
compojoom: Good point! Fixed.

katspaugh (approved):

src/features/walletconnect/components/WcSessionMananger/index.tsx
katspaugh: Thanks!

Btw please don't resolve threads unless you started them so that the author can come back and see your response (and hopefully resolve it) 😄

PR #3322, opened 2024-02-23
feat: One-click add safes to watchlist

### LINKED ISSUE #3304: [Watchlist] One-click add Safe Accounts
## What is the feature about

It takes a multi-step form to add a safe to the watchlist. This is especially annoying when trying to add an already open safe to the watchlist. In that case the user is completely taken out of their current context and has to go through 3 steps in order to get back to where they were:

1. Input the safe address
2. Review the owners and change their names
3. Review the whole thing

For starters it would be good to offer a 1-click add option from inside opened safes. This would add the current safe to the list of added safes in the background.

## The list of requirements

- [ ] Enhance the `Add to watchlist` button to add a safe with one click

## Designs/sketches
Screenshot 2024-02-21 at 14 27 49

usame-algan:
## What it solves

Resolves #3304 

## How this PR fixes it

- Adds a new function `addSafeToWatchlist` that requires a `SafeInfo` object

## How to test it

1. Open any safe with a disconnected wallet
2. Click on the "Add to watchlist" button in the sidebar
3. Observe not leaving the page
4. Observe the safe is part of the watchlist and has a mnemonic name assigned
5. Remove the safe from watchlist
6. Observe the safe is removed from watchlist and from the address book

## Screenshots

https://github.com/safe-global/safe-wallet-web/assets/5880855/171596ca-fb1c-4679-bec2-e8a8d4c56605

## Checklist

### DIFF
--- a/src/components/new-safe/load/logic/index.ts
+++ b/src/components/new-safe/load/logic/index.ts
@@ -1,0 +1,43 @@
+import type { SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
+import type { AppDispatch } from '@/store'
+import { addOrUpdateSafe } from '@/store/addedSafesSlice'
+import { upsertAddressBookEntry } from '@/store/addressBookSlice'
+import { defaultSafeInfo } from '@/store/safeInfoSlice'
+
+export const addSafeToWatchlist = (dispatch: AppDispatch, safe: SafeInfo, safeName: string) => {
+  // Add safe to watchlist
+  dispatch(
+    addOrUpdateSafe({
+      safe: {
+        ...defaultSafeInfo,
+        address: { value: safe.address.value },
+        threshold: safe.threshold,
+        owners: safe.owners.map((owner) => ({
+          value: owner.value,
+          name: owner.name,
+        })),
+        chainId: safe.chainId,
+      },
+    }),
+  )
+
+  // Add safe to address book
+  dispatch(
+    upsertAddressBookEntry({
+      chainId: safe.chainId,
+      address: safe.address.value,
+      name: safeName,
+    }),
+  )
+
+  // Add all owners of the safe to address book
+  for (const { value } of safe.owners) {
+    dispatch(
+      upsertAddressBookEntry({
+        chainId: safe.chainId,
+        address: value,
+        name: `${safeName} owner`,
+      }),
+    )
+  }
+}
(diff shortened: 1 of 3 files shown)


usame-algan:
Not sure if we should add all owners to the address book or not since removing them when the safe is removed requires us to have the `SafeInfo` available. With the current add to watchlist feature we are also not cleaning them up.

src/components/new-safe/load/logic/index.ts
```diff
@@ -0,0 +1,43 @@
+import type { SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
+import type { AppDispatch } from '@/store'
+import { addOrUpdateSafe } from '@/store/addedSafesSlice'
+import { upsertAddressBookEntry } from '@/store/addressBookSlice'
+import { defaultSafeInfo } from '@/store/safeInfoSlice'
+
+export const addSafeToWatchlist = (dispatch: AppDispatch, safe: SafeInfo, safeName: string) => {
```
usame-algan, line 7: Ideally, I would like to reuse this for the add safe form but we don't have access to the `SafeInfo` there. We could create a function to build the `SafeInfo` out of the parameters though.

src/components/sidebar/WatchlistAddButton/index.tsx
```diff
@@ -1,34 +1,25 @@
...
   const [open, setOpen] = useState(false)
-  const router = useRouter()
-  const chain = useCurrentChain()
-  const address = useSafeAddress()
-  const chainId = chain?.chainId || ''
-  const addedSafes = useAppSelector((state) => selectAddedSafes(state, chainId))
-  const isInWatchlist = !!addedSafes?.[address]
+  const { safe } = useSafeInfo()
+  const dispatch = useAppDispatch()
+  const safeName = useMnemonicSafeName()
```
katspaugh, line 16: What if a Safe already has an actual name in the address book?

usame-algan: Do you think its still useful to add the safe and its owners to the address book now that the "add safe" feature has become a watchlist i.e. is supposed to be used only for non-owned safes?

In any case lets check the address book for a name before defaulting to the mnemonic.

katspaugh:
I'll close this for now to clean up the PR list, but the idea will be incorporated in the new Safe List iteration.

Issue #3321, opened 2024-02-23
[Recovery] Verify an email address


usame-algan:
## Requirements

- [x] There should be a 6 character CodeInput for the verification code
- [x]  It should be possible to resend the verification code
- [x] The component should be reusable in the setup recovery flow

## Screenshots

<img width="612" alt="Screenshot 2024-02-23 at 14 01 40" src="https://github.com/safe-global/safe-wallet-web/assets/5880855/ed005148-0eeb-431c-bf0a-7654a8b88d87">

usame-algan:
We can reuse the `CodeInput` and `CooldownLink` from the SMS OTP Branch: https://github.com/safe-global/safe-wallet-web/pull/2968/files#diff-4391498d03b80b5e75ad3c6faec862a75878220caf4269530d9628942c874db6R90

liliya-soroka:
Verified
a card was added to the Recovery v2 QA board

usame-algan:
Closing this as the project was removed from the roadmap

PR #3324, opened 2024-02-23
feat: [Recovery] Add register email form

### LINKED ISSUE #3320: [Recovery] Register an email address
## Requirements

- [x] There should be an input on the settings page for email address
- [x] Registering an email address is only possible for safe owners

## Screenshots
Screenshot 2024-02-23 at 14 00 04

usame-algan:
## What it solves

Resolves #3320

## How this PR fixes it

- Adds a form to enter an email address and Cancel and Continue buttons
- On Continue, registers the email
- On Cancel, hides the form and displays the "Sign to view" button again

## How to test it

1. Open a safe with recovery enabled
2. Go to the settings page
3. Observe the "Sign to view" button is disabled if no wallet is connected or wallet is not an owner
4. Press "Sign to view" and sign the message
5. An email form should be visible
6. Press cancel
7. The "Sign to view" button should be visible again
8. Do Step 4 again
9. Enter a valid email and submit
10. The email address should be visible

Note: With the addition of #3321, Step 9 will open a new dialog to verify the email.

## Screenshots
Screenshot 2024-02-23 at 16 37 55

## Checklist

### DIFF
--- a/src/features/recovery/components/RecoveryEmail/RegisterEmail.tsx
+++ b/src/features/recovery/components/RecoveryEmail/RegisterEmail.tsx
@@ -1,0 +1,80 @@
+import useSafeInfo from '@/hooks/useSafeInfo'
+import useOnboard from '@/hooks/wallets/useOnboard'
+import { asError } from '@/services/exceptions/utils'
+import { getAssertedChainSigner } from '@/services/tx/tx-sender/sdk'
+import { isWalletRejection } from '@/utils/wallets'
+import { Button, Grid, TextField } from '@mui/material'
+import Stack from '@mui/material/Stack'
+import { registerEmail } from '@safe-global/safe-gateway-typescript-sdk'
+import { useForm } from 'react-hook-form'
+
+const RegisterEmail = ({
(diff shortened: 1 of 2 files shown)


usame-algan:
> There is no user feedback when I reject the signature. Might be confusing.

Good point. I left a comment on Figma. Imo not a huge issue when signing as its still the same context for the user but we could show a notification.

> The PR's title is "add register email", but I would be also ok, if I was able to also edit the email.

Edit email will be added later on


> No e-mail validation. I guess the server will validate, but was not able to test whether we show any feedback on wrong email.

Good question. I don't think we have email validation in our app yet so this is something to discuss. Right now we don't show any error in case the request fails. I left a comment in Figma for this.

> I'm kind of confused that if I delete a recovery account, then add a new one and then click on "view email" I see the email. Shouldn't this actually be visible the whole time?

I think this also came up during one of our sync meetings and we decided to keep it simple for now as email notifications will be extracted later on anyway. In its current state it makes sense to only show the email section if recovery is set up as its not used for anything else.

PR #3333, opened 2024-02-26
fix: Catch errors during gas limit estimation

### LINKED ISSUE #3183: App crash on tx creation using tx builder with "Custom data" when tx nonce=0
## Bug description

When attempting to create and sign a Base mainnet transaction, the SDK fails to initialize with `Code 103: Error creating a SafeTransaction (The Safe SDK could not be initialized. Please be aware that we only support v1.0.0 Safe Accounts and up.)`.

Attempting the same transaction on Polygon mainnet works, so it appears to be a Base-specific issue. The Base Safe contract was deployed in the same way as the Polygon Safe contract, using the "Safe: Proxy Factory 1.3.0".

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Base mainnet

## Steps to reproduce
 1. Choose a Safe account on Base mainnet
 2. Click "New Transaction"
 3. Go to "Transaction Builder"
 4. Enter a contract address (e.g. `0x4e59b44847b379578588920cA78FbF26c0B4956C`).
 5. Toggle "Custom data" and enter call data
 6. Set an "ETH Value" of 0
 7. Click "Add Transaction"
 8. Click "Create Batch"
 9. Click "Send Batch"
 10. Observe the error

## Expected result

I should be able to sign the transaction.

## Obtained result

I receive an error indicating that the SDK failed to initialize.

## Screenshots

Screenshot by Dropbox Capture

usame-algan:
## What it solves

Resolves #3183 

`getEncodedSafeTx` can throw for invalid values like uneven byteslike lengths and since it didn't catch those errors it would crash the app.

## How this PR fixes it

Moves the `getEncodedSafeTx` and `operationType` calls into the `useAsync` so errors are caught and displayed to the user instead. We also need to make the callback async to achieve this. Since both variables are only used within the `useAsync`, having the memoized isn't necessary.

## How to test it

1. Open the tx-builder safe app
2. Create a transaction with custom data and choose `0x0` as the data
3. Send the transaction
4. Observe the app not crashing
5. Observe an error shown in the tx-flow modal instead

## Screenshots
Screenshot 2024-02-26 at 12 57 08

## Checklist

### DIFF
--- a/src/hooks/useGasLimit.ts
+++ b/src/hooks/useGasLimit.ts
@@ -1,4 +1,4 @@
-import { useEffect, useMemo } from 'react'
+import { useEffect } from 'react'
 import type Safe from '@safe-global/protocol-kit'
 import { encodeSignatures } from '@/services/tx/encodeSignatures'
 import type { SafeTransaction } from '@safe-global/safe-core-sdk-types'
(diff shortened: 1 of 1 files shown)


schmanu (approved):
Looks good to me!

francovenica:
LGTM. 

The app doesn't crash anymore and shows the error in the tx as described in the ticket's description

PR #3335, opened 2024-02-26
fix: do not modify signatures for fully signed txs when estimating

### LINKED ISSUE #3332: Different gaslimit is used for diff type of users during tx execution
## Bug description
Different gaslimit is used for diff type of users during tx execution 

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Sepolia

## Steps to reproduce
 1. Try to execute fully sign tx as
 - the owner who created and fully signed txs
 -  second owner who is trying to execute fully signed tx
 -  as non-owner who is trying to execute fully sign tx

Current result: The gas limit is different in all 2 cases
- 1st case - the gas limit is 74 XXX
- 2nd case - the gas limit is 72 XXX
- 3rd case - the gas limit is 76 XXX
As the result the tx execution is failed in some case because of the low gaslimit - https://sepolia.etherscan.io/tx/0xdc290620369f80a3cf601a02067337eb57b2bb28b7d3da2a704c01bda56e16f9

The next tx was used during testing https://safe-wallet-web.dev.5afe.dev/transactions/tx?safe=sep:0x2a73e61bd15b25B6958b4DA3bfc759ca4db249b9&id=multisig_0x2a73e61bd15b25B6958b4DA3bfc759ca4db249b9_0x3fe5da1d59bb76ea00a231eeb46441303ae712dd9d074c0fccc4be43451ac3e0 

## Expected result
Make sure that gaslimit is correct in all cases

schmanu:
## What it solves

Resolves #3332 

## How this PR fixes it
- Checks if a signature is needed before adding a pre-validated-sig.

## How to test it
- Open a 1/2 Safe where you control both owners
- Sign a transaction with the owner that has a higher address (higher hex number)
- Switch to the other owner with a lower address
- Try to execute the tx
- Observe that the gas estimation now matches

## Checklist

### DIFF
--- a/src/hooks/__tests__/useGasLimit.test.ts
+++ b/src/hooks/__tests__/useGasLimit.test.ts
@@ -1,0 +1,115 @@
+import { mockWeb3Provider, renderHook, waitFor } from '@/tests/test-utils'
+import useGasLimit from '../useGasLimit'
+import * as safeCoreSDK from '../coreSDK/safeCoreSDK'
+import * as useWallet from '../wallets/useWallet'
+import * as useSafeInfo from '../useSafeInfo'
+import * as useIsSafeOwner from '../useIsSafeOwner'
+
+import { mockContractManager } from '@/tests/mocks/contractManager'
+import type Safe from '@safe-global/protocol-kit'
+import { faker } from '@faker-js/faker'
+import { connectedWalletBuilder } from '@/tests/builders/wallet'
+import { createMockSafeTransaction } from '@/tests/transactions'
+import { safeInfoBuilder } from '@/tests/builders/safe'
+import { type JsonRpcProvider, zeroPadValue } from 'ethers'
+import { Gnosis_safe__factory } from '@/types/contracts/factories/@safe-global/safe-deployments/dist/assets/v1.3.0'
+
+const contractManager = mockContractManager()
+
+const walletAddress = faker.finance.ethereumAddress()
+const prevSignerAddress = faker.finance.ethereumAddress()
+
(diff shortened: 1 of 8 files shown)


usame-algan (approved):
Looks good!

There is a lint warning about `threshold` missing in the deps array and a couple of unrelated? tests failing.

francovenica:
I still see the issue. check the gif
for context:
The safe is in Sepolia, is a 1/2 safe, the tx is "Send funds" that is queued and with its creation is fully signed
The owner that created it was 0x8aEf2f5c3F17261F6F1C4dA058D022BE92776af8
In the gif I try to sign with that first owner and I have a gas limit of 65k
When I switch to the other owner "0x0D65139Da4B36a8A39BF1b63e950038D42231b2e" I get a gas limit of 67k. The same gas limit shows if I try to execute with a non-owner of the safe

gas estimation

schmanu:
Could you check again @francovenica?
Not sure if it was some deployment issue but for me I always see the same gas limit with the first signer, another owner who did not sign or some connected wallet that isn't an owner.

francovenica:
We tested with Manuel and we end up figuring out that if you send funds to yourself (so the receiver is the same EOA wallet as the one executing) the tx is slightly cheaper than if the person who execute is not the receiver. We also noticed that if the receiver is not an owner of the safe it also cost more.
We attribute all this to something on the contract that must control the gas limit for these different scenarios, so all of this is outside of the scope of this ticket and not considered an issue that must be fixed

I checked with other tx, like owner swapping and rejecting a tx and there I was able to confirm that, no matter who tries to sign the gas limit is always the same, so I consider this ticket done.

PR #3336, opened 2024-02-26
Feat: delete transactions

### LINKED ISSUE #3329: [Delete tx] Deletion logic
* Enable deletion only for proposers
* Sign an EIP1271 message to verify owner
* Send a request to CGW

Screenshot 2024-02-26 at 11 32 02

Design: https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?type=design&node-id=5794-183021&mode=design&t=EUCtPuMbQTVxoewx-0

Specs: https://www.notion.so/safe-global/Deleting-transactions-e62a1f6a6d2547eab25243663e0325f1

API docs: https://safe-transaction-mainnet.safe.global/ (search for /v1/multisig-transactions/{safe_tx_hash}/)

katspaugh:
## What it solves

Resolves #3329

## How this PR fixes it

It's not working yet, I'm getting Error 42 Bad Request from the backend.

## Screenshots
Screenshot 2024-02-26 at 13 47 48

### DIFF
--- a/src/components/transactions/TxDetails/index.tsx
+++ b/src/components/transactions/TxDetails/index.tsx
@@ -55,6 +55,10 @@ const TxDetailsBlock = ({ txSummary, txDetails }: TxDetailsProps): ReactElement 
   // If we have no token list we always trust the transfer
   const isTrustedTransfer = !hasDefaultTokenlist || isTrustedTx(txSummary)
 
+  const safeTxHash = isMultisigDetailedExecutionInfo(txDetails.detailedExecutionInfo)
+    ? txDetails.detailedExecutionInfo.safeTxHash
+    : undefined
+
   return (
     <>
       {/* /Details */}
@@ -111,7 +115,7 @@ const TxDetailsBlock = ({ txSummary, txDetails }: TxDetailsProps): ReactElement 
           {isQueue && (
             <Box display="flex" alignItems="center" justifyContent="center" gap={1} mt={2}>
               {awaitingExecution ? <ExecuteTxButton txSummary={txSummary} /> : <SignTxButton txSummary={txSummary} />}
-              <RejectTxButton txSummary={txSummary} />
+              <RejectTxButton txSummary={txSummary} safeTxHash={safeTxHash} />
             </Box>
           )}
         </div>
--- a/src/components/tx-flow/flows/ReplaceTx/index.tsx
+++ b/src/components/tx-flow/flows/ReplaceTx/index.tsx
@@ -1,4 +1,4 @@
-import { Box, Button, SvgIcon, Tooltip, Typography } from '@mui/material'
+import { Box, Button, Divider, SvgIcon, Tooltip, Typography } from '@mui/material'
 
 import InfoIcon from '@/public/images/notifications/info.svg'
 import ReplaceTxIcon from '@/public/images/transactions/replace-tx.svg'
@@ -7,12 +7,13 @@ import { useQueuedTxByNonce } from '@/hooks/useTxQueue'
 import { isCustomTxInfo } from '@/utils/transaction-guards'
 
 import css from './styles.module.css'
-import { useContext } from 'react'
+import { useContext, useState } from 'react'
 import { TxModalContext } from '../..'
 import TokenTransferFlow from '../TokenTransfer'
 import RejectTx from '../RejectTx'
 import TxLayout from '@/components/tx-flow/common/TxLayout'
 import TxCard from '@/components/tx-flow/common/TxCard'
+import DeleteTxModal from './DeleteTxModal'
 
 // TODO: Move this to the status widget
 /*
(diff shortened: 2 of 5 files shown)


src/components/transactions/TxDetails/index.tsx
```diff
@@ -55,6 +57,14 @@ const TxDetailsBlock = ({ txSummary, txDetails }: TxDetailsProps): ReactElement
   // If we have no token list we always trust the transfer
   const isTrustedTransfer = !hasDefaultTokenlist || isTrustedTx(txSummary)
 
+  const proposedSafeTxHash =
+    isMultisigDetailedExecutionInfo(txDetails.detailedExecutionInfo) &&
+    isMultisigExecutionInfo(txSummary.executionInfo) &&
+    wallet &&
+    wallet.address === txSummary.executionInfo.proposer?.value
```
schmanu, line 64: Let's use the `sameAddress` helper here

schmanu: A test for this would be good as well as this code decides if we offer deletion for the owner or not.

src/components/tx-flow/flows/ReplaceTx/index.tsx
```diff
@@ -1,114 +1,90 @@
...
         </Box>
-        <Typography variant="h4" fontWeight="700" textAlign="center">
-          Select how you would like to replace this transaction
-        </Typography>
-        <Typography variant="body1" textAlign="center">
-          A signed transaction cannot be removed but it can be replaced with a new transaction with the same nonce.
+
+        <Typography variant="body2" mt={-1} mb={1}>
+          You can replace or reject this transaction on-chain. It requires gas fees and your signature.{' '}
+          <ExternalLink href="https://help.safe.global/">Read more</ExternalLink>{' '}
```
katspaugh, line 37: TODO: add the correct help article link here.

katspaugh:
@schmanu could you review again pls, I've added a redesign for the Replace Tx modal since your last review.

schmanu (approved):
Looks good. I really like the new Replace modal buttons as well :)

schmanu:
I just took a look again in dark mode and the colors / icon does not seem right there:
The recommended pill should have a dark font and the icon should have a dark mode version.
Screenshot 2024-03-14 at 07 44 02

TanyaEfremova:
1. **Read more** link on the rejection page should probably lead to [this article](https://help.safe.global/en/articles/40836-why-do-i-need-to-pay-for-cancelling-a-transaction)
2. The tag **Recommended** should be 32px height (the same size as tag `New`).
2. Word **OR** should use `Overline` texts style (the same as on the Welcome page)
Suggestion: When a user chooses to reject or cancel a transaction, having a back button on the next page or being able to return to the initial screen with choices would be convenient. 
Screenshot 2024-03-14 at 09 47 29
5c50d0a7b4">

katspaugh:
@TanyaEfremova btw I think it's not very clear that `Recommended` refers to the first two options, it's kinda disconnected from any of them in fact.

Screenshot 2024-03-14 at 11 37 46

TanyaEfremova:
Let's try it [differently](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?type=design&node-id=10439%3A21907&mode=design&t=m0geNsE34DxClPVP-1), but not overload it visually, by adding the same colorful tag in both options twice

francovenica:
I tried in Sepolia, in a safe 1/x with a send funds tx type

I had 2 issues:
The first one is I queued 2 "send funds" tx and a rejection, all of them in the same nonce. I tried the "Remove from queue" option and even tho it did work for the rejection and one of the "send funds" (the tx was removed), the modal never closed. I had to manually go to the queue and see that the tx was removed.
I was only able to see this issue once, never able to reproduce it again.
When I tried to remove the last "Send funds" tx I got the 2nd issue.

The second issue that is being happening consistently with every tx I've been trying so far is that I got an "error 400: provided owner not the proposer of the transaction".I don't change signers at all at any point, during the tx creation or rejection
Gif:
error 400

I'll stop here before checking other types of tx.

Uxio0:
@francovenica Could you provide details about the transaction so I can debug? Can you link the safe?

francovenica:
@Uxio0 
So this is the safe, is in sepolia. It was a regular "send funds" tx type 
https://delete_tx--walletweb.review-wallet-web.5afe.dev/transactions/history?safe=sep:0x8f4A19C85b39032A37f7a6dCc65234f966F72551

I can make you owner if you want, but I don't think this safe has anything special

src/utils/gateway.ts
```diff
@@ -31,3 +34,45 @@ export const getExplorerLink = (
...
+  }
+  const types = {
+    DeleteRequest: [
+      { name: 'safeTxHash', type: 'bytes32' },
+      { name: 'totp', type: 'uint256' },
+    ],
+  }
+  const message = {
+    safeTxHash: safeTxHash,
+    totp: Math.round(Date.now() / 3600e3),
```
schmanu, line 58: We have to `Math.floor` here according to the algorithm (https://en.wikipedia.org/wiki/Time-based_one-time_password)

katspaugh:
@francovenica both issues are fixed now!

katspaugh:
Fixed both of those edge cases, @francovenica. ✅

francovenica:
The issues reported were fixed

new issues:

Signing with Trezor and Ledger X is not supported.
I get no other errors in the console or network tabs
image

katspaugh:
I'll merge this PR, will add Ledger support in #3470.

PR #3343, opened 2024-02-27
chore: remove old sidebar components

jmealy:
## What it solves
Remove components from the old sidebar which are no longer used.


usame-algan:
Not sure if I am missing something but I can't find these components on dev. Looks like they've already been deleted?

katspaugh:
It was already merged once: #3317

PR #3344, opened 2024-02-27
fix: Stay on the same route when switching safes

### LINKED ISSUE #3341: Stay on current page when switching to another Safe
## Bug description
Currently when switching to another Safe via the Sidebar we always open up the dashboard.
Previously we stayed on the current page.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Open the Queue of a Safe
 2. Switch to another Safe via the Sidebar

## Expected result
The queue of the selected Safe opens up

## Obtained result
The dashboard of the selected Safe opens up

usame-algan:
## What it solves

Resolves #3341

## How this PR fixes it

- Restores the `getHref` function that was previously in the `SafeList` component

## How to test it

1. Open a safe
2. Navigate to any route
3. Switch to a different safe
4. Observe being on the same route
5. Open the single tx view
6. Switch to a different safe
7. Observe being navigated to the tx history
8. Go to the welcome page
9. Select a safe there
10. Observe being navigated to the dashboard

## Checklist

### DIFF
--- a/src/components/welcome/MyAccounts/AccountItem.tsx
+++ b/src/components/welcome/MyAccounts/AccountItem.tsx
@@ -1,4 +1,5 @@
-import { useMemo } from 'react'
+import type { ChainInfo } from '@safe-global/safe-gateway-typescript-sdk'
+import { useCallback, useMemo } from 'react'
 import { ListItemButton, Box, Typography } from '@mui/material'
 import Link from 'next/link'
 import SafeIcon from '@/components/common/SafeIcon'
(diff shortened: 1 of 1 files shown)


src/components/welcome/MyAccounts/AccountItem.tsx
```diff
@@ -39,10 +35,27 @@ const AccountItem = ({ onLinkClick, chainId, address, ...rest }: AccountItemProp
   const isCurrentSafe = chainId === currChainId && sameAddress(safeAddress, address)
   const trackingLabel =
     router.pathname === AppRoutes.welcome.accounts ? OVERVIEW_LABELS.login_page : OVERVIEW_LABELS.sidebar
+  const isWelcomePage = router.pathname === AppRoutes.welcome.index || router.pathname === AppRoutes.welcome.socialLogin
```
katspaugh, line 38: This can be removed because we don't have a SafeList on the Welcome page anymore.

usame-algan: Ah nice catch! We have to change it to `AppRoutes.welcome.accounts` since we are displaying them there

francovenica:
Ah, didn't realize that this broke. My bad.

LGTM

PR #3345, opened 2024-02-27
refactor: update typescript to version 5.3.3

compojoom:
## What it solves
When working on the stake feature we ran into a conflict with stakekit and typescript v4. Since sooner or later we have to update, I decided to give typescript v5 a chance (in the past I was running in a lot of errors). This time it seems that it could be done in a painless way.


katspaugh (approved):
Thank you sir!

PR #3349, opened 2024-02-27
Update next.js to v 14.1

compojoom:
## What it solves
Keeps us up to date with the latest and greatest next.js release.


usame-algan (approved):
Looks good! I checked out the Branch locally and played around with the app and nothing seems broken. Lets still run the regression e2e and do some smoke tests with QA.

PR #3356, opened 2024-02-28
Fix: remove broken index redirect

katspaugh:
## What it solves

The redirect fron '/` to the last opened Safe hasn't been working reliably recently, but also it conceptually complicates the user flow.

We want the Welcome page to be the starting point of the app, w/o a hidden redirect logic.

### DIFF
--- a/cypress/e2e/pages/create_wallet.pages.js
+++ b/cypress/e2e/pages/create_wallet.pages.js
@@ -4,15 +4,14 @@ import * as main from '../pages/main.page'
 const welcomeLoginScreen = '[data-testid="welcome-login"]'
 const expandMoreIcon = 'svg[data-testid="ExpandMoreIcon"]'
 const nameInput = 'input[name="name"]'
-const selectNetworkBtn = '[data-cy="create-safe-select-network"]'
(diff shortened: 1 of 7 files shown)


usame-algan (approved):
There is this weird behaviour now when being on the `/welcome/accounts` page and pressing the Logo it looks like its loading but keeps the user trapped on that page. The only way to get back to `/welcome` is by disconnecting the wallet and pressing the Logo again.

Not sure how to best solve this. We could change the Logo link to always go to `/welcome` or `/welcome/accounts` but generally I think it is confusing that we now have two "home" pages instead of just one.

katspaugh:
Yeah, the cause is that the Welcome page itself redirects to the Accounts page if you're connected.

Perhaps we should remove that redirect as well, and only go to the Accounts page if the user clicks on Continue.

francovenica:
LGTM

also +1 to not redirect from the /welcome page automatically to the /welcome/accounts page. Is produces a flickr effect that doesn't look pretty.

katspaugh:
I've removed the auto-redirect from Welcome to Accounts. It will now redirect only if you explicitely connect a wallet via the Connect button on that page.

PR #3364, opened 2024-02-29
Analytics: track number of owned safes and watchlist safes from the login page

jmealy:
## What it solves
The `Total added Safes' event was being tracked too much. Replace it with two events to track the number of safes in the watchlist and the number of owned safes. 

## How this PR fixes it
- From the login page (/welcome/accounts) emit two events to track the number of safes:
  - `Total Safes owned`
  - `Total Safes watchlist`
- Emit these events only once when logging in.
- Remove the `Total Added Safe` event

## How to test it
- visit `/welcome/accounts` 
- Check that the two events are emitted in the console.
- Check that the number of safes specified is correct in each case. The count is tracked in the `eventLabel` field
- Check that these events are not emitted elsewhere (such as when opening the sidebar from the console.)
- Check that `Total Added Safe` is no longer emitted anywhere.

## Screenshots
image


src/components/welcome/MyAccounts/index.tsx
```diff
@@ -13,21 +13,31 @@ import AddIcon from '@/public/images/common/add.svg'
...
+  const isLoginPage = router.pathname === AppRoutes.welcome.accounts
+  const trackingLabel = isLoginPage ? OVERVIEW_LABELS.login_page : OVERVIEW_LABELS.sidebar
+
+  useEffect(() => {
+    // track owned and watchlist safe counts once when the user logs in
+    if (watchlistSafes && ownedSafes && isLoginPage) {
+      trackEvent({ ...OVERVIEW_EVENTS.TOTAL_SAFES_OWNED, label: ownedSafes.length })
+      trackEvent({ ...OVERVIEW_EVENTS.TOTAL_SAFES_WATCHLIST, label: watchlistSafes.length })
+    }
+  }, [isLoginPage, ownedSafes, watchlistSafes])
```
katspaugh, line 40: This effect will run as many times as any of these dependencies change, right? I thought @alicecha wanted to emit this only one per hard page load.

jmealy: true it shouldn't send the events when you add a safe to the watchlist or anything like that. I removed ownedSafes watchlistSafes from the dependency array

compojoom: but isn't this now always going to send the safes when you navigate to the welcome page? I can click a safe, navigate to it, then hit back and then click on another safe.
In your first iteration it was when "on login page and watched safes and owned safes" change. Now you would always send it when on login page.
I'm not sure what we want here, but the way I understood is - once per session. What do we treat as a session? I think - when connected with walletX, if that is the case we need this somewhere higher in the component hierarchy?

jmealy: yes so it sends the events once on visiting the login page. The idea being that the frequency people usually visit the login page is about the frequency @alicecha wanted these data tracked. Tracking on wallet connect could work also but it is missing out on the case where the user uses safes in localstorage without connecting a wallet

src/components/welcome/MyAccounts/index.tsx
```diff
@@ -13,21 +13,32 @@ import AddIcon from '@/public/images/common/add.svg'
...
+  const trackingLabel = isLoginPage ? OVERVIEW_LABELS.login_page : OVERVIEW_LABELS.sidebar
+
+  useEffect(() => {
+    // track owned and watchlist safe counts once when the user logs in
+    if (watchlistSafes && ownedSafes && isLoginPage) {
+      trackEvent({ ...OVERVIEW_EVENTS.TOTAL_SAFES_OWNED, label: ownedSafes.length })
+      trackEvent({ ...OVERVIEW_EVENTS.TOTAL_SAFES_WATCHLIST, label: watchlistSafes.length })
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isLoginPage])
```
katspaugh, line 41: This won't work if safes aren't defined on the initial component render. You should keep all the dependencies but add a global flag to do it only once per page.

Besides, please extract that logic into a hook.

src/components/welcome/MyAccounts/index.tsx
alicecha: We need to think about what purpose this data will serve. Based on recent conversations, we want to know the number of Safes in the list(s) so we can
1) decide how to display them properly in the UI
2) understand how many users will be impacted by design choices, and how often (e.g. how many users needed to do an extra click when we started showing only 5 Safes)
3) estimate the impact of increasing the number of CGW requests.

Others??

Based on this, not sure whether it makes more sense to trigger the event every time a list is loaded rather than once per session. Wdyt?

jmealy: to clarify, by every time list is loaded you mean emit for:
- opening the sidebar.
- viewing accounts page
- adding or creating a safe.
If that's the case my guess is that it would be a lot of events with mostly the same data. Personally I think once per session would probably be enough to get an idea, but it's up to you @alicecha. We can also pare it back later if it's too much.

alicecha: Thanks for clarifying. Let's emit once per session!

src/components/welcome/MyAccounts/useTrackedSafesCount.ts
```diff
@@ -0,0 +1,21 @@
+import { AppRoutes } from '@/config/routes'
+import { OVERVIEW_EVENTS, trackEvent } from '@/services/analytics'
+import { useRouter } from 'next/router'
+import { useEffect, useState } from 'react'
+import type { SafeItems } from './useAllSafes'
+
+const useTrackSafesCount = (ownedSafes: SafeItems | undefined, watchlistSafes: SafeItems | undefined) => {
+  const [isTracked, setIsTracked] = useState(false)
```
katspaugh, line 8: This is not global tho, is it? It will run as many times as this hook runs. I.e. on every Safe List render.

jmealy: you mean like this?

```
let isTracked = false

const useTrackSafesCount = (ownedSafes: SafeItems | undefined, watchlistSafes: SafeItems | undefined) => {
  const router = useRouter()
  const isLoginPage = router.pathname === AppRoutes.welcome.accounts

  useEffect(() => {
    if (watchlistSafes && ownedSafes && isLoginPage && !isTracked) {
     ...
      isTracked = true
    }
  }, [isLoginPage, ownedSafes, watchlistSafes])
}
```

katspaugh: Yes, please test it, though. In some build systems imported modules are functions that are executed every time on import.

jmealy: Tested, and the only time it sends them again is on browser reload. Navigating back to the login page doesn't send them again, which I think is what we want here.

katspaugh: Did you test it locally or on the PR preview? Dev & prod build are quite different.

jmealy: With the PR preview link

francovenica:
Issue:

So the event triggers if you visit "/welcome/account" by copy/pasting the URL, refresh while being in that URL or if you switch wallet (like switching your owner in the MM extension). But If you reach "/welcome/account" by clicking "view my accounts" in the welcome page or "add+" button in the sidebar (to add the safe to the watchlist) the events won't trigger.


Side note:
This PR has a bug that was already fixed in #3398, about a "Create new safe" event that was triggering twice. Take this in account whenever this ticket has to be merged in the RC.

jmealy:
> Issue:
> 
> So the event triggers if you visit "/welcome/account" by copy/pasting the URL, refresh while being in that URL or if you switch wallet (like switching your owner in the MM extension). But If you reach "/welcome/account" by clicking "view my accounts" in the welcome page or "add+" button in the sidebar (to add the safe to the watchlist) the events won't trigger.
> 
> Side note: This PR has a bug that was already fixed in #3398, about a "Create new safe" event that was triggering twice. Take this in account whenever this ticket has to be merged in the RC.

The idea is to send the events once per "session" so this is the expected behaviour. The event these are replacing was being tracked too much so we wanted to reduce the frequency.

PR #3369, opened 2024-02-29
Fix: zksync nested safe gas estimation

compojoom:
## What it solves

Resolves https://github.com/safe-global/safe-wallet-web/issues/3141

## How this PR fixes it
Instead of using the safe wallet as "from" for the gas estimation, we use a random EOA address. 
Also, I don't think that we should show an error for when we determine that the connected walelt might not have enough funds. I changed it to info, maybe warning would be even more appropriate, opinions? 

## How to test it
You need a nested safe setup: parent/child. The parent should be owner of the child. Create a tx in the child (e.g. add owner etc) and try to sign in the parent safe. Without this fix the process will fail with an estimation error. With the fix, the transaction should go through. 

## Checklist

### DIFF
--- a/src/hooks/useGasLimit.ts
+++ b/src/hooks/useGasLimit.ts
@@ -11,6 +11,13 @@ import { useSafeSDK } from './coreSDK/safeCoreSDK'
 import useIsSafeOwner from './useIsSafeOwner'
 import { Errors, logError } from '@/services/exceptions'
 import useSafeInfo from './useSafeInfo'
+import { estimateTxBaseGas } from '@safe-global/protocol-kit/dist/src/utils/transactions/gas'
+import {
+  getCompatibilityFallbackHandlerContract,
+  getSimulateTxAccessorContract,
+} from '@safe-global/protocol-kit/dist/src/contracts/safeDeploymentContracts'
+import { type JsonRpcProvider } from 'ethers'
+import { type ExtendedSafeInfo } from '@/store/safeInfoSlice'
 
 const getEncodedSafeTx = (
   safeSDK: Safe,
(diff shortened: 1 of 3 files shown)


src/hooks/useGasLimit.ts
```diff
@@ -40,6 +47,79 @@ const incrementByPercentage = (value: bigint, percentage: bigint) => {
...
+  const safeFunctionToEstimate: string = fallbackHandlerContract.encode('simulate', [
+    await simulateTxAccessorContract.getAddress(),
+    transactionDataToEstimate,
+  ])
+
+  const gas = await web3.estimateGas({
+    to: safe.address.value,
+    // use a random EOA address as the from address
+    // https://github.com/zkSync-Community-Hub/zksync-developers/discussions/144
+    from: '0x330d9F4906EDA1f73f668660d1946bea71f48827',
```
katspaugh, line 112: Could you move this to a constant on top of this function?

src/hooks/useGasLimit.ts
```diff
@@ -40,6 +47,79 @@ const incrementByPercentage = (value: bigint, percentage: bigint) => {
...
+    value: '0',
+    data: safeFunctionToEstimate,
+  })
+
+  // The estimateTxBaseGas function seems to estimate too low for zkSync
+  const baseGas = BigInt(await estimateTxBaseGas(safeSDK, safeTx)) * 20n
```
katspaugh, line 118: So this is the second hack we have to do for zkSync? Can you put the multiplier into a constant?

src/hooks/useGasLimit.ts
```diff
@@ -68,6 +148,11 @@ const useGasLimit = (
       safeTx.signatures.size < threshold,
     )
 
+    // if we are dealing with zksync and the walletAddress is a Safe, we have to do some magic
+    if (safe.chainId === chains.zksync && (await web3ReadOnly.getCode(walletAddress)) !== '0x') {
+      return getGasLimitForZkSync(safe, web3ReadOnly, safeSDK, safeTx)
+    }
```
katspaugh, line 154: Why do we have to reimplement the entire gas estimation and not just substitute the [from address](https://github.com/safe-global/safe-wallet-web/blob/32f9038386419ec31d65dd06eee64d4340ba53b5/src/hooks/useGasLimit.ts#L159) in the existing estimation call?

Also, there's already [a multiplier for Gnosis Chain](https://github.com/safe-global/safe-wallet-web/blob/32f9038386419ec31d65dd06eee64d4340ba53b5/src/hooks/useGasLimit.ts#L165-L168), it would be good to do it for zkSync in the same place, using some config.

katspaugh (changes requested):

src/hooks/useGasLimit.ts
schmanu: The problem here is:
On Zksync every EOA is actually a Smart account deployed through the same proxy factory.
You can only use `eth_estimateGas` if the `from` address was created through that proxy factory. Safes do not use that factory and thus can not be used.

If we now put a random EOA as `from` address and call `eth_estimateGas` the transaction will fail as the signature is missing because that random EOA is not an owner and therefore we can not use a prevalidated signature.
That's why we have to use this alternative estimation function. 

@compojoom requested to add all this logic to the core-sdk. That would be the ideal solution as this affects all dapps built on top of Safe using the protocol-kit.

compojoom: Because it's not that easy :) If we just change the `from` address we run into'GS025: Hash has not been approved' (https://github.com/safe-global/safe-smart-account/blob/12136fd54940892bf43258f86b6e8f9d0eff69ea/contracts/Safe.sol#L302). If we also change the `from` also for the enoded tx we get`GS026: Invalid owner provided`. 

So, instead what we do for zksync is we use the simulate function on the callbackHandler: https://github.com/safe-global/safe-smart-account/blob/12136fd54940892bf43258f86b6e8f9d0eff69ea/contracts/handler/CompatibilityFallbackHandler.sol#L88
So we are not passing the encodedSafeTx for the estimateGas function, but rather "wrapping" it in a "simulate" call and then calling the estimate gas for "simulate" call. 

That's my current understanding of the issue. @schmanu did the same thing for counterfactual deployments, but there he had a multisend. 

About the multiplier - the multiplier for GNO is not a config. Also the incrementByPercentage function just increases the gasLimit by that percentage. Note that I'm multiplying the baseFee by 20 to get it to a value that passes.  So, should I use incrementByPercentage and pass 2000%?

src/hooks/useGasLimit.ts
compojoom: check again please

src/hooks/useGasLimit.ts
compojoom: check again please

src/hooks/useGasLimit.ts
katspaugh: @compojoom I would stick to floats, i.e.

```js
const GasMultipliers = {
  [chains.gno]: 1.2,
  [chains.zksync]: 20,
}
```

compojoom: why float and not Bigint? The returned values from the estimate functions are bigints

francovenica:
I'm getting another error about gas estimation, which is weird since I'm not even trying to execute the tx, just sign it.
I'm trying a safe where is a 2/2, one owner is a EOA, the other is the safe.
I'm trying to sign with the safe in this case

image

schmanu:
> I'm getting another error about gas estimation, which is weird since I'm not even trying to execute the tx, just sign it. I'm trying a safe where is a 2/2, one owner is a EOA, the other is the safe. I'm trying to sign with the safe in this case
> 
> image

@francovenica could you link to your safe / the transaction? Woud love to try to simulate this with tenderly to see why it is failing.

liliya-soroka:
Same issue on my side in case of "only signing" with execution later
image

schmanu:
@liliya-soroka @francovenica 
I found the issue and fixed it. It was a related but separate issue within the protocol-kit on zksync.

Signing for nested Safes should work now!

liliya-soroka:
Verified

PR #3373, opened 2024-03-01
fix: not possible to click on a safe in listItem

compojoom:
Since the .listItem had a padding, the <a> tag inside of it was only clickable if you manage to hit it directly. But ListItem actually looks like a clickable button, so you won’t expect that you have to be clicking directly in the center.

## What it solves
Since a picture is worth 1000 words, a gif is worth ????

Resolves #

## Checklist


katspaugh (approved):
🙌

PR #3375, opened 2024-03-01
rename owner to signer

jmealy:
WIP

## What it solves

Resolves #

## Checklist


schmanu (approved):
🦾

francovenica:
In the settings > setup there is an option to Export as CSV the list of signers. That file is called "owners". Change it to signers as well if possible 
image

francovenica:
The "owner" text is in the "redefine" evaluations, but I'm not sure of the text there is something we have control over of if they come from redefine itself. Check if it is something we can change or not
To see that text try to add a new owner to any safe

image

francovenica:
The rest looks good

jmealy:
> In the settings > setup there is an option to Export as CSV the list of signers. That file is called "owners". Change it to signers as well if possible image

Good spot, thanks 🙌

PR #3376, opened 2024-03-01
Feat: if there are local safes redirect from index route `/` to accounts page 

jmealy:
## What it solves
- Allow users to access their locally stored safes without requiring them to connect a wallet. 
- redirect automatically to the accounts page if the user connects a wallet via the wallet connect button on the `/welcome` page. This change was included in [this PR](https://github.com/safe-global/safe-wallet-web/pull/3356) but the redirect is not working.

## How this PR fixes it
When a user visits the main route of the site, check if they have safes in localstorage.
- if they do, redirect to `/welcome/accounts`
- if they don't redirect to `/welcome` as before
- redirect to the login page (/wecome/accounts) when a user explicitly connects using the wallet connect button on the welcome page.


katspaugh (approved):
Looks good but there's a failing test.

PR #3380, opened 2024-03-04
fix: mainnet gas price calculation

schmanu:
## What it solves
The gas estimation on mainnet is sometimes wrong.

## How this PR fixes it
Takes `maxPriorityFeePerGas` and `baseFee` from etherscan instead of using ethers `maxPriorityFeePerGas`.

## How to test it
- Try to execute a transaction on mainnet.
- Compare the estimated `maxFeePerGas` and maxPriorityFeePerGas with etherscan's [gas tracker](https://etherscan.io/gastracker)

## Checklist

### DIFF
--- a/src/hooks/useGasPrice.ts
+++ b/src/hooks/useGasPrice.ts
@@ -1,5 +1,5 @@
 import { formatVisualAmount } from '@/utils/formatters'
-import { type FeeData, parseUnits } from 'ethers'
+import { type FeeData } from 'ethers'
 import type {
   ChainInfo,
   GasPrice,
@@ -33,8 +33,6 @@ type GasFeeParams = {
 // Update gas fees every 20 seconds
 const REFRESH_DELAY = 20e3
 
-const ETHER_SCAN_URL = 'https://api.etherscan.io/api?module=gastracker&action=gasoracle'
-
 type EtherscanResult = {
   LastBlock: string
   SafeGasPrice: string
@@ -42,6 +40,10 @@ type EtherscanResult = {
   FastGasPrice: string
   suggestBaseFee: string
   gasUsedRatio: string
+}
+
+const isEtherscanResult = (data: any): data is EtherscanResult => {
+  return 'FastGasPrice' in data && 'suggestBaseFee' in data
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/hooks/useGasPrice.ts
```diff
@@ -45,7 +74,11 @@ const fetchGasOracle = async (gasPriceOracle: GasPriceOracle): Promise<bigint> =
 
   const json = await response.json()
   const data = json.data || json.result || json
-  return BigInt(data[gasParameter] * Number(gweiFactor))
+
+  if (uri.startsWith(ETHER_SCAN_URL)) {
```
katspaugh, line 78: I would do duck typing instead of checking the URL. This will work even of the URL change, and it will also protect you if the API changes.

```suggestion
  if (isEtherscanResult(data)) {
    return parseEtherscanOracleResult(result)
  }
```

```js
function isEtherscanResult(data): data is EtherscanResult {
  return 'FastGasPrice' in data && 'suggestBaseFee' in data
}
```

src/hooks/useGasPrice.ts
```diff
@@ -33,10 +33,39 @@ type GasFeeParams = {
...
+ * Parses result from etherscan oracle.
+ * Since EIP 1559 it returns the `maxFeePerGas` as gas price and the current network baseFee as `suggestedBaseFee`.
+ * The `maxPriorityFeePerGas` can then be computed as `maxFeePerGas` - `suggestedBaseFee`
+ *
+ * @param result {@link EtherscanResult}
+ * @see https://docs.etherscan.io/api-endpoints/gas-tracker
+ */
+const parseEtherscanOracleResult = (result: EtherscanResult): EstimatedGasPrice => {
+  const maxFeePerGas = parseUnits(result.FastGasPrice, 9)
+  const baseFee = parseUnits(result.suggestBaseFee, 9)
```
usame-algan, line 57: Is it possible to use `gweiFactor` here instead of hardcoding 9?

katspaugh (approved):
Thanks!

Could you raise with backend that we should revise the config-service setup for gas oracles?

liliya-soroka:
hm, I got timeout for usual ETH transfer tx again. The Max base fee is still too low ( 99-113 instead of 130-160-Market)
But the tx was executed on-chain successfully in 1.5-2 mints 
What do you think about increasing timeout to 2 mints with this fix?
image

liliya-soroka:
Verified, the estimation on Ethereum is better now

PR #3390, opened 2024-03-05
Fix: Remove show more button from safe lists

jmealy:
## What it solves
Users have complained that it is difficult to find the safe they're looking for with the pagination on the safe lists.

## How this PR fixes it
Remove the show more button from the safes list. Show all safes in the list by default


katspaugh (approved):
👍

Issue #3145, opened 2024-01-23
[Recovery] Register and verify an email address via Settings


usame-algan:
## What is the feature about

Allow users to register and verify an email address on the settings page in order to receive email notifications about pending recovery attempts.

## The list of requirements

- [ ] #3291
- [ ] #3320
- [ ] #3321

## Designs/sketches

- [Figma](https://www.figma.com/file/pO43pjVwR34Uy1ihMsY3OM/Recovery-Hub?type=design&node-id=1816-48897&mode=design&t=syybxe0VIclJTlmh-0)

## Links

Tech spec: [Notion](https://www.notion.so/safe-global/As-a-Safe-owner-I-want-to-register-an-email-address-so-that-I-can-receive-recovery-notifications-80f8336053694dec917db5d3072a6610)

usame-algan:
Closing this as the project was removed from the roadmap

PR #3392, opened 2024-03-05
fix: [Recovery] Register recovery module for alerts

usame-algan:
## What it solves

Part of #3145 

## How this PR fixes it

- Registers the recovery module for Alerts whenever an email address is added

## How to test it

1. Set up recovery
2. Add an email address and verify it
3. Propose a recovery
4. You should receive an email for the recovery proposal

## Checklist

### DIFF
--- a/src/features/recovery/components/RecoveryEmail/VerifyEmail.tsx
+++ b/src/features/recovery/components/RecoveryEmail/VerifyEmail.tsx
@@ -1,4 +1,3 @@
-import { asError } from '@/services/exceptions/utils'
 import { useAppDispatch } from '@/store'
 import { showNotification } from '@/store/notificationsSlice'
 import { useState } from 'react'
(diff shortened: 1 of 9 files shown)


src/features/recovery/components/RecoveryEmail/VerifyEmail.tsx
```diff
@@ -23,45 +22,35 @@ export const NotVerifiedMessage = ({ onVerify }: { onVerify: () => void }) => {
...
+      await verifyEmailAddress(verificationCode)
+      dispatch(
+        showNotification({
+          variant: 'success',
+          groupKey: 'verify-email-complete',
+          message: 'Your email address is verified',
+        }),
+      )
+      onSuccess()
+      await registerRecovery()
```
usame-algan, line 52: We could also register the recovery one step sooner when the email is registered and not when its verified.

src/features/recovery/components/RecoveryEmail/useRecoveryEmail.ts
```diff
@@ -100,13 +103,20 @@ const useRecoveryEmail = () => {
     )
   }
 
+  const registerRecovery = () => {
+    if (!recovery || recovery.length === 0) return
```
katspaugh, line 107:
```suggestion
    if (!recovery?.length) return
```

katspaugh (approved):

PR #3397, opened 2024-03-06
Fix: remove dynamic sorting for safe lists

jmealy:
## What it solves
For users with a lot of safes it is disorienting for the order of their accounts to change after selecting a different safe. Make the list order persistent.

Resolves #

## How this PR fixes it
- Remove sorting by current chain, so that the list order does not change after selecting a new safe.
- Now it is ordered primarily by chains you have added safes on, and secondarily by chainId.

### DIFF
--- a/src/components/welcome/MyAccounts/useAllSafes.ts
+++ b/src/components/welcome/MyAccounts/useAllSafes.ts
@@ -45,7 +45,7 @@ const useAllSafes = (): SafeItems => {
   const undeployedSafes = useAppSelector(selectUndeployedSafes)
 
   return useMemo<SafeItems>(() => {
-    const chains = uniq([currentChainId].concat(Object.keys(allAdded)).concat(Object.keys(allOwned)))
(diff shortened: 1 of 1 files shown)


katspaugh (commented):
Lint

src/components/welcome/MyAccounts/useAllSafes.ts
```diff
@@ -45,7 +45,7 @@ const useAllSafes = (): SafeItems => {
   const undeployedSafes = useAppSelector(selectUndeployedSafes)
 
   return useMemo<SafeItems>(() => {
-    const chains = uniq([currentChainId].concat(Object.keys(allAdded)).concat(Object.keys(allOwned)))
```
katspaugh, line 48: currentChainId can be removed alltogether then.

katspaugh (approved):
👍

francovenica:
LGTM. The safe order is the same everytime. Tested in the sidebar and login page

PR #3399, opened 2024-03-06
Fix: always redirect index route to accounts page when there are safes in localstorage

jmealy:
## What it solves

Resolves [notion](https://www.notion.so/safe-global/4cbe48f69de44fd3ae8682b2c45cdcbe?v=d1599a9fe4ff413a9b7ccdd1af19b9cb&p=9c5d6c1fe1bc4928b5c5e33313aa2ab3&pm=s)

## How this PR fixes it
When landing on the index route for the first time, the redirect to the welcome page happens before the check for local safes is complete. Make sure the check for local safes has completed **before**  redirecting.

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -10,7 +10,7 @@ const IndexPage: NextPage = () => {
   const { hasSafes } = useHasSafes()
 
   useEffect(() => {
-    if (!router.isReady || router.pathname !== AppRoutes.index) {
+    if (!router.isReady || router.pathname !== AppRoutes.index || hasSafes === undefined) {
       return
     }
 


src/pages/index.tsx
```diff
@@ -10,7 +10,7 @@ const IndexPage: NextPage = () => {
   const { hasSafes } = useHasSafes()
 
   useEffect(() => {
-    if (!router.isReady || router.pathname !== AppRoutes.index) {
+    if (!router.isReady || router.pathname !== AppRoutes.index || hasSafes === undefined) {
```
usame-algan, line 13: The issue now is that if I don't have any safes added I will be stuck on the index page. I can't reproduce the issue that Franco found. I have safes added and opening the index page redirects me to the account page and not the welcome page which is what we want.

src/pages/index.tsx
```diff
@@ -2,25 +2,28 @@ import { useEffect } from 'react'
 import type { NextPage } from 'next'
 import { useRouter } from 'next/router'
 import { AppRoutes } from '@/config/routes'
-import { useHasSafes } from '@/components/welcome/MyAccounts/useAllSafes'
+import { isEmpty } from 'lodash'
+
+const ADDED_SAFES_KEY = 'SAFE_v2__addedSafes'
```
katspaugh, line 7: 🙅 
We have a hook called useLocalStorage that adds this prefix, please use that.

usame-algan: Does the hook return undefined initially until the data is loaded from local storage? In that case we would need to find a way to stall the useEffect from running since we don't have a loading flag

katspaugh: Hooks run sequentially, and localStorage itself isn't async.

jmealy: When using the useLocalStorage hook I get the same issue as there was initially, the value comes as undefined initially and it redirects to the welcome page where it should go to accounts

usame-algan (approved):

src/pages/index.tsx
usame-algan: I think in this case we have to access the local storage directly as otherwise the useEffect hook will run and redirect away from the index page since the initial value of what useLocalStorage returns is undefined but it will also be undefined if there is nothing in local storage so we can't differentiate that.

katspaugh: Then you can use the Storage service directly, you can copy if from the the useLocalStorage hook.

usame-algan (approved):

src/pages/index.tsx
usame-algan: Quickly checked and it looks like we can also adjust the `useLocalStorage` hook to use an initialValue. I will open a separate refactor PR since that hook is used in a lot of places and we should run regression tests on that change.

katspaugh (approved):

src/pages/index.tsx
```diff
@@ -2,25 +2,30 @@ import { useEffect } from 'react'
 import type { NextPage } from 'next'
 import { useRouter } from 'next/router'
 import { AppRoutes } from '@/config/routes'
-import { useHasSafes } from '@/components/welcome/MyAccounts/useAllSafes'
+import { isEmpty } from 'lodash'
+import local from '@/services/local-storage/local'
+import type { AddedSafesState } from '@/store/addedSafesSlice'
+
+const ADDED_SAFES_KEY = 'addedSafes'
```
katspaugh, line 9: This could be `addedSlice.name` because that's how it's stored.

PR #3400, opened 2024-03-06
Fix: don't fetch manifest for unknown Safe Apps

katspaugh:
## What it solves

When you open a URL like `/apps/open?appUrl=https://...` or `/apps/share?appUrl=https://...` , we were previously immediately fetching a manifest.json from that external URL before the user had accepted a risk disclaimer.

## How this PR fixes it
With this PR, we'll be fetching manifest.json only for the apps that are listed in our config service. This is also not ideal. Ideally, we should fetch manifest.json only after the disclaimer has been accepted even for "official" apps, but that's a larger refactor and I'll do it in another PR.

## How to test it

* Open /apps/open?appUrl=https://...` or `/apps/share?appUrl=https://...` with an arbitrary URL that is not in the Safe Apps list
* Check in the network panel that no manifest.json is loaded until the consent is given


francovenica:
Checked with Ivan together. Looks good.

Note, this is only for the "/apps/open" endpoint, not the /apps/share one

cypress/e2e/safe-apps/browser_permissions.cy.js
```diff
@@ -18,13 +18,14 @@ describe('Browser permissions tests', () => {
...
-  it('Verify a permissions slide to the user is displayed', () => {
+  // @TODO: unknown apps don't have permissions
+  xit('Verify a permissions slide to the user is displayed', () => {
     safeapps.clickOnContinueBtn()
     safeapps.verifyCameraCheckBoxExists()
     safeapps.verifyMicrofoneCheckBoxExists()
   })
 
-  it('Verify the selection can be changed, accepted and stored', () => {
+  xit('Verify the selection can be changed, accepted and stored', () => {
```
katspaugh, line 28: @safe-global/safe-qa FYI I've commented out these tests temporarily as unknown apps won't have browser permissions displayed anymore, but this'll be restored in another PR.

usame-algan (approved):
👍

PR #3402, opened 2024-03-06
fix: detect native transfers

schmanu:
## What it solves

Resolves [Notion issue](https://www.notion.so/safe-global/NativeToken-transfer-ETH-is-marked-as-Unknown-contract-interaction-1f5ed836073245cbb9962daf47485c41)

## How this PR fixes it
Does not require tx.data to be defined when checking for native transfers.

## How to test it
- Send a multisend with native transfers
- Observe that they are shown correctly

## Checklist


liliya-soroka:
Verified. Drain account tests are also passed in the regression run

usame-algan (approved):
Good catch!

PR #3408, opened 2024-03-07
fix: max button not applying correct value

### LINKED ISSUE #3326: "Max" button doesn't work if a token with spending limit was previously selected
## Bug description
TL;DR: Check the gif

In send funds, the "Max" sets the value to the max amount you have of a specific token or the max allowed by the spending limit created for that user.
If you have a token that has a spending limit, select using it, then switch to another token that doesn't have a spending limit set for the user and use the "Max" button, the amount would remain at 0
I assume that, even the spending limit option is not being displayed, is still "selected" for that new token

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet
 - Env: Current prod 1.30

## Steps to reproduce
 1. Go to safe where you have 2 tokens
 2. Create a spending limit for one of them, for the user you are corrently connected to
 3. Go to send funds
 4. Select the token you created the spending limit for
 5. Select "use spending limit"
 6. Switch to the other token with the token selector
 7. Click max

## Expected result
It should autocomplete the amount input to the max of that token

## Obtained result
Sets a 0

## Screenshots
max button

compojoom:
## What it solves
fixes #3326

If you have a token that has a spending limit, select it, then switch to another token that doesn't have a spending limit set then the "Max" button would update the amount field to 0 instead of the max available.

## How this PR fixes it
The whole problem stems from the fact that we watch for the type of the transfer with react-hook-form, but if we just unmount the component that displays those limit controls the watch doesn't know that the type has to be reset back to default.

## How to test it
follow the steps in #3326

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/TokenTransfer/SpendingLimitRow/index.tsx
+++ b/src/components/tx-flow/flows/TokenTransfer/SpendingLimitRow/index.tsx
@@ -11,7 +11,7 @@ import { HelpCenterArticle } from '@/config/constants'
 
 import css from './styles.module.css'
 import { TokenAmountFields } from '@/components/common/TokenAmountInput'
-import { useContext } from 'react'
+import { useContext, useEffect } from 'react'
 import { SafeTxContext } from '@/components/tx-flow/SafeTxProvider'
 
 const SpendingLimitRow = ({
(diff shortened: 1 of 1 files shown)


francovenica:
LGTM.

usame-algan (approved):

PR #3409, opened 2024-03-07
fix: Use chain logos from config service

### LINKED ISSUE #3379: Load chain logos from CGW
Chain logos in the NetworkSelector and ChainIndicator should be taken from the chains config from CGW/config-service.
They are currently hardcoded.

Unfortunately, CGW failed to implement SVG support, so we'll have to convert the existng SVG logos to PNGs.

## Requirements
* Covert all SVGs to 2x res PNGs
* Tinify them via tinypng.com
* Upload to config-service
* Replace hardcoded files with chainLogoUri field form the configs

usame-algan:
## What it solves

Resolves #3379

## How this PR fixes it

- Uses `chainConfig.chainLogoUri` as the primary image source

Note: We can delete `getChainLogo` once logos are uploaded to the config service on staging and prod

## How to test it

1. Open the app
2. Open the network selector
3. All logos should still be visible and have the same quality

## Screenshots

Screenshot 2024-03-07 at 12 50 44

## Checklist


usame-algan:
Not a huge issue but we could also add a placeholder Skeleton since the area is blank as long as the logos are loading.

src/components/common/ChainIndicator/index.tsx
```diff
@@ -67,7 +68,7 @@ const ChainIndicator = ({
     >
       {showLogo && (
         <img
-          src={getChainLogo(chainConfig.chainId)}
+          src={chainConfig.chainLogoUri || getChainLogo(chainConfig.chainId)}
```
katspaugh, line 71: The fallback should be removed now, since they icons are in prod.

katspaugh (approved):
👍

katspaugh:
Tested, looks exactly the same as before. ✅ 

Screenshot 2024-03-13 at 15 33 22
Screenshot 2024-03-13 at 15 34 03

PR #3416, opened 2024-03-11
Combine safe lists

### LINKED ISSUE #3418: Combine Watchlist and owned safes into one list
## What is the feature about
Redesign the safe lists on the welcome page and in the sidebar according to the designs.

## The list of requirements
- [ ] combine watchlist and owned safes into one list. 
- [ ] add a read-only tag for safes that are not owned.
- [ ] add a bookmarking feature which saves safes locally. Bookmarking and un-bookmarking is a single click. It does not add names to the safes or its owners.

## Designs/sketches
https://www.figma.com/file/WSpulDIM1EoSCI6Ph9QdWq/Seedless-onboarding?type=design&node-id=2365-22980&mode=design&t=ZFkC1LQCa6UPcSio-0

jmealy:
## What it solves

Resolves #3418

## How this PR fixes it
- combines the watchlist and owned safes into one list. 
- allows users to bookmark and un-bookmark safes in their list. Bookmarked safes are stored locally and will be displayed also when there is no external wallet connected.

## How to test it
- Open the safe list either in the accounts page of the sidebar of the dashboard with a wallet connected.
- Safes that were added previously should be shown as bookmarked
- Bookmark a safe. Disconnect from your wallet. The bookmarked safes should still be shown (with read-only tag). The rest should not be shown. 
- Click to un-bookmark a safe. It should be removed from the list. 

## Screenshots
connected:
image

disconnected:
image

## Checklist


katspaugh:
@jmealy thanks for the PR! Looks like the design will be changed yet again, so I'll close this PR for now.

PR #3417, opened 2024-03-11
chore: <p> cannot appear as a descendant of <p>.

compojoom:
note to testers: I can't see this error in production, it seems to only appear with the local dev release, so you'll have to test this locally

fix: Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.

In some locations we had <Typography variant=“body1”> nested inside another <Typography> this renders 2 <p> tags inside of each-other and that is not valid. The other way around was having a <Box> that is nested inside of a Typograpühy which also seems to be not valid.

## How to test it
run the website with 'yarn dev' -> open the wallet box on top -> you shouldn't see:
grafik

## Checklist

### DIFF
--- a/src/components/common/WalletInfo/index.tsx
+++ b/src/components/common/WalletInfo/index.tsx
@@ -90,7 +90,7 @@ export const WalletInfo = ({
         ) : (
           <>
             <WalletIdenticon wallet={wallet} size={36} />
-            <Typography variant="body2" className={css.address}>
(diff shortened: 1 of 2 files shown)


katspaugh (commented):
This solution isn’t very good for semantics/a11y though, is it? I would rather we find the places where this warning is thrown and fix just them.

katspaugh (approved):
👍

PR #3421, opened 2024-03-12
Tests: fix recover test

mike10ca:
## How this PR fixes it

- This task improves recovery tests by using API calls. It also adds some waits to help the tests run.
- We are still facing issues with timeouts from external which out of our control. This will be addressed later.

## How to test it


francovenica (approved):
LGTM

PR #3425, opened 2024-03-12
feat: speedup pending transactions

### LINKED ISSUE #3107: Speed-up pending transaction
## What is the feature about

Currently, when submitting a transaction it will just say "Processing" and there is no option to speed-up that transaction through the UI. Some wallets like MetaMask offer this option but others don't which can result in the transaction being stuck until it times out on our end.

## The list of requirements

- Offer a "speed-up" option when a transaction is processing that resubmits it with the same wallet nonce

## Designs/sketches

TBD

## Links

TBD

schmanu:
## What it solves
Stuck transactions due to moving gas prices disappear after a 1 minute timeout.

Resolves #3107 

## How this PR fixes it
- Shows pending transactions and offers a Speedup button after the timeout instead of showing an error and removing them
- Sped up transactions use the current gasPrice but speed it up by
   - Multiplying the `maxPriorityFeePerGas`by two
   - Adjust the `maxFeePerGas` accordingly

## How to test it
- Submit a transaction and overwrite the baseFee in MetaMask to a very low value (e.g. 1)
- Click the speedup button and resubmit the tx

## Checklist


src/components/common/Notifications/index.tsx
```diff
@@ -4,17 +4,61 @@ import groupBy from 'lodash/groupBy'
...
 import Track from '../Track'
 import { isRelativeUrl } from '@/utils/url'
+import { type PendingStatus } from '@/store/pendingTxsSlice'
+import useChainId from '@/hooks/useChainId'
+import useSafeAddress from '@/hooks/useSafeAddress'
+import { STATUS_LABELS } from '@/hooks/useTransactionStatus'
 
 const toastStyle = { position: 'static', margin: 1 }
 
+const CircularProgressWithLabel = (props: CircularProgressProps & { value: number }) => {
```
schmanu, line 22: I think these 2 components can be removed as they were part of the sticky transactions

src/features/speedup/components/SpeedUpMonitor.tsx
```diff
@@ -0,0 +1,56 @@
...
+  modalTrigger: 'alertBox' | 'alertButton'
+}
+
+const SPEED_UP_THRESHOLD_IN_SECONDS = 15
+
+export function SpeedUpMonitor({ txDetails, txId, pendingTx, modalTrigger = 'alertBox' }: Props) {
+  const [openSpeedUpModal, setOpenSpeedUpModal] = useState(false)
+  const counter = useCounter(pendingTx?.submittedAt)
+
+  if (!counter || counter < SPEED_UP_THRESHOLD_IN_SECONDS) {
```
schmanu, line 22: We should also check the pendingTx status to be PROCESSING as we cannot really speed up relaying txs.  
For relayed tx Gelato is handling the gas price and payments.

compojoom: oh, yes. Forgot about this here

src/features/speedup/components/SpeedUpModal.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const safeAddress = useSafeAddress()
+  const hasActions = signerAddress && signerNonce !== undefined && signerAddress === wallet?.address
+  const dispatch = useAppDispatch()
+
+  const isDisabled = waitingForConfirmation || !wallet || !safeTx || !gasPrice || !txDetails || !onboard
+
+  const speedUpFee = useMemo(
+    () => ({
+      ...gasPrice,
+      maxFeePerGas: BigInt(gasPrice?.maxFeePerGas ?? 10n) + BigInt(gasPrice?.maxPriorityFeePerGas ?? 1n),
```
schmanu, line 47: We have a feature toggle for EIP 1559. 
If this is turned on we provide `maxFeePerGas` and `maxPriorityFeePerGas`
otherwise just a `gasPrice`.

(See the useGasPrice hook)

It might be easier to add a new boolean to the useGasPrice hook `speedUp: boolean` or `speedUpFactor: number` and multiply the fee there.

schmanu:
> Works really nice already! As discussed with @compojoom we need to find a way to clean up pending transactions if they were sped-up from outside the UI. Our idea is to check the txHistory whenever it updates and compare the newest nonce with the pending tx nonce. If its the same we can remove the pending tx.
> 
> I also noticed that the notifications in Figma have a linear progress bar. Did we decide not to implement that?
> Screenshot 2024-03-20 at 16 22 01

We decided to not go with sticky notifications yet as we have too many popups / overlays already. Focusing on the new logic by just adding speed up buttons to all components currently showing pending txs.

src/components/transactions/TxSummary/PendingActions.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { SpeedUpMonitor } from '@/features/speedup/components/SpeedUpMonitor'
+import useAsync from '@/hooks/useAsync'
+import { useAppSelector } from '@/store'
+import { selectPendingTxById } from '@/store/pendingTxsSlice'
+
+export const PendingActions = ({ tx }: { tx: TransactionSummary }) => {
+  const pendingTx = useAppSelector((state) => selectPendingTxById(state, tx.id))
+
+  const [txDetails, txDetailsLoading, txDetailError] = useAsync(() => {
+    if (!pendingTx) return Promise.resolve(undefined)
```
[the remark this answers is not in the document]
schmanu: This block will be shown for every tx in the queue.
Pending txs are only txs that are currently executing, processing or indexing though.
For the others it will be undefined. Sadly the types do not reflect that as we use a `Record<string, PendingTx>` here.

src/features/speedup/components/SpeedUpMonitor.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const SPEED_UP_THRESHOLD_IN_SECONDS = 15
+
+export function SpeedUpMonitor({ txDetails, txId, pendingTx, modalTrigger = 'alertBox' }: Props) {
+  const [openSpeedUpModal, setOpenSpeedUpModal] = useState(false)
+  const counter = useCounter(pendingTx?.submittedAt)
+
+  // We only care about processing txs, for everything else we don't show the speed up button
+  if (!pendingTx || pendingTx.status !== PendingStatus.PROCESSING) {
```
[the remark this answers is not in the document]
schmanu: I made sure that this component does not get rendered without pendingTx being defined and removed this checkj.

src/utils/SimpleTxWatcher.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Singleton class for watching pending txs.
+ *
+ * Offers two methods:
+ * - {@linkplain watchTxHash} to watch a new pending tx
+ * - {@linkplain stopWatchingTxHash} to stop an active watcher for a pending tx
+ */
+export class SimpleTxWatcher {
+  private static INSTANCE: SimpleTxWatcher | undefined
+  private unsubFunctions: Record<string, () => void>
```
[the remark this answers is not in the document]
compojoom: it's actually readonly so we don't need a setter.

src/components/tx-flow/flows/SuccessScreen/index.tsx
```diff
@@ -32,17 +35,33 @@ const SuccessScreen = ({ txId }: { txId: string }) => {
...
+  switch (status) {
+    case PendingStatus.PROCESSING:
+    case PendingStatus.RELAYING:
+      StatusComponent = <ProcessingStatus txId={txId} pendingTx={pendingTx} />
+      break
+    case PendingStatus.INDEXING:
+      StatusComponent = <IndexingStatus />
+      break
+    default:
+      StatusComponent = <DefaultStatus error={error} />
```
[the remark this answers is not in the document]
compojoom: the statusMessage component was packing everything in a blue container (the message: "this tx was confirmed and is now ..."), which according to @TanyaEfremova was not correct. I couldn't figure out a way how to refactor it to accomodate this, so I went for writing as is now in the successScreen.
grafik

The StatusMessage component is actually no longer used anywhere. If you have ideas how to make it work with the new requirements let me know, otherwise I would actually remove it.

src/components/transactions/TxSummary/QueueActions.tsx
```diff
@@ -8,7 +8,7 @@ const QueueActions = ({ tx }: { tx: TransactionSummary }) => {
   const awaitingExecution = isAwaitingExecution(tx.txStatus)
 
   return (
-    <Box my={-1}>
+    <Box my={-1} mr={2} display={'flex'} flexDirection={'row'} alignItems={'center'} justifyContent={'flex-end'}>
```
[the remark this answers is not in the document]
compojoom: I automatically reformat with prettier and it does display={'flex'} instead of display="flex". If we can find a way to make it so that prettier reformats it the way you suggest I'm all for this change :)

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -182,12 +188,37 @@ export const dispatchTxExecution = async (
...
   return result.hash
 }
 
+export const dispatchTxSpeedUp = async (
+  safeTx: SafeTransaction,
+  txOptions: TransactionOptions,
+  txId: string,
+  onboard: OnboardAPI,
+  chainId: SafeInfo['chainId'],
+  safeAddress: string,
```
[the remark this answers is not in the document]
compojoom: so you want:
```
interface DispatchOptions {
  txOptions: TransactionOptions;
  txId: string;
  to: string;
  data: string;
  onboard: OnboardAPI;
  chainId: SafeInfo['chainId'];
  safeAddress: string;
}
```
and then 
`export const dispatchTxSpeedUp = async (options: DispatchOptions) => {`

src/utils/SimpleTxWatcher.ts
```diff
@@ -0,0 +1,74 @@
...
+        // try to retrieve the receipt
+        const testReceipt = await provider.getTransactionReceipt(txHash)
+        if (testReceipt !== null) {
+          unsubscribe()
+          resolve(testReceipt)
+        } else {
+          // Check if tx was replaced
+          const currentNonce = await provider.getTransactionCount(walletAddress)
+          if (currentNonce > walletNonce) {
+            if (replacedBlockCount >= 2) {
```
katspaugh, line 52: Put 2 into a named constant pls.

compojoom: No, it's a magic constant number 😛

src/components/transactions/TxSummary/index.tsx
```diff
@@ -75,7 +75,7 @@ const TxSummary = ({ item, isGrouped }: TxSummaryProps): ReactElement => {
         </Box>
       )}
 
-      <Box gridArea="status">{isQueue && !isPending ? <QueueActions tx={tx} /> : <TxStatusLabel tx={tx} />}</Box>
+      <Box gridArea="status">{isQueue && !isPending ? <QueueActions tx={tx} /> : <PendingActions tx={tx} />}</Box>
```
usame-algan, line 78: I think it is a bit misleading that we are rendering `PendingActions` for transactions that are in the history because the `TxStatusLabel` was moved in there. It would be more clear to keep the separation in this component i.e.

```
<Box gridArea="status">
    {isQueue ? isPending ? <PendingActions tx={tx} /> : <QueueActions tx={tx} /> : <TxStatusLabel tx={tx} />}
</Box>
```

src/features/speedup/components/SpeedUpMonitor.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const counter = useCounter(pendingTx.submittedAt)
+  const web3ReadOnly = useWeb3ReadOnly()
+  const [transaction] = useAsync(async () => {
+    if (!pendingTx.txHash || !web3ReadOnly) return null
+    return web3ReadOnly?.getTransaction(pendingTx.txHash)
+  }, [pendingTx.txHash, web3ReadOnly])
+
+  const [smartContract] = useAsync(async () => {
+    if (!pendingTx.signerAddress || !web3ReadOnly) return false
+    return await isSmartContract(web3ReadOnly, pendingTx.signerAddress)
```
usame-algan, line 32:
```suggestion
    return isSmartContract(web3ReadOnly, pendingTx.signerAddress)
```

No need to await this

src/features/speedup/components/SpeedUpMonitor.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export const SpeedUpMonitor = ({ txId, pendingTx, modalTrigger = 'alertBox' }: SpeedUpMonitorProps) => {
+  const [openSpeedUpModal, setOpenSpeedUpModal] = useState(false)
+  const wallet = useWallet()
+  const counter = useCounter(pendingTx.submittedAt)
+  const web3ReadOnly = useWeb3ReadOnly()
+  const [transaction] = useAsync(async () => {
+    if (!pendingTx.txHash || !web3ReadOnly) return null
+    return web3ReadOnly?.getTransaction(pendingTx.txHash)
+  }, [pendingTx.txHash, web3ReadOnly])
```
usame-algan, line 28:
```suggestion
  
  const [transaction] = useAsync(async () => {
    if (!pendingTx.txHash || !web3ReadOnly) return null
    return web3ReadOnly.getTransaction(pendingTx.txHash)
  }, [pendingTx.txHash, web3ReadOnly])
```

src/features/speedup/components/SpeedUpMonitor.tsx
```diff
@@ -0,0 +1,81 @@
...
+  // We only care about processing txs, for everything else we don't show the speed up button
+  if (
+    pendingTx.status !== PendingStatus.PROCESSING ||
+    !pendingTx.txHash ||
+    !transaction ||
+    !sameAddress(pendingTx.signerAddress, wallet?.address) ||
+    smartContract
+  ) {
+    return null
+  }
```
usame-algan, line 44: I suggest to test this logic

src/features/speedup/components/SpeedUpModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+          <SvgIcon inheritViewBox component={RocketSpeedup} sx={{ width: 90, height: 90 }} />
+        </Box>
+        <Typography data-testid="speedup-summary">
+          This will speed up the pending transaction by{' '}
+          <Typography component={'span'} fontWeight={700}>
+            replacing
+          </Typography>{' '}
+          the original gas parameters with new ones.
+        </Typography>
+        <Box mt={2}>
```
usame-algan, line 117:
```suggestion
        </Box>
        
        <Typography data-testid="speedup-summary">
          This will speed up the pending transaction by{' '}
          <Typography component={'span'} fontWeight={700}>
            replacing
          </Typography>{' '}
          the original gas parameters with new ones.
        </Typography>
        
        <Box mt={2}>
```

A newline between blocks makes it easier to read.

src/components/tx-flow/flows/SuccessScreen/index.tsx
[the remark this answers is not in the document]
compojoom: I see that you use a similar StatusMessage component in safe creation. The main difference is in the steps and a bit in the UI. The problem here is that if we are in the processing state we need the pendingTx to display the speed up. So it is no longer possible to just pass a `status` and be done with it. If you have some ideas how to abstract this let me know.

src/hooks/usePendingTxs.ts
```diff
@@ -62,6 +62,17 @@ export const usePendingTxsQueue = (): {
...
+      if (results.filter((item) => isTransactionListItem(item)).length === 0) {
+        results.splice(0, 1)
+      }
+    }
+
+    if (results[1] && isConflictHeaderListItem(results[1])) {
+      // Check if we both conflicting txs are still pending
+      if (results.filter((item) => isTransactionListItem(item)).length <= 1) {
+        results.splice(1, 1)
+      }
```
usame-algan, line 75: Not sure I understand these two conditions. If the first element is a label and there is nothing else we remove the label. Why not check for `results.length === 1` in that case?

For the second condition I don't see how it is connected to pending transactions. If we filter all of them out anyway does it matter if one or two are pending?

src/store/txHistorySlice.ts
```diff
@@ -34,10 +34,13 @@ export const txHistoryListener = (listenerMiddleware: typeof listenerMiddlewareI
 
         const txId = result.transaction.id
 
+        const txHash = pendingTxs[txId]?.txHash
+
         if (pendingTxs[txId]) {
           txDispatch(TxEvent.SUCCESS, {
             txId,
             groupKey: pendingTxs[txId].groupKey,
+            txHash: txHash,
```
usame-algan, line 43:
```suggestion
            txHash,
```

src/components/transactions/TxSummary/QueueActions.tsx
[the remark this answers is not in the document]
usame-algan: I think there is an [eslint plugin](https://github.com/jsx-eslint/eslint-plugin-react/blob/HEAD/docs/rules/jsx-curly-brace-presence.md) out there to automate this but so far we manually made sure to use one over the other. Imo it doesn't matter as long as it is consistent.

src/components/tx-flow/flows/SuccessScreen/index.tsx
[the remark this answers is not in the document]
usame-algan: We could render the `pendingTx` part as a child i.e.

```
<StatusMessage status={status} error={error}>
    {pendingTx && <SpeedUpMonitor txId={txId} pendingTx={pendingTx} modalTrigger={'alertBox'} />}
</StatusMessage>
```

That way we could reuse all the other elements and their styles.

src/services/tx/tx-sender/dispatch.ts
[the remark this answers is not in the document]
usame-algan: I just noticed that `dispatchTxExecution` and `dispatchTxSpeedUp` don't have the same props so abstracting (part) of the props wouldn't really give us much. Lets leave it as is.

src/hooks/usePendingTxs.ts
schmanu: You could execute the same tx twice. Once with i.e. metamask and once with e.g. ledger. So there is in theory the possibility to have conflicting pending txs. But if only one of the untrusted queue txs is pending we have to remove the conflicting tx label.

compojoom:
> Works really nice already! As discussed with @compojoom we need to find a way to clean up pending transactions if they were sped-up from outside the UI. Our idea is to check the txHistory whenever it updates and compare the newest nonce with the pending tx nonce. If its the same we can remove the pending tx.
> 
> I also noticed that the notifications in Figma have a linear progress bar. Did we decide not to implement that?
> Screenshot 2024-03-20 at 16 22 01

We said that we are going to refactor the notifications and then we are going this notification.

src/components/transactions/TxSummary/index.tsx
compojoom: Good point. I'll try to come up with something. `?? ::` too complicated :D

compojoom: Now should be better. I realized that queue and pending are basically the same. So moved the pending stuff to the queue (you can't have pending in the history tab). Also added a new grid cell for actions. Now the UI no longer jumps

usame-algan (approved):

src/components/transactions/TxSummary/index.tsx
usame-algan: Thanks!

src/components/tx-flow/flows/SuccessScreen/index.tsx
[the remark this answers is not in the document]
usame-algan: How about having it inside `StatusMessage` and only rendering it if `status == PROCESSING` then?

src/features/speedup/components/SpeedUpMonitor.tsx
compojoom: Ok, check again.

francovenica:
The execute button looks broken in this PR. It doesn't look like that in current dev:
image

francovenica:
Question. If the tx is sped up in the wallet itself (like MM extension or MM mobile), should the safe app catch that and show the updated tx hash?
Is not happening, but since is in the test plan in Notion, I wonder if it is even possible to do
image

compojoom:
> Question. If the tx is sped up in the wallet itself (like MM extension or MM mobile), should the safe app catch that and show the updated tx hash? Is not happening, but since is in the test plan in Notion, I wonder if it is even possible to do image

Unfortunately this is not possible. If the user speeds up in MM we don't know this. We only find out about it later when suddenly a tx is included on the chain with the same nonce.

francovenica:
An issue I have with MM mobile.

Speeding up the tx in the Safe App creates a 2nd transaction in the MM phone app that is meant to replace the first one. This makes the safe app to get stuck waiting for that sped up transaction
This doesn't happen if you speed it up in the MM phone app. When you speed it up there, there is no "2nd transaction" shown in the list, is just the first one being replaced right away. With this the tx is executed normally.

Steps to make it clearly:
Speeding it up in the MM phone app
- Create a tx (send funds) with 0.000000001 gas fee
- Execute in the phone
- Check the list of tx in the phone app. 
- Tap the 1st tx and check in blockexplorer to see the Tx hash
- Come back and speed it up with "aggresive" settings (it sets the gas fee as 2 Gwei like we do in our speed up). Save and execute
- Check the list of tx in the phone app again. Still only 1 tx is at the top as pending
- Check the tx hash on that transaction. It should be a new hash
- The tx executes

Speeding it up in the Safe app
- Create a tx (send funds) with 0.000000001 gas fee
- Execute in the phone
- Check the list of tx in the phone app. 
- 1 transaction is at the top being pending
- Check the safe app and speed up the transaction there
- Check the list of tx in the phone app. 
- Now 2 transactions are in peding
- Wait here as long as you want, the transaction will not execute
- In the phone app, tap the 2nd tx in the list (The one that was created first) and cancel it
- Now the newest transaction (the one created as you speed it up in the safe app) should execute in a few seconds

image

francovenica:
Question: In polygon the speed up option doesn't show up. Waited for like 5 mins. Is the feature not present there? (I've checked the admin and I didn't see a feature toggle for it)

francovenica:
Execution of a batched tx works fine, but looks confusing. The speed up feature show in all of them and if you use one it just dissapears from one of the tx

Idk if there is anythign here to fix without design helping out

Note: This is an edge case since we don't allow gas fee edition in tx in bulk, you have to edit it manually in your wallet. So this is low prio

image

There are less tx in this snapshot because it was a 2nd attempt, but it happened the same for the 3 tx above
image

compojoom:
> An issue I have with MM mobile.
> 
> Speeding up the tx in the Safe App creates a 2nd transaction in the MM phone app that is meant to replace the first one. This makes the safe app to get stuck waiting for that sped up transaction This doesn't happen if you speed it up in the MM phone app. When you speed it up there, there is no "2nd transaction" shown in the list, is just the first one being replaced right away. With this the tx is executed normally.
> 
> Steps to make it clearly: Speeding it up in the MM phone app
> 
>     * Create a tx (send funds) with 0.000000001 gas fee
> 
>     * Execute in the phone
> 
>     * Check the list of tx in the phone app.
> 
>     * Tap the 1st tx and check in blockexplorer to see the Tx hash
> 
>     * Come back and speed it up with "aggresive" settings (it sets the gas fee as 2 Gwei like we do in our speed up). Save and execute
> 
>     * Check the list of tx in the phone app again. Still only 1 tx is at the top as pending
> 
>     * Check the tx hash on that transaction. It should be a new hash
> 
>     * The tx executes
> 
> 
> Speeding it up in the Safe app
> 
>     * Create a tx (send funds) with 0.000000001 gas fee
> 
>     * Execute in the phone
> 
>     * Check the list of tx in the phone app.
> 
>     * 1 transaction is at the top being pending
> 
>     * Check the safe app and speed up the transaction there
> 
>     * Check the list of tx in the phone app.
> 
>     * Now 2 transactions are in peding
> 
>     * Wait here as long as you want, the transaction will not execute
> 
>     * In the phone app, tap the 2nd tx in the list (The one that was created first) and cancel it
> 
>     * Now the newest transaction (the one created as you speed it up in the safe app) should execute in a few seconds
> 
> 
> image

This is a bug in MM Mobile. I've reported it here https://github.com/MetaMask/metamask-mobile/issues/9118
Right now there is nothing in our side we could do to overcome it. Maybe we can follow up the suggestion to offer speed up only on whitelisted wallets.

compojoom:
@francovenica I just rebased, please try again once a new version is deployed. With the latest changes that @schmanu did you should be able to speed up txs on polygon as well. The problem there was that we were polling the rpc node for the transaction status, but if the transaction is submitted with a too low gas then no RPC is indexing the TX. Now, we are storing the tx data locally and are able to offer speed up even if the RPC doesn't know the TX.

francovenica:
@compojoom @schmanu 
Cool. I'll give it a try.

Can we fix the visual issues like this one? so we can create the RC
image

francovenica:
Now that Poly works I decided to try a few phone apps
Zerion:
It completely ignores our suggestion from safe. So if the tx is slow, even the speedup we send is going to have the fee ignored. Is up to the user to change the fee manually in his phone

1inch:
If you lower the suggested fee we send it by just 20% the app would show an error, it won't even allow changing the fee we suggest
image

Rainbow:
It also ignores our suggestion. the dapp doesn't even allow custom gas fee changes, it gives you 3 speeds and that's it. So our speed up feature cannot be used here
image

Trust:
This receives our suggestion, and you can edit them in the phone app as well, but if you set something too low it won't even let you try to execute the tx.
image

Also this trust app has issues.
First it already tells you the tx will fail in the form:
image

The tx doesn't go through in our app and a ton of errors show up in the console, but in the blockexplorer you can see the tx actually executed just fine
https://polygonscan.com/tx/0x8c4908720dd7063fb6d1dae6c2f5803f49fc89414197b53c5c380570e1545759
image


Conclussion:
It seems that beyond MM that kinda allows anything for the gas fee, this dapps seems to have safety features to assure users don't get stuck by setting the gas too low, so probably we won't have to worry about Mobile phone apps connected through WC

compojoom:
Btw, just tested with Rabby and it has the same issue as MM Mobile. It ignores the nonce we provide and instead sets one higher. This leads to a stuck tx.
grafik

The user would have to speed up the first tx and then the second will execute.

src/components/common/Notifications/useCounter.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from 'react'
+
+export const useCounter = (startTimestamp: number | undefined) => {
```
katspaugh, line 3: Can't we use the existing `useIntervalCounter`?

src/hooks/useTxPendingStatuses.ts
```diff
@@ -74,42 +79,160 @@ const useTxPendingStatuses = (): void => {
...
+              safeAddress,
+              txId,
+              status: PendingStatus.PROCESSING,
+              txHash: detail.txHash,
+              signerAddress: detail.signerAddress,
+              signerNonce: detail.signerNonce,
+              submittedAt: Date.now(),
+              gasLimit: detail.gasLimit,
+              txType: PendingTxType.SAFE_TX,
+            }
```
katspaugh, line 144: Is it possible to reduce the code duplication here?

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -132,6 +132,93 @@ export const dispatchOnChainSigning = async (
...
+
+  const provider = getWeb3ReadOnly()
+
+  if (provider) {
+    // don't await as we don't want to block
+    waitForTx(provider, [txId], result.hash, safeAddress, wallet.address, signerNonce)
+  }
+
+  return result.hash
+}
```
katspaugh, line 220: Again, these are extremely similar, possible to exract the common part?

francovenica:
The ticket is not in QA, but still I've checked the fix regarding the buttons positions.

I don't know if I agree with the last commit. I think a big point of this feature is that wallets that do not have the speed up option like physical devices (trezor/ledger) have now the chance of speeding it up.
I understand the concern that causes being able to speed up a tx that might get replaced by another, but at least on the safe side we don't allow a new tx to be executed while another one is in the process of being executed, so the only scenario where I think it could happen what is being mentioned in the commit is if the user uses his wallet in another app that is not the safe.

katspaugh:
@francovenica speeding up is not disabled for hw wallets. It's only disabled for immediately executed tx (unsigned txs) with all wallets.

francovenica:
@katspaugh I'm aware of that. but most people in safes 1/1 dont' sign, queue the tx and then execute, they execute right away, making these particular trezor/ledger users prone to see a message saying "go and speed it up in your wallet" whne they cannot do that. I still think the risk we are trying to avoid are not worth the disconfort of these users.

katspaugh:
I see, thanks for clarifying. That's a fair point. However, consider the following:

- This feature wasn't there before, so we're not making existing UX worse.
- Ledger is 1.2% of our users, Trezor is 0.36%
- The risk when speeding up an unsigned tx is that you can submit the same tx twice (due to the nonce not being validated for immediate executions on the contract) – so you can potentially send your money twice if we don't disable this – IMHO it's definitely worth it to avoid this risk.

PR #3426, opened 2024-03-12
fix: block new social login creations

schmanu:
## What it solves
Approaching the deprecation of Social logins, we do not want new users to sign up with social signers.

## How this PR fixes it
If a single factor (initial setup) account it younger than 5 minutes we error out of the login / social signer creation.

## How to test it
- If you have an old account, delete it first
- Login with an unused google login
- observe new error

## Screenshots
Screenshot 2024-03-12 at 15 21 51

## Checklist

### DIFF
--- a/src/services/mpc/SocialWalletService.ts
+++ b/src/services/mpc/SocialWalletService.ts
@@ -10,6 +10,8 @@ import { asError } from '../exceptions/utils'
 import { type ISocialWalletService } from './interfaces'
 import { isSocialWalletOptions, SOCIAL_WALLET_OPTIONS } from './config'
 
+const ERROR_MSG_NO_NEW_ACCOUNTS = 'No new account creations are allowed as this feature will be deprecated soon.'
+
 /**
  * Singleton Service for accessing the social login wallet
  */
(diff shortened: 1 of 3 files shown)


katspaugh (approved):

src/services/mpc/SocialWalletService.ts
```diff
@@ -10,6 +10,8 @@ import { asError } from '../exceptions/utils'
 import { type ISocialWalletService } from './interfaces'
 import { isSocialWalletOptions, SOCIAL_WALLET_OPTIONS } from './config'
 
+const ERROR_MSG_NO_NEW_ACCOUNTS = 'No new account creations are allowed as this feature will be deprecated soon.'
```
katspaugh, line 13:
```suggestion
const ERROR_MSG_NO_NEW_ACCOUNTS = 'Social Login is deprecated and will be removed on 01.05.2024. New accounts may not be created.'
```

francovenica:
This PR needs to be whitelisted for me to use the Social login feature

katspaugh:
Let's test it on dev

francovenica:
Tested it in dev.

For existing account I can connect just fine
Deleted my account, then try to connect again re-creating the account. I got the expected error message 
LGTM

image

PR #3427, opened 2024-03-12
fix: Don't allow duplicate signers in safe creation

### LINKED ISSUE #3415: Creating a New Safe Account Allows Duplicate Signers
## Bug description
When attempting to create a new Safe Account, the system permits users to select multiple signers using the autocomplete feature from address entries.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
1. Go to the Safe Account creation page.
2. Begin adding signers using the autocomplete feature from address entries.
3. Select the same address for multiple signers.

## Expected result
The system should prevent the selection of duplicate signers.

## Obtained result
The system allows selection of duplicate signers.

## Screenshots
image

usame-algan:
## What it solves

Resolves #3415 

The fix implemented in #3060 added a new `onChange` call to the `Autocomplete` to render the `AddressInputReadOnly` whenever the user selects an address book entry in the input field and display validation errors in this separate component. 

However if a `deps` array is also passed to `AddressBookInput` it triggers a revalidation which is not propagated because `AddressInputReadOnly` is rendered instead of `AddressInput`.

## How this PR fixes it

- Simplify and move `AddressInputReadOnly` into `AddressInput`
- Add the read-only logic into `AddressInput`
- Adds a new `useChain` hook

## How to test it

1. Create a new Safe
2. Add the same owner twice
3. Observe an error is visible

Make sure that the fix from #3060 is still working:

1. Open an existing safe
2. Go to the owner settings and add a new owner
3. Choose the safe itself as an owner
4. Observe an error is visible

Check that other places where `AddressInput` and `AddressBookInput` are used are still working
- Create address book entry
- Load safe
- TxHistory filter
- Create transactions
- Recovery

## Screenshots
Screenshot 2024-03-12 at 16 10 22
Screenshot 2024-03-12 at 16 10 43

## Checklist


schmanu (commented):
I think it really helps to merge these components 👍 

Could we add a test for this new fix? There is a AddressBookInput unit test already.

francovenica:
I created a e2e test for this. I'll make my PR once this ticket is merged to dev

francovenica:
LGTM
The duplication error works fine for the autocomplete feature and by copy/pasting the address as well
dup signer fix

schmanu (approved):

PR #3429, opened 2024-03-12
fix: allow subscribing to owned but not added safes

### LINKED ISSUE #3288: Can't subscribe to notifications for owned un-added safe
## Bug description
can't subscribe to notifications for owned un-added safe

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: any

## Steps to reproduce
 1. Connect your owner that has a few safes on different networks ( watchlist is empty)
 2. go to the settings->notifications -> global preferences 
 Current result: there is no possibility to subscribe to notifications  for owned un-added safes

## Expected result
Make sure that it's possible to subscribe to notifications for all owned safes of connected owner

jmealy:
## What it solves

Resolves #3288

## How this PR fixes it
- Do not show added but not owned safes in global notification settings.
- Show owned safes plus added & owned safes.
- Remove the notification banner completely.

## How to test it
1. Go to the settings->notifications -> global preferences
2. You should see the option to add notifications to only safes that are owned by the connected wallet.
---
1. Observe that the notification banner does not show up under any conditions.

## Checklist


src/components/settings/PushNotifications/GlobalPushNotifications.tsx
```diff
@@ -93,22 +94,30 @@ export const _sanitizeNotifiableSafes = (
...
+    ]),
+  )
 
+  let notifiableSafes: NotifiableSafes = {}
+  for (const chainId of chains) {
+    const ownedSafesOnChain = ownedSafes?.[chainId] ?? []
+    const addedSafesOnChain = added[chainId] || []
+    const currentSubscriptionsOnChain = currentSubscriptions?.[chainId] || []
+    const uniqueSafeAddresses = Array.from(
+      new Set([...currentSubscriptionsOnChain, ...addedSafesOnChain, ...ownedSafesOnChain]),
```
usame-algan, line 119: I think we need to filter `addedSafesOnChain` and remove all safes that are not owned otherwise the user can see un-owned safes that are in their watchlist on the global preferences page.

cypress/e2e/regression/add_owner.cy.js
```diff
@@ -2,11 +2,19 @@ import * as constants from '../../support/constants'
...
-    cy.visit(constants.setupUrl + constants.SEPOLIA_TEST_SAFE_1)
     cy.clearLocalStorage()
+    cy.visit(constants.setupUrl + constants.SEPOLIA_TEST_SAFE_1, {
+      onBeforeLoad: (win) => {
+        win.localStorage.setItem(
+          constants.localStorageKeys.SAFE_v2__dismissPushNotifications,
+          JSON.stringify(ls.dismissedNotifications),
+        )
+      },
+    })
```
usame-algan, line 17: Do we have to add this and similar changes to our e2e tests because the banner is now always visible for owned safes and we want to hide it there?

src/components/settings/PushNotifications/PushNotificationsBanner/index.tsx
usame-algan, line 1: The `Enable all` button in this component is still checking `addedSafes` and is not visible to the user unless they have something in their watch list. We should discuss if we want to adjust its logic to check `ownedSafes` instead.

usame-algan (commented):
Considering that we have the notification feature for a while now, how complex the `PushNotificationBanner` component is and that we already received feedback in the past to reduce the number of popups I would be in favour of removing it completely. Lets discuss it with the team.

src/components/settings/PushNotifications/GlobalPushNotifications.tsx
jmealy: I assumed there was some reason for allowing to subscribe to notifications even if they don't own the safe but I guess there wouldn't be use for this. So yeah I can just show owned

cypress/e2e/regression/add_owner.cy.js
jmealy: yeah this is just because the banner is not showing up for some of the tests and it blocks elements that it's trying to click. So if we end up getting rid of the banner we can get rid of this also

usame-algan:
@jmealy discussed it this morning and others are also in favour of removing the banner. @katspaugh suggests to rename the   "Settings" button to "Push notification settings"
Screenshot 2024-03-18 at 10 08 17

jmealy:
> @jmealy discussed it this morning and others are also in favour of removing the banner. @katspaugh suggests to rename the "Settings" button to "Push notification settings"

brilliant, thanks Usame 🙏

usame-algan (approved):

usame-algan:
@mike10ca do we need to update something in Testrail now that the Push Notification Banner has been removed?

mike10ca:
> @mike10ca do we need to update something in Testrail now that the Push Notification Banner has been removed?

@usame-algan no, since automation does not cover notifications to date.

francovenica:
Are we renaming this button as "Push notifications settings" or I missunderstood and it will remain as "settings"?
image

francovenica:
The rest looks good

Safe you added but don't own doesn't show up in the global settings
The notification popUp doesn't show up anymore (checked for CF safes after deployed as well)
Tested notifications in general (incomming, outgoing tx and tx that need signing)

If the button will remain as "Settings" then the ticket can be merged. If not, change the test and don't put it in QA, just merge it

PR #3432, opened 2024-03-13
Feat: spam filter for incoming txs

### LINKED ISSUE #3334: "Hide/show unknown" button in the history tab is not working
## Bug description
The button is supposed to hide/show bogus tx's, but is not doing that when you click it

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to this safe
 2. Filter by "incoming tx" so thelist show all the receivings tx (this safe has a lot of bogus tx of that kind)
 3. Use the "Hide/Show unknown" button

## Expected result
bogus tx should be hidden/shown

## Obtained result
The list just refreshes, nothing is hidden

## Screenshots
hide tokens

katspaugh:
## What it solves

Resolves #3334

## How this PR fixes it

Passes the trusted flag to the filtered history endpoints (although it only works for the incoming tx filter).

## How to test it

* Open https://filter_incoming--walletweb.review-wallet-web.5afe.dev/transactions/history?safe=eth:0x220866b1a2219f40e72f5c628b65d54268ca3a9d&type=Incoming
* Press "Hide unknown transactions"
* The tx list should be filtered

---

Note: this PR is based off of the Tx Deletion PR because it requires an updated Gateway SDK.

### DIFF
--- a/src/hooks/useTxHistory.ts
+++ b/src/hooks/useTxHistory.ts
@@ -22,6 +22,7 @@ const useTxHistory = (
   const [filter] = useTxFilter()
   const { showOnlyTrustedTransactions } = useAppSelector(selectSettings)
   const hasDefaultTokenlist = useHasFeature(FEATURES.DEFAULT_TOKENLIST)
+  const onlyTrusted = (hasDefaultTokenlist && showOnlyTrustedTransactions) || false
 
   const {
     safe: { chainId },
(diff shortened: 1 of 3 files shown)


iamacook (approved):

francovenica:
LGTM
The show/hide button works fitler is working (only with incomming tx as we said)

Note:
I noticed that the counter of "X tx found" is only working with page 1 (so the first 20 txs), so if you have 15 tokens spam, it will tell you that only "> 5 tx has been found" even tho there might be a lot more pages of proper tx to show, so it should still say "> 20 tx found"
I've never seen this issue since without the show/hide button wasn't working

But I consider this a different issue. I'll report it in another ticket

PR #3437, opened 2024-03-13
fix: cannot connect with WC to Opensea

compojoom:
If the user is trying to connect with WC to Opensea and the user is on a chain other than mainnet the first connection attempt fails. The second one succeeds. This is because when opensea sees the list of optional chains, after establishing a session it sends a request, but the request is for the optional chain and not for the required one and we end up blocking it. By giving up on the optional namespace Opensea sends the request to the requried chain.

## How to test it
Go to opensea to a collection that is on let's say OP. Try to establish WC connection to your safe that is also on OP. Establishing a connection should work.

## Checklist

### DIFF
--- a/src/features/walletconnect/services/WalletConnectWallet.ts
+++ b/src/features/walletconnect/services/WalletConnectWallet.ts
@@ -89,7 +89,8 @@ class WalletConnectWallet {
 
     const supportedChainIds = [currentChainId].concat(
       requiredChains.map(stripEip155Prefix),
-      optionalChains.map(stripEip155Prefix),
+      // disable optional chains for now as some dapps end up sending request after the session is approved to the wrong chain
+      // optionalChains.map(stripEip155Prefix),
     )
 
     const eip155ChainIds = supportedChainIds.map(getEip155ChainId)
--- a/src/features/walletconnect/services/WalletConnectWallet.ts
+++ b/src/features/walletconnect/services/WalletConnectWallet.ts
@@ -81,16 +81,18 @@ class WalletConnectWallet {
     })
   }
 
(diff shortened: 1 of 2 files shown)


compojoom:
@katspaugh - I've been thinking about the failing tests. 
It's obviously expected that those 2 will fail as now I don't pass the optional namespace. This actually made me wonder whether the app will continue to work as expected. While reading the WC docs I realized that one can update the session with additional chains and accounts. So I thought - ok, so if we switch from let's say OP to ETH, we could pass the new account to WC and it would pass it down to the dapap. To my surprise we already do that :D So the dapp recognises that we are switching the account in the wallet and updates respectively. Now with my fix this no longer works, because we have an isUnsupportedChain check:
https://github.com/safe-global/safe-wallet-web/blob/e25b770d406ff9fe5c48adfcec1a346d234df7c2/src/features/walletconnect/services/WalletConnectWallet.ts#L150

I've find another way to make this work. First we approve with requiredChain, then immediately after that we update the session with the Optional namespace. Then later our changes to the accounts will be properly propagated to all supported chains. 

The test still fails though, but that is I guess expected as we are not really testing whether approve has succeeded, but rather whether it was called with the "correct" arguments :(

compojoom:
ok, now it should be fine everywhere :(

usame-algan (approved):
Code looks good and we tested the happy path together (Connecting to OpenSea). I suggest we do regression tests for other WC cases before merging. Things like switching to a different Safe on the same network, switching to a different Safe on a different network etc.

katspaugh (approved):
👍

francovenica:
LGTM

Connecting with a different network than the one expected by OpenSea, it changes the network in openSea itself and connects fine.

Issue #3395, opened 2024-03-06
[WalletConnect] Remember session approvals


katspaugh:
Once you approve a connection to a dapp, remeber this choice **per chain**. Per chain because the same dapp might not work on a different chain.

PR #3441, opened 2024-03-14
fix: Add loading states to WC

### LINKED ISSUE #3200: [WalletConnect] Add loaders when connecting and signing
Users complain that it's slow, so let's add loading indicators and also measure the latency.

usame-algan:
## What it solves

Resolves #3200

## How this PR fixes it

- Adds loading indicators for `disconnect`, `reject` and `approve`
- Moves the `onDisconnect`, `onReject` and `onApprove` handler to the components where they are used instead of prop drilling them

Other changes:

💡 These aim to reduce complexity and improve the UX. We should discuss if we want to keep them or not. 💡 

- Removes the `handleBeforeUnload` call as there is no error and no such handling in the [example app](https://react-wallet.walletconnect.com/pairings). Reloading the page during an approval doesn't show any errors either, just an "Unknown" pairing but no active session. Using the same uri again establishes a session and updates the "Unknown" pairing which is expected.
- Removes `WcConnectionState` and associated logic. Switching back and forth with the dialogs when approving/rejecting takes the user out of the current context which is confusing imo. We should rather keep the user in the current dialog and display a loading state there.

## How to test it

1. Open the app, open the WC widget and paste a uri
2. Press Reject or Approve
3. Observe a loading indicator
4. Press Disconnect
5. Observe a loading indicator 

We should check that after pasting the URI into the input field and reloading the page before pressing reject and approve doesn't cause any errors.

## Checklist

### DIFF
--- a/src/features/walletconnect/WalletConnectContext.tsx
+++ b/src/features/walletconnect/WalletConnectContext.tsx
@@ -1,3 +1,4 @@
+import type { WCLoadingState } from '@/features/walletconnect/components/WalletConnectProvider'
 import { createContext, type Dispatch, type SetStateAction } from 'react'
 import type WalletConnectWallet from '@/features/walletconnect/services/WalletConnectWallet'
 
@@ -7,6 +8,8 @@ type WalletConnectContextType = {
   setError: Dispatch<SetStateAction<Error | null>>
   open: boolean
   setOpen: (open: boolean) => void
+  isLoading: WCLoadingState | undefined
+  setIsLoading: Dispatch<SetStateAction<WCLoadingState | undefined>>
 }
 
 export const WalletConnectContext = createContext<WalletConnectContextType>({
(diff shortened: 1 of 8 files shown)


katspaugh (approved):
👍

francovenica:
The only issue I found by playing with the page reload is that, if you click "Disconnect" and reload the page at the same time, then after the reload you still look connected in the safe app, while you got disconnected in the 3rd party app.
Still I think this is a very nish case, and you can still click Disconnect again with no errors
WC connection

Beyond this the rest worked fine for me connecting and disconnecting from apps.
Even tried executing tx in sepolia with cowswap with no issues from the safe side (cowswap is taking ages to execute the tx tho...)

usame-algan:
While working on #3395 I realised it would be nicer to only have one loading state and keep it inside the context. When a user approves a session, the reject button should be disabled and vice versa. If there was one loading state e.g. with an indicator that its an "approval" or a "rejection" or a "disconnection" we could disable all other action buttons. I will implement it in this PR since it would otherwise revert most of the changes done here. Sorry for the back and forth!

usame-algan:
@francovenica could you test again? The only difference should be that now when pressing "Approve", the "Reject" button should be disabled and when pressing "Reject", the "Approve" button should be disabled.

francovenica:
Yes, works fine for me.

PR #3446, opened 2024-03-15
fix: Remove counterfactual backup option

usame-algan:
## What it solves

Follow up on #3203 

Less than 4% of users who created a counterfactual safe did a backup according to our analytics. Furthermore we received feedback that this option is confusing. As there is currently no strategy to restore those backups this option seems more confusing than it is useful. In case users with an existing backup reach out to us we can still handle those manually.

## How to test it

2. Observe no Backup option

## Screenshots
Screenshot 2024-03-15 at 15 13 47

## Checklist

### DIFF
--- a/cypress/e2e/pages/create_wallet.pages.js
+++ b/cypress/e2e/pages/create_wallet.pages.js
@@ -26,7 +26,6 @@ const networkFeeSection = '[data-tetid="network-fee-section"]'
 const nextBtn = '[data-testid="next-btn"]'
 const backBtn = '[data-testid="back-btn"]'
 const cancelBtn = '[data-testid="cancel-btn"]'
-const safeBackupAlert = '[data-testid="safe-backup-alert"]'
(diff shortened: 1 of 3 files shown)


cypress/e2e/pages/create_wallet.pages.js
```diff
@@ -41,7 +40,7 @@ export const defaultSepoliaPlaceholder = 'Sepolia Safe'
 const welcomeToSafeStr = 'Welcome to Safe'
 
 export function verifyNewSafeDialogModal() {
-  main.verifyElementsIsVisible([safeBackupAlert, dialogConfirmBtn])
+  main.verifyElementsIsVisible([dialogConfirmBtn])
```
usame-algan, line 43: @mike10ca I have to adjust this line in order for the test to work again since the backup option for counterfactual safes is removed in this PR.

jmealy (approved):

liliya-soroka:
Verified

PR #3447, opened 2024-03-15
fix: Redirect to home if safe address is part of the URL path

usame-algan:
## What it solves

Follow up on #3356 

## How this PR fixes it

#3356 removed the redirect to last used safe but also the redirect in case there is a safe address in the URL. This adds it back as part of the 404 page so instead of redirecting to `/` there we redirect the user to their `/home` to bypass the new redirection logic of the index page.

## How to test it

1. Open `/sep:<safe address>`
2. You should be redirected to `/home?safe=sep:<safe address>`

## Checklist

### DIFF
--- a/src/pages/404.tsx
+++ b/src/pages/404.tsx
@@ -10,7 +10,7 @@ export const _getRedirectUrl = (location: Location): string | undefined => {
   const [, pathSafe] = pathname.match(re) || []
 
   if (pathSafe) {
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
👍

francovenica:
LGTM
The redirection works fine
I checked that other redirections have not being affected (They shouldn't for what I've seen in the code, but still, worth checking)

PR #3455, opened 2024-03-19
feat: Auto approve verified WC sessions

### LINKED ISSUE #3395: [WalletConnect] Remember session approvals

usame-algan:
## What it solves

Resolves #3395 

## How this PR fixes it

- Stores approvals for verified dApps per chain and auto approves future sessions

## How to test it

1. Open a Safe
2. Open a verified dApp (e.g. CowSwap)
3. Connect CowSwap to your Safe
4. A green message should be visible on the Approve screen
6. Press Approve
7. Disconnect the session and connect to CowSwap again
8. Observe no Approve screen and automatically connecting
9. Switch to a different chain and connect with CowSwap again
10. Observe the session is not auto approved
11. Oberve that the dialog closes on approval 

## Screenshots

https://github.com/safe-global/safe-wallet-web/assets/5880855/a302c57d-71d3-4af6-9895-9239014f43e8

## Checklist

### DIFF
--- a/src/features/walletconnect/components/WcProposalForm/index.tsx
+++ b/src/features/walletconnect/components/WcProposalForm/index.tsx
@@ -24,13 +24,14 @@ import css from './styles.module.css'
 type ProposalFormProps = {
   proposal: Web3WalletTypes.SessionProposal
   setProposal: Dispatch<SetStateAction<Web3WalletTypes.SessionProposal | undefined>>
+  onApprove: () => void
 }
 
-const WcProposalForm = ({ proposal, setProposal }: ProposalFormProps): ReactElement => {
+const WcProposalForm = ({ proposal, setProposal, onApprove }: ProposalFormProps): ReactElement => {
   const { walletConnect, setError, setIsLoading, isLoading } = useContext(WalletConnectContext)
 
   const { configs } = useChains()
-  const { safeLoaded, safe, safeAddress } = useSafeInfo()
+  const { safeLoaded, safe } = useSafeInfo()
   const { chainId } = safe
   const [understandsRisk, setUnderstandsRisk] = useState(false)
   const { proposer } = proposal.params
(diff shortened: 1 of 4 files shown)


katspaugh (approved):
👍

src/features/walletconnect/components/WcProposalForm/index.tsx
```diff
@@ -162,7 +141,7 @@ const WcProposalForm = ({ proposal, setProposal }: ProposalFormProps): ReactElem
           {isLoading === WCLoadingState.REJECT ? <CircularProgress size={20} /> : 'Reject'}
         </Button>
 
-        <Button variant="contained" onClick={onApprove} className={css.button} disabled={disabled}>
+        <Button variant="contained" onClick={() => onApprove()} className={css.button} disabled={disabled}>
```
katspaugh, line 144:
```suggestion
        <Button variant="contained" onClick={onApprove} className={css.button} disabled={disabled}>
```

usame-algan: `onApprove` takes an optional parameter and in this case we don't want to pass it but `onClick` passes an event object so we have to wrap the onApprove call.

francovenica:
Looks good. It works fine but I still have a question on this

Question:
I noticed that CowSwap has the green squre with the checkmark which I assume means is a trusted one.
Which apps are trusted and which ones are not? I tried with Uniswap and it seems to not be a trusted one, since it doesn't have that green checkbox, but Uniswap I think is a very well known app

Another question is, should we tell the user that "trusted apps will autoconnect after the first approve"? Should we let the user wonder why this app connects automatically and that other app does not. I don't think that the green checkbox conveys that message clearly

Comparison between Cowswap and Uniswap
image
image

usame-algan:
> I noticed that CowSwap has the green squre with the checkmark which I assume means is a trusted one.
Which apps are trusted and which ones are not? I tried with Uniswap and it seems to not be a trusted one, since it doesn't have that green checkbox, but Uniswap I think is a very well known app


> Another question is, should we tell the user that "trusted apps will autoconnect after the first approve"? Should we let the user wonder why this app connects automatically and that other app does not. I don't think that the green checkbox conveys that message clearly


I think a lot of legit dApps are not verified unfortunately. We discussed it this morning and decided to auto approve all dApps (verified and unverified). The only dApps we don't auto-approve are either scams or where there is a [domain mismatch](https://docs.walletconnect.com/web3wallet/verify#domain-risk-detection).

PR #3457, opened 2024-03-19
fix: Propose relay transactions without a signature

### LINKED ISSUE #3422: transaction duplication in case of relayer error
## Bug description
During the transaction "approve" sending, 2 "approve" transactions were created in the transaction queue (when 
I used a relayer).

## Environment
 - Browser: Chrome
 - Wallet: WalletConnect + Metamask
 - Chain: Sepolia

## Steps to reproduce
 - Create approve transaction
 - Select __Sponsored by Safe__
 - Try to execute

## Expected result
transaction execution
## Obtained result
2 signed transactions require execution, WalletConnect will return a GnosisSafe Tx Hash from the sendTransaction function for a transaction with a higher nonce.

https://github.com/safe-global/safe-wallet-web/assets/43392281/02dda1df-ba24-4d02-a0cf-5f06f80658ce

If you try to execute a transaction with a higher nonce, you will catch an error.

image

After several attempts to "execute", the transaction with the higher nonce disappears. (transaction with nonce 115 is in the video, but it's not in the screenshot.)

image

image

usame-algan:
## What it solves

Resolves #3422

Relaying a transaction first prompts the user to sign and then proposes the signed transaction to the backend which makes it appear in the queue with one signature but in case the call to the relay service fails, the tx dialog is still open and the user can either Execute the transaction again or only sign if the recommended nonce updated fast enough. This is confusing because the same transaction is already in the queue. Instead the transaction should not be queued unless the relay call succeeds.

## How this PR fixes it

Propose the unsigned transaction to the backend in case the user tries to relay. That way it doesn't show up in the queue unless executed and the recommended nonce also won't increase in case there is an error from the relay call.

## How to test it

1. Open a Safe on Sepolia
2. Create a transfer to the same Safe
3. Try to relay
4. Observe an error
5. Observe no transaction in the queue
6. Create a transfer to a different Account
7. Try to relay
8. Observe no error and the transaction showing up in the queue

## Screenshots
Screenshot 2024-03-19 at 16 32 51

## Checklist

### DIFF
--- a/src/components/tx/SignOrExecuteForm/hooks.test.ts
+++ b/src/components/tx/SignOrExecuteForm/hooks.test.ts
@@ -539,7 +539,9 @@ describe('SignOrExecute hooks', () => {
       await expect(executeTx({ gasPrice: 1 }, tx, '123', 'origin.com', true)).rejects.toThrowError(
         'Cannot relay an unsigned transaction from a smart contract wallet',
       )
-      expect(proposeSpy).not.toHaveBeenCalled()
+
+      // We are proposing before signing
+      expect(proposeSpy).toHaveBeenCalled()
       expect(signSpy).not.toHaveBeenCalled()
       expect(relaySpy).not.toHaveBeenCalled()
     })
--- a/src/components/tx/SignOrExecuteForm/hooks.ts
+++ b/src/components/tx/SignOrExecuteForm/hooks.ts
@@ -102,8 +102,8 @@ export const useTxActions = (): TxActions => {
       let tx: TransactionDetails | undefined
       // Relayed transactions must be fully signed, so request a final signature if needed
       if (isRelayed && safeTx.signatures.size < safe.threshold) {
-        safeTx = await signRelayedTx(safeTx)
         tx = await proposeTx(wallet.address, safeTx, txId, origin)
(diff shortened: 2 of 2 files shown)


katspaugh (approved):
Good idea, it makes them behave the same as regular immediately executed 1/X transactions.

On the other hand, it also brings all the problems of those txs, namely that the nonce isn't updated immediately as you pointed out. Can we maybe still sign but not propose a new tx when trying to relay again?

francovenica:
So, I was able to get the expected results as the description says:
Creating a send funds towards the safe itself, using relayer, seing the error, not getting any tx in queue after the error. So far so good.

Still it doesn't seem to be the same scenario of what the user initially reported.

I'll like to check with Usame how this fix addresses the reported issue

liliya-soroka:
Note about 115 tx. Most likely, it was a missing TX bug on the CGW side. The wallet infra team knows about the issue and debugging it

francovenica:
Checked with Usame. It showed me how, reproducin the issue the user has, it creates an untrusted tx now that doesn't show in the queue tab and keeps the same nonce over and over, until the user finally creates a proper tx that will replace the untrusted ones once executed

I think the solution is fine for the end user.

LGTM

PR #3458, opened 2024-03-20
chore: update ethers to v6.11.1

compojoom:
This is a long shot, but maybe it could fix this: https://github.com/5afe/safe-support/issues/408

In the stack trace there I saw 
EE error: unexpected tokens... coming from the useIsSafeTokenPaused hook.
It appears that ethers has some issues with some abis and those have been fixed in 6.11.0
I don't know why this error is not in read in the console, but looking at the ethers js it throws an exception there. I think this is when we do new Contract. And we are not catching this - so I could speculate that this causes the calls down the line to fail

## How to test it
@liliya-soroka  - maybe ask the user to test with the deployed version in this PR?

But even if it doesn't fix his issue we should still merge this. 

## Checklist


katspaugh:
I have a Ledger, will test on this branch.

compojoom:
Btw, this error is also visible when one tries to create a new tx:
grafik

katspaugh (approved):
I misplaced my Ledger and can't find it, so sending straight to QA. :)

schmanu (approved):

francovenica:
I wasn't able to see the issue the user reported in dev or prod, so there seems to be no easy way to reproduce it.

For this PR I just made sure I can sign tx with different wallets (ledger X, trezor one, MM, MM phone app through WC). I had no issues.

So I think is fine to merge this. I'll check it again during regression

PR #3473, opened 2024-03-25
Chore: update web3-onboard packages

katspaugh:
## What it solves

Updated onboard and walletconnect packages. The latest onboard core does something weird with type imports. Internally, it imports types from its own `common/types` dependency and doesn't like our `common/types` import.

@web3-onboard/core/node_modules/@web3-onboard/common

vs 

@web3-onboard/common/dist/types

So I had to cast some of WalletInit types.


src/hooks/wallets/wallets.ts
```diff
@@ -18,7 +17,10 @@ const prefersDarkMode = (): boolean => {
   return window?.matchMedia('(prefers-color-scheme: dark)')?.matches
 }
 
-const walletConnectV2 = (chain: ChainInfo): WalletInit => {
+type WalletInits = InitOptions['wallets']
+type WalletInit = WalletInits extends Array<infer U> ? U : never
```
compojoom, line 21: so WalletInit is not an Array of something, but just something?

katspaugh: This conditional expression means:
- if `WalletInits` is an array, `WalletInit` is an inferred type of the array's members

So it's a weird way to extract the type of an individual WalletInit from an array.

compojoom: ah, thanks!

PR #3475, opened 2024-03-25
Update eslint rules

compojoom:
## What it solves
Reduces nitpicking on syntax during reviews.

Ads rules for enforcing shortharnd object notation and removing curly brace when having a single prop e.g.
`something={"somtihng"}` becomes `sometihng="something"`

I also tried adding jsx-newline, but it conflicts with prettier and there doesn't seem to be a setting that can make those 2 co-exist. So I gues nitpicking on non-readable jsx syntax is fine there :).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,7 +14,8 @@
         "additionalHooks": "useAsync"
       }
     ],
-    "no-only-tests/no-only-tests": "error"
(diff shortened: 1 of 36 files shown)


katspaugh (approved):
LGTM! I've fixed the failing test on another branch.

usame-algan (approved):

PR #3474, opened 2024-03-25
Feat: show Social Login deprecation banner on all pages

katspaugh:
## What it solves

The Social Login warning will be now shown on all pages, not just the Dashboard.

PR #3471, opened 2024-03-25
fix: Overlapping buttons when signing a transaction

Eshwarj:
## What it solves
Overlapping buttons when signing a transaction #3403

Resolves #3403

## How this PR fixes it: 
As per the designs provided removed box component and used stack component to achieve the responsive design for different screen widths.  Added media queries for respective screen sizes to render the buttons as asked in the design.

## How to test it
Tested on various devices, screen sizes also by resizing the window.

## Screenshots

Screenshot 2024-03-25 at 1 54 14 PM
Screenshot 2024-03-25 at 2 00 17 PM
Screenshot 2024-03-25 at 1 54 30 PM

## Checklist
* [x] I've tested the branch on mobile 📱
* [ ] I've documented how it affects the analytics (if at all) 📊
* [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #3454, opened 2024-03-18
fix: Remove display version on unsupported contract

daniel0ar:
## What it solves

Resolves #3436

## How this PR fixes it
It adds an additional check on safe.version before displaying version tag (green mark and label).

## How to test it
Open https://localhost:3000/settings/setup?safe=matic:0x68d231A531BcF34199a825AE568aa79b695f7965 and check there is no "Latest Version" label

## Screenshots
image

PR #3453, opened 2024-03-18
fix: Hamburger-icon

puneetbagewadi:
## What it solves
Hamburger icon is not left padded

Resolves #3372
Padding on left of hamburger icon matches that above and below it.

## How this PR fixes it
1. Removing edge="start" which uses a negative margin to counteract the padding on one side (this is often helpful for aligning the left or right side of the icon with content above or below, without ruining the border size and shape).

2. Removing unnecessary padding-left provided to MUI IconButton and adding padding 16 px to the Button child element of IconButton.

## How to test it
As it is a ui/css change you can test the issue with the steps mentioned in reproducing it.

## Screenshots

Screenshot 2024-03-19 at 8 55 51 PM

Screenshot 2024-03-19 at 8 57 40 PM

Screenshot 2024-03-19 at 8 58 03 PM

## Checklist
* [x] I've tested the branch on mobile 📱
* [ ] I've documented how it affects the analytics (if at all) 📊
* [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #3477, opened 2024-03-26
Release 1.32.0

katspaugh:
## WalletConnect imrprovements
* feat: Auto approve verified WC sessions (#3455)
* fix: cannot connect with WC to Opensea (#3437)
* fix: Add loading states to WC (#3441)

## Features
* feat: make gas estimation error more prominent (#3404)
* Feat: delete transactions (#3336) – disabled for now
* Feat: show Social Login deprecation banner on all pages (#3474)

## Bug fixes
* fix: Overlapping buttons when signing a transaction (#3471) by @eshwarj
* fix: block new social login creations (#3426)
* fix: Revert warning.background color in dark mode (#3468)
* fix: allow subscribing to owned but not added safes (#3429)
* fix: Redirect user to accounts page on social login (#3460)
* fix: Propose relay transactions without a signature (#3457)
* Fix: always switch wallet chain before signing/executing (#3440)
* fix: Hamburger icon on mobile (#3453) by @puneetbagewadi
* Fix: color styles updates (#3410)
* fix: hydration error on dashboard (#3444)
* fix: Remove display version on unsupported contract (#3454) by @daniel0ar
* fix: Redirect to home if safe address is in url path (#3447)
* fix: Remove counterfactual backup option (#3446)
* fix: move scroll.io from blocked bridges to the warned bridges (#3445) by @ElvisKrop 
* fix: Don't allow duplicate signers in safe creation (#3427)
* fix: max button not applying correct value (#3408)
* fix: Use chain logos from config service (#3409)
* fix: Address overflow in read-only inputs (#3423)
* fix: Recovery not visible on settings page (#3414)
* #3452 
* #3462 
* #3484 

## Other improvements
* chore: update ethers to v6.11.1 (#3458)
* Chore: update web3-onboard packages (#3473)
* chore: `<p>` cannot appear as a descendant of `<p>` (#3417)
* tests: add send funds happy pathwith connected wallet (#3466)
* Test: add tx signing for NFTs (#3430)
* Test: detect duplicated signer in safe creation (#3439)
* Test: Add "sign action" to the replace signer e2e test (#3419)
* Tests: fix recover test (#3421)


mike10ca:
Regress automation tests passed. There is a failing test: [SMOKE] Verify that "Hide token" button is present and opens the "Hide tokens menu" in smoke tests. This is due to delay in response in CI. Locally this test passes just fine. This is not a blocker and is safe to be ignored. Will be fixed later.

francovenica:
Board:
https://www.notion.so/safe-global/63420452940c454b809f3b5872e678ef?v=7044dcf36cdb4656b7b00878ede3974a

francovenica:
The RC looks good from the QA side

The issue with the relayer and colors were fixed and the issue with the password of social login was a false positive, so it was dismissed

Automation worked
The list of tickets that can be tested were checked and they are properly deployed
The regression went well. Mostly signing with different wallets in a few different networks and few different type of tx, plus anything else automation doesn't cover.

Issue #3431, opened 2024-03-13
[Optimism] Check/update gas limit estimation


katspaugh:
Optmism is rolling out a new update that might affect our gas estimation function.

> The proposed Ecotone upgrade to the OP Stack and OP Mainnet changes the formula for estimating the [L1 Data Fee](https://docs.optimism.io/stack/transactions/fees#l1-data-fee) component of the [OP Stack Transaction Fee](https://docs.optimism.io/stack/transactions/fees) in preparation for the introduction of [EIP-4844](https://eip4844.com/) support.

> This new formula is designed to account for the option to post transaction data within [data blobs](https://docs.optimism.io/builders/chain-operators/management/blobs) instead of as Ethereum calldata, a feature enabled by [EIP-4844](https://eip4844.com/).

https://docs.optimism.io/builders/notices/ecotone-changes

PR #3490, opened 2024-03-27
feature: better gasLimit estimation for OP

compojoom:
## What it solves
According to the OP docs: https://docs.optimism.io/builders/app-developers/transactions/fees transaction fees on OP are composed of Execution Gas Fee and L1 Data fee. using provider.estimateGas just returns the execution gas fee, but not the L1 Data fee. Which means that in some cases our gas estimation could be too low. With this change we take the L1 Data fee into consideration when estimating transaction on OP.

Currently there is one open question - why aren't we currently aware of any OP transaction that has wrong gas, despite the fact that we never added L1 data fee. https://github.com/ethereum-optimism/developers/discussions/320 

Resolves #
#3431 

## How this PR fixes it
We now do both estimateGas and call the OP oracle for the gas needed to post on l1

## How to test it
Transactions on OP will now have a slightly higher gaslimit than they used to have

## Checklist

### DIFF
--- a/src/hooks/useGasLimit.ts
+++ b/src/hooks/useGasLimit.ts
@@ -18,6 +18,7 @@ import {
 } from '@safe-global/protocol-kit/dist/src/contracts/safeDeploymentContracts'
 import { type JsonRpcProvider } from 'ethers'
 import { type ExtendedSafeInfo } from '@/store/safeInfoSlice'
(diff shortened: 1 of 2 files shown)


compojoom:
Closing as according to my discussion with the OP developers:
https://github.com/ethereum-optimism/developers/discussions/320
this is not necessary

PR #3495, opened 2024-03-28
feat: [Counterfactual] Show pending status on dashboard

### LINKED ISSUE #3494: [Counterfactual] Display activation status on the dashboard
## What is the feature about

We want to make it more obvious to the user when a safe activation is under way and the status of it.

This is also the first task to enable a speed-up option.

## The list of requirements

- [ ] The dashboard widgets and title should change if a safe activation is in progress

## Designs/sketches
Screenshot 2024-03-28 at 12 09 51

usame-algan:
## What it solves

Resolves #3494

## How this PR fixes it

- Shows a different title and a loading spinner on the dashboard when activating a safe
- Shows a Pending status widget on the dashboard when activating a safe
- Shows a Useful Information widget on the dashboard when activating a safe

## Out of scope

Not included in this PR but coming later is an additional screen in case the activation is taking too long and an option to speed-up the transaction. See [Figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=10822%3A2654&mode=dev) design.

## How to test it

2. Go to the dashboard
3. Activate the safe
4. Observe a new loading state

## Screenshots
Screenshot 2024-03-28 at 12 05 00

## Checklist


katspaugh (approved):

francovenica:
LGTM.

Question regarding this:
> Not included in this PR but coming later is an additional screen in case the activation is taking too long and an option to speed-up the transaction. See [Figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=10822%3A2654&mode=dev) design.

Right now is not possible to speed up tx with no signatures, and I assume that the safe creation is the same as other tx where the signature is not there at the moment of execution. So, it will be possible to use speed up in this type of tx?

usame-algan:
> Right now is not possible to speed up tx with no signatures, and I assume that the safe creation is the same as other tx where the signature is not there at the moment of execution. So, it will be possible to use speed up in this type of tx?

Since we require the user to sign the safe creation tx I think speed-up will always be an option there.

Btw. I just noticed a bug with the widget not changing when relaying the transaction but it should be fixed now.

Issue #3482, opened 2024-03-27
Mobile viewport - Sidebar menu doesn't close when clicking on the items


francovenica:
## Bug description
While in a mobile view port, when you click the items in the sidebar, like assets, settings, etc, the menu stays open

## Environment
 - Browser: Chrome (Emulate the mobile viewport)
 - Wallet: Any
 - Chain: Any
 - Env: Prod 1.31

## Steps to reproduce
 1. Go to any safe
 2. Emulate the mobile viewport 
 3. Open the menu on the top left corner
 4. Click any section like settings

## Expected result
The sidebar collapses after taking the user to the selected section

## Obtained result
The sidebar remains after the redirection. The user must close it manually

## Screenshots
![image](https://github.com/safe-global/safe-wallet-web/assets/4503659/7479cbcb-bc43-4477-9a5f-1717ec35f531)

PR #3498, opened 2024-03-28
fix: Close sidebar drawer when user navigates on mobile

usame-algan:
## What it solves

Closes the sidebar on mobile when navigating to a different page

## How to test it

1. Open a Safe on mobile or with a mobile viewport
2. Navigate to a different page
3. Observe the sidebar closes

## Checklist


usame-algan:
Ideally, we would call `onToggle` on button click but it requires us to prop drill the function 3 times which is not as clean as having a single `useEffect` imo.

katspaugh (approved):
👍

katspaugh (approved):
✅

francovenica:
Nice, this solves this ticket #3482 . 

LGTM

PR #3500, opened 2024-03-29
Fix: Change disabled tx button tooltip text

### LINKED ISSUE #3381: [Spending Limit] Change tooltip for disabled buttons
## What is the feature about

For non-owners who have a spending limit, only specific actions in the app are enabled like `New transaction` from the sidebar or from the assets table. On a lot of other buttons they see a Tooltip saying `You can only create ERC-20 transactions within your spending limit`. This makes sense if those buttons lead to a transaction flow (e.g. Add owner) but with the addition of Recovery emails, there are now buttons that should only be enabled for owners but are not tied to transactions e.g. `Sign to view`. Therefore this tooltip can be confusing on those buttons.

## The list of requirements

The `CheckWallet` component needs to be adjusted.

The buttons that are available to spending limit users are already prominently displayed in the UI, telling them that they can only do that action is not necessary. Instead we should tell them why they cannot do the specific action of the button that they are trying to click on. So the tooltip should just say `Your connected wallet is not an owner of this Safe Account`.

e.g. if they try to add an owner, the tooltip would say `Your connected wallet is not an owner of this Safe Account` which still makes sense but now also makes more sense for buttons that are not tied to a transaction flow.

jmealy:
## What it solves

Resolves #3381

## How this PR fixes it
- Use the message `Your connected wallet is not an owner of this Safe Account` for limiting spending limit users as well as non owners.

## How to test it
- Check that the create tx button is enabled for spending limit users.
- Check that the correct tooltip message is shown on disabled buttons for spending limit users. Eg. adding an owner button.

## Checklist


katspaugh (commented):
When you're a beneficiary, will the New Transaction button still work?

jmealy:
> When you're a beneficiary, will the New Transaction button still work?

It was broken there actually but is working now after updating form dev

katspaugh (approved):

usame-algan (approved):

francovenica:
LGTM

PR #3505, opened 2024-04-02
Fix: extend bridge warnings list

### LINKED ISSUE #3499: Blacklist Core.app bridge
## What is the feature about
Preventing users from losing their funds on bridging.

## The list of requirements
Gotta add Core.app bridge to the list of dangerous bridges as they don't support receiving bridged tokens on a different address.

## Designs/sketches
Not needed.

## Links
https://www.notion.so/safe-global/Native-WalletConnect-Dangerous-bridge-research-22298ea428934846814e5b6599437dd8?pvs=4#4419d4f601a74833bfa0dd2544266a55

katspaugh:
## What it solves

Resolves #3499

## How this PR fixes it

Adds core.app to the list of warned bridges as it doesn't support specifying a different recepient address.

### DIFF
--- a/src/features/walletconnect/constants.ts
+++ b/src/features/walletconnect/constants.ts
@@ -64,6 +64,7 @@ export const BlockedBridges = [
 
 // Bridges that initially select the same address on the destination chain but allow changing it
 export const WarnedBridges = [
+  'core.app',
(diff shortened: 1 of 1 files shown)


katspaugh:
Tested ✅ 

Screenshot 2024-04-02 at 14 09 52

PR #3506, opened 2024-04-02
Fix: adjust bridge warning wording

katspaugh:
## What it solves

Make the warning more specifically refer to bridging as opposed to using a dapp.


katspaugh:
Approved by @kirkkonen

PR #3507, opened 2024-04-02
Chore: adjust timezone in tests

katspaugh:
## What it solves

TZ=CET stopped working and doesn't seem to affect the return value of `getTimezoneOffset()` anymore. It probably never worked, and it became apparent with the switch to the summer time last Sunday.


katspaugh:
@schmanu yeah, something weird is going on there, I haven't fully figured it out but got it fixed with the mockdata package.

PR #3510, opened 2024-04-03
Fix: don't show onboard popup on reconnect for MetaMask, Rabby and Zerion

### LINKED ISSUE #3508: Disable web3-onboard when re-connecting
When connecting a previously connected wallet, don't show the onboard popup.

`disableModals` is currently true when the wallet is unlocked, but AFAIK only MetaMask allows checking if it's unlocked, so the modal pops up for all other wallets.

```js
const connectLastWallet = async (onboard: OnboardAPI) => {
  const lastWalletLabel = lastWalletStorage.get()
  if (lastWalletLabel) {
    const isUnlocked = await isWalletUnlocked(lastWalletLabel)

    if (isUnlocked === true || isUnlocked === undefined) {
      connectWallet(onboard, {
        autoSelect: { label: lastWalletLabel, disableModals: isUnlocked || false },
      })
    }
  }
}
```

katspaugh:
## What it solves

Resolves #3508

## How this PR fixes it

Rabby and Zerion expose the same `window.ethereum._metamask` API as MetaMask itself, so it's possible to check if they are unlocked. If they are, we won't show a reconnect modal.

### DIFF
--- a/src/utils/wallets.ts
+++ b/src/utils/wallets.ts
@@ -48,10 +48,10 @@ export const isSmartContractWallet = memoize(
 
 /* Check if the wallet is unlocked. */
 export const isWalletUnlocked = async (walletName: string): Promise<boolean | undefined> => {
-  const METAMASK = 'MetaMask'
+  const METAMASK_LIKE = ['MetaMask', 'Rabby Wallet', 'Zerion']
 
   // Only MetaMask exposes a method to check if the wallet is unlocked
-  if (walletName === METAMASK) {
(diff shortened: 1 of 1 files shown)


compojoom (approved):
Looks good. I also tested it with Rabby and it works as expected

francovenica:
I was confused because I thought this was about phone apps, but it was about the browser extension. Daniel helped me out to check the fix and looks good. I don't see the reconection again when navigating/refreshing the page using Rabby

PR #3512, opened 2024-04-03
Fix: [Tx Deletion] signTypedData fallback for Ledger

katspaugh:
## What it solves

In Ethers v6, the signTypedData method unconditionally calls `eth_signTypedData_v4` that Ledger (and probably also Trezor) doesn't support. So I've added a fallback that calls `eth_signTypedData` if `eth_signTypedData_v4` fails.

## How to test it

Case 1:
* Propose a transaction with Ledger
* Delete that transaction with Ledger

Case 2:
* Sign a message with one of the signers being a Ledger

### DIFF
--- a/src/utils/__tests__/web3.test.ts
+++ b/src/utils/__tests__/web3.test.ts
@@ -44,20 +44,33 @@ describe('web3', () => {
     })
 
     it('should fall back to signTypedData if signTypedData_v4 is not available', async () => {
+      const error = new Error('error') as Error & { code: string }
+      error.code = 'UNSUPPORTED_OPERATION'
+
       const signer = {
-        signTypedData_v4: undefined,
-        signTypedData: jest.fn().mockResolvedValue(mockSignature),
+        signTypedData: jest.fn().mockRejectedValue(error),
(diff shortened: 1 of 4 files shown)


schmanu (approved):
👍 
Not sure if they are flaky but our AddressInput tests seem to fail. But there is no connection to the changes.

francovenica:
Cancelling a tx with ledger works fine

Signing a message has Issues:

Signing messages is failing with ledger for me.
I tried with a LedgerX in Ethereum, singing the tx that Opensea asks to sign when connecting to it.
image

francovenica:
Note:
We decided to test Trezor as well in this ticket. 
Trezor has issues for both, signing the Tx deletion and signing the Open sea message

Tx Deletion error:
image

For the openSea message, is the same error as the ledger

francovenica:
Correction. It seems my ledger was locked at the time of signing the message. 
I was able to sign just fine with the ledgerX
image

francovenica:
Signatures work again with both trezor one and ledgerX

francovenica:
Given the issue of the Trezor is going to be fixed in another ticket we can merge this one

PR #3518, opened 2024-04-05
fix: web3-onboard support for EIP6963

compojoom:
## What it solves
The previous version of web3-onboard we were using had a bug with EIP6963 and wallets supporting the standard were doing the wrong thing and overriding Metamask.

Resolves #

## How to test it
1. Next to metamask, install other wallets such as Rabby, coinbase, OKX etc
2. Web3-onboard should now display all those wallets as options to connect to
3. Clicking on metamask should open MM, clicking on Rabby should open Rabby etc

Previously the wallets were overriding metamask and it was disappearing from the list. 

## Screenshots
grafik

## Checklist


katspaugh (approved):
💪

PR #3526, opened 2024-04-08
chore: storybook

compojoom:
This commit ads Storybook support: https://storybook.js.org

Files ending with .stories.tsx will appear inside of the storybook UI where we get preview for those UI components.

## What it solves
Currently it is very hard to figure out which component can be reused when we build a new feature. We have a lot of designs in figma, but it is hard to find out if we have implemented reusable components that could be used to speed up the development of a new feature. With storyboard we get a place where we can display those reusable components. In addition to this storyboard forces us to think in "composable components" starting from the smallest UI elements and composing up. One first things about the props a component should receive, builds the UI around it with the different possible stats and later just feeds this component with the necessary data.

## How to test it
Run 
1. yarn
2. yarn storybook -> this should open localhost preview of components for which we have a story. 

This is still very rough - for example the CopyButton component doesn't render fully correctly:
grafik

but this will be tackled in a separate PR.

## Screenshots
grafik

grafik

## Checklist

### DIFF
--- a/src/components/common/CopyButton/index.tsx
+++ b/src/components/common/CopyButton/index.tsx
@@ -4,14 +4,7 @@ import CopyIcon from '@/public/images/common/copy.svg'
 import { IconButton, SvgIcon } from '@mui/material'
 import CopyTooltip from '../CopyTooltip'
 
-const CopyButton = ({
-  text,
-  className,
-  children,
-  initialToolTipText = 'Copy to clipboard',
-  onCopy,
-  dialogContent,
-}: {
+export interface ButtonProps {
   text: string
   className?: string
   children?: ReactNode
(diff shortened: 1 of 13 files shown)


src/components/common/CopyButton/index.tsx
```diff
@@ -4,22 +4,24 @@ import CopyIcon from '@/public/images/common/copy.svg'
...
 
-const CopyButton = ({
-  text,
-  className,
-  children,
-  initialToolTipText = 'Copy to clipboard',
-  onCopy,
-  dialogContent,
-}: {
+export interface ButtonProps {
```
usame-algan, line 7: Is this something we have to do for Storybook and does it make a different if we use type or interface here?

usame-algan (approved):
I suggest we disable telemetry: https://storybook.js.org/docs/configure/telemetry#how-to-opt-out

katspaugh (approved):
LGTM!
It would be good to also build the storybook on CI, so that we can host it next to the actual app in PR previews. In another PR, of course.

src/components/common/CopyButton/index.tsx
compojoom: No, I was plaing with this tutorial for enabling MUI in the preview:
https://storybook.js.org/recipes/@mui/material
and this is a left over. 

Basically with code like this:
```
import { Button as MuiButton, ButtonProps as MuiButtonProps } from '@mui/material';

export interface ButtonProps extends MuiButtonProps {
  label: string;
}
```

our Props would be also extended with all the available MUI buttonProps. Then you get 100s of MUI props next to your own component props and can filter with Pick or Omit.
```
// Only include variant, size, and color
type ButtonBaseProps = Pick<MuiButtonProps, 'variant' | 'size' | 'color'>;
```

PR #3532, opened 2024-04-09
Fix: onboard wallet icon (2)

katspaugh:
## What it solves

A follow-up on #3527 

## How this PR fixes it

Apparently some icons still come as a data-URI, so we need to support both raw SVG and URIs.

## How to test it
* Connect with MM
* Connect with WC


usame-algan (approved):
Looks good. I tested with MM, WC and Coinbase Wallet.

PR #3534, opened 2024-04-09
fix: status|speed up button too cramped up

compojoom:
## What it solves
grafik

Resolves #

## How this PR fixes it
grafik

## How to test it
Have a look at a pending transaction in the queue - both execute and speed up should have enough space from the elements above them.

## Checklist

### DIFF
--- a/src/components/transactions/TxSummary/styles.module.css
+++ b/src/components/transactions/TxSummary/styles.module.css
@@ -63,6 +63,6 @@
   }
 
   .status {
-    margin: 0  var(--space-1);
+    margin: 0 var(--space-1);
   }
 }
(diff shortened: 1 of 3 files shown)


src/components/transactions/TxSummary/styles.module.css
```diff
@@ -57,4 +61,8 @@
   .date {
     width: 100%;
   }
+
+  .status {
+    margin: 0  var(--space-1);
```
katspaugh, line 66:
```suggestion
    margin: 0 var(--space-1);
```

katspaugh (commented):
On mobile, it still looks a bit too close to the previous line.

compojoom:
define too close? 
grafik

compojoom:
yes, because you have the word ago in there, and the g would be the lowest point for the whole line.It's a visual illusion. As you can see from my screenshot that draws around the flex around the elements - it seems to be the same space between all elements

katspaugh (approved):
OK, doesn't matter much anyway

PR #3543, opened 2024-04-10
Feat: update voting power calculation to include locking contract tokens

jmealy:
## What it solves

Resolves https://www.notion.so/safe-global/Change-voting-power-computation-08023afef09e4676b7a9020d0d7ba7f3?pvs=4

## How this PR fixes it
Changes voting power calculation to include SAFE tokens that are in the locking contract. This includes locked tokens and unlocked but not yet withdrawn tokens.

## How to test it
- Open a safe with some SAFE tokens allocated.
- Check that the number shown in the header is correct according to the formula: `(allocation - claimed) + token balance + locking balance`

## Checklist


schmanu (approved):

liliya-soroka:
checking ...

liliya-soroka:
Verified.
We will need the PR to be released before TTE
cc @katspaugh

schmanu:
This can already be merged as the contract is already deployed and it should return 0 for all users at the moment.

PR #3544, opened 2024-04-10
Fix: use first queue nonce for queue label

katspaugh:
## What it solves

When deleting the oldest transaction in the queue, the label wasn't updated accordingly.

## How this PR fixes it

The queue label was taking the safe nonce but it should be actually taking the nonce of the oldest tx in the queue.

Screenshot 2024-04-10 at 13 20 21

NB: this doesn't fix that there's no Next label when deleting the top tx (it's a CGW issue).


schmanu (approved):

francovenica:
i don't know if I agree with this change. 
You have now this message showing for example:
image

Instead of this one: (this is current dev)
image

The 2nd one makes sense because it tells the user why he cannot execute that tx he has in the list.

If we want to implemente this change we should do something like show the "Next" block all the time with an empty state saying "the next tx to execute must be [Current_nonce]" or something like that

katspaugh:
As I mentioned in the description, the absence of a Next label is a backend bug and will be fixed on CGW (see https://github.com/safe-global/safe-client-gateway/issues/1387).

However, good point about the Execute button being disabled – we should fix that too. It currently also relies on the `safe.nonce` insted of the top queue nonce.

katspaugh:
The Execute button will now be active for the top tx regardless of its nonce.

francovenica:
I don't see what's the issue telling the user that, if he has a tx with a higher nonce than the one the safe requires to first create a tx with [CURRENT_NONCE], execute that and then he can execute whatever is in the queue.

I don't agree with the "Execute" button enabled for tx that are not the [CURRENT_NONCE]. It could create situations like this one:
image

image

katspaugh:
Not sure what you mean @francovenica. If you delete a tx with the current nonce (i.e. the top of the queue tx), the next top of the queue tx's nonce becomes the current nonce.

francovenica:
First. Lili filled me up in the issue about deleting tx in the queue, I didn't know what we were trying to fix.

By Current nonce, I mean the nonce that is in the settings, the next one the safe has (mandatory) to execute next in sequence.

I think the aproach should be to let know the user that the next tx to execute must be nonce "X" and if it's not in the queue (because it was deleted or for what ever other reason) he has to create it manually again. 

During tx creation we make the nonce to go +1 from the latest tx in queue for convenience (since most people won't go around skipping nonces), but we have little to say about what is the next nonce in the sequence (current nonce for that safe) that must be executed. The only way to know is to set a low nonce, see the tooltip saying "Nonce cannot be lower than  X" and adjust accordingly. Probably we should address that part

francovenica:
Here. Nothing in the tx creation tells you that the next nonce is 206 (in this case) until you set a low nonce on purpose.

nonce

katspaugh:
I'll close this PR, it looks like transactions with skipped nonces aren't executable.

PR #3554, opened 2024-04-12
Fix: allow deleting only last tx and duplicates

### LINKED ISSUE #3551: [Delete tx] Avoid nonce gaps in the queue
If you delete a tx from the middle of the queue, it results in a gap in the nonce seqence. Which makes it impossible to execute the transaction that comes after the deleted one, unless you create another tx with the skipped nonce (which is a bad UX).

Given this problem, we should limit what transactions can be deleted.

## Requirements
Only allow deletion if:
- There's another tx with the same nonce
- Or it's the highest nonce tx (i.e. the end of the queue)

---
A related bug report: https://github.com/safe-global/safe-wallet-web/issues/3549

katspaugh:
## What it solves

Resolves #3551

## How this PR fixes it

Only the last transaction in the queue and transactions with a duplicate nonce are allowed to be deleted.

I used `useRecommendedNonce() - 1` to calculate the last transaction nonce.

If a tx is in principle deletable but isn't the last one or a duplicate, we show a tooltip.

Screenshot 2024-04-12 at 14 30 37

### DIFF
--- a/src/components/tx-flow/flows/ReplaceTx/index.tsx
+++ b/src/components/tx-flow/flows/ReplaceTx/index.tsx
@@ -24,6 +24,7 @@ import { useHasFeature } from '@/hooks/useChains'
 import { FEATURES } from '@/utils/chains'
 import Track from '@/components/common/Track'
 import { REJECT_TX_EVENTS } from '@/services/analytics/events/reject-tx'
+import { useRecommendedNonce } from '@/components/tx/SignOrExecuteForm/hooks'
 
 const goToQueue = (router: NextRouter) => {
   if (router.pathname === AppRoutes.transactions.tx) {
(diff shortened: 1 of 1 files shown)


compojoom (approved):

francovenica:
If there are 2 tx with the same nonce, but one of them is a signed Rejection, then I think we shouldn't allow the tx deletion of the "good transaction", since the user would be left with only the rejection, and we know users still don't know that they can create a new tx with an old nonce, so chances are they will end up executing that deletion to get rid of it.

image

Beyond that the ticket is fine.

katspaugh:
> the user would be left with only the rejection, ..., so chances are they will end up executing that deletion to get rid of it.

That's the purpose of a rejection tx though, isn't it?

PR #3555, opened 2024-04-12
fix: [Counterfactual] Call onSubmit handler when submitting a transaction

usame-algan:
## What it solves

We didn't call the `onSubmit` handler in `CounterfactualForm` so safe apps didn't receive a `safeTxHash` when submitting transactions.

## How this PR fixes it

- Call `onSubmit` when submitting in `CounterfactualForm`

## How to test it

2. Send some WETH to it
3. Open the CowSwap safe app
4. Create a TWAP order
5. Submit the transaction in the tx-flow
6. Observe that CowSwap shows a success message and is not stuck

## Checklist

### DIFF
--- a/src/features/counterfactual/CounterfactualForm.tsx
+++ b/src/features/counterfactual/CounterfactualForm.tsx
@@ -66,7 +66,7 @@ export const CounterfactualForm = ({
   // On modal submit
   const handleSubmit = async (e: SyntheticEvent) => {
     e.preventDefault()
-    onSubmit?.('mockId')
+    onSubmit?.(Math.random().toString())
 
     if (needsRiskConfirmation && !isRiskConfirmed) {
       setIsRiskIgnored(true)
--- a/src/features/counterfactual/CounterfactualForm.tsx
+++ b/src/features/counterfactual/CounterfactualForm.tsx
@@ -38,6 +38,7 @@ export const CounterfactualForm = ({
   isOwner,
   isExecutionLoop,
   txSecurity,
+  onSubmit,
 }: SignOrExecuteProps & {
   isOwner: ReturnType<typeof useIsSafeOwner>
   isExecutionLoop: ReturnType<typeof useIsExecutionLoop>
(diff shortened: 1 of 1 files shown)


src/features/counterfactual/CounterfactualForm.tsx
```diff
@@ -65,6 +66,7 @@ export const CounterfactualForm = ({
   // On modal submit
   const handleSubmit = async (e: SyntheticEvent) => {
     e.preventDefault()
+    onSubmit?.('mockId')
```
usame-algan, line 69: This is not ideal but also doesn't break anything. We can also adjust the type but we are prop drilling this handler multiple times and will have to change it in all of these places and adjust related logic.

katspaugh: Can it be `Math.random()` instead?

usame-algan:
This fixes the issue mentioned here https://github.com/cowprotocol/cowswap/pull/4063#issuecomment-2039588336

katspaugh (approved):

francovenica:
Tried to swap WETH for USDC and I got a timeout on the tx and now the app is stuck here:
image

The cowSwap app still shows a tx is pending in the UI, but the address never got any order in cowswap at all:
https://explorer.cow.fi/sepolia/address/0xd134A0150f82C11f2B4368Ee0fA8133Fa450b5a6
image

Also the link to the tx in the sepolia block explorer is emtpy as well:
https://sepolia.etherscan.io/tx/0x22f82e47fd2f4564b5af08d81f4f98dd3481af78c3203996643c7bea3cf9a877

francovenica:
Given Usame's advice I tried again with Rabby and this time it worked. I was able to do the token approval for swapping + safe activation just fine and then tried the swap (the swap end up failing for low amount being swapped, so it was expected)

The safe used:
https://cf_onsubmit--walletweb.review-wallet-web.5afe.dev/transactions/history?safe=sep:0xfeAacD8DDFb0E9B26136971C244e782b2EB05675

So, the ticket is fine, but given that most people use MM they might find the issue I reported first.

usame-algan:
I will merge this so the CoW team is unblocked but we should follow up on this once Metamask works again.

PR #3556, opened 2024-04-12
feat: add a build of storybook to PRs

compojoom:
This PR ads a github action that builds our storybook together with the site and publishes it.

The URL to the storybook can be viewed in the comment.


katspaugh (approved):
🎉

PR #3558, opened 2024-04-13
fix: Hide "add to address book" when address already in the address book

### LINKED ISSUE #3531: Option to "add to address book" shown when the address already is in the addressbook
## Bug description
In the address input (of send funds or any other tx), if the address pasted is unknown an option to add it to the address book is shown. Currently the option is shown even if the address is already in the address book

## Environment
 - Browser: Chrome
 - Wallet: Any
 - Chain: Any
 - Env: Current prod 1.32

## Steps to reproduce
 1. Go to a safe you own
 2. Add an address to the address book
 3. Start a send funds tx
 4. Copy/paste that address in the input

## Expected result
No option to add to the address book is shown

## Obtained result
The option to add it to the AB is shown

## Screenshots
image

daniel0ar:
## What it solves

Resolves #3531 

## How this PR fixes it
It adds a memo function value (isInAddressBook) to find address in address book (returns true if found) and adds this check before displaying the message (if not isInAddressBook then display message).
## How to test it
1. Go to a safe you own
2. Add an address to the address book
3. Start a send funds tx
4. Copy/paste that address in the input
5. Check that "add to address book" is not showing anymore.

## Screenshots
image

### DIFF
--- a/src/components/common/AddressBookInput/index.tsx
+++ b/src/components/common/AddressBookInput/index.tsx
@@ -35,6 +35,11 @@ const AddressBookInput = ({ name, canAdd, ...props }: AddressInputProps & { canA
     [addressBookEntries, addressValue],
   )
 
+  const isInAddressBook = useMemo(
+    () => addressBookEntries.some((entry) => entry.label === addressValue),
+    [addressBookEntries, addressValue],
+  )
+
   const customFilterOptions = (options: any, state: any) => {
     // Don't show suggestions from the address book once a valid address has been entered.
     if (isValidAddress(addressValue)) return []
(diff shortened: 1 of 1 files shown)


src/components/common/AddressBookInput/index.tsx
```diff
@@ -35,6 +35,11 @@ const AddressBookInput = ({ name, canAdd, ...props }: AddressInputProps & { canA
     [addressBookEntries, addressValue],
   )
 
+  const isInAddressBook = useMemo(
```
katspaugh, line 38: Is this checking the eth address or the displayed name of an entry?

daniel0ar: Eth address previously assigned to entry.label

katspaugh: Thanks! A follow-up question: will it work it I enter a lowercase address but the address book contains a checksummed version of the same address? In some other places we use the sameAddress function for both variants to match.

daniel0ar: It is working (not sure why tho) but anyways I will use the sameAddress function for consistency :)

PR #3563, opened 2024-04-16
Fix: use wallet_addEthereumChain if cannot switch chain

### LINKED ISSUE #3515: MetaMask connection: unable to switch to the network, that isn't added in to current wallet
## Bug description
When user has selected network that isn't added to connected MetaMask wallet, clicking on "Switch to _Network name_" button  doesn't do anything.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Any

## Steps to reproduce
 1. Go to https://app.safe.global/welcome;
 2. Connect your MetaMask Wallet;
 3. In Network switcher dropdown select any Network, that isn't added to your Metamask;
 4. Click on your address in header;
 5. Click on "Switch to *Network name*".

## Expected result
MetaMask window appears, with request to "Allow this site to add a network?".

## Obtained result
Nothing happens on UI, MM throws next error in console:
`VM3088:1 MetaMask - RPC Error: Unrecognized chain ID "0xa4ec". Try adding the chain using wallet_addEthereumChain first. `
## Screenshots
image

katspaugh:
## What it solves

Resolves #3515

## How this PR fixes it

If `wallet_switchEthereumChain` throughs an "Unrecognized chain" error, we'll attempt to add that chain to the wallet.

### DIFF
--- a/src/services/tx/tx-sender/sdk.ts
+++ b/src/services/tx/tx-sender/sdk.ts
@@ -5,7 +5,7 @@ import type { JsonRpcSigner } from 'ethers'
 import { ethers } from 'ethers'
 import { isWalletRejection, isHardwareWallet, isWalletConnect } from '@/utils/wallets'
 import { OperationType, type SafeTransaction } from '@safe-global/safe-core-sdk-types'
-import type { SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
+import { getChainConfig, type SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import { SAFE_FEATURES } from '@safe-global/protocol-kit/dist/src/utils/safeVersions'
 import { hasSafeFeature } from '@/utils/safe-versions'
 import { createWeb3 } from '@/hooks/wallets/web3'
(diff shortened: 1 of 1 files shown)


src/services/tx/tx-sender/sdk.ts
```diff
@@ -26,6 +26,37 @@ export const getAndValidateSafeSDK = (): Safe => {
...
+  const UNKNOWN_CHAIN_ERROR_CODE = 4902
+  const hexChainId = toQuantity(parseInt(chainId))
+
+  try {
+    return await walletProvider.request({
+      method: 'wallet_switchEthereumChain',
+      params: [{ chainId: hexChainId }],
+    })
+  } catch (error) {
+    if ((error as Error & { code: number }).code !== UNKNOWN_CHAIN_ERROR_CODE) {
```
usame-algan, line 39: Is it possible to use `asError` here?

src/services/tx/tx-sender/sdk.ts
```diff
@@ -26,6 +26,37 @@ export const getAndValidateSafeSDK = (): Safe => {
...
+      throw error
+    }
+
+    const chain = await getChainConfig(chainId)
+
+    return await walletProvider.request({
```
usame-algan, line 45: I don't think `await` is needed here

src/services/tx/tx-sender/sdk.ts
```diff
@@ -26,6 +26,37 @@ export const getAndValidateSafeSDK = (): Safe => {
   return safeSDK
 }
 
+async function switchOrAddChain(walletProvider: ConnectedWallet['provider'], chainId: string): Promise<void> {
+  const UNKNOWN_CHAIN_ERROR_CODE = 4902
```
usame-algan, line 30: I assume this is Metamask specific. Do we need to handle this with other wallets as well?

usame-algan (approved):

src/services/tx/tx-sender/sdk.ts
katspaugh: Not much value here, we'd still have to cast it to have a code.

src/services/tx/tx-sender/sdk.ts
usame-algan: I guess with MM we already cover a lot of cases and not sure how WC would behave if we try to call `wallet_addEthereumChain`.

francovenica:
LGTM

Tested with Cello that was the network that gave me troubles on the first place
Deleted a few networks from the MM extension and add them again by switching to them

PR #3566, opened 2024-04-17
Tests: add core sdk integration to cypress tests

mike10ca:
## What it solves

Resolves #

## Checklist


compojoom (approved):
Looks good to me. Just one nitpick
```
const originatingSafe = constants.SEPOLIA_TEST_SAFE_44_SEND_FUNDS_HP14.substring(4)
```
It took me few seconds to realize why substring 4. The constant name is SEPOLIA_... so I was expecting to have an address without the chain prefix. So 4 looked like wrong amount - I was thinking you wanted to remove the 0x.
A better way would be to have a custom function
getAddressWithoutChainPrefix(SEPOLIA...)

this way it would be obvious why you did substring(4)

PR #3569, opened 2024-04-17
feat: add storybook for SrcEthHashInfo

compojoom:
## What it solves
add storybook for SrcEthHashInfo


usame-algan (approved):

katspaugh:
`SrcEthHashInfo` wasn't meant to be used as an end-component. `EthHashInfo` should be used instead. We should eventually remove `SrcEthHashInfo` as a public export.

PR #3577, opened 2024-04-18
Fix: link to the accounts page from the logo

katspaugh:
## What it solves

The logo has been historically linking to the Index or Welcome page which is now not so useful anymore.

It will now link to the Home page if on a different Safe page, or to the Accounts page (`/welcome/accounts`) if on Home already.

### DIFF
--- a/src/components/common/Header/index.tsx
+++ b/src/components/common/Header/index.tsx
@@ -33,8 +33,7 @@ const Header = ({ onMenuToggle, onBatchToggle }: HeaderProps): ReactElement => {
   const router = useRouter()
   const enableWc = useHasFeature(FEATURES.NATIVE_WALLETCONNECT)
 
-  // Logo link: if on Dashboard, link to Welcome, otherwise to the root (which redirects to either Dashboard or Welcome)
-  const logoHref = router.pathname === AppRoutes.home ? AppRoutes.welcome.index : AppRoutes.index
+  const logoHref = router.pathname === AppRoutes.welcome.accounts ? AppRoutes.welcome.index : AppRoutes.welcome.accounts
 
   const handleMenuToggle = () => {
     if (onMenuToggle) {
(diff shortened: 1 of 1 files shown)


src/components/common/Header/index.tsx
```diff
@@ -26,15 +27,23 @@ type HeaderProps = {
   onBatchToggle?: Dispatch<SetStateAction<boolean>>
 }
 
+function getLogoLink(router: ReturnType<typeof useRouter>): Url {
+  return router.pathname === AppRoutes.home || !router.query.safe
+    ? router.pathname === AppRoutes.welcome.accounts
+      ? AppRoutes.welcome.index
```
usame-algan, line 33: I don't understand why we are linking to the welcome page. The only thing the user can do there is go to the accounts page again so its an unnecessary click.

katspaugh: 1. That page is more focused on loging in, so if the person didn't find what they wanted on the accounts page, they probably want to "reset", i.e. go to the "beginning"
2. I think it's better than just staying on the same page (Accounts), alghough we could also make the logo unclickable in that case

usame-algan (approved):

francovenica:
I'm confused abut the term "home page". We got either "accounts page" or "welcome page". 
Currently if you are in the "accounts page" the logo redirects you to the "Welcome page" and if you are anywhere else you are redirected to the "Accounts page"

I think the behavior is fine, and I'll pass the ticket as it is, but I still wanted to know why in the description of the ticket we say "It will now link to the Home page if on a different Safe page"

katspaugh:
By Home page I meant the Dashboard. If you’re on e.g. the Transactions page and click on the logo, it will take you to the Dashboard of that Safe.

francovenica:
LGTM

PR #3583, opened 2024-04-19
fix: Disable Execute button for expired swaps

usame-algan:
## What it solves

Resolves [Notion issue](https://www.notion.so/safe-global/View-pending-expired-swap-in-Queue-00dd0e0e820f4040aab6dea6995f4d73)

## How this PR fixes it

- Disables the Execute button if a swap order is expired
- Shows a red Cancelled label for cancelled swaps

## How to test it

1. Create a cow swap and queue it
2. Wait until its expired
3. Observe the Execute button is disabled and a message is shown

## Screenshots

Screenshot 2024-04-19 at 10 07 17
Screenshot 2024-04-19 at 10 49 49

## Checklist

### DIFF
--- a/src/components/transactions/ExecuteTxButton/index.tsx
+++ b/src/components/transactions/ExecuteTxButton/index.tsx
@@ -4,7 +4,7 @@ import { type TransactionSummary } from '@safe-global/safe-gateway-typescript-sd
 import { Button, Tooltip } from '@mui/material'
 
 import useSafeInfo from '@/hooks/useSafeInfo'
-import { isMultisigExecutionInfo } from '@/utils/transaction-guards'
+import { isMultisigExecutionInfo, isSwapTxInfo } from '@/utils/transaction-guards'
 import Track from '@/components/common/Track'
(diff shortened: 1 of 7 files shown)


src/components/transactions/ExecuteTxButton/index.tsx
```diff
@@ -26,8 +26,10 @@ const ExecuteTxButton = ({
   const { setSelectedTxId } = useContext(ReplaceTxHoverContext)
   const safeSDK = useSafeSDK()
 
+  const isExpiredSwap = isSwapTxInfo(txSummary.txInfo) && txSummary.txInfo.status === 'expired'
```
compojoom, line 29: Maybe deserves a helper since you are doing the same in the TxDetails?

usame-algan: Good idea, I will extract them into helper methods.

usame-algan: done

PR #3590, opened 2024-04-19
fix: Set default owner on safe creation when entering step

### LINKED ISSUE #3571: Safe creation: wrong address is used as owner during safe creation
## Bug description
wrong address is used as owner during safe creation . It may lead to the safe creation with wrong owner set up
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: any

## Steps to reproduce
 1.Connect signer 1
2. click create a safe
3. change the signer 1 to the signer 2
Current result: Your account preview in the right side of the screen is updated with signer 2 instead of signer 1

4. go to the next step
Current result: The signer 1 is used as an owner of the future safe . At the same time Account preview shows signer2 as the safe set up
image

## Expected result
1. Wallet set up previous and owners set up should be the same
2.  if the connected signer was changed on the step 1 , the new connected signer should be used as default owner in the safe set up 
## Screenshots

Uploading Screen Recording 2024-04-17 at 12.37.19.mov…

usame-algan:
## What it solves

Resolves #3571

## How this PR fixes it

- Moves the defaultOwner logic from the parent component to the owner step so that it updates if the wallet was changed during one of the steps

## How to test it

1. Go to safe creation with wallet 1
2. Switch to wallet 2 in the first step
3. Go to the owner step
4. Observe wallet 2 is the default owner
5. Switch back to wallet 1
6. Observe wallet 2 is still the default owner
7. Go back one step and forth again
8. Observe wallet 2 is still the default owner

## Checklist

### DIFF
--- a/src/components/new-safe/create/index.tsx
+++ b/src/components/new-safe/create/index.tsx
@@ -9,7 +9,6 @@ import SetNameStep from '@/components/new-safe/create/steps/SetNameStep'
 import OwnerPolicyStep from '@/components/new-safe/create/steps/OwnerPolicyStep'
 import ReviewStep from '@/components/new-safe/create/steps/ReviewStep'
 import { CreateSafeStatus } from '@/components/new-safe/create/steps/StatusStep'
-import useAddressBook from '@/hooks/useAddressBook'
 import { CardStepper } from '@/components/new-safe/CardStepper'
 import { AppRoutes } from '@/config/routes'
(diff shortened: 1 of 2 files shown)


katspaugh (approved):

francovenica:
LGTM.

I've checked that, once a "first owner" is set in the owner step that one does't change at all when changing wallets connected.

PR #3596, opened 2024-04-22
fix: Set swap widget to full height

usame-algan:
## What it solves

Sets the swap widget page to full height so that the widget is not cut off

## How to test it

1. Open the Swap page
2. Click "My orders"
3. Observe the overlay taking up the available space

## Screenshots

Screenshot 2024-04-22 at 12 36 52

## Checklist

### DIFF
--- a/src/features/swap/index.tsx
+++ b/src/features/swap/index.tsx
@@ -16,6 +16,8 @@ import { useCustomAppCommunicator } from '@/hooks/safe-apps/useCustomAppCommunic
 import { showNotification } from '@/store/notificationsSlice'
 import { useAppDispatch } from '@/store'
 
+import css from './styles.module.css'
+
 const supportedChains = [1, 100, 11155111]
 
 const isSupportedChainForSwap = (chainId: number) => supportedChains.includes(chainId)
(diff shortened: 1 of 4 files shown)


src/features/swap/index.tsx
```diff
@@ -124,7 +126,7 @@ const SwapWidget = ({ sell }: Params) => {
       height: '860px',
       // provider: safeAppWeb3Provider, // Ethereum EIP-1193 provider. For a quick test, you can pass `window.ethereum`, but consider using something like https://web3modal.com
       chainId, // 1 (Mainnet), 5 (Goerli), 100 (Gnosis)
-      // standaloneMode: false,
+      standaloneMode: true,
```
compojoom, line 129: This should be false. If it is true the app displays a button to connect to the widget, we don't want this since we want the users to connect through our wallet connectors

compojoom: The default when not set is "false", but Anxo was suggesting to be explicit and have it set to "false"

compojoom, line 129:
```suggestion
      standaloneMode: false,
```

usame-algan: Ah true it was the opposite! Thanks, I just changed it. Now its also not possible anymore to "Disconnect" from the widget which is good.

compojoom (approved):

PR #3608, opened 2024-04-24
fix: do not show nonce conflict warning for batch transactions.

### LINKED ISSUE #3502: Batch tx show "conflicting nonce" warning during execution in bulk
## Bug description
In queue, while tx in bulk are executed, if any of the tx is a batch of tx, they will show the "nonce conflicting" warning message

## Environment
 - Browser: Chrome
 - Wallet: Any
 - Chain: Any (tested in sepolia)
 - Env: Current dev 1.32
 
## Steps to reproduce
 1. Go to safe 1/1, with the queue tx list empty
 2. Create a tx (like send funds) and queue it
 3. Create 2 more tx, but as add them as a batch instead of signing
 4. Confirm the batch of those 2 tx
 5. Check the Queue list that you have 2 tx, the 2nd being the batch of 2 tx created before
 6. Execute in bulk
 7. Check the queue list

## Expected result
Tx should not be "fighting for the nonce", since they are one transaction

## Obtained result
The "Conflicting nonce" warning shows

## Screenshots
image

jmealy:
## What it solves

Resolves #3502 

## How this PR fixes it
Removes the conflicting nonce header for pending transactions.

## How to test it
- Create a batch tx and sign
- Create any other tx and sign
- bulk execute the two Txs
- There should be no warning about conflicting nonces within the bulk tx, as they are all executed as one tx with the same nonce

## Screenshots
image

## Checklist


usame-algan:
I didn't understand how this can happen so I looked a bit more into it. From what I can see this only happens when using the Batch execute feature in a 1/n safe:

1. Whenever a user presses "Add to Batch", an unsigned transaction is proposed to the backend
2. When eventually confirming the batch, a multisend is created, signed and proposed to the backend
3. All of these transactions have the same nonce so there is a conflict header
4. When the user executes a transaction we fetch the **untrusted** queue which contains the previously mentioned transactions. We filter all of them out except the one(s) that are pending.

If the user were to execute this multisend on its own it would work because we have logic in place inside `usePendingTxs` to remove the conflict header item if only one transaction is pending. **Not sure if this is intended or a "feature".**

However if they queue another transaction and use the "Bulk execute" feature there is more than one pending transaction so the conflict header is not removed but we remove all transactions that are not pending so we end up with only one transaction that has a conflict header. Inside `groupConflictingTxs` we detect that and then push that transaction into an array which causes this display issue.

Imo the logic inside `TxList` and `groupConflictingTxs` is correct because we assume that if there is a conflict header that there are at least 2 transactions conflicting. Instead we should look into `usePendingTxs` and why we are removing the conflict header in some cases but not others and why we are only looking at the first two items there.

```js
if (results[1] && isConflictHeaderListItem(results[1])) {
  // Check if we both conflicting txs are still pending
  if (results.filter((item) => isTransactionListItem(item)).length <= 1) {
    results.splice(1, 1)
  }
}
```

jmealy:
> I didn't understand how this can happen so I looked a bit more into it. From what I can see this only happens when using the Batch execute feature in a 1/n safe:
> 
> 1. Whenever a user presses "Add to Batch", an unsigned transaction is proposed to the backend
> 2. When eventually confirming the batch, a multisend is created, signed and proposed to the backend
> 3. All of these transactions have the same nonce so there is a conflict header
> 4. When the user executes a transaction we fetch the **untrusted** queue which contains the previously mentioned transactions. We filter all of them out except the one(s) that are pending.
> 
> If the user were to execute this multisend on its own it would work because we have logic in place inside `usePendingTxs` to remove the conflict header item if only one transaction is pending. **Not sure if this is intended or a "feature".**
> 
> However if they queue another transaction and use the "Bulk execute" feature there is more than one pending transaction so the conflict header is not removed but we remove all transactions that are not pending so we end up with only one transaction that has a conflict header. Inside `groupConflictingTxs` we detect that and then push that transaction into an array which causes this display issue.
> 
> Imo the logic inside `TxList` and `groupConflictingTxs` is correct because we assume that if there is a conflict header that there are at least 2 transactions conflicting. Instead we should look into `usePendingTxs` and why we are removing the conflict header in some cases but not others and why we are only looking at the first two items there.
> 
> For example if I change this block it also resolves the issue (but might break other things)
> 
> ```js
> if (results[1] && isConflictHeaderListItem(results[1])) {
>   // Check if we both conflicting txs are still pending
>   if (results.filter((item) => isTransactionListItem(item)).length <= 1) {
>     results.splice(1, 1)
>   }
> }
> ```
> 
> to
> 
> ```js
> if (results[1] && isConflictHeaderListItem(results[1])) {
>     results.splice(1, 1)
> }
> ```

@usame-algan Thanks, I didn't see this header removal part you pointed out. After looking into it again I can't see any reason for why we are assuming the conflict header is always in the same position, so I'm assuming it's just an oversight. ie. it is not taking into account when there are single transactions preceding the batch, or when there are more than one batches. I made the conflict header removal section you pointed out more general to handle these cases where the header doesn't have index 1 in the array or where there is more than one of them.

This seems to work and I haven't found anything so far that it breaks, but it seems like it would be better if these cases were accounted for at the point where these headers are added (in the CGW I'm assuming).

usame-algan (approved):
Thanks! Looks good 👍

francovenica:
@jmealy The environment is showing a 404 when you access it

francovenica:
So I currently have a problem that won't allow me to fully test this ticket

Right now executing a tx in bulk is not exiting the form (I read about this in another ticket or a slack conversation).
Also the tx in queue are not showing the "pending" status that usually shows when tx are being executed, so the scenario where this was failing before is not met, therefore is not possible to test the fix.

I'm not saying that the fix is not correct or there is something wrong with it, just saying that, given how tx in bulk is working here, is not possible to test the fix

I'll keep this ticket here for the time

francovenica:
True, I tried with Rabby and I didn't have that issue so I was able to test the fix

LGMT. The "Conflicting nonce" message doesn't show anymore for tx with multiple actions in it.

PR #3615, opened 2024-04-25
Feat: block addresses on ofac list from swapping

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Block-OFAC-Addresses-a0a641560784420ea65fa72c1e036d74?pvs=4

## How this PR fixes it
- For addresses that are on the block list, do not show the cowswap widget. Show information on why they cannot swap instead.

## How to test it
- The component will only show for addresses in the block list. It could be tested by impersonating one of those addresses.
- Can work this out with QA when the time comes.

## Screenshots
image

## Checklist


compojoom:
codewise looks good to me 👍 I would just move it to components/common, as I can imagine that we will need it at other places as well. Also I would add a short description to the storybook, so it is clear what this component is for.

katspaugh:
`blockedAddressList.json` shouldn't be in components. I'd create a "service" for it:

```
src/
  services/
    ofac/
      blockedAddressList.json
      index.ts
```

In index.ts, export a function that checks if an address is in the list. Please use `sameAddress` for that, not `includes` as addresses can be both lowercased and mixed-case (checksummed).

katspaugh (approved):
👍

usame-algan (approved):

usame-algan:
Is it possible to add the source where we got the addresses from to the .json file as a comment?

compojoom:
@usame-algan https://sentry.io/answers/comments-in-json/

The Problem

> Can comments be used in JSON?
> The Solution
> 
> The short answer is no. The [JSON specification](https://www.json.org/json-en.html) does not include comments, and they cannot be used in JSON files (i.e. files with a .json extension). Comments were removed from the JSON specification to prevent developers from using them to specify parser directives, which complicated the data format.

We can do it in the .ts file

Since I had to rebase the feat-cow-swap branch this branch now has conflicts. I've locally squashed James commits and pushed them directly to the feat-cow-swap branch avoiding unnecessary conflict resolution now.

PR #3616, opened 2024-04-26
Refactor: walletconnect card on dashboard

katspaugh:
## What it solves

We no longer have to keep the old WalletConnect Safe App visible, so I replaced the dashboard widget with a hardcoded WalletConnect card.

Screenshot 2024-04-26 at 14 45 55


src/components/dashboard/FeaturedApps/FeaturedApps.tsx
```diff
@@ -53,17 +45,28 @@ export const FeaturedApps = ({ stackedLayout }: { stackedLayout: boolean }): Rea
...
+                    description={txBuilder.app.description}
+                    iconUrl={txBuilder.app.iconUrl}
+                  />
                 </NextLink>
               </Grid>
-            ))}
+            )}
+            {enableWc && (
+              <Grid item xs={12} md={stackedLayout ? 12 : 6}>
+                <a onClick={onWcWidgetClick}>
```
usame-algan, line 61: Without href the cursor is not a pointer even though the whole card is clickable. I suggest we add some hover effect. We could also do `href="#"` and `e.preventDefault()` onClick.

src/components/dashboard/FeaturedApps/FeaturedApps.tsx
```diff
@@ -39,12 +34,9 @@ const onWcWidgetClick = (e: SyntheticEvent) => {
 }
 
 export const FeaturedApps = ({ stackedLayout }: { stackedLayout: boolean }): ReactElement | null => {
-  const router = useRouter()
-  const [featuredApps, _, remoteSafeAppsLoading] = useRemoteSafeApps(SafeAppsTag.DASHBOARD_FEATURED)
+  const txBuilder = useTxBuilderApp()
   const enableWc = useHasFeature(FEATURES.NATIVE_WALLETCONNECT)
```
usame-algan, line 38: nit: Unrelated but this variable name is confusing like its a user action to enable WC. Something like `isWcEnabled` would be more descriptive

usame-algan (approved):
I have one suggestion but looks good otherwise!

liliya-soroka:
Verified.
Don't see any problem with hardcoded card even for a new network , as WC should be enabled for any chain

PR #3626, opened 2024-04-30
feat: hide finish button on success screen for swaps

compojoom:
## What it solves
This is an attempt to make the swap a bit less confusing after performing a swap. We incentivise the user to go to "view transaction" instead of staying on the swap page. This way he won't be on the swap page that just displays "order submitted" with no further information about the order itself. 

Resolves
https://www.notion.so/safe-global/Submit-order-redirect-5e826c01903249fdb3cfc116811fa472?pvs=4

## How this PR fixes it
Hides the "Finish" button on the Success screen

## How to test it
Perform a swap. On the success screen you should only see "view transaction"

## Screenshots
grafik

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/SuccessScreen/index.tsx
+++ b/src/components/tx-flow/flows/SuccessScreen/index.tsx
@@ -29,7 +29,7 @@ const SuccessScreen = ({ txId, safeTx }: { txId: string; safeTx?: SafeTransactio
   const { status } = pendingTx || {}
   const txHash = pendingTx && 'txHash' in pendingTx ? pendingTx.txHash : undefined
   const txLink = chain && getTxLink(txId, chain, safeAddress)
-  const [decodedData, decodedDataError, decodedDataLoading] = useDecodeTx(safeTx)
+  const [decodedData] = useDecodeTx(safeTx)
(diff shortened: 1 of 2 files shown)


katspaugh:
Should we add a secondary button with something like "Back to swaps"?

katspaugh:
No, it's good that the main CTA at the bottom is View transaction.

compojoom:
Ok, added

src/components/tx-flow/flows/SuccessScreen/index.tsx
```diff
@@ -26,6 +29,8 @@ const SuccessScreen = ({ txId }: { txId: string }) => {
   const { status } = pendingTx || {}
   const txHash = pendingTx && 'txHash' in pendingTx ? pendingTx.txHash : undefined
   const txLink = chain && getTxLink(txId, chain, safeAddress)
+  const [decodedData, decodedDataError, decodedDataLoading] = useDecodeTx(safeTx)
```
usame-algan, line 32:
```suggestion
  const [decodedData] = useDecodeTx(safeTx)
```

usame-algan (approved):

PR #3627, opened 2024-04-30
Feat: Add token symbols to swap transaction summary

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Display-token-icons-in-the-tx-list-12098c01ba6743b7a3669e6262c9800c?pvs=4

## How this PR fixes it
- display the sell token symbol and then the buy token symbol in the transaction info 

## How to test it
- Excecute a swap
- View the swap tx in transaction history.

## Screenshots
image

## Checklist

### DIFF
--- a/src/components/common/TokenIcon/index.tsx
+++ b/src/components/common/TokenIcon/index.tsx
@@ -1,6 +1,7 @@
 import { type ReactElement } from 'react'
 import ImageFallback from '../ImageFallback'
 import css from './styles.module.css'
+import classNames from 'classnames'
 
 const FALLBACK_ICON = '/images/common/token-placeholder.svg'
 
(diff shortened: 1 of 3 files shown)


katspaugh:
I this per the design? I feel like they should overlap a little more.

Screenshot 2024-04-30 at 10 51 46

jmealy:
> I this per the design? I feel like they should overlap a little more. Also there's not enough margin between the icons and the text on the right.
> 
> Screenshot 2024-04-30 at 10 51 46

It's an overlap of 4px in the design, but here it is with 6px, and the increased margin. It does look closer to the design.

image

katspaugh:
We should probably add a negative offset to the swap pairs, so that the text to the right is always aligned to the grid.

jmealy:
> We should probably add a negative offset to the swap pairs, so that the text to the right is always aligned to the grid.

This is what it looks like with the text aligned:

image

katspaugh:
Also not ideal but definitely better 👍
I would make the icons overlap even more, like 40%. But yeah let’s wait for a designer’s decision. 🙂

katspaugh:
IMG_4354

We should probably make all token icons circular in dark mode, otherwise some of them are circles and some rounded rectangles.

TanyaEfremova:
The latest changes look good! Approved :)

PR #3630, opened 2024-04-30
fix: limit batch rpc max count

schmanu:
## What it solves
When safe apps request many RPC calls, the RPC batch limit of e.g. Infura gets reached on mainnet.

## How this PR fixes it
- Limits batch requests to max 10 requests

## How to test it
- Open a Safe app that does many RPC requests (e.g. Cow Swap)

## Checklist


katspaugh (approved):
Good find!
Does it mean that there will be more batches sent?

schmanu:
>

> Good find! Does it mean that there will be more batches sent?

Yes exactly. With some of my Safes I cannot use CoW Swap anymore as it can not load the balances due to too big batches and Infuras restrictions. This fixes the app for those Safes for me.

Ethers will manage the batch size internally. The BrowserProvider for the connected wallet uses a limit of 1 btw disabling batch requests.

PR #3634, opened 2024-05-01
Feat: persist last used order type in cowswap widget

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/limit-etc-Remember-last-used-trade-type-6cd4087ac75947b2972e726653e32eb4?pvs=4

## How this PR fixes it
- Remember the last used trade type (swap or limit)
- persist across page visits.

## How to test it
- Change the trade type in the cow swap widget. (default is swap)
- Navigate away from the page and back again, or refresh the page. The trade type should stay as it was set before.

## Checklist


compojoom:
The issue was mainly for "swap|limit" change and not about what sell|buy token the user has selected. 
I think that if you want to remember the sell|buy token we would have some more work to do, to actually make this useful. For example - I would reset the the sell and buy token if a user makes a successful swap. If I play around with tokens, then maybe it makes sense to remember them until I've performed the swap. 
The feature implemented like this also makes it impossible to set the source token from the assets page.

jmealy:
> The issue was mainly for "swap|limit" change and not about what sell|buy token the user has selected. I think that if you want to remember the sell|buy token we would have some more work to do, to actually make this useful. For example - I would reset the the sell and buy token if a user makes a successful swap. If I play around with tokens, then maybe it makes sense to remember them until I've performed the swap. The feature implemented like this also makes it impossible to set the source token from the assets page.

Gotcha, changing swap|limit makes more sense. I'll get rid of keeping the tokens then for now

compojoom (approved):

compojoom (changes requested):
Can you check why the unit test fails

compojoom (approved):

PR #3643, opened 2024-05-02
fix: Hide slippage for swap limit orders

usame-algan:
## What it solves

Resolves [Notion issue](https://www.notion.so/safe-global/Slippage-is-displayed-for-the-Limit-Orders-on-Confirmation-page-during-tx-creation-a2d2a8a3bdb74964950b3611b85ae498)

## How this PR fixes it

- Hides the Slippage if the `orderClass` is `limit`

## How to test it

1. Create a limit order
2. Submit the order in the widget
3. Observe no Slippage in the order details

## Screenshots
Screenshot 2024-05-02 at 15 50 04

## Checklist

### DIFF
--- a/src/features/swap/components/SwapOrderConfirmationView/index.tsx
+++ b/src/features/swap/components/SwapOrderConfirmationView/index.tsx
@@ -7,7 +7,13 @@ import { compareAsc } from 'date-fns'
 import { Alert, Typography } from '@mui/material'
 import { formatAmount } from '@/utils/formatNumber'
 import { formatVisualAmount } from '@/utils/formatters'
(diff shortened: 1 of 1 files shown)


compojoom (changes requested):
The orderClass from the order is actually always "limit". Every order in cowswap is basically a limit order. To have the Ui's "swap|limit" option you need to read the value from the parameters. In utils.ts we have a function to calculate them.

usame-algan:
> The orderClass from the order is actually always "limit". Every order in cowswap is basically a limit order. To have the Ui's "swap|limit" option you need to read the value from the parameters. In utils.ts we have a function to calculate them.

Thanks! I switched to using the helper function `getOrderClass` 👍

compojoom (approved):
I'm starting to think that we need a storybook for limit order for orderConfirmationView

PR #3646, opened 2024-05-02
Analytics: swap analytics

schmanu:
## How this PR fixes it
- Adds event for opening safe apps modals with app name as label
- Adds token address as label to event for the Swap button in the assets table
- adds native_swap as tx type for swap transactions

## How to test it
- Create a Swap through our modal and observe the new /adjusted events

## Checklist


src/hooks/safe-apps/useCustomAppCommunicator.tsx
```diff
@@ -72,6 +72,7 @@ export const useCustomAppCommunicator = (
       }
 
       setCurrentRequestId(requestId)
+      trackEvent({ ...SAFE_APPS_EVENTS.OPEN_TRANSACTION_MODAL, label: appData.name })
```
compojoom, line 75: 👍

compojoom (commented):
so, with this change we can track when people open the app and when they submit a tx through it? But what about the order statuses?

compojoom (approved):
let's mege as it is and maybe later update with further events for the order status if necessary

PR #3648, opened 2024-05-03
Fix: cow swap disclaimer communication

schmanu:
## What it solves
- the app communicator was not correctly initialized when the disclaimer is displayed

Resolves #

## How this PR fixes it
- Adds the consent flag to the dependencies of the useEffect that sets the iframeRef.

## How to test it
- Delete the disclaimer consent from local storage
- Open the CoW widget
- Witness that the connection succeeds

## Checklist

### DIFF
--- a/src/features/swap/index.tsx
+++ b/src/features/swap/index.tsx
@@ -194,7 +194,7 @@ const SwapWidget = ({ sell }: Params) => {
     if (iframeElement) {
       iframeRef.current = iframeElement as HTMLIFrameElement
     }
-  }, [params])
+  }, [params, isConsentAccepted])
 
   useCustomAppCommunicator(iframeRef, appData, chain)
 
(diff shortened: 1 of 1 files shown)


src/features/swap/index.tsx
```diff
@@ -194,7 +194,7 @@ const SwapWidget = ({ sell }: Params) => {
     if (iframeElement) {
       iframeRef.current = iframeElement as HTMLIFrameElement
     }
-  }, [params])
+  }, [params, isConsentAccepted])
```
usame-algan, line 197: This makes it work for me when there is a disclaimer but not when reloading the swap page. Adding `safeLoading` here as well seems to make it work though.

usame-algan (changes requested):

usame-algan (approved):
Thanks!

PR #3649, opened 2024-05-03
feat: Use different titles depending on the type of swap

usame-algan:
## What it solves

Resolves [Notion issue](https://www.notion.so/safe-global/Issue-No-icon-and-wrong-name-on-the-confirmation-screen-7d49cf4e40794dd4b9a628e192c24f43)

Note: This doesn't add titles for specific actions like "Wrap & Swap" and "Approve & Swap"

## How this PR fixes it

- Shows either `Swap order` or `Limit order` as the title when creating a swap tx via the swap widget
- Adjusts the `AppTitle` component to be consistent with the styles of other transaction titles and icons

## How to test it

1. Create a limit and a swap order
2. Observe the correct title showing
3. Queue them
4. Go to the queue and press sign/execute
5. Observe the correct title showing

## Checklist


src/features/swap/index.tsx
```diff
@@ -61,6 +50,22 @@ const SwapWidget = ({ sell }: Params) => {
   const wallet = useWallet()
   const { isConsentAccepted, onAccept } = useSwapConsent()
 
+  const appData: SafeAppData = useMemo(
+    () => ({
+      id: 1,
+      url: 'https://app.safe.global',
+      name: `${capitalize(tradeType.toLowerCase())} order`,
```
compojoom, line 57: No, I don't think this is a good idea as the name and url is submitted as origin to cgw:
grafik

also the label for GTM is also dependent on it. 

@schmanu what do you think about this?

usame-algan: We could also keep the name the same here and add an exception to the `AppTitle` component.

usame-algan: I kept the `Safe Swap` name in the config and handle it in `AppTitle` now

src/features/swap/index.tsx
```diff
@@ -61,6 +55,22 @@ const SwapWidget = ({ sell }: Params) => {
...
+      name: SWAP_TITLE,
+      iconUrl: darkMode ? './images/common/safe-swap-dark.svg' : './images/common/safe-swap.svg',
+      description: 'Safe Apps',
+      chainIds: ['1', '100'],
+      accessControl: { type: SafeAppAccessPolicyTypes.NoRestrictions },
+      tags: ['safe-apps'],
+      features: [SafeAppFeatures.BATCHED_TRANSACTIONS],
+      socialProfiles: [],
+    }),
+    [darkMode, tradeType],
```
compojoom, line 71:
```suggestion
    [darkMode],
```

the eslint warning seems correct here.

src/features/swap/index.tsx
```diff
@@ -127,7 +137,7 @@ const SwapWidget = ({ sell }: Params) => {
         event: CowEvents.ON_CHANGE_TRADE_PARAMS,
         handler: (newTradeParams) => {
           const { orderType: tradeType } = newTradeParams
-          dispatch(setSwapParams({ tradeType }))
+          dispatch(setSwapParams({ tradeType: tradeType.toLowerCase() }))
```
compojoom, line 140: how about we automatically do this conversion in the reducer and not here. I'm sure that if we have to dispatch this somewhere else we'll forget to lowercase it.

compojoom (approved):

PR #3651, opened 2024-05-03
Feat cow swap better notifications

compojoom:
## What it solves
Original issue: we had a listener for the different order statuses on the swap widget. That meant that if the user would leave the widget page he wouldn’t receive any notifications about the order. What is even worse - the user would be on the transaction page, would see the order updating and later when navigating to the widget he would now see a notification for that order. This was extremely confusing.

Resolves #
https://www.notion.so/safe-global/07996126a3c041db90e36e01eb4ce46c?v=03b2eaffc9af4b48b1bf57e50546ac3d&p=e3f19e8ddcaf4263ab4164a4eb95e333&pm=s

## How this PR fixes it
I’ve changed the way we deal with events. Instead of directly dispatching a notification whenever we receive a new event from the widget, we now store those order updates in the redux state. We update those orders based on our txHistory and on the widget updates we receive. The middleware that listens on those updates has access to the state prior to the state update and can decide if a notification needs to be shown to the user or not.

**Known issue: notifications for expired or cancelled orders can first be diplayed when the user refreshes the page or receives/executes a tx. We rely on the tx history to get the order state. The tx history won't be polled if the cgw doesn't see a new tx.**

## How to test it
Create an order and leave the swap page. You should still get updates for the order. 

## Checklist


src/store/swapOrderSlice.ts
```diff
@@ -0,0 +1,197 @@
...
+
+          listenerApi.dispatch({
+            type: slice.actions.setSwapOrder.type,
+            payload: {
+              orderUid: swapOrder.uid,
+              status: swapOrder.status,
+              txId: result.transaction.id,
+            },
+          })
+        }
```
usame-algan, line 193: The `swapOrderListener` is very similar to the `txHistoryListener`. Anything against moving this part into the `txHistoryListener` instead of defining another one on the same slice?

src/store/swapOrderSlice.ts
```diff
@@ -0,0 +1,197 @@
...
+      switch (newStatus) {
+        case 'created':
+          const safeInfo = selectSafeInfo(listenerApi.getState())
+
+          dispatch(
+            showNotification({
+              title: 'Order created',
+              message:
+                safeInfo.data?.threshold === 1
+                  ? ' Waiting for the transaction to be executed'
```
usame-algan, line 88:
```suggestion
                  ? 'Waiting for the transaction to be executed'
```

src/features/swap/index.tsx
```diff
@@ -61,62 +61,52 @@ const SwapWidget = ({ sell }: Params) => {
...
             case 'ORDER_CREATED':
               dispatch(
-                showNotification({
-                  title: 'Swap transaction created',
-                  message: 'Waiting for confirmation from signers of your Safe',
-                  groupKey,
-                  variant: 'info',
+                setSwapOrder({
+                  orderUid: event.data.orderUid,
+                  status: 'created',
```
usame-algan, line 77: I don't think we need a switch case here since we always call the same function. Instead we could pass the `messageType`:

```
dispatch(
    setSwapOrder({
        orderUid: event.data.orderUid,
        status: messageType,
    })
)
```

Or if we want to have a different status notation we could add a mapper to go from `ORDER_CREATED -> created` etc.

src/store/swapOrderSlice.ts
```diff
@@ -0,0 +1,197 @@
...
+import type { OrderStatuses } from '@safe-global/safe-gateway-typescript-sdk'
+import type { RootState } from '@/store'
+import { isSwapTxInfo, isTransactionListItem } from '@/utils/transaction-guards'
+import { txHistorySlice } from '@/store/txHistorySlice'
+import { showNotification } from '@/store/notificationsSlice'
+import { selectSafeInfo } from '@/store/safeInfoSlice'
+
+type AllStatuses = OrderStatuses | 'created'
+type Order = {
+  orderUid: string
```
usame-algan, line 12: With this structure we have `orderUid` twice. Is it needed or can we remove the object attribute?

src/store/swapOrderSlice.ts
```diff
@@ -0,0 +1,197 @@
...
+    actionCreator: slice.actions.setSwapOrder,
+    effect: (action, listenerApi) => {
+      const { dispatch } = listenerApi
+      const swapOrder = action.payload
+      const oldStatus = selectSwapOrderStatus(listenerApi.getOriginalState(), swapOrder.orderUid)
+      const newStatus = swapOrder.status
+
+      if (oldStatus === newStatus) {
+        return
+      }
```
usame-algan, line 77: How can this case happen like calling `setSwapOrder` twice with the same status?

src/store/swapOrderSlice.ts
compojoom: I wouldn't do this. I would prefer to keep them separated. I know that they are similar, but still distinct and I would like to avoid touching "non-swap" parts. This would also complicate the original middleware. 

I think that also this separation is ok, as you would have everything related to the swap at one place and won't be surprised to find logic related to it in an unrelated place.

src/features/swap/index.tsx
compojoom: Not that easy. You would at least need a mapping of the cow eventListener type to the allStatusesType (which is the cgw status type + few more)

src/store/swapOrderSlice.ts
compojoom: It's twice, but get used to that. It's a normal pattern in redux and it simplifies working with the object. I know you have it in the key, but then you would always have to do Object.keys() to get it... It's very annoying. The duplication here is fine.

src/store/swapOrderSlice.ts
compojoom: Easy. The state is written to from the txHistory and from the widget. The widget doesn't know anything about our state, it has it's own state. And if you leave the widget when the status is open and return with a fullfilled state, then the widget will notify us and we'll try to update the order status. But if the status was updated by txHistory before that... it won't be necessary to show any notification

src/store/swapOrderSlice.ts
usame-algan: Agreed, after thinking about it it makes sense to have the slice as the source of truth and separate listeners by feature instead of having one listener that handles multiple things. Its much easier to look up that swap orders are affected by the history than having to know the context that the history is affecting swap orders.

src/store/swapOrderSlice.ts
usame-algan: Agree that its more convenient and we are not optimizing for space so lets follow this pattern in the future.

src/store/swapOrderSlice.ts
```diff
@@ -0,0 +1,206 @@
...
+        case 'fulfilled':
+          if (oldStatus === undefined) {
+            return
+          }
+          dispatch(
+            showNotification({
+              title: 'Order executed',
+              message: 'Your order has been successful',
+              groupKey,
+              variant: 'info',
```
usame-algan, line 130:
```suggestion
              variant: 'success',
```

src/store/swapOrderSlice.ts
```diff
@@ -0,0 +1,206 @@
...
+          }
+          dispatch(
+            showNotification({
+              title: 'Order executed',
+              message: 'Your order has been successful',
+              groupKey,
+              variant: 'success',
+            }),
+          )
+          dispatch(slice.actions.deleteSwapOrder(swapOrder.orderUid))
```
usame-algan, line 133: I think we need to dispatch the deletion as the first thing in case `oldStatus === undefined` we wouldn't delete the entry. This can happen when a swap gets filled and the user navigates back to the swap page. The widget emits a fulfilled event so we dispatch the state again but never remove it after that.

usame-algan (approved):

PR #3652, opened 2024-05-03
feat: fetch contract info for contract interactions

schmanu:
## What it solves
Contract interactions where not decoded using known addresses.

## How this PR fixes it
- Fetches contract info from known addresses for every `to` address during tx decoding.

## How to test it
- Initiate txs and observe the correct contract names being shown.

## Checklist

### DIFF
--- a/src/components/common/NamedAddress/index.tsx
+++ b/src/components/common/NamedAddress/index.tsx
@@ -1,0 +1,21 @@
+import useAsync from '@/hooks/useAsync'
+import useChainId from '@/hooks/useChainId'
+import { getContract } from '@safe-global/safe-gateway-typescript-sdk'
+import EthHashInfo from '../EthHashInfo'
+import type { EthHashInfoProps } from '../EthHashInfo/SrcEthHashInfo'
+
+const NamedAddressInfo = ({ address, name, customAvatar, ...props }: EthHashInfoProps) => {
+  const chainId = useChainId()
+  const [contract] = useAsync(() => getContract(chainId, address), [address, chainId])
+
+  return (
+    <EthHashInfo
+      address={address}
+      name={name || contract?.displayName || contract?.name}
+      customAvatar={customAvatar || contract?.logoUri}
+      {...props}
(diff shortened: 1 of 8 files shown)


jmealy (approved):

src/components/common/NamedAddress/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  return (
+    <EthHashInfo
+      address={address}
+      name={name || contract?.displayName || contract?.name}
+      customAvatar={customAvatar || contract?.logoUri}
+      {...props}
+    />
+  )
+}
```
usame-algan, line 19: Is this doing the same as the `onlyName` property we added to `EthHashInfo`? I couldn't test it since the swap widget is disabled on prod cgw and known names are not available on staging cgw

schmanu: It fetches the name from the CGW `/contract/{address}` endpoint if no name is given. 
I realized that we never fetch the contract name when we show the to address in "Interact with" fields of transactions or inside batch transactions.

src/components/common/NamedAddress/index.tsx
```diff
@@ -0,0 +1,21 @@
+import useAsync from '@/hooks/useAsync'
+import useChainId from '@/hooks/useChainId'
+import { getContract } from '@safe-global/safe-gateway-typescript-sdk'
+import EthHashInfo from '../EthHashInfo'
+import type { EthHashInfoProps } from '../EthHashInfo/SrcEthHashInfo'
+
+const NamedAddressInfo = ({ address, name, customAvatar, ...props }: EthHashInfoProps) => {
+  const chainId = useChainId()
+  const [contract] = useAsync(() => getContract(chainId, address), [address, chainId])
```
katspaugh, line 9: Seems like this would increase the amount of requests per page quite a lot. Should we do it only if there's no name passed in props maybe?

usame-algan (approved):

PR #3654, opened 2024-05-03
Feat: Block custom swap recipients if the are in the ofac block list

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/07996126a3c041db90e36e01eb4ce46c?v=03b2eaffc9af4b48b1bf57e50546ac3d&p=a0a641560784420ea65fa72c1e036d74&pm=s

Related: https://github.com/safe-global/safe-wallet-web/pull/3615

## How this PR fixes it
- When the recipient address is changed, check against the OFAC list.
    - If so, show the blocked address disclaimer

## Checklist


src/features/swap/index.tsx
```diff
@@ -126,8 +133,12 @@ const SwapWidget = ({ sell }: Params) => {
...
         event: CowEvents.ON_CHANGE_TRADE_PARAMS,
         handler: (newTradeParams) => {
-          const { orderType: tradeType } = newTradeParams
-          dispatch(setSwapParams({ tradeType }))
+          const { orderType: tradeType, recipient } = newTradeParams
+          if (recipient && isBlockedAddress(recipient)) {
+            setBlockedAddress(recipient)
+          } else {
+            dispatch(setSwapParams({ tradeType }))
+          }
```
compojoom, line 141: nit: I would move this a bit around. I would always disptach a setSwapParams(this is indipendent of the recipient). Then afterwards check if we need to block the recipient.

compojoom (commented):
since we are reusing the block for "signer address", the message that we display for custom recipient is: "this signer address". Personally I feel like this is not something we have to optimize for good user experience, but on the other hand... a correct error message is a basic right in this world :)

usame-algan:
> since we are reusing the block for "signer address", the message that we display for custom recipient is: "this signer address". Personally I feel like this is not something we have to optimize for good user experience, but on the other hand... a correct error message is a basic right in this world :)

I removed the "signer" part so it just says "This address ..." which sounds more accurate now.

PR #3662, opened 2024-05-07
Fix: do not show pending status if swap order is expired

jmealy:
## What it solves

Resolves:https://www.notion.so/safe-global/Loader-icon-on-top-of-the-expired-tag-4ddba4d2f2564ab0b1481b92ca6451a7?pvs=4

## How this PR fixes it
- If a pending swap transaction takes longer than the allotted time to execute
    - show the transaction as expired only. 
    - Do not give the option to speed up the transaction.

## Checklist


compojoom (approved):
I think it is fine.

PR #3666, opened 2024-05-07
feat: add partially filled state [SWT-51]

compojoom:
Resolves: https://www.notion.so/safe-global/8f5927d10d3643f3a0d2ebc02c8eb425?v=0177347b11784ea69b1883fd88be37ab&p=3b1c5384735c405d8ba91a7fd47772b8&pm=s

Cowswap is showing partially filled limit orders - both expired and open as "partially filled". We now do the same. 

Note: i was not able to test with buy limit orders as on sepolia non of the orders I created got filled at all. 

## How to test it
Create a limit order and it it gets partially filled, you should see it as partially filled. An easy way to create a partially filled order would be to try to sell more tokens than you own. Then cowswap would just sell what you have and will say that it can't fulfil the rest. 

## Checklist

### DIFF
--- a/src/features/swap/helpers/__tests__/utils.test.ts
+++ b/src/features/swap/helpers/__tests__/utils.test.ts
@@ -197,7 +197,7 @@ describe('Swap helpers', () => {
       expect(result).toBe(true)
     })
 
-    it('returns false if a buy order is fully filled', () => {
+    it('returns false if a buy order is not fully filled or fully filled', () => {
       const mockOrder = {
         executedBuyAmount: '0',
         buyAmount: '100000000000000000000', // 100 tokens
@@ -209,6 +209,12 @@ describe('Swap helpers', () => {
       const result = isOrderPartiallyFilled(mockOrder)
 
       expect(result).toBe(false)
+
+      const result1 = isOrderPartiallyFilled({
+        ...mockOrder,
+        executedBuyAmount: '100000000000000000000', // 100 tokens
+      })
+      expect(result1).toBe(false)
     })
 
     it('returns true if a sell order is partially filled', () => {
@@ -225,7 +231,7 @@ describe('Swap helpers', () => {
       expect(result).toBe(true)
     })
 
-    it('returns false if a sell order is fully filled', () => {
+    it('returns false if a sell order is not fully filled or fully filled', () => {
       const mockOrder = {
         sellAmount: '100000000000000000000',
         executedSellAmount: '0',
@@ -237,6 +243,13 @@ describe('Swap helpers', () => {
       const result = isOrderPartiallyFilled(mockOrder)
 
       expect(result).toBe(false)
+
+      const result1 = isOrderPartiallyFilled({
+        ...mockOrder,
+        executedSellAmount: '100000000000000000000', // 100 tokens
+      })
+
+      expect(result1).toBe(false)
     })
   })
 })


src/features/swap/helpers/utils.ts
```diff
@@ -114,3 +114,18 @@ export const getOrderClass = (order: Pick<SwapOrder, 'fullAppData'>): latest.Ord
...
+  const buyAmount = BigInt(order.buyAmount)
+  const executedSellAmount = BigInt(order.executedSellAmount)
+  const sellAmount = BigInt(order.sellAmount)
+
+  if (order.kind === OrderKind.BUY) {
+    return executedSellAmount !== 0n && executedSellAmount < sellAmount
+  }
+
+  return BigInt(executedBuyAmount) !== 0n && executedBuyAmount < buyAmount
+}
```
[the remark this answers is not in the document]
compojoom: We've agreed that cgw forwards the data from the cow order API and augments it with token information. Everything else, we do on our end.

src/features/swap/helpers/__tests__/utils.test.ts
```diff
@@ -175,4 +181,62 @@ describe('Swap helpers', () => {
...
+
+      const result = isOrderPartiallyFilled(mockOrder)
+
+      expect(result).toBe(true)
+    })
+
+    it('returns false if a sell order is fully filled', () => {
+      const mockOrder = {
+        sellAmount: '100000000000000000000',
+        executedSellAmount: '0',
```
usame-algan, line 231: Wouldn't this order be not filled at all?

src/features/swap/helpers/__tests__/utils.test.ts
```diff
@@ -175,4 +181,62 @@ describe('Swap helpers', () => {
...
+
+      const result = isOrderPartiallyFilled(mockOrder)
+
+      expect(result).toBe(true)
+    })
+
+    it('returns false if a buy order is fully filled', () => {
+      const mockOrder = {
+        executedBuyAmount: '0',
+        buyAmount: '100000000000000000000', // 100 tokens
```
usame-algan, line 203: Same here, shouldn't `executedBuyAmount` be the same as `buyAmount` for it to be fully filled?

src/features/swap/components/SwapOrder/index.tsx
```diff
@@ -105,7 +113,7 @@ export const SellOrder = ({ order }: { order: SwapOrderType }) => {
           <OrderId orderId={uid} href={explorerUrl} />
         </DataRow>,
         <DataRow key="Status" title="Status">
-          <StatusLabel status={status} />
+          <StatusLabel status={isPartiallyFilled ? 'partiallyFilled' : status} />
```
usame-algan, line 116: nit: This could be extracted into a `getOrderStatus` helper together with the other helper functions so that the logic is in one place.

usame-algan (approved):

src/features/swap/components/SwapOrder/index.tsx
compojoom: It's a bit confusing and that's why I left like this.
The orderbook api only returns an expired status, there is no partiallyFilled. PartiallyFilled is a "meta" status only when expired and filled > 0. 
defining a function named `getOrderStatus` could imply that this is the order status, but it isn't. 
defining a function named `getOrderMetaStatus` could work, but I would have to explain what a "metaStatus" is, so I opted to have the more direct thing in the StatusLabel as this is basically the only place where we care about it.

src/features/swap/helpers/__tests__/utils.test.ts
compojoom: test title is wrong. The condition for a partial fill is

executedBuyAmount !== 0 and executedBuyAmount < buyAmount.

With the test as is, I've just tested the left side of the check. 🤦

PR #3671, opened 2024-05-08
Feat: resolve Safe address ENS in sidebar

katspaugh:
## What it solves

If a Safe doesn't have a name in the Address Book, do a reverse ENS lookup in the Sidebar.

## How to test it

Can be tested with the Safe from the PR preview link.

## Screenshots

Screenshot 2024-05-08 at 09 48 27


usame-algan (approved):

francovenica:
Looks good

Question. Is the scope of this ticket only the sidebar? I noticed that the ENS name is not loaded for the safe in the watchlist (if it is there) like the address book entry does.
If is not, then I can create a ticket for that and pass this one

katspaugh:
Yes, it's only in this one place. I'm planning to extend it to the Safe List (including the watchlist) but need to check that it's not too many RPC requests. No need for a ticket at this time.

PR #3672, opened 2024-05-08
Fix: Account for when fullAppData is undefined on swap orders 

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Look-at-our-appData-usage-714b0f9b7639465ab39b5d13d2d102b0

According to the cowswap team, fullAppData can not be assumed to always be defined, so we must include default values for when it isn't.

## Checklist

### DIFF
--- a/src/features/swap/helpers/utils.ts
+++ b/src/features/swap/helpers/utils.ts
@@ -105,12 +105,14 @@ export const getFilledAmount = (
 
 export const getSlippageInPercent = (order: Pick<SwapOrder, 'fullAppData'>): string => {
   const fullAppData = order.fullAppData as AnyAppDataDocVersion
-  const slippageBips = (fullAppData.metadata.quote as latest.Quote).slippageBips || 0
+  const slippageBips = (fullAppData?.metadata?.quote as latest.Quote)?.slippageBips || 0
 
   return (Number(slippageBips) / 100).toFixed(2)
 }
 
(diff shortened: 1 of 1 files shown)


src/features/swap/helpers/utils.ts
```diff
@@ -105,12 +105,14 @@ export const getFilledAmount = (
...
 
   return (Number(slippageBips) / 100).toFixed(2)
 }
 
 export const getOrderClass = (order: Pick<SwapOrder, 'fullAppData'>): latest.OrderClass1 => {
   const fullAppData = order.fullAppData as AnyAppDataDocVersion
-  return (fullAppData.metadata.orderClass as latest.OrderClass).orderClass
+  const orderClass = (fullAppData?.metadata?.orderClass as latest.OrderClass)?.orderClass
+
+  return orderClass || 'limit'
```
compojoom, line 117: No, I think we should assume default to be "market", no? That's the default state of the widget

jmealy: I was basing it off of this: https://5afe.slack.com/archives/C03DVQREAH0/p1714383629475399?thread_ts=1714139366.988719&cid=C03DVQREAH0

But looks like that is just because they treat all orders as limit orders behind the scenes which is not really relevant to us here. So we can go with market since that will presumably be more common

compojoom: Interesting! 

Thanks for changing it. We'll monitor this. I don't think that we'll end up in a situation without it, but who knows.

PR #3676, opened 2024-05-08
fix: Close transaction modals as soon as the transaction is submitted

### LINKED ISSUE #3675: Transaction modal doesn't close when not using an unchecked signer
## Bug description

When executing a transaction without an unchecked signer e.g. via "Bulk execute", the modal doesn't close until the transaction has been executed.

This is most likely an issue introduced with a recent MetaMask update and is also reported [here](https://github.com/ethers-io/ethers.js/issues/4722).

## Steps to reproduce
 1. Go to a safe and queue at least two transactions
 2. Press Bulk execute and submit the transaction in your wallet

## Expected result

The modal closes

## Obtained result

The modal is visible

usame-algan:
## What it solves

Resolves #3675 

## How this PR fixes it

- Uses the unchecked signer when executing transactions

## How to test it

This needs to be tested with MetaMask and at least one other wallet thats not affected e.g. Rabby.
We should also make sure it still works when executing transactions with hardware wallets and nested safes.

The following transaction types need to be tested:
- Spending limit execution
- Bulk execution
- Tx speedup
- Recovery setup
- Recovery proposal

The transaction modal should close when the transaction is executed from the wallet.

## Checklist


usame-algan:
I've added a new helper `getUncheckedSigner` and used it in all dispatch calls where we previously relied on the signer. We need to test these cases with 1/n and n/m setups and other types of wallets like hardware wallets and nested safes in order to ensure backwards compatibility.

schmanu (approved):

jmealy (approved):

katspaugh:
Could you please rename the PR to reflect what exactly is fixed on the UI level? Without using the word “unchecked signer” because it’s an implementation detail.

francovenica:
LGTM

I tried several types of tx: 
Send funds, owner editing, recovery setup and recovery execution, spending limit creation and use, Bulk execution, batch tx, Tx rejection execution, speeding up tx
I tried with both MM and Rabby

PR #3679, opened 2024-05-08
Fix: queue changes to mobile layout on desktop resolutions

jmealy:
## What it solves
- Queue switches to mobile/tablet view on too large screens
- grouped transactions are not aligned

resolves: https://www.notion.so/safe-global/Issue-dancing-view-for-the-Queued-because-of-expired-state-f81f11c9114d4e8e9975708b3fe899db [SWT-69]

## Screenshots
image

## Checklist

### DIFF
--- a/src/components/transactions/TxSummary/styles.module.css
+++ b/src/components/transactions/TxSummary/styles.module.css
@@ -26,11 +26,13 @@
 
 .gridContainer.history {
   grid-template-columns: var(--grid-nonce) var(--grid-type) var(--grid-info) var(--grid-date) var(--grid-status);
-  grid-template-areas: 'nonce type info date status';
+  grid-template-areas: 'nonce type info date status ';
 }
 
 .gridContainer.grouped {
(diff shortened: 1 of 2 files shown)


src/components/transactions/TxSummary/styles.module.css
```diff
@@ -46,7 +48,7 @@
   color: var(--color-text-secondary);
 }
 
-@media (max-width: 1460px) {
+@media (max-width: 1350px) {
```
usame-algan, line 51: Previously we had 1200px here which is a default MUI breakpoint. We should try to stick to that as much as possible so the whole UI changes layout at the same viewports.

jmealy: In this case there is no way to fit everything in one line with 1200px without the execute button at the end getting cut off

jmealy: that is assuming all fields are as wide as they can potentially be

compojoom: I'm for merging this as is @usame-algan and opening a new issue for 1200px. This is good enough and @liliya-soroka confirmed that it is fine on her 13 inch macbook.

PR #3684, opened 2024-05-10
fix: Send orderUid with analytics events

usame-algan:
## What it solves

Resolves [Notion issue](https://www.notion.so/safe-global/Send-order-status-to-GA-740c268ea08f4c929adc3d6c4b3fb041)

## How this PR fixes it

- Sends the `orderUid` as a predefined dimension called `transaction_id` with certain events such as `tx_created`, `tx_confirmed`, `tx_executed`, `on-chain interaction`

## How to test it

1. Create a swap order
2. Press submit in the modal
3. Observe the GTM event containing `transaction_id` with the `orderUid` as the value

## Checklist

### DIFF
--- a/src/hooks/useTxTracking.ts
+++ b/src/hooks/useTxTracking.ts
@@ -1,6 +1,7 @@
 import { trackEvent, WALLET_EVENTS } from '@/services/analytics'
 import { getTxDetails } from '@/services/transactions'
 import { TxEvent, txSubscribe } from '@/services/tx/txEvents'
+import { isSwapTxInfo } from '@/utils/transaction-guards'
 import { useEffect } from 'react'
 import useChainId from './useChainId'
 
@@ -22,16 +23,20 @@ export const useTxTracking = (): void => {
         const id = txId || txHash
 
         let origin = ''
+        let transaction_id
+
         if (id) {
           try {
             const txDetails = await getTxDetails(chainId, id)
             origin = txDetails.safeAppInfo?.url || ''
+            transaction_id = isSwapTxInfo(txDetails.txInfo) ? txDetails.txInfo.uid : undefined
           } catch {}
         }
 
         trackEvent({
           ...analyticsEvent,
           label: origin,
+          transaction_id,
         })
       }),
     )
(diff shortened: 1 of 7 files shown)


src/services/analytics/gtm.ts
```diff
@@ -107,28 +107,35 @@ type SafeAppGtmEvent = ActionGtmEvent & {
...
     ...commonEventParams,
-    event: eventData.event || EventType.CLICK,
-    eventCategory: eventData.category,
-    eventAction: eventData.action,
-    chainId: eventData.chainId || commonEventParams.chainId,
+    event: event || EventType.CLICK,
+    eventCategory: category,
+    eventAction: action,
+    chainId: chainId || commonEventParams.chainId,
+    ...rest,
```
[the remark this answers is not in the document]
usame-algan: We have to manually add each parameter to GTM otherwise it won't be forwarded.

src/hooks/useTxTracking.ts
```diff
@@ -22,16 +23,20 @@ export const useTxTracking = (): void => {
         const id = txId || txHash
 
         let origin = ''
+        let transaction_id
+
         if (id) {
           try {
             const txDetails = await getTxDetails(chainId, id)
             origin = txDetails.safeAppInfo?.url || ''
+            transaction_id = isSwapTxInfo(txDetails.txInfo) ? txDetails.txInfo.uid : undefined
```
compojoom, line 32: this is a bit confusing to me - why do we have to return transaction_id as undefined? Can't we just not trigger trackEvent later without it?

usame-algan: If we set the `transaction_id` in the data layer once it will always be there for all subsequent events even if they don't set a `transaction_id` so we have to set it to undefined explicitely.

compojoom: confusing. The signature of the function looks like we should be tracking separate events. The dataLayer is an array of Events. Why would the transaction_id of event 1 in that array apply to event 10 in the array?

compojoom (approved):

francovenica:
Tested this in the RC directly.
Question:
When the tx is executed you get a tx id just fine
image

But when you just sign the tx there is no tx_id. 
I'd assume that that value should be at least "defined" with the tx proposal.
image

PR #3685, opened 2024-05-10
Fix: Set initial sell token when entering swaps through assets page

jmealy:
## What it solves
Resolves: https://www.notion.so/safe-global/WETH-is-selected-in-the-widget-when-pressing-Swap-for-ETH-on-the-assets-page-1e9981aed97340979538bfd2c43b51c2?pvs=4 [SWT-73]

When clicking the swap option on an asset, that asset was not set as the sell token in the widget. It was always initialised with WETH.

## How this PR fixes it
- Include the chainId in the swap widget params. It seems to be necessary to be able to set the sell token.

## How to test it
- click on the swap button on an asset in the assets page
- the swap widget should open with the chosen asset as the sell token, with the amount set to the token balance.

## Checklist

### DIFF
--- a/src/features/swap/index.tsx
+++ b/src/features/swap/index.tsx
@@ -25,6 +25,7 @@ import LegalDisclaimerContent from '@/components/common/LegalDisclaimerContent'
 import { isBlockedAddress } from '@/services/ofac'
 import { selectSwapParams, setSwapParams, type SwapState } from './store/swapParamsSlice'
 import { setSwapOrder } from '@/store/swapOrderSlice'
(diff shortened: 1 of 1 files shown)


usame-algan (approved):
`chainId` needs to be included in the dependency array but other than that it works nicely, thanks!

@compojoom do you remember why we removed the `chainId` from the config? I guess it is needed after all

compojoom:
Yes, I remember. It was because of this PR feedback https://github.com/safe-global/safe-wallet-web/pull/3231#pullrequestreview-2008448234 from @anxolin

PR #3688, opened 2024-05-13
fix: remove swap from counterfactual safes [SWAP-66]

compojoom:
Resolves #

Due to a bug, the swap widget is not picking up the creation of a counterfactual safe. 

## How this PR fixes it
Until this is resolved, we’ll disable the swap feature for counterfactual safes

## How to test it
Create a counterfactual safe. The swap button should be missing from the dashboard, assets and navigation. For the first transaction modal there is a swap button, but it should lead to the safe apps page and not the native integration.

## Checklist

### DIFF
--- a/src/components/balances/AssetsTable/index.tsx
+++ b/src/components/balances/AssetsTable/index.tsx
@@ -27,6 +27,7 @@ import { TxModalContext } from '@/components/tx-flow'
 import { TokenTransferFlow } from '@/components/tx-flow/flows'
 import AddFundsCTA from '@/components/common/AddFunds'
 import SwapButton from '@/features/swap/components/SwapButton'
+import useIsCounterfactualSafe from '@/features/counterfactual/hooks/useIsCounterfactualSafe'
 
 const skeletonCells: EnhancedTableProps['rows'][0]['cells'] = {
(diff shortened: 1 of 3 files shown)


src/components/dashboard/Overview/Overview.tsx
```diff
@@ -47,7 +48,8 @@ const Overview = (): ReactElement => {
   const router = useRouter()
   const theme = useTheme()
   const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'))
-  const isSwapFeatureEnabled = useHasFeature(FEATURES.NATIVE_SWAPS)
+  const isCounterfactualSafe = useIsCounterfactualSafe()
+  const isSwapFeatureEnabled = useHasFeature(FEATURES.NATIVE_SWAPS) && !isCounterfactualSafe
```
usame-algan, line 52: The overview buttons are already hidden for counterfactual safes so this is not necessary.

usame-algan (approved):

liliya-soroka:
@compojoom , could we navigate the user on the networks with Native Swap to the safe-apps filtered list as we do for the networks without native swap?

PR #3703, opened 2024-05-14
refactor: Remove safe token paused logic

usame-algan:
## What it solves

Removes the logic that checks if the safe token is paused since its unpaused now.

## How to test it

1. Create a transaction and select the safe token
2. Observe no warning

## Checklist


katspaugh (approved):
Thanks!

francovenica:
Looks good.

Tested on send funds, the "Send" button in assets, the creation of a spending limit for the Safe token and the use of that spending limit. I tried also a swap using the new Swap integration and that one worked fine too.

mike10ca:
> @mike10ca do any e2e tests need to be adjusted for this change? I couldn't find anything.

@usame-algan we don't have such tests

PR #3707, opened 2024-05-14
Slim Down Docker Image

nlordell:
## What it solves

When playing around with the Safe web interface I noticed that the Docker image that it produces is **over 6GB in size**! This is very large for a docker image that just hosts static files.

## How this PR fixes it

This PR slims down the docker image significantly by using a multistage build where the first stage builds the static website, and the actual image just hosts the static webside (here using BusyBox `httpd` which is nice and lightweight).

This gets the image down to around 72MB, around 1% of the original image size _(sizes computed from Docker images built on amd64 Linux)_.

Note that there is one weird detail is that static HTTP servers typically don't have support for automatically adding `.html` endings to URL paths. It was worked around here with symlinks. See comment in the Dockerfile for more details.

## How to test it

Build the docker image and run the container locally:

```sh
docker build . -t safe-wallet-web
docker run --rm -p 3000:3000 safe-wallet-web
```

This should give you a local interface at <http://localhost:3000>.

## "Screenshots"

```
$ docker images
REPOSITORY                            TAG          IMAGE ID      CREATED            SIZE
localhost/safe-web                    small        2bb37df44d9f  14 minutes ago     71.9 MB
localhost/safe-web                    big          ae2940e99e19  3 hours ago        6.52 GB
```

## Checklist

    * Analytics are not affected
    * Not applicable

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM node:18-alpine
+FROM node:18-alpine AS build
+
 RUN apk add --no-cache libc6-compat git python3 py3-pip make g++ libusb-dev eudev-dev linux-headers
 WORKDIR /app
 COPY . .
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Amazing, thank you!

PR #3708, opened 2024-05-14
fix: Safe stats tracking for owned Safes

### LINKED ISSUE #3702: Tracking of owned and added Safes logs wrong values
## Bug description
There are certain conditions in which we log 0 owned / added Safes:
- If no wallet is connected, we always log that the user owns 0 Safes, after connecting a wallet and loading the actual owned Safes they are not logged again
- First page load
  - If a user navigates directly to the welcome page (default) with a connected wallet we log 0 owned and watched Safes - even if there are Safes in the local storage watchlist
  - We also always log 0 owned Safes on page load with connected wallet.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
Open the welcome page on dev and observe the GTM events

## Expected result
- If a user has no wallet connected we should not log any owned Safes
- If a wallet gets connected / owner changed we should log the correct owned Safes
- On page load we should log correct owned and watched Safes

## Obtained result
See above

schmanu:
## What it solves

Resolves #3702 

## How this PR fixes it
- Track owned Safes whenever the connected wallet address changes
- Do not track 0 owned Safes or 0 watchlist Safes as these usually relate to loading states and are not really interesting
- Fixes a bug that caused the cached owned Safes to get overwritten with wrong values

## How to test it
- Load the /welcome/accounts page and check analytics
- Switch the connected wallet whilst being on this page and observe the correct owned Safes to be tracked

## Checklist

### DIFF
--- a/src/components/welcome/MyAccounts/useAllSafes.ts
+++ b/src/components/welcome/MyAccounts/useAllSafes.ts
@@ -7,8 +7,6 @@ import useAllOwnedSafes from './useAllOwnedSafes'
 import useChains from '@/hooks/useChains'
 import useWallet from '@/hooks/wallets/useWallet'
 import { selectUndeployedSafes } from '@/store/slices'
-import { sameAddress } from '@/utils/addresses'
-
 export type SafeItem = {
   chainId: string
   address: string
@@ -53,10 +51,8 @@ const useAllSafes = (): SafeItems | undefined => {
       const uniqueAddresses = uniq(addedOnChain.concat(ownedOnChain)).filter(Boolean)
 
       return uniqueAddresses.map((address) => {
-        const owners = allAdded?.[chainId]?.[address]?.owners
-        const isOwner = owners?.some(({ value }) => sameAddress(walletAddress, value))
         const isUndeployed = undeployedOnChain.includes(address)
-        const isOwned = (ownedOnChain || []).includes(address) || isOwner
+        const isOwned = (ownedOnChain || []).includes(address)
         return {
           address,
           chainId,
@@ -64,7 +60,7 @@ const useAllSafes = (): SafeItems | undefined => {
         }
       })
     })
-  }, [allAdded, allOwned, configs, undeployedSafes, walletAddress])
+  }, [allAdded, allOwned, configs, undeployedSafes])
 }
 
 export default useAllSafes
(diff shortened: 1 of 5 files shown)


src/components/welcome/MyAccounts/useAllSafes.ts
```diff
@@ -53,18 +51,16 @@ const useAllSafes = (): SafeItems | undefined => {
       const uniqueAddresses = uniq(addedOnChain.concat(ownedOnChain)).filter(Boolean)
 
       return uniqueAddresses.map((address) => {
-        const owners = allAdded?.[chainId]?.[address]?.owners
-        const isOwner = owners?.some(({ value }) => sameAddress(walletAddress, value))
         const isUndeployed = undeployedOnChain.includes(address)
-        const isOwned = (ownedOnChain || []).includes(address) || isOwner
+        const isOwned = (ownedOnChain || []).includes(address)
```
katspaugh, line 55: Could you clarify this fix? It looks like it's now doing less than it did before.

schmanu: Due to this we get owned Safes from the added Safes while the reply from the owned Safes endpoint is still pending leading to us logging fewer owned Safes than we actually have.
E.g a user has 30 owned Safes but only 5 Safes added of which the connected wallet is also the owner.
We will return 5 owned Safes initially and then the 30 owned Safes leading to the tracking only capturing 5/30 Safes.

schmanu: But thinking about it some more this might affect counterfactual Safes as they are not part of the backend reply. Will double check that and def. write some tests for this hook. It has so much untested logic.

katspaugh (approved):
👍

francovenica:
So the main fix is fine. The track now shows amount of safes owned and watchlist. Also when you change owner it throws new events with the amount of safe owned for the new connected owner

An issue I'm having is that "Watchlist" looks empty when there is no owner connected at all (even tho there are safes added to the watchlist)
watchlist

schmanu:
> So the main fix is fine. The track now shows amount of safes owned and watchlist. Also when you change owner it throws new events with the amount of safe owned for the new connected owner
> 
> An issue I'm having is that "Watchlist" looks empty when there is no owner connected at all (even tho there are safes added to the watchlist) Currently in dev the safe in the watchlist show when you are disconnected

@francovenica 
I also cannot seem to reproduce this. Could you recheck?
here is a recording of the account list when I disconnect. The watchlist stays in place as expected (I have 3 Safes on it)

https://github.com/safe-global/safe-wallet-web/assets/2670790/55012d9b-a52b-4bdf-a098-9032e2e25b11

francovenica:
I still can see the issue. let's check it in a call to see what's different on my side
watchlist

francovenica:
The issue was fixed. I can see the watched safes as a disconnected user just fine.

LGTM

PR #3710, opened 2024-05-15
fix: Hide signed transaction from untrusted pending queue

### LINKED ISSUE #3215: TX execution doubles categories/txs on UI while execution is in progress
## Bug description
Bulk execution doubles categories/txs on UI while execution is in progress

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Sepolia

## Steps to reproduce
 1. Add a few txs in queue and sign fully
 2. click bulk execution with relay -> start the process
 3. check the queue view
 Current result: 
 The queue list contains:
 pending
queued
 next
 queued
categories at the same time and the txs are displayed twice on UI

## Expected result
Make sure that the categories and txs are not doubled , as this is quite confusing 
## Screenshots
image

usame-algan:
## What it solves

Resolves #3215

## How this PR fixes it

- Only shows unsigned transactions in the untrusted pending queue
- Changes the first label of the queue to always show "Pending" when a transaction is pending

## ToDos

- [x] Write tests for `usePendingTxsQueue`

## How to test it

A few variations in 1/n and n/m safes should be tests
1. Queue 2 transactions and execute the first one
2. Queue 2 transactions and bulk execute
3. Queue 1 transaction, create another one and immediately execute that one to replace the first

## Checklist

### DIFF
--- a/src/components/transactions/TxList/index.tsx
+++ b/src/components/transactions/TxList/index.tsx
@@ -1,10 +1,13 @@
+import GroupedTxListItems from '@/components/transactions/GroupedTxListItems'
+import { useHasPendingTxs } from '@/hooks/usePendingTxs'
+import { isLabelListItem } from '@/utils/transaction-guards'
+import { groupConflictingTxs } from '@/utils/tx-list'
+import { Box } from '@mui/material'
+import type { LabelValue, TransactionListPage } from '@safe-global/safe-gateway-typescript-sdk'
+import { TransactionListItemType } from '@safe-global/safe-gateway-typescript-sdk'
 import type { ReactElement, ReactNode } from 'react'
 import { useMemo } from 'react'
-import { Box } from '@mui/material'
-import type { TransactionListPage } from '@safe-global/safe-gateway-typescript-sdk'
 import TxListItem from '../TxListItem'
-import GroupedTxListItems from '@/components/transactions/GroupedTxListItems'
-import { groupConflictingTxs } from '@/utils/tx-list'
 import css from './styles.module.css'
 
 type TxListProps = {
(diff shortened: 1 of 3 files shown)


src/components/transactions/TxList/index.tsx
```diff
@@ -15,8 +18,24 @@ export const TxListGrid = ({ children }: { children: ReactNode }): ReactElement
   return <Box className={css.container}>{children}</Box>
 }
 
+const adjustPendingLabel = (items: TransactionListPage['results']) => {
+  const result = [...items]
+
+  if (isLabelListItem(result[0])) {
+    result[0] = {
+      label: 'Pending' as LabelValue,
```
katspaugh, line 26: Why is it in two places now?

src/hooks/usePendingTxs.ts
```diff
@@ -62,20 +67,24 @@ export const usePendingTxsQueue = (): {
     // Adjust the first label ("Next" -> "Pending")
     if (results[0] && isLabelListItem(results[0])) {
       results[0].label = 'Pending' as LabelValue
```
katspaugh, line 69: Here's the second place where it sets the label.

katspaugh (commented):
Tests are failing.

src/components/transactions/TxList/index.tsx
usame-algan: We have it as part of `usePendingTxsQueue` which is only visible for 1/n safes. I didn't want to hide/filter items from the normal queue so instead I am adjusting the label there and hide items from this untrusted queue thats only visible for 1/n safes.

usame-algan: As discussed lets not change this label and keep it as "Next" since this is the same behaviour in n/m safes anyway.

katspaugh (approved):
💪

jmealy (approved):

francovenica:
Listing scenarios as a checklist:

Safe 1/x
1 queued tx with nonce X, execute a 2nd tx with also the nonce X (basically replacing the one in queue)
3 queued txs with consecutive nonces and executed in bulk
3 queued txs, 2 of them consecutive nonces and the last one with a non-consecutive nonce to the others (so only the first 2 execute in bulk)
3 queued tx, only executing the 1st one

Safe 2/x
2 queued tx, first fully signed, second not fully signed. Execute the first one
2 queued tx, 2nd one with 2 tx with conflicting nonce, one of those fully signed the other not. Execute in bulk

Other: (Things that should not be affected, but checked just in case)
Check that the recovery tx is not affected by these changes
Delete tx and replace them, then execute those tx replacement in bulk

francovenica:
Question:
This case is queuing the tx with nonceX and then executing a 2nd tx also with nonceX, replacing the one in queue.
This is how the queue table looks:
image

I'm not sure it looks good having it duplicated, but also I'm not sure if there is a way to fix this situation. So not a bug, but a weird situation to open a debate on if there is something to be done

katspaugh:
It's always been like this in 1/X Safes, let's not change it now.

PR #3712, opened 2024-05-15
Fix: Show loading state in WC input between pairing and receiving proposal

### LINKED ISSUE #3598: [WalletConnect] Loader disappears too early after pasting
- The loader should appear right after pasting
- And disappaear only after the proposal screen is shown or an auto-confirmed session is established

Slack thread: https://5afe.slack.com/archives/C0462F65CLD/p1713802598287739

### LINKED ISSUE #3731: Add timeout to WalletConnect proposal
## Bug description
Sometimes after pasting the WC pairing code, it will be stuck in the loading state until the user refreshes the page. 
Possibly because we do not receive a proposal back from WC.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
I was not able to reliably reproduce this, it appears to happen sporadically when using the WC widget.

## Expected result
- the connection should timeout and show and error message after a certain amount of time (eg. 30 seconds)

## Obtained result
The widget will be stuck in a loading state and the widget is unusable until the user refreshes the page.

jmealy:
## What it solves

Resolves #3598
Resolves #3731

## How this PR fixes it
After successfully pairing with wallet connect, and before receiving a proposal, keep the paste button in its loading state.
Only remove the loading state after the proposal is received and handled.

## How to test it
note: This bug only appears when Walletconnect is slow to send the proposal from their side.
- connect a safe to any dapp using walletconnect.
- after pasting the pairing code, the loading state should appear.
- the loading state should persist until the WC modal closes.

## Checklist

### DIFF
--- a/src/features/walletconnect/components/WalletConnectUi/index.tsx
+++ b/src/features/walletconnect/components/WalletConnectUi/index.tsx
@@ -5,7 +5,7 @@ import useWalletConnectSessions from '@/features/walletconnect/hooks/useWalletCo
 import { WalletConnectContext } from '@/features/walletconnect/WalletConnectContext'
 import useWcUri from '../../hooks/useWcUri'
 import WcHeaderWidget from '../WcHeaderWidget'
-import WcSessionManager from '../WcSessionMananger'
+import WcSessionManager from '../WcSessionManager'
 import { WalletConnectProvider } from '../WalletConnectProvider'
 
 const WalletConnectWidget = () => {
--- a/src/features/walletconnect/components/WcInput/index.tsx
+++ b/src/features/walletconnect/components/WcInput/index.tsx
@@ -48,6 +48,7 @@ const WcInput = ({ uri }: { uri: string }) => {
         await walletConnect.connect(val)
       } catch (e) {
         setError(asError(e))
+        setIsLoading(undefined)
       }
     },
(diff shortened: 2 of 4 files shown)


usame-algan (changes requested):
We are disabling certain buttons when loading is set to true so I think we need to set loading to undefined in `WcSessionManager` as soon as we receive a proposal otherwise the user can't click anything

Screenshot 2024-05-16 at 10 19 51

usame-algan:
As a future idea: We could also add a timer for the loading state and give the user more feedback if it takes too long after connecting and not receiving a proposal or when trying to use an invalid wc uri. Right now we are showing an infinite loading state until the user reloads the app.

jmealy:
> Yes, the change Usame requested is necessary for this to pass. I'll kick it back.
> 
> Also I don't think the timeout should be something to do in the future. Maybe a 30 secs timeout at least?

I fixed the disabled buttons and added in the timeout of 30 seconds. After it times out it shows this error:

image

usame-algan (approved):
Nice, thanks for implementing the timeout! 🚀 

@francovenica could you do a last check with the timeout? Good to go from my side!

francovenica:
LGTM

PR #3722, opened 2024-05-15
Fix: make safe app id an int

katspaugh:
## What it solves

Custom Safe Apps get a random id which might be a problem when validating messages.

### DIFF
--- a/src/services/safe-apps/manifest.ts
+++ b/src/services/safe-apps/manifest.ts
@@ -99,7 +99,7 @@ const fetchSafeAppFromManifest = async (
   const iconUrl = getAppLogoUrl(normalizedAppUrl, appManifest)
 
   return {
-    id: Math.random(),
+    id: Math.round(Math.random() * Math.pow(10, 18)),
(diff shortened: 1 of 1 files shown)


src/services/safe-apps/manifest.ts
```diff
@@ -99,7 +99,7 @@ const fetchSafeAppFromManifest = async (
   const iconUrl = getAppLogoUrl(normalizedAppUrl, appManifest)
 
   return {
-    id: Math.random(),
+    id: Math.round(Math.random() * Math.pow(10, 18)),
```
schmanu, line 102: This leads to a number space that can still include low numbers which are given to actual Safe apps.
Lets add a billion or so to the random number: `id: Math.round(1_000_000_000 + Math.random() * Math.pow(10, 18))` to avoid collisions.

schmanu (changes requested):

schmanu (approved):

liliya-soroka:
Verified , works with https://safe-apps-test-app.pages.dev/

PR #3724, opened 2024-05-16
fix: fetch single message if safe message is not in the messagesSlice

### LINKED ISSUE #3718: Fully signed message is not returned to the dapp if it's not on the 1st page from back-end
## Bug description
Fully signed message is not returned to the dapp if it's not on the 1st page from back-end

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Connect https://app.eigenlayer.xyz/ via WC 
 2. sign T&C => sign off-chain message fully 
 3. doscinnest the https://app.eigenlayer.xyz/ app
 4. create many other messages, so the signed message from the https://app.eigenlayer.xyz/  appears on the 2nd page in the response from CGW
 5.  Connect https://app.eigenlayer.xyz/ via WC 
 6. sign T&C => try to sign
 Current result: The error that the message exists in the DB is displayed and there is no possibility to sign t&C to continue work with the dapp

Expected result: The fully signed message should be returned to the connected dapp regardless of pagination

schmanu:
## What it solves

Resolves #3718 

## How this PR fixes it
- Looks in the messageSlice and additionally fetches the message from the single message endpoint

## How to test it
- Sign 21+ messages in a Safe
- Then request the first message again

## Checklist

### DIFF
--- a/src/hooks/messages/useSafeMessage.ts
+++ b/src/hooks/messages/useSafeMessage.ts
@@ -1,10 +1,17 @@
 import { isSafeMessageListItem } from '@/utils/safe-message-guards'
-import type { SafeMessage } from '@safe-global/safe-gateway-typescript-sdk'
+import { type SafeMessage } from '@safe-global/safe-gateway-typescript-sdk'
 import { useState, useEffect } from 'react'
 import useSafeMessages from './useSafeMessages'
+import useAsync from '../useAsync'
+import useSafeInfo from '../useSafeInfo'
+import { fetchSafeMessage } from './useSyncSafeMessageSigner'
 
 const useSafeMessage = (safeMessageHash: string) => {
   const [safeMessage, setSafeMessage] = useState<SafeMessage | undefined>()
+
+  safeMessage?.type
+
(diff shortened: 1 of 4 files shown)


src/hooks/messages/useSafeMessage.ts
```diff
@@ -1,20 +1,32 @@
...
 import useSafeMessages from './useSafeMessages'
+import useAsync from '../useAsync'
+import useSafeInfo from '../useSafeInfo'
+import { fetchSafeMessage } from './useSyncSafeMessageSigner'
 
 const useSafeMessage = (safeMessageHash: string) => {
   const [safeMessage, setSafeMessage] = useState<SafeMessage | undefined>()
 
+  safeMessage?.type
+
```
katspaugh, line 13:
```suggestion
```

katspaugh (approved):
For posterity, this will always fetch the individual message from the backend, not only as a fallback, but es ist gut so.

liliya-soroka:
Verified.

PR #3725, opened 2024-05-16
fix: (WC) do not return Safe as available on all optional Namespaces

schmanu:
## What it solves
Sometimes the chainChanged event is ignored if the connected dapp did not acknowledge the new session yet.
Then the dapp is stuck with a wrong chain on which the Safe is not deployed.

## How this PR fixes it
- Only uses our workaround if the Safe is not present in the requiredNamespace

## How to test it
- Connect to uniswap with a polygon Safe
- observe that the Safe is correctly recognized on Polygon

## Checklist

### DIFF
--- a/src/features/walletconnect/services/WalletConnectWallet.ts
+++ b/src/features/walletconnect/services/WalletConnectWallet.ts
@@ -82,8 +82,7 @@ class WalletConnectWallet {
   }
 
   private getNamespaces(proposal: Web3WalletTypes.SessionProposal, currentChainId: string, safeAddress: string) {
-    // Most dApps require mainnet, but we aren't always on mainnet
-    // As workaround, we pretend include all required with the Safe chainId
+    // As workaround, we pretend to support all the required chains plus the current Safe's chain
     const requiredChains = proposal.params.requiredNamespaces[EIP155]?.chains || []
 
     const supportedChainIds = [currentChainId].concat(requiredChains.map(stripEip155Prefix))
--- a/src/features/walletconnect/services/WalletConnectWallet.ts
+++ b/src/features/walletconnect/services/WalletConnectWallet.ts
@@ -83,14 +83,10 @@ class WalletConnectWallet {
 
   private getNamespaces(proposal: Web3WalletTypes.SessionProposal, currentChainId: string, safeAddress: string) {
     // Most dApps require mainnet, but we aren't always on mainnet
-    // As workaround, we pretend include all required and optional chains with the Safe chainId
+    // As workaround, we pretend include all required with the Safe chainId
     const requiredChains = proposal.params.requiredNamespaces[EIP155]?.chains || []
-    const optionalChains = proposal.params.optionalNamespaces[EIP155]?.chains || []
 
-    const supportedChainIds = [currentChainId].concat(
-      requiredChains.map(stripEip155Prefix),
-      optionalChains.map(stripEip155Prefix),
-    )
+    const supportedChainIds = [currentChainId].concat(requiredChains.map(stripEip155Prefix))
 
     const eip155ChainIds = supportedChainIds.map(getEip155ChainId)
     const eip155Accounts = eip155ChainIds.map((eip155ChainId) => `${eip155ChainId}:${safeAddress}`)
--- a/src/features/walletconnect/services/__tests__/WalletConnectWallet.test.ts
+++ b/src/features/walletconnect/services/__tests__/WalletConnectWallet.test.ts
@@ -155,21 +155,13 @@ describe('WalletConnectWallet', () => {
 
       await wallet.approveSession(
         proposal,
-        '69420', // Not in proposal, therefore not supported
+        '43114', // Not in proposal, therefore not supported
         toBeHex('0x123', 20),
       )
 
       const namespaces = {
         eip155: {
-          chains: [
-            'eip155:1',
-            'eip155:43114',
-            'eip155:42161',
-            'eip155:8453',
-            'eip155:100',
-            'eip155:137',
-            'eip155:1101',
-          ],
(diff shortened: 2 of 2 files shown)


src/features/walletconnect/services/WalletConnectWallet.ts
```diff
@@ -83,14 +83,10 @@ class WalletConnectWallet {
 
   private getNamespaces(proposal: Web3WalletTypes.SessionProposal, currentChainId: string, safeAddress: string) {
     // Most dApps require mainnet, but we aren't always on mainnet
-    // As workaround, we pretend include all required and optional chains with the Safe chainId
+    // As workaround, we pretend include all required with the Safe chainId
```
katspaugh, line 86:
```suggestion
    // As workaround, we pretend to support all the required chains plus the current Safe's chain
```

katspaugh (approved):
🦸

src/features/walletconnect/services/WalletConnectWallet.ts
```diff
@@ -82,15 +82,10 @@ class WalletConnectWallet {
...
-    // As workaround, we pretend include all required and optional chains with the Safe chainId
+    // As workaround, we pretend to support all the required chains plus the current Safe's chain
     const requiredChains = proposal.params.requiredNamespaces[EIP155]?.chains || []
-    const optionalChains = proposal.params.optionalNamespaces[EIP155]?.chains || []
 
-    const supportedChainIds = [currentChainId].concat(
-      requiredChains.map(stripEip155Prefix),
-      optionalChains.map(stripEip155Prefix),
-    )
+    const supportedChainIds = [currentChainId].concat(requiredChains.map(stripEip155Prefix))
```
compojoom, line 88: If we don't add the optional chains wouldn't that mean that the moment you switch your network the connection will be closed?

schmanu: I will check

schmanu: You are right. I will readd those :)

schmanu: You can not return a chainId in the namespace without an account. So we would be stuck with the same workaround in the end which caused the bug.
I feel like having to reconnect after switching to another Safe is the lesser devil than not being able to connect on any chain other than mainnet. But We should discuss this.

francovenica:
I got this alert/prompt when I tried to connect in Uniswap with Polygon which I don't think is intended

image

francovenica:
I'm having different behaviors between current prod, this PR and Uniswap and I don't know if they are intended. 

In prod, you can easily change from one network to the other no problem, but if you try in this PR the Uniswap app will tell you that you wallet doesn't support the networks (tried also in prod CGW since it has all networks but with same result.

Although it might be considered an issue on Uniswap side, is weird that the behavior changes depending if we use the PR or current Prod
In prod:
uni in prod

Using this PR:
uni in PR

francovenica:
I tried to connect to OpenSea several times. In the safe app you can see the connection (in the top bar) but in openSea, or the WC modal with the QR code doesn't even close or the WC option in the modal where you select the wallet shows a infinite loading icon. 
I never got the message that OpenSea makes you sign to connect in the safe

I'm not don't know if this an issue from OpenSea side, but in current prod the connection works fine (the message pop's up, I signed, I connect just fine...)

schmanu:
> I got this alert/prompt when I tried to connect in Uniswap with Polygon which I don't think is intended
> 
> image
> 
> After accepting it (I wrote "Continue" but idk what to write because it doesn't say) didn't tirgger anymore, the only way to trigger it is by clearing the LS and cache of the safe app and trying again
> 
> Go to Uniswap, set it to ethereum there Open the WC modal and copy the hash Go to the safe app in a safe in Polygon Paste the code in the WC modal to connect The popUp should show up

@francovenica 
I cannot reproduce this. Tried it multiple times for me the following happens, which is also expected and correct:
1. I open Uniswap on Ethereum Mainnet
2. I click on connect wallet via Wallet Connect and copy the hash
3. I go to our Safe Wallet interface and open a Safe on Polygon
4. Now I copy the hash into the WC Modal in that Safe
5. The Safe connects, no pop up shows and uniswap switches to Polygon

schmanu:
> I'm having different behaviors between current prod, this PR and Uniswap and I don't know if they are intended.
> 
> In prod, you can easily change from one network to the other no problem, but if you try in this PR the Uniswap app will tell you that you wallet doesn't support the networks (tried also in prod CGW since it has all networks but with same result.
> 
> Although it might be considered an issue on Uniswap side, is weird that the behavior changes depending if we use the PR or current Prod In prod: 

This is expected and this is actually what this fix does:
- In current Prod we pretend that the Safe exists on all networks, but in fact it only exists on Mainnet. Switching to a network where the Safe does not exist should not be offered by Uniswap

schmanu:
> I tried to connect to OpenSea several times. In the safe app you can see the connection (in the top bar) but in openSea, or the WC modal with the QR code doesn't even close or the WC option in the modal where you select the wallet shows a infinite loading icon. I never got the message that OpenSea makes you sign to connect in the safe
> 
> I'm not don't know if this an issue from OpenSea side, but in current prod the connection works fine (the message pop's up, I signed, I connect just fine...)

I also can not reproduce this:
I tried different Safes now on Mainnet, Optimism and Polygon and connection / signing the login message always worked without any issues :/

francovenica:
Well, I tried to reproduce again the issues I reported, but they are working fine now

I cannot make that prompt to show up anymore
OpenSea is connecting fine now in different networks

I already tried Curve and Safe-as-owner and those work fine as well
Checked that chaning network in the safe app disconnects the current connected dapp in the WC topbar

LGTM

PR #3727, opened 2024-05-16
POC: Add web3modal

### LINKED ISSUE #3701: Migrate to AppKit from WalletConnect
## What is the feature about

Based on the research from #2953, we took the decision to migrate from web3-onboard to [AppKit](https://walletconnect.com/appkit).

## The list of requirements

* Remove onboard and add AppKit
* Refactor all the code that references onboard

usame-algan:
## What it solves

Resolves #3701

- Currently only activated on Sepolia and Mainnet until we can get static chain configs
- No balances visible yet
- Wallet tracking most likely not working yet
- Switch wallet not working
- Social Login not working
- No Ledger or Trezor support

## How this PR fixes it

- Replaces all instances of `useOnboard` with new hooks
- Adjusts function signatures to expect a `Eip1193Provider` instead of a custom `onboard` object

## How to test it

WIP

## Checklist


.env.example
```diff
@@ -45,3 +45,4 @@ NEXT_PUBLIC_REDEFINE_API=
 NEXT_PUBLIC_SOCIAL_WALLET_OPTIONS_STAGING=
 NEXT_PUBLIC_SOCIAL_WALLET_OPTIONS_PRODUCTION=
 
+NEXT_PUBLIC_WEB3MODAL_PROJECT_ID=
```
katspaugh, line 48: Looks like this would be a required env var. We should mention it in the readme.

katspaugh (commented):
Maybe we should have an abstraction layer between our code and a connector library. Otherwise we'd have to edit 48 files every time we switch (which admittedly doesn't happen often).

usame-algan:
> Maybe we should have an abstraction layer between our code and a connector library. Otherwise we'd have to edit 25 files (counting all the web3modal imports) every time we switch (which admittedly doesn't happen often).

I could imagine that we can adjust our `WalletProvider` to expose this via the `useWallet` hook i.e.

`const { wallet, connectWallet, disconnectWallet, switchNetwork, ...  } = useWallet()`

usame-algan:
Closing this for now. We can use this as a basis once there is Ledger support.

PR #3728, opened 2024-05-17
Feat: allow custom recovery delay

### LINKED ISSUE #3489: Recovery setup: add custom delay
## What is the feature about

Enhancing the Recovery setup

## The list of requirements

Following a relatively frequent user request, we are adding the custom option for the recovery delay. It will be accessible from the dropdown and counted in days.

## Designs/sketches

https://www.figma.com/file/pO43pjVwR34Uy1ihMsY3OM/Recovery-Hub?type=design&node-id=2143-33660&mode=design&t=NhpKUSBw2yoDKGoP-0

## Links

Added to notion spec:
https://www.notion.so/safe-global/Recovery-v1-feature-spec-d873c33a27c24cfabc91c85dbcf3ea35?pvs=4#a921767010c84523b07acd71d5eb9553

Related Slack thread: https://5afe.slack.com/archives/C03DAGWJCR5/p1710940492213439

jmealy:
## What it solves

Resolves #3489

## How this PR fixes it
- Add an option in the recovery period dropdown for a custom delay.
- Open a number input on selection

## How to test it
- Go through the steps for setting up account recovery
- Make sure the next button is disabled when no recovery delay, or an invalid recovery delay, is provided.
- Make sure the custom delay shows up correctly in the transaction review step and in account settings after executing the tx.
- Make sure the correct delay is used when executing recovery from the second account.

## Screenshots
image

## Checklist


schmanu (approved):
🚀

francovenica:
Blocked:
This ticket has an issue that it was fixed a while back and has to be updated in every PR about some infura issues. so I cannot sign/execute tx, so I cannot create a recovery
This is a thread when that happened for context
https://5afe.slack.com/archives/C069MCS78QL/p1716306882188049

Issue:
1 - The "Next" button is grayed out unless you pick a custom value. Basically the user cannot pick any other option that is not a "Custom" value
image

Also that is not part of this ticket but it could be fix here if possible:
The text above the input is innacurate. It says that "In this period of time you can cancel any recovery proposal..." which is technically correct, but the actual use of that timer is to let the user know when the recovery proposal is available for execution, and none of that is mentioned in that text. 
My suggestion is to add:
"The recovery proposal will be available for execution after this period of time." and then the "You can cancel any...."

jmealy:
I updated from dev and it appears to be working now. The greyed out button problem is not happening for me. 
I also updated the text to include your suggestion.

Also fixed a typo in the tooltip text:
`A period that begins after a recovery submitted on-chain, during which..` -> `A period that begins after a recovery *is* submitted on-chain, during which`

PR #3729, opened 2024-05-17
Fix: memoize Safe List data

katspaugh:
## What it solves

Balances and pending actions in the Safe List are now cached until the new data arrives or the page is reloaded, so that there's less jumping every time you open the sidebar.

## How to test

* Connect a wallet with many Safes
* Open the Sidebar
* Disconnect or switch to another wallet
* The list should be updated accordingly
* Close the sidebar and open it again
* Balances and pending actions should be visible immediately


jmealy (approved):
Looks good 👍  I don't have enough safes to see a significant different but it does seem to load more smoothly

francovenica:
LGTM

Note: this PR has to be updated because of the recent infura issues

PR #3742, opened 2024-05-24
Fix: back button style

### LINKED ISSUE #3717: Replace primary button with the secondary one on the confirmation screen
## Bug description
`Back` button is one of the main CTAs on this page, but it's a secondary action, and the button should be consistent for this use case.

## Environment
 - Browser: Any
 - Wallet: Any
 - Chain: Any

## Steps to reproduce
 1. Create a New transaction
 2. Fill in the address and token details, or contract details, etc.
 3. See confirmation screen

## Expected result
The `Back` button should use the same style as everywhere else in the UI (see [create page](https://app.safe.global/new-safe/create))

## Screenshots
Screenshot 2024-05-15 at 17 33 45

katspaugh:
## What it solves

Resolves #3717

## How this PR fixes it

Make the Back button in the tx modal look consistent with the same button in Safe Creation.

## How to test it

* Create a new tx
* Proceed to the Sign step

Test both on mobile and desktop.

Screenshot 2024-05-24 at 12 06 00


jmealy (approved):

katspaugh:
Good catch! I've restored the "text" style on small screens.

Screenshot 2024-05-29 at 11 56 13

Issue #3743, opened 2024-05-24
Display a Switch Network button if wallet is connected to a different chain


usame-algan:
## What is the feature about

The connected wallet needs to be on the same network as the safe in order for the user to sign/execute transactions. The current implementation to ensure this consists of multiple parts:

- We display a `WrongChainWarning` in some transaction modals
- We call `assertWalletChain` in some `dispatch` functions

This approach is not consistent e.g. we don't call `assertWalletChain` when creating a safe apps transactions and its easy to miss when adding a new feature. Another issue is that we combine switching the chain and executing/signing a transaction. While this "saves" the user a click, it can also be confusing if one click leads to the wallet popping up twice.

## The list of requirements

Ideally, we would have this in one place like `CheckWallet` and don't show a warning at all but instead show a different button than the action button to prompt the user to switch their network.

## Designs/sketches
<img width="732" alt="Screenshot 2024-05-24 at 13 03 23" src="https://github.com/safe-global/safe-wallet-web/assets/5880855/efd6cc56-5306-41f5-a09d-d1d33cb12fef">

PR #3745, opened 2024-05-24
Refactor: Replace onboard with Eip1193Provider in dispatch calls

### LINKED ISSUE #3744: Use Eip1193Provider instead of onboard for dispatch calls
## Description

Transaction dispatch calls are tightly coupled with onboard even though they are low-level functions and only need a standard `Eip1193Provider`. They also handle chain switching in case the wallet is on a different chain than the safe. This will be tackled as part of #3743

The goal is to not rely on our custom `useOnboard` state anymore but instead either switch to a different wallet solution or start using pre-defined hooks that onboard provides.

## Requirements

- All `dispatch` functions should expect a `Eip1193Provider` instead of an `onboard` object
- Extract the `assertWalletChain` calls to switch the chain higher up the call tree to the components calling the dispatch functions

usame-algan:
## What it solves

Resolves #3744 

This ports part of the changes over from #3727 

## How this PR fixes it

- Changes all `dispatch` functions to receive `wallet.provider` instead of `onboard`
- Some `dispatch` functions also receive `wallet.address` now which is needed for `TxEvents`
- Moves `assertWalletChain` calls to the component calling the dispatch functions. This is a temporary solution until #3743.

## How to test it

1. All transaction flows and message flows need to be tested
2. Gas estimation should still work
3. Being prompted to switch the chain when transacting should still work

## Checklist


tsconfig.json
usame-algan, line 1: This was automatically changed when running `lint --fix` and affected a few files that I had to adjust.

katspaugh (approved):

src/services/tx/tx-sender/sdk.ts
```diff
@@ -93,6 +93,7 @@ export const switchWalletChain = async (onboard: OnboardAPI, chainId: string): P
 
 export const assertWalletChain = async (onboard: OnboardAPI, chainId: string): Promise<ConnectedWallet> => {
   const wallet = getConnectedWallet(onboard.state.get().wallets)
+  console.log(wallet)
```
schmanu, line 96:
```suggestion
```

src/services/tx/tx-sender/sdk.ts
```diff
@@ -126,8 +123,9 @@ export const getAssertedChainSigner = async (
  * most of the values of transactionResponse which is needed when
  * dealing with smart-contract wallet owners
  */
-export const getUncheckedSafeSDK = async (onboard: OnboardAPI, chainId: SafeInfo['chainId']): Promise<Safe> => {
-  const signer = await getAssertedChainSigner(onboard, chainId)
+export const getUncheckedSafeSDK = async (provider: Eip1193Provider): Promise<Safe> => {
+  const browserProvider = createWeb3(provider)
+  const signer = await browserProvider.getSigner()
```
schmanu, line 128: If we update to the latest protocol-kit it actually expects EIP 1193 Providers and removes all adapters.
Should we use the opportunity to do this in this PR?

https://docs.safe.global/sdk/protocol-kit/reference/migrate-to-v4

usame-algan: Thats really nice! Doesn't look like too big of a change but just to be safe I would do the migration in a follow-up PR after this is merged.

liliya-soroka:
Switch network doesn't work for rabby wallet. On prod we are showing the error with information about wrong chain in the wallet. In the PR there is no such error

usame-algan:
> Switch network doesn't work for rabby wallet. On prod we are showing the error with information about wrong chain in the wallet. In the PR there is no such error. @usame-algan

Thanks for reporting! I can reproduce that the error notification is not visible anymore. I found a way however to fix this error for adding networks in Rabby so I will create a separate PR to dev instead of only fixing the notification.

PR #3747, opened 2024-05-25
Feat: site to export Google private key

katspaugh:
## What it solves

This branch renders the Social Login settings page on the index route and doesn't allow doing anything else.

## How to test it

* Go to https://google--walletweb.review.5afe.dev
* Log in with google
* Export your PK

> Ignore the failing tests, this PR won't be merged anyway, it's just for preview.


schmanu (changes requested):
After logging in you are requested to "Swap your signer" but there is no way to choose an owned Safe.
Screenshot 2024-05-27 at 09 46 20

I also cannot export the pk without having a Password setup. We should simply always allow it so it is a one click action.

PR #3748, opened 2024-05-27
Fix: missing feature toggles

### LINKED ISSUE #3746: ERC721, Safe Apps and Contract Interaction feature flags aren't working
Turning off these features doesn't hide the corresponding features in the UI.

katspaugh:
## What it solves

Resolves #3746

## How this PR fixes it

* I've added a few missing feature toggle checks.
* CONTRACT_INTERACTION is deprecated and removed

## How to test

Can be tested on Sepolia by turning off the following features (don't forget to turn them back on):

* ERC721
* SAFE_APPS
* PUSH_NOTIFICATION


src/components/balances/AssetsHeader/index.tsx
```diff
@@ -5,15 +5,20 @@ import PageHeader from '@/components/common/PageHeader'
 import { balancesNavItems } from '@/components/sidebar/SidebarNavigation/config'
 
 import css from '@/components/common/PageHeader/styles.module.css'
+import { useHasFeature } from '@/hooks/useChains'
+import { FEATURES } from '@/utils/chains'
 
 const AssetsHeader = ({ children }: { children?: ReactNode }): ReactElement => {
+  const isNftsEnabled = useHasFeature(FEATURES.EIP1271)
```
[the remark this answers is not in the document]
katspaugh: Oops

src/components/dashboard/index.tsx
```diff
@@ -14,14 +14,16 @@ import { CREATION_MODAL_QUERY_PARM } from '../new-safe/create/logic'
...
+import { FEATURES } from '@/utils/chains'
 const RecoveryHeader = dynamic(() => import('@/features/recovery/components/RecoveryHeader'))
 const RecoveryWidget = dynamic(() => import('@/features/recovery/components/RecoveryWidget'))
 
 const Dashboard = (): ReactElement => {
   const router = useRouter()
   const { safe } = useSafeInfo()
   const { [CREATION_MODAL_QUERY_PARM]: showCreationModal = '' } = router.query
-
+  const showSafeApps = useHasFeature(FEATURES.SAFE_APPS)
```
usame-algan, line 26: I suggest we also hide the navigation item in the sidebar. There should be a helper `isRouteEnabled`.

usame-algan (approved):
We need to make sure to enable these feature toggles before pushing to prod since they weren't in use so far.

src/components/dashboard/index.tsx
katspaugh: That's already there.

francovenica:
Issues/questions:

The "Push notifications" toggle, works fine, but the "Notification" tab is still present in the settings, so clicking on it shows an empty page. I assume the notification tab should also dissapear if the push notification is disabled

The "Contract interaction" toggle also removes the Tx builder from the dashboard, but is still visible in the Apps sections and can be used.
The question is: should the "Contract interaction" toggle still show the tx builder in the dashboard? or should it hide it along with the Tx builder app in the "Apps" page?

katspaugh:
Good catch with the settings tab, I'll hide it too.

Regarding the CONTRACT_INTERACTION flag, I've just realized it's not even necessary. We can just toggle Transaction Builder on and off, and it will automatically hide all links to it, including that button. So I'll deprecate CONTRACT_INTERACTION altogether.

PR #3751, opened 2024-05-27
Fix: rearrange Settings tabs

katspaugh:
## What it solves

* Security & Login -> Security
* Moved spending limits to Setup
* Setup items are more compact now

Screenshot 2024-05-27 at 10 13 42

Screenshot 2024-05-27 at 09 56 10


schmanu (changes requested):
Looks good.

There is a lint error and a test is failing tho.

usame-algan:
Do we also need to adjust e2e tests for spending limits not that it was moved to a different page?

schmanu (approved):

katspaugh:
I agree regarding a separate tab for limits. There's a bigger project in the works to revamp the settings, so it will be addressed there.

PR #3753, opened 2024-05-27
Chore: remove social login

### LINKED ISSUE #3705: Clean up seedless feature from the code
The feature has been deprecated but the code is still there.

katspaugh:
## What it solves

Resolves #3705

The feature has been discontinued, so we're removing the unused code.

## How to test it

* Connect a wallet
* Switch to another wallet w/o disconnecting the previous one
* Try executing/relaying


package.json
```diff
@@ -75,7 +74,6 @@
     "@web3-onboard/ledger": "2.3.2",
     "@web3-onboard/trezor": "^2.4.2",
     "@web3-onboard/walletconnect": "^2.5.4",
-    "@web3auth/mpc-core-kit": "^1.1.3",
     "blo": "^1.1.1",
     "bn.js": "^5.2.1",
```
schmanu, line 78: This dependency was also only used by social login.

schmanu (changes requested):

package.json
katspaugh: Removed. Looks like quite a few other package still depend on it.

schmanu (approved):

francovenica:
LGTM.

I see conflicts so I won't pass it. Let me know if after solving those I should re check

PR #3754, opened 2024-05-27
Chore: remove batch reordering

katspaugh:
## What it solves

Very few people actually reorder batched transactions, so no need to keep this heavy dependency on framer-motion.

Last 90 days:
Screenshot 2024-05-27 at 16 32 28

Screenshot 2024-05-27 at 16 31 11


usame-algan (approved):
Looks good 👍

katspaugh:
I'll merge this directly, since it's just a removal and I tested it.

PR #3756, opened 2024-05-28
feat: Execute transaction through role 

jfschwarz:
## What it solves

With [SEP-14](https://forum.safe.global/t/sep-14-obra-role-based-access-control-pilot-browser-extension-gnosis-guild/4676) the community voted to implement role-based access control in Safe{Wallet} based on the Zodiac Roles Modifier.

## How this PR fixes it

This PR allows role members to use Safe{Wallet} to execute transaction from the Safe using their role. For this purpose, a new section has been added to the transaction sign/execute form, which only shows up if the connected wallet is a member of any role enabled on the Safe.

## How to test it

- Deploy a Roles Modifier through the Zodiac Safe app
- Set up some permissions for a testing role, e.g. using [permissions-starter-kit](https://github.com/gnosisguild/permissions-starter-kit), apply and assign the role to an EOA using the mod's `assignRoles` function

Alternatively, send me your EOA address so I can assign a role in my [Sepolia test setup](https://app.safe.global/apps/open?safe=sep:0xB9289969bd138b802cfD48466e0cCc8aCD528Ba1) to it.

## Screenshots

A new card appears in the transaction sign/execute form only if the Safe has a Roles mod enabled and the connected wallet account is a role member.

##### Connected with role member wallet, permission checks pass:
image

##### Connected with role member wallet, permission checks fail:
image

(Status can be any value from the Roles mods' [Status enum](https://github.com/gnosisguild/zodiac-modifier-roles/blob/00b68f361e9bfbbd2cd3d527c084f39e6fd85606/packages/evm/contracts/PermissionChecker.sol#L727))

## Checklist

### DIFF
--- a/src/components/tx/SignOrExecuteForm/PermissionsCheck.tsx
+++ b/src/components/tx/SignOrExecuteForm/PermissionsCheck.tsx
@@ -1,21 +1,31 @@
 import { useContext, useState } from 'react'
+import { type ChainId, chains, fetchRolesMod } from 'zodiac-roles-deployments'
+
 import { SafeTxContext } from '@/components/tx-flow/SafeTxProvider'
 import { Box, Button, CardActions, CircularProgress, Divider, Typography } from '@mui/material'
 import commonCss from '@/components/tx-flow/common/styles.module.css'
-
 import CheckWallet from '@/components/common/CheckWallet'
-import WalletRejectionError from './WalletRejectionError'
 import TxCard from '@/components/tx-flow/common/TxCard'
-import ErrorMessage from '../ErrorMessage'
 import { getTransactionTrackingType } from '@/services/analytics/tx-tracking'
 import { TX_EVENTS } from '@/services/analytics/events/transactions'
 import { trackEvent } from '@/services/analytics'
 import useChainId from '@/hooks/useChainId'
+import useSafeInfo from '@/hooks/useSafeInfo'
+import useWallet from '@/hooks/wallets/useWallet'
+import useAsync from '@/hooks/useAsync'
+import WalletRejectionError from './WalletRejectionError'
+import ErrorMessage from '../ErrorMessage'
 
 const PermissionsCheck: React.FC<{}> = ({}) => {
   const chainId = useChainId()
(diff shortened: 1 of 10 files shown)


jfschwarz:

src/components/tx/SignOrExecuteForm/PermissionsCheck.tsx
```diff
@@ -0,0 +1,290 @@
...
+    <TxCard>
+      <Typography variant="h5">Execute through role</Typography>
+
+      {allowingRole && (
+        <Typography>
+          As a member of the <Role>{allowingRole.roleKey}</Role> role you can execute this transaction immediately.
+        </Typography>
+      )}
+
+      {!allowingRole && (
```
schmanu, line 82: What is the use-case behind displaying the role if the user has a not allowing role?

I'd prefer adding a new section "Roles" in the settings where we list all roles with their permissions and only showing the Card in the modal if the user is allowed to use it.

jfschwarz: Appreciate the feedback. Our thinking was: If a roles member is connected to the Safe web app it's pretty likely they are trying to execute something through the role–especially if the connected wallet is not even an owner on the Safe. For connected wallets that are both, a Safe owner as well as a role member, we could indeed consider hiding the "Execute through role" card if it doesn't pass role permissions checks, even though I'd prefer to still show it for improved transparency.

It's pretty common that permissions become outdated/have not been designed with all cases in mind. In that case it's helpful if the user can see the permission check status to request an update to the role permissions.

____
> I'd prefer adding a new section "Roles" in the settings where we list all roles with their permissions and only showing the Card in the modal if the user is allowed to use it.

We definitely also plan to add a "Roles" section in settings as a next iteration. We gonna share some mock-ups for that soon to kick off the discussion.

schmanu: @jfschwarz 
Thanks for the detailed explanation on the motivation of this. That makes a lot of sense!

jfschwarz:
superseded by #3768

PR #3760, opened 2024-05-28
Feat: display contract names for WalletConnect txs

katspaugh:
## What it solves

For transactions made via WalletConnect, show the contract name instead of just WalletConnect in the tx summary.

## How this PR fixes it

I've changed the `appUrl` or these txs from a hard-coded WalletConnect Safe App URL to the original dapp URL.
Since our backend doesn't know anything about arbitrary dapp origins, it will not be able to annotate these txs as WalletConnect txs anymore, so we'll display a generic contract name instead. If the contract isn't verified, it will just say Contract interaction as usual.

## How to test it

* Make txs via WalletConnect
* Check how they are displayed in the queue/history

### Before
Screenshot 2024-05-28 at 13 39 33

### After
Screenshot 2024-05-28 at 14 12 26


src/services/analytics/tx-tracking.ts
```diff
@@ -63,12 +62,14 @@ export const getTransactionTrackingType = async (chainId: string, txId: string):
...
     if (details.safeAppInfo) {
-      return isWalletConnectSafeApp(details.safeAppInfo.url) ? TX_TYPES.walletconnect : details.safeAppInfo.url
+      return details.safeAppInfo.url
     }
 
     if (isMultiSendTxInfo(txInfo)) {
       return TX_TYPES.batch
     }
+
+    return TX_TYPES.walletconnect
```
usame-algan, line 72: Is it safe to assume wc here as the fallback? Previously it would fallback to custom which is more generic.

src/features/walletconnect/components/WalletConnectProvider/index.tsx
```diff
@@ -90,9 +88,9 @@ export const WalletConnectProvider = ({ children }: { children: ReactNode }) =>
 
         // Get response from Safe Wallet Provider
         return safeWalletProvider.request(event.id, event.params.request, {
-          id: WalletConnectSafeApp.id,
-          url: WalletConnectSafeApp.url,
-          name: getPeerName(session.peer) || 'Unknown dApp',
+          id: -1,
+          url: session.peer.metadata.url,
+          name: getPeerName(session.peer) || 'WalletConnect',
```
usame-algan, line 93: Afaiu we are sending this through WC and not proposing it with the transaction to our backend. How does this affect what we display in the tx history?

src/services/analytics/tx-tracking.ts
katspaugh: Not ideal but I think it's OK to assume that if it's not an app, it must be WalletConnect. I don't think there's any other way to sign or execute a tx from our UI.

src/features/walletconnect/components/WalletConnectProvider/index.tsx
katspaugh: No, this is what we pass to our tx modal. It will use the icon URL and name in the title of the modal. Since the URL won't be a known Safe App (unless the main dapp and a safe app share the same URL), CGW will just ignore it.

francovenica:
I think these are ok when the tx name are generic like "Contract interaction" or "WalletConnect" , but if the name is the specific dapp I'd leave that one:
For example in this one, I think the "CowSwap" name says more than Wrapped ETH
image
image

The safe these were taken: sep:0x8f4A19C85b39032A37f7a6dCc65234f966F72551

katspaugh:
I've also adjusted the Messages to display the EIP712 domain name and primaryType if available.

Screenshot 2024-06-04 at 10 32 44

schmanu (approved):
Nice addition to show the Domains for WalletConnect messages!

francovenica:
@katspaugh Check in both snapshots the TxHash, is the same, so it is the same transaction

PR #3758, opened 2024-05-28
Tests: add component tests for tx history filter

mike10ca:
## What it solves

Resolves #3759 

## How this PR fixes it

- Add component tests for TxFilterForm

## How to test it

- Run tests and observe outcome

PR #3739, opened 2024-05-24
Chore: update release action

katspaugh:
## What it solves

The action `actions/upload-release-asset@v1` stopped working and it's not maintained, so I replaced it with `Shopify/upload-to-release@v1`. See https://github.com/marketplace/actions/upload-to-release-action

PR #3766, opened 2024-05-29
Release 1.37.1

katspaugh:
## Features
* feat: add new label to swaps button (#3766)

## Analytics
* Differentiate bulk executions from batches (#3779)
* fix: Safe stats tracking for owned Safes (#3708)
* fix: counterfactual safe tracking (#3783)

## Bug fixes
* **WalletConnect**
  - fix: (WC) do not return Safe as available on all optional Namespaces (#3725) 
  - Fix: Show loading state in WC input between pairing and receiving proposal (#3712)

* Fix: back button style (#3742)
* Fix: missing feature toggles (#3748)
* fix: do not show nonce conflict warning for batch transactions. (#3608)
* Fix: rearrange Settings tabs (#3751)
* Fix: memoize Safe List data (#3729)
* Fix: move rhino.fi to warned bridges (#3740)
* fix: Hide signed transaction from untrusted pending queue (#3710)
* fix: Add network to Rabby if unknown (#3776)

## Chores & tests
* Chore: remove social login (#3753)
* Chore: remove batch reordering (#3754)
* Chore: update release action (#3739)
* Tests: add component tests for tx history filter (#3758)
* Chore: use separate Infura tokens for dev (#3773)
* Chore: fix branch conditionals (#3782)


francovenica:
Question regarding the infura token for dev.
Currently in the PR where this was developed initially, the infura URL you can see in the settings is, as expected, different than the one in current prod
The thing is that after merging that to the release, the infura URL in the settings is the same as the one in current prod. Could this be because the release branch is tagged as prod? or should this release also have the same infura URL in the settings as the original PR and we have an issue?

francovenica:
I tried to check that las commit regarding the change in how tx in bulk is tracked in analytics, but I don't see any GTM calls all of the sudden. Those were there before since I even reported an issue about them in the board.
Any recent changes that might have disable/hide them from showing up in the dev console?

katspaugh:
@francovenica yes, exactly, the RC branches will use all the production settings. CGW will also be the prod version.

francovenica:
It seems that the GTM calls are not being shown anymore because now the release branch is "like prod".

I went back to the PR's that did the changes in the GTM tracks 3779 and 3783 and test those fixed there and they work fine.

francovenica:
The rest of the regression went fine, and the last commits were checked and they worked fine. 

There is an ongoing discussion if we should keep the "use different infura for dev" change, so if it is decied to roll back that change let me know

If not, then the RC looks good for release

PR #3768, opened 2024-05-29
feat: Execute transaction through role 

jfschwarz:
## What it solves

With [SEP-14](https://forum.safe.global/t/sep-14-obra-role-based-access-control-pilot-browser-extension-gnosis-guild/4676) the community voted to implement role-based access control in Safe{Wallet} based on the Zodiac Roles Modifier.

## How this PR fixes it

This PR allows role members to use Safe{Wallet} to execute transaction from the Safe using their role. For this purpose, a new section has been added to the transaction sign/execute form, which only shows up if the connected wallet is a member of any role enabled on the Safe.

## How to test it

- Deploy a Roles Modifier through the Zodiac Safe app
- Set up some permissions for a testing role, e.g. using [permissions-starter-kit](https://github.com/gnosisguild/permissions-starter-kit), apply and assign the role to an EOA using the mod's `assignRoles` function

Alternatively, send me your EOA address so I can assign a role in my [Sepolia test setup](https://app.safe.global/apps/open?safe=sep:0xB9289969bd138b802cfD48466e0cCc8aCD528Ba1) to it.

## Screenshots

A new card appears in the transaction sign/execute form only if the Safe has a Roles mod enabled and the connected wallet account is a role member.

##### Connected with role member wallet, permission checks pass:
image

##### Connected with role member wallet, permission checks fail:
image

(Status can be any value from the Roles mods' [Status enum](https://github.com/gnosisguild/zodiac-modifier-roles/blob/00b68f361e9bfbbd2cd3d527c084f39e6fd85606/packages/evm/contracts/PermissionChecker.sol#L727))

## Checklist


TanyaEfremova:
Hey @jfschwarz, great addition! A few suggestions to improve the texts:

**Execute without confirmations**
As a member of the `eth_wrapping` role you can execute this transaction immediately without confirmations from other owners.
Button label: Execute

src/components/tx/SignOrExecuteForm/PermissionsCheck.tsx
```diff
@@ -0,0 +1,473 @@
...
+      } else {
+        trackError(Errors._815, err)
+        setSubmitError(err)
+      }
+      setIsPending(false)
+      return
+    }
+
+    // On success, wait for module tx to be indexed
+    const transactionService = currentChain?.transactionService
```
katspaugh, line 126: We don't use the transaction service directly. We use the Client Gateway, which has the [module txs endpoint](https://github.com/safe-global/safe-gateway-typescript-sdk/blob/main/src/index.ts#L103) as well.

src/components/tx/SignOrExecuteForm/PermissionsCheck.tsx
katspaugh, line 1: There's a lot happening in this file, do you think it would make sense to create a folder `PermissionsCheck` and extract some of the hooks and helper functions into separate files?

src/components/tx/SignOrExecuteForm/PermissionsCheck.tsx
jfschwarz: I checked the client gateway, but it doesn't seem to expose the required `transaction_hash` query param: https://github.com/safe-global/safe-gateway-typescript-sdk/blob/main/src/types/api.ts#L151

jfschwarz: Or is it just a matter of updating the types in safe-gateway-typescript-sdk?

katspaugh (approved):
Thank you @jfschwarz! Excellent work!

PR #3769, opened 2024-05-29
Fix: improve chain switch redirect

katspaugh:
## What it solves

After we removed the automatic redirect from `/welcome` to `/welcome/accounts` if wallet is connected, the Network Selector started requiring an extra click to see the list of your Safes. This PR fixes that.

The Network Select will now take you straight to `/welcome/accounts` if wallet is connected.

### DIFF
--- a/src/components/common/NetworkSelector/index.tsx
+++ b/src/components/common/NetworkSelector/index.tsx
@@ -15,6 +15,7 @@ import { type ReactElement, useMemo } from 'react'
 import { useCallback } from 'react'
 import { AppRoutes } from '@/config/routes'
 import { trackEvent, OVERVIEW_EVENTS } from '@/services/analytics'
+import useWallet from '@/hooks/wallets/useWallet'
 
 const NetworkSelector = (props: { onChainSelect?: () => void }): ReactElement => {
   const isDarkMode = useDarkMode()
@@ -22,6 +23,7 @@ const NetworkSelector = (props: { onChainSelect?: () => void }): ReactElement =>
   const { configs } = useChains()
   const chainId = useChainId()
   const router = useRouter()
+  const isWalletConnected = !!useWallet()
 
   const [testNets, prodNets] = useMemo(() => partition(configs, (config) => config.isTestnet), [configs])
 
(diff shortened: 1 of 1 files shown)


src/components/common/NetworkSelector/index.tsx
```diff
@@ -30,7 +32,11 @@ const NetworkSelector = (props: { onChainSelect?: () => void }): ReactElement =>
       const shouldKeepPath = !router.query.safe
 
       const route = {
-        pathname: shouldKeepPath ? router.pathname : AppRoutes.index,
+        pathname: shouldKeepPath
+          ? router.pathname
+          : isWalletConnected
+          ? AppRoutes.welcome.accounts
+          : AppRoutes.welcome.index,
```
usame-algan, line 39: I am not sure if this works because on non-safe routes, `shouldKeepPath` is always true so there won't be a redirect even if the wallet is connected. What is the behaviour we are trying to achieve?

katspaugh: It keeps the path only for non-safe specific routes, like the welcome page or global settings.

katspaugh: The behavior I want to achieve: you're connected, and you're viewing a Safe on Arbitrum. You switch the chain in the NetworkSelector. I want it to go to the account list. Before it would go to the Welcome page.

usame-algan: Ok, that makes sense! I was confused because thats already the behaviour for me but realized it happens because of the index page. So with this PR we now support both connected and unconnected wallets.

- If connected and they switch the network from a safe we redirect them to the accounts page
- If they are not connected and switch the network we redirect them to the index page
-> There we check if they have added safes or not and redirect them either to the accounts page or to the welcome page

usame-algan (approved):
🚀

src/components/common/NetworkSelector/index.tsx
katspaugh: Almost. If they are not connected, we're redirecting straight to /welcome.

francovenica:
Had to read the conversation in the review to understand what it was the purpose of the ticket.

LGTM
Being in a safe of NetworkA and switching to networkB in the top right corner takes the user to the safe selection (/welcome/accounts) instead of the welcome page (/welcome).

It seems this only happens if the user is connected. I think diconnected users also can be taken to the /welcome/accounts page since the user might have safes in his watchlist, but is not a blocker

PR #3771, opened 2024-05-29
feat: add promo widget for swaps [SWAP-87]

compojoom:
## What it solves
Displays a swaps promo block(that can be hidden) on the dashboard. 
Right now we are loading a .png, but @TanyaEfremova would give us an svg.

Resolves #

## How this PR fixes it

- Adds new Swap promo widget on the dashboard
- Adjusts the `Chip` style to match Figma design

Other changes:
- Adjusts the Queue counter style to match Figma design
- Adjusts the sidebar badges to match Figma design

## How to test it

1. Open the dashboard and observe the new widget
2. Press "Try it now" and be navigated to the swaps page
3. Press "Don't show again", the widget should permanently disappear

## Screenshots

Screenshot 2024-05-31 at 10 52 43
Screenshot 2024-05-31 at 10 52 58
Screenshot 2024-05-31 at 10 53 15

## Checklist


jmealy (approved):
Looks good! 
just one small thing. It looks a bit odd next to the safe pass widget, since they are very similar but the buttons are way bigger for the safe pass one. Do we plan to show both of these together or is it one promotion at a time?

image

usame-algan:
> Looks good! 
> 
> just one small thing. It looks a bit odd next to the safe pass widget, since they are very similar but the buttons are way bigger for the safe pass one. Do we plan to show both of these together or is it one promotion at a time?
> 
> 
> 
> image
> 
> 

Noticed this too. AFAIK we will show both at the same time. @TanyaEfremova do you think we should adjust either one of them to make it look more consistent?

francovenica:
The widget should be network sensitive, so it should only show in ETH, GNO and Sepolia. 
Or if we want to show it everywhere for the sake of promotion, we should have "try it now" button grayed out with a tooltip saying that is only supported in those networks (But personally I'd prefer the widget to only show in the supported networks)

image

compojoom:
Good catch! Check again please!

TanyaEfremova:
Left margin is indeed different, let's align with the Safe{Pass} widget:` 40px` on the left side.
Buttons should use style `Large` in both widgets.

katspaugh:
As a side note, we should avoid using margins/paddings that are not a multiple of 8 px.

francovenica:
LGTM

PR #3772, opened 2024-05-30
Feat: visually group transactions executed in bulk in the transaction history

### LINKED ISSUE #3443: Visually group multisend-executed txs in the history
Currently, when you execute several Safe transacitons using the Bulk Execute button, they are displayed as separate transactions in the history.

However, they all share the same tx hash (because they were executed by the Multisend contract as one tx). And if you click on the Etherscan link in the tx details of these txs, they obviously lead to the same Etherscan page that shows them as a Multisend contract interaction.

Since this discrepancy can be confusing to users, we should probably somehow mark bulk-executed txs in the UI, perhaps visually group them like we do with conflicting nonce txs.

## Design
[Figma](https://www.figma.com/proto/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?page-id=5794%3A183021&type=design&node-id=10780-98066&viewport=-111%2C14953%2C0.88&t=N0YFEO05M4pwpkfQ-1&scaling=min-zoom&starting-point-node-id=10780%3A98066)

Screenshot 2024-04-26 at 08 07 07

jmealy:
## What it solves

Resolves #3443

## How this PR fixes it
- Transactions with the same transaction hash are displayed in a group, similar to conflicting transactions in the queue.

## How to test it
- Queue multiple transactions and then execute them in bulk. 
- After execution these transactions should be grouped together
- If the transactions are sending tokens to another safe, the receive transactions should also be grouped.

## Screenshots
image

## Checklist


src/components/transactions/TxList/index.tsx
```diff
@@ -1,29 +1,62 @@
...
+    const transactions = items.filter(isTransactionListItem)
+    return await getTxDetailsWithBackoff(transactions, chainId).then((txDetailsList) => {
+      return txDetailsList.reduce((accumulator: Record<string, string>, txDetails: TransactionDetails) => {
+        const { txId, txHash } = txDetails
+        if (!txId || !txHash) return accumulator
+        accumulator[txId] = txHash
+        return accumulator
+      }, {})
+    })
+  }, [chainId, items])
```
usame-algan, line 45: This is an impactful change as we would now fetch txDetails for every transaction in the history whenever the user opens the page and/or scrolls. If I understand correctly all thats missing is the `txHash` of each transaction?

usame-algan (changes requested):

src/components/transactions/TxList/index.tsx
jmealy: yeah it would be much better to get the tx hash /history if that's possible

iamacook: > As far as I can see the `/all-transactions` endpoint on the transaction service returns the `txHash`. @iamacook is it possible to also return it from the `/history` endpoint on the CGW?

It is and whilst it makes sense to return it in the summary, I'm not sure why it isn't. I'll bring it up for discussion next week.

On a separate note: you might want to consider ensuring the transaction has execution details before grouping under "bulk execution". I'm not certain but I think the current implementation may group transfers that occur in the same transaction.

jmealy: It's updated now to use the changes from the CGW. Cheer Aaron 👍

jmealy:
Aaron gave some feedback that it is confusing to have the txHash displayed in the `actions` column
@TanyaEfremova what do you think?

TanyaEfremova:
Sure, that does make sense!
How about we show the amount of transactions batched instead, and show a button separately? A couple ideas:
- [Idea 1](https://www.figma.com/design/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=11151-47993&t=gfp1I4vZwEzFOo7i-1)
- [Idea 2](https://www.figma.com/design/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=11151-49354&t=gfp1I4vZwEzFOo7i-1)

cc @katspaugh could you also provide feedback here please?

katspaugh:
@TanyaEfremova I like this version:
Screenshot 2024-06-07 at 16 40 00

It also has a much bigger clickable are on the explorer link, so it's better overall.

Showing a single transaction hash was an important detail, because that's the confusing part about these transactions. They all have the same tx hash, so the user might be going in circles when debugging transactions. So I would keep the hash where it was but also add this View on explorer button instead of the tiny arrow icon next to the hash.

TanyaEfremova:
Perfect, this exact layout can work well for grouping [swap transactions](https://www.figma.com/design/VyA38zUPbJ2zflzCIYR6Nu/Swap?node-id=6980-41655&t=UIJrfrXB5RpSqTIN-1) (with limitations) in hitory.

src/components/transactions/BulkTxListGroup/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <Paper className={css.container}>
+      <Box gridArea="icon">
+        <SvgIcon className={css.icon} component={BatchIcon} inheritViewBox fontSize="medium" />
+      </Box>
+      <Box gridArea="info">
+        <Typography noWrap>Transactions executed in bulk</Typography>
+      </Box>
+      <Box className={css.action}>{groupedListItems.length} Transactions</Box>
```
katspaugh, line 31:
```suggestion
      <Box className={css.action}>{groupedListItems.length} transactions</Box>
```

src/components/transactions/TxList/index.tsx
```diff
@@ -1,29 +1,41 @@
...
-import { groupConflictingTxs } from '@/utils/tx-list'
+import { groupTxs } from '@/utils/tx-list'
 import { Box } from '@mui/material'
-import type { TransactionListPage } from '@safe-global/safe-gateway-typescript-sdk'
+import type { Transaction, TransactionListPage } from '@safe-global/safe-gateway-typescript-sdk'
 import type { ReactElement, ReactNode } from 'react'
 import { useMemo } from 'react'
 import TxListItem from '../TxListItem'
 import css from './styles.module.css'
+import { uniq } from 'lodash'
```
katspaugh, line 9: I'm not sure if our build is able to tree-shake lodash properly, it's better to help it:

```suggestion
import uniq from 'lodash/uniq'
```

src/utils/tx-list.ts
```diff
@@ -7,6 +7,11 @@ import type { RecoveryQueueItem } from '@/features/recovery/services/recovery-st
 
 type GroupedTxs = Array<TransactionListItem | Transaction[]>
 
+export const groupTxs = (list: TransactionListItem[]) => {
```
katspaugh, line 10: It would be good to unit-test this fn.

katspaugh (commented):
Nice work!

katspaugh (approved):

usame-algan (approved):
Looks good 👍

TanyaEfremova:
The last thing, that I just noticed, but minor. The number of transactions displayed in the 2nd column should use `label-secondary` color style.

PR #3776, opened 2024-05-30
fix: Add network to Rabby if unknown

usame-algan:
## What it solves

Part of #3745

## How this PR fixes it

In Rabby, the error when trying to switch to a network that is not yet added is nested and doesn't pass our condition. It now checks for a nested `code` in the error that is thrown.

## How to test it

1. Remove Sepolia network from Rabby
2. Connect to a Sepolia Safe
3. Create / Execute a transaction or sign a message
4. Observe there is no error and instead Rabby pops up to add Sepolia as a new network

## Screenshots

Screenshot 2024-05-30 at 13 52 15

## Checklist

### DIFF
--- a/src/components/common/BuyCryptoButton/index.tsx
+++ b/src/components/common/BuyCryptoButton/index.tsx
@@ -21,7 +21,7 @@ const useOnrampAppUrl = (): string | undefined => {
 
 const useBuyCryptoHref = (): LinkProps['href'] | undefined => {
   const query = useSearchParams()
-  const safe = query.get('safe')
+  const safe = query?.get('safe')
   const appUrl = useOnrampAppUrl()
 
   return useMemo(() => {
--- a/src/services/tx/tx-sender/sdk.ts
+++ b/src/services/tx/tx-sender/sdk.ts
@@ -15,6 +15,7 @@ import { type OnboardAPI } from '@web3-onboard/core'
 import type { ConnectedWallet } from '@/hooks/wallets/useOnboard'
 import { asError } from '@/services/exceptions/utils'
 import { UncheckedJsonRpcSigner } from '@/utils/providers/UncheckedJsonRpcSigner'
+import get from 'lodash/get'
 
 export const getAndValidateSafeSDK = (): Safe => {
   const safeSDK = getSafeSDK()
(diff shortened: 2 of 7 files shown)


usame-algan:
The second commit is still the lint issues that happen to me whenever I try to push something. Would be nice if we could get this merged to dev soon because I've added these lint fixes probably 5 times now in different PRs 😪

src/components/common/BuyCryptoButton/index.tsx
```diff
@@ -21,7 +21,7 @@ const useOnrampAppUrl = (): string | undefined => {
 
 const useBuyCryptoHref = (): LinkProps['href'] | undefined => {
   const query = useSearchParams()
-  const safe = query.get('safe')
+  const safe = query?.get('safe')
```
katspaugh, line 24: You seem to fix these randomly in various PRs. Is it really required? Can we fix _all_ of them in one separate PR?

src/services/tx/tx-sender/sdk.ts
```diff
@@ -36,24 +37,29 @@ async function switchOrAddChain(walletProvider: ConnectedWallet['provider'], cha
       params: [{ chainId: hexChainId }],
     })
   } catch (error) {
-    if ((error as Error & { code: number }).code !== UNKNOWN_CHAIN_ERROR_CODE) {
-      throw error
+    const errorCode = get(error, 'code') as number | undefined
+
+    // Rabby emits the same error code as MM, but it is nested
+    const nestedErrorCode = get(error, 'data.originalError.code') as number | undefined
```
katspaugh, line 43: Is this better than `data?.originalError?.code`?

src/components/common/BuyCryptoButton/index.tsx
usame-algan: Whenever I run `git push` I get this and it causes the same lint errors. I am fixing all of them in each of my PRs and its always the same ones so once we have this merged to dev it should be resolved.

Screenshot 2024-05-30 at 14 54 15

katspaugh (approved):
Let's merge it directly then, it's a straightforward fix, no need to QA.

src/services/tx/tx-sender/sdk.ts
usame-algan: I think it is definitely more readable. As an alternative I tried to check for each key before accessing it i.e.`'data' in error && 'originalError' in (error.data as ...) && error.data.originalError` but it got messy really fast. We could ofc also go with `data?.originalError?.code` and ts-ignore it.

PR #3778, opened 2024-05-30
feat: Improve safe creation status screen

### LINKED ISSUE #3612: [Safe creation] Improve the status screen
## What is the feature about

The status screen when creating a safe has multiple UX issues as documented [here](https://www.notion.so/safe-global/Better-handling-of-safe-creation-051abb7ffe8948da8bd638017af96d97) and #1484, #1871, [Slack](https://5afe.slack.com/archives/C04MQSC3R7Y/p1710951036300699)

## The list of requirements

- [ ] Only move to the status screen once the user has confirmed the transaction in their wallet
- [ ] Rework the Retry and Cancel button
- [ ] Retry should navigate the user back to the review screen with the previous setup filled out
- [ ] Cancel should take the user back to the welcome page
- [ ] Remove the stepper and instead only show the a "Check Status" link for the block explorer
- [ ] Show a clear pending state in the UI and inform the user about the status of their transaction

## Designs/sketches

- [Figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?type=design&node-id=10822-4396&mode=design&t=WMKNizUHCrqhabci-0)

## Links

- [Notion](https://www.notion.so/safe-global/Better-handling-of-safe-creation-051abb7ffe8948da8bd638017af96d97)

usame-algan:
## What it solves

Resolves #3612

## How this PR fixes it

- Removes the timeout for safe creation transactions
- Only navigates to the status screen after the user submits the safe creation in their wallet
- Renames the Cancel button to `Back to homepage`
- Renames the Retry button to `Try again`
- Navigates the user to the safe creation review screen when they press `Try again`
- Adds a new screen for when the transaction has failed
- Removes the `StatusStepper`
- Removes `useSafeCreation`
- Removes `useSafeCreationEffects`
- Shows a speed-up message if the transaction takes longer than 15 seconds
- Removes `CreationModal` and instead shows the new Success screen with the address attached

## ToDos

- [x] Need to figure out how to select the correct pending safe creation as there can be two now with counterfactual
- [x] Still pass `safeViewRedirectURL` on SUCCESS
- [x] Navigate user to dashboard on SUCCESS
- [x] Check that analytics events are still mapped correctly
- [x] Add safe (+ owners) to address book on SUCCESS

## How to test it

- create a safe through a safe
- the user should only proceed to the status screen once they submit the tx in their wallet

## Screenshots
Screenshot 2024-05-30 at 16 24 29
Screenshot 2024-05-30 at 16 31 20
Screenshot 2024-05-30 at 16 36 44
Screenshot 2024-06-18 at 10 40 50

https://github.com/safe-global/safe-wallet-web/assets/5880855/19a1c895-2749-4a7e-b077-86abb1ea15af

## Checklist


jmealy (approved):

francovenica:
This Scroll bar shoudn't be there
image

francovenica:
If the user reject the tx we show an error message saying that "Something went wrong". For tx executions in a safe we say that "the user rejected the tx". Can't we do the same here?
image
image

francovenica:
When the tx fails it says that the tx "was rejected by the connected wallet". I think this makes belive that the user rejected the tx on purpose, and not that the tx failed for some reason outside of the user's control.
I think it should just say that the tx has failed without the "rejected by the wallet" text.

In this snapshot I, on purpose, set the gas limit too low to make the tx fail:
image

francovenica:
Beyond the comments I made, the rest looks fine. 

Regarding this comment in the Notion doc:
"We remove the status page and instead navigate the user to their safe even while its deploying. This is already how its done for Counterfactual safes and would unblock the UI for the user. We could show the deployment status as a sticky notification."

I'm not 100% in removing the "creation done" modal that used to show what the user can do with the safe and also not fully sold about dumping the user into the safe as it is being deployed. I don't think the user will start to navigate the UI as it sees that the safe is deploying with all those loaders. 

My suggestion would be to show a modal saying "your safe is being deployed, it might take a few minutes, on the meanwhile you can still explore it". And if the safe is deployed then you show the "Your safe has been deployed: [ADDRESS]" modal.

usame-algan:
@francovenica thanks for testing! I pushed a fix to address the issues you found. Could you try again?

> I'm not 100% in removing the "creation done" modal that used to show what the user can do with the safe and also not fully sold about dumping the user into the safe as it is being deployed. I don't think the user will start to navigate the UI as it sees that the safe is deploying with all those loaders.

> My suggestion would be to show a modal saying "your safe is being deployed, it might take a few minutes, on the meanwhile you can still explore it". And if the safe is deployed then you show the "Your safe has been deployed: [ADDRESS]" modal.

You are talking about this screen right? We should document why we want to remove it. cc @TanyaEfremova From my understanding it overloads the user with information about what they can do with their Safe while also blocking them from those actions and essentially forcing them to do an onboarding which can be frustrating.
Screenshot 2024-06-17 at 10 42 37

TanyaEfremova:
I think, we agreed to show the success screen with the account address and a welcoming text instead of this one. I don't think, the current screen provides a lot of error. We also have another success screen for CF accounts, which is completely different from this one, so they need to be aligned anyway, there is no reason in keeping both.

TanyaEfremova:
Yes, exactly! Before, we had a modal with the overview, when an account was created, then once the CF account is _deployed_ a different modal was shown, and the overview wasn't. So we would like to merge them, and only show the one you posted on the screenshot :) So you are correct.

francovenica:
Ok, Given Tanya's comment and a talk I had with Usame I see that this success screen is the only one and correct. I was afraid I was missing some success screen somewhere

LGTM, we can pass this ticket to done

PR #3780, opened 2024-05-30
fix: Optimize RPC requests when predicting safe address

usame-algan:
## What it solves

Noticed this when debugging failing e2e tests the other day. This change reduces the number of RPC requests when calling `getAvailableSaltNonce` from 3 to 1.

## How this PR fixes it

- Calling `SafeFactory.create` always checks if the factory contract(s) are deployed on the current network
- Doesn't create a new instance of `SafeFactory` every time we want to predict the safe address
- Instead it now calls the utils function that hte protocol-kit provides directly

## How to test it

1. Creating a Safe should still work
2. Creating a counterfactual safe should still work
3. The predicted safe address should still be correct and what the user ends up with eventually once its deployed on chain

## Checklist

### DIFF
--- a/src/components/new-safe/create/logic/index.ts
+++ b/src/components/new-safe/create/logic/index.ts
@@ -19,7 +19,7 @@ import { AppRoutes } from '@/config/routes'
 import { SAFE_APPS_EVENTS, trackEvent } from '@/services/analytics'
 import type { AppDispatch, AppThunk } from '@/store'
 import { showNotification } from '@/store/notificationsSlice'
-import { SafeFactory } from '@safe-global/protocol-kit'
+import { predictSafeAddress, SafeFactory } from '@safe-global/protocol-kit'
 import type Safe from '@safe-global/protocol-kit'
 import type { DeploySafeProps } from '@safe-global/protocol-kit'
 import { createEthersAdapter, isValidSafeVersion } from '@/hooks/coreSDK/safeCoreSDK'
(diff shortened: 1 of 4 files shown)


src/components/new-safe/create/logic/index.ts
```diff
@@ -86,9 +86,18 @@ export const createNewSafe = async (
...
-  const safeFactory = await getSafeFactory(ethersProvider)
-  return safeFactory.predictSafeAddress(props.safeAccountConfig, props.saltNonce)
+  const ethAdapter = await createEthersAdapter(ethersProvider)
+
+  return predictSafeAddress({
+    ethAdapter,
+    chainId: BigInt(chainId),
+    safeAccountConfig: props.safeAccountConfig,
+    safeDeploymentConfig: {
+      saltNonce: props.saltNonce,
```
schmanu, line 98: If we do not pass in the `safeVersion` here the SDK will use its current `DEFAULT_SAFE_VERSION`.
But we store our `LATEST_SAFE_VERSION` in the local storage. 
Updating the protocol-kit could or our interface's safe version could lead to a mismatch  which would lead to a mismatch of the predicted Safe address.

schmanu (changes requested):

src/components/new-safe/create/logic/index.ts
usame-algan: We call this function before a safe is created where we use `LATEST_SAFE_VERSION` and don't read anything from local storage. Still makes sense to explicitely add `LATEST_SAFE_VERSION` here in case there is some unexpected change on the protocol-kit side.

schmanu (approved):

francovenica:
LGTM

Created saveral Safes and CF safes, I didn't had any issues

PR #3782, opened 2024-05-30
Chore: fix branch conditionals

katspaugh:
## What it solves

The condition to check the PR branch was wrong.

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -41,7 +41,7 @@ jobs:
       - uses: ./.github/workflows/build
         with:
           secrets: ${{ toJSON(secrets) }}
-          prod: ${{ github.ref == 'refs/heads/main' }}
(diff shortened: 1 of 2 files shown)


katspaugh:
Tested on this branch, it works correctly now. ✅

.github/workflows/deploy-dev.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - uses: ./.github/workflows/build
         with:
           secrets: ${{ toJSON(secrets) }}
-          prod: ${{ github.ref == 'refs/heads/main' }}
+          prod: ${{ github.base_ref == 'main' }} # PRs to main are Release Candidates and must be tested with prod settings
```
usame-algan, line 44: Will this still work for both staging and dev deployments? Is base_ref dev when merging something from dev to main or only when merging into dev?

usame-algan (approved):

PR #3784, opened 2024-05-31
Feat: private key signer

katspaugh:
## What it solves

An experimental web3-onboard module to allow connecting directly with a private key as a Safe signer.

The idea is that you can have a (non-primary) signer key stored in your password manager/OS keychain and log in with it like with a password.

Screenshot 2024-05-31 at 12 03 18

The private key is then stored in the session storage and it reconnects if the page is reloaded.

Screenshot 2024-05-31 at 11 58 24
adc2f">


francovenica:
It looks really promising! and it will going to be really useful for automation to be able to switch owners, so test on 2/x safes can be performed

My initial feedback.
- Clear the storage when you disconnect. One for safty reasons (hopefully no laptop is stolen ever, but just in case) and second because now when you disconnect and connect again, it takes the same Pk you already introduced, so there is actually no way to connect with a different owner beyond clearing the LS
- Switching networks is not possible. (maybe not possible since there is not a wallet to ask for a network changing, but still something to take in account)

katspaugh:
@francovenica fixed both, thank you!

src/hooks/wallets/wallets.ts
```diff
@@ -45,6 +46,7 @@ const WALLET_MODULES: { [key in WALLET_KEYS]: (chain: ChainInfo) => WalletInit }
   [WALLET_KEYS.LEDGER]: () => ledgerModule() as WalletInit,
   [WALLET_KEYS.TREZOR]: () => trezorModule({ appUrl: TREZOR_APP_URL, email: TREZOR_EMAIL }) as WalletInit,
   [WALLET_KEYS.KEYSTONE]: () => keystoneModule() as WalletInit,
+  [WALLET_KEYS.PK]: (chain) => pkModule(chain.chainId, chain.rpcUri) as WalletInit,
```
schmanu, line 49: I think we ensure that this wallet can only be enabled when `IS_PRODUCTION` is `false` as it is a tool for our e2e tests.

src/services/private-key-module/index.ts
```diff
@@ -0,0 +1,126 @@
...
+              },
+
+              personal_sign: async ({ params }) => {
+                const signedMessage = wallet.signingKey.sign(params[0])
+                return signedMessage.serialized
+              },
+
+              eth_signTypedData: async ({ params }) => {
+                const signedMessage = await wallet.signTypedData(params[1].domain, params[1].data, params[1].value)
+                return signedMessage
```
schmanu, line 109: Nit:
```suggestion
                return wallet.signTypedData(params[1].domain, params[1].data, params[1].value)
```

schmanu (commented):
If I understand correctly this is meant to mainly be a tool for our e2e tests to use multiple signers.

I think we should make sure that this won't accidentally turned on in prod.

src/hooks/wallets/wallets.ts
katspaugh: To enable it, you'd have to go to the production config service, create a PK wallet, and enable it on a chain. Don't think this can be done accidentally.

katspaugh:
@schmanu I ended up added a prod check as you suggested because this will allow us to use this wallet on dev even if the CGW toggle is set to prod.

PR #3785, opened 2024-05-31
Fix: remove the New chip from Recovery settings

katspaugh:
## What it solves

The Recovery feature is not "new" anymore so this label should be removed.


katspaugh:
Screenshot 2024-05-31 at 16 04 04

PR #3786, opened 2024-05-31
Feat: add swaps card to safe apps list

jmealy:
## What it solves

Resolves https://www.notion.so/safe-global/Add-a-block-to-apps-page-985e220c975b42b7950929ecb02a6715?pvs=4

## How this PR fixes it
- Add a component for showing native swaps in the safe apps list
- Allow the card to be dismissed. This decision is persisted in localstorage.
- Make the cards respond to the search and filter like the other safe apps.

## How to test it
- Open the safe apps list.
- By default the new swaps card should be shown first in the list
- It should be shown when searching with the keyword `swap`, along with other swapping apps.
- It should show when filtering for the `defi` category
- Clicking `try now` should bring you to the swaps feature
- Clicking `Don't show` should remove the card from the list. It will remain hidden unless localstorage is cleared.

## Screenshots
image

## Checklist


src/components/safe-apps/hooks/useNativeSwapsAppCard.ts
```diff
@@ -0,0 +1,33 @@
...
+    id: 100_000,
+    url: AppRoutes.swap,
+    name: 'Native swaps are here!',
+    description: 'Experience seamless trading with better decoding and security in native swaps.',
+    accessControl: { type: SafeAppAccessPolicyTypes.NoRestrictions },
+    tags: ['DeFi'],
+    features: [],
+    socialProfiles: [],
+    developerWebsite: '',
+    chainIds: [chainId],
```
jmealy, line 23: here I have the card shown on whatever chain is selected. Is swaps enabled on all chains?

compojoom: yes, you can use: `  const isSwapFeatureEnabled = useHasFeature(FEATURES.NATIVE_SWAPS)` to determine if swaps are enabled on the current chain.
Also how about we don't go for a custom "isNativeFeature" flag, but instead expose the swap app id as a custom const NATIVE_SWAP_APP_ID = 100_000
and then later you can compare against that id, instead of taht new isNativeFeature flag.

compojoom:
It's ok for me now. @schmanu ?

francovenica:
Ok, looks good. I've checked the things asked in the description plus what manu asked to be changed: be filtered by DEX and "Optimized with batched tx"

Question:
Currently when you search, the card leaves the first place and places just among the other apps. Would it be possible to make it appear always first since is technically not an app, so it shouldn't be mixed with them.
image

I don't consider this a blocker, so if too outside of the scope we can create a new ticket for just that and pass this one

compojoom:
> Currently when you search, the card leaves the first place and places just among the other apps. Would it be possible to make it appear always first since is technically not an app, so it shouldn't be mixed with them.

Actually this seems to be a bug, we have to fix this :(

src/components/safe-apps/SafeAppList/index.tsx
```diff
@@ -70,18 +86,28 @@ const SafeAppList = ({
...
+            <li key={safeApp.id}>
+              <SafeAppCard
+                safeApp={safeApp}
+                isBookmarked={bookmarkedSafeAppsId?.has(safeApp.id)}
+                onBookmarkSafeApp={onBookmarkSafeApp}
+                removeCustomApp={removeCustomApp}
+                onClickSafeApp={handleSafeAppClick(safeApp)}
+                openPreviewDrawer={openPreviewDrawer}
+              />
+            </li>
```
katspaugh, line 100: Why does this card have to be added into the `safeAppsList`? Why can't it just be rendered outside of the map? Feels very hacky to me.

jmealy: It gets mixed in with the safe apps in the parent component just so that it responds to filtering.
I changed it here to render outside of the map, hopefully it's a bit better

jmealy:
> Ok, looks good. I've checked the things asked in the description plus what manu asked to be changed: be filtered by DEX and "Optimized with batched tx"

Should be ok now

src/pages/apps/index.tsx
```diff
@@ -13,18 +13,21 @@ import useSafeAppsFilters from '@/hooks/safe-apps/useSafeAppsFilters'
 import SafeAppsFilters from '@/components/safe-apps/SafeAppsFilters'
 import { useHasFeature } from '@/hooks/useChains'
 import { FEATURES } from '@/utils/chains'
+import { useNativeSwapsAppCard } from '@/components/safe-apps/hooks/useNativeSwapsAppCard'
 
 const SafeApps: NextPage = () => {
   const router = useRouter()
+  const { swapsCardDetails } = useNativeSwapsAppCard()
   const { remoteSafeApps, remoteSafeAppsLoading, pinnedSafeApps, pinnedSafeAppIds, togglePin } = useSafeApps()
+  const allApps = useMemo(() => [swapsCardDetails, ...remoteSafeApps], [remoteSafeApps, swapsCardDetails])
```
katspaugh, line 22: This can be reverted as well then?

src/components/safe-apps/SafeAppList/index.tsx
katspaugh: I think it's totally fine if it disappears when the list is filtered.

src/pages/apps/index.tsx
jmealy: This is still needed to apply the filters

jmealy: ah just saw you other comment. I will remove the filtering then.

src/components/safe-apps/NativeFeatureCard/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <Typography className={css.title} variant="h5">
+          {details.name}
+        </Typography>
+
+        <Typography className={css.description} variant="body2" color="text.secondary">
+          {details.description}
+        </Typography>
+
+        <Stack direction="row" gap={2} className={css.buttons}>
+          <Button onClick={onClick} size="small" variant="contained" sx={{ px: '16px' }}>
```
katspaugh, line 37: There should be a tracking event on this button (see other Swap buttons).

src/components/safe-apps/SafeAppList/index.tsx
```diff
@@ -69,19 +87,30 @@ const SafeAppList = ({
...
-              isBookmarked={bookmarkedSafeAppsId?.has(safeApp.id)}
-              onBookmarkSafeApp={onBookmarkSafeApp}
-              removeCustomApp={removeCustomApp}
-              onClickSafeApp={handleSafeAppClick(safeApp)}
-              openPreviewDrawer={openPreviewDrawer}
-            />
-          </li>
-        ))}
+        {safeAppsList.map((safeApp) => {
+          if (safeApp.id === NATIVE_SWAPS_APP_ID) return null
```
katspaugh, line 100: This is not needed anymore.

src/components/safe-apps/SafeAppList/index.tsx
```diff
@@ -48,6 +56,16 @@ const SafeAppList = ({
...
 
+  const handleNativeAppClick = useCallback(
+    (route: string) => {
+      router.push({
+        pathname: route,
+        query: router.query,
+      })
+    },
+    [router],
+  )
```
katspaugh, line 67: This can be replaced with a regular link inside the NativeFeatureCard

src/pages/apps/index.tsx
jmealy: I now have the swaps card being passed down separately from the safe apps list, and it disappears when a filter is applied

src/components/safe-apps/hooks/useNativeSwapsAppCard.ts
katspaugh, line 1: I would suggest to remove this and simply hardcode the text in `src/components/safe-apps/NativeFeatureCard/index.tsx` directly. `NativeFeatureCard` can then be renamed to `NativeSwapsCard`.

src/pages/apps/index.tsx
```diff
@@ -24,7 +26,7 @@ const SafeApps: NextPage = () => {
 
   const nonPinnedApps = useMemo(
     () => remoteSafeApps.filter((app) => !pinnedSafeAppIds.has(app.id)),
-    [remoteSafeApps, pinnedSafeAppIds],
+    [pinnedSafeAppIds, remoteSafeApps],
```
katspaugh, line 29:
```suggestion
    [remoteSafeApps, pinnedSafeAppIds],
```

katspaugh:
I don't understand why so many abstraction layers were added just to show a static card in the beginning of the apps list...

src/components/safe-apps/SafeAppList/index.tsx
```diff
@@ -31,9 +38,12 @@ const SafeAppList = ({
   removeCustomApp,
   title,
   query,
+  isFiltered = false,
 }: SafeAppListProps) => {
   const { isPreviewDrawerOpen, previewDrawerApp, openPreviewDrawer, closePreviewDrawer } = useSafeAppPreviewDrawer()
   const { openedSafeAppIds } = useOpenedSafeApps()
+  let [isSwapsCardVisible = true, setIsSwapsCardVisible] = useLocalStorage<boolean>(SWAPS_APP_CARD_STORAGE_KEY)
```
katspaugh, line 45: Why can't this be inside `NativeSwapsCard` itself?

src/components/safe-apps/SafeAppList/index.tsx
```diff
@@ -31,9 +38,12 @@ const SafeAppList = ({
...
+  const isSwapFeatureEnabled = useHasFeature(FEATURES.NATIVE_SWAPS)
```
katspaugh, line 46: Same for this. If not enabled, just return null.

src/features/swap/index.tsx
```diff
@@ -94,7 +95,7 @@ const SwapWidget = ({ sell }: Params) => {
 
   const appData: SafeAppData = useMemo(
     () => ({
-      id: 1,
+      id: NATIVE_SWAPS_APP_ID,
```
katspaugh, line 98:
```suggestion
      id: 1,
```

francovenica:
@jmealy The deploy has failed, can you re-trigger it to see if it at least creates the environment again to test?

src/features/swap/index.tsx
```diff
@@ -27,6 +27,7 @@ import { isBlockedAddress } from '@/services/ofac'
 import { selectSwapParams, setSwapParams, type SwapState } from './store/swapParamsSlice'
 import { setSwapOrder } from '@/store/swapOrderSlice'
 import useChainId from '@/hooks/useChainId'
+import { NATIVE_SWAPS_APP_ID } from '@/features/swap/config/constants'
```
katspaugh, line 30:
```suggestion
```

src/features/swap/config/constants.ts
```diff
@@ -0,0 +1 @@
+export const NATIVE_SWAPS_APP_ID = 100_000
```
katspaugh, line 1:
```suggestion
```

jmealy:
> I don't understand why so many abstraction layers were added just to show a static card in the beginning of the apps list...

It was mostly so the filtering would work. Hopefully it's ok now, sorry for the mess!

katspaugh (approved):
Thanks for refactoring it! 🙏

compojoom:
But wait, so now we no longer show the safe widget when we filter?

francovenica:
> But wait, so now we no longer show the safe widget when we filter?
I was to comment this.

Also I just saw that, if the card is too thin (for some viewports) the button text are in 2 rows. If you reduce the padding to 21px it fits just fine with the thiniest cards
image
image

katspaugh:
Yes, it's fine to now show it when you filter.

francovenica:
Ok, the filtering is out.

See if the padding of that button can be fixed. 

Beyond that the ticket is fine to be merged

PR #3789, opened 2024-06-03
Fix: Error initializing the Safe Core SDK -> Error connecting to the blockchain

### LINKED ISSUE #3587: Safe SDK couldn't be initialized
When attempting a transaction creation very quickly after loading the page, an error appears:

```
Error initializing the Safe Core SDK
```

It has been one of the top 5 errors [on Sentry](https://safe-global.sentry.io/issues/5045301034/?project=6769196&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=3) for a while now.

## Requirements

* Find out why it breaks so often, looks like a race condition
* Make the error message clearer in case this error is unavoidable (e.g. an RPC failure)

katspaugh:
## What it solves

Resolves #3587

Judging by the Sentry logs, this error is happening when an RCP responds with an error.

There's nothing we can do about this from the frontend, but the error message should be clearer.

### DIFF
--- a/src/hooks/coreSDK/useInitSafeCoreSDK.ts
+++ b/src/hooks/coreSDK/useInitSafeCoreSDK.ts
@@ -43,7 +43,7 @@ export const useInitSafeCoreSDK = () => {
         const e = asError(_e)
         dispatch(
           showNotification({
-            message: 'Please try connecting your wallet again.',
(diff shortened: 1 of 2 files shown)


compojoom (approved):
I approve this message!

PR #3791, opened 2024-06-03
fix: switch between swap & limit loses token [SWAP-97]

compojoom:
## What it solves
When switching betwenn Swap/Limit forms, selected tokens are reset. On CoW Swap they  are synced between all forms.

## How this PR fixes it
We were re-rendering the widget when the user switches between swap and a limit swap and this was causing the sellToken to reset. Now, we no longer do this

## How to test it
1. Navigate to swaps
2. Select a different token
3. switch to limit order

## Checklist

### DIFF
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,6 +1,5 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference types="next/navigation-types/compat/navigation" />
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/basic-features/typescript for more information.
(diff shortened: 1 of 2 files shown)


francovenica:
Question. I see it here as it is in QA in the CowSwap board in notion, but is not currently in the github board, any reason no not having it there?

I've checked and the tokens are preserverd when switching from the "Swap" form to the "limits" form.

Question: The values seems to not be maintained. is this intended? I tried to see in the original cowswap app, but currently it seems to be having problems (is not connecting the wallet and it is working really slow) so I cannot check right now

schmanu (approved):

compojoom:
> Question. I see it here as it is in QA in the CowSwap board in notion, but is not currently in the github board, any reason no not having it there?

Only because our workflows suck... And I have to do the same things twice

> Question since it's not in the ticket description and it might be out of the scope: The values seems to not be maintained. is this intended?

I wanted to also maintain the values, but the swap widget behaves strange there. It seems to keep separate values for swap and limit, so it doesn't apply the swap value to limit and vice versa. I'll ask anxo about this.

PR #3795, opened 2024-06-04
Fix: same addresses showing same balance

### LINKED ISSUE #3794: Same safe addresses on different chains show same balance
## Bug description
If you have multiple safes with the same address on different chains (L1, L2) the balance of the L2 safe is the same as the one on L1 - which is wrong. 

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Create the same safe on multiple networks
 2. Observe that both have the same balance

## Expected result
The safes should have different balances

katspaugh:
## What it solves

Resolves #3794

## How this PR fixes it

`chainId` wasn't taken into account when looking up Safes in the balances list.


compojoom (approved):
muy bueno!

katspaugh:
I’ve tested it with the Daniel’s safes.

PR #3796, opened 2024-06-04
Feat: top 5 assets on the dashboard

katspaugh:
## What it solves

A helpful widget to get a quick glance at Safe's assets + promote the Swaps feature on the Dashboard.

Screenshot 2024-06-04 at 13 31 40


compojoom (approved):

francovenica:
The list doesn't take in account if tokens were hidden in the assets tab:
image
image

Suggestion:
Adding an empty state, so if you have no tokens in the safe it doesn't show an empty native token at the top.
swap token reset

Suggestion 2:
In networks where the swap feature is not available no swap button is shown (which is correct), but makes the widget less useful. Maybe the "Send" button can be put there for those networks.

katspaugh:
Done, @francovenica 👍

francovenica:
Looks good! approved

PR #3801, opened 2024-06-05
Fix: filter out undefined address in useSafeOverviews

katspaugh:
## What it solves

I'm not sure what the root cause of undefined addresses is, they seem to be already [filtered by Boolean](https://github.com/safe-global/safe-wallet-web/blob/dev/src/components/welcome/MyAccounts/useAllSafes.ts#L55) before being passed to `useSafeOverviews`. But here's a fix that just filters out bad values before sending a request to CGW.

### DIFF
--- a/src/components/welcome/MyAccounts/__tests__/useSafeOverviews.test.ts
+++ b/src/components/welcome/MyAccounts/__tests__/useSafeOverviews.test.ts
@@ -1,0 +1,47 @@
+import useSafeOverviews from '../useSafeOverviews'
+import * as balances from '@/hooks/loadables/useLoadBalances'
+import * as sdk from '@safe-global/safe-gateway-typescript-sdk'
+import * as useWallet from '@/hooks/wallets/useWallet'
+import * as store from '@/store'
+import type { Eip1193Provider } from 'ethers'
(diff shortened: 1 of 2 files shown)


src/components/welcome/MyAccounts/__tests__/useSafeOverviews.test.ts
```diff
@@ -0,0 +1,47 @@
...
+import { act } from 'react-dom/test-utils'
+
+jest.spyOn(balances, 'useTokenListSetting').mockReturnValue(false)
+jest.spyOn(store, 'useAppSelector').mockReturnValue('USD')
+jest
+  .spyOn(useWallet, 'default')
+  .mockReturnValue({ label: 'MetaMask', chainId: '1', address: '0x1234', provider: null as unknown as Eip1193Provider })
+
+describe('useSafeOverviews', () => {
+  it('should filter out undefined addresses', async () => {
```
iamacook, line 17: Nit: whilst I can tell that the code is correct, I'd add a test for `undefined` `chainId` values as well - those are the errors we're seeing in the logs.

iamacook (approved):

PR #3802, opened 2024-06-05
Analytics: remove 0x from signer address

katspaugh:
## What it solves

Like we do with Safe addresses, signer addresses must be trimmed so that GA/BigQuery doesn't try to convert them to integers.


alicecha (approved):

francovenica:
LGTM:
A before/after comparison
image

image

PR #3803, opened 2024-06-05
feat: add link to tx for order notifications [SWAP-61]

compojoom:
## What it solves
We don’t always have the txId in the order information, but when we do we add a link to the notification.
It’s highly likely that order created events won’t ever have a txId as this event is created before submitting the tx to the blockchain.

I had to extract getTxLink to a separate file as having it in useTxNotifications was causing a circular dependency error in our tests.

Resolves #

## How this PR fixes it
This PR cannot guarantee that the push notification displayed will have a "view transaction" link. It all depends in what order the updates to the order will come through (whether they will be triggered through the widget or through the transaction service). If the user leaves the widget page after submitting a tx, then updates will most probably come from the tx service and there will be "view transaction" link in the notification.

## How to test it
Submit an order and watch the push notifications. The "order executed" one should have a link to "view transaction".

## Screenshots
grafik

## Checklist

### DIFF
--- a/src/store/swapOrderSlice.ts
+++ b/src/store/swapOrderSlice.ts
@@ -6,6 +6,8 @@ import { isSwapTxInfo, isTransactionListItem } from '@/utils/transaction-guards'
 import { txHistorySlice } from '@/store/txHistorySlice'
 import { showNotification } from '@/store/notificationsSlice'
 import { selectSafeInfo } from '@/store/safeInfoSlice'
+import { selectChainById } from '@/store/chainsSlice'
+import { getTxLink } from '@/utils/tx-link'
 
 type AllStatuses = OrderStatuses | 'created'
 type Order = {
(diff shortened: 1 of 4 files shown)


src/store/swapOrderSlice.ts
```diff
@@ -94,6 +103,7 @@ export const swapOrderStatusListener = (listenerMiddleware: typeof listenerMiddl
                   : 'Waiting for confirmation from signers of your Safe',
               groupKey,
               variant: 'info',
+              link: link ? link : undefined,
```
schmanu, line 106: I think if we initialize the `link` as 
`let link = undefined`
we could get rid of all these checks and simply do in all these lines

```suggestion
              link,
```

src/store/swapOrderSlice.ts
```diff
@@ -80,11 +82,18 @@ export const swapOrderStatusListener = (listenerMiddleware: typeof listenerMiddl
       if (oldStatus === newStatus || newStatus === undefined) {
         return
       }
+      const safeInfo = selectSafeInfo(listenerApi.getState())
+
+      let link = null
+      if (swapOrder.txId && safeInfo.data?.chainId && safeInfo.data?.address) {
+        const chainInfo = selectChainById(listenerApi.getState(), safeInfo.data?.chainId)
+        if (chainInfo !== undefined) {
+          link = getTxLink(swapOrder.txId, chainInfo, safeInfo.data?.address.value)
```
schmanu, line 91: I feel like it would be more re-usable to add a new optional field `txId` to the `notificationSlice`.
Then our redux middleware does not need to know how to create URIs and fetch all this data here. The Notification component could handle that.

compojoom: yes, I agree! I was thinking about the same. I think that we also need to pass the safe address this way we can fix the bug where we display a notification for the wrong safe.

Doing this kind of modifications in this PR however seemed like to much of a refactoring and wouldn't be clear why it was done. 

Also there are some instances where we have more complex links than just to the transaction:
grafik

So, you are suggesting we keep a txId and link. If no link is passed, but a txId exist, we try to link to the transaction?

schmanu: Hmm we can also for now leave it as is.
I think reworking the whole notifications and merging them with pendingTxs is also still an idea floating around and would be the proper way of doing this.

I'll approve for now.

schmanu (approved):

PR #3804, opened 2024-06-05
Feat: calculate surplus for partially filled orders

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Better-Surplus-calculation-2e3bf20feeed46938c6efa6728c5ea94?pvs=4

## How this PR fixes it
- Calculate surplus for partially filled orders instead of only for fully fulfilled orders, like CoW Swap are doing.

## How to test it
- Create a partially fillable limit order.
- When the order is partially filled, check that the surplus price is displayed. Check that it is the same as the order on cowswap.
- You can set the sell amount to more than your current balance to make sure that it will be partially filled

## Screenshots
image

## Checklist

### DIFF
--- a/src/features/swap/helpers/utils.ts
+++ b/src/features/swap/helpers/utils.ts
@@ -37,16 +37,13 @@ export const getExecutionPrice = (
 }
 
 export const getLimitPrice = (
-  order: Pick<SwapOrder, 'sellAmount' | 'buyAmount' | 'buyAmount' | 'buyToken' | 'sellToken'>,
+  order: Pick<SwapOrder, 'sellAmount' | 'buyAmount' | 'buyToken' | 'sellToken'>,
 ): number => {
   const { sellAmount, buyAmount, buyToken, sellToken } = order
 
   const ratio = calculateRatio(
     { amount: sellAmount, decimals: sellToken.decimals },
-    {
-      amount: buyAmount,
-      decimals: buyToken.decimals,
-    },
+    { amount: buyAmount, decimals: buyToken.decimals },
   )
 
   return ratio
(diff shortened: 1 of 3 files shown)


src/features/swap/helpers/utils.ts
```diff
@@ -72,6 +69,49 @@ export const getSurplusPrice = (
...
+  if (order.kind === OrderKind.BUY) {
+    return _getPartiallyFilledBuySurplus(order)
+  } else if (order.kind === OrderKind.SELL) {
+    return _getPartiallyFilledSellSurplus(order)
+  } else {
+    return 0
+  }
+}
+
+const _getPartiallyFilledBuySurplus = (
```
compojoom, line 82: @katspaugh do we have a convention for _? If not I would remove it. As this function is not being exported it is clear that it is "private".

katspaugh: Yes, no need for _ for not exported functions. The convention is the same as in Python, e.g. if you export something only for unit tests, it can be prefixed with an _.

compojoom:
@francovenica or @liliya-soroka it can be e a bit tricky to get a partially filled order on sepolia. The easiest way to do it in my opinion is to make a limit order for an amount higher than what you have in the safe. Then when the first part gets filled, you can fund the safe and would get a second fill order.

compojoom (changes requested):
@jmealy please remove the _ in the private functions and we are good to go.

compojoom (approved):
👍

francovenica:
LGTM 

First I forced an order to fill partially by swapping more tokens than I had and it worked fine, the surpluss in cowSwap and the tx details in safe were the same
Then I transfer to the safe more of that tokens to make the order fill completely. After that I've checked that the tx updated the surplus in the safe tx details as it was updated in cowswap.
Snapshots and links to the tx:
https://feat_swap_surplus--walletweb.review.5afe.dev/transactions/tx?safe=sep:0x8f4A19C85b39032A37f7a6dCc65234f966F72551&id=multisig_0x8f4A19C85b39032A37f7a6dCc65234f966F72551_0x9399daa69f6c03036021e25e6dd6e2a139a7e5cf65e4b610eea2291e967d85e4
https://explorer.cow.fi/sepolia/orders/0xfb01850d548708b6a1bcc863f1d9226aa37f0ecde1a09028c6a5fceb8c5226838f4a19c85b39032a37f7a6dcc65234f966f72551666675b6?tab=overview

Partially filled:
image
image

Fully filled and updated:
image
image

PR #3807, opened 2024-06-06
Feat: better fiat and token amount formatting

katspaugh:
## What it solves

A revised fiat and token amount formatting rules as outlined [here](https://www.notion.so/safe-global/Revised-number-formatting-cd2f32bcd9174626851c27bb34cb163a).

## How this PR fixes it

This PR simplifies the formatting logic quite a lot, and relies on browser's default formatting plus some tweaks.

Additionally, it right-aligns fiat values in Asset tables.

Screenshot 2024-06-06 at 14 49 21

### DIFF
--- a/src/components/balances/AssetsTable/index.tsx
+++ b/src/components/balances/AssetsTable/index.tsx
@@ -80,6 +80,7 @@ const headCells = [
     id: 'value',
     label: 'Value',
     width: '20%',
+    align: 'right',
   },
   {
     id: 'actions',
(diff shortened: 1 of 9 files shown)


compojoom (approved):

francovenica:
There are concerns about numbers "dancing" given the different space that numbers take, so a $11 will take much less space than a $88, so the numbers look "not aligned" even when they are
https://format--walletweb.review.5afe.dev/balances?safe=eth:0x8675B754342754A30A2AeF474D114d8460bca19b

image

francovenica:
It was agreed that the "dancing" will be addressed in the future:
https://5afe.slack.com/archives/C03DAGWJCR5/p1718031746930849?thread_ts=1718025802.019199&cid=C03DAGWJCR5

LGTM

PR #3808, opened 2024-06-06
Fix: move the execute button on recovery transactions to the correct column [SWAP-75]

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Reposition-execute-button-for-the-recovery-transaction-ba6c3968a2c44d2f93f9f9e848bda873

## How this PR fixes it
- puts the recovery transactions execute button in the 'actions' column.
- uses the same styles as the normal recovery button to make sure they are aligned.
- removes unused css class (unrelated)

## Screenshots
image

## Checklist


katspaugh (approved):

francovenica:
Looks like the snapshot. Passed

PR #3809, opened 2024-06-06
feat: relaying for native swaps only [SWAP-100]

schmanu:
## What it solves

Resolves SWAP-100

## How this PR fixes it
New feature flag to offer relaying for native swaps only.

## How to test it
- On staging gnosis chain swaps are temporily only available for native swaps to test this PR
- Test a tx from the native swaps
- Test a tx from other sources / Safe apps

## Checklist

### DIFF
--- a/src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
+++ b/src/components/tx/SignOrExecuteForm/ExecuteForm.tsx
@@ -41,13 +41,11 @@ export const ExecuteForm = ({
   isCreation,
   isOwner,
   isExecutionLoop,
-  relays,
   txActions,
   txSecurity,
 }: SignOrExecuteProps & {
   isOwner: ReturnType<typeof useIsSafeOwner>
   isExecutionLoop: ReturnType<typeof useIsExecutionLoop>
-  relays: ReturnType<typeof useRelaysBySafe>
   txActions: ReturnType<typeof useTxActions>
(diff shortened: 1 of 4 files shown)


compojoom (approved):
👍

francovenica:
I've checked the Relayer feature is only enabled for native swap, you can access it by the swap button on the sidebar or the Swap button in the "New tx" modal.
Any other tx, even a CowSwap tx from the "Safe apps" page doesn't offer using relayer.

Looks good to me

Note: (Probably not part of the scope of this ticket)
I tried a relayed swapping and it worked on the cowswap side (the order was filled), but on our side the tx never finished to be signed, still reads as 0 out 1 and with an "Expired" status even tho the order was filled.
I understand that gnosis chain in stg is not really staging, but is using prod backend, so that might be a factor in all this.
Here is the tx, see in which state is tomorrow:
https://relay_swaps--walletweb.review.5afe.dev/transactions/tx?safe=gno:0xB8d760a90a5ed54D3c2b3EFC231277e99188642A&id=multisig_0xB8d760a90a5ed54D3c2b3EFC231277e99188642A_0xed4b9d156121e6a01b1b86fc4945d9a7e9819658e3c106de876ff61d65ec4724

schmanu:
@francovenica 
The transaction you shared is displayed correctly to me: fully signed and with a fulfilled order.
That might have been an caching issue on our backend, but def. not related to this change.

PR #3811, opened 2024-06-07
Refactor: Consistently use latest MultiSendCallOnly address 

jfschwarz:
Refactoring to generally use the latest version of the MultiSendCallOnly available for the given network. This simplification enables removing some logic for determining the MultiSendCallOnly version from the codebase.

There is no dependency between the Safe version and the multisend version, meaning any Safe can use any multisend version. Assuming that the latest MultiSendCallOnly contract version is most optimized, it doesn't make sense to use legacy versions with older Safes.

A consistent multisend address also makes the configuration of the Roles Modifier more straightforward. (Users of the Roles mod need to explicitly enable all allowed multisend addresses.)

## How to test

Make some multisend transactions on different versions of Safe (1.3.0, 1.1.1, 1.4.1)

### DIFF
--- a/src/services/contracts/safeContracts.ts
+++ b/src/services/contracts/safeContracts.ts
@@ -1,6 +1,5 @@
 import {
   getFallbackHandlerContractDeployment,
-  getMultiSendCallOnlyContractDeployment,
   getProxyFactoryContractDeployment,
   getSafeContractDeployment,
   getSignMessageLibContractDeployment,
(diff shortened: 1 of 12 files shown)


src/services/contracts/safeContracts.ts
```diff
@@ -70,37 +69,11 @@ export const getReadOnlyGnosisSafeContract = async (chain: ChainInfo, safeVersio
...
-export const getReadOnlyMultiSendCallOnlyContract = async (chainId: string, safeVersion: SafeInfo['version']) => {
+export const getReadOnlyMultiSendCallOnlyContract = async (chainId: string) => {
   const ethAdapter = createReadOnlyEthersAdapter()
-  const multiSendVersion = _getMinimumMultiSendCallOnlyVersion(safeVersion)
-
   return ethAdapter.getMultiSendCallOnlyContract({
-    singletonDeployment: getMultiSendCallOnlyContractDeployment(chainId, multiSendVersion),
-    ..._getValidatedGetContractProps(safeVersion),
+    singletonDeployment: getMultiSendCallOnlyDeployment({ network: chainId }),
+    safeVersion: '1.4.1', // we generally use the latest multisend version (there is no dependency on the Safe version)
```
katspaugh, line 76: Can we get this version from safe-deployments instead of hardcoding?

liliya-soroka:
Verified

PR #3818, opened 2024-06-10
fix: display swap button in top assets only on hover

compojoom:
## What it solves
Display the swap button only on hover in the "top asset" block. 
@TanyaEfremova @kirkkonen looks a bit weird or like a bug, because we have this space on the right now without the button.

## Screenshots
grafik
grafik

## Checklist

### DIFF
--- a/src/components/dashboard/PendingTxs/styles.module.css
+++ b/src/components/dashboard/PendingTxs/styles.module.css
@@ -11,9 +11,17 @@
   min-height: 50px;
 }
 
+.container button {
+  opacity: 0;
+}
+
 .container:hover {
   background-color: var(--color-background-light);
   border-color: var(--color-secondary-light);
+}
+
+.container:hover button {
+  opacity: 1;
 }
 
 .list {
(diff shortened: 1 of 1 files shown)


src/components/dashboard/PendingTxs/styles.module.css
```diff
@@ -11,11 +11,19 @@
   min-height: 50px;
 }
 
+.container button {
+  opacity: 0;
```
katspaugh, line 15:
```suggestion
  opacity: 0;
  transition: opacity .1s ease-in;
```

katspaugh:
I would keep it w/o hover tbh.

katspaugh (approved):

francovenica:
I think the change is fine, but for mobile viewport, since there is no "hover", can we keep it always showing?

Note: technically if you keep pressing the asset row the button shows up, but with no indicator that that's there, nobody will do that.

image

katspaugh:
I'll go ahead and close this PR because:
* people do click on these buttons (second most popular entry point to Swaps after the sidebar)
* the layout become unbalanced until you hover due to the empty space on the right

PR #3821, opened 2024-06-10
feat: basic gnosis pay support

schmanu:
## What it solves
Gnosis Pay Safes can not be controlled from the Safe{Wallet} interface.
Resolves #

## How this PR fixes it

- It detects, that you are an owner of a Gnosis Pay Safe.
  - Owner of a Gnosis Pay Safe means that your connected wallet is a module in the DelayModifier module of the Safe.
- It offers transaction execution for all types of Safe Transactions via that module if you are a Gnosis Pay owner.
- It shows a small section highlighting that this is a Gnosis Pay Safe, and that there is a 3 minute delay between queueing and executing the transaction
- It supports enqueueing and executing the transaction after 3 minutes

## Checklist


katspaugh:
I'll close this PR because it became abandoned due to product priorities but it's good to have the code as a reference.

Some thoughts on a general approach to modules:

* We now have 2 officially supported modules – Spending Limits and Roles & Permissions
* Both have custom components integrated in the tx flow
* Both have custom analytics events (or rather custom event actions for the regular tx events)
* Custom branding (in case of Roles and Gnosis Pay)
* It would be great to generalize/formalize this so that each module implements a common interface + parametrized specifics.

PR #3828, opened 2024-06-11
Fix: explicitly pass a locale to NumberFormat

katspaugh:
## What it solves

Firefox doesn't play nicely with `undefined` as a locale parameter.


compojoom (approved):
seems to work 👍

PR #3834, opened 2024-06-12
fix: malicious recovery detection

schmanu:
## What it solves

Resolves https://www.notion.so/safe-global/Recovery-tx-is-being-considered-a-malicious-tx-724664e7de4e4a8da4d88b5606af047c?pvs=4

## How this PR fixes it
- Uses correct multiSend deployment when trying to detect if recovery is malicious

## How to test it
- Queue a recovery through the interface

## Checklist


katspaugh:
I guess it's a regression after #3811. So we cannot just always use the latest multisend contract?

schmanu:
Closing this as we reverted the change that caused this issue.

PR #3836, opened 2024-06-12
fix: do not pass negative app id

schmanu:
## What it solves

Resolves https://www.notion.so/safe-global/Signing-message-shows-a-too_small-error-8d4a815756e142e194e56fb57638b211

## How this PR fixes it
- Does not pass -1 as appId but undefined instead

## How to test it
- Sign a message
- Sign a tx

## Checklist


compojoom (approved):
beautiful 👌

katspaugh (approved):
Thanks, my bad.

katspaugh:
Tested, signing with WalletConnect works ✅

PR #3839, opened 2024-06-13
feat: return more info in appCommunicator getInfo

compojoom:
The return type of the getInfo endpoint now returns more data about the safe.

## What it solves
https://github.com/cowprotocol/cowswap/pull/4508 

The appCommunicator getInfo endpoint now returns the same data as api-kit's getSafeInfo. this makes it easier for safe dapp developers to just use the appcommunicator when it is there (and don't have to do an api-kit call to get info such as nonce or modules. 

## How to test it
This modifies the getInfo endpoint to return additional information. There is no way to test this through the UI.
## Checklist


katspaugh (approved):
🎉

PR #3842, opened 2024-06-14
Fix: remove empty span in balance changes

katspaugh:
## What it solves

Resolves https://github.com/5afe/safe-support/issues/461

## How this PR fixes it

I haven't been able to reproduce the ▯▯▯▯▯▯▯▯ bug but my best guess is that this empty span is the culprit.

Screenshot 2024-06-14 at 09 15 54


katspaugh:
@francovenica could you test this in Chrome on Windows? It seems to affect only Windows.

francovenica:
I tried it in Chrome Windows, I can still see it somewhat often. Is not consistent, but it can happen by just maxing the amount of tokens. You might have to try a few times

tokens

katspaugh:
Apparently it’s not the span.

PR #3844, opened 2024-06-17
chore: Upgrade protocol-kit to v4

### LINKED ISSUE #3750: Upgrade protocol-kit to v4
## What is the feature about

The protocol-kit v4 has some nice improvements like getting rid of adapters.

Migration guide: https://docs.safe.global/sdk/protocol-kit/reference/migrate-to-v4

## The list of requirements

- [ ] Upgrade the protocol-kit package to v4 and do the necessary migration

usame-algan:
## What it solves

Resolves #3750

## ToDos

- [x] Fix failing tests
- [x] Test that unchecked sdk still works
- [x] Mock RPC calls in unit-tests

## How this PR fixes it

- Removes `EthersAdapter` from the code and uses `SafeProvider` instead

## How to test it

All the flows that require the protocol-kit need to be tested
- Creating, signing and executing transactions
- Execute transaction via nested safe
- Creating safes and counterfactual safes
- Signing on-chain messages
- Batch and bulk execution
- Spending limit, add owner
- Relaying and tx simulation

## Checklist


mike10ca:
@usame-algan , we use protocol kit in our Cypress tests, in happy path: `sendfunds_connected_wallet.cy.js, sendfunds_queue_1.cy.js, sendfunds_relay.cy.js`

These are happy path tests and are not included into Smoke tests and won’t be there. Please have a look at the tests for breaking changes and make adjustments. 

You can run them locally off your branch inside Cypress client. I tried running a tests and got this first error right away (enclosed). 

Screenshot 2024-06-19 at 19 00 26

usame-algan:
> These are happy path tests and are not included into Smoke tests and won’t be there. Please have a look at the tests for breaking changes and make adjustments.

Thanks for noticing! I will take a look and update those as well.

src/features/counterfactual/hooks/useDeployGasLimit.ts
```diff
@@ -92,6 +92,7 @@ export const estimateBatchDeploymentTransaction = async (
   // 2. Add a simulate call to the predicted SafeProxy as second transaction
   const transactionDataToEstimate: string = simulateTxAccessorContract.encode('simulate', [
     safeTransaction.data.to,
+    // @ts-ignore
     safeTransaction.data.value,
```
usame-algan, line 96: It complains that safeTransaction.data.value needs to be bigint instead of string. I am not sure why this change is suddenly necessary or why it worked before. We could try changing it and see if it still works.

src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -95,15 +73,15 @@ export const initSafeSDK = async ({
   }
 
   if (undeployedSafe) {
-    return Safe.create({
-      ethAdapter: createReadOnlyEthersAdapter(provider),
+    return Safe.init({
+      provider: provider._getConnection().url,
```
usame-algan, line 77: Looking into this further I think whenever we call `_getConnection()` it actually does a fetch request. I will try to find another way to get the URL.

src/components/new-safe/create/logic/index.ts
```diff
@@ -112,9 +107,10 @@ export const encodeSafeCreationTx = async ({
   chain,
 }: SafeCreationProps & { chain: ChainInfo }) => {
   const readOnlySafeContract = await getReadOnlyGnosisSafeContract(chain, LATEST_SAFE_VERSION)
-  const readOnlyProxyContract = await getReadOnlyProxyFactoryContract(chain.chainId, LATEST_SAFE_VERSION)
-  const readOnlyFallbackHandlerContract = await getReadOnlyFallbackHandlerContract(chain.chainId, LATEST_SAFE_VERSION)
+  const readOnlyProxyContract = await getReadOnlyProxyFactoryContract(LATEST_SAFE_VERSION)
+  const readOnlyFallbackHandlerContract = await getReadOnlyFallbackHandlerContract(LATEST_SAFE_VERSION)
 
+  // @ts-ignore union type is too complex
```
katspaugh, line 113: I counted 10 ts-ignores added in this PR, are we planning to fix them?

usame-algan: This has to be fixed in the protocol-kit. ([Discussion on Slack](https://5afe.slack.com/archives/C0660KAEG49/p1718626730277609))

src/hooks/coreSDK/safeCoreSDK.ts
usame-algan: Can confirm that providing a real RPC URL for the tests does indeed make requests to that RPC so the fetches are not mocked. Need to figure this out before we can merge this.

src/features/counterfactual/hooks/useDeployGasLimit.ts
usame-algan: Tested this with calling `BigInt(safeTransaction.data.value)` which also works so I will remove the @ts-ignore

src/hooks/coreSDK/__tests__/safeCoreSDK.test.ts
```diff
@@ -137,16 +144,17 @@ describe('safeCoreSDK', () => {
           implementationVersionState: ImplementationVersionState.UP_TO_DATE,
         })
 
-        expect(sdk).toBeInstanceOf(Safe)
```
usame-algan, line 140: Had to change these tests because we can't create an actual instance of the SDK anymore. Basically, the protocol-kit creates a provider internally now that we can't mock so real RPC requests would happen. Instead we can check that the SDK was called with the parameters we expect.

src/services/tx/tx-sender/sdk.ts
```diff
@@ -134,30 +133,18 @@ export const getUncheckedSigner = async (provider: Eip1193Provider) => {
...
   const uncheckedJsonRpcSigner = new UncheckedJsonRpcSigner(signer.provider, await signer.getAddress())
-  const sdk = getAndValidateSafeSDK()
-
-  const ethAdapter = new EthersAdapter({
-    ethers,
-    signerOrProvider: uncheckedJsonRpcSigner,
-  })
 
-  return sdk.connect({ ethAdapter })
+  return sdk.connect({ provider, signer: uncheckedJsonRpcSigner.address })
```
usame-algan, line 141: We need to test if this actually does anything now since we are not passing the unchecked signer but just the signer address. Maybe there is a way to pass an "unchecked provider".

An easy way to test this is to execute a tx with a nested safe. The Tx modal should close immediately.

usame-algan:
One thing we should decide is how to access the RPC URL.

- We can derive it from a `JsonRpcProvider` by calling `provider._getConnection().url` -> This is how its currently done and convenient because we can use `useWeb3ReadOnly` to access it
- Call `const chain = useCurrentChain()` and `getRpcServiceUrl(chain.rpcUri)` in order to access the url

Imo `_getConnection()` seems like its intended to only be used internally so it might not be as stable/robust.

src/services/contracts/safeContracts.ts
```diff
@@ -80,65 +98,61 @@ export const _getMinimumMultiSendCallOnlyVersion = (safeVersion: SafeInfo['versi
...
-  })
+  return getMultiSendCallOnlyContractInstance(_getValidatedGetContractProps(safeVersion).safeVersion, safeProvider)
 }
 
 // GnosisSafeProxyFactory
 
-export const getReadOnlyProxyFactoryContract = (chainId: string, safeVersion: SafeInfo['version']) => {
-  const ethAdapter = createReadOnlyEthersAdapter()
+export const getReadOnlyProxyFactoryContract = (safeVersion: SafeInfo['version']) => {
+  const safeSDK = getSafeSDK()
```
schmanu, line 115: Isn't the SafeSDK usually `undefined` here as well (as in `getReadonlyFallbackHandlerContract`)?  
We call this function during Safe creation where we have no context of a Safe yet.

usame-algan: Nice catch! It was indeed not working e.g. for the gas estimation during safe creation. I created a helper function `getSafeProvider` and use that instead which we can also mock more easily in our tests.

schmanu (approved):
We def have to run a bigger regression on this but it looks good now.

katspaugh (approved):
Kudos Usame! 🦾

francovenica:
Tested a ton of different tx, signing messages and safe creation. I even run automation on this branch (a few test failed, but they have nothing to do with tx signing, so they can be dismissed)

I found a behavior that is different from dev or prod. It involves nested safe, where the tx stays in the modal with the "sign" button loading indefinitely

Steps:
SafeOwned - 2/x policy, owned by another safe and an EOA
SafeOwner - 1/x policy, owned by an EOA

- In SafeOwned create a send funds tx
- Sign it with the SafeOwner, the sign button goes to a "loading state"
- In the SafeOwner the tx pop's up so it can be signed by its owner, an EOA
- Instead of executing that tx right away, set it to "no, later" and sign it

In current prod/dev, what happens in the safeOwned is that the modal goes away and a green notification pop's up. 

In this PR, the modal stays there with the "Sign" button loading. The problem here also is that, even if you fully execute it in the SafeOwner, the modal still "doesn't realize" that the tx has been executed and stays in the tx modal.

Gif:
nested safes

usame-algan:
Debugged this further with @dasanra and the issue with nested safes will be fixed with the viem migration on the protocol-kit. In the meantime I suggest to add a workaround.

For this I created two new functions `prepareTxExecution` and `prepareApproveTxHash` which are almost identical to `sdk.executeTransaction` and `sdk.approveTransactionHash` but instead of calling the contract to do the execution which would use ethers `sendTransaction` and block they return the encoded transaction. We can then execute this on our side through the `Eip1193Provider` by calling `provider.request({ method: 'eth_sendTransaction', params: [encodedTx] })`

@schmanu could you have another look?

@francovenica I've touched three flows with this change that we need to test again.

- normal transaction execution
- nested safe transaction execution
- normal transaction signing
- nested safe transaction signing
- normal transaction speed up
- nested safe transaction speed up

These should not block the UI for the user anymore.

francovenica:
@usame-algan 

I got a problem.

Having a 2/x safe, being another safe oner of that first one. If there is a tx proposed with already 1 signature, when I try to sign&execute with the safe as owner, the prompt never shows up in the safe owner so the EOA can sign that tx.

Steps:
SafeOwned 2/x, owned by SafeOwner and a EOA.
SafeOwner 1/x owned by an EOA
EOA creates a tx and signes in SafeOwned. So there is a 1/2 safe in the SafeOwned in queue
Switch to SafeOwner. Try to sign and execute that transaction (So pick "Yes, execute" in the modal)
Wait for the modal to pop up in the SafeOwner, so the EOA can execute that transaction.

What happens is that the pop up in the safeOwner never shows up.

image

francovenica:
Looks good now

Note: At some point I had an issue with nested safes, where the tx modal would not popUp on the safe that is signing the tx, but I couldn't point out why it happened (I even made a comment that I removed because I realized that the steps I described were not correct)
I don't want to stop the whole ticket for something that happened a few times during a specific amount of time.
I'd say lets put this ticket in the RC and during the regression I'll check again

usame-algan:
> Note: At some point I had an issue with nested safes, where the tx modal would not popUp on the safe that is signing the tx, but I couldn't point out why it happened (I even made a comment that I removed because I realized that the steps I described were not correct)

Thanks @francovenica. I will give it a quick test to see if I can encounter this issue as well and merge the PR otherwise.

PR #3848, opened 2024-06-18
Fix: full token amount in tx details

katspaugh:
## What it solves

Resolves https://github.com/5afe/safe-support/issues/464

A user reported a regression after #3807. Token amounts should not be abbreviated in tx details.

## How this PR fixes it
It makes an exception for token amounts displayed in tx details to show the full number up to 30 digits. If more than 30, it will be still abbreviated.

After the fix:

Screenshot 2024-06-18 at 21 16 43


compojoom:
On the test site I see this:
grafik
No idea what we are sending and doesn't look formatted?

src/components/common/TokenAmount/index.tsx
```diff
@@ -12,16 +12,18 @@ const TokenAmount = ({
...
   decimals?: number
   logoUri?: string
   tokenSymbol?: string
   direction?: TransferDirection
   fallbackSrc?: string
+  maxLength?: number
 }): ReactElement => {
   const sign = direction === TransferDirection.OUTGOING ? '-' : ''
-  const amount = decimals !== undefined ? formatVisualAmount(value, decimals) : value
+  const amount = decimals !== undefined ? formatVisualAmount(value, decimals, undefined, maxLength) : value
```
schmanu, line 26: `formatVisualAmount` always drops amounts `< 0.00001` and just shows "< .00001".
We could consider giving the TokenAmount component a `precise: boolean` prop and use `formatAmountPrecise` instead.
That's what we use within the `SendAmountBlock` component.
This would make it consistent to what we show in the Review step of transaction creation.

katspaugh: Good point, `formatAmountPrecise` works for both big and small numbers and is consistent with  `SendAmountBlock` 👍

schmanu (commented):
The code is looking good, but the e2e history tests expect different values right now.

schmanu (approved):

liliya-soroka:
Verified

PR #3851, opened 2024-06-19
Feat: Show warning for address poisoning transactions

### LINKED ISSUE #3832: Implement Address Poisoning detection in history
CGW implement the detection on the backend. On the frontend, we need to display the detected transactions as per the designs here: https://www.notion.so/safe-global/Address-poisoning-0761b7f531534d3c80b41863ba48be78?d=fe664d66c89a40b1988a5f66bc010dcd

jmealy:
## What it solves

Resolves #3832

## How this PR fixes it
Check for transactions that have been flagged as imitations of previous legitimate transactions. If they are flagged, display a warning on the tx in the history.

## How to test it
- Open a safe with malicious imitation transactions, (https://address_poisoning--walletweb.review.5afe.dev/transactions/history?safe=eth:0xf0ddc6435b47d161d46f51f2ea7c5f32f90b98be)
- See that the imitation transactions have the warning displayed, as in the designs:
https://www.figma.com/design/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=11151-49354&t=gfp1I4vZwEzFOo7i-1

## Screenshots
image

## Checklist


usame-algan (approved):
👍

francovenica:
I got some question regarding things that are in the designs and are not here, and I don't see a discussion where it was decided not to be implemented. 
If it was decided that those were not going to be implemented then it's fine, I'm just asking for clarification

- I don't see the "Don't show again" when you copy an address. 
image

- I tried sending tokens to a poisoned address to see if a warning popped up in the tx signing, but I didn't see any. I saw a comment in the design that we might not implement it, but I don't see if it was decied or not.
image


Another question regarding this specific case:
It seems here that some USDC were sent somewhere, and 2 malicious tx were created in the history as well. What I don't get is why one has the red warning label and the other only the "!" icon with a tooltip. what is the difference between these to show the warning differently?
Tx nonce 21:
https://address_poisoning--walletweb.review.5afe.dev/transactions/history?safe=eth:0xf0ddc6435b47d161d46f51f2ea7c5f32f90b98be
image

A suggestion (see if it can be implemented or we can discuss it in another ticket if there are doubts on if it should be implemented or not):
We show a warning when the user tries to even copy the address. I think that we should also warn the user if he tries to save the address in the address book, using the "..." buttons.
image

jmealy:
> * I don't see the "Don't show again" when you copy an address.
For the don't show again checkbox, we discussed it in standup this morning and decided it would be better to be safe and show the warning every time. Unless we get complaints about it being annoying then we can leave it out. @TanyaEfremova 

> * I tried sending tokens to a poisoned address to see if a warning popped up in the tx signing, but I didn't see any. I saw a comment in the design that we might not implement it, but I don't see if it was decied or not.
For the warning when creating a tx, this is an idea for displaying blockaid warnings, so out of scope for this feature.

> Another question regarding this specific case: It seems here that some USDC were sent somewhere, and 2 malicious tx were created in the history as well. What I don't get is why one has the red warning label and the other only the "!" icon with a tooltip. what is the difference between these to show the warning differently? Tx nonce 21: https://address_poisoning--walletweb.review.5afe.dev/transactions/history?safe=eth:0xf0ddc6435b47d161d46f51f2ea7c5f32f90b98be 
These two transactions do look the same so I don't know why one was flagged and the other not. @iamacook do you know what the difference is here?

> A suggestion (see if it can be implemented or we can discuss it in another ticket if there are doubts on if it should be implemented or not): We show a warning when the user tries to even copy the address. I think that we should also warn the user if he tries to save the address in the address book, using the "..." buttons.
I like this idea 👍  And yes, makes sense as a separate ticket I think

TanyaEfremova:
1. I like the idea of showing the warning every time, let's omit the checkbox. This is an outdated design, and is not implemented now, so let's keep it this way.
4. Also a good suggestion! Let's reuse the modal, I will prepare the prototype.

francovenica:
1 - Noted, no checkbox for the copy address modal
2 - I'll make a ticket after this one is approved
3 - It was noted in [THIS](https://5afe.slack.com/archives/C03DAGWJCR5/p1719321051035429) thread that the scenario where 2 tx might show different warnings is a limitation on how this was implemented in the cgw side, and it seems it's going to be unavoidable, so nothing to fix here.

question: I tried also to see if this warning message would popup in a transaction when dealing with suspicious addresses (the ones that you get a warning if you try to copy them) but I was never able to make it show up.
For what I see in the code it wasn't implemented. Is this correct?
image

jmealy:
>question: I tried also to see if this warning message would popup in a transaction when dealing with suspicious addresses (the ones that you get a warning if you try to copy them) but I was never able to make it show up.
For what I see in the code it wasn't implemented. Is this correct?

That's right, the scope for this PR is just the warning on transactions and not for addresses specifically. Using the current detection on the gateway for this you'd need to search through the entire tx history to see if an address was involved in an address poisoning attack, which isn't practical. I don't know what the plan is for this warning when sending tokens exactly but my understanding is that the detection would be coming from blockaid, not from our own detection on the gateway. @TanyaEfremova could you give more context on this maybe?

francovenica:
Leaving this comment after merging

I think the implementation on our side is fine, I see the warning just fine in the tx but I'm still effy about how the cgw detects these tx.
Some of them have the icon of "dubious token" and even copying the address gives you the warning of the tx being an address involved in a tx with a suspicious token, but the warning is not there, so is difficult for me to predict which type of tx should have the red warning message and which one not.
I'm aware of a document that describes how this tx are flagged, but that's something the end user will not see or read, so it has to be obvious  from the UI itself if a tx is suspicious or not

Take an example the swap in nonce 418 of this safe:
https://address_poisoning--walletweb.review.5afe.dev/transactions/history?safe=sep:0x8f4A19C85b39032A37f7a6dCc65234f966F72551

Here you have the icon of dubious token, and if you try to copy the address you get the warning, but there is no warning message.

TanyaEfremova:
The warning should come from our side, Blockaid would detect it differently. We should still warn users in one way or another about a malicious address they may be interacting with. 

Only transactions that interact with the poisoning address should have that warning. If we detect them in the tx list, can we also detect the same way in the transaction summary?

src/components/transactions/TxDetails/TxData/Transfer/index.tsx
```diff
@@ -51,6 +54,7 @@ const TransferTxInfo = ({ txInfo, txStatus, trusted }: TransferTxInfoProps & { t
           <TransferActions address={address.value} txInfo={txInfo} trusted={trusted} />
         </EthHashInfo>
```
schmanu, line 55: @jmealy 
I know this is already closed but the trusted flag is used for address poisoning as well here.

I think we should pass the new imitation flag also into the `EthHashInfo` so we get the "Before you copy" modal if someone tries to copy the address.
```ts
        <EthHashInfo
          address={address.value}
          name={address.name}
          customAvatar={address.logoUri}
          shortAddress={false}
          hasExplorer
          showCopyButton
          trusted={trusted || imitation}
        >
 ```

jmealy: Good spot, thanks!

PR #3853, opened 2024-06-20
feat: Add method to Sign in with Ethereum [SW-28]

usame-algan:
## What it solves

Adds a new function `signInWithEthereum` that can be used to prompt the user to sign an EIP-4361 message and request an access_token from the CGW that is stored as a cookie and can be used in subsequent requests via the `credentials: include` header.

## How to test

You can add a new button and trigger it manually e.g.

```ts
const SignInButton = () => {
    const provider = useWeb3()

    if (!provider) return null

    return <button onClick={() => signInWithEthereum(provider)}>Sign in</button>
}
```

## Checklist


jmealy (approved):

src/services/siwe/index.ts
```diff
@@ -0,0 +1,58 @@
...
+import type { BrowserProvider } from 'ethers'
+
+const GATEWAY_URL = IS_PRODUCTION || cgwDebugStorage.get() ? GATEWAY_URL_PRODUCTION : GATEWAY_URL_STAGING
+
+/**
+ * Prompt the user to sign in with their wallet and set an access_token cookie
+ * @param provider
+ */
+async function signInWithEthereum(provider: BrowserProvider) {
+  const { nonce } = await fetch(`${GATEWAY_URL}/v1/auth/nonce`, {
```
katspaugh, line 12: We should move both fetches to the CGW SDK.

usame-algan:
The lint issues will be fixed after https://github.com/safe-global/safe-wallet-web/pull/3865 is merged

PR #3854, opened 2024-06-20
Fix: ignore nonce for risk scanning requests

katspaugh:
## What it solves

Tx nonce is updated in real-time when creating a new transaction. At first it takes the Safe nonce, then a recommended nonce is fetched and the safeTx object is updated with that.

This triggers two risk scanning requests where the only difference is the nonce.

## How this PR fixes it

Since the risk API doesn't seem to care about nonces, I made it always send 0, as well as memoizing the payload as JSON.

Screenshot 2024-06-20 at 11 45 40

## How to test it

On prod (so without this fix):
* Go to Swaps
* Create a swap of any two tokens
* Open the console and filter Network Requests by `risk`
* Confirm the swap to see the Safe Tx confirmation modal
* Observe two `/messages` requests

On this branch:
* Only one `/messages` request

### DIFF
--- a/src/components/tx/security/redefine/useRedefine.ts
+++ b/src/components/tx/security/redefine/useRedefine.ts
@@ -80,22 +80,33 @@ export const useRedefine = (
   const [retryCounter, setRetryCounter] = useState(0)
   const isFeatureEnabled = useHasFeature(FEATURES.RISK_MITIGATION)
 
+  // Memoized JSON data to avoid unnecessary requests
+  const jsonData = useMemo(() => {
+    if (!data) return ''
+    let adjustedData = data
+    if ('data' in data) {
+      // We need to set nonce to 0 to avoid repeated requests with an updated nonce
+      adjustedData = { ...data, data: { ...data.data, nonce: 0 } }
+    }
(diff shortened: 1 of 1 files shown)


src/components/tx/security/shared/TxSecurityContext.tsx
```diff
@@ -34,7 +34,12 @@ export const TxSecurityContext = createContext<{
 
 export const TxSecurityProvider = ({ children }: { children: JSX.Element }) => {
   const { safeTx, safeMessage } = useContext(SafeTxContext)
-  const [redefineResponse, redefineError, redefineLoading] = useRedefine(safeTx ?? safeMessage)
+  const data = useMemo(
```
schmanu, line 37: Should we move overwriting the `nonce` into the `useMemo` that creates `jsonData`?
Reason being that while Redefine does not care about the nonce - other providers probably do (as nonce 0 makes txs revert). Then the overwriting happens in `useRedefine` which is handling all redefine specific stuff.

liliya-soroka:
Verified.
The regression run is also passed.
 one failed case related to the tooltip for remove owner can be ignored ( not related to the fix)

PR #3855, opened 2024-06-20
Fix: adjust approval editor with swaps

katspaugh:
## What it solves
A few UI adjustments for the Approval Editor in swap transactions.

* The approval editor is now shown below Swap details
* The entire Unlimited amount block is clickable (becomes editable)
* The tracking event on click is now differentiated by label: `edit` or `save`.

### Before
Screenshot 2024-06-20 at 15 06 12

### After
Screenshot 2024-06-20 at 15 02 16

## Links
Internal discussion: https://5afe.slack.com/archives/C06GBBW39QU/p1718885834196179

### DIFF
--- a/src/components/tx/ApprovalEditor/EditableApprovalItem.tsx
+++ b/src/components/tx/ApprovalEditor/EditableApprovalItem.tsx
@@ -32,7 +32,6 @@ const EditableApprovalItem = ({
   const handleSave = () => {
     onSave()
     setReadOnly(true)
-    MODALS_EVENTS.EDIT_APPROVALS
   }
 
   const handleEditMode = () => {
--- a/src/components/tx/ApprovalEditor/EditableApprovalItem.tsx
+++ b/src/components/tx/ApprovalEditor/EditableApprovalItem.tsx
@@ -32,6 +32,7 @@ const EditableApprovalItem = ({
   const handleSave = () => {
     onSave()
     setReadOnly(true)
+    MODALS_EVENTS.EDIT_APPROVALS
   }
 
   const handleEditMode = () => {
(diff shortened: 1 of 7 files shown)


src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -109,7 +105,11 @@ export const SignOrExecuteForm = ({
           decodedDataLoading={decodedDataLoading}
           showMultisend={!props.isBatch}
           showToBlock={props.showToBlock}
-        />
+        >
+          <ErrorBoundary fallback={<div>Error parsing data</div>}>
```
compojoom, line 109: why is this now inside of the decoded TX?

katspaugh: To place it under the swap into but above multisend details like it was for other tx types.
Alternatively we move it all the way down

src/components/tx/ApprovalEditor/EditableApprovalItem.tsx
```diff
@@ -32,6 +32,7 @@ const EditableApprovalItem = ({
   const handleSave = () => {
     onSave()
     setReadOnly(true)
+    MODALS_EVENTS.EDIT_APPROVALS
```
schmanu, line 35:
```suggestion
```

src/components/tx/SignOrExecuteForm/index.tsx
schmanu: Yeah its a bit surprising. Also with rendering the children if the decoded data does not exist.
I would not block it and will still approve but I think at some point we should maybe let decodedData only decode the transaction and move the SwapOrder details outside of it and render it separately. Then we have smaller components. I am also not sure if we need all those conditions within DecodedTx.
We suppress the `SendToBlock` for instance just to render our own `SendToBlock` (which looks identical) within the Swap order.

schmanu (approved):

src/components/tx/SignOrExecuteForm/index.tsx
katspaugh: Agree! I didn’t want to do a big refactoring in this scope but we can do it as a follow up.

liliya-soroka:
Reload makes a new view ащк for the approver worse than the current approval position on prod

https://github.com/safe-global/safe-wallet-web/assets/338622/0fc90ae8-b706-4b00-8a03-16daf00c9ee4

src/hooks/useDecodeTx.ts
```diff
@@ -23,13 +23,17 @@ const useDecodeTx = (
...
+  >(
+    () => {
+      if (!encodedData || isEmptyData) {
+        const nativeTransfer = isEmptyData && !isRejection ? getNativeTransferData(tx?.data) : undefined
+        return Promise.resolve(nativeTransfer)
+      }
+      return getConfirmationView(chainId, safeAddress, encodedData, tx.data.to)
+    },
+    [chainId, encodedData, isEmptyData, tx?.data, isRejection, safeAddress],
+    false,
```
katspaugh, line 35: This fixes the reload glitch.

schmanu (approved):

liliya-soroka:
Verified

PR #3861, opened 2024-06-21
Feat: TWAP order decoding in tx history/queue [SW-1]

compojoom:
## What it solves
- [x] Adds decoding of twaps on history 
- [x] Adds decoding of twaps in queue

Resolves #
https://www.notion.so/safe-global/76f265123c0a4ad680947ce489804342?v=17c30f9503cd4d48a622209408de56d6&p=0f273a0f06f2465589628c7299f44e05&pm=s

## How to test it
Create a twap order throw the native swap feature. 

## Screenshots
grafik

## Checklist

### DIFF
--- a/src/components/common/Table/styles.module.css
+++ b/src/components/common/Table/styles.module.css
@@ -7,14 +7,14 @@
 }
 
 .gridEmptyRow {
-    display: grid;
-    grid-template-columns: 35% auto;
-    gap: var(--space-1);
-    justify-content: flex-start;
-    max-width: 900px;
-    margin-top: var(--space-1);
-    margin-bottom: var(--space-1);
-    border-top: 1px solid var(--color-border-light);
+  display: grid;
+  grid-template-columns: 35% auto;
+  gap: var(--space-1);
+  justify-content: flex-start;
+  max-width: 900px;
+  margin-top: var(--space-1);
+  margin-bottom: var(--space-1);
+  border-top: 1px solid var(--color-border-light);
 }
 .title {
   color: var(--color-primary-light);
--- a/src/components/common/Table/styles.module.css
+++ b/src/components/common/Table/styles.module.css
@@ -6,6 +6,16 @@
   max-width: 900px;
 }
 
+.gridEmptyRow {
+    display: grid;
+    grid-template-columns: 35% auto;
+    gap: var(--space-1);
+    justify-content: flex-start;
+    max-width: 900px;
+    margin-top: var(--space-1);
+    margin-bottom: var(--space-1);
+    border-top: 1px solid var(--color-border-light);
+}
 .title {
   color: var(--color-primary-light);
   font-weight: 400;
--- a/src/components/transactions/TxInfo/index.tsx
+++ b/src/components/transactions/TxInfo/index.tsx
@@ -103,7 +103,7 @@ const SwapTx = ({ info }: { info: Order }): ReactElement => {
         <TokenIcon logoUri={sellToken.logoUri || undefined} tokenSymbol={sellToken.symbol} />
       </Box>
       <Typography component="span" fontWeight="bold">
-        {formatVisualAmount(sellAmount, sellToken.decimals)} {sellToken.symbol} {" "}
+        {formatVisualAmount(sellAmount, sellToken.decimals)} {sellToken.symbol}{' '}
       </Typography>
     </>
   )
--- a/src/components/transactions/TxInfo/index.tsx
+++ b/src/components/transactions/TxInfo/index.tsx
@@ -10,7 +10,7 @@ import type {
 import { SettingsInfoType } from '@safe-global/safe-gateway-typescript-sdk'
 import TokenAmount from '@/components/common/TokenAmount'
 import {
-  isCoWOrderTxInfo,
+  isOrderTxInfo,
   isCreationTxInfo,
   isCustomTxInfo,
   isERC20Transfer,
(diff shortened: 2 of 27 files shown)


src/hooks/useTransactionType.ts
```diff
@@ -70,6 +70,12 @@ export const getTransactionType = (tx: TransactionSummary, addressBook: AddressB
         text: orderClass === 'limit' ? 'Limit order' : 'Swap order',
       }
     }
+    case 'TwapOrder': {
+      return {
+        icon: '/images/common/swap.svg',
+        text: 'Twap order',
```
katspaugh, line 76:
```suggestion
        text: 'Swap order (TWAP)',
```

src/components/common/Table/styles.module.css
```diff
@@ -6,6 +6,16 @@
...
 
+.gridEmptyRow {
+    display: grid;
+    grid-template-columns: 35% auto;
+    gap: var(--space-1);
+    justify-content: flex-start;
+    max-width: 900px;
+    margin-top: var(--space-1);
+    margin-bottom: var(--space-1);
+    border-top: 1px solid var(--color-border-light);
```
katspaugh, line 17: Please run `yarn prettier`, this should be 2 spaces not 4.

src/components/transactions/TxInfo/index.tsx
```diff
@@ -91,23 +92,45 @@ const MultiSendTx = ({ info }: { info: MultiSend }): ReactElement => {
...
+      <Typography component="span" fontWeight="bold">
+        {buyToken.symbol}
+      </Typography>
+    </>
+  )
+  if (!isSellOrder) {
+    // switch them around for buy order
+    ;[from, to] = [to, from]
+  }
+
```
katspaugh, line 125: Should we move this component to `features/swaps` and import here?

src/components/transactions/TxInfo/index.tsx
```diff
@@ -144,7 +167,7 @@ const TxInfo = ({ info, ...rest }: { info: TransactionInfo; omitSign?: boolean;
     return <CreationTx info={info} />
   }
 
-  if (isSwapTxInfo(info)) {
+  if (isCoWOrderTxInfo(info)) {
```
katspaugh, line 170: I noticed we mix the CoW and Swaps terminology in the code (probably due to what CGW sends us?). This might shoot us in the foot once we have other types of swaps, e.g. on chains that CoW doesn't support.

src/features/swap/components/SwapOrder/index.tsx
```diff
@@ -25,103 +31,243 @@ import {
...
+
+  if (receiver && receiver !== safeAddress) {
+    return (
+      <DataRow key="Recipient" title="Recipient">
+        <EthHashInfo address={receiver} showAvatar={false} />
+      </DataRow>
+    )
+  }
+
+  return <></>
```
katspaugh, line 184: This also needs a key, otherwise you'll get a warning.

src/components/transactions/TxInfo/index.tsx
compojoom: No, I went for it because we have 3 kind of orders:
swap, limit & twap.

swap is actually a limit order with price set to maket, but this is another story.

swap & limit orders have one structure
TWAP orders have a slightly different structure

When we only had swap and limit -> isSwapTxInfo was basically applying to those 2 order types. But now that I have Twaps I still want to treat this as an order and than later decide what to show and what not. That's why I went for a isCoWOrder...

Right now we are basically forwarding the data from CoWSwap. If at one point we have a different provider for a chain, then we will most probably have to come with our own format and CGW would have to convert the data to our format. But we are not there yet...

Do you have better ideas?

src/components/common/Table/styles.module.css
compojoom: we'll we need this in our git hooks and also a check for this on CI, then. 

css was missing in my prettier conifg
grafik

src/components/transactions/TxInfo/index.tsx
katspaugh: I would just use `swap, limit & twap` as prefixes.

src/hooks/useTransactionType.ts
compojoom: No, designs say TWAP order
https://www.figma.com/design/VyA38zUPbJ2zflzCIYR6Nu/Swap?node-id=6655-24390&t=UftNhqyLfN8CZdnb-4

src/components/transactions/TxInfo/index.tsx
compojoom: on an API level there is no difference betwen swap and limit. Also the difference between swap,limit and twap is so small that I don't want to do:
isSwapOrder || isLimitOrder || isTwapOrder...

All those orders are basically managed through the same component at the end.

katspaugh (approved):

francovenica:
For me it throws an error right where the type of tx is not recognized and won't allow me to execute. Given this I can see the tx in the queue (which it loosk fine btw) but not in the history tab.

image

The safe:
https://feat_twaps--walletweb.review.5afe.dev/transactions/queue?safe=sep:0x8f4A19C85b39032A37f7a6dCc65234f966F72551

Is a sepolia safe with only 1 owner
Is a tx that was queued first and I'm trying to execute now. I didn't had issues signing the tx in the first place.
I had to check the checkbox to update the fallbackhandler during the TWAP creation, so "SetFallbackHandler" is also in the bulk for execution

compojoom:
@francovenica - I was able to reproduce & fix the bug you encountered: https://github.com/safe-global/safe-wallet-web/pull/3865

PR #3862, opened 2024-06-24
Tests: Fix regression tests

mike10ca:
## How this PR fixes it

- Add minor fixes to regression tests

## How to test it

### DIFF
--- a/cypress/e2e/regression/remove_owner.cy.js
+++ b/cypress/e2e/regression/remove_owner.cy.js
@@ -25,12 +25,10 @@ describe('Remove Owners tests', () => {
     owner.verifyRemoveBtnIsEnabled().should('have.length', 2)
   })
 
-  //TODO: Clarify correct behaviour for non-owner button visibility
-  it.skip('Verify remove owner button is disabled for Non-Owner', () => {
-    cy.visit(constants.setupUrl + staticSafes.SEP_STATIC_SAFE_1)
+  it('Verify remove button does not exist for Non-Owner when there is only 1 owner in the safe', () => {
+    cy.visit(constants.setupUrl + staticSafes.SEP_STATIC_SAFE_3)
     main.waitForHistoryCallToComplete()
-    owner.waitForConnectionStatus()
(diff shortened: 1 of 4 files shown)


cypress/e2e/regression/remove_owner.cy.js
```diff
@@ -25,7 +25,8 @@ describe('Remove Owners tests', () => {
     owner.verifyRemoveBtnIsEnabled().should('have.length', 2)
   })
 
-  it('Verify Tooltip displays correct message for Non-Owner', () => {
+  //TODO: Clarify correct behaviour for non-owner button visibility
+  it.skip('Verify remove owner button is disabled for Non-Owner', () => {
```
liliya-soroka, line 29: what an issue do we have currently with the case? Not possible to disconnect signer?

cypress/e2e/regression/swaps_history.cy.js
```diff
@@ -36,16 +36,14 @@ describe('[SMOKE] Swaps history tests', () => {
...
-    create_tx.clickOnTransactionItemByName('11:14 AM')
+    create_tx.clickOnTransactionItemByName('14')
     create_tx.verifyExpandedDetails([
       swapsHistory.sellOrder,
       swapsHistory.sell,
       swapsHistory.oneCOW,
       swapsHistory.forAtLeast,
       swapsHistory.dai,
       swapsHistory.filled,
-      swapsHistory.actionApprove,
```
liliya-soroka, line 47: why do we remove the check for actions?

cypress/e2e/regression/remove_owner.cy.js
mike10ca: This is now updated as discussed.

cypress/e2e/regression/swaps_history.cy.js
mike10ca: This is now updated as discussed.

liliya-soroka (approved):

Issue #3323, opened 2024-02-23
Optimize CGW requests


katspaugh:
## Problem

We're doing a lot of CGW requests on each page load, as well as polling data.

Here's a full list: https://kinopio.club/cgw-requests-per-page-f4bmV--EJjhXwzaEQAa-D

Let's see what we can optimize.

## Solutions

* Some calls can be avoided altogether
* Some calls can be delayed until that data is actually needed (e.g. tx history)
* Some repeated calls should be cached (e.g. Safe App list)

katspaugh:
Resolved in various PRs.

PR #3864, opened 2024-06-24
Refactor: lazy-load governance widget

katspaugh:
## What it solves

Part of #3323

The governance widget makes a few requests to CGW from below the fold. This PR defers the rendering of this widget until the user scrolls to it.

### DIFF
--- a/src/components/dashboard/GovernanceSection/GovernanceSection.tsx
+++ b/src/components/dashboard/GovernanceSection/GovernanceSection.tsx
@@ -1,10 +1,6 @@
-import { useEffect, useRef, useState } from 'react'
-import { Typography, Card, Box, Alert, IconButton, Link, SvgIcon } from '@mui/material'
+import { useCallback, useEffect, useRef, useState } from 'react'
+import { Typography, Card, Box, Link, SvgIcon } from '@mui/material'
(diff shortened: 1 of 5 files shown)


src/components/dashboard/GovernanceSection/GovernanceSection.tsx
```diff
@@ -85,73 +82,76 @@ const GovernanceSection = () => {
...
+    window.addEventListener('scroll', handleScroll)
+    return () => window.removeEventListener('scroll', handleScroll)
   }, [])
 
-  if (!isLayoutReady) {
-    return null
-  }
+  return (
+    <>
+      {hasScrolled && <InfiniteScroll onLoadMore={onVisible} />}
```
compojoom, line 134: Why do you need hasScrolled? Can't we just render the infinitescroll all the time?

katspaugh: The page is small in the beginning and this block is always in the viewport initially.

compojoom (approved):
👍

francovenica:
LGTM. 
The widget doesn't load anymore as soon as you enter in the dashboard (since the widget is visible for a second), instead it loads as you scroll down.

PR #3865, opened 2024-06-24
Feat: TWAP decoding on confirmation screen [SW-2]

compojoom:
known issue:
- won't build due to the use of a wrong typescirpt-sdk

## What it solves
Ads decoding for twap transactions on the confirm transaction screen

## How to test it
Create a twap order with the native swap. On the confirmation screen the decoding of the order should be displayed and it should be according to the designs.

## Screenshots
grafik

## Checklist


src/components/tx/DecodedTx/index.tsx
```diff
@@ -59,7 +59,7 @@ const DecodedTx = ({
 }: DecodedTxProps): ReactElement | null => {
   const chainId = useChainId()
   const chain = useCurrentChain()
-  const isSwapOrder = isSwapConfirmationViewOrder(decodedData)
+  const isSwapOrder = isCoWConfirmationViewOrder(decodedData)
```
katspaugh, line 62: Why rename this?

compojoom: because isCoWConfirmationViewOrder can contain both a swap and twap order.

src/features/swap/components/SwapOrderConfirmationView/index.tsx
```diff
@@ -107,8 +117,32 @@ export const SwapOrderConfirmationView = ({ order, settlementContract }: SwapOrd
...
       />
+
+      {isTwapOrder && (
+        <div className={css.partsBlock}>
+          <DataTable
+            rows={[
+              <Typography key="title" variant="body1" className={css.partsBlockTitle}>
+                <strong>
+                  Order will be split in{' '}
+                  <span className={css.numberOfPartsLabel}>${order.numberOfParts} equal parts</span>
```
usame-algan, line 128:
```suggestion
                  <span className={css.numberOfPartsLabel}>{order.numberOfParts} equal parts</span>
```
Screenshot 2024-06-25 at 14 56 18

src/features/swap/components/SwapOrderConfirmationView/index.tsx
```diff
@@ -107,8 +117,32 @@ export const SwapOrderConfirmationView = ({ order, settlementContract }: SwapOrd
...
+                </strong>
+              </Typography>,
+              <PartSellAmount order={order} addonText="per part" key="sell_part" />,
+              <PartBuyAmount order={order} addonText="per part" key="buy_part" />,
+              <DataRow title="Start time" key="Start time">
+                {order.startTime.startType === StartTimeValue.AtMiningTime ? 'At mining time' : 'At block number'}
```
usame-algan, line 134: Not sure how useful this information is for the user. In the CoW review screen it just says `Start time first part: Now` which is easier to understand imo.

src/features/swap/components/SwapOrder/index.tsx
```diff
@@ -236,20 +229,10 @@ export const TwapOrder = ({ order }: { order: SwapTwapOrder }) => {
...
-            {formatVisualAmount(partSellAmount, sellToken.decimals)} {sellToken.symbol}
-          </Typography>
-        </DataRow>,
-        <DataRow title="Buy amount" key="buy_amount_part">
-          <Typography component="span" fontWeight="bold">
-            {formatVisualAmount(minPartLimit, buyToken.decimals)} {buyToken.symbol}
-          </Typography>
-        </DataRow>,
+        <PartSellAmount order={order} key="part_sell_amount" />,
+        <PartBuyAmount order={order} key="part_buy_amount" />,
```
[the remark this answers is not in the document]
compojoom: those components are part of an array. If the elements don't have keys you get some key warnings...

src/features/swap/components/SwapOrderConfirmationView/index.tsx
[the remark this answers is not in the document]
compojoom: Yes, you can't have 1 order for 100 and a second one for 200. At least at the moment.

package.json
```diff
@@ -58,7 +58,7 @@
     "@safe-global/protocol-kit": "^3.1.1",
     "@safe-global/safe-apps-sdk": "^9.1.0",
     "@safe-global/safe-deployments": "^1.36.0",
-    "@safe-global/safe-gateway-typescript-sdk": "^3.21.5",
+    "@safe-global/safe-gateway-typescript-sdk": "3.21.6",
```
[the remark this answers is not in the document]
compojoom: well, this was not intended, but maybe a good thing. 
Normally I install it with @latest, which would add ^ before the version number. But this time because 3.21.6 was missing from npm and I ended up with 3.21.5, I tried installing it with @3.21.6 to see if the version is really missing or not... And once it was available on npm we ended up with "3.21.6".

With that said - as you can see an update to the sdk could brake other people's PR's because of type incompatibilities, so maybe it is better to be explicit here and specify the exact version?

src/features/swap/components/SwapOrderConfirmationView/index.tsx
compojoom: As I wrote on the designs - we need to come with something better than "now", because "now" in my view means nothing. It can be now, it can be tomorrow - it is at the moment this tx gets minted. This needs to be clear to the user.

usame-algan: Imo `Now` is easy to understand because it means as soon as the transaction is executed whereas differentiating between mining and block number is quite technical. Where is the difference between those two?

compojoom: It cannot be "now" if the transaction is not submitted, right? If I have n/m safe it's not now. It's a point of time in the future :)

usame-algan (approved):

francovenica:
I was asked to take a look at this ticket now before being in Ready for QA.
I tested this on Sepolia since is the only network I have enough funds to even trigger a TWAP

When I created a TWAP order for the first time, I had to use the checkbox that edits the fallbackhandler in the CowSwap form.
The tx at that moment looked like this instead on how the design looks

image

Once I've executed this tx I tried again, with the correct fallbackhandler now in place, the 2nd tx actually had the decoding expected in this ticket, but there are still some issues:

image

- The title is not "TWAP" but "Twap order". Maybe is trying to be consistent with the other types of orders (swap and limit), but in this case it should be at least "TWAP" all in uppercase
- The text in the decoding is centered instead of being aligned left. I assume again that is to be consistent with the other types of orders. If that's the case then the designs should be updated @TanyaEfremova 
- Check the alignment of the text above with the text below. I draw a line to see how the texts are not aligned.
- In the designs, the "actions" of the tx are displayed. Again, probably they are not there to be consistent with the other orders so once again probably the designs needs to be updated and remove them

I didn't execute this transaction, I just queued it first and then I went to execute it in the queue. When I did that the form looks nothing like it looked before: 
- No decoding at all, 
- The title looks messy, with the name of the order and the tokens being swap tossed there not aligned
Also in my case an error won't allow me to execute the tx. I didn't see this error while I was creating the tx, so maybe is there an issue with proposed transactions that are already signed?
image

compojoom:
> When I created a TWAP order for the first time, I had to use the checkbox that edits the fallbackhandler in the CowSwap form.
> The tx at that moment looked like this instead on how the design looks

Did you maybe use the production gateway instead of staging? I just tested it again and for me it looks fine:

grafik

compojoom:
> The title looks messy, with the name of the order and the tokens being swap tossed there not aligned
> Also in my case an error won't allow me to execute the tx. I didn't see this error while I was creating the tx, so maybe is there an issue with proposed transactions that are already signed?

This should now be fixed

francovenica:
@compojoom Did you try a safe where you have to enable the update of the fallbackhandler? When I had to do that for the 1st tx it would look without any decoding. The 2nd tx (once the fallbackhandler was updated) the decoding was showing properly. 
I used stg CGW, but since I was able to see the decoding in my 2nd attempt I assume that's not a factor.

francovenica:
I can execute the tx now, so that's fixed.

The issue I reported at the beggining is still there:
It seems that, when you upgrade fallbackhandler along the creation of the TWAP, the decoding in the tx form doesn't show as expected.
I tried with Prod CGW this time

image
image

compojoom:
I tested with setting the fallback handler - it works on my end. We fixed something on the gateway yesterday that might have caused the issue for you. Please try again.

usame-algan:
Just tested using TWAP on a new safe that doesn't have the fallback handler yet. This is the view I am getting which looks correct to me. @francovenica it looks like you are not seeing the decoded data. Could you share what the transaction-confirmation network request returns for you?
localhost_3000_swap_safe=sep_0x361A14639600e025321893158437E15a30D6CE9A (2)

compojoom:
> The title looks messy, with the name of the order and the tokens being swap tossed there not aligned

Not sure how this got there. Fix it. Hope that it doesn't break something else, but tested with swap and limit and it looks fine.

compojoom:
> The title is not "TWAP" but "Twap order". Maybe is trying to be consistent with the other types of orders (swap and limit), but in this case it should be at least "TWAP" all in uppercase
I fixed that

> The text in the decoding is centered instead of being aligned left. I assume again that is to be consistent with the other types of orders. If that's the case then the designs should be updated @TanyaEfremova
I talked to Tania, I reduced the margin between the elements, also moved them a bit more to the left. Now they are closer to the figma design.

> Check the alignment of the text above with the text below. I draw a line to see how the texts are not aligned.
That's ok. They cannot be vertically aligned as the parts are inside of the other table so it is expected that there is few px missmatch.

francovenica:
Ok, works now.
The alignments and text changing is fine as well

image

LGTM

PR #3866, opened 2024-06-25
Fix: update meta description

katspaugh:
## What it solves

The old text was outdated. New text approved by Ani.

### DIFF
--- a/src/components/common/MetaTags/index.tsx
+++ b/src/components/common/MetaTags/index.tsx
@@ -3,8 +3,7 @@ import { ContentSecurityPolicy, StrictTransportSecurity } from '@/config/securit
 import lightPalette from '@/components/theme/lightPalette'
 import darkPalette from '@/components/theme/darkPalette'
 
-const descriptionText =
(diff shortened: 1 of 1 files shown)


src/components/common/MetaTags/index.tsx
```diff
@@ -3,8 +3,7 @@ import { ContentSecurityPolicy, StrictTransportSecurity } from '@/config/securit
 import lightPalette from '@/components/theme/lightPalette'
 import darkPalette from '@/components/theme/darkPalette'
 
-const descriptionText =
-  'Safe (prev. Gnosis Safe) is the most trusted platform to manage digital assets on Ethereum and multiple EVMs. Over $40B secured.'
+const descriptionText = 'Safe{Wallet} is the most trusted smart account wallet on Ethereum. Over $100B secured.'
```
usame-algan, line 6: nit: ...wallet on Ethereum with over $100B secured.

Reads a bit nicer instead of two sentences imo.

usame-algan (approved):

PR #3868, opened 2024-06-25
Fix: wrong number formatting in SingleTxDecoded

katspaugh:
## What it solves

The wrong number formatting function was called in SingleTxDecoded. It resulted in double formatting which failed in locales where the decimal separator is a comma (e.g. pt-PT).

## How to test it

* Set your browser language to Portuguese (Portugal)
* Go to tx history and expand a multisend transaction containing fractional amount transfers
* In prod: they would be displayed as NaN
* On this branch: they are displayed correctly

### DIFF
--- a/src/components/transactions/TxDetails/TxData/DecodedData/SingleTxDecoded/index.tsx
+++ b/src/components/transactions/TxDetails/TxData/DecodedData/SingleTxDecoded/index.tsx
@@ -7,7 +7,7 @@ import {
 } from '@safe-global/safe-gateway-typescript-sdk'
 import type { AccordionProps } from '@mui/material/Accordion/Accordion'
 import { useCurrentChain } from '@/hooks/useChains'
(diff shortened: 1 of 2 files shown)


src/components/transactions/TxDetails/TxData/DecodedData/SingleTxDecoded/index.tsx
```diff
@@ -56,7 +56,6 @@ export const SingleTxDecoded = ({
 
   const addressInfo = txData.addressInfoIndex?.[tx.to]
   const name = addressInfo?.name
-  const avatarUrl = addressInfo?.logoUri
```
katspaugh, line 59: Unused variable.

jmealy (approved):

francovenica:
LGTM. Leaving here a before/after snapshot for the sake of documentation:

image

image

PR #3869, opened 2024-06-25
Feat: design improvements for Roles integration

jfschwarz:
Taking into account the feedback points from design & QA, this PR improves the integration of role-based access control into the execution flow previously delivered with #3768.

### Integrate into the main "execute or sign" card

Previously we've integrated the execution through a role as new card appearing above the default "execute or sign" card. As suggested by the wallet design team, we've now improved the user experience by enabling the execution through a role directly from the default form.

**Call is allowed for the role:**

image

**Call is not allowed:**

image

#### Logic: default execution as owner vs execution as role member

- This PR only affects creating new transactions. The confirm flow for previously proposed transaction remains unchanged.
- If the connected wallet is not an owner but a role member, we present the form for execution as role member.
- If the connected wallet is both, an owner and a role member, we present the form for execution as role member only if the threshold is >1 (no direct execution as owner possible).

### Handle edge cases with batch transactions

So far we didn't correctly handle cases where the Role mod expects a different multi-send address than the one used per default in Safe Wallet. This PR fixes it through a lookup of the multi-send address configured on the Roles mod and routing the batch transaction accordingly. 

### More user-friendly error messages for failing permission checks

Review the complete list of potential messages [here](https://github.com/safe-global/safe-wallet-web/pull/3869/files#diff-4de71d09890c621a8418ab97fd0d278831cf0959da6d668a7b0ef443beec3cdeR272-R290).


katspaugh (approved):
Looks great, thank you!

jfschwarz:
> Expected result: The option to execute via Role should appear only for the tx allowed by the role . The same as we have for the Spending limits

Thank you, @liliya-soroka, this was a little misunderstanding. Based on the mockups I assumed we want to show the permission error in that case. I agree, though, it's better to not show it at all for Safe owners but only if the connected wallet is a role member and not a Safe owner.

I just pushed a fix and updated the specs:

image

liliya-soroka:
approved from the QA side

katspaugh (approved):
Thanks @jfschwarz, amazing work!

PR #3872, opened 2024-06-26
fix: update recommendedNonce when history tag changes

schmanu:
## What it solves
Updates `useRecommendedNonce` for 1/1 Safes that execute a transaction directly. 

## How this PR fixes it
Fetched the recommended nonce when the safe's `txHistoryTag` changes.
Reason being that 1/1 Safe transactions that get executed immediately never hit the queue and thus do not update the `txQueuedTag`.

Adds a unit test for `useRecommendedNonce`

## How to test it
- Open a 1/1 Safe with an empty queue
- Create a transaction and execute it immediately
- Quicky create a second transaction and wait in the tx modal for the first one to finish
- The nonce should automatically get updated to the next one once the first tx gets indexed if the nonce has not been changed manually

## Checklist

### DIFF
--- a/src/components/tx/SignOrExecuteForm/hooks.test.ts
+++ b/src/components/tx/SignOrExecuteForm/hooks.test.ts
@@ -1,5 +1,5 @@
-import { extendedSafeInfoBuilder } from '@/tests/builders/safe'
-import { renderHook } from '@/tests/test-utils'
+import { extendedSafeInfoBuilder, safeInfoBuilder } from '@/tests/builders/safe'
+import { renderHook, waitFor } from '@/tests/test-utils'
 import { zeroPadValue } from 'ethers'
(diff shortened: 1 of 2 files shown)


src/components/tx/SignOrExecuteForm/hooks.test.ts
```diff
@@ -566,24 +575,151 @@ describe('SignOrExecute hooks', () => {
       const { result } = renderHook(() => useAlreadySigned(tx))
       expect(result.current).toEqual(true)
     })
+
+    it('should return false if wallet has not signed a tx yet', () => {
```
schmanu, line 579: This change is unrealted. But I realised that the case was not nested in the matching `describe` block.

katspaugh (approved):
👍

liliya-soroka:
Verified

PR #3877, opened 2024-06-27
Display swap info on sent/receive events [SW-3]

compojoom:
## What it solves
Displays swap info on sent/receive events

## How to test it
sent/receive events that are created in response to a swap order should display additional info about the swap

## 
grafik

## Checklist

### DIFF
--- a/src/features/swap/components/SwapOrder/index.tsx
+++ b/src/features/swap/components/SwapOrder/index.tsx
@@ -11,7 +11,6 @@ import {
   type SwapOrder as SwapOrderType,
   type Order,
   type TransactionData,
-  TransactionInfoType,
 } from '@safe-global/safe-gateway-typescript-sdk'
 import { DataRow } from '@/components/common/Table/DataRow'
 import { DataTable } from '@/components/common/Table/DataTable'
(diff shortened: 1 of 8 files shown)


src/features/swap/components/SwapOrder/index.tsx
```diff
@@ -256,13 +254,13 @@ export const TwapOrder = ({ order }: { order: SwapTwapOrder }) => {
 }
 
 export const SwapOrder = ({ txData, txInfo }: SwapOrderProps): ReactElement | null => {
```
usame-algan, line 256:
```suggestion
export const SwapOrder = ({ txInfo }: SwapOrderProps): ReactElement | null => {
```

usame-algan (approved):

francovenica:
For me the "filled" field with the green bar and % is not necesary for the "Parts transactions", since those can be filled 100% or not at all, there is no in between, since if you don't have enough funds to fullfill a part (like you send tokens from your safe to another place, leaving your safe with not enough funds) the system will tell you that the order cannot be filled until you top it up

image

So in my opinion, that field is irrelevant in the parts. 
It IS relevant tho, in the main TWAP tx (the one with nonce), since that one actually show the % of how many tx have been fullfilled

compojoom:
Thanks @francovenica ! Yes, Usame and I had the same discussion and we agree with you. I don't want to do this in this PR as we have a lot of other things on our plate. I've made an issue for https://www.notion.so/safe-global/Make-order-components-view-aware-5e1ee21d5ff145c795308524d7ed80bd and we'll tackle this in the next release.

PR #3880, opened 2024-06-28
Feat swaps fees [SW-34]

compojoom:
## What it solves
Swaps now support dynamic fees.

## How to test it
Navigate to the swaps widget - enter different amounts to sell. The applied fee should depend on the fiat amount swapped.

## Checklist

### DIFF
--- a/src/features/swap/helpers/data/stablecoins.ts
+++ b/src/features/swap/helpers/data/stablecoins.ts
@@ -2,187 +2,503 @@ export const stableCoinAddresses: {
   [address: string]: {
     name: string
     symbol: string
-    chain: 'gnosis' | 'mainnet' | 'arbitrum' | 'sepolia'
+    chains: Array<'gnosis' | 'ethereum' | 'arbitrum-one' | 'sepolia'>
   }
 } = {
   '0xdd96b45877d0e8361a4ddb732da741e97f3191ff': {
     name: 'BUSD Token from BSC',
     symbol: 'BUSD',
-    chain: 'gnosis',
+    chains: ['gnosis'],
   },
   '0x44fa8e6f47987339850636f88629646662444217': {
     name: 'Dai Stablecoin on Gnosis',
     symbol: 'DAI',
-    chain: 'gnosis',
+    chains: ['gnosis'],
   },
   '0x1e37e5b504f7773460d6eb0e24d2e7c223b66ec7': {
     name: 'HUSD on Gnosis',
     symbol: 'HUSD',
-    chain: 'gnosis',
+    chains: ['gnosis'],
   },
   '0xb714654e905edad1ca1940b7790a8239ece5a9ff': {
     name: 'TrueUSD on Gnosis',
     symbol: 'TUSD',
-    chain: 'gnosis',
+    chains: ['gnosis'],
   },
   '0xddafbb505ad214d7b80b1f830fccc89b60fb7a83': {
     name: 'USD//C on Gnosis',
     symbol: 'USDC',
-    chain: 'gnosis',
+    chains: ['gnosis'],
   },
   '0x4ecaba5870353805a9f068101a40e0f32ed605c6': {
     name: 'Tether on Gnosis',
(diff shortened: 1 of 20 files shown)


src/features/swap/helpers/data/stablecoins.ts
```diff
@@ -0,0 +1,188 @@
...
+  [address: string]: {
+    name: string
+    symbol: string
+    chain: 'gnosis' | 'mainnet' | 'arbitrum' | 'sepolia'
+  }
+} = {
+  '0xdd96b45877d0e8361a4ddb732da741e97f3191ff': {
+    name: 'BUSD Token from BSC',
+    symbol: 'BUSD',
+    chain: 'gnosis',
```
anxolin, line 11: nit: maybe it is easier to maintain and less redundant if you use a `Record<Chain, Token[]>`?

compojoom: yes! The reason I did it like this is that I didn't want to pass the chainId/chainName. 
Also, name, symbol, chain - not necessary actually. We should maybe just go with token addresses.

src/features/swap/helpers/fee.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const isStableCoin = stableCoinAddresses[sellToken?.address] && stableCoinAddresses[buyToken?.address]
+
+  const fiatAmount = Number(orderKind == 'sell' ? sellTokenFiatAmount : buyTokenFiatAmount) || 0
+
+  if (fiatAmount < 100000) {
+    return isStableCoin ? 10 : 35
+  }
+
+  if (fiatAmount < 1000000) {
```
usame-algan, line 31:
```suggestion
  if (fiatAmount < 1_000_000) {
```

nit: Its easier to read that way

src/features/swap/helpers/fee.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export const calculateFeePercentageInBps = (orderParams: OnTradeParamsPayload) => {
+  const { sellToken, buyToken, buyTokenFiatAmount, sellTokenFiatAmount, orderKind } = orderParams
+
+  const isStableCoin = stableCoinAddresses[sellToken?.address] && stableCoinAddresses[buyToken?.address]
+
+  const fiatAmount = Number(orderKind == 'sell' ? sellTokenFiatAmount : buyTokenFiatAmount) || 0
+
+  if (fiatAmount < 100000) {
+    return isStableCoin ? 10 : 35
```
usame-algan, line 28: Do you think it makes sense to have an enum for these magic numbers? e.g. `StableCoinFee.High`

src/features/swap/index.tsx
```diff
@@ -69,19 +76,63 @@ const SwapWidget = ({ sell }: Params) => {
...
+      alert: palette.warning.main,
+    },
+    partnerFee: {
+      bps: 35,
+      recipient: SWAP_FEE_RECIPIENT,
+    },
+    content: {
+      feeLabel: 'Widget Fee',
+      feeTooltipMarkdown:
+        'The [tiered widget fee](https://help.safe.global/en/articles/178530-how-does-the-widget-fee-work-for-native-swaps) incurred here and charged by CoW DAO for the operation of the CoW Swap Widget is automatically calculated into this quote. It will contribute to a license fee that supports the Safe Community. Neither the Safe Ecosystem Foundation nor Safe (Wallet) operate the CoW Swap Widget and/or CoW Swap.',
```
usame-algan, line 133: This doesn't seem to work (including formatted links into the tooltip)
Screenshot 2024-07-01 at 11 21 48

usame-algan (approved):

src/features/swap/index.tsx
compojoom: Yes, I've reported this to CoW on friday

src/features/swap/helpers/fee.ts
compojoom: good point!

PR #3882, opened 2024-07-01
fix: Display cow fallback handler message [SW-33]

usame-algan:
## What it solves

Resolves [SW-33](https://www.notion.so/safe-global/CoW-s-fallback-handler-99044779a28d46cfb69552103bee6e68)

## How this PR fixes it

- Checks if the fallback handler is cows fallback handler for TWAPs and displays a message instead of the unofficial fallback handler warning
- Removes the check icon and adjusts the spacing according to Figma design

## How to test it

1. Open a Safe that has TWAP enabled
2. Go to Settings -> Modules
3. Observe a green message above the fallback handler address

## Screenshots
Screenshot 2024-07-01 at 16 50 58

## Checklist

### DIFF
--- a/src/components/settings/FallbackHandler/index.tsx
+++ b/src/components/settings/FallbackHandler/index.tsx
@@ -1,3 +1,4 @@
+import { TWAP_FALLBACK_HANDLER } from '@/features/swap/helpers/utils'
 import NextLink from 'next/link'
 import { Typography, Box, Grid, Paper, Link, Alert } from '@mui/material'
 import semverSatisfies from 'semver/functions/satisfies'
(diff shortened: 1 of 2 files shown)


compojoom:
@usame-algan can you ask infra to add title and icon as in the figma design?

usame-algan:
> @usame-algan can you ask infra to add title and icon as in the figma design? Or we hardcode all of those in the code?

Good idea! I asked infra to add the fallback handler address to the known contract list.

compojoom (approved):

francovenica:
LGTM
Looks like this right now, so I assume infra did give it the proper icon an title
image

The icon is different than in the designs, but I think CowSwap changed it recently so I think is ok:
image

TanyaEfremova:
@francovenica CoW changed their branding slightly, so I will need to update their logos in our design system. The difference is OK for now.

Could we add rounded corners to the icons? E.g. 4px should be fine

PR #3883, opened 2024-07-01
fix: correctly format native transfer amounts

schmanu:
## What it solves
On some locales native values are displayed as NaN.

## How this PR fixes it
Always uses `safeFormatUnits` instead of `formatVisualAmount` before passing the value to `SendAmountBlock`

## How to test this PR
1. On firefox open the tx builder Safe app
2. Enter any EOA address as `to` address
3. Enter a big amount (e.g. 123456789 - your Safe does not need to actually hold this amount as we just want to see the decoding)
4. Witness `NaN` on prod but now the correct amount in this PR deployment

## Checklist

### DIFF
--- a/src/components/tx/DecodedTx/index.tsx
+++ b/src/components/tx/DecodedTx/index.tsx
@@ -33,7 +33,6 @@ import ExternalLink from '@/components/common/ExternalLink'
 import { HelpCenterArticle } from '@/config/constants'
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
 import accordionCss from '@/styles/accordion.module.css'
-import { safeFormatUnits } from '@/utils/formatters'
 import SendAmountBlock from '@/components/tx-flow/flows/TokenTransfer/SendAmountBlock'
 import { ZERO_ADDRESS } from '@safe-global/protocol-kit/dist/src/utils/constants'
 
@@ -75,7 +74,7 @@ const DecodedTx = ({
 
   if (!decodedData) return null
 
-  const amount = tx?.data.value ? safeFormatUnits(tx.data.value, chain?.nativeCurrency.decimals) : '0'
+  const amount = tx?.data.value ?? '0'
 
   return (
     <Stack spacing={2}>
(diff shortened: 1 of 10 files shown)


src/components/tx/DecodedTx/index.tsx
```diff
@@ -75,7 +75,7 @@ const DecodedTx = ({
 
   if (!decodedData) return null
 
-  const amount = tx?.data.value ? formatVisualAmount(tx.data.value, chain?.nativeCurrency.decimals) : '0'
+  const amount = tx?.data.value ? safeFormatUnits(tx.data.value, chain?.nativeCurrency.decimals) : '0'
```
usame-algan, line 78: I've tested this and can confirm that for large numbers it fixes the `NaN` issue but it seems to also change the formatting for smaller numbers e.g. sending `468651  ETH` via transaction builder results in `468,651` if `formatVisualAmount` is used and `468.651` if `safeFormatUnits` is used. My `navigator.language` is set to `de`.

usame-algan:
A more general observation: It is not clear what amount format components like `SendAmountBlock` expect. For example if I want to send `1 ETH`, the component expects the value to be `1` and not `1` with 18 zeroes but this is not clear from looking at its props. Imo we should find a way to define this either via types or naming so that regressions are less likely to happen.

src/components/tx/DecodedTx/index.tsx
schmanu: I will refactor it and comment that `SendAmountBlock` always expects the WEI amount. And then move all formatting inside there.

usame-algan (approved):
Nice solution, thanks!

francovenica:
I'll need more information here. 

There was a ticket not long ago where for the Portuguese language some amounts would look like the NaN because of the " , " or " . " in the numbers. 
Is this a similar case? is this for a specific language number formatting?

katspaugh:
@francovenica yes, it's the same thing. Manu found more places where the wrong function was called.

schmanu:
> I'll need more information here.
> 
> There was a ticket not long ago where for the Portuguese language some amounts would look like the NaN because of the " , " or " . " in the numbers. Is this a similar case? is this for a specific language number formatting?

@francovenica I added steps to reproduce.
It is related to the same thing, but in different places.

Here is a screenshot in our current prod after the described steps.
Screenshot 2024-07-15 at 11 05 19

francovenica:
Ok, Thanks for the explanation. never tried such numbers in the tx builder, is good to know I can do that.
I was able to reproduce the issue in prod and check that here the fix works fine.

With the fix:
image

LGTM

PR #3889, opened 2024-07-02
feat: add a warning if changing the fallback handler [SW-53]

compojoom:
## What it solves
If the user is submitting a TWAP order for the first time he would most probably need to change his safe's fallback handler. Whenever we detect this we now show a warning:
grafik

Resolves #

## How to test it
Try to create a TWAP order on a safe that has the standard fallback handler. The first tx should display the warning in the screenshot.

## Checklist

### DIFF
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -29,6 +29,8 @@ import useChainId from '@/hooks/useChainId'
 import PermissionsCheck from './PermissionsCheck'
 import { isConfirmationViewOrder } from '@/utils/transaction-guards'
 import SwapOrderConfirmationView from '@/features/swap/components/SwapOrderConfirmationView'
+import { isSettingTwapFallbackHandler } from '@/features/swap/helpers/utils'
+import { TwapFallbackHandlerWarning } from '@/features/swap/components/TwapFallbackHandlerWarning'
 
 export type SubmitCallback = (txId: string, isExecuted?: boolean) => void
 
@@ -80,6 +82,7 @@ export const SignOrExecuteForm = ({
 
   const { safe } = useSafeInfo()
   const isCounterfactualSafe = !safe.deployed
+  const isChangingFallbackHandler = isSettingTwapFallbackHandler(decodedData)
 
   // If checkbox is checked and the transaction is executable, execute it, otherwise sign it
   const canExecute = isCorrectNonce && (props.isExecutable || isNewExecutableTx)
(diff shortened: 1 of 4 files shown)


src/features/swap/helpers/utils.ts
```diff
@@ -173,3 +173,18 @@ export const isOrderPartiallyFilled = (
 
   return BigInt(executedSellAmount) !== 0n && executedSellAmount < sellAmount
 }
+
+export const isSettingTwapFallbackHandler = (decodedData: DecodedDataResponse | undefined) => {
+  const isCow =
```
usame-algan, line 178: nit: This variable assignment is not needed since the function name is already descriptive enough.

src/features/swap/components/TwapFallbackHandlerWarning/index.tsx
```diff
@@ -0,0 +1,14 @@
+import { Alert, Box, SvgIcon } from '@mui/material'
+import InfoOutlinedIcon from '@/public/images/notifications/info.svg'
+
+export const TwapFallbackHandlerWarning = () => {
+  return (
+    <Box mt={1} mb={1}>
+      <Alert severity="warning" icon={<SvgIcon component={InfoOutlinedIcon} inheritViewBox color="error" />}>
+        <b>Enable TWAPs and submit order .</b>
```
usame-algan, line 8:
```suggestion
        <b>Enable TWAPs and submit order.</b>{' '}
```

usame-algan: My IDE actually auto-corrects this to `<b>Enable TWAPs and submit order.</b>  To enable TWAP orders you need to set a custo...` which also seems to work.

src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -100,6 +103,8 @@ export const SignOrExecuteForm = ({
       <TxCard>
         {props.children}
 
+        {isChangingFallbackHandler && <TwapFallbackHandlerWarning />}
```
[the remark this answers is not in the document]
compojoom: Yes, I thought about this, but think that SignOrExecute is the better place. It's possible for a user to change the fallback handler to CoW's one outside of an Order and I still wanted to show the warning. That's why it is here.

src/features/swap/helpers/utils.ts
```diff
@@ -173,3 +173,18 @@ export const isOrderPartiallyFilled = (
...
+    decodedData?.parameters?.some((item: any) =>
+      item.valueDecoded?.some(
+        (decoded: any) =>
+          decoded.dataDecoded?.method === 'setFallbackHandler' &&
+          decoded.dataDecoded.parameters?.some(
+            (parameter: any) => parameter.name === 'handler' && parameter.value === TWAP_FALLBACK_HANDLER,
+          ),
```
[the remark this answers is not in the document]
compojoom: good point!

usame-algan (approved):
Thanks!

PR #3891, opened 2024-07-02
fix: address poison copy confirmation modal

schmanu:
## What it solves
We currently do not show the "Before you copy" confirmation modal for transactions using real tokens to perform address poisoning.

## How this PR fixes it
Show the Copy modal warning for transactions that are detected as imitations but use trusted tokens.

## How to test it
Open a Safe with an address poisoning attempt using a real token (e.g. USDC)

## Checklist


jmealy (approved):
Thanks for adding the tests!

francovenica:
We tested it with Manu, since finding a real address that had the issue to see it was difficult. I cannot share the address of the safe  since it is private, but I can ensure that, for tx where you receive real tokens from obviously dubious addresses you see the "are you sure" modal before copying such address
image

PR #3892, opened 2024-07-02
Feat: Do not show swaps feature to users from blocked counties [SW-48]

jmealy:
## What it solves
We need to block access to the swaps feature for users on the OFAC list of sanctioned countries. 

## How this PR fixes it
- make a request to the /swap endpoint when the app loads. If there is a 403 error, set a flag.
- If this flag is set, do not show the entry points to the swap feature (sidebar, dashboard banner, etc.)
- If the user visits the swaps page directly, redirect to a 403 error page.

## How to test it
- You can use https://app.requestly.io/ to make the swaps endpoint return a 403 for testing.
- As a blocked users there should be no links to the swaps page. It should be removed from the sidebar, assets page, dashboard, safe apps etc.
- As a blocked user, visiting the swaps page directly should show a 403 error page.

## Screenshots
image

## Checklist


compojoom (approved):
Looks good to me

PR #3895, opened 2024-07-03
Fix buy order [SW-50]

compojoom:
## What it solves
We were displaying wrong amount in the queue and history when the user was making a buy order.

grafik

## How to test it
1. Create a buy order
2. in queue and history you should see "{sell token icon} to {buy token icon} {amount buy token}"
grafik

## Checklist


usame-algan (approved):
Looks good 👍

PR #3901, opened 2024-07-03
fix: Show DateTime tooltip even for older transactions

### LINKED ISSUE #3115: Tx history - Date tooltips don't show after using pagination
## Bug description
In the tx history, if you hover the date a tooltip showing a more accurate date is displayed. This tooltip only works in the first page, if you start to scroll, the new items that show after using the pagination don't have that tooltip

## Environment
 - Browser: Chrome
 - Wallet: Any
 - Chain: Ethereum mainnet
 - Env: Current prod build

## Steps to reproduce
 1. Go to this safe that has a lot of tx's
 2. Hover the date on the first element and check the tooptip shows
 3. Scroll to load new elements on the list
 4. Hover the date in those new elements

## Expected result
The tooltip is displayed

## Obtained result
The Tooltip doesn't show up

## Screenshots
The ones that don't show the tooltip are elements loaded by pagination
tooltips

usame-algan:
## What it solves

Resolves #3115

Transactions are displayed either with a **timestamp in hours**, a **datetime** or **time in words**

timestamp in hours
-> In the tx history and only if no filter is applied

datetime
-> In the tx queue if older than 60 days
-> In the tx history if a filter is applied

time in words
-> In the tx queue if within 60 days

The tooltip contains the datetime so it should only be visible if either timestamp in hours is shown or time in words

## How this PR fixes it

- Adjusts the condition to include the case where a transaction is older than 60 days but where the timestamp in hours is still shown

## How to test it

1. Open the tx history
2. Scroll enough to get to transactions older than 60 days
3. Hover the timestamp
4. Observe a tooltip with datetime showing

## Screenshots
Screenshot 2024-07-03 at 15 22 53

## Checklist

### DIFF
--- a/src/components/common/DateTime/DateTime.tsx
+++ b/src/components/common/DateTime/DateTime.tsx
@@ -9,8 +9,10 @@ type DateTimeProps = {
 }
 
 export const DateTime = ({ value, showDateTime, showTime }: DateTimeProps): ReactElement => {
+  const showTooltip = !showDateTime || showTime
+
   return (
-    <Tooltip title={showDateTime ? '' : formatDateTime(value)} placement="top">
+    <Tooltip title={showTooltip && formatDateTime(value)} placement="top">
       <span>{showTime ? formatTime(value) : showDateTime ? formatDateTime(value) : formatTimeInWords(value)}</span>
     </Tooltip>
   )


jmealy (approved):
looks good 👍

francovenica:
LGTM. Tooltips show when the should now

PR #3906, opened 2024-07-04
chore: update terms and conditions request [SW-57]

compojoom:
## What it solves
We have to ask all our users to reconfirm the cookies and terms and conditions. 

## How to test it
Open the page you should immediately see the cookie and terms prompt. There is no checkbox for terms, but it is implied that by using the page they are accepted

## Checklist


src/components/common/CookieAndTermBanner/index.tsx
```diff
@@ -72,8 +74,11 @@ export const CookieBanner = ({
         <Grid container alignItems="center">
           <Grid item xs>
             <Typography variant="body2" mb={2}>
-              By clicking &quot;Accept all&quot; you agree to the use of the tools listed below and their corresponding
-              cookies. <ExternalLink href={AppRoutes.cookie}>Cookie policy</ExternalLink>
+              By browsing this page, you accept our{' '}
+              <ExternalLink href={AppRoutes.terms}>Terms & Conditions</ExternalLink> and the use of necessary cookies.
+              By clicking &quot;Accept al &quot; or &quot;Save settings &quot;, you additionally agree to the use of
```
schmanu, line 79:
```suggestion
              By clicking &quot;Accept all &quot; or &quot;Save settings &quot;, you additionally agree to the use of
```

schmanu, line 79: I feel like this sounds as if "Accept all" and "Save settings" would both accept all options.
I was confused by that text and would drop the Save settings part:
`By clicking "Accept all" you additionally agree to the use of Beamer and Analytics cookies as listed below.`

schmanu (changes requested):
There is for sure a small typo but in general I was a bit confused by how it is phrased now.

schmanu (approved):

PR #3907, opened 2024-07-04
Fix: correct filtering of history for txs marked trusted and imitation [SW-100]

jmealy:
## What it solves
If there is a transaction marked with `trusted: true` and `imitation: true`, then it does not get filtered with the `hide malicious` toggle in the transaction history page.

## How this PR fixes it
- Specify the `imitation` query param for the `/history` endpoint, based on the value of the `hide malicious` toggle button on the transaction history page.
- Also rename the `hide malicious` toggle in user settings to reflect the new name in the UI, and differentiate from the `trusted` flag on transactions.


src/hooks/useTxHistory.ts
```diff
@@ -20,9 +20,9 @@ const useTxHistory = (
   // The latest page of the history is always in the store
   const historyState = useAppSelector(selectTxHistory)
   const [filter] = useTxFilter()
-  const { showOnlyTrustedTransactions } = useAppSelector(selectSettings)
+  const { hideSuspiciousTransactions } = useAppSelector(selectSettings)
   const hasDefaultTokenlist = useHasFeature(FEATURES.DEFAULT_TOKENLIST)
-  const onlyTrusted = (hasDefaultTokenlist && showOnlyTrustedTransactions) || false
+  const hideSuspicious = (hasDefaultTokenlist && hideSuspiciousTransactions) || false
```
schmanu, line 25:
```suggestion
  const hideSuspicious = (hideSuspiciousTransactions) || false
```

Same here

src/services/transactions/index.ts
```diff
@@ -18,13 +18,15 @@ export const getTxDetails = memoize(
   (id: string, chainId: string) => `${chainId}-${id}`,
 )
 
-export const getTxHistory = (chainId: string, safeAddress: string, trusted = false, pageUrl?: string) => {
+export const getTxHistory = (chainId: string, safeAddress: string, hideSuspicious = false, pageUrl?: string) => {
```
schmanu, line 21: We should maybe have two flags for trusted and imitation that we pass here.

The trusted filtering requires token lists to be available. But imitations don't.

jmealy: Created separate flags here and also in fetchFilteredTxHistory

schmanu (approved):

jmealy:
I noticed an issue with the filtering, but it is pre-existing so I created a separate ticket for it: https://github.com/safe-global/safe-wallet-web/issues/4005

liliya-soroka:
Could you explain me the impact on UI for the users, please?

liliya-soroka:
Verified

PR #3939, opened 2024-07-12
Fix: prompt -> confirm

katspaugh:
## What it solves

We were showing the wrong type of the native confirmation popup.

image (5)


compojoom (approved):

liliya-soroka:
could I get more details , because I am not sure where and what to test from the description?

katspaugh:
@liliya-soroka this is a fix for the bug you mentioned here: https://5afe.slack.com/archives/C03DAGWJCR5/p1720687961207999

When a dapp connected over WalletConnect wants to switch the chain.

liliya-soroka:
Verified 
image

PR #3942, opened 2024-07-12
Tests: fix tx deletion

katspaugh:
## What it solves

This fixes the Private Key signer wallet in e2e tests, specifically transaction deletion.

JSON parse wasn't necessary.


mike10ca:
@katspaugh , this does not seem to work. I tried to delete a tx from the queue and got the error. See screenshot. 
tx

PR #3945, opened 2024-07-12
fix: Append timestamp to static files for cache busting

### LINKED ISSUE #3944: Don't cache pages and resources on new deployment
## What is the problem

When deploying a new version of the app, old files are still referenced for some users e.g. if they kept the app open before deployment which can lead to unexpected behaviour.

## Steps to reproduce

Need to figure out the root cause of this issue. Some assumptions:

- Next PWA could cache old files
- The HTML files keep their filename between deployments and could reference old chunks

usame-algan:
## What it solves

Resolves #3944

## How this PR fixes it

- Appens the current timestamp to all static files on build step

## How to test it

1. Inspect network requests of the deployed PR
2. Observe static files contain a timestamp in their filename

## Checklist

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -34,7 +34,7 @@ const nextConfig = {
   experimental: {
     optimizePackageImports: ['@mui/material', '@mui/icons-material', 'lodash', 'date-fns', '@sentry/react', '@gnosis.pm/zodiac'],
   },
-  webpack(config) {
+  webpack(config, {dev}) {
     config.module.rules.push({
       test: /\.svg$/i,
       issuer: { and: [/\.(js|ts|md)x?$/] },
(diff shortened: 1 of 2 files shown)


usame-algan:
I tried extracting the `version` from the package.json inside `next.config.mjs` but kept getting the following error: `Error: EISDIR: illegal operation on a directory, open '/web-core/.next/static/chunks/pages/apps'`

katspaugh:
Aren't chunk names just hashes of the contents of these files? If the hash didn't change, the cache shouldn't be busted for that chunk.

usame-algan:
> Aren't chunk names just hashes of the contents of these files? If the hash didn't change, the cache shouldn't be busted for that chunk.

You are right. By default next.js uses webpacks `contenthash` so unless the contents of the file changed the filename will stay the same. If its not the chunks themselves I assume it has to do with the service worker that caches both chunks and also HTML files. I will look into this next.

usame-algan:
I will close this for now as its unlikely that the static files are the issue but instead the HTML page files are cached and need to be refreshed on deployment.

PR #3946, opened 2024-07-15
feat: Add MetaMask SDK to onboard.js

### LINKED ISSUE #3841: Web3onboard metamask
## What is the feature about
Adapt web3onboard/metamask: https://onboard.blocknative.com/docs/wallets/metamask

Benefits:
- Strong Connection Health: Persistent connections with MetaMask mobile app for up to 30 days via a new async communication layer.
- Natively supports EIP6963.

If we use it, in our connect modal users would see metamask:
grafik

usame-algan:
## What it solves

Resolves #3841

## How this PR fixes it

- Adds the MetaMask SDK as a wallet module to onboard.js

## How to test it

1. Open Safe and try to connect with MetaMask
2. The MM extension should open normally

- Test what happens when no MM extension is installed
- Test what happens on mobile with MM app installed and without
-> iOS: If no MM app is installed it opens the app store page of MM
-> iOS: If MM app is installed it opens it and connects

## Screenshots

No extension installed -> User sees the mobile QR code to scan and a dialog to download the extension
Screenshot 2024-07-15 at 12 04 47
Screenshot 2024-07-15 at 12 04 51

## Checklist


src/hooks/wallets/wallets.ts
```diff
@@ -38,7 +39,17 @@ const walletConnectV2 = (chain: ChainInfo) => {
   })
 }
 
+const metamaskSDKWallet = metamaskSDK({
+  options: {
+    extensionOnly: true,
+    dappMetadata: {
+      name: 'Safe',
```
katspaugh, line 46:
```suggestion
      name: 'Safe Wallet',
```

src/hooks/wallets/wallets.ts
```diff
@@ -38,7 +39,17 @@ const walletConnectV2 = (chain: ChainInfo) => {
...
+    },
+  },
+})
+
 const WALLET_MODULES: Partial<{ [key in WALLET_KEYS]: (chain: ChainInfo) => WalletInit }> = {
+  [WALLET_KEYS.METAMASK]: () => metamaskSDKWallet,
```
katspaugh, line 52: Dis not a function call like other modules?

katspaugh:
Looks like the bundle analyzer bot is giving fake news on this one. It clearly adds lots of dependencies down the supply chain, and should increase the bundle size, not decrease it.

usame-algan:
> Looks like the bundle analyzer bot is giving fake news on this one. It clearly adds lots of dependencies down the supply chain, and should increase the bundle size, not decrease it.

True, bundlephobia says its 1.3kb minified so it should increase unless onboard.js is doing some optimizations in the background in case the sdk is added 🤔

src/hooks/wallets/wallets.ts
usame-algan: It is I just extracted it because we are passing a rather large options object otherwise it would look like this:

```
[WALLET_KEYS.METAMASK]: () =>
    metamaskSDK({
      options: {
        extensionOnly: true,
        dappMetadata: {
          name: 'Safe Wallet',
        },
      },
    }),
```

usame-algan:
I am running into an issue when trying to connect to my mobile MM. @compojoom @liliya-soroka could you try and see if it works for you?

compojoom:
I tested it on mobile. For me it was slow. First attempt worked fine, then tried to sign - took few seconds to do so. Then disconnected and tried to reconnect again - it failed. Tried to connect a second time by just rescanning the QR, now I see "Connecting to Metamask..." 

On refresh of our UI the popup for connection to metamask (in the metamask app) comes up - I don't have to scan the QR code. Very weird. If this is the superior WC experience I'm not convinced :D

src/hooks/wallets/wallets.ts
katspaugh: I would keep it consistent with the other modules.

PR #3947, opened 2024-07-15
Fix: remove "could not coalesce error" prefix from error messages

### LINKED ISSUE #3935: "could not coalesce error" error on UI is shown when balance=0 in the trezor wallet
## Bug description
 "could not coalesce error" error on UI is shown when balance=0 in the trezor wallet
## Environment
 - Browser: Chrome
 - Wallet: Trezor
 - Chain: any

## Steps to reproduce
 1. connect Trezor signer with balance=0
 2. try to execute tx using such signer
## Expected result
The user should be informed that the balance is 0 and this is the reason why execution is not possible 
## Obtained result
image

katspaugh:
## What it solves

Resolves #3935

## How this PR fixes it

Ethers adds `could not coalesce error` in the beginning of error messages which isn't useful to the user. This PR simply cuts it off.

### DIFF
--- a/src/components/tx/ErrorMessage/index.tsx
+++ b/src/components/tx/ErrorMessage/index.tsx
@@ -5,6 +5,8 @@ import WarningIcon from '@/public/images/notifications/warning.svg'
 import InfoIcon from '@/public/images/notifications/info.svg'
 import css from './styles.module.css'
 
+const ETHERS_PREFIX = 'could not coalesce error'
+
 const ErrorMessage = ({
   children,
   error,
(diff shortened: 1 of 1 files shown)


src/components/tx/ErrorMessage/index.tsx
```diff
@@ -46,7 +48,7 @@ const ErrorMessage = ({
 
           {error && showDetails && (
             <Typography variant="body2" className={css.details}>
-              {error.reason || error.message.slice(0, 300)}
+              {(error.reason || error.message).replace(ETHERS_PREFIX, '').trim().slice(0, 500)}
```
usame-algan, line 51: Do you remember the reason why we sliced the message at 300 characters? Is it possibly a limitation of Sentry?

katspaugh: Just to not blow it up visually.

francovenica:
Since the "could not coalesce error" error is easy to reproduce and since this ticket only changes the way the error message is display I'm fine passing it now and let regression catch anything that might pop up

PR #3953, opened 2024-07-15
Feat: auto-connect to nested Safe as a Safe App

katspaugh:
## What it solves

When a nested Safe is opened as a Safe App:
* Auto-connect to the parent Safe
* Show a thick green border to differentiate the two identical UIs

## Screenshots
Screenshot 2024-07-15 at 20 43 09


src/components/dashboard/NestedSafeBanner/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const [nestedSafeOwner] = useAsync<string | null>(async () => {
+    if (!safeLoaded) return null
+
+    if (isIframe()) return null
+
+    const ownerAddresses = safe?.owners.slice(0, MAX_CHECKED_OWNERS).map((owner) => owner.value)
+
+    const infos = await Promise.allSettled(
+      ownerAddresses.map((ownerAddress) => getSafeInfo(safe.chainId, ownerAddress)),
```
katspaugh, line 65: This will make up to 3 requests to CGW each time Dashboard is loaded. 🤔

schmanu: We could fetch the (already cached) owned Safes of the current wallet instead which is just one request (or none if cached) and then do an intersection of the owners and the owned Safes.

src/components/dashboard/NestedSafeBanner/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+    )
+
+    const nestedSafe = infos.find((info) => info.status === 'fulfilled')
```
schmanu, line 68: If you have a Safe with two Safe owners where you are an owner in both, you can only connect to the first one through this. You'd move to WC for the second signature.

schmanu (commented):
While this is simple I think the UX is a bit clunky as having the same app loaded twice is quite overwhelming.
Maybe we could use a similar approach but create a ParentSafe-Onboard module. Where we wrap the actual connected wallet somehow in a Safe provider and use that as connected wallet.

src/components/dashboard/NestedSafeBanner/index.tsx
katspaugh: This would work if:
* You’re connected to a wallet
* That wallet owns the parent Safe

So a little bit less reliable but much cheaper indeed.

src/components/dashboard/NestedSafeBanner/index.tsx
katspaugh: Yes, that’s a trade off. My (not validated) assumption is that typically one Safe controls sub-acccount Safes.

katspaugh:
> While this is simple I think the UX is a bit clunky as having the same app loaded twice is quite overwhelming.
Maybe we could use a similar approach but create a ParentSafe-Onboard module. Where we wrap the actual connected wallet somehow in a Safe provider and use that as connected wallet.

That was my initial thought as well. I then realized we'd still have to show both UIs somehow. Two tx confirmtion modals, some indication of which parent Safe you're connected to, etc. Ultimately, the best experience would probably be with a browser extension.

PR #3958, opened 2024-07-16
Tests: mock fetch in PermissionsCheck tests

katspaugh:
## What it solves

The PermissionsCheck tests were doing actual fetch requests to a gas oracle which sometimes failed on CI. Fetch is now mocked to return an empty object.

### DIFF
--- a/src/components/tx/SignOrExecuteForm/PermissionsCheck/__test__/PermissionsCheck.test.tsx
+++ b/src/components/tx/SignOrExecuteForm/PermissionsCheck/__test__/PermissionsCheck.test.tsx
@@ -18,6 +18,17 @@ import { FEATURES } from '@/utils/chains'
 import { chainBuilder } from '@/tests/builders/chains'
 import { useHasFeature } from '@/hooks/useChains'
 
+// Mock fetch
+Object.defineProperty(window, 'fetch', {
+  writable: true,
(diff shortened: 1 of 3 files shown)


src/components/tx/SignOrExecuteForm/PermissionsCheck/__test__/PermissionsCheck.test.tsx
```diff
@@ -18,6 +18,17 @@ import { FEATURES } from '@/utils/chains'
...
+// Mock fetch
+Object.defineProperty(window, 'fetch', {
+  writable: true,
+  value: jest.fn(() =>
+    Promise.resolve({
+      ok: false,
+      json: () => Promise.resolve({}),
+    }),
+  ),
+})
```
usame-algan, line 30: Do we know which call ends up doing the fetch? Imo we shouldn't mock fetch but instead mock our own functions that wrap the fetch calls. It seems we are overwriting the global object, do you know if this is cleaned up after the tests run?

katspaugh: We use the same type of mock in the GasPrice tests and it seems to be fine.

usame-algan (approved):

PR #3961, opened 2024-07-17
Fix: allow keyboard scrolling for in Dialog

### LINKED ISSUE #2690: Cannot easily scroll to end on Confirm Transaction screen
## Bug description
When signing transactions with a large number of steps, the Sign button, located on the bottom of the page, becomes less accessible. While a mouse scrollwheel works, you would typically be able to use End or Cmd-Down (Mac) on the keyboard to quickly move the viewport to the bottom; however, the structure of the page does not allow this.

## Environment
 - Browser: Chrome
 - Wallet: Rabby
 - Chain: Optimism

## Steps to reproduce
 1. Create a MultiSend or other transaction that enables a scrollbar on the Confirm Transaction screen
 2. Try to quickly access the Sign button, located at the bottom of the page

## Expected result
One should be able to use End or Cmd-Down to move the viewport to the bottom of the page.

## Obtained result
The viewport does not move, despite the scrollbar.

## Screenshots
Screenshot 2023-10-24 at 10 55 07 AM

jmealy:
## What it solves
Cannot use keyboard scroll in the tx modal.
Resolves #2690

## How this PR fixes it
- Wrap the dialog contents in the DialogContent and DialogContentText components to get the default MUI modal behaviour.

## How to test it
- Create a transaction with many steps (https://github.com/safe-global/safe-wallet-web/issues/2690#issuecomment-1945821479)
- Open the tx modal. Focus on the modal (click or tab).
- Scroll using the arrow buttons should work. Scrolling to the end of the page using `CMD`+`down` or `end` should also work.

## Checklist

### DIFF
--- a/src/components/common/TxModalDialog/index.tsx
+++ b/src/components/common/TxModalDialog/index.tsx
@@ -1,6 +1,6 @@
-import { type ReactElement } from 'react'
-import { IconButton, Dialog, DialogTitle, type DialogProps } from '@mui/material'
+import { Dialog, DialogContent, DialogContentText, DialogTitle, IconButton, type DialogProps } from '@mui/material'
 import classnames from 'classnames'
+import type { ReactElement } from 'react'
 import CloseIcon from '@mui/icons-material/Close'
 import css from './styles.module.css'
 
@@ -42,8 +42,9 @@ const TxModalDialog = ({
           </IconButton>
         </div>
       </DialogTitle>
-
-      {children}
+      <DialogContent dividers={false}>
+        <DialogContentText tabIndex={-1}>{children}</DialogContentText>
+      </DialogContent>
     </Dialog>
   )
 }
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
👍

francovenica:
@jmealy 
Question:
When you hit the "confirm" button to go to the modal, the modal is not selected, so you cannot scroll until you puposedly click on it. Is there a way to focus the modal as soon as you enter without the any action from the user, so he can scroll right away with the keys?

jmealy:
> @jmealy Question: When you hit the "confirm" button to go to the modal, the modal is not selected, so you cannot scroll until you puposedly click on it. Is there a way to focus the modal as soon as you enter without the any action from the user, so he can scroll right away with the keys?

This appears to be a bug in MUI. When fullScreen is set to true on the Dialog, it doesn't allow you to scroll straight away and focusing on the Dialog contents using JS doesn't work either. I created an issue for it here: https://github.com/mui/material-ui/issues/42989

PR #3962, opened 2024-07-17
Feat: Kiln widget

compojoom:
Screenshot 2024-08-26 at 16 02 15

## Other changes

* Staking and Swaps share a lot of components, so we've extracted some of the Swaps components to generic components.
* Method call and parameters for Swaps were moved to the Advanced details to be consistent with other tx types (see https://dinky.dog/?q=tx-details_f481530497111930).
* All other tx types, where we previously weren't displaying method call and parameters at all, will have that block also in Advanced details.


src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -152,15 +160,17 @@ const AppFrame = ({ appUrl, allowedFeaturesList, safeAppFromManifest }: AppFrame
...
         </div>
 
-        <TransactionQueueBar
-          expanded={queueBarExpanded}
-          visible={queueBarVisible && !queueBarDismissed}
-          setExpanded={setExpanded}
-          onDismiss={dismissQueueBar}
-          transactions={transactions}
-        />
+        {!isNativeEmbed && (
```
compojoom, line 163: It's a bit unclear why a native Embed wouldn't have the transactionQueueBar. I remember talking to @schmanu and I think he likes the bar and even wanted to have it for the swap feature. 

Maybe a more explicit name such as "hideTransactionBar" is more appropriate? But then I see above that we are also removing the page title... and also not emitting tracking events.

schmanu: I think native swaps and native staking are both basically full screen Safe apps. So I do not see a reason to hide the queue bar if there are transactions in the queue.
Especially if some actions require multiple interactions, forcing users to switch context means that likely the widgets will load back into their original state so some context could be lost.

src/components/tx/ConfirmationOrder/TokenInfoPair/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  value: string
+  label: string
+  tokenInfo?: {
+    decimals: number
+    symbol: string
+    logoUri?: string | null
+  }
+}
+
+const TokenInfoPair = ({ blocks, showArrow }: { blocks: [InfoBlock, InfoBlock]; showArrow?: boolean }) => {
```
compojoom, line 16: That name is a small lie & you know it :)
Maybe a better name would be BlockInfoPair?

src/components/tx/DecodedTx/index.tsx
```diff
@@ -41,9 +41,9 @@ const DecodedTx = ({
   showMultisend = true,
   showMethodCall = false,
 }: DecodedTxProps): ReactElement => {
-  const isMultisend = !!decodedData?.parameters?.[0]?.valueDecoded
   const chainId = useChainId()
-  const isMethodCallInAdvanced = !showMethodCall || isMultisend
+  const isMultisend = !!decodedData?.parameters?.[0]?.valueDecoded
+  const isMethodCallInAdvanced = !showMethodCall || (isMultisend && showMultisend)
```
compojoom, line 46: The showMultisend here is confusing. Now the name is `isMethodCallInAdvanced`, but since showMultisend could be false this would also be false, despite the method call being in advanced. I'm more for havign showMultisend inside of the render.

src/features/stake/components/StakingStatus/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    text: 'Withdrawn',
+  },
+  [NativeStakingStatus.UNKNOWN]: undefined,
+}
+
+const capitalizedStatus = (status: string) =>
+  status
+    .toLowerCase()
+    .replace(/_/g, ' ')
+    .replace(/^\w/g, (l) => l.toUpperCase())
```
compojoom, line 54: isn't it easier to do capitalize with text-transform in css?

src/components/safe-apps/AppFrame/index.tsx
katspaugh: I'm not against keeping it. It's just not consistent with other pages. Why donn't we show it on all pages except for the queue itself?

src/components/tx/DecodedTx/index.tsx
katspaugh: Just to make sure we understand the same logic. We move the method name and parameters to Advanced details in two cases:
* When `showMethodCall` is set to false in props
* When it's a multi-send and `showMultisend` from props is true

How would you make this less confusing?

src/components/tx/ConfirmationOrder/TokenInfoPair/index.tsx
katspaugh: Maybe `ConfirmationOrderHeader`?

src/components/safe-apps/AppFrame/index.tsx
katspaugh: Anyway, I removed the check that hides the queue bar.

src/components/tx/DecodedTx/index.tsx
katspaugh: Let's leave it like this for now as @clovisdasilvaneto and I are refactoring this more comprehensively.

katspaugh (approved):

liliya-soroka:
@katspaugh , regression issue :
call .... on contract is missing on the swap tx details after refactoring 
image

liliya-soroka:
2. colour of the state is also changed to the different from the expected according tot he design :
image
Expected:
image

PR #3963, opened 2024-07-17
Fix: make swap icon white in dark mode [SW-81]

jmealy:
## What it solves
In the transaction summary, the swap icon is black in dark mode.

## How this PR fixes it
- Use MUI SvgIcon to render the swap icon so that it changes color according to the theme.

## How to test it
- Find or create a swap transaction in the tx history.
- See the the icon is the correct color in both light and dark mode. Make sure the other tx icons are unaffected.

## Screenshots
image

## Checklist

### DIFF
--- a/src/components/transactions/TxType/index.tsx
+++ b/src/components/transactions/TxType/index.tsx
@@ -3,6 +3,7 @@ import type { TransactionSummary } from '@safe-global/safe-gateway-typescript-sd
 import { Box } from '@mui/material'
 import css from './styles.module.css'
 import SafeAppIconCard from '@/components/safe-apps/SafeAppIconCard'
+import { isValidElement } from 'react'
 
 type TxTypeProps = {
   tx: TransactionSummary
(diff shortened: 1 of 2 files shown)


compojoom (approved):
Looks fine to me, but maybe @usame-algan wants to give his 2 cents as the SVG and CSS Master :)

usame-algan:
There are css classes in `globals.css`  like `illustration-text-primary-fill` that can be used to overwrite the color of svg icons without having to change anything in components.

src/components/transactions/TxType/index.tsx
```diff
@@ -13,13 +14,17 @@ const TxType = ({ tx }: TxTypeProps) => {
...
     <Box className={css.txType}>
-      <SafeAppIconCard
-        src={type.icon}
-        alt={type.text}
-        width={16}
-        height={16}
-        fallback="/images/transactions/custom.svg"
-      />
+      {isValidElement(type.icon) && type.icon}
+      {typeof type.icon == 'string' && (
```
usame-algan, line 18: nit: Could use an if else here since `isValidElement(string)` returns false.

usame-algan (approved):

PR #3966, opened 2024-07-17
Chore: switch to pull_request_target

compojoom:
Github actions on forks fail, because they require secrets that are not available. By relying on pull_request_target maintainers will be able to start those action workflows and they shouldn’t fail.

From the [github docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target):
This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks.

## What it solves
With this we should be able to manually run actions on github PR from forks. This will allow for unit tests, preview deployments to work.

## How to test
I think that we can only test it by first merging and then trying to create a PR from a fork.

## Checklist

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -1,7 +1,7 @@
 name: Deploy to dev/staging
 
 on:
-  pull_request:
+  pull_request_target:
 
   push:
     branches:
--- a/.github/workflows/e2e-regression.yml
+++ b/.github/workflows/e2e-regression.yml
@@ -1,7 +1,7 @@
 name: Regression tests
 
 on:
-  pull_request:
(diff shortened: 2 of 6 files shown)


compojoom:
> Very cool, thank you! What about the unit tests workflow?

I think that I mixed it up with the `lint` and that's why I didn't change it. I update the PR.

katspaugh (approved):

katspaugh:
Looks like the changed jobs got stuck in this PR. Perhaps because the changes aren't on main yet? Let's merge and see.

PR #3967, opened 2024-07-18
Feat: redesign tx details (pt 1)

katspaugh:
An incremental improvement on how we display the decoded method and contract parameters.

Based on https://www.figma.com/design/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=5794-183021&t=sKcrFPtbaWnr5jhr-0

## Screenshot
Screenshot 2024-07-17 at 15 54 35

(In part 2, we'll redesign the expanded Advanced details.)

## How to test it

* We should test all kinds of transactions (transfers, Safe App txs, multisend txs, batches, settings changes etc)
* Both how they look in the history/queue
* And when you create a tx and when you sign an existing tx

Screenshot 2024-07-23 at 10 23 36


katspaugh:
Fixing the tests.

src/components/common/Table/styles.module.css
```diff
@@ -5,6 +5,11 @@
   justify-content: flex-start;
   max-width: 900px;
   overflow-x: auto;
+  margin-bottom: var(--space-1);
```
usame-algan, line 8: According to Figma this is only supposed to be 4px

src/components/transactions/TxDetails/TxData/DecodedData/MethodCall.tsx
```diff
@@ -0,0 +1,55 @@
...
+        on
+        <NamedAddressInfo
+          address={contractAddress}
+          name={contractName}
+          customAvatar={contractLogo}
+          onlyName
+          hasExplorer
+          showCopyButton
+          avatarSize={26}
+        />
```
usame-algan, line 42: What do you think about using this component in `generateDataRowValue` if there is a name for an address to make it better fit with the rest of the layout?

Current:
Screenshot 2024-07-22 at 12 14 59

With NamedAddressInfo:
Screenshot 2024-07-22 at 12 14 35

src/components/transactions/TxDetails/TxData/DecodedData/MethodCall.tsx
```diff
@@ -0,0 +1,55 @@
...
+}: {
+  method: string
+  contractAddress: string
+  contractName?: string
+  contractLogo?: string
+}) => {
+  return (
+    <>
+      <Typography fontWeight="bold" display="flex" alignItems="center" gap=".5em" pb={1.5} component="div">
+        Call{' '}
```
usame-algan, line 18: I suggest we change the display to block or similar to make it readable on mobile:

Screenshot 2024-07-22 at 12 17 10

src/components/transactions/TxDetails/TxData/DecodedData/MethodCall.tsx
```diff
@@ -0,0 +1,55 @@
...
+        </Typography>{' '}
+        on
+        <NamedAddressInfo
+          address={contractAddress}
+          name={contractName}
+          customAvatar={contractLogo}
+          onlyName
+          hasExplorer
+          showCopyButton
+          avatarSize={26}
```
usame-algan, line 41:
```suggestion
          avatarSize={24}
```

src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
```diff
@@ -35,23 +46,22 @@ export const DecodedData = ({ txData, txInfo }: Props): ReactElement | null => {
...
+        </Box>
       )}
-      <SendToBlock
-        address={txData.to.value}
-        title="Interact with"
-        name={isCustomTxInfo(txInfo) ? txInfo.to.name : undefined}
-        avatarSize={26}
-      />
+
+      {method && <MethodCall contractAddress={toAddress} contractName={name} contractLogo={avatar} method={method} />}
```
usame-algan, line 64: In case there is an amount there is already good information visible for the user. I suggest we either hide this block in that case or move it to the top to avoid layout issues:
Screenshot 2024-07-22 at 12 21 53

usame-algan: Seeing this screenshot, similar idea like before: In case there is an address name we could use `NamedAddressInfo` inside `SendToBlock`.

src/components/common/Table/styles.module.css
```diff
@@ -5,6 +5,11 @@
...
+}
+
+.gridRow:last-of-type {
+  margin-bottom: 0;
```
usame-algan, line 12: This causes a notable jump when toggling advanced details. I suggest to invert it and use `margin-top` instead and target the first element instead of the last.

src/components/transactions/TxDetails/TxData/DecodedData/SingleTxDecoded/index.tsx
```diff
@@ -39,23 +28,19 @@ export const SingleTxDecoded = ({
...
 }: SingleTxDecodedProps) => {
-  const chain = useCurrentChain()
   const isNativeTransfer = tx.value !== '0' && (!tx.data || isEmptyHexData(tx.data))
   const method = tx.dataDecoded?.method || (isNativeTransfer ? 'native transfer' : 'contract interaction')
-  const amountInWei = tx.value ?? '0'
-
-  let details
-  if (tx.dataDecoded) {
-    details = <MethodDetails data={tx.dataDecoded} addressInfoIndex={txData.addressInfoIndex} />
-  } else if (tx.data) {
```
usame-algan, line 50: We are now checking `txData.hexData` inside `DecodedData` to render this block. Is this still the same behaviour?

src/components/transactions/TxDetails/TxData/DecodedData/MethodCall.tsx
katspaugh: I think this section _must_ display the address. In fact I would make it long.

src/components/tx/DecodedTx/index.tsx
```diff
@@ -119,63 +89,34 @@ const DecodedTx = ({
             expandIcon={<ExpandMoreIcon />}
             className={accordionCss.accordion}
           >
-            <span style={{ flex: 1 }}>Transaction details</span>
-
-            {decodedData
-              ? decodedData.method
-              : tx?.data.operation === OperationType.DelegateCall
-              ? 'Delegate call'
```
schmanu, line 127: This information is gone now. 
I think we should always display if an operation is a Delegate call. We should also display the Delegate Call warning here, if `txDetails?.txData?.trustedDelegateCallTarget` is false.

Currently we only display such warning in `SingleTxDecoded`. But there were cases were scammers would try to bait you into a high level delegateCall - which can be submitted through other interfaces / phishing sites to our tx service.

schmanu (changes requested):
We need to handle top-level DelegateCalls better.

src/components/transactions/HexEncodedData/index.tsx
```diff
@@ -18,22 +19,29 @@ export const HexEncodedData = ({ hexData, title, limit = 20 }: Props): ReactElem
     setShowTxData((val) => !val)
   }
 
+  const firstBytes = <b>{hexData.slice(0, 8)}</b>
+  const restBytes = hexData.slice(8)
```
schmanu, line 23: It is the first 4 bytes. So including the 0x-prefix it is 10 chars.
```suggestion
  const firstBytes = <b>{hexData.slice(0, 10)}</b>
  const restBytes = hexData.slice(10)
```

src/components/transactions/HexEncodedData/index.tsx
```diff
@@ -18,22 +19,29 @@ export const HexEncodedData = ({ hexData, title, limit = 20 }: Props): ReactElem
...
         </span>
       )}
+
+      <CopyButton text={hexData} />
+
+      {firstBytes}
       {showExpandBtn ? (
         <>
-          {showTxData ? hexData : shortenText(hexData, 25)}{' '}
+          {showTxData ? restBytes : shortenText(restBytes, 25)}{' '}
```
schmanu, line 38: As we now always show the first 10 characters separately we have to shorten more.
```suggestion
          {showTxData ? restBytes : shortenText(restBytes, 15)}{' '}
```

src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
usame-algan: Should we also update the existing titles as seen in the screenshot above? "Send" and "Recipient" are both smaller and have a different color than the rest of them.

src/components/tx/DecodedTx/index.tsx
```diff
@@ -43,72 +22,63 @@ type DecodedTxProps = {
...
   }, [chainId, txId])
 
-  const addressInfoIndex = txDetails?.txData?.addressInfoIndex
-
   const onChangeExpand = (_: SyntheticEvent, expanded: boolean) => {
     trackEvent({ ...MODALS_EVENTS.TX_DETAILS, label: expanded ? 'Open' : 'Close' })
   }
 
-  if (!decodedData) return null
+  if (!decodedData || !tx) return null
```
schmanu, line 50: We cannot return `null` if there is no `decodedData` as this causes the sign modal to show no tx info at all when calling unverified contracts or contracts where we have no ABI.
We should always at least show the `SendToBlock`, `SendAmountBlock` if value > 0 and the raw hex `data`.

src/components/tx/DecodedTx/index.tsx
```diff
@@ -119,63 +75,34 @@ const DecodedTx = ({
...
             className={accordionCss.accordion}
           >
-            <span style={{ flex: 1 }}>Transaction details</span>
-
-            {decodedData
-              ? decodedData.method
-              : tx?.data.operation === OperationType.DelegateCall
-              ? 'Delegate call'
-              : ''}
+            <Box display="flex" alignItems="center">
```
schmanu, line 78: This does not look aligned:
Screenshot 2024-07-25 at 11 35 48

src/components/tx/DecodedTx/index.tsx
```diff
@@ -119,63 +75,34 @@ const DecodedTx = ({
...
           </AccordionSummary>
 
           <AccordionDetails data-testid="decoded-tx-details">
-            {decodedData ? (
-              <MethodDetails data={decodedData} addressInfoIndex={addressInfoIndex} />
-            ) : decodedDataError ? (
-              <ErrorMessage error={decodedDataError}>Failed decoding transaction data</ErrorMessage>
-            ) : (
-              decodedDataLoading && <Skeleton />
+            {isMethodCallInAdvanced && decodedData?.method && (
```
schmanu, line 85: I think this would be more for v2 but I find it strange that now for some transactions I find the method call and params in the advanced details and for some I find it on the top.

IMO having this consistent would help to quickly find the data users search for. Like this its a bit guessing around where to look for which transaction type.

src/components/transactions/TxDetails/TxData/DecodedData/MethodCall.tsx
```diff
@@ -0,0 +1,55 @@
...
+          avatarSize={24}
+        />
+      </Typography>
+
+      {/* Divider */}
+      <Box
+        borderBottom="1px solid var(--color-border-light)"
+        width="calc(100% + 32px)"
+        sx={{ ml: '-16px !important' }}
+      />
```
schmanu, line 50: The way we divide content is a bit misleading sometimes. For instance in this transaction:
Screenshot 2024-07-25 at 11 50 03

This divider separates the Method Call from the Parameters (No Parameters). But there is no Divider splitting the parameters from the Advanced Details and no Divider splitting the Info text (Sign or immediately execute this transaction) from the Method Call.
This looks like the Parameters are part of the Details and Balance Change block - even though they are in fact a part of the Method Call.

schmanu (approved):
Minor layouting issues so I'll already approve.

src/components/tx/DecodedTx/index.tsx
katspaugh: The idea is that for the txs that we have a distinct type on CGW (e.i. transfers, settings changes, swaps, etc), showing a raw method call is redundant and too much information, so it goes to advanced details. For arbitrary contract interactions via Safe Apps and WalletConnect, the raw method call is the only decoding we can show in the main summary.
But I see your point also.

src/components/transactions/TxDetails/TxData/DecodedData/MethodCall.tsx
katspaugh: I've put it in a rounded border to differentiate from the other blocks.

Screenshot 2024-07-25 at 12 59 42

src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
katspaugh: @usame-algan done, I made it look like the parameters.

katspaugh: In history:
Screenshot 2024-07-26 at 10 18 19

In the tx modal:
Screenshot 2024-07-26 at 10 18 59

usame-algan: Thanks!

usame-algan (approved):
Thanks for doing all the changes!

Not a blocker but have we explored the order for Batches? It seems like it would be better if the method call block was moved to the very top. Will leave it up to you.
Screenshot 2024-07-26 at 11 47 47

katspaugh:
@usame-algan yes, I tried that, and I don't think it's good. The most important info for the user is what transactions are being batched, not the method signature with a hex-encoded parameter.

francovenica:
Small issue on mobile viewport
Some texts are too long and the copy to clipboard and go to block explorer icons are out of bounds.
Also if the tx is too wide there is horizontal scrolling, but the color of the block where the action's title is, is cut if you do scroll:
image

An opinion, completely subjective:
I really prefer to have also the icon and address of the contract visible. Maybe I'm just used to see it. Feels more confortable.
Maybe bellow the "Call XXXX on YYYY"
a comparison before/after:
image
image

katspaugh:
@francovenica fixed both, thank you!

liliya-soroka:
IMO: safe apps info is better than Batch title and provides me more information 
image
image

liliya-soroka:
2. On MackBook 13", the first parameters is displayed in 2 lines
image

liliya-soroka:
3. when there are many actions the information about untrusted delegate call or other messages about interaction with will be too low 
4. Also Module info might be quite important , why is it under advanced data ?
image

liliya-soroka:
there are 6 failing cases in regression suite that require update because of the change sin PR - https://cloud.cypress.io/projects/exhdra/runs/7607/overview/0c2b3091-f9de-48d3-88f3-05997d92ea9a?roarHideRunsWithDiffGroupsAndTags=1&runStatus=PASSED&runStatus=FAILED&runStatus=RUNNING&runStatus=CANCELLED&runStatus=ERRORED&runStatus=TIMEDOUT&runStatus=NOTESTS
@mike10ca

katspaugh:
I'll go ahead and merge this PR but we can fix the regression tests together in another PR.

PR #3976, opened 2024-07-22
Fix: add correct contents to section 18 of terms page

jmealy:
Section 18 had repeated content.
Also fixes reference mistakes.

Requested by legal


katspaugh (approved):

francovenica:
Ticket was merged but I was asked to take a look. 

Text changes Looks good to me
Also since is merged in dev I was able to check that the pop up to accept the terms and conditions shows again even after being dismissed before

PR #3977, opened 2024-07-22
Feat: dynamically block sactioned addresses [SW-11]

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Dynamic-OFAC-addresses-blocking-b4f02f154d8140e6aeeaea9947c69090

## How this PR fixes it
- Uses a chainalysis contract to check if an address is blocked: https://go.chainalysis.com/chainalysis-oracle-docs.html. 
- The RPC request is done with RTK query, which stores the result for each address in state.
- The blocking is for the swaps feature. It will check based on the wallet address, safe address, and recipient address (if given).

## How to test it

Blocked wallet address:
- Impersonate an address on the OFAC block list.
- Try to use the swap feature.
- The blocked message should show.

Blocked recipient address:
- Using a non blocked wallet address, go to the swaps feature.
- Enable custom recipients and provide a blocked address.
- The blocked message should show. 

Blocked safe address:
- Not sure how we can test this one, I don't think are aware of any safe addresses on the block list.

## Checklist


src/utils/helpers.ts
```diff
@@ -27,3 +27,11 @@ export function assertOnboard(onboard: OnboardAPI | undefined): asserts onboard
 export function assertProvider(provider: Eip1193Provider | undefined): asserts provider {
   return invariant(provider, 'Provider not found')
 }
+
+export const getKeyWithTrueValue = (obj: Record<string, boolean>) => {
+  for (const [key, value] of Object.entries(obj)) {
+    if (value) {
+      return key
+    }
+  }
```
katspaugh, line 36:
```suggestion
  return Object.entries(obj).find([,value] => !!value)[0]
```

src/store/ofac.ts
```diff
@@ -0,0 +1,72 @@
...
+const noopBaseQuery = async () => ({ data: null })
+
+export const ofacApi = createApi({
+  reducerPath: 'ofacApi',
+  baseQuery: noopBaseQuery,
+  endpoints: (builder) => ({
+    getIsSanctioned: builder.query({
+      async queryFn(address, { getState }) {
+        const state = getState()
+        const chain = selectChainById(state as RootState, '1')
```
katspaugh, line 45:
```
import chains from '@/config/chains'
```

```suggestion
        const chain = selectChainById(state as RootState, chains.eth)
```

src/store/ofac.ts
```diff
@@ -0,0 +1,72 @@
...
+
+        if (!chain)
+          return {
+            error: { status: 400, statusText: 'Bad Request', data: 'Chain info not found' },
+          }
+        if (!address)
+          return {
+            error: { status: 400, statusText: 'Bad Request', data: 'No address provided' },
+          }
```
katspaugh, line 54: Possible to de-duplicate some of this code?

jmealy: Something like this?
```
if (!chain) return createBadRequestError('Chain info not found')
if (!address) return createBadRequestError('No address provided')
```

src/features/swap/index.tsx
```diff
@@ -81,12 +83,23 @@ const SwapWidget = ({ sell }: Params) => {
...
   const { safeAddress, safeLoading } = useSafeInfo()
-  const [blockedAddress, setBlockedAddress] = useState('')
+  const [recipientAddress, setRecipientAddress] = useState('')
   const wallet = useWallet()
   const { isConsentAccepted, onAccept } = useSwapConsent()
   const feeEnabled = useHasFeature(FEATURES.NATIVE_SWAPS_FEE_ENABLED)
   const useStagingCowServer = useHasFeature(FEATURES.NATIVE_SWAPS_USE_COW_STAGING_SERVER)
 
+  const { data: isSafeAddressBlocked } = useGetIsSanctionedQuery(safeAddress ?? skipToken)
+  const { data: isWalletAddressBlocked } = useGetIsSanctionedQuery(wallet?.address ?? skipToken)
```
compojoom, line 93: what I discovered is that skipToken, silences TS for when the value could be undefined, but since we don't want to do API request if the value is an empty string, then we still need to pass {skip: !wallet?.address} as second argument. Otherwise, we are dispatching the event

jmealy: Ah true. Also the recipient is initialized to an empty string so there's the same problem there. Replacing `??` with `||` fixes it, and then we don't need the double skip.

`const { data: isWalletAddressBlocked } = useGetIsSanctionedQuery(wallet?.address || skipToken)`

compojoom: Ok, cool!

compojoom (approved):

PR #3979, opened 2024-07-22
Feat: full amount tooltip

### LINKED ISSUE #3978: Show full asset amount on hover
As per the feedback here: https://github.com/5afe/safe-support/issues/491

katspaugh:
## What it solves

Resolves #3978

## How this PR fixes it

Show precise full token and fiat amounts on hover.

Screenshot 2024-07-22 at 14 51 52

### DIFF
--- a/src/features/swap/components/SwapTxInfo/SwapTx.tsx
+++ b/src/features/swap/components/SwapTxInfo/SwapTx.tsx
@@ -14,7 +14,10 @@ const Amount = ({ value, token }: { value: string; token: OrderToken }) => (
 )
 
 const OnlyToken = ({ token }: { token: OrderToken }) => (
-  <TokenIcon tokenSymbol={token.symbol} logoUri={token.logoUri ?? undefined} />
+  <Typography fontWeight="bold" component="span" display="flex" alignItems="center" gap={1}>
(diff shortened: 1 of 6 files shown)


src/features/swap/components/SwapTxInfo/SwapTx.tsx
```diff
@@ -1,65 +1,24 @@
...
-      <>
-        <Box style={{ paddingLeft: 5, paddingRight: 5, display: 'inline-block' }}>
-          <TokenIcon logoUri={buyToken.logoUri || undefined} tokenSymbol={buyToken.symbol} />
-        </Box>{' '}
-        <Typography component="span" fontWeight="bold">
-          {formatVisualAmount(buyAmount, buyToken.decimals)} {buyToken.symbol}{' '}
-        </Typography>
-      </>
-    )
+    ;[from, to] = [to, from]
```
usame-algan, line 14: @compojoom I remember we tried this in the first iteration of swaps but had to change it. Do you remember what the issue was?

usame-algan (approved):
Looks good from my side although the original support request was regarding the aggregated value in the sidebar which seems to be off for large values. Maybe more of an issue we need to solve on the CGW since we are just using `balances.fiatTotal` there so no further precision is possible.

francovenica:
LGTM

PR #3981, opened 2024-07-22
fix: Harden condition for loading spending limits

usame-algan:
## What it solves

Part of #2414 

## How this PR fixes it

- `getSpendingLimits` was called twice because the condition only checked for the existence of `tokenInfoFromBalances` but the initial value is an empty array so it should check the length of the array instead

## How to test it

1. Open a Safe with a spending limit
2. It should still be visible under settings and when trying to create a transaction

## Checklist

### DIFF
--- a/src/hooks/loadables/useLoadSpendingLimits.ts
+++ b/src/hooks/loadables/useLoadSpendingLimits.ts
@@ -103,7 +103,7 @@ export const useLoadSpendingLimits = (): AsyncResult<SpendingLimitState[]> => {
 
   const [data, error, loading] = useAsync<SpendingLimitState[] | undefined>(
     () => {
-      if (!provider || !safeLoaded || !safe.modules || !tokenInfoFromBalances) return
+      if (!provider || !safeLoaded || !safe.modules || tokenInfoFromBalances.length === 0) return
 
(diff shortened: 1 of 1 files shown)


usame-algan:
I wonder what is the expected behaviour in case there are no tokens in a safe? Assuming `tokenInfoFromBalances` would always return an empty array we would not display the spending limits to the user anymore.

compojoom (approved):
I think that this should get our RPC count significantly down as due to this we were doing the queries twice.

schmanu:
> Assuming `tokenInfoFromBalances` would always return an empty array we would not display the spending limits to the user anymore.

@usame-algan I think the `tokenInfoFromBalances` always contains an entry for the native token once fetched from the CGW - even if it is 0. I just setup Spending Limits on a Safe without any balances and they get displayed correctly in the settings.

katspaugh:
@schmanu I think that's fine because the purpose of that check is to verify that the balances are fetched from CGW.
This could probably be more cleanly achieved via `const { balances, loading } = useBalances()`.

francovenica:
Spending limit looks fine

Creation, edition, removal. 
With Native and ERC 20 token. 
Use of the spending limt with the user
Use all the tokens allowed
Check the refresh of the allowance

LGTM

PR #3984, opened 2024-07-23
Fix: align assets table

katspaugh:
## What it solves

Visually align columns and headers in the assets table.

Before:
Screenshot 2024-07-23 at 09 27 32

After:
Screenshot 2024-07-23 at 09 35 51


jmealy (approved):
To me the right align in the balance column looks a bit weird but AFAIR there was already a discussion about this so all good 👍

PR #3991, opened 2024-07-23
fix: Show approval editor for ERC721 approvals

### LINKED ISSUE #3774: Error decoding NFT approval transaction
## Bug description
When executing a transaction that has within it an approval of erc721 to enable its transfer to another wallet, this error comes up:
`Error while decoding approval transaction`

image (11)

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Sepolia testnet
 
## Steps to reproduce
 1. Setup a contract function that has a transfer of nft within it. While doing the multisend to the gnosis module, you must first execute the approval to transfer this nft to another wallet. Then this decoding error message comes up

usame-algan:
## What it solves

Resolves #3774 

## How this PR fixes it

- When the ERC20 token info request fails we try the ERC721 contract and fetch the symbol, if that fails we check the ERC-165 function `supportsInterface` with the ERC721 identifier to at least know if its an ERC721 token or not.

Caveat: If its a multisend that contains both ERC20 and ERC721 approvals all of them will be read-only now.

## How to test it

1. Open the transaction build with an ERC721 contract address
2. Create an `approve` transaction and continue
3. Observe no error messages in the approval editor

## Screenshots
Screenshot 2024-07-26 at 11 41 04

## Checklist


src/components/tx/ApprovalEditor/index.tsx
```diff
@@ -11,11 +11,16 @@ import { decodeSafeTxToBaseTransactions } from '@/utils/transactions'
 import Approvals from '@/components/tx/ApprovalEditor/Approvals'
 import { type EIP712TypedData } from '@safe-global/safe-gateway-typescript-sdk'
 
-const Title = () => {
+const Title = ({ isErc721 }: { isErc721: boolean }) => {
+  const title = isErc721 ? 'Allow access to tokens?' : 'Allow access to tokens?'
+  const subtitle = isErc721
+    ? 'This allows the spender to transfer the specified token.'
+    : 'This allows the spender to spend the specified amount of your tokens.'
```
usame-algan, line 18: I think we can keep the title the same for both erc20 and erc721.

src/components/tx/ApprovalEditor/hooks/useApprovalInfos.ts
```diff
@@ -49,16 +55,22 @@ export const useApprovalInfos = (payload: {
...
           )?.tokenInfo
 
+          let isErc721 = false
+
           if (!tokenInfo) {
-            tokenInfo = await getERC20TokenInfoOnChain(approval.tokenAddress)
+            try {
+              tokenInfo = await getERC20TokenInfoOnChain(approval.tokenAddress)
+            } catch (e) {
+              isErc721 = await isErc721Token(approval.tokenAddress)
```
schmanu, line 64: We should also try to fetch some minimal token information here. At least the name or symbol of the collection for instance.

Also we need to display the `id` that you are giving your approval to.
Currently the modal says "This allows the spender to transfer the specified token."
But we do not specify any token underneath.

schmanu (changes requested):

src/components/tx/ApprovalEditor/hooks/useApprovalInfos.ts
usame-algan: Good point, I added a call to fetch the symbol and decided to only fetch the symbol instead of the name or both because they are part of the same contract so either both exist or both don't and symbol is already part of the approval object.

I've also added a different message and included the token id. Let me know if this works for you:
Screenshot 2024-07-26 at 11 38 12

liliya-soroka:
@usame-algan , I see only a collection name. Do we agree to display only it ?
image

liliya-soroka:
Verified. We are agreed to display the collection name and TokenID
The combination of ERC20 approve and ERC721 approve for different spenders are also checked
image

usame-algan:
> @usame-algan , I see only a collection name. Do we agree to display only it ?

I would suggest for simplicity that we only display the symbol since we only load NFTs when the user visits the NFT page so they are most likely not there.

liliya-soroka:
# verified

PR #3994, opened 2024-07-24
fix: Don't parse amount when removing spending limit

### LINKED ISSUE #3992: Spending limit removal - Too many 0's when removing a spending limit
## Bug description
When the spending limit is removed, the amount being removed is shown. That number is showing a lot of 0's. 
The same doesn't happen when creating the spending limit, just when removing it

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Sepolia
 - Env: Current prod 

## Steps to reproduce
 1. Go to safe you own
 2. Create a spending limit of the native token
 3. Execute the tx
 4. Once is in the spending limit is in the list try to remove it
 5. Check the number in the review screen

## Expected result
It should look like when creating it, the number set by the user with not that many 0's
image

## Obtained result
image

usame-algan:
## What it solves

Resolves #3992

## How this PR fixes it

- The amount when removing a spending limit is already in wei so no need to parse it

## How to test it

1. Open a safe with a spending limit
2. Remove the spending limit
3. Observe the amount is shown correctly

## Screenshots
Screenshot 2024-07-24 at 15 29 23

## Checklist


katspaugh (approved):
👍

francovenica:
LGTM

PR #3999, opened 2024-07-25
feat: add an option to cancel twaps [SW-102]

compojoom:
## What it solves
Ads an option to cancel a twap order. It currently it's hard for other signers that haven't submitted the transaction to cancel a TWAP. By having a cancel button in the Safe Interface every signer would be able to propose a transaction to cancel it. 

## Checklist


katspaugh:
An update from @compojoom:

> We have the endpoint on CGW now that gives the currently active part for canceling. 

So this PR has a chance of eventually being merged. Don't close.

katspaugh:
I'll close this PR for now since we don't have capacity to work in this atm, the branch can be restored if needed.

PR #4001, opened 2024-07-26
Refactor: update react-redux and redux-toolkit to latest versions

compojoom:
## What it solves
While I was tracking why a component re-renders I found out that it is due to a reducer always returning a new array. Those kind of errors should actually be reported in dev mode, but our version of redux-tookit was not doing that. After I quickly looked at the changelog it seemed that the relevant code for this was added in redux tookit version 5, but we were on 4. 

I had to bite the bullet and do the update. I did try to do this in the past, but back then didn't want to deal with the type problems this update was bringing with it... Now I no longer had a choice. :)

## Checklist

### DIFF
--- a/src/store/broadcast.ts
+++ b/src/store/broadcast.ts
@@ -1,22 +1,23 @@
 import type { Store } from 'redux'
-import type { Middleware, PreloadedState } from '@reduxjs/toolkit'
+import type { Middleware } from '@reduxjs/toolkit'
 import type { RootState } from '@/store'
-
-type PreloadedRootState = PreloadedState<RootState>
 
 const BC_NAME = 'SAFE__store-updates'
 const tabId = Math.random().toString(32).slice(2)
 let broadcast: BroadcastChannel | undefined
 
-export const broadcastState = <K extends keyof PreloadedRootState>(sliceNames: K[]): Middleware<{}, RootState> => {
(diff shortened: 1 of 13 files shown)


src/store/broadcast.ts
```diff
@@ -1,22 +1,23 @@
...
     // Broadcast actions that aren't being already broadcasted
-    if (!action._isBroadcasted) {
-      const sliceType = action.type.split('/')[0]
-      if (sliceNames.includes(sliceType)) {
-        broadcast?.postMessage({ action, tabId })
+    if (typeof action === 'object' && action !== null) {
+      const actionObj = action as { _isBroadcasted?: boolean; type?: string }
+      if (!actionObj._isBroadcasted && actionObj.type) {
+        const sliceType = actionObj.type.split('/')[0]
+        if (sliceNames.includes(sliceType as K)) {
```
katspaugh, line 18: Was it doing more broadcasts than needed before?

katspaugh (approved):
Looks good!

src/store/broadcast.ts
compojoom: Those changes are more due to the new ts types. I had to silence ts. I don't think that there was an error in the previous code.

katspaugh:
Let's merge it directly, since it's mostly mechanical changes.

PR #4003, opened 2024-07-29
fix: useIsExpiredSwap causing too many rerenders

compojoom:
## What it solves
The Hook was using IntervalCounter and despite the state of the hook not changing, the interval counter was being updated, which in turn caused every component using this hook to rerender.

Since this hook is used in TXSummary it meant that the whole TX History or TX Queue were rerendered every 10s degrading the performance.

No unnecessary rerenders every 10s

## How to test it
Easiest way to see the improvement is to use the react devtools extension and make sure that "highlight updates when component render" is turned on. Without the fixes in this PR you will see the History page constantly rerenering
grafik

## Checklist

### DIFF
--- a/src/features/swap/hooks/useIsExpiredSwap.ts
+++ b/src/features/swap/hooks/useIsExpiredSwap.ts
@@ -1,9 +1,6 @@
-import { useEffect, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import type { TransactionInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import { isSwapOrderTxInfo } from '@/utils/transaction-guards'
-import useIntervalCounter from '@/hooks/useIntervalCounter'
-
-const INTERVAL_IN_MS = 10_000
 
 /**
  * Checks the expiry time of a swap every 10s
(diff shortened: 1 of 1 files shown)


src/features/swap/hooks/useIsExpiredSwap.ts
```diff
@@ -1,9 +1,6 @@
-import { useEffect, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import type { TransactionInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import { isSwapOrderTxInfo } from '@/utils/transaction-guards'
-import useIntervalCounter from '@/hooks/useIntervalCounter'
```
katspaugh, line 4: Does this mean that all other places where we use `useIntervalCounter` may cause an unnecessary re-render?

src/features/swap/hooks/useIsExpiredSwap.ts
```diff
@@ -12,13 +9,33 @@ const INTERVAL_IN_MS = 10_000
...
 const useIsExpiredSwap = (txInfo: TransactionInfo) => {
   const [isExpired, setIsExpired] = useState<boolean>(false)
-  const [counter] = useIntervalCounter(INTERVAL_IN_MS)
+  const timerRef = useRef<NodeJS.Timeout | null>(null)
 
   useEffect(() => {
     if (!isSwapOrderTxInfo(txInfo)) return
 
-    setIsExpired(Date.now() > txInfo.validUntil * 1000)
-  }, [counter, txInfo])
```
usame-algan, line 21: This updates on every `counter` increase but will say `true` at some point so `useIsExpiredSwap` would then always return the same boolean value. How is this causing rerenders?

usame-algan: Thanks for looking into this! Makes a lot more sense to only check this once if we have the exact timestamp. No need to do this every 10s with a poll.

src/features/swap/hooks/useIsExpiredSwap.ts
```diff
@@ -1,9 +1,6 @@
-import { useEffect, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import type { TransactionInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import { isSwapOrderTxInfo } from '@/utils/transaction-guards'
-import useIntervalCounter from '@/hooks/useIntervalCounter'
-
-const INTERVAL_IN_MS = 10_000
 
 /**
  * Checks the expiry time of a swap every 10s
```
usame-algan, line 6: This comment is outdated now

usame-algan (approved):

katspaugh (approved):
👍

PR #4006, opened 2024-07-29
fix: swap token selection flickers

compojoom:
## What it solves
Tokens can end up having the same token abbreviation, but they always have different token addresses. If the user select a token that has the same abbreviation as another token in the list the widget starts flickering between the selected tokens and resets to WETH for input token and no selected output token. By opting to use addresses instead the problem seems to be resolved.

## How to test it
Go to the swaps page.
For source token select: USDC
for target token select: APE

Without this commit the token selection will flicker and would reset to WETH for source token and nothing for output token. With this PR source token would be USDC and target token would be the APE token selected.

## Checklist


katspaugh (approved):
👍

Issue #3988, opened 2024-07-23
Copy safe address on mobile view doesn't work


usame-algan:
## Bug description

Trying to copy the safe address from the sidebar when having the app open on a mobile phone doesn't do anything.

## Environment
 - Browser: Safari
 - Wallet: MetaMask
 - Chain: Any

## Steps to reproduce
 1. Open app.safe.global on mobile safari
 2. Open a Safe
 3. Open the Sidebar and press the copy button

## Expected result

The safe address should be copied and there should be some ui feedback to the user

## Obtained result

Nothin happens and nothing is copied

## Screenshots
![IMG_7209](https://github.com/user-attachments/assets/102b2e15-61c4-4df6-97f3-ec29d660484c)

HarshS1611:
Hey, could you assign this issue to me?

jmealy:
> Hey, could you assign this issue to me?

Thanks @HarshS1611!

HarshS1611:
Hey @jmealy, thanks for assigning the issue. I have raised the PR #4009 . Kindly review and let me know your thoughts.


https://github.com/user-attachments/assets/f2ab20a6-5780-48a1-92d4-a788d2b31868

PR #4009, opened 2024-07-30
fix:Copy safe address on mobile view

HarshS1611:
## What it solves
Copy safe address on mobile view doesn't work #3988

## How this PR fixes it
- I have added a fallback function for IOS devices and ensure that it copied the address for every device.
## How to test it
- Open the Safari browser on your mobile device or the simulator from the Safari browser if you have XCode from the App Store.
- Connect your wallet (metamask) or view any account (0x17D47f5a76e9601C4179963E18cc17918564864F) on sepolia.
- Open the sidebar and click on the copy icon.
- Now the safe address is copied.

## Screenshots

https://github.com/user-attachments/assets/c2712aee-a037-40b4-8eac-6719260cef93

## Checklist

### DIFF
--- a/src/components/common/CopyTooltip/index.tsx
+++ b/src/components/common/CopyTooltip/index.tsx
@@ -15,7 +15,7 @@ const CopyTooltip = ({
   text: string
   children?: ReactNode
   initialToolTipText?: string
-  onCopy?: () => void
+  onCopy?: (e: SyntheticEvent) => void
   dialogContent?: ReactElement
 }): ReactElement => {
   const [tooltipText, setTooltipText] = useState(initialToolTipText)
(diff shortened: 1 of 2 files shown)


src/components/common/CopyTooltip/index.tsx
```diff
@@ -32,24 +32,69 @@ const CopyTooltip = ({
...
+        document.body.appendChild(textArea)
+        textArea.focus()
+        textArea.select()
+
+        try {
+          // Using touch events for iOS
+          const event = new TouchEvent('touch')
+          textArea.dispatchEvent(event)
+          document.execCommand('selectall')
+          const successful = document.execCommand('copy')
```
katspaugh, line 65: This method is deprecated: https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand

I would rather not add this fallback. Any idea why navigator.clipboard might not work on iOS?

compojoom: > Secure context: This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard#browser_compatibility).

Is it possible that we are testing on non-https page?

HarshS1611: > > Secure context: This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard#browser_compatibility).
> 
> Is it possible that we are testing on non-https page?
Yeah, it works on https pages, but not on localhost and non-https pages

HarshS1611: > This method is deprecated: https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand
> 
> I would rather not add this fallback. Any idea why navigator.clipboard might not work on iOS?

`navigator.clipboard` is working on https pages only not on localhost.

katspaugh: Then I suggest we just log a warning if clipboard fails but not introduce a fallback just for dev environments. Thanks for figuring out the root cause @HarshS1611!

katspaugh:
I'll go ahead and close this PR as per the discussion above. Thanks for investigating this @HarshS1611!

PR #4011, opened 2024-07-30
Fix: do not show redefine block when no scan is performed

### LINKED ISSUE #3622: Messages: Hide "Scan for Risk" block for the EIP-712 messages
## Bug description
Messages: Hide "Scan for Risk" block for the EIP-712 messages

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet/any

## Steps to reproduce
 1. Go to OpenSea
 2. try to connect to the safe 
 3. check "Cinfirm messages" page
 Current result:
 there is "Scan for Risk" block without any action

## Expected result
Check with Redefine if the scan for risk is supported for the EIP-712 message. If the scanning is not supported, hide the "Scan for risk" block on Confirm Messages page
## Screenshots
image

jmealy:
## What it solves

Resolves #3622

Also fixes the tooltip being partially obscured by the redefine logo, causing the tooltip text to not show up when hovering over the bottom part of the tooltip icon.

## How this PR fixes it
- If there is nothing to show in the redefine block then don't render the component at all.
- set z-index on the tooltip icon so the tooltip text always shows up.

## How to test it
- Open the message signing confirmation screen for a message where redefine checks are not performed. For example signing in to opensea.
- See that the redefine block is not shown. [1]
- When the redefine block is shown (eg. executing a send tx on a supported network), see that the redefine tooltip text is shown when hovering over any part of the tooltip icon. [2]

## Screenshots
[1]
image

[2] 
image

## Checklist

### DIFF
--- a/src/components/tx-flow/flows/SignMessage/SignMessage.tsx
+++ b/src/components/tx-flow/flows/SignMessage/SignMessage.tsx
@@ -305,7 +305,7 @@ const SignMessage = ({ message, safeAppId, requestId }: ProposeProps | ConfirmPr
           </Typography>
           <DecodedMsg message={decodedMessage} isInModal />
 
-          <Accordion sx={{ my: 2, '&.Mui-expanded': { mt: 2 } }}>
+          <Accordion sx={{ mt: 2 }}>
             <AccordionSummary data-testid="message-details" expandIcon={<ExpandMoreIcon />}>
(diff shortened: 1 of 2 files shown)


schmanu (approved):

liliya-soroka:
Verified

PR #4013, opened 2024-07-30
fix: accept canonical mastercopy

schmanu:
## What it solves
Safe deployments 1.37.2 adds canonical deployment addresses.
Those are used when i.e. replaying a 1.3.0 L1 Safe on a L2 chain and on some networks such as Optimism they are not detected currently.

## How this PR fixes it
- Updates safe-deployments
- Also accepts the canonical L1 master copy address

## How to test it
- Open a replayed 1.3.0 L1 Safe on e.g. Optimism

## Checklist

### DIFF
--- a/src/hooks/coreSDK/safeCoreSDK.ts
+++ b/src/hooks/coreSDK/safeCoreSDK.ts
@@ -1,6 +1,6 @@
 import chains from '@/config/chains'
 import type { UndeployedSafe } from '@/features/counterfactual/store/undeployedSafesSlice'
-import { getSafeSingletonDeployment, getSafeL2SingletonDeployment } from '@safe-global/safe-deployments'
+import { getSafeSingletonDeployments, getSafeL2SingletonDeployments } from '@safe-global/safe-deployments'
 import ExternalStore from '@/services/ExternalStore'
 import { Gnosis_safe__factory } from '@/types/contracts'
 import { invariant } from '@/utils/helpers'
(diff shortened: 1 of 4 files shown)


src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -56,11 +64,11 @@ export const initSafeSDK = async ({
...
 
-    const safeL1Deployment = getSafeSingletonDeployment({ network: chainId, version: safeVersion })
-    const safeL2Deployment = getSafeL2SingletonDeployment({ network: chainId, version: safeVersion })
+    const safeL1Deployment = getSafeSingletonDeployments({ network: chainId, version: safeVersion })
+    const safeL2Deployment = getSafeL2SingletonDeployments({ network: chainId, version: safeVersion })
 
-    isL1SafeSingleton = masterCopy === safeL1Deployment?.networkAddresses[chainId]
-    const isL2SafeMasterCopy = masterCopy === safeL2Deployment?.networkAddresses[chainId]
+    isL1SafeSingleton = isInDeployments(masterCopy, safeL1Deployment?.networkAddresses[chainId])
+    const isL2SafeMasterCopy = isInDeployments(masterCopy, safeL2Deployment?.networkAddresses[chainId])
```
compojoom, line 71: I'm fine with the change, but if you have don't have the context on the L1/L2 deployment situation I doubt that anyone would be able to figure out why we do all this magic :D

compojoom (approved):
👍

src/hooks/coreSDK/safeCoreSDK.ts
schmanu: Our deployment structure changed in general. Now there can be multiple contract deployments per contract. So this would be correct regardless of this case in the future.

katspaugh (approved):
👍

PR #4017, opened 2024-07-31
feat: Add support for proposing transactions as a delegate

### LINKED ISSUE #1825: Propose a transaction if you are a delegate
## What is the feature about
Being able to propose a transaction from a wallet that is a delegate for the safe

## The list of requirements
Right now, if you open a safe from a wallet that is a delegate but not an owner, it is not possible to propose any transaction.
The safe stays in `readonly` mode.

Proposing a transaction is possible from the API with a delegate wallet

usame-algan:
## What it solves

Resolves #1825

## How this PR fixes it

- Check if a connected wallet is a delegate and store it as a flag in `useWallet`
- Enable transaction buttons for delegates
- Display a `DelegateForm` when creating a transaction and only propose the transaction

## ToDos

- [x] The transaction should be accessible via the UI even with 0 confirmations (untrusted)
- [x] Adjust message content on the review tx screen
- [x] Add v2 endpoint to the SDK instead of using deprecated v1 endpoint via `getDelegates`

## How to test it

1. Open a Safe that has a delegate who is not an owner
2. Connect with that delegate
3. Observe the New transaction button is enabled
4. Observe being able to create a transaction

## Screenshots
Screenshot 2024-07-31 at 12 39 25

## Checklist


src/components/common/WalletProvider/index.tsx
```diff
@@ -7,19 +10,27 @@ const WalletProvider = ({ children }: { children: ReactNode }): ReactElement =>
...
     if (!onboard) return
 
-    const walletSubscription = onboard.state.select('wallets').subscribe((wallets) => {
+    const walletSubscription = onboard.state.select('wallets').subscribe(async (wallets) => {
       const newWallet = getConnectedWallet(wallets)
+
+      if (newWallet && chainId && safeAddress) {
+        const delegates = await getDelegates(chainId, { safe: safeAddress })
+        newWallet.isDelegate = delegates.results.some((delegate) => delegate.delegate === newWallet.address)
+      }
```
katspaugh, line 25: This would delay wallet connection for all users to serve a very minor portion of users. I would not add isDelegate to the wallet object and instead use a separate hook.

katspaugh (commented):
Code look good!
There was a discussion about how we should call delegates in the UI (e.g. proposers). Let's further align on this with the team.

njelich:
@katspaugh Does this mean Delegate management is coming to the web ui as well?

francovenica:
I tested it with a safe where I already have delegates set and it worked fine. Tested with MM extension, trezor and ledger.

LGTM. Solve the conflicts before merging it.

src/hooks/useDelegates.ts
```diff
@@ -0,0 +1,22 @@
...
+import { useGetDelegatesQuery } from '@/store/gateway'
+import { skipToken } from '@reduxjs/toolkit/query/react'
+
+const useDelegates = () => {
+  const {
+    safe: { chainId },
+    safeAddress,
+  } = useSafeInfo()
+
+  return useGetDelegatesQuery(chainId && safeAddress ? { chainId, safeAddress } : skipToken)
```
katspaugh, line 12: Is this cached?

usame-algan: Yes, the data is cached and updates only if `chainId` or `safeAddress` changes. Even if we have `useDelegates` in multiple components it will still access the cached data instead of making copies.

PR #4018, opened 2024-07-31
Feat: advanced Safe creation page

schmanu:
## What it solves
Offers a way to create Safes with old version in case a user e.g. has to recover a counterfactual Safe or needs a different Safe version than our used one.

## How this PR fixes it
- Adds hidden advanced Safe creation route and flow: `/new-safe/advanced-create`
- Refactors Safe creation. All creations now add the Safe as counterfactual Safe. But "Pay now" option first creates a deployment tx.
- MultiSig counterfactual Safes only offer Activation and no other txs as we cannot collect sigs.

## How to test it
- Navigate to `/new-safe/advanced-create`
- Create a Safe

## Screenshots
Screenshot 2024-07-31 at 14 36 12
Screenshot 2024-07-31 at 14 35 59

## Checklist


src/components/common/CheckWallet/index.tsx
```diff
@@ -23,11 +25,16 @@ const CheckWallet = ({ children, allowSpendingLimit, allowNonOwner, noTooltip }:
   const isSpendingLimit = useIsOnlySpendingLimitBeneficiary()
   const connectWallet = useConnectWallet()
 
+  const { safe } = useSafeInfo()
+  const isCounteractualMultiSig = !allowNonOwner && !safe.deployed && safe.threshold > 1
```
usame-algan, line 29: Small typo here, should be `isCounterfactualMultiSig`.

src/components/new-safe/create/useSyncSafeCreationStep.ts
```diff
@@ -5,29 +5,17 @@ import type { NewSafeFormData } from '@/components/new-safe/create/index'
 import useWallet from '@/hooks/wallets/useWallet'
 import useIsWrongChain from '@/hooks/useIsWrongChain'
 import { useRouter } from 'next/router'
-import { AppRoutes } from '@/config/routes'
 
 const useSyncSafeCreationStep = (setStep: StepRenderProps<NewSafeFormData>['setStep']) => {
-  const [safeAddress, pendingSafe] = useUndeployedSafe()
+  const [_, pendingSafe] = useUndeployedSafe()
```
usame-algan, line 10: Is `pendingSafe` still needed here?

usame-algan (commented):
There seems to be some conflicts that prevent deployment.

src/components/new-safe/create/useSyncSafeCreationStep.ts
schmanu: Good spot. Strange that no linter complained about the unused dependency in the hook dep array.

I will remove it

src/features/counterfactual/ActivateAccountButton.tsx
```diff
@@ -24,24 +25,28 @@ const ActivateAccountButton = () => {
...
-            </>
-          ) : (
-            'Activate now'
+        <CheckWallet allowNonOwner>
+          {(isOk) => (
+            <Button
+              data-testid="activate-account-btn"
+              variant="contained"
+              size="small"
+              onClick={activateAccount}
```
usame-algan, line 34:
```suggestion
              onClick={activateAccount}
                sx={{ minHeight: '40px' }}
```

So that both buttons have the same size on the dashboard

src/components/new-safe/create/AdvancedCreateSafe.tsx
```diff
@@ -0,0 +1,196 @@
...
+import type { CreateSafeInfoItem } from '@/components/new-safe/create/CreateSafeInfos'
+import CreateSafeInfos from '@/components/new-safe/create/CreateSafeInfos'
+import { type ReactElement, useMemo, useState } from 'react'
+import ExternalLink from '@/components/common/ExternalLink'
+import { HelpCenterArticle, LATEST_SAFE_VERSION } from '@/config/constants'
+import { type NewSafeFormData } from '.'
+import { type SafeVersion } from '@safe-global/safe-core-sdk-types'
+import AdvancedOptionsStep from './steps/AdvancedOptionsStep'
+
+const staticHints: Record<
```
usame-algan, line 24: Wdyt about removing hints for this advanced flow or at least reusing the existing data to reduce code duplication?

src/components/new-safe/create/steps/AdvancedOptionsStep/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+  const formMethods = useForm<AdvancedOptionsStepForm>({
+    mode: 'onChange',
+    defaultValues: {
+      [AdvancedOptionsFields.safeVersion]: LATEST_SAFE_VERSION as SafeVersion,
+      [AdvancedOptionsFields.saltNonce]: 0,
+    },
+  })
+
+  const { handleSubmit, control, watch, formState, getValues, setValue, trigger, register } = formMethods
```
usame-algan, line 61:
```suggestion
  const { handleSubmit, control, watch, formState, getValues, register } = formMethods
```

src/components/new-safe/create/steps/AdvancedOptionsStep/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+export type AdvancedOptionsStepForm = {
+  [AdvancedOptionsFields.safeVersion]: SafeVersion
+  [AdvancedOptionsFields.saltNonce]: number
+}
+
+const ADVANCED_OPTIONS_STEP_FORM_ID = 'create-safe-advanced-options-step-form'
+
+const AdvancedOptionsStep = ({ onSubmit, onBack, data, setStep }: StepRenderProps<NewSafeFormData>): ReactElement => {
+  const wallet = useWallet()
```
usame-algan, line 48:
```suggestion
```

src/components/new-safe/create/steps/AdvancedOptionsStep/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+              )}
+            />
+          </Box>
+
+          <Divider />
+          <Box className={layoutCss.row}>
+            <Typography variant="h4" fontWeight={700} display="inline-flex" alignItems="center" gap={1}>
+              Salt nonce
+              <Tooltip
+                title="The salt nonce changes the predicted Safe address. It can be used to re-create a Safe from anothet chain or to create a specifcic Safe address"
```
usame-algan, line 153:
```suggestion
                title="The salt nonce changes the predicted Safe address. It can be used to re-create a Safe from another chain or to create a specific Safe address."
```

src/components/new-safe/create/steps/AdvancedOptionsStep/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+              name="New Safe address"
+              value={
+                predictedSafeAddress ? (
+                  <EthHashInfo address={predictedSafeAddress} hasExplorer showCopyButton />
+                ) : (
+                  <Skeleton />
+                )
+              }
+            />
+          </Grid>
```
usame-algan, line 196: I would suggest wrapping this in a `<Box className={layoutCss.row}>` as well so that it is aligned with the other rows.
Screenshot 2024-08-05 at 11 18 52

src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -207,13 +209,14 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
...
-            safeVersion: LATEST_SAFE_VERSION as SafeVersion,
+            safeVersion: data.safeVersion,
           },
         },
       }
 
       const onSubmitCallback = async (taskId?: string, txHash?: string) => {
-        dispatch(addUndeployedSafe(undeployedSafe))
+        // Create a counterfactual Safe
+        await createCounterfactualSafe(chain, safeAddress, saltNonce, data, dispatch, props)
```
usame-algan, line 219:
```suggestion
        createCounterfactualSafe(chain, safeAddress, saltNonce, data, dispatch, props)
```

Now that it doesn't return a promise anymore.

katspaugh:
Please update the PR so that we can move it to QA.

usame-algan (approved):
Thanks!

francovenica:
Question: Is this always going to be an internal tool? or do we actually intend for user to be able to use this at some point?

francovenica:
I'm not getting a redirection after creating a safe. Is expected from me to copy the address given during the steps? 
I'm stuck here. I created a safe with a 1/1 safe 1.3.0 with a "Pay now" option.

image

schmanu:
@francovenica good catch! I fixed the reported error. Now the status page correctly forwards to the Safe after creation.

The page is intended to be hidden but we will send it to users who want to deploy a legacy Safe version or who need to recover a lost counterfactual Safe on 1.3.0

francovenica:
The issue has been fixed for safes 1/1, but if you create a safe 1/2 the redirection fails in the same way as it was reported

Other issue: the salt nonce is not saved in between steps, so if you go back it resets to 0. Same with the safe version 

I don't understand this point:
"MultiSig counterfactual Safes only offer Activation and no other txs as we cannot collect sigs."

CF are always 1/1 safes, I don't get the option of "pay later" with safes with more of one owner. Am I missunderstanding this bullet point?

schmanu:
> I don't understand this point: "MultiSig counterfactual Safes only offer Activation and no other txs as we cannot collect sigs."
> 
> CF are always 1/1 safes, I don't get the option of "pay later" with safes with more of one owner. Am I missunderstanding this bullet point?

@francovenica This PR refactors the Safe creation logic a bit. Instead of keeping a pendingSafe and always reopening it if you close the Create Safe modal while still creating - we just add all Safes as counterfactual Safes to re-use all that logic. 
To reproduce this you can create a 2/3 Safe with too low gas costs - then navigate away from the status page and cancel the transaction in metamask. This will make it a 2/3 Counterfactual Safe.

francovenica:
Ok tried what you suggested and yes. the safe shows in the list as a CF safe to be activated. Also checked that the only thing you can do as a tx is the activation:

Leaving some snapshots for the record
image
image

LGTM

francovenica:
Question given this in the user story:

"Users which lost access to their undeployed Safe data which want to deploy a old counterfactual Safe need a way to deploy a 1.3.0 Safe through our UI."

I think the only way to do this is by knowing the "salt nonce" before hand. Is there a way for the user to know or find out the salt nonce of his lost CF safe?

schmanu:
> I think the only way to do this is by knowing the "salt nonce" before hand. Is there a way for the user to know or find out the salt nonce of his lost CF safe?

@francovenica We usually start with 0 and then increment. So the user could do that here.

francovenica:
Ok LGTM then

PR #4020, opened 2024-07-31
refactor: usage of getTransactionDetails

compojoom:
Today with @usame-algan we found out that our use of useIntervalCounter was causing the txDetails component to also rerender every 15s despite no data changing. All this was introduced when we wanted to poll the getTransactionDetails for swap transactions (as they can change their status from open to fullfilled, canceled etc)
It turned out to be very hard to rewrite our useAsync pattern to poll & I thought that this could be good exercise to test rtk query. I'm very pleased with the result.

This:
```
  const [pollCount] = useIntervalCounter(POLLING_INTERVAL)
  const swapPollCount = isOpenSwapOrder(txSummary.txInfo) ? pollCount : 0

  const [txDetailsData, error, loading] = useAsync<TransactionDetails>(
    async () => {
      if (txDetails && swapPollCount <= 0) {
        return txDetails
      }
      return getTransactionDetails(chainId, txSummary.id)
    },
    // eslint-disable-next-line react-hooks/exhaustive-deps
    [txDetails, chainId, txSummary.id, safe.txQueuedTag, swapPollCount],
    false,
  )
```
turned into this
```
  const {
    data: txDetailsData,
    error,
    isLoading: loading,
    refetch,
  } = useGetTransactionDetailsQuery(
    { chainId, txId: txSummary.id },
    {
      pollingInterval: isOpenSwapOrder(txSummary.txInfo) ? 2000 : undefined,
    },
  )

  useEffect(() => {
    refetch()
  }, [safe.txQueuedTag, refetch, txDetails])

```

Notice that we are outsourcing the polling to rtk if we are dealing with a swap order. If not, we just fetch the information and keep it in the store. Any other component that needs the same data can access it and won't have to do a new ajax request for that. 

The data is kept for as long as there is a component on the page using the data. If the user navigates away the data will be kept in the store for 60s and will be removed. The next time the user goes tries to view the data it will be refetched. All this is configurable. 

I've changed few more places where we use getTransactionDetails. I also noticed that we use the getTxDetails function as a memoized version of getTransactionDetails in tracking or notification code. I'll try to think if we could be move this to also use rtk. It should be possible to use it without hooks, so this would be also cool.

## What it solves
Fixes unnecessary rerenders, makes our code easier to read, reduces network requests.

Notice that I've opted for calling this reducer gatewayAPI. The idea is to add the getBalance, getChains calls in there.

## How to test it
Run the site with redux devtools extension in your browser. You should see redux events being dispatched that fetch the data from the server and save it in the store.

## Checklist


src/components/transactions/TxDetails/index.tsx
```diff
@@ -167,21 +166,22 @@ const TxDetails = ({
...
-      return getTransactionDetails(chainId, txSummary.id)
+  const {
+    data: txDetailsData,
+    error,
+    isLoading: loading,
+    refetch,
+  } = useGetTransactionDetailsQuery(
+    { chainId, txId: txSummary.id },
+    {
+      pollingInterval: isOpenSwapOrder(txSummary.txInfo) ? 2000 : undefined,
```
usame-algan, line 177:
```suggestion
      pollingInterval: isOpenSwapOrder(txSummary.txInfo) ? POLLING_INTERVAL : undefined,
```

compojoom: ah, thanks!

src/components/transactions/TxDetails/index.tsx
```diff
@@ -159,21 +158,22 @@ const TxDetails = ({
...
-      return getTransactionDetails(chainId, txSummary.id)
+  const {
+    data: txDetailsData,
+    error,
+    isLoading: loading,
+    refetch,
+  } = useGetTransactionDetailsQuery(
+    { chainId, txId: txSummary.id },
+    {
+      pollingInterval: isOpenSwapOrder(txSummary.txInfo) ? POLLING_INTERVAL : undefined,
```
usame-algan, line 169: Previously if it was not a swap order and there were already `txDetails` we would just return the existing `txDetails` without fetching them again. Is this still covered with the new hook?

usame-algan (approved):
Code looks good! Can you add a list of cases that should be tested by QA? e.g. single transaction view should still load, executing a transaction should still track the type correctly etc.

src/components/transactions/TxDetails/index.tsx
compojoom: good point. You are right the new code is not equivalent and we actually aways do the the network request for it. I'll try to come up with a variant that has the old behavior.

compojoom: With @usame-algan we agreed that this is actually no longer necessary. I've written a detailed explanation why here: https://github.com/safe-global/safe-wallet-web/issues/4056

compojoom:
@safe-global/safe-qa - this modification shouldn't change anything in the user interface. The places that are touched by the change are

SingleTx view, History & Queue view - here the tx details should load as before. Swap txs should also properly display the swap info and change from open to fullfilled.

ReviewBatch
ExecuteTrhoughRoleForm -> here tracking shuold further work as expected
SignOrExecuteForm -> tracking, but also tx details should be there
SpeedUpModal -> tracking
txNotifiations -> notifications should further have humandDescriptions as before
Tracking -> whenever the useTxTracking function is used we should be sending the correct origin taken from the txDetails

Hope that this clearly outlines the possible places where you would expect stuff to break.

katspaugh:
@safe-global/safe-qa, translating:

* Regression for tx flows
* Speeding up
* Tx details in history and queue
* tx_created, tx_confirmed and tx_executed events should have the correct label

liliya-soroka:
regression run is green - [Regression on demand tests](https://github.com/safe-global/safe-wallet-web/actions/runs/10303812275)

francovenica:
@compojoom Hi there, can we solve the conflicts? I'd like to be sure the ticket is clear before testing it

usame-algan:
> @compojoom Hi there, can we solve the conflicts? I'd like to be sure the ticket is clear before testing it

@francovenica conflicts are resolved. Can be tested now 🙏

francovenica:
The button to open your owned safes is not working. The GTM calls are being triggered, but nothing else happens.
I get this error after a bit of staying in the welcome page. Probably is related:
Tested in current Dev and I don't have that issue
image

button

francovenica:
Question.

These elements come from the tx redesign right? I was told that nothing from the redesign will be deployed in this release, so is this going to be removed from this ticket?
image

katspaugh:
Nothing redesign-related can be reverted at this point as it touches too many files that were already fused with other features.
I'll move the new block to the Advanced Details for swaps.

Regarding this PR itself, I suggest we not include it in the release.

francovenica:
The issue of the owned safes list not showing up is solved. Works fine now

I agree with not merging this. Anything that is not necesary for safes 1.4.1 I'd like to leave it out until the next release

francovenica:
This ticket is fine, the issues we had were already discussed and it was stated that the changes of tx details will stay. 
Solve the conflict and the ticket can pass. 
I'll re check during regression

PR #4023, opened 2024-08-01
fix: update terms page last updated date

jmealy:
## What it solves
Since the latest release went out later than expected, the changes to the terms page went live in August rather than July. 

## How this PR fixes it
Update the 'last updated' date in the terms popup and the terms page itself


katspaugh (approved):

liliya-soroka:
Verified

PR #4024, opened 2024-08-01
Tests: Fix hp tests

mike10ca:
## How this PR fixes it

- Add accept cookies key to local storage to maintain cookie acceptance throughout tests
- Add timeouts to test workflow to max 60 mins
- Add logout steps to keep login state fresh in further tests

## How to test it


liliya-soroka (approved):

cypress/e2e/pages/batches.pages.js
```diff
@@ -62,7 +62,8 @@ function executeTransaction() {
 }
 
 function addToBatchButton() {
-  cy.contains(addToBatchBtn).should('be.visible').and('not.be.disabled').click()
+  cy.wait(1000)
```
katspaugh, line 65: I don't think this timeout is necessary. The button is rendered immediately, it's just not in the viewport.

cypress/e2e/pages/batches.pages.js
```diff
@@ -62,7 +62,8 @@ function executeTransaction() {
...
+  cy.get('button').contains(addToBatchBtn).scrollIntoView().should('not.be.disabled').click()
```
katspaugh, line 66: Here you don't check if the button is visible, so no need to scroll to it either.

PR #4027, opened 2024-08-01
Feat: display a Network switch button when connected on the wrong chain

### LINKED ISSUE #3743: Display a Switch Network button if wallet is connected to a different chain

jmealy:
## What it solves

Resolves #3743

## How this PR fixes it
- Shows a button to switch network instead of the execution/sign button, whenever the wallet is connected on the wrong chain.
- When the wallet network has been switched to the same as the safe, the execution or sign button is displayed as before.

## How to test it
This affects any place where you perform an action using your connected wallet, where you wallet network must match your safe network:
- Executing a tx (send, batch, review spending limit tx, recover an account, etc.)
- Signing a message
- Changing safe notification settings
- Activating a CF safe.
- Speed up tx modal 

## Screenshots
Execute tx modal while wallet is on the wrong chain:
image

Changing notifications settings:
image

Activating CF safe:
image

## Checklist


TanyaEfremova:
Looks good, my suggestions are:
- Show a network logo instead of the color
- Align text styles: both the label and the network name should use `Body 1`

Question: once the network is switched, will the signature request pop up automatically?

jmealy:
> Looks good, my suggestions are:
> 
> * Show a network logo instead of the color
> * Align text styles: both the label and the network name should use `Body 1`
> 
> Question: once the network is switched, will the signature request pop up automatically?

After the network is switched, the normal execute button will show so it would be a separate action.

src/components/common/CheckWallet/index.test.tsx
```diff
@@ -50,6 +58,8 @@ describe('CheckWallet', () => {
 
     const { container } = renderButton()
 
+    console.log(container)
+
```
katspaugh, line 62:
```suggestion
```

src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -45,17 +45,17 @@ import { useMemo, useState } from 'react'
 export const NetworkFee = ({
   totalFee,
   chain,
-  willRelay,
+  strikethrough,
```
katspaugh, line 48: `strikethrough` refers to how it looks instead of its function. It should be something like `isWaived` instead.

src/features/speedup/components/SpeedUpModal.tsx
```diff
@@ -196,9 +195,19 @@ export const SpeedUpModal = ({
...
+            <CheckWallet checkNetwork={!isDisabled}>
+              {(isOk) => (
+                <Button
+                  color="primary"
+                  disabled={!isOk || isDisabled}
+                  onClick={onSubmit}
+                  variant="contained"
+                  disableElevation
+                >
+                  {isDisabled ? 'Waiting on confirmation in wallet...' : 'Confirm'}
```
katspaugh, line 207: `Waiting on confirmation in wallet...` – that's a long piece of text for a button, how does it look in this state?

katspaugh (approved):
Code looks good.

@TanyaEfremova could you do a final design and usability review?

jmealy:
> I like how the button looks like, however, it seems that the button and the message are too far from each other. After testing it out, it looks like they are disconnected from each other. I wonder if it would make sense to combine the alert and the button together following the pattern that we already have? (e.g. see creating a wallet screen)

Yeah sure, makes sense! So the warning would have the switch network button and then the execute/sign button would just be disabled, is that right?

src/features/speedup/components/SpeedUpModal.tsx
jmealy: Changed to a loading spinner like we have it on the execute/sign buttons

TanyaEfremova:
Looks good, @jmealy!

francovenica:
The button to go to the safe creation page doesn't work. This was an issue reported in the 4020 ticket a few weeks ago.
image

francovenica:
The rest looks fine.
I'll pass the ticket to done, just make sure to check the button I mentioned in the previous comment

PR #4028, opened 2024-08-02
Refactor: plain img tag for safe app icons

katspaugh:
## What it solves

Safe App icons were previously inserted as an iframe to isolate the cookies. Our Safe App icons are now hosted on the config service and not hot-loaded from Safe Apps' sites, so there's no need to isolate them.

I've updated the CSP accordingly, it will now allow iframes only from https and http (for local custom apps).

## How to test it

App icons should be still visible.


katspaugh:
@usame-algan I should've tested it 🤦 

Looks like config service doesn't allow `crossorigin="anonymous"` so I just removed that attribute altogether since we host these images. For custom apps, we'll assume people trust them.

src/components/safe-apps/SafeAppIconCard/index.tsx
```diff
@@ -46,23 +16,7 @@ const SafeAppIconCard = ({
...
-      sandbox="allow-scripts"
-      referrerPolicy="strict-origin"
-      width={width}
-      height={height}
-      style={{ pointerEvents: 'none', border: 0 }}
-      tabIndex={-1}
-      loading="lazy"
-    />
-  )
+  return <ImageFallback src={src} alt={alt} width={width} height={height} fallbackSrc={fallback} />
```
usame-algan, line 19: Wdyt about keeping the `_isSafeSrc` check? We wouldn't be able to detect if something changes and it could accidentally set cookies or other things for the user.

usame-algan (approved):

src/components/safe-apps/SafeAppIconCard/index.tsx
katspaugh: Restored 👍

src/components/safe-apps/SafeAppIconCard/index.tsx
```diff
@@ -46,21 +32,14 @@ const SafeAppIconCard = ({
...
+    <ImageFallback
+      src={src}
+      alt={alt}
       width={width}
       height={height}
-      style={{ pointerEvents: 'none', border: 0 }}
-      tabIndex={-1}
-      loading="lazy"
+      fallbackSrc={fallback}
+      crossOrigin={_isSafeSrc(src) ? undefined : 'anonymous'}
```
usame-algan, line 42: I would expect that we show the fallback image in case it is an external source. However, from my tests so far the behaviour is that the network request doesn't fail but the image is also not shown. Since the network request didn't fail we don't show the fallback image so it looks broken in the app.

Should it instead say: `src={_isSafeSrc(src) ? src : undefined}`?

Screenshot 2024-08-05 at 15 33 27

katspaugh: Some sites support `crossorigin=anonymous`, some don't. I think ideally it should handle this as an error.

katspaugh: @usame-algan I decided to remove both `crossorigin=anonymous` entirely, as well as `isSafeSrc` because this will not work for forks hosted on other domains. Let's assume config service icons are safe, and custom apps are added at user's own risk.

usame-algan (approved):

francovenica:
LGTM

PR #4029, opened 2024-08-02
feat: blockaid integration

schmanu:
## What it solves

Resolves #

## How this PR fixes it
- exchanges redefine for blockaid
- redesigns warning component

## How to test it
Do any tx or message signing

## TODO
- [ ] Better error handling

## Checklist


katspaugh:
Some unit tests are failing but we can already send this to QA.

schmanu:
> Some unit tests are failing but we can already send this to QA.

They are not really failing. The job just fails when trying to publish results -.- but the test cases all pass.

katspaugh (approved):

.env.example
```diff
@@ -38,8 +38,8 @@ NEXT_PUBLIC_FIREBASE_VAPID_KEY_PRODUCTION=
 NEXT_PUBLIC_FIREBASE_OPTIONS_STAGING=
 NEXT_PUBLIC_FIREBASE_VAPID_KEY_STAGING=
 
-# Redefine
-NEXT_PUBLIC_REDEFINE_API=
+# Blockaid
+NEXT_PUBLIC_BLOCKAID_CLIENT_ID
```
katspaugh, line 42: This var has to be added to the build.yml action too.

francovenica:
I tried using the tool provided and tried to trigger some transactions, but all of them would show this message (bening and malicious ones the same)
image

It seems that the validation of the tx is failing, thus not being able to tell you if the tx is malicious or not.

I got the message to show up once when the SDK initialialize correctly, but I'd assume this is a mistake because the message seems to be there by default as soon as the tx pop's up and then it should be updated depending on if the tx is malicious or not
image

Also I don't think the message should be green if the tx is malicious

schmanu:
@francovenica could you share what those transactions did? 
I cannot reproduce any of those cases :/

francovenica:
I used the tool you shsared with Lily https://examples.blockaid.io/. Every malicious tx would show what I reported in my comment

schmanu:
> I used the tool you shsared with Lily https://examples.blockaid.io/. Every malicious tx would show what I reported in my comment

@francovenica And the last case where it says "Transaction is malicious" in the success color? Do you remember how that happened?

katspaugh:
Blocked on BlockAid's side due to some errors (we reached out for help).

francovenica:
@schmanu I never got to see it twice. There was that SDK error that happened once and the green message just popped. 
I tested several times after and it seems that there is a split second that the green text shows up, like if it is loaded by default, and then is covered by the "could not check" warning message

liliya-soroka:
approved internally together with @kirkkonen

PR #4033, opened 2024-08-03
Fix: multisend position in tx details; "to" field wrapping on small screens; 

katspaugh:
## What it solves

A follow-up on #3967.

Moved multisend to the very bottom (like it was before) and fixed the "To" field wrapping on small screens.

Screenshot 2024-08-03 at 08 53 27

Screenshot 2024-08-03 at 08 52 54


jmealy (commented):
At around 1200px the etherscan link overflows the container:
image

PR #4035, opened 2024-08-05
Feat: total asset value cents in grey

katspaugh:
## What it solves

A bit of extra precision in the total asset value.

Screenshot 2024-08-05 at 12 04 34

### DIFF
--- a/src/components/common/FiatValue/index.tsx
+++ b/src/components/common/FiatValue/index.tsx
@@ -1,13 +1,21 @@
 import type { CSSProperties, ReactElement } from 'react'
 import { useMemo } from 'react'
-import { Tooltip } from '@mui/material'
+import { Tooltip, Typography } from '@mui/material'
 import { useAppSelector } from '@/store'
 import { selectCurrency } from '@/store/settingsSlice'
 import { formatCurrency, formatCurrencyPrecise } from '@/utils/formatNumber'
 
 const style = { whiteSpace: 'nowrap' } as CSSProperties
 
-const FiatValue = ({ value, maxLength }: { value: string | number; maxLength?: number }): ReactElement => {
+const FiatValue = ({
+  value,
+  maxLength,
+  precise,
+}: {
+  value: string | number
+  maxLength?: number
+  precise?: boolean
+}): ReactElement => {
   const currency = useAppSelector(selectCurrency)
 
   const fiat = useMemo(() => {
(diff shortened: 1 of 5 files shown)


src/components/common/FiatValue/index.tsx
```diff
@@ -18,10 +26,26 @@ const FiatValue = ({ value, maxLength }: { value: string | number; maxLength?: n
     return formatCurrencyPrecise(value, currency)
   }, [value, currency])
 
+  const [whole, decimals] = useMemo(() => {
+    const match = preciseFiat.match(/(.+)(\D\d+)$/)
```
usame-algan, line 30: I think we have to consider that this separator can be different across languages, either comma or period.
Screenshot 2024-08-05 at 12 15 57

katspaugh: That's why it matches `\D`.

usame-algan: You are right. It didn't work in my example because of the € at the end but in our code the `preciseFiat` doesn't contain the currency symbol.

src/utils/__tests__/formatNumber.test.ts
```diff
@@ -31,31 +31,31 @@ describe('formatNumber', () => {
 
   describe('formatCurrency', () => {
     it('should format a 0', () => {
-      expect(formatCurrency(0, 'USD')).toBe('$ 0')
+      expect(formatCurrency(0, 'USD')).toBe('$ 0')
```
usame-algan, line 34: Not sure why there is a diff here 🤷

src/components/common/FiatValue/index.tsx
usame-algan: I am actually running into this issue because for me the $ sign is at the end of the value and not in front.

Screenshot 2024-08-05 at 12 44 57

src/utils/formatNumber.ts
```diff
@@ -74,10 +74,10 @@ export const formatCurrency = (number: string | number, currency: string, maxLen
     result = getCurrencyFormatter(currency, true, 2).format(float)
   }
 
-  return result.replace(/^(\D+)/, '$1 ')
+  return result.replace(/^(\D+)/, '$1 ')
```
usame-algan, line 77: This is what it looks like for me now 🤔

Screenshot 2024-08-05 at 12 47 27

katspaugh: I made the space thinner (it was a thin space, now a hair-width space).

src/components/common/FiatValue/index.tsx
katspaugh: Fixed, thanks!

src/utils/formatNumber.ts
usame-algan: Interesting, TIL!

usame-algan: This also only seems to apply if the currency symbol is in front of the value.

src/components/common/FiatValue/index.tsx
```diff
@@ -18,10 +26,27 @@ const FiatValue = ({ value, maxLength }: { value: string | number; maxLength?: n
...
+    <Tooltip title={precise ? undefined : preciseFiat}>
       <span suppressHydrationWarning style={style}>
-        {fiat}
+        {precise ? (
+          <>
+            {whole}
+            {decimals && (
+              <Typography component="span" color="text.secondary" fontSize="inherit">
+                {decimals}
+              </Typography>
```
usame-algan, line 43:
```suggestion
              <Typography component="span" color="text.secondary" fontSize="inherit" fontWeight="bold">
                {decimals}
              </Typography>
```

According to the Figma design it should be bold: https://www.figma.com/design/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=10805-28960&t=fxUQpHt0EkGqo8i4-0

usame-algan (approved):

francovenica:
LGTM

PR #4037, opened 2024-08-05
Analytcs: execution via role event

katspaugh:
## What it solves

Role-based executions were previously tracked as separate events preceding a `tx_executed` event but it's more inline with our existing events to track them as `tx_executed` with a custom `eventAction`.

Edit: extended the same to spending limit transactions.

### 1
* Make a token transfer using Spending Limits
* In the console, the `tx_created` and `tx_executed` events should have a `Created/Executed via spending limit` action

### 2
* Create a Roles-based tx and execute it
* In the console, the `tx_created` and `tx_executed` events should have a `Created/Executed via role` action

### DIFF
--- a/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
+++ b/src/components/tx-flow/flows/TokenTransfer/ReviewSpendingLimitTx.tsx
@@ -119,7 +119,7 @@ const ReviewSpendingLimitTx = ({
     }
 
     trackEvent({ ...TX_EVENTS.CREATE, label: TX_TYPES.transfer_token })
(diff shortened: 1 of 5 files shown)


schmanu (approved):
Makes sense. I think we should also refactor and add `EXECUTE_VIA_SPENDING_LIMIT` to be consistent.

alicecha:
Could we add the same eventAction to the other transaction Custom Events: tx-created (and tx_confirmed + tx_signed if applicable)?

jfschwarz (approved):

alicecha (approved):
Looks good.

liliya-soroka:
image
image

PR #4040, opened 2024-08-06
Tests: add a test for getPeriod

yuzuswapteam002:
## What it solves

Add corresponding test cases for the getPeriod method in date.ts


katspaugh:
@yuzuswapteam002 can you sign the CLA?

PR #4045, opened 2024-08-06
Analytcs: restore WalletConnect tx origin

### LINKED ISSUE #4043: WalletConnect no longer tracked as Safe transaction origin
## Bug description
When a Safe transaction is created through WalletConnect, we no longer "https://apps-portal.safe.global/wallet-connect" as the origin url to the tx-service (since 12/06/24). Instead, we send the url of the WC-connected dApp. This makes it impossible to differentiate between transactions made through a SafeApp and transactions made through a third-party dApp. In the case of Uniswap, for example, the origin url field now returns "https://app.uniswap.org/"  in both cases.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

katspaugh:
## What it solves

Resolves #4043

## How this PR fixes it

To the app, it doesn't matter what URL is being sent in the `origin` of WC-initialted tx proposals. The old WC Safe App doesn't exist anymore, so CGW won't match these txs to anything.
But it matters for data analysis on the backend, so we're restoring the behavior unintentionally altered in #3760.

## How to test it

* Create any transaction via WalletConnect
* Filter the Network Requests in devtools by "propose"
* Observe that the `origin` field in the propose request is set to `https://apps-portal.safe.global/wallet-connect`

### DIFF
--- a/src/features/walletconnect/__tests__/WalletConnectContext.test.tsx
+++ b/src/features/walletconnect/__tests__/WalletConnectContext.test.tsx
@@ -357,7 +357,7 @@ describe('WalletConnectProvider', () => {
             metadata: {
               name: 'name',
               description: 'description',
-              url: 'https://apps-portal.safe.global/wallet-connect',
+              url: 'https://safe-apps.dev.5afe.dev/wallet-connect',
               icons: ['iconUrl'],
             },
           },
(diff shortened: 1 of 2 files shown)


alicecha (approved):
Lgtm

francovenica:
LGTM

A comparison between the cowSwap app in the safe apps list and then connected through WC:
image
image

PR #4050, opened 2024-08-07
Tests: enable PK module for tests in prod

katspaugh:
## What it solves

We now run [scheduled tests in production](https://github.com/safe-global/safe-wallet-web/pull/4049), and these tests require the Private Key wallet module that's been only available on dev so far.

I've removed the `IS_PROD` check so that we can control this wallet via the config service like other wallets, and enable this module on prod Sepolia.


schmanu (approved):
Not a big fan of test strategies leading to unwanted features in prod :/ 
We have to be really careful with not enabling this.

PR #4052, opened 2024-08-08
Fix: no balance skeleton for counterfactual safes

### LINKED ISSUE #4039: CF safes balances on sidebar never finish loading
## Bug description
In the left side where all the safes are listed, if a counter factual safe is there it shows the skeleton of the balance being loaded.

## Environment
 - Browser: Chrome
 - Env: current Dev 1.40

## Steps to reproduce
 1. Create a CF safe
 2. Open the safe list on the sidebar

## Expected result
Show the current balance of the safe since it can have tokens associated to the address 
In case is not possible to fetch balances from a non-activated safe, make it show something different than the skeleton since is never gonna end the loading (maybe not don't show anything at all until activation)

## Obtained result
image

katspaugh:
## What it solves

Resolves #4039

## How this PR fixes it

Don't show a loader for CF safes.


jmealy (approved):

francovenica:
LGTM

PR #4054, opened 2024-08-08
feat: support custom networks from config service [SW-136]

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/76f265123c0a4ad680947ce489804342?v=17c30f9503cd4d48a622209408de56d6&p=e7fed2501e9e4650b30ba9b4f618a804&pm=s

## How this PR fixes it
- If there are safe contract addresses set for a network in the config service, use them to initialize the protocol-kit.
- Only applies when the app is run as a fork, or in dev/staging. Does not apply to the production app.

## Checklist

### DIFF
--- a/src/hooks/coreSDK/safeCoreSDK.ts
+++ b/src/hooks/coreSDK/safeCoreSDK.ts
@@ -5,12 +5,14 @@ import ExternalStore from '@/services/ExternalStore'
 import { Gnosis_safe__factory } from '@/types/contracts'
 import { invariant } from '@/utils/helpers'
 import type { JsonRpcProvider } from 'ethers'
+import type { ContractNetworksConfig } from '@safe-global/protocol-kit'
 import Safe from '@safe-global/protocol-kit'
 import type { SafeVersion } from '@safe-global/safe-core-sdk-types'
-import type { SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
+import type { ChainInfo, SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import semverSatisfies from 'semver/functions/satisfies'
 import { isValidMasterCopy } from '@/services/contracts/safeContracts'
 import { sameAddress } from '@/utils/addresses'
+import { IS_OFFICIAL_HOST, IS_PRODUCTION } from '@/config/constants'
 
 export const isLegacyVersion = (safeVersion: string): boolean => {
   const LEGACY_VERSION = '<1.3.0'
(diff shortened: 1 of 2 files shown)


src/hooks/coreSDK/__tests__/useInitSafeCoreSDK.test.ts
```diff
@@ -44,33 +42,33 @@ describe('useInitSafeCoreSDK hook', () => {
       .mockReturnValue({ query: { safe: `gno:${mockSafeAddress}` } } as unknown as router.NextRouter)
   })
 
-  it('initializes a Core SDK instance', async () => {
-    const mockSafe = {} as Safe
-    const initMock = jest.spyOn(coreSDK, 'initSafeSDK').mockReturnValue(Promise.resolve(mockSafe))
-    const setSDKMock = jest.spyOn(coreSDK, 'setSafeSDK')
+  // it('initializes a Core SDK instance', async () => {
```
katspaugh, line 45: You can use `xit` to comment out a test. Why is it commented out though?

src/hooks/coreSDK/useCustomNetworkContracts.ts
```diff
@@ -0,0 +1,32 @@
+import type { ChainInfo } from '@safe-global/safe-gateway-typescript-sdk'
+import { useCurrentChain } from '../useChains'
+import type { ContractNetworkConfig, ContractNetworksConfig } from '@safe-global/protocol-kit'
+import { IS_OFFICIAL_HOST, IS_PRODUCTION } from '@/config/constants'
+
+const isContractNetworkConfig = (obj: Omit<ChainInfo['contractAddresses'], 'safeWebAuthnSignerFactoryAddress'>) => {
+  return Object.values(obj).every((value) => !!value)
```
katspaugh, line 7: Is this a replacement for `_.isEmpty()`?

The function name is `isContractNetworkConfig` but all this function does is check if all object's properties are truthy.

src/hooks/coreSDK/useCustomNetworkContracts.ts
```diff
@@ -0,0 +1,32 @@
...
+export const useCustomNetworkContracts = () => {
+  const currentChain = useCurrentChain()
+  if (!currentChain || (IS_PRODUCTION && IS_OFFICIAL_HOST)) return
+
+  // Exclude safeWebAuthnSignerFactoryAddress as it is not yet supported.
+  const { safeWebAuthnSignerFactoryAddress, ...contractAddresses } = currentChain.contractAddresses
+
+  if (!isContractNetworkConfig(contractAddresses)) return
+  return contractAddresses as ContractNetworkConfig
+}
```
katspaugh, line 23:
```suggestion
export const useCustomNetworkContracts = IS_PRODUCTION && IS_OFFICIAL_HOST ? () => {
} : () => undefined
```

katspaugh:
@schmanu please also review, there might be an overlap with some of the multichain logic.

src/hooks/coreSDK/useCustomNetworkContracts.ts
jmealy: true, the name doesn't make sense 👍 

The reason for the check is that if the contracts are not defined in the config service admin, the contractNetworks object comes back with all the properties defined but set to null: 
```
const contractAddresses: {
    safeSingletonAddress:  null;
    ...
}
```

src/hooks/coreSDK/useCustomNetworkContracts.ts
jmealy: changed but it should be the opposite way I think:
export const useCustomNetworkContracts = IS_PRODUCTION && IS_OFFICIAL_HOST ? () => undefined  : () => {
...}

schmanu (approved):

PR #4057, opened 2024-08-09
fix: Watch nested safe transactions

### LINKED ISSUE #4025: Improve cleanup of pending transactions state
## Bug description

A pending transaction state is cleaned up whenever one of the following transaction events are emitted:

`TxEvent.SIGNATURE_INDEXED | TxEvent.SUCCESS | TxEvent.REVERTED | TxEvent.FAILED`

As long as there is still a pending transaction we don't allow users to execute another transaction. There are reports that users end up in this state even though there is seemingly no pending transaction.

We should implement a more robust way of cleaning up this state instead of only relying on events that are emitted if the app is opened and the tx watcher is running.

## Steps to reproduce

Need to figure out steps to reproduce this. Simply submitting a transaction and closing the app doesn't work because as soon as the app is reopened, we start the watcher based on the pending tx state and the watcher emits those events.

## Expected result

If there is no pending transaction in the queue there should also be no pending transaction in the local storage

## Obtained result

Transaction is executed but the pending transaction state is still there

usame-algan:
## What it solves

Resolves #4025 

## How this PR fixes it

- Checks for each pending transactions if it is a nested safe transaction and in case it is it polls the CGW until it returns a `txHash` which then starts the usual tx watcher
- Removes `waitForTx` calls from dispatchers and instead runs the `useTxMonitor` hook whenever a pending tx is added

## How to test it

1. Open a 1/n safe that has another safe as an owner
2. Connect to the parent safe
3. Create and execute a transaction
4. Observe in the networks tab that the CGW is called every 15s
5. Once the parent safe executes the transaction the pending state should resolve

## Checklist


katspaugh:
Didn't we decide that the most robust way to cleanup stuck pending txs is checking the safe nonce in the tx flow + add a TTL?

usame-algan:
> Didn't we decide that the most robust way to cleanup stuck pending txs is checking the safe nonce in the tx flow + add a TTL?

I think they are not mutually exclusive. I've added a longer explanation in the issue but TLDR with this PR we would ensure that there are no stuck pending txs in the first place. We can still add a listener on top of the safeInfo and clean up any rogue pending txs.

usame-algan:
Also just realized we don't have any tests for `useTxMonitor` and `useTxPendingStatuses` so I will write them as these two hooks are quite critical.

src/hooks/__tests__/useTxPendingStatus.ts
schmanu, line 1: All other hooks tests follow the `*.test.ts` name scheme

src/hooks/__tests__/useTxPendingStatus.ts
```diff
@@ -0,0 +1,105 @@
...
+  })
+
+  it('should monitor processing transactions', () => {
+    const mockWaitForTx = jest.spyOn(txMonitor, 'waitForTx')
+    const mockWaitForRelayedTx = jest.spyOn(txMonitor, 'waitForRelayedTx')
+
+    const pendingTx: PendingTxsState = {
+      '123': {
+        chainId: '11155111',
+        safeAddress: '0x123',
```
schmanu, line 48: Let's avoid using wrong addresses that are not valid but instead generate legit data.

```suggestion
        safeAddress: faker.finance.ethereumAddress(),
```

schmanu (approved):
I just left some minor comments regarding the test.
Otherwise it looks good. This should fix some stuck pending txs.

The only case that's not covered is if a Safe is using a invalid implementation. In such case the tx will never get indexed and thus never arrive in the tx service. 
For those we could only clear them once the nonce is used up.

francovenica:
Tested with Usame. It seems there is an issue with te multisg call, where is trying to fetch a txhash that doesn't exist in the safe where the tx is trying to be executed
Also the call every 15 secs doesn't happen, it calls twice for me and then it stopped
image

usame-algan:
@francovenica just had another look into this. TLDR: The watcher monitors the correct transaction and those additional calls are happening because of other unrelated logic. Easiest way to test is to go through this flow and stay in the status screen without navigating to the transaction.

Slightly longer version:

The calls to the "wrong" transaction details happen for a couple of reasons.

- The tx details page is open which fetches the details whenever the history tag / queued tag updates
- There is one fetch at the beginning for the notifications and one at the end because the pending state updates

There are some optimizations we can do here which will be addressed in https://github.com/safe-global/safe-wallet-web/issues/4056 so I would leave it out of scope for this PR.

francovenica:
TLDR: I don't find issues with the tx usage (the pending state thing), still see wrong calls even by not switching to the tx view

Long report 
The issue I still have is that there is a call that shows a 404 of the tx that is being executed, and 200 calls of a tx that don't exist. If these are about this optimization you mentioned then we can skip it and close this ticket and re check once the 4056 is done

image

image

Here the txt that showed a 404 that ends up being the one you can see in the etherscan:
https://sepolia.etherscan.io/tx/0xf22efd2f838a0d16d56f324c51ae7efa6e45783eeb3ee0bdff465218b820a1e5

And the one that shows as 200 that doesn't exist  in etherscan:
https://sepolia.etherscan.io/tx/0xde91fc8ac70f74c4b3bc5e4a82ae4e59580bba82be519af18472a051879f4c2b

usame-algan:
> The issue I still have is that there is a call that shows a 404 of the tx that is being executed

This is expected since for every pending transaction we now have to check if its a nested transaction or not.

- Safe A executes Transaction X
- Now Transaction X is pending as it awaits Safe B to sign
- Safe B signs Transaction Y and returns the txHash of Y to Safe A
- Safe A checks if the txHash belongs to a safe transaction - it does so it polls for Transaction Y until it is executed

The same thing happens for Safe B with Transaction Y. Once Y is getting executed it receives a txHash which is an actual txHash. It checks for that txHash if its a safe transaction (this is where the 404 error comes from). Since its not it will start the tx watcher right away.

PR #4064, opened 2024-08-13
feat: show Safe{Pass} points in header

schmanu:
## What this PR changes
- Shows the Safe{Pass} points in the header on Sepolia and Mainnet

## How to test it
- Open a Safe with points on Mainnet
- Check the header

## Screenshots
Screenshot 2024-08-13 at 12 29 32
Screenshot 2024-08-13 at 12 29 28

## Checklist


katspaugh:
@schmanu given the OFAC blocking, should we maybe hide this whole widget if the app is blocked?

schmanu:
> @schmanu given the OFAC blocking, should we maybe hide this whole widget if the app is blocked?

@katspaugh I added it to the ofac PR as that contains the logic for checking for sanctions: https://github.com/safe-global/safe-wallet-web/pull/4066/commits/4a9818d7240bb0219b7f13b5f9aead2c61aabf80

katspaugh:
I meant the geo filter.

schmanu:
> I meant the geo filter.

That is not configured yet. So I wouldn't bother as we cannot even test it rn.

src/components/common/SafeTokenWidget/index.tsx
```diff
@@ -28,25 +31,22 @@ export const getSafeLockingAddress = (chainId: string): string | undefined => {
...
-const canRedeemSep5Airdrop = (allocation?: Vesting[]): boolean => {
-  const sep5Allocation = allocation?.find(({ tag }) => tag === 'user_v2')
-
-  if (!sep5Allocation) {
-    return false
-  }
-
-  return !sep5Allocation.isRedeemed && !sep5Allocation.isExpired
-}
-
```
usame-algan, line 40: Nice!

usame-algan (commented):
Imo we don't need to be this accurate in the header
Screenshot 2024-08-15 at 14 41 26
Instead it could be rounded down or up without any decimals. I've seen other platforms do that as well so I don't think its unexpected to be missing a fraction of a few points/tokens.
Screenshot 2024-08-15 at 14 41 50

src/store/safePass.ts
```diff
@@ -0,0 +1,36 @@
...
+    getOwnGlobalCampaignRank: builder.query<
+      CampaignLeaderboardEntry,
+      { chainId: '1' | '11155111'; safeAddress: string }
+    >({
+      query: (request) => ({
+        url: `v1/community/campaigns/${GLOBAL_CAMPAIGN_IDS[request.chainId]}/leaderboard/${request.safeAddress}`,
+      }),
+    }),
+  }),
+})
```
usame-algan, line 32: I think this could be part of a general `gatewayApi` slice like we added in #4020 and #4017

src/store/safePass.ts
```diff
@@ -0,0 +1,36 @@
+import { cgwDebugStorage } from '@/components/sidebar/DebugToggle'
+import { IS_PRODUCTION, GATEWAY_URL_PRODUCTION, GATEWAY_URL_STAGING } from '@/config/constants'
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react'
+
+const GATEWAY_URL = IS_PRODUCTION || cgwDebugStorage.get() ? GATEWAY_URL_PRODUCTION : GATEWAY_URL_STAGING
```
usame-algan, line 5: I suggest to add a TODO to replace this with an SDK function once the new Gateway SDK is ready.

src/store/safePass.ts
```diff
@@ -0,0 +1,36 @@
...
+const GLOBAL_CAMPAIGN_IDS: Record<'1' | '11155111', string> = {
+  '11155111': 'fa9f462b-8e8c-4122-aa41-2464e919b721',
+  '1': '9ed78b8b-178d-4e25-9ef2-1517865991ee',
+}
```
usame-algan, line 9: Do you know if its possible to fetch points without specifying a campaign id? i.e. the CGW returning points by campaign id for a given address. That way we wouldn't need to hard-code this.

usame-algan (approved):
Looks good! Left some refactor suggestions but no blockers.

src/store/safePass.ts
schmanu: Those PRs are not merged yet. We should consolidate it into once slice when all those PRs are merged.

src/store/safePass.ts
schmanu: Its not possible. There is no global endpoint for points.

src/components/common/SafeTokenWidget/index.tsx
liliya-soroka: Let's use the same rules for rounding in the header as we use in the safe{Pass} to have the exact numbers

PR #4066, opened 2024-08-13
feat: restrict safe pass safe app access for ofac blocked addresses

schmanu:
## What this PR changes
- Reuses ofac detection for Safe{Pass} Safe app
- Reject WalletConnect request from Safe{Pass} for sanctioned addresses.

## How to test it
- Impersonate a blocked address
  - Try to open the Safe{Pass} app (community.safe.global)
  - Try to connect via WalletConnect to community.safe.global

## Screenshots
Screenshot 2024-08-13 at 17 57 00

## Checklist


src/hooks/useSanctionedAddress.ts
```diff
@@ -0,0 +1,22 @@
...
+  const { data: isWalletSanctioned } = useGetIsSanctionedQuery(wallet ? wallet.address : skipToken)
+
+  const { data: isSafeSanctioned } = useGetIsSanctionedQuery(safeAddress !== '' ? safeAddress : skipToken)
+
+  if (isSafeSanctioned) {
+    return safeAddress
+  }
+  if (isWalletSanctioned) {
+    return wallet?.address
+  }
```
usame-algan, line 19: `data` contains the error message in case the `useGetIsSanctionedQuery` throws so I think it would falsely say that an address is sanctioned if the network request fails.

schmanu: I think that is not true. The error seems to be transformed and returned in the `error` field of `useGetIsSanctionedQuery`

usame-algan: Should we remove the data assignment on error then inside the `ofacApi` query?

`return { error: { status: 'CUSTOM_ERROR', data: (error as Error).message } }`

src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -43,6 +45,8 @@ const AppFrame = ({ appUrl, allowedFeaturesList, safeAppFromManifest }: AppFrame
   const chainId = useChainId()
   const chain = useCurrentChain()
   const router = useRouter()
+  const sanctionedAddress = useSanctionedAddress()
```
katspaugh, line 48: This will make a request for all apps not just Safe Apps.

src/features/walletconnect/services/utils.ts
```diff
@@ -50,6 +50,10 @@ export const isBlockedBridge = (origin: string) => {
   return BlockedBridges.some((bridge) => origin.includes(bridge))
 }
 
+export const isSafePassApp = (origin: string) => {
+  return origin.includes('community.safe.global')
```
katspaugh, line 54: Can we if not avoid hardcode then at least make it a constant?

katspaugh:
I think checking blocked addresses from within apps/open and WalletConnect is a bit excessive. It would be better if we just hide all links to this app, and inside the app itself check for blocked addresses (would have to port that hook there).

schmanu:
> I think checking blocked addresses from within apps/open and WalletConnect is a bit excessive. It would be better if we just hide all links to this app, and inside the app itself check for blocked addresses (would have to port that hook there).

You cannot receive the connected wallet within the Safe app. We only get the `safeAddress`. That's why the check is within Safe{Wallet}.

src/features/walletconnect/services/utils.ts
usame-algan: Can we use the constant `SAFE_PASS_URL` here?

usame-algan (approved):

PR #4070, opened 2024-08-14
Fix: native transfer only if value > 0

katspaugh:
## What it solves

Resolves https://www.notion.so/safe-global/Rejection-has-advanced-details-and-it-shouldn-t-f7c7ae2514354c34860d9c3565c20b58

## How this PR fixes it
The value check was incorrect as value is a string.

Screenshot 2024-08-14 at 16 13 19

### DIFF
--- a/src/components/tx-flow/flows/RejectTx/RejectTx.tsx
+++ b/src/components/tx-flow/flows/RejectTx/RejectTx.tsx
@@ -19,7 +19,7 @@ const RejectTx = ({ txNonce }: RejectTxProps): ReactElement => {
   }, [txNonce, setNonce, setSafeTx, setSafeTxError])
 
   return (
-    <SignOrExecuteForm isBatchable={false}>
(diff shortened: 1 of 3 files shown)


usame-algan (approved):
Not a blocker but I would prefer if we didn't show this "Advanced details" block at all. Do you remember why we removed the `if (!decodedData) return null` here? https://github.com/safe-global/safe-wallet-web/pull/3967/files#diff-a8d2c595615c9a0b3cc38b6c9642bf761cc1a63eedf6bbf5e19c3b7049f523aeL75

francovenica:
LGTM

francovenica:
I agree with Usame. If it is possible I'd remove it only for rejections

katspaugh:
I made always appear for consistency with all other transaction. But we can remove it, no problem.

francovenica:
LGTM

PR #4071, opened 2024-08-15
fix: update latest Safe version when switching chains during creation

schmanu:
## What it solves
If a user switches the network during Safe creation the latest Safe version of the first chain was still being used.

## How this PR fixes it
- Updates the latest Safe version when the network changes.

## How to test it
Currently on Prod 1.4.1 is only used on Sepolia.
- Switch network to Sepolia
- Start Safe creation flow
- Go to owner Step
- Now switch network to e.g. Base Sepolia
- Continue Safe creation on Base Sepolia and witness that 1.3.0 is correctly used instead of Sepolia's 1.4.1 version

## Checklist


usame-algan (approved):

katspaugh:
Tested, works correctly. ✅

PR #4075, opened 2024-08-19
feat: SetupToL2 during Safe creation

schmanu:
## ATTENTION THIS IS A DRAFT PR AND NOT MEANT TO BE MERGED YET
This PR hardcodes some addresses and is using an un-audited contract.
DO NOT MERGE.

## What this PR changes
Changes the Safe creation transaction for new Safes to use the SafeToL2Setup library

## Checklist

### DIFF
--- a/src/components/new-safe/create/logic/index.ts
+++ b/src/components/new-safe/create/logic/index.ts
@@ -1,5 +1,5 @@
 import type { SafeVersion } from '@safe-global/safe-core-sdk-types'
-import { type Eip1193Provider, type Provider } from 'ethers'
+import { Interface, type Eip1193Provider, type Provider } from 'ethers'
 
 import { getSafeInfo, type SafeInfo, type ChainInfo, relayTransaction } from '@safe-global/safe-gateway-typescript-sdk'
 import {
@@ -18,6 +18,7 @@ import { isValidSafeVersion } from '@/hooks/coreSDK/safeCoreSDK'
 import { backOff } from 'exponential-backoff'
 import { EMPTY_DATA, ZERO_ADDRESS } from '@safe-global/protocol-kit/dist/src/utils/constants'
 import { getLatestSafeVersion } from '@/utils/chains'
+import { getSafeL2SingletonDeployment } from '@safe-global/safe-deployments'
 
 export type SafeCreationProps = {
   owners: string[]
@@ -25,11 +26,15 @@ export type SafeCreationProps = {
   saltNonce: number
 }
 
-const getSafeFactory = async (provider: Eip1193Provider, safeVersion: SafeVersion): Promise<SafeFactory> => {
+const getSafeFactory = async (
+  provider: Eip1193Provider,
+  safeVersion: SafeVersion,
+  isL1SafeSingleton?: boolean,
+): Promise<SafeFactory> => {
   if (!isValidSafeVersion(safeVersion)) {
     throw new Error('Invalid Safe version')
   }
-  return SafeFactory.init({ provider, safeVersion })
+  return SafeFactory.init({ provider, safeVersion, isL1SafeSingleton })
 }
 
 /**
@@ -39,8 +44,9 @@ export const createNewSafe = async (
   provider: Eip1193Provider,
   props: DeploySafeProps,
   safeVersion: SafeVersion,
+  isL1SafeSingleton?: boolean,
 ): Promise<Safe> => {
-  const safeFactory = await getSafeFactory(provider, safeVersion)
+  const safeFactory = await getSafeFactory(provider, safeVersion, isL1SafeSingleton)
   return safeFactory.deploySafe(props)
 }
 
@@ -63,8 +69,12 @@ export const computeNewSafeAddress = async (
       saltNonce: props.saltNonce,
       safeVersion: safeVersion ?? getLatestSafeVersion(chain),
     },
+    isL1SafeSingleton: true,
   })
 }
+
+export const SAFE_TO_L2_SETUP_ADDRESS = '0x80E0d1577aD3d982BF2F49aAB00BfA161AA763c4'
+export const SAFE_TO_L2_SETUP_INTERFACE = new Interface(['function setupToL2(address l2Singleton)'])
 
 /**
  * Encode a Safe creation transaction NOT using the Core SDK because it doesn't support that
@@ -76,18 +86,19 @@ export const encodeSafeCreationTx = async ({
   saltNonce,
   chain,
   safeVersion,
-}: SafeCreationProps & { chain: ChainInfo; safeVersion?: SafeVersion }) => {
+}: SafeCreationProps & { chain: ChainInfo; safeVersion?: SafeVersion; to?: string; data?: string }) => {
   const usedSafeVersion = safeVersion ?? getLatestSafeVersion(chain)
-  const readOnlySafeContract = await getReadOnlyGnosisSafeContract(chain, usedSafeVersion)
+  const readOnlyL1SafeContract = await getReadOnlyGnosisSafeContract(chain, usedSafeVersion, true)
+  const l2Deployment = getSafeL2SingletonDeployment({ version: safeVersion, network: chain.chainId })
   const readOnlyProxyContract = await getReadOnlyProxyFactoryContract(usedSafeVersion)
   const readOnlyFallbackHandlerContract = await getReadOnlyFallbackHandlerContract(usedSafeVersion)
 
   // @ts-ignore union type is too complex
-  const setupData = readOnlySafeContract.encode('setup', [
+  const setupData = readOnlyL1SafeContract.encode('setup', [
     owners,
     threshold,
-    ZERO_ADDRESS,
-    EMPTY_DATA,
+    SAFE_TO_L2_SETUP_ADDRESS,
+    SAFE_TO_L2_SETUP_INTERFACE.encodeFunctionData('setupToL2', [l2Deployment?.defaultAddress]),
     await readOnlyFallbackHandlerContract.getAddress(),
     ZERO_ADDRESS,
     '0',
@@ -95,7 +106,7 @@ export const encodeSafeCreationTx = async ({
   ])
 
   return readOnlyProxyContract.encode('createProxyWithNonce', [
-    await readOnlySafeContract.getAddress(),
+    await readOnlyL1SafeContract.getAddress(), // always L1 Mastercopy
     setupData,
     BigInt(saltNonce),
   ])
--- a/src/components/new-safe/create/steps/ReviewStep/index.tsx
+++ b/src/components/new-safe/create/steps/ReviewStep/index.tsx
@@ -5,7 +5,13 @@ import { safeCreationDispatch, SafeCreationEvent } from '@/features/counterfactu
 import { getTotalFeeFormatted } from '@/hooks/useGasPrice'
 import type { StepRenderProps } from '@/components/new-safe/CardStepper/useCardStepper'
 import type { NewSafeFormData } from '@/components/new-safe/create'
-import { computeNewSafeAddress, createNewSafe, relaySafeCreation } from '@/components/new-safe/create/logic'
+import {
+  computeNewSafeAddress,
+  createNewSafe,
+  relaySafeCreation,
+  SAFE_TO_L2_SETUP_ADDRESS,
+  SAFE_TO_L2_SETUP_INTERFACE,
+} from '@/components/new-safe/create/logic'
 import { getAvailableSaltNonce } from '@/components/new-safe/create/logic/utils'
 import NetworkWarning from '@/components/new-safe/create/NetworkWarning'
 import css from '@/components/new-safe/create/steps/ReviewStep/styles.module.css'
@@ -38,7 +44,7 @@ import { type ChainInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import classnames from 'classnames'
 import { useRouter } from 'next/router'
 import { useMemo, useState } from 'react'
-import { getLatestSafeVersion } from '@/utils/chains'
+import { getSafeL2SingletonDeployment } from '@safe-global/safe-deployments'
 
 export const NetworkFee = ({
   totalFee,
@@ -130,8 +136,6 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
   const canRelay = hasRemainingRelays(minRelays)
   const willRelay = canRelay && executionMethod === ExecutionMethod.RELAY
 
-  const latestSafeVersion = getLatestSafeVersion(chain)
-
   const safeParams = useMemo(() => {
     return {
       owners: data.owners.map((owner) => owner.address),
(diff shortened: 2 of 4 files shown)


src/features/counterfactual/PayNowPayLater.tsx
```diff
@@ -42,6 +45,12 @@ const PayNowPayLater = ({
       <Typography variant="h4" fontWeight="bold">
         Before you continue
       </Typography>
+      {isMultiChain && (
+        <ErrorMessage level="info">
+          You will need to <b>activate your account</b> separately, on each network. Make sure you have funds on your
```
schmanu, line 50: I feel like "Make sure you have funds on your wallet..." sounds like the user needs to have funds to continue. I might drop that sentence as the user only needs funds at activation time.

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -0,0 +1,76 @@
...
+import css from './styles.module.css'
+
+const NetworkMultiSelector = ({
+  selectedNetworks,
+  setSelectedNetworks,
+}: {
+  selectedNetworks: ChainInfo[]
+  setSelectedNetworks: (selectedChains: ChainInfo[]) => void
+}): ReactElement => {
+  const { configs } = useChains()
```
schmanu, line 18: We cannot offer all chains.

We have to make sure that the chains have matching contracts deployed.
Those are all chains that have a canonical 1.4.1 deployment.

src/features/counterfactual/PayNowPayLater.tsx
schmanu: This component still renders the text "If you choose to pay later, the fee will be included with the first transaction you make." for multichain Safes.

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -0,0 +1,76 @@
...
+import useWallet from '@/hooks/wallets/useWallet'
+import { getNetworkLink } from '.'
+import css from './styles.module.css'
+
+const NetworkMultiSelector = ({
+  selectedNetworks,
+  setSelectedNetworks,
+}: {
+  selectedNetworks: ChainInfo[]
+  setSelectedNetworks: (selectedChains: ChainInfo[]) => void
```
schmanu, line 16: We could get rid of those props if this component uses `useFormContext` to pass the values to the react-hook-form.

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
schmanu: We still need to offer a way to deploy Safes on networks that do not support 1.4.1 yet though. 
Maybe we could make it dependent on the first selected network and disable all networks that are not compatible.

schmanu (commented):
It looks and works well.
There are some caveats though that we have to consider.

src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -239,6 +272,7 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
             },
           },
           data.safeVersion,
```
schmanu, line 274: We have to adjust the available `saltNonce` computation as the nonce has to be available on all selected chains and also not exist on any other chains yet.

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
jmealy: I made the changes to the options here, can you check it again and make sure it behaves as expected?

src/components/new-safe/create/steps/ReviewStep/index.tsx
schmanu: I adjusted the `saltNonce` computation in https://github.com/safe-global/safe-wallet-web/pull/4075/commits/8aa29c6c382cda64c105abefe1bb6ea8b6042d82

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -35,6 +42,20 @@ const NetworkMultiSelector = ({ name }: { name: string }): ReactElement => {
     router.push(networkLink)
   }
 
+  const isOptionDisabled = (chain: ChainInfo) => {
```
schmanu: Instead of hardcoding zkSync to be excluded we should check the `safe-deployments` and check that they have a canonical address deployed for version 1.4.1 .

Here is a snippet that does that would be this for instance:
```ts
  const hasCanonicalSingletonDeployment = Boolean(
    getSafeSingletonDeployment({ version: data.safeVersion })?.deployments.canonical,
  )
```

The reason is that there could always be chains that do not have 1.4.1 support yet or that do not share the canonical deployment.

src/components/new-safe/create/logic/utils.ts
```diff
@@ -3,23 +3,52 @@ import { isSmartContract } from '@/utils/wallets'
...
+  const allRPCs = await Promise.all(
+    chains.map((chain) => {
+      const rpcUrl = customRpcs?.[chain.chainId] || getRpcServiceUrl(chain.rpcUri)
+      // Turn into Eip1993Provider
+      return {
+        rpcUrl,
+        chainId: chain.chainId,
+      }
+    }),
+  )
```
usame-algan, line 29: Does this need to be wrapped in a Promise.all? It doesn't seem to be async.

src/components/new-safe/create/steps/OwnerPolicyStep/useSafeSetupHints.ts
```diff
@@ -18,18 +19,26 @@ export const useSafeSetupHints = (
     }
```
usame-algan, line 19: Do we need to also adjust this condition in line 13 because of the change to optional parameters? In theory it could happen now that we display something like `1/undefined policy` in case we pass a threshold but no noOwners.

src/components/new-safe/create/useSyncSafeCreationStep.ts
```diff
@@ -2,19 +2,20 @@ import { useEffect } from 'react'
...
+const useSyncSafeCreationStep = (setStep: StepRenderProps<NewSafeFormData>['setStep'], networks: ChainInfo[]) => {
   const wallet = useWallet()
-  const isWrongChain = useIsWrongChain()
+  const currentChain = useCurrentChain()
 
   useEffect(() => {
-    // Jump to choose name and network step if the wallet is connected to the wrong chain and there is no pending Safe
-    if (!wallet || isWrongChain) {
+    // Jump to choose name and network step if there is no pending Safe or if the selected network does not match the connected network
+    if (!wallet || (networks.length === 1 && currentChain?.chainId !== networks[0].chainId)) {
```
usame-algan, line 14: Not sure if we should add this length 1 limitation or not. Currently, it is possible to select more than one network, go to the next step and switch to a network that is not selected and nothing happens. I think it would me more intuitive if it jumped back to the network selection in this case so that the user can add the current network to the selected networks as well. Wdyt?

jmealy: I'm not sure we even need the network selector during safe creation anymore. There is another PR that changes the network selector while you have a safe open to just show other chains in the multichain safe group, and additional chains you can deploy to, so the purpose of the network selector has changed a bit.

But before you've created a safe it doesn't really serve any purpose. If you want to change the networks you're deploying to I think the intuitive thing to do would be to go back to that step using the back button. And for multichain safes it doesn't matter what the currently selected network is.

The only thing the current network is used for now is if you're deploying a single chain safe, it will check the current network (which should be aligned with the network to be deployed) against the wallet network and show a warning if they're different. but we could just check the network to be deployed directly instead, and then the current network wouldn't be needed at all.

@schmanu wdyt?

usame-algan: I agree that the network selector in the header on non-safe routes could be removed now. Probably makes sense to discuss it in #4124.

In the context of safe creation the behaviour as I understand it:

- If more than one network is selected there is only the pay later option for which the wallet network doesn't matter. If the user however changes the network through the header we ignore whatever they select. My suggestion: Remove the header network selector as part of #4124 

- If only one network is selected and if the user wants to pay now the wallet network matters as they have to execute a transaction. Since they might also have the option to pay later there is no reason to jump back to step 1 but instead show an info box and let them switch the network before submitting so this is in line with the current implementation.

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -0,0 +1,129 @@
...
+import css from './styles.module.css'
+import { Controller, useFormContext, useWatch } from 'react-hook-form'
+import { useRouter } from 'next/router'
+import { getNetworkLink } from '.'
+import useWallet from '@/hooks/wallets/useWallet'
+import { SetNameStepFields } from '@/components/new-safe/create/steps/SetNameStep'
+import { LATEST_SAFE_VERSION } from '@/config/constants'
+import { getSafeSingletonDeployment } from '@safe-global/safe-deployments'
+
+const ZKSYNC_ID = '324'
```
usame-algan, line 15: This is not used. Do we need to have add an exception for zksync?

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const isWalletConnected = !!useWallet()
+
+  const {
+    formState: { errors },
+    control,
+    getValues,
+    setValue,
+  } = useFormContext()
+
+  const selectedNetworkSafeVersion = useWatch({ control, name: SetNameStepFields.safeVersion })
```
usame-algan, line 35:
```suggestion

```

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -0,0 +1,129 @@
...
+    router.push(networkLink)
+  }
+
+  const isOptionDisabled = (optionNetwork: ChainInfo) => {
+    if (selectedNetworks.length === 0) return false
+
+    // do not allow multi chain safes for advanced setup flow.
+    if (isAdvancedFlow) return optionNetwork.chainId != selectedNetworks[0].chainId
+
+    // const hasCanonicalSingletonDeployment = getSafeSingletonDeployment({ version: LATEST_SAFE_VERSION })?.deployments
```
usame-algan, line 58:
```suggestion

```

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const optionHasCanonicalSingletonDeployment = Boolean(
+      getSafeSingletonDeployment({ network: optionNetwork.chainId, version: LATEST_SAFE_VERSION })?.deployments
+        .canonical,
+    )
+    const selectedHasCanonicalSingletonDeployment = Boolean(
+      getSafeSingletonDeployment({ network: selectedNetworks[0].chainId, version: LATEST_SAFE_VERSION })?.deployments
```
usame-algan, line 64: I suggest extracting `selectedNetworks[0]` into a variable for better readability. Afaiu it is the default selected network that we compare all other networks against.

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const updatedValues = currentValues.filter((chain) => chain.chainId !== deletedChainId)
+    updateSelectedNetwork(updatedValues)
+    setValue(name, updatedValues)
+  }
+
+  const updateSelectedNetwork = (chains: ChainInfo[]) => {
+    if (chains.length !== 1) return
+    const shortName = chains[0].shortName
+    const networkLink = getNetworkLink(router.pathname, router.query, shortName, isWalletConnected)
+    router.push(networkLink)
```
usame-algan, line 49: I think router.replace would make more sense here. I tried going back in the history and while the url parameter updates, the selector doesn't so its not working anyway.

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -0,0 +1,129 @@
...
+        control={control}
+        defaultValue={[]}
+        render={({ field }) => (
+          <Autocomplete
+            {...field}
+            multiple
+            value={field.value || []}
+            disableCloseOnSelect
+            options={configs}
+            renderTags={(selectedOptions, getTagProps) =>
```
usame-algan, line 86:
```suggestion
            renderTags={(selectedOptions) =>
```

src/components/new-safe/create/logic/index.ts
```diff
@@ -65,9 +71,13 @@ export const computeNewSafeAddress = async (
       saltNonce: props.saltNonce,
       safeVersion: safeVersion ?? getLatestSafeVersion(chain),
     },
+    isL1SafeSingleton: true,
   })
 }
 
+export const SAFE_TO_L2_SETUP_ADDRESS = '0x80E0d1577aD3d982BF2F49aAB00BfA161AA763c4'
```
usame-algan, line 78: Is it already clear where this address will be stored and if it will be the same on every chain?

usame-algan (commented):
Found a few small things but looks good overall!

One thing to note is that with this PR we effectively remove the limitation to only have one counterfactual safe at the same time which can lead to increased RPC requests. We should define the behaviour of what to expect if any of those RPC requests fail during safe creation and whether we want that limitation or not.

src/components/new-safe/create/logic/index.ts
jmealy: Might be better to store it in constants.ts? 
I assume it will be the same on every chain but @schmanu may be better able to say

usame-algan (approved):
Thanks!

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  const updateSelectedNetwork = (chains: ChainInfo[]) => {
+    if (chains.length !== 1) return
+    const shortName = chains[0].shortName
+    const networkLink = getNetworkLink(router.pathname, router.query, shortName, isWalletConnected)
+    router.replace(networkLink)
+  }
+
+  const isOptionDisabled = (optionNetwork: ChainInfo) => {
```
schmanu, line 49: I am not sure if this function and the others could cause re-renders if they are not using `useCallback`

src/components/common/NetworkSelector/NetworkMultiSelector.tsx
```diff
@@ -0,0 +1,126 @@
...
+    if (selectedNetworks.length === 0) return false
+    const firstSelectedNetwork = selectedNetworks[0]
+
+    // do not allow multi chain safes for advanced setup flow.
+    if (isAdvancedFlow) return optionNetwork.chainId != firstSelectedNetwork.chainId
+
+    const optionHasCanonicalSingletonDeployment = Boolean(
+      getSafeSingletonDeployment({ network: optionNetwork.chainId, version: LATEST_SAFE_VERSION })?.deployments
```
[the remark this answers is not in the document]
schmanu: It needs to have a canonical deployment for the same Safe version that the first selected chain supports.

src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -147,40 +170,82 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
...
+      if (!safeL2Address) {
+        throw new Error('No Safe deployment found')
+      }
 
       const props: DeploySafeProps = {
         safeAccountConfig: {
           threshold: data.threshold,
           owners: data.owners.map((owner) => owner.address),
           fallbackHandler: await readOnlyFallbackHandlerContract.getAddress(),
+          to: SAFE_TO_L2_SETUP_ADDRESS,
```
schmanu, line 201: If the network / safe version is only 1.3.0 or lower we cannot use the setupToL2 flow.

schmanu (commented):
I think we need to adjust the check for the canonical Safe version a little bit.

It could be that the first selected network only supports 1.3.0. In that case we should at least check that the other networks support the 1.3.0 canonical network.
Or we should not allow selecting any other networks in general.

For Safes < 1.4.1 we should also not use the SetupToL2 flow.

PR #4077, opened 2024-08-20
Analytics: restore safe address in the safe_created event

katspaugh:
## What it solves

#3954 restored the `safe_created` event but not the safe address for it.

## How this PR fixes it

I've added a `gtmSetSafeAddress` call right before the `safe_created` event is emitted.


schmanu (approved):

francovenica:
LGTM

I tried safe creation with wallet, with relayer and CF safe creation. The safe-create event is there with the safe address.

Leaving a snapshot for documentation.
image

PR #4073, opened 2024-08-16
Analytics: ecosystem id in safe creation params

katspaugh:
## What it solves

Resolves [Onchain-Safe-creation-tracking](https://www.notion.so/safe-global/Onchain-Safe-creation-tracking-9fff9569fbb843cda6357854ce733236)

## How this PR fixes it
I've added a new env variable that sets a custom `paymentReceiver` value in the Safe creation tx. This acts like an affiliate link to tag Safe creations via the Wallet.

## How to test it

Create a Safe on Sepolia using 3 different methods:

* Counterfactual
* Regular creation paid from the wallet
* Relayed creation

Open the creation transaction in Etherscan, it should have `0x5afe7A11E7000000000000000000000000000000` in the `initializer` field.

Screenshot 2024-08-16 at 14 24 01

PR #4062, opened 2024-08-12
feat: upgrade zodiac to enable recovery on Base

schmanu:
## What it solves

Resolves #4061 

## How this PR fixes it
Upgrades Zodiac to 4.0.3

## How to test it
- Enable Recovery on Base (Feature flag)
- Test recovery

## Screenshots
Screenshot 2024-08-12 at 14 48 20

## Checklist
* [x] I've tested the branch on mobile 📱
* [x] I've documented how it affects the analytics (if at all) 📊
* [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #4078, opened 2024-08-20
Release 1.41.2

katspaugh:
## Features
* feat: upgrade zodiac to enable recovery on Base (#4062)

## Analytics
* Analytics: restore safe address in the safe_created event (#4077)
* Analytics: ecosystem id in safe creation params (#4073)

## Refactors
* Refactor: plain img tag for safe app icons (#4028)

## Tests
* Tests: add a test for getPeriod (#4040)
* Tests: fix create_safe_simple (#4074)


francovenica:
Note: I wasn't able to check here the recovery in Base network. The zodiac app is not present, only zodiac roles.
@schmanu can you try here the same tests you did in your original PR?

Besides that the RC looks good. 
I've checked the issues that automation found but it seems is just flaky tests. They work fine when tested manually

Also did a small tx execution regression to make sure nothing broke

katspaugh:
I've tested recovery by temporarily enabling it on Base, it works. ✅ 
Screenshot 2024-08-21 at 09 00 56

PR #519, opened 2022-09-06
Feature: Share safe app

mikhailxyz:
## What it solves

Resolves #

PR #4085, opened 2024-08-22
Fix: limit safeViewRedirectURL redirects to Safe App URLs

katspaugh:
## What it solves

Limits redirects to known URLs.

## How this PR fixes it

The redirect was introduced here: #519 – to allow creating a Safe from a Safe App share URL and then going back to that Safe App.

## How to test it
* Open a share link of any Safe App, e.g. https://app.safe.global/share/safe-app?appUrl=https%3A%2F%2Fsafe-eth.everstake.one%2F&chain=eth
* Connect a wallet that doesn't have Safes on that chain
* Click on Create new Safe Account
* Complete a Safe creation
* You'll be redirected back to that Safe App

Screenshot 2024-08-22 at 12 43 42


schmanu (approved):
Good catch!

francovenica:
Works fine. LGTM

Note: I found an unexpected behaviour with that list of safes in the App page. I created an issue for that here:
https://github.com/safe-global/safe-wallet-web/issues/4093

PR #4089, opened 2024-08-22
Release 1.41.3

katspaugh:
A hot fix for tx details: contract method parameters must be always displayed.

Screenshot 2024-08-22 at 15 58 05


schmanu (approved):

schmanu:
Tested this PR with two transactions that have a value whilst calling a Payable method. 
It now shows the parameters correctly!

PR #4090, opened 2024-08-22
feat: show multichain accounts in sidebar

schmanu:
## What this PR changes
Groups Safes with the same address in the sidebar.

## How to test it
- Create Safes with the same address on some networks.
- Check and observe the sidebar:
  - Multichain Safes should appear first
- The "add network" button has no function yet. It will be added in the PR that implements the action!

## Screenshots
Screenshot 2024-08-22 at 16 13 44

## Checklist

### DIFF
--- a/src/components/welcome/MyAccounts/SubAccountItem.tsx
+++ b/src/components/welcome/MyAccounts/SubAccountItem.tsx
@@ -10,10 +10,8 @@ import { OVERVIEW_EVENTS, OVERVIEW_LABELS } from '@/services/analytics'
 import { AppRoutes } from '@/config/routes'
 import { useAppSelector } from '@/store'
 import { selectChainById } from '@/store/chainsSlice'
-import ChainIndicator from '@/components/common/ChainIndicator'
 import css from './styles.module.css'
 import { selectAllAddressBooks } from '@/store/addressBookSlice'
-import { shortenAddress } from '@/utils/formatters'
 import SafeListContextMenu from '@/components/sidebar/SafeListContextMenu'
 import useSafeAddress from '@/hooks/useSafeAddress'
 import useChainId from '@/hooks/useChainId'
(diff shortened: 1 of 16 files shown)


src/components/welcome/MyAccounts/SubAccountItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const isSingleTxPage = router.pathname === AppRoutes.transactions.tx
+
+  const trackingLabel = isWelcomePage ? OVERVIEW_LABELS.login_page : OVERVIEW_LABELS.sidebar
+
+  /**
+   * Navigate to the dashboard when selecting a safe on the welcome page,
+   * navigate to the history when selecting a safe on a single tx page,
+   * otherwise keep the current route
+   */
+  const getHref = useCallback(
```
jmealy, line 50: We could make this a hook since it's used again in the other list

jmealy, line 50: We could make this a hook since it's used again in the other account item component

src/services/analytics/events/overview.ts
```diff
@@ -117,6 +117,12 @@ export const OVERVIEW_EVENTS = {
     category: OVERVIEW_CATEGORY,
     //label: OPEN_SAFE_LABELS
   },
+  // Track clicks on links to Safe Accounts
+  EXPAND_MULTI_SAFE: {
+    action: 'Expand multi Safe',
+    category: OVERVIEW_CATEGORY,
+    //label: OPEN_SAFE_LABELS
```
jmealy, line 124: should the label be included?

jmealy (changes requested):
The selected state looks a bit off.
- The vertical line seems to be thicker than the horizontal:
- ~I would leave the lines on the left as grey in the selected state.~
- ~I think the divider line before the `add another network` button should be green also.~
- In the designs the outer border is not highlighted at all

image

jmealy:
A suggestion: when I was testing it I wanted to delete the whole group, but the only way to do it was one by one. Could it be useful to have a `delete group` button somewhere?

src/services/analytics/events/overview.ts
schmanu: Its just a comment that should signal that its an even which uses one of OPEN_SAFE_LABELS as label. (Copied it from a related event)

schmanu:
> A suggestion: when I was testing it I wanted to delete the whole group, but the only way to do it was one by one. Could it be useful to have a `delete group` button somewhere?

@jmealy Usually not the entire group is counterfactual and currently only counterfactual Safes can be removed. So we could only offer it if all Safes are counterfactual which would be inconsistent (some safes have that button and some don't).
I'd say lets keep it like this for now. I think in a later iteration we wanted to add a context menu for the parent Safe.

jmealy (approved):

PR #4112, opened 2024-08-26
Fix: don't show "verified by WalletConnect"

katspaugh:
## What it solves

As requested by WalletConnect, we should not show `has been verified by WalletConnect` as it's not 100% bullet-proof. See https://docs.walletconnect.com/walletkit/web/verify#disclaimer

So from now on, we'll be showing only if a dApp is a known scam or there's a domain mismatch.

Screenshot 2024-08-26 at 10 14 15


schmanu (approved):

francovenica:
I was able to verify that the "verified by WC" doesn't show anymore. It can be tested with uniswap and sablier apps

I cannot find apps that trigger the "is a scam" error message tho.

PR #4116, opened 2024-08-27
Feat: replay safe creation with same address

schmanu:
##  What this PR changes
- Adds logic to replay a Safe creation with its initial setup to get the same address
- Adds logic to activate replayed counterfactual Safes through the connected wallet and through relayers.
- Adds "Add new network" to Safe's context menu in the sidebar

## How to test it
- Click "Add new network"
- Select a available network

## TODO
- [x] Better error handling
- [x] Replaying Counterfactual Safes

## Checklist

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -5,6 +5,7 @@
 // Learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom/extend-expect'
 import { TextEncoder, TextDecoder } from 'util'
+import { Headers, Request, Response } from 'node-fetch'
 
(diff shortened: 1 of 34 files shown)


jmealy (commented):
I created a multichain safe and it seems to always be displayed under owned safes regardless of which wallet I'm connected with

jmealy (commented):
We could remove the testnets header if there's no testnets available:
image

schmanu:
> I created a multichain safe and it seems to always be displayed under owned safes regardless of which wallet I'm connected with

This is how the current Sidebar already works. Counterfactual Safes never appear on the watchlist but instead always under "My accounts". So I'd leave it as is for now.

PR #4118, opened 2024-08-27
Chore: upgrade protocol-kit

katspaugh:
## What it solves

An upgrade of the protocol-kit that has static network enabled for the JSON RPC provider to reduce the number of RPC calls.


katspaugh:
Looks like the passkeys support has some breaking changes. Need to adjust our code.

PR #4125, opened 2024-08-30
[Multichain] feat: redesign of multiaccounts and context menu for multiaccounts

schmanu:
## What this PR changes
- Redesigns how sub-items are displayed under multi-accounts.
- Adds context menu for multi accounts offering renaming and adding a new network
- Some UX fixes around the context menu

## How to test it
- Open the new Sidebar and observe multi chain accounts
- Try to rename an account

## Checklist


src/components/sidebar/SafeListContextMenu/MultiAccountContextMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+const MultiAccountContextMenu = ({
+  name,
+  address,
+  chainIds,
+}: {
+  name: string
+  address: string
+  chainIds: string[]
+}): ReactElement => {
+  const [anchorEl, setAnchorEl] = useState<HTMLElement | undefined>()
```
clovisdasilvaneto, line 35:
```suggestion
  const [anchorEl, setAnchorEl] = useRef<HTMLElement>(null)
```

src/components/sidebar/SafeListContextMenu/MultiAccountContextMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const handleCloseModal = () => {
+    setOpen(defaultOpen)
+  }
+
+  return (
+    <>
+      <IconButton data-testid="safe-options-btn" edge="end" size="small" onClick={handleOpenContextMenu}>
+        <MoreVertIcon sx={({ palette }) => ({ color: palette.border.main })} />
+      </IconButton>
+      <ContextMenu anchorEl={anchorEl} open={!!anchorEl} onClose={handleCloseContextMenu}>
```
clovisdasilvaneto, line 69:
```suggestion
      <ContextMenu anchorEl={anchorEl.current}>
```

src/components/sidebar/SafeListContextMenu/MultiAccountContextMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+      trackEvent({ ...event, label: trackingLabel })
+    }
+
+  const handleCloseModal = () => {
+    setOpen(defaultOpen)
+  }
+
+  return (
+    <>
+      <IconButton data-testid="safe-options-btn" edge="end" size="small" onClick={handleOpenContextMenu}>
```
clovisdasilvaneto, line 66:
```suggestion
      <IconButton data-testid="safe-options-btn" edge="end" size="small" onClick={handleOpenContextMenu} ref={anchorEl}>
```

src/components/sidebar/SafeListContextMenu/MultiAccountContextMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { CreateSafeOnNewChain } from '@/features/multichain/components/CreateSafeOnNewChain'
+
+enum ModalType {
+  RENAME = 'rename',
+  ADD_CHAIN = 'add_chain',
+}
+
+const defaultOpen = { [ModalType.RENAME]: false, [ModalType.ADD_CHAIN]: false }
+
+const MultiAccountContextMenu = ({
```
clovisdasilvaneto, line 26: about the architecture of this component, since it's a button that shows a dropdown with some options which triggers a modal, could we separate the dropdown in generic renderProps component which provides the `openMenu` function to the rendered child and takes an array of options? something like: 

```jsx
<DropdownMenu items={items}>
{(openMenu) =>  (
<IconButton data-testid="safe-options-btn" edge="end" size="small" onClick={handleOpenContextMenu}>
      <MoreVertIcon sx={({ palette }) => ({ color: palette.border.main })} />
</IconButton>
)}
</DropdownMenu>
```

in this case, `items` would be a common type like this:

```typescript
type item = {
 onClick: (e: MouseEvent) => void,
 label: string
 icon?: React.ReactNode
}[]
```

src/components/welcome/MyAccounts/MultiAccountItem.tsx
```diff
@@ -111,20 +113,35 @@ const MultiAccountItem = ({ onLinkClick, multiSafeAccountItem, safeOverviews }:
...
+            <Tooltip
+              title={
+                <Box>
+                  <Typography fontSize="14px">Multichain account on:</Typography>
+                  {safes.map((safeItem) => (
+                    <Box p="4px 0px" key={safeItem.chainId}>
+                      <ChainIndicator chainId={safeItem.chainId} />
+                    </Box>
+                  ))}
+                </Box>
```
clovisdasilvaneto, line 132: Would be nice if we can move it to be a separated component outside the `MultiAccountItem`? So we just need to call the reference of the component inside the tooltip instead of re-creating the component every-time the parent has a re-render?

src/components/welcome/MyAccounts/styles.module.css
```diff
@@ -44,6 +44,12 @@
   background-color: var(--color-background-light) !important;
 }
 
+.currentListItem.multiListItem {
+  border-left-width: 1px;
```
clovisdasilvaneto, line 48: I think you don't need this first line, since the next one overwrites this one 🤔

src/components/address-book/EntryDialog/index.tsx
```diff
@@ -41,7 +41,11 @@ function EntryDialog({
   const { handleSubmit, formState } = methods
 
   const submitCallback = handleSubmit((data: AddressEntry) => {
-    dispatch(upsertAddressBookEntry({ ...data, chainId: chainId || currentChainId }))
+    if (chainIds) {
+      chainIds?.forEach((chainId) => dispatch(upsertAddressBookEntry({ ...data, chainId })))
```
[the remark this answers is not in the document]
schmanu: Makes sense. Especially considering that we cannot really handle the case that one of those update unexpectedly fails otherwise.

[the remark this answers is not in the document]
clovisdasilvaneto: true

src/components/welcome/MyAccounts/styles.module.css
schmanu: There is the single account component which uses only `currentListItem` without the `multiListItem` css class. 
I think its needed there.

src/components/sidebar/SafeListContextMenu/MultiAccountContextMenu.tsx
schmanu: Hmm internally we use the Mui Menu component (https://mui.com/material-ui/react-menu/)
All examples how to use it use the anchorEl property and do not work with refs. I think this is the intended way. 

I will leave it as is for now. If you feel like it could be improved we could do it separately as we use this ContextMenu component in other places in the app as well :)

src/components/sidebar/SafeListContextMenu/MultiAccountContextMenu.tsx
schmanu: The architecture also comes from the MUI component library. I do not think we can easily rewrite it to this. 
So I'll stick to how the examples use the component for the time being.

src/components/sidebar/SafeListContextMenu/MultiAccountContextMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+    setAnchorEl(e.currentTarget)
+  }
+
+  const handleCloseContextMenu = (event: MouseEvent) => {
+    event.stopPropagation()
+    setAnchorEl(undefined)
+  }
+
+  const handleOpenModal =
+    (type: keyof typeof open, event: typeof OVERVIEW_EVENTS.SIDEBAR_RENAME | typeof OVERVIEW_EVENTS.SIDEBAR_RENAME) =>
```
usame-algan, line 52:
```suggestion
    (type: ModalType, event: typeof OVERVIEW_EVENTS.ADD_NEW_NETWORK | typeof OVERVIEW_EVENTS.SIDEBAR_RENAME) =>
```

Can we use the enum directly for the type?

src/components/sidebar/SafeListContextMenu/MultiAccountContextMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+}: {
+  name: string
+  address: string
+  chainIds: string[]
+}): ReactElement => {
+  const [anchorEl, setAnchorEl] = useState<HTMLElement | undefined>()
+  const [open, setOpen] = useState<typeof defaultOpen>(defaultOpen)
+
+  const trackingLabel =
+    router.pathname === AppRoutes.welcome.accounts ? OVERVIEW_LABELS.login_page : OVERVIEW_LABELS.sidebar
```
usame-algan, line 39: This is only needed in `handleOpenModal` so I suggest moving it there.

src/store/addressBookSlice.ts
```diff
@@ -33,6 +33,20 @@ export const addressBookSlice = createSlice({
...
+      const { chainIds, address, name } = action.payload
+      if (name.trim() === '') {
+        return
+      }
+      chainIds.forEach((chainId) => {
+        if (!state[chainId]) state[chainId] = {}
+        state[chainId][address] = name
+      })
+    },
+
```
usame-algan, line 49: This is mostly identical with `upsertAddressBookEntry`. I tried replacing it with the following but for some reason the unit test for the slice fails.

```
const { chainIds, address, name } = action.payload

chainIds.forEach((chainId) => {
  upsertAddressBookEntry({ chainId, address, name })
})
```

usame-algan: Although it might be an anti-pattern to dispatch another action from a reducer.

src/components/sidebar/SafeListContextMenu/MultiAccountContextMenu.tsx
clovisdasilvaneto: Hmmm got it 👍

usame-algan (approved):

src/components/welcome/MyAccounts/styles.module.css
clovisdasilvaneto: @schmanu I mean, in the same selector, line 49, you overwrite it with `border: 1px solid var(--color-border-light)` 🤔

clovisdasilvaneto (approved):
✅

src/components/welcome/MyAccounts/styles.module.css
schmanu: Ah sorry I misunderstood. I thought you meant the entire block not just the line -.- 
You are right. I will remove it :)

PR #4144, opened 2024-09-09
Tests: Fix failing regression tests

mike10ca:
## How this PR fixes it

- Fix failing regression tests
- Skip a swap test due to a bug - fix WIP

## How to test it

- Run Cypress tests and observe outcome

PR #4145, opened 2024-09-09
Fix: (Swaps) sell token reactivity

katspaugh:
## What it solves

Resolves #4144

## How this PR fixes it

The sell token query parameter wasn't properly pass to the Swaps widget. It will now update the widget params when the URL query becomes available.

### DIFF
--- a/src/features/swap/index.tsx
+++ b/src/features/swap/index.tsx
@@ -274,6 +274,14 @@ const SwapWidget = ({ sell }: Params) => {
     }))
   }, [palette, darkMode, chainId])
 
+  useEffect(() => {
+    if (!sell) return
+    setParams((params) => ({
+      ...params,
+      sell,
+    }))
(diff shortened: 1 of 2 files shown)


src/features/swap/index.tsx
```diff
@@ -274,6 +274,14 @@ const SwapWidget = ({ sell }: Params) => {
...
   }, [palette, darkMode, chainId])
 
+  useEffect(() => {
+    if (!sell) return
+    setParams((params) => ({
+      ...params,
+      sell,
+    }))
+  }, [sell])
+
```
compojoom, line 284: If the user is navigating from assets/dashboard to the swap page then this doesn't seem necessary. 
If however the user has copied the swap link with the asset it in it and now tries to load this page in a new tab then the asset is not properly set.

compojoom (approved):
👍

PR #4148, opened 2024-09-10
Fix: hide SAP and Swaps if geoblocked

katspaugh:
## What it solves

When Swaps and Safe Activity Program are unavailable in a sanctioned country, hide the corresponding widgets.

## How to test

Install Urban VPN for mac and simulate Russia.


schmanu (approved):

francovenica:
I tried with the Urban VPN, setting it to Russian and I still can se the Swap feature.

I see also a lot of errorin the console that keep popping up constantly every second

image
image

katspaugh:
Ah, I forgot this would only work on prod. On dev/staging there's no blocking.

francovenica:
Ok, let's wait for prod then

PR #4149, opened 2024-09-10
fix: avoid blinking screen in the confirmation modal view

### LINKED ISSUE #4119: Tx creation: Confirm screen blinks during data loading  loading ( from interact with to actions list)
## Environment
 - Browser: Chrome
 - Wallet: any
 - Chain: any

## Steps to reproduce
 1. Go to the drain app ( any tx with many actions - starting from 2)
 2. select a few tokens
 3. enter recipient 
 4. Click Create tx
 Current result: The confirm screen loads interact with the first before actions list, as result, there is a blinking of the data on the confirm screen 

## Screenshots

https://github.com/user-attachments/assets/92596dd7-21b7-44f9-b09b-e7144962d0dd

clovisdasilvaneto:
## What it solves

Resolves #4119

## How this PR fixes it
This problem was happening because the call to get the `decodedData` is an asynchronous call which uses the `useAsync` hook. This `useAsync` hook was clearing the data everytime the component re-renders, what was making the component who presents the decoded transaction details (`<DecodedTx />`) to be unmounted. Diving in the three of this component and its dependencies, I found we have a parameter in the `useAsync` hook, which was added specially to avoiding cleaning the data after the consumer is re-rendered, and the default behaviour of it is to clear the data. 

The solution was to simply provide this parameter in the execution of `useAsync` hook inside the `useDecodeTx` hook, and also just show the `<DecodedTx />` component once the `decodedData` exists.

## How to test it
1) Go to the drain app ( any tx with many actions - starting from 2)
2) select a few tokens
3) enter recipient
4) Click Create tx

## Screenshots
Bellow is a video of the solution working:
[screen-recorder-tue-sep-10-2024-11-56-27.webm](https://github.com/user-attachments/assets/46b291db-6ec6-4281-8010-1182050d9b68)

## Checklist

### DIFF
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -94,6 +94,7 @@ export const SignOrExecuteForm = ({
   const isNewExecutableTx = useImmediatelyExecutable() && isCreation
   const isCorrectNonce = useValidateNonce(safeTx)
   const [decodedData] = useDecodeTx(safeTx)
+
   const isBatchable = props.isBatchable !== false && safeTx && !isDelegateCall(safeTx)
   const isSwapOrder = isConfirmationViewOrder(decodedData)
   const { data: txDetails } = useGetTransactionDetailsQuery(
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
Nice! 😉

katspaugh:
Tested, no more blinking. ✅

PR #4151, opened 2024-09-11
Feat: show warning when changing signer setup in a multichain safe [SW-150]

jmealy:
## What it solves
Resolves https://www.notion.so/safe-global/Multichain-Safes-V1-change-ownership-91f4347bbf764d35b7fdfa54179d7972?pvs=4

## How this PR fixes it
On the tx submit page, check if the tx is changing the signer setup and if the safe is multichain. If so, display a warning.

## How to test it
- for a multichain safe, create a new tx to add or remove a signer on one of the chains.
- see the warning on the sign or execute form.

## Screenshots
image

## Checklist


katspaugh (approved):
👍

PR #4156, opened 2024-09-11
Fix: continue with wallet button on welcome page sometimes not working

jmealy:
## What it solves
Resolves: #4247
@francovenica flagged it here: https://github.com/safe-global/safe-wallet-web/pull/4027#issuecomment-2333405462

## How this PR fixes it
Explicitly redirect when the continue button is pressed and adjust the redirect condition, so that when the safes have not yet loaded, you will still be taken to the accounts page.

## How to test it
- simulate a slow CGW response. I did this by changing the useHasSafes to always return `{ isLoaded: false, hasSafes: false }`
- Connect your wallet and visit the welcome page. Click the `Continue with..` button. You should be redirected to the accounts page.

## Checklist


src/components/welcome/WelcomeLogin/index.tsx
```diff
@@ -17,22 +17,29 @@ const WelcomeLogin = () => {
...
+    }
+  }, [hasSafes, isLoaded, router, wallet])
+
   const onLogin = useCallback(() => {
     setShouldRedirect(true)
   }, [])
 
+  const onContinue = useCallback(() => {
+    redirect()
+  }, [redirect])
```
katspaugh, line 37:
```suggestion
  const onContinue = useCallback(() => {
    redirect()
  }, [redirect])
```

Essentially it's the same as `const onContinue = redeirect`. So you could just do `onContinue={redirect}` in the JSX below.

katspaugh (approved):

francovenica:
LGTM.

Tried several times and I didn't had issues reaching the /accounts URL.

PR #4069, opened 2024-08-14
feat: automatically migrate incompatible Safes from L1 to L2 if possible

schmanu:
## ATTENTION THIS IS A DRAFT PR AND NOT MEANT TO BE MERGED YET
This PR hardcodes some addresses and is using an un-audited contract.
DO NOT MERGE.

## What it solves
Makes replayed Mainnet Safes compatible on L2s.

## How this PR fixes it
- If a Safe is using an incompatible MasterCopy on a L2 chain, a migration delegateCall is prepended to the first transaction the Safe creates
- This only works for Safes which are still at nonce 0

## How to test it
- Replay a L1 Safe on Polygon or Optimism
- Create any transaction in that new Safe and observe a migration tx being added to it.

## Checklist
* [x] I've tested the branch on mobile 📱
* [ ] I've documented how it affects the analytics (if at all) 📊
* [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #4158, opened 2024-09-11
fix: Cleanup pending txs if nonce already executed

usame-algan:
## What it solves

Follow up to #4025

This is needed in order for #4069 to work since migration transactions are always replaced and their pending state is stuck indefinitely which bricks the safe.

## How this PR fixes it

- Stores the `nonce` for each pending transaction
- Extends the pending tx cleanup logic in the tx history middleware by not just checking if there is a matchin txId but also if there is a matching nonce and in case there is and txIds don't match it means the pending transaction got replaced so it is safe to clear it

## ToDo

- [x] Extend tests for `txHistorySlice`
- [x] (optional) Simplify test mocks by using a builder

## How to test it

1. Open the same safe in two different browsers
2. Execute a transaction with the same nonce at the same time from each browser
3. Observe both pending states to be cleared as soon as one of the transactions is executed

## Checklist

### DIFF
--- a/src/store/__tests__/txHistorySlice.test.ts
+++ b/src/store/__tests__/txHistorySlice.test.ts
@@ -84,13 +84,6 @@ describe('txHistorySlice', () => {
         dispatch: jest.fn(),
       }
 
-      const transaction = {
-        type: TransactionListItemType.TRANSACTION,
-        transaction: {
-          id: '0x123',
-        },
-      } as TransactionListItem
-
       const action = txHistorySlice.actions.set({
         loading: false,
         data: undefined, // Cleared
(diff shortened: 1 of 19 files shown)


src/store/__tests__/txHistorySlice.test.ts
```diff
@@ -25,6 +25,7 @@ describe('txHistorySlice', () => {
       const state = {
         pendingTxs: {
```
schmanu, line 26: Lets add a test for the new pending state clearing here.

schmanu (approved):
The fix looks good 💯 
Should improve a lot of stuck Safes with pending txs.

francovenica:
LGTM.

Opened the same safe in 2 browsers, create a tx that is ready for execution and in each browser I connected with a different owner.
I executed the tx at the same time in both browsers. I see that as soon as one tx is accepted and is being indexed, the other browser shows the tx not in pending anymore. The current behavior is, if the tx takes a bit to index, the tx in the other browser would look as ready to be executed again, showing the error message that the tx was reverted, but as soon the indexing ends, it shows the success status.

PR #4161, opened 2024-09-12
[Multichain] Feat: disable adding new chains to safes created from old mastercopy versions [SW-167]

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Do-not-offer-adding-networks-for-1-1-1-created-Safes-9608b9ee16a549e891b795fe18b5e831?pvs=4

## How this PR fixes it
- Gets the mastercopy/factory address used to create a given safe, and checks if there is a corresponding deployment within the list of supported versions.
- If there are no matching deployments, display a message informing the user that this safe cannot be replayed.

## How to test it
- Open a safe that was created as a 1.1.1 safe. 
- Click the network selector in the top right corner. It should show a message saying that the safe cannot be replayed.
- Click on the context menu of the safe in the sidebar or the accounts page. Click the button to add another network. It should show a message saying that the safe cannot be replayed.

## Screenshots
image

image

## Checklist


src/features/multichain/hooks/__tests__/useCompatibleNetworks.test.ts
```diff
@@ -184,55 +183,7 @@ describe('useReplayableNetworks', () => {
         saltNonce: '0',
         setupData,
       }
-      const { result } = renderHook(() => useReplayableNetworks(creationData, []))
-      expect(result.current).toHaveLength(4)
-      expect(result.current.map((chain) => chain.chainId)).toEqual(['1', '10', '100', '480'])
-    }
-  })
-
-  it('should remove already deployed chains from result', () => {
```
jmealy, line 193: I moved this filtering to outside the hook because I needed to distinguish between
- there being no more chains left to deploy to (already deployed on all chains)
- there being no chains at all (unsupported safe creation version)

src/features/multichain/components/CreateSafeOnNewChain/index.tsx
```diff
@@ -102,17 +106,28 @@ const ReplaySafeDialog = ({
...
             <ErrorMessage error={safeCreationDataError} level="error">
               Could not determine the Safe creation parameters.
             </ErrorMessage>
+          ) : safeCreationDataLoading ? (
+            <Stack direction="column" alignItems="center" gap={1}>
+              <CircularProgress />
+              <Typography variant="body2">Loading Safe data</Typography>
+            </Stack>
+          ) : isUnsupportedSafeCreationVersion ? (
+            <Typography>
```
schmanu, line 128: Lets wrap this in an `ErrorMessage`.

src/features/multichain/components/CreateSafeOnNewChain/index.tsx
```diff
@@ -175,18 +198,21 @@ export const CreateSafeOnNewChain = ({
   )
 
   const safeCreationResult = useSafeCreationData(safeAddress, deployedChains)
-  const replayableChains = useReplayableNetworks(safeCreationResult[0], deployedChainIds)
+  const allCompatibleChains = useCompatibleNetworks(safeCreationResult[0])
+  const isUnsupportedSafeCreationVersion = Boolean(!allCompatibleChains?.length)
+  const replayableChains = allCompatibleChains?.filter((config) => !deployedChainIds.includes(config.chainId)) || []
```
schmanu, line 203: I think we need to memoize this with `useMemo`

src/features/multichain/components/CreateSafeOnNewChain/index.tsx
```diff
@@ -175,18 +198,21 @@ export const CreateSafeOnNewChain = ({
   )
 
   const safeCreationResult = useSafeCreationData(safeAddress, deployedChains)
-  const replayableChains = useReplayableNetworks(safeCreationResult[0], deployedChainIds)
+  const allCompatibleChains = useCompatibleNetworks(safeCreationResult[0])
+  const isUnsupportedSafeCreationVersion = Boolean(!allCompatibleChains?.length)
```
schmanu, line 202: This will also be true for zkSync Safes as they have no compatible networks.

jmealy: Also, I created a ticket for when there are no available chains to deploy to: https://www.notion.so/safe-global/76f265123c0a4ad680947ce489804342?v=f9f164524ec4464286fdc38f31589d0c&p=4ec720b334ea410bbe8abb083509bf7b&pm=s

Right now it just shows the network selector with no options

schmanu: Okay we can keep it as is then.

liliya-soroka:
@jmealy , could you re-check "add network" in side menu
I am getting red error instead of text from the ticket ( 1.1.1 safe on Gnosis - gno:0x0A2f9630369120dbb021516889c96cdf80F94844 )
image

liliya-soroka:
3. In the PR , I am also getting the error for 1.3.0L2 Gnosis safe (gno:0x62AB1CEF969F9459E8fFbb5E0E8289f4101ace1c) - 

https://github.com/user-attachments/assets/538f1a57-8e32-4039-8d90-77cd49299fae

schmanu:
@liliya-soroka All errors are expected.

1. The Gnosis RPC can be super flaky. Often it returns that the tx hash does not exist. Thats why the error says the transaction could not be found.

2. Same here.

3. The Safe you are trying to replay here was not created with create2 (`createProxyWithNonce`) it was created by calling the `createProxy` method. So the error correctly states that the Safe was created with an unsupported method.

schmanu (approved):

PR #4163, opened 2024-09-12
[Multichain] fix: Hide header network selector on non-safe routes [SW-171]

usame-algan:
## What it solves

Resolves [SW-171](https://www.notion.so/safe-global/Hide-network-selector-on-non-safe-routes-a3071123877f43b5a56ae630e073298d)

## How this PR fixes it

- Hides the network selector in the header if there is no safe address present

## How to test it

1. Open the welcome page, safe creation, imprint and other non-safe specific pages
2. Observe there is no network selector in the header
3. Open a Safe route
4. Observe there is a network selector in the header

## Screenshots
Screenshot 2024-09-12 at 12 07 09
Screenshot 2024-09-12 at 12 07 22

## Checklist


jmealy (approved):

src/components/common/NetworkSelector/index.tsx
```diff
@@ -179,46 +181,6 @@ const UndeployedNetworks = ({
...
 }
 
-const DeployedNetworkMenuItem = ({
-  chainId,
-  chainConfigs,
-  isSelected = false,
-  onClick,
-  safeOverviews,
-  getNetworkLink,
-}: {
```
usame-algan, line 189: @clovisdasilvaneto I had to move this component back to be a function call because the network selector wasn't closing anymore. From what I could find it is because MUI Select expects the first child to be a `MenuItem` so extracting it into a custom component broke that. I tried adding a `forwardRef` without success. Do you know of anything we could still try?

PR #4167, opened 2024-09-12
Feat: staking claim [SW-160]

compojoom:
## What it solves
Ads decoding views for stake claim txs

Resolves #

## Checklist

### DIFF
--- a/src/features/stake/components/StakingWidget/index.tsx
+++ b/src/features/stake/components/StakingWidget/index.tsx
@@ -4,7 +4,7 @@ import AppFrame from '@/components/safe-apps/AppFrame'
 import { getEmptySafeApp } from '@/components/safe-apps/utils'
 
 const widgetAppData = {
-  url: 'https://safe.widget.testnet.kiln.fi/earn',
+  url: 'https://safe.widget.kiln.fi/earn',
   name: 'Stake',
   iconUrl: '/images/common/stake.svg',
   chainIds: ['17000', '11155111', '1', '42161', '137', '56', '8453', '10'],
--- a/src/features/stake/components/StakingWidget/index.tsx
+++ b/src/features/stake/components/StakingWidget/index.tsx
@@ -2,9 +2,13 @@ import { useMemo } from 'react'
 import { useDarkMode } from '@/hooks/useDarkMode'
 import AppFrame from '@/components/safe-apps/AppFrame'
 import { getEmptySafeApp } from '@/components/safe-apps/utils'
+import useChainId from '@/hooks/useChainId'
+import useChains from '@/hooks/useChains'
 
+const WIDGET_PRODUCTION_URL = 'https://safe.widget.kiln.fi/earn'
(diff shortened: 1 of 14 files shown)


src/features/stake/helpers/utils.ts
```diff
@@ -2,10 +2,12 @@ import { id } from 'ethers'
 import type { BaseTransaction } from '@safe-global/safe-apps-sdk'
 
 const WITHDRAW_SIGHASH = id('requestValidatorsExit(bytes)').slice(0, 10)
+const CLAIM_SIGHASH = id('batchWithdrawCLFee(bytes)').slice(0, 10)
 
 export const getStakeTitle = (txs: BaseTransaction[] | undefined) => {
   const hashToLabel = {
     [WITHDRAW_SIGHASH]: 'Withdraw request',
+    [CLAIM_SIGHASH]: 'Claim',
```
katspaugh, line 10: Can’t these titles be deduced from the txInfo/txConfirmation type?

compojoom: As discussed in the call - we currently do this because the app title is a component that is rendered before the modal content and as such it doesn't have access to the txInfo. This would need a refactoring (since you guys are working on the txflow #4157 you can consider it there)

katspaugh (approved):

katspaugh:
Tested, it works correctly ✅ 

Tx comfirmation:
Screenshot 2024-09-13 at 14 27 08

Tx details in the queue:
Screenshot 2024-09-13 at 14 27 38

I'm not convinced it's a good idea to say "Receive" in one place but "Amount" in another for the exact same thing (rewards mount). @TanyaEfremova shall we unify the phrasing maybe?

katspaugh:
@compojoom also I would make the gap between these two blocks a tiny bit bigger:
Screenshot 2024-09-13 at 14 29 30

TanyaEfremova:
@katspaugh You're right, it's not intended to say 'Amount'. Both screens (transaction summary and transaction confirmation) are [now aligned in Figma](https://www.figma.com/design/ih2mM8bt8rNwbwV9EMmWF5/Staking?node-id=3265-33900&t=4UZz068V9AfDPJlw-1).

src/features/stake/components/StakingWidget/index.tsx
```diff
@@ -12,12 +14,21 @@ const widgetAppData = {
...
 const StakingWidget = () => {
   const isDarkMode = useDarkMode()
+  let url = widgetAppData.url
+  const currentChainId = useChainId()
+  const { configs } = useChains()
+  const testChains = useMemo(() => configs.filter((chain) => chain.isTestnet), [configs])
+
+  // if currentChainId is in testChains, then set the url to the testnet version
+  if (testChains.some((chain) => chain.chainId === currentChainId)) {
+    url = 'https://safe.widget.testnet.kiln.fi/earn'
```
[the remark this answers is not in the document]
compojoom: good point

katspaugh (approved):

PR #4168, opened 2024-09-13
Tests: Fix regression cy 2

mike10ca:
## How this PR fixes it

- Use ls to accept user consent
- Adjust history tests
- Remove propose request in create tx tests

## How to test it


cypress/e2e/regression/swaps_history_2.cy.js
liliya-soroka, line 1: will you add check for the contract in the advanced details ?

mike10ca: This will be done in scope of another task, I added it to our board.

liliya-soroka (approved):

PR #4178, opened 2024-09-16
Legal: update the terms page [SW-185]

katspaugh:
## What it solves

A few minor edits on the Terms page.

### DIFF
--- a/src/components/common/CookieAndTermBanner/index.tsx
+++ b/src/components/common/CookieAndTermBanner/index.tsx
@@ -75,9 +75,10 @@ export const CookieAndTermBanner = ({
           <Grid item xs>
             <Typography variant="body2" mb={2}>
               By browsing this page, you accept our{' '}
-              <ExternalLink href={AppRoutes.terms}>Terms & Conditions</ExternalLink> (last updated August 2024) and the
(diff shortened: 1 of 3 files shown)


src/components/common/CookieAndTermBanner/index.tsx
```diff
@@ -75,9 +75,10 @@ export const CookieAndTermBanner = ({
           <Grid item xs>
             <Typography variant="body2" mb={2}>
               By browsing this page, you accept our{' '}
-              <ExternalLink href={AppRoutes.terms}>Terms & Conditions</ExternalLink> (last updated August 2024) and the
-              use of necessary cookies. By clicking &quot;Accept all&quot; you additionally agree to the use of Beamer
-              and Analytics cookies as listed below. <ExternalLink href={AppRoutes.cookie}>Cookie policy</ExternalLink>
+              <ExternalLink href={AppRoutes.terms}>Terms & Conditions</ExternalLink> (last updated September 2024) and
```
katspaugh, line 78: We should include last updated month into the terms slice, otherwise it has to be hardcoded here and on the terms page.

compojoom (approved):

compojoom:
How about we close this in favour of #4183 ?

PR #4181, opened 2024-09-16
Fix: back button breakpoint in tx flow

clovisdasilvaneto:
## What it solves

Resolves 
release blocker: https://github.com/safe-global/safe-wallet-web/pull/4162#issuecomment-2349136056

## How this PR fixes it
It decreases the breakpoint of the backButton css and removes a magic number from the txLayout component. Now the `isDesktop` variable is getting it value based on the `lg` measure.

## How to test it
1) Go to any transaction screen where the back button appears
2) Resize the screen 
3) The back button should break only in the 1199 resolution

## Screenshots

Video: https://www.awesomescreenshot.com/video/31567985?key=75c8d42a80cec6163ca9679d129950cf

## Checklist

### DIFF
--- a/src/components/tx-flow/common/TxLayout/index.tsx
+++ b/src/components/tx-flow/common/TxLayout/index.tsx
@@ -83,7 +83,7 @@ const TxLayout = ({
 
   const theme = useTheme()
   const isSmallScreen = useMediaQuery(theme.breakpoints.down('md'))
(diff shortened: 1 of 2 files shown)


katspaugh (approved):

katspaugh:
Tested, the Back button now wraps at the same breakpoint as when the buttons become full-width. ✅

PR #4182, opened 2024-09-16
fix: do not check network for confirm / execute buttons in the queue

schmanu:
## What it solves
If an owner is connected to the wrong chain the confirm and execute button is disabled.

## How this PR fixes it
Removes network check  for those buttons.

## How to test it
- Open a Safe with a queued and signable tx
- Connect with an owner wallet on a wrong chain

## Checklist


katspaugh (approved):
✌️

PR #4183, opened 2024-09-16
Refactor: generate terms page from md file

compojoom:
## What it solves
It's very cumbersome to update the Terms page as it is in html and we have to find the changed areas and make sure to correctly update them. Instead with this change we can always export the complete terms document to md and then leave it to remark to render it as html

Resolves #

## Checklist


src/markdown/terms/terms.md
```diff
@@ -0,0 +1,384 @@
+# Terms and Conditions
+
+Last updated: SeptemberAugust, 2024
```
katspaugh, line 3: > SeptemberAugust

Looks like you exported this file in Suggest mode. We should accept all those suggestions and export the final version.

katspaugh (commented):
Looks great but we should move some (all?) of these new dependencies in devDependencies. Nonce of them seem to be imported in any runtime module.

katspaugh (approved):
🔥

katspaugh (approved):

PR #4187, opened 2024-09-17
fix: add network switch warning to recovery execution

jmealy:
## What it solves

Resolves: https://github.com/safe-global/safe-wallet-web/pull/4162#issuecomment-2353248074

## How this PR fixes it
- Since the recovery tx is executed directly from the queue, we need a network warning to be shown at this point.
- Adds a network switch warning above the execute button.

## How to test it
- Create a recovery for another wallet.
- Open the safe with the recoverer wallet, wait for the recovery delay to elapse, then view the recovery transaction in the queue.
- If you're on the wrong network it should show a network switch warning, and the tooltip for the disabled execute button should show the relevant message.

## Screenshots
image

## Checklist


katspaugh (approved):

usame-algan:
I wasn't aware that we don't have a review screen for when the recovery is executed. Imo something worth adding.

I tried testing this but the app crashes for me. Could you try to reproduce? Here are the steps I took:

1. Create a recovery attempt with wallet B
2. Switch to wallet A who is an owner and go to queue
3. Switch wallet A network to something other than the safe

After a while before the recovery attempt appears in the queue the app just freezes for me. The Tab takes up 2.2gb so I assume some kind of memory leak is going on 🤔

katspaugh (commented):
Why does the network switch button have to be right in the tx details? All other tx flows have it on the review screen.

jmealy:
> I wasn't aware that we don't have a review screen for when the recovery is executed. Imo something worth adding.
> 
> I tried testing this but the app crashes for me. Could you try to reproduce? Here are the steps I took:
> 
> 1. Create a recovery attempt with wallet B
> 2. Switch to wallet A who is an owner and go to queue
> 3. Switch wallet A network to something other than the safe
> 
> After a while before the recovery attempt appears in the queue the app just freezes for me. The Tab takes up 2.2gb so I assume some kind of memory leak is going on 🤔

Happening for me also, looking into it

katspaugh:
Tested, the execute button is now working. The crash is also in production, so I'm going to merge this PR.

PR #4198, opened 2024-09-18
Fix(Multichain): Show owner setup warning also when replacing owners and changing threshold [SW-150]

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Multichain-Safes-V1-change-ownership-91f4347bbf764d35b7fdfa54179d7972?pvs=4

## How this PR fixes it
- Shows the warning message whenever signer setup is changed.
- Also changes the blockaid warning to have the same icon as other warnings on the page.

## How to test it
- create a multichain safe
- change the signer setup of one of the chains.
- On the tx confirmation screen for adding, removing, replacing, or changing threshold, you should see the warning.

## Screenshots

Making icons consistant (before):
image

After:
image

## Checklist


schmanu (approved):
Will approve as it fixes the described issue.
But once the refactoring of the Settings Change review screens is merged we should move this warning up to the block that describes the settings change. 
I don't think this error makes much sense in the Execute / Sign block.

jmealy:
> Will approve as it fixes the described issue. But once the refactoring of the Settings Change review screens is merged we should move this warning up to the block that describes the settings change. I don't think this error makes much sense in the Execute / Sign block.

True probably makes more sense there!

TanyaEfremova (commented):
@schmanu had a valid point, see [Figma design](https://www.figma.com/design/aUYQwSp7bKJDE7hBUxREzQ/Multichain-accounts?node-id=1455-22443&t=cxUZRW4npAUObx1X-1)

It looks good, but let's move it to the top under the card "new signer" and change alert style to info.

src/features/multichain/helpers/utils.ts
```diff
@@ -1,5 +1,8 @@
 import type { DecodedDataResponse } from '@safe-global/safe-gateway-typescript-sdk'
 
 export const isChangingSignerSetup = (decodedData: DecodedDataResponse | undefined) => {
-  return decodedData?.method === 'addOwnerWithThreshold' || decodedData?.method === 'removeOwner'
+  return (
+    decodedData?.method &&
+    ['swapOwner', 'addOwnerWithThreshold', 'removeOwner', 'changeThreshold'].includes(decodedData?.method)
```
clovisdasilvaneto, line 6: it would be nice if could we move the declaration of this array to be outside the function body, just to avoid having this array re-created on every function call.

jmealy:
> @schmanu had a valid point, see [Figma design](https://www.figma.com/design/aUYQwSp7bKJDE7hBUxREzQ/Multichain-accounts?node-id=1455-22443&t=cxUZRW4npAUObx1X-1)
> 
> It looks good, but let's move it to the top under the card "new signer" and change alert style to info.

Made the suggested changes:

image
image

src/features/multichain/helpers/utils.ts
jmealy: This hook isn't needed anymore after the other requested changes

PR #4199, opened 2024-09-18
Fix(Multichain): No network is visible in activate account flow. [SW-192]

jmealy:
## What it solves
When activating a CF safe, the network is not shown as it was before.

## How this PR fixes it
Show the network in `ActivateAccountFlow` when activating a safe

## How to test it
- Create a CF safe and then activate it.
- On the confirmation screen, the network of the safe should be shown in the `Network` section

## Screenshots
image

## Checklist


usame-algan (approved):
Thanks!

PR #4200, opened 2024-09-18
Refactor: a review screen for recovery attempts

### LINKED ISSUE #4188: Tx confirmation screen for Recovery execution
[It turned out](https://github.com/safe-global/safe-wallet-web/pull/4187#issuecomment-2354896608) that Recovery execution doesn't have a tx confirmation screen. It gets executed directly from the queue.

This is not consistent with all other txs and leads to code duplication and is error-prone. So we should add a confirmation view for this case.

katspaugh:
## What it solves

Resolves #4188

Screenshot 2024-09-20 at 11 03 03

## How this PR fixes it

Adds a simple review screen similar to the Spening Limit review component.

## How to test it
* Set up recovery
* Initiate a recovery tx
* Confirm/execute the recovery tx


src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
```diff
@@ -0,0 +1,93 @@
...
+import ErrorMessage from '@/components/tx/ErrorMessage'
+import TxCard from '@/components/tx-flow/common/TxCard'
+import FieldsGrid from '@/components/tx/FieldsGrid'
+import { TxModalContext } from '../..'
+
+export type RecoveryAttemptReviewProps = {
+  params: {
+    args: TransactionAddedEvent.Log['args']
+    address: string
+  }
```
clovisdasilvaneto, line 19: It would be nice if we could avoid having nested parameters in the props, so this component would just be re-rendered only when it is necessary. So instead of passing the address and args inside params, you could pass them as main props, like: 

```suggestion

export type RecoveryAttemptReviewProps = {
    args: TransactionAddedEvent.Log['args']
    address: string
}
```

src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+const RecoveryAttemptReview = ({ params }: RecoveryAttemptReviewProps) => {
+  const [isPending, setIsPending] = useState(false)
+  const [error, setError] = useState<Error>()
+  const wallet = useWallet()
+  const { safe } = useSafeInfo()
+  const { setTxFlow } = useContext(TxModalContext)
+
+  const onFormSubmit = useCallback(
```
clovisdasilvaneto, line 29: I think we do not need to use a `useCallback` here, because this function doesn't do expansive calculation and also it is only used as a DOM event.

src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
```diff
@@ -0,0 +1,93 @@
...
+          delayModifierAddress: params.address,
+          signerAddress: wallet.address,
+        })
+        setTxFlow(undefined)
+      } catch (err) {
+        trackError(Errors._812, err)
+        setError(err as Error)
+      }
+
+      setIsPending(false)
```
clovisdasilvaneto, line 52: It would be nice if we could move the logic of handling the `error` and `pending` states of the request to a separated hook, so we don't need to explicitly handle this cases every time we have to send a "dispatch request"

src/features/recovery/components/ExecuteRecoveryButton/index.tsx
```diff
@@ -22,37 +17,17 @@ export function ExecuteRecoveryButton({
...
-        delayModifierAddress: recovery.address,
-        signerAddress: wallet.address,
-      })
-    } catch (_err) {
-      const err = asError(_err)
-
-      trackError(Errors._812, e)
-      setSubmitError(err)
-    }
+    setTxFlow(<RecoveryAttemptFlow params={recovery} />)
```
clovisdasilvaneto, line 30: Its not part of this PR, but just to bring one idea if it wasn't noticed yet:

It would be nice if we could change the way how we set the current transaction flow component, like setting a enum on the state or any other primitive value instead of the component call would save us some undesired reconciliations and re-renders because react doesn't memoize non-primitive states.

src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
katspaugh: It's not about whether the function itself is heavy or not. It's not going to be called on render anyway, only created. We memoize callbacks because otherwise it will trigger a re-rendering of the JSX tree as the onSubmit prop would always receive a new reference. While it might be only a slight optimization in this case, it doesn't hurt.

src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
katspaugh: Could you give me an example of how you envision this? I would use the useAsync hook here but it has to happen on form submit, and I'm not sure what would trigger the hook.

src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
katspaugh: It now takes (an even more nested) RecoveryQueueItem as a prop.

src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
clovisdasilvaneto: similar to the way of how useQuery does, something like:

```jsx

function useAsyncDispatch<T, E>(dispatch: (props: T) => Promise<E>) {
  const [isLoading, toggleIsLoading] = useReducer((isLoading: boolean) => !isLoading, false)
  const [error, setError] = useState<Error>()

  const asyncDispatch = async (args: T) => {
    toggleIsLoading()
    setError(undefined)

    try {
      const data = await dispatch(args)

      toggleIsLoading()

      return data
    } catch (e) {
      setError(e as Error)
      toggleIsLoading()
    }
  }

  return { isLoading, error, asyncDispatch }
}
```
and in your component you could do:

```jsx
 const MyComponent = () => {
  const { asyncDispatch, isLoading, errror } = useAsyncDispatch<MyExpectedArgs, MyExpectedResult>();

 const onSubmit = async () => {
  const data = await asyncDispatch();

  // do somthing else
 }
 }
```

katspaugh: I see, thanks!

src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
```diff
@@ -18,9 +18,35 @@ type RecoveryAttemptReviewProps = {
   item: RecoveryQueueItem
 }
 
-const RecoveryAttemptReview = ({ item }: RecoveryAttemptReviewProps) => {
-  const [isPending, setIsPending] = useState(false)
+function useAsyncDispatch<T extends (...args: any) => Promise<any>>(dispatchFn: T) {
```
clovisdasilvaneto, line 21: Nice! Shall we move this hook to the globals hook folder? Or shall we wait until we have a similar use case?

src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
```diff
@@ -0,0 +1,112 @@
...
+  )
+
+  return (
+    <TxCard>
+      <form onSubmit={onFormSubmit}>
+        <Stack gap={3} mb={2}>
+          {item.address && (
+            <FieldsGrid title="Initiated by">
+              <EthHashInfo address={item.address} showAvatar hasExplorer showName />
+            </FieldsGrid>
```
usame-algan, line 78: I would expect to see the recoverer address here (the wallet that initiated this recovery attempt) but `item.address` is the contract address of the delay modifier.

src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
katspaugh: Moved to `useAsync.ts` and renamed to `useAsyncCallback`.

src/components/tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx
katspaugh: Oh, good catch.

usame-algan: We could even include the validity time since its part of `item` i.e. say `Confirm or reject this transaction until 20.11.2024` or if it never expires i.e. `item.expiresAt` is null we could say `Confirm or reject this transaction`

usame-algan:
I just noticed that it is possible to change the nonce of the transaction even though it is already queued. Can we make it a read-only field?
Screenshot 2024-09-20 at 10 33 13

katspaugh:
@usame-algan good catch, I've added hideNonce as it's a module transaction.

Screenshot 2024-09-20 at 11 03 03

usame-algan (approved):
Looks good! 👍

francovenica:
Minor issue.
The tx status still shows how many confirmations are needed for this tx to go through, even tho only the current owner has to sign and execute it.
image

For comparison, you can see how in spending limits, the amount of confirmations go away when the option is selected. This is the same case
transaction status+

katspaugh:
Fixed ✅ 

Screenshot 2024-09-26 at 10 33 20

francovenica:
Looks good!

PR #4203, opened 2024-09-19
Feat stake OFAC

compojoom:
## What it solves
Checks whether the currently connected wallet or safe is on the OFAC list and blocks access to the stake feature.

## How to test it
You can use impersonator.xyz with any of the addresses on the https://github.com/ultrasoundmoney/ofac-ethereum-addresses/blob/main/data.csv list and it should display a block letting you know that the address is blocked.

## Screenshots
grafik

## Checklist

### DIFF
--- a/src/store/ofac.ts
+++ b/src/store/ofac.ts
@@ -62,6 +62,7 @@ export const ofacApi = createApi({
           return { error: { status: 'CUSTOM_ERROR', data: (error as Error).message } }
         }
       },
+      keepUnusedDataFor: 24 * 60 * 60, // 24 hours
     }),
   }),
 })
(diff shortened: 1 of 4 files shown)


src/store/ofac.ts
```diff
@@ -62,6 +62,7 @@ export const ofacApi = createApi({
           return { error: { status: 'CUSTOM_ERROR', data: (error as Error).message } }
         }
       },
+      keepUnusedDataFor: 24 * 60 * 60, // 2
```
jmealy, line 65: Outdated comment (should be 1 day?)

jmealy (approved):

src/store/ofac.ts
compojoom: No, I somehow managed to delete the "4 hours" from "24 hours" :D

PR #4208, opened 2024-09-19
[Multichain] Fix: multichain disable cf creation

schmanu:
## What this PR changes
- new `allowUndeployedSafe` prop in `CheckWallet`
- If false the buttons get disabled for undeployed Safes
- Unit tests

## How to test it
- Create a new CF multichain Safe
- Go to e.g. Settings and observe the "Add owner" button to be disabled

## Screenshots
Screenshot 2024-09-19 at 12 45 29

## Checklist

### DIFF
--- a/src/components/common/CheckWallet/index.test.tsx
+++ b/src/components/common/CheckWallet/index.test.tsx
@@ -5,6 +5,10 @@ import useIsSafeOwner from '@/hooks/useIsSafeOwner'
 import useIsWrongChain from '@/hooks/useIsWrongChain'
 import useWallet from '@/hooks/wallets/useWallet'
 import { chainBuilder } from '@/tests/builders/chains'
+import { useIsWalletDelegate } from '@/hooks/useDelegates'
+import { faker } from '@faker-js/faker'
+import { extendedSafeInfoBuilder } from '@/tests/builders/safe'
+import useSafeInfo from '@/hooks/useSafeInfo'
 
 // mock useWallet
 jest.mock('@/hooks/wallets/useWallet', () => ({
@@ -36,6 +40,25 @@ jest.mock('@/hooks/useChains', () => ({
 jest.mock('@/hooks/useIsWrongChain', () => ({
   __esModule: true,
   default: jest.fn(() => false),
+}))
+
+jest.mock('@/hooks/useDelegates', () => ({
+  __esModule: true,
+  useIsWalletDelegate: jest.fn(() => false),
+}))
+
+jest.mock('@/hooks/useSafeInfo', () => ({
+  __esModule: true,
+  default: jest.fn(() => {
+    const safeAddress = faker.finance.ethereumAddress()
+    return {
+      safeAddress,
+      safe: extendedSafeInfoBuilder()
+        .with({ address: { value: safeAddress } })
+        .with({ deployed: true })
+        .build(),
+    }
+  }),
 }))
 
 const renderButton = () =>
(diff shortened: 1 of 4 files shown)


src/components/common/CheckWallet/index.tsx
```diff
@@ -28,28 +29,39 @@ const CheckWallet = ({
...
+  const message = useMemo(() => {
+    if (!wallet) {
+      return Message.WalletNotConnected
+    }
+    if ((!allowNonOwner && !isSafeOwner && !isDelegate) || (isOnlySpendingLimit && !allowSpendingLimit)) {
+      return Message.NotSafeOwner
+    }
+    if (isUndeployedSafe && !allowUndeployedSafe) {
+      return Message.SafeNotActivated
+    }
```
usame-algan, line 54: I would reorder these two conditions to preserve the same order as previously since the check for NotSafeOwner is more specific than for SafeNotActivated. 

Do we need the empty string return still or does it also work with undefined?

src/features/counterfactual/ActivateAccountButton.tsx
```diff
@@ -25,7 +25,7 @@ const ActivateAccountButton = () => {
   return (
     <Tooltip title={isProcessing ? 'The safe activation is already in process' : undefined}>
       <span>
-        <CheckWallet allowNonOwner>
+        <CheckWallet allowNonOwner allowUndeployedSafe>
```
usame-algan, line 28: Do we need CheckWallet here at all? Afaiu the only thing we need to make sure is that the user has any wallet connected. If we could ensure this without CheckWallet we wouldn't need another flag.

schmanu: But isn't that exactly what CheckWallet is for? So I don't need to use the `useWallet` hook here and worry about rendering a tooltip around the button etc?

src/components/common/CheckWallet/index.test.tsx
```diff
@@ -110,6 +133,63 @@ describe('CheckWallet', () => {
...
+    }
+
+    ;(useSafeInfo as jest.MockedFunction<typeof useSafeInfo>).mockReturnValueOnce(
+      mockSafeInfo as unknown as ReturnType<typeof useSafeInfo>,
+    )
+
+    const { container } = renderButton()
+
+    expect(container.querySelector('button')).toBeDisabled()
+    expect(container.querySelector('span[aria-label]')).toHaveAttribute(
```
clovisdasilvaneto, line 164: you can use getByLabelText instead of querySelector. It's more performant than using query selectors: https://testing-library.com/docs/queries/bylabeltext/

src/features/counterfactual/ActivateAccountButton.tsx
usame-algan: Yes, makes sense. I created a ticket #4210 to refactor this component.

src/components/common/CheckWallet/index.test.tsx
compojoom: `getByLabelText` is nice, but in my opinion it makes tests brittle as a change to the text would cause the test to fail. Unless one explicitly cares about the button's text I wouldn't go for it. I personally prefer to have `data-testid` makes life way easier.

usame-algan: I disagree. If possible we should select elements by their content as it reflects what the end user sees. If there is no content I would also fall back to using test ids. Imo its more brittle if we select by something like the HTML element. Changing the span to a p would make the test fail even though the element might still look the same to the user.

schmanu: This case especially is interested in seeing specific error messages in specific cases.
If we want to avoid it breaking if we just rephrase an error we could use constants for each individual error and use the same constants in the test.

src/components/common/CheckWallet/index.tsx
schmanu: We do not need an empty string, it also works with undefined:
`  if (!message) return children(true)
`
I'll adjust the order 👍

src/components/common/CheckWallet/index.test.tsx
schmanu: The only downside of `getByLabelText` vs `querySelect` is that the assertion error in case of a wrong expectation is worse:

With query selector:
```
Expected the element to have attribute:
  aria-label="Wrong expectation"
Received:
  aria-label="You need to activate the Safe before transacting"
```

With getByLabelText:
```
TestingLibraryElementError: Unable to find a label with the text of: Wrong expectation

Ignored nodes: comments, script, style
<div>
  <span
    aria-label="You need to activate the Safe before transacting"
    class=""
    data-mui-internal-clone-element="true"
  >
    <button
      disabled=""
    >
      Continue
    </button>
  </span>
</div>
```

If the container is a big html element it becomes really hard to see why the field wasn't found.

PR #4211, opened 2024-09-19
[Multichain] fix: Disable submit button while replaying safe [SW-173]

usame-algan:
## What it solves

Resolves [SW-173](https://www.notion.so/safe-global/Close-networks-selector-in-the-header-after-choosing-network-8300245b7058465ea6de4a9e9171f094)

## How this PR fixes it

- Adds a `isSubmitDisabled` state

## How to test it

1. Open a replayable safe
2. Add another network
3. Observe the submit button becoming disabled when clicked
4. Observe the submit button becoming enabled when finished

## Checklist

### DIFF
--- a/src/features/multichain/components/CreateSafeOnNewChain/index.tsx
+++ b/src/features/multichain/components/CreateSafeOnNewChain/index.tsx
@@ -69,49 +69,50 @@ const ReplaySafeDialog = ({
   const customRpc = useAppSelector(selectRpc)
   const dispatch = useAppDispatch()
   const [creationError, setCreationError] = useState<Error>()
-  const [isSubmitDisabled, setIsSubmitDisabled] = useState<boolean>(false)
+  const [isSubmitting, setIsSubmitting] = useState<boolean>(false)
 
   // Load some data
   const [safeCreationData, safeCreationDataError, safeCreationDataLoading] = safeCreationResult
 
   const onFormSubmit = handleSubmit(async (data) => {
-    setIsSubmitDisabled(true)
+    setIsSubmitting(true)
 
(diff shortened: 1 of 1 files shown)


src/features/multichain/components/CreateSafeOnNewChain/index.tsx
```diff
@@ -69,27 +69,33 @@ const ReplaySafeDialog = ({
   const customRpc = useAppSelector(selectRpc)
   const dispatch = useAppDispatch()
   const [creationError, setCreationError] = useState<Error>()
+  const [isSubmitDisabled, setIsSubmitDisabled] = useState<boolean>(false)
```
schmanu, line 72: This is very similar to the `submitDisabled` variable.
Should we rename this to `isSubmitting` and inverse it?

src/features/multichain/components/CreateSafeOnNewChain/index.tsx
```diff
@@ -69,27 +69,33 @@ const ReplaySafeDialog = ({
...
+      setIsSubmitDisabled(false)
       return
     }
 
     // 1. Double check that the creation Data will lead to the correct address
     const predictedAddress = await predictAddressBasedOnReplayData(safeCreationData, provider)
     if (!sameAddress(safeAddress, predictedAddress)) {
       setCreationError(new Error('The replayed Safe leads to an unexpected address'))
+      setIsSubmitDisabled(false)
       return
```
schmanu, line 99: Should we just use a try-catch-finally where we `setIsSubmitDisabled(false)` in the finally block?

schmanu (approved):

PR #4221, opened 2024-09-20
Feat: cached safe overviews

schmanu:
## What it solves
Rewrites `useSafeOverviews` hook using RTK query enabling:
- caching
- more efficient batching of requests
- easier interface to query safe overviews

The goal is to have two hooks:
- One that loads a single Safe overview for any Safe.
- One that loads multiple Safe overviews

Both of them use a shared queue that waits for more requests to come in for 50ms or triggers a requests once 10 requests are in the queue.

Other changes
- Do not fetch Safe overviews for undeployed Safes

## How to test it
While monitoring the outgoing network requests via e.g. developer tools in the browser:
- Connect a wallet and open the sidebar / account list.
- Select and open any Safe
- Now re-open the sidebar. 
- This should not trigger any new requests as the data should still be cached.

## Stats
On a full sidebar with multichain Safes this optimization with a fresh browser state reduces the safe overview requests from 38 to 18.

## Checklist


src/store/safeOverviews.ts
```diff
@@ -0,0 +1,164 @@
...
+      return
+    }
+
+    nextBatch.forEach((item) => {
+      const overview = overviews.find(
+        (entry) => sameAddress(entry.address.value, item.safeAddress) && entry.chainId === item.chainId,
+      )
+
+      item.callback({ data: overview })
+    })
```
clovisdasilvaneto, line 73: since we're already calling the item.callback in the catch block, shouldn't this line be inside the try block?

src/features/multichain/components/SignerSetupWarning/InconsistentSignerSetupWarning.tsx
```diff
@@ -37,14 +36,19 @@ export const InconsistentSignerSetupWarning = () => {
...
   const multiChainGroupSafes = useMemo(
     () => allMultiChainSafes?.find((account) => sameAddress(safeAddress, account.safes[0].address))?.safes ?? [],
     [allMultiChainSafes, safeAddress],
   )
-  const [safeOverviews] = useSafeOverviews(multiChainGroupSafes)
+  const deployedSafes = useMemo(
+    () => multiChainGroupSafes.filter((safe) => undeployedSafes[safe.chainId]?.[safe.address] === undefined),
+    [multiChainGroupSafes, undeployedSafes],
+  )
+  const { data: safeOverviews } = useGetMultipleSafeOverviewsQuery({ safes: deployedSafes, currency })
```
usame-algan, line 51: Do we need to pass the walletAddress here as well?

src/store/index.ts
```diff
@@ -53,6 +53,7 @@ const rootReducer = combineReducers({
   [ofacApi.reducerPath]: ofacApi.reducer,
   [safePassApi.reducerPath]: safePassApi.reducer,
   [slices.gatewayApi.reducerPath]: slices.gatewayApi.reducer,
+  [slices.safeOverviewApi.reducerPath]: slices.safeOverviewApi.reducer,
```
usame-algan, line 56: Wdyt about making this part of the `gatewayApi` slice to keep all related calls in one place? The `SafeOverviewFetcher` could be moved to its own file.

usame-algan (approved):
Great idea to batch and cache these requests! We should do the same for `useAllOwnedSafes`.

src/features/multichain/components/SignerSetupWarning/InconsistentSignerSetupWarning.tsx
schmanu: I do not pass it to have less dependencies and more likely cache hits. 
This component is only interested in the signers and the threshold. The wallet address fetches the pending actions which is not needed here.

PR #4229, opened 2024-09-23
fix(Multichain): do not disable currently selected network in the network selector

jmealy:
## What it solves
In the network multi select for safe creation, if the selected network does not have the multichain feature enabled, then the network is disabled after selection, so you cannot deselect it.

## How this PR fixes it
If the feature is not enabled on the currently selected network, all other networks are disabled, but the currently selected one is not.

## How to test it
- In the safe creation flow, use the network selector to select a chain without multichain enabled (eg. Mantle, Holesky on dev)
- The selected option should not become disabled.

## Screenshots
Before:
image

After:
image

## Checklist


schmanu (approved):
Good catch.

PR #4233, opened 2024-09-23
fix(Multichain): add more unsupported Safe cases

schmanu:
## What it solves
Disables adding networks for:
- Safes using reimbursement on creation
- Safes with unknown fallback handler

## How to test it
- Try to add a network to a Safe that used reimbursements (`payment > 0`)
- Try to add a network to a Safe with a custom fallback handler

## Checklist

### DIFF
--- a/src/features/multichain/hooks/useCompatibleNetworks.ts
+++ b/src/features/multichain/hooks/useCompatibleNetworks.ts
@@ -1,9 +1,9 @@
 import { type ReplayedSafeProps } from '@/features/counterfactual/store/undeployedSafesSlice'
 import useChains from '@/hooks/useChains'
-import { sameAddress } from '@/utils/addresses'
+import { hasMatchingDeployment } from '@/services/contracts/deployments'
 import { type SafeVersion } from '@safe-global/safe-core-sdk-types'
 import {
-  type SingletonDeploymentV2,
+  getCompatibilityFallbackHandlerDeployments,
   getProxyFactoryDeployments,
   getSafeL2SingletonDeployments,
   getSafeSingletonDeployments,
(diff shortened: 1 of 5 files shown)


src/features/multichain/hooks/useCompatibleNetworks.ts
```diff
@@ -35,27 +25,25 @@ export const useCompatibleNetworks = (
...
 
-  const allL2SingletonDeployments = SUPPORTED_VERSIONS.map((version) =>
-    getSafeL2SingletonDeployments({ version }),
-  ).filter(Boolean) as SingletonDeploymentV2[]
+  const { fallbackHandler } = safeAccountConfig
 
-  const allProxyFactoryDeployments = SUPPORTED_VERSIONS.map((version) =>
-    getProxyFactoryDeployments({ version }),
-  ).filter(Boolean) as SingletonDeploymentV2[]
+  console.log('Fallbackhandler', fallbackHandler)
```
usame-algan, line 32:
```suggestion

```

src/services/contracts/deployments.ts
```diff
@@ -28,6 +29,31 @@ export const hasCanonicalDeployment = (deployment: SingletonDeploymentV2 | undef
...
+/**
+ * Checks if any of the deployments returned by the `getDeployments` function for the given `network` and `versions` contain a deployment for the `contractAddress`
+ *
+ * @param getDeployments function to get the contract deployments
+ * @param contractAddress address that should be included in the deployments
+ * @param network chainId that is getting checked
+ * @param versions supported Safe versions
+ * @returns true if a matching deployment was found
+ */
+export const hasMatchingDeployment = (
```
usame-algan, line 41: Should we test this function on its own as well or do you think it has enough coverage through testing `useCompatibleNetworks`?

usame-algan (approved):

src/services/contracts/deployments.ts
schmanu: According to the coverage report those lines are covered so I'd say it should be enough.

PR #4237, opened 2024-09-23
Fix(Multichain): Keep focus within text input when typing name in replay modal [SW-205]

jmealy:
## What it solves
When typing some letters in the name input within the replay safe modal, the focus is moved away from the input.
Related: https://github.com/mui/material-ui/issues/19116

## How this PR fixes it
Stops keyboard events in the text input.

## How to test it
- Open a multichain safe.
- In the network selector, open the modal to replay the safe on another chain
- In the text input for setting the name, highlight the existing text and start typing. No matter what letter you start with, it should type as expected. (problematic letters were s, e, and p)

## Checklist

### DIFF
--- a/src/components/common/NameInput/index.tsx
+++ b/src/components/common/NameInput/index.tsx
@@ -27,6 +27,7 @@ const NameInput = ({
       fullWidth
       required={required}
       className={inputCss.input}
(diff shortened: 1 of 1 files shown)


schmanu (approved):
Wow. Good find!

PR #4254, opened 2024-09-26
fix: Push notifications toggle

### LINKED ISSUE #4253: Can not enable notifications for one safe
## Bug description
Can not enable notifications for one safe 
## Environment
 - Browser: Chrome/FF
 - Wallet: MetaMask
 - Chain: Sepolia mainnet

## Steps to reproduce
 1. Go to Sepolia safe
 2. go to the notifications
 3. turn on notification toggle
 4. sign message
 Current result: the message is signed but the toggle is not changed to on 

## Expected result
The notification after signing message should be enabled in the browser for the safe

usame-algan:
## What it solves

Resolves #4253 

## How this PR fixes it

- https://github.com/safe-global/safe-gateway-typescript-sdk/pull/189 changed the response from being undefined to be an empty object so we check for `isEmpty` instead of `null`. 

## How to test it

1. Open a Safe
2. Go to Notification settings
3. Turn on notifications
4. Sign the message
5. Observe the Toggle turns on

## Screenshots
Screenshot 2024-09-26 at 11 44 21

## Checklist

### DIFF
--- a/src/components/settings/PushNotifications/hooks/useNotificationRegistrations.ts
+++ b/src/components/settings/PushNotifications/hooks/useNotificationRegistrations.ts
@@ -1,4 +1,5 @@
 import { registerDevice, unregisterDevice, unregisterSafe } from '@safe-global/safe-gateway-typescript-sdk'
+import isEmpty from 'lodash/isEmpty'
 
 import { useAppDispatch } from '@/store'
 import { showNotification } from '@/store/notificationsSlice'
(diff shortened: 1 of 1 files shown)


src/components/settings/PushNotifications/hooks/useNotificationRegistrations.ts
```diff
@@ -19,7 +20,7 @@ const registrationFlow = async (registrationFn: Promise<unknown>, callback: () =
 
     // Gateway will return 200 with an empty payload if the device was (un-)registered successfully
     // @see https://github.com/safe-global/safe-client-gateway-nest/blob/27b6b3846b4ecbf938cdf5d0595ca464c10e556b/src/routes/notifications/notifications.service.ts#L29
-    success = response == null
+    success = isEmpty(response)
```
jmealy, line 23: Why do we need this check? If we get to this point then we've gotten a 200, no?

usame-algan: I also think so otherwise if it threw we would end up in the catch. I assume this was added in case the response is successful but not empty. For now I would leave it as is since the notifications are being reworked on the service side so we will have to touch this again anyway.

katspaugh (approved):

liliya-soroka:
Verified
Toggle works but , notifications are not received.
The issue was reported to the protocol team

PR #4260, opened 2024-09-26
fix(display-raw-data): Show raw data when the method name and parameters is missing in a tx

### LINKED ISSUE #4215: Show raw data for empty methods without params
## Bug description
If a contract is missing ABI information (e.g. unverified contracts) or if `fallback` functions are getting called our decoder currently returns an empty method name without parameters.
In such cases we should display the raw `data`.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Use  e.g. Bungee Safe App to bridge native ETH or call any other `fallback()` function with data.
 2. Observe that we do not display the `data` but instead just display "No parameters"

## Expected result
- Show the `to` address and the hex `data`
OR
- show `fallback()` as function name and raw data (if the contract has a `fallback()` function)

## Obtained result
We display the `to` address and "No Parameters"

## Screenshots
Screenshot 2024-09-20 at 09 40 20

clovisdasilvaneto:
## What it solves

Resolves #4215

## How this PR fixes it
it adds the hexData in the txData of the `DecodedTx` component and check if there is parameters when the method name is missing, if there are no parameters and method name is missing, we show the raw hexData

## How to test it
1) Go to transaction builder
2) Create a new transacation with custom data
3) Send it

**Expected**: The raw data should appears in the confirmation screen as screenshot bellow

## Screenshots
Screenshot 2024-09-26 at 17 03 53

## Checklist


src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
```diff
@@ -50,7 +50,7 @@ export const DecodedData = ({ txData, toInfo }: Props): ReactElement | null => {
   const avatar = addressInfo?.logoUri || toInfo?.logoUri || txData.to.logoUri
 
   let decodedData = <></>
-  if (txData.dataDecoded) {
+  if (txData.dataDecoded && (method || (!method && txData.dataDecoded.parameters))) {
```
katspaugh, line 53:
```suggestion
  if (txData.dataDecoded && (method || txData.dataDecoded.parameters)) {
```

katspaugh (approved):

src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
clovisdasilvaneto, line 53: this check for the parameters and method name is needed because for a native token transfer (from transaction builder) the decoded data comes also with an empty method: 

```js
{
    "type": "GENERIC",
    "method": "",
    "parameters": [
        {
            "name": "to",
            "type": "address",
            "value": "0xDa5e9FA404881Ff36DDa97b41Da402dF6430EE6b"
        },
        {
            "name": "value",
            "type": "uint256",
            "value": "1000000000000000000"
        }
    ]
}
```

clovisdasilvaneto: true haha

schmanu (commented):
Works well!
Could you add a unit test case to e.g. `DecodedTx/index.test.ts`?

schmanu (approved):

francovenica:
LGTM

Snapshots comparing current prod with this PR:
image
image

PR #4264, opened 2024-09-27
Chore: pull-requests: read permisison

katspaugh:
## What it solves

A follow-up on #4236.

Replacing `pull-requests: write` with a safer read permission.


katspaugh:
Looks like `read` isn't sufficient for this job, it's failing now. Closing.

PR #4270, opened 2024-09-27
fix(staking): fee displayed in wrong format [SW-242]

compojoom:
## What it solves
The displayed fee was wrong.

## How this PR fixes it
the fee is returned by the API in decimal format, so we have to multiply by 100 to get the percentage

## How to test it
Make a deposit and check that the values in the widget match the values in the deocding

## Checklist

### DIFF
--- a/src/features/stake/components/StakingConfirmationTx/Deposit.tsx
+++ b/src/features/stake/components/StakingConfirmationTx/Deposit.tsx
@@ -20,6 +20,8 @@ const CURRENCY = 'USD'
 const StakingConfirmationTxDeposit = ({ order }: StakingOrderConfirmationViewProps) => {
   const isOrder = order.type === ConfirmationViewTypes.KILN_NATIVE_STAKING_DEPOSIT
 
+  // the fee is returned in decimal format, so we multiply by 100 to get the percentage
+  const fee = (order.fee * 100).toFixed(2)
   return (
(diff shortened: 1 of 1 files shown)


src/features/stake/components/StakingConfirmationTx/Deposit.tsx
```diff
@@ -20,6 +20,8 @@ const CURRENCY = 'USD'
 const StakingConfirmationTxDeposit = ({ order }: StakingOrderConfirmationViewProps) => {
   const isOrder = order.type === ConfirmationViewTypes.KILN_NATIVE_STAKING_DEPOSIT
 
+  // the fee is returned in decimal format, so we multiply by 100 to get the percentage
+  const fee = (order.fee * 100).toFixed(2)
```
clovisdasilvaneto, line 24: shall we put it into a utility function, in order to re-use that in future scenarios?

clovisdasilvaneto (approved):
Codewise looks good

src/features/stake/components/StakingConfirmationTx/Deposit.tsx
compojoom: I thought about this, but don't want to "optimize prematurely". That's the only place where we are showing this right now (as far as I remember). If we need it later, we can extract it.

PR #4280, opened 2024-09-30
fix: duplicated advanced data

compojoom:
## What it solves
when you had a queued TX and you tried to execute it the "call method on contract" was shown twice.

grafik

Resolves #

## How this PR fixes it
DecodedTx was rendering a decodedData block, which was also rendered by the Summary component.

Right now I’m hiding the decodedData block in the summary based on a prop that is passed by the parent component.

This will be refactored and tested over here:
https://github.com/safe-global/safe-wallet-web/issues/4091

## How to test it
Create a new tx, sign it. Then go in the queue and try to execute it. Now the "call x method on y contract" should be shown just once.

## Checklist

### DIFF
--- a/src/components/tx/DecodedTx/index.tsx
+++ b/src/components/tx/DecodedTx/index.tsx
@@ -45,7 +45,6 @@ const DecodedTx = ({
   const isMultisend = !!decodedData?.parameters?.[0]?.valueDecoded
   const isMethodCallInAdvanced = !showMethodCall || (isMultisend && showMultisend)
 
-  console.log('DecodedTx', tx, txId, decodedData, isMultisend, showMultisend, showMethodCall, isMethodCallInAdvanced)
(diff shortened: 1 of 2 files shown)


src/components/tx/DecodedTx/index.tsx
```diff
@@ -45,6 +45,7 @@ const DecodedTx = ({
   const isMultisend = !!decodedData?.parameters?.[0]?.valueDecoded
   const isMethodCallInAdvanced = !showMethodCall || (isMultisend && showMultisend)
 
+  console.log('DecodedTx', tx, txId, decodedData, isMultisend, showMultisend, showMethodCall, isMethodCallInAdvanced)
```
usame-algan, line 48:
```suggestion

```

usame-algan (approved):

PR #4281, opened 2024-09-30
[Multichain] fix: Check all fallbackHandler deployments [SW-236]

usame-algan:
## What it solves

Resolves [SW-236](https://www.notion.so/safe-global/Unofficial-fallback-handler-message-for-cloned-safe-10e8180fe57380d4aa37fcd48feda821)

## How this PR fixes it

Uses `getCompatibilityFallbackHandlerDeployments` to get all deployment addresses in case there is a canonical and eip155 deployment and checks those addresses to decide whether its an official deployment or not

## How to test it

1. Open a Safe on a network with multiple deployments e.g. oeth:0xa4Fea8eF4a5dc35C5627Da7784b065934a5693a7
2. Go to the settings page and observe no unofficial fallback handler warning

## Screenshots

Screenshot 2024-09-30 at 11 32 13

## Checklist


src/components/settings/FallbackHandler/index.tsx
```diff
@@ -35,7 +36,8 @@ export const FallbackHandler = (): ReactElement | null => {
 
   const hasFallbackHandler = !!safe.fallbackHandler
   const isOfficial =
-    hasFallbackHandler && safe.fallbackHandler?.value === fallbackHandlerDeployment?.networkAddresses[safe.chainId]
+    safe.fallbackHandler &&
+    fallbackHandlerDeployments?.networkAddresses[safe.chainId].includes(safe.fallbackHandler.value)
   const isTWAPFallbackHandler = safe.fallbackHandler?.value === TWAP_FALLBACK_HANDLER
```
schmanu, line 41: I know you did not introduce this. But this check is too simplified. There are many chains where no Extensible Fallback handler is deployed e.g. Base. 
Maybe we should only allow this on Mainnet, Gnosis Chain, Sepolia and Arbitrum for now as those are the chains where CoW exists?

usame-algan: Wouldn't `isTWAPFallbackHandler` be false on those other chains where its not deployed which is the expected behaviour?

schmanu: But it just compares an hard coded address:
```ts
export const TWAP_FALLBACK_HANDLER = '0x2f55e8b20D0B9FEFA187AA7d00B6Cbe563605bF5'
```

There are no checks that this address is a contract on a chain so this will always be true when you replay a twap Safe for instance in case someone sets one up with the twap fallback handler.

schmanu: These are the networks where that fallback handler is deployed:
https://github.com/cowprotocol/composable-cow/blob/main/networks.json

usame-algan: Ah I see. Then lets harden the condition by also checking that the chainId matches one of the deployed networks. I hard-coded the list of those networks provided by the link for now.

usame-algan: Also added some tests for the twap fallback handler message

src/features/swap/helpers/utils.ts
```diff
@@ -24,6 +24,9 @@ function asDecimal(amount: number | bigint, decimals: number): number {
 
 export const TWAP_FALLBACK_HANDLER = '0x2f55e8b20D0B9FEFA187AA7d00B6Cbe563605bF5'
 
+// https://github.com/cowprotocol/composable-cow/blob/main/networks.json
+export const TWAP_FALLBACK_HANDLER_NETWORKS = ['1', '100', '11155111', '42161']
```
schmanu, line 28: Nit: We could also use a mapping chainId -> address here. But maybe we can change it if there are ever multiple addresses.

schmanu (approved):

PR #4284, opened 2024-09-30
[Multichain] fix: Adjust multichain design [SW-244]

usame-algan:
## What it solves

Resolves [SW-244](https://www.notion.so/safe-global/Small-design-tweaks-1118180fe57380e08ee4f1f3477f5c6b)

## How this PR fixes it

Addresses feedback from this document: https://www.notion.so/safe-global/Insights-Feedback-on-Multichain-10c8180fe57380eba5efded223e0aa70

- Reuses `NetworkLogosList` inside the SafeList
- Adjusts `NetworkLogosList` to optionally only display the first 4 networks

## How to test it

Compare screenshots to Figma design

## Screenshots
Screenshot 2024-09-30 at 15 05 46
Screenshot 2024-09-30 at 15 06 00
Screenshot 2024-09-30 at 15 07 10
Screenshot 2024-09-30 at 15 07 21

## Checklist

### DIFF
--- a/src/components/common/ModalDialog/index.tsx
+++ b/src/components/common/ModalDialog/index.tsx
@@ -20,7 +20,11 @@ interface DialogTitleProps {
 
 export const ModalDialogTitle = ({ children, onClose, hideChainIndicator = false, ...other }: DialogTitleProps) => {
   return (
-    <DialogTitle data-testid="modal-title" sx={{ m: 0, p: 2, display: 'flex', alignItems: 'center' }} {...other}>
+    <DialogTitle
+      data-testid="modal-title"
+      sx={{ m: 0, px: 3, pt: 3, pb: 2, display: 'flex', alignItems: 'center', fontWeight: 'bold' }}
+      {...other}
+    >
       {children}
       <span style={{ flex: 1 }} />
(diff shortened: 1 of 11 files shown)


schmanu (commented):
Its looking really nice.

Do you think we should revert this tooltip for the multichain Safes? Or is it still interesting for the case that there are 5+ chains?
Screenshot 2024-09-30 at 15 18 48

cc @TanyaEfremova

usame-algan:
> Do you think we should revert this tooltip for the multichain Safes? Or is it still interesting for the case that there are 5+ chains?

I think we should at least show it if there are more than 5 safes otherwise it would just say +2 or something without the user knowing which other 2 networks there are.

schmanu (approved):
Vamos 🚀

TanyaEfremova:
Agreed with @usame-algan regarding the tooltip. It still has additional value for >5 networks.

The only remark I have is the border color, it should be 'background-main' (so #ffffff). Could we do this small change? 
Otherwise, it looks great!

usame-algan:
> The only remark I have is the border color, it should be 'background-main' (so #ffffff). Could we do this small change?

Thanks! Adjusted the border color to be `background.paper` so its not visible in the list against the background

PR #4287, opened 2024-09-30
fix: Handle special characters for prefix in address input

usame-algan:
## What it solves

Noticed that the `AddressInput` unit test randomly fails sometimes. This happens when faker generates a shortName with a hyphen or other special characters in it so I adjusted our cleanup function to handle such characters.

## How this PR fixes it

- Updates `cleanInputValue` to not only handle `a-z0-9` characters in the prefix but any character that is not a `:`
- Adds unit tests for `cleanInputValue`

## How to test it

1. Create a send transaction
2. Input `test-prefix:<some address>` into the recipient field
3. Observe that the value is not changed
4. Input a real chain prefix like `eth:<some address>`
5. Observe that it is the same behaviour as previously

## Screenshots

On prod when typing `test-hyphen:<some address>`
Screenshot 2024-09-30 at 16 44 36

With the change when typing `test-hyphen:<some address>`
Screenshot 2024-09-30 at 16 44 23

## Checklist

### DIFF
--- a/src/utils/__tests__/addresses.test.ts
+++ b/src/utils/__tests__/addresses.test.ts
@@ -1,4 +1,4 @@
-import { checksumAddress, isChecksummedAddress, parsePrefixedAddress, sameAddress } from '../addresses'
+import { checksumAddress, cleanInputValue, isChecksummedAddress, parsePrefixedAddress, sameAddress } from '../addresses'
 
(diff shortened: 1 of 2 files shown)


schmanu (approved):

francovenica:
Never thought about entering a network with a special character. LGTM

PR #4288, opened 2024-09-30
Feat(Multichain): add feature flag checks when adding a new network [SW-240]

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Add-feature-toggle-for-adding-networks-10e8180fe57380b4b27df647dae49742

## How this PR fixes it
- adds new feature flag `MULTI_CHAIN_SAFE_ADD_NETWORK`.
- Checks that the current safe network has the feature enabled anywhere we show the option to add a network.
- Checks that other networks have the feature enabled before showing them as an option to add to the group.

## How to test it
- When the feature flag is not enabled on the current chain, you should not see the option to add a network:
    - in the safe list context menu
    - at the bottom of multichain groups in the safe list
    - in the network selector
- When the feature flag is enabled on the current chain but not enabled on another compatible chain, disabled chains should not be shown:
    - in the network selector
    - in the add network modal

## Screenshots
image

## Checklist


schmanu (changes requested):
I think feature flags should ideally hide the feature completely.

So in this case disabling the new feature should:
- Hide the "Add new network" button in the sidebar 
  - In the context menu and multichain groups
- Not offer adding networks through the network selector

jmealy:
>   * In the context menu and multichain groups

Makes sense, I hid the option to add networks in the selector and the context menu for single safes.
For multichain groups, if it's a group already then the feature must be enabled on the already deployed chains so it would make sense to still show it in this case I think. And then for other chains that don't have the flag, show them as disabled in the network input (or hidden maybe). wdyt?

schmanu (approved):

PR #4289, opened 2024-09-30
Fix(Multichain): ignore empty safe address when loading safeOverviews

schmanu:
## What it solves
We use `""` as fallback value for `useSafeAddress`. This caused empty requests to be triggered for the safe overview endpoint.

This fixes it by checking the safeAddress parameter before loading the safe overviews.

## How to test it
- Observe no more empty requests while the safe information is loading

## Checklist

### DIFF
--- a/src/store/api/gateway/safeOverviews.ts
+++ b/src/store/api/gateway/safeOverviews.ts
@@ -5,6 +5,7 @@ import { sameAddress } from '@/utils/addresses'
 import type { RootState } from '../..'
 import { selectCurrency } from '../../settingsSlice'
 import { type SafeItem } from '@/components/welcome/MyAccounts/useAllSafes'
+import { asError } from '@/services/exceptions/utils'
 
 type SafeOverviewQueueItem = {
   safeAddress: string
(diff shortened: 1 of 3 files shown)


src/store/api/gateway/safeOverviews.ts
```diff
@@ -125,22 +125,25 @@ export const safeOverviewEndpoints = (
...
-        return { error: { status: 'CUSTOM_ERROR', error: (error as Error).message } }
-      }
+  getSafeOverview: builder.query<SafeOverview | null, { safeAddress: string; walletAddress?: string; chainId: string }>(
+    {
+      async queryFn({ safeAddress, walletAddress, chainId }, { getState }) {
+        const state = getState()
+        const currency = selectCurrency(state as RootState)
+
+        if (!safeAddress) {
+          return { data: null }
```
schmanu, line 135: Had to change the return type to `null` because `undefined` is not allowed as return type for `queryFn`

iamacook (approved):

iamacook (approved):

usame-algan (approved):

PR #4300, opened 2024-10-02
Fix(Multichain): do not migrate to L2 if no lib contract is available

schmanu:
## What it solves
On e.g. Gnosis Chiado no migration libs are deployed. We still tried to migrate from L1 to L2 though.

## How this PR fixes it
- Does not prepend a migration call if the chain does not have the migration contracts deployed.

## How to test it
- Create a L1 1.3.0 Mainnet Safe
- Add Gnosis Chiado to it
- Activate and do any tx
- It should not try to migrate

## Checklist


usame-algan (approved):
There is a failing unit-test

PR #4302, opened 2024-10-02
Fix(Multichain): validate account data

schmanu:
## What it solves
Validates the `safeAccountConfig` of counterfactual Safes as well.

## How to test it
- Try to replay a counterfactual Safe that uses a `payment` or unknown `to` address.

## Checklist


usame-algan (approved):

src/features/multichain/hooks/useSafeCreationData.ts
usame-algan, line 1: Out of scope here but I wonder if we need this nested try catch structure inside `useSafeCreationData`. If there is an error for one of the chains, shouldn't it already propagate to the outer catch where we log the _816 error?

PR #4311, opened 2024-10-04
[Multichain] fix: Add more analytics events for multichain [SW-165]

usame-algan:
## What it solves

Resolves [SW-165](https://www.notion.so/safe-global/Check-analytics-7f1cdc962c6f45e1a1a79dc79055e948)

## How this PR fixes it

- Adds the missing events as described in the ticket

## How to test it

Check that the events described in SW-165 are emitted correctly

## Checklist

### DIFF
--- a/src/components/welcome/MyAccounts/SubAccountItem.tsx
+++ b/src/components/welcome/MyAccounts/SubAccountItem.tsx
@@ -6,7 +6,7 @@ import { ListItemButton, Box, Typography, Chip, Skeleton } from '@mui/material'
 import Link from 'next/link'
 import SafeIcon from '@/components/common/SafeIcon'
 import Track from '@/components/common/Track'
-import { OVERVIEW_EVENTS, OVERVIEW_LABELS } from '@/services/analytics'
+import { OVERVIEW_EVENTS, OVERVIEW_LABELS, trackEvent } from '@/services/analytics'
 import { AppRoutes } from '@/config/routes'
 import { useAppSelector } from '@/store'
 import { selectChainById } from '@/store/chainsSlice'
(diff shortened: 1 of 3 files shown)


src/components/welcome/MyAccounts/SubAccountItem.tsx
```diff
@@ -55,11 +55,16 @@ const SubAccountItem = ({ onLinkClick, safeItem, safeOverview }: SubAccountItem)
 
   const cfSafeSetup = extractCounterfactualSafeSetup(undeployedSafe, chain?.chainId)
 
+  const onSwitchSafe = () => {
+    trackEvent({ ...OVERVIEW_EVENTS.SWITCH_NETWORK, label: chainId })
```
schmanu, line 59: I switch network the correct event here?
I could also switch to another Multichain Safe and therefore switch Safe and network or only switch to another Safe on the same network.

Maybe we should call it "choose Safe from sidebar" or something like that instead?

usame-algan: It should be fine here since its in the `SubAccountItem` so only emitted if the user switches between the same safe on a different network. I added this to match the network selector in the header which also allows switching between the same safe on different networks.

schmanu: But I can open a `SubAccountItem` of a different Safe. 
So for instance switch from Safe "0x00...01" on Gnosis Chain to Safe "0x00...02" on Polygon. 

Also now its impossible to distinguish between switching networks in the sidebar or through the network selector.

usame-algan: Good point! Maybe we should leave it out then since we already emit an `Open Safe` when a Safe is opened from the Sidebar.

schmanu (approved):

PR #4312, opened 2024-10-04
[ignore, just a poc] Safenet Opt-in in Safe Wallet Settings Page

mmv08:
## What it solves

Resolves https://github.com/safe-global/safenet/issues/37

## Checklist


src/components/tx-flow/flows/EnableSafenet/ReviewEnableSafenet.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  useEffect(() => {
+    async function getTxs(): Promise<SafeTransaction> {
+      if (params.tokensForPresetAllowances.length === 0) {
+        return await createEnableGuardTx(params.guardAddress)
+      }
+
+      const txs: MetaTransactionData[] = [(await createEnableGuardTx(params.guardAddress)).data]
+      txs.push(
+        ...params.tokensForPresetAllowances.map((tokenAddress) => {
```
rmeissner, line 27: this is a better approach in JS than using a `forEach`?

rmeissner (approved):

src/config/constants.ts
```diff
@@ -109,3 +109,6 @@ export const CHAINALYSIS_OFAC_CONTRACT = '0x40c57923924b5c5c5455c48d93317139adda
 
 export const ECOSYSTEM_ID_ADDRESS =
   process.env.NEXT_PUBLIC_ECOSYSTEM_ID_ADDRESS || '0x0000000000000000000000000000000000000000'
+
+// SafeNet
+export const SAFENET_API_URL = process.env.NEXT_PUBLIC_SAFENET_API_URL || 'https://safenet-testnet.h00.me'
```
gjeanmart, line 114: Can we keep that private until we have a proper environment 👀

src/components/tx-flow/flows/EnableSafenet/ReviewEnableSafenet.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const ERC20_INTERFACE = ERC20__factory.createInterface()
+
+export const ReviewEnableSafenet = ({ params }: { params: EnableSafenetFlowProps }) => {
+  const { setSafeTx, safeTxError, setSafeTxError } = useContext(SafeTxContext)
+
+  useEffect(() => {
+    async function getTxs(): Promise<SafeTransaction> {
+      if (params.tokensForPresetAllowances.length === 0) {
+        return await createEnableGuardTx(params.guardAddress)
```
gjeanmart, line 22: nitpick: no need to have `await` with return

gjeanmart (approved):
Looks good!

src/components/tx-flow/flows/EnableSafenet/ReviewEnableSafenet.tsx
mmv08: I think it's needed. We want to return the transaction, not a promise that resolves to a transaction.

If we remove the await, the caller must handle all the cases (fulfilment, rejection, etc.), but we want to do this in this function.

mmv08: https://jakearchibald.com/2017/await-vs-return-vs-return-await/

gjeanmart: > Note: Outside of try/catch blocks, return await is redundant. There's even an [ESLint rule to detect it](https://eslint.org/docs/latest/rules/no-return-await), but it allows it in try/catch.

Hmm, I'm confused now. You don't have a try/catch block for this call. Always thought it was redundant as mentioned.

mmv08: I meant that if we were to catch an error, then the await block would be needed. 

After a second look, I don't think there are errors we'd want to catch in this function, so I'll remove the await in the next PR

Btw, from the ESLint rule link, TIL:
> This rule was deprecated in ESLint v8.46.0 with no replacement. The original intent of this rule was to discourage the use of return await, to avoid an extra microtask. However, due to the fact that JavaScript now handles native Promises differently, there is no longer an extra microtask. More technical information can be found in [this V8 blog entry](https://v8.dev/blog/fast-async).

PR #4339, opened 2024-10-08
fix(Multichain): detect migration txs and mark them as trusted

schmanu:
## What it solves
The MultiSends of migrations are not flagged as trusted because they contain delegateCalls.

## How this PR fixes it
Adds a workaround that detects that the tx only contains a trusted migration delegateCall and nothing else.

## How to test it
Open e.g. this Safe:
`aurora:0x51b49d7812478A2eaa25fA1d5F7352600d3ABfD2`

## Checklist


usame-algan (approved):
Unrelated to this PR but we should add a small gap between the migration info and the delegate call info box
Screenshot 2024-10-09 at 09 33 31

PR #4344, opened 2024-10-09
fix: Track number of delegate transactions

### LINKED ISSUE #4206: Add analytics for delegate support
## What is the feature about

In #1825 we added delegate support in the UI. It would be good to know how much this is being used.

## The list of requirements

- Track an event when a delegate proposes a transaction

usame-algan:
## What it solves

Resolves #4206

## How this PR fixes it

- Extends `trackTxEvents` to handle delegate form submissions

## How to test it

1. Open a Safe that has a delegate
2. Create a transaction as a delegate
3. Observe a GA event being emitted once the user submits the tx in their wallet

## Checklist

### DIFF
--- a/src/components/tx/SignOrExecuteForm/index.tsx
+++ b/src/components/tx/SignOrExecuteForm/index.tsx
@@ -63,8 +63,13 @@ const trackTxEvents = (
   isCreation: boolean,
   isExecuted: boolean,
   isRoleExecution: boolean,
+  isDelegateCreation: boolean,
 ) => {
-  const creationEvent = isRoleExecution ? TX_EVENTS.CREATE_VIA_ROLE : TX_EVENTS.CREATE
+  const creationEvent = isRoleExecution
+    ? TX_EVENTS.CREATE_VIA_ROLE
+    : isDelegateCreation
+    ? TX_EVENTS.CREATE_VIA_DELEGATE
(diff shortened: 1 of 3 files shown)


src/components/tx/SignOrExecuteForm/index.tsx
```diff
@@ -63,8 +63,13 @@ const trackTxEvents = (
...
   isExecuted: boolean,
   isRoleExecution: boolean,
+  isDelegateCreation: boolean,
 ) => {
-  const creationEvent = isRoleExecution ? TX_EVENTS.CREATE_VIA_ROLE : TX_EVENTS.CREATE
+  const creationEvent = isRoleExecution
+    ? TX_EVENTS.CREATE_VIA_ROLE
+    : isDelegateCreation
+    ? TX_EVENTS.CREATE_VIA_DELEGATE
+    : TX_EVENTS.CREATE
```
compojoom, line 72: you know that nested nested ternaries trigger give me such anxiety :)

compojoom (approved):

src/components/tx/SignOrExecuteForm/index.tsx
usame-algan: Lets hope this is the last flag we have to pass to this function 🤞

liliya-soroka:
Verified

PR #4355, opened 2024-10-10
[Multichain] fix: Copy owners to address book when replaying safe [SW-270]

usame-algan:
## What it solves

Resolves [SW-270](https://www.notion.so/safe-global/Adding-new-network-does-not-update-address-book-1198180fe5738025b835d3aeb84d5162)

## How this PR fixes it

- Looks up the owner entry in the address book when replaying a safe and copies it over if it exists

## How to test it

1. Create a Safe and add owner names
2. Observe them in the address book
3. Replay the safe on another chain
4. Observe them still being in the address book on the new chain

## Checklist

### DIFF
--- a/src/components/new-safe/create/steps/ReviewStep/index.tsx
+++ b/src/components/new-safe/create/steps/ReviewStep/index.tsx
@@ -263,7 +263,7 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
       gtmSetChainId(chain.chainId)
 
       if (isCounterfactualEnabled && payMethod === PayMethod.PayLater) {
-        router?.push({
+        await router?.push({
           pathname: AppRoutes.home,
           query: { safe: `${data.networks[0].shortName}:${safeAddress}` },
         })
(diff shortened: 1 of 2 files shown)


src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -263,7 +263,7 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
       gtmSetChainId(chain.chainId)
 
       if (isCounterfactualEnabled && payMethod === PayMethod.PayLater) {
-        router?.push({
+        await router?.push({
```
usame-algan, line 266: I thought I added this already in the past but this prevents the success screen from being shown until the dashboard is opened otherwise it pops up immediately even while still seeing the review screen in the background

src/features/multichain/components/CreateSafeOnNewChain/index.tsx
```diff
@@ -73,6 +75,7 @@ const ReplaySafeDialog = ({
     trackEvent({ ...OVERVIEW_EVENTS.CANCEL_ADD_NEW_NETWORK })
     onClose()
   }
+  console.log(addressBook)
```
schmanu, line 78:
```suggestion
```

PR #4356, opened 2024-10-10
refactor: use rtk query for loading owned safes

### LINKED ISSUE #4350: Add caching to owned Safes
## Overview
Currently the owned Safe are not cached and fetched whenever someone opens the sidebar. This causes a lot of unnecessary requests on our CGW.

We should refactor the `useAllOwnedSafes` such that it uses RTK Query and caches its result for a while.

schmanu:
## What it solves

Resolves #4350 

## How this PR fixes it
Uses RTK query and its caching to load owned Safes.

## How to test it
- Toggle the sidebar and observe only one request for loading the owned Safes

## Checklist

### DIFF
--- a/src/store/gateway.ts
+++ b/src/store/gateway.ts
@@ -1,6 +1,13 @@
 import { createApi } from '@reduxjs/toolkit/query/react'
 
-import { getTransactionDetails, type TransactionDetails } from '@safe-global/safe-gateway-typescript-sdk'
+import {
+  type AllOwnedSafes,
+  type OwnedSafes,
+  getAllOwnedSafes,
+  getOwnedSafes,
+  getTransactionDetails,
+  type TransactionDetails,
+} from '@safe-global/safe-gateway-typescript-sdk'
 import type { BaseQueryFn } from '@reduxjs/toolkit/dist/query/baseQueryTypes'
 import type { FetchBaseQueryError } from '@reduxjs/toolkit/dist/query/react'
 import { getDelegates } from '@safe-global/safe-gateway-typescript-sdk'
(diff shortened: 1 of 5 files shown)


src/store/gateway.ts
```diff
@@ -48,6 +55,26 @@ export const gatewayApi = createApi({
...
+        }
+      },
+    }),
+    getAllOwnedSafes: builder.query<AllOwnedSafes, { address: string }>({
+      async queryFn({ address }) {
+        try {
+          const ownedSafes = await getAllOwnedSafes(address)
+          return { data: ownedSafes }
+        } catch (error) {
+          return { error: error as FetchBaseQueryError }
```
iamacook, line 74: Nit: I don't think these correctly match the [`FetchBaseQueryError` type](https://redux-toolkit.js.org/rtk-query/usage-with-typescript).

I'd suggest using the `asError` helper and retuen a `SerializedError` instead. You may not even need to cast.

iamacook (approved):

compojoom:
@iamacook we are not persisting this data? On refresh of the page we should fetch it from the server again. 

What I noticed that is that when the sidebar is closed there is no query that need the data so rtk is unsubscribing from the query after 60s. We should actually increase the caching time to maybe something like 10mins?

iamacook:
> @iamacook we are not persisting this data? On refresh of the page we should fetch it from the server again.

The cached [owners](https://github.com/safe-global/safe-wallet-web/blob/main/src/hooks/useOwnedSafes.ts#L20) and [all Safes](https://github.com/safe-global/safe-wallet-web/blob/main/src/components/welcome/MyAccounts/useAllOwnedSafes.ts#L16) are initially returned. This means that a returning user would already have a populated sidebar, for instance.

However, whether we want to keep this behavior is open for discussion, particularly when considering the complexity of implementing it with RTK Query.

iamacook:
I've since realised we should definitely test creating/being added to a Safe now that the caching has been adjusted, and if it correctly updates.

schmanu:
Closing it until we rework the sidebar. 

Reason:
This refactoring changes the behavior of the sidebar loading as well:
- Previously we always triggered a request but if we had a cached value we served that value until the new data came in.
- This behavior would be lost now

I think to properly improve this we should redesign the sidebar and rework the sidebar logic such that we do not require all owned Safes anymore to render the basic Sidebar.

PR #4360, opened 2024-10-11
refactor: simplify RTK Query `endpoints`

### LINKED ISSUE #4359: Avoid casting in RTK Query
## Bug description

We currently cast `error`s within our `gatewayApi` to `FetchBaseQueryError`. As such, we canno be certain that they match the intended type.

## Expected result

No casting is needed.

iamacook:
## What it solves

Resolves #4359

## How this PR fixes it

Our `gatewayApi` has been refactored to use the `fakeBaseQuery` of RTK Query. It accepts a generic, with which we can specify the error type we expect - an `Error` instance in our case. A builder for the `queryFn` within `endpoints` then wraps all Client Gateway calls, returning such instances.

## How to test it

There should be no noticeable difference in the UI. Usage of the above can, however, be checked as such:

- [ ] Loading the transaction details of a transaction in the transaction list works as expected, and caches the response.
- [ ] Loading the transaction details of batched transactions works as expected, and caches the response.
- [ ] Speeding up a Safe transaction dispatches a "Speed up transaction" tracking event correctly.
- [ ] No tooltip warning regarding connecting a supported wallet is shown in the UI for deletgates.

## Checklist

### DIFF
--- a/src/store/gateway.ts
+++ b/src/store/gateway.ts
@@ -1,51 +1,35 @@
-import { createApi } from '@reduxjs/toolkit/query/react'
+import { createApi, fakeBaseQuery } from '@reduxjs/toolkit/query/react'
 
 import { getTransactionDetails, type TransactionDetails } from '@safe-global/safe-gateway-typescript-sdk'
-import type { BaseQueryFn } from '@reduxjs/toolkit/dist/query/baseQueryTypes'
-import type { FetchBaseQueryError } from '@reduxjs/toolkit/dist/query/react'
+import { asError } from '@/services/exceptions/utils'
 import { getDelegates } from '@safe-global/safe-gateway-typescript-sdk'
 import type { DelegateResponse } from '@safe-global/safe-gateway-typescript-sdk/dist/types/delegates'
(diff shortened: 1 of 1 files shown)


schmanu:
Closing it until we rework the sidebar. 

Reason:
This refactoring changes the behavior of the sidebar loading as well:
- Previously we always triggered a request but if we had a cached value we served that value until the new data came in.
- This behavior would be lost now

I think to properly improve this we should redesign the sidebar and rework the sidebar logic such that we do not require all owned Safes anymore to render the basic Sidebar.

iamacook:
I'm reopening [at the request of @schmanu](https://github.com/safe-global/safe-wallet-web/pull/4360#issuecomment-2406688770).

compojoom (approved):
cool stuff @iamacook! 

I didn't know that the fakeBaseQuery function exists & removing the duplication in the code... awesome!

compojoom (approved):

francovenica:
Checked the cases described in the description. LGTM

PR #4370, opened 2024-10-14
[Multichain] fix: Only show network logo of activated safe [SW-287]

usame-algan:
## What it solves

Resolves [SW-287](https://www.notion.so/safe-global/all-added-networks-are-displayed-instead-one-network-where-the-safe-was-activated-11c8180fe573802bb404e5e6bff339a6)

## How this PR fixes it

- Narrows down `networks` when receiving an INDEXED event

## How to test it

1. Create a multichain safe on at least 2 networks
2. Observe all the network logos on the success screen
3. Activate one of them
4. Observe only the activated network logo on the success screen
5. Reload the page
6. Activate another safe
7. Observe the correct network logo on the success screen

## Checklist

### DIFF
--- a/src/features/counterfactual/CounterfactualSuccessScreen.tsx
+++ b/src/features/counterfactual/CounterfactualSuccessScreen.tsx
@@ -28,6 +28,7 @@ const CounterfactualSuccessScreen = () => {
         if (event === SafeCreationEvent.INDEXED) {
           if ('chainId' in detail) {
             setChainId(detail.chainId)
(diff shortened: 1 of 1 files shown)


usame-algan:
This will only work if the user hasn't reloaded the app between creation and activation and only for the first activation. If they activate a second safe on a different network the information is "lost" and we fall back to the current chain to display the network logo.

schmanu (approved):

PR #4371, opened 2024-10-14
fix: allowSpendingLimit in CheckWallet

schmanu:
## What it solves
`allowSpendingLimits` in `CheckWallet` did not work anymore.

## How this PR fixes it
Fixes the conditions in `CheckWallet` and the unit test that did not pick up the error

## How to test it
- Connect with a wallet that only has a spending limit
- Observe the "New transaction" button to be available

## Checklist

### DIFF
--- a/src/components/common/CheckWallet/index.tsx
+++ b/src/components/common/CheckWallet/index.tsx
@@ -50,7 +50,7 @@ const CheckWallet = ({
       return Message.SafeNotActivated
     }
 
-    if (!allowNonOwner && !isSafeOwner && !isDelegate && !isOnlySpendingLimit && !allowSpendingLimit) {
+    if (!allowNonOwner && !isSafeOwner && !isDelegate && (!isOnlySpendingLimit || !allowSpendingLimit)) {
       return Message.NotSafeOwner
     }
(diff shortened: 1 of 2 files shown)


usame-algan:
There is still a difference to prod where a spending limit benefiaciary can remove owners etc. These buttons should be disabled for non-owners.

src/components/common/CheckWallet/index.tsx
```diff
@@ -49,7 +49,8 @@ const CheckWallet = ({
     if (isUndeployedSafe && !allowUndeployedSafe) {
       return Message.SafeNotActivated
     }
-    if ((!allowNonOwner && !isSafeOwner && !isDelegate) || (isOnlySpendingLimit && !allowSpendingLimit)) {
+
+    if (!allowNonOwner && !isSafeOwner && !isDelegate && !isOnlySpendingLimit && !allowSpendingLimit) {
```
usame-algan, line 53: I suspect it should be `... && isOnlySpendingLimit && !allowSpendingLimit`

usame-algan (approved):
Looks good 👍

PR #4377, opened 2024-10-15
fix: Add fallbacks when parsing erc20 symbol and decimals

### LINKED ISSUE #4376: Spending limits are not visible for some token contracts
## Bug description

Some token contracts like MKR use bytes32 for the symbol instead of string so our ABI throws when trying to fetch them.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to a Safe that has a spending limit for the MKR token (0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2)
 2. Observe an error in the console and nothign showing in the spending limit list

## Expected result

Spending limits show up

## Obtained result

Spending limits are not showing up

usame-algan:
## What it solves

Resolves #4376 

## How this PR fixes it

- Tries to parse erc20 symbols with `parseBytes32String` in case it throws
- Adds fallbacks for symbol and decimals

## How to test it

1. Open a safe with MKR as a spending limit
2. Observe it showing up in the table

## Checklist

### DIFF
--- a/src/utils/tokens.ts
+++ b/src/utils/tokens.ts
@@ -1,5 +1,6 @@
 import { getWeb3ReadOnly } from '@/hooks/wallets/web3'
 import { ERC20__factory, ERC721__factory } from '@/types/contracts'
+import { parseBytes32String } from '@ethersproject/strings'
 import { type TokenInfo, TokenType } from '@safe-global/safe-gateway-typescript-sdk'
 
 export const UNLIMITED_APPROVAL_AMOUNT = 2n ** 256n - 1n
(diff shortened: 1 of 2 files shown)


src/utils/tokens.ts
```diff
@@ -19,7 +20,18 @@ export const getERC20TokenInfoOnChain = async (
...
+  const symbol = await erc20
+    .symbol()
+    .then((symbol) => symbol)
+    .catch((error) => parseBytes32String(error.value)) // Some older contracts use bytes32 instead of string
+    .finally(() => '')
+
+  const decimals = await erc20
+    .decimals()
+    .then((decimals) => decimals)
+    .catch(() => BigInt(18))
```
schmanu, line 33: This could potentially lead to wrong decimals. Does the contract also use a exotic decimal function?

usame-algan: Not that I know of. I will revert this for the decimals

schmanu (approved):

francovenica:
LGTM. I've checked in the safe where the issue was reported since it has the MKR tokens already. The allowances show in the list just fine now

image

PR #4385, opened 2024-10-15
fix: Show different message in success screen when safe is deployed

### LINKED ISSUE #4383: Wrong screen is displayed after the first safe activation in the group
## Bug description
The account is almost set is displayed for the activated safe
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1.Create one CF safe or group of CF safes
2. activate the first safe
Current result: Your account is almost set! is displayed instead of your account is set .
Expected result: make sure that correct screen is displayed for the first activated safe

## Screenshots
image

usame-algan:
## What it solves

Resolves #4383 

## How this PR fixes it

- If there is a `chainId` we assume it is not a counterfactual deployment and show a different message

## How to test it

1. Create a single safe and deploy immediately
2. Observe the success screen is saying "Your Account is all set"
3. Create a single safe and pay later
4. Observe the success screen is saying "Your account is almost set"
5. Create a multichain safe
6. Observe the success screen is saying "Your account is almost set"
7. Deploy one of the accounts
8. Observe the success screen is saying "Your account is all set"

## Checklist

### DIFF
--- a/src/features/counterfactual/CounterfactualSuccessScreen.tsx
+++ b/src/features/counterfactual/CounterfactualSuccessScreen.tsx
@@ -18,7 +18,7 @@ const CounterfactualSuccessScreen = () => {
   const [networks, setNetworks] = useState<ChainInfo[]>([])
   const addressBooks = useAllAddressBooks()
   const safeName = safeAddress && chain ? addressBooks?.[chain.chainId]?.[safeAddress] : ''
-  const isCFCreation = !!networks.length
+  const isCFCreation = !!networks.length && !chainId
   const isMultiChain = networks.length > 1
   const chainName = isMultiChain ? '' : isCFCreation ? networks[0].chainName : chain?.chainName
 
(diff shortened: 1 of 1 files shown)


src/features/counterfactual/CounterfactualSuccessScreen.tsx
```diff
@@ -18,7 +18,7 @@ const CounterfactualSuccessScreen = () => {
   const [networks, setNetworks] = useState<ChainInfo[]>([])
   const addressBooks = useAllAddressBooks()
   const safeName = safeAddress && chain ? addressBooks?.[chain.chainId]?.[safeAddress] : ''
-  const isCFCreation = !!networks.length
+  const isCFCreation = !!networks.length && !chainId
```
schmanu, line 21: I had to do some digging to understand why this works. It feels a bit disconnected / hard to follow.

Could we not store a state with the `SafeCreationEvent` instead and set it to `SafeCreationEvent.INDEXED` if we receive an indexed event and to `SafeCreationEvent.AWAITING_EXECUTION` if we receive that event?

Then we could change this quite easily to 
```suggestion
  const isCFCreation = event === SafeCreationEvent.AWAITING_EXECUTION
```

schmanu (approved):

PR #4406, opened 2024-10-22
feat: allow signing with owned parent Safes

schmanu:
## What it solves
Currently to sign a transaction with a parent Safe, you have to connect those two Safes via WalletConnect and multiple browser windows.

## How this PR fixes it
- Detects that the connected wallet owns a Safe that owns the currently opened Safe
- Lets users select a Safe to sign with if there are multiple potential owner Safes
- Signs transitively with the parent Safe instead of the the owner Wallet.

## How to test it
- Setup at least two Safes where one is owned by the other.
- Create a transaction in the "child" Safe
- Witness that you can choose to sign as the "parent" Safe

## Screenshots
Screenshot 2024-10-28 at 09 46 15
Screenshot 2024-10-28 at 09 48 43
Screenshot 2024-10-28 at 09 46 22

Nested signing flow with new status screen

https://github.com/user-attachments/assets/4f8ce40d-f621-4a40-827b-ecd12368dbac

## Checklist


src/components/tx-flow/flows/NestedTxSuccessScreen/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+            },
+          }}
+          passHref
+        >
+          <Button variant="outlined">Open the transaction</Button>
+        </Link>
+      </Box>
+    </Container>
+  ) : (
+    <ErrorMessage>No nested signing data found</ErrorMessage>
```
iamacook, line 110: Nit: the ternary can be removed with an early return, e.g.

```ts
if (pendingTx.status !== PendingStatus.NESTED_SIGNING) {
  return <ErrorMessage>No nested signing data found</ErrorMessage>
}

return ...
```

src/components/tx/SignOrExecuteForm/SignerForm/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const onChange = (event: SelectChangeEvent<string>) => {
+    setNestedSafeAddress(event.target.value)
+  }
+
+  const isOwner = useMemo(
+    () => safe.owners.map((owner) => owner.value).includes(wallet?.address ?? ''),
+    [wallet?.address, safe.owners],
+  )
+
+  const isNotNestedOwner = useMemo(() => nestedSafeOwners && nestedSafeOwners.length === 0, [nestedSafeOwners])
```
iamacook, line 37: Here is another place we could use the proposed `useIsNestedSafeOwner` hook.

src/components/tx/SignOrExecuteForm/SignerForm/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+  )
+
+  useEffect(() => {
+    if (nestedSafeAddress) {
+      return
+    }
+    if (!isOwner && nestedSafeOwners && nestedSafeOwners.length > 0) {
+      setNestedSafeAddress(nestedSafeOwners[0])
+    }
+    if (isOwner && (!nestedSafeOwners || nestedSafeOwners.length === 0)) {
```
iamacook, line 54: We could simplify these conditions as well.

src/components/tx/SignOrExecuteForm/SignerForm/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+  if (!wallet || isNotNestedOwner) {
+    return null
+  }
+
+  return (
+    <TxCard>
+      <Typography variant="h5">
+        Select signer
+        <Tooltip
+          title="Your connected wallet controls other Safe accounts, which can sign this transaction. You can select which account to sign with."
```
iamacook, line 68:
```suggestion
          title="Your connected wallet controls other Safe Accounts, which can sign this transaction. You can select which Account to sign with."
```

src/components/tx/SignOrExecuteForm/SignerForm/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+                ml: 0.5,
+              }}
+            />
+          </span>
+        </Tooltip>
+      </Typography>
+
+      <Box display="flex" alignItems="center" gap={1}>
+        <FormControl fullWidth size="medium">
+          <InputLabel id="signer-label">Signer account</InputLabel>
```
iamacook, line 89:
```suggestion
          <InputLabel id="signer-label">Signer Account</InputLabel>
```

src/components/tx/SignOrExecuteForm/SignerForm/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+            className={css.signerForm}
+            labelId="signer-label"
+            label="Signer account"
+            fullWidth
+            onChange={onChange}
+            value={nestedSafeAddress ?? options[0]}
+          >
+            {options?.map((owner) => (
+              <MenuItem key={owner} value={owner}>
+                <EthHashInfo address={owner} avatarSize={32} onlyName />
```
iamacook, line 100: Is there a reason we don't want to show the address? (In reference to the `onlyName` prop.)

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -58,48 +59,56 @@ export const useTxActions = (): TxActions => {
...
 
     const signTx: TxActions['signTx'] = async (safeTx, txId, origin) => {
       assertTx(safeTx)
-      assertWallet(wallet)
+      assertProvider(signer?.provider)
       assertOnboard(onboard)
 
       // Smart contract wallets must sign via an on-chain tx
-      if (await isSmartContractWallet(wallet.chainId, wallet.address)) {
+      if (await isSmartContractWallet(signer.chainId, signer.address)) {
```
iamacook, line 86: We could add the `signer.isSafe` flag here as well meaning Safes don't require the `isSmartContractWallet` call.

src/hooks/useNestedSafeOwners.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const [allOwned] = useAllOwnedSafes(wallet?.address ?? '')
+  allOwned
+
+  const nestedSafeOwner = useMemo(() => {
+    if (!safeLoaded || !allOwned) return null
+
+    // Find an intersection of owned safes and the owners of the current safe
+    const ownerAddresses = safe?.owners.map((owner) => owner.value)
+
+    return allOwned[safe.chainId]?.filter((ownedSafe) => ownerAddresses?.includes(ownedSafe))
```
iamacook, line 18: As we're only concerned about the current chain, we should use the `[v1/chains/{chainId}/owners/{ownerAddress}/safes` endpoint instead of getting all the Safes across all chains.

src/services/tx/tx-sender/dispatch.ts
```diff
@@ -141,18 +141,19 @@ export const dispatchOnChainSigning = async (
...
   safeAddress: string,
+  isNestedSafe: boolean,
 ) => {
   const sdk = await getSafeSDKWithSigner(provider)
   const safeTxHash = await sdk.getTransactionHash(safeTx)
   const eventParams = { txId, nonce: safeTx.data.nonce }
 
   const options = chainId === chains.zksync ? { gasLimit: ZK_SYNC_ON_CHAIN_SIGNATURE_GAS_LIMIT } : undefined
-
+  let txResponse: any
```
iamacook, line 151: If you type this as a `string`, you don't need to reassign the type in `parentSafeTxHash`.

src/tests/builders/safeTx.ts
```diff
@@ -65,3 +74,40 @@ export function safeSignatureBuilder(): IBuilder<SafeSignature> {
...
+  const num2 = faker.number.int({ min: 1, max: 10 })
+
+  return Builder.new<MultisigExecutionInfo>().with({
+    nonce: faker.number.int(),
+    type: DetailedExecutionInfoType.MULTISIG,
+    confirmationsRequired: Math.max(num1, num2),
+    confirmationsSubmitted: Math.min(num1, num2),
+    missingSigners: Array.from(new Array(Math.min(num1, num2))).map(() => ({
+      value: faker.finance.ethereumAddress(),
+    })),
```
iamacook, line 98: You can achieve this within `Array.from`:

```suggestion
    missingSigners: Array.from({ length: Math.min(num1, num2) }, () => ({
      value: faker.finance.ethereumAddress(),
    })),
```

src/utils/nested-safe-wallet.ts
```diff
@@ -0,0 +1,148 @@
...
+          if (safeInfo.threshold === 1) {
+            // Directly execute the tx
+            await connectedSDK.executeTransaction(safeTx)
+          } else {
+            const signedTx = await connectedSDK.signTransaction(safeTx)
+            await proposeTx(safeInfo.chainId, safeInfo.address.value, actualWallet.address, signedTx, safeTxHash)
+          }
+        }
+      } catch (err) {
+        console.error(err)
```
iamacook, line 94: We should `logError` here so that we can track issues.

src/utils/nested-safe-wallet.ts
```diff
@@ -0,0 +1,148 @@
...
+      chainId: Number(safeInfo.chainId),
+      safeAddress: safeInfo.address.value,
+    },
+    nestedSafeSdk,
+  )
+
+  return {
+    provider: {
+      async request(request) {
+        const result = await nestedSafeProvider.request(69420, request, {
```
iamacook, line 132: 🤣

iamacook (commented):
We need to emphasise the required testing of this. There are lots of signing changes across edge cases, e.g. batching, relaying, signing, proposing, executing.

src/components/tx/SignOrExecuteForm/SignerForm/index.tsx
schmanu: It just looks a bit more compact. `onlyName` only shows the name or the address as fallback if there is no given name.

src/components/tx-flow/flows/NestedTxSuccessScreen/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const pendingTx = useAppSelector((state) => (txId ? selectPendingTxById(state, txId) : undefined))
+  const addressBook = useAddressBook()
+
+  const { data: parentTxData } = useGetTransactionDetailsQuery(
+    pendingTx?.status === PendingStatus.NESTED_SIGNING
+      ? { chainId: pendingTx.chainId, txId: pendingTx.signingSafeTxHash }
+      : skipToken,
+  )
+  if (!pendingTx) {
+    return 'No tx found'
```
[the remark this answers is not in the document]
schmanu: This was actually useless and a leftover from my first iteration.

I removed this and changed the other condition to:
```
if (pendingTx?.status !== PendingStatus.NESTED_SIGNING) {
```

src/hooks/useNestedSafeOwners.tsx
schmanu: I also refactored the `useOwnedSafes` hook to use rtk query under the hood but kept the return type the same for now. Its something we should refactor as well though in a later iteration as the return type makes no sense currently.

src/components/tx-flow/flows/NestedTxSuccessScreen/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+    >
+      <Box padding={3} mt={3} display="flex" flexDirection="column" alignItems="center" gap={2}>
+        <Box className={css.icon}>
+          <SvgIcon component={NestedSafeIcon} inheritViewBox fontSize="large" alt="Nested Safe" />
+        </Box>
+        <Typography data-testid="transaction-status" variant="h6" marginTop={2} fontWeight={700}>
+          A nested transaction was created
+        </Typography>
+        <Typography variant="body2" mb={3}>
+          Once confirmed and executed this signer transaction will confirm the child Safe&apos;s transaction.
```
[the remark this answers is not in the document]
schmanu: I will leave this open for now as its more a product question.
But I agree. We need one term for the top level / parent Safe and one for the child Safe / Sub Safe.

src/components/common/WalletProvider/index.tsx
```diff
@@ -1,13 +1,49 @@
...
+import { useRouter } from 'next/router'
+import ExternalStore from '@/services/ExternalStore'
+import { type Eip1193Provider } from 'ethers'
+import { getNestedWallet } from '@/utils/nested-safe-wallet'
 
-export const WalletContext = createContext<ConnectedWallet | null>(null)
+const { setStore, useStore } = new ExternalStore<string | null>(null)
+
+export const setNestedSafeAddress = setStore
+export const useNestedSafeAddress = useStore
```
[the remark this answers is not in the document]
schmanu: I agree. I am generally not happy with how that works as of now.
The problem is that the selected signer should ideally only live within our transaction modal context. But now it lives globally and with your suggestion it still lives inside the `WalletProvider`. 

Because as of now we have to be careful to always reset it to `undefined` when the modal is closed as we otherwise still use the selected Safe as signer and some buttons turn disabled as we think your parent Signer is your connected signer.

I will give it a second thought.

src/components/common/ExternalLink/index.tsx
```diff
@@ -9,17 +9,25 @@ import { Box, Link, type LinkProps } from '@mui/material'
...
         {children}
         {!noIcon && <OpenInNewRounded fontSize="small" />}
       </Box>
     </Link>
+  ) : (
+    <Button variant="outlined" href={href} rel="noreferrer noopener" target="_blank" sx={props.sx}>
+      <Box display="inline-flex" alignItems="center" gap={0.2} component="span">
+        {children}
+        {!noIcon && <OpenInNewRounded fontSize="small" />}
+      </Box>
```
iamacook, line 29: Nit: this duplicates the `children` of the above `Link`. Let's move it to a variable and use it inside both.

src/components/tx-flow/flows/NestedTxSuccessScreen/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+            </Typography>
+          </Stack>
+          <Box display="flex" flexDirection="column" alignItems="start" gap={1}>
+            <Typography variant="body2" color="text.secondary">
+              Current Safe
+            </Typography>
+            <EthHashInfo
+              address={pendingTx.safeAddress}
+              name={addressBook[pendingTx.safeAddress]}
+              isAddressBookName={Boolean(addressBook[pendingTx.safeAddress])}
```
iamacook, line 83: Nit: you use `addressBook[pendingTx.safeAddress]` a lot. What do you think about extracting it to a variable?

src/components/tx/SignOrExecuteForm/SignerForm/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+    if (nestedSafeAddress) {
+      return
+    }
+    if (!isOwner && nestedSafeOwners && nestedSafeOwners.length > 0) {
+      setNestedSafeAddress(nestedSafeOwners[0])
+    }
+    if (isOwner && (!nestedSafeOwners || nestedSafeOwners.length === 0)) {
+      setNestedSafeAddress(undefined)
+    }
+  }, [isOwner, nestedSafeAddress, nestedSafeOwners])
```
iamacook, line 62: What do you think about moving this to the context that sets the `signer`? Afaiu, this sets the state of the nested Safe address only on the `SignerForm`.

iamacook (approved):

liliya-soroka:
Verified

src/components/tx/SignOrExecuteForm/SignerForm/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const signerAddress = signer?.address
+  const { safe } = useSafeInfo()
+  const { safeTx } = useContext(SafeTxContext)
+
+  const onChange = (event: SelectChangeEvent<string>) => {
+    trackEvent(MODALS_EVENTS.CHANGE_SIGNER)
+    setSignerAddress?.(event.target.value)
+  }
+
+  const isNotNestedOwner = useMemo(() => nestedSafeOwners && nestedSafeOwners.length === 0, [nestedSafeOwners])
```
usame-algan, line 38: Nit: There is a hook `useIsNestedSafeOwner` that can be used here.

src/utils/signers.ts
```diff
@@ -0,0 +1,28 @@
...
+  nestedSafeOwners: string[] | null,
+  safe: SafeInfo,
+  tx: SafeTransaction | undefined,
+) => {
+  if (!wallet || !nestedSafeOwners || !tx) {
+    return []
+  }
+
+  const isDirectOwner =
+    typeof wallet?.address === 'string' && safe.owners.map((owner) => owner.value).includes(wallet.address)
```
usame-algan, line 16: Why do we need to check the type of `wallet.address` here? Can it be something other than string?

src/utils/signers.ts
```diff
@@ -0,0 +1,28 @@
+import type { ConnectedWallet } from '@/hooks/wallets/useOnboard'
+import type { SafeTransaction } from '@safe-global/safe-core-sdk-types'
+import type { SafeInfo } from '@safe-global/safe-gateway-typescript-sdk'
+
+export const getAvailableSigners = (
```
usame-algan, line 5: I would suggest testing this function

src/utils/signers.ts
```diff
@@ -0,0 +1,28 @@
...
+    return []
+  }
+
+  const isDirectOwner =
+    typeof wallet?.address === 'string' && safe.owners.map((owner) => owner.value).includes(wallet.address)
+  const isFullySigned = tx.signatures.size >= safe.threshold
+  const availableSigners = nestedSafeOwners ? [...nestedSafeOwners] : []
+  if (wallet?.address && isDirectOwner && !tx?.signatures.has(wallet.address.toLowerCase())) {
+    availableSigners?.push(wallet.address)
+  }
```
usame-algan, line 21: I think at this point `wallet.address` and `tx` should be defined so this should also work:

```
if (isDirectOwner && !tx.signatures.has(wallet.address.toLowerCase())) {
    availableSigners.push(wallet.address)
}
```
Nit: I would also add an empty line above the condition for better readability.

src/utils/signers.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  if (!isFullySigned) {
+    // Filter signers that already signed
+    return availableSigners.filter((signer) => !tx?.signatures.has(signer.toLowerCase()))
```
usame-algan, line 25:
```suggestion
    return availableSigners.filter((signer) => !tx.signatures.has(signer.toLowerCase()))
```

katspaugh (approved):

katspaugh:
@iamacook could you resolve the conflict and merge this please?

src/components/tx/SignOrExecuteForm/SignerForm/index.tsx
iamacook: Addressed in 56de0ed1f7cc052f870738d6184a6b7e77981ca1.

src/utils/signers.ts
iamacook: Addressed in 56de0ed1f7cc052f870738d6184a6b7e77981ca1.

src/utils/signers.ts
iamacook: Addressed in 56de0ed1f7cc052f870738d6184a6b7e77981ca1.

src/utils/signers.ts
iamacook: Addressed in 56de0ed1f7cc052f870738d6184a6b7e77981ca1.

src/utils/signers.ts
iamacook: Addressed in 56de0ed1f7cc052f870738d6184a6b7e77981ca1.

PR #4412, opened 2024-10-23
Feat(Sidebar): Add pinned safes list to sidebar and accounts page [SW-304]

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Create-pinned-accounts-list-in-sidebar-and-accounts-page-11f8180fe573806da36dc2327ce063ad?pvs=4

## How this PR fixes it
Replaces the existing safe lists in the sidebar with `Accounts`, containing all owned and read only accounts, and `Pinned` containing safes that have been manually added to the list.
- Adds a pinned flag to the added safes slice
- When you pin a safe, it adds it to the added safes list (if not already added) and sets this flag to true.
- Pinned safes appear in both the pinned list and also in the full list below.

## How to test it
On the accounts page or the sidebar:
- Make sure you can pin and unpin safes (single chain and multichain should behave the same)
- Pinned safes should persist, even when connecting with a different wallet, or in the disconnected state.
- Pinned safes should be duplicated in both lists. 

## Screenshots
Sidebar:
image

Accounts page:
image

Mobile:
image

## Checklist

### DIFF
--- a/src/components/welcome/MyAccounts/AccountItem.tsx
+++ b/src/components/welcome/MyAccounts/AccountItem.tsx
@@ -2,13 +2,13 @@ import { LoopIcon } from '@/features/counterfactual/CounterfactualStatusButton'
 import { selectUndeployedSafe } from '@/features/counterfactual/store/undeployedSafesSlice'
 import type { SafeOverview } from '@safe-global/safe-gateway-typescript-sdk'
 import { useMemo } from 'react'
-import { ListItemButton, Box, Typography, Chip, Skeleton } from '@mui/material'
+import { ListItemButton, Box, Typography, Chip, Skeleton, Tooltip, IconButton, SvgIcon } from '@mui/material'
 import Link from 'next/link'
 import SafeIcon from '@/components/common/SafeIcon'
 import Track from '@/components/common/Track'
 import { OVERVIEW_EVENTS, OVERVIEW_LABELS } from '@/services/analytics'
 import { AppRoutes } from '@/config/routes'
-import { useAppSelector } from '@/store'
+import { useAppDispatch, useAppSelector } from '@/store'
 import { selectChainById } from '@/store/chainsSlice'
 import ChainIndicator from '@/components/common/ChainIndicator'
 import css from './styles.module.css'
@@ -30,6 +30,10 @@ import { useGetSafeOverviewQuery } from '@/store/api/gateway'
 import useWallet from '@/hooks/wallets/useWallet'
 import { skipToken } from '@reduxjs/toolkit/query'
 import { hasMultiChainAddNetworkFeature } from '@/features/multichain/utils/utils'
+import BookmarkIcon from '@/public/images/apps/bookmark.svg'
+import BookmarkedIcon from '@/public/images/apps/bookmarked.svg'
+import { addOrUpdateSafe, pinSafe, selectAddedSafes, unpinSafe } from '@/store/addedSafesSlice'
+import { defaultSafeInfo } from '@/store/safeInfoSlice'
 type AccountItemProps = {
   safeItem: SafeItem
   safeOverview?: SafeOverview
@@ -37,7 +41,7 @@ type AccountItemProps = {
 }
 
 const AccountItem = ({ onLinkClick, safeItem }: AccountItemProps) => {
-  const { chainId, address } = safeItem
+  const { chainId, address, isPinned } = safeItem
   const chain = useAppSelector((state) => selectChainById(state, chainId))
   const undeployedSafe = useAppSelector((state) => selectUndeployedSafe(state, chainId, address))
   const safeAddress = useSafeAddress()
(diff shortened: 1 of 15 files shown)


src/store/addedSafesSlice.ts
```diff
@@ -49,10 +51,35 @@ export const addedSafesSlice = createSlice({
...
+      const { chainId, address } = payload
+      const removeOnUnpin = state[chainId]?.[address]?.removeOnUnpin
+
+      if (removeOnUnpin) {
+        delete state[chainId]?.[address]
+        if (Object.keys(state[chainId]).length === 0) {
+          delete state[chainId]
+        }
+      } else {
+        state[chainId][address] = {
```
usame-algan, line 73: This crashes for me sometimes when unpinning a multichain safe because `state[chainId]` is undefined.

src/components/welcome/MyAccounts/QueueActions.tsx
```diff
@@ -44,27 +44,25 @@ const QueueActions = ({
   }
 
   return (
-    <Box width="100%">
```
usame-algan, line 47: I think we need the Box as a wrapper otherwise the Chips are not below the address:
Screenshot 2024-10-29 at 10 04 12

src/components/welcome/MyAccounts/index.tsx
```diff
@@ -51,6 +61,13 @@ const AccountsList = ({ safes, onLinkClick }: AccountsListProps) => {
     ],
     [safes, watchlistMultiChainSafes],
   )
+  const pinnedSafes = useMemo<(MultiChainSafeItem | SafeItem)[]>(
+    () => [
+      ...(safes.allSingleSafes?.filter(({ isPinned }) => isPinned) ?? []),
+      ...(safes.allMultiChainSafes?.filter(({ isPinned }) => isPinned) ?? []),
+    ],
+    [safes],
+  )
```
usame-algan, line 70: I am getting a flicker in the accounts list when loading the app. It looks like `pinnedSafes` are there but then for a few render cycles it is an empty array again and then switches back to the correct result. Is there a way we can pass in a loading flag from the `useAllSafes` hook and wait until everything is ready before rendering the list?

src/components/welcome/MyAccounts/index.tsx
```diff
@@ -59,60 +76,89 @@ const AccountsList = ({ safes, onLinkClick }: AccountsListProps) => {
...
+              />
+              <Typography variant="h5" fontWeight={700} mb={2}>
+                Pinned
+              </Typography>
+            </div>
+            {pinnedSafes.length > 0 ? (
+              <SafesList safes={pinnedSafes} onLinkClick={onLinkClick} />
+            ) : (
+              <Box className={css.noPinnedSafesMessage}>
+                <Typography color="text.secondary" maxWidth="350px" textAlign="center" fontSize="14px">
```
usame-algan, line 123: nit: You can also use `variant="body2"` to achieve 14px font-size instead of setting it separately.

src/components/welcome/MyAccounts/index.tsx
```diff
@@ -59,60 +76,89 @@ const AccountsList = ({ safes, onLinkClick }: AccountsListProps) => {
...
+                  {allSafes && allSafes.length > 0 && (
+                    <Typography component="span" color="text.secondary" fontSize="inherit" fontWeight="normal" mr={1}>
+                      {' '}
+                      ({allSafes.length})
+                    </Typography>
+                  )}
+                </Typography>
+              </div>
+            </AccordionSummary>
+            <AccordionDetails sx={{ padding: 0 }}>
```
usame-algan, line 155: As an idea should we keep the accounts list open by default if there are no pinned safes?

jmealy: good idea!

src/components/welcome/MyAccounts/index.tsx
```diff
@@ -135,7 +135,7 @@ const AccountsList = ({ safes, onLinkClick, isSidebar = false }: AccountsListPro
           </Box>
 
           {/* All Accounts */}
-          <Accordion sx={{ border: 'none' }}>
+          <Accordion expanded={pinnedSafes.length === 0} sx={{ border: 'none' }}>
```
usame-algan, line 138: When the user pins their first safe it closes now. We can use `defaultExpanded` to keep it open in that case

src/components/welcome/MyAccounts/AccountItem.tsx
```diff
@@ -90,8 +130,8 @@ const AccountItem = ({ onLinkClick, safeItem }: AccountItemProps) => {
           <Box pr={2.5}>
             <SafeIcon
               address={address}
-              owners={safeOverview?.owners.length ?? counterfactualSetup?.owners.length}
-              threshold={safeOverview?.threshold ?? counterfactualSetup?.threshold}
+              owners={safeOwners.length ?? undefined}
```
usame-algan, line 133: `safeOwners` is always an array now so `safeOwners.length` could resolve to 0 and fail the check. We should instead check `safeOwners.length > 0 ? safeOwners.length : undefined` in order to keep the same behaviour.

src/components/welcome/MyAccounts/AccountItem.tsx
```diff
@@ -90,8 +130,8 @@ const AccountItem = ({ onLinkClick, safeItem }: AccountItemProps) => {
...
+              threshold={safeThreshold ?? undefined}
```
usame-algan, line 134: Same as above I think. The default value is 0 which would pass the `??` check.

usame-algan (approved):
Looks good!

PR #4419, opened 2024-10-24
Add Safenet Check Summary to Signature Form

nlordell:
## What it solves

This PR adds the Safenet checks to the transaction execution flow in the wallet interface.

## How this PR fixes it

Currently, the checks are shown for any Safe that has Safenet enabled (i.e. has the SafeNet guard set on a supported chain). The checks are done using the Safenet service backend.

Results are grouped into categories which allows the UI to show high level summaries based on a collection of guarantees as well as grouping any unknown guarantees into an "other" category.

Note that I implemented `SafeTx` hash computation as I couldn't find a utility method that did it for me anywhere - which was surprising. If there is an existing utility method for this - please let me know and I will change my code to use it.

## How to test it

:shrug: 

Also, if you have "what the hell is this code" reaction - please forgive me, I am not a React developer (but I did make sure that the POST to the simulate endpoint only happens once!).

## Screenshots

image

## Checklist


src/components/tx/SignOrExecuteForm/SafeNetTxChecks.tsx
```diff
@@ -0,0 +1,29 @@
...
+import useIsSafeNetEnabled from '@/hooks/useIsSafeNetEnabled'
+import { Typography } from '@mui/material'
+import useChainId from '@/hooks/useChainId'
+import useSafeAddress from '@/hooks/useSafeAddress'
+
+const SafeNetTxChecks = (): ReactElement | null => {
+  const safe = useSafeAddress()
+  const chainId = useChainId()
+  const { safeTx } = useContext(SafeTxContext)
+  const isSafeNetEnabled = useIsSafeNetEnabled()
```
mmv08, line 14: I am not a big fan of using hooks to retrieve the data accessible in the parent component. It gets rather hard to follow where the data is coming from. Can we pass them as a prop?

mmv08: Ah, I see what's up. You followed the example of an existing component. Nevermind then.

src/components/tx/security/safenet/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+      ...groups,
+      [display]: (groups[display] ?? true) && success,
+    }
+  }, {} as Record<string, boolean>)
+  return Object.entries(groups)
+    .map(([display, success]) => ({ display, success }))
+    .sort((a, b) => a.display.localeCompare(b.display))
+}
+
+function _getSafeTxHash({ safe, chainId, safeTx }: Required<SafeNetTxSimulationProps>): string {
```
mmv08, line 45: I'm sure there should be an existing function for this somewhere.

src/components/tx/security/safenet/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+    if (!safeTx || !dataDecoded) {
+      return
+    }
+
+    const safeTxHash = _getSafeTxHash({ safe, chainId, safeTx })
+    simulate({
+      chainId,
+      tx: {
+        safe,
+        safeTxHash,
```
mmv08, line 116: Can't the backend compute it? Feel like it's not required for simulation

nlordell: I checked and _technically_ it isn't required for the simulation (I passed in a fake hash and it worked :sweat_smile:). One reason I like it is that it gives a caching tag for the Redux toolkit `createApi`.

src/components/tx/SignOrExecuteForm/SafeNetTxChecks.tsx
nlordell: I made the inner component `SafeNetTxSimulation` work on passed in props instead of reading state with hooks for this reason.

PR #4424, opened 2024-10-25
test: add unit test for NetworkMultiSelector

schmanu:
Adds a test for the `NetworkMultiSelector` component as it contains a lot of logic and had some issues in the past.


jmealy (approved):
nice!

PR #4429, opened 2024-10-25
fix: Disable add proposer and delete proposer [SW-400] [SW-396]

usame-algan:
## What it solves

Resolves [SW-400](https://www.notion.so/safe-global/As-a-proposer-I-can-go-throught-the-add-proposer-12a8180fe57380b4810cec062aaaa98c)
Resolves [SW-396](https://www.notion.so/safe-global/Only-owner-who-created-delegate-can-delete-them-12a8180fe57380ca84e9c8fb60e35004)

## How this PR fixes it

- Disables the `Add proposer` button for proposers
- Only enables the Delete proposer button if any of these conditions meet:
- - Connected wallet is the proposer itself
- - Connected wallet is the owner who created the proposer

## How to test it

1. Open a safe with proposers
2. Connect with one of the proposers
3. Observe you can't add new proposers
4. Observe you can only delete yourself
5. Connect with an owner that didn't create any proposers
6. Observe you can't delete them
7. Connect with an owner that created a proposer
8. Observe you can delete them

## Checklist

### DIFF
--- a/src/components/common/CheckWallet/index.tsx
+++ b/src/components/common/CheckWallet/index.tsx
@@ -57,7 +57,7 @@ const CheckWallet = ({
       return Message.NotSafeOwner
     }
 
-    if (!allowProposer && isProposer) {
+    if (!allowProposer && isProposer && !isSafeOwner) {
       return Message.NotSafeOwner
     }
   }, [
--- a/src/components/common/CheckWallet/index.tsx
+++ b/src/components/common/CheckWallet/index.tsx
@@ -15,6 +15,7 @@ type CheckWalletProps = {
   noTooltip?: boolean
   checkNetwork?: boolean
   allowUndeployedSafe?: boolean
+  allowProposer?: boolean
 }
 
 enum Message {
@@ -30,6 +31,7 @@ const CheckWallet = ({
   noTooltip,
   checkNetwork = false,
   allowUndeployedSafe = false,
+  allowProposer = true,
 }: CheckWalletProps): ReactElement => {
   const wallet = useWallet()
   const isSafeOwner = useIsSafeOwner()
@@ -46,6 +48,7 @@ const CheckWallet = ({
     if (!wallet) {
       return Message.WalletNotConnected
     }
+
     if (isUndeployedSafe && !allowUndeployedSafe) {
       return Message.SafeNotActivated
     }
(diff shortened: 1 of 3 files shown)


src/components/common/CheckWallet/index.tsx
```diff
@@ -30,6 +31,7 @@ const CheckWallet = ({
   noTooltip,
   checkNetwork = false,
   allowUndeployedSafe = false,
+  allowProposer = true,
```
usame-algan, line 34: By default, proposers should have all the same permissions as owners

schmanu, line 34: Just an observation: All other `allowX` props default to `false` when not set.

I understand that it makes more sense but we need to be aware of this :D

src/components/common/CheckWallet/index.tsx
```diff
@@ -46,15 +48,21 @@ const CheckWallet = ({
...
+
     if (isUndeployedSafe && !allowUndeployedSafe) {
       return Message.SafeNotActivated
     }
 
     if (!allowNonOwner && !isSafeOwner && !isProposer && (!isOnlySpendingLimit || !allowSpendingLimit)) {
       return Message.NotSafeOwner
     }
+
+    if (!allowProposer && isProposer) {
```
schmanu, line 60: Is it possible to be proposer and owner at the same time?
e.g. by adding an address as proposer and then later on adding the same address as owner?

This could lead to false warnings in that case when `allowProposer` is `false` and `isProposer` and `isOwner` is `true`.

src/components/settings/ProposersList/index.tsx
```diff
@@ -51,7 +53,9 @@ const ProposersList = () => {
                         onClick={() => onDelete(proposer.delegate)}
                         color="error"
                         size="small"
-                        disabled={!isOk}
+                        disabled={
+                          !isOk || (wallet?.address !== proposer.delegate && wallet?.address !== proposer.delegator)
```
schmanu, line 57:
```suggestion
                          !isOk || wallet?.address !== proposer.delegator
```

If I understand correctly, we could rewrite like this if we add `allowProposer={false}` to `CheckWallet`? Or can proposers remove themselves?

src/components/common/CheckWallet/index.tsx
usame-algan: Good point! Technically, it is possible to be owner and proposer although we added validation to prevent adding owners as proposers. Still someone could do this via other means so lets extend the condition to `!allowProposer && isProposer && !isSafeOwner`

src/components/settings/ProposersList/index.tsx
usame-algan: Yes, the API allows proposers to remove themselves so I left that possiblity in the UI too

schmanu (approved):

PR #4437, opened 2024-10-28
[ SW-501 ] fix: Update privacy policy

usame-algan:
## What it solves

Updating the privacy policy according to [this request](https://5afe.slack.com/archives/C045P36MXQC/p1730108275091469)

## How this PR fixes it

- Adds markdown support for privacy policy

## How to test it

1. Open the privacy page
2. Observe the updates from the document

## Checklist


src/pages/privacy.tsx
```diff
@@ -1,1004 +1,30 @@
...
+  React.PropsWithChildren<Omit<MUILinkProps, 'href'> & Pick<NextLinkProps, 'href' | 'as'>>
+> = ({ href = '', as, children, ...other }) => {
+  const isExternal = href.toString().startsWith('http')
+  return (
+    <NextLink href={href} as={as} passHref legacyBehavior>
+      <MUILink target={isExternal ? '_blank' : ''} rel="noreferrer" {...other}>
+        {children}
+      </MUILink>
+    </NextLink>
+  )
```
compojoom, line 21: How about we move this to util function as we have the same function in the terms.tsx page?

src/markdown/privacy/privacy.md.d.ts
```diff
@@ -0,0 +1,5 @@
+export { default } from '*.md'
+export const metadata = {
+  version: string,
+  last_update_date: string,
+}
```
compojoom, line 5: You are specifying a frontmatter with this, but your md file doesn't have one. Look at terms.md

```
---
version: 1.2
last_update_date: September, 2024
---
```

usame-algan: Afaik we are not using those anywhere in the app like we do for terms so I will remove the `.d.ts` file

src/pages/privacy.tsx
usame-algan: done

compojoom (approved):
👍

liliya-soroka:
The formatting is broken in a new version. Please check why it looks so different and sometimes with the wrong numbering

usame-algan:
> The formatting is broken in a new version. Please check why it looks so different and sometimes with the wrong numbering

Updated the document. The numbering should be fixed. The styles might look a bit different than before but it is more consistent with the styling of the terms page now.

francovenica:
I was able to see that the changes are there, but I don't have access to the doc, so I cannot confirm everythign is correct.

I requested access and I'll wait to have it

liliya-soroka:
verified by the product - mhttps://www.notion.so/safe-global/fix-Update-privacy-policy-1378180fe573801a9df1d33eb19d6e4c?pvs=4

PR #4447, opened 2024-10-30
fix: Poll proposers when on the settings page [SW-427]

usame-algan:
## What it solves

Resolves SW-427

## How this PR fixes it

- Adds polling every 15 seconds for the get proposers call

## How to test it

1. Open the settings page in two different tabs
2. Create a proposer
3. Switch to the other tab and observe the proposer appearing within 15 seconds

## Checklist


usame-algan:
Changed this to draft until we make a decision if we want to move forward with this change as it would significantly increase the number of requests to the CGW and tx-service for every safe and every wallet

usame-algan:
Closing this for now until we have a CGW cache in place: https://www.notion.so/safe-global/Cache-Proposers-data-in-the-CGW-1308180fe57380c2a524f3e50c68ef23

PR #4449, opened 2024-10-30
Feat: decode execTransaction SafeTransactions

schmanu:
## What this PR changes
Properly decodes the child transaction when a parent Safe is calling execTransaction

## How to test it
- execute a signed Safe Transaction with another Safe

## Screenshots
_tbd_

## Checklist


src/components/transactions/TxDetails/TxData/ExecTransaction/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  return {
+    addSignature: () => {},
+    encodedSignatures: () => params[9],
+    getSignature: () => undefined,
+    data: {
+      to: params[0],
+      value: params[1],
+      data: params[2],
+      operation: params[3] as number,
```
iamacook, line 33: We don't need to cast this as the type is inherently `any`. (I'm quite surprised it's missing the types.)

src/components/transactions/TxDetails/TxData/ExecTransaction/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+      safeTxGas: params[4],
+      baseGas: params[5],
+      gasPrice: params[6],
+      gasToken: params[7],
+      refundReceiver: params[8],
+      nonce: 0,
```
iamacook, line 39: Nit: what do you think about setting this to `-1` to emphasise we don't know it?

src/components/transactions/TxDetails/TxData/ExecTransaction/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { AppRoutes } from '@/config/routes'
+import { useMemo } from 'react'
+import type { SafeTransaction } from '@safe-global/safe-core-sdk-types'
+import useAsync from '@/hooks/useAsync'
+import NestedTransactionIcon from '@/public/images/transactions/nestedTx.svg'
+import ExternalLink from '@/components/common/ExternalLink'
+
+const safeInterface = Safe__factory.createInterface()
+
+const extractTransactionData = (data: string | undefined): SafeTransaction | undefined => {
```
iamacook, line 19: The conditional `data` is specific to the component logic. I'd suggest moving the check for `undefined` inside the relevant `useMemo` instead.

src/components/transactions/TxDetails/TxData/ExecTransaction/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export const ExecTransaction = ({ data }: { data?: TransactionData }) => {
+  const chain = useCurrentChain()
+
+  const childSafeTx = useMemo<SafeTransaction | undefined>(() => extractTransactionData(data?.hexData), [data?.hexData])
+
+  const [dataDecoded, error] = useAsync(async () => {
+    if (chain?.chainId && data?.to.value && childSafeTx) {
+      return await getConfirmationView(
```
iamacook, line 52: Can we try getting this from the preview endpoint instead? We want to eventually [deprecate the confirmation view](https://github.com/safe-global/safe-wallet-web/issues/4331).

src/components/transactions/TxDetails/TxData/ExecTransaction/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+        chain.chainId,
+        data.to.value,
+        childSafeTx.data.data,
+        childSafeTx.data.to,
+        childSafeTx.data.value.toString(),
+      )
+    }
+  }, [chain?.chainId, data?.to.value, childSafeTx])
+
+  const decodedDataBlock = dataDecoded ? (
```
iamacook, line 62: Nit: I was confused between `data?.dataDecoded` and `dataDecoded`. Can we rename this be more explicit?

src/components/transactions/TxDetails/TxData/ExecTransaction/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+          </>
+        ) : error ? (
+          <ErrorMessage>Could not load details on executed transaction.</ErrorMessage>
+        ) : (
+          <Skeleton />
+        )}
+      </Stack>
+    </Stack>
+  )
+}
```
iamacook, line 106: There's a lot of similarity with the [`OnChainConfirmation` component](https://github.com/safe-global/safe-wallet-web/blob/c8b4f51089f77619413b8c01b339fb40df874d62/src/components/transactions/TxDetails/TxData/OnChainConfirmation/index.tsx). We could create a wrapper of sorts that you pass the transaction-to-display and reuse it for both.

src/components/transactions/TxDetails/TxData/index.tsx
```diff
@@ -82,6 +83,10 @@ const TxData = ({
     return <OnChainConfirmation data={txDetails.txData} />
   }
 
+  if (isExecTxData(txDetails.txData)) {
+    return <ExecTransaction data={txDetails.txData} />
+  }
```
iamacook, line 88: Have you considered how this will affect all other `execTransaction` details?

src/hooks/useTransactionType.tsx
```diff
@@ -131,7 +132,7 @@ export const getTransactionType = (tx: TransactionSummary, addressBook: AddressB
         }
       }
 
-      if (isOnChainConfirmationTxInfo(tx.txInfo)) {
+      if (isOnChainConfirmationTxInfo(tx.txInfo) || isExecTxInfo(tx.txInfo)) {
```
iamacook, line 135: As before re. the influence on existing `execTransaction` calls. Won't this try to decode "standard" transactions as well?

src/utils/transaction-guards.ts
```diff
@@ -428,3 +428,18 @@ export const isOnChainConfirmationTxInfo = (info: TransactionInfo): info is Cust
   }
   return false
 }
+
+/**
+ * True if the tx calls `execTransaction`
+ */
+export const isExecTxData = (data?: TransactionData): boolean => {
+  const execTransactionSelector = Safe__factory.createInterface().getFunction('execTransaction').selector
```
iamacook, line 436: Nit: we can move this outside the helper.

src/hooks/useTransactionType.tsx
schmanu: Which others? Sadly we lack detailed data in the TxInfo type. Ideally we would also get the method selector which is less ambiguous as it also hashes the parameter types.

But I am not aware of other execTransaction transactions.

src/components/transactions/TxDetails/TxData/index.tsx
schmanu: Which other details?

iamacook: As per discussion: we always display the transaction executed so the `execTransaction` here is of the nested Safe.

src/hooks/useTransactionType.tsx
iamacook: As per discussion: we always display the transaction executed so the `execTransaction` here is of the nested Safe.

src/components/transactions/TxDetails/TxData/ExecTransaction/index.tsx
schmanu: I tried to change it quickly but the returned types are not fully compatible. So we need to adjust the component that renders the result.

Lets do it in a separate tech debt issue to reduce the scope.

PR #4454, opened 2024-10-31
Recipient signature verification

gjeanmart:
## What it solves

Resolves [#70](https://github.com/safe-global/safenet/issues/70)

## Checklist

### DIFF
--- a/src/components/transactions/TxDetails/Summary/index.tsx
+++ b/src/components/transactions/TxDetails/Summary/index.tsx
@@ -13,7 +13,7 @@ import DecodedData from '../TxData/DecodedData'
 import { calculateSafeTransactionHash } from '@safe-global/protocol-kit/dist/src/utils'
 import useSafeInfo from '@/hooks/useSafeInfo'
 import { SafeTxHashDataRow } from './SafeTxHashDataRow'
-import SafenetTxChecks from '@/components/tx/SignOrExecuteForm/SafenetTxChecks'
+import { SafenetTxSimulation } from '@/components/tx/security/safenet'
 
 interface Props {
   txDetails: TransactionDetails
@@ -74,15 +74,21 @@ const Summary = ({ txDetails, defaultExpanded = false, hideDecodedData = false }
         </TxDataRow>
       )}
 
-      <SafenetTxChecks
-        safeTx={{
-          data: safeTxData!,
-          signatures: new Map(),
-          getSignature: () => undefined,
-          addSignature: () => {},
-          encodedSignatures: () => '',
-        }}
-      />
+      <Box mt={1}>
+        <TxDataRow title="Safenet Simulation:">
+          <SafenetTxSimulation
+            safe={safe.address.value}
+            chainId={safe.chainId}
+            safeTx={{
+              data: safeTxData!,
+              signatures: new Map(),
+              getSignature: () => undefined,
+              addSignature: () => {},
+              encodedSignatures: () => '',
+            }}
+          />
+        </TxDataRow>
+      </Box>
 
       {/* Advanced TxData */}
       {txData && (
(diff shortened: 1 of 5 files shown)


src/components/tx/SignOrExecuteForm/SafenetTxChecks.tsx
```diff
@@ -1,16 +1,15 @@
...
 import TxCard from '@/components/tx-flow/common/TxCard'
-import { SafeTxContext } from '@/components/tx-flow/SafeTxProvider'
 import useIsSafenetEnabled from '@/hooks/useIsSafenetEnabled'
 import { Typography } from '@mui/material'
 import useChainId from '@/hooks/useChainId'
 import useSafeAddress from '@/hooks/useSafeAddress'
+import type { SafeTransaction } from '@safe-global/safe-core-sdk-types'
 
-const SafenetTxChecks = (): ReactElement | null => {
+const SafenetTxChecks = ({ safeTx }: { safeTx?: SafeTransaction }): ReactElement | null => {
```
mmv08, line 10: why is the `safeTx`param optional? can it execute any checks if safeTx is undefined ?

mmv08 (approved):
Looks good, only one minor issue

src/components/transactions/TxDetails/Summary/index.tsx
```diff
@@ -73,6 +74,16 @@ const Summary = ({ txDetails, defaultExpanded = false, hideDecodedData = false }
         </TxDataRow>
       )}
 
+      <SafenetTxChecks
+        safeTx={{
+          data: safeTxData!,
```
nlordell, line 79: Maybe, we can change `SafenetTxChecksProps` to use `Pick<SafeTransaction, "data">` so that only the `data` field is required (since that is all we use anyway).

src/components/transactions/TxDetails/Summary/index.tsx
```diff
@@ -73,6 +74,16 @@ const Summary = ({ txDetails, defaultExpanded = false, hideDecodedData = false }
         </TxDataRow>
       )}
 
+      <SafenetTxChecks
```
nlordell, line 77: From what I understand, components in `src/components/tx/SignOrExecuteForm/...` are for the signing and executing form itself.

I think it would make more sense to not use the `SafenetTxChecks` component here (it uses things like `TxCard` which are MUI cards for the sign-or-execute form), and instead just embed the `SafenetTxSimulation` component (which is where all the logic for displaying the checks is anyway). This way, it can be laid out differently in the transaction summary to better match the UI there.

src/components/tx/SignOrExecuteForm/SafenetTxChecks.tsx
gjeanmart: I kept the same data definition as before 
image

src/components/transactions/TxDetails/Summary/index.tsx
gjeanmart: It will likely not work without going deeper in existing hooks and changing `useDecodeTx`

src/components/transactions/TxDetails/Summary/index.tsx
gjeanmart: Agree, I'll make a couple of change to reflect this

src/components/tx/SignOrExecuteForm/SafenetTxChecks.tsx
mmv08: But this is the context definition, it can be empty or contain a transaction. I don't think they're related

gjeanmart: yeah you're right, let's not make it optional

PR #4456, opened 2024-10-31
feat: Subaccounts

iamacook:
## What it solves

Implementation of creating (and funding) Subaccounts

## How this PR fixes it

A new sidebar icon, context menu (accessible via the icon/drawer) and settings have been added that both list and open a Subaccount creation flow.

## How to test it

Open a Safe and observe:

- A new header icon in the sidebar that, when clicked, opens either a promo message (if no Subaccounts are owned) or a list, with an option to add more. (The presence of Subaccounts is signified by a badge number.)
- If the **current** Safe has Subaccounts, a new item in the sidebar drawer menu (accessible via the three dots) should open the same component.
- A new settings area under "Setup" has the option to add Subaccounts, and list/rename those which exist already.

## Screenshots

image

image

image

image

image

image

image

image

## Checklist


schmanu (commented):
Looking and working great 🔥 
I added some comments regarding multichain compatibility

Feature wise:
I think we could already add a new page on e.g. route `/subaccounts ` which fetches and lists all SubAccounts / owned Safes of the currently opened Safe and simply offers links to open those. 
We could also include the Create button there.

iamacook:
> Looking and working great 🔥 I added some comments regarding multichain compatibility

Thank you for the thorough review (I agree with everything). Since this is a PoC and subject to change, I’ll leave your comments as unresolved for the time being and reference them when we plan this out properly.

src/components/settings/SubaccountsList/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+            </Typography>
+          </Grid>
+
+          <Grid item xs>
+            <Typography sx={{ mb: 3 }}>
+              Subaccounts are separate wallets owned by your main Account, perfect for organizing different funds and
+              projects.{' '}
+              <ExternalLink
+                // TODO: Add link
+                href="#"
```
iamacook, line 83: Marking this so we don't forget.

iamacook:
I am going to close this as we intend to approach it differently. However, it will be used as reference.

PR #4465, opened 2024-11-04
[ SW-467 ] fix(threshold-info): Display right info in the change threshold confirmation screen

clovisdasilvaneto:
## What it solves
Confirm threshold screen is showing 0 of (n) when user comes from the transaction list.

## How this PR fixes it
Before we were grabbing the threshold information from the changeThreshold screen flow, but this information is not available when user comes from the transactions list (for eg, in order to confirm a `change threshold` transaction)

## How to test it
- Change the threshold of your account
- sign but do not execute it
- go to the transaction list and try to confirm or execute it from there.
- The amount of required signers now should be working as expected instead of showing "0 out (n)"

## Screenshots
Screenshot 2024-11-04 at 12 21 11

## Checklist

### DIFF
--- a/src/components/tx/confirmation-views/ChangeThreshold/index.tsx
+++ b/src/components/tx/confirmation-views/ChangeThreshold/index.tsx
@@ -5,8 +5,8 @@ import commonCss from '@/components/tx-flow/common/styles.module.css'
 import useSafeInfo from '@/hooks/useSafeInfo'
 import { ChangeThresholdReviewContext } from '@/components/tx-flow/flows/ChangeThreshold/context'
 import { ChangeSignerSetupWarning } from '@/features/multichain/components/SignerSetupWarning/ChangeSignerSetupWarning'
-import { SettingsInfoType, type TransactionDetails } from '@safe-global/safe-gateway-typescript-sdk'
-import { isSettingsChangeTxInfo } from '@/utils/transaction-guards'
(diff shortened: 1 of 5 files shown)


src/components/tx/confirmation-views/ChangeThreshold/index.tsx
```diff
@@ -5,10 +5,21 @@ import commonCss from '@/components/tx-flow/common/styles.module.css'
...
+}
+
+function ChangeThreshold({ txDetails }: ChangeThresholdProps) {
   const { safe } = useSafeInfo()
   const { newThreshold } = useContext(ChangeThresholdReviewContext)
+  const threshold =
+    txDetails &&
+    isSettingsChangeTxInfo(txDetails.txInfo) &&
+    txDetails.txInfo.settingsInfo?.type === SettingsInfoType.CHANGE_THRESHOLD &&
+    txDetails.txInfo.settingsInfo.threshold
```
usame-algan, line 22: Can we use the `isChangeThresholdView` guard instead?

usame-algan (approved):

src/components/tx/confirmation-views/ChangeThreshold/index.tsx
clovisdasilvaneto: Sure, didn't know we have this one ehehe

francovenica:
Looks good to me.

A before/after screenshots:
image
image

PR #4471, opened 2024-11-04
feat: Display creator in the proposer list [SW-408] [SW-470]

usame-algan:
## What it solves

Resolves SW-408
Resolves SW-470

## How this PR fixes it

- Tightens validation when doing optimistic updates to the proposer list
- Displays the creator address in the proposer list

## How to test it

1. Open the settings and create a proposer
2. Observe the creator address is shown next to them
3. Add the same proposer twice with 2 different owners
4. Delete one of them
5. Observe the other proposer is still visible in the list
6. Add the same proposer twice with 2 different owners
7. Update the name of one of them
8. Observe the correct entry is updated

## Screenshots
Screenshot 2024-11-04 at 17 58 48

## Checklist

### DIFF
--- a/src/store/api/gateway/proposers.ts
+++ b/src/store/api/gateway/proposers.ts
@@ -15,18 +15,29 @@ export const proposerEndpoints = (
   }),
   deleteProposer: builder.mutation<
     void,
-    { chainId: string; safeAddress: string; delegateAddress: string; signature: string }
+    {
+      chainId: string
+      safeAddress: string
+      delegateAddress: string
+      delegator: string
+      signature: string
+    }
   >({
-    queryFn({ chainId, safeAddress, delegateAddress, signature }) {
+    queryFn({ chainId, safeAddress, delegateAddress, delegator, signature }) {
       return buildQueryFn(() =>
-        deleteDelegateV2({ params: { path: { chainId, delegateAddress } }, body: { safe: safeAddress, signature } }),
+        deleteDelegateV2({
+          params: { path: { chainId, delegateAddress } },
+          body: { safe: safeAddress, signature, delegator },
+        }),
       )
     },
     // Optimistically update the cache and roll back in case the mutation fails
-    async onQueryStarted({ chainId, safeAddress, delegateAddress }, { dispatch, queryFulfilled }) {
+    async onQueryStarted({ chainId, safeAddress, delegateAddress, delegator }, { dispatch, queryFulfilled }) {
       const patchResult = dispatch(
         gatewayApi.util.updateQueryData('getProposers', { chainId, safeAddress }, (draft) => {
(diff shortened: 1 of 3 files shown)


src/store/api/gateway/proposers.ts
```diff
@@ -15,18 +15,29 @@ export const proposerEndpoints = (
...
-        deleteDelegateV2({ params: { path: { chainId, delegateAddress } }, body: { safe: safeAddress, signature } }),
+        deleteDelegateV2({
+          params: { path: { chainId, delegateAddress } },
+          body: { safe: safeAddress, signature, delegator },
+        }),
       )
     },
     // Optimistically update the cache and roll back in case the mutation fails
-    async onQueryStarted({ chainId, safeAddress, delegateAddress }, { dispatch, queryFulfilled }) {
+    async onQueryStarted({ chainId, safeAddress, delegateAddress, delegator }, { dispatch, queryFulfilled }) {
```
compojoom, line 35: This is cool. Just wondering - can we use cache tags here as well and once the query is complete also invalidate the cache for the getDelegates query?

src/store/api/gateway/proposers.ts
```diff
@@ -52,7 +70,11 @@ export const proposerEndpoints = (
     async onQueryStarted({ chainId, safeAddress, delegate, delegator, label }, { dispatch, queryFulfilled }) {
       const patchResult = dispatch(
         gatewayApi.util.updateQueryData('getProposers', { chainId, safeAddress }, (draft) => {
-          const existingProposer = draft.results.findIndex((proposer: Delegate) => proposer.delegate === delegate)
+          const existingProposer = draft.results.findIndex(
+            (proposer: Delegate) => proposer.delegate === delegate && delegator === proposer.delegator,
+          )
+
+          console.log(existingProposer)
```
compojoom, line 77: I guess you forgot this :)

src/store/api/gateway/proposers.ts
usame-algan: I think it is possible to use tags here as well to do a re-fetch whenever a proposer is added or deleted. However, to avoid too many requests until our [backend is optimized](https://www.notion.so/safe-global/Cache-Proposers-data-in-the-CGW-1308180fe57380c2a524f3e50c68ef23) I followed this guide of optimistically updating the cache: https://redux-toolkit.js.org/rtk-query/usage/manual-cache-updates#optimistic-updates

jmealy (commented):
There is a couple of bugs here:

- ~~If you are on the wrong chain when you open the add proposer modal, the network switch button does not work properly. It shows the wallet connect modal instead of prompting you to change network~~

- Again if you're on the wrong network and you try to delete a proposer, it gives an error

- related to the above, if you get an error on the delete proposer dialog, then close it and reopen it, the previous error is still shown.

jmealy (approved):
Looks good 👍

PR #4481, opened 2024-11-05
Slim Down Docker Image (Attempt 2)

nlordell:
## What it solves

This is a long awaited follow up to #3707, which reduces the Docker image size of the web interface from ~6GB to ~70MB. This iteration addresses build issues that it introduced in the previous version.

In particular, we now use a multi-stage Docker build, where building the web assets is done on the `BUILDPLATFORM` (i.e. the native build platform) instead of the target platform. This means that the image for `linux/arm64` will first build the front-end on the native architecture of the CI runner (`linux/amd64` by default) before copying. This should make building the Docker image MUCH quicker, while still producing a multi-platform image, and avoid the Docker build timeouts that were happening in CI. Documentation on how `BUILDPLATFORM` works and can be used for cross-compilation can be found [here](https://docs.docker.com/build/building/multi-platform/#cross-compilation).

## How this PR fixes it

The docker image size is significantly reduced by using a multistage build where the first stage builds the static website, and the actual image just hosts the static webside (here using BusyBox `httpd` which is nice and lightweight).

This gets the image down to around 70MB, around 1% of the original image size _(sizes computed from Docker images built on amd64 Linux)_.

Note that there is one weird detail is that static HTTP servers typically don't have support for automatically adding `.html` endings to URL paths. It was worked around here with symlinks. See comment in the Dockerfile for more details.

## How to test it

Build the docker image and run the container locally:

```sh
docker build . -t safe-wallet-web
docker run --rm -p 3000:3000 safe-wallet-web
```

This should give you a local interface at <http://localhost:3000>.

It was also able to successfully [build in CI](https://github.com/safe-global/safe-wallet-web/actions/runs/11686228763/job/32541499128?pr=4481) without timing out:

image

## "Screenshots"

```
$ docker images
REPOSITORY                            TAG          IMAGE ID      CREATED             SIZE
localhost/safe-wallet-web             latest       a78bd066aacd  About a minute ago  67.1 MB
```

## Checklist

    * Analytics are not affected
    * Not applicable

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,39 @@
-FROM node:18-alpine
+FROM --platform=$BUILDPLATFORM node:18-alpine AS build
+ARG BUILDPLATFORM
+
+# install build dependencies
 RUN apk add --no-cache libc6-compat git python3 py3-pip make g++ libusb-dev eudev-dev linux-headers
 WORKDIR /app
 COPY . .
 
-# Fix arm64 timeouts
-RUN yarn config set network-timeout 300000 && yarn global add node-gyp
-
-# install deps
+# install NPM dependencies
 RUN yarn install --frozen-lockfile
 RUN yarn after-install
 
-ENV NODE_ENV production
+ENV NODE_ENV=production
 
-# Next.js collects completely anonymous telemetry data about general usage.
-# Learn more here: https://nextjs.org/telemetry
-# Uncomment the following line in case you want to disable telemetry during the build.
-ENV NEXT_TELEMETRY_DISABLED 1
+# disable Next.js telemetry
+ENV NEXT_TELEMETRY_DISABLED=1
+
+RUN yarn build
+
+FROM alpine
+
+RUN apk add busybox-extras tini
+WORKDIR /www
+COPY --from=build /app/out /www
+
+# Next.js automatically adds `.html` extensions to URLs, but static HTTP file
+# servers don't generally do this. You can work around this by creating symbolic
+# links from `my-page.html` to either `my-page/index.html` or `my-page`
+# depending on whether or not a `my-page` directory exists.
+RUN find /www -name '*.html' -exec sh -c 'f="{}"; b="${f%.*}"; [ -d "$b" ] && ln -s "$f" "$b/index.html" || ln -s "$f" "$b"' ';'
 
 EXPOSE 3000
 
-ENV PORT 3000
+ENV PORT=3000
+RUN echo '#!/bin/sh' > /usr/local/bin/docker-entrypoint.sh &&\
+    echo 'busybox-extras httpd -fvv -h /www -p 0.0.0.0:${PORT:-3000}' >> /usr/local/bin/docker-entrypoint.sh &&\
+    chmod +x /usr/local/bin/docker-entrypoint.sh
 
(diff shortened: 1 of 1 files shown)


nlordell:
Testing Docker build in CI here: https://github.com/safe-global/safe-wallet-web/actions/runs/11686228763/job/32541499128?pr=4481

Dockerfile
```diff
@@ -1,24 +1,39 @@
...
+FROM --platform=$BUILDPLATFORM node:18-alpine AS build
+ARG BUILDPLATFORM
+
+# install build dependencies
 RUN apk add --no-cache libc6-compat git python3 py3-pip make g++ libusb-dev eudev-dev linux-headers
 WORKDIR /app
 COPY . .
 
-# Fix arm64 timeouts
-RUN yarn config set network-timeout 300000 && yarn global add node-gyp
```
nlordell, line 7: no longer needed, we build on native arch.

Dockerfile
```diff
@@ -1,24 +1,39 @@
-FROM node:18-alpine
+FROM --platform=$BUILDPLATFORM node:18-alpine AS build
+ARG BUILDPLATFORM
```
nlordell, line 2: This is a built-in Docker arg that resolves to the native platform of the builder: https://docs.docker.com/build/building/multi-platform/#cross-compilation

Uxio0: I didn't know that, good to know

Uxio0:
@nlordell but this way we will not be able to set up environment variables on the run, right? Like we do here when running the docker image: https://github.com/safe-global/safe-infrastructure/blob/main/container_env_files/ui.env

nlordell:
Ah, I hadn't noticed (I thought it was just serving static files), but you are that it will not.

It looks like it does a `yarn build` to rebuild the interface on `static-serve` on container startup. So the container ships both a development and build environment including pre-built static files, instead of just the Safe{Wallet} interface. Doing a `docker pull` for a 6GB image is too much IMO. I would argue that `safe-infrastructure` could be changed to something like:

```yaml
  ui:
    build: https://github.com/safe-global/safe-wallet-web.git#${UI_VERSION}
    env_file:
      - container_env_files/ui.env
    depends_on:
      - nginx
    expose:
      - 8080
```

The alternative, would be to implement env value substitution ONLY before serving the static 
files (instead of pulling in the entire DEV environment). It _may_ be possible with Next.js standalone builds.

Uxio0:
> It may be possible with Next.js standalone builds.

I tried that in the past but I couldn't make it, maybe @katspaugh knows a workaround

nlordell:
> It may be possible with Next.js standalone builds

Just tried it out, and it isn't because of how the environment variables are used (they are directly replaced in the static files). Some hand-rolled thing could work - but may be brittle.

> Agree, but also building was so slow for M1 computers in the past

I think the speed should be similar-ish to pulling a 6GB image :sweat_smile:. I guess one strategy would be to split the build into different layers so that when replacing the `.env` file, most of the layers can be used from cache (so rebuilding is relatively inexpensive).

katspaugh:
@nlordell so does this work or not? Code looks good.

nlordell:
> so does this work or not?

It does not.

PR #4484, opened 2024-11-05
Feat(sidebar): filter safes by search query [SW-305]

jmealy:
## What it solves

Resolves: https://www.notion.so/safe-global/Add-search-to-safes-list-11f8180fe573800b952bd3189dca3b67?pvs=4

## How this PR fixes it
- When the user types in the search bar, renders a filtered list of all safes based on the search term.

## How to test it
- On the accounts page or sidebar, start typing in the search bar.
- Only safes with names matching the search query should be shown. It uses a fuzzy search, like the safe apps search.
- It should also match chain name and address

## Screenshots
image

## Checklist

### DIFF
--- a/src/components/welcome/MyAccounts/index.tsx
+++ b/src/components/welcome/MyAccounts/index.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react'
+import { useCallback, useMemo, useState } from 'react'
 import {
   Accordion,
   AccordionDetails,
@@ -13,6 +13,7 @@ import {
   TextField,
   Typography,
 } from '@mui/material'
+import debounce from 'lodash/debounce'
 import madProps from '@/utils/mad-props'
 import CreateButton from './CreateButton'
 import AddIcon from '@/public/images/common/add.svg'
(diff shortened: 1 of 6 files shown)


src/components/welcome/MyAccounts/useSafesSearch.ts
```diff
@@ -0,0 +1,22 @@
+import { useMemo } from 'react'
+import Fuse from 'fuse.js'
+import type { MultiChainSafeItem } from './useAllSafesGrouped'
+import type { SafeItem } from './useAllSafes'
+
+const useSafesSearch = (safes: (SafeItem | MultiChainSafeItem)[], query: string) => {
+  const fuse = useMemo(
+    () =>
+      new Fuse(safes, {
+        keys: [{ name: 'name' }],
```
usame-algan, line 10: Should we also allow searching by address? I quickly tested this and it only worked when typing from the start e.g. `0x7a...` but not if I type characters at the end of the address I want to find e.g. if my address ends in `86E2` and I type `86E` there is no result. Is this possible with fuse.js?

src/components/welcome/MyAccounts/index.tsx
```diff
@@ -148,61 +170,84 @@ const AccountsList = ({ safes, onLinkClick, isSidebar = false }: AccountsListPro
...
-                  fontSize="small"
-                  sx={{ mt: '2px', mr: 1, strokeWidth: 2 }}
-                />
-                <Typography variant="h5" fontWeight={700} mb={2}>
-                  Pinned
+            {/* Search results */}
+            {searchQuery ? (
+              <>
+                <Typography variant="h5" fontWeight="normal" mb={1} color="primary.light">
+                  Found {filteredSafes.length} result{filteredSafes.length === 1 ? '' : 's'}
```
usame-algan, line 177: Not a big issue but the list is slightly "jumping" when switching between the searchQuery view and the default view because the spacing of "Found x results" and "Pinned" is a bit different.

src/components/welcome/MyAccounts/useSafesSearch.ts
usame-algan: Another idea to include searching by prefix e.g. if the address includes `sep:0x7a...` searching for "sep" would only show sepolia safes.

jmealy: Good idea! We decided against searching for addresses to keep things simple but it seems like it could be useful and it's not so much extra effort. And having a way to filter by chain would be quite useful I think. Let's add them in

usame-algan (approved):
Should we add an analytics event whenever the search runs?

PR #4489, opened 2024-11-06
Docs: update release docs with missing step for creating final release

jmealy:
A draft release is created automatically but a final release needs to be created manually from the draft. Adds in this extra step.


usame-algan (approved):
The diff looks like more than it is. Not sure if `*` or `-` is needed for bullet points but otherwise looks good!

jmealy:
> The diff looks like more than it is. Not sure if `*` or `-` is needed for bullet points but otherwise looks good!

These changes were autoformatting, but they don't change the markdown so I guess it's just the preferred alternative

PR #4495, opened 2024-11-06
fix: fetch safe overviews including untrusted token fiatTotal

### LINKED ISSUE #4479: Aggregated Multichain balances are wrong for networks without default tokenlists
## Bug description
When requesting the Safe overviews to display fiat balances in the network selector and sidebar we always pass `trusted=true` regardless of the default tokenlist setting and feature flag.

This leads to always 0$ balance for e.g. Base Safes.

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Open a Base safe with non zero fiat balance.
 2. Observe that the sidebar and network selector always shows 0$ while the assets list and dashboard shows the actual value.

## Expected result
The same fiat value as shown on the dashboard.

## Obtained result
Always a 0$ value

## Screenshots
Screenshot 2024-11-05 at 14 48 30

schmanu:
## What it solves

Resolves #4479

## How this PR fixes it
Requests `fiatTotal` in `SafeOverviews` using `trusted=false` as we otherwise receive 0 balances for Safes on networks without Default tokenlists.

## How to test it
- Open a Safe with ERC20 tokens on a network without default token list.
- Compare the balance in the sidebar / network selector with the one in e.g. the dashboard

## Checklist

### DIFF
--- a/src/store/__tests__/safeOverviews.test.ts
+++ b/src/store/__tests__/safeOverviews.test.ts
@@ -175,7 +175,7 @@ describe('safeOverviews', () => {
           ],
           {
             currency: 'usd',
-            trusted: true,
+            trusted: false,
             exclude_spam: true,
           },
         )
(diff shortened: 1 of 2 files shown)


iamacook (approved):
Nit: let's add the reasoning to the desc.

liliya-soroka:
Verified in the notion

katspaugh:
Shouldn't this flag be taken from Redux? Depending on the user toggle.

PR #4499, opened 2024-11-07
chore: update mui to v6

compojoom:
## What it solves
The nextjs v15 update (https://github.com/safe-global/safe-wallet-web/pull/4485) builds on top of react 19. There was a concern raised that mui v5 is not compatible with react 19. 

## How this PR fixes it
Updates to mui v6 and applies codemons and manual changes to make the old v5 code compatible.

## How to test it
This updates touches nearly every UI file. The app should continue to look as before. 

## Checklist


usame-algan:
The font looks different now. Seems like `DM Sans` is not loaded anymore?
Screenshot 2024-11-07 at 16 18 52

Not sure if critical for production but when I change the viewport it seems there is a presentation layer which prevents almost anything in the UI from being clicked. 
Screenshot 2024-11-07 at 16 19 37

usame-algan (approved):

francovenica:
LGTM.
I compared this with current prod and I don't see any differences.

I've check the font change that was reported and looks good

PR #4500, opened 2024-11-07
Feat(sidebar): Add tracking events for search, sorting, and pinning/unpinning safes [SW-308]

jmealy:
## What it solves

Resolves [SW-308]

## How this PR fixes it
- Adds events for new interactions added in the epic

## How to test it
- Open the accounts page. Observe that the number of owned safes and watchlist safes are tracked as before (once on the first load).
- In addition, the number of pinned safes should now be tracked also.
- Use the search bar to filter safes. An event should be emmitted for each search.
- Use the sort by button. Whenever the sort by value changes, an event should be emitted with the new sort by value.
- Pin or unpin a safe. An event should be emitted specifying which action took place.

## Checklist

### DIFF
--- a/src/components/welcome/MyAccounts/index.tsx
+++ b/src/components/welcome/MyAccounts/index.tsx
@@ -59,29 +59,6 @@ const AccountsList = ({ safes, onLinkClick, isSidebar = false }: AccountsListPro
   )
   const filteredSafes = useSafesSearch(allSafes ?? [], searchQuery).sort(sortComparator)
 
-  // We consider a multiChain account owned if at least one of the multiChain accounts is not on the watchlist
-  const ownedMultiChainSafes = useMemo(
-    () => safes.allMultiChainSafes?.filter((account) => account.safes.some(({ isWatchlist }) => !isWatchlist)),
-    [safes],
-  )
-
-  // If all safes of a multichain account are on the watchlist we put the entire account on the watchlist
-  const watchlistMultiChainSafes = useMemo(
-    () => safes.allMultiChainSafes?.filter((account) => !account.safes.some(({ isWatchlist }) => !isWatchlist)),
-    [safes],
-  )
-
-  const ownedSafes = useMemo<(MultiChainSafeItem | SafeItem)[]>(
-    () => [...(ownedMultiChainSafes ?? []), ...(safes.allSingleSafes?.filter(({ isWatchlist }) => !isWatchlist) ?? [])],
-    [safes, ownedMultiChainSafes],
-  )
-  const watchlistSafes = useMemo<(MultiChainSafeItem | SafeItem)[]>(
-    () => [
-      ...(watchlistMultiChainSafes ?? []),
-      ...(safes.allSingleSafes?.filter(({ isWatchlist }) => isWatchlist) ?? []),
-    ],
-    [safes, watchlistMultiChainSafes],
-  )
   const pinnedSafes = useMemo<(MultiChainSafeItem | SafeItem)[]>(
     () => [...(allSafes?.filter(({ isPinned }) => isPinned) ?? [])],
     [allSafes],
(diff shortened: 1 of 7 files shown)


src/components/welcome/MyAccounts/index.tsx
```diff
@@ -59,29 +59,6 @@ const AccountsList = ({ safes, onLinkClick, isSidebar = false }: AccountsListPro
   )
   const filteredSafes = useSafesSearch(allSafes ?? [], searchQuery).sort(sortComparator)
 
-  // We consider a multiChain account owned if at least one of the multiChain accounts is not on the watchlist
-  const ownedMultiChainSafes = useMemo(
```
usame-algan, line 63: Good catch moving these into the tracking hook!

src/components/welcome/MyAccounts/useSafesSearch.ts
```diff
@@ -11,11 +11,13 @@ const useSafesSearch = (safes: (SafeItem | MultiChainSafeItem)[], query: string)
...
-    trackEvent({
-      category: OVERVIEW_EVENTS.SEARCH.category,
-      action: OVERVIEW_EVENTS.SEARCH.action,
-      label: query,
-    })
+    if (query) {
+      trackEvent({
+        category: OVERVIEW_EVENTS.SEARCH.category,
+        action: OVERVIEW_EVENTS.SEARCH.action,
+        label: query,
```
usame-algan, line 18: I would not track the search query as the label as it can contain sensitive user information like address book names.

src/components/welcome/MyAccounts/useTrackedSafesCount.ts
```diff
@@ -34,6 +59,17 @@ const useTrackSafesCount = (
     }
   }, [isLoginPage, ownedSafes, wallet])
 
+  useEffect(() => {
+    const totalSafesPinned = pinnedSafes?.reduce(
+      (prev, current) => prev + (isMultiChainSafeItem(current) ? current.safes.length : 1),
+      0,
+    )
+    if (wallet && !isPinnedSafesTracked && pinnedSafes && pinnedSafes.length > 0 && isLoginPage) {
+      trackEvent({ ...OVERVIEW_EVENTS.TOTAL_SAFES_PINNED, label: totalSafesPinned })
```
usame-algan, line 68: For visibility: I noticed there is now an overlap of watchlist safes tracking and pinned safes tracking i.e. the watchlist safes contain the pinned safes + any safe that has a name while the pinned safes only contain pinned safes.

I would suggest filtering pinned safes from watchlisted safes to account for this e.g. if I have 3 pinned safes and 2 additional watched safes we would track 3 for pinned and 2 for watchlisted instead of 5 for watchlisted.

src/components/welcome/MyAccounts/useSafesSearch.ts
jmealy: ah I had this to debug the debounce, thanks!

src/components/welcome/MyAccounts/useTrackedSafesCount.ts
jmealy: makes sense, thanks!

src/components/welcome/MyAccounts/useTrackedSafesCount.ts
```diff
@@ -1,23 +1,51 @@
...
+  const ownedMultiChainSafes = useMemo(
+    () => safes.allMultiChainSafes?.filter((account) => account.safes.some(({ isWatchlist }) => !isWatchlist)),
+    [safes],
+  )
+
+  // If all safes of a multichain account are on the watchlist we put the entire account on the watchlist
+  const watchlistMultiChainSafes = useMemo(
+    () =>
+      safes.allMultiChainSafes?.filter(
+        (account) => !account.safes.some(({ isWatchlist, isPinned }) => !isWatchlist && !isPinned),
```
usame-algan, line 32:
```suggestion
      safes.allMultiChainSafes?.filter(
        (account) => account.safes.some(({ isWatchlist, isPinned }) => isWatchlist && !isPinned),
```

usame-algan: I think this only keeps safes that are watchlisted and not pinned.

usame-algan (approved):
LGTM 🚀

PR #4509, opened 2024-11-08
fix(sidebar):  fix designs for data widget and empty/disconnected state [SW-507]

jmealy:
## What it solves

Resolves SW-507

## How to test it
See that the accounts page and sidebar matches the designs for:
- when no wallet is connected
- when there are no safes in the list. 
- the datawidget when no data is available.

## Screenshots
No safes:
image

Not connected:
image

Sidebar data widget:
image

## Checklist

### DIFF
--- a/src/features/myAccounts/index.tsx
+++ b/src/features/myAccounts/index.tsx
@@ -42,6 +42,7 @@ import { useSafesSearch } from '@/features/myAccounts/hooks/useSafesSearch'
 import useTrackSafesCount from '@/features/myAccounts/hooks/useTrackedSafesCount'
 import { DataWidget } from '@/features/myAccounts/components/DataWidget'
 import OrderByButton from '@/features/myAccounts/components/OrderByButton'
+import ConnectWalletButton from '@/components/common/ConnectWallet/ConnectWalletButton'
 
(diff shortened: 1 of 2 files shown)


src/features/myAccounts/index.tsx
```diff
@@ -191,7 +199,7 @@ const AccountsList = ({ safes, onLinkClick, isSidebar = false }: AccountsListPro
                 </Box>
 
                 {/* All Accounts */}
-                <Accordion sx={{ border: 'none' }}>
+                <Accordion sx={{ border: 'none' }} defaultExpanded={!isSidebar}>
```
usame-algan, line 202: In the sidebar it is now always collapsed even if there are no pinned safes. Didn't we want to expand it if there are no pinned safes and keep it collapsed if there is at least one pinned safe?

usame-algan (approved):

src/features/myAccounts/index.tsx
jmealy: The defaultExpanded prop wasn't working because it only applies to the first render, and the safes haven't loaded yet. I'll look into it more tomorrow on the flight

PR #4529, opened 2024-11-18
Fix: add a guard for a possibly undefined safeAccountConfig

katspaugh:
## What it solves

The app might crash when coming back after a 6-week vacation with an old localStorage.

Screenshot 2024-11-16 at 18 34 38

## How to test

* Check out an old version of the app (from September 2024)
* Connect a wallet
* Add a few Safes to the watchlist
* Then check out the latest production code
* Connect wallet
* Open the Safe List sidebar
* App crashes

### DIFF
--- a/src/features/counterfactual/utils.ts
+++ b/src/features/counterfactual/utils.ts
@@ -356,27 +356,20 @@ export const extractCounterfactualSafeSetup = (
       saltNonce: string | undefined
     }
   | undefined => {
-  if (!undeployedSafe || !chainId) {
+  if (!undeployedSafe || !chainId || !undeployedSafe.props.safeAccountConfig) {
     return undefined
   }
-  if (isPredictedSafeProps(undeployedSafe.props)) {
-    return {
-      owners: undeployedSafe.props.safeAccountConfig.owners,
-      threshold: undeployedSafe.props.safeAccountConfig.threshold,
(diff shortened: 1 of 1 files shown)


src/features/counterfactual/utils.ts
```diff
@@ -356,27 +356,20 @@ export const extractCounterfactualSafeSetup = (
...
-      owners,
-      threshold: Number(threshold),
-      fallbackHandler,
-      safeVersion: undeployedSafe.props.safeVersion,
-      saltNonce: undeployedSafe.props.saltNonce,
-    }
+  const { owners, threshold, fallbackHandler } = undeployedSafe.props.safeAccountConfig
+  const { safeVersion, saltNonce } = isPredictedSafeProps(undeployedSafe.props)
+    ? undeployedSafe.props.safeDeploymentConfig ?? {}
+    : undeployedSafe.props
```
compojoom, line 365: I'm all for turning this on: https://eslint.org/docs/latest/rules/no-nested-ternary. Those ternaries always give me the chills.

compojoom: Let's agree that different people have different views on what is easier to read :)

compojoom (approved):
Haven't tested it, but code looks ok.

Issue #4331, opened 2024-10-08
Migrate from `/transaction-confirmation` to `/preview` endpoint


iamacook:
## What is the feature about

The `/transaction-confirmation` endpoint only returns swap and staking-related data. However, the `/preview` endpoint returns this _and_ the `TransactionInfo` of all other types. By migrating, we can provide a more details transaction confirmation overview.

Note: the `/transaction-confirmation` endpoint has been marked as deprecated and, after this, will be removed.

## The list of requirements

- Migrate from `/transaction-confirmation` to `/preview` endpoint
- Add transaction confirmation mapping for non-swap/staking `TransactionInfo`

## Links

- [Dreprecation PR that adds swaps/staking to `/preview` endpoint](https://github.com/safe-global/safe-client-gateway/pull/1973)

PR #4177, opened 2024-09-16
Refactor: unify confirmations in the transaction flow

clovisdasilvaneto:
## What it solves

Unify all the Review screens from the tx-flow floder.

## How this PR fixes it

It introduces a refactor in the way how we handle the confirmation views on the SignOrExecuteForm component. This PR introduces the first part of the refactor, by creating a central component to handle the presentation of the Review screens. It also makes all modals from the approval owner flow to use the same SettingsChange component (which was part of the ReviewOwner component.

## How to test it
1- Go to the add signer flow (in a modal)
2- Go to the confirmation screen.

You will see the new screen shows a loading screen before the confirmation screen, it happens because we had to move some data outside the SignOrExecuteForm component, and these data was asynchronous, so in order to prevent some glitch in the screen, we show the loading component before we have load the necessary data to show in the component.

## Checklist
* [x] I've tested the branch on mobile 📱
* [ ] I've documented how it affects the analytics (if at all) 📊
* [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #4541, opened 2024-11-20
Refactor: rm unnecessary tx decoding

katspaugh:
## What it solves

A follow-up on #4177 that partially addresses #4331.

## How it solves it

* Remove unnecessary useDecodeTx calls because we now always propose txs, so they already have txInfo and txData and don't need to be decoded from raw tx data
* Unify the handling of Swap and Stake transaction confirmations

## How to test it

* Make a swap tx
* Make a staking tx
* Make a threshold change tx

All the txs should work exactly as before.


src/features/stake/components/StakingConfirmationTx/Exit.tsx
```diff
@@ -28,7 +28,7 @@ const StakingConfirmationTxExit = ({ order }: StakingOrderConfirmationViewProps)
             label: 'Exit',
           },
           {
-            value: order.value || '0',
+            value: (order as unknown as { value: string }).value || '0',
```
katspaugh, line 31: This is needed because the type in the gateway SDK is incorrect, it omits `value` from the TxInfo incorrectly.

katspaugh:
Todo: test how pre-proposals work in counterfactual Safes.

src/components/tx/SignOrExecuteForm/hooks.ts
```diff
@@ -146,14 +146,8 @@ export const useTxActions = (): TxActions => {
       let tx: TransactionDetails | undefined
       // Relayed transactions must be fully signed, so request a final signature if needed
       if (isRelayed && safeTx.signatures.size < safe.threshold) {
-        if (txId) {
-          safeTx = await signRelayedTx(safeTx)
-          tx = await _propose(wallet.address, safeTx, txId, origin)
-        } else {
-          tx = await _propose(wallet.address, safeTx, txId, origin)
-          safeTx = await signRelayedTx(safeTx)
-        }
```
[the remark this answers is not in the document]
usame-algan: Yeah it looks weird but I remember there was a good reason to add this. Will take a look into this again.

src/features/stake/components/StakingConfirmationTx/Exit.tsx
usame-algan: It works for me with the new SDK like so:

```
import type { components } from '@safe-global/safe-client-gateway-sdk'

type StakingOrderConfirmationViewProps = {
  order: components['schemas']['NativeStakingValidatorsExitConfirmationView']
}
```

katspaugh: Migrating to the new SDK is out of scope for this PR. Besides, the type in question isn't `NativeStakingValidatorsExitConfirmationView`, it's the StakingTxInfo.

src/components/tx/SignOrExecuteForm/hooks.ts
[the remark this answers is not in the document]
usame-algan: Double checked this against the test cases in #3462 and it still works 👍

src/components/tx/confirmation-views/SwapOrder/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+interface SwapOrderProps extends NarrowConfirmationViewProps {}
+
+function SwapOrder({ txDetails, txInfo }: SwapOrderProps) {
+  return (
+    <SwapOrderConfirmation
+      order={txInfo as unknown as AnySwapOrderConfirmationView}
+      settlementContract={txDetails?.txData?.to?.value ?? ''}
+    />
+  )
+}
```
usame-algan, line 14: I've checked the usage of `SwapOrder` and we only use it in places where we have guards for checking `txInfo`. Can't we type `txInfo` here to be more narrow (i.e. `Order`) instead of still resolving to `TransactionInfo`? That way we wouldn't need to cast.

This would match the structure of `StakingTxProps`

src/components/tx/confirmation-views/index.tsx
```diff
@@ -51,12 +52,15 @@ const getConfirmationViewComponent = ({
 
   if (isExecTxData(txDetails.txData)) return <ExecTransaction data={txDetails.txData} isConfirmationView />
 
+  if (isSwapOrderTxInfo(txInfo)) return <SwapOrder txDetails={txDetails} txInfo={txInfo} />
+
+  if (isAnyStakingTxInfo(txInfo)) return <StakingTx txDetails={txDetails} txInfo={txInfo as StakingTxInfo} />
```
usame-algan, line 57:
```suggestion
  if (isAnyStakingTxInfo(txInfo)) return <StakingTx txDetails={txDetails} txInfo={txInfo} />
```

usame-algan: It works for me without casting

usame-algan (approved):
There is a unit-test failing because faker randomly generated an address that contains 420. We might want to choose a number thats more "random" to make that test more robust.

src/components/tx/confirmation-views/SwapOrder/index.tsx
katspaugh: Good point, fixed.

francovenica:
Tried Swap, limit, TWAP, staking and claiming and changing a safe threshold

Everything worked just fine

PR #4548, opened 2024-11-22
Chore: replace @walletconnect packages with @reown/walletkit

katspaugh:
## What it solves

WalletConnect packages are now under a new namespace on npm. See https://docs.reown.com/walletkit/upgrade/from-web3wallet-web


package.json
```diff
@@ -66,7 +66,6 @@
     "@sentry/react": "^7.91.0",
     "@spindl-xyz/attribution-lite": "^1.4.0",
     "@walletconnect/utils": "^2.17.2",
```
parius, line 68: could this be also taken from reown?

katspaugh: I've asked them about this.

usame-algan: Looks like they still recommend this package according to their docs: https://docs.reown.com/walletkit/web/installation

usame-algan (approved):
👍

package.json
usame-algan: Although I would suggest to add an explicit dependency to `@walletconnect/core` since we are using it as well.

katspaugh:
I've tested that WalletConnect still works. ✅ 

Screenshot 2024-11-22 at 11 42 43

PR #4550, opened 2024-11-22
Fix: replace "signer(s)" with "N signers" or "1 signer"

katspaugh:
## What it solves

Writing `signer(s)` is a little bit strange given that we know if it's plural or not, like it's a book and not a computer UI.

Screenshot 2024-11-22 at 10 13 33


iamacook (approved):
We can also use this for [remaining relays](https://github.com/safe-global/safe-wallet-web/blob/dev/src/components/tx/RemainingRelays/index.tsx#L9-L23) as well.

katspaugh:
I'll merge directly as it's mostly a mechanical change covered by unit and e2e tests.

PR #4552, opened 2024-11-22
Refactor: pinned = added safes [SW-304]

katspaugh:
## What it solves

Simplify the code and the UI by treating all Added Safes as pinned.

There's now no distinction between "watchlist" Safes and "pinned" Safes. Watchlist Safes are simply pinned read-only Safes.

Counterfactual Safes will be now always displayed under Accounts because they are added to the `useAllSafes` hook.


francovenica:
@katspaugh 

The CF safes are added twice
image

katspaugh:
@francovenica fixed 👍

francovenica:
The issue was fixed. 
There rest of the PR looks good to me too, so from QA side is fine

PR #4562, opened 2024-11-25
Fix: catch erros in message decoding

### LINKED ISSUE #4561: SW-578 Messages: The app is crashed on Messages opening
## Bug description
The app is crashed on Messages opening 

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Base

## Steps to reproduce
 1. Go to the messages tab in the safe from the support report
 Current result: the app is crashed  

## Expected result
The messages should be displayed

katspaugh:
## What it solves

Resolves #4561

## How this PR fixes it

I've added a catch-all error boundary around message details but also smaller localized try-catches for a couple places where it was failing in a specific safe.


usame-algan (approved):

liliya-soroka:
Verified

PR #4564, opened 2024-11-25
fix: Advanced safe creation address computation

### LINKED ISSUE #4560: Advanced safe creation saltNonce is different
## Bug description

When setting a saltNonce during advanced safe creation the generated address is different than when creating the same safe through normal safe creation. This is due to differences in how the address is generated in `ReviewStep` and in `AdvancedOptionsStep`

- The `ECOSYSTEM_ID_ADDRESS` is not used in the advanced safe creation
- The saltNonce is computed in both steps which can result in a different safe address being deployed

## Steps to reproduce
 1. Go to `/new-safe/advanced-create`
 2. Create a 1/1 safe with saltNonce 0
 3. Observe the address
 4. Go to `/new-safe/create`
 5. Create the same safe
 6. Observe a different address

usame-algan:
## What it solves

Resolves #4560 

## How this PR fixes it

- Adds a `paymentReceiver` field to the advanced safe creation
- Uses the `ECOSYSTEM_ID_ADDRESS` as the default value for `paymentReceiver` during advanced safe creation
- Reuses `predictAddressBasedOnReplayData` during advanced safe creation for computing the safe address
- Uses `paymentReceiver` and `saltNonce` in `ReviewStep` if passed

## How to test it

Make sure that the advanced creation flags already created safes with and without the new paymentReceiver address

1. Go the `/new-safe/advanced-create`
2. Create an existing safe and go to the advanced options step
3. Observe that there is an error when using the same saltNonce
4. Choose any salt nonce, safe version and payment receiver
5. Go to the next step and submit the safe creation tx
6. Observe that the transaction data reflects previously chosen values for salt nonce, safe version and payment receiver

---

Make sure that multichain safe creation still works as expected

## Screenshots
Screenshot 2024-11-25 at 13 15 59

## Checklist

### DIFF
--- a/src/components/new-safe/create/steps/AdvancedOptionsStep/index.tsx
+++ b/src/components/new-safe/create/steps/AdvancedOptionsStep/index.tsx
@@ -1,6 +1,8 @@
+import { predictAddressBasedOnReplayData } from '@/features/multichain/utils/utils'
+import { useWeb3ReadOnly } from '@/hooks/wallets/web3'
 import { Button, MenuItem, Divider, Box, TextField, Stack, Skeleton, SvgIcon, Tooltip, Typography } from '@mui/material'
 import { Controller, FormProvider, useForm } from 'react-hook-form'
-import type { ReactElement } from 'react'
+import { type ReactElement, useMemo } from 'react'
 
 import type { StepRenderProps } from '@/components/new-safe/CardStepper/useCardStepper'
 import type { NewSafeFormData } from '@/components/new-safe/create'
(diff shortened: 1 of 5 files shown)


src/components/new-safe/create/steps/AdvancedOptionsStep/index.tsx
```diff
@@ -183,15 +176,57 @@ const AdvancedOptionsStep = ({ onSubmit, onBack, data, setStep }: StepRenderProp
...
+                required: true,
+              })}
+              label="Payment receiver"
+              error={Boolean(formState.errors[AdvancedOptionsFields.paymentReceiver]) || Boolean(isDeployed)}
+              helperText={
+                formState.errors[AdvancedOptionsFields.paymentReceiver]?.message ??
+                (Boolean(isDeployed) ? 'The Safe is already deployed. Use a different payment receiver.' : undefined)
+              }
+              fullWidth
+            />
```
katspaugh, line 225: Was that someone's requirement?

katspaugh (approved):

src/components/new-safe/create/steps/AdvancedOptionsStep/index.tsx
katspaugh: Should we add a tip along the lines of "if you created your Safe before 09.2024, leave this empty"?

francovenica:
LGTM.

Created several safes, editing the saltNonce and leaving it to the creation flow (normal safe creation)
Made sure that in all of them you can see in the blockexplorer that the salt nonce is the one set. 
Checked that you cannot re use a salt nonce
Checked that trying to use a used saltNonce shows the error of "safe already deployed". Also the salt nonce "re-guesses" the used safe correctly.

PR #4565, opened 2024-11-26
feat: Update header safe token widget for SAP [SW-586]

usame-algan:
## What it solves

Resolves [SW-586](https://www.notion.so/safe-global/Update-safe-token-header-widget-14a8180fe57380e9a399cad809b8dc16)

## How this PR fixes it

- Adds an `UnreadBadge` to the safe token header widget in case there is an allocation
- Shows a different tooltip text in case there is an allocation

## How to test it

1. Open the app with a safe that has a SAP allocation on staging
2. Observe the updated header widget for safe tokens
3. Open the app on prod or with a safe without an allocation
4. Observe the header widget looks as before

## Screenshots
Screenshot 2024-11-26 at 10 38 06

## Checklist

### DIFF
--- a/src/components/common/SafeTokenWidget/index.tsx
+++ b/src/components/common/SafeTokenWidget/index.tsx
@@ -1,6 +1,8 @@
+import UnreadBadge from '@/components/common/UnreadBadge'
 import { IS_PRODUCTION, SAFE_TOKEN_ADDRESSES, SAFE_LOCKING_ADDRESS } from '@/config/constants'
 import { AppRoutes } from '@/config/routes'
 import useChainId from '@/hooks/useChainId'
+import type { Vesting } from '@/hooks/useSafeTokenAllocation'
 import useSafeTokenAllocation, { useSafeVotingPower } from '@/hooks/useSafeTokenAllocation'
 import { OVERVIEW_EVENTS } from '@/services/analytics'
 import { formatVisualAmount } from '@/utils/formatters'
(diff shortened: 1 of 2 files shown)


usame-algan:
> Can you insert a screenshot? In my safes, the widget looks the same as before.

Added a screenshot to the description. The copy needs to be updated before we merge this so I would leave the PR open for now.

katspaugh (approved):

francovenica:
I think the tooltip should not show for non-owners and disconnected users:

image

Beyond that it looks good

usame-algan:
> I think the tooltip should not show for non-owners and disconnected users:

Changed it so it only displays the update for safe owners

francovenica:
LGMT

usame-algan:
@francovenica I will push another change to exclude the allocation of the new Airdrop as long as the start date has not been reached otherwise it is misleading to include this number imo

francovenica:
I saw the number changing on the top bar, but I still see the old number in the "Governance & claiming":

Is this out of the scope? of this ticket?

image

francovenica:
I still see the old number in this section. Is it correct? 

image

usame-algan:
> The top bar is fine, but I still see the old number in this section. Is it correct?

Yes, the 10k number in the header widget is correct because the allocation for SAP has not started yet. I will also update the number in the governance app so it will say "Total voting power is 10k" in this case.

Issue #4404, opened 2024-10-22
Error when trying to sign/execute transaction too fast after page refresh


usame-algan:
## Bug description

When loading the app and pressing the Confirm button in the queue there is sometimes an error that the SDK could not be initialized. The SDK takes a bit until ready but if a user tries to "interact" with it too fast it will error out. This especially happens with e2e tests that do this step programmatically and run into this error more often which makes tests more flaky.

## Steps to reproduce
 1. Add a transaction to the queue
 2. Go to the queue
 3. Reload the app
 4. Press the confirm button right after reloading
 5. Observe an error

## Expected result

There should not be an error

## Obtained result

There is an error visible in the UI

PR #4567, opened 2024-11-26
fix: Init Safe SDK for replayed safes

usame-algan:
## What it solves

Follow-up on #4404 

The local storage structure of undeployed safes changed with multichain but during SDK initialization we only checked for the old structure instead of also the new one. This results in the Activate now button always being disabled for new counterfactual safes.

## How this PR fixes it

- Init the Safe SDK if undeployed safe `isReplayedSafeProps` or `isPredictedSafeProps`

## How to test it

2. Observe that the Activate now button is enabled

## Screenshots

Screenshot 2024-11-26 at 11 51 07

## Checklist


katspaugh (approved):

francovenica:
The button is enabled once the user reaches the CF safe main hub.
LGTM

PR #4568, opened 2024-11-26
fix: only return "next" transactions as pending

iamacook:
## What it solves

Resolves [duplicate "Queue" section](https://www.notion.so/safe-global/Nested-safe-v1-Mess-in-the-queue-with-pending-section-2-queued-sections-and-next-in-between-1378180fe57380cb868be95e7a4f46a2)

## How this PR fixes it

The "Pending" section of the queue contains immediately executed (untrusted) transactions.

Our `usePendingTxsQueue` [assumes there is only one pending transaction](https://github.com/safe-global/safe-wallet-web/blob/only-next-pending/src/hooks/usePendingTxs.ts#L58). However, if multiple untrusted transactions are proposed (particularly via nested signing), there can be >1 pending transactions, e.g. next label, transaction, queue label, transaction.

The current implementation therefore returns the "Pending" ("Next") section and "Queue" - duplicating the "Queue" section:

image

This returns only the "Next" transaction(s) as "Pending" meaning there will be no duplicate "Queue" section.

## How to test it

Create multiple nested Safe transactions, observing no duplicate "Queue" transaction.

## Checklist

### DIFF
--- a/src/hooks/__tests__/usePendingTxs.test.ts
+++ b/src/hooks/__tests__/usePendingTxs.test.ts
@@ -14,7 +14,8 @@ import {
   ConflictType,
 } from '@safe-global/safe-gateway-typescript-sdk'
 import * as useSafeInfoHook from '@/hooks/useSafeInfo'
-import { filterUntrustedQueue, useHasPendingTxs, usePendingTxsQueue } from '../usePendingTxs'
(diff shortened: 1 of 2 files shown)


iamacook:
I'm going to merge this and over to https://github.com/safe-global/safe-wallet-web/pull/4406 as it is integral there.

PR #4571, opened 2024-11-26
fix: Don't break array elements in tx details into multiple lines if …

usame-algan:
## What it solves

In case a transaction method parameter is an empty array we should not break the brackets into multiple lines to improve readability.

## How this PR fixes it

- Only renders the div for the array contents if its not empty

## How to test it

1. Open a transaction that has array method parameters that are empty
2. Observe the array brackets don't break into the next line

## Screenshots

Before:
Screenshot 2024-11-26 at 15 50 11

After:
Screenshot 2024-11-26 at 15 50 45

## Checklist


katspaugh (approved):

katspaugh:
Tested ✅

PR #4580, opened 2024-11-29
Chore: fix lodash imports

katspaugh:
## What it solves

Reduces the bundle size by ~400kb.


usame-algan (approved):
I thought `optimizePackageImports` fixes this already 🙈

PR #4593, opened 2024-12-03
Refactor: white label branding + domain checks

katspaugh:
## What it solves

To show or hide trademark elements like the Safe Wallet logo, legal pages etc, do a runtime check of the current domain in addition to an env var.

Previously, we were only checking an env variable which apparently wasn't enough as some forks are still using Safe{Wallet} branding. I've also extended this check to the logo in the header.

## How does it solve it

I've added two new environment variables:
* `NEXT_PUBLIC_BRAND_NAME` – set to `Safe{Wallet}` for the official distribution, falls back to `Wallet fork`.
* `NEXT_PUBLIC_BRAND_LOGO` – takes the URL of the main logo in the header


src/config/constants.ts
```diff
@@ -99,6 +99,7 @@ export const TWITTER_URL = 'https://twitter.com/safe'
 
 // Legal
 export const IS_OFFICIAL_HOST = process.env.NEXT_PUBLIC_IS_OFFICIAL_HOST === 'true'
+export const OFFICIAL_HOSTS = /app\.safe\.global|.+\.5afe\.dev|localhost:3000/
```
katspaugh, line 102: Todo: move this to GitHub Secrets so that it's a bit more difficult to overwrite.

usame-algan:
Should we add some kind of switch in the UI to make it easier to test this?

usame-algan:
An additional idea: Write a blog post / help article / document on how to set up your own fork of the wallet app since there are a couple of steps involved.

francovenica:
Is there any change in the UI or behaviour that the user might find?

katspaugh:
@francovenica actually no, it's mostly a refactoring for developers.

PR #4595, opened 2024-12-03
fix: Keep Sidebar drawer mounted

usame-algan:
## What it solves

The Sidebar has a delay when opening with a lot of safes. This PR keeps the Sidebar Drawer mounted to resolve it. It does not refactor the hooks that are slow.

## How this PR fixes it

- Adds `keepMounted: true` to the Sidebar Drawer

## How to test it

1. Open a Safe with a wallet that owns a lot of safes
2. Open and close the sidebar a few times
3. Observe no delay when doing so

## Checklist


src/components/sidebar/Sidebar/index.tsx
```diff
@@ -58,7 +58,13 @@ const Sidebar = (): ReactElement => {
 
         <IndexingStatus />
       </div>
-      <Drawer variant="temporary" anchor="left" open={isDrawerOpen} onClose={onDrawerToggle}>
+      <Drawer
+        variant="temporary"
+        anchor="left"
+        open={isDrawerOpen}
+        onClose={onDrawerToggle}
+        ModalProps={{ keepMounted: true }}
```
katspaugh, line 66: Will it remain mounted after it was rendered once, or will it be mounted from the start? If it's the latter, I would instead look into removing the animation delay like I did for the BarchSidebar.

usame-algan: The animation doesn't cause the delay but the computations we do inside the nested components. The Sidebar unmounts every time the user closes it so it keeps doing those computations every time it opens again. We should definitely look into improving performance on those computations but this is out of scope for this fix.

katspaugh: To avoid making extra requests on page load, I would suggest doing something like this then:

```js
const [keepMounted, setKeepMounted] = useState(false)

useEffect() => setKeepMounted(true), [])
```

katspaugh:
This can be closed now thanks to #4604.

PR #4598, opened 2024-12-03
fix: do not offer executing unsigned txs with connected wallet

schmanu:
## What it solves
We currently offer executing unsigned txs through the connected wallet whilst having a subaccount opened.

## How this PR fixes it
Only adds the connected wallet as option if the transaction is already fully signed or the wallet is an owner.

## How to test it
1. Open a 1/n Subaccount whose parent is owned by the connected wallet
2. Create a new tx with the current Safes nonce
3. Select to Execute the tx directly
4. Observe that the connected wallet is not offered
5. Now only Sign the tx
6. When signed, click execute 
7. Observe that the connected wallet is now an option

## Checklist


src/components/tx/SignOrExecuteForm/SignerForm/index.tsx
```diff
@@ -58,13 +58,14 @@ export const SignerForm = ({ willExecute }: { willExecute?: boolean }) => {
       return []
     }
     const owners = new Set(nestedSafeOwners ?? [])
+    const isFullySigned = safeTx ? safeTx.signatures.size >= safe.threshold : false
```
katspaugh, line 61: Why only fully signed and not also `threshold - 1`?

schmanu: Because only fully signed transactions can be executed by non-owners.
There is another condition in case the connected wallet is an owner itself - in which case we always offer to chose it.

katspaugh (approved):
You should pull from dev to fix the tests.

iamacook (changes requested):
We also need to consider delegates. I am now no longer able to propose a nested transaction when connected as a parent delegate to a child.

iamacook (approved):
After discussion, and considering we purposely excluded other signing cases such as spending limites, recovery, roles, etc., we will disregard delegates. If requested, we will revisit it later

liliya-soroka:
Verified

PR #4604, opened 2024-12-04
fix: remove unnecessary deps. from `useAllOwnedSafes` cache

iamacook:
## What it solves

Resolves unnecessary cache updates

## How this PR fixes it

We cache the owned Safes for a given address. This is used as an initial value for the sidebar, before the list of owned Safes is returned.

Currently, we are updating the cache if the aforementioned list is returned, as well as when the request address changes and **when the cache updates**. This results in an endless update to the cache, broadcasting across browser tabs.

This removes the unnecessary deps.: `address` and `cache`.

## How to test it

1. Open the same Safe in two different browser tabs, ensuring wallet connection with the same owner.
2. Create a new Safe, owned by the same owner.
3. Refresh both tabs and observe no `storage` event warning in the console.

## Checklist

### DIFF
--- a/src/features/myAccounts/hooks/useAllOwnedSafes.ts
+++ b/src/features/myAccounts/hooks/useAllOwnedSafes.ts
@@ -17,7 +17,7 @@ const useAllOwnedSafes = (address: string): AsyncResult<AllOwnedSafes> => {
     if (data != undefined) {
       setCache(data)
     }
-  }, [address, cache, data, setCache])
+  }, [data, setCache])
 
   return [cache, asError(error), isLoading]
 }


schmanu (approved):

katspaugh:
Looks like this circular dependency was added in #3708, quite some time ago. Strange that we noticed only now.

compojoom:
Not using `currentData` on this line:
const { data, error, isLoading } = useGetAllOwnedSafesQuery(address === '' ? skipToken : { walletAddress: address })

would mean that when the user disconnects you would be still served the last data available for this query. And I don't think you want this. Also I don't think you need the locastorage cache at all.

iamacook:
> Not using `currentData` on this line: const { data, error, isLoading } = useGetAllOwnedSafesQuery(address === '' ? skipToken : { walletAddress: address })
> 
> would mean that when the user disconnects you would be still served the last data available for this query. And I don't think you want this. Also I don't think you need the locastorage cache at all.

Whilst I agree, implementing either change would alter the current behavior. The primary goal of this PR is to address the existing bug.

src/features/myAccounts/hooks/useAllOwnedSafes.ts
```diff
@@ -17,7 +17,7 @@ const useAllOwnedSafes = (address: string): AsyncResult<AllOwnedSafes> => {
     if (data != undefined) {
       setCache(data)
     }
-  }, [address, cache, data, setCache])
+  }, [data, setCache])
```
katspaugh, line 20: As per @compojoom comment, should we restore `address` here?
Only `cache` was causing a recursion.

iamacook: I don't think it's necessary as keeping `address` as a dep. results in two `setCache` calls each time it changes:

1. The first happens immediately when `address` changes.
2. The second happens when `useGetAllOwnedSafesQuery` fetches and updates `data`.

iamacook: `data` remains populated when the wallet is disconnected though. [Passing `skipToken` to `useGetAllOwnedSafesQuery`](https://github.com/safe-global/safe-wallet-web/blob/c33e74a9c44b1d15a409091ea62a561f585170b2/src/features/myAccounts/hooks/useAllOwnedSafes.ts#L14) prevents the query from fetching new data or clearing the existing cached data.

katspaugh: OK, then let's treat it as a separate bug. I'll merge the PR now, thanks a lot!

PR #4606, opened 2024-12-04
Tests: fix safe-wallet-provider unit tests

katspaugh:
## What it solves

This test was passing locally but not on CI, likely because of the fuzzy output from faker. If a hex string doesn't contain alphabetical characters, it passes.


iamacook (commented):
Did we check if this is happening because of the [`casing` of `faker.string.hexadecimal`](https://fakerjs.dev/api/string.html#hexadecimal) (it defaults to `mixed`)?

A `casing` of `lower` would match the output of `numberToHex`.

PR #4620, opened 2024-12-06
Refactor: restore old MUI props

katspaugh:
## What it solves

Allow using restructured sx props like in MUI v5 (upgraded to v6 in #4499).

Atm only two components:
* Box
* Typography

More can be added if we see they are frequently used.


clovisdasilvaneto (approved):
Nice, can we cover it with unit tests?

clovisdasilvaneto (approved):

PR #4622, opened 2024-12-06
fix: Show raw data if decodedData doesn't contain any parameters [SW-435]

usame-algan:
## What it solves

Resolves [SW-435](https://www.notion.so/safe-global/Display-raw-fallback-data-12f8180fe57380fc941cd73042bb0837)

## How this PR fixes it

- Removes the `isFallback` condition and instead shows `HexEncodedData` additionally when there are no parameters
- Case from https://github.com/safe-global/safe-wallet-web/pull/4260 is still covered
Screenshot 2024-12-06 at 15 52 16

## How to test it

1. Open a transaction where the method name is there but no parameters
2. Observe it displays Hex encoded data instead of nothing

## Screenshots
Screenshot 2024-12-06 at 15 52 24

## Checklist

### DIFF
--- a/src/components/transactions/TxDetails/TxData/DecodedData/MethodDetails/index.tsx
+++ b/src/components/transactions/TxDetails/TxData/DecodedData/MethodDetails/index.tsx
@@ -1,3 +1,4 @@
+import { HexEncodedData } from '@/components/transactions/HexEncodedData'
 import type { ReactElement } from 'react'
 import { generateDataRowValue, TxDataRow } from '@/components/transactions/TxDetails/Summary/TxDataRow'
 import { isAddress, isArrayParameter, isByte } from '@/utils/transaction-guards'
@@ -7,21 +8,25 @@ import { Value } from '@/components/transactions/TxDetails/TxData/DecodedData/Va
 
 type MethodDetailsProps = {
   data: DataDecoded
+  hexData?: string
   addressInfoIndex?: {
     [key: string]: AddressEx
   }
 }
 
-export const MethodDetails = ({ data, addressInfoIndex }: MethodDetailsProps): ReactElement => {
+export const MethodDetails = ({ data, hexData, addressInfoIndex }: MethodDetailsProps): ReactElement => {
   if (!data.parameters?.length) {
     return (
-      <Typography
-        sx={{
-          color: 'text.secondary',
-        }}
-      >
-        No parameters
-      </Typography>
+      <>
+        <Typography
+          sx={{
+            color: 'text.secondary',
+          }}
+        >
+          No parameters
+        </Typography>
+        {hexData && <HexEncodedData title="Data (hex-encoded)" hexData={hexData} />}
+      </>
     )
   }
 
--- a/src/components/transactions/TxDetails/TxData/DecodedData/MethodDetails/index.tsx
+++ b/src/components/transactions/TxDetails/TxData/DecodedData/MethodDetails/index.tsx
@@ -25,6 +25,7 @@ export const MethodDetails = ({ data, hexData, addressInfoIndex }: MethodDetails
         >
           No parameters
         </Typography>
+
         {hexData && <HexEncodedData title="Data (hex-encoded)" hexData={hexData} />}
       </>
     )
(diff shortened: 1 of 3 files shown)


src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
```diff
@@ -38,8 +38,6 @@ export const DecodedData = ({ txData, toInfo }: Props): ReactElement | null => {
     )
   }
 
-  if (!txData) return null
```
usame-algan, line 41: While writing tests I realised the only reason this condition is here is to satisfy Typescript so I refactored it. Now it will show that `txData` is defined further down.

src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
```diff
@@ -49,7 +47,7 @@ export const DecodedData = ({ txData, toInfo }: Props): ReactElement | null => {
     ? 'this Safe Account'
     : addressInfo?.name || toInfo?.name || txData.to.name
   const avatar = addressInfo?.logoUri || toInfo?.logoUri || txData.to.logoUri
-  const isFallback = !method && !txData?.dataDecoded?.parameters
+  const isFallback = !method || txData?.dataDecoded?.parameters?.length === 0
```
usame-algan, line 50: This will cover the case listed where there is a method but no parameters. I am not sure if we have the case where there is no method but parameters (don't think so) but in that case it would also fallback to raw data now.

katspaugh: There are plenty of methods with no parameters.
E.g. WETH wrap/unwrap (value is passed as `value`, not in parameters), NFT mints etc.

Screenshot 2024-12-06 at 15 07 54

usame-algan: That case is fine and should now display the raw data instead of "No parameters"
Screenshot 2024-12-06 at 15 20 09

I was talking about the opposite case where potential transactions don't have a method name but parameters for some reason but feels like if the method name is missing then surely the ABI will be missing so there won't be decoded parameters either.

src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
```diff
@@ -49,7 +47,7 @@ export const DecodedData = ({ txData, toInfo }: Props): ReactElement | null => {
...
 
   let decodedData = <></>
   if (txData.dataDecoded && !isFallback) {
```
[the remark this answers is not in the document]
usame-algan: Typescript can't deal with that for some reason.

```
if (!txData && !toInfo) return null

if (!txData && toInfo) return <XYZ />

// txData should be defined from here on out but isn't
```

src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
katspaugh: I wouldn't replace "No parameters" with a raw data hash. The raw data could be added to the list below.

usame-algan: Ah I undertand what you mean now. There are legitimate transactions without parameters so we should clearly tell the user in those cases that there are no parameters but still have the fallback to raw data. Lets display both to them.

usame-algan: Like so
Screenshot 2024-12-06 at 15 31 25

katspaugh: Yes, exactly. And I would move it down, below the separator. It's a bit weird that the copy button would be in front though.

usame-algan: The part below the separator is the `Summary` which we don't render for Multisends. I checked if we can just add another divider between but imo it looks strange:
Screenshot 2024-12-09 at 10 17 19
Screenshot 2024-12-09 at 10 17 29

I would keep it as is since it has a very limited scope (only rendered when we also render "No parameters")

src/components/transactions/TxDetails/TxData/DecodedData/MethodDetails/index.tsx
```diff
@@ -7,21 +8,25 @@ import { Value } from '@/components/transactions/TxDetails/TxData/DecodedData/Va
...
-      </Typography>
+      <>
+        <Typography
+          sx={{
+            color: 'text.secondary',
+          }}
+        >
+          No parameters
+        </Typography>
+        {hexData && <HexEncodedData title="Data (hex-encoded)" hexData={hexData} />}
```
katspaugh, line 28:
```suggestion

        {hexData && <HexEncodedData title="Data (hex-encoded)" hexData={hexData} />}
```

katspaugh (approved):

francovenica:
Lgtm:

Before after comparison:
image

image

PR #4623, opened 2024-12-06
fix: Disable Add to Batch button for safe apps [SW-363]

usame-algan:
## What it solves

Resolves [SW-363](https://www.notion.so/safe-global/Disable-add-to-batch-button-for-Dapps-1238180fe57380ac935aec33de026aef)

## How this PR fixes it

- Disables the `BatchButton` if there is an `origin` passed (it is a safe app or walletconnect)

## How to test it

1. Create a transaction through a safe app
2. Choose Sign
3. Observe Add to batch button is disabled
4. Create a transaction through the native UI
5. Observe there is an Add to batch button when signing and it is enabled

## Checklist

### DIFF
--- a/src/components/tx/SignOrExecuteForm/SignForm.tsx
+++ b/src/components/tx/SignOrExecuteForm/SignForm.tsx
@@ -101,6 +101,8 @@ export const SignForm = ({
   const submitDisabled =
     !safeTx || !isSubmittable || disableSubmit || cannotPropose || (needsRiskConfirmation && !isRiskConfirmed)
 
+  const isSafeAppTransaction = !!origin
+
   return (
     <form onSubmit={handleSubmit}>
       {hasSigned && <ErrorMessage level="warning">You have already signed this transaction.</ErrorMessage>}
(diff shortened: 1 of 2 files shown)


src/components/tx/SignOrExecuteForm/SignForm.tsx
```diff
@@ -101,6 +101,8 @@ export const SignForm = ({
   const submitDisabled =
     !safeTx || !isSubmittable || disableSubmit || cannotPropose || (needsRiskConfirmation && !isRiskConfirmed)
 
+  const isSafeAppTransaction = !!origin
```
katspaugh, line 104: I think origin is also non-null in case of WalletConnect txs and probably swaps/staking/bridging as well.

usame-algan: I think we should hide it for those transactions as well to be on the safe side. At least for swaps the button is already disabled since it is a delegate call:
Screenshot 2024-12-09 at 10 29 49

katspaugh (approved):

usame-algan:
Slightly adjusted it so we always show the button but it is disabled to keep it consistent with previous behaviour.

francovenica:
LGTM

Checked that dapps don't have the add to batch button. Same for other types of tx like reject tx, swap, stake, bridge....
Made sure that the tx that must have the add to batch are still there: send funds, edit owner, edit policies, remove modules...

PR #4629, opened 2024-12-09
Refactor: Box flex -> Stack

katspaugh:
Replace all instances of `<Box display="flex" ... >` with the more idiomatic and shorter `<Stack ... >`.


src/components/balances/AssetsTable/index.tsx
```diff
@@ -185,7 +185,7 @@ const AssetsTable = ({
               sticky: true,
               collapsed: item.tokenInfo.address === hidingAsset,
               content: (
-                <Box display="flex" flexDirection="row" gap={1} alignItems="center">
+                <Stack direction="row" spacing={1} alignItems="center">
```
usame-algan, line 188: Looks like it didn't work here for some reason it keeps using Box.
Screenshot 2024-12-09 at 16 27 58

katspaugh: I noticed a couple other places. Will close the PR, not worth the trouble.

PR #4634, opened 2024-12-10
Chore: update protocol-kit and safe-deployments

katspaugh:
Needed to roll out a new chain.


package.json
```diff
@@ -55,10 +55,9 @@
     "@reduxjs/toolkit": "^2.2.6",
     "@reown/walletkit": "^1.1.1",
     "@safe-global/api-kit": "^2.4.6",
-    "@safe-global/protocol-kit": "^4.1.1",
+    "@safe-global/protocol-kit": "^4.1.2",
     "@safe-global/safe-apps-sdk": "^9.1.0",
     "@safe-global/safe-client-gateway-sdk": "v1.60.1",
-    "@safe-global/safe-deployments": "1.37.12",
```
katspaugh, line 61: I've removed this explicit dependency so that we just always use the package version that `@safe-global/protocol-kit` depends on.

usame-algan: We are importing from this dependency though. IMO it should still be there in the package.json. Does it make sense to mark it as a peer dependency in that case?

katspaugh: What for? It works like this. If it stops working, we'll know.

PR #4637, opened 2024-12-10
fix: CheckWallet sdk only if safe is also loaded

### LINKED ISSUE #4636: Notifications: Can not enable/disable notifications from the global preferences page ( dev env)
## Bug description
Notifications: Can not enable/disable notifications from the global preferences page ( dev env)
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet/any
 - dev env

## Steps to reproduce
 1. Go to global preferences notification
 2. select all safes
 3. try to enable notifications 
 Current result: save button is dsabled 

## Expected result
It should be possible to use global preferences for the notifications enabling/diabling
## Screenshots
image

usame-algan:
## What it solves

Resolves #4636 

## How this PR fixes it

- Also checks for `safeLoaded` when evaluating if the sdk exists or not in `CheckWallet` since the sdk won't exist if there is no safe opened

## How to test it

This should not break existing behaviour for transaction related buttons

1. Go to the global preferences page
2. Activate notifications
3. Observe the Save button is enabled

## Screenshots
Screenshot 2024-12-10 at 16 43 12

## Checklist

### DIFF
--- a/src/components/common/CheckWallet/index.test.tsx
+++ b/src/components/common/CheckWallet/index.test.tsx
@@ -231,8 +231,17 @@ describe('CheckWallet', () => {
     expect(getByText('Continue')).toBeDisabled()
   })
 
-  it('should disable the button if SDK is not initialized', () => {
+  it('should disable the button if SDK is not initialized and safe is loaded', () => {
     mockUseSafeSdk.mockReturnValue(undefined)
+
+    const mockSafeInfo = {
+      safeLoaded: true,
+      safe: extendedSafeInfoBuilder(),
+    }
+
+    ;(useSafeInfo as jest.MockedFunction<typeof useSafeInfo>).mockReturnValueOnce(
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
👍 
Please make sure this component isn't used anywhere outside of an individual Safe. E.g. Safe creation.

usame-algan:
> Please make sure this component isn't used anywhere outside of an individual Safe. E.g. Safe creation.

I wonder why we use `CheckWallet` there in the first place since we just need to make sure that the users wallet is connected before they can press the Save button. I just looked into it a bit more and noticed some strange behaviour in case I have added safes. Will create another ticket to look into this separately.

francovenica:
LGTM

You can disable notifcations from the global settings
Checked that disabling the notifications from there it works (no more notifications arrive after disabling notifications from there)

PR #4641, opened 2024-12-11
Refactor sidebar hooks [SW-647]

usame-algan:
## What it solves

Part of [SW-647](https://www.notion.so/safe-global/Refactor-Sidebar-components-1548180fe57380d08e77d05a3a121aec)

## How this PR fixes it

- Adds tests for `useAllSafes`
- Removes the cache from `useAllOwnedSafes` as it is not needed anymore
- Removes the sorting in `useAllSafes` since it is already done in `MyAccounts`
- Moves `useHasSafes` into its own file

## ToDos

- [x] Write more tests for `useAllSafes`
- [x] Write tests for `buildSafeItem`
- [x] Write tests for `prepareAddresses`
- [x] Refactor `useAllSafes` and split up the logic


src/features/myAccounts/hooks/useAllSafes.ts
```diff
@@ -20,63 +21,71 @@ export type SafeItem = {
...
+  allOwned: AllOwnedSafes,
+  allVisitedSafes: VisitedSafesState,
+  allSafeNames: AddressBookState,
+): SafeItem => {
+  const addedSafe = allAdded[chainId]?.[address]
+  const addedSafeOwners = addedSafe?.owners || []
+  const isPinned = Boolean(addedSafe) // Pinning a safe means adding it to the added safes storage
+
+  // Determine if the user is an owner
+  const isOwnerFromAdded = addedSafeOwners.some(({ value }) => sameAddress(walletAddress, value))
```
jmealy, line 53: This is just for CF safes, right?

usame-algan: Good question. From what I can find is that we are using the `isReadOnly` flag for two things:

1. To show the Add network button
2. To show the queue actions

With that in mind I would expect the following behaviour:

CF Safes:
- Show the Add network button if the connected wallet is an owner regardless of pinned state 🔴 
- No queue actions required

Owned Safes:
- Show the Add network button regardless of pinned state ✅ 
- Show the queue actions regardless of pinned state ✅ 

Added but not owned safes:
- Show neither the Add network button nor the queue actions ✅ 

Anything I missed? I will implement the CF Safes case

jmealy: That behaviour makes sense, thanks for clarifying!

jmealy: >Good question. From what I can find is that we are using the isReadOnly flag for two things:
It's also used to display the read-only chip for the item in the safe list

src/features/myAccounts/hooks/useAllSafes.ts
```diff
@@ -42,17 +42,20 @@ export const _buildSafeItem = (
...
   const addedSafeOwners = addedSafe?.owners || []
+  const undeployedSafeOwners = allUndeployed[chainId]?.[address]?.props.safeAccountConfig.owners || []
   const isPinned = Boolean(addedSafe) // Pinning a safe means adding it to the added safes storage
 
   // Determine if the user is an owner
   const isOwnerFromAdded = addedSafeOwners.some(({ value }) => sameAddress(walletAddress, value))
+  const isOwnerFromCF = undeployedSafeOwners.some((ownedAddress) => sameAddress(walletAddress, ownedAddress))
   const isOwnedSafe = (allOwned[chainId] || []).includes(address)
-  const isOwned = isOwnedSafe || isOwnerFromAdded
+  const isOwned = isOwnedSafe || isOwnerFromAdded || isOwnerFromCF
```
jmealy, line 58: I think `isOwnerFromAdded` is not needed anymore. If it's deployed we check ownership from allOwned, if it's CF we check ownership from allUndeployed, so isOwnerFromAdded is redundant

jmealy (approved):
nice work 🚀

PR #4642, opened 2024-12-11
Refactor: use recommendedMasterCopyVersion

katspaugh:
## What it solves

Instead of the SAFE_141 feature toggle, use the recommendedMasterCopyVersion field from the ChainInfo.

This field from the config service is also used by CGW to determine whether a Safe is UP_TO_DATE or OUTDATED.

Screenshot 2024-12-11 at 17 09 23

## NB

After we merge this, we'll need to make sure to update the config service to switch all chains where SAFE_141 was enabled to 1.4.1. This will have the side effect of enabling the upgrade to 1.4.1 for old Safes – must be tested.

### DIFF
--- a/src/components/common/NetworkSelector/__tests__/NetworkMultiSelector.test.tsx
+++ b/src/components/common/NetworkSelector/__tests__/NetworkMultiSelector.test.tsx
@@ -33,30 +33,35 @@ describe('NetworkMultiSelector', () => {
       .with({ chainName: 'Ethereum' })
       .with({ shortName: 'eth' })
       .with({ features: [FEATURES.COUNTERFACTUAL, FEATURES.MULTI_CHAIN_SAFE_CREATION] as any })
+      .with({ recommendedMasterCopyVersion: '1.4.1' })
       .build(),
     chainBuilder()
       .with({ chainId: '10' })
       .with({ chainName: 'Optimism' })
       .with({ shortName: 'oeth' })
       .with({ features: [FEATURES.COUNTERFACTUAL, FEATURES.MULTI_CHAIN_SAFE_CREATION] as any })
+      .with({ recommendedMasterCopyVersion: '1.4.1' })
       .build(),
     chainBuilder()
       .with({ chainId: '100' })
       .with({ chainName: 'Gnosis Chain' })
       .with({ shortName: 'gno' })
       .with({ features: [FEATURES.COUNTERFACTUAL, FEATURES.MULTI_CHAIN_SAFE_CREATION] as any })
+      .with({ recommendedMasterCopyVersion: '1.4.1' })
       .build(),
     chainBuilder()
       .with({ chainId: '324' })
       .with({ chainName: 'ZkSync Era' })
       .with({ shortName: 'zksync' })
       .with({ features: [FEATURES.COUNTERFACTUAL] as any })
+      .with({ recommendedMasterCopyVersion: '1.4.1' })
       .build(),
     chainBuilder()
       .with({ chainId: '480' })
       .with({ chainName: 'Worldchain' })
       .with({ shortName: 'wc' })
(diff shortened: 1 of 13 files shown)


src/components/common/NetworkSelector/__tests__/NetworkMultiSelector.test.tsx
```diff
@@ -32,31 +32,31 @@ describe('NetworkMultiSelector', () => {
...
       .with({ shortName: 'zksync' })
-      .with({ features: [FEATURES.SAFE_141, FEATURES.COUNTERFACTUAL] as any })
+      .with({ features: [FEATURES.COUNTERFACTUAL] as any })
       .build(),
     chainBuilder()
       .with({ chainId: '480' })
       .with({ chainName: 'Worldchain' })
       .with({ shortName: 'wc' })
-      .with({ features: [FEATURES.SAFE_141, FEATURES.COUNTERFACTUAL, FEATURES.MULTI_CHAIN_SAFE_CREATION] as any })
+      .with({ features: [FEATURES.COUNTERFACTUAL, FEATURES.MULTI_CHAIN_SAFE_CREATION] as any })
```
iamacook, line 59: Nit: as multichain is >=1.4.1 specific, I think we should also set the `recommendedMasterCopyVersion` to `'1.4.1'` in these mocks.

src/components/new-safe/create/logic/index.test.ts
```diff
@@ -301,7 +298,7 @@ describe('create/logic', () => {
           chainBuilder()
             .with({ chainId: '137' })
             // Multichain creation is toggled on
-            .with({ features: [FEATURES.SAFE_141, FEATURES.COUNTERFACTUAL, FEATURES.MULTI_CHAIN_SAFE_CREATION] as any })
+            .with({ features: [FEATURES.COUNTERFACTUAL, FEATURES.MULTI_CHAIN_SAFE_CREATION] as any })
```
iamacook, line 301: Nit: as before regarding mocking the `recommendedMasterCopyVersion` as `'1.4.1'`.

src/features/multichain/utils/utils.ts
```diff
@@ -121,11 +121,7 @@ export const predictAddressBasedOnReplayData = async (safeCreationData: Replayed
 }
 
 export const hasMultiChainCreationFeatures = (chain: ChainInfo): boolean => {
-  return (
-    hasFeature(chain, FEATURES.MULTI_CHAIN_SAFE_CREATION) &&
-    hasFeature(chain, FEATURES.COUNTERFACTUAL) &&
-    hasFeature(chain, FEATURES.SAFE_141)
-  )
+  return hasFeature(chain, FEATURES.MULTI_CHAIN_SAFE_CREATION) && hasFeature(chain, FEATURES.COUNTERFACTUAL)
```
iamacook, line 124: As multichain required >1.4.1, what do you think about including a semver comparison of `recommendedMasterCopyVersion` too?

iamacook: Why do we not include it here and [only `hasMultiChainAddNetworkFeature`](https://github.com/safe-global/safe-wallet-web/pull/4642/commits/9421c23a41cb8dec1cf2fd322ed767909a7c302c#diff-e224730418c4d512f934ecdd399575e7b8786239ada01bcfc9a7632ee7c8d3a4R133)? Without the comparison here, [it will affect the `NetworkMultiSelector`](https://github.com/safe-global/safe-wallet-web/blob/5c1307c082447e93abb9da1bc59bbe1d7c85656e/src/components/common/NetworkSelector/NetworkMultiSelector.tsx#L72). I think our tests would've picked this up if we weren't [only mocking 1.4.1](https://github.com/safe-global/safe-wallet-web/blob/9421c23a41cb8dec1cf2fd322ed767909a7c302c/src/components/common/NetworkSelector/__tests__/NetworkMultiSelector.test.tsx).

PR #4643, opened 2024-12-11
Fix: allow removing counterfactual safes from safe list

### LINKED ISSUE #4611: Sidebar: Remove CF doesn't work from the sidebar
## Bug description
Remove CF doesn't work from the sidebar 
## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet/any

## Steps to reproduce
 1. Unpin the CF safe
 2. disconnect the owner
 3. try to remove the CF safe
 Current result: The safe is not deleted from the app

## Expected result
Make sure that the user can delete the CF safes

jmealy:
## What it solves

Resolves #4611

## How this PR fixes it
- Removes the counterfactual safe from the `undeployedSafes` store as well as from the `addedSafes` store.
- If the safe being removed is the currently opened safe, redirect to the accounts page.

## How to test it
- Create a counterfactual safe
- it should be possible to remove the safe from the context menu of the CF safe in the sidebar.
- If the CF safe being removed is currently opened, redirect to the accounts page.

## Checklist

### DIFF
--- a/src/components/sidebar/SafeListRemoveDialog/index.tsx
+++ b/src/components/sidebar/SafeListRemoveDialog/index.tsx
@@ -12,8 +12,9 @@ import { OVERVIEW_EVENTS, OVERVIEW_LABELS } from '@/services/analytics'
 import { AppRoutes } from '@/config/routes'
 import router from 'next/router'
 import { removeAddressBookEntry } from '@/store/addressBookSlice'
-import { removeUndeployedSafe } from '@/store/slices'
+import { removeSafe, removeUndeployedSafe } from '@/store/slices'
 import useSafeAddress from '@/hooks/useSafeAddress'
+import useChainId from '@/hooks/useChainId'
 
 const SafeListRemoveDialog = ({
   handleClose,
@@ -26,6 +27,7 @@ const SafeListRemoveDialog = ({
 }): ReactElement => {
   const dispatch = useAppDispatch()
   const safeAddress = useSafeAddress()
+  const safeChainId = useChainId()
   const addressBook = useAddressBook()
   const trackingLabel =
     router.pathname === AppRoutes.welcome.accounts ? OVERVIEW_LABELS.login_page : OVERVIEW_LABELS.sidebar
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
👍

src/components/sidebar/SafeListRemoveDialog/index.tsx
```diff
@@ -24,13 +26,20 @@ const SafeListRemoveDialog = ({
...
   const safe = addressBook?.[address] || address
 
-  const handleConfirm = () => {
+  const handleConfirm = async () => {
+    // When removing the current counterfactual safe, redirect to the accounts page
+    if (safeAddress === address && safeChainId === chainId) {
+      await router.push(AppRoutes.welcome.accounts)
+    }
+    dispatch(removeUndeployedSafe({ chainId, address }))
     dispatch(removeSafe({ chainId, address }))
```
usame-algan, line 43: Seems a bit strange that regardless of the type of safe we always call both. I would expect something like so:

```
if (isUndeployedSafe) {
    dispatch(removeUndeployedSafe)
} else {
    dispatch(removeSafe)
}
```

I assume inside the dispatch we don't do anything since we can't find an entry in the respective storage?

katspaugh: We have to do both because the Safe is both "added" and CF.

usame-algan: You are right. In case an undeployed safe is removed while being pinned we need to do both. If its not pinned and removed we only need to remove it from the undeployed storage. Lets keep it as is as long as those dispatch calls don't have unexpected side-effects.

usame-algan (approved):
Looks good! 👍 

I tested the following cases:

1. Remove pinned, undeployed safe while open -> Redirects to the accounts page
2. Remove unpinned, undeployed safe while open -> Redirects to the accounts page
3. Remove pinned, undeployed safe on the accounts page -> Removes it and stays there
4. Remove unpinned, undeployed safe on the accounts page -> Removes it and stays there

PR #4644, opened 2024-12-12
fix: rank featured Safe Apps first on dashboard

iamacook:
## What it solves

Resolves [inconsistent display of featured Safe Apps on dashboard](https://5afe.slack.com/archives/C03CQ5ABLQM/p1733926283956939)

## How this PR fixes it

We display Safe Apps on the dashboard by ranking (the amount of times an App is opened) and whether they are pinned. This does not take into account whether they are featured.

This updates the ranking of Safe Apps to put any which are featured in the initial position(s). Thereafter are only (ranked|) pinned Apps, and the "Explore" card.

## How to test it

Open the dashboard on a network which has featured Safe Apps and observe featured Apps at the beginning of the list on the dashboard.

## Screenshots

image

## Checklist

### DIFF
--- a/src/components/dashboard/SafeAppsDashboardSection/__tests__/SafeAppsDashboardSection.test.tsx
+++ b/src/components/dashboard/SafeAppsDashboardSection/__tests__/SafeAppsDashboardSection.test.tsx
@@ -23,6 +23,7 @@ jest.mock('@safe-global/safe-gateway-typescript-sdk', () => ({
         features: [],
         socialProfiles: [],
         developerWebsite: '',
+        featured: true,
       },
       {
         id: 3,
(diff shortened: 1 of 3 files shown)


katspaugh:
I think this kills the whole point of ranking. There are just 5 slots there and they'll be fully taken by featured apps. Where is this product requirement documented?

I would suggest showing featured apps only _after_ pinned/most used apps to fill up the grid when the former aren't enough.

iamacook:
> I think this kills the whole point of ranking. There are just 5 slots there and they'll be fully taken by featured apps. Where is this product requirement documented?
> 
> I would suggest showing featured apps only _after_ pinned/most used apps to fill up the grid when the former aren't enough.

For visibility, the following has since been [clarified by product](https://5afe.slack.com/archives/C03CQ5ABLQM/p1733997767628739?thread_ts=1733926283.956939&cid=C03CQ5ABLQM):

1. Feature Safe Apps
2. Pinned
3. Explore card

katspaugh (approved):
Now I get it. With both featured and pinned apps ranked, it still makes sense to keep the ranking.

liliya-soroka:
Verified

PR #4645, opened 2024-12-12
Feat: upgrade to 1.4.1

katspaugh:
## What it solves

This PR contains the changes from #4642

* Use the now audited Safe_migration contract to migrate 1.3.0 Safes to 1.4.1
* Move all migration-related functions (including multi-chain l2 migration) to a separate util file
* Instead of the `SAFE_141` feature toggle, use the `recommendedMasterCopyVersion` field from `ChainInfo`

1.3.0 Safes:
Screenshot 2024-12-13 at 14 50 05

1.1.1 Safes:
Screenshot 2024-12-13 at 14 49 14

## How to test

Case 1:
* Open a 1.3.0 Safe on mainnet
* Go to Settings and upgrade it to 1.4.1

Case 2:
* In the config service on staging, pick an L2 chain and set its recommended master copy version to 1.4.1
* Open a 1.3.0 Safe on that chain and upgrade it to 1.4.1

Case 3:
* Open a 1.3.0 Safe with a custom fallback handler (e.g. TWAP)
* Upgrade it to 1.4.1
* Verify that the fallback handler is still installed


src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
```diff
@@ -59,11 +58,9 @@ export const DecodedData = ({ txData, toInfo }: Props): ReactElement | null => {
     decodedData = <HexEncodedData title="Data (hex-encoded)" hexData={txData.hexData} />
   }
 
-  const isL2Migration = isMigrateToL2TxData(txData, chainInfo?.chainId)
-
   return (
     <Stack spacing={2}>
-      {isDelegateCall && <DelegateCallWarning showWarning={!txData.trustedDelegateCallTarget && !isL2Migration} />}
```
katspaugh, line 66: We shouldn't hard-code delegate warning exceptions. These contracts must be added as trusted in tx-service (maybe they have been already).

usame-algan: This was added because users would see an unexpected delegate call warning when they tried to create their first tx and the migration was included. https://github.com/safe-global/safe-wallet-web/pull/4339

If this hasn't changed I would not remove it.

src/services/tx/safeUpdateParams.ts
```diff
@@ -32,6 +34,10 @@ const getChangeFallbackHandlerCallData = async (
 export const createUpdateSafeTxs = async (safe: SafeInfo, chain: ChainInfo): Promise<MetaTransactionData[]> => {
   assertValidSafeVersion(safe.version)
 
+  if (semverSatisfies(safe.version, '>=1.3.0')) {
```
usame-algan, line 37: Are we sure this will also work with future safe version e.g. going from 1.4.1 to 1.5.0? Otherwise I would limit it to only target 1.3.0 versions instead of >=1.3.0

src/utils/chains.ts
```diff
@@ -71,7 +70,7 @@ export const isRouteEnabled = (route: string, chain?: ChainInfo) => {
 }
 
 export const getLatestSafeVersion = (chain: ChainInfo | undefined): SafeVersion => {
-  const latestSafeVersion = chain && hasFeature(chain, FEATURES.SAFE_141) ? LATEST_SAFE_VERSION : FALLBACK_SAFE_VERSION
+  const latestSafeVersion = chain?.recommendedMasterCopyVersion || LATEST_SAFE_VERSION
```
usame-algan, line 73: What if a network doesn't have 1.4.1? Previously we would use `FALLBACK_SAFE_VERSION` which defaulted to 1.3.0 but now this won't be the case anymore.

katspaugh: It will fail somewhere down the line where it gets an actual deployment. 1.3.0 is also not guaranteed to exist on all chains.

src/components/transactions/TxDetails/TxData/DecodedData/index.tsx
katspaugh: Like I wrote above, we should request to add these contracts as "known" and "trusted", it shouldn't be hard-coded.

katspaugh: @usame-algan looks like it [was already added](https://github.com/safe-global/safe-transaction-service/blob/ec11de08be6acf5b110de4a8680f526ae56650ed/safe_transaction_service/contracts/management/commands/setup_safe_contracts.py#L13) so there should be no warning.

usame-algan (approved):

PR #4652, opened 2024-12-13
refactor: Extract useAllSafesGrouped logic and write tests [SW-647]

usame-algan:
## What it solves

Resolves [SW-647](https://www.notion.so/safe-global/Refactor-Sidebar-components-1548180fe57380d08e77d05a3a121aec)

## How this PR fixes it

- Extracts the logic from `useAllSafesGrouped` into a function and tests it

## Checklist

### DIFF
--- a/src/features/myAccounts/hooks/useAllSafesGrouped.ts
+++ b/src/features/myAccounts/hooks/useAllSafesGrouped.ts
@@ -16,18 +16,28 @@ export type AllSafesGrouped = {
   allMultiChainSafes: MultiChainSafeItem[] | undefined
 }
 
-const getMultiChainAccounts = (safes: SafeItems): MultiChainSafeItem[] => {
+export const _buildMultiChainSafeItem = (address: string, safes: SafeItems): MultiChainSafeItem => {
+  const isPinned = safes.some((safe) => safe.isPinned)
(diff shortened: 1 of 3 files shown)


src/features/myAccounts/hooks/useAllSafesGrouped.ts
```diff
@@ -16,18 +16,28 @@ export type AllSafesGrouped = {
   allMultiChainSafes: MultiChainSafeItem[] | undefined
 }
 
-const getMultiChainAccounts = (safes: SafeItems): MultiChainSafeItem[] => {
+export const _buildMultiChainSafeItem = (address: string, safes: SafeItems): MultiChainSafeItem => {
+  const isPinned = safes.some((safe) => safe.isPinned)
+  const lastVisited = safes.reduce((acc, safe) => Math.max(acc, safe.lastVisited || 0), 0)
+  const name = safes.find((safe) => safe.name !== undefined)?.name
```
usame-algan, line 22: Don't think it makes a huge difference but we could also optimize this by only iterating once instead of three times.

jmealy: Yeah I think it's ok like this

PR #4654, opened 2024-12-13
Fix: pass sx prop to Box and Typography

katspaugh:
## What it solves

The `sx` prop wasn't passed down to MUI Box and Typography because it was overwritten by individual props.


usame-algan (approved):
Is it intended that when passing the sx prop it would overwrite whatever inline prop is set?

PR #4656, opened 2024-12-13
Fix: add chain check in Delete Tx modal

katspaugh:
## What it solves

This modal was lacking a check for the current chain causing cryptic error messages.

Screenshot 2024-12-13 at 12 53 00


src/components/tx-flow/flows/ReplaceTx/DeleteTxModal.tsx
```diff
@@ -136,17 +142,21 @@ const InternalDeleteTxModal = ({
...
+        <CheckWallet checkNetwork>
+          {(isOk) => (
+            <Button
+              data-testid="delete-tx-btn"
+              size="small"
+              variant="contained"
+              color="primary"
+              onClick={onConfirm}
+              disabled={!isOk || isLoading}
+              sx={{ minWidth: '122px', minHeight: '36px' }}
```
clovisdasilvaneto, line 154: can we extend this component to avoid "inline" styles?

clovisdasilvaneto (approved):

francovenica:
Gotta say never thought about this. Well spotted.

Looks good to me.

PR #4659, opened 2024-12-13
fix: Add sorting back to useAllSafes hook

usame-algan:
## What it solves

We should return safes sorted from `useAllSafes` after all otherwise there is a jump when pinning a safe while there are other safes in the same network

## How this PR fixes it

- Adds back the sorting of safes that was removed in #4641

## How to test it

1. Open the safe list
2. Pin a safe thats not the first one from its chain
3. Observe no jump

## Checklist

### DIFF
--- a/src/features/myAccounts/hooks/useAllSafes.ts
+++ b/src/features/myAccounts/hooks/useAllSafes.ts
@@ -33,7 +33,7 @@ export const _prepareAddresses = (
 
   const combined = [...addedOnChain, ...ownedOnChain, ...undeployedOnChain]
 
-  return [...new Set(combined)]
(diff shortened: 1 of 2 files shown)


jmealy (approved):

src/features/myAccounts/hooks/useAllSafes.ts
```diff
@@ -33,7 +33,7 @@ export const _prepareAddresses = (
 
   const combined = [...addedOnChain, ...ownedOnChain, ...undeployedOnChain]
 
-  return [...new Set(combined)]
+  return [...new Set(combined)].sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()))
```
jmealy, line 36: might be worth adding a comment explaining why we do this, since it's not obvious

PR #4666, opened 2024-12-16
fix: Revert add read-only button naming

usame-algan:
## What it solves

Reverts change introduced in https://github.com/safe-global/safe-wallet-web/pull/4657

## How this PR fixes it

- Reverts the naming of the read-only button in the sidebar to "Add read-only" instead of "Add as Read-only"

## How to test it

1. Open a Safe without a wallet connected
2. Observe the button in the sidebar says "Add read-only"

## Checklist

### DIFF
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -2,4 +2,4 @@
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/pages/building-your-application/configuring/typescript for more information.
+// see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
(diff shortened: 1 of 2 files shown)


next-env.d.ts
```diff
@@ -2,4 +2,4 @@
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/pages/building-your-application/configuring/typescript for more information.
+// see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
```
usame-algan, line 5: This was an automatic change

katspaugh (approved):

PR #4668, opened 2024-12-16
Fix: box border color in tx details

katspaugh:
## What it solves

Border color wasn't properly passed to a Box because of the prop order.

I've also added omitBy to exclude unnecessary undefined properties passed to sx.

Screenshot 2024-12-16 at 14 53 11


usame-algan:
Can you include a before screenshot please?

usame-algan (approved):
Not sure why omitBy and isUndefined are needed. Is it causing performance issues without?

PR #4671, opened 2024-12-16
refactor: Split up MyAccounts component

usame-algan:
## What it solves

Part of [SW-647](https://www.notion.so/safe-global/Refactor-Sidebar-components-1548180fe57380d08e77d05a3a121aec)

## How this PR fixes it

- Renames `AccountsList` to `MyAccounts`
- Creates a new component called `AccountsList`
- Creates `AllSafes`, `PinnedSafes`, `FilteredSafes` components and moves logic
- Creates `AccountsFilter` and `AccountsHeader` components and moves logic
- Renames type `AllSafesGrouped` to `AllSafeItemsGrouped`
- Creates new type callde `AllSafeItems`
- Removes the sorting from filtered safe items
- Changes `QueueActions` to be a button instead of an a element to resolve react error

## ToDos

- [x] Revisit watchlist and pinned safes logic inside `useTrackedSafesCount`
- [x] Write tests for each component

## How to test it

1. Open the accounts page
2. Connect a wallet
3. Observe Safes are listed as before
4. Pin and unpin should still work
5. Pinned safes should be tracked
6. Sidebar should still work as before

## Checklist


src/features/myAccounts/hooks/useTrackedSafesCount.ts
usame-algan, line 1: @jmealy I remember we talked about how we handle pinned and watchlist safes. In this hook we say watchlist safes are safes that are not pinned and read-only. I think this doesn't work anymore since an unpinned safe that is not owned or undeployed is removed from the list so watchlist === pinned.

src/features/myAccounts/components/AccountsFilter/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+                  }}
+                />
+              </InputAdornment>
+            ),
+            disableUnderline: true,
+          }}
+          fullWidth
+          size="small"
+        />
+        <OrderByButton orderBy={orderBy} onOrderByChange={handleOrderByChange} />
```
jmealy, line 54: Took me a minute to find the OrderByButton component here. To me `AccountsFilter`,  suggests it only contains the search. Could we rename to reflect that it has both search and sort?  `AccountListFilters` or something

src/features/myAccounts/hooks/useTrackedSafesCount.ts
```diff
@@ -46,6 +42,15 @@ const useTrackSafesCount = (
     [safes, watchlistMultiChainSafes],
   )
 
+  // TODO: This is computed here and inside PinnedSafes now. Find a way to optimize it
```
jmealy, line 45: A small improvement could be to extract the calculations of allSafes and pinnedSafes into helper functions since they are both done twice, but nbd. IMO it's ok as is

jmealy (approved):
Nice refactoring 🚀

PR #4673, opened 2024-12-16
fix: Refetch owned safes after deployment

usame-algan:
## What it solves

Fixes [issue found in the release](https://github.com/safe-global/safe-wallet-web/pull/4667#issuecomment-2545810281)

## How this PR fixes it

- Calls `refetch` on the RTK query for all owned safes when receiving a `SafeCreationEvent.INDEXED`

## How to test it

1. Create a safe with the pay later option
2. Activate the Safe
3. Open the Sidebar
4. Observe there is no "read-only" chip next to the safe item

## Checklist

### DIFF
--- a/src/features/counterfactual/hooks/usePendingSafeNotifications.ts
+++ b/src/features/counterfactual/hooks/usePendingSafeNotifications.ts
@@ -1,5 +1,8 @@
 import { SafeCreationEvent, safeCreationSubscribe } from '@/features/counterfactual/services/safeCreationEvents'
+import useWallet from '@/hooks/wallets/useWallet'
+import { useGetAllOwnedSafesQuery } from '@/store/api/gateway'
 import { getBlockExplorerLink } from '@/utils/chains'
+import { skipToken } from '@reduxjs/toolkit/query'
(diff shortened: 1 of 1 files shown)


src/features/counterfactual/hooks/usePendingSafeNotifications.ts
```diff
@@ -26,6 +29,8 @@ const usePendingSafeNotifications = (): void => {
   const dispatch = useAppDispatch()
   const chain = useCurrentChain()
   const safeAddress = useSafeAddress()
+  const { address = '' } = useWallet() || {}
+  const { refetch } = useGetAllOwnedSafesQuery(address === '' ? skipToken : { walletAddress: address })
```
usame-algan, line 33: Its probably cleaner to create a new listener middleware for this. Wdyt @katspaugh @compojoom

usame-algan: Although it does require access to the connected wallet which I am not sure is possible inside a middleware. We would have to move the connected wallet state into Redux as well.

katspaugh (approved):

PR #4674, opened 2024-12-16
fix: Cap max expiry for setTimeout in useIsExpiredSwap

### LINKED ISSUE #4650: Swaps with a large expiry appear as expired
## Bug description

When creating a swap that expires in the far future the transaction displays an `Expired` status. This is most likely due to the ms time we pass to `setTimeout` inside `useIsExpiredSwap` which can't handle numbers that are too big.

https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#maximum_delay_value

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Create a limit swap with a large expiry time
 2. Go to the transaction
 3. Observe it shows Expired as the status

## Expected result

It should not display Expired

## Obtained result

It displays Expired

## Screenshots
Screenshot 2024-12-12 at 16 00 34

usame-algan:
## What it solves

Resolves #4650

## How this PR fixes it

- If the `validUntil` value is too far in the future (larger than a custom MAX_TIMEOUT value) we use MAX_TIMEOUT instead for the setTimeout value

## How to test it

1. Create a limit order swap that expires far in the future (25 days+)
2. Observe no expired label
3. Create a limit order that expires not so far in the future (< 25 days)
4. Observe no expired label
5. Create a limit order that expires short in the future
6. Observe the expired label shows up after it is expired

## Checklist

### DIFF
--- a/src/features/swap/hooks/__tests__/useIsExpiredSwap.test.ts
+++ b/src/features/swap/hooks/__tests__/useIsExpiredSwap.test.ts
@@ -1,0 +1,106 @@
+import { act } from 'react'
+import useIsExpiredSwap from '@/features/swap/hooks/useIsExpiredSwap'
+import { renderHook } from '@/tests/test-utils'
+import * as guards from '@/utils/transaction-guards'
+import type { TransactionInfo } from '@safe-global/safe-gateway-typescript-sdk'
+
+describe('useIsExpiredSwap', () => {
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


katspaugh (approved):

liliya-soroka:
Verified

PR #4686, opened 2024-12-18
refactor: switch to Yarn v4

compojoom:
## What it solves
The first step to switching to a monorepo structure is to update yarn as newer versions offer:
- improved speed
- better monorepo management

We were stuck with yarn classic (which is deprecated for more than a couple of years now). Yarn 4 feels significantly faster and will lead to a better developer experience

## How this PR fixes it
I've added a packagerManager field to the yarn.lock file. Had to dedupe some dependencies as they were wrong in yarn v1 and also had to remove `@web3-onboard/keystone` as it has one incorrectly defined dependency (the dependency version is not following semantic versioning and I couldn't find a way for yarn to ignore this). I have created a fix for it https://github.com/blocknative/web3-onboard/pull/2322, but I doubt that it will be merged any time soon. 

About github actions. 
Our github actions were incorrectly placed in the workflows folder. I moved them to a separate github actions folder. This also made it possible to use tools such as act https://github.com/nektos/act to tests the flows locally. 

The caching dependency action is a bit weird, but I had my reasons :). Using the `actions/cache@v4` directly turned out to be very slow. I think that the main issue is the use of hashFiles in the key. For example restoring next.js build cache the key there is composed of `hashFiles('**/*.js', '**/*.jsx', '**/*.ts', '**/*.tsx')` - we are dealing with a lot of files here and that apparently takes a lot of time. That's why I'm caching the key once it is generated by the restore action and then using it for the save action.
 
## How to test it
Checkout the branch, make sure that you have run `corepack enable` and then just run `yarn` in the terminal. This should install the latest yarn v4 version and all necessary project dependencies.

## Checklist


.github/actions/cache-deps/action.yml
```diff
@@ -0,0 +1,85 @@
...
+
+    - name: Set composite outputs nc
+      if: ${{ inputs.mode == 'restore-nc' }}
+      shell: bash
+      run: | 
+        echo "cache-hit-nc=${{ steps.restore-nc.outputs.cache-hit }}" >> $GITHUB_OUTPUT
+        echo "computed-cache-key-nc=${{ steps.restore-nc.outputs.cache-primary-key }}" >> $GITHUB_OUTPUT
+
+
+    - name: Save Nextj.js
```
usame-algan, line 79:
```suggestion
    - name: Save Next.js
```

usame-algan (approved):
Should we make an announcement that Keystone is going to be removed?

package.json
```diff
@@ -26,21 +26,21 @@
     "cypress:canary": "cross-env TZ=UTC cypress open --e2e -b chrome:canary",
     "cypress:run": "cypress run",
     "cypress:ci": "yarn cypress:run --config baseUrl=http://localhost:8080 --spec cypress/e2e/smoke/*.cy.js",
-    "serve": "npx -y serve out -p ${REVERSE_PROXY_UI_PORT:=8080}",
+    "serve": "sh -c 'npx -y serve out -p ${REVERSE_PROXY_UI_PORT:=8080}'",
```
katspaugh, line 29: Why do you need this?

package.json
```diff
@@ -70,7 +70,6 @@
     "@web3-onboard/coinbase": "^2.4.1",
     "@web3-onboard/core": "^2.21.4",
     "@web3-onboard/injected-wallets": "^2.11.2",
-    "@web3-onboard/keystone": "^2.3.7",
```
katspaugh, line 73: Looks like we won't be migrating to AppKit until after Q1 2025, so we'll eventually have to find a way to restore this package.

katspaugh (approved):
🤘

package.json
compojoom: Because it was choking on the curly brackets. It seem that they are invalide in their Posix-like interpreter 🤷 
https://yarnpkg.com/features/scripting#script-arguments

that's why I'm starting a shell and passign the command to it.

PR #4679, opened 2024-12-17
fix: always estimate relay `gasLimit`

iamacook:
## What it solves

Resolves inconsistent relays on Arbitrum

## How this PR fixes it

When relaying, a `gasLimit` can be optionally provided in the request. If provided, [a 150k buffer is added to it](https://github.com/safe-global/safe-client-gateway/blob/b7640ed4bd7416ecb7696d21ba105fcb5af52a10/src/datasources/relay-api/gelato-api.service.ts#L62-L75) for the Gelato execution overhead. If none is provided, Gelato estimates this on our behalf and no buffer is added.

This explicitly provides the `gasLimit` to relay requests (if present), otherwise manually estimates it if none is provided.

## How to test it

Ensure relay attempts do not fail (particularly on Arbitrum) for the following cases, and that the `gasLimit` is present in each `POST` request to the `/v1/chains/:chainId/relay`:

1. Safe creation/activation - uses estimation from UI
2. Bulk execution - estimation occurs in the background

## Checklist
* [ ] I've tested the branch on mobile 📱
* [ ] I've documented how it affects the analytics (if at all) 📊
* [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #4678, opened 2024-12-17
Fix: add Base stablecoins for swaps

katspaugh:
## What it solves

Resolves #4649.

Top stable coins on Base, data from https://basescan.org/tokens

PR #4416, opened 2024-10-24
Feat: natively support clear signing with Ledger

iamacook:
## What it solves

The inability to perform clear signing with Ledger devices.

## How this PR fixes it

This replaces the [outdated `@web3-onboard/ledger`](https://www.npmjs.com/package/@web3-onboard/ledger/v/2.3.2) module, which relied on an [archived Ledger package](https://github.com/LedgerHQ/ledgerjs), with a custom module built on Ledger's new [Web Management Device Kit](https://github.com/LedgerHQ/device-sdk-ts).

## How to test it

Connect a Ledger device and test each RPC method (which corresponds to each SDK method):

- Successful connection (accounts)
- Signing a message (signing)
- Signing a transaction (signing typed data)
- Executing a transaction (sending a transaction)

## Checklist
* [ ] I've tested the branch on mobile 📱
* [ ] I've documented how it affects the analytics (if at all) 📊
* [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #4688, opened 2024-12-18
Release 1.48.1

katspaugh:
* fix: always estimate relay `gasLimit` (#4679)
* Fix: add chain check in Delete Tx modal (#4656)
* Fix: add Base stablecoins for swaps (#4678)
* Feat: natively support clear signing with Ledger (#4416)


liliya-soroka:
Verified

PR #4690, opened 2024-12-18
Monorepo

compojoom:
## What it solves

Resolves #

## Checklist


README.md
```diff
@@ -0,0 +1,158 @@
...
+Example:
+
+```bash
+yarn workspace @safe-global/web start
+```
+
+- **Add a dependency to a specific workspace:**
+
+```bash
+yarn workspace <workspace-name> add <package-name>
```
usame-algan, line 79: Is it not possible to cd into the app and run `yarn add <package-name>` there directly?

README.md
```diff
@@ -0,0 +1,158 @@
...
+
+- **Remove a dependency from a specific workspace:**
+
+```bash
+yarn workspace <workspace-name> remove <package-name>
+```
+
+> [!Note]
+>
+> Yarn treats commands that contain a semicolon as global commands. For example if you have a
```
usame-algan, line 90: should be colon instead of semicolon right?

katspaugh (approved):

README.md
compojoom: thanks, I fixed this in the #4698 PR.

README.md
compojoom: No. When you execute the command it would run in the package directory context (meaning root will be the directory). Sometimes the output coloring is a bit off and yarn is adding some prefixes to every line (but I can't tell you when this exactly occurs)

PR #4691, opened 2024-12-18
docs: Clarify the purpose of the remark-heading-id Update update-term…

mdqst:
## What it solves

This update improves the clarity of the documentation by expanding on why the `remark-heading-id` plugin is required. Previously, the statement:  

> "Note that because the Markdown was generated from Google Docs, we require the remark-heading-id plugin."  

was somewhat incomplete and could lead to confusion for readers unfamiliar with the context.  

The updated text now explicitly states:  

> "Note that because the Markdown was generated from Google Docs, we require the remark-heading-id plugin to ensure proper handling of heading IDs generated by Google Docs."  

This small change enhances readability and provides better guidance for developers working with this part of the project.  

## How this PR fixes it

- Expanded the explanation of the `remark-heading-id` plugin's purpose in the "How to update Terms & Conditions" documentation.  

## Checklist


katspaugh:
Thanks, we try to avoid merging tiny edits like this. Good bot though.

PR #4693, opened 2024-12-19
Feat: public transaction notes

katspaugh:
## What it solves

Resolves [SW-700](https://www.notion.so/safe-global/Public-transaction-notes-16d8180fe573808ea865f0028559dd2b)

This PR adds the ability to add a public note when creating a transaction. It's sent to the backend in the previously unused `origin.name` field as JSON.

The feature is behind a feature flag called `TX_NOTES`.

### Adding a note
Screenshot 2024-12-27 at 12 19 49

### Displaying a note
Screenshot 2024-12-27 at 12 18 22

## How to test

1.
* Create a transaction
* Attach a note
* Submit the tx

2.
* Open a tx with a note in the queue
* Confirm a tx with a note (it should not be editable!)


src/components/transactions/TxDetails/TxNote.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const txService = currentChain?.transactionService
+
+  let safeTxHash = ''
+  if (isMultisigDetailedExecutionInfo(txDetails.detailedExecutionInfo)) {
+    safeTxHash = txDetails.detailedExecutionInfo?.safeTxHash
+  }
+
+  const [data] = useAsync(() => {
+    if (!safeTxHash || !txService) return
+    return fetch(`${txService}/api/v1/multisig-transactions/${safeTxHash}`).then((res) => res.json())
```
katspaugh, line 19: I'm using the tx-service directly here because CGW doesn't mirror the `origin` field. It can be easily added there.

apps/web/src/components/transactions/TxDetails/TxNote.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  let safeTxHash = ''
+  if (txDetails && isMultisigDetailedExecutionInfo(txDetails.detailedExecutionInfo)) {
+    safeTxHash = txDetails.detailedExecutionInfo?.safeTxHash
+  }
+
+  const [data] = useAsync(() => {
+    if (!safeTxHash || !txService) return
+    return fetch(`${txService}/api/v1/multisig-transactions/${safeTxHash}`).then((res) => res.json())
+  }, [safeTxHash, txService])
```
katspaugh, line 20: This will be replaced by https://github.com/safe-global/safe-client-gateway/pull/2223

apps/web/src/components/transactions/TxDetails/TxNote.tsx
```diff
@@ -0,0 +1,26 @@
...
+          </Typography>
+        </Tooltip>
+      </Typography>
+
+      <Typography p={2} mt={1} borderRadius={1} bgcolor="background.main">
+        {note}
+      </Typography>
+    </div>
+  ) : null
+}
```
clovisdasilvaneto, line 24: can we adopt early return here? So we avoid having another ident level inside the component
```suggestion
 if(!note) return null;

  return (
    <div>
      <Typography variant="h5" display="flex" alignItems="center" justifyItems="center">
        Note
        <Tooltip title="This note is left by the transaction creator." arrow>
          <Typography color="text.secondary" component="span" height="1em">
            <InfoIcon height="1em" />
          </Typography>
        </Tooltip>
      </Typography>

      <Typography p={2} mt={1} borderRadius={1} bgcolor="background.main">
        {note}
      </Typography>
    </div>
  )
}
```

apps/web/src/components/transactions/TxDetails/TxNote.tsx
```diff
@@ -0,0 +1,26 @@
...
+const TxNote = ({ txDetails }: { txDetails: TransactionDetails | undefined }) => {
+  const note = (txDetails as TransactionDetails & { note: string | null })?.note
+
+  return note ? (
+    <div>
+      <Typography variant="h5" display="flex" alignItems="center" justifyItems="center">
+        Note
+        <Tooltip title="This note is left by the transaction creator." arrow>
+          <Typography color="text.secondary" component="span" height="1em">
+            <InfoIcon height="1em" />
```
clovisdasilvaneto, line 14: maybe moving the 1em to a constant and re-using it in both places?

apps/web/src/components/tx/SignOrExecuteForm/SignOrExecuteForm.tsx
```diff
@@ -124,6 +127,55 @@ export const SignOrExecuteForm = ({
...
+          safeTxError={safeTxError}
+          onSubmit={onRoleExecutionSubmit}
+        />
+      )
+    }
+
+    if (!isCounterfactualSafe && !willExecute && !willExecuteThroughRole && !isProposing) {
+      return <SignForm {...commonProps} isBatchable={isBatchable} />
+    }
+
```
clovisdasilvaneto, line 168: One suggestion to improve this bunch of "ifs" later would be adopt an buiilder pattern here. Where you could have an mapper where the key would be the builder name and the value the component you wanna build. And to render the builder you could use a director to receive the builder name and map it according to the value you find inside the map... Just an idea. not a blocker though.

For reference https://refactoring.guru/design-patterns/builder

clovisdasilvaneto (approved):

apps/web/src/utils/transactions.ts
```diff
@@ -240,3 +240,19 @@ export const getSafeTransaction = async (safeTxHash: string, chainId: string, sa
     return undefined
   }
 }
+
+export const encodeTxNote = (note: string, origin?: string): string => {
+  let originalOrigin = { url: '' }
+  if (origin) {
+    try {
+      originalOrigin = JSON.parse(origin)
+    } catch {}
```
clovisdasilvaneto, line 249: shouldn't we log anything in the catch block?

francovenica:
@katspaugh The deployment has failed. Can you try again?

katspaugh:
@francovenica fixed! ✅

francovenica:
Issue:

When a tx is just signed and then executed by another owner, the "Note" field shows up again.
- With ownerA create a tx with a note
- With ownerB try to execute it
- The note field is visible during execution

The filed should not show up or show as a read only input

francovenica:
Issue:

Notes created by proposers don't show in the tx. I'm not sure if it is possible since the user is just proposing the tx so it's only in the backend.
If is possible the note added by the proposer should show. 
If it is not possible (because of the nature of the type of user) then we should remove the field when a tx is just proposed with no signature

katspaugh:
Fixed both, thank you. ✅ 
Notes now work for proposers too.

francovenica:
New issue I just noticed

Tx cannot be "Added to a batch". I tried in dev with the same type of tx and it is possible there, so it seems is only the changes done here that are causing this.

Could the note field make it impossible to add to a batch?
If the issues is the notes, maybe just allow "noteless" tx to be added to a batch

image

francovenica:
Looks good now!

apps/web/src/features/tx-notes/TxNoteInput.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const [note, setNote] = useState('')
+
+  const onInput = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+    setNote(e.target.value)
+  }, [])
+
+  const onChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      onSubmit(e.target.value.slice(0, MAX_NOTE_LENGTH))
+      trackEvent(MODALS_EVENTS.ADD_TX_NOTE)
```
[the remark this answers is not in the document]
katspaugh: Yes, absolutely correct. Fixed here: https://github.com/safe-global/safe-wallet-monorepo/pull/4771

PR #4694, opened 2024-12-19
Fix: display Safe App icon and name in history

katspaugh:
## What it solves

We weren't displaying Safe App info in the queue/history due to a regression. Safe App txs had default icons and names.

Before this PR:
Screenshot 2024-12-19 at 14 08 06

After:
Screenshot 2024-12-19 at 14 07 52


usame-algan (approved):
👍

PR #4697, opened 2024-12-19
Fix: rm circular dependency

katspaugh:
## What it solves

#4645 introduced a circular dependency: transaction-guards imported something from safe-migration utils which indirectly imports transaction-guards.

I've extracted the common part into a separate module.


compojoom (approved):
👍

PR #4708, opened 2024-12-23
Fix(Safe upgrade): preserve fallback handler + show queue warning only for <1.3.0

katspaugh:
## What it solves

* Preserve an existing fallback handler, or set the default one if none is present.
* Don't show a queue warning when upgrading 1.3.0 Safes.

Screenshot 2024-12-23 at 14 37 36


schmanu (approved):
Looks good to me :)

liliya-soroka:
Verified

PR #4725, opened 2025-01-03
Feat(sidebar): change safe search placeholder text

jmealy:
## What it solves
It is not clear that the search bar can be used for filtering safe names, addresses, and chains all using the same input

## How this PR fixes it
Changes the placeholder text to explain what the search can be used for. 

## Screenshots
image

## Checklist

### DIFF
--- a/apps/web/src/features/myAccounts/components/AccountListFilters/index.tsx
+++ b/apps/web/src/features/myAccounts/components/AccountListFilters/index.tsx
@@ -23,7 +23,7 @@ const AccountListFilters = ({ setSearchQuery }: { setSearchQuery: Dispatch<SetSt
       <Box display="flex" justifyContent="space-between" width="100%" gap={1}>
         <TextField
           id="search-by-name"
-          placeholder="Search"
+          placeholder="Search Safes by name, address, or chain"
           aria-label="Search Safe list by name"
           variant="filled"
           hiddenLabel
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
👍

apps/web/src/features/myAccounts/components/AccountListFilters/index.tsx
```diff
@@ -23,7 +23,7 @@ const AccountListFilters = ({ setSearchQuery }: { setSearchQuery: Dispatch<SetSt
       <Box display="flex" justifyContent="space-between" width="100%" gap={1}>
         <TextField
           id="search-by-name"
-          placeholder="Search"
+          placeholder="Search Safes by name, address, or chain"
```
katspaugh, line 26: We usually call them `Safe Accounts` though.

katspaugh: Also maybe we should say "chain prefix" instead of just chain?

jmealy: the search also matches the whole chain name, not just the prefix. So searching "gon" will give polygon safes. Could say "chain name" to make it clearer?

katspaugh: Ah I see. Then chain is clear!

PR #4726, opened 2025-01-03
Fix: larger coingecko icons

katspaugh:
## What it solves

* Token icons from Coingecko are now `referrerPolicy="no-referrer"`
* I also made them larger (50x50) because they were too low-res (25x25)

### DIFF
--- a/apps/web/src/components/common/TokenIcon/index.tsx
+++ b/apps/web/src/components/common/TokenIcon/index.tsx
@@ -1,9 +1,8 @@
-import { memo, type ReactElement } from 'react'
+import { useMemo, type ReactElement } from 'react'
 import ImageFallback from '../ImageFallback'
 import css from './styles.module.css'
 
 const FALLBACK_ICON = '/images/common/token-placeholder.svg'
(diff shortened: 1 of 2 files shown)


apps/web/src/components/common/TokenIcon/index.tsx
```diff
@@ -15,15 +18,26 @@ const TokenIcon = ({
...
       className={css.image}
+      loading="lazy"
+      referrerPolicy={crossOrigin ? 'no-referrer' : undefined}
+      crossOrigin={crossOrigin ? 'anonymous' : undefined}
     />
   )
 }
 
-export default TokenIcon
+export default memo(TokenIcon)
```
clovisdasilvaneto, line 43: small question, do we really need to use memo here? Because since this component just renders an image, I'm don't think it would be a problem in this case

katspaugh: Because I added a replace there. There can be hundreds of icons on each history page so it’s better not to rerender them.

PR #4730, opened 2025-01-05
fix(Nested signing): do not copy addresses in SignerForm

schmanu:
## What it solves
Clicking directly on the address in the `SignerForm` copied the address instead of chosing the signer.

## How this PR fixes it
Disables copying the address in the `SignerForm`

## How to test it
- Create a nested Safe tx
- Try to choose a signer which is not in the address book and click on the address

## Checklist


katspaugh (approved):
👍

PR #4737, opened 2025-01-06
WIP: Adjust meta tag to fix PWA header position

usame-algan:
DO NOT MERGE


usame-algan:
Adding viewport-fit=cover results in this which is not what we want

image

PR #4738, opened 2025-01-06
fix: Adjust meta tag to fix PWA header position

### LINKED ISSUE #4434: Page content is visible above the header on mobile view
## Bug description

When scrolling the app on a mobile viewport, the content is visible above the header

## Steps to reproduce
 1. Open the app on a mobile screen
 2. Scroll down
 3. Observe the content is visible above the header

## Expected result

The header should be at the top and no content visible above it

## Screenshots
IMG_1163

usame-algan:
## What it solves

Resolves #4434 

## How this PR fixes it

- Adds the `viewport-fit=cover` directive to the meta tags -> This moves the whole app up into the "Dynamic island" on iOS
- Adds a padding (for PWA only) to the header and main element to move them down again and adds a background so the content is not visible when scrolling

## How to test it

1. Open the app on mobile (iOS and Android and iOS with a device without the notch), add it as a PWA to your Home Screen and open it again
2. Scroll up and down
3. Observe nothing is visible above the header
4. Open the app on web
5. Scroll up and down
6. Observe the header is at the very top
7. Make sure it works both in landscape and portrait mode on mobile

## Screenshots

image

image

## Checklist

### DIFF
--- a/apps/web/src/components/common/MetaTags/index.tsx
+++ b/apps/web/src/components/common/MetaTags/index.tsx
@@ -30,7 +30,7 @@ const MetaTags = ({ prefetchUrl }: { prefetchUrl: string }) => (
     <link rel="preconnect" href={prefetchUrl} crossOrigin="" />
 
     {/* Mobile tags */}
-    <meta name="viewport" content="width=device-width" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
     <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent" />
 
     {/* PWA primary color and manifest */}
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
👍

francovenica:
I'm trying to reproduce the issue first in current prod and dev and I don't see it, so I cannot confirm the fix does what it is supposed to do.
Am I'm missing something?

Check this gif. It was taken in a chrome emulator, but it looks the same in an iPhone and Android device:
safe twitching

usame-algan:
> I'm trying to reproduce the issue first in current prod and dev and I don't see it, so I cannot confirm the fix does what it is supposed to do. Am I'm missing something?

This needs to be tested on an actual phone, not in the Browser with a mobile emulator

francovenica:
Ok, The app needed to be added to the homepage as a "stand alone app" from the browser. 
Thanks Usame for the clarification 

I was able to compare current prod with this PR and the fix looks good

PR #4740, opened 2025-01-07
Chore: fix docker build

katspaugh:
## What it solves

The yarn.lock file is in the root of the monorepo, so the Dockerfile should also be there.


compojoom:
that is weird. This docker file is for the web app, so it doesn't make sense to have it at the root of the repository.

PR #4744, opened 2025-01-07
fix: Enable Add proposer button for owners only

### LINKED ISSUE #4743: Add proposer button is enabled for nested safe signers
## Bug description

When connected with a wallet that is the owner of a nested safe the Add proposer button is enabled but it is not possible to add a proposer through a nested safe signature so the button needs to be disabled.

## Steps to reproduce

1.  Open a Safe A that has another Safe B as owner
2. Connect with a wallet that is an owner of Safe B
3. Go to the settings page

## Expected result

The Add proposer button is enabled

## Obtained result

The Add proposer should be disabled

usame-algan:
## What it solves

Resolves #4743 

## How this PR fixes it

- Adds a new HoC called `OnlyOwner` and uses it for the Add proposer button instead of the `CheckWallet` component

## How to test it

1. Open a Safe
2. Go to the Settings page
3. Connect with an owner
4. The Add proposer button is enabled
5. Connect with a different wallet e.g. spending limit beneficiary, proposer, nested safe owner, recoverer
6. Observe that the add proposer button is disabled

## Checklist

### DIFF
--- a/apps/web/src/components/common/OnlyOwner/index.tsx
+++ b/apps/web/src/components/common/OnlyOwner/index.tsx
@@ -9,6 +9,7 @@ type CheckWalletProps = {
 }
 
 enum Message {
+  WalletNotConnected = 'Please connect your wallet',
   NotSafeOwner = 'Your connected wallet is not a signer of this Safe Account',
 }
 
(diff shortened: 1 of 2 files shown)


usame-algan:
I decided to write a new component instead of adjusting `CheckWallet` because its already way too complex and I don't want to take the risk of breaking another place where `CheckWallet` is used.

tmjssz:
Looks good to me. 
Perhaps we could also use the new OnlyOwner component in other places that are connected to the proposer logic? e.g. `UpsertProposer`

tmjssz (approved):

francovenica:
Isssue: 
When I'm not connected it tells me that I have a wallet connecte that is not a signer:
image

The rest looks good

usame-algan:
> When I'm not connected it tells me that I have a wallet connecte that is not a signer:

Pushed a fix. It should now say to connect a wallet

francovenica:
The fix works fine.

LGTM

PR #4750, opened 2025-01-08
fix: initialize Firebase only if `projectId` is present in options

tmjssz:
## What it solves

Fixes error message when Firebase is initialized:
> FirebaseError: Installations: Missing App configuration value: "projectId" (installations/missing-app-config-values).

Screenshot 2025-01-08 at 16 30 31

## How this PR fixes it
[`apps/web/src/services/push-notifications/firebase.ts`](diffhunk://#diff-3be0086de4433400ec5c7c60f44c1230c7bc6da197c50edb0b8acbd480181378L22-R28): Added the `isFirebaseOptions` type guard function to ensure that the `FIREBASE_OPTIONS` object contains a `projectId`.

## How to test it (on localhost)
1. Clear `NEXT_PUBLIC_FIREBASE_OPTIONS_STAGING` in .env file
2. Load app while monitoring the dev console
3. Observe that the error message (see above) is not logged

## Checklist


katspaugh (approved):
Thanks!

PR #4753, opened 2025-01-10
Refactor: less css in tx summary

katspaugh:
## What it solves

Just a mechanical refactoring of TxSummary to use less inline styles.


usame-algan (approved):
👍

PR #4756, opened 2025-01-12
Feat(Mobile): add logger and MMKV wrapper

Jonathansoufer:
## What it solves

This PR implements a base reusable wrapper on console.log (on steroids) as well as a MMKV one. Both will be crucial to the implementation of the upcoming Notifications feature as well as localization when the time comes.

Resolves #

## Checklist

### DIFF
--- a/apps/mobile/src/store/storage.ts
+++ b/apps/mobile/src/store/storage.ts
@@ -23,7 +23,7 @@ export const reduxStorage: Storage = {
 }
 
 export class safeMMKVStorage {
-  static getLocal(key: string) {
+  static getLocal(key: STORAGE_IDS) {
     if (!key) {
       return
     }


apps/mobile/src/store/constants.ts
```diff
@@ -278,3 +278,47 @@ export const mockedChains = [
     transactionService: 'https://safe-transaction-arbitrum.safe.global',
   },
 ]
+
+export const STORAGE_IDS = {
```
katspaugh, line 282: Should this be an enum?

apps/mobile/src/store/constants.ts
```diff
@@ -278,3 +278,47 @@ export const mockedChains = [
...
+  ANNOUNCEMENT_NOTIFICATION_CHANNEL_ID: 'ANNOUNCEMENT_NOTIFICATION_CHANNEL_ID',
+  DEFAULT_PUSH_NOTIFICATION_CHANNEL_PRIORITY: 'high',
+  REQUEST_PERMISSION_ASKED: 'REQUEST_PERMISSION_ASKED',
+  REQUEST_PERMISSION_GRANTED: 'REQUEST_PERMISSION_GRANTED',
+  NOTIFICATION_DATE_FORMAT: 'DD/MM/YYYY HH:mm:ss',
+  NOTIFICATIONS_SETTINGS: 'notifications-settings',
+  PN_USER_STORAGE: 'safePnUserStorage',
+}
+
+export const STORAGE_TYPES = {
```
katspaugh, line 300: Same here.

apps/mobile/src/utils/logger.ts
```diff
@@ -0,0 +1,89 @@
...
+      }
+      console.groupEnd()
+
+      if (this.shouldLogErrorToSentry) {
+        // TODO: implement Sentry error logging
+      }
+    }
+  }
+
+  setShouldLogErrorToSentry = (shouldLogErrorToSentry: boolean): void => {
```
katspaugh, line 84: Why do these functions need to be bound functions? As opposed to regular methods?

```
setShouldLogErrorToSentry(shouldLogErrorToSentry: boolean): void {
}
```

katspaugh: I don't fully follow this argument. It's equally easy to extend either of these class syntaxes. The only difference is that `this` is "bound" with your syntax so you can do things like `const warn = logger.warn` and it will still refer to the right `this`. At the cost of a tiny performance decrease.
Just making sure this choice is intentional.

apps/mobile/src/store/constants.ts
katspaugh: I don't see the change yet, so will unresolve the thread. Please generally don't resolve others' comments (let the author do it themselves) so that it's easier to follow up. :)

Once it's an enum, you can also use it as the type of the `key` in the functions below for extra type safety.

apps/mobile/src/utils/logger.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.shouldLogErrorToSentry = shouldLogErrorToSentry
+  }
+}
+
+export default new Logger()
```
clovisdasilvaneto, line 89: shouldn't we use the same logger between web and mobile projects? It might be a good idea move this logger to the `packages` folder, if the web project doesn't have one.

apps/mobile/src/store/storage.ts
```diff
@@ -1,19 +1,69 @@
...
   },
   removeItem: (key) => {
-    storage.delete(key)
+    safeStorage.delete(key)
     return Promise.resolve()
   },
 }
+
+export class safeMMKVStorage {
+  static getLocal(key: string) {
```
katspaugh, line 26: Are you expecting the key to ever be an arbitrary thing or is it going it be only `STORAGE_TYPES`?

apps/mobile/src/store/storage.ts
```diff
@@ -1,19 +1,69 @@
...
 import { Storage } from 'redux-persist'
 import { MMKV } from 'react-native-mmkv'
+import { mapStorageTypeToIds, STORAGE_IDS, STORAGE_TYPES } from './constants'
 
-const storage = new MMKV()
+const safeStorage = new MMKV({
+  id: STORAGE_IDS.SAFE,
+})
 
 export const reduxStorage: Storage = {
```
katspaugh, line 10: Btw, have you seen that we already have a very similar abstraction here: https://github.com/safe-global/safe-wallet-monorepo/blob/dev/apps/web/src/services/local-storage/Storage.ts

Perhaps `redux-persist` could be another "backend" for that storage class. Currently it only supports localStorage and sessionStorage.

katspaugh (approved):
👍

PR #4762, opened 2025-01-13
feat(Safe Apps): Add option for Safe Apps to include Telegram ID as a social media link

PooyaRaki:
## What it solves
Resolves the lack of support for Telegram as a social media option for Safe Apps.

## How this PR fixes it
This PR adds an option for Safe Apps to include their Telegram ID in the social media section.

## How to test it
1. Navigate to the config service where social media options of Safe Apps are managed.
2. Check for the new option to add a Telegram ID.
3. Add a Telegram ID and verify that it is saved.
4. Navigate to the web app and check that the ID is displayed correctly.
5. Ensure there are no UI or functional issues with the existing social media options.

## Screenshot
Screenshot 2025-01-13 at 12 03 47

## Checklist

- [-] I've tested the branch on mobile 📱
- [-] I've documented how it affects the analytics (if at all) 📊
- [+] I've written a unit/e2e test for it (if applicable) 🧑‍💻

### DIFF
--- a/apps/web/src/components/safe-apps/SafeAppSocialLinksCard/index.tsx
+++ b/apps/web/src/components/safe-apps/SafeAppSocialLinksCard/index.tsx
@@ -7,6 +7,7 @@ import Divider from '@mui/material/Divider'
 import { default as MuiLink } from '@mui/material/Link'
 import HelpOutlineRoundedIcon from '@mui/icons-material/HelpOutlineRounded'
 import GitHubIcon from '@mui/icons-material/GitHub'
+import TelegramIcon from '@mui/icons-material/Telegram'
 import TwitterIcon from '@mui/icons-material/Twitter'
 import { SafeAppSocialPlatforms } from '@safe-global/safe-gateway-typescript-sdk'
 import type { SafeAppData, SafeAppSocialProfile } from '@safe-global/safe-gateway-typescript-sdk'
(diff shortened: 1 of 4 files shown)


apps/web/src/components/safe-apps/SafeAppSocialLinksCard/index.tsx
```diff
@@ -80,6 +82,18 @@ const SafeAppSocialLinksCard = ({ safeApp }: SafeAppSocialLinksCardProps) => {
...
                 </IconButton>
               )}
+
+              {telegramSocialLink && (
+                <IconButton
+                  aria-label="Telegram link"
+                  component="a"
+                  href={telegramSocialLink.url}
+                  target="_blank"
+                  style={{ height: '40px', width: '40px' }}
```
katspaugh, line 92: Could you add `rel="noopener noreferrer"` to this and the other IcoButtons in this file?

Might be good to create a component to not repeat the same 6 lines for each button.


```
function IconButtonLink({ href, label, children }) {
  return (
    <IconButton
                  aria-label={label}
                  component="a"
                  href={href}
                  target="_blank"
                  rel="noopener noreferrer"
                  style={{ height: '40px', width: '40px' }}>
                    {children}
    </IconButton>
  )
}
```

PooyaRaki: @katspaugh I added the `rel` attribute.
https://github.com/safe-global/safe-wallet-monorepo/pull/4762/commits/7922961ccefe78b78f1612ba4b798d842dbf7341

I suggest postponing the creation of a new component to a separate PR.

katspaugh (approved):
LGTM!

PR #4765, opened 2025-01-13
chore: rename API generation command

iamacook:
## What it solves

Inconsistency with README

## How this PR fixes it

The `@safe-global/store` package's README states that the generation command is `generate-api`. This changes the command to match instead of `api-generate`.

## How to test it

Run `generate-api` command in `@safe-global/store` package and observe no error.

## Checklist

### DIFF
--- a/packages/store/package.json
+++ b/packages/store/package.json
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "scripts": {
     "test": "jest",
-    "api-generate": " npx @rtk-query/codegen-openapi scripts/openapi-config.ts"
+    "generate-api": " npx @rtk-query/codegen-openapi scripts/openapi-config.ts"
   },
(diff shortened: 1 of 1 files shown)


packages/store/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "scripts": {
     "test": "jest",
-    "api-generate": " npx @rtk-query/codegen-openapi scripts/openapi-config.ts"
+    "generate-api": " npx @rtk-query/codegen-openapi scripts/openapi-config.ts"
```
iamacook, line 7: This reads better so I opted to change the command, rather than the README.

compojoom (approved):

iamacook:
I'm going to merge this directly as it renames a command.

PR #4769, opened 2025-01-13
fix: copy the LICENSE back in the monorepo root

compojoom:
## What it solves
When we restructured the repo the LICENSE was moved to apps/web. I'm adding a copy of it also in the root of the repo. 
I'm living the original LICENSE in the apps/web as well. This way the docker image will also has the LICENSE file. (otherwise we need to update our build scripts to copay the LICENSE file around)


katspaugh (approved):
Thanks!

PR #4770, opened 2025-01-14
Refactor: use balances RTK query

katspaugh:
## What it solves

* Replace the fetching function from the old SDK with a new RTK query
* Polling is done via the `pollingInterval` query option

## How to test
Test that:
* Balances are loaded and displayed correctly
* Balances are updated every 15 seconds
* Spam filter
* Manually hidden tokens
* Fiat total in the sidebar
* Sending tokens with and w/o spending limits


apps/web/src/hooks/useBalances.ts
```diff
@@ -1,24 +1,70 @@
...
+  return data
+}
+
+const useBalances = () => {
+  const currency = useAppSelector(selectCurrency)
+  const isTrustedTokenList = useTokenListSetting()
+  const { safe, safeAddress, safeLoaded } = useSafeInfo()
+  const isReady = safeLoaded && isTrustedTokenList !== undefined
+
+  const { data, isLoading, error } = useBalancesGetBalancesV1Query(
```
usame-algan, line 46: Are we still polling the balances?

katspaugh: Part two of this PR is going to be to replace balances polling with a Safe Overview fetching because we actually only need the fiat total, not all the balances.

apps/web/src/hooks/loadables/useLoadBalances.ts
```diff
@@ -24,46 +24,56 @@ export const useTokenListSetting = (): boolean | undefined => {
...
-      })
+  const { data, isLoading, error } = useBalancesGetBalancesV1Query(
+    {
+      chainId: safe.chainId,
+      safeAddress,
+      fiatCode: currency,
+      trusted: isTrustedTokenList,
+    },
+    {
+      skip: !isReady,
```
usame-algan, line 55: We also need to skip / prevent this request in case it is a counterfactual safe otherwise we are getting a 404 and an unhandled error
Screenshot 2025-01-27 at 16 27 57

apps/web/src/components/sidebar/IndexingStatus/index.tsx
```diff
@@ -16,6 +16,7 @@ const useIndexingStatus = () => {
 
   return useAsync(
     () => {
+      if (count === undefined) return
```
usame-algan, line 19: How can this case happen? Afaik the default value is 0 so it can never be undefined.

apps/web/src/hooks/loadables/useLoadBalances.ts
```diff
@@ -24,46 +24,56 @@ export const useTokenListSetting = (): boolean | undefined => {
...
-  // Reset the counter when safe address/chainId changes
-  useEffect(() => {
-    resetPolling()
-  }, [resetPolling, safeAddress, chainId])
+  // Counterfactual balances
+  const cfData = useCounterfactualBalances(safe)
+
+  return useMemo(
+    () => ({
+      balances: data || cfData || { items: [], fiatTotal: '' },
```
usame-algan, line 65: nit: I would extract the object fallback into a `defaultBalances` or `emptyBalances` variable

apps/web/src/components/sidebar/IndexingStatus/index.tsx
katspaugh: I added it only to satisfy the linter as it's a dependency.

PR #4771, opened 2025-01-14
Fix(Tx notes): fire analytics on tx submit + adjust design

katspaugh:
## What it solves

A follow-up on #4693.

Triggering the `Add tx note` on input change is firing too many times, so I'm changing it to fire only when you submit a tx with a note.

Also, we've adjusted the design as per internal feedback:

Screenshot 2025-01-14 at 16 42 31

## How this PR fixes it

* GA event is moved to tx modal submit callback
* GA event is renamed to `Submit tx note` to differentiate it from the previous, incorrectly tracked, event
* I also renamed the `onSubmit` prop name to `onChange` for clarity
* Slight adjustments to the design
* Character limit is reduced to 60 to fit in the `origin` limit better + align with the size of the input.

### DIFF
--- a/apps/web/src/components/tx/SignOrExecuteForm/SignOrExecuteForm.tsx
+++ b/apps/web/src/components/tx/SignOrExecuteForm/SignOrExecuteForm.tsx
@@ -33,7 +33,7 @@ import ConfirmationView from '../confirmation-views'
 import { SignerForm } from './SignerForm'
 import { useSigner } from '@/hooks/wallets/useWallet'
 import { trackTxEvents } from './tracking'
-import { TxNoteForm, encodeTxNote } from '@/features/tx-notes'
+import { TxNoteForm, encodeTxNote, trackAddNote } from '@/features/tx-notes'
 
 export type SubmitCallback = (txId: string, isExecuted?: boolean) => void
 
@@ -112,8 +112,12 @@ export const SignOrExecuteForm = ({
         !!signer?.isSafe,
         customOrigin,
       )
+
+      if (customOrigin !== props.origin) {
+        trackAddNote()
+      }
     },
-    [chainId, isCreation, onSubmit, trigger, signer?.isSafe, customOrigin],
+    [chainId, isCreation, onSubmit, trigger, signer?.isSafe, customOrigin, props.origin],
   )
 
   const onRoleExecutionSubmit = useCallback<typeof onFormSubmit>(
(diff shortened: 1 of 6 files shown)


apps/web/src/features/tx-notes/index.tsx
```diff
@@ -1,7 +1,12 @@
...
 import { TxNoteForm as TxNoteFormComponent } from './TxNoteForm'
+import { MODALS_EVENTS, trackEvent } from '@/services/analytics'
 
 export const TxNote = featureToggled(TxNoteComponent, FEATURES.TX_NOTES)
 export const TxNoteForm = featureToggled(TxNoteFormComponent, FEATURES.TX_NOTES)
 export * from './encodeTxNote'
+
+export function trackAddNote() {
+  trackEvent(MODALS_EVENTS.ADD_TX_NOTE)
+}
```
usame-algan, line 12: nit: Not sure if we need this abstraction. Everywhere else we use trackEvent directly.

apps/web/src/components/tx/SignOrExecuteForm/SignOrExecuteForm.tsx
```diff
@@ -112,8 +112,12 @@ export const SignOrExecuteForm = ({
         !!signer?.isSafe,
         customOrigin,
       )
+
+      if (customOrigin !== props.origin) {
+        trackAddNote()
+      }
```
usame-algan, line 118: What happens if the user didn't add a note and its not a safe apps transaction. Are both values undefined and we would try to track something?

usame-algan (approved):

apps/web/src/components/tx/SignOrExecuteForm/SignOrExecuteForm.tsx
usame-algan: Just trying to understand why we initialize `customOrigin` with `props.origin` instead of just leaving it undefined and if it exists we track an event.

katspaugh: undefined equals undefined though.

katspaugh: Ah, simply to avoid doing `customOrigin || props.origin` when passing it down.

francovenica:
Tested the event changes. The event only triggers when a tx is submitted. 
If the tx was first signed and then executed, the event only triggers once (after signing) which is expected 
If the tx is executed right away the event triggers after siging which is expected

Also verified the fix for the empty block you can see during tx execution when no tx note was created during tx proposal 

LGTM

PR #4785, opened 2025-01-16
Fix(Tx notes): hide for 1/X safes

### LINKED ISSUE #4779: Tx notes - Notes are not being saved if the tx executes right away
## Bug description
Tx notes are not showing when a tx is executed right away (1/X safes)

## Environment

- Browser: Chrome
- Wallet: MetaMask
- Chain: Ethereum mainnet
- evn: dev 1.49
## Steps to reproduce

1.  Create any send funds tx you can execute right away
2. Write a note
3. Execute the tx

## Expected result
The note shows

## Obtained result
Note is not in the tx

## Screenshots
no notes swapp

https://safe-wallet-web.dev.5afe.dev/transactions/tx?safe=sep:0x8f4A19C85b39032A37f7a6dCc65234f966F72551&id=multisig_0x8f4A19C85b39032A37f7a6dCc65234f966F72551_0xe818effa0c5ca528c1c155e392a5febf40bf5bc542693be3b625929de15eacda

katspaugh:
## What it solves

Resolves #4779.

* Disable tx notes for 1/X Safes

## Why?
Notes cannot be added when doing an immediate execution in 1/X Safes because we propose the tx before a note is added and then it's impossible to overwrite it unless the tx is signed.

So for now, we're disabling notes for all 1/X Safes. A better solution will be done later in #4783.

### DIFF
--- a/apps/web/src/features/tx-notes/TxNoteForm.tsx
+++ b/apps/web/src/features/tx-notes/TxNoteForm.tsx
@@ -2,6 +2,7 @@ import type { TransactionDetails } from '@safe-global/safe-gateway-typescript-sd
 import TxCard from '@/components/tx-flow/common/TxCard'
 import { TxNote } from './TxNote'
 import { TxNoteInput } from './TxNoteInput'
+import useSafeInfo from '@/hooks/useSafeInfo'
 
 export function TxNoteForm({
   isCreation,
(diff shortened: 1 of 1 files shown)


usame-algan (approved):

liliya-soroka:
verified

PR #4788, opened 2025-01-16
feat(safenet): Multi-recipient token transfers

germartinez:
## What it solves

This PR adds a multi-recipient token transfer tx flow for Safenet-enabled Safe accounts

Resolves https://github.com/safe-global/safenet/issues/269

## How to test it

Create a new transaction on a Safenet-enabled Safe.

## Checklist


apps/web/src/components/tx-flow/flows/SafenetTokenTransfers/utils.ts
```diff
@@ -0,0 +1,21 @@
+import { useVisibleBalances } from '@/hooks/useVisibleBalances'
+import { type SafeBalanceResponse } from '@safe-global/safe-gateway-typescript-sdk'
+
+export const useTokenAmount = (selectedToken: SafeBalanceResponse['items'][0] | undefined) => {
+  const totalFiatAmount =
```
schmanu, line 5: I do not understand why we need to consider `decimals` for the `fiatBalance`.
The decimals should only matter for the `balance`.

Generally I feel like the maxAmount is a bit more complex. IMO it should consider the upper bound of e.g. 100.000$ in fiat.
But also consider how much unified balance the user has and how much is already used up by previous transfers.

apps/web/src/components/tx-flow/flows/SafenetTokenTransfers/CreateTokenTransfers.tsx
```diff
@@ -0,0 +1,169 @@
...
+      </Typography>
+
+      <Typography variant="caption" component="p">
+        {formatVisualAmount(item.balance, item.tokenInfo.decimals)} {item.tokenInfo.symbol}
+      </Typography>
+    </Grid>
+  </Grid>
+)
+
+export const CreateTokenTransfers = ({
```
schmanu, line 45: We need a component to check that the total transfer amount is not higher than the users balance

apps/web/src/components/tx-flow/flows/SafenetTokenTransfers/RecipientRow/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const fieldName = `${groupName}.${index}`
+  const {
+    watch,
+    formState: { errors },
+  } = useFormContext<TokenTransfersParams>()
+
+  const recipient = watch(TokenTransfersFields.recipients)
+
+  // TODO: Review tokens available for selection and max amount
+  const selectedToken = balancesItems.find((item) => item.tokenInfo.symbol === 'USDC')
```
schmanu, line 32: I think we should use the addresses from the Safenet config instead of the name / symbol. There are many fake tokens that are called USDC for instance.

apps/web/src/hooks/safe-apps/useRemoteSafeApps.ts
```diff
@@ -40,9 +45,15 @@ const useRemoteSafeApps = (tag?: SafeAppsTag): AsyncResult<SafeAppsResponse> =>
...
   const apps = useMemo(() => {
-    if (!remoteApps || !tag) return remoteApps
-    return remoteApps.filter((app) => app.tags.includes(tag))
-  }, [remoteApps, tag])
+    if (!remoteApps) return remoteApps
+    if (tag) {
+      return remoteApps.filter((app) => app.tags.includes(tag))
+    }
+    if (name) {
+      return remoteApps.filter((app) => app.name.includes(name))
```
schmanu, line 53: Should we not check for equality instead of just including the name?

schmanu (changes requested):

apps/web/src/components/tx-flow/flows/SafenetTokenTransfers/utils.ts
germartinez: As we decided in our call, let's move this topic out of the happy path: https://github.com/safe-global/safenet/issues/290

apps/web/src/components/tx-flow/flows/SafenetTokenTransfers/CreateTokenTransfers.tsx
germartinez: As we decided in our call, let's move this topic out of the happy path: https://github.com/safe-global/safenet/issues/290.
For now I implemented the UI alert, without any functionality.

schmanu (approved):
Ty for adjusting! I think its good to be merged for now.
We'll address further improvements in subsequent PRs!

PR #4791, opened 2025-01-17
feat(Safenet): Add Safenet opt-in flow

germartinez:
## What it solves

Implement the Safenet opt-in user flow (1.a and 1.b) starting in the Safe creation flow with the current `CardStepper` component. Create all the new UI components. Call the deployment endpoint. Check the Safe creation events.

Resolves https://github.com/safe-global/safenet/issues/276

## How to test it

- Safenet banner displayed on the account list.
- Regular flow without Safenet
- Flow 1.a to create new Safenet accounts.
- Flow 1.b to create skip Safenet accounts creation.

## Screenshots

Flow 1.a
Image

Flow 1.b
Image

## Checklist


apps/web/src/features/safenet/components/new-safe/SafenetReview/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+    router?.push({
+      pathname: AppRoutes.home,
+      query: { safe: `${data.networks[0].shortName}:${deploymentData.safeAddress}` },
+    })
+  }, [data, deploymentData, dispatch, onSubmit, router])
+
+  const handleCreateSafeClick = async () => {
+    setIsCreating(true)
+
+    deploy({
```
schmanu, line 110: Instead of handling the result processing through `useEffect(func, [deploymentData])` we could also do 
```ts
   const {data: deploymentData} = await deploy()
   // Process deploymentData.
```

This way we have less side-effects handling.

apps/web/src/features/safenet/components/new-safe/SafenetReview/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+  onSubmit,
+  setSubmitError,
+  setIsCreating,
+}: UseSubmitSafenetReviewHandlerProps): SafenetReviewType => {
+  const router = useRouter()
+  const dispatch = useAppDispatch()
+  const [deploy, { data: deploymentData, error }] = useLazyDeploySafenetAccountQuery()
+
+  useEffect(() => {
+    if (!error) return
```
schmanu, line 42: See comment below: I think we should move this handling to the `handleSafeCreationClick` function or maybe even extract it into a new `safeCreationCallback` function that gets called with the result of the deployment call.

apps/web/src/features/safenet/components/new-safe/SafenetReview/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+        PayMethod.PayNow,
+      )
+
+      safeCreationDispatch(SafeCreationEvent.RELAYING, {
+        groupKey: CF_TX_GROUP_KEY,
+        safeAddress: deploymentData.safeAddress,
+        chainId: chain.chainId,
+      })
+
+      //trackEvent({ ...OVERVIEW_EVENTS.PROCEED_WITH_TX, label: 'deployment', category: CREATE_SAFE_CATEGORY })
```
schmanu, line 84: We should create a new github issue regarding analytics events for Safenet, but can leave it out of scope in the PR.

apps/web/src/features/safenet/components/new-safe/SafenetInfoCard/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+import BellIcon from '@/public/images/safenet/bell.svg'
+import CloseIcon from '@mui/icons-material/Close'
+import { Box, Button, IconButton, Typography } from '@mui/material'
+import { useRouter } from 'next/router'
+import { useState } from 'react'
+import css from './styles.module.css'
+
+function SafenetInfoCard() {
+  const router = useRouter()
+  const [displayBanner, setDisplayBanner] = useState<boolean>(true)
```
schmanu, line 11: Usually we use the local storage to remember that a user closed a banner.

```suggestion
  const [displayBanner = true, setDisplayBanner] = useLocalStorage<boolean>("showSafenetBanner")
```

apps/web/src/features/safenet/components/new-safe/SafenetBanner/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Box, Button, Grid, Typography } from '@mui/material'
+import { useTheme } from '@mui/material/styles'
+import { useRouter } from 'next/router'
+import { useState } from 'react'
+import css from './styles.module.css'
+
+const SafenetBanner = () => {
+  const router = useRouter()
+  const { palette } = useTheme()
+  const [displayBanner, setDisplayBanner] = useState<boolean>(true)
```
schmanu, line 15: Usually we use the local storage to remember that a user closed a banner.

```suggestion
  const [displayBanner = true, setDisplayBanner] = useLocalStorage<boolean>("showSafenetBanner")
```

apps/web/src/features/counterfactual/hooks/usePendingSafeStatuses.ts
```diff
@@ -57,7 +61,7 @@ const usePendingSafeMonitor = (): void => {
           } = undeployedSafe
 
           const isProcessing = status === PendingSafeStatus.PROCESSING && txHash !== undefined
-          const isRelaying = status === PendingSafeStatus.RELAYING && taskId !== undefined
+          const isRelaying = status === PendingSafeStatus.RELAYING
```
schmanu, line 64: We could consider adding a new `PendingSafeStatus.SAFENET_RELAYING` in case we run into problems with re-using `RELAYING`

schmanu (commented):
It is looking and working well and a huge lift 🥇 

But it is a bit hard to follow the Steppers with the new SafenetFlow property.

Maybe it would be better to duplicate the entire `StepperCard` in a new `SafenetCreation` component similar to the `AdvancedSafeCreation` component.
Then we could add new Steps for all Steps that are different when using Safenet (e.g. `SetNameStep`, `ReviewStep`) but keep using the same `OwnerPolicyStep`

This should lead to simpler components as we do not have those `isSafenet ? (<SafenetComponent />) : (<NormalComponent />)` switches.

apps/web/src/features/safenet/components/new-safe/SafenetInfoCard/index.tsx
germartinez: The main Safenet banner and this card are the only access points to Safenet creation flow. This makes sense for the main banner in the dashboard, but it should be visible here every time the creation flow is started (with the dismiss option), otherwise there is no access point.

apps/web/src/features/safenet/components/new-safe/SafenetReview/index.tsx
germartinez: https://github.com/safe-global/safenet/issues/369

schmanu (approved):
Nice one! Vamos!

PR #4800, opened 2025-01-21
Fix(Blockaid): pass chain id instead of chain name for Blast and others

katspaugh:
## What it solves

We noticed that Blockaid wasn't working on Blast because we were limiting supported chains to a hardcoded list.
Since we already have a per-chain feature toggle for this, we don't need that list. This enables Blockaid on Blast.

## How to test it

Create a tx on Blast and check that Blockaid now works
Do the same on some other chain where tx scanning is enabled


schmanu (approved):
Nice that they support chainIds instead of names now.

liliya-soroka:
1. getting Post request to blockaid with 422 error on networks where the feature is turned off

katspaugh:
I've tested on Aurora where Blockaid is disabled via the feature flag, and I don't see any Blockaid requests.
The code is also 100% checking the RISK_MITIGATION  flag before doing any requests.

PR #4803, opened 2025-01-21
Fix(Swaps): TWAP order decoding

### LINKED ISSUE #4802: Twap tx decoding is missing on the TX confirmation
## Bug description
Seems, that after the Tx confirmation refactoring we lost the Twap tx decoding on the Tx confirmation screen
## Environment

- Browser: Chrome
- Wallet: any
- Chain: any

## Steps to reproduce

1.  Go to Native swap or Cowswap safe app
2. start the Twap tx creation flow
3. go to the Tx confirm page

Current result: There is no decoding 

Image

katspaugh:
## What it solves

Resolves #4802

## How this PR fixes it

TWAP confirmation view is the same as the Swap one so it just needed an extra mapping.

### DIFF
--- a/apps/web/src/components/tx/confirmation-views/index.tsx
+++ b/apps/web/src/components/tx/confirmation-views/index.tsx
@@ -11,6 +11,7 @@ import {
   isSafeToL2MigrationTxData,
   isSafeUpdateTxData,
   isSwapOrderTxInfo,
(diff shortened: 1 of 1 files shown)


liliya-soroka:
Verified

clovisdasilvaneto (approved):

PR #4810, opened 2025-01-22
feat(Safenet): Safenet feature flag

germartinez:
## What it solves

This PR creates a feature flag for Safenet, to hide or display all the related components.

Resolves https://github.com/safe-global/safenet/issues/277

## How to test it

Add `NEXT_PUBLIC_IS_SAFENET_ENABLED` to your `.env` file and try `NEXT_PUBLIC_IS_SAFENET_ENABLED=true` vs. `NEXT_PUBLIC_IS_SAFENET_ENABLED=false`. When set to `true` all Safenet features/components should be displaed, otherwise the Safenet features should be hidden.

## Checklist


apps/web/src/hooks/useIsSafenetEnabled.ts
```diff
@@ -6,6 +7,10 @@ const useIsSafenetEnabled = () => {
   const { safe } = useSafeInfo()
   const { data: safenetConfig } = useGetSafenetConfigQuery()
```
schmanu, line 8: I think we should pass the `skipToken` here in case safenet is not enabled such that RTK query skips the request.

schmanu (changes requested):

katspaugh:
Please also make sure the entire feature is encapsulated in one folder and is exported from a single, lazy-loaded index.ts module as per https://www.notion.so/safe-global/Feature-toggles-970c96d18e544fd1a01a362f0535146b

schmanu (commented):
Should me move some more components into the safenet feature folder?
e.g. EnableSafenet, GradientBoxSafenet, the Safenet hooks

schmanu (approved):
Thanks for adjusting!

PR #4813, opened 2025-01-22
Fix(Gas estimation): gas price from oracles can be a float

### LINKED ISSUE #4754: Gas price can't be converted to BigInt
## Bug description

In some cases, the `fetchGasOracle` function is producing a floating-point result and tries to convert it to BigInt which fails

## Environment

- Wallet: MetaMask through WC
- Chain: Ethereum mainnet

## Steps to reproduce

1.  Open a Safe on mainnet
2. Try to execute a transaction through MM with WC (TBD)

## Screenshots
image (9)

katspaugh:
## What it solves

Resolves #4754

## How this PR fixes it

Gas oracles return prices as a float inside a string. Converting it directly to BigInt fails. So we're now trimming the decimal part before converting them.


apps/web/src/hooks/useGasPrice.ts
```diff
@@ -46,6 +46,10 @@ const isEtherscanResult = (data: any): data is EtherscanResult => {
   return 'FastGasPrice' in data && 'suggestBaseFee' in data
 }
 
+const stringToBigInt = (value: string): bigint => {
+  return BigInt(value.split('.')[0])
+}
```
usame-algan, line 51: This could result in a too low gas price estimation as we are "rounding down" by cutting of the decimal part. Alternatively we could also round up. Wdyt?

Nit: The name of the function is not entirely accurate as it not only converts the value but also transforms it

usame-algan (approved):

apps/web/src/hooks/useGasPrice.ts
katspaugh: The point of this fn is to not convert to a number to avoid overflows but we could do something like `val === '0' ? '1' : val`.

francovenica:
Question:

This only happens with the stg CGW enabled but still:
in the Max Prio Fee I see a " - ". If you edit the values the field is empty (not even a 0) and in the wallet (MM Phone app) the fee is shown as 0.000000001 Gwei
Could this become an issue in prod? I assume not since in prod CGW works fine, but idk why it fails in the stg cgw to begin with

image

If it is expected/unavoidable or at least we are sure that it wont be an issue in prod then the PR is fine and can be merged

liliya-soroka:
@francovenica , could you provide the steps to reproduce?

katspaugh:
I've added a fallback of 1 Gwei so that it's never a `-`.

PR #4821, opened 2025-01-24
Docs: add a template for tasks

katspaugh:
## What it solves

For some structure and linking to Notion.


usame-algan (approved):
Should we add a section on "How to test" for QA?

PR #4834, opened 2025-01-27
Chore: automatically fetch CGW schema

katspaugh:
## What it solves

Automate the fetching of the CGW API schema.

The new command to fetch and generate RTK queries is `yarn build`.


packages/store/scripts/fetch-schema.ts
```diff
@@ -0,0 +1,15 @@
+const PRODUCTION_CGW_API_URL = process.env.PRODUCTION_CGW_API_URL || 'https://safe-client.safe.global/api-json'
+const STAGING_CGW_API_URL = process.env.STAGING_CGW_API_URL || 'https://safe-client.staging.5afe.dev/api-json'
```
compojoom, line 2: Can we maybe do something like
const urls = {prod: ... , staging: ..., local: ...}

and then have 
build:prod
build:staging
build:local

It's weird that we do build:dev, but we use teh staging url for it. 

At least up till now I've mostly used my localhost to build the API and now having Production_GW_API_URL and STAGING_CGW_API - which one do I override?

katspaugh: If you just run `yarn build`, it will use the stable version (prod). If you need some latest changes from dev/staging, use `yarn build:dev`.

compojoom (approved):
Thanks!

PR #4837, opened 2025-01-27
feat(Mobile): add notifications opt-In screen

Jonathansoufer:
## What it solves

This ticket implements the notifications onboarding screen, enabling users to toggle the feature ON or deciding to leave for later. Its important to mention though that after three rejected attempts the user will not be prompt again within a month.

P.S: Although out of the scope of this PR, it was implemented a reusable [FloatingContainer](https://github.com/safe-global/safe-wallet-monorepo/pull/4837/files#diff-664fee1a63f48c42b7fadd22859d485d29e8569d20601581bea3998e1a3ff479) which is intent to be used to group and stick elements on the bottom of the screen. It has also a awareness of keyboard elements on those cases when keyboard is necessary to proceed with an action.

Resolves #

## Screenshots
Screenshot 2025-01-28 at 13 07 34

## Checklist

### DIFF
--- a/apps/mobile/src/features/Assets/Assets.container.tsx
+++ b/apps/mobile/src/features/Assets/Assets.container.tsx
@@ -1,10 +1,14 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 
 import { SafeTab } from '@/src/components/SafeTab'
 
 import { TokensContainer } from '@/src/features/Assets/components/Tokens'
 import { NFTsContainer } from '@/src/features/Assets/components/NFTs'
 import { AssetsHeaderContainer } from '@/src/features/Assets/components/AssetsHeader'
+import useNotifications from '@/src/hooks/useNotifications'
+import { useRouter } from 'expo-router'
+import { useAppDispatch } from '@/src/store/hooks'
+import { updatePromptAttempts } from '@/src/store/notificationsSlice'
 
 const tabItems = [
   {
@@ -18,5 +22,22 @@ const tabItems = [
 ]
 
 export function AssetsContainer() {
+  const { isAppNotificationEnabled, promptAttempts } = useNotifications()
+  const dispatch = useAppDispatch()
+  const router = useRouter()
+
+  /*
+   * If the user has not enabled notifications and has not been prompted to enable them,
+   * redirect to the opt-in screen
+   * */
+
+  const shouldShowOptIn = !isAppNotificationEnabled && !promptAttempts
+
+  useEffect(() => {
+    if (shouldShowOptIn) {
+      dispatch(updatePromptAttempts(1))
+      router.navigate('/notifications-opt-in')
+    }
+  }, [])
   return <SafeTab items={tabItems} headerHeight={200} renderHeader={AssetsHeaderContainer} />
 }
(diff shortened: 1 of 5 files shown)


apps/mobile/src/features/Onboarding/components/OnboardingCarousel/OnboardingCarousel.test.tsx
```diff
@@ -1,8 +1,18 @@
...
+jest.mock('@notifee/react-native', () => ({
+  AndroidImportance: {
+    NONE: 0,
+    MIN: 1,
+    LOW: 2,
+    DEFAULT: 3,
+    HIGH: 4,
+  },
+}))
+
```
compojoom, line 15: I guess that we should move this to a global mock as you have it also in the Onboarding.container.test.tsx?

apps/mobile/src/features/Onboarding/components/OnboardingCarousel/items.tsx
```diff
@@ -17,6 +18,9 @@ const styles = StyleSheet.create({
   signing: {
     height: Math.abs(windowHeight * 0.3),
   },
+  notifications: {
+    height: Math.abs(windowHeight * 0.32),
```
compojoom, line 22: what's up with those magic contants here? For signing it is 0.3, here it is 0.32? Why are we multiplying with that? How does it look on a smaller/bigger device?

apps/mobile/src/hooks/useNotifications.ts
```diff
@@ -55,6 +57,11 @@ const useNotifications = (): NotificationsProps => {
           if (isAppNotificationEnabled) {
             dispatch(toggleAppNotifications(false))
           }
+
+          if (isOnboarding) {
+            router.navigate('/(tabs)')
```
compojoom, line 62: I think that we should be using replace instead of navigate. With navigate on devices with a hardware back button you would be able to go bath to the onboarding screeen. Where with replace you won't be.

katspaugh:
@Jonathansoufer could you please adjust the PR title as per our guidelines?

compojoom (commented):
What is the plan with this screen? When are we going to show it? Immediately after the user goes through the onboarding or at a later stage? After that if the user cancels are we prompting him every time?

apps/mobile/src/features/Assets/Assets.container.tsx
```diff
@@ -18,5 +22,22 @@ const tabItems = [
...
+   * */
+
+  const shouldShowOptIn = !isAppNotificationEnabled && !promptAttempts
+
+  useEffect(() => {
+    if (shouldShowOptIn) {
+      dispatch(updatePromptAttempts(1))
+      router.navigate('/notifications-opt-in')
+    }
+  }, [])
```
compojoom, line 41: This is weird. The AssetsContainer should not care about notifications. 
Get some inspiration from here:
https://docs.expo.dev/router/reference/authentication/
https://reactnavigation.org/docs/auth-flow
I think that the _layout.tsx should be the place where we decide what first screen we show to the user.

compojoom: When the user clicks on "get started/continue" in the onboarding flow we should set a flag "onboarding_done": true. Next time the user starts we should not show the onboarding screens. So we will be deciding in the _layout.tsx which screen to show anyway.

compojoom (approved):
Jonathan and I, agreed to merge this PR as is, but create a follow up Ticket that would move the logic of the first screen to the _layout or to a better place.

PR #4838, opened 2025-01-27
feat(Queue): add share block to transaction details

iamacook:
## What it solves

Improves [transaction sharing](https://www.notion.so/safe-global/Add-tx-copy-link-for-multisig-transactions-1888180fe57380ba929edb166cbff24f)

## How this PR fixes it

Sharing a transaction can assist in signature requests, but the current share button is not super clear.

This adds a new share block to QUEUED MULTISIG transactions according to [the designs](https://www.figma.com/design/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=12735-56883&t=1mKANji8TAad1EkP-0)

## How to test it

Each of the following should emit distinctly labelled tracking events:

1. Open a QUEUED MULTISIG transaction and observe the new share block, with a functioning copy button. When expanding it, an event should also be emitted.
2. Ensure the existing copy button works and is present on all transaction types.

## Screenshots

image

image

image

## Checklist


apps/web/src/components/transactions/TxShareLink/TxShareBlock.tsx
```diff
@@ -0,0 +1,37 @@
...
+import TxShareLink from '.'
+
+import css from './styles.module.css'
+
+export function TxShareBlock({ txId }: { txId: string }): ReactElement | null {
+  return (
+    <Paper className={css.wrapper}>
+      <Accordion className={css.accordion}>
+        <AccordionSummary expandIcon={<ExpandMoreIcon />} className={css.summary}>
+          <Typography className={css.header}>Share link with other owners</Typography>
```
katspaugh, line 16:
```suggestion
          <Typography className={css.header}>Share link with other signers</Typography>
```

We've been calling them signers throughout the UI.

iamacook: Updated in fe946dd8909741ccba923ac5b402c9b99ebb8249.

apps/web/src/components/transactions/TxShareLink/TxShareBlock.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <AccordionSummary expandIcon={<ExpandMoreIcon />} className={css.summary}>
+          <Typography className={css.header}>Share link with other signers</Typography>
+        </AccordionSummary>
+        <AccordionDetails className={css.details}>
+          If signers have previously subscribed to notifications, they will be notified about signing this transaction.
+          You can also share the link with them to speed up the process.
+        </AccordionDetails>
+      </Accordion>
+      <div className={css.copy}>
+        <TxShareLink id={txId} event={TX_LIST_EVENTS.COPY_SHARE_LINK}>
```
iamacook, line 30: Should this instead be the same event as the button, using a label to differentiate? cc @katspaugh

katspaugh: I would keep a separate event.

katspaugh (approved):
👍

apps/web/src/components/transactions/TxShareLink/TxShareBlock.tsx
alicecha: Actually the way we've been organising events lately is that each action is linked to a single event, and we use labels to differentiate where the action is done. How this would apply in this case is having an event (e.g. "copy share link") with labels to distinguish between "icon" and "button".

No strong feelings though - so leaving this up to you.

apps/web/src/components/transactions/TxShareLink/index.tsx
tmjssz, line 1: This component is only used internally by TxShareBlock/Button right? So maybe as a suggestion we could rename the file to its component name and not be `index` anymore. Just for consistency, but not blocking

tmjssz (approved):
LGTM! Just left a small suggestion

apps/web/src/components/transactions/TxShareLink/TxShareBlock.tsx
iamacook: I changed it to use the same event ("Copy deeplink") in 20df0b7f2747548a249c346a9ff65a42f30b520d with different labels: `button` and `share-block`. The "Open share block" remains though.

apps/web/src/components/transactions/TxShareLink/index.tsx
iamacook: `TxShareLink` is a wrapper for `TxShareBlock` and `TxShareButton`. We could separate them into independant components but I think keeping them together with the wrapper makes sense. What do you think?

tmjssz: Yeah all good then, lets keep them together

liliya-soroka:
Checked cases:
1. Tracking 
2. not executed txs in different states of signing
3. executed txs
4. nested safes
5. Chrome, Firefox
6. Mobal view
7. diff themes
8. queued/history and tx details
9. proposed txs
--
Test cases were created- https://gno.testrail.io/index.php?/suites/view/36&group_by=cases:section_id&group_id=7002&group_order=asc&display_deleted_cases=0

----- 
The task to add automation was created  - https://github.com/safe-global/safe-test-management/issues/17

PR #4842, opened 2025-01-28
Upload build output using parallel threads

louis-md:
## What it solves

Identical to safe-global/safe-docs#680 and safe-global/safe-homepage#541, this fix aims to drastically reduce deployment times to AWS.

## How this PR fixes it

The issue is solved by uploading files in parallel using the `&` operator.

### DIFF
--- a/apps/web/scripts/github/s3_upload.sh
+++ b/apps/web/scripts/github/s3_upload.sh
@@ -13,7 +13,9 @@ aws s3 sync . $BUCKET --delete
 
 # Upload all HTML files again but w/o an extention so that URLs like /welcome open the right page
 for file in $(find . -name '*.html' | sed 's|^\./||'); do
-    aws s3 cp ${file%} $BUCKET/${file%.*} --content-type 'text/html'
+    aws s3 cp ${file%} $BUCKET/${file%.*} --content-type 'text/html' &
 done
+
+wait
 
 cd -


katspaugh:
Pretty sure it's not good to launch so many processes in parallel. There are hundreds of files.

apps/web/scripts/github/s3_upload.sh
```diff
@@ -13,7 +13,9 @@ aws s3 sync . $BUCKET --delete
 
 # Upload all HTML files again but w/o an extention so that URLs like /welcome open the right page
 for file in $(find . -name '*.html' | sed 's|^\./||'); do
-    aws s3 cp ${file%} $BUCKET/${file%.*} --content-type 'text/html'
+    aws s3 cp ${file%} $BUCKET/${file%.*} --content-type 'text/html' &
 done
 
+wait
+
```
katspaugh, line 20: ChatGPT suggested we can run them in batches like so:

```
# Max number of parallel processes
MAX_PARALLEL=20

# Counter to track the number of running processes
counter=0

for file in $(find . -name '*.html' | sed 's|^\./||'); do
    # Launch the process
    aws s3 cp "${file%}" "$BUCKET/${file%.*}" --content-type 'text/html' &

    # Increment the counter
    ((counter++))

    # If the counter reaches the max limit, wait for all processes to complete
    if ((counter >= MAX_PARALLEL)); then
        wait
        counter=0
    fi
done

# Wait for any remaining processes to complete
wait
```

louis-md:
Closed in favor of https://github.com/safe-global/safe-wallet-monorepo/pull/4843

PR #4843, opened 2025-01-28
fix: upload build output to AWS using up to 20 parallel threads

louis-md:
## What it solves

Identical to https://github.com/safe-global/safe-docs/pull/680 and https://github.com/safe-global/safe-homepage/pull/541, this fix aims to drastically reduce deployment times to AWS.

## How this PR fixes it

The issue is solved by uploading files in parallel using the `&` operator. A limit of 20 threads has been added for more safety.


katspaugh (approved):
Thanks Louis! 🦊

PR #4851, opened 2025-01-29
Fix: nested tx preview

### LINKED ISSUE #4840: Nested safe: The Executehas nested tx details can not be loaded and the browser hangs
## Environment

- Browser: Chrome
- Wallet: MetaMask
- Chain: Ethereum mainnet

## Steps to reproduce

1. create a tx in the child safe ( remove owner in my case)
2. try to sign and execute using the owner of the parent safe
3. go to the parent safe 
4. open Executehash nested tx 
5. open tx details and try to sign and execute 
Current result : details can not be loaded and the browser tab hangs. The problem appears only with staging CGW

Image

## Expected result
It should be possible to see the nested tx details and sign in the parent safe 
## Screenshots
https://safe-wallet-web.dev.5afe.dev/transactions/queue?safe=sep:0x2a73e61bd15b25B6958b4DA3bfc759ca4db249b9

katspaugh:
## What it solves

Resolves #4840

It was a regression after #4783 (not live yet).

## How this PR fixes it

The `operation` field in a nested tx was BigInt and it wasn't working well with the tx preview endpoint. So I converted it to a number.

Also optimized the dependencies of useAsync in useTxPreview to avoid an infinite loop if a passed tx object is re-created.

### DIFF
--- a/apps/web/src/components/transactions/TxDetails/TxData/NestedTransaction/ExecTransaction/index.tsx
+++ b/apps/web/src/components/transactions/TxDetails/TxData/NestedTransaction/ExecTransaction/index.tsx
@@ -58,7 +58,7 @@ export const ExecTransaction = ({
   const [txPreview, error] = useTxPreview(
     childSafeTx
       ? {
-          operation: childSafeTx.data.operation,
+          operation: Number(childSafeTx.data.operation),
           data: childSafeTx.data.data,
           to: childSafeTx.data.to,
           value: childSafeTx.data.value.toString(),
--- a/apps/web/src/components/tx/SignOrExecuteForm/__tests__/__snapshots__/SignOrExecute.test.tsx.snap
+++ b/apps/web/src/components/tx/SignOrExecuteForm/__tests__/__snapshots__/SignOrExecute.test.tsx.snap
@@ -8,6 +8,7 @@ exports[`SignOrExecute should display a confirmation screen 1`] = `
   >
     <div
       class="MuiCardContent-root cardContent css-1lt5qva-MuiCardContent-root"
+      data-testid="card-content"
     >
       <div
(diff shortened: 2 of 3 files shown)


liliya-soroka:
image

liliya-soroka:
@katspaugh , the signer info  is missing, so execution returns an error GS020
Steps:
1. connect to the child safe as a parent safe owner
2. sign a new tx as parent safe owner
3. go to the parent safe(2/N) to provide the second signature and execute ApproveHash tx
4. the execution shows GS020 error
5. check details - no signature from the step 2
6. check tx service - https://safe-transaction-sepolia.staging.5afe.dev/api/v1/multisig-transactions/0xcbf45521d05ab65fc45ad1385c9ab7ed8d74206d14dbcc1db11ea62f48102f67/ - signature information is here
Current result: 
 https://fix_nested--walletweb.review.5afe.dev/transactions/tx?safe=sep:0x2a73e61bd15b25B6958b4DA3bfc759ca4db249b9&id=multisig_0x2a73e61bd15b25B6958b4DA3bfc759ca4db249b9_0xcbf45521d05ab65fc45ad1385c9ab7ed8d74206d14dbcc1db11ea62f48102f67
 
 Reported to the wallet-infra team - https://5afe.slack.com/archives/C04MJJM9S3U/p1738168543580799

francovenica:
I have an issue that maybe is related to what Lily reported, but is reproduced in a different way

Safe child is 1/x
Safe parent is 2/x

Trying to create a tx and execute right away in the child safe shows this error:
image

katspaugh:
I'll merge this one because it just fixes decoding. Btw the same bug existed on prod. We can fix the signature problem in another issue.

PR #4857, opened 2025-01-31
Feat(Mobile): import signers

clovisdasilvaneto:
## What it solves
It creates the import signers flow.

## How to test it
- Go to settings
- Click in the "Signers >" optiojn
- Click in "Import Signer
- paste a valid private key
- click in continue
- If your private key is from a signer of the safe you're using, you should see the success screen, otherwise the error screen.
- 
## Checklist


apps/mobile/app/_layout.tsx
```diff
@@ -61,6 +62,31 @@ function RootLayout() {
...
+                          headerShown: false,
+                        }}
+                      />
+                      <Stack.Screen
+                        name="import-signers/import-private-key-error"
+                        options={{
+                          presentation: 'modal',
+                          headerShown: false,
+                        }}
+                      />
```
compojoom, line 89: Just a nit. 

We'll have quite some typing with those names. 
the folder name is import-signers. Repeating import in the file name afterwards looks redundant. 
import-signers/private-key
import-signers/private-key-success
import-signers/loading
import-signers/private-key-error

convey the same meaning in my opinion.

apps/mobile/app/import-signers/index.tsx
```diff
@@ -2,60 +2,66 @@ import { SectionTitle } from '@/src/components/Title'
...
- * waiting for the design to exprot - figma is crazy
- */
 const items = [
   {
     name: 'seed',
     title: 'Import seed phrase or a private key',
     description: 'Enter a private key or a 12-24 word seed phrase.',
     icon: <SafeFontIcon name="wallet" size={16} />,
     Image: Seed,
+    imageProps: { marginBottom: -31 },
```
compojoom, line 21: why is it -31 here and -48 for the next image. Is this going to look ok on a smaller, bigger display? I guess you are adding those because of the safe area?

apps/mobile/src/features/ImportPrivateKey/ImportPrivateKey.test.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import { render, screen, fireEvent, act, waitFor } from '@/src/tests/test-utils'
+import { ImportPrivateKey } from './ImportPrivateKey.container'
+
+jest.mock('@/src/navigation/useScrollableHeader', () => ({
+  useScrollableHeader: () => ({
+    handleScroll: jest.fn(),
+  }),
+}))
```
compojoom, line 9: maybe we should move this to a global mock. We have a lot of screens with this hook now.

apps/mobile/src/features/ImportPrivateKey/ImportPrivateKey.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  it('enables import button when private key is entered', async () => {
+    render(<ImportPrivateKey />)
+
+    const input = screen.getByPlaceholderText('Paste here or type...')
+    const button = screen.getByText('Import signer')
+
+    act(() => fireEvent.press(button))
+
+    await waitFor(() => {
+      expect(screen.getByTestId('safe-input').props.style.borderTopColor).toBe('#FF5F72')
```
compojoom, line 32: I don't like this. I see that we did the same in the SafeWalletInput PR, but the problem is that the moment we go for a different border color those tests will fail. I think that instead we should read the color variable in tamagui here.

apps/mobile/src/features/ImportPrivateKey/components/ImportError/ImportError.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { Text, View } from 'tamagui'
+
+export function ImportError() {
+  return (
+    <View flex={1} paddingBottom={80} justifyContent="space-between">
+      <View flex={1}>
+        <ScrollView contentContainerStyle={{ flexGrow: 1 }}>
+          <View flex={1} flexGrow={1} alignItems="center" justifyContent="center" paddingHorizontal="$3">
+            <View flexDirection="row" alignItems="center" gap="$3">
+              <Identicon address="0x0000000000000000000000000000000000000000" size={64} />
```
compojoom, line 17: Why are we generating an identicon of the 0x0 address? Later we say - the address is not a signer of this account. I would expect that we generate an identicon from the provided private key

apps/mobile/src/features/ImportPrivateKey/components/ImportError/ImportError.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+              <Text fontSize="$4" color="$error">
+                . . . .
+              </Text>
+
+              <Badge
+                circleProps={{ backgroundColor: '#2F2527' }}
+                themeName="badge_success"
```
compojoom, line 25: badge success on error page?

apps/mobile/src/features/ImportPrivateKey/components/LoadingImport/LoadingImport.container.tsx
```diff
@@ -0,0 +1,56 @@
...
+export function LoadingImport() {
+  const { address } = useLocalSearchParams()
+  const dispatch = useAppDispatch()
+  const router = useRouter()
+
+  const activeSafe = useAppSelector(selectActiveSafe)
+  const { data } = useSafesGetSafeV1Query({
+    safeAddress: activeSafe.address,
+    chainId: activeSafe.chainId,
+  })
```
compojoom, line 18: What happens if that call fails with a network error? And why do we need to do a network request here? How did we get to the import page? We start from a safe, so we most probably have the safe in the store, no?

package.json
```diff
@@ -27,5 +27,8 @@
     "msw": "^2.7.0",
     "prettier": "^3.4.2"
   },
-  "packageManager": "yarn@4.5.3"
+  "packageManager": "yarn@4.5.3",
+  "dependencies": {
+    "expo-linear-gradient": "~14.0.2"
```
compojoom, line 32: this must be some kind of a mistake. It should be in apps/mobile

apps/mobile/src/features/ImportPrivateKey/ImportPrivateKey.test.tsx
clovisdasilvaneto: I'll use the input theme reference here 👍

apps/mobile/src/features/ImportPrivateKey/components/ImportError/ImportError.tsx
clovisdasilvaneto: Oh noh, this might be a result of the merge that I did, will fix it. thanks for catching it

apps/mobile/src/features/ImportPrivateKey/components/LoadingImport/LoadingImport.container.tsx
clovisdasilvaneto: No, because if the safe is added / removed during the time the user is importing it, it will not reflect here. So using rtk here will first give us the cached data and then will return the updated safe list, so there is no needing for adding it in the redux.

package.json
clovisdasilvaneto: true

apps/mobile/src/features/ImportPrivateKey/components/LoadingImport/LoadingImport.container.tsx
compojoom: How can a safe be added/removed during the time the user is importing a signer? 

I'm fine with leaving the code as is, I just don't understand how the above could happen?

apps/mobile/src/features/ImportPrivateKey/components/ImportError/ImportError.tsx
clovisdasilvaneto: 🔥

apps/mobile/src/features/ImportPrivateKey/components/LoadingImport/LoadingImport.container.tsx
clovisdasilvaneto: It can happen in the following scenario:

Suppose it is a safe with some owners and small threshold

- User opens the import signer screen but takes longer to import the private key, maybe he went to the bath or something 😆 
- One of the signers removed the signer the user wanted to add there
- Suppose the other signers were faster than the app user to approve this remove signer transaction
- The signer is removed
- User is still on this screen, and when he tries to add the signer, the app would add the signer however the signer is not part in the safe anymore, and since we list the signers in the signer list, the user would have "you do not have signers imported" message again, what would be weird.

But I agree with you, it might not happen though. Wdyt?

compojoom (approved):
Great stuff @clovisdasilvaneto !

PR #4858, opened 2025-01-31
Feat(Safenet): Safenet transaction simulations

schmanu:
Related to https://github.com/safe-global/safenet/pull/318

## What it solves
The standard Tenderly simulations do not work with Safenet transactions because
- The Safe does not have the sufficient balance on the current chain (but will get them provided by Safenet)
- The Safenet guard blocks transaction execution

## How this PR fixes it
Therefore we have a new simulation endpoint that simulates the transaction through the EntryPoint.
This PR uses that new endpoint to do transaction simulations.

### Other changes
The new endpoint only sends the required data instead of the full result as results tend to have > 2MB.
Therefore this PR adds a new type `RequiredTenderySimulation` which only has the fields that we actually use in components.

## How to test it
1. Open a Safenet enabled Safe
2. Create any transaction. Either one that requires funds from other chains or a regular Safe transaction
3. Simulate the transaction
4. Check the result on tenderly and observe that it includes the spending transactions / is simulated through the EntryPoint

## Checklist


apps/web/src/components/tx/security/tenderly/useSimulation.ts
```diff
@@ -44,10 +51,35 @@ export const useSimulation = (): UseSimulationReturn => {
...
+            throw new Error('Batch execution is not possible with Safenet Accounts')
+          }
+          const safeTxHash = calculateSafeTransactionHash(
+            params.safe.address.value,
+            params.transactions.data,
+            params.safe.version ?? LATEST_SAFE_VERSION,
+            BigInt(params.safe.chainId),
+          )
+
+          const { data: safeNetSimulation } = await simulateSafenet({
```
germartinez, line 71:
```suggestion
          const { data: safenetSimulationData } = await simulateSafenet({
```

apps/web/src/components/tx/security/tenderly/useSimulation.ts
```diff
@@ -44,10 +51,35 @@ export const useSimulation = (): UseSimulationReturn => {
...
+            chainId: params.safe.chainId,
+            tx: {
+              safe: params.safe.address.value,
+              safeTxHash,
+            },
+          })
 
-        const data = await getSimulation(simulationPayload, tenderly)
+          if (safeNetSimulation) {
+            data = safeNetSimulation
```
germartinez, line 80:
```suggestion
          if (safenetSimulationData) {
            data = safenetSimulationData
```

germartinez (approved):

PR #4861, opened 2025-01-31
chore(dependencies): set explicit version for @safe-global/safe-deployments

dasanra:
## What it solves

Ensures safe-deployments version used is the expected one

## How this PR fixes it
Set a explicit version for @safe-global/safe-deployments

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -55,6 +55,7 @@
     "@safe-global/protocol-kit": "^4.1.4",
     "@safe-global/safe-apps-sdk": "^9.1.0",
     "@safe-global/safe-client-gateway-sdk": "v1.60.1",
+    "@safe-global/safe-deployments": "^1.37.26",
     "@safe-global/safe-gateway-typescript-sdk": "3.22.7",
     "@safe-global/safe-modules-deployments": "^2.2.5",
     "@safe-global/store": "workspace:^",
--- a/yarn.lock
+++ b/yarn.lock
@@ -8337,16 +8337,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@safe-global/safe-deployments@npm:^1.37.22":
-  version: 1.37.22
-  resolution: "@safe-global/safe-deployments@npm:1.37.22"
-  dependencies:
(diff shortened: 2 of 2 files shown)


compojoom:
How about we use the resolutions in the global package.json?
    "@safe-global/safe-deployments": "1.37.26"

It looks like we end up with 2 versions of safe-deployments in yarn.lock because 2 different versions are needed by the packages that get included.

katspaugh:
@dasanra why is it necessary? The idea was to fetch safe-deployments at the same version as protocol-kit because we know it depends on it.

dasanra:
@compojoom it shouldn't be necessary to add it explicitly but I consider yarn failed here to resolve the version.

The problem is that protocol-kit@v4 is the one explicitly installed, having the direct dependency with safe-deployments@1.37.26. This is also the package that does indeed use the dependency.

safe-deployments @1.37.22 is a dependency created by api-kit@2.5.6, not because the api-kit depends on it but because api-kit also depends in the protocol-kit which is the one having this dependency. As the web did not migrate yet to protocol-kit@v5 there is an inconsistency, because api-kit@2.5.X uses protocol-kit@v5, although in the package.json api-kit is matching v2.4.6. Using the ^ allows yarn to resolve to a higher minor version, which is fine, and it shouldn't be an issue as the api-kit only uses small utils from protocol-kit that weren't affected by breaking changes in v5. The problem is that yarn, for some reason don't apply this ^ logic to safe-deployments, as it should be fine to use the latest patch version, but instead is bundling safe-deployments v1.37.22, which it even makes less sense considering the app explicitly calls protocol-kit@v4.1.4 which says it needs at least safe-deployments@v1.37.26

As I was trying to mess the less I could with the dependencies I prefer this approach of just making sure it bumps safe-deployments affecting the less to the other dependencies to avoid unexpected side effects (not really sure if I can downgrade api-kit to v2.4.6 to avoid this inconsistency without adding safe-deployments explicitly as a dependency)

@katspaugh so the explanation above, it's picking the wrong safe-deployments version because for me yarn is failing to resolve the correct versions. So I explicitly set this one until web migrates to next major of protocol-kit

PR #4868, opened 2025-02-03
Fix(Spending limits): check selected token

katspaugh:
## What it solves

Resolves this bug report by @liliya-soroka:

> 1. The bug with defining the spending limits amount -
Steps;
> 2. add spending limits in a new safe
> 3. start the send funds and select spending limits

https://github.com/user-attachments/assets/28c115b8-c79e-4229-a6d0-80de589e61ba

> Current result: The amount is 0 undefined


liliya-soroka:
verified, the list is empty if the token is hidden under spam

PR #4873, opened 2025-02-04
Fix(Proposers): allow deleting tx proposals

katspaugh:
## What it solves

Transactions proposed by delegates were previously not deletable because the proposer value was wrong.

## How to test it

* Create a tx as a proposer
* Delete that tx as a proposer


compojoom (approved):
👍

katspaugh:
Well previously neither could. Now at least the proposer can. Generally the rule is: he who hath created a tx can also delete it.

PR #4875, opened 2025-02-04
refactor(permissions): Move types to config

tmjssz:
## What it solves

Move Role and Permission enums to the config file. This is to make it easier to find corresponding permission functions defined in the config.

The other permission related types had to be moved as well to avoid circular dependencies.


katspaugh (approved):
Thanks!

PR #4878, opened 2025-02-04
feat(mobile): add integration with gateway backend for notifications feature

Jonathansoufer:
## What it solves

Tis PR implements the logic necessary to integrate (enable/disable) notifications from the backend (CGW) perspective, as well as making sure Firebase integration/flow works.

Resolves #

## How to test it

P.S: Since we're **pausing** this PR up the point where we register de device on the database, but unable to receive notifications from there, the only way to test this PR is triggering a Push Notification from the Firebase Dashboard.

## Screenshots

https://github.com/user-attachments/assets/1988125f-a1b0-4f43-bab0-f00336fda6f2

## Checklist

### DIFF
--- a/apps/mobile/src/hooks/useNotificationManager.ts
+++ b/apps/mobile/src/hooks/useNotificationManager.ts
@@ -53,7 +53,7 @@ export const useNotificationManager = () => {
       const deviceNotificationStatus = await NotificationsService.isDeviceNotificationEnabled()
 
       if (!deviceNotificationStatus && !isAppNotificationEnabled) {
-        await NotificationsService.requestPushNotificationsPermission()
+        await NotificationsService.getAllPermissions()
       } else if (deviceNotificationStatus && !isAppNotificationEnabled) {
         await registerForNotifications()
       } else {


apps/mobile/src/hooks/useDelegateKey.ts
```diff
@@ -0,0 +1,117 @@
...
+
+      // Step 0: Get users private key from keychain. TODO: need to pass ownerAddress to getPrivateKey
+      const privateKey = await getPrivateKey()
+
+      if (!privateKey) {
+        throw Logger.error('useDelegateKey: Something went wrong', error)
+      }
+
+      // Step 1: Generate delegate key
+      const delegateMnemonic = ethers.Wallet.createRandom().mnemonic?.phrase
```
compojoom, line 34: Why do we need mnemonic here? Why don#t we go for creating a random private key?
```
const wallet = Wallet.createRandom(provider);
console.log(wallet.privatekey)
```

apps/mobile/src/hooks/useDelegateKey.ts
```diff
@@ -0,0 +1,117 @@
...
+
+      // Step 4: Send to backend
+      await createOnBackEnd(ownerAddress, delegatedAccount, signature)
+
+      // Step 5: Store delegate key in keychain
+      const encryptedDelegateKey = await DeviceCrypto.encrypt(asymmetricKey, delegatedAccount.privateKey, {
+        biometryTitle: 'Authenticate',
+        biometrySubTitle: 'Saving delegated key',
+        biometryDescription: 'Please authenticate yourself',
+      })
```
compojoom, line 65: we don't need this kind of protection for the delegate key. It is not going to sign transactions and can be exchanged at any point.

compojoom: No, I meant that we don't need biometry protection. Users should be able to remove the keys as in the old app.

apps/mobile/package.json
```diff
@@ -90,6 +90,7 @@
     "react-native-device-info": "^14.0.1",
     "react-native-draggable-flatlist": "^4.0.1",
     "react-native-gesture-handler": "~2.20.2",
+    "react-native-get-random-values": "^1.11.0",
```
compojoom, line 93: before we add this package I would actually want to know why is quick-crypto not doing this. Maybe we are calling the polyfills from it too late?

apps/mobile/src/hooks/useDelegateKey.ts
```diff
@@ -0,0 +1,176 @@
...
+  // Step 0 - Get the nonce to be included in the message to be sent to the backend
+  const { data } = useAuthGetNonceV1Query()
+
+  const createDelegate = useCallback(
+    async (ownerAddress: Address) => {
+      setLoading(true)
+      setError(null)
+      const nonce = data?.nonce
+      // Step 1 - Try to get the owner's private key
+      const ownerPrivateKey = await getPrivateKey()
```
compojoom, line 48: How is this working? Aren't we supposed to pass an user id for this function? 

```
const getPrivateKey = async (userId: string) => {
    try {
      const keychainKey = getKeychainKey(userId)
      const user = await Keychain.getGenericPassword({
        service: keychainKey,
      })
      if (!user) {
        throw 'user password not found'
      }
      const { encryptyedPassword, iv } = JSON.parse(user.password)
      const decryptedKey = await DeviceCrypto.decrypt(userId, encryptyedPassword, iv, {
        biometryTitle: 'Authenticate',
        biometrySubTitle: 'Signing',
        biometryDescription: 'Authenticate yourself to sign the text',
      })
      return decryptedKey
    } catch (err) {
      console.log(err)
    }
}
```

apps/mobile/src/hooks/useDelegateKey.ts
```diff
@@ -0,0 +1,176 @@
...
+        const siweMessage = `SafeWallet wants you to sign in with your Ethereum account:
+${signerAccount.address}
+
+Sign in with Ethereum to the app.
+
+URI: https://safe.global
+Version: 1
+Chain ID: ${activeSafe.chainId}
+Nonce: ${nonce}
+Issued At: ${new Date().toISOString()}`
```
compojoom, line 73: I would extract this in a separate function for example generateSiwseMessage

apps/mobile/src/hooks/useDelegateKey.ts
```diff
@@ -0,0 +1,176 @@
...
+          throw Logger.info(ERROR_MSG)
+        }
+
+        // Step 2 - Create a new random (delegated) private key in case the owner's private key is not available
+        //TODO: Double check if we have a wallet stored already avoiding to create a new one
+        const signerAccount = ownerPrivateKey ? new Wallet(ownerPrivateKey) : Wallet.createRandom()
+
+        if (!signerAccount) {
+          throw Logger.error(ERROR_MSG, error)
+        }
```
compojoom, line 61: this could be moved to a getSignerAccount function

apps/mobile/src/hooks/useDelegateKey.ts
```diff
@@ -0,0 +1,176 @@
...
+  const createDelegate = useCallback(
+    async (ownerAddress: Address) => {
+      setLoading(true)
+      setError(null)
+      const nonce = data?.nonce
+      // Step 1 - Try to get the owner's private key
+      const ownerPrivateKey = await getPrivateKey()
+
+      try {
+        if (!ownerAddress || !activeSafe || !nonce || !fcmToken) {
```
compojoom, line 51: could be moved to a validate function.

compojoom: I think that if the try block is structured as:
```
        validateParameters(...)

        const ownerPrivateKey = await getPrivateKey(...)
        const signerAccount = await createSignerAccount(ownerPrivateKey)
        if (!signerAccount) throw new Error(ERROR_MSG)

        const siweMessage = generateSiweMessage(...)

        // Verify the delegate on the backend and subscribe for notifications.
        await verifyAndSubscribeDelegate(...)

```
it would make it way easier to read and follow

compojoom: Also what you get is separation of conserns. We abstract everything in small, testable functions. It not only improves readibility, but also makes it easier to unit test.

compojoom (changes requested):
Overall looks very good 👍 !  I sadly could not test it yet as some of your dependencies are not 100% correct in the PR. (I guess due to a rebase or something)

compojoom:
@Jonathansoufer - I built the app locally and installed on my phone, but when I click on "turn on" notificatoins and navigate to the settings I don't have any notifications to choose from? The permission is not in the list?

apps/mobile/src/services/notifications/FCMService.ts
```diff
@@ -1,7 +1,9 @@
+//@ts-ignore
+globalThis.RNFB_SILENCE_MODULAR_DEPRECATION_WARNINGS = true
```
compojoom, line 2: I think we better move this to the entry point of the application. It's a bit weird to have this global var first in the FCMService. It seems to be a global thing that react-native-firebase "needs".

apps/mobile/src/app/_layout.tsx
```diff
@@ -1,3 +1,5 @@
+import '../../shim'
```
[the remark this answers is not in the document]
compojoom: Did you forget to push? github is still showing it.

apps/mobile/src/features/Notifications/Notifications.container.tsx
```diff
@@ -2,14 +2,20 @@ import React, { useCallback } from 'react'
...
 export const NotificationsContainer = () => {
   const dispatch = useAppDispatch()
+  const { deleteDelegate } = useDelegateKey()
   const isAppNotificationEnabled = useAppSelector(selectAppNotificationStatus)
 
-  const handleToggleAppNotifications = useCallback(() => {
+  const handleToggleAppNotifications = useCallback(async () => {
     dispatch(toggleAppNotifications(!isAppNotificationEnabled))
+    if (!isAppNotificationEnabled) {
+      await deleteDelegate()
```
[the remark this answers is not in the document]
compojoom: @Jonathansoufer I have my doubts that this if(!error) check is goin to work reliably. the useDelegateKey hook returns an error. here you await the deleteDelegate and thje deleteDelegate function might set the error, but if does the notificationContainer needs to rerender. I'm not sure how the flow of the code is going to be. Isn't delegate always be null, since you are in the callback and at the moment the callback has started it is null 🤷 

Wouldn't it be easier of the deleteDelegate and createDelegate return the error or throw?

apps/mobile/src/hooks/useGTW.ts
```diff
@@ -0,0 +1,120 @@
...
+            chainId,
+            createDelegateDto: {
+              safe: safeAddress,
+              delegator: signer.address,
+              delegate: delegatedAccount.address,
+              signature,
+              label: DELEGATED_ACCOUNT_TYPE.OWNER,
+            },
+          })
+        }
```
compojoom, line 75: I don't fully understand this block. Why does a regular user need to authenticate, but an owner doesn't?

apps/mobile/src/services/notifications/NotificationService.ts
```diff
@@ -87,6 +105,18 @@ class NotificationsService {
     return isAuthorized
   }
 
+  async openDeviceSettings() {
+    try {
+      if (Platform.OS === 'ios') {
+        Linking.openSettings()
+      } else {
+        notifee.openNotificationSettings()
```
compojoom, line 113: why is it necessary that we use the notifee openNotificationSettings? isn't Linking.openSettings() also working on android?

apps/mobile/src/features/Notifications/Notifications.container.tsx
```diff
@@ -1,15 +1,53 @@
...
+      if (!error) {
+        dispatch(toggleAppNotifications(!isAppNotificationEnabled))
+      }
+    }
+  }, [isAppNotificationEnabled])
+
+  useEffect(() => {
+    const subscription = AppState.addEventListener('change', async (nextAppState) => {
+      if (appState.current.match(/inactive|background/) && nextAppState === 'active') {
```
katspaugh, line 37: This code is repeated in two places, should we make a hook for it?

apps/mobile/src/hooks/useDelegateKey.ts
```diff
@@ -0,0 +1,120 @@
...
+
+      // Step 2.3 - Define the signer account
+      const signerAccount = getSigner(safeOwnerPK, randomDelegatedAccount)
+
+      // Step 3 - Create a message following the SIWE standard
+      const siweMessage = createSiweMessage({
+        address: signerAccount.address,
+        chainId: Number(activeSafe.chainId),
+        nonce,
+        statement: 'SafeWallet wants you to sign in with your Ethereum account',
```
katspaugh, line 77:
```suggestion
        statement: 'Safe Wallet wants you to sign in with your Ethereum account',
```

apps/mobile/src/app/_layout.tsx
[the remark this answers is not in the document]
katspaugh: Still showing...

apps/mobile/src/hooks/useNotificationPayload.ts
```diff
@@ -20,11 +21,18 @@ export function useNotificationPayload(appSigners: Record<string, AddressInfo>)
   const fcmToken = useAppSelector(selectFCMToken)
 
   const getNotificationRegisterPayload = useCallback(
-    async (nonce: string | undefined) => {
+    async ({
+      nonce,
+      ramdomPK
```
compojoom, line 26: I don't like that we pass the random PK here. Can't we pass a signer instead and move the calculation of the signer to the useRegisterForNotificaiton hook?

Having to find the owner of the safe and then going for a random PK if none is found is a weird thing for the getNotificationRegisterPayload function. 

this fucntion also returns accountType, but it doesn't seem that the account type is actually needed for the SiWE message, rather it is needed later when proposing to the backend. Same goes for the fcmToken and signer?

If you need those things up, you will have a pretty simple funciton signature
nonce, signer. and the function would return just the payload (which is currently implied by the function name)

apps/mobile/src/hooks/useNotificationGTWPermissions.ts
```diff
@@ -0,0 +1,22 @@
...
+export function useNotificationGTWPermissions() {
+    const appSigners = useAppSelector(selectSigners)
+    const activeSafe = useDefinedActiveSafe()
+    const activeSafeInfo = useAppSelector((state: RootState) => selectSafeInfo(state, activeSafe.address))
+    const ownerFound = activeSafeInfo.SafeInfo.owners.find((owner) => appSigners[owner.value]) ?? null
+    const accountType = ownerFound ? NOTIFICATION_ACCOUNT_TYPE.OWNER : NOTIFICATION_ACCOUNT_TYPE.REGULAR
+
+    const getAccountType = useCallback(() => {
+        return accountType
+    }, [accountType])
```
compojoom, line 19: wouldn't it make more sense to have the activeSafeInfo.SafeInfo.owners as dependency here and do the onwerFound and accountType calculation inside?

Right now the func will recalculate whenever something in the safe changes.

apps/mobile/src/hooks/useNotificationManager.ts
```diff
@@ -0,0 +1,92 @@
+// src/hooks/useNotificationManager.ts
```
compojoom, line 1:
```suggestion

```

apps/mobile/src/hooks/useNotificationPayload.ts
```diff
@@ -0,0 +1,56 @@
...
+    async ({
+      nonce,
+      signer,
+    }: {
+      nonce: string | undefined;
+      signer: Wallet | HDNodeWallet;
+    }) => {
+      // 2. Construct payload to register for notifications
+      // 2.1 - Get the nonce to be included in the SiWe as per
+      // https://www.notion.so/safe-global/Authentication-implementation-f26c6040be5148748ac19655da5e4842
```
compojoom, line 29: this comment is no longer at the correct place

apps/mobile/src/hooks/useRegisterForNotifications.ts
```diff
@@ -0,0 +1,221 @@
...
+      await FCMService.registerAppWithFCM()
+      await FCMService.saveFCMToken()
+      const fcmToken = await FCMService.getFCMToken()
+      FCMService.listenForMessagesBackground()
+
+      /* Step 2 - Check if the imported signers is the owner of the active safe.
+      * Based on the previous step, we determine the account type, which will be used to register for notifications types
+      */
+      const ownerFound = activeSafeInfo.SafeInfo.owners.find((owner) => appSigners[owner.value]) ?? null
+      const accountType = ownerFound ? NOTIFICATION_ACCOUNT_TYPE.OWNER : NOTIFICATION_ACCOUNT_TYPE.REGULAR
```
compojoom, line 76: maybe we can use the useNotificationGTWPermissions hook here now?

apps/mobile/src/features/Notifications/components/NotificationPermissions.tsx
```diff
@@ -0,0 +1,44 @@
...
+            Queued transactions
+          </Text>
+        </View>
+        {!isOwner && (
+          <View flexDirection="row" alignItems="center" gap={8} marginBottom="$4">
+            <Text fontWeight={400} color={'$colorSecondary'} fontSize="$3">
+              You need to import at least one signer to receive transaction requests.
+            </Text>
+          </View>
+        )}
```
compojoom, line 40: Is this still working if you enable push notifications, but haven't imported a PK. Then you import one and subscribe - are we already updating the permissions on cgw?

apps/mobile/src/hooks/useNotificationPayload.test.ts
```diff
@@ -0,0 +1,57 @@
...
+jest.mock('@/src/utils/logger')
+
+describe('useNotificationPayload', () => {
+  const mockCreateSiweMessage = jest.fn()
+  const mockUseAuthGetNonceV1Query = useAuthGetNonceV1Query as jest.Mock
+  const mockUseSiwe = useSiwe as jest.Mock
+  const mockUseDefinedActiveSafe = useDefinedActiveSafe as jest.Mock
+
+  beforeEach(() => {
+    mockUseAuthGetNonceV1Query.mockReturnValue({ data: 'mockNonce' })
```
compojoom, line 30: fyi: you should be able to use msw and just mock a return object for the nonce endpoint. 

by defining a handler in config/test/msw/handlers.ts this mocked data would be available in any test you do and you won't have to explicitly mock @safe-global/store/gateway/AUTO_GENERATED/auth

apps/mobile/src/hooks/useRegisterForNotifications.ts
```diff
@@ -0,0 +1,272 @@
...
+   */
+  const registerForNotifications = useCallback(async () => {
+    try {
+      setLoading(true)
+      setError(null)
+      // Step 1 - Set up Firebase Cloud Messaging steps
+      await FCMService.registerAppWithFCM()
+      await FCMService.saveFCMToken()
+      const fcmToken = await FCMService.getFCMToken()
+      FCMService.listenForMessagesBackground()
```
compojoom, line 65: if you create a function `setUpFirebaseCloudMessaging` and move this code to it, there won't be a need for the comment as the title of the function already explains what is going on.

apps/mobile/src/hooks/useRegisterForNotifications.ts
```diff
@@ -0,0 +1,272 @@
...
+
+      /* Step 3 - Create a new random (delegated) private key to avoid exposing the subscriber's private key
```
compojoom, line 67: Since we changed the code order, the comments no longer reflect the code ( I mean the steps..) where idd step 2 and 4 go :D ?

Becuase of that I view comments as debt as they have to be kept up to date with the code... :)

apps/mobile/src/hooks/useRegisterForNotifications.ts
```diff
@@ -0,0 +1,272 @@
...
+       *
+       * This key will be used to register for notifications on the backend
+       * avoiding the prompt to grant notifications permission
+       */
+      const randomDelegatedAccount = Wallet.createRandom()
```
compojoom, line 72: I don't understand one thing here. Where are we storing the private key for this random address?

apps/mobile/src/hooks/useRegisterForNotifications.ts
```diff
@@ -0,0 +1,272 @@
...
+
+      // Step 5 - Store the random delegated account in the Redux store
+      dispatch(
+        addOrUpdateDelegatedAccount({
+          accountDetails,
+          safes: [activeSafe],
+        }),
+      )
+
+      const signer = getSigner(randomDelegatedAccount.privateKey)
```
compojoom, line 97: isn't wallet.createRandom equal to new Wallet(PK)? So the random Account is the signer?

apps/mobile/src/hooks/useSign/useSign.ts
```diff
@@ -15,7 +15,7 @@ export const storePrivateKey = async (userId: string, privateKey: string) => {
       invalidateOnNewBiometry: true,
     })
 
-    const encryptyedPrivateKey = await DeviceCrypto.encrypt(userId, privateKey, {
+    const encryptedPrivateKey = await DeviceCrypto.encrypt(userId, privateKey, {
```
compojoom, line 18: just a nit: Those changes should have been a separate PR. The implications of this change mostly later in the const {encryptedPassword, iv } ... is that previously stored private keys are no longer properly decrypted :) When we merge this PR it won't be obvious why a previously imported PK would no longer work since there is no commit that does anything with the PK.

compojoom (changes requested):
Looks very good. and I think that we could merge it after few small changes
- remove the second shim import in _layout.tsx :)
- move the delegateKey PK to the keychain with default protection (no need for biometrics, the app needs to be able to access it in the background)
- once a user is subscribed for the notifications the bell icon on the dashboard does nothing. A quick fix would be to redirect to the upcoming "notification center" screen in the design with a text "comming soon"

apps/mobile/src/services/key-storage/key-storage.service.ts
```diff
@@ -0,0 +1,181 @@
...
+        accessible: Keychain.ACCESSIBLE.WHEN_UNLOCKED_THIS_DEVICE_ONLY,
+        service: keychainKey,
+      },
+    )
+
+    // This enrols the biometry authentication in the device
+    await Keychain.getGenericPassword({
+      accessControl: Keychain.ACCESS_CONTROL.BIOMETRY_CURRENT_SET_OR_DEVICE_PASSCODE,
+      service: keychainKey,
+    })
```
compojoom, line 107: I don't understand this. Why are we accessing the password we just stored here?

compojoom (approved):
Thank you for your hard work on this! 🔥 
There are few small things that we would have to still address, but we can do this in separate PR. 

👍

packages/store/src/gateway/cgwClient.ts
```diff
@@ -11,6 +11,12 @@ export const getBaseUrl = () => {
 }
 export const rawBaseQuery = fetchBaseQuery({
   baseUrl: '/',
+  credentials: 'include',
+  headers: {
+    'Content-Type': 'application/json',
+    Accept: 'application/json',
+    'Set-Cookie': 'HttpOnly;Secure;SameSite=None',
```
schmanu, line 18: This has no effect as it is a response header. This needs to be set by the service when setting e.g. an auth cookie.

PR #4883, opened 2025-02-05
Refactor(web): faster SafeInfo loading

katspaugh:
## What it solves

Improves the initial page loading time by optimizing how we load the SafeInfo.

* Decouple React dependencies between undeployed safes and regular safes, so SafeInfo loading doesn't depend on chain configs – it will load in parallel with it
* Use `location.search` to get a safe address from the URL faster

Before:
Screenshot 2025-02-05 at 11 37 43

After:
Screenshot 2025-02-05 at 17 05 21


katspaugh:
Merging directly as it was trivial to test.

PR #4911, opened 2025-02-07
refactor(mobile): move screens inside of src

compojoom:
It turns out that expo is fine with having the app folder inside of the src folder (I always assumed that it has to live outside). It always bugged me that app is outside of where the other code was living. This way we have all the code sitting tight together :)

## What it solves
Mental overhead for searching for the screens.

## How to test it
Nothing changes, app should continue to function as usual.

## Checklist


katspaugh (approved):
👍

PR #4914, opened 2025-02-07
Chore(Web): make dev workflows depend on linter

katspaugh:
## What it solves

In the linter fails, everything else should not run. Otherwise it's hard to see what exactly failed.


katspaugh:
Closing because this approach runs workflows w/o reporting back to the PR.

PR #4925, opened 2025-02-10
chore(dependencies): bump protocol-kit and safe-deplyoments

dasanra:
## What it solves

Bumps protocol-kit and safe-modules-deployments to the latest versions to support new chains


katspaugh (approved):
Thanks!

PR #4928, opened 2025-02-10
Feat(Sidebar): always show current safe; consistent New transaction button

katspaugh:
Based on user feedback, see [proposal on notion](https://www.notion.so/safe-global/Sidebar-improvements-1968180fe57380abb911e2a1ec91fe46).

## What it solves

* The "New transaction" button is now always just "New transaction" and never "Add readonly" – this makes it consistent, especially when wallet is being connected.
* The currently opened Safe will always appear on top of the Safe list if it's not pinned. You can then pin it.
Screenshot 2025-02-10 at 15 00 28


jmealy (approved):

francovenica:
LGTM

Checked that the Current Safe account shows, well, the current safe
This space also only shows the current safe when it is a multichain safe (so if the safe is deployed in several networks it only shows the network you are currently in)
The "Current safe account" goes away if the safe is pinned. Is not described how it should work, but I agree with how it works
"Add read only" and "Remove account" buttons don't show up anymore. Now is the "New transaction" button shows grayed out.
The "New transaction button" still is enabled for spending limit users and proposers

Created this ticket to update the "Pinned safes" test cases here https://github.com/safe-global/safe-test-management/issues/31

Issue #4881, opened 2025-02-05
fix(Sub-Safes): rename “Subaccounts” and “nested (action)” to “Sub-Safe (action)”


iamacook:
## Links

- [Problem statement](https://www.notion.so/safe-global/Sub-Safes-easier-defi-access-maintaing-main-account-security-1208180fe5738018a9fdde8cd5423ed2)

## What must be done

All mention of “nested (action)” or “Subaccount” should be renamed to “Sub-Safe (action)”.

## Designs/sketches

- [Designs](https://www.figma.com/design/VX9wcNdEOj6sg4kH73bBP4/Sub-accounts)

## How to test it

- The creation of (and listing) of “Subaccounts” should be named “Sub-Safes
- All “nested” actions should be “Sub-Safe” actions

liliya-soroka:
Verified

PR #4932, opened 2025-02-10
feat: Nested Safes

### LINKED ISSUE #4879: feat: Sub-Safes
## What is the feature about

We want to natively support Sub-Safes (Safes owned by Safes) in the interface. This support should allow simple creation/funding of children, clearly indicate relationship(s) and allow easy navigation between them.

## The list of requirements

The following are to-be-created:

- A new context menu listing Sub-Safes
- Aforementioned context menu accessible via sidebar and Safe list drawer
- Sub-Safe creation (and funding) flow
- Aforementioned flow accessible from previously mentioned context menu and Safe settings
- All mention of “nested” actions referred to as “Sub-Safe” actions

Note: this feature is to-be-released to a pre-defined list Safes. It should be feature flagged according to a targeted outreach.

## Designs/sketches

- [Designs](https://www.figma.com/design/VX9wcNdEOj6sg4kH73bBP4/Sub-accounts)

## Links

- [Problem statement](https://www.notion.so/safe-global/Sub-Safes-easier-defi-access-maintaing-main-account-security-1208180fe5738018a9fdde8cd5423ed2)
- [Iteration](https://www.notion.so/safe-global/Subaccount-Iteration-18f8180fe57380b1bc67c6cc5edecb71)

iamacook:
## What it solves

Resolves #4879, #4881 inclusive

## How this PR fixes it

This implements "Nested Safes" ([previously "Subaccounts"](https://github.com/safe-global/safe-wallet-monorepo/pull/4456)).

A new sidebar icon, context menu (accessible via the icon/drawer) and settings have been added that both list and open a "Nested Safe" creation flow.

## How to test it

Open a Safe and observe:

- A new header icon in the sidebar that, when clicked, opens either a promo message (if no "Nested Safes" are owned) or a list, with an option to add more. (The presence of "Nested Safes" are signified by a badge number.)
- If the current Safe has "Nested Safes", a new item in the sidebar drawer menu (accessible via the three dots) should open the same component.
- A new settings area under "Setup" has the option to add "Nested Safes", and list/rename those which exist already.

## Screenshots

image

image

image

image

image

image

image

image

image

## Checklist


apps/web/src/components/sidebar/NestedSafesPopover/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+      </ModalDialogTitle>
+      <Box
+        sx={{
+          p: 3,
+          pt: 2,
+          display: 'flex',
+          flexDirection: 'column',
+          maxHeight: '590px',
+        }}
+      >
```
katspaugh, line 60: Hinweise: Box and Typography take sx props directly (which may look a bit neater).

```suggestion
      <Box
        p={3}
        pt={2}
        display="flex"
        flexDirection="column"
        maxHeight="590px"
      >
```

Though in this particular case, the Stack component would be even better.

apps/web/src/components/sidebar/SafeListContextMenu/index.tsx
```diff
@@ -78,6 +90,20 @@ const SafeListContextMenu = ({
         <MoreVertIcon sx={({ palette }) => ({ color: palette.border.main })} />
       </IconButton>
       <ContextMenu anchorEl={anchorEl} open={!!anchorEl} onClose={handleCloseContextMenu}>
+        {!undeployedSafe && nestedSafes?.safes && nestedSafes.safes.length > 0 && (
```
katspaugh, line 93:
```suggestion
        {!undeployedSafe && nestedSafes?.safes?.length > 0 && (
```

apps/web/src/components/tx-flow/flows/CreateNestedSafe/ReviewNestedSafe.tsx
```diff
@@ -0,0 +1,150 @@
...
+        address: predictedSafeAddress,
+        name: params.name,
+      }),
+    )
+  }
+
+  return (
+    <SignOrExecuteForm onSubmit={onSubmit}>
+      {predictedSafeAddress && (
+        <Grid
```
katspaugh, line 119: Grid is deprecated, it's better to use Grid2 for new code.

apps/web/src/components/sidebar/SafeListContextMenu/index.tsx
iamacook: TypeScript doesn't like that:

image

apps/web/src/components/tx-flow/flows/CreateNestedSafe/ReviewNestedSafe.tsx
```diff
@@ -0,0 +1,150 @@
...
+            <EthHashInfo
+              name={params.name}
+              address={predictedSafeAddress}
+              shortAddress={false}
+              hasExplorer
+              showCopyButton
+              showAvatar={false}
+            />
+          </Grid>
+        </Grid>
```
katspaugh, line 146: Our general approach for various "Review" components is to just create and pass a SafeTx to `SignOrExecuteForm` which in turn renders an appropriate ["confirmation view"](https://github.com/safe-global/safe-wallet-monorepo/blob/dev/apps/web/src/components/tx/confirmation-views/index.tsx) based on the transaction-preview payload. Please check if this works in this case, otherwise the other signers will not see this nice review component.

iamacook: Is there a defined pattern to pass arbitrary values to the the `ConfirmationView`? Here we need to show the predicted Safe address.

katspaugh: No, the point is to not pass and parameters because the goal is to render the same view for creators and signers of a tx. So you can rely only on the decoded txData.

apps/web/src/components/tx-flow/flows/CreateNestedSafe/ReviewNestedSafe.tsx
iamacook: Changed in 4b3cd6a762afbbce0d3bd543605654e8f8578b3f.

apps/web/src/components/tx-flow/flows/CreateNestedSafe/ReviewNestedSafe.tsx
iamacook: Included your suggestion in 4b3cd6a762afbbce0d3bd543605654e8f8578b3f. The Safe address is now predicted from the decoded data.

katspaugh (approved):
Thanks!

apps/web/src/components/sidebar/NestedSafeInfo/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </Avatar>
+        <Typography variant="body2" fontWeight={700}>
+          Nested Safes allow you to:
+        </Typography>
+      </Box>
+      <List sx={{ p: 0, display: 'flex', flexDirection: 'column', gap: 2 }}>
+        {[
+          'rebuild your organizational structure onchain',
+          'explore new DeFi opportunities without exposing your main Account',
+          'deploy specialized modules and enable easier access through passkeys and other signers',
```
liliya-soroka, line 44: Replace the 44  with "deploy specialized modules and extend Safe functionality"

iamacook: Changed in e5021b56804135a71b37c739f9f09c50141a9ad8.

liliya-soroka:
1. There is navigation problem when Add Nested safe is opened from the pinned safes list :
steps:
1. there are 2 not multichain safes in the side bar as pinned
2. open safe 1 as the current open
3. open side bar drawer menu and click 3dotcs for the second safe
4. click Add Nested safe button 
Current result: Create a nested safe flow is started in the safe 1 instead of the safe from the step 3
I would suggest to hide " Add Nested safe" button or switch to the correct safe before opeining the Add Nested safe 

https://github.com/user-attachments/assets/4edbb913-6718-4800-bb73-d80f047aecb3

liliya-soroka:
2. Navigation to the nested safe from the sidebar drawer menu is not completed:
- menu displayed 
- I would prefere to see the dashboard of the opened nested safe and closed sidebar drawer menu
https://github.com/user-attachments/assets/bfb923c5-c2ea-46fd-9927-55a1758ee4ab

PR #4936, opened 2025-02-11
feat: Nested Safe breadcrumbs

### LINKED ISSUE #4880: feat(Sub-Safes): navigation between parent <> child Safes
## Links

- [Problem statement](https://www.notion.so/safe-global/Sub-Safes-easier-defi-access-maintaing-main-account-security-1208180fe5738018a9fdde8cd5423ed2)

## What must be done

The [initial implementation](https://github.com/safe-global/safe-wallet-monorepo/pull/4456) of “Subaccounts” did not include navigation between parent <> children. This has since been deemed a requirement of the feature.

A banner should be added to Sub-Safes with breadcrumbs that allow navigation back to the parent. This will be shown according to a search parameter containing the parent address.

Note: this will only be shown for immediate relations

## Designs/sketches

- [Designs](https://www.figma.com/design/VX9wcNdEOj6sg4kH73bBP4/Sub-accounts)

## How to test it

- When navigating from a parent to a child, a banner is shown with a link back to the parent
- Back/forward should work as expected

iamacook:
## What it solves

Resolves #4880

## How this PR fixes it

This adds navigation between parent <> child Safes by means of breadcrumbs according to [the designs](https://www.figma.com/design/VX9wcNdEOj6sg4kH73bBP4/Sub-accounts?node-id=11839-10918&t=I5XKQGSCLhMNsGs5-0).

When navigating to a Nested Safe, either via the sidebar or drawer, a new `parent` search parameter is appended to the URL and referenced to display the breadcrumbs.

Note: this only works with direct relations.

## How to test it

Open a Nested Safe via the sidebar/drawer and observe the new breadcrumbs, from which you can navigate back to the parent.

## Screenshots

image

image

## Checklist


katspaugh:
* I created a [nested safe](https://feat_nested_safe_breadcrumbs--walletweb.review.5afe.dev/home?safe=sep:0xF5dD6e5A4c96CA9549ED09efBf7D85E4ED4Cf7eE&parent=) but don't see the breadcrumbs, what am I doing wrong?
* As per the design, the parent link is grey and the current safe is black. I don't get this. Shouldn't it be vice versa? An active link (the parent safe) black, and the current unclickable safe greyed out?
* Looking at the code, there's a lot of places where `parent` is added to the query. Why is that necessary? Can't the parent address be inferred from the owner list? Or perhaps from the localStorage? I would try to reduce the code changes to the minimum and not complicate the maintenance of the internal routing/linking.

iamacook:
> * I created a [nested safe](https://feat_nested_safe_breadcrumbs--walletweb.review.5afe.dev/home?safe=sep:0xF5dD6e5A4c96CA9549ED09efBf7D85E4ED4Cf7eE&parent=) but don't see the breadcrumbs, what am I doing wrong?

As mentioned in the description, breadcrumbs are only shown when navigating from a parent to a child. This was implemented intentionally to avoid the overhead of checking every owner against the backend on each visit. If you have a more efficient approach in mind, I'm open to suggestions.

> * As per the design, the parent link is grey and the current safe is black. I don't get this. Shouldn't it be vice versa? An active link (the parent safe) black, and the current unclickable safe greyed out?

cc @IvaLukan for design input.

> * Looking at the code, there's a lot of places where `parent` is added to the query. Why is that necessary? Can't the parent address be inferred from the owner list? Or perhaps from the localStorage? I would try to reduce the code changes to the minimum and not complicate the maintenance of the internal routing/linking.
The places modified already overwrite query parameters before my changes, so explicitly setting the parent parameter follows that existing pattern. That said, I agree that simplifying the logic in general would be beneficial.

If we cache locally, we'd still need to check for ownership changes, and links would no longer be shareable - unless we check each owner against the backend. An alternative could be a centralised helper function, which might also improve the existing handling of the `safe` parameter. What do you think?

I'll mark this as a draft and revisit these points as this epic is not yet in our sprint. Appreciate the input.

katspaugh:
I would go for checking the owners on safe load, and not appending `parent` to every safe link.

iamacook:
> I would go for checking the owners on safe load, and not appending `parent` to every safe link.
> 
> Ideally it should be on the backend, a part of SafeInfo or SafeOverview like “parentSafe” (but not now of course).

I tweaked the implementation to check whether singular owners are Safes against the backend and it works smoothly. I considered caching the response but it works relatively well with RTK. Please let me know your thoughts.

I'll follow up re. the link colour and update accordingly.

IvaLukan:
> * As per the design, the parent link is grey and the current safe is black. I don't get this. Shouldn't it be vice versa? An active link (the parent safe) black, and the current unclickable safe greyed out?

The Ux principle of having the current safe black is to display the current ‘location’ - that’s why it is visually dominant since its job is to indicate where user currently is. At the same time the parent account is grey (still clickable) not to visually compete with ‘active’ account to indicate a secondary option, but not the main focus.

Since both patterns are commonly used across apps, I’m not against revisiting this if you have a strong opinion on it.

katspaugh (approved):
👍

PR #4947, opened 2025-02-11
Fix(Tx history filter): incoming amount with decimals

### LINKED ISSUE #4942: [Tx filtering] Wrong decimals when filtering by amount
A bug was reported by a user that filtering incoming USDC txs by value doesn't work correctly.

Upon inspecting the request and response, it looks like the frontend is sending a request to CGW with a `value` converted to decimals with **too many zeroes** because it takes the default 18 decimals which is not correct for USDC.

Given that the filter is supposed to work without specifying a token address, i.e. only by amount, we cannot predict how many decimals a token must have.

I would suggest to fix this on the backend by allowing to specify an amount in its original formatted form, so that the history is filtered post-formatting.

E.g.

Txs:
* Received 10 ETH 
* Received 1500 USDC
* Received 1500 ETH
* Received 100 USDT

When filtering by `amount: 1500`, it should find both the `1500 USDC` and `1500 ETH` transfers even though these tokens have different decimals.

## How to reproduce

* Go to a Safe with incoming USDC transfers
* Open tx history
* Filter by Incoming + Amount = some amount
* 0 txs are shown

## Links
A backend ticket: https://github.com/safe-global/safe-client-gateway/issues/2340

katspaugh:
## What it solves

Resolves #4942

## How this PR fixes it

For the incoming filter, the amount now must be specified in a decimal format (with zeroes) and not assume the default 18 decimals. This allows filtering by amounts for various tokens w/o specifying a token address.

Screenshot 2025-02-11 at 12 20 08

## How to test it

* Open a Safe with incoming USDC transfers
* Filter by incoming with an amount with 6 trailing zeroes
* Filter for incoming ETH transfers with 18 zeroes

### DIFF
--- a/apps/web/src/components/transactions/TxFilterForm/index.tsx
+++ b/apps/web/src/components/transactions/TxFilterForm/index.tsx
@@ -22,6 +22,7 @@ import NumberField from '@/components/common/NumberField'
 
 import css from './styles.module.css'
 import inputCss from '@/styles/inputs.module.css'
+import AddressInput from '@/components/common/AddressInput'
 
 enum TxFilterFormFieldNames {
   FILTER_TYPE = 'type',
@@ -172,13 +173,14 @@ const TxFilterForm = ({ toggleFilter }: { toggleFilter: () => void }): ReactElem
                           }}
                         />
                       </Grid>
+
                       <Grid item xs={12} md={6}>
                         <Controller
                           name={TxFilterFormFieldNames.AMOUNT}
                           control={control}
                           rules={{
                             validate: (val: TxFilterFormState[TxFilterFormFieldNames.AMOUNT]) => {
-                              if (val.length > 0) {
+                              if (val?.length > 0) {
                                 return validateAmount(val)
                               }
                             },
@@ -189,7 +191,9 @@ const TxFilterForm = ({ toggleFilter }: { toggleFilter: () => void }): ReactElem
                               className={inputCss.input}
                               label={
                                 fieldState.error?.message ||
-                                `Amount${isMultisigFilter && chain ? ` (only ${chain.nativeCurrency.symbol})` : ''}`
+                                (isIncomingFilter
+                                  ? 'Amount (with decimals)'
+                                  : `Amount (only ${chain?.nativeCurrency.symbol || 'ETH'})`)
                               }
                               error={!!fieldState.error}
                               {...field}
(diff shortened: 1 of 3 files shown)


apps/web/src/components/transactions/TxFilterForm/index.tsx
```diff
@@ -189,7 +191,9 @@ const TxFilterForm = ({ toggleFilter }: { toggleFilter: () => void }): ReactElem
                               className={inputCss.input}
                               label={
                                 fieldState.error?.message ||
-                                `Amount${isMultisigFilter && chain ? ` (only ${chain.nativeCurrency.symbol})` : ''}`
+                                (isIncomingFilter
+                                  ? 'Amount (with decimals)'
```
katspaugh, line 195: Not sure if "with decimals" is clear enough. Maybe it should say "with zeroes"?

apps/web/src/utils/tx-history-filter.ts
```diff
@@ -45,6 +45,10 @@ export const _isModuleFilter = (filter: TxFilter['filter']): filter is ModuleTxF
   return 'module' in filter
 }
 
+const NATIVE_DECIMALS = 18
+const parseNativeValue = (value: string) => safeParseUnits(value, NATIVE_DECIMALS)?.toString()
+const formatNativeValue = (value: string) => safeFormatUnits(value, NATIVE_DECIMALS)
```
iamacook, line 50: I see that this was 18 before but the native currency decimals can be set on the Config Service. Do you think it's worth reading the config instead?

apps/web/src/components/transactions/TxFilterForm/index.tsx
iamacook: I think "with zeros" is clearer because "decimals" suggest a float. We also could add a tooltip explaining it's the "base amount".

apps/web/src/utils/tx-history-filter.ts
katspaugh: Yes, that would be more correct, although a quick google/chatgpt indicates they try to stick to 18. I'll pass decimals from chain info anyway.

apps/web/src/components/transactions/TxFilterForm/index.tsx
katspaugh: Done:
Screenshot 2025-02-11 at 15 25 33

liliya-soroka:
Verified

PR #4978, opened 2025-02-13
feat: Create org when submitting the form

### LINKED ISSUE #4977: [Orgs] Integrate org creation API endpoint
## Links

## What must be done

- When submitting the org creation form the endpoint should be called to create an org
- The user should be navigated to the org dashboard after creation

## How to test it

- Org is created when the form is submitted
- User is navigated to the dashboard when the form is submitted

usame-algan:
## What it solves

Resolves #4977

## How this PR fixes it

- Calls the endpoint to create an org when submitting the org creation form
- Navigates user to the org dashboard

## How to test it

1. Create an org either from the dashboard or from the org list page
2. Observe being navigated to the org dashboard

## Checklist

### DIFF
--- a/packages/store/src/gateway/AUTO_GENERATED/users.ts
+++ b/packages/store/src/gateway/AUTO_GENERATED/users.ts
@@ -53,7 +53,7 @@ export type UserWallet = {
 }
 export type UserWithWallets = {
   id: number
-  status: 1
+  status: 0 | 1
   wallets: UserWallet[]
 }
 export type CreatedUserWithWallet = {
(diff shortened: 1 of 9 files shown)


packages/store/src/gateway/AUTO_GENERATED/users.ts
```diff
@@ -53,7 +53,7 @@ export type UserWallet = {
 }
 export type UserWithWallets = {
   id: number
-  status: 0 | 1
+  status: 1
```
jmealy, line 56: Do you know the reason for this change?

jmealy (approved):

packages/store/src/gateway/AUTO_GENERATED/users.ts
usame-algan: The pending status (0) was removed at some point but is now added back so I will revert this change. Thanks!

Issue #4965, opened 2025-02-12
[Orgs] Implement sign in button


usame-algan:
## Links

Epic on Notion: https://www.notion.so/safe-global/Org-onboarding-flow-1208180fe57380078591f5380a0c5d2a

## What must be done

- Adjust the current sign in button to match the design of the continue with wallet button
- Handle disconnected wallet fallback

## Designs/sketches

<img width="255" alt="Image" src="https://github.com/user-attachments/assets/3c7d06cb-f328-4660-9b0a-5e41b55c09be" />

## How to test it

- If the user has no connected wallet they should not sign in but instead connect first
- If they are connected the sign in button should look like the continue with wallet button

francovenica:
@jmealy 

Question. This ticket says we should handle the disconnect (which currently works in the RC), but shouldn't also handle changing the wallet's owner?
Currently if you switch owners in your MM extension, you are still logged in as previous user. You are forced to disconnect and then reconnect with already the user you want selected in MM.

PR #4986, opened 2025-02-14
Feat(Orgs): Implement orgs sign in state

### LINKED ISSUE #4985: [Orgs] Implement sign in state
## Links

## What must be done

When users do SiWE they store an access token cookie for 24 hours. This cookie is not accessible from the app so we need to store an additional state to indicate that the current wallet is authenticated in order to display different views to the user e.g. on the organization list page.

## How to test it

- Sign in on the org list page
- Observe the view changing

usame-algan:
## What it solves

Resolves #4985 

## How this PR fixes it

- Adds a new `auth` slice that stores the session expiry
- Sets the session expiry to now + 24 hours but this will come from the `/auth/verify` endpoint as a return value eventually
- Updates the `OrgsList` component to store and retrieve the auth state in order to display a different view

## How to test it

1. Go to `/welcome/organizations`
2. Press sign in
3. Observe the view changing
4. Reload the app
5. Observe still seeing the updated view

## Checklist


usame-algan:
> When you are not connected and try to sign in, nothing happens. We might need to have a state for this.

It seems we are checking for the existence of `provider` when pressing that button which doesn't exist if the wallet is not connected. I would resolve this as part of https://github.com/safe-global/safe-wallet-monorepo/issues/4965

> We still get the CORS error on the deployed test branch. Presumably it would also not work on staging when we deploy it there.

I couldn't reproduce this issue anymore. The `chains` fetch succeeds even on the PR deployment. Can you try it again?

jmealy (approved):

Issue #4894, opened 2025-02-05
[Mass payouts] Same-chain mass payouts


tmjssz:
We want to extend the create transaction flow to allow users to send tokens to multiple recipients within the same transaction.

The feature originates from the [Safenet MVP scope](https://www.notion.so/Safenet-MVP-Epics-and-Features-1378180fe57380bcbfbfd5a884fa0832?pvs=21) but is designed to function independently and will not affect the Safenet scope.

## Feature details

* An "Add recipient" button is displayed below the send token form.
* Up to 5 recipients with a token and a respective amount can be defined to be included in the same transaction.
* Redirect to CSV Airdrop app if more than 5 recipients are needed.
* Total amounts are not allowed to exceed the available balance or spending limit.

### Additional requirements

* The feature should be **partially released** for a selected group of users. - it was decided to release the feature for all users
* It should be available regardless of the access point - “New transaction”, address book entry or asset in the list (in 2 latter cases respective fields will be pre-populated)

## Links

* [Designs](https://www.figma.com/design/RtoyskMC7r2Dhs1BzX0TN5/TEMP%3A-Same-chain-mass-payouts?node-id=0-1&p=f&t=CcKEIAkG5CHWHa4w-0)
* [Notion page](https://www.notion.so/safe-global/Same-chain-mass-payouts-18c8180fe5738050a85bff68e8d22934?pvs=4)
* Safenet feature PR: #4788

PR #4992, opened 2025-02-14
feat: targeted features

iamacook:
## What it solves

Partially resolves https://github.com/safe-global/safe-client-gateway/issues/2323

## How this PR fixes it

As we intend to launch [Nested Safes](https://github.com/safe-global/safe-wallet-monorepo/issues/4879) and [mass payouts](https://github.com/safe-global/safe-wallet-monorepo/issues/4894) as targeted features, we need a way to distinguish Safe-specific features.

This introduces two new hooks: one for determining whether a Safe is targeted (via [targeted messaging](https://github.com/safe-global/safe-client-gateway/pull/2343)) - `useIsOutreachSafe` - and the other whether a feature flag is enabled for a targeted Safe - `useIsTargetedFeature`.

Note: new store types were generated in order to use the targeted Safe endpoint of the Client Gateway.

### `useIsOutreachSafe`

This queries the Client Gateway to determine whether the current Safe is targeted by an `outreachId`. It returns `true` if the given `outreachId` and Safe address are targeted, otherwise `false`.

### `useIsTargetedFeature`

Leveraging the above, this combines feature flags of the Config Service. If the current Safe is targeted AND the feature flag is enabled, it returns `true`, otherwise `false`.

If `true` was ever previously returned, the feature is "unlocked" regardless of the current Safe for consistency in the UI (particularly important when navigating to Nested Safes). This means that the feature remains enabled until disabled in the Config Service.

## How to test it

Until integrated with either Nested Safes or mass payouts, this is covered by unit tests.

## Checklist


apps/mobile/src/tests/mocks.ts
```diff
@@ -159,7 +159,6 @@ export const mockSwapTransfer: SwapTransferTransactionInfo = {
   sellToken: fakeToken2,
   buyToken: fakeToken,
   explorerUrl: 'http://google.com',
-  executedSurplusFee: '',
```
iamacook, line 162: This is no longer in the types.

tmjssz (approved):
Nice work!
Thanks for already adding the MASS_PAYOUTS targeted feature as well!

apps/web/src/utils/chains.ts
```diff
@@ -41,6 +41,8 @@ export enum FEATURES {
   BRIDGE = 'BRIDGE',
   RENEW_NOTIFICATIONS_TOKEN = 'RENEW_NOTIFICATIONS_TOKEN',
   TX_NOTES = 'TX_NOTES',
+  NESTED_SAFES = 'NESTED_SAFES',
+  MASS_PAYOUTS = 'MASS_PAYOUTS',
```
katspaugh, line 45: Should we prefix this with `TARGETED_` for clarity? Otherwise someone might toggle them in the config and expect the features to be enabled.

apps/web/src/features/targetedOutreach/constants.ts
```diff
@@ -1,5 +1,12 @@
+import { FEATURES } from '@/utils/chains'
+
 export const ACTIVE_OUTREACH = { id: 2, url: 'https://wn2n6ocviur.typeform.com/to/nlQlP7lU', targetAll: true }
 
+export const TARGETED_FEATURES = [
+  { id: 3, feature: FEATURES.NESTED_SAFES },
+  { id: 4, feature: FEATURES.MASS_PAYOUTS },
+] as const
```
katspaugh, line 8: It's a little bit confusing that the outreach popup feature is mixed with a more generic "targeted features" feature.

The idea of [feature folders](https://www.notion.so/safe-global/Feature-toggles-970c96d18e544fd1a01a362f0535146b) was to isolate features so that they can be added, removed and toggled independently from everything else. With this change, the outreach popup will be coupled with targeting.

P.S. btw "a targeted Safe" is an ominous combination of words.

apps/web/src/utils/chains.ts
iamacook: Renamed in 927d53384247b970800270b9d19992306ea79470.

apps/web/src/features/targetedOutreach/constants.ts
iamacook: Everything has been renamed/relocated in 927d53384247b970800270b9d19992306ea79470.

- `useIsTargetedSafe` -> `useIsOutreachSafe`
- Feature-targeting logic -> `@/features/targetedFeatures`

katspaugh: Thanks! However, you're still importing and using a hook from the popup feature which creates a strange dependency. That they use the same endpoint doesn't mean we should be using a hook to check whether to render a survey popup to toggle other features.

katspaugh (approved):
LGTM

liliya-soroka:
It will be available for full testing together with new features

Issue #3876, opened 2024-06-26
UI fixes for transaction flows


TanyaEfremova:
## Bug description
1. Align progress bars styles in _Transaction flow_ and with the one in _Create/Add Safe Account_. The progress bar (MuiLinearProgress) should use `linear-gradient(#5FDDFF,#12FF80)` in both light and dark modes.
<img width="720" alt="Screenshot 2024-06-26 at 18 21 53" src="https://github.com/safe-global/safe-wallet-web/assets/93707525/642e23cb-33f0-4722-9fe8-c79be7fecb1f">
<img width="800" alt="Screenshot 2024-06-26 at 18 24 04" src="https://github.com/safe-global/safe-wallet-web/assets/93707525/d9b3366d-f480-48d3-aa91-45fe55ec0e8e">

[For reference see Figma. ](https://www.figma.com/design/HKW5iarMST9rL7Kxkf59qY/MUI-for-Safe?node-id=12894-65214&t=B8MKulq81S3hLvtw-1)

2. Icons in the transaction status should use` success-main` depending on their status (e.g. when a transaction is being confirmed, the checkmark icon should turn green instead of grey).
<img width="410" alt="Screenshot 2024-06-26 at 18 31 10" src="https://github.com/safe-global/safe-wallet-web/assets/93707525/715778d5-03ab-46cd-9349-1420bff9980e">

## Environment
 - Browser: Chrome
 - Wallet: MetaMask
 - Chain: Ethereum mainnet

## Steps to reproduce
 1. Go to a new transaction screen and see the status bar above the transaction card header and the transaction status on the right. 

## Expected result
Described above.

## Screenshots
See above.

SebastianLF:
Hello, can i work on this ?

liliya-soroka:
existing 18.11.24

SamarthJ2004:
Hi @TanyaEfremova , I have opened a pull request solving the same. Kindly review.

PR #4996, opened 2025-02-16
Fix(Tx flow): progress bar color

SamarthJ2004:
## What it solves

Resolves the issue: #3876. The progress bar and the checklist icons color in the light mode are fixed.

## How this PR fixes it

I have added the conditional rendering of the variable color as per the theme. Earlier in both the themes the color name was but , the color values defined in the var.css were different. I added conditions in css so that the required color is used.

## How to test it
Go to a new transaction screen and see the status bar above the transaction card header and the transaction status on the right.

## Screenshots
Screenshot 2025-02-17 at 12 14 12 AM

now both the things are renders with the required color.
## Checklist

### DIFF
--- a/apps/web/src/components/common/ProgressBar/styles.module.css
+++ b/apps/web/src/components/common/ProgressBar/styles.module.css
@@ -9,6 +9,10 @@
   border-radius: 6px;
 }
 
+[data-theme="light"] .progressBar :global .MuiLinearProgress-bar {
+  background: var(--color-secondary-main);
+}
+
 @media (max-width: 599.95px) {
   .progressBar {
     border-radius: 0;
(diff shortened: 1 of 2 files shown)


apps/web/src/components/common/ProgressBar/styles.module.css
```diff
@@ -9,6 +9,10 @@
   border-radius: 6px;
 }
 
+[data-theme="light"] .progressBar :global .MuiLinearProgress-bar {
+  background: var(--color-secondary-main);
+}
```
katspaugh, line 14: Why only for the light theme?

katspaugh (approved):

katspaugh:
Thanks again! ❤️

PR #5043, opened 2025-02-19
Feat(Safenet): Add Safenet balance overview

germartinez:
## What it solves

Resolves https://github.com/safe-global/safenet/issues/359

## How to test it
Open a Safenet-enabled Safe account.
Go to the dashboard and check the cross-chain balance overview.

## Screenshots
Screenshot 2025-02-19 at 18 34 48

## Checklist


apps/web/src/features/safenet/components/SafenetBalanceOverview/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    },
+    {} as { [chainId: string]: number },
+  )
+}
+
+const SafenetBalanceOverview = () => {
+  const { balances } = useBalances()
+  const isDarkMode = useDarkMode()
+
+  const totalSafenetValueFiat = useMemo(() => {
```
schmanu, line 27: I do not think we can actually sum up fiat balances here.
If we do want this, we need to consider the correct fiat conversion rates.

I would suggest to only go for USDC balances for now until we support multiple tokens.

PR #5060, opened 2025-02-20
feat(mobile): display alert before deleting safe

compojoom:
## What it solves
- display an alert before deleting a safe
- makes it possible to delete all safes in the app

## Screenshots
grafik

## Checklist


apps/mobile/src/features/AccountsSheet/AccountItem/AccountItem.tsx
```diff
@@ -52,7 +52,23 @@ export function AccountItem({ account, drag, chains, isDragging, activeAccount,
         <AccountCard
           leftNode={
             isEdit && (
-              <TouchableOpacity onPress={onSafeDeleted(account.address.value as Address)}>
+              <TouchableOpacity
+                onPress={() => {
+                  Alert.alert('Delete Safe', 'Are you sure you want to delete this safe?', [
```
clovisdasilvaneto, line 57: Can we move this function to be outside the render layer?

apps/mobile/src/features/AccountsSheet/MyAccounts/hooks/useMyAccountsSortable.ts
```diff
@@ -13,24 +12,13 @@ export const useMyAccountsSortable = (): useMyAccountsSortableReturn => {
   const dispatch = useDispatch()
   const safes = useSelector(selectAllSafes)
   const [sortableSafes, setSortableSafes] = useState(() => Object.values(safes))
-  const isEdit = useSelector(selectMyAccountsMode)
-
-  useEffect(() => {
-    const newSafes = Object.values(safes)
```
clovisdasilvaneto, line 19: Did you check the drag and drop of the sortable when removing this effect? I remember we were having a issue after sorting a wallet... That is why I created this useEffect 🤔

clovisdasilvaneto (approved):

apps/mobile/src/features/AccountsSheet/MyAccounts/hooks/useMyAccountsSortable.ts
compojoom: Yes, I did. As you can see few lines later I moved the setSortableSafes to the onDragEnd callback. 

I removed the shouldGoToListMode as now it is possible to remove a safe even if it is the only one in the app.

apps/mobile/src/features/AccountsSheet/AccountItem/AccountItem.tsx
compojoom: good point. I guess I was in a hurry yesterday evening :)

PR #5062, opened 2025-02-21
feat(web): add keystone wallet back

compojoom:
## What it solves
We had to remove keystone from the list of wallets in this PR: https://github.com/safe-global/safe-wallet-monorepo/pull/4686/files as the previous package had a faulty dependency that didn’t resolve with yarn v4. This has been resolved here:
https://github.com/blocknative/web3-onboard/pull/2322

And I’m adding keystone back

## Screenshots
grafik

## Checklist

### DIFF
--- a/apps/web/jest.setup.js
+++ b/apps/web/jest.setup.js
@@ -5,6 +5,7 @@ import { server } from '@/tests/server'
 
 jest.mock('@web3-onboard/coinbase', () => jest.fn())
 jest.mock('@web3-onboard/injected-wallets', () => ({ ProviderLabel: { MetaMask: 'MetaMask' } }))
+jest.mock('@web3-onboard/keystone/dist/index', () => jest.fn())
 jest.mock('@web3-onboard/ledger/dist/index', () => jest.fn())
 jest.mock('@web3-onboard/trezor', () => jest.fn())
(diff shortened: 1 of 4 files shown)


clovisdasilvaneto (approved):

liliya-soroka:
image

katspaugh:
The usage of this wallet was very low but it comes with many (partially outdated) dependencies. I suggest we park this PR for now and discuss this with the product.

BenAffleck:
Removing one of the options that doesn't rely on blind signing in times of crisis and large hacks isn't the best move, IMO. The partnership between Safe and Keystone was strongly marketed on SoMe. I encourage everyone to get this fixed so the space isn't left with even fewer secure options.

katspaugh:
@compojoom thanks for taking the time to fix this integration (including the PR to web3-onboard).

The product team ultimately decided not to include this for the following reasons:
* The module comes with lots of outdated dependencies, some of which have critical vulnerabilities flagged by dependabot.
* The usage of this wallet as per our analytics is 28 users in the last 90 days. This is 100 times lower than e.g. Ledger, and 22 times lower than Trezor.
* Keystone users can still connect to Safe via MetaMask or WalletConnect.

PR #5081, opened 2025-02-22
fix(web): appUrl query was getting lost while switching network in safeApp 

### LINKED ISSUE #4980: `appUrl` removed when switching chains
## Bug description

Providing the current Safe is multi-chain and a Safe App is open, when switching network, we lose the `appUrl` search parameter. This results in a black screen.

## Steps to reproduce

1.  Open a Safe App in a multi-chain Safe.
2. Switch network.
3. Observe black screen.

## Expected result

Providing the Safe App is present on the destination chain, the Safe App should remain open, otherwise the dashboard navigated to, for example.

## Obtained result

The screen is black after switching.

## Screenshots

https://github.com/user-attachments/assets/ae0979d6-280b-4c41-85a5-98bbfe372510

chiranjeev13:
## What it solves

Resolves #4980

## How this PR fixes it
In File _apps/web/src/components/common/NetworkSelector/index.tsx_
```
export const getNetworkLink = (router: NextRouter, safeAddress: string, networkShortName: string) => {
  const isSafeOpened = safeAddress !== ''

  const query = (
    isSafeOpened
      ? {
          safe: `${networkShortName}:${safeAddress}`,
        }
      : { chain: networkShortName }
  ) as {
    safe?: string
    chain?: string
    safeViewRedirectURL?: string
    appUrl?: string // Added this to the query
  }

  const route = {
    pathname: router.pathname,
    query,
  }

  if (router.query?.safeViewRedirectURL) {
    route.query.safeViewRedirectURL = router.query?.safeViewRedirectURL.toString()
  }

  if (router.query?.appUrl) {
    route.query.appUrl = router.query.appUrl.toString()
  } // if present then appUrl should be populated

  return route
}
```
## How to test it

yarn workspace @safe-global/web start

create a safe multi chain account
go inside a safeApp
switch network in there 

## Screenshots

https://github.com/user-attachments/assets/9ad91240-32f9-439e-b380-1770616202aa

## Checklist

### DIFF
--- a/apps/web/src/components/common/NetworkSelector/index.tsx
+++ b/apps/web/src/components/common/NetworkSelector/index.tsx
@@ -81,6 +81,7 @@ export const getNetworkLink = (router: NextRouter, safeAddress: string, networkS
     safe?: string
     chain?: string
     safeViewRedirectURL?: string
+    appUrl?: string
   }
 
   const route = {
@@ -90,6 +91,10 @@ export const getNetworkLink = (router: NextRouter, safeAddress: string, networkS
 
   if (router.query?.safeViewRedirectURL) {
     route.query.safeViewRedirectURL = router.query?.safeViewRedirectURL.toString()
+  }
+
+  if (router.query?.appUrl) {
+    route.query.appUrl = router.query.appUrl.toString()
   }
 
   return route
--- a/apps/web/src/pages/apps/open.tsx
+++ b/apps/web/src/pages/apps/open.tsx
@@ -1,6 +1,6 @@
 import type { NextPage } from 'next'
 import { useRouter } from 'next/router'
-import { useCallback } from 'react'
+import { useCallback, useEffect } from 'react'
 import { Box, CircularProgress, Typography } from '@mui/material'
 
 import { useSafeAppUrl } from '@/hooks/safe-apps/useSafeAppUrl'
@@ -15,7 +15,7 @@ import { useBrowserPermissions } from '@/hooks/safe-apps/permissions'
 import useChainId from '@/hooks/useChainId'
 import { AppRoutes } from '@/config/routes'
 import { getOrigin } from '@/components/safe-apps/utils'
-import { useHasFeature } from '@/hooks/useChains'
+import { useCurrentChain, useHasFeature } from '@/hooks/useChains'
 import { FEATURES } from '@/utils/chains'
 
 const SafeApps: NextPage = () => {
@@ -26,6 +26,7 @@ const SafeApps: NextPage = () => {
   const safeAppData = allSafeApps.find((app) => app.url === appUrl)
   const { safeApp, isLoading } = useSafeAppFromManifest(appUrl || '', chainId, safeAppData)
   const isSafeAppsEnabled = useHasFeature(FEATURES.SAFE_APPS)
+  const currentChain = useCurrentChain()
 
   const { addPermissions, getPermissions, getAllowedFeaturesList } = useBrowserPermissions()
   const origin = getOrigin(appUrl)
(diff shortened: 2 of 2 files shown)


chiranjeev13:
> hereby

I agree

chiranjeev13:

apps/web/src/components/common/NetworkSelector/index.tsx
```diff
@@ -92,6 +93,10 @@ export const getNetworkLink = (router: NextRouter, safeAddress: string, networkS
     route.query.safeViewRedirectURL = router.query?.safeViewRedirectURL.toString()
   }
 
+  if (router.query?.appUrl) {
+    route.query.appUrl = router.query.appUrl.toString()
```
katspaugh, line 97: This preserves the query indeed but ignores the fact that a Safe App might be not enabled on all chains.

From the original issue requirements:
> Providing the Safe App is present on the destination chain, the Safe App should remain open, otherwise the dashboard navigated to, for example.

chiranjeev13: Yes, I was thinking the same. However, in the case of other EOAs, when we switch to a different network within an app, it's the app that indicates it as an unsupported network.

Since we're displaying the app within Safe's dashboard, I guess it's a better user experience for Safe to indicate an unsupported network. Working on this fix !!

chiranjeev13: Updated the unsupported chain handling logic in `apps/web/src/pages/apps/open.tsx`:

In this commit 82118790aa5e34e96b69357ce8c60bd0525a8d0d

### Key Modifications:
- Refined the condition to check when `safeApp.chainIds` is empty
- Maintained 3-second redirect timeout
- ensured the `remoteSafeAppsLoading` is false before we check for chainIds

## Video Preview


https://github.com/user-attachments/assets/5bfcd7bd-be11-48df-9b2a-03c34137cf96


Let me know if any thing else needs to be updated @katspaugh

katspaugh: Good idea!

apps/web/src/pages/apps/open.tsx
```diff
@@ -88,6 +88,32 @@ const SafeApps: NextPage = () => {
...
+        display="flex"
+        flexDirection="column"
+        alignItems="center"
+        justifyContent="center"
+        height="100%"
+        textAlign="center"
+        p={2}
+      >
+        <Typography variant="body1" gutterBottom>
+          Chain {chainId} is not supported in this app. <br />
```
katspaugh, line 109: Can we display chainName here? It can be taken from `useCurrentChain()`.

chiranjeev13: Sure let me just do that instead!

chiranjeev13: Updated fix:

- Using `useEffect` for routing back to safeApp Dashboard
- Displaying chain name instead of chainId using `useCurrentChain()`

check it out here: c288de4b5122a011b47979375b2ea7e9fd6dc5fc

katspaugh (approved):
Thanks so much!
Sending to our QA.

chiranjeev13:
Not sure why the cypress tests failed @katspaugh

katspaugh:
@chiranjeev13 that's normal, they can't run on external forks.

francovenica:
The fix itself looks good to me, the URL is maintained when switching networks using the topright selector.

NOTE: The connect button doesn't work tho, because it wasn't working on dev at the time of creating this branch. It was already fixed in dev so do we know if this PR will break dev when merged?

francovenica:
QA has nothing more to do on this ticket, so I'll move it to done. It can be merged once the rebase is done

katspaugh:
Thank you @chiranjeev13!

PR #5104, opened 2025-02-24
Feat(Orgs): sign out on disconnect wallet

jmealy:
## What it solves
Currently there is no way to go from the signed in state to the signed out state. We should allow the user to be able to sign out.

## How this PR fixes it
- Call the sign out endpoint when disconnecting the connected wallet.
- Set the redux state to logged out.

## How to test it
- go to the orgs list page in the signed in state. you should see the list of orgs.
- Disconnect your wallet.
- You should see the signed out state of the orgs list

## Checklist


usame-algan (approved):
Might just be a "visual" issue but for me the `access_token` cookie is not deleted when logging out. Even after reloading the app. Its not affecting this PR but we should figure out if this needs fixing on the CGW or not.

PR #5109, opened 2025-02-25
Fix(Orgs): use correct format for add accounts request

jmealy:
## What it solves
The payload type for the add accounts request from the CGW is wrong.

## How this PR fixes it
- Use the correct type and add a todo to update the type from the CGW when it is fixed.
- also adds back in the empty state for the accounts page when the org has no accounts.

## How to test it
- Go to the accounts page of an org. 
- If there are no accounts added yet you should see the empty accounts component.
- Click the add accounts button, select a safe, and submit. See that the form submits succesfully, and the safe now appears in the accounts page.

## Checklist


usame-algan (approved):
Thanks!

PR #5114, opened 2025-02-25
Fix(Blockaid): pass dapp origin

katspaugh:
## What it solves

We were passing the wrong `metadata.domain` to Blockaid API:
* For txs made via Safe Wallet itself, it should be `non_dapp`
* For Safe Apps and apps connected via WalletConnect – their URL


usame-algan (approved):
Looks good! I tested all three cases:

- non_dapp in case its an internal tx
- safe app url in case its a safe app
- wallet connect url in case its an app through wallet connect

PR #5118, opened 2025-02-25
fix(web): Updated re-verifying of address in new-safe/load when chain is switched fixes (#5067)

### LINKED ISSUE #5067: Address input error is not cleared when input changes to the correct value
## Bug description
When adding a read only safe, if you paste a safe address with the wrong chain selected (where the safe doesn't exist), you get an error. If you then select the correct chain, the error should be cleared, but it isn't.

## Environment

- Browser: Chrome
- Wallet: MetaMask
- Chain: Ethereum mainnet

## Steps to reproduce

1.  Go to `/new-safe/load`
2. In the address input, paste an address that does not exist on the currently selected chain. You will see an input error
3. Then switch the chain input to the correct chain. The input error will remain and the submit button is still disabled. It will only update if you click somewhere on the screen.

## Expected result
The error should be cleared when the chain is changed.

## Obtained result
The error is not cleared.

chiranjeev13:
## What it solves
Resolves #5067

## How this PR fixes it
`In apps/web/src/components/common/AddressInput/index.tsx
`

The code change adds a dependency on currentChain?.chainId in the useEffect hook for the AddressInput component. This ensures that the input validation is re-triggered when the chain changes, preventing potential stale validation states across different blockchain networks.

```
  useEffect(() => {
    trigger(address)
  }, [currentChain?.chainId, trigger, address])
```

## How to test it
- Switch between different blockchain networks in the application
- Verify that address validation works correctly after network changes

## Screen Recordings

https://github.com/user-attachments/assets/3543c8c9-533f-493b-9e5d-13675f482373

## Checklist

### DIFF
--- a/apps/web/src/components/common/AddressInput/index.tsx
+++ b/apps/web/src/components/common/AddressInput/index.tsx
@@ -64,6 +64,11 @@ const AddressInput = ({
   const isDomainLookupEnabled = !!currentChain && hasFeature(currentChain, FEATURES.DOMAIN_LOOKUP)
   const { address, resolverError, resolving } = useNameResolver(isDomainLookupEnabled ? watchedValue : '')
 
+  // Re-validate when chainId changes
+  useEffect(() => {
+    trigger(address)
+  }, [currentChain?.chainId, trigger, address])
+
   // errors[name] doesn't work with nested field names like 'safe.address', need to use the lodash get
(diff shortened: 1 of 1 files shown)


chiranjeev13:
Hey @katspaugh can you review this PR?

katspaugh (approved):
Sorry, missed it. Code looks good, thanks a lot!

PR #5119, opened 2025-02-25
added margin for continue button in ImportAccount and AddSigners fixes #5069

### LINKED ISSUE #5069: IOS: Some buttons are to low at the bottom and hidden under keyboard
## Bug description
Some buttons are to low at the bottom and hidden under keyboard
## Environment
iphone 11

## Steps to reproduce
1.  start the Import safe flow
Current result: Continue button is partly hidden as it's located too low 

2. click Paste address
Current result: The keyboard appear,s and the button is hidden under the keyboard

3. go to the signer list
Current result: the button is too low

## Screenshots

Image

Image

https://github.com/user-attachments/assets/f7cf4a86-52bf-49a7-b882-16fdf59952e2

chiranjeev13:
## What it solves

Resolves #5069

## Environment

Iphone 16 (Xcode Simulator)

## How this PR fixes it

kept bottom margin to 70px in `<SafeButton />` component which adjusts to the keyboard properly in both
`apps/mobile/src/features/ImportReadOnly/components/ImportAccountFormView.tsx`
`apps/mobile/src/features/ImportReadOnly/components/AddSignersFormView.tsx`

## How to test it
Add an account in the ios app
## Screen Recordings

https://github.com/user-attachments/assets/1f23d274-af1a-456d-b4d4-37157cafdaa8

## Checklist

### DIFF
--- a/apps/mobile/src/features/ImportReadOnly/components/ImportAccountFormView.tsx
+++ b/apps/mobile/src/features/ImportReadOnly/components/ImportAccountFormView.tsx
@@ -91,6 +91,7 @@ export const ImportAccountFormView: React.FC<ImportAccountFormViewProps> = ({
         onPress={onContinue}
         disabled={!isEnteredAddressValid || !safeExists}
         testID={'continue-button'}
+        marginBottom={70}
       >
         Continue
       </SafeButton>
--- a/apps/mobile/src/features/ImportReadOnly/components/AddSignersFormView.tsx
+++ b/apps/mobile/src/features/ImportReadOnly/components/AddSignersFormView.tsx
@@ -27,7 +27,7 @@ export const AddSignersFormView = ({
       />
 
       <View paddingHorizontal={16}>
-        <SafeButton onPress={onPress} testID={'continue-button'}>
+        <SafeButton onPress={onPress} testID={'continue-button'} marginBottom={70}>
           Continue
         </SafeButton>
       </View>


chiranjeev13:
Hey @katspaugh so sorry for pinging again.
Can you review this PR aswell?

chiranjeev13:
Hey @clovisdasilvaneto Can you review this PR?
Thanks!

apps/mobile/src/features/ImportReadOnly/components/ImportAccountFormView.tsx
```diff
@@ -91,6 +91,7 @@ export const ImportAccountFormView: React.FC<ImportAccountFormViewProps> = ({
         onPress={onContinue}
         disabled={!isEnteredAddressValid || !safeExists}
         testID={'continue-button'}
+        marginBottom={70}
```
clovisdasilvaneto, line 94: Thanks for the contribution @chiranjeev13 . However I think a more concise approach would be better here, we need to understand what is causing the button to go way down the screen instead of just adding a big margin bottom to it. Might be due to the usage of this hook `useSafeAreaPaddingBottom` in our component.

clovisdasilvaneto (changes requested):
Thanks for the contribution @chiranjeev13 , however increasing the margin on the Button might not be a good approach. We need to figure out why the button is going to far in the first place and fix the root cause of the issue.

compojoom:
@chiranjeev13 thank you for looking into this. Unfortunately the issue is more complicated. It's a bug in the react-native-screens library https://github.com/software-mansion/react-native-screens/issues/2587
We use the same screens in multiple locations and just going for a single marginBottom won't be enough.

chiranjeev13:
Got it thanks for the feedback! @compojoom @clovisdasilvaneto

compojoom:
We've made some modifications to this in a different PR and in most cases it should be correct now. Closing this PR. Thanks for looking into this!

PR #5133, opened 2025-02-26
chore(ci): generate artifact attestation

compojoom:
## What it solves
This will generate an attestation over the archive file. One can then verify that this archive was generated on github by using the `gh attestation verify PATH/TO/YOUR/BUILD/ARTIFACT-BINARY -R ORGANIZATION_NAME/REPOSITORY_NAME` command. 

More info on attestation here:

https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds

### DIFF
--- a/.github/workflows/web-deploy-production.yml
+++ b/.github/workflows/web-deploy-production.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   release:
     permissions:
+      attestations: write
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
👍

PR #5158, opened 2025-02-27
Fix(Analytics): Remove GTM

### LINKED ISSUE #5159: [Analytics] Remove Google Tagmanager
## Overview

Tagmanager was introduced to handle more integrations other than GA. This has not happened. All events are sent to Tagmanager where they are mapped onto different GA properties. This adds additional complexity and another product that needs to be maintained.

## What must be done

- Remove `TagManager` initialization
- Move the mapping logic into the web frontend

## How to test it

- All events should still be tracked with the correct names and attributes
- The three different properties should still be used as before

usame-algan:
## What it solves

Resolves #5159

## How this PR fixes it

- Removes `TagManager` class and moves required logic into the gtm.ts file
- Adds `@next/third-parties` package
- Adds new `Analytics` component to initialize GA
- Replaces `dataLayer` calls with `sendGAEvent`

## ToDos

- [ ] Add env variables to CI

## How to test it

1. Ensure that the same events are still sent to GA
- Test this PR deployment vs. dev/staging/prod
- Do the same action in both deployments e.g. Create new transaction
- Compare `collect` network call data

2. Ensure events are still sent to the correct GA property
- Test property on localhost, dev, staging, PRs
- Prod property on prod
- Safe apps property when inside a safe app

## Checklist


apps/web/src/services/analytics/Analytics.tsx
```diff
@@ -0,0 +1,23 @@
...
+      ad_storage: 'denied',
+      analytics_storage: 'denied',
+      functionality_storage: 'granted',
+      personalization_storage: 'denied',
+      security_storage: 'granted',
+      wait_for_update: 500,
+    })
+  }, [])
+
+  return <GoogleAnalytics gaId={GA_TRACKING_ID} debugMode={true} />
```
katspaugh, line 20: Should `debugMode` be false in prod?

apps/web/src/services/analytics/Analytics.tsx
```diff
@@ -0,0 +1,23 @@
+import { GA_TRACKING_ID, SAFE_APPS_GA_TRACKING_ID } from '@/config/constants'
+import { GoogleAnalytics } from '@next/third-parties/google'
+import { useEffect } from 'react'
+
+const Analytics = () => {
+  useEffect(() => {
+    // This needs to be added once in order for events with send_to: SAFE_APPS_GA_TRACKING_ID to work
+    window.gtag?.('config', SAFE_APPS_GA_TRACKING_ID, { debug_mode: true })
+
+    window.gtag?.('consent', 'default', {
```
katspaugh, line 10: How does this work wrt cookies? The current code upgrades the cookie consent once it's given via the banner, and also cookies were deleted once the consent is withdrawn.

katspaugh:
Please rebase it to the `release` branch.

usame-algan:
Closing this in favour of #5164

apps/web/src/services/analytics/Analytics.tsx
usame-algan: This is the default consent config we set whenever we called `gtmInit` which we did regardless of cookies being set or not so I think the behaviour should still be the same.

PR #5160, opened 2025-02-27
chore(ci): add SBOM next to archive

compojoom:
A Software Bill of Materials (SBOM) is a formal record of all components -  libraries and all dependencies - that are used within a software product. It’s essentially an “ingredient list” that helps track and manage vulnerabilities, licensing issues, and compliance requirements throughout the software supply chain.

The SBOM is generated out of the package.json in apps/web. After it is generated we create an attestation that embeds and links it to the generated archive. 

For easier usage with other tools the generated SBOM is attached to the release and can then be imported in tools such [dependencytrack](https://dependencytrack.org/).

### DIFF
--- a/.github/workflows/web-deploy-production.yml
+++ b/.github/workflows/web-deploy-production.yml
@@ -17,6 +17,7 @@ jobs:
 
     env:
       ARCHIVE_NAME: ${{ github.event.repository.name }}-${{ github.event.release.tag_name }}
+      SBOM_FILE_NAME: sbom
 
     steps:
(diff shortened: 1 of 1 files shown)


katspaugh:
I'm reluctant to add yet another dependency, especially one that has only 21 stars on GitHub and has to be run during the build in the production release workflow. The resulting file seems to be no different than the `yarn.lock`. Perhaps there are tools that monitor yarn.lock instead?

katspaugh:
Closing as per the comment above.

PR #5164, opened 2025-02-28
Fix(Analytics): Remove GTM

### LINKED ISSUE #5159: [Analytics] Remove Google Tagmanager

usame-algan:
## What it solves

Resolves #5159

## How this PR fixes it

- Removes `TagManager` class and moves required logic into the gtm.ts file
- Adds `@next/third-parties` package
- Adds new `Analytics` component to initialize GA
- Replaces `dataLayer` calls with `sendGAEvent`

## ToDos

- [ ] Add env variables to CI

## How to test it

1. Ensure that the same events are still sent to GA
- Test this PR deployment vs. dev/staging/prod
- Do the same action in both deployments e.g. Create new transaction
- Compare `collect` network call data

2. Ensure events are still sent to the correct GA property
- Test property on localhost, dev, staging, PRs
- Prod property on prod
- Safe apps property when inside a safe app

## Checklist


apps/web/src/services/analytics/Analytics.tsx
```diff
@@ -0,0 +1,23 @@
+import { GA_TRACKING_ID, IS_PRODUCTION, SAFE_APPS_GA_TRACKING_ID } from '@/config/constants'
+import { GoogleAnalytics } from '@next/third-parties/google'
+import { useEffect } from 'react'
+
+const Analytics = () => {
+  useEffect(() => {
+    // This needs to be added once in order for events with send_to: SAFE_APPS_GA_TRACKING_ID to work
+    window.gtag?.('config', SAFE_APPS_GA_TRACKING_ID, { debug_mode: !IS_PRODUCTION })
+
+    window.gtag?.('consent', 'default', {
```
katspaugh, line 10: We initialize it with these default settings and never change them. The previous implementation was upgrading them when a cookie consent is given (and we were also deleting cookies if the consent is withdrawn).

I think it's OK for the time being but it would be good to have a full feature parity with the previous implementation.

usame-algan: This is still being done inside the `useGtm` hook. All I did here is move the needed logic out of the `TagManager` to keep full parity

usame-algan: We should probably rename Gtm to Ga to make it more clear

PR #5165, opened 2025-02-28
chore(ci): New workflow to publish the web packages to IPFS

gjeanmart:
## What it solves

Resolves #

## Checklist

### DIFF
--- a/.github/workflows/web-deploy-ipfs.yml
+++ b/.github/workflows/web-deploy-ipfs.yml
@@ -3,7 +3,7 @@ name: Web Deploy to IPFS
 on:
   pull_request:
     branches:
-      - ipfs-deploy
+      - dev
     paths:
       - apps/web/**
       - packages/**
@@ -16,9 +16,12 @@ concurrency:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+
     permissions:
-      pull-requests: write
-      id-token: write
+      contents: read       
+      statuses: write      
+      checks: write        
+      pull-requests: write 
 
     name: Deploy to IPFS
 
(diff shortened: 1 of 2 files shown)


katspaugh:
Seems to work!

Not sure if intended but it still uploads .html files in subfolders.

E.g.: https://ipfs.io/ipfs/QmRvQGWAVwptkpvwUPJMamaCKjw5RCwYDUGDCNAMToxjDo/apps/

apps/web/next.config.mjs
```diff
@@ -26,6 +26,7 @@ const withPWA = withPWAInit({
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   output: 'export', // static site export
+  trailingSlash: true,
```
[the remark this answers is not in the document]
gjeanmart: let me do one test again without.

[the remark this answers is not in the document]
gjeanmart: it works without it

2color:
Great to see you embracing CIDs and deploying the frontend to IPFS. 

The main problem with this approach is that you rely on the pinning service to handle the merkleisation of the build. This places implicit trust in the pinning service. If the pinning service is hacked they can modify the build and return a CID of a malicious version. 

Our general recommendation is to merkleise the build output in GitHub actions, making the CID generation a local operation, and then uploading the CAR to the pinning service (note that Pinata doesn't support CAR uploads, yet). We recently released a new GitHub action that encapsulates these best practices: https://github.com/ipfs/ipfs-deploy-action.

Happy to answer any questions you might have.

.github/workflows/web-deploy-ipfs.yml
```diff
@@ -0,0 +1,61 @@
...
+    name: Deploy to IPFS
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/yarn
+
+      - uses: ./.github/actions/build
+        with:
+          secrets: ${{ toJSON(secrets) }}
```
katspaugh, line 32:
```suggestion
          secrets: ${{ toJSON(secrets) }}
          prod: ${{ true }}
```

katspaugh:
We should probably add IPFS gateways to the "official" hosts list: https://github.com/safe-global/safe-wallet-monorepo/blob/dev/apps/web/src/config/constants.ts#L109

Otherwise it will be branded as a "Wallet fork" with no logo in the header.

.github/workflows/web-deploy-ipfs.yml
```diff
@@ -0,0 +1,56 @@
...
+          prod: ${{ true }}
+
+      - name: Remove .html extension from all files
```
2color, line 38: Do you remove the `.html` extension so that next page routes are loaded correctly via IPFS Gateways? 

If so, you can fix this more elegantly by setting [`trailingSlash` to `true` in your `next.config.js` file](https://nextjs.org/docs/app/api-reference/config/next-config-js/trailingSlash). 

The reason is that IPFS Gateways expect either a filename with an extension, and if none is provided, they will look for the `index.html`. 

Setting [`trailingSlash` to `true` in your `next.config.js` file](https://nextjs.org/docs/app/api-reference/config/next-config-js/trailingSlash) will create a folder for each next route with an `index.html` making it play nicely with gateways.

gjeanmart: As far as I know, this is something we cannot use which would break the traditional hosting app. Maybe @katspaugh you can explain a bit more (this was already discussed in this [comment](https://github.com/safe-global/safe-wallet-monorepo/pull/5165#discussion_r1975585903))

gjeanmart:
Close in favor of https://github.com/5afe/safe-wallet-ipfs

PR #5168, opened 2025-02-28
Fix(Web): use only validated safe address

katspaugh:
## What it solves

`useSafeAddress` was used throughout the app to get the current Safe address. Some places use `safeAddress` from Safe Info instead.

This hook reads directly from the URL params which violates the single source of truth principle and also it doesn't validate that the Safe actually exists.


usame-algan (approved):
Looks good! One possible improvement for the future is to have some kind of skeleton for the sidebar navigation as its now "empty" considerable longer than before.

PR #5172, opened 2025-02-28
chore(yarn): prevent postinstall by default

compojoom:
Initially, I was thinking of using lavamoat for this, but then discovered that lavamoat by default actually sets enableScripts to false, so it seemed like kind of unnecessary to drag lavamoat into this. Having this option set to false means that postinstall scripts from 3rd party packages won't automatically run. Yarn is warning us about this:

grafik

I'm not sure what the scripts are doing and initially thought that nothing is going to work without them, but it turns out that at least on my machine everything I testesd seems to work:
- starting the app in dev mode
- building the web app
- building the mobile app
- running tests on both web and mobile
- linting

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -4,24 +4,26 @@ enableGlobalCache: true
 
 nodeLinker: node-modules
 
+enableScripts: false
+
 packageExtensions:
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
Looks good!

PR #5184, opened 2025-03-03
Feat(mobile): Add signing functionality 

clovisdasilvaneto:
It allows users to sign transactions in the mobile app.

## How to test it
- Create any transaction in the web app
- Go to pending transactions screen
- Select the pending transaction
- Click in "Confirm"

## Screenshots
415225501-7a388839-d573-4d68-8003-52dac84dfff9

## Checklist


apps/mobile/src/features/ConfirmTx/ConfirmTx.container.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const { data, isFetching, isError } = useTransactionsGetTransactionByIdV1Query({
+    chainId: activeSafe.chainId,
+    id: txId,
+  })
+
+  const detailedExecutionInfo = data?.detailedExecutionInfo as MultisigExecutionDetails
+  const { activeSigner, hasSigned } = useTxSigner(detailedExecutionInfo)
+  const hasEnoughConfirmations =
+    detailedExecutionInfo?.confirmationsRequired === detailedExecutionInfo?.confirmations?.length
```
katspaugh, line 35:
```suggestion
    detailedExecutionInfo?.confirmationsRequired <= detailedExecutionInfo?.confirmations?.length
```

There can be more signatures than needed, our backend allows that.

apps/mobile/src/features/ConfirmTx/ConfirmTx.container.tsx
```diff
@@ -0,0 +1,73 @@
...
+    children: <NavBarTitle paddingRight={5}>Confirm transaction</NavBarTitle>,
+    alwaysVisible: true,
+  })
+
+  const { data, isFetching, isError } = useTransactionsGetTransactionByIdV1Query({
+    chainId: activeSafe.chainId,
+    id: txId,
+  })
+
+  const detailedExecutionInfo = data?.detailedExecutionInfo as MultisigExecutionDetails
```
katspaugh, line 32: You shouldn't cast like this, use guards instead.

apps/mobile/src/features/ConfirmTx/components/ConfirmationView/ConfirmationView.tsx
```diff
@@ -0,0 +1,81 @@
...
+  txDetails: TransactionDetails
+}
+
+export function ConfirmationView({ txDetails }: ConfirmationViewProps) {
+  const confirmationViewType = getTransactionType({ txInfo: txDetails.txInfo })
+
+  switch (confirmationViewType) {
+    case ETxType.TOKEN_TRANSFER:
+      return (
+        <QueueSummary
```
katspaugh, line 31: Why is this called QueueSummary? I would expect something like TokenTransfer.

apps/mobile/src/features/ConfirmTx/components/ExecuteForm/ExecuteForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+  chainId: string
+}
+
+export function ExecuteForm({ safeAddress, chainId }: ExecuteFormProps) {
+  const chain = useAppSelector((state) => selectChainById(state, chainId))
+
+  return (
+    <YStack justifyContent="center" gap="$4" alignItems="center" paddingHorizontal={'$4'}>
+      <Text fontSize="$4" fontWeight={400} width="70%" textAlign="center" color="$textSecondaryLight">
+        The transaction can be executed in the web app only.
```
katspaugh, line 19:
```suggestion
        Transactions can be executed only in the web app at the moment.
```

apps/mobile/src/features/ConfirmTx/components/ExecuteForm/ExecuteForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+      </Text>
+
+      <View display="flex" flexDirection="row" alignItems="center" justifyContent="center" gap="$2">
+        <Anchor
+          href={`https://app.safe.global/home?safe=${formatPrefixedAddress(safeAddress, chain?.shortName)}`}
+          target={'_blank'}
+        >
+          Go to Web App
```
katspaugh, line 27:
```suggestion
          Go to web app
```

Don't use Capital Case.

apps/mobile/src/features/ConfirmTx/components/SignTransaction/SignError.tsx
```diff
@@ -0,0 +1,67 @@
...
+import React from 'react'
+import { SafeAreaView } from 'react-native-safe-area-context'
+import { ScrollView, Text, View } from 'tamagui'
+import { Badge } from '@/src/components/Badge'
+import { SafeFontIcon } from '@/src/components/SafeFontIcon'
+import { LargeHeaderTitle } from '@/src/components/Title'
+import { SafeButton } from '@/src/components/SafeButton'
+import { router } from 'expo-router'
+
+const colors: [string, string] = ['#221818', 'transparent']
```
katspaugh, line 12: Is this color not in the theme?

apps/mobile/src/features/ConfirmTx/components/confirmation-views/AddSigner/utils.tsx
```diff
@@ -0,0 +1,68 @@
...
+export const formatAddSignerItems = (
+  txInfo: NormalizedSettingsChangeTransaction,
+  chain: Chain,
+  executionInfo: MultisigExecutionDetails,
+) => {
+  const newSignerAddress = getSignerName(txInfo)
+
+  return [
+    {
+      label: 'New Signer',
```
katspaugh, line 35:
```suggestion
      label: 'New signer',
```

apps/mobile/src/features/ConfirmTx/components/confirmation-views/RemoveSigner/utils.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Text, View } from 'tamagui'
+
+import { Chain } from '@safe-global/store/gateway/AUTO_GENERATED/chains'
+import { Identicon } from '@/src/components/Identicon'
+import { getSignerName } from '../AddSigner/utils'
+
+import { NormalizedSettingsChangeTransaction } from '../../ConfirmationView/types'
+
+export const formatRemoveSignerItems = (txInfo: NormalizedSettingsChangeTransaction, chain: Chain) => {
+  const newSignerRemoveress = getSignerName(txInfo)
```
katspaugh, line 12: What's a "removeress"? New removed signers?

apps/mobile/src/features/ConfirmTx/components/confirmation-views/RemoveSigner/utils.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  return [
+    {
+      label: 'Removed Signer',
```
katspaugh, line 16:
```suggestion
      label: 'Removed signer',
```

apps/mobile/src/features/ConfirmTx/components/confirmation-views/SendNFT/utils.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { ellipsis } from '@/src/utils/formatters'
+import { Text, View } from 'tamagui'
+import { Address } from '@/src/types/address'
+import { Identicon } from '@/src/components/Identicon'
+import { Chain } from '@safe-global/store/gateway/AUTO_GENERATED/chains'
+import { shortenAddress } from '@safe-global/utils/formatters'
+export const formatSendNFTItems = (txInfo: TransferTransactionInfo, chain: Chain) => {
+  return [
+    {
+      label: 'New Signer',
```
katspaugh, line 14:
```suggestion
      label: 'New signer',
```

apps/mobile/src/features/ConfirmTx/components/ConfirmationView/ConfirmationView.tsx
clovisdasilvaneto: Just added this name to follow the same name we have on figma screen. Agree, TokenTransfer makes more sense

apps/mobile/src/config/constants.ts
```diff
@@ -3,7 +3,7 @@ import { Platform } from 'react-native'
 
 // export const isProduction = process.env.NODE_ENV === 'production'
 // TODO: put it to get from process.env.NODE_ENV once we remove the mocks for the user account.
-export const isProduction = true
+export const isProduction = false
```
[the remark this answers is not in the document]
clovisdasilvaneto: yes, just so we can try to execute the transaction in staging env. However it will fails anyway hehe

apps/mobile/src/features/ConfirmTx/components/ExecuteForm/ExecuteForm.tsx
clovisdasilvaneto: Shall we ask @TanyaEfremova to update the designs on figma as well?

apps/mobile/src/app/confirm-transaction.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { SafeAreaView } from 'react-native-safe-area-context'
+import { ConfirmTxContainer } from '@/src/features/ConfirmTx'
+
+function ConfirmTransactionPage() {
+  return (
+    <SafeAreaView style={{ flex: 1 }} edges={['bottom']}>
```
compojoom, line 7: Btw, I don't know if you've seen this doc: https://reactnavigation.org/docs/handling-safe-area/#use-the-hook-for-more-control, but the recommendation is to use the hook instead of the `<SafeAreaView`.

apps/mobile/src/components/TxInfo/TxInfo.tsx
```diff
@@ -19,17 +19,28 @@ import { TxRejectionCard } from '@/src/components/transactions-list/Card/TxRejec
...
   const txInfo = tx.txInfo
+
+  const onCardPress = () => {
+    if (onPress) {
+      onPress({
+        tx,
+        type: txType,
+      })
+    }
+  }
```
compojoom, line 42: maybe wrap it in a useCallback?

apps/mobile/src/components/transactions-list/Card/TxConflictingCard/TxConflictingCard.tsx
```diff
@@ -3,27 +3,37 @@ import { Theme, View } from 'tamagui'
...
+function TxConflictingComponent({ transactions, inQueue, onPress }: TxConflictingCard) {
+  const handleConflictTxPress = () => {
+    onPress(
+      {
+        transactions,
+      },
+      true,
+    )
+  }
+
```
compojoom, line 22: also needs to be wrapped in useCallback?

apps/mobile/src/app/confirm-transaction.tsx
clovisdasilvaneto: hmm I dind't! Will take a look, thanks

apps/mobile/src/features/ConfirmTx/components/confirmation-views/RemoveSigner/utils.tsx
clovisdasilvaneto: removeress is a typo 😆 it should be `newRemovedSigners`

apps/mobile/src/features/ConfirmTx/ConfirmTx.container.tsx
clovisdasilvaneto: Yes, but in the app we need to change the "confirm" button in case the user has enough signatures. We show this layout here: https://www.figma.com/design/yVshG47sZRM7ztI66bDII4/New-Mobile-App?node-id=5000-25641&t=4NQx81UKm8wXZ3Bm-0 , so the user will see that he can only execute it in the web app, that is why I check for enough confirmations.

apps/mobile/src/features/ConfirmTx/ConfirmTx.container.tsx
clovisdasilvaneto: @katspaugh you're right, however I found we have a types cohesion problem, where the client gateway is returning a type from a package that is different that we're using in the guards. 

Since it will require us to change some files in the application (because of the types that needs to be changed, I suggest us to handle that in a separated PR, wdyt?

apps/mobile/src/features/ConfirmTx/components/SignTransaction/SignError.tsx
clovisdasilvaneto: it's not, because the error colors are different.

clovisdasilvaneto: Just checked how it looks with the theme colors, and looks ok as well, will use the theme ones and ask @TanyaEfremova to update Figma screens

apps/mobile/src/features/ConfirmTx/ConfirmTx.container.tsx
katspaugh: Why do you mean? In the web app, there's a `isMultisigExectionDetails` that simply checks the `type` property.
That said, I'm OK with doing it in another PR, this one is too big already.

katspaugh (approved):
LFG! 🚀

apps/mobile/src/features/ConfirmTx/ConfirmTx.container.tsx
katspaugh: What I'm saying is, that a tx can have MORE confirmations than needed.

apps/mobile/src/features/ConfirmTx/ConfirmTx.container.tsx
clovisdasilvaneto: @katspaugh if you take a look in the transaction-guards file, this condition uses some types (`DetailedExecutionInfo, MultisigExecutionDetails`) from the `@safe-global/safe-gateway-typescript-sdk` package, but the type of `data` is from `@safe-global/store/gateway/AUTO_GENERATED/transactions` package, which ended up conflicting with the types in the transaction guard. We need to put the guard to use the types from the  `@safe-global/store/gateway/AUTO_GENERATED/` ones.

Let me know if it is clear, otherwise we can jump in a quick call and I can show to you what I mean.

PR #5186, opened 2025-03-04
fix: add sri integrity plugin to webpack

roninjin10:
## What it solves

Fixes issue of unexpected content being loaded by using [html sri integrity functionality](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity). If unexpected content is loaded from something like swapping out a javascript file attack it will block the content

Resolves #

Helps prevent against issues that caused a 1.5 billion hack. Note: it doesn't protect against the html file itself being compromised so it might be worth hashing that so people can check that the hmtl file matches what they expect manually

## How this PR fixes it

If a javascript file is swapped for another the browser will block it.

## How to test it

Run the build and double check that integrity properties exist in final build

## Checklist

### DIFF
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -39,6 +39,9 @@ const nextConfig = {
     dirs: ['src', 'cypress'],
   },
   experimental: {
+    sri: {
+      algorithm: 'sha256',
+    },
     optimizePackageImports: [
       '@mui/material',
       '@mui/icons-material',


usame-algan:
@roninjin10 thanks for the contribution. Unfortunately, only adding the experimental flag doesn't generate the integrity hashes. It works with NextJS v15.2.0 but only for app router pages so all pages would need to be migrated from the current pages router to the new app router structure ([Docs](https://nextjs.org/docs/pages/building-your-application/upgrading/app-router-migration#migrating-from-pages-to-app))

An alternative approach is to write a custom script that generates the integrity hashes for the static export. I will close this PR for now but feel free to suggest another solution.

PR #5190, opened 2025-03-04
chore: Add script to generate SRI hashes

### LINKED ISSUE #5154: Add support to Subresource Integrity (SRI)
## What is the feature about

Enhancing web interface security implementing a battle-tested security feature of modern browsers, the so called  Subresource Integrity (SRI).

## The list of requirements

During build step, NextJS must add an `integrity` attribute to any `<script>` or `<link>` tag containing the base64-encoded cryptographic hash (I suggest `sha512`) of the remote resource.

## Links

- [Subresource Integrity on MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity)
- [SRI on NextJS](https://github.com/vercel/next.js/issues?q=is%3Aissue%20state%3Aopen%20SRI)

usame-algan:
## What it solves

Resolves #5154

## How this PR fixes it

- Adds a new script called `integrity-hashes.cjs`
- Adds a dev dependency for `cheerio` in order to parse HTML
- When run it iterates over each file in the `out` directory, generates a sha-256 hash for each internal script and inserts the hash with the integrity attribute to the script element
- Runs this script for each PR deployment

## ToDo

- [x] Run the script for Dev, Staging and Prod deployments as well

## How to test it

1. Open the PR deployment
2. Inspect the DOM
3. Observe no errors in the console
4. Observe integrity hashes are added to the script elements

## Screenshots
Screenshot 2025-03-04 at 11 41 51
Screenshot 2025-03-04 at 11 45 32

## Checklist

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -30,7 +30,8 @@
     "static-serve": "yarn build && yarn serve",
     "prepare": "husky",
     "storybook": "storybook dev -p 6006",
-    "build-storybook": "storybook build --quiet"
+    "build-storybook": "storybook build --quiet",
+    "integrity": "node scripts/integrity-hashes.cjs"
   },
   "engines": {
     "node": ">=18"
@@ -130,6 +131,7 @@
     "@types/semver": "^7.3.10",
     "@typescript-eslint/eslint-plugin": "^7.6.0",
     "@typescript-eslint/parser": "^8.18.1",
+    "cheerio": "^1.0.0",
     "cross-env": "^7.0.3",
     "cypress": "^13.15.2",
     "cypress-file-upload": "^5.0.8",
--- a/apps/web/scripts/integrity-hashes.cjs
+++ b/apps/web/scripts/integrity-hashes.cjs
@@ -1,0 +1,65 @@
+const fs = require('fs')
+const path = require('path')
+const crypto = require('crypto')
+const cheerio = require('cheerio')
+
+const OUT_DIR = 'out'
+
+/**
+ * Given a local file path, compute the SHA-256 integrity hash.
+ */
+function generateIntegrityHash(filePath) {
+  const fileContents = fs.readFileSync(filePath)
+  const hash = crypto.createHash('sha256').update(fileContents).digest('base64')
+  return `sha256-${hash}`
+}
+
+/**
+ * Process a single .html file to add SRI attributes to local script tags.
+ */
+function processHtmlFile(htmlFilePath) {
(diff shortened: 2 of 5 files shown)


apps/web/scripts/integrity-hashes.cjs
```diff
@@ -0,0 +1,65 @@
...
+function processHtmlFile(htmlFilePath) {
+  const html = fs.readFileSync(htmlFilePath, 'utf8')
+  const $ = cheerio.load(html)
+
+  $('script[src]').each((_, scriptEl) => {
+    const scriptSrc = $(scriptEl).attr('src')
+    // Skip external or protocol-based (http/https) scripts:
+    if (!scriptSrc || scriptSrc.startsWith('http')) {
+      return
+    }
```
usame-algan, line 29: We don't load any external scripts right now but if we wanted to support integrity hashes we would need to fetch the scripts here e.g. via curl.

usame-algan: I added a comment and a log but I suggest to leave it out of scope for this PR.

apps/web/package.json
```diff
@@ -130,6 +131,7 @@
     "@types/semver": "^7.3.10",
     "@typescript-eslint/eslint-plugin": "^7.6.0",
     "@typescript-eslint/parser": "^8.18.1",
+    "cheerio": "^1.0.0",
```
usame-algan, line 134: If we don't want this dependency we could also try simple regex to match script tags but I think it would be more brittle than an actual html parser.

katspaugh (approved):
Looks great! Quite simple too.

katspaugh:
Looks like this doesn't work for lazy-loaded scripts.

Screenshot 2025-03-04 at 14 24 26

usame-algan:
> Looks like this doesn't work for lazy-loaded scripts.

Good point. Looks like when we import via `next/dynamic` the corresponding chunk is added to the document in runtime which makes sense but it will be missed by the script. I can double-check with an empty project if this works when using the nextjs experimental flag and app router. In that case I suggest we evaluate to do a follow-up with the app router migration.

roninjin10:
You likely could do this with a webpack plugin rather than a script

roninjin10:
Highly recommend just trying the webpack plugin first

roninjin10:
https://www.npmjs.com/package/webpack-subresource-integrity

usame-algan:
> Highly recommend just trying the webpack plugin first

The plugin doesn't work for statically generated files (SSG) so unfortunately not a viable option.

PR #5194, opened 2025-03-05
Feat(Tx flow): better advanced details

katspaugh:
## What it solves

* Always show `to`, `value` and `data` in Advanced Details
* Group hashes under their own section

Screenshot 2025-03-05 at 11 26 23


apps/web/src/components/transactions/TxDetails/Summary/SafeTxHashDataRow/index.tsx
```diff
@@ -1,28 +1,39 @@
...
+  const { safe, safeAddress } = useSafeInfo()
   const chainId = useChainId()
-  const safeAddress = useSafeAddress()
+  const safeVersion = safe.version as SafeVersion
+
+  const computedHash = useMemo(() => {
+    if (safeTxHash) return safeTxHash
+    if (!safe.version) return
+    try {
+      return calculateSafeTransactionHash(safeAddress, safeTxData, safe.version, BigInt(safe.chainId))
```
schmanu, line 24: If we are rendering a SafeTx in the history of a Safe this could lead to wrong hashes if the Safe has been upgraded from e.g. 1.1.0 to 1.3.0 / 1.4.1 as the domain separator changed.

Considering that we recompute and validate the `safeTxHash` anyways in `useValidateTxData` in the signing modals I would use the given `safeTxHash`

katspaugh: It doesn't exist when creating a new tx. This branch of code runs for this case.

schmanu: Oh I am blind. Never mind! I missed the early return -.-

apps/web/src/components/transactions/TxDetails/Summary/index.tsx
```diff
@@ -75,98 +73,86 @@ const Summary = ({ txDetails, defaultExpanded = false, hideDecodedData = false }
...
-      <Box mt={1}>
-        <TxDataRow datatestid="tx-raw-data" title="Raw data:">
-          {generateDataRowValue(txData.data, 'rawData')}
-        </TxDataRow>
-      </Box>
+          <TxDataRow datatestid="tx-to" title="to:">
+            {generateDataRowValue(safeTxData.to, 'address', true)}
+          </TxDataRow>
+
+          <TxDataRow datatestid="tx-to" title="value:">
```
tmjssz, line 100: This should probably be:
```suggestion
          <TxDataRow datatestid="tx-value" title="value:">
```

iamacook (approved):

francovenica:
Was the "Parameters" title removed on purpose?

Comparison between current prod and this PR, same transaction (an owner removal)

image
image

Also given that all the long addresses seems to replace the sort ones, in this particular tx, the "Prev owner" is all 0's, so they go out of bounds
image

katspaugh:
@francovenica title was removed on purpose, yes – to group it visually with the method call.

I'll fix the visual overflow.

PR #5196, opened 2025-03-05
Feat(web): validate tx preview against mitm

schmanu:
## What this PR changes
- validates the decodedData of the transaction preview against the raw transaction data
- If it does not match it shows an error and disabled the sign and execute button

## How to test it
- Create different transactions
- No error should appear

## Screenshots
Screenshot 2025-03-04 at 14 06 10

## Checklist


apps/web/src/components/tx/confirmation-views/useTxPreview.ts
```diff
@@ -2,6 +2,17 @@ import { Operation, getTxPreview } from '@safe-global/safe-gateway-typescript-sd
...
+export const InvalidPreviewErrorName = 'InvalidPreviewError'
+class InvalidPreviewError extends Error {
+  constructor(message: string) {
+    super(message)
+
+    this.name = InvalidPreviewErrorName
+  }
+}
 
 const useTxPreview = (
```
katspaugh, line 17: This hook was very simple – just an RTK call. Now it does more. Can we extract the validation part into a separate hook (can be in the same file) and call it from the old hook?

schmanu:
Closing in favor of https://github.com/safe-global/safe-wallet-monorepo/pull/5197

PR #5197, opened 2025-03-05
Feat(web): validate tx preview

schmanu:
## What this PR changes
- validates the decodedData of the transaction preview against the raw transaction data
- If it does not match it shows an error and disabled the sign and execute button

## How to test it
- Create different transactions
- No error should appear

## Screenshots
Screenshot 2025-03-04 at 14 06 10

## Checklist

### DIFF
--- a/apps/web/src/components/tx/confirmation-views/useTxPreview.ts
+++ b/apps/web/src/components/tx/confirmation-views/useTxPreview.ts
@@ -2,17 +2,6 @@ import { Operation, getTxPreview } from '@safe-global/safe-gateway-typescript-sd
 import type { SafeTransaction } from '@safe-global/safe-core-sdk-types'
 import useAsync from '@/hooks/useAsync'
 import useSafeInfo from '@/hooks/useSafeInfo'
-import { sameAddress } from '@/utils/addresses'
-import { Interface } from 'ethers'
-
-export const InvalidPreviewErrorName = 'InvalidPreviewError'
-class InvalidPreviewError extends Error {
-  constructor(message: string) {
-    super(message)
-
-    this.name = InvalidPreviewErrorName
-  }
-}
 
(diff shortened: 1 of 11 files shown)


katspaugh (approved):
👍

apps/web/src/components/tx/confirmation-views/useTxPreview.ts
```diff
@@ -21,10 +32,40 @@ const useTxPreview = (
...
+        dataDecoded.parameters?.map((param) => param.value),
+      )
+
+      if (rawDataFromDecodedData !== safeTxData.data) {
+        throw new InvalidPreviewError('Decoded data does not match raw data')
+      }
+    }
+
+    return txPreview
+  }, [skip, chainId, address, operation, data, to, value, safeTxData])
```
katspaugh, line 68: I suggest we extract this into a separate hook. Apart from `useTxPreview`, which is used only when _creating_ a tx, we should call this hook also when confirming proposed txs – in that case they use TransactionDetails.

katspaugh (approved):
Although I pointed out that the same code could be used for proposed txs to verify CGW response integrity, I'm fine with only doing it for tx creation – that's where the most value is.
Great idea and implementation!

katspaugh:
I've tested with various tx types (send tokens, send ETH, remove owner, a multi send, a custom contract interaction) – there were no false positives. ✅

liliya-soroka:
Checked type of txs:
1. TransactionBuilder -> contract interaction(46 actions)
2. settings: owners, policy, enable/disable modules
3. spending limits
4. native token transfer
5. erc20 transfer
6. Swap( swap, Limit order, TWAP)
7. spending limit
8. fallbackhandler
9. update mastercopy from 1.1.1 to 1.3.0

PR #5198, opened 2025-03-05
chore: Add SRI hashes for dynamic chunks

usame-algan:
## What it solves

Follow up #5154

## How this PR fixes it

- Extends the `integrity-hashes.cjs` script to iterate over all chunks, generate sha256 hashes and store them in a manifest js file
- Monkey patches `appendScript` from next `router-loader.js` to check each src that is to be appended against the generated map and insert the integrity hash if it exists
- Pins next version so that the patch process doesn't fail

## Checklist


usame-algan:
Closing in favour of #5199

PR #5199, opened 2025-03-05
chore: Add SRI hashes to dynamic chunks

usame-algan:
## What it solves

Follow up #5154

## How this PR fixes it

- Extends the `integrity-hashes.cjs` script to iterate over all chunks, generate sha384 hashes and store them in a manifest js file
- Patches `appendScript` from next `router-loader.js` to check each src that is to be appended against the generated map and insert the integrity hash if it exists
- Pins next version so that the patch process doesn't fail

## How to test it

1. Open the deployment
2. Open the dev console
3. Observe all loaded scripts have an integrity hash
4. Observe there are no script loading errors in the console

## Screenshots
Screenshot 2025-03-05 at 19 15 41

## Checklist


apps/web/scripts/integrity-hashes.cjs
```diff
@@ -4,14 +4,109 @@ const crypto = require('crypto')
...
+        const html = fs.readFileSync(entryPath, 'utf8')
+        const $ = cheerio.load(html)
+
+        // Ideally, put it in <head> so it loads early
+        // so the manifest is available by the time Next tries dynamic chunks
+        if ($('head').length) {
+          $('head').append(`\n<script src="${manifestScriptPath}"></script>\n`)
+        } else {
+          // Fallback: if no <head>, put it at start of <body>
+          $('body').prepend(`\n<script src="${manifestScriptPath}"></script>\n`)
```
katspaugh, line 101: Not sure this is a good idea. This script itself lacks a hash.

Alternatively, we could store the manifest file as JSON, fetch and cache the manifest in runtime.

```js
        // Set integrity if we have a matching hash
        const hashManifest = window. __CHUNK_SRI_MANIFEST || await fetch('/out/sri-manifest.json').then(r => r.json())
        if (hashManifest[src]) {
            script.integrity = hashManifest[src];
        }
```

usame-algan: I was following the implementation of the `__buildManifest.js` that Next generates and imports in the same way. After we add the script to the document we run `addSRIToAllHtmlFiles` which hashes all the static imports including the manifest script
Screenshot 2025-03-06 at 09 17 33

katspaugh (approved):
🎉

apps/web/scripts/integrity-hashes.cjs
```diff
@@ -4,14 +4,109 @@ const crypto = require('crypto')
...
+        }
```
katspaugh, line 102:
```suggestion
        const container = $('head').length ? $('head') : $('body')
        container.append(`\n<script src="${manifestScriptPath}"></script>\n`)
```

PR #5201, opened 2025-03-06
fix: Remove unsafe-inline and GA domains from CSP

usame-algan:
## What it solves

Follow up #5159

## How this PR fixes it

- Removes `unsafe-inline` from script-src since the GTM script that was inlined before has been removed
- Removes `https://www.google-analytics.com` and `https://ssl.google-analytics.com` from script-src because nothing is loaded from those domains

## How to test it

1. Open the app
2. Observe no errors in the console regarding CSP
3. Observe that GA events are still tracked

## Checklist


katspaugh (approved):
Good catch!

schmanu (approved):
🚀

PR #5205, opened 2025-03-06
Fix(mobile): Create a new fiat formatter and implement its usage

Jonathansoufer:
## What it solves
This PR resolves an issue in formatting fiat balance above 1000.

Resolves #

## How this PR fixes it
It implements a new format util to handle all possible scenarios.

## How to test it
Load the app and check if balance is shown correctly.
This [wallet](:https://app.safe.global/transactions/history?safe=eth:0x91c32893216dE3eA0a55ABb9851f581d4503d39b ) can be imported and used to see a different amount.

## Screenshots
https://github.com/user-attachments/assets/d299cc2d-6f06-438d-8952-51acd2d770ce

## Checklist


apps/mobile/src/utils/formatters.ts
```diff
@@ -47,3 +47,51 @@ const calculateRatio = (a: Quantity, b: Quantity) => {
...
+  } else if (numericValue >= 1_000_000) {
+    // Millions
+    formattedValue = `${(numericValue / 1_000_000).toFixed(decimalPlaces)}M`
+  } else if (numericValue >= threshold) {
+    // Thousands
+    formattedValue = `${(numericValue / 1_000).toFixed(decimalPlaces)}k`
+  } else {
+    // Regular numbers
+    formattedValue = numericValue.toFixed(decimalPlaces)
+  }
```
katspaugh, line 93: Any chance we can use `Intl.NumberFormat` [like the web app](https://github.com/safe-global/safe-wallet-monorepo/blob/dev/apps/web/src/utils/formatNumber.ts) instead of custom logic? The `compact` option does exactly the same.

katspaugh: The fiat/token amount formatting should be consistent across clients. I'm not suggesting here to make it a shared package (although this would be ideal) as it's out of scope, but we should re-use the code as much as possible.

apps/mobile/src/components/Fiat/Fiat.tsx
```diff
@@ -6,16 +7,23 @@ interface FiatProps {
...
 
   return (
     <View flexDirection="row" alignItems="center">
       <H3 fontWeight="600">$</H3>
-      <H1 fontWeight="600">{amount[0]}</H1>
 
-      {amount[1] && (
-        <H1 fontWeight={600} color="$textSecondaryDark">
-          .{amount[1].slice(0, 2)}
+      {formattedAmount.includes('k') ? (
```
katspaugh, line 17: Why is it checking specifically for `k`?

apps/mobile/src/utils/formatters.ts
compojoom: @Jonathansoufer why can't we use /packages/utils/src/formatNumber.ts
It has a formatAmount|formatCurrency functions and it will format the number the same way it is done in the web app. 
Since it relies on Intl.NumberFormat it will respect the user's region specified in the OS settings.

compojoom:
Nice, thanks for the update @Jonathansoufer 
you need to pass the precise flag for the balance. Otherwise you will be missing the decimals 
grafik

compojoom:
also notice that it no longer fits the design:
grafik

and if we are viewing a safe with 100M+
grafik

We should accomodate bigger values.

apps/mobile/src/utils/formatters.ts
```diff
@@ -47,3 +47,42 @@ const calculateRatio = (a: Quantity, b: Quantity) => {
...
+  })
+
+  const formattedCurrency = new Intl.NumberFormat(locale, {
+    style: 'currency',
+    currency: currency,
+    minimumFractionDigits: 0,
+    maximumFractionDigits: 2,
+  }).format(Number(formattedNumber))
+
+  return `${formattedCurrency}${units[unitIndex]}`
```
katspaugh, line 74: Why do you need this custom unit handling? The `compact` option of NumberFormat does exactly that. That's what the web app is using. Please don't reinvent the wheel.

Screenshot 2025-03-13 at 15 48 29

compojoom: @katspaugh there is a bug in Hermes and the compact mode doesn't work on iOS. Jonathan and I discussed this and went this route: https://github.com/facebook/hermes/issues/1035

apps/mobile/src/utils/formatters.ts
```diff
@@ -47,3 +47,42 @@ const calculateRatio = (a: Quantity, b: Quantity) => {
   }
   return asDecimal(BigInt(a.amount), a.decimals) / asDecimal(BigInt(b.amount), b.decimals)
 }
+
+export const customIntl = (number = 0, locale = 'en-US', currency = 'USD') => {
```
katspaugh, line 51: This should be covered with unit tests.

compojoom (approved):

PR #5211, opened 2025-03-07
Release 1.52.2

katspaugh:
* Fix(Web): txData can be null


apps/web/src/components/transactions/TxDetails/Summary/index.tsx
```diff
@@ -73,7 +73,7 @@ const Summary = ({
       )}
 
       {/* Advanced TxData */}
-      {isTxDetailsPreview && (
+      {txData && isTxDetailsPreview && (
```
schmanu, line 76: Should we move the entire advanced details into a component and only display it if `txData` is defined? Instead of having the condition on both components?

schmanu (approved):

Issue #5166, opened 2025-02-28
[Orgs] Change dynamic route to query param


usame-algan:
## Links

Epic on Notion: https://www.notion.so/safe-global/Org-onboarding-flow-1208180fe57380078591f5380a0c5d2a

## What must be done

We use a dynamic route for the `orgId` but this might not be supported with SSG on prod build

- Change the dynamic route to a query param the same how we handle txId routes
- Adjust the sidebar navigation logic
- Adjust the AppRoutes generation

## How to test it

1. Open an org
2. Observe the page showing even when running the prod build

usame-algan:
To test the current behaviour:

1. Open the epic branch deployment
2. Go to an org page
3. Reload the page
4. Observe a 404 error

liliya-soroka:
Verified

PR #5213, opened 2025-03-07
Feat(Orgs): replace dynamic routing with query params

jmealy:
## What it solves

#5166
Dynamic routing does not work in the production build.

## How this PR fixes it
Use query params to store the current org ID instead of dynamic routing. 
Routes will now take the from `/organizations/members?orgId={orgId}`

## How to test it
- Open the orgnanizations feature.
- Make sure all the links between different pages are working (members, accounts, dashboard etc.)

## Checklist


katspaugh (approved):
👍

PR #5227, opened 2025-03-10
Fix(Orgs): Update orgs list when signing in

usame-algan:
## What it solves

Small fix when signing in on the orgs list page. Previously it would try to fetch user info without the cookie which resulted in an empty orgs list after signing in.

## How this PR fixes it

- Skip the user info fetch if the user is not signed in

## How to test it

1. Go to the orgs list page
2. Sign in
3. Observe your orgs showing up without having to reload

## Checklist


jmealy (approved):
nice!

Issue #5231, opened 2025-03-10
[Orgs] added safes are rendered in orgs safe list


jmealy:
## Bug description
On the safe accounts list for an org, before the request for the org safes is resolved, the added safes list is briefly shown. It results in a flicker when refreshing the org safes list.

## Steps to reproduce

1.  Go to the accounts page for an org
2. Refresh the page

It is easiest to see in orgs with no safe accounts added yet.

## Expected result
The org accounts are rendered.
## Obtained result
The added safes will show briefly before the correct list is shown.

liliya-soroka:
Verified

PR #5233, opened 2025-03-10
Fix(Orgs): added safes are shown on the org accounts list

jmealy:
## What it solves

Resolves [#5231](https://github.com/safe-global/safe-wallet-monorepo/issues/5231)

## How this PR fixes it
- Sends an empty array to `useAllSafesGrouped` instead of undefined so that the fallback (added safes) is not used.

## How to test it
- Go to the accounts page for an org
- Refresh the page
- See that the added safes are not shown briefly before the page has finished loading.


usame-algan (approved):
Nice solution!

Issue #5234, opened 2025-03-10
[Bugs Orgs] signing in with a wallet that has no user account


jmealy:
## What must be done
- On the /welcome/organizations page, if you sign in using a wallet with no user account created yet, nothing happens. You are still shown the sign in button.
- You should be shown the empty orgs list page.

## How to test it
- go to /welcome/organizations
- switch to a wallet that has not created or been invited to any orgs yet.
- click the sign in button.
- You should be shown the empty orgs list view.

PR #5235, opened 2025-03-10
Fix(Orgs): show empty orgs list state when user account is not found

jmealy:
## What it solves
[5234](https://github.com/safe-global/safe-wallet-monorepo/issues/5234)

## How this PR fixes it
Shows the empty orgs list if no user account is found for the signed in wallet address.

## How to test it
- go to /welcome/organizations using a wallet with no account
- click connect and then sign in.
- You should be shown the "no orgs" state.

## Checklist

### DIFF
--- a/apps/web/src/features/organizations/components/OrgsList/index.tsx
+++ b/apps/web/src/features/organizations/components/OrgsList/index.tsx
@@ -117,7 +117,7 @@ const OrgsList = () => {
 
         {isUserSignedIn ? (
           <Grid2 container spacing={2} flexWrap="wrap">
-            {currentUser && activeOrganizations.length > 0 ? (
+            {activeOrganizations.length > 0 ? (
               activeOrganizations.map((org) => (
                 <Grid2 size={6} key={org.name}>
                   <OrgsCard org={org} />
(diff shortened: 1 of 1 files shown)


apps/web/src/features/organizations/components/OrgsList/index.tsx
```diff
@@ -115,9 +115,9 @@ const OrgsList = () => {
             <OrgListInvite key={invitingOrg.id} org={invitingOrg} />
           ))}
 
-        {isUserSignedIn && organizations ? (
+        {isUserSignedIn ? (
           <Grid2 container spacing={2} flexWrap="wrap">
-            {activeOrganizations.length > 0 ? (
+            {currentUser && activeOrganizations.length > 0 ? (
```
usame-algan, line 120: Do we need to check for `currentUser` here if we already checked for `isUserSignedIn`? I think currentUser could still be cached by Redux even if technically the user has logged out. We should double-check if this is cleaned up correctly.

usame-algan (approved):

apps/web/src/features/organizations/components/OrgsList/index.tsx
jmealy: For the case where you sign in but you haven't created an org yet, then `isUserSignedIn` will be true but `currentUser` will be undefined (GET `/v1/users` will return 404). Could also check specifically for this 404 response if it's more clear

usame-algan: But wouldn't `activeOrganizations` in that case be empty if the user is signed in but we can't find any orgs for them?

PR #5251, opened 2025-03-10
Fix(mobile): infinite scroll cache invalidation duplicates the list data

clovisdasilvaneto:
## What it solves
The `useInfiniteScroll` hook is duplicating the `list` data every time the useEffect is triggered. The reason why it is happening is because we do not check if the consumer component is really going to the next page or just refetching the list data again.

## How this PR fixes it
Now we're comparing the previous url of the data with the previous url of the list, if both are the same, we just set whatever is in the data, otherwise we concatenate the data with the previous data, just like before.

## How to test it
- This issue is only reproducible within the Signer functionality, since before we were not invalidating the pending transactions data.

## Checklist


compojoom:
@clovisdasilvaneto - looks good, but I don't understand why we have so many changes in the PR. It seems that it does way more than just the infinite scroll fix?

clovisdasilvaneto:
@compojoom it was pointing to the wrong target branch 😮‍💨

compojoom (approved):

PR #5258, opened 2025-03-10
Fix(Orgs): Add signed out state to every orgs view

### LINKED ISSUE #5252: [Orgs] Sign in screen should be visible on all org views
## Links

## What must be done

- When not signed in with an authorized wallet the org view should display a sign in screen on all (sub) pages

## How to test it

1. Open an org dashboard while not being signed in
2. Observe the sign in screen
3. Go to safe accounts
4. There should be the same sign in screen
5. Go to Members and Settings
6. There should be the same sign in screen

usame-algan:
## What it solves

Resolves #5252

## How this PR fixes it

- Extracts `SignedOutState` into its own file
- Uses that component on every view if user is not signed in

## How to test it

1. Open an org
2. Disconnect your wallet
3. Navigate through each page
4. Observe the sign in view

## Checklist


jmealy (approved):
👍

PR #5264, opened 2025-03-11
Fix(Orgs): Display selected safes number, reset form state on cancel

### LINKED ISSUE #5140: [Orgs] Safe account counter is misscounting
## Bug description
When selecting a multichain safe, the safe counter in the button is counting also the first row, when it should only count the selected safes

## Steps to reproduce

1.  Have an org created and a user that is owner of multichain safes
2. Go to "Add safe accounts" list
3. Click on a multichain account 
4. See the counter

## Screenshots

Here I'd expect to count just 2 when selecting the upper row
Image

### LINKED ISSUE #5253: [Orgs] Search filter is not cleaned without page refresh
## Bug description
Search filter is not cleaned without page refresh 
## Environment

- Browser: Chrome
- Wallet: any 
- Chain: any

## Steps to reproduce

1.  Go to the add account 
2. search the safe  by the name 
3. add found safe
4. click "Add account" again
Current result: 
only one safe filtered by name is displayed even with empty search field
The filter is removed after the page refresh 

## Expected result
the filter should be cleaned after closing add account flow 

## Screenshots

https://github.com/user-attachments/assets/2537ee48-14bf-41eb-b324-7a9ff067816d

usame-algan:
## What it solves

Resolves #5140 
Resolves #5253

## How this PR fixes it

- Filters out elements starting with `multichain_` in the selectedSafes object
- Resets form state and search query when cancelling the form

## How to test it

1. Open an org
2. Go to the safe accounts page
3. Add account
4. Select multichain accounts
5. Observe the correct number is displayed in the submit form
6. Cancel the form
7. Reopen it
8. Observe the form state is reset

## Checklist

### DIFF
--- a/apps/web/src/features/organizations/components/AddAccounts/index.tsx
+++ b/apps/web/src/features/organizations/components/AddAccounts/index.tsx
@@ -85,7 +85,7 @@ const AddAccounts = () => {
     } catch (e) {
       console.log(e)
     } finally {
-      setOpen(false)
+      handleClose()
     }
   })
 
@@ -106,7 +106,7 @@ const AddAccounts = () => {
     setValue(`selectedSafes.${safeId}`, true, { shouldValidate: true })
   }
 
-  const handleCancel = () => {
+  const handleClose = () => {
     setSearchQuery('')
     reset()
     setOpen(false)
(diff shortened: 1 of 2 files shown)


apps/web/src/features/organizations/components/AddAccounts/index.tsx
```diff
@@ -58,18 +62,16 @@ const AddAccounts = () => {
...
   })
 
-  const { handleSubmit, watch, setValue } = formMethods
+  const { handleSubmit, watch, setValue, reset } = formMethods
 
   const selectedSafes = watch(`selectedSafes`)
-  const selectedSafesLength = Object.values(selectedSafes).filter(Boolean).length
+  const selectedSafesLength = getSelectedSafes(selectedSafes).length
 
   const onSubmit = handleSubmit(async (data) => {
```
jmealy, line 70: we should probably clear the search on submit as well

usame-algan: Good catch! Pushed a fix

jmealy (approved):

PR #5276, opened 2025-03-11
Refactor(Web): set trailing slashes to true in the Next config

katspaugh:
## What it solves

Switch on the trailingSlashes Next config to avoid renaming HTML files when deploying.


katspaugh:
It seems to be still creating html files instead of folders.

https://slashes--walletweb.review.5afe.dev/welcome.html works
https://slashes--walletweb.review.5afe.dev/welcome/ doesn't.

compojoom:
Look ok, but what happens with all previous links in existance? They will now all lead to a 404? Isn't there a way in next to do a redirect when we detect an url doesn't have a trailing link?

katspaugh:
@compojoom good point. I think this won't work in the PR environment, but in prod there's a custom 404 page where we can add this redirect (it would have to be a custom one).

compojoom (approved):

liliya-soroka:
1. My accounts instead of dashboard on the safe opening from the My accounts list 

https://github.com/user-attachments/assets/beec6e44-b73f-4e4e-ae37-de5eaf5024db

liliya-soroka:
2. UI navigation is broken on the Transactions page – there is no indication when switching between tabs.

https://github.com/user-attachments/assets/0a54d9f5-d47a-4811-a96c-54f559be9e7f

PR #5277, opened 2025-03-11
Fix(mobile): Use CGW for signing transactions

clovisdasilvaneto:
## What it solves
We were using api-kit for signing transactions, which stop working some weeks ago due to the server being down. After talking to @iamacook he mentioned that we actually should be using CGW for signing transactions, since that is the reason it was designed for.

## How this PR fixes it
It removes the lazy query we were using and uses the `addConfirmation` mutation instead.

## How to test it
- Add any wallet with a pending transaction
- go to your pending transactions
- click in the transaction that is pending
- click in the "Confirm" button

#### Available testing video
https://github.com/user-attachments/assets/27111040-055a-4eae-a2e8-b2a5ea4fe3ee

## Checklist

### DIFF
--- a/apps/mobile/src/features/ConfirmTx/components/SignTransaction/SignTransaction.tsx
+++ b/apps/mobile/src/features/ConfirmTx/components/SignTransaction/SignTransaction.tsx
@@ -6,27 +6,44 @@ import { selectChainById } from '@/src/store/chains'
 import { RootState } from '@/src/store'
 import type { ChainInfo } from '@safe-global/safe-gateway-typescript-sdk'
 import { useLocalSearchParams } from 'expo-router'
-import { useLazySignTransactionQuery } from '@/src/store/signersBalance'
 import { getPrivateKey } from '@/src/hooks/useSign/useSign'
 import SignError from './SignError'
 import SignSuccess from './SignSuccess'
+import { signTx } from '@/src/services/tx/tx-sender/sign'
+import { useTransactionsAddConfirmationV1Mutation } from '@safe-global/store/gateway/AUTO_GENERATED/transactions'
+import logger from '@/src/utils/logger'
 
 export function SignTransaction() {
   const { txId, signerAddress } = useLocalSearchParams<{ txId: string; signerAddress: string }>()
   const activeSafe = useDefinedActiveSafe()
   const activeChain = useAppSelector((state: RootState) => selectChainById(state, activeSafe.chainId))
 
-  const [trigger, { isFetching, data, isError }] = useLazySignTransactionQuery()
+  const [addConfirmation, { isLoading, data, isError }] = useTransactionsAddConfirmationV1Mutation()
 
   const sign = useCallback(async () => {
-    const privateKey = await getPrivateKey(signerAddress)
+    try {
+      const privateKey = await getPrivateKey(signerAddress)
 
-    trigger({
-      chain: activeChain as ChainInfo,
-      activeSafe,
-      txId,
-      privateKey,
-    })
+      const signedTx = await signTx({
+        chain: activeChain as ChainInfo,
+        activeSafe,
+        txId,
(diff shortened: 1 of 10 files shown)


apps/mobile/src/features/ConfirmTx/components/SignTransaction/SignTransaction.tsx
```diff
@@ -6,27 +6,44 @@ import { selectChainById } from '@/src/store/chains'
...
-      chain: activeChain as ChainInfo,
-      activeSafe,
-      txId,
-      privateKey,
-    })
+    try {
+      const privateKey = await getPrivateKey(signerAddress)
+
+      const signedTx = await signTx({
+        chain: activeChain as ChainInfo,
```
iamacook, line 28: Nit: can this be achieved with a type guard instead?

apps/mobile/src/features/ConfirmTx/components/SignTransaction/SignTransaction.tsx
```diff
@@ -6,27 +6,44 @@ import { selectChainById } from '@/src/store/chains'
...
+          // @ts-ignore
+          signature: signedTx.signature,
+        },
+      })
+    } catch (error) {
+      logger.error('Error signing transaction:', error)
+    }
   }, [activeChain, activeSafe, txId, signerAddress])
 
   useLayoutEffect(() => {
```
iamacook, line 49: Not a mobile dev but why is this happening in an effect? Is there an [event handler that can be used](0x6e4c29936929629df7af5986078b71f879c1694a9bba03c2f983d4ff21bc483a695f04341b1f1533d4420fb5b91f13ea66291901197d20ac19420a5f81d70d151f)?

apps/mobile/src/services/tx/tx-sender/sign.ts
```diff
@@ -37,56 +28,29 @@ export const signTx = async ({
   }
 
   const { protocolKit, wallet } = await createConnectedWallet(privateKey, activeSafe, chain)
-  const { safeTx, signatures } = await createExistingTx({
+  const { safeTx } = await createExistingTx({
```
iamacook, line 31: Nit: I can't add a comment to the function itself but I'd suggest renaming this to something like `proposeTx` for clarity.

apps/mobile/src/features/ConfirmTx/components/SignTransaction/SignTransaction.tsx
clovisdasilvaneto: @iamacook this is done in a effect because this transaction is done in a "loading screen" (you can see it in the video). So there is no explicit trigger for this call in the screen.

apps/mobile/src/features/ConfirmTx/components/SignTransaction/SignTransaction.tsx
clovisdasilvaneto: we're having some types conflict in the app 😞 We will address it in a separated PR soon

apps/mobile/src/features/ConfirmTx/components/SignTransaction/SignTransaction.tsx
iamacook: Is it possible to do it when pressing "Confirm"? It's generally [good practice](https://react.dev/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers) to avoid event-specific logic in an effect.

clovisdasilvaneto: yeah I know, the thing is that when pressing confirm we change the screen of the app, its like going from one page to another in a web project, so I lose context of everything including the `data` which is returned from the mutation.

compojoom (approved):

PR #5280, opened 2025-03-11
fix(web): notification registration was failing

compojoom:
## What it solves
Some users were getting a „code 633: Notification (un-)registration faield“ when trying to register for push notifications. This seems to be caused by an error coming from the service worker happening right before that: „uncaught (in promise) bad-precaching-response: … dynamic-css-manifest.json status 404

## How this PR fixes it
It excludes some files that were never generated. 

## How to test it
Most of the time this happens in private mode. 
- start a session in private mode
- try to register for push notifications

## Checklist

### DIFF
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -13,6 +13,24 @@ const withPWA = withPWAInit({
   dest: 'public',
   workboxOptions: {
     mode: 'production',
+    exclude: [
+      // Some users report bad-precaching-responses. We exclude the problematic files.
+      // https://github.com/shadowwalker/next-pwa/issues/424#issuecomment-1332258575
+      ({ asset }) => {
+        // Add here any file that fails pre-caching
+        const excludeList = [
+          // Default Serwist https://serwist.pages.dev/docs/next/configuring/exclude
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Thanks!

compojoom:
Something seems to be wrong with the build now:
grafik

next is not generating an `<a>` tag anymore

compojoom:
Actually I see this behavior on other PRs as well, but not on the release 1.53 branch. @katspaugh any ideas what is going on?

katspaugh:
That branch is synced with `dev` so it should be the same.

usame-algan (approved):
I don't see the pre-caching issue with the manifest file in inkognito anymore. However the notification registration is still failing there because of notification permissions in the browser not being turned on. Not sure if thats even possible in a private window so I would suggest not fixing it.

compojoom:
Incognito mode was maybe a bad call for reproduction steps. I was not able to subscribe to push notifications in FF in "normal mode", with the changes in this PR it worked fine. However @Jonathansoufer isn't able to use the web notifications even with this PR. This needs further debugging.

PR #5288, opened 2025-03-12
fix(Tx flow): do not render SignerForm for fully signed transactions

### LINKED ISSUE #5204: Signer selector keeps reseting

schmanu:
## What it solves

Resolves #5204 

## How this PR fixes it
- Does not render `SignerForm` for fully signed transactions
- Simplifies selecting the first option if none is selected

## How to test it
- Setup a nested Safe
- Create a new transaction and fully sign it
- Observe the SignerForm being available during signing
- Observe the SignerForm not being rendered when executing the fully signed transaction

## Checklist


katspaugh (approved):

liliya-soroka:
@schmanu :
1. The drop-down is empty by default and the error is displayed during the tx creation 
image

PR #5300, opened 2025-03-12
fix(Signing UX): improvements and fixes

tmjssz:
Various fixes, e.g.:
* fix conditions to show batch button
* setTxOrigin in ReviewTransactionContent component to make transaction notes work
* rename components to `SignOrExecuteFormV2` and `SignFormV2` for better clarity


katspaugh (approved):
Personally not a fan of "v2" – for me it makes it even more confusing. But approving already as it fixes the lint and build.

tmjssz:
> Personally not a fan of "v2" – for me it makes it even more confusing. But approving already as it fixes the lint and build.

Yes I agree V2 suffixes are pretty ugly but the idea is to have it only temporarily. Didn't want to override the existing components and break all the other transaction flows that are still using them. As soon as we migrated all flows to use the new components we can rename them and remove the old ones.

PR #5187, opened 2025-03-04
Release 1.53.0

katspaugh:
## Features
* Feat(Tx flow): show blockaid simulation results on tx confirmation screen (#5203)
* feat(Assets): mask fiat balance when missing conversion (#5127)
* Feat(Sidebar): always show current safe; consistent New transaction button (#4928)
* feat: targeted features (#4992)
* feat(env var): allow for control over the default chainid (#5029)

## Bug fixes
* fix(Safe Apps): keep appUrl query param when switching network (#5081)
* fix(permissions): check spending limits for connected wallet (#5030)
* fix(store): RTK query had wrong return types (#5034)
* Fix(Tx flow): progress bar color (#4996)
* Fix(Simulation): change the error styling (#4994)
* Fix(Balances): regression after useLoadBalances refactoring (#4913)
* Fix(Proposers): allow deleting proposed txs (#4873)
* Fix(Dashboard): hide widget if empty
* Fix(CSS): fallback handler component styles (#4896)
* fix(Safe creation): revalidate AddressInput when chain is switched (#5118)

## Refactoring
* Refactor(web): faster SafeInfo loading (#4883)
* Refactor(Tx flow): simplify extractTxInfo (#4876)
* Refactor: use balances RTK query (#4770)
* Refactor(Staking): rename hook file
* feat(store): add a safeGetOverviewForMany endpoint
* feat(store): generate lazy RTK hooks for endpoints
* feat: create utils package
* Refactor(Web): rename app component (#4910)
* refactor(permissions): Move types to config (#4875)

<details>
<summary><h2>Chores</h2></summary>

* fix(web): storybook was not compiling
* chore(e2e): Select events from new dataLayer structure (#5182)
* chore(yarn): prevent postinstall by default (#5172)
* chore(ci): generate artifact attestation (#5133)
* Chore(deps): Bump CatChen/eslint-suggestion-action from 4.1.9 to 4.1.10 (#5083)
* Chore(deps): Bump @web3-onboard/walletconnect from 2.6.1 to 2.6.2 (#5085)
* Chore(deps-dev): Bump jest-expo from 52.0.2 to 52.0.4 (#5086)
* Chore(deps-dev): Bump typescript-eslint from 8.22.0 to 8.24.1 (#5087)
* Chore(deps): Bump semver from 7.6.3 to 7.7.1 (#5088)
* Chore(deps): Bump expo-constants from 17.0.5 to 17.0.6 (#4999)
[...]

PR #5305, opened 2025-03-13
Fix: validation on address input being triggered on load

jmealy:
## What it solves

Resolves https://github.com/safe-global/safe-wallet-monorepo/pull/5187#issuecomment-2720305035

## How this PR fixes it
- only triggers validation on chain change if the address is present.

## How to test it
- Go anywhere that address input is used and see that validation is not triggered on load

## Checklist


katspaugh (approved):
👍 
@chiranjeev13 FYI

Issue #5315, opened 2025-03-13
[Orgs] Hide certain buttons for non admins


jmealy:
## Bug description
Action buttons are visible to members who are not admins of an org, even when they cannot perform the action in question.
- add/delete members buttons
- add/delete accounts buttons
- all buttons on the settings page

## Environment

- Browser: Chrome
- Wallet: MetaMask
- Chain: Ethereum mainnet

## Steps to reproduce

1.  Go to on org as a non admin member
2. See that there are buttons on the members page, admin page, and accounts page which are visible but will fail with an unauthorized error if you try to perform the action.

PR #5316, opened 2025-03-13
Fix(Orgs): Hide buttons for non admins

jmealy:
## What it solves

partially Resolves [#5315](https://github.com/safe-global/safe-wallet-monorepo/issues/5315)
note: the admin page will be tackled separately once there are designs for it

## How this PR fixes it
For non admin users:
- Hides the add members button on the members page
- Hide the delete invited member button on the members page
- Hides the add accounts button on the accounts page

## How to test it
- open an org as a non admin user.
- Check that the buttons mentioned above are not visible
- Change to an admin user
- Check that the buttons are visible

## Screenshots
image

## Checklist


usame-algan (approved):
Nice one! 👍

PR #5323, opened 2025-03-13
feat(nextjs): sri support

sambacha:
# SRI Hash Support

> [!IMPORTANT]
> Please do not merge this just yet, I am opening it to get feedback first!

## Background

Currently safe ui uses a post-process workflow for inserting the SRI hash, this has some potential issues. NextJS v15 has experimental support for SRI support, however I am proposing a different way of doing this that will actually be of better benefit. 

```
safe-wallet-monorepo/apps/web/scripts/integrity-hashes.cjs
```

There is support for it in NextJS now, but Let me suggest a better way of doing this.

### SRI Import Map

> read more here: <https://shopify.engineering/shipping-support-for-module-script-integrity-in-chrome-safari>

https://github.com/sambacha/sri-import-map#how-it-works

This library uses the import map specification to associate module URLs with their integrity hashes:

```json
{
  "imports": {
    "square": "./module/shapes/square.js"
  },
  "integrity": {
    "./module/shapes/square.js": "sha384-oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8wC"
  }
}
```

#### During the build process, the NextJS plugin:

Identifies modules to be included in the import map
Calculates SHA-384 integrity hashes for each module
Generates an import map with integrity information
Optionally updates security headers in Vercel configuration

#### At runtime, the client-side library:

Loads the import map
Verifies module integrity before execution
Prevents execution of compromised modules

## NextJS SRI config flag

Use the nextjs experimental flag,

```js
    sri: {
      algorithm: 'sha384',
    },
```

### NextJS SRI Issue

Currently webpack chunks loaded via flight runtime do not get integrity
hashes. This was previously unobservable in this test because these scripts
are inserted by the webpack runtime and immediately removed from the document.
However with the advent of preinitialization of chunks used during SSR there are
some script tags for flight loaded chunks that will be part of the initial HTML
but do not have integrity hashes. Flight does not currently support a way to
provide integrity hashes for these chunks. When this is addressed in React upstream
we can revisit this tests assertions and start to ensure it actually applies to
all SSR'd scripts. For now we will look for known entrypoint scripts and assume
everything else in the <head> is part of flight loaded chunks

> source: <https://github.com/vercel/next.js/pull/73891/files#diff-3c2194b9775dc490d232fd610c84e85bf320eac7a0a38dc4968d38cc67ae0cb1R153-R163>

### DIFF
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -1,4 +1,5 @@
-import path from 'path'
+// @ts-check
+import path from 'node:path'
 import withBundleAnalyzer from '@next/bundle-analyzer'
 import withPWAInit from '@ducanh2912/next-pwa'
 import remarkGfm from 'remark-gfm'
@@ -17,12 +18,13 @@ const withPWA = withPWAInit({
   reloadOnOnline: false,
   /* Do not precache anything */
   publicExcludes: ['**/*'],
-  buildExcludes: [/./],
+ // buildExcludes: [/./],
   customWorkerSrc: SERVICE_WORKERS_PATH,
   // Prefer InjectManifest for Web Push
+  // @ts-ignore
   swSrc: `${SERVICE_WORKERS_PATH}/index.ts`,
+  disable: false
 })
-
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   output: 'export', // static site export
@@ -33,7 +35,8 @@ const nextConfig = {
   },
 
   pageExtensions: ['js', 'jsx', 'md', 'mdx', 'ts', 'tsx'],
-  reactStrictMode: false,
+  reactStrictMode: true,
+  poweredByHeader: false,
   productionBrowserSourceMaps: true,
   eslint: {
     dirs: ['src', 'cypress'],
@@ -47,7 +50,17 @@ const nextConfig = {
       '@sentry/react',
       '@gnosis.pm/zodiac',
     ],
+    sri: {
+      algorithm: 'sha384',
+    },
+
   },
+  deploymentId: 'web',
+  bundlePagesRouterDependencies: true,
+//  generateBuildId: async () => {
+    // This could be anything, using the latest git hash
+  //  return process.env.GIT_HASH
+ // },
   webpack(config, { dev }) {
     config.module.rules.push({
       test: /\.svg$/i,
@@ -103,13 +116,11 @@ const nextConfig = {
 }
 const withMDX = createMDX({
   extension: /\.(md|mdx)?$/,
-  jsx: true,
   options: {
     remarkPlugins: [remarkFrontmatter, [remarkMdxFrontmatter, { name: 'metadata' }], remarkHeadingId, remarkGfm],
     rehypePlugins: [],
   },
 })
-
 export default withBundleAnalyzer({
   enabled: process.env.ANALYZE === 'true',
 })(withPWA(withMDX(nextConfig)))


katspaugh:
Thanks! Have you tested locally if this works with our SSG setup? `yarn build & yarn serve`.

usame-algan:
@sambacha thanks for the contribution! I wasn't aware of an import-map specification but it looks a lot like the manifest object we generate that maps each chunk name to its computed hash. We patch the low-level `appendScript` function from Next.js to include this integrity hash for every import. One of the reasons we decided against using the experimental Next.js flag is that it only supports app router pages while our codebase is still using the pages router. However it would be nice to have a more robust way of including those hashes. Is this what your client library achieves with the module integrity middleware that wraps the next config? Could you elaborate a bit more on how it works?

sambacha:
> Thanks! Have you tested locally if this works with our SSG setup? `yarn build & yarn serve`.

No, I first wanted to get additional feedback before doing anything more involved. 

It seems @usame-algan explains my reservation ("surely they looked into NextJS support for SRI hashes and have a reason to not use this configuration setting"). Thank you for the additional context, yes it makes sense.


Import Maps are now part of the HTML standard, <https://html.spec.whatwg.org/multipage/webappapis.html#import-maps>, more specifically the import map sri support has recently landed (this also means ESM modules can use SRI hashes directly via import url within a javascript file, e.g. via a CDN)[^1]. They also enforce deterministic module resolution to ensure that new import maps cannot define the module resolution for modules that were already defined by past import maps, or for ones that were already resolved. 

HTML Example using SRI and Import Map:

```html
<script type="importmap">
  {
    "imports": {
      "square": "./modules/shapes/square.js"
    },
    "integrity": {
      "./modules/shapes/square.js": "sha384-oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8wC"
    }
  }
</script>
```


Screenshot 2025-03-14 at 3 28 10 AM


Each `<script type="module"`> tag on a web page initiates the creation of a module graph. This graph represents the relationships between JavaScript modules, where each module can import and depend on others. Unlike a traditional tree structure, a module graph allows circular dependencies, meaning modules can reference each other in loops.

The module graph originates from the module specified in the <script type="module"> tag, which serves as the starting point.

The execution of this module graph follows a deterministic order, dictated by a post-order depth-first traversal starting from the root module. This means that dependencies are evaluated in a structured sequence after all their sub-dependencies have been resolved. For example, given the following dependency structures:

Example 1: Modules are evaluated in the order: ` a.js, b-2.js, b-1.js, root.js.`
Example 2: Modules are evaluated in the order: `a-2.js, a-1.js, b.js, root.js.`

I mention this because import maps effect this resolution, meaning you can create an import map for dependencies that are 'sensitive' and an import map for dependencies that are not quite security sensitive. 


More generally, the build artifact that is produced on GitHub perhaps can be used to provide provenance against the artifact served on the server instance, which may be facilitated by a service worker or 3rd party wallet (e.g. metamask already does this in practice for updating its phishing database). 

A lot to digest, please ask away I am sure I did not explain everything, I do not think simply providing SRI hashes will move the bar enough w.r.t providing a more secured frontend environment.  




[^1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap

sambacha:
For some previous artwork regarding out of band verification, Facebook does this with WhatsApp, they have a Chrome extension: https://github.com/facebookincubator/meta-code-verify you can read more about it here: https://blog.cloudflare.com/cloudflare-verifies-code-whatsapp-web-serves-users/

Basically, it works like this:

1. WhatsApp publishes the latest version of their JavaScript libraries to their servers, and the corresponding hash for that version to Cloudflare’s audit endpoint.

2. A WhatsApp web client fetches the latest libraries from WhatsApp.

3. The Code Verify browser extension subsequently fetches the hash for that version from Cloudflare over a separate, secure connection.

4. Code Verify compares the “known good” hash from Cloudflare with the hash of the libraries it locally computed.

sambacha:

> Thank you for your submission, we really appreciate it. Like many open-source projects, we ask that you sign our [Contributor License Agreement](https://safe.global/cla) before we can accept your contribution. You can sign the CLA by just posting a Pull Request Comment same as the below format.
> 
> I have read the CLA Document and I hereby sign the CLA
> 
> You can retrigger this bot by commenting **recheck** in this Pull Request. Posted by the **CLA Assistant Lite bot**.

katspaugh:
That sounds great! We could probably also build a separate website or browser extension to verify that the Safe app hasn’t been modified. This could even extend to HTML file verification, which, as far as I know, lacks an existing SRI solution.

katspaugh:
Closing the pull request at this time, thanks for the fruitful discussion. We'll follow up on this topic internally.

PR #5327, opened 2025-03-14
Revert "fix(web): appUrl query was getting lost while switching network in safeApp "

### LINKED ISSUE #5350: Deep links for listed Safe Apps stopped working
The Zodiac Safe App uses deep links to their app and this stopped working after #5081

We should restore the previous behavior.

katspaugh:
Reverts #5081

Resolves #5350.

It broke the use case when you want to test an unlisted Safe App, e.g. running on localhost.

### DIFF
--- a/apps/web/src/components/common/NetworkSelector/index.tsx
+++ b/apps/web/src/components/common/NetworkSelector/index.tsx
@@ -81,7 +81,6 @@ export const getNetworkLink = (router: NextRouter, safeAddress: string, networkS
     safe?: string
     chain?: string
     safeViewRedirectURL?: string
-    appUrl?: string
   }
 
   const route = {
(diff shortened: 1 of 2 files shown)


liliya-soroka:
@katspaugh , question:
I have tried to add custom app on the prod with the https://github.com/safe-global/safe-wallet-monorepo/pull/5081 fix and there is no issue with adding custom ( unlisted) safe app.
Are we sure that the problem is related to our custom app's functionality and not the user-safe app on localhost?

katspaugh:
Hmm, I only tested pasting a Safe App URL directly in the main URL. If adding a custom app still works, we can keep it this way and close this PR.

katspaugh:
Reopening because of #5350.

usame-algan (approved):

PR #5328, opened 2025-03-14
feat(web): clear pending txs

### LINKED ISSUE #5326: Option to clear pending transactions of a Safe
## What is the feature about
Sometimes transactions get stuck in a pending State. Currently users would have to wait for the timeout or clear the local storage of the browser.
The latter bears the risk that the user also deletes the address book or other important data.
Therefore we should have a button in the settings to clear the pending transaction state of the currently opened Safe.

## The list of requirements
- New Button in the _Data_ section of the Safe settings to clear Safe's pending transaction state

## Designs/sketches
tbd

schmanu:
## What it solves

Resolves #5326 

## How this PR fixes it
- Adds new section for clearing the pending transactions to the _Data_ tab of the _settings_ page
- New button that triggers action to clear pending transactions of the **currently opened** Safe.

## How to test it
- Queue a transaction and fully sign it
- Execute the transaction with a too low gas price (e.g. 0.0001 Max Fee on Sepolia)
- Go to the Settings / Data page 
- Button should show one pending transaction
- Clearing the pending transaction should reset the transactions state to being executable again.

## Screenshots
Screenshot 2025-03-14 at 09 17 42

## Checklist

### DIFF
--- a/apps/web/src/components/settings/ClearPendingTxs/index.test.tsx
+++ b/apps/web/src/components/settings/ClearPendingTxs/index.test.tsx
@@ -1,0 +1,104 @@
+import React, { act } from 'react'
+import { fireEvent, screen } from '@testing-library/react'
+import { ClearPendingTxs } from '../ClearPendingTxs'
+import { render } from '@/tests/test-utils'
+import { faker } from '@faker-js/faker'
+import { PendingStatus, PendingTxType } from '@/store/pendingTxsSlice'
+import { extendedSafeInfoBuilder } from '@/tests/builders/safe'
+
+const safeAddress = faker.finance.ethereumAddress()
+
+describe('ClearPendingTxs', () => {
+  it('clear single transaction', () => {
+    render(<ClearPendingTxs />, {
+      initialReduxState: {
+        pendingTxs: {
+          ['0x123']: {
+            chainId: '1',
+            safeAddress,
+            nonce: 0,
+            data: faker.string.hexadecimal({ length: 64 }),
+            to: faker.finance.ethereumAddress(),
+            status: PendingStatus.PROCESSING,
+            txHash: faker.string.hexadecimal({ length: 64 }),
+            signerAddress: faker.finance.ethereumAddress(),
+            submittedAt: Date.now(),
+            signerNonce: 0,
+            txType: PendingTxType.CUSTOM_TX,
+          },
+        },
+        safeInfo: {
+          data: extendedSafeInfoBuilder()
+            .with({ address: { value: safeAddress } })
+            .with({ chainId: '1' })
+            .build(),
+          loading: false,
+        },
+      },
+    })
+    expect(screen.getByText('Clear 1 transaction')).toBeInTheDocument()
+    act(() => {
+      fireEvent.click(screen.getByText('Clear 1 transaction'))
+    })
+    expect(screen.getByText('No pending transactions')).toBeInTheDocument()
+  })
+  it('clear multiple transactions', () => {
+    render(<ClearPendingTxs />, {
+      initialReduxState: {
+        pendingTxs: {
+          ['0x123']: {
+            chainId: '1',
+            safeAddress,
+            nonce: 0,
+            data: faker.string.hexadecimal({ length: 64 }),
+            to: faker.finance.ethereumAddress(),
+            status: PendingStatus.PROCESSING,
+            txHash: faker.string.hexadecimal({ length: 64 }),
+            signerAddress: faker.finance.ethereumAddress(),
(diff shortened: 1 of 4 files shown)


katspaugh:
Wouldn't it be better to show this button directly in the Queue?

schmanu:
> Wouldn't it be better to show this button directly in the Queue?

@katspaugh  I wanted to keep it semi-hidden for now as it should usually not be necessary to do this if there is no problem on bug on our end. If we realize that it is very frequently required, we should ideally fix the root cause than making this button more prominent. WDYT?

I will add a new analytics event though to see how often this is used.

apps/web/src/components/settings/ClearPendingTxs/index.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { ClearPendingTxs } from '../ClearPendingTxs'
+import { render } from '@/tests/test-utils'
+import { faker } from '@faker-js/faker'
+import { PendingStatus, PendingTxType } from '@/store/pendingTxsSlice'
+import { extendedSafeInfoBuilder } from '@/tests/builders/safe'
+
+const safeAddress = faker.finance.ethereumAddress()
+
+describe('ClearPendingTxs', () => {
+  it('clear single transaction', () => {
```
iamacook, line 12:
```suggestion
  it('clears a single transaction', () => {
```

apps/web/src/components/settings/ClearPendingTxs/index.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+        },
+      },
+    })
+    expect(screen.getByText('Clear 1 transaction')).toBeInTheDocument()
+    act(() => {
+      fireEvent.click(screen.getByText('Clear 1 transaction'))
+    })
+    expect(screen.getByText('No pending transactions')).toBeInTheDocument()
+  })
+  it('clear multiple transactions', () => {
```
iamacook, line 45:
```suggestion
  it('clears multiple transactions', () => {
```

apps/web/src/components/settings/ClearPendingTxs/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const ClearPendingTxs = () => {
+  const pendingTxIds = usePendingTxIds()
+  const pendingTxCount = pendingTxIds.length
+  const dispatch = useAppDispatch()
+
+  const clearPendingTxs = useCallback(() => {
+    pendingTxIds.forEach((txId) => {
+      dispatch(clearPendingTx({ txId }))
+    })
+    trackEvent(SETTINGS_EVENTS.DATA.CLEAR_PENDING_TXS)
```
iamacook, line 17: What do you think about including the number? Afaiu, we are also aware of the chain so we can isolate troublesome chains easier.

apps/web/src/components/settings/ClearPendingTxs/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }, [dispatch, pendingTxIds])
+  return (
+    <Stack spacing={2}>
+      <Typography>Clear this Safe&apos;s pending transactions.</Typography>
```
iamacook, line 21:
```suggestion
      <Typography>Clear this Safe Account&apos;s pending transactions.</Typography>
```

apps/web/src/components/settings/ClearPendingTxs/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <Alert severity="warning">
+        <Typography>
+          This action does not delete any transactions but only resets their local state. It does not stop any pending
+          transactions from executing. If you want to cancel a execution you have to do this in your connected wallet.
```
iamacook, line 25:
```suggestion
          transactions from executing. If you want to cancel an execution, you have to do so in your connected wallet.
```

apps/web/src/components/settings/ClearPendingTxs/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+      </Alert>
+      <Box>
+        {pendingTxCount > 0 ? (
+          <Button
+            variant="text"
+            color="error"
+            onClick={clearPendingTxs}
+            sx={{ backgroundColor: ({ palette }) => palette.error.background }}
+          >
+            Clear {pendingTxCount} transaction{pendingTxCount > 1 ? 's' : ''}
```
iamacook, line 36:
```suggestion
            Clear {pendingTxCount} transaction{maybePlural(pendingTxCount)}
```

iamacook (approved):

PR #5330, opened 2025-03-14
Fix(Orgs): Handle unauthorized users in org views

### LINKED ISSUE #5329: [Orgs] Handle unauthorized users
## Links

## What must be done

- When a user is on an org page and signed in but not authorized to view this org we should display a message and a link back to the overview page

## How to test it

1. Open an org
2. Disconnect your wallet
3. Sign in with a different wallet that is not a member of this org
4. Observe a message and a button that brings you back to the overview page

usame-algan:
## What it solves

Resolves #5329

## How this PR fixes it

- Adds a `isUnauthorized` helper method
- Adds a `UnauthorizedState` component
- Handles errors returned from RTK Queries

## How to test it

1. Open an org
2. Disconnect your wallet
3. Connect with a wallet thats not a member of this org
4. Navigate through each page
5. Observe a message and a button that brings you back to the overview page

## Screenshots

Screenshot 2025-03-14 at 10 42 16

## Checklist

### DIFF
--- a/apps/web/src/features/organizations/components/Dashboard/index.tsx
+++ b/apps/web/src/features/organizations/components/Dashboard/index.tsx
@@ -17,6 +17,8 @@ import ChevronRightIcon from '@mui/icons-material/ChevronRight'
 import DashboardMembersList from '@/features/organizations/components/Dashboard/DashboardMembersList'
 import { useOrgMembers } from '@/features/organizations/hooks/useOrgMembers'
 import SignedOutState from '@/features/organizations/components/SignedOutState'
+import { isUnauthorized } from '@/features/organizations/utils'
+import UnauthorizedState from '@/features/organizations/components/UnauthorizedState'
 
 const ViewAllLink = ({ url }: { url: LinkProps['href'] }) => {
   return (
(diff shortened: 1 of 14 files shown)


apps/web/src/features/organizations/components/Dashboard/index.tsx
```diff
@@ -40,17 +42,17 @@ const ViewAllLink = ({ url }: { url: LinkProps['href'] }) => {
...
 
   const safesToDisplay = safes.slice(0, DASHBOARD_LIST_DISPLAY_LIMIT)
   const membersToDisplay = activeMembers.slice(0, DASHBOARD_LIST_DISPLAY_LIMIT)
 
-  if (!isUserSignedIn) {
-    return <SignedOutState />
-  }
+  if (!isUserSignedIn) return <SignedOutState />
+
+  if (isUnauthorized(error)) return <UnauthorizedState />
```
usame-algan, line 55: We do this in every page so I was thinking to extract them into a higher order component but the error comes from different RTK queries so its hard to pull this up but it might be ok if RTK manages to handle the caching.

usame-algan:
> If you view the org as an invited member you get the "nothing to see here" component on the dashboard and safes list (members and settings are ok)

I think this has to do with the issue that [I reported](https://5afe.slack.com/archives/C088LF9850E/p1741872187750229). Nothing we can do about it right now in the Frontend.

apps/web/src/features/organizations/components/Dashboard/index.tsx
usame-algan: Extracted the two conditions into a new component called `AuthState`

jmealy (approved):

PR #5331, opened 2025-03-14
Feat(mobile): transaction actions flow

clovisdasilvaneto:
## What it solves
It create all the missing screens for the transaction actions.

## How to test it
- Go to any pending tx that has actions
- Click in the actions section
- You should be redirected to the actions list and inside of it, click in any of the item
- You should be redirected to the action details

## Video

https://github.com/user-attachments/assets/57bf4364-f2a1-406d-9d79-b071c094baeb

## Checklist

### DIFF
--- a/apps/mobile/src/components/EncodedData/EncodedData.tsx
+++ b/apps/mobile/src/components/EncodedData/EncodedData.tsx
@@ -1,0 +1,23 @@
+import React, { useReducer } from 'react'
+import { Text } from 'tamagui'
+import { TouchableOpacity } from 'react-native'
+
+interface EncodedDataProps {
+  data: string
+}
+
+export function EncodedData({ data }: EncodedDataProps) {
+  const [truncated, toggleTruncate] = useReducer((state: boolean) => !state, true)
+
+  return (
+    <>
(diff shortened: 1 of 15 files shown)


apps/mobile/src/components/EncodedData/EncodedData.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { useReducer } from 'react'
+import { Text } from 'tamagui'
+import { TouchableOpacity } from 'react-native'
+
+interface EncodedDataProps {
+  data: string
+}
+
+export function EncodedData({ data }: EncodedDataProps) {
+  const [truncated, toggleTruncate] = useReducer((state: boolean) => !state, true)
```
compojoom, line 10: just curious - why did you go for useReducer instead of useState?

compojoom (approved):
👍

apps/mobile/src/components/EncodedData/EncodedData.tsx
clovisdasilvaneto: Since the state is a toggle, `useReducer` is a better approach for that.

PR #5360, opened 2025-03-17
Fix(Orgs): Optimize orgs views for mobile screens

### LINKED ISSUE #5358: [Orgs]Mobile view: invite and top tabs are not fit the screen
Steps:
1. send invite to the EOA 1
2. sign in as EOA 1 and go to the organizations page in the mobile view ( iPhone 11 in the dev tool)

Current result:

https://github.com/user-attachments/assets/692d2e11-383a-4977-b394-e6e1af7152d2

usame-algan:
## What it solves

Resolves #5358

## How this PR fixes it

- Adds `flex-wrap: wrap` for certain components so they break under each other on smaller screens
- Moves `invitedBy` logic to `InviteBanner` and removes unused `DashboardInvite` component

## How to test it

1. Open the orgs list on mobile viewport
2. Observe everything is visible
3. Go to an org and navigate through each page
4. Observe all elements are visible and an appropriate size

## Screenshots

Screenshot 2025-03-17 at 10 08 22
Screenshot 2025-03-17 at 10 08 35
Screenshot 2025-03-17 at 10 09 16

## Checklist


apps/web/src/features/organizations/components/SafeAccounts/index.tsx
```diff
@@ -25,7 +25,7 @@ const OrganizationSafeAccounts = () => {
       <Typography variant="h1" mb={3}>
         Safe Accounts
       </Typography>
-      <Stack direction="row" justifyContent="space-between" alignItems="center" spacing={2} mb={3}>
+      <Stack direction="row" justifyContent="space-between" alignItems="center" gap={2} mb={3} flexWrap="wrap">
         <TextField
```
jmealy, line 29: To me it feels more natural to have the add button before the search input. Otherwise when searching you have the button in between the input and the results:
image

jmealy (approved):
💯

apps/web/src/features/organizations/components/Members/index.tsx
```diff
@@ -26,7 +26,15 @@ const OrganizationMembers = () => {
...
       </Typography>
-      <Stack direction="row" justifyContent="space-between" alignItems="center" mb={3} flexWrap="wrap" gap={2}>
+      <Stack
+        direction="row"
+        justifyContent="space-between"
+        alignItems="flex-start"
+        mb={3}
+        flexWrap="wrap"
+        gap={2}
+        flexDirection={{ xs: 'column-reverse', md: 'row' }}
```
jmealy, line 36: is the `direction="row"` still required with this addition?

usame-algan: Unfortunately we need to set it "back" to the default value of row if we override it on a lower viewport because MUI is applying up breakpoints

PR #5361, opened 2025-03-17
feat(Signing UX): adds receipt step to NftTransferFlow + RemoveGuardFlow

tmjssz:
## What it solves

Adds the new transaction receipt step to the following flows:
* NftTransferFlow
* RemoveGuardFlow

**Note**: I haven't been able to test it yet, because I'm missing a Safe that has the conditions for both flows (holds an NFT or has a guard defined).


katspaugh:
Screenshot 2025-03-18 at 07 17 31

Sending NFTs works 👍

katspaugh (approved):

katspaugh:
@tmjssz approved, but there's a jest snapshot failing.

tmjssz:
> @tmjssz approved, but there's a jest snapshot failing.

Rebased onto epic branch, so they should pass this time

PR #5362, opened 2025-03-17
fix(mobile): Use a polyfill for Intl class and use currencyFormat

compojoom:
## What it solves
Hermes doesn't support all features of the web Intl class: https://github.com/facebook/hermes/issues/1035
After some research I think that the correct thing to do would be to polyfill those missing features.

## How this PR fixes it
Use polyfill from formatjs


apps/mobile/src/config/polyfills.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Hermes's implementation of Intl is not up to date and is missing some features. We rely mainly on `notation: "compact"`
+ * to properly format currency values, but it is not supported in Hermes.
+ * https://github.com/facebook/hermes/blob/5911e8180796d3ccb2669237ca441da717ad00b2/doc/IntlAPIs.md#limited-ios-property-support
+ *
+ * We need to polyfill some Intl APIs to make it work, and it seems that formatjs is doing the trick for us
+ */
+
+// Don't remove -force from these because detection is VERY slow on low-end Android.
+// https://github.com/formatjs/formatjs/issues/4463#issuecomment-2176070577
```
clovisdasilvaneto, line 10: interesting

apps/mobile/src/features/Assets/components/Balance/Balance.tsx
```diff
@@ -38,7 +38,7 @@ export function Balance({ activeChainId, chains, isLoading, balanceAmount, chain
         {isLoading ? (
           <Spinner />
         ) : balanceAmount ? (
-          <Fiat value={balanceAmount} currency="usd" />
+          <Fiat value={balanceAmount} currency="usd" precise={balanceAmount.length < 6} />
```
clovisdasilvaneto, line 41: nice

apps/mobile/src/features/Assets/components/Tokens/Tokens.container.tsx
```diff
@@ -36,17 +36,17 @@ export function TokensContainer() {
       <AssetsCard
         name={item.tokenInfo.name}
         logoUri={item.tokenInfo.logoUri}
-        description={`${formatValue(item.balance, item.tokenInfo.decimals as number)} ${item.tokenInfo.symbol}`}
+        description={`${formatVisualAmount(item.balance, item.tokenInfo.decimals as number)} ${item.tokenInfo.symbol}`}
         rightNode={
           <Text fontSize="$4" fontWeight={400} color="$color">
-            ${item.fiatBalance}
+            {formatCurrency(item.fiatBalance, 'usd')}
```
clovisdasilvaneto, line 42: can't we say that the `formatCurrenty` function would take `usd` by default?

clovisdasilvaneto (approved):

apps/mobile/src/features/Assets/components/Tokens/Tokens.container.tsx
compojoom: We currently only have USD, but we have a Settings screen where the user will be able to select the currency. So it makes no sense to enforce the default in the formatCurrency function as we will have to always get the currency selected in the settings.

PR #5368, opened 2025-03-17
Fix(Orgs): add declined chip back to invites list

jmealy:
## What it solves
The chip was accidentally removed during a refactor in a previous PR

## How this PR fixes it
Adds the chip back in when a member in the invites list has declined the invite.

## How to test it
- Invite a user to an org
- decline the invite
- from the inviter wallet, view the pending invites list

## Checklist


usame-algan (approved):
Thanks!

PR #5372, opened 2025-03-17
Fix(mobile): allow user to sign the conflicting transaction when it is pressed

clovisdasilvaneto:
## What it solves
It allows users to sign a conflicting transaction in the pending tx list component.

## How this PR fixes it
In. the `onPress` function of the pending transaction list item component, we just need to check if the transaction exists or not, if it exists, we redirect the user to the confirmation page, if it doesn't we show the bottom sheet.

## Checklist


app.json
```diff
@@ -0,0 +1,3 @@
+{
+  "expo": {}
+}
```
compojoom, line 3: that was most probably added because of running expo in the root of the monorepo.

clovisdasilvaneto: nooo

Issue #5371, opened 2025-03-17
[Orgs] Limit members to the EOA


jmealy:
Steps:
1. Open Add member
2. add any valid address ( EOA, safe, token20 and etc)
Current result: any address can be added without prefix validation for safes and any address can be added
Expected result: Limit to EOA address, remove prefix and add validation for the EOA address
https://5afe.slack.com/archives/C07RWQXH9KL/p1741872511067949

usame-algan:
We are not limiting members to EOAs for now. ([Context](https://github.com/safe-global/safe-wallet-monorepo/pull/5375#discussion_r1999085238))

PR #5375, opened 2025-03-17
Feat(Orgs): do not show prefix for address input in  members invite

jmealy:
## What it solves

Resolves [#5371](https://github.com/safe-global/safe-wallet-monorepo/issues/5371)

## How this PR fixes it
- Adds a prop to the AddressInput component for not showing the chain prefix
- Removes the chain prefix in AddressInput
- ~Adds validation to check if the address is an EOA. If it isn't an EOA show a validation error.~

## How to test it
- open the add member modal
- see that the chain prefix is no longer visible.
- ~enter a safe address or any other contract address. See the validation error~
- ~Enter an EOA address. See that no validation error is shown and the form can be submitted.~

## Checklist


apps/web/src/components/common/AddressInput/index.tsx
```diff
@@ -78,6 +82,20 @@ const AddressInput = ({
   // Validation function based on the current chain prefix
   const validatePrefixed = useMemo(() => validatePrefixedAddress(currentShortName), [currentShortName])
 
+  const validateEOA = useCallback(
+    async (address: string): Promise<string | undefined> => {
+      if (!isEOAOnly) return undefined
+
+      try {
+        const isContract = await isSmartContract(address)
```
usame-algan, line 90: This only works if its a smart contract on the chain as either the wallet provider (if passed as a parameter) or the read-only provider otherwise the validation doesn't catch it.

I was thinking we could check if its a safe contract via `getSafeInfo` but we need to provide a chainId there as well.

So I am not sure how else to check that the address is an EOA but maybe its not that big of an issue and we don't need to validate this. If someone adds an address there thats not an EOA then that address must do SiWE in order to accept/decline the invite which might not be possible. So worst case is that the admin has to remove that invite at some point and even if they don't it doesn't do any harm.

jmealy: removed the EOA validation

francovenica:
The prefix was removed and that's correct, but we still make the validation of prefix (if the address was copied with one)

1 - Be connected to sepolia
2 - Copy/paste an address from eth:
3 - See the error

image

Issue #5275, opened 2025-03-11
[Orgs] Add member form: 2 fields are red by default on form opening


liliya-soroka:
## Bug description
Add member form: 2 fields are red by default on form opening 
## Environment

- Browser: Chrome
- Wallet: any
- Chain: any

## Steps to reproduce

1.  Go to the organization-> Members
2. click Add Member
Current result: add member form is opened and 2 fields are red by default

## Expected result
The fields should be validated only on the wrong data entering and not by default on opening 
## Screenshots

<img width="672" alt="Image" src="https://github.com/user-attachments/assets/7f082e67-126b-4791-ad83-90e2489b9602" />

jmealy:
related: https://github.com/safe-global/safe-wallet-monorepo/pull/5305

liliya-soroka:
Verified

PR #5376, opened 2025-03-17
Fix(Orgs): re-trigger validation of safe address when chain id changes

jmealy:
## What it solves

Resolves https://github.com/safe-global/safe-wallet-monorepo/issues/5275#issuecomment-2724543297

## How this PR fixes it
- re triggers validation of the address input in the add account modal when the selected chain is changed.

## How to test it
- go to the add address modal. 
- Click add manually
- paste the address of a safe that exists on the currently selected chain. No validation error should be shown and the submit button enabled
- change the chain. A validation error should show and the submit should be disabled.

## Checklist

### DIFF
--- a/apps/web/src/features/organizations/components/AddAccounts/AddManually.tsx
+++ b/apps/web/src/features/organizations/components/AddAccounts/AddManually.tsx
@@ -78,7 +78,7 @@ const AddManually = ({ handleAddSafe }: { handleAddSafe: (data: AddManuallyFormV
     if (address) {
       trigger('address')
     }
-  }, [chainId, trigger, watch])
+  }, [address, chainId, trigger])
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


apps/web/src/features/organizations/components/AddAccounts/AddManually.tsx
```diff
@@ -72,6 +73,13 @@ const AddManually = ({ handleAddSafe }: { handleAddSafe: (data: AddManuallyFormV
 
   const chainIdField = register('chainId')
 
+  // Re-validate when chainId changes
+  useEffect(() => {
+    if (address) {
+      trigger('address')
+    }
+  }, [chainId, trigger, watch])
```
usame-algan, line 81: Is `watch` needed here?

PR #5377, opened 2025-03-17
feat(web): add module for Ledger devices

iamacook:
## What it solves

Ledger device support

## How this PR fixes it

This adds a proprietary Web3-Onboard module for Ledger devices to facilitate direct transacting and signin.

In comparison to the "official" module, it supports `eth_signTypedData`, meaning 712 structs are shown instead of hashes for clearer signing.

It is based on our [initial implementation](https://github.com/safe-global/safe-wallet-monorepo/pull/4416), using the latest version of the [Web Device Management Kit](https://github.com/LedgerHQ/device-sdk-ts).

## How to test it

As this is a wallet module, we should test interactions according to RPC methods, across various Ledger devices. The following actions cover supported RPC methods:

- Successful connection (accounts)
- Signing a message (signing)
- Signing a transaction (signing typed data)
- Executing a transaction/creating a Safe (sending a transaction)

Outside of "standard" transacting, we should confirm other areas that require wallet signatures work: notification and proposer (un-)registration.

Note: the module should be enabled/disabled via the `ledger` key.

## Checklist


apps/web/src/components/settings/PushNotifications/logic.ts
```diff
@@ -34,35 +32,18 @@ export const requestNotificationPermission = async (): Promise<boolean> => {
...
-export const _adjustLedgerSignatureV = (signature: string): string => {
-  const split = splitSignature(signature)
-
-  // @ts-ignore
-  if (split.v === 0 || split.v === 1) {
-    split.v += 27
-  }
-
-  return joinSignature(split)
-}
```
iamacook, line 50: The correct `v` is now returned.

apps/web/src/hooks/wallets/consts.ts
```diff
@@ -16,7 +14,6 @@ export const CGW_NAMES: { [_key in WALLET_KEYS]: string | undefined } = {
   [WALLET_KEYS.WALLETCONNECT_V2]: 'walletConnect_v2',
   [WALLET_KEYS.COINBASE]: 'coinbase',
   [WALLET_KEYS.LEDGER]: 'ledger',
```
schmanu, line 16: Should Ledger (v1) be removed?

apps/web/src/services/onboard/ledger-module.ts
```diff
@@ -0,0 +1,388 @@
...
+            eth_signTransaction: async (args) => {
+              const txParams = args.params[0]
+
+              const gasLimit = txParams.gas ?? txParams.gasLimit
+              const nonce =
+                txParams.nonce ??
+                // Safe creation does not provide nonce
+                ((await eip1193Provider.request({
+                  method: 'eth_getTransactionCount',
+                  params: [currentAccount!.address, 'latest'],
```
schmanu, line 144: Would `pending` be better here in case you still got pending transactions?
I think `latest` only considers already mined blocks.

```suggestion
                  params: [currentAccount!.address, 'pending'],
```

apps/web/src/hooks/wallets/consts.ts
iamacook: I removed all reference and logic related to the previous implementation as this replaces it. Do you think we should keep it instead?

apps/web/src/services/onboard/ledger-module.ts
iamacook: Updated in 9f6c12e9509068ac6be90ee199fb54de3f5ba022.

633kh4ck:
Could you elaborate on the following, please?
- What was the reason for https://github.com/safe-global/safe-wallet-monorepo/pull/5091?
- What's the ETA for this?

apps/web/src/services/onboard/ledger-module.ts
```diff
@@ -0,0 +1,389 @@
...
+                // Safe signs bytes
+                getBytes(message),
+              )
+              return Signature.from(signature).serialized
+            },
+            personal_sign: async (args) => {
+              // personal_sign params are the inverse of eth_sign
+              const [message, address] = args.params
+              return await eip1193Provider.request({
+                method: 'eth_sign',
```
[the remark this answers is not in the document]
iamacook: The method is named `personal_sign` because that's the expected interface, but it intentionally calls `eth_sign` with reversed `params`.

`personal_sign` applies a prefix and expects the `params` in reverse order compared to `eth_sign`, which is why we're deliberately calling `eth_sign` here instead of defining a separate method.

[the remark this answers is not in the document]
tmjssz: Gotcha! Thanks for the explanation.

tmjssz (approved):

schmanu (approved):

mbertin-ledger (commented):
LGTM 👍

katspaugh (approved):

PR #5383, opened 2025-03-18
Release 1.53.1

usame-algan:
## Fixes

- Fix(web): remove warning banner and popup (#5384)
- Revert "fix(Safe Apps): keep appUrl query param when switching network (#5081)" (#5327)
- Fix(Legal): update CC address (#5380)
- fix(GA): Add unsafe-inline to script-src CSP (#5339)
- fix: Restore sticky banner for tx verification


liliya-soroka:
1. Wrong amount  validation we don't have on prod 
image

PR #5385, opened 2025-03-18
feat(Signing UX): tracking

### LINKED ISSUE #5283: [Signing UX] Tracking
## Links

[Epic on Notion](https://www.notion.so/safe-global/Signing-UX-security-improvements-1ab8180fe573806aa382f7dddb3c7d49?pvs=4)

## What must be done

Send analytics events when:
* User checks the checkbox "I understand what I will sign ... "
* User goes Back to the previous screen
* User clicks on external link to the article

tmjssz:
## What it solves

Resolves #5283 

Add tracking for the following events:
* User checks the checkbox "I understand what I will sign ... "
* User clicks on external link to the article about how to perform basic transaction checks
* User navigates back or next on any transaction flow

### DIFF
--- a/apps/web/src/services/analytics/events/modals.ts
+++ b/apps/web/src/services/analytics/events/modals.ts
@@ -100,6 +100,11 @@ export const MODALS_EVENTS = {
     category: MODALS_CATEGORY,
     event: EventType.CLICK,
   },
+  CONFIRM_SIGN_CHECKBOX: {
+    action: 'Confirm sign checkbox',
+    category: MODALS_CATEGORY,
+    event: EventType.CLICK,
+  },
 }
 
 export enum MODAL_NAVIGATION {
--- a/apps/web/src/services/analytics/types.ts
+++ b/apps/web/src/services/analytics/types.ts
@@ -44,8 +44,24 @@ export enum AnalyticsUserProperties {
 
 // These are used for the generic stepper flow events (Next, Back)
 export enum EventCategory {
-  SETUP_RECOVERY = 'setup-recovery',
-  START_RECOVERY = 'propose-recovery',
-  REMOVE_RECOVERY = 'remove-recovery',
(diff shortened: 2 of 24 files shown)


apps/web/src/services/analytics/events/modals.ts
```diff
@@ -96,6 +96,16 @@ export const MODALS_EVENTS = {
     category: MODALS_CATEGORY,
     event: EventType.CLICK,
   },
+  CONFIRM_SIGN_CHECKBOX: {
+    action: 'Confirm sign checkbox',
+    category: MODALS_CATEGORY,
+    event: EventType.CLICK,
+  },
+  OPEN_TX_CHECK_ARTICLE_LINK: {
+    action: 'Open the basic transactions checks article link',
```
katspaugh, line 105: That's too long for an action name. Also I moved this link to another place in another PR, and it already has a tracking event there, so this can be removed.

apps/web/src/services/analytics/types.ts
```diff
@@ -41,3 +41,27 @@ export enum AnalyticsUserProperties {
   WALLET_LABEL = 'walletLabel',
   WALLET_ADDRESS = 'walletAddress',
 }
+
+// These are used for the generic stepper flow events (Next, Back)
+export enum EventCategory {
```
katspaugh, line 46: The name of the enum is too generic. TxFlowType maybe?

tmjssz: Good point! I originally chose a generic name thinking the enum might be useful for other events as well. But I'm totally fine with making it more specific, TxFlowType works for me if that makes things clearer.

apps/web/src/services/analytics/events/modals.ts
tmjssz: Ok will remove it

Issue #5055, opened 2025-02-20
[Orgs] Analytics


usame-algan:
## Links

Epic on Notion: https://www.notion.so/safe-global/Org-onboarding-flow-1208180fe57380078591f5380a0c5d2a

Analytics events: https://www.notion.so/safe-global/Analytics-events-1b98180fe573804bafa1c1a7e0f2b297

## What must be done

- Create a Notion page for analytics, which metrics to track and which GA events are needed in the frontend
- Implement the required events

## How to test it

1. Go through the list of events in Notion and check each event

francovenica:
Opening the Spaces from the tab in the welcome/account pages is not showing the event.
I see that there is a GA event, but it displays nothing so I cannot tell if it is the `open space list page` event.

The other 2 ways to create the event (From the safe dashboard and from inside a space with the "view spaces" link") work fine
![Image](https://github.com/user-attachments/assets/2057411b-4fbc-4789-a6b5-46403d273ffb)

francovenica:
Expanding the multichain safes in the account widget or the safe account's list triggers an event "expand" with an eventLabel: "Sidebar". I assume is because the component is the same from the sidebar, but it shouldn't trigger in this context.

![Image](https://github.com/user-attachments/assets/8e923202-99bb-43f3-b004-551f4ff11fff)

francovenica:
The delete event for spaces have issues:
1 - Minor: Is called "Open remove space modal" instead of "open delete space modal"
2 - The event "Open modal" doesn't trigger when used from the spaces list (the "..." menu
3 - The "Delete event" doesn't trigger at all, no matter where you use the delete (settings or the spaces list)

francovenica:
Also I added notes to the document. Event's that are not wrong, but some missing eventLabels or and some events that were not listed
https://www.notion.so/safe-global/Analytics-events-1b98180fe573804bafa1c1a7e0f2b297

francovenica:
Looks good now.
The events were fixed, the comments in the document were addressed and the new event triggered by using the breadcrum also was checked here.

PR #5392, opened 2025-03-18
Feat(Orgs): add analytics events

jmealy:
## What it solves

Resolves [#5055](https://github.com/safe-global/safe-wallet-monorepo/issues/5055)
https://www.notion.so/safe-global/Orgs-analytics-events-1b98180fe573804bafa1c1a7e0f2b297

## How this PR fixes it
- adds events for important actions in the orgs feature 

## How to test it
- make sure all the listed events are succesfully sending when the action happens.

## Checklist


apps/web/src/features/myAccounts/components/AccountsNavigation/index.tsx
```diff
@@ -23,13 +25,15 @@ const AccountsNavigation = () => {
...
 
       <Typography variant="h1" fontWeight={700} className={css.title}>
-        <Link
-          href={AppRoutes.welcome.organizations}
-          className={classNames(css.link, { [css.active]: isActiveNavigation(AppRoutes.welcome.organizations) })}
-        >
-          Organizations
-          <Chip label="Beta" size="small" sx={{ ml: 1, fontWeight: 'normal', borderRadius: '4px' }} />
-        </Link>
+        <Track {...ORG_EVENTS.OPEN_ORGS_LIST_PAGE} label={ORG_LABELS.accounts_page}>
```
usame-algan, line 28: In my tests this event was not emited when clicking the button but it might also be because the app seems to do a "hard" refresh when clicking this link and so it disappears after being emitted.

apps/web/src/features/myAccounts/components/FilteredSafes/index.tsx
```diff
@@ -15,6 +18,12 @@ const FilteredSafes = ({
 }) => {
   const filteredSafes = useSafesSearch(allSafes ?? [], searchQuery)
 
+  useEffect(() => {
+    if (searchQuery) {
+      trackEvent({ category: OVERVIEW_EVENTS.SEARCH.category, action: OVERVIEW_EVENTS.SEARCH.action })
```
usame-algan, line 23: Not critical but this event is tracked as a "customClick" event instead of a meta event.

usame-algan (approved):
Nice one! I tested each event and apart from this one issue with the hard reload it looks good 🚀 

Wdyt about adding a few more events to make it more granular?

Split ADD_ACCOUNT_MANUALLY:
- Open add account address manually
- Submit add account address manually
---
- Delete safe account
- Rename org
- Open delete org modal

apps/web/src/features/myAccounts/components/AccountsNavigation/index.tsx
jmealy: From debugging it seems like it does get emitted (or would do in production), but the console log gets cleared when we redirect. So I think this is ok

PR #5393, opened 2025-03-18
feat(mobile): add signer view

compojoom:
## What it solves
Ads the signer detail view

Resolves #
https://github.com/orgs/safe-global/projects/1/views/1?filterQuery=signer&pane=issue&itemId=102700913

## How this PR fixes it
You can navigate to this view both from the safe's settings or from the import safe flow. The implementation diverges from the figma, becasue we currently don't support "backup, view private key, view seedphrase or remove signer"

## Checklist


apps/mobile/src/components/SafeFontIcon/SafeFontIcon.tsx
```diff
@@ -4,7 +4,7 @@ import { useFonts } from 'expo-font'
 import { IconName } from '@/src/types/iconTypes'
 import { getVariable, useTheme } from 'tamagui'
 
-export const SafeIcon = createIconSetFromIcoMoon(
+const SafeIcon = createIconSetFromIcoMoon(
```
compojoom, line 7: It is not necessary to export this. It's only needed here and often I was erroniously importing it.

PR #5394, opened 2025-03-19
Fix(Signing UX): adjust tx data parameter design

katspaugh:
## What it solves

There were some design inconsistencies on NFT review screen and in Advanced details.
Also I removed an unused component.

Screenshot 2025-03-19 at 08 53 11


tmjssz:
Looks good! Just a small thing: the NFT address in the screenshot is overflowing.

francovenica:
From the QA side it looks good as well in the EPIC

PR #5395, opened 2025-03-19
fix(Spending limits): rename disabled spending limit text

iamacook:
## What it solves

Resolves confusing text

## How this PR fixes it

When the spending limits _feature_ is disabled, but a module remains activated, a message and module are shown. This is for security reasons: allowing an owner to disable the module:

image

However, the message is unclear because the _module_ is available but it states that it isn't. As such, this renames _module_ to _feature_.

## How to test it

1. Ensuring the spending limits feature is disabled, observe the new message.

## Checklist


schmanu (approved):
🫡

Issue #5400, opened 2025-03-19
disable dashboard buttons for non admins


jmealy:
## What must be done
Non admins can't add members or accounts, so they should not be shown these buttons. Disable the buttons on the dashboard with a tooltip to explain

## Designs/sketches
https://www.figma.com/design/PeChIDi5UKPjf58IOChtET/Organisations?node-id=4413-34765&t=IZwFHUzfeGPGpnna-1

## How to test it
- view the dashboard page as a non admin (normal member or invited member)
- see that the add member and accounts buttons are disabled with a tootlip to explain why

jmealy:
Moving back, found another button

francovenica:
LGTM

The buttons are disabled, they show a tooltip indicating that only admins can use those actions.

PR #5401, opened 2025-03-19
Fix(Orgs): disable buttons on dashboard for non admins

jmealy:
## What it solves

Resolves [#5400](https://github.com/safe-global/safe-wallet-monorepo/issues/5400)

## How this PR fixes it
- diables the buttons for non admins
- adds tooltips to explain why

## How to test it
view the dashboard page as a non admin (normal member or invited member)
see that the add member and accounts buttons are disabled with a tootlip to explain why

## Screenshots
image

## Checklist

### DIFF
--- a/apps/web/src/features/organizations/components/AddAccounts/index.tsx
+++ b/apps/web/src/features/organizations/components/AddAccounts/index.tsx
@@ -123,7 +123,7 @@ const AddAccounts = () => {
 
   return (
     <>
-      <Tooltip title={!isAdmin ? 'You need an Admin role to add accounts' : ''} placement="top">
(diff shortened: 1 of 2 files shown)


usame-algan (approved):
nit: We hide those buttons on other pages instead of disabling them so the behaviour is a bit inconsistent now but we can iterate on it.

jmealy:
> nit: We hide those buttons on other pages instead of disabling them so the behaviour is a bit inconsistent now but we can iterate on it.

True, personally I would not show these cards at all for non admins since they are telling you to add accounts and members, which you can't actually do. But we would need to show something here instead. Let's come back to it

Issue #5240, opened 2025-03-10
[Signing UX] Additional transaction details review screen


tmjssz:
## Links

[Epic on Notion](https://www.notion.so/safe-global/Signing-UX-security-improvements-1ab8180fe573806aa382f7dddb3c7d49?pvs=4)

## What must be done

Add an additional screen that appears after the current transaction confirmation screen and before the transaction can be signed. The new screen shows all transaction data that will be signed in an overview, mimicking the way it is shown by the connected signing wallet, to allow easy validation of all the data that will be signed.

* Show all the data that will be signed like in the designs
* The transaction details "receipt" component view can be switched to show the raw JSON data as well by using the toggle button
* Add a checkbox that must be checked for the Sign button to be clickable
* Nonce can not be edited in this step anymore

## Designs/sketches

[Designs](https://www.figma.com/design/cMDccuR9aZFqAiBd4dKsKg/Transaction-flow?node-id=1091-7&p=f&t=Raunnkog5vhJCJvi-0)

## How to test it

1. Create a transaction and go to the review step (that is currently the last one before signing)
2. Observe that the button is now "Continue" instead of "Sign"
3. Click that button and observe the transaction details confirmation step
4. Sign button is disabled as long as the checkbox is unchecked
5. The nonce can not be edited anymore
6. When clicking the Sign button the signing process with the connected wallet is triggered
7. The wallet data is presented in a similar way as in our confirmation step

francovenica:
This ticket is fine. The elements are there and it looks good

The check of every type of tx is going to be done on the 5303

francovenica:
Putting this back because of a new issue:

In the screen where you can see the "Continue" button, that button should be disabled until the user is in the correct network:
![Image](https://github.com/user-attachments/assets/d0938cf7-489e-49a1-ba8d-53b8f4f95993)

katspaugh:
Isn't the Sign button enabled in that case in prod? It should switch the chain automatically before signing.

francovenica:
@katspaugh No, Current prod looks like this:

![Image](https://github.com/user-attachments/assets/2a2d2d8a-9eea-44b4-a5d6-cf7c8de739ca)

tmjssz:
#5402 disables the continue button if on wrong network.

francovenica:
The fix was tested and looks good

PR #5402, opened 2025-03-19
fix(ReviewTransaction): check wallet for continue button 

tmjssz:
Disable the continue button if wallet check fails.

Fixes [this bug](https://github.com/safe-global/safe-wallet-monorepo/issues/5240#issuecomment-2735581247).


katspaugh (approved):

francovenica:
LGTM

PR #5285, opened 2025-03-12
Feat(Tx flow): Signing UX improvements

katspaugh:
Resolves #5238

[Designs](https://www.figma.com/design/cMDccuR9aZFqAiBd4dKsKg/Transaction-flow?node-id=1091-7&p=f&t=Raunnkog5vhJCJvi-0)
[Epic on Notion](https://www.notion.so/safe-global/Signing-UX-security-improvements-1ab8180fe573806aa382f7dddb3c7d49?pvs=4)

PR #5404, opened 2025-03-19
Release 1.54.0

katspaugh:
* Feat(Tx flow): Signing UX improvements (#5285)


francovenica:
Whenever you open a the new tx form, it puts an black overlay on the screen. This overlay was shorten down a bit for the warning message we put in prod about "Always verify what you sign". 
In this RC that warning is not present so now there is a gap on the top.

Question is: is that warning message will also disappear from prod? if that's the case we need to stretch up the overlay again to fill that gap.
If the warning will still be in prod after this deployment then we can leave it as it is.

image
image

francovenica:
Beyond the previous comment the rest is fine

Automation was run, many test failed as expected because of the last step added to tx. Other test that also failed were verified manually

francovenica:
The fix looks good.

Approved

PR #5406, opened 2025-03-19
feat(Signing UX): color coded transaction details

tmjssz:
## What it solves

Resolves #5241 

Displays advanced details accordion in different colors, depending on the transaction method.

## Screenshots
Screenshot 2025-03-19 at 17 19 49
Screenshot 2025-03-19 at 15 10 11
Screenshot 2025-03-19 at 17 21 36

## Checklist

- [ ] I've tested the branch on mobile 📱
- [ ] I've documented how it affects the analytics (if at all) 📊
- [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #5421, opened 2025-03-20
Release 1.54.0

katspaugh:
* Feat(Tx flow): Signing UX improvements (#5285)
* feat(Signing UX): color coded transaction details (#5406)
* Fix(web): remove warning banner


katspaugh:
This is a re-creation of #5404 because I forgot to bump the version there, so the release action didn't kick in.

PR #5430, opened 2025-03-20
fix(CGW SDK): only include `credentials` for specific routes

iamacook:
## What it solves

Resolves addition of `credentials` to every request

## How this PR fixes it

The `dynamicBaseQuery` of the store now only includes `credentials` for the following routes:

- `/v1/auth/*`
- `/v2/register/notifications`
- `/v2/chains/:chainId/notifications/devices/*`

Note: these will need to be adjusted should we require `credentials` on other routes, or if we restrict domains for the Client Gateway.

## How to test it

Interaction with the app should be as expected, and mobile registration for notifications should still work.

## Checklist


katspaugh (approved):

iamacook:
I'm going to merge this directly to unblock development.

PR #5431, opened 2025-03-20
Fix(Orgs): Only list owned accounts when adding to space

### LINKED ISSUE #5425: [Orgs] Don't list counterfactual safes in the safe accounts list
## Links

## What must be done

- When adding safe accounts to a space there should be no counterfactual safes in that list

## How to test it

1. Open an org
2. Add safe accounts
3. Observe no counterfactual safes are in the list
4. Add a bunch of different Safes
5. Observe no jumping network icons

usame-algan:
## What it solves

Resolves #5425 

## How this PR fixes it

- Adds a new hook `useOwnedSafesGrouped` and uses that instead of `useAllSafesGrouped` within spaces

## How to test it

1. Create an undeployed safe
2. Observe it is listed under accounts
1. Open a space
3. Add safe accounts
4. Observe the undeployed safe is not showing up in the list

## Checklist

### DIFF
--- a/apps/web/src/features/multichain/utils/utils.ts
+++ b/apps/web/src/features/multichain/utils/utils.ts
@@ -30,6 +30,10 @@ export const isMultiChainSafeItem = (safe: SafeItem | MultiChainSafeItem): safe 
   return false
 }
 
+export const isSafeItem = (safe: SafeItem | MultiChainSafeItem): safe is SafeItem => {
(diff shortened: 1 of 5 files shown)


jmealy (approved):
Looks good! It would be nice also if we didn't show safes that have already been added in the list. Currently if you try to add a previously added safe if just gives a vague error. Can be done separately though

PR #5448, opened 2025-03-20
feat(mobile): Safes can have a name now

compojoom:
## What it solves
With this PR safes added to the mobile app can have names.

Resolves  https://github.com/safe-global/safe-wallet-mobile/issues/62

## How this PR fixes it
During account import there is a field to enter the safe name. You can change that Safe name later in the Safe's settings. 
During the import the safe is created as a contact in the address book. We look for that contact everywhere where we need to display the safe name. (we could later optimise this and make it a hook)

## How to test it
Look at the added video

## Screenshots

https://github.com/user-attachments/assets/95a97367-e0c8-4be6-9f50-2fd943e61216

## Checklist

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -129,6 +129,7 @@
     "@babel/preset-env": "^7.26.0",
     "@babel/preset-react": "^7.26.3",
     "@eslint/js": "^9.18.0",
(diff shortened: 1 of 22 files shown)


compojoom:
you are right I forgot to add a restriction to the zod validation.

Issue #5144, opened 2025-02-27
[Orgs] Safe list needs a skeleton while loading


francovenica:
Not necesarely a bug, but it would be nice if the safe list inside an org show a skeleton while it is loading.

Currently it shows the whole safelist the user owns and then it loads the safes that are part of the org

Gif: 
![Image](https://github.com/user-attachments/assets/3078a7e4-cc5b-40d1-b5d1-4a16c281f601)

usame-algan:
We removed the issue that unrelated safes are shown while loading. I think its fine without a skeleton because we can't know the number of safes someone has added before loading them.

PR #5449, opened 2025-03-20
Feat(Orgs): add loading state for dashboard and accounts page

jmealy:
## What it solves

Resolves [#5144](https://github.com/safe-global/safe-wallet-monorepo/issues/5144)
The empty and logged out states are rendering while some of the requests are loading causing the page to flicker on refresh.

## How this PR fixes it
Adds a loading spinner to 
- [x] accounts page
- [x] dashboard
- [ ] spaces list

## How to test it
- go to one of the pages listed above      
- refresh the page.
- see that a loading spinner is shown while the data loads instead of the empty or logged out state.

## Checklist


usame-algan (approved):
I have removed the LoadingState from the dashboard and accounts page and only kept it in the AuthState component since it already wraps all other pages. Looks good otherwise!

Note: There is no loading state on the space list page because it relies on the `isUserSignedIn` state that comes directly from redux and doesn't have a loading flag. I think its fine for now.

PR #5459, opened 2025-03-21
fix(Signing UX): allow non-owner execution in transaction review

tmjssz:
## What it solves

Allow to continue to the transaction receipt step if connected wallet is not an owner of the Safe.


katspaugh (approved):

francovenica:
LGTM

PR #5460, opened 2025-03-21
fix(History filters): don't parse/format filter values

iamacook:
## What it solves

Partially resolves https://github.com/safe-global/safe-client-gateway/issues/2340

## How this PR fixes it

We format/parse the `value` for incoming/outgoing filters because the Client Gateway requires "raw" values. When https://github.com/safe-global/safe-client-gateway/pull/2485 is merged, this value parsing will happen on the Client Gateway.

This removes the client-side format/parse accordingly.

## How to test it

Filter incoming/outgoing and observe the value "as is" in the request.

## Checklist


katspaugh:
Merging to fix smoke tests. Thanks!

PR #5465, opened 2025-03-21
fix(mobile): import signer crash

Jonathansoufer:
## What it solves

Signer import was crashing the app due the auto update on notification's permission right after an safe import.

Resolves #

## How this PR fixes it
We replaced the source of SafeInfo from Redux to glob directly. 

## Screenshots

https://github.com/user-attachments/assets/83aa9506-56f1-48b3-b9a9-94e1ad51f876

## Checklist

### DIFF
--- a/apps/mobile/src/hooks/useRegisterForNotifications.ts
+++ b/apps/mobile/src/hooks/useRegisterForNotifications.ts
@@ -55,10 +55,6 @@ const useRegisterForNotifications = (): NotificationsProps => {
 
   const glob = useGlobalSearchParams<{ safeAddress?: string; chainId?: string; import_safe?: string }>()
 
-  if (!glob) {
-    throw new Error('useRegisterForNotifications :: glob is undefined')
-  }
-
   if (!glob.safeAddress) {
     glob.safeAddress = activeSafe?.address
   }


apps/mobile/src/hooks/useRegisterForNotifications.ts
```diff
@@ -49,9 +50,27 @@ const useRegisterForNotifications = (): NotificationsProps => {
...
-  const activeSafe = useDefinedActiveSafe()
+  const activeSafe = useAppSelector(selectActiveSafe)
   const delegatedAccounts = useAppSelector(selectDelegatedAccounts)
-  const { ownerFound, accountType } = useNotificationGTWPermissions().getAccountType()
+
+  const glob = useGlobalSearchParams<{ safeAddress?: string; chainId?: string; import_safe?: string }>()
+
+  if (!glob) {
+    throw new Error('useRegisterForNotifications :: glob is undefined')
+  }
```
compojoom, line 60: Why is it necessary to throw here? Few lines later you set the glob safeAddress to whatever was in the activeSafe?

compojoom (approved):
Looks good. Tested it and it no longer crashes

PR #5469, opened 2025-03-21
(mobile):fix - add skeleton placeholder on main balance & fix UI on Settings account

Jonathansoufer:
## What it solves

[x] - Add skeleton placeholder on main balance;
[x] - Fix badge and containers for Account/Settings (Signers, threshold)

Resolves #

## Screenshots

Screenshot 2025-03-21 at 22 44 04
Screenshot 2025-03-21 at 22 43 55
Screenshot 2025-03-21 at 23 02 39
Screenshot 2025-03-21 at 23 02 27

## Checklist

### DIFF
--- a/apps/mobile/src/components/Fiat/Fiat.tsx
+++ b/apps/mobile/src/components/Fiat/Fiat.tsx
@@ -1,8 +1,6 @@
 import React, { useMemo } from 'react'
-import { Skeleton } from 'moti/skeleton'
 import { H1, H2, View, XStack } from 'tamagui'
 import { formatCurrency, formatCurrencyPrecise } from '@safe-global/utils/formatNumber'
-import { useColorScheme } from 'react-native'
 
 interface FiatProps {
   value: string
@@ -12,8 +10,6 @@ interface FiatProps {
 }
 
 export const Fiat = ({ value, currency, maxLength, precise }: FiatProps) => {
-  const colorScheme = useColorScheme()
-
   const fiat = useMemo(() => {
     return formatCurrency(value, currency, maxLength)
   }, [value, currency, maxLength])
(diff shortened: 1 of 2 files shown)


apps/mobile/src/features/Settings/Settings.tsx
```diff
@@ -74,15 +74,16 @@ export const Settings = ({ address, data, onImplementationTap, displayDevMenu, c
                       </Text>
                     </Skeleton>
                   </View>
-                  <Text color="$colorHover" fontSize={'$3'}>
+                  <Text color="$colorHover" fontSize={'$3'} fontWeight="400">
```
compojoom, line 77: nit: fontWeight is by default 400.

apps/mobile/src/components/Fiat/Fiat.tsx
```diff
@@ -23,25 +27,25 @@ export const Fiat = ({ value, currency, maxLength, precise }: FiatProps) => {
...
-          <H1 fontWeight="600">{whole}</H1>
-          {decimals && (
-            <H1 fontWeight={600} color="$textSecondaryDark">
-              {decimals}
-            </H1>
-          )}
-          <H1 fontWeight={600}>{endCurrency}</H1>
-        </XStack>
+        <Skeleton.Group show={!fiat}>
+          <Skeleton colorMode={colorScheme === 'dark' ? 'dark' : 'light'} width={220}>
```
compojoom, line 34: I don't understand when this skeleton kicks in? 

fiat always end up being a string, so the Skeleton will never kick in? I think that you should move the skeleton logic here:
https://github.com/safe-global/safe-wallet-monorepo/blob/2fcc6735a94a02cbe227eba5e2c25fb696f6f841/apps/mobile/src/features/Assets/components/Balance/Balance.tsx#L62-L68

Also when you move it there, then I'm not sure we need to display the alert "error while getting the balance of your wallet", since if you have the skeleton it would be clear that it's loading and we keep trying to load the balance every couple of seconds.

compojoom (approved):
👍

PR #5480, opened 2025-03-24
fix(test): failing Cypress pipeline 

tmjssz:
## What it solves

Cypress pipeline was failing with error:
> TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /home/runner/work/***-wallet-monorepo/***-wallet-monorepo/apps/web/src/markdown/terms/version.ts

Solve it by renaming version.ts to .js.


compojoom:
Cool, but it would actually be good to know why this happens. 
I suspect this PR https://github.com/safe-global/safe-wallet-monorepo/pull/5393/files#diff-2d5976b4b1723453d48fb571fafaa3f0408bcb4ceadd86570761104b370e2bc0L4 and the change in the tsconfig file fo the store package.

mike10ca (approved):

mike10ca:
Since we are blocked by this, I would suggest we merge it for a quick fix and go from there

PR #5484, opened 2025-03-24
fix(Nested Safes): only allow adding nested Safe for deployed Safes

### LINKED ISSUE #5075: Nested safes: Add Nested safe button should be disabled for non-active safes
## Bug description
Nested safes: Add Nested safe button should be disabled for non-active safes
## Environment

- Browser: Chrome
- Wallet: MetaMask
- Chain: any

## Steps to reproduce

1.  Go to the not activated safe when the "Nested safe" feature is on for the user through other whitelisted safe
2.  go to the settings->set up
Current result : the add nested safe is disabled with the description tooltip

3. go to the nested safe menu from the side bar
Current result: Add nested button is available and the process can be started but execute button is disabled 

https://github.com/user-attachments/assets/0982e5d5-6920-4cfe-81b5-ca38771fc71e

iamacook:
## What it solves

Resolves #5075

## How this PR fixes it

Only deployed Safes can add nested Safes. This hides the creation buttons (in the sidebar and settings) if the current Safe is not deployed.

## How to test it

In a counterfactual Safe, observe no "Add nested Safe" button in the sidebar popover or settings.

## Checklist

### DIFF
--- a/apps/web/src/components/settings/NestedSafesList/index.tsx
+++ b/apps/web/src/components/settings/NestedSafesList/index.tsx
@@ -92,19 +92,21 @@ export function NestedSafesList(): ReactElement | null {
               </Typography>
             )}
 
-            <CheckWallet>
-              {(isOk) => (
-                <Button
-                  onClick={() => setTxFlow(<CreateNestedSafe />)}
-                  variant="text"
-                  startIcon={<SvgIcon component={AddIcon} inheritViewBox fontSize="small" />}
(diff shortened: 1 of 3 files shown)


tmjssz (approved):

katspaugh:
I think it would make sense to hide the Nested Safes icon button and popup entirely for undeployed safes.

Screenshot 2025-03-26 at 10 10 22

PR #5498, opened 2025-03-24
feat(Signing UX): receipt screen for CreateNestedSafe + MigrateSafeL2 flows

### LINKED ISSUE #5496: [Signing UX v2] Make nested safes' safe creation have a receipt screen
##Context
Nested safe feature is being deployed. This feature is able to create a safe from another safe. So SafeA can create a SafeB and send tokens to it. The result is SafeB being created with SafeA as its only owner, with the tokens transferred to it

## What must be done
Creating a safe through nested safe feature must show the receipt screen at the end of it

## Designs/sketches
Nothing particular for this tx, is the same receipt screen like the other tx's

## How to test it
Here, instead of clicking sign I'd expect the "Continue" button to get to the receipt screen
Image

tmjssz:
## What it solves

Resolves #5496 

Adds the transaction receipt screen to the **CreateNestedSafe** and **MigrateSafeL2** flows.

## Checklist

### DIFF
--- a/apps/web/src/components/tx-flow/flows/CreateNestedSafe/ReviewNestedSafe.tsx
+++ b/apps/web/src/components/tx-flow/flows/CreateNestedSafe/ReviewNestedSafe.tsx
@@ -1,15 +1,12 @@
 import { skipToken } from '@reduxjs/toolkit/query'
-import { useContext, useEffect, useMemo } from 'react'
+import { useCallback, useContext, useEffect, useMemo } from 'react'
 import type { ReactElement } from 'react'
 import type { MetaTransactionData, SafeTransaction } from '@safe-global/safe-core-sdk-types'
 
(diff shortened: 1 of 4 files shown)


iamacook (approved):

francovenica:
LGTM

Tested the nested safes' safe creation, updating a safe version and mass payout as well

katspaugh:
@tmjssz does the Receipt in this case show a wrapped Safe transaction (i.e. data inside data inside message) or the inner message?

tmjssz:
> @tmjssz does the Receipt in this case show a wrapped Safe transaction (i.e. data inside data inside message) or the inner message?

Not exactly sure what you mean, but for nested Safe creation it shows the regular transaction data I would say:

Screenshot 2025-03-25 at 10 47 47

PR #5502, opened 2025-03-25
Feat(Orgs): Show breadcrumbs if navigating to a safe from a space

usame-algan:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/58

## How this PR fixes it

- Adds a new `SpaceBreadcrumbs` component
- Adds a new `Breadcrumbs` component that combines the two existing breadcrumbs
- Adjusts sidebar navigation to include a possible `spaceId` query param when navigating
- Adjusts the queue actions to incldue the `spaceId` param when navigating

## How to test it

1. Open a space
2. Open a safe account from that safe e.g. by pressing the queued tag
3. Observe breadcrumbs on that safe
4. There should be a link back to the spaces list and the space itself
5. This should work in combination with nested safes and nested safes should work without being signed in as well

## Screenshots

Nested safe breadcrumbs + spaces breadcrumbs
Screenshot 2025-03-25 at 15 10 22

Only nested safes breadcrumbs
Screenshot 2025-03-25 at 15 11 10

Only spaces breadcrumbs
Screenshot 2025-03-25 at 15 12 35

## Checklist


TanyaEfremova:
I don't think the icon is correct. Could we use this one instead? 
ic-space

PR #5503, opened 2025-03-25
fix(Mass payouts): sum calculation for token amounts with large number of decimals

tmjssz:
### Issue  
When a user enters a token amount with a large number of decimal places, the form validation fails. As a result, clicking the **MAX** button incorrectly triggers the "Insufficient balance" alert, even when the user should have sufficient funds.  

### Root Cause  
The `validateAmount` function calculates the total sum of all defined amounts for a given token by converting input strings into numbers. However, for values with many decimal places, JavaScript rounds the number, leading to inaccurate calculations and an incorrect validation outcome.  

### Fix  
Instead of converting input strings to floating-point numbers, this fix ensures that values are transformed into **BigInt** representations without decimals before summing. This approach eliminates rounding errors and ensures accurate balance validation.


katspaugh (approved):

francovenica:
LGTM.

Checked that the Max value when there are 18 decimals are calculated properly, this for 1 field or more

PR #5504, opened 2025-03-25
feat(mobile):  add biometrics feature on onboarding

Jonathansoufer:
## What it solves

This PR implements Biometrics enablement from ONBOARDING ONLY. The Settings and feature restrictions based on the state of the biometrics will follow in different PRs

## How to test it
Install the app and go through the onboarding flow.

## Screenshots

https://github.com/user-attachments/assets/cab6ccdb-87c4-42aa-a43c-b86f330289e5

## Checklist

### DIFF
--- a/apps/mobile/src/hooks/useBiometrics.ts
+++ b/apps/mobile/src/hooks/useBiometrics.ts
@@ -60,7 +60,9 @@ export function useBiometrics() {
       const isSupported = await checkBiometricsSupport()
 
       if (!isSupported) {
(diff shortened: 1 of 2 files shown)


compojoom (approved):
I'm OK with merging this like this, but I don't think we need to display this screen after the onboarding. It needs to be displayed right before the user wants to add a PK. 
The user can do this during the onboarding - so by the time you show this it will be too late. If the user then later doesn't enable it he can't sign.

PR #5505, opened 2025-03-25
Fix(Signing UX v2): show checkbox just once till checked

katspaugh:
## What it solves

Show the signing acknowledgment checkbox only until the user checks it once, and don't show afterwards.

## How to test

* Make a tx, proceed to the Receipt screen
* Observe the checkbox is in place
* Go back to the previous screen
* Go to Receipt again
* Checkbox still there
* Accept the checkbox
* Make another tx
* No checkbox anymore


apps/web/src/components/tx/ConfirmTxDetails/index.tsx
```diff
@@ -65,18 +66,24 @@ const HardwareWalletStep = [
...
   const showHashes = wallet ? isHardwareWallet(wallet) || isLedgerLive(wallet) : false
   const steps = showHashes ? HardwareWalletStep : InfoSteps
 
   const handleCheckboxChange = useCallback(({ target: { checked } }: React.ChangeEvent<HTMLInputElement>) => {
     trackEvent({ ...MODALS_EVENTS.CONFIRM_SIGN_CHECKBOX, label: checked })
-    setChecked(checked)
-  }, [])
+    setCheckboxAccepted(checked)
+    setCheckboxAcceptedNow(true)
+  }, [setCheckboxAccepted])
```
tmjssz, line 86: Nit: Can't we leave the dep array empty? I think the setter function is stable

tmjssz (approved):

apps/web/src/components/tx/ConfirmTxDetails/index.tsx
katspaugh: It's indeed stable but eslint complains...

PR #5510, opened 2025-03-26
Fix(Balances): add skipPollingIfUnfocused to polled queries

katspaugh:
## What it solves

We're seeing a 3x increase in balances requests after migrating the Balances fetching to RTK Query.
@compojoom's hypothesis is that it's because the old `useIntervalCounter` polling was being suspended in a background tab and the query wasn't.

## How to test

* Open the Console -> Network and filter by `balances`
* Observe that balances are polled every 15s
* Go to another tab, wait for 1 minute
* Come back to the Safe tab
* Observe that there's just one new request, not 4+


compojoom (approved):
nice 👍

francovenica:
Looks good to me. Compared the calls with current prod and I see that only 1 call was done while the tab as inactive

Issue #5477, opened 2025-03-24
[Blockaid] Allow reporting false positives


katspaugh:
## What is the feature about

Users sometimes complain that Blockaid is giving them false positives for benign transactions.
Blockaid supports reporting such cases via the API after which Blockaid reviews them.

Allow the user to report false scans with a click of a button.

## The list of requirements

* Add a button near the malicious tx warning saying "Report wrong results" or similar
* Send an API request to report the scan. Docs: [https://docs.blockaid.io/reference/evm-address-report](https://docs.blockaid.io/reference/evm-address-report)
  * A workaround would be to prefill a UI developed by Blockaid, just like Metamask does it (see *Linked Slack*)

**Additionally:**

* When the user gets a malicious tx warning but clicks the checkbox "Proceed anyway", also report it to Blockaid via another request: [https://docs.blockaid.io/reference/scan-status](https://docs.blockaid.io/reference/scan-status-1) with the status `accepted`
* When the user abandons a tx with a warning, report it with the status `rejected`

## Design

[Transaction_Review - Transaction flow (Figma)](https://www.figma.com/design/cMDccuR9aZFqAiBd4dKsKg/Transaction-flow?node-id=2179-48429&t=7idBwuMJ8lXTxCNI-4)

- [X] @simon ask blockaid if we can pass parameters to the Blockaid form [https://report.blockaid.io/mistake](https://report.blockaid.io/mistake)  
  [https://5afe.slack.com/archives/C05UN2C71RV/p1754935876173429](https://5afe.slack.com/archives/C05UN2C71RV/p1754935876173429)

katspaugh:
@TanyaEfremova could you prepare a design for this?

the-real-chrizzo:
The reporting feature is not relevant right now as we are evaluating another solution. Happy to resurface this at a later point.

mulligan-madness:
Heya wanted to check if you have any updates here... Had someone in our orgs address pop up as malicious even though litterally nothing has happened and they are still in possession of the multisig. Noticed in the gnosis ui. Even a link or something to know what actually caused the flag...

As it stands this is actually counterproductive for security because you are teaching people that "malicious" tags are inaccurate and can "usually" be safely ignored. Of course the onus is always on the user to verify, but presenting unqualified false information with no recourse to correct the error will only hurt.

katspaugh:
cc @the-real-chrizzo

simabeh:
[...]

PR #5523, opened 2025-03-27
Feat(Blockaid): report scan reactions

katspaugh:
## What it solves

Part of #5477

## How this PR fixes it

When a Blockaid risk warning is confirmed or ignored, send a report to Blockaid.

## How to test it

1.
* Create a malicious tx
* Ignore the risk and sign or execute it
* Observe a new request to Blockaid

2.
* Create a malicious tx
* Look at the warning and close the tx
* Observe a new request to Blockaid


apps/web/src/components/tx/security/blockaid/useBlockaid.ts
```diff
@@ -22,23 +22,29 @@ export const useBlockaid = (
   const { safe, safeAddress } = useSafeInfo()
   const signer = useSigner()
   const isFeatureEnabled = useHasFeature(FEATURES.RISK_MITIGATION)
+  const jsonData = data && JSON.stringify(data)
```
katspaugh, line 25: Stringified this because it was spamming requests every time data is updated, and data was updating every 15s.

apps/web/src/components/tx/security/shared/TxSecurityContext.tsx
```diff
@@ -56,11 +58,26 @@ export const TxSecurityContext = createContext<TxSecurityContextProps>(defaultSe
...
+  const [blockaidResponse, blockaidError, blockaidLoading] = useBlockaid(txData, txOrigin)
+  const reportScan = useBlockaidReportScan(blockaidResponse?.payload?.requestId)
+  const needsRiskConfirmation = !!blockaidResponse && blockaidResponse.severity >= SecuritySeverity.HIGH
 
   const [isRiskConfirmed, setIsRiskConfirmed] = useState(false)
   const [isRiskIgnored, setIsRiskIgnored] = useState(false)
 
+  // Report scan status when risk is confirmed or ignored
+  useEffect(() => {
+    // Risk is accepted, meaning the user rejected the warning
```
usame-algan, line 71: Took me a second to understand this part. User accepts the risk = user rejects the warning but in case it is a high severity issue the user shouldn't be able to ignore the warning anyway so if the user doesn't accept the risk we can be sure that they won't submit the transaction.

PR #5524, opened 2025-03-27
feat(mobile): add general settings + theme control + biometrics on settings

Jonathansoufer:
## What it solves

This PR implements a General Settings session on the app, granting access to a variety of options (some enabled already, some coming soon). This also implements a guard flow using Biometrics.

## How this PR fixes it

[x] - Implements guarding on sign import conditional to biometrics enabled.
[x] - Implements general settings to the app
[x] - disable not implemented features on settings
[x] - Implements theme choice from within the app
[x] - Implements about items on settings

### Testing Biometrics Guard
Case a
1 - On a fresh install, try to import a signer
2 - After enabling on onboarding, import a signer

Case b
1 - On a fresh install, don't import an signer
2 - Go to settings --> Signers --> try to import

### Testing General Settings
1 - Go to settings --> Gear icon
2 - Change theme
3 - Click on Address Book
4 - Click on General Session itens (last three)

## Recording
https://github.com/user-attachments/assets/56d8595a-d3f2-48a2-a01c-5fd15e1634be

## Screenshots
Screenshot 2025-03-27 at 12 04 41
Screenshot 2025-03-27 at 12 04 25
Screenshot 2025-03-27 at 12 04 18

## Checklist


compojoom (commented):
No longer crashes on initial import of a PK, but the navigation is wrong. It should lead the user abck to teh "import your signers to unlock account" screen and not the "import a signer" screen.

compojoom (commented):
I enable FaceId. I close the app. Go to the OS Settings. Disable face id. Come back to the app, go to the app settings ->  face Id is still enabled. 

When I now try to sign a tx I'm actually asked to enable faceId (which is good), but the first press on "enable" does nothing. The second press sends me to the "app settings" in the OS. 
If I come back now the app resets. I navigate to confirm the tx -> I'm asked again for biometric. I click on enable and this time it works. 

I know it is an edge case. Maybe we should open a new ticket for reference and can improve the behavior in a separate PR.

compojoom (commented):
The bug with scrolling on small devices where teh "Settings" view is jumping is still there.

apps/mobile/src/components/ScreenHeader/ScreenHeader.tsx
compojoom, line 1: Why do you need this? We have teh useScrollableHeader component together with the LargeHeaderTitle and those should be enough. Check how we use it in the PendingTxListContainer.

apps/mobile/src/features/Settings/components/AppSettings/AppSettings.container.tsx
```diff
@@ -1,24 +1,132 @@
...
+                {
+                  id: 'dark',
+                  title: 'Dark',
+                },
+                {
+                  id: 'light',
+                  title: 'Light',
+                },
+              ]}
+            >
```
compojoom, line 53: That implementation doesn't work very well. In the designs the arrow for appeareace points down indicating it is a dropdown. In our implementation it is arrow right, which makes me think that it is a menu. I can click on the "appearance" text itself and it does nothing and that is confusing. I have to click directly on the "auto" button and then I see the dropdown. 

If hard to fix right now, open an issue and let's do it in a different sprint.

apps/mobile/src/theme/provider/safeTheme.tsx
```diff
@@ -8,32 +8,57 @@ import { NavDarkTheme, NavLightTheme } from '@/src/theme/navigation'
...
 
-      <TamaguiProvider config={config} defaultTheme={colorScheme ? colorScheme : 'dark'}>
+      <TamaguiProvider config={config} defaultTheme={currentTheme ?? 'light'}>
         {themeProvider}
       </TamaguiProvider>
     </FontProvider>
   )
 }
+
+export const useTheme = () => {
```
compojoom, line 45: Nice, but maybe in a separate file?

compojoom (approved):
Thank you 👏 !

PR #5526, opened 2025-03-27
Fix(Web): Lazy load my accounts list

### LINKED ISSUE #5508: Lots of 429 on My Accounts page
The Accounts page is doing too many requests to the /safes endpoint (Safe Overviews).

If one fails, it keeps fetching them because of infinite scroll.

Image

usame-algan:
## What it solves

Resolves #5508 

## How this PR fixes it

- Refactors the `useOnceVisible` hook and tests it
- Removes the `TransitionGroup` and `Collapse` from `SafesList`
- - This doesn't seem to affect UX and fixes the lazy load issue
- Increases the fetch timeout for the `SafeOverviewFetcher`
- - The previous timeout of 50ms was too short and resulted in overview requests for single safe accounts

## How to test it

1. Go to `/welcome/accounts`
2. Observe safe overview requests only happening for visible safe items
3. Scroll through your safe list very fast
4. Observe few batched requests happening for your safes
5. Reload the page and scroll through the list slower
6. Observe more requests happening for your safes (The goal is not to eliminate this case entirely but find a good middleground)

## Screenshots

https://github.com/user-attachments/assets/285359eb-9ed6-4ad7-bc33-103712df612f

## Checklist


katspaugh:
I've tested it, the Accounts list now works as intended. ✅ 
Infinite scroll in the Tx History also still works.

PR #5530, opened 2025-03-28
refactor(web): use utility functions from @safe-global/utils

compojoom:
## What it solves

This PR looks large but it only moves files around. Each commit moves a different utility file.

**Context**

We introduced the @safe-global/utils package to share TypeScript utilities between the web and mobile apps. Over the past few months, I’ve been copying shared code into this package as needed, but I neglected to update the web app’s tsconfig and didn’t replace the usage of the moved code in the web app.

This PR fixes that:
- The moved utilities are now used by both the web and mobile apps.
- Each moved file corresponds to one commit for easier review.

**Limitations**

Not all code can be moved yet due to type incompatibilities:
- The mobile app relies on types from @safe-global/store.
- The web app still uses @safe-global/typescript-sdk in many places.

For example, I can’t use features/swap/helpers/utils.ts in the web app because I’ve updated its types to rely on @safe-global/store, which causes type conflicts.

Going forward, we should move only code that relies on the new shared types into @safe-global/utils.

**Structure**

The folder structure in @safe-global/utils mirrors the structure in the web app. This keeps it intuitive and reduces mental overhead when navigating shared code.

**Please Note**

If you’re writing pure utility functions (just TypeScript), add them to @safe-global/utils, not directly into the web app. This makes future code reuse easier across platforms.

**Resolves**
Code sharing between web and mobile.

## How this PR fixes it
Just moves files around.

## How to test it
Our automated tests should catch any errors if any. 

## Checklist


compojoom:
Hey guys, I've tagged multiple people. I don't need a review from everyone. Just want to make everyone aware of the change.

katspaugh (approved):
tl;dr but looks good!

packages/utils/src/utils/gateway.ts
```diff
@@ -26,3 +27,28 @@ export const getExplorerLink = (
 
   return { href, title }
 }
+export const signTxServiceMessage = async (
```
iamacook, line 30: Nit: this is not a generic function for signing Transaction Service messages.
```suggestion
export const signTxDeletionMessage = async (
```

packages/utils/src/utils/validation.ts
```diff
@@ -63,8 +63,8 @@ export const validateAmount = (amount?: string, includingZero: boolean = false)
   }
 }
 
-export const validateLimitedAmount = (amount: string, decimals?: number, max?: string) => {
-  if (typeof decimals === 'undefined' || !max) return
+export const validateLimitedAmount = (amount: string, decimals?: number | null, max?: string, errorMsg?: string) => {
+  if (decimals == null || !max) return
```
iamacook, line 67: Are we sure that this isn't used in places where `decimals` isn't declared?

packages/utils/src/utils/validation.ts
```diff
@@ -74,12 +74,12 @@ export const validateLimitedAmount = (amount: string, decimals?: number, max?: s
...
   if (value !== undefined && value > BigInt(max)) {
-    return `Maximum value is ${safeFormatUnits(max, decimals)}`
+    return errorMsg || `Maximum value is ${safeFormatUnits(max, decimals)}`
   }
 }
 
-export const validateDecimalLength = (value: string, maxLen?: number, minLen = 1) => {
-  if (typeof maxLen === 'undefined' || !value.includes('.')) {
+export const validateDecimalLength = (value: string, maxLen?: number | null, minLen = 1) => {
+  if (maxLen == null || !value.includes('.')) {
```
iamacook, line 82: As above regarding undeclared `maxLen`.

packages/utils/src/utils/formatters.ts
```diff
@@ -36,16 +37,16 @@ export const safeFormatUnits = (value: BigNumberish, decimals: number | string =
...
-  const amount = safeFormatUnits(value, decimals)
+  const amount = safeFormatUnits(value, decimals ?? GWEI)
   return precision !== undefined ? formatAmountPrecise(amount, precision) : formatAmount(amount)
 }
 
-export const safeParseUnits = (value: string, decimals: number | string = GWEI): bigint | undefined => {
+export const safeParseUnits = (value: string, decimals?: number | string | null): bigint | undefined => {
   try {
-    return parseUnits(value, decimals)
+    return parseUnits(value, decimals ?? GWEI)
```
iamacook, line 49: Nit: It's subjective, but default values are clearer at the top of the function (whether as a fallback or reassignment). I'll leave this up to your discretion though.

packages/utils/src/utils/validation.ts
compojoom: same as above. Those changes are in the validation.ts in production. I've just applied them to utils/validation.ts file.

packages/utils/src/utils/formatters.ts
compojoom: Same as with validation.ts.

packages/utils/src/utils/validation.ts
iamacook: Understood. Thank you for the clarification.

apps/web/src/utils/__tests__/addresses.test.ts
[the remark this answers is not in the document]
compojoom: because the same test 1:1 is already in the utils package (it was copied few months ago).

PR #5536, opened 2025-03-28
Refactor(web): use TypedData type from the @safe-global/utils package

compojoom:
## What it solves
Refactors the usage of the old types to the new ones. Why? Because the new ones are "more correct & auto-generated" and the mobile app already uses them. If we want to reuse code, types need to match, otherwise all hell breaks lose and developers loose their sanity & joy of life. 

Resolves #

## How this PR fixes it
There is a codemod that maps old types to new types and also exchanges the use of the old enum types. After running the script I went through the files and did manual modifications that were necessary for the types to be correctly used

## How to test it
Everything should continue to work as before. Unit tests and e2e tests should catch any errors.

## Checklist


apps/web/src/components/safe-messages/MsgDetails/index.tsx
```diff
@@ -46,7 +46,7 @@ const MsgDetails = ({ msg }: { msg: SafeMessage }): ReactElement => {
             <EthHashInfo
               address={msg.proposedBy.value || ''}
               name={msg.proposedBy.name}
-              customAvatar={msg.proposedBy.logoUri}
+              customAvatar={msg.proposedBy.logoUri || undefined}
```
usame-algan, line 49: Afaiu logoUri now has the type of `string | null | undefined`. Is there a benefit of having a fallback here instead of adjusting the `EthHashInfo` type to also accept `null` as a value for `customAvatar`?

apps/web/src/components/safe-messages/MsgSigners/MsgSigners.test.tsx
```diff
@@ -31,8 +31,8 @@ describe('MsgSigners', () => {
       proposedBy: {
         value: '',
       },
-      status: SafeMessageStatus.NEEDS_CONFIRMATION,
-      type: SafeMessageListItemType.MESSAGE,
+      status: 'NEEDS_CONFIRMATION' as const,
+      type: 'MESSAGE' as const,
```
usame-algan, line 35: Does it throw errors for you without the `as const`? For me its working without.
Screenshot 2025-03-31 at 11 19 20

compojoom: Initially it was, that's why the codemod is setting it to "as const", not sure why it works without it now.

compojoom: Ok, removed.

apps/web/src/components/safe-messages/MsgDetails/index.tsx
usame-algan: Although thinking about it it makes sense to change it here instead of everywhere downstream so we don't have to deal with a null value in potentially many components.

compojoom: Good question. The EthHashInfo component has specific requirements. if you pass a prop to it, it should be a string, otherwise you should not pass a customAvatar. The component defines concrete function arguments. 
I'm not sure changing the function arguments should be done, because an API we use might return an incompatible type. 
On the other hand I agree that it is a bit weird, but not sure that deserves changing the EthHashInfo component.

apps/web/src/utils/safe-message-guards.ts
```diff
@@ -1,10 +1,10 @@
-import { SafeMessageListItemType } from '@safe-global/safe-gateway-typescript-sdk'
-import type { SafeMessageListItem, SafeMessage, SafeMessageDateLabel } from '@safe-global/safe-gateway-typescript-sdk'
+import type { MessageItem, DateLabel } from '@safe-global/store/gateway/AUTO_GENERATED/messages'
+import type { SafeMessageListItem } from '@safe-global/store/gateway/types'
 
-export const isSafeMessageListDateLabel = (item: SafeMessageListItem): item is SafeMessageDateLabel => {
-  return item.type === SafeMessageListItemType.DATE_LABEL
+export const isSafeMessageListDateLabel = (item: SafeMessageListItem): item is DateLabel => {
+  return item.type === 'DATE_LABEL'
```
katspaugh, line 5: Using `DateLabel` for comparison wouldn't work?

katspaugh (approved):

apps/web/src/components/safe-messages/MsgDetails/index.tsx
```diff
@@ -23,9 +23,9 @@ import CopyButton from '@/components/common/CopyButton'
 import NamedAddressInfo from '@/components/common/NamedAddressInfo'
 import MsgShareLink from '../MsgShareLink'
 
-const MsgDetails = ({ msg }: { msg: SafeMessage }): ReactElement => {
+const MsgDetails = ({ msg }: { msg: MessageItem }): ReactElement => {
   const wallet = useWallet()
-  const isConfirmed = msg.status === SafeMessageStatus.CONFIRMED
+  const isConfirmed = msg.status === ('CONFIRMED' as const)
```
iamacook, line 28: Nit: You don't need `as const` here because the string is inherently a constant. (There are a few other `isConfirmed` intances that are the same but I won't mark them.)

iamacook (approved):

apps/web/src/components/safe-messages/MsgDetails/index.tsx
compojoom: fixed.

PR #5537, opened 2025-03-29
refactor(web): move services/security to @safe-global/utils

compojoom:
## What it solves
We need to implement the blockaid and tenderly checks in the mobile app. To minimize the amount of the same code we need to write again, I want to reuse the security services files from the web app. For that I need to move them to @safe-global/utils. Unfortunately they relied on other utils files, constants etc that were not already moved. So I had to move them first. This branch builds on top of the data-type branch as the change to the data type is also necessary for this to work properly. 

Resolves #

## How this PR fixes it
- This PR moves utility functions and constants to the utils package. Those functions and constants can be used by bothe the app/web and app/mobile packages.

On top of this I've been able to use the tenderly useSimulation function in the mobile app with as little copy/paste as possible. The changes to the mobile app will come in a separate PR.  

## Checklist


config/test/presets/jest-preset.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
   // setupFilesAfterEnv: ['<rootDir>/../../config/jest-presets/jest/setup.js'],
 
   transformIgnorePatterns: [
-    'node_modules/(?!((jest-)?react-native|@react-native(-community)?)|expo(nent)?|@expo(nent)?/.*|@expo-google-fonts/.*|react-navigation|@react-navigation/.*|@sentry/react-native|native-base|react-native-svg|react-redux|moti/.*)',
+    'node_modules/(?!((jest-)?react-native|@react-native(-community)?)|expo(nent)?|@expo(nent)?/.*|@expo-google-fonts/.*|react-navigation|@react-navigation/.*|@sentry/react-native|native-base|react-native-svg|react-redux|moti/.*|@safe-global/safe-apps-sdk|ethereum-cryptography|@safe-global/protocol-kit|@safe-global/safe-gateway-typescript-sdk)',
```
usame-algan, line 19: Tests still run for me even without these changes. Are they needed?

compojoom: the tests in @safe-global/utils were failing without it.

apps/web/.gitignore
```diff
@@ -1,6 +1,3 @@
-# Types
-src/types/contracts
```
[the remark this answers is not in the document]
compojoom: If I remove this everyone will see those types in their git commit and need to explicitly delete them. You want me to do this?

PR #5545, opened 2025-03-31
Fix(Orgs): Nested safe breadcrumbs, add breadcrumb analytics events

usame-algan:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/58

## How this PR fixes it

- Fixes the parent safe link in the nested safe breadcrumbs by including the chain prefix in the pathname
- Includes all other query params in the parent safe link as well
- Adds two new analytics events to the space breadcrumbs when navigating to a space or to the space list
- Adds a link to the safe when clicking a safe item in spaces

## How to test it

1. Open a nested safe so that the breadcrumbs are visible
2. Add a spaceId parameter to the url
3. Observe the spaces breadcrumbs also appear
4. Press the parent safe button
5. Observe the spaceId stays in the url and the breadcrumbs are still visible
6. Press the space button and the space overview button
7. Observe analytics events for both

## Checklist

### DIFF
--- a/apps/web/src/components/common/NestedSafeBreadcrumbs/index.tsx
+++ b/apps/web/src/components/common/NestedSafeBreadcrumbs/index.tsx
@@ -5,22 +5,30 @@ import type { ReactElement } from 'react'
 import useSafeInfo from '@/hooks/useSafeInfo'
 import { useParentSafe } from '@/hooks/useParentSafe'
 import { BreadcrumbItem } from '@/components/common/Breadcrumbs/BreadcrumbItem'
+import { formatPrefixedAddress } from '@safe-global/utils/utils/addresses'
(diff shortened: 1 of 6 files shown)


jmealy (commented):
I like the breadcrumbs, useful stuff 👍 

Is it correct that you can only open a safe from a space if there is pending transactions on it? Seems like an arbitrary restriction if so. Could we have it like in the sidebar where we can just click the safe and it will open?

usame-algan:
> Could we have it like in the sidebar where we can just click the safe and it will open?

Yes, agree. Having that link for every safe would be better. I will add it in this PR

jmealy (approved):
Looks like there is an unused import in `myAccounts/components/AccountItems/SingleAccountItem.tsx`

PR #5546, opened 2025-03-31
Fix(Spaces): add missing tracking event and remove misfiring events

jmealy:
## What it solves

QA feedback for: https://github.com/safe-global/safe-wallet-monorepo/issues/5055

## How this PR fixes it
- do not send a tracking event when clicking the `Spaces` tab on the accounts page, if it has already been selected
- do not send an event for expanding a multi chain safe in the space accounts list.
- add missing events in the delete space flow.
- Minor name changes.

## How to test it
- go to the accounts page and click the spaces tab. See that the event is fired correctly. Clicking the tab again immediately should not fire the event a second time.
- go to the accounts list for a space. Expand a multichain safe item. See that no event is fired.
- Open the `delete space` dialog from the context menu in the spaces list. See that an event is fired.
- Click the confirm deletion from the `delete space` dialog. See that an event is fired.

## Checklist

### DIFF
--- a/apps/web/src/features/spaces/components/SpaceCard/SpaceContextMenu.tsx
+++ b/apps/web/src/features/spaces/components/SpaceCard/SpaceContextMenu.tsx
@@ -12,6 +12,8 @@ import type { GetSpaceResponse } from '@safe-global/store/gateway/AUTO_GENERATED
 import css from './styles.module.css'
 import DeleteSpaceDialog from '@/features/spaces/components/SpaceSettings/DeleteSpaceDialog'
 import UpdateSpaceDialog from '@/features/spaces/components/SpaceSettings/UpdateSpaceDialog'
+import Track from '@/components/common/Track'
+import { SPACE_EVENTS, SPACE_LABELS } from '@/services/analytics/events/spaces'
 
 enum ModalType {
   RENAME = 'rename',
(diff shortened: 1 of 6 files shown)


apps/web/src/features/spaces/components/SpaceCard/SpaceContextMenu.tsx
```diff
@@ -57,12 +59,14 @@ const SpaceContextMenu = ({ space }: { space: GetSpaceResponse }) => {
           <ListItemText>Rename</ListItemText>
         </MenuItem>
 
-        <MenuItem onClick={(e) => handleOpenModal(e, ModalType.REMOVE)}>
-          <ListItemIcon>
-            <SvgIcon component={DeleteIcon} inheritViewBox fontSize="small" color="error" />
-          </ListItemIcon>
-          <ListItemText>Remove</ListItemText>
-        </MenuItem>
+        <Track {...SPACE_EVENTS.DELETE_SPACE_MODAL} label={SPACE_LABELS.space_dashboard_card}>
```
usame-algan, line 62: I would suggest adding a new label here like `space_context_menu` because this is also used on the space list page and not just the dashboard/sidebar

apps/web/src/features/spaces/components/SpaceSettings/DeleteSpaceDialog.tsx
```diff
@@ -40,6 +42,7 @@ const DeleteSpaceDialog = ({ space, onClose }: { space: GetSpaceResponse | undef
     if (!space) return
 
     setError(undefined)
+    trackEvent({ ...SPACE_EVENTS.DELETE_SPACE })
```
usame-algan, line 45: I would move this after the await call to make sure its only fired if the CGW request succeeded to keep it consistent with the other submit events.

usame-algan (approved):

usame-algan (approved):
Thanks!

PR #5550, opened 2025-03-31
Fix(Spaces): Re-trigger address transformation on chain change

jmealy:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/77

## How this PR fixes it
- triggers the address transformation to remove the chain prefix anytime the chain is changed, not just when the address itself is changed. 
- This has the side effect of re-triggering validation, so the manual re-trigger of validation is removed.

## How to test it
- Open the `add address manually` dialog (or the add safe flow).
- Copy an address with the chain prefix included
- paste the address into the address input, but with the incorrect chain selected.
- It will show a validation error
- then select the correct chain. The chain prefix should be removed from the displayed value. You should be able to submit and add the safe to your space without issues.

## Checklist

### DIFF
--- a/apps/web/src/components/common/AddressInput/index.test.tsx
+++ b/apps/web/src/components/common/AddressInput/index.test.tsx
@@ -181,9 +181,10 @@ describe('AddressInput tests', () => {
       fireEvent.change(input, { target: { value: 'zero.eth' } })
     })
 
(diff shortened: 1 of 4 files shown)


usame-algan (approved):
Some tests were failing. I pushed a slight fix to check for the existence of an ens `address` before running the new hook. Still one `AddressInput` test keeps failing but I can't reproduce the issue when manually testing so I am skipping it for now.

PR #5565, opened 2025-04-01
Fix(Spaces): Add a CTA to empty spaces card on the spaces list page

jmealy:
## What it solves
The CTA for the logged out and disconnected states is in the center of the screen, but for the empty state, the CTA jumps to the top right.

## How this PR fixes it
- When there are no spaces in the list, do not show the CTA at the top right, add it to the empty spaces card instead.
- Also adds some padding above the CTA for this screen and the logged out and disconnected states

## How to test it
- Go to the spaces page and sign in with a EOA that is not part of any spaces yet.
- See that the CTA to create a space is in the middle of the screen

## Screenshots
image

## Checklist

### DIFF
--- a/apps/web/src/features/spaces/components/SpacesList/index.tsx
+++ b/apps/web/src/features/spaces/components/SpacesList/index.tsx
@@ -84,7 +84,7 @@ const NoSpacesState = () => {
           </Link>
         </Box>
         <Track {...SPACE_EVENTS.CREATE_SPACE_MODAL} label={SPACE_LABELS.space_list_page}>
-          <AddSpaceButton disabled={false} size="medium" />
+          <AddSpaceButton disabled={false} />
         </Track>
       </Card>
       {isInfoOpen && (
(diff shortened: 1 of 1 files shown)


apps/web/src/features/spaces/components/SpacesList/index.tsx
```diff
@@ -74,13 +74,18 @@ const NoSpacesState = () => {
...
+          <Typography color="text.secondary" mb={1}>
+            No spaces found.
+            <br />
+          </Typography>
+          <Link onClick={() => setIsInfoOpen(true)} href="#">
+            What are spaces?
+          </Link>
+        </Box>
+        <Track {...SPACE_EVENTS.CREATE_SPACE_MODAL} label={SPACE_LABELS.space_list_page}>
+          <AddSpaceButton disabled={false} size="medium" />
```
usame-algan, line 87: Do we only show this if the user is signed in? In that case we can keep it static non disabled

usame-algan (approved):

francovenica:
Checked in the RC. Looks good

PR #5566, opened 2025-04-01
Refactor(Spaces): Use current spaceId from local storage

usame-algan:
## What it solves

Instead of pushing the `spaceId` query parameter whenever the user "leaves" a space and use that as a source of truth we store it in local storage and update it whenever the `spaceId` changes.

## How this PR fixes it

- Removes the `spaceId` query parameter outside of spaces
- Stores it as part of the `authSlice`
- Checks that the currently opened safe is part of the space

## How to test it

1. Open a space
2. Navigate to one of the safes
3. Observe no spaceId in the url
4. Open a safe that is not part of the space
5. Observe the breadcrumbs are not showing

## Checklist

### DIFF
--- a/apps/web/src/features/myAccounts/hooks/useGetHref.ts
+++ b/apps/web/src/features/myAccounts/hooks/useGetHref.ts
@@ -23,7 +23,7 @@ export const useGetHref = (router: NextRouter) => {
             : isSingleTxPage
               ? AppRoutes.transactions.history
               : router.pathname,
-        query: { ...router.query, safe: `${chain.shortName}:${address}` },
+        query: { ...(!isSpacePage && router.query), safe: `${chain.shortName}:${address}` },
       }
     },
     [isSingleTxPage, isWelcomePage, isSpacePage, router.pathname, router.query],
--- a/apps/web/src/features/spaces/components/AuthState/index.tsx
+++ b/apps/web/src/features/spaces/components/AuthState/index.tsx
@@ -1,10 +1,10 @@
-import { type ReactNode } from 'react'
+import { type ReactNode, useEffect } from 'react'
 import SignedOutState from '@/features/spaces/components/SignedOutState'
 import { isUnauthorized } from '@/features/spaces/utils'
 import UnauthorizedState from '@/features/spaces/components/UnauthorizedState'
 import LoadingState from '@/features/spaces/components/LoadingState'
-import { useAppSelector } from '@/store'
-import { isAuthenticated } from '@/store/authSlice'
+import { useAppDispatch, useAppSelector } from '@/store'
+import { isAuthenticated, setLastUsedSpace } from '@/store/authSlice'
 import { useSpacesGetOneV1Query } from '@safe-global/store/gateway/AUTO_GENERATED/spaces'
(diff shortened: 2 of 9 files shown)


apps/web/src/features/myAccounts/hooks/useGetHref.ts
```diff
@@ -23,7 +23,7 @@ export const useGetHref = (router: NextRouter) => {
             : isSingleTxPage
               ? AppRoutes.transactions.history
               : router.pathname,
-        query: { ...router.query, safe: `${chain.shortName}:${address}` },
+        query: { ...(!isSpacePage && router.query), safe: `${chain.shortName}:${address}` },
```
katspaugh, line 26: Should this just pass a `safe` param? What other params does it need?

apps/web/src/features/spaces/components/SignInButton/index.tsx
```diff
@@ -28,7 +28,7 @@ const SignInButton = () => {
 
       if (result) {
         const oneDayInMs = 24 * 60 * 60 * 1000
-        dispatch(setAuthenticated({ sessionExpiresAt: Date.now() + oneDayInMs }))
+        dispatch(setAuthenticated(Date.now() + oneDayInMs))
```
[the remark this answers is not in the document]
usame-algan: I added another key into that slice and because of the function signature it now expected both values to be present so narrowed down the function signature which requires this change.

apps/web/src/features/myAccounts/hooks/useGetHref.ts
usame-algan: The `...router.query` was there before so I didn't want to touch it but at the same time I don't want to keep the `spaceId` param that we show when on a space and if the user then clicks on one of the safe items so I had to add this exception.

apps/web/src/features/spaces/components/AuthState/index.tsx
```diff
@@ -23,6 +24,10 @@ const AuthState = ({ spaceId, children }: { spaceId: string; children: ReactNode
     (member) => member.user.id === currentUser?.id && member.status === MemberStatus.DECLINED,
   )
 
+  useEffect(() => {
+    dispatch(setLastUsedSpace(spaceId))
+  }, [dispatch, spaceId])
+
```
usame-algan, line 30: I added it here because we wrap this component around every space detail page so as soon as the user navigates to one of them it will be set. IMO this should be enough to cover subsequent navigations away from spaces.

usame-algan:
@katspaugh regarding our discussion today to use the backend as the source of truth. This is already in place in the breadcrumbs where we fetch the data about the space from the CGW via the space id (that now comes from the local storage). If the user was removed from the space or the space doesn't exist anymore the CGW will return nothing and we won't render the breadcrumbs.

katspaugh (approved):

katspaugh:
@usame-algan you misunderstood me. I meant that when you open a Safe, you should look up (on the backend) if it's a part of a Space. If yes, show the breadcrumbs (and use the localStorage only to single out which space in case of many).

usame-algan:
Had to force push because my commits stopped being verified

PR #5569, opened 2025-04-01
refactor(Signing UX): modular transaction flow

### LINKED ISSUE #5428: [Signing UX v2] Refactor SignOrExecute
Refactor the tx flow logic and UI for extensibility and maintainability.

Details here: https://www.notion.so/safe-global/Tx-flow-refactoring-for-extensibility-1b38180fe573808bbb02f6d45451115b

tmjssz:
## What It Solves  

Resolves #5428  

This POC introduces a more flexible and modular transaction flow setup, making it easier to create custom flows. Features and actions are now structured as middleware components, ensuring better separation of concerns. For example, logic for specific features (such as transaction notes) is encapsulated within independent components that register themselves and load only when needed.  

## Overview  

image

### **`TxFlowProvider`**  
A shared context provider that maintains the state and common properties required across all transaction flow components.  

### **`SlotProvider`**  
Manages dynamic slots within the transaction flow. Components can register/unregister themselves in specific slots, allowing for greater flexibility in flow customization.  

### **`TxFlow`**  
A wrapper component for the transaction flow, responsible for:  
- Managing transaction state and security context.  
- Handling transaction steps and progress.  
- Rendering child components dynamically within the flow.  

### **`TxFlowContent`**  
Renders the main content of the transaction flow, utilizing `TxFlowContext` to manage state and layout properties. It also orchestrates transaction steps and progress while rendering child components as needed.  

## How to Test  

This PR migrates three transaction flows - **ChangeThreshold**, **ConfirmTx**, and **TokenTransfer** - to the new structure. These, along with all other flows, should function identically to their previous implementations.  

## Checklist


katspaugh:
I suggest we iterate a bit more on this as per our discussion this morning.
Action points:

* Remove the eventBus – tx notes can be tracked simply on input blur
* Reduce the complexity by removing createTxFlow and parametrized actions/features
* Instead, just render all existing features and actions and let them turn themselves on/off

In a future iteration (not in this PR):
*  Create an API for Actions to plug in the UI in a certain place (e.g. the submit button slot) to override its behavior. E.g. ExecuteViaRole hides the regular execute button, replaces it with a select and its own submit button with conditional logic.

https://dinky.dog/?q=tx-flow-refactoring_9025693fa164ded0

katspaugh (approved):

apps/web/src/components/tx-flow/TxFlow.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const TxFlow = <T extends unknown>({
+  children = [],
+  initialData,
+  txId,
+  onSubmit,
+  onlyExecute,
+  isExecutable,
+  showMethodCall,
+  isRejection,
+  ReviewTransactionComponent = ReviewTransaction,
```
usame-algan, line 46: I noticed that we overwrite this Review component sometimes e.g. when confirming an existing tx. What are the differences in that flow that require a different review component?

apps/web/src/components/tx-flow/actions/Counterfactual.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return isCounterfactualSafe && !isProposing
+}
+
+const CounterfactualSlot = withSlot({
+  Component: Counterfactual,
+  slotName: SlotName.Submit,
+  id: 'counterfactual',
+  useSlotCondition: useShouldRegisterSlot,
+})
```
usame-algan, line 44: I really like this pattern!

apps/web/src/components/tx-flow/actions/Batching.tsx
```diff
@@ -0,0 +1,66 @@
...
+    !isProposing
+  )
+}
+
+const BatchingSlot = withSlot({
+  Component: Batching,
+  slotName: SlotName.Action,
+  id: 'batching',
+  useSlotCondition: useShouldRegisterSlot,
+})
```
usame-algan, line 64: Do you think its possible to test this somehow to make sure there is no regression to the previous behaviour? It looks like all the conditions were moved into the `useShouldRegisterSlot` but it would give us more confidence if there were tests for it.

apps/web/src/components/tx-flow/flows/ChangeThreshold/ChooseThreshold.tsx
```diff
@@ -11,151 +11,155 @@ import {
...
-          <Controller
-            control={formMethods.control}
-            rules={{
-              validate: (value) => {
-                if (value === safe.threshold) {
-                  return `Current policy is already set to ${safe.threshold}.`
-                }
-              },
+  return (
+    <TxFlowStep title="New transaction">
```
usame-algan, line 45: Wdyt about moving the `TxFlowStep` wrapper out to where the `TxFlow` is since they belong together and that way its more clear? e.g.

```
<TxFlow ...>
    <TxFlowStep ...>
        <ChooseThreshold ...>
    </TxFlowStep>
</TxFlow>
```

apps/web/src/components/tx-flow/slots/withSlot.tsx
```diff
@@ -0,0 +1,22 @@
...
+ */
+export const withSlot = <T extends SlotName>({
+  Component,
+  slotName,
+  id,
+  useSlotCondition = () => true,
+}: Omit<UseRegisterSlotProps<T>, 'condition'> & {
+  useSlotCondition: () => boolean
+}) => {
+  const WithSlot = () => {
```
usame-algan, line 16: nit: You can directly return the function since the variable is not used and named the same as the function. My IDE is throwing a warning for that: 
Screenshot 2025-04-04 at 11 16 10

apps/web/src/components/tx-flow/slots/hooks/useRegisterSlot.ts
```diff
@@ -0,0 +1,27 @@
...
+}: UseRegisterSlotProps<T>) => {
+  const { registerSlot, unregisterSlot } = useSlotContext()
+
+  useEffect(() => {
+    if (condition) {
+      registerSlot(slotName, id, Component)
+    } else {
+      unregisterSlot(slotName, id)
+    }
+  }, [condition, registerSlot, unregisterSlot, slotName, Component, id])
```
usame-algan, line 26: Should there be a cleanup where we ensure that we always unregister on unmount?

apps/web/src/components/tx-flow/slots/hooks/useSlotContext.ts
```diff
@@ -0,0 +1,10 @@
+import { useContext } from 'react'
+import { SlotContext } from '../SlotProvider'
+
+export const useSlotContext = () => {
+  const context = useContext(SlotContext)
+  if (!context) {
+    throw new Error('useSlotContext must be used within a SlotProvider')
```
usame-algan, line 7: Can this ever happen in runtime? In that case we should catch it.

apps/web/src/components/tx/ReviewTransactionV2/ReviewTransactionContent.tsx
```diff
@@ -0,0 +1,173 @@
...
+    isProposing,
+    isRejection,
+    canExecuteThroughRole,
+  } = useContext(TxFlowContext)
+
+  const [readableApprovals] = useApprovalInfos({ safeTransaction: safeTx })
+  const isApproval = readableApprovals && readableApprovals.length > 0
+  const isCounterfactualSafe = useIsCounterfactualSafe()
+  const [ActionComponent] = useSlot(SlotName.Action)
+  const features = useSlot(SlotName.Feature)
```
usame-algan, line 64: I am trying to understand why we only access the first action component. Right now it seems that `Batching` is the only action but what happens if there is a second component registered with `SlotName.Action`?

apps/web/src/features/tx-notes/TxNoteInput.tsx
```diff
@@ -1,14 +1,20 @@
 import { useCallback, useState } from 'react'
 import { InputAdornment, Stack, TextField, Typography, Alert } from '@mui/material'
+import { MODALS_EVENTS, trackEvent } from '@/services/analytics'
 
 const MAX_NOTE_LENGTH = 60
 
 export const TxNoteInput = ({ onChange }: { onChange: (note: string) => void }) => {
   const [note, setNote] = useState('')
+  const [isDirty, setIsDirty] = useState(false)
```
usame-algan, line 9: nit: If we use react-hook-form this is built-in

apps/web/src/features/tx-notes/TxNoteInput.tsx
```diff
@@ -17,6 +23,18 @@ export const TxNoteInput = ({ onChange }: { onChange: (note: string) => void })
...
+    setIsDirty(false)
+  }, [])
+
+  const onBlur = useCallback(() => {
+    if (isDirty) {
+      // Track the event only if the note is dirty
+      // This prevents tracking the event when the user focuses and blurs the input without changing the note
+      trackEvent(MODALS_EVENTS.SUBMIT_TX_NOTE)
+    }
+  }, [isDirty])
```
usame-algan, line 36: We also still call `trackAddNote`. Couldn't we check in there if the note is empty and only track then?

usame-algan (commented):
Good job! The new structure with the slots and tx flow is very neat

apps/web/src/components/tx-flow/actions/Batching.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const isBatchable = !!safeTx && !isDelegateCall(safeTx)
+
+  const onBatchClick = async (e: SyntheticEvent) => {
+    e.preventDefault()
+
+    if (!safeTx) return
+
+    setIsSubmittable(false)
+
+    await addToBatch(safeTx, origin)
```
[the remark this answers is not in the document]
tmjssz: Good idea. Added error handling in [7af53ff](https://github.com/safe-global/safe-wallet-monorepo/pull/5569/commits/7af53ff1103f1e1d4496d084fb9c2040d9b4e733)

apps/web/src/components/tx-flow/TxFlow.tsx
tmjssz: We need some customized versions of that component in some cases. For TokenTransfer e.g. we also display all the recipients and for the ConfirmTx flow we show an additional text on top of the page. But both customized components use the default `ReviewTransaction` under the hood.

apps/web/src/components/tx-flow/actions/Batching.tsx
tmjssz: Yes makes sense! Will add tests, but I think I will do it in a separate PR as this one is already quite big.

apps/web/src/components/tx-flow/flows/ChangeThreshold/ChooseThreshold.tsx
tmjssz: Yes I like that! Done in [601a0bd](https://github.com/safe-global/safe-wallet-monorepo/pull/5569/commits/601a0bd410d409bcffb81473cf31eb10d460223c)

apps/web/src/components/tx-flow/slots/withSlot.tsx
tmjssz: Done

apps/web/src/components/tx-flow/slots/hooks/useRegisterSlot.ts
tmjssz: Nice catch! Added a cleanup function

apps/web/src/components/tx/ReviewTransactionV2/ReviewTransactionContent.tsx
tmjssz: You’re right, we can simply render all the registered actions here. Initially I only rendered the first one, assuming we needed just a single action. But in this case, supporting an arbitrary number of actions makes sense.

apps/web/src/features/tx-notes/TxNoteInput.tsx
tmjssz: I’ve removed the trackAddNote call from `ReviewTransactionContentV2`. Currently, it remains only in `ReviewTransactionContent`, which is used for the old flow logic. Good point about adding a check for empty notes, I’ll make sure to include that.

apps/web/src/features/tx-notes/TxNoteInput.tsx
tmjssz: Done!

apps/web/src/components/tx-flow/slots/hooks/useSlotContext.ts
usame-algan: In that case I think its fine to leave it.

usame-algan (approved):
Thanks!

PR #5575, opened 2025-04-02
Fix(Spaces): Only update auth slice if user was signed in

usame-algan:
## What it solves

There were a lot of requests and flickering happening whenever the user disconnected their wallet (even if not signed in). This is because we are resetting the logged in state whenever that happens which rerenders every component where we access the authenticated state.

This PR changes this to only update the state if the user is logged in.

Note: The app still flickers in case the user is signed in and disconnects. This will be optimized in a future PR.

## How this PR fixes it

- Check for `isAuthenticated` before calling `setUnauthenticated`

## How to test it

1. Open a Safe
2. Connect your wallet
3. Disconnect your wallet
4. Observe no flicker

## Checklist


katspaugh (approved):

usame-algan:
While this PR limits the scope of the issue it doesn't solve the underlying problem of many state updates on signing out. That happens because of the `authListener` that calls `dispatch(cgwClient.util.resetApiState())`. This targets all data from our auto-generated api and if its used in more places than just Spaces those will be affected as well.

Instead we should target specific queries and reset those when disconnecting (spaces, space safes, members)

PR #5602, opened 2025-04-04
Fix(Spaces): Check address book when adding a member

usame-algan:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/115

## How this PR fixes it

- Changes the `NameInput` to be wrapped by a `Controller` to fix the shrink issue when setting a value
- Checks the address book whenever the address changed and sets the name value if there is a hit
- Uses the `AddressBookInput` for the address when adding a member

## How to test it

1. Open a space
2. Add a member
3. Observe suggestions from the address book
4. Select one
5. Observe the name is filled automatically

## Screenshots
Screenshot 2025-04-04 at 14 09 21
Screenshot 2025-04-04 at 14 09 32

## Checklist

### DIFF
--- a/apps/web/src/components/common/NameInput/index.tsx
+++ b/apps/web/src/components/common/NameInput/index.tsx
@@ -24,7 +24,7 @@ const NameInput = ({
         maxLength: 50,
         required,
         validate: (value) => {
-          if (value?.trim() === '') return 'Required'
+          if (value?.trim() === '' && required) return 'Required'
           return true
         },
       }}
--- a/apps/web/src/components/common/NameInput/index.tsx
+++ b/apps/web/src/components/common/NameInput/index.tsx
@@ -12,7 +12,7 @@ const NameInput = ({
   name: string
   required?: boolean
 }) => {
-  const { register, formState, control } = useFormContext() || {}
+  const { formState, control } = useFormContext() || {}
   // the name can be a path: e.g. "owner.3.name"
   const fieldError = get(formState.errors, name) as FieldError | undefined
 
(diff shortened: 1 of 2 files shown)


apps/web/src/components/common/NameInput/index.tsx
```diff
@@ -12,25 +12,32 @@ const NameInput = ({
...
-      label={<>{fieldError?.type === 'maxLength' ? 'Maximum 50 symbols' : fieldError?.message || props.label}</>}
-      error={Boolean(fieldError)}
-      fullWidth
-      required={required}
-      className={inputCss.input}
-      onKeyDown={(e) => e.stopPropagation()}
-      {...register(name, {
-        maxLength: 50,
-        required,
-        setValueAs: (value) => value.trim(),
```
usame-algan, line 32: This trim started to cause issues after I added the Controller. When typing something like `Safe name` it would change it to `Safename`. Not sure why it worked before but now it is possible again to add empty member names which we need to solve in a different way.

jmealy (commented):
What do you think about showing all address book entries, not just from the current chain? When you're in a space there is still a "current chain" but it doesn't really have any relevance to a space.

usame-algan:
> What do you think about showing all address book entries, not just from the current chain? When you're in a space there is still a "current chain" but it doesn't really have any relevance to a space.

I like the idea of accessing the "full" address book but ran into this problem where the same address is added on different chains and potentially with different names (see screenshot). In that case it looks more confusing IMO. I suggest to leave it out of scope in this PR and discuss it more in-depth with @TanyaEfremova as its also relevant for the space address book.

Screenshot 2025-04-07 at 10 59 32

PR #5629, opened 2025-04-08
fix(mobile): activate QR camera on user interaction

Jonathansoufer:
## What it solves

This PR implements fix the QRCode Scan flow while importing a Safe

Resolves #
From Notion
[x] - Update text in QR code scanner: You can find it in the web app’s sidebar on top.
[x] - Avoid camera to be prompt before user's interaction.

## How this PR fixes it
 Add a button on CRScanner so the user trigger a camera permission request, instead of automatically on render.

## How to test it
Install a fresh version and try import a Safe.

## Screenshots

Screenshot 2025-04-08 at 01 00 26
Screenshot 2025-04-08 at 01 00 14
Screenshot 2025-04-08 at 01 00 08

## Checklist

### DIFF
--- a/apps/mobile/src/components/Camera/QrCamera.tsx
+++ b/apps/mobile/src/components/Camera/QrCamera.tsx
@@ -1,6 +1,6 @@
 import { Camera, useCodeScanner, useCameraDevice, Code, CameraPermissionStatus } from 'react-native-vision-camera'
-import { View, Theme, H3 } from 'tamagui'
-import { Dimensions, Linking, Pressable, StyleSheet, useWindowDimensions } from 'react-native'
+import { View, Theme, H3, getTokenValue } from 'tamagui'
+import { Dimensions, Linking, Pressable, StyleSheet, useColorScheme, useWindowDimensions } from 'react-native'
 import React, { useCallback, useEffect } from 'react'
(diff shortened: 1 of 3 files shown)


apps/mobile/src/components/Camera/QrCamera.tsx
```diff
@@ -48,28 +51,67 @@ function CameraFooter(props: { footer: React.ReactNode }) {
...
       <View borderColor={denied ? '$error' : '$success'} style={[styles.corner, styles.bottomRight]} />
 
-      {denied && (
+      {/* Show button/icon only if permission denied, not granted, or granted but inactive */}
+      {(denied || !hasPermission || (hasPermission && !isCameraActive)) && (
         <View style={styles.deniedCameraContainer}>
-          <SafeFontIcon name={'camera'} size={40} color={'$error'} />
-          <SafeButton rounded secondary onPress={onPress} marginTop={20}>
-            Enable camera
+          <SafeFontIcon name={'camera'} size={40} color={denied ? '$error' : '$white50'} />
```
compojoom, line 95: You should get some color from the theme and not assign $white50 like this (you can use the helper getTokenValue for example). In dark mode this results to:
grafik

compojoom (approved):
👍

PR #5630, opened 2025-04-08
chore(mobile): add account badge on selector

Jonathansoufer:
## What it solves

Resolves #
From Notion
[x] - Asset screen: badge threshold missing New Mobile App

## How this PR fixes it
Added badge with signers count on account selector

## How to test it
Open the app and on the Home, check if the account selector shows a badge.

## Screenshots
Screenshot 2025-04-08 at 01 13 37

## Checklist


compojoom (commented):
The IdenticonWithBadge component will need adjustments for when it is displaying a very small badge.
grafik

Notice that the owners info is nearly centered on top of the identicon and how much space there is to the name. 

In your implementation
grafik

Note that the implementation of the IdenticonWithBadge is correct according to this screen
grafik

We should slightly modify it to be usable on both screens. Could be a theme variant. "Small" "large"

compojoom (approved):
👍

PR #5631, opened 2025-04-08
feat(mobile): import menu conditionally and UI fix

Jonathansoufer:
## What it solves
This PR implements a condition to avoid show import on item menu if the item was imported already.

Resolves #
https://github.com/safe-global/wallet-private-tasks/issues/99

## How this PR fixes it
Adds state and conditionals to prevent showing unnecessary menu action to the user.

## How to test it
Import and account and click on the 3dots.

## Screenshots
Screenshot 2025-04-08 at 01 24 33
Screenshot 2025-04-08 at 01 24 24
Screenshot 2025-04-08 at 01 23 58

## Checklist

### DIFF
--- a/apps/mobile/src/features/Signers/components/SignersList/SignersListItem.tsx
+++ b/apps/mobile/src/features/Signers/components/SignersList/SignersListItem.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { MenuView, NativeActionEvent } from '@react-native-menu/menu'
+import { MenuView, NativeActionEvent, MenuAction } from '@react-native-menu/menu'
 import { useSignersActions } from './hooks/useSignersActions'
 import { SafeFontIcon } from '@/src/components/SafeFontIcon'
 import { SignersCard } from '@/src/components/transactions-list/Card/SignersCard'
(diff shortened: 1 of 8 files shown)


apps/mobile/src/features/Signers/components/SignersList/SignersListItem.tsx
```diff
@@ -24,7 +24,15 @@ function SignersListItem({ item, index, signersGroup }: SignersListItemProps) {
   const colorScheme = useColorScheme()
   const contact = useAppSelector(selectContactByAddress(item.value))
   const local = useLocalSearchParams<{ safeAddress: string; chainId: string; import_safe: string }>()
-  const actions = useSignersActions()
+
+  // Check if the current item belongs to the 'My signers' section
+  const isMySigner = signersGroup.some(
+    (section) => section.title === 'My signers' && section.data.some((signer) => signer.value === item.value),
```
compojoom, line 30: Can we maybe add a key/id to the section instead? "My signers" were named something else at the begining and I can see it changing to something else. Plus, what if it is multilingual? Better to rely on a stable key than wonder why this condition is not working.

compojoom (approved):
nice!

PR #5634, opened 2025-04-08
Fix(Spaces): Remove tags from auth and verify queries

usame-algan:
DO NOT MERGE

## What it solves

Resolves TBD

## Checklist


packages/store/src/gateway/AUTO_GENERATED/auth.ts
```diff
@@ -8,15 +8,12 @@ const injectedRtkApi = api
     endpoints: (build) => ({
       authGetNonceV1: build.query<AuthGetNonceV1ApiResponse, AuthGetNonceV1ApiArg>({
         query: () => ({ url: `/v1/auth/nonce` }),
-        providesTags: ['auth'],
```
compojoom, line 11: How did you remove them @usame-algan ? This is auto-generated. We should not modify it directly

usame-algan: This was done for testing only. Unfortunately not the reason why it fails for Michael so no need to merge this.

PR #5635, opened 2025-04-08
feat(mobile): Use RTK infinite queries API

compojoom:
## What it solves
We had some weird bugs in our history. When switching between the safes we were not displaying the correct data. Now RTK handles this for us.

## How this PR fixes it
Uses the latest RTK infinite API

## How to test it
Load a safe, scroll down and you should see that all entries are loading. Switch to a different safe - it should display the correct history items.

### DIFF
--- a/packages/store/src/gateway/transactions.ts
+++ b/packages/store/src/gateway/transactions.ts
@@ -24,16 +24,29 @@ export const txHistoryApi = api.injectEndpoints({
             return undefined
           }
 
-          // Extract the cursor from the next URL
-          // The URL format is something like: /v1/chains/{chainId}/safes/{safeAddress}/transactions/history?cursor=XYZ
-          const nextUrl = lastPage.next
-          const cursor = nextUrl.split('cursor=')[1]
+          // Extract the cursor from the next URL using URLSearchParams
+          // This is more robust than using string.split when dealing with complex URLs
+          try {
+            // The URL might be a relative URL like /v1/chains/{chainId}/safes/{safeAddress}/transactions/history?cursor=XYZ&other=param
+            // or a full URL with hostname
+            const urlParts = lastPage.next.split('?')
+            if (urlParts.length < 2) {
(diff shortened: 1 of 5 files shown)


packages/store/src/gateway/transactions.ts
```diff
@@ -0,0 +1,56 @@
...
+          }
+
+          // Extract the cursor from the next URL
+          // The URL format is something like: /v1/chains/{chainId}/safes/{safeAddress}/transactions/history?cursor=XYZ
+          const nextUrl = lastPage.next
+          const cursor = nextUrl.split('cursor=')[1]
+
+          if (!cursor) {
+            return undefined
+          }
```
clovisdasilvaneto, line 34: unless you need to do type checking , since you return `cursor` at the end, there will be no difference here if you remove this return and let the `return cursor`...

compojoom: I need explicitly undefined to tell RTK that there are no more pages. If I have null or empty string this won't do it.

packages/store/src/gateway/transactions.ts
```diff
@@ -24,16 +24,29 @@ export const txHistoryApi = api.injectEndpoints({
...
+            const searchParams = new URLSearchParams(queryString)
+            const cursor = searchParams.get('cursor')
+
+            if (!cursor) {
+              return undefined
+            }
+
+            return cursor
+          } catch (error) {
+            console.error('Error extracting cursor from next URL:', error)
```
katspaugh, line 47: Don't we have a logging library?

katspaugh: Ah, it's a shared package.

packages/store/src/gateway/transactions.ts
clovisdasilvaneto: got it

clovisdasilvaneto (approved):

PR #5636, opened 2025-04-08
feat(mobile): add pull down to refresh history

compojoom:
## What it 
Ads pull down to refresh on the history page

Resolves
https://github.com/safe-global/wallet-private-tasks/issues/128

## How to test it
Open safe, navigate to the history, pull down - you would see a spinner and the tx will load from the server. (you could try sending a tx to the safe and then pulling down to see the newest tx at the top.

## Checklist


apps/mobile/src/features/TxHistory/TxHistory.container.tsx
```diff
@@ -31,8 +33,29 @@ export function TxHistoryContainer() {
...
+    try {
+      // Refetch will reset the data and start fresh with page 1
+      await refetch()
+    } catch (error) {
+      Logger.error('Error refreshing transaction history:', error)
+    } finally {
+      setIsRefreshing(false)
+    }
+  }, [refetch])
+
```
clovisdasilvaneto, line 48: It would be nice to move this logic to a hook, so we can combine the infinite scroll with pull-to-refresh as well 🤔 wdyt?

compojoom: Let's abstract this when we work on the pendingTx view and need to do the same there.

PR #5638, opened 2025-04-08
fix(web): lint issue

compojoom:
## What it solves
There was some empty space and lint was complaining


katspaugh (approved):
My bad, thank you!

PR #5659, opened 2025-04-09
feat(web): show 24h asset value change

### LINKED ISSUE #5650: [Portfolio] Show 24h value change
## Links
Epic on Notion: https://www.notion.so/safe-global/Proto-portfolio-tracking-displaying-Coingecko-24h-price-change-1c38180fe573803ea8ffccafe3ba8081

## What must be done
Add relative 24h change of asset value to items in Safe's balances.
They are already returned by the CGW and need to be displayed.

## Designs/sketches
https://www.figma.com/design/0KN3LnWKHT0r9JgCXXG9jk/DeFi-Positions?node-id=3265-2305&t=73y5phnAQLkD9UIJ-4

schmanu:
## What it solves
Resolves #5650 

## How this PR fixes it
- Displays relative 24h change in assets lists on assets page and dashboard

## How to test it
- Open a Safe on a production network for realistic prices and look at the Safe's assets

## Screenshots
Screenshot 2025-04-09 at 13 48 28
Screenshot 2025-04-09 at 13 48 16

## Checklist


apps/web/src/components/balances/AssetsTable/FiatChange.tsx
```diff
@@ -0,0 +1,43 @@
+import { Chip, SvgIcon, Typography } from '@mui/material'
+import { type Balance } from '@safe-global/store/gateway/AUTO_GENERATED/balances'
+import { formatPercentage } from '@safe-global/utils/utils/formatters'
+import ArrowDown from '@/public/images/balances/change-down.svg'
+import ArrowUp from '@/public/images/balances/change-up.svg'
+export const FiatChange = ({ balanceItem }: { balanceItem: Balance }) => {
```
iamacook, line 6:
```suggestion

export const FiatChange = ({ balanceItem }: { balanceItem: Balance }) => {
```

apps/web/src/components/balances/AssetsTable/FiatChange.tsx
```diff
@@ -0,0 +1,43 @@
...
+  if (!balanceItem.fiatBalance24hChange) {
+    return (
+      <Typography variant="caption" color="text.secondary" textAlign="center" display="block">
+        n/a
+      </Typography>
+    )
+  }
+
+  const changeAsNumber = Number(balanceItem.fiatBalance24hChange) / 100
+  const changeLabel = formatPercentage(Number(balanceItem.fiatBalance24hChange ?? 0) / 100)
```
iamacook, line 16:
```suggestion
  const changeLabel = formatPercentage(changeAsNumber)
```

apps/web/src/components/balances/AssetsTable/index.tsx
```diff
@@ -82,6 +91,12 @@ const headCells = [
     width: '20%',
     align: 'right',
   },
+  {
+    id: 'change',
+    label: '(24h change)',
+    width: '20%',
+    align: 'left',
+  },
```
iamacook, line 99: The total `width` is above 100%. However, it was like this before? Maybe we should try to reduce these values?

apps/web/src/components/balances/AssetsTable/index.tsx
```diff
@@ -157,6 +172,11 @@ const AssetsTable = ({
               collapsed: item.tokenInfo.address === hidingAsset,
               content: <FiatBalance balanceItem={item} />,
             },
+            change: {
+              rawValue: Number(item.fiatBalance24hChange ?? -9999999999999999),
```
iamacook, line 176: I think sorting will work as expected if we use `null`. If not, we should add a comment re. the arbitrary number. We could also use `-1 * Number.MAX_SAFE_INTEGER`. What do you think?

apps/web/src/components/balances/AssetsTable/index.tsx
```diff
@@ -82,6 +91,12 @@ const headCells = [
     width: '20%',
     align: 'right',
   },
+  {
+    id: 'change',
+    label: '(24h change)',
```
katspaugh, line 96: Why do we parenthesize this label?

schmanu: @IvaLukan should we keep the parenthesize? I sticked to the designs here.

apps/web/src/components/balances/AssetsTable/FiatChange.tsx
schmanu: Changed it. We should put this into our lint rules.

apps/web/src/components/balances/AssetsTable/index.tsx
IvaLukan: Since the 24h % is value related I wanted to keep them close. If we remove the parentheses we should increase the space between the two columns. Any strong opinion here?
Screenshot 2025-04-09 at 15 38 10

katspaugh: Ah, I see, thanks for clarifying.

apps/web/src/components/balances/AssetsTable/index.tsx
schmanu: The current comparator handled `null` the same as `0` so they would always get sorted between positive and negative relative changes.

I adjusted the comparator to always sort null / undefined last.

apps/web/src/components/balances/AssetsTable/index.tsx
schmanu: I redid the math and split it such that it adds up to 100%.

iamacook (approved):

PR #5662, opened 2025-04-09
Fix(PK Module): Use `signMessageSync` for personal_sign with PK module

usame-algan:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/162

This now uses the same method as on mobile for the PK module that we use for testing.

## How this PR fixes it

- In case the connected wallet is the PK module we pass the signable message directly via `personal_sign` instead of using `signer.signMessage`

## How to test it

1. Open spaces
2. Connect a PK module wallet
3. Sign in
4. Observe it working
5. Connect with any other wallet
6. Sign in
7. Observe it still working

## Checklist

### DIFF
--- a/apps/web/src/services/private-key-module/index.ts
+++ b/apps/web/src/services/private-key-module/index.ts
@@ -100,7 +100,7 @@ const PrivateKeyModule = (chainId: ChainInfo['chainId'], rpcUri: ChainInfo['rpcU
               },
 
               personal_sign: async ({ params }) => {
-                return wallet.signMessageSync(params[0])
+                return wallet.signMessage(params[0])
               },
 
               eth_signTypedData: async ({ params }) => {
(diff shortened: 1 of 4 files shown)


katspaugh (approved):

apps/web/src/services/private-key-module/index.ts
```diff
@@ -100,7 +100,7 @@ const PrivateKeyModule = (chainId: ChainInfo['chainId'], rpcUri: ChainInfo['rpcU
               },
 
               personal_sign: async ({ params }) => {
-                return wallet.signMessageSync(params[0])
+                return wallet.signMessage(params[0])
```
usame-algan, line 103: This calls signMessageSync

usame-algan:
@katspaugh I pushed a bigger fix after your review. The problem was that `signer.signMessage` was converting the string message into a byte array before passing it to personal_sign but we need to pass the raw string message there in order to sign it.

PR #5663, opened 2025-04-09
fix: add check for Lens to use ZKsync flows

dasanra:
## What it solves

Resolves #address calculation for Lens, which uses ZKsync

## How this PR fixes it
Adds Lens chains to the checks where we use ZKsync flows.

This is just a temporary solution to unlock QA, not an ideal solution as the chances of adding new ZKsync chains are higher over time. Already requested that a config parameter is added to the config service so we can check from ChainInfo if the chain uses ZKsync allowing dinamically to configure them in the future.


katspaugh (approved):
Thanks @dasanra!

PR #5671, opened 2025-04-10
Feat(Spaces): add dialog for editing contacts in the address book

jmealy:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/113#issue-2969298203

## How this PR fixes it
- adds a dialog to edit a contact. Address cannot be changed, only the networks and the contact name.
- If a network is deselected, the corresponding entry in the backend will be deleted. If a new network is selected, then an entry will be created. If the name changes but the network stays the same, then the entry in the backend will be updated.
- For now it will just log which entries are to be updated, deleted, or created.

## How to test it
- Click the edit icon in the context menu of a contact in the space address book
- You should be shown a dialog allowing you to change the networks and name
- On submit, it should console log for each call to the backend, along with the operation.

## Screenshots
image

## Checklist


apps/web/src/features/spaces/components/SpaceAddressBook/EditContactDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+      const selectedChainIds = data.networks.map((network) => network.chainId)
+      const originalChainIds = entry.networks.map((network) => network.chainId)
+
+      // Determine which networks need to be created, updated, or deleted
+      const toCreate = selectedChainIds.filter((chainId) => !originalChainIds.includes(chainId))
+      const toDelete = originalChainIds.filter((chainId) => !selectedChainIds.includes(chainId))
+      // Only update if the name has changed
+      const nameHasChanged = data.name !== entry.name
+      const toUpdate = nameHasChanged ? selectedChainIds.filter((chainId) => originalChainIds.includes(chainId)) : []
```
usame-algan, line 98: It feels a bit complex to combine these two actions into one form i.e. if a user deselects networks or adds more networks and also changes the name we would have to delete/add some entries while updating the name in others. IMO it would be easier to only allow changing the name. If a multi-network entry is edited we would update each entry with that new name and if its a contact on a single network we just update one name.

usame-algan: After our discussion today we decided to keep it this way so it will be possible to add/remove networks when editing an address and there will be one endpoint where we just have to send all selected networks and it will do the upsertion so I think this good.

apps/web/src/features/spaces/components/SpaceAddressBook/EditContactDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+    try {
+      setIsSubmitting(true)
+      trackEvent({ ...SPACE_EVENTS.ADD_ADDRESS_SUBMIT })
+
+      const selectedChainIds = data.networks.map((network) => network.chainId)
+      const originalChainIds = entry.networks.map((network) => network.chainId)
+
+      // Determine which networks need to be created, updated, or deleted
+      const toCreate = selectedChainIds.filter((chainId) => !originalChainIds.includes(chainId))
+      const toDelete = originalChainIds.filter((chainId) => !selectedChainIds.includes(chainId))
```
usame-algan, line 95: It looks like this will be handled by the CGW where we will have one upsert endpoint that we send all the selected chainIds to and it will then decide what to delete/create/edit.

usame-algan (approved):
Looks good 👍

PR #5674, opened 2025-04-10
Fix(Spaces): Use EthHashInfo for SafeList items to support copy

usame-algan:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/169
Resolves https://github.com/safe-global/wallet-private-tasks/issues/170

## How this PR fixes it

- Uses `EthHashInfo` within `SingleAccountItem` and `MultiAccountItem`
- Adds a more useful error message when reaching the space creation limit

## How to test it

1. Open a space
2. Add safes to that space
3. Hover over the address in the list
4. Observe a "Copy to clipboard" tooltip
5. Click the address
6. Observe the address is copied and the user is not navigated to that safe
7. Click anywhere else
8. The user is navigated to that safe
9. Accounts page and safe sidebar will also have this change
10. Add a safe address (multi-chain and single chain) to your address book
11. Observe the name shows up only once in the list

## Screenshots
Screenshot 2025-04-10 at 15 34 12
Screenshot 2025-04-10 at 15 34 27

## Checklist

### DIFF
--- a/apps/web/src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
+++ b/apps/web/src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
@@ -97,7 +97,7 @@ const SrcEthHashInfo = ({
 
       <Box overflow="hidden" className={onlyName ? css.inline : undefined} gap={0.5}>
         {name && (
-          <Box title={name} display="flex" alignItems="center" gap={0.5}>
+          <Box title={name} className="ethHashInfo-name" display="flex" alignItems="center" gap={0.5}>
             <Box overflow="hidden" textOverflow="ellipsis">
               {name}
             </Box>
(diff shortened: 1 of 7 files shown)


francovenica:
Some issues:

1 - The copy address works fine, but if you copy an multichain address it copies it without prefix for obvious reasons, but copying it from a single network safe it DOES copies with prefix. I don't know the perfect solution, but I think it should be all with prefix or none with prefix

2 - The copy should NOT be on the safe list in the sidebar. It was taken out on purpose because for a mobile viewport there is already little space, so is more likely that the user will end up copying the address instead of entering into the safe he wants.

3 - The message about the limit of Spaces is fine, but there should be also a message in the form indicating the limit.
For example in add accounts we have a message saying "You can add any Safe Account to your Space. This is currently limited to 10 Safe Accounts."

francovenica:
Ok, the copy is disabled for mobile

I get the prefix argument. Let's leave it as it is

I saw in the other ticket the request to change the message from the backend side so that will be fix later.

LGTM

jmealy:
> Ok, the copy is disabled for mobile
> 
> I get the prefix argument. Let's leave it as it is
> 
> I saw in the other ticket the request to change the message from the backend side so that will be fix later.
> 
> LGTM

Looks like there are conflicts, I will merge dev into the epic branch and it should work after

PR #5682, opened 2025-04-11
Fix(Proposer): Hide signer form for proposers and select proposer address

### LINKED ISSUE #5679: [Web] Creating transaction as proposer not possible with nested safe owners
## Bug description

If a proposer tries to create a transaction in a safe where there are nested safe owners there is no option to choose the proposer in the dropdown

## Steps to reproduce

1. Open a safe with a nested safe owner and a proposer
2. Create a transaction as a proposer
3. Observe the proposer wallet is not an option in the selector

## Screenshots

Image

usame-algan:
## What it solves

Resolves #5679 

## How this PR fixes it

- Doesn't render the `SignerForm` when a proposer is creating a transaction
- Selects the wallet address in the `SignerForm` if a proposer is creating a transaction

## How to test it

1. Open a safe that has a nested safe owner and a proposer
2. Connect with the proposer
3. Create a transaction
4. Observe the Signer form is not visible
5. Observe that the proposer creates the transaction
6. Connect with an owner and do the same
7. Observe the Signer form is visible
8. Connect with an owner that is also a proposer and do the same
9. Observe the Signer form is visible

## Checklist


katspaugh:
@usame-algan there's a merge conflict.

schmanu (approved):

katspaugh (approved):

francovenica:
Question/Issue.
If you have a EOA 1 that owns SafeA and then a SafeB that is owned by safeA, shouldn't that EOA 1 user, even him being a proposer, be able to sign/execute the transaction through SafeA?
What I said is that, EOA 1 was able to sign/execute tx just fine in SafeB through SafeA, but when it was added as a proposer of this SafeB, all of the sudden the only thing he can do is propose.
This seems like a downgrade to me, or at least he should be able to choose if he wants to propose or sign through SafeA

usame-algan:
> This seems like a downgrade to me, or at least he should be able to choose if he wants to propose or sign through SafeA

You are right. I changed it so that its possible to select both options in case the EOA is owner of the parent safe and a proposer.

francovenica:
We have new issues with this.

The user now can select his EOA to propose a tx, which is correct, but if the tx is already proposed he should not be able to select himself anymore, only the safe he owns should be listed in this case to sign or execute the proposed tx

image

Another issue (minor in my opinion) is the tooltip and title of the selector that says "Execute with" or "Sign with", when the EOA selected cannot do either.
image

usame-algan:
> The user now can select his EOA to propose a tx, which is correct, but if the tx is already proposed he should not be able to select himself anymore

Good find! This makes sense to me as it doesn't add anything (signature) to the already proposed transaction. In that case we should disable this option or not show it at all. I will take a look at it. Overall this logic is getting more complex so I will make sure to cover it with tests as well.

usame-algan:
@francovenica I fixed it now to only show the proposer wallet if its a tx creation.

francovenica:
Nice. LGTM

PR #5688, opened 2025-04-11
feat(Signing UX): show combo button also on receipt step

tmjssz:
Show the combo button also on the receipt step and don't navigate to the previous step on sign/execute error.


katspaugh:
Tested works well!

PR #5701, opened 2025-04-14
Feat(Spaces): Add import address book card to the dashboard

jmealy:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/183#issue-2992973175

## How this PR fixes it
Adds a card for importing the address book on the dashboard

## Screenshots
image

## Checklist


usame-algan (approved):
Thanks! Agree that we can remove the last card that teases new features coming. I noticed on Figma also that the icons for address book and members changed to blue but not sure if its final yet. Not a blocker.

Issue #5416, opened 2025-03-20
[Signing UX v2] Epic


katspaugh:
A parent ticket for the next iteration of Signing UX improvements. See [v1](https://github.com/safe-global/safe-wallet-monorepo/issues/5238).

## Links

[Designs](https://www.figma.com/design/cMDccuR9aZFqAiBd4dKsKg/Transaction-flow?node-id=1091-7&p=f&t=Raunnkog5vhJCJvi-0)
[Epic on Notion](https://www.notion.so/safe-global/Signing-UX-security-improvements-1ab8180fe573806aa382f7dddb3c7d49?pvs=4)

mike10ca:
Please do not merge to dev until we run automated regression tests. @francovenica @liliya-soroka fyi

katspaugh:
@liliya-soroka @francovenica it's merged into dev and can be tested on there.

francovenica:
Question. For that new flow (where the button as a display to select the options) when in the review step you decide to execute a tx you click the button and as you reach the receipt step the action to sign is already triggered, leaving the button in a loading status. Is this intended?
I think is confusing. In MM is ok cuz the popup shows up fine, but for native ledger you gotta see that the device is asking for the signing, you reach to that point of "What I do now?" until you realize
Also for GNO network you sitill see the "Sponsored" selector that you can still click. I think this is the only network (beyond Sepolia) where this shows up, but I think that if we are going to leave this flow, we should remove that block

francovenica:
I tried nested safes and the data shown in the final tx is the same as the tx executed in the parent safe instead of the one shown during the signiging of the tx in the child safe

In this gif a send funds was executed in the parent safe, so in the one in the gif (the child safe) first you see in the receipt all the data of the signature the AOE will sign, but then in the tx details you see the same data you see in the parent safe, which is just the "send funds" tx

![Image](https://github.com/user-attachments/assets/df35176e-f0c4-4f13-9f1f-156f766aa538)

PR #5741, opened 2025-04-22
fix(Signing UX): revert to show combo button only on final receipt step

tmjssz:
## What it solves

Show the sign/execute/batch submit button on the Receipt step only and remove it from the transaction review step.
This is to avoid confusion in the signing process (see [this comment](https://github.com/safe-global/safe-wallet-monorepo/issues/5416#issuecomment-2819974099)).

## How to test it

1. Open update threshold transaction flow.
2. Continue to the transaction review step
3. Observe a "Continue" button instead of the combo button

## Screenshots

Screenshot 2025-04-22 at 15 12 14

## Checklist

### DIFF
--- a/apps/web/src/components/transactions/TxDetails/Summary/index.tsx
+++ b/apps/web/src/components/transactions/TxDetails/Summary/index.tsx
@@ -53,18 +53,22 @@ const Summary = ({ safeTxData, txData, txInfo, txDetails }: Props): ReactElement
 
       {submittedAt && (
         <TxDataRow datatestid="tx-created-at" title="Created">
-          <Typography variant="body2">{dateString(submittedAt)}</Typography>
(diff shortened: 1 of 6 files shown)


katspaugh (approved):
Thanks!

francovenica:
Question. This ticket was merged a while back in dev, but in both dev and this PR the combo button is still there. Was the button added back at some point?

image

cc @katspaugh @tmjssz

tmjssz:
@francovenica The combo button wasn’t removed entirely — it was just moved to the final step. Previously, it was shown on the transaction review step (the one before the final one), but with this change, it’s now displayed on the final receipt step instead (like on your screenshot).

francovenica:
Noted

PR #5745, opened 2025-04-23
Analytics(Signing UX): track time spent on receipt

katspaugh:
## What it solves

* Add a custom event on the Continue button click
* Track time spent on the Receipt screen

## Continue to receipt
Screenshot 2025-05-15 at 08 33 13

## Time spent on receipt
Screenshot 2025-05-15 at 08 32 13


apps/web/src/components/tx/ReviewTransactionV2/ReviewTransactionContent.tsx
```diff
@@ -47,6 +48,11 @@ export const ReviewTransactionContent = ({
   const isApproval = readableApprovals && readableApprovals.length > 0
   const isCounterfactualSafe = useIsCounterfactualSafe()
 
+  const onContinueClick = useCallback(() => {
```
tmjssz, line 51: Should we also add this to `apps/web/src/components/tx/ReviewTransaction/ReviewTransactionContent.tsx`, at least temporarily until we remove that component as part of the refactor? Or do we not care since it won’t go live anyway before the refactor is complete?

apps/web/src/components/tx/ReviewTransactionV2/ReviewTransactionContent.tsx
```diff
@@ -47,6 +48,11 @@ export const ReviewTransactionContent = ({
...
+    trackEvent(MODALS_EVENTS.CONTINUE_CLICKED)
```
tmjssz, line 52: Isn’t this duplicating the “Next click” event that’s already tracked by `useTxStepper`?

apps/web/src/components/tx/ReviewTransactionV2/ReviewTransactionContent.tsx
katspaugh: Sure, I'll add it there too. We'll go live asap, porting other flows isn't a blocker.

apps/web/src/components/tx/ReviewTransactionV2/ReviewTransactionContent.tsx
katspaugh: Yes, but "Next click" is fired on all steps, and this one only here.

tmjssz (approved):

katspaugh:
"Next click" is fired on all steps, and this one only here @liliya-soroka

PR #5750, opened 2025-04-24
Fix(mobile): truncate the amount if it is bigger than expected

clovisdasilvaneto:
## What it solves
Resolves https://github.com/safe-global/wallet-private-tasks/issues/176

## How this PR fixes it
It truncates the amount of the token in case it is bigger than expected.

## How to test it
- Create a transaction with a broken token amount, like: 0.00000000002 ETH
- Go to pending transactions or transaction history screen (in case you already have some on your wallet)
- Scroll to find the transaction

**Excpected result**
- The tx amount should be truncated, like: -0.00000000...ETH

## Screenshots
Screenshot 2025-04-24 at 10 03 32

## Checklist


katspaugh:
@clovisdasilvaneto please solve the merge conflicts.

PR #5753, opened 2025-04-24
Tests(web): add proposedByDelegate to mock transaction details

tmjssz:
Fixes TS errors in the test files that I noticed when running the linter.


usame-algan (approved):
Thanks!

PR #5754, opened 2025-04-24
Feat(Swaps): swap parameters for KyberSwap

katspaugh:
## What it solves

Add URL params to the Swap app fallback so that a token can be preselected.


usame-algan:
Code looks good but how to test this? When I navigate to the swaps page on a BNB safe for example I can see it redirects me to the kyberswap.com safe app but I don't see any additional params in the URL. Is this expected?

usame-algan (approved):

PR #5768, opened 2025-04-28
Feat(Spaces): Implement deletion of address book items

usame-algan:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/193

## How this PR fixes it

- Integrates the DELETE endpoint in the delete dialog

## How to test it

1. Open a space with address book items
2. Delete one
3. Observe a success message
4. Observe that the item is gone from the list

## Checklist

### DIFF
--- a/apps/web/src/features/spaces/components/SpaceAddressBook/DeleteContactDialog.tsx
+++ b/apps/web/src/features/spaces/components/SpaceAddressBook/DeleteContactDialog.tsx
@@ -7,6 +7,12 @@ import ModalDialog from '@/components/common/ModalDialog'
 import { trackEvent } from '@/services/analytics'
 import { SPACE_EVENTS } from '@/services/analytics/events/spaces'
 import { ChainIndicatorList } from '@/features/multichain/components/SignerSetupWarning/InconsistentSignerSetupWarning'
+import { useAddressBooksDeleteByAddressV1Mutation } from '@safe-global/store/gateway/AUTO_GENERATED/spaces'
+import { useCurrentSpaceId } from '@/features/spaces/hooks/useCurrentSpaceId'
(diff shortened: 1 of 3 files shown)


apps/web/src/features/spaces/components/SpaceAddressBook/DeleteContactDialog.tsx
```diff
@@ -16,15 +22,38 @@ type DeleteContactDialogProps = {
...
-      // TODO: Implement the API call to remove the address book entry
-      //   await Promise.all(entryIds.map((id) => removeAddressBookEntry(id)))
-      console.log('remove address book entries', name, address, networks)
+      setIsSubmitting(true)
       trackEvent({ ...SPACE_EVENTS.REMOVE_ADDRESS_SUBMIT })
+      const response = await deleteEntry({ spaceId: Number(spaceId), address })
+
+      if (response.error) {
+        setError('Something went wrong deleting the contact. Please try again.')
+        return
```
tmjssz, line 41: I think we should set isSubmitting to false here. Alternatively, we could throw an error in this case to trigger the catch block, which would handle both setError and setIsSubmitting. WDYT?

usame-algan: I think because it is inside the finally block it will always be reached even if we return early here but I also see the duplication. Ideally we can show a more descriptive error if `result.error` and show a generic error if the whole request throws but currently this is missing on the CGW.

tmjssz: True, you're right! I missed that the finally block will be executed despite the early return.

PR #5770, opened 2025-04-28
Chore: git ignore types/contracts

katspaugh:
Contract types were not ignored in the shared packages folder and they were committed accidentally.


.gitignore
```diff
@@ -7,6 +7,7 @@
 # testing
 /coverage
 /packages/utils/coverage
+/packages/utils/src/types/contracts
```
compojoom, line 10: this is being added to the wrong block (#testing)

PR #5776, opened 2025-04-29
fix(web): decode multisend txs

### LINKED ISSUE #5775: [Web] MultiSend txs are not decoded anymore
## Bug description
On our current `dev` environment we lack the multisend tx decoding.

## Environment

- Browser: Chrome
- Wallet: MetaMask
- Chain: Ethereum mainnet
- dev environment

## Steps to reproduce

1.  Create any multisend tx
2. Witness that you only see the raw Hex data in the tx preview / confirmation screen.

## Expected result
- We see each inner tx of the multisend decoded

## Obtained result
Only hex data

schmanu:
## What it solves

Resolves #5775

## How this PR fixes it
Adds new case in `ConfirmationView` for Multisends and renders the compact `Multisend` component

## How to test it
See issue

## Checklist


iamacook (approved):

francovenica:
I see the twap order with no decoding.

This is a comparison between dev and this PR

Dev:
image

This PR:
image

francovenica:
The fix Looks good. leaving a snapshot how it looks now
The rest of the tx looks good as well

image

schmanu:
We found an issue with transactions in the queue / history that now show the multisends decoded twice.

I will re-add the showMultisend parameter to the Summary component to address this.

francovenica:
What it works
The issue about not showing the data that was discussed in slack was solved, and the tx now shows the "Data" field.
The duplicated decoding was fixed, only shows one
The decoding not showing for some tx also was fixed

Question/issue:
I don't know if this is part of this ticket or not, so let me know if it is not in the scope.
In the same way we don't show twice the method of the contract being called in the tx flow, shouldn't we do the same for the tx in queue?

Tx flow: showing this only once:
image

Tx in queue: is shown twice
image

This is the tx, must be seen in stg cgw
https://fix_multisend_decoding--walletweb.review.5afe.dev/transactions/tx?safe=eth:0x8675B754342754A30A2AeF474D114d8460bca19b&id=multisig_0x8675B754342754A30A2AeF474D114d8460bca19b_0x9fd741c6000123af42fba91746258a3cda34135f43ee60c53a8850b63f92d572

schmanu:
> Question/issue:
> I don't know if this is part of this ticket or not, so let me know if it is not in the scope.
> In the same way we don't show twice the method of the contract being called in the tx flow, shouldn't we do the same for the tx in queue?

@francovenica this is another unrelated regression which is also happening in the current dev environment without the changes from this PR.
As it is not about the decoding of multiSends which works fine in the queue I would consider it out of scope for this PR.

francovenica:
Ok. Then this PR LGTM

PR #5782, opened 2025-04-30
fix(web): use recommended nonce when editing the nonce is allowed

schmanu:
## What it solves
The recommended nonce is not used when creating a new transaction

## How this PR fixes it
Fixes the condition such that the `SafeTxProvider` uses the `recommendedNonce` when the form can be edited.

## How to test it
- Create a new transaction in a Safe that has queued transactions
- Observe that the nonce is chosen correctly to be at the end of the queue

## Checklist


iamacook (approved):

francovenica:
LGTM

PR #5789, opened 2025-05-02
fix: remove localState from useTheme hook

clovisdasilvaneto:
## What it solves

Resolves #XYZ (please replace with actual issue number)

## How this PR fixes it

This PR fixes a theme management issue in the mobile app by removing local state from the `useTheme` hook. Instead of maintaining a separate local state for the current theme, the hook now directly calculates and returns the current theme based on the theme preference and system color scheme.

Key changes:

- Removed `useState` hook and `currentTheme` local state from `useTheme`
- Modified the return value to calculate the theme dynamically
- Simplified the AppState change handler to only refresh the theme preference
- Removed unnecessary imports for `useState` and `ColorSchemeName`

This change eliminates potential state synchronization issues between the app's stored preferences and the displayed theme.

## How to test it

1. Open the app and navigate to the theme settings
2. Toggle between different theme options (Light, Dark, Auto)
3. Put the app in the background and then back to foreground
4. Verify that the theme correctly applies and persists after app state changes
5. If using "System" preference, verify that the app respects the device theme changes

## Checklist

### DIFF
--- a/apps/mobile/src/theme/hooks/useTheme.tsx
+++ b/apps/mobile/src/theme/hooks/useTheme.tsx
@@ -1,5 +1,5 @@
-import { useCallback, useEffect, useState } from 'react'
-import { AppState, ColorSchemeName, useColorScheme } from 'react-native'
+import { useCallback, useEffect } from 'react'
+import { AppState, useColorScheme } from 'react-native'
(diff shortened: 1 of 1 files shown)


compojoom (approved):
Tested on emulator. Works awesome! 👍

PR #5791, opened 2025-05-02
refactor(SetupNestedSafeFlow): migrate to new TxFlow logic

tmjssz:
Refactors SetupNestedSafeFlow to use the new logic.

Also updates the ComboSubmit component to ensure all props are properly passed through and that `onSubmitSuccess` is triggered as expected.

### DIFF
--- a/apps/web/src/components/tx-flow/flows/CreateNestedSafe/index.tsx
+++ b/apps/web/src/components/tx-flow/flows/CreateNestedSafe/index.tsx
@@ -1,75 +1,66 @@
-import { useCallback, useMemo, useState, type ReactElement } from 'react'
-
+import { useCallback, useContext, useMemo, useState, type ReactElement } from 'react'
 import NestedSafeIcon from '@/public/images/sidebar/nested-safes-icon.svg'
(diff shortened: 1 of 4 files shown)


apps/web/src/components/tx-flow/flows/CreateNestedSafe/index.tsx
```diff
@@ -1,75 +1,66 @@
...
+        return
+      }
+      dispatch(
+        upsertAddressBookEntries({
+          chainIds: [safe.chainId],
+          address: predictedSafeAddress,
+          name: args?.data?.name ?? '',
+        }),
+      )
+    },
```
usame-algan, line 49: I wonder if we could do this inside the onSubmit of `ReviewNestedSafe` to reduce complexity here. How would the behaviour change?

tmjssz: It would upsert the address book entry already one step earlier, instead of on the final one when the transaction is signed/executed. I think that would not be ideal, since the user could still cancel the nested Safe creation at that point, which would leave an address book entry for something that was never finalized.

usame-algan (approved):

PR #5827, opened 2025-05-06
fix(TxFlow): use correct ReviewTransaction component

tmjssz:
Migrate ReviewRemoveGuard, ReviewRemoveModule, and ReviewRemoveOwner to use ReviewTransactionV2 and pass down children.


katspaugh (approved):
🙏

PR #5829, opened 2025-05-06
fix(mobile): wrong placeholder on 403 error

compojoom:
## What it solves
If the image loading fails, we don’t show the fallback image. This is a quirk of tamagui (I think), so adding some local state to keep track of things and display the fallback if image errors out.

Resolves https://github.com/safe-global/wallet-private-tasks/issues/104

## How to test it

1. add real Ethereum safe 0x84443F61efc60D10DA9F9a2398980CD5748394BB
2. go to the history
3. See the fallback icon instead of a white/dark circle

## Screenshots
before:
grafik
after:
grafik

## Checklist


clovisdasilvaneto (approved):
Nice, I think there might be some other places where we want to use the SafeAvatar component

compojoom:
Let's merge for now and updated the other places on a case-by-case bases.

PR #5837, opened 2025-05-07
fix(mobile): crash when there are no active signers

compojoom:
## What it solves
The recent fix for the "ready to execute view" caused a crash when there is no active signer https://github.com/safe-global/safe-wallet-monorepo/pull/5828

## How this PR fixes it
We now accept the fact that the user might not have a signer imported and don't render the signer selector component that was causing the crash. 

## How to test it
1. Import a safe with pending txs
2. Make sure that you have no signers imported for it
3. Navigate to a pending tx
4. Observe no crash

## Checklist

### DIFF
--- a/apps/mobile/src/features/ConfirmTx/ConfirmTx.container.tsx
+++ b/apps/mobile/src/features/ConfirmTx/ConfirmTx.container.tsx
@@ -57,7 +57,7 @@ function ConfirmTxContainer() {
         <TransactionInfo txId={txId} detailedExecutionInfo={detailedExecutionInfo} />
       </ScrollView>
 
-      <View paddingTop="$1" backgroundColor="$background">
+      <View paddingTop="$1">
         <ConfirmTxForm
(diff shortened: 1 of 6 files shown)


clovisdasilvaneto (approved):

apps/mobile/src/features/ConfirmTx/ConfirmTx.container.tsx
```diff
@@ -57,7 +57,7 @@ function ConfirmTxContainer() {
         <TransactionInfo txId={txId} detailedExecutionInfo={detailedExecutionInfo} />
       </ScrollView>
 
-      <View paddingTop="$1" backgroundColor="$background">
+      <View paddingTop="$1">
```
clovisdasilvaneto, line 60: Did you check this in the light theme?

compojoom: yes, in my opinion it looks good.

PR #5842, opened 2025-05-07
Fix(mobile): import safe refactor

clovisdasilvaneto:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/218 and https://github.com/safe-global/wallet-private-tasks/issues/86

## How this PR fixes it
It changes the way of how we're using react-hook-form on this screen to be according to its docs. Previously we were having conflicts with react-hook-form internal state due to validating/request the address outside the zod schema and the Controller component.

## How to test it
- Get an cleaned app
- Try to import a valid safe: button should be green and the chains should appear
- Try to import a EOA address: button should be disabled and the error message should appears
- Try to import a valid safe again: button should be green and the chains should appear again.

## Screenshots
Screenshot 2025-05-07 at 10 19 40

## Checklist

### DIFF
--- a/apps/mobile/src/app/(import-accounts)/form.tsx
+++ b/apps/mobile/src/app/(import-accounts)/form.tsx
@@ -1,11 +1,27 @@
 import React from 'react'
 import { ImportAccountFormContainer } from '@/src/features/ImportReadOnly'
 import { View } from 'tamagui'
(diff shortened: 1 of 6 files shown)


compojoom (approved):
good work @clovisdasilvaneto !

PR #5850, opened 2025-05-09
fix: redirect user to the import signer flow if he cames from import safe flow

clovisdasilvaneto:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/130

## How this PR fixes it
Now we check in the item click, if the user is coming from the import safe flow or not. If the `import_safe` parameter is passed and the signer that we clicked is not a imported signer, we redirect him to the import signer flow.

## How to test it
Check https://github.com/safe-global/wallet-private-tasks/issues/130

## Video

https://github.com/user-attachments/assets/e3875d99-3068-410d-8e39-64881f945d5f

## Checklist


compojoom (approved):
nice!

PR #5851, opened 2025-05-09
fix(Signing UX): hide sign option if transaction is fully signed

### LINKED ISSUE #5847: [Signing UX v2] Combo button - disabled tooltip overlaps the dropdown menu
## Bug description

The combo button dropdown menu is overlapped by the tooltip when no signer is connected.

## Environment

- Browser: Chrome
- Wallet: MetaMask
- Chain: Ethereum mainnet

## Steps to reproduce

1.  Go through the transaction flow until you reach the final receipt step
2. Disconnect the wallet or switch to an EOA that is not a signer of the Safe
3. Open the combo button drop down menu
4. Observe that a tooltip overlaps the dropdown menu when hovering the button

## Expected result

Multiple options, e.g.:
- Disable the dropdown arrow as well, so that the menu can't be opened in case all of the combo button options are disabled.
- Show the tooltip on top of the button, not below where the dropdown options would appear.

## Obtained result

Tooltip overlaps the dropdown items.

## Screenshots

Image

tmjssz:
## What it solves

Hides the Sign option when a transaction is already fully signed.

This also fixes #5847 where the combo button dropdown menu was overlapped by the tooltip when no signer was connected. 

## How to test it

1. Create a transaction and fully sign it without executing
2. Go to the receipt step with the execute button
3. Switch the connected EOA to a wallet that is not an owner of the Safe
4. Observe that only the "Execute" button is visible, instead of the combo button

## Before

Screenshot 2025-05-09 at 15 07 18

## Now

Screenshot 2025-05-09 at 15 36 20

## Checklist

### DIFF
--- a/apps/web/src/components/tx-flow/actions/Sign/index.tsx
+++ b/apps/web/src/components/tx-flow/actions/Sign/index.tsx
@@ -6,6 +6,7 @@ import useIsCounterfactualSafe from '@/features/counterfactual/hooks/useIsCounte
 import { type SlotComponentProps, SlotName, withSlot } from '../../slots'
 import type { SubmitCallback } from '../../TxFlow'
 import { useAlreadySigned } from '@/components/tx/SignOrExecuteForm/hooks'
+import useSafeInfo from '@/hooks/useSafeInfo'
 
 export const Sign = ({
   onSubmit,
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
👍

francovenica:
LGTM
Looks the same as the snapshot. 
Users that are not owners but can execute the tx (since is fully signed) don't see the "Sign" action grayed out, it sees the Execute right away.

PR #5860, opened 2025-05-12
fix(web): hide queue bar and change bridge icon/position

iamacook:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/219

## How this PR fixes it

This updates the bridge-related UI by:

- Hiding the queue bar on the `/bridge` route.
- Changing the sidebar icon.
- Moving the sidebar element between "Swap" and "Stake".

## How to test it

Observe a new, correctly located (thicker) icon in the sidebar that, when opened, has no queue bar.

## Screenshots

grafik

grafik

## Checklist


apps/web/src/components/safe-apps/AppFrame/index.tsx
```diff
@@ -30,8 +30,10 @@ import { useSanctionedAddress } from '@/hooks/useSanctionedAddress'
 import BlockedAddress from '@/components/common/BlockedAddress'
 import { isSafePassApp } from '@/features/walletconnect/services/utils'
 import { BRAND_NAME } from '@/config/constants'
+import { AppRoutes } from '@/config/routes'
 
 const UNKNOWN_APP_NAME = 'Unknown Safe App'
+const NON_QUEUE_BAR_ROUTES = [AppRoutes.bridge]
```
katspaugh, line 36: Why is this needed for Bridging but not for Swaps? Are they using a different iframe component?

iamacook: > Are they using a different iframe component?

Yes, we [use their `CowSwapWidget` and access the `iframe` element](https://github.com/safe-global/safe-wallet-monorepo/blob/fix/bridge-ui/apps/web/src/features/swap/index.tsx#L287-L313) instead of embedding it ourselves in an `AppFrame`.

katspaugh: What about Staking? I see it also uses the AppFrame. Does it currently show the queue bar?

iamacook: Yes.

grafik

katspaugh: Should we unify this behavior and hide the queue based on the `isNativeEmbed` prop?

usame-algan: Agree, I don't think we deliberately decided to show the queue bar in any of these pages so hiding it if `isNativeEmbed` makes sense to me as well.

usame-algan (approved):

apps/web/src/components/safe-apps/AppFrame/index.tsx
iamacook: Changed in 04c2dde433709c61291082037e0bb83492eb25d6.

katspaugh (approved):
👍

PR #5864, opened 2025-05-12
Feat(web): don't notify about 1.4.1 upgrades

katspaugh:
## What it solves

The Safe version upgrade from 1.3.0 to 1.4.1 isn't critical, so we shouldn't show a notification for it. This will allow us to enable 1.4.1 in the config service.

## How it solves it

* Skip notifications for 1.3.0 Safes
* Use the latest Safe version specified in the config service for both Safe upgrades and new Safe creations (previously it was only used for upgrades).


usame-algan (approved):
Code looks good 👍

PR #5868, opened 2025-05-13
Fix(Balances): refresh cache when switching Safes

katspaugh:
## What it solves

There was a bug after migrating to RTK Query for balances:

* Open a Safe, go to Assets
* Switch to another Safe
* Previous balances are still shown until new balances are fetched

We were using the wrong response field.


compojoom (approved):

francovenica:
LGTM

PR #5870, opened 2025-05-13
Feat(Spaces): Plug space address book into existing UI

usame-algan:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/195

## How this PR fixes it

- Creates a new hook `useAllMergedAddressBooks` that merges the space address book with the local address book (for the current network)
- Creates a new hook `useAddressBookItem` that returns a specific item from the above hook
- Uses the newly created hooks inside `AddressBookInput` and `EthHashInfo` among other places where names need to be shown
- If the user is not signed in or no space has been visited or the current safe doesn't belong to a space fall back to the local address book

## How to test it

1. Open a space with an address book and safe accounts attached to it
2. Sign in and navigate to one of the safes
3. Observe the names shown either come from the local address book or space address book
4. Observe a different icon next to the name depending on its source
5. Create a new transaction
6. Click into the recipient field
7. Observe entries from both sources
8. Disconnect your wallet
9. Observe a fallback to the local address book
10. Navigate to the address book within a safe
11. Observe everything is still the same
12. Observe a CTA that leads to the space address book in case you are an admin

## Screenshots

Screenshot 2025-05-13 at 12 24 04
Screenshot 2025-05-13 at 12 24 11
Screenshot 2025-05-15 at 10 44 58

## Checklist

### DIFF
--- a/apps/web/src/components/common/AddressBookInput/index.tsx
+++ b/apps/web/src/components/common/AddressBookInput/index.tsx
@@ -27,10 +27,14 @@ const AddressBookInput = ({ name, canAdd, ...props }: AddressInputProps & { canA
   const { setValue, control } = useFormContext()
   const addressValue = useWatch({ name, control })
 
-  const allAddressBookEntries = mergedAddressBook.map((entry) => ({
-    label: entry.address,
-    name: entry.name,
-  }))
+  const allAddressBookEntries = useMemo(
+    () =>
+      mergedAddressBook.map((entry) => ({
+        label: entry.address,
+        name: entry.name,
+      })),
+    [mergedAddressBook],
+  )
 
   const hasVisibleOptions = useMemo(
     () => !!allAddressBookEntries.filter((entry) => entry.label.includes(addressValue)).length,
--- a/apps/web/src/components/common/AddressBookInput/index.tsx
+++ b/apps/web/src/components/common/AddressBookInput/index.tsx
@@ -2,7 +2,6 @@ import { type ReactElement, useState, useMemo } from 'react'
 import { Controller, useFormContext, useWatch } from 'react-hook-form'
 import { SvgIcon, Typography } from '@mui/material'
 import Autocomplete, { createFilterOptions } from '@mui/material/Autocomplete'
-import useAddressBook from '@/hooks/useAddressBook'
 import AddressInput, { type AddressInputProps } from '../AddressInput'
 import EthHashInfo from '../EthHashInfo'
 import InfoIcon from '@/public/images/notifications/info.svg'
@@ -11,6 +10,7 @@ import css from './styles.module.css'
 import inputCss from '@/styles/inputs.module.css'
 import { isValidAddress } from '@safe-global/utils/utils/validation'
 import { sameAddress } from '@safe-global/utils/utils/addresses'
+import { useAllMergedAddressBooks } from '@/hooks/useAllAddressBooks'
 
 const abFilterOptions = createFilterOptions({
   stringify: (option: { label: string; name: string }) => option.name + ' ' + option.label,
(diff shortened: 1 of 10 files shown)


apps/web/src/components/common/AddressBookInput/index.tsx
```diff
@@ -20,25 +20,26 @@ const abFilterOptions = createFilterOptions({
...
   const [openAddressBook, setOpenAddressBook] = useState<boolean>(false)
+  const mergedAddressBook = useAllMergedAddressBooks()
+
+  const { setValue, control } = useFormContext()
+  const addressValue = useWatch({ name, control })
 
-  const addressBookEntries = Object.entries(addressBook).map(([address, name]) => ({
-    label: address,
-    name,
+  const allAddressBookEntries = mergedAddressBook.map((entry) => ({
```
compojoom, line 30: can't we do this maping in the hook?

apps/web/src/features/spaces/components/SpaceAddressBook/SpaceAddressBookTable.tsx
```diff
@@ -27,14 +27,7 @@ function SpaceAddressBookTable({ entries }: SpaceAddressBookTableProps) {
...
             <Stack direction="column" spacing={0.5}>
-              <EthHashInfo
-                spaceName={entry.name}
-                showAvatar={false}
-                address={entry.address}
-                shortAddress={false}
-                hasExplorer
-                showCopyButton
-              />
+              <EthHashInfo showAvatar={false} address={entry.address} shortAddress={false} hasExplorer showCopyButton />
```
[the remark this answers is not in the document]
usame-algan: Name is handled inside EthHashInfo now so no need to pass it anymore

apps/web/src/components/common/AddressBookInput/index.tsx
usame-algan: We only need this mapping for rendering the list inside the `Autocomplete` while the hook is used in more places so I would keep it outside. We could probably put a `useMemo` around it though

francovenica:
Have some issues / questions:

1 - When I try to edit an entry that is already in other space I get an error

Have 2 spaces
Add an entry in SpaceA
Go to SpaceB and try to add the same entry

image

2 - 
In the address book I see the message "This data is stored in your local storage. Do you want to manage your [SPACENAME] space address book instead? Click here", but I see that message even if I enter a Safe from the accounts list.
I'd assume that, if I enter a safe account from the accounts page or by copy/Pasting the URL, I should not see that message and probably I should not see the addresses that are stored in the cloud (in the sidebar for example)


3 - To discuss: 
Whenever I enter into a safe I see everywhere the name from the cloud. For example if you name a safe account you see the name stored in the cloud in: The sidebar, the breadcrumb, the list of recipients in a tx, the name of the contract in the tx history tab.... The only place where I see it as it is in the LS is in the addressbook list, which in my opinion is kinda pointless.

usame-algan:
> 1 - When I try to edit an entry that is already in other space I get an error

I was not able to reproduce this one. Can you try again? 500 error without an error message looks like some unhandled server issue. It might have resolved itself now.

> In the address book I see the message "This data is stored in your local storage. Do you want to manage your [SPACENAME] space address book instead? Click here", but I see that message even if I enter a Safe from the accounts list.

I will tighten the check to only show this if its a safe thats part of the space like we do with the breadcrumbs

> Whenever I enter into a safe I see everywhere the name from the cloud. For example if you name a safe account you see the name stored in the cloud in: The sidebar, the breadcrumb, the list of recipients in a tx, the name of the contract in the tx history tab.... The only place where I see it as it is in the LS is in the addressbook list, which in my opinion is kinda pointless.

This is intended. We want to keep the changes as minimal as possible so we didn't want to touch that page but instead just add a CTA to redirect the user to their space address book.

francovenica:
1- I cannot reproduce it anymore either. We can leave it out
2 - Looks good. I See the info message if you come from a space or you are logged into spaces in general
3 - Noted, we leave it this way

PR #5871, opened 2025-05-13
feat(mobile): rehydrate chainConfig slice

compojoom:
## What it solves
Currenlty we had to wait for the app to fetch the chainConfig. It doesn’t change that often, so it makes sense that we persist it and use it directly when the app starts. Then we dispatch a new fetch and can update it if something has changed.

## How to test
Launch the app, hit cmd + shift in the terminal and open redux dev tools. Add a safe, Reload the app and inspect the rehydrate action. You should see it having the query and config objects
grafik

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -35,9 +35,13 @@ configureReanimatedLogger({
   strict: false,
 })
 
-// Initialize store-side effects
-store.dispatch(apiSliceWithChainsConfig.endpoints.getChainsConfig.initiate())
-
+persistor.subscribe(() => {
+  const { bootstrapped } = persistor.getState()
+  if (bootstrapped) {
+    // The chain config is persisted in the store, but might be outdated.
+    store.dispatch(apiSliceWithChainsConfig.endpoints.getChainsConfig.initiate(undefined, { forceRefetch: true }))
+  }
+})
 const HooksInitializer = () => {
   useInitWeb3()
   useInitSafeCoreSDK()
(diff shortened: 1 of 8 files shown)


apps/mobile/src/app/_layout.tsx
```diff
@@ -35,9 +35,13 @@ configureReanimatedLogger({
...
-// Initialize store-side effects
-store.dispatch(apiSliceWithChainsConfig.endpoints.getChainsConfig.initiate())
-
+persistor.subscribe(() => {
+  const { bootstrapped } = persistor.getState()
+  if (bootstrapped) {
+    // The chain config is persisted in the store, but might be outdated.
+    store.dispatch(apiSliceWithChainsConfig.endpoints.getChainsConfig.initiate(undefined, { forceRefetch: true }))
+  }
+})
```
compojoom, line 44: Previously we were dispatching the fetch whenever the store was ready, now it's better to wait for the rehydrate action to complete and to dispatch after that.

packages/store/src/utils/persistTransformFilter.ts
```diff
@@ -0,0 +1,154 @@
+// this is copy-paste from https://github.com/edy/redux-persist-transform-filter/tree/master
+// I've changed to typescript and modified the tests to work with jest
```
compojoom, line 2: I decided to copy-paste the code as it is based on the transform https://github.com/rt2zz/redux-persist?tab=readme-ov-file#transforms, is not going to change often, is not typescript and I don't want to rely on yet another package.

katspaugh (approved):
Good optimization! And it's nice you were able to use redux-persist, albeit with a custom filter.

compojoom:
I had to revert the place where I call persistor.subscribe. Doing it in the useEffect as was suggested in the AI review doesn't work properly as it is too late in the lifecycle and the subscribe is not triggered. 

I could use redux-persist without a custom filter, but this would mean that we would store the complete "api" slice, which contains all kind of network request we do. This will backfire at one point :D

PR #5872, opened 2025-05-14
Fix(Tx flow): don't show hex data as a fallback in multisend

katspaugh:
## What it solves

Resolves https://www.notion.so/safe-global/Repeated-data-field-in-the-Send-NFT-tx-1f38180fe5738034871cef3bc475a073

## How this PR fixes it

I've removed the raw data fallback in the Multisend component because we display the same data in the advanced details.


tmjssz (approved):

francovenica:
LGTM

PR #5874, opened 2025-05-14
Fix(Tx flow): fix tx flow styling

katspaugh:
## What it solves

* Rename the Receipt component
* Unify spending limits flow components (use `TxDetailsRow`)
* Increase the uncollapsed raw data length to 140 symbols (~4 lines) which fits most ERC20 transfers
* Fix the font of the "Advanced details" heading (see [ticket](https://www.notion.so/safe-global/Change-font-size-in-expanded-details-Signing-UX-v2-1f38180fe57380838a4ed0639ee9e858))


tmjssz (approved):
Just a small linter error, but otherwise LGTM!

francovenica:
LGTM

PR #5876, opened 2025-05-14
fix(Signing UX): hide batch button instead of disable when batching is not allowed

tmjssz:
## What it solves

Resolves: https://www.notion.so/safe-global/Add-to-batch-in-batch-txs-Signing-UX-v2-issue-1f38180fe57380a08f7aed114f552dc9?pvs=4

Hide the "Add to batch" button if a transaction is not batchable, instead of showing the button disabled.

## How this PR fixes it

Moves the conditions whether the transaction is batchable to the `useShouldRegisterSlot` hook. That way the button will only be rendered if the respective conditions are met.


katspaugh (approved):
👍

katspaugh:
I've tested it with different types of txs and batches – it's fixed and works correctly now. ✅

PR #5882, opened 2025-05-16
fix(Recovery): pass data to second step of recovery flow

schmanu:
## What it solves
https://www.notion.so/safe-global/Safe-recovery-Signing-UX-v2-1f28180fe573806b9a4bff880e8f8016

## How this PR fixes it
- Passes the data correctly between steps

## How to test it
- Setup recovery on an account without recovery
- Change the recovery setup on an account with existing recoverers

## Screenshots
Screenshot 2025-05-16 at 10 20 34

## Checklist

### DIFF
--- a/apps/web/src/components/tx-flow/TxFlowStep.tsx
+++ b/apps/web/src/components/tx-flow/TxFlowStep.tsx
@@ -13,7 +13,7 @@ export const TxFlowStep = ({ children, ...txLayoutProps }: TxFlowStepProps) => {
   useEffect(() => {
     updateTxLayoutProps(txLayoutProps)
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [])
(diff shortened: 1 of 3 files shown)


katspaugh (approved):
👍

francovenica:
1 - The title is wrong for me. It says "How does recovery works" which was the title of the previous step that shows the infoguide. I'd epxect here to show the ticket "Set up recovery settings"

image

2 - Trying to save a new recovery shows this error in the console. This shows every time I try to click Next.
I tried with brand new safes and old safes as well
image

schmanu:
@francovenica 
I fixed the first issue. It was a general bug in the new TxFlow. 

I cannot reproduce the second issue though. I was able to create and edit recovery settings without problems. Can you still reproduce it? And if yes: How do you reproduce it?

francovenica:
Ok, looks good now. It looks the same as the snapshot in the description

PR #5897, opened 2025-05-19
fix(web): do not mark address book addresses as untrusted in NameChip

schmanu:
## What it solves
https://www.notion.so/safe-global/Why-is-the-name-of-an-address-book-item-red-during-transaction-review-Signing-UX-v2-issue-1f38180fe57380c49d55eb1776b1f85e

### **`NameChip.ts`**
- Does not mark addresses as unverified / untrusted when they are in the address book.
- Adds unit test to component
## Checklist

### DIFF
--- a/apps/web/src/components/tx/ConfirmTxDetails/NameChip.tsx
+++ b/apps/web/src/components/tx/ConfirmTxDetails/NameChip.tsx
@@ -1,5 +1,6 @@
 import EthHashInfo from '@/components/common/EthHashInfo'
 import { useAddressName } from '@/components/common/NamedAddressInfo'
+import useAddressBook from '@/hooks/useAddressBook'
 import { isCustomTxInfo } from '@/utils/transaction-guards'
 import { Chip } from '@mui/material'
 import type { TransactionData, TransactionDetails } from '@safe-global/safe-gateway-typescript-sdk'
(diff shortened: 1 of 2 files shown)


apps/web/src/components/tx/ConfirmTxDetails/NameChip.tsx
```diff
@@ -22,15 +23,16 @@ const NameChip = ({
   const name = toName || contractInfo?.name
   const logo = toLogo || contractInfo?.logoUri
 
+  const addressBook = useAddressBook()
+  const isInAddressBook = toAddress !== undefined && !!addressBook[toAddress]
+  const isUntrusted = !isInAddressBook && contractInfo.isUnverifiedContract
```
katspaugh, line 28: An optimization would be to add the address book name to the `toName` definition a few lines above. Then it will skip the requests for a name/contract info in the `useAddressName` hook.

apps/web/src/components/tx/ConfirmTxDetails/NameChip.tsx
```diff
@@ -13,24 +14,27 @@ const NameChip = ({
...
   withBackground?: boolean
 }) => {
+  const addressBook = useAddressBook()
   const toAddress = txData?.to.value
   const customTxInfo = txInfo && isCustomTxInfo(txInfo) ? txInfo : undefined
   const toInfo = customTxInfo?.to || txData?.addressInfoIndex?.[txData?.to.value] || txData?.to
-  const toName = toInfo?.name || (toInfo && 'displayName' in toInfo ? String(toInfo.displayName || '') : undefined)
+  const nameFromAb = toAddress !== undefined ? addressBook[toAddress] : undefined
+  const toName =
+    toInfo?.name || (toInfo && 'displayName' in toInfo ? String(toInfo.displayName || '') : undefined) || nameFromAb
```
katspaugh, line 23: `nameFromAb` should probably come first.

schmanu: Changed priorities and added a test case for this.

katspaugh (approved):

katspaugh:
Tested, it's fixed ✅

PR #5899, opened 2025-05-19
fix(Signing UX): duplicate contract interaction info

### LINKED ISSUE #5780: [Signing UX v2] Duplicate contract interaction info for multiSend transactions
## Bug description

The transaction confirmation view shows the contract information twice for **mutliSend** transactions. 

## Environment

- Browser: Chrome
- Wallet: MetaMask
- Chain: Ethereum mainnet

## Steps to reproduce

1. Create a token transfer with multiple recipients
2. Proceed to the confirmation view
3. Expand the "Transaction details"
4. Observe the same info ("Call multiSend on...") twice
5. Observe that the list of actions included in the multiSend are missing

## Expected result

Instead of the contract interaction info shown twice, the confirmation view for multiSend transaction should include the list of actions.

## Obtained result

The contract interaction info is shown twice on the confirmation view for multiSend transactions.

### Before

This is how it looks currently on production:

Image

### Now

This is the same view on current staging:

Image

tmjssz:
## What it solves

Resolves #5780 

Hide the contract interaction info in the transaction details accordion if it is already being rendered above. 

## How to test it

1. Create a transaction that has no custom decoding (e.g. a multiSend transaction)
2. In the confirmation view observe that the transaction details accordion does not show the contract interaction info as it is already being rendered above
3. Same should be the case for a transaction in the queue/history
4. The contract interaction info should still be visible for transactions that have a custom decoding (such as a token transfer)

### Before
Screenshot 2025-05-19 at 17 23 16

### After

Screenshot 2025-05-19 at 17 22 55

### DIFF
--- a/apps/web/src/components/transactions/TxDetails/Summary/index.tsx
+++ b/apps/web/src/components/transactions/TxDetails/Summary/index.tsx
@@ -8,7 +8,7 @@ import { ZERO_ADDRESS } from '@safe-global/protocol-kit/dist/src/utils/constants
 import { Receipt } from '@/components/tx/ConfirmTxDetails/Receipt'
 import DecodedData from '../TxData/DecodedData'
 import ColorCodedTxAccordion from '@/components/tx/ColorCodedTxAccordion'
-import { Box, Divider, Typography } from '@mui/material'
+import { Box, Divider, Stack, Typography } from '@mui/material'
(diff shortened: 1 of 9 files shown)


francovenica:
LGTM, same as the snapshots in the description

PR #5906, opened 2025-05-20
fix(mobile): Qr code scanning

clovisdasilvaneto:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/221

## How this PR fixes it

This PR fixes two issues:

- Navigating multiple times for the import safe form - We introduced a flag to control whenever we should push the import safe form to the router or not. I've tried to se throttle but for some reason it was still calling the function, even memoizing it.
- Removes the camera background usage - the main cause of the issue, was because after updating to the expo version the previous layout is still mounted and running in background.

## How to test it

Please check: https://github.com/safe-global/wallet-private-tasks/issues/221 and try to reproduce the same steps described into the ticket.

## Screenshots

https://github.com/user-attachments/assets/119a15fc-21be-4521-a6f5-a7b12dd224cb

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/ImportReadOnly/hooks/useScan/useScan.test.tsx
+++ b/apps/mobile/src/features/ImportReadOnly/hooks/useScan/useScan.test.tsx
@@ -4,6 +4,28 @@ import { Code } from 'react-native-vision-camera'
 import { parsePrefixedAddress } from '@safe-global/utils/utils/addresses'
 import { isValidAddress } from '@safe-global/utils/utils/validation'
 
+// Store the focus callback for later testing
+let mockFocusCallback: (() => void) | null = null
+
+// Mock react-native-vision-camera
+jest.mock('react-native-vision-camera', () => ({
+  Camera: {
+    getCameraDevice: jest.fn(),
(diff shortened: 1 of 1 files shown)


compojoom (changes requested):
We are now navigating once to the next screen after scannign a QR code (as it is supposed to be), but what I'm noticing is that whenever we go to the camera screen for the first time the OS is prompting us to enable the camera. We had feedback early on that this is scary and that is why by default the camera is turned off, until the user press on the "enable camera" button. Then we prompt for permissions to enable the camera. We should keep this behavior. 
Also in light mode the switch off of the camera is very noticeable as the area where the camera is becomes white. I would remove the {isCameraActive && device && (... 
condition from the render - this way the camera will freeze at the last scanned frame.

clovisdasilvaneto:
@compojoom regarding this comment:

 >  I would remove the {isCameraActive && device &&

We can not remove it, because in case the permission is denied we will have an error in the app. Also I had to adjust the condition in there, because if the camera is mounted in the component (even if it is not active) it will prompt the user before he clicks in the "enable" button.

PR #5921, opened 2025-05-22
feat: optimize rpc requests through Multicall3

### LINKED ISSUE #5051: Batch RPC requests through Multicall3
## Problem
A lot of features require multiple RPC requests to the same contract:
- Fetching ERC20 symbol and decimals
- Fetching spending limits
- Recovery

## Proposed solution
We could reduce our RPC requests by batching multiple requests into one using the [Multicall3](https://github.com/mds1/multicall) contract.
Therefore we should create a helper function to batch RPC call, which tries to use Multicall3 and fetched individually as a fallback.

## Dependencies & risks
- We need to make sure that the Multicall3 contract is deployed on every network (it should be as it is deployed on 250+ chains)

schmanu:
## What it solves

Resolves #5051 

## How this PR fixes it
Data that can be fetched simultanously is being fetched through multicall3:
- SpendingLimits
- ERC20 Token infos
- Safe Token allocations

Introduces a new multicall utility for this.

## How to test it
- Test that Spending Limits are fetched correctly (across multiple chains)
- Test that the Safe Token balance and locked amount is fetched correctly

Everything should look and feel the same but require fewer RPC calls.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/hooks/useSafeTokenAllocation.ts
+++ b/apps/web/src/hooks/useSafeTokenAllocation.ts
@@ -1,7 +1,6 @@
 import { getSafeTokenAddress, getSafeLockingAddress } from '@/components/common/SafeTokenWidget'
 import { IS_PRODUCTION } from '@/config/constants'
 import { ZERO_ADDRESS } from '@safe-global/protocol-kit/dist/src/utils/constants'
-import { isPast } from 'date-fns'
 import { AbiCoder, Interface, type JsonRpcProvider } from 'ethers'
 import { useMemo } from 'react'
 import useAsync, { type AsyncResult } from '@safe-global/utils/hooks/useAsync'
@@ -9,6 +8,7 @@ import useSafeInfo from './useSafeInfo'
 import { getWeb3ReadOnly } from './wallets/web3'
 import memoize from 'lodash/memoize'
 import { cgwDebugStorage } from '@/config/gateway'
+import { multicall } from '../../../../packages/utils/src/utils/multicall'
 
 export const VESTING_URL =
   IS_PRODUCTION || cgwDebugStorage.get()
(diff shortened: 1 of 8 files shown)


apps/web/src/hooks/useSafeTokenAllocation.ts
```diff
@@ -43,49 +42,42 @@ const tokenInterface = new Interface(['function balanceOf(address _owner) public
...
 
-  const redeemDeadlineDate = new Date(Number(BigInt(redeemDeadline) * BigInt(1000)))
+    const isRedeemed = decodedVestingData[0].toLowerCase() !== ZERO_ADDRESS.toLowerCase()
+    if (isRedeemed) {
+      return { ...allocation, isRedeemed, isExpired: false, amountClaimed: decodedVestingData[6] }
+    }
 
-  // Allocation is valid if redeem deadline is in future
-  return { ...allocation, isRedeemed, isExpired: isPast(redeemDeadlineDate), amountClaimed: '0' }
+    // All allocations are expired by now. We do not load the redeem deadline anymore
```
usame-algan, line 78: What about the SAP airdrop that can be redeemed until end of this year?

apps/web/src/hooks/useSafeTokenAllocation.ts
```diff
@@ -116,41 +108,41 @@ const useSafeTokenAllocation = (): AsyncResult<Vesting[]> => {
...
+      },
+    ]
+
+    const results = await multicall(web3ReadOnly, calls)
+
+    if (!results[0].success || !results[1].success) {
+      throw new Error('Failed to fetch token balances')
+    }
+
+    return [BigInt(results[0].returnData), BigInt(results[1].returnData)]
```
usame-algan, line 143: nit: For better readability I would suggest to destructure results into named consts like you do in `useSafeVotingPower`

```
const [tokenBalance, lockingContractBalance] = await multicall(...)

if (!tokenBalance.success || !lockingContractBalance.success) ...
```

apps/web/src/utils/tokens.ts
```diff
@@ -4,33 +4,60 @@ import { parseBytes32String } from '@ethersproject/strings'
...
-    decimals: Number(decimals),
-    type: TokenType.ERC20,
+  const results = await multicall(web3, calls)
+
+  const tokenInfos: Omit<Erc20Token, 'name' | 'logoUri'>[] = []
+  for (let i = 0; i < results.length; i += 2) {
+    const address = tokenAddresses[i / 2]
+    if (!address) break
+    let symbol: string
+    try {
```
usame-algan, line 42: Not sure I understand why we had to change the existing logic. Why do we have to iterate over results like this i.e. increasing by 2 and dividing by 2? Aren't there always exactly two elements in the `results` array? The first one for symbol the second one for decimals?

apps/web/src/hooks/useSafeTokenAllocation.ts
schmanu: @usame-algan whilst re-introducing the redeemDeadline check I realized that we only use these allocations to compute the total voting power of the current Safe and the SAP airdrop allocation does not count towards that voting power (until claimed). For this reason it does not make a different if we return it or filter it out as expired.

We can leave it as is for this reason and save these RPC requests.

PR #5923, opened 2025-05-22
Fix(mobile): balance of all chains

clovisdasilvaneto:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/103

## How this PR fixes it
It removes the mulitples fetches we do for getting the balances of all chains.

## How to test it
Go to https://github.com/safe-global/wallet-private-tasks/issues/103 to get more details about it

## Checklist

---

## CLA signature


apps/mobile/src/hooks/services/useSafeOverviewService.ts
```diff
@@ -5,19 +5,24 @@ import { useMemo } from 'react'
 import { makeSafeId } from '@/src/utils/formatters'
 import { useSafesGetOverviewForManyQuery } from '@safe-global/store/gateway/safes'
 
-export const useSafeOverviewService = (safeAddress: string) => {
+export const useSafeOverviewService = (safeAddress?: string) => {
   const chainIds = useAppSelector(selectAllChainsIds)
   const safes = useMemo(
-    () => chainIds.map((chainId: string) => makeSafeId(chainId, safeAddress)),
+    () => (safeAddress ? chainIds.map((chainId: string) => makeSafeId(chainId, safeAddress)) : []),
```
clovisdasilvaneto, line 11: it was necessary because since we fetch the safe information in the beginning of the app, the address might not exists and hooks can not be conditional, so there were only 2 possible options, either create a wrapper component and put it under a render condition or either make this type optional.

I've decided to do it like that because if I create a wrapper component and put it under a a render condition, the whole app tree would re-render when the active safe changes (because the render component would trigger that), and to avoid that we would need to wrap the children under an useMemo and so on, which does not makes sense IMO.

apps/mobile/src/theme/provider/DataFetchProvider.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useSafesGetOverviewForManyQuery } from '@safe-global/store/gateway/safes'
+import { SafeOverviewResult } from '@safe-global/store/gateway/types'
+import { useChainsBalance } from '@/src/hooks/useChainsBalance'
+import { selectActiveSafe } from '@/src/store/activeSafeSlice'
+
+export const DataFetchProvider = ({ children }: { children: React.ReactNode }) => {
+  const chains = useAppSelector(selectAllChains)
+  const activeSafe = useAppSelector(selectActiveSafe)
+
+  useChainsBalance(activeSafe?.address)
```
compojoom, line 14: I don't undrestand this hook. Why do we need to do useChainsBalance here? 

useChainBalance does:  const data = useSafeOverviewService(safeAddress)

useSafeOverviewService does:
const { data } = useSafesGetOverviewForManyQuery<SafeOverviewResult>(
    {
      safes,
      currency: 'usd',
      trusted: true,
      excludeSpam: true,
    },
    {
      skip: !safeAddress,
    },
  )
  
  nearly the same as the line you have below. What useChainsBalance also does is updating the SafeInfo from the call. 
  
The purpose of the different hooks is a mess. It's not obvious why we need one over the other. Something is wrong in our architeture.

apps/mobile/src/theme/provider/DataFetchProvider.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
```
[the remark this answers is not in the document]
clovisdasilvaneto: 🤦🏻‍♂️ indeed, I put it to be inside `theme/providers` folder by mistake, it should be inside `src/providers`.

apps/mobile/src/theme/provider/DataFetchProvider.tsx
clovisdasilvaneto: @compojoom I didn't created this hook, I just moved where it is located and renamed it better.

PR #5928, opened 2025-05-23
Refactor(WalletConnect): centralize session state in context

katspaugh:
## What it solves

The WalletConnect session management logic was spread across the WalletConnectContext and WcSessionManager + useSessions. This was counter-intuitive when coming back to this feature after a while, so I've refactored it for all the logic to be inside the WalletConnectContext.


apps/web/src/features/walletconnect/WalletConnectContext.tsx
```diff
@@ -1,23 +1,344 @@
...
+      }
+
+      setIsLoading(undefined)
+    })
+  }, [walletConnect, safeWalletProvider, chainId, safeAddress, setOpen])
+
+  //
+  // --- Sessions
+  //
+  const [sessions, setSessions] = useState<SessionTypes.Struct[]>([])
```
usame-algan, line 230: nit: Is there a reason why some of the useState calls are not at the top?

apps/web/src/features/walletconnect/WalletConnectContext.tsx
```diff
@@ -1,23 +1,344 @@
...
+    walletConnect && setSessions(walletConnect.getActiveSessions())
+  }, [walletConnect])
+
+  // Initial sessions
+  useEffect(updateSessions, [updateSessions])
+
+  // On session add
+  useEffect(() => {
+    return walletConnect?.onSessionAdd(updateSessions)
+  }, [walletConnect, updateSessions])
```
usame-algan, line 242: nit: No comment needed. The function calls pretty much says the same thing. There are more instances like this.

apps/web/src/features/walletconnect/WalletConnectContext.tsx
```diff
@@ -1,23 +1,344 @@
...
+
+  // On session delete
+  useEffect(() => {
+    return walletConnect?.onSessionDelete(updateSessions)
+  }, [walletConnect, updateSessions])
+
+  //
+  // --- Proposals
+  //
+  const [sessionProposal, setSessionProposal] = useState<WalletKitTypes.SessionProposal | null>(null)
```
usame-algan, line 252: Is there a reason why this was changed from undefined to null?

usame-algan (approved):

apps/web/src/features/walletconnect/WalletConnectContext.tsx
```diff
@@ -1,23 +1,344 @@
...
+import { getSdkError } from '@walletconnect/utils'
+import { formatJsonRpcError } from '@walletconnect/jsonrpc-utils'
+import type { SessionTypes } from '@walletconnect/types'
+import type { WalletKitTypes } from '@reown/walletkit'
+
+import useSafeInfo from '@/hooks/useSafeInfo'
+import useSafeWalletProvider from '@/services/safe-wallet-provider/useSafeWalletProvider'
+import { IS_PRODUCTION } from '@/config/constants'
+import { getEip155ChainId, getPeerName, stripEip155Prefix } from '@/features/walletconnect/services/utils'
+import { trackRequest } from '@/features/walletconnect//services/tracking'
```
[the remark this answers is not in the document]
usame-algan: This one also needs fixing. Good Copilot!

usame-algan:
Not sure yet if this is because of the refactor but the Approve and Reject button stay disabled whereas on prod it still works:

Screenshot 2025-05-23 at 14 40 51

apps/web/src/features/walletconnect/WalletConnectContext.tsx
katspaugh: The context had most values as `null` by default. Only isLoading was `undefined`. I'd rather change the latter to `false` by default.

apps/web/src/features/walletconnect/WalletConnectContext.tsx
katspaugh: I'll keep them because they separate code blocks nicely. Agree they are somewhat superfluous though.

PR #5930, opened 2025-05-23
Fix(Tests): Fix and update recovery and nested safes e2e tests

liliya-soroka:
## What it solves
Fixes:
2 suites in the regression run
recovery suite
nestedsafes suite

Updates according to Signing UX v2
added check for the "Call to" line for multisend( a few actions) on Tx details for create nested safe middle step

## How to test it
Run cypress and verify that the updated suites are passed

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/cypress/e2e/regression/nested_safes_review.cy.js
+++ b/apps/web/cypress/e2e/regression/nested_safes_review.cy.js
@@ -21,7 +21,7 @@ describe('Nested safes review step tests', () => {
     nsafes.clickOnAddNestedSafeBtn()
   })
 
-  it.only('Verify middle step with Fund new assets in create nestedsafe tx flow', () => {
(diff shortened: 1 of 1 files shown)


apps/web/cypress/e2e/regression/nested_safes_review.cy.js
```diff
@@ -21,20 +21,20 @@ describe('Nested safes review step tests', () => {
     nsafes.clickOnAddNestedSafeBtn()
   })
 
-  it('Verify review step with Fund new assets', () => {
+  it.only('Verify middle step with Fund new assets in create nestedsafe tx flow', () => {
```
usame-algan, line 24: Is this intended or a leftover from testing? It will result in only this test being run.

usame-algan (approved):

PR #5943, opened 2025-05-27
feat(Spaces): leave a space

tmjssz:
## What it solves

Resolves [#116](https://github.com/safe-global/wallet-private-tasks/issues/116) - Allow active members to leave a space voluntarily
https://linear.app/safe-global/issue/EN-29/spaces-allow-members-to-leave-a-space
This PR enables users who are active members of a space to remove themselves from that space. Safeguards are included to prevent the last active admin from leaving.

> 🔗 **Note:** This change works in conjunction with related CGW updates [#2600](https://github.com/safe-global/safe-client-gateway/pull/2600).

### UI updates
- Added a “Leave space” button to the `SpaceSettings` view.
- Display a tooltip explaining why the button is disabled for the last active admin.
- Introduced `LeaveSpaceDialog` to confirm the action before proceeding.
- Display the Settings page in the sidebar navigation for all active members, not just active admins.
  - Otherwise the "leave space" functionality would not be reachable for non-admins.

### Logic & hooks
- Created a `useIsActiveMember` hook to determine if the user is considered active.
- Added utility functions in `useSpaceMembers` to assess admin roles and active member count.
- Implemented logic to detect whether the current user is the last active admin.

### Behavioral updates
- Disabled the leave button for users who are not active members or are the last active admin.
- Ensured `MembersList` leverages the new hooks for accurate admin role detection.

### Analytics
- Track user actions related to leaving a space.

### ✅ Scenario 1: Active member
1. Log into a space as an active member or admin (with at least one other active admin).
2. Navigate to **Settings**.
3. Confirm that a **"Leave space"** button is visible.
4. Click the button – a confirmation dialog should appear.
5. Confirm the dialog – the user should be removed from the space.

### ⚠️ Scenario 2: Last active admin
1. Log in as the sole active admin in a space.
2. Navigate to **Settings**.
3. **"Leave space"** button should be disabled.
4. Tooltip on hover explains that the last active admin cannot leave.

### 🚫 Scenario 3: Invited user
1. Log in as a user who has only been invited (not yet accepted).
2. **Settings** should not appear in the sidebar.

### Scenario 1 (active member - not admin)
Screenshot 2025-05-27 at 23 57 06

### Scenario 1 (active admin)
Screenshot 2025-05-28 at 00 01 59

### Scenario 2 (last active admin)
Screenshot 2025-05-28 at 00 02 39

### Scenario 3 (invited user)
Screenshot 2025-05-28 at 00 00 25

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/features/spaces/components/SpaceSidebarNavigation/index.tsx
+++ b/apps/web/src/features/spaces/components/SpaceSidebarNavigation/index.tsx
@@ -9,18 +9,19 @@ import {
   SidebarListItemText,
 } from '@/components/sidebar/SidebarList'
 import { useCurrentSpaceId } from '@/features/spaces/hooks/useCurrentSpaceId'
-import { useIsAdmin } from '@/features/spaces/hooks/useSpaceMembers'
+import { useIsActiceMember, useIsAdmin } from '@/features/spaces/hooks/useSpaceMembers'
 import { navItems } from './config'
 
 const Navigation = (): ReactElement => {
   const router = useRouter()
   const spaceId = useCurrentSpaceId()
   const isAdmin = useIsAdmin()
+  const isActiveMember = useIsActiceMember()
 
   return (
     <SidebarList>
       {navItems.map((item) => {
-        const hideItem = item.adminOnly && !isAdmin
+        const hideItem = (item.adminOnly && !isAdmin) || (item.activeMemberOnly && !isActiveMember)
         const isSelected = router.pathname === item.href
 
         if (hideItem) return null
--- a/apps/web/src/features/spaces/components/SpaceSidebarNavigation/index.tsx
+++ b/apps/web/src/features/spaces/components/SpaceSidebarNavigation/index.tsx
@@ -9,19 +9,18 @@ import {
   SidebarListItemText,
 } from '@/components/sidebar/SidebarList'
 import { useCurrentSpaceId } from '@/features/spaces/hooks/useCurrentSpaceId'
-import { useIsActiceMember, useIsAdmin } from '@/features/spaces/hooks/useSpaceMembers'
+import { useIsActiceMember } from '@/features/spaces/hooks/useSpaceMembers'
 import { navItems } from './config'
 
 const Navigation = (): ReactElement => {
   const router = useRouter()
   const spaceId = useCurrentSpaceId()
-  const isAdmin = useIsAdmin()
   const isActiveMember = useIsActiceMember()
 
   return (
     <SidebarList>
       {navItems.map((item) => {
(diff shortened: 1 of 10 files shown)


apps/web/src/features/spaces/components/SpaceSidebarNavigation/index.tsx
```diff
@@ -9,18 +9,19 @@ import {
...
   const router = useRouter()
   const spaceId = useCurrentSpaceId()
   const isAdmin = useIsAdmin()
+  const isActiveMember = useIsActiceMember()
 
   return (
     <SidebarList>
       {navItems.map((item) => {
-        const hideItem = item.adminOnly && !isAdmin
+        const hideItem = (item.adminOnly && !isAdmin) || (item.activeMemberOnly && !isActiveMember)
```
usame-algan: Is the first part of the condition still needed?

apps/web/src/features/spaces/hooks/useIsLastActiveAdmin.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { Member } from '@safe-global/store/gateway/AUTO_GENERATED/spaces'
+import { useCurrentMembership } from './useSpaceMembers'
+
+export const useAdminCount = (members?: Member[]) => {
+  const { activeMembers } = useSpaceMembersByStatus()
+  const membersToUse = members ?? activeMembers
+  return useMemo(() => membersToUse.filter(isAdmin).length, [membersToUse])
+}
+
+export const useIsLastActiveAdmin = (member?: Member) => {
```
usame-algan: It looks like this is only used in one place without passing a member prop. Can we remove it for now and always use `useCurrentMembership`?

usame-algan (approved):
Tested it with the backend and works nicely 👍

apps/web/src/features/spaces/components/SpaceSidebarNavigation/index.tsx
tmjssz: Not really, removed it along with the `adminOnly` flag.

apps/web/src/features/spaces/hooks/useIsLastActiveAdmin.ts
tmjssz: Done!

francovenica:
Probably an unlikely scenario but I was able to login with 2 admins in 2 different browsers. I left the space with one of them and then in the other browser (since the frontend doesn't update right away the leave button wasn't grayed out) I was able to leave with the other admin, leaving this space empty.

So, some validation has to be done before executing the action.
Is this scenario something we fix in the frontend? or is a backend validation?

PR #5946, opened 2025-05-28
Fix(Signing UX): execute status + CSS fixes

katspaugh:
## What it solves

* The Execute step was always green on step 2
* Wrong font for tx data rows
* Name chip background

Screenshot 2025-05-28 at 14 28 46
Screenshot 2025-05-28 at 14 30 46
Screenshot 2025-05-28 at 14 30 01

### DIFF
--- a/apps/web/src/components/tx/FieldsGrid/index.tsx
+++ b/apps/web/src/components/tx/FieldsGrid/index.tsx
@@ -19,7 +19,7 @@ const FieldsGrid = ({ title, children }: { title: string | ReactNode; children: 
       ]}
     >
       <Grid item data-testid="tx-row-title" style={{ wordBreak: 'break-word' }} sx={gridSx}>
-        <Typography color="primary.light" variant="body1" component="span">
+        <Typography color="primary.light" variant="body2" component="span">
           {title}
         </Typography>
       </Grid>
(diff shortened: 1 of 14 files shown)


apps/web/src/components/tx/FieldsGrid/index.tsx
```diff
@@ -19,7 +19,7 @@ const FieldsGrid = ({ title, children }: { title: string | ReactNode; children:
       ]}
     >
       <Grid item data-testid="tx-row-title" style={{ wordBreak: 'break-word' }} sx={gridSx}>
-        <Typography color="primary.light" variant="body1" component="span">
```
usame-algan, line 22: This was updated as part of the native earn decoding to always have 16px of font-size on the confirmation screen: https://www.figma.com/design/K1EOvrgmK59lIPTM3WDHpm/Lending?node-id=5078-16132&m=dev
Are there exceptions where we want to keep the 14px and can we list them? cc @TanyaEfremova

francovenica:
LGTM

The execute only lights up if execution is the only possible action (Note, It doesn't light up if 1 signature is missing and the user opts for a sign&execute. The tx must be fully signed and ready to be executed for that text to show **bold**)
The background color for the contract being interacted with is showing fine 
The font size of the data field looks fine

apps/web/src/components/tx/FieldsGrid/index.tsx
usame-algan: Thanks for clarifying. I agree for long data strings like that going with a smaller font makes more sense.

PR #5767, opened 2025-04-28
feat(web): add optional custom paramaters for safe apps

valens-carpentier:
## What it solves
Safe Apps cannot properly handle URL parameters (like UTM tracking) because they interfere with manifest loading. This PR implements a solution that allows Safe Apps to maintain URL parameters while ensuring correct manifest loading.

Resolves #187

## How this PR fixes it
- Added `stripURLParams` to strip URL parameters, returning only the base URL
- Added `originalUrl` to `SafeAppDataWithPermissions` type to track URLs with parameters
- Modified manifest fetching to use clean URLs while preserving original parameters
- Updated iframe component to use the full URL with parameters

## How to test it
- Start local development server
- Add a Custom Safe App
- Use URL https://app.venus.io/#/?utm_source=SafeWallet
- App can be added
- Apps loads correctly

## Screenshots
Screenshot 2025-04-28 at 15 36 38

## Checklist

- [] I've tested the branch on mobile (not applicable) 📱
- [] I've documented how it affects the analytics (not applicable) 📊
- [] I've written a unit/e2e test for it (if applicable) 🧑‍💻

PR #5947, opened 2025-05-28
Fix(Tests):update test app url according to changes

liliya-soroka:
## What it solves
Contains the fix for :

- message-popup regression suite 
- tx-details-createtx regression suite
- tx-builder happypath2

Resolves #
https://github.com/safe-global/safe-wallet-monorepo/pull/5767 broke the e2e tests where the safe app URL is used
## How this PR fixes it
1. Update iframe URL view
2. wrong selector was used to define the message field in the test safe app
## How to test it
run message-popup regression suite 
## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/cypress/e2e/regression/tx_details_createtx.cy.js
+++ b/apps/web/cypress/e2e/regression/tx_details_createtx.cy.js
@@ -40,7 +40,7 @@ describe('Transaction details create tests', { defaultCommandTimeout: 30000 }, (
     safeapps.verifyUntrustedHandllerWarningVisible()
   })
 
-  it.only('Verify that when the tx contains the action with an official 1.4.1 fallbackhandler contract there is no error', () => {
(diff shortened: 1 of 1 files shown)


apps/web/cypress/e2e/regression/tx_details_createtx.cy.js
```diff
@@ -39,7 +40,7 @@ describe('Transaction details create tests', { defaultCommandTimeout: 30000 }, (
     safeapps.verifyUntrustedHandllerWarningVisible()
   })
 
-  it('Verify that when the tx contains the action with an official 1.4.1 fallbackhandler contract there is no error', () => {
+  it.only('Verify that when the tx contains the action with an official 1.4.1 fallbackhandler contract there is no error', () => {
```
schmanu, line 43:
```suggestion
  it('Verify that when the tx contains the action with an official 1.4.1 fallbackhandler contract there is no error', () => {
```

schmanu (changes requested):
I think you forgot to remove two `only` test modifiers.

PR #5949, opened 2025-05-30
feat(web): simulate transactions from queue

schmanu:
## How this PR fixes it
New button in queued transactions' details to simulate the transaction.

New component `QueuedTxSimulation` which offers simulation without requiring a Transaction in the context and renders it inline.

## How to test it
1. Queue any transaction
2. Simulate it in the tx details
3. Test edge cases 
  - by queueing txs that will fail (by using e.g. tx builder)
  - by blocking the simulation request in the browser dev tools

## Screenshots
Screenshot 2025-05-30 at 16 55 24
Screenshot 2025-05-30 at 16 55 49
Screenshot 2025-05-30 at 16 55 32

## Checklist

---

## CLA signature


apps/web/src/components/tx/security/tenderly/utils.ts
```diff
@@ -106,3 +109,41 @@ export const getSimulationPayload = async (params: SimulationTxParams): Promise<
...
+
+export type SimulationStatus = {
+  isLoading: boolean
+  isFinished: boolean
+  isSuccess: boolean
+  isCallTraceError: boolean
+  isError: boolean
+}
+
+export const getSimulationStatus = (simulation: UseSimulationReturn): SimulationStatus => {
```
compojoom: Can't this live in the utils package? It looks like something that we would also need in the mobile app.

schmanu: There is no issue moving it there.
But should we move it there when the mobile app starts using it? Or should I do it now?

usame-algan:
What is the expected behaviour if there is a transaction in the queue that I already signed with my wallet? In that case the simulation fails for me when trying to simulate again.

schmanu:
> What is the expected behaviour if there is a transaction in the queue that I already signed with my wallet? In that case the simulation fails for me when trying to simulate again.

This should not happen. Seems like the overwrites are somehow wrong in that case.
I will take a look and fix it.

apps/web/src/components/tx/security/tenderly/utils.ts
compojoom: Let's move it now. If we have utility code that is not dependent on any mobile/web stuff it should live in the utils package. This makes it easier to use it later.

schmanu:
@usame-algan 
It was a general bug in the simulation utils. It should be fixed now.
Good catch!

compojoom (approved):

PR #5956, opened 2025-06-02
feat(utils): overwrite guard with zero address when simulating

schmanu:
## What it solves
Resolves https://github.com/safe-global/wallet-private-tasks/issues/226

## How this PR fixes it
When a guard is present, we overwrite the guard slot with the zero address it when simulating transactions.

## How to test it
- Setup a Safe with a Guard that blocks some transaction
- Queue a transaction that meets the blocking criteria
- Observe a successful simulation despite the guard

## Checklist

---

## CLA signature

### DIFF
--- a/packages/utils/src/components/tx/security/tenderly/utils.ts
+++ b/packages/utils/src/components/tx/security/tenderly/utils.ts
@@ -13,7 +13,7 @@ import type {
   TenderlySimulation,
 } from '@safe-global/utils/components/tx/security/tenderly/types'
 import type { EnvState } from '@safe-global/store/settingsSlice'
-import { toBeHex } from 'ethers'
+import { toBeHex, ZeroAddress } from 'ethers'
 
 export const getSimulationLink = (simulationId: string): string => {
   return `https://dashboard.tenderly.co/public/${TENDERLY_ORG_NAME}/${TENDERLY_PROJECT_NAME}/simulator/${simulationId}`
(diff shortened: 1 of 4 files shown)


packages/utils/src/components/tx/security/tenderly/utils.ts
```diff
@@ -135,6 +144,9 @@ export const getStateOverwrites = (params: SimulationTxParams) => {
   if (nonceOverwrite !== undefined) {
     storageOverwrites[NONCE_STORAGE_POSITION] = toBeHex('0x' + BigInt(nonceOverwrite).toString(16), 32)
   }
+  if (guardOverwrite !== undefined) {
+    storageOverwrites[GUARD_STORAGE_POSITION] = toBeHex(guardOverwrite, 32)
```
usame-algan, line 148: Why don't we always overwrite the guard with the zero address regardless if they have a guard enabled or not? Then we don't need `getGuardOverwrite`.

schmanu: No specific reason tbh. Its the same pattern as with the overwritten nonce.
We could just always overwrite it to match the txs nonce. But we only do it when 
```ts
  if (txNonce > safeNonce) {
    return txNonce
  }
```

Should I remove the check and always set it to the Zero address?

usame-algan: Ah I didn't see that. Lets keep it as is for consistency sake.

usame-algan (approved):
👍

liliya-soroka:
vefrified, details in PMS

PR #5962, opened 2025-06-02
tests(mobile): add safes slice tests

compojoom:
## What it solves
add safes slice tests

---

## CLA signature


usame-algan (approved):
nit: It might be easier to maintain this test if we used a builder pattern for the mock data

Issue #5724, opened 2025-04-16
Credential route logic breaks IAP proxy


cen1:
## Bug description
Cgw client decides on whether to include credentials or not based on the path, see https://github.com/safe-global/safe-wallet-monorepo/blob/dev/packages/store/src/gateway/cgwClient.ts#L50

This breaks google IAP proxy because requests going to the gateway are missing the required IAP cookies.

IAP essentially allows you to transparently put the Safe UI behind an OAuth screen and block public access.

Idea to fix this is to simply add an ENV config option that disable this credentials logic and uses default behaviour (same-origin). Since all calls when using IAP are authenticated, default client logic will work as expected. Any security implications bypassing this logic? I don't see any for this particular use case.

## Environment
Deploy UI behind google IAP

## Steps to reproduce
1. Open UI
2. Open network
3. Observe that XHR calls to /cgw/v1/<...> are blocked due to missing cookies

## Expected result
IAP setup works

## Obtained result
IAP setup fails

sara554-beep:
Hola..
Consulta esto va  solamente dirigido a mi??

atte. sara silva

katspaugh:
Pull requests are welcome!

cen1:
```
const IS_BEHIND_IAP = process.env.NEXT_PUBLIC_IS_BEHIND_IAP === 'true'

export function isCredentialRoute(url: string) {
  return IS_BEHIND_IAP || CREDENTIAL_ROUTES.some((route) => url.match(route))
}
```

This simple 2-line patch solves the problem, however, since I am not knowledgeable on the code base, before opening a PR it would be good to know why this protections exist and why it was added in the first place. Sending credentials on certain CGW paths is apparently not desirable and I suspect there must have been a good reason why it was added.

This was introduced in https://github.com/safe-global/safe-wallet-monorepo/pull/5430 so perhaps cc @iamacook for insight

sara554-beep:
Te agradezco tu información.

Hace un tiempo atrás me robaron toda la información según los datos que me
entregó guihub, te soy sincera yo no manejo bien este sistema, si me puedes
ayudar te lo agradecería mucho.
He perdido más de un año tratando de conseguir mis datos perdidos. Si es
necesario te puedo enviar mis claves o códigos de verdad necesito ayuda

atte. sara silva

iamacook:
@cen1, I don't think it should be an issue but we will need to test it with our (non-)authenticated routes.

PR #5972, opened 2025-06-04
fixes for deployments protected behind IAP proxy

cen1:
## What it solves
Safe App deployment behind IAP proxy.

IAP proxy allows you to protect a publicly facing Safe App instance from random access by adding an OAuth screen in front of the deployment. After user authenticates, cookie credentials are transparently added to all HTTP traffic to allow access to the instance.

Resolves #
https://github.com/safe-global/safe-wallet-monorepo/issues/5724

## How this PR fixes it
1. Skip credential route check when IAP is enabled, fixes https://github.com/safe-global/safe-wallet-monorepo/issues/5724
2. When behind IAP, manifest fetch fails because credentials are not sent. Please note that I am not 100% sure on the security implications on setting `crossOrigin="use-credentials"`, to the best of my knowledge it should be OK.

## How to test it
1. Deploy unpatched Safe App behind IAP. Observe that calls to CGW fail due to missing authentication.
2. In browser network, observe that fetching manifest fails due to missing credentials

## Screenshots
n/a
## Checklist

---

## CLA signature


apps/web/src/components/common/MetaTags/index.tsx
```diff
@@ -36,7 +36,7 @@ const MetaTags = ({ prefetchUrl }: { prefetchUrl: string }) => (
     {/* PWA primary color and manifest */}
     <meta name="theme-color" content={lightPalette.background.main} media="(prefers-color-scheme: light)" />
     <meta name="theme-color" content={darkPalette.background.main} media="(prefers-color-scheme: dark)" />
-    <link rel="manifest" href="/safe.webmanifest" />
+    <link rel="manifest" href="/safe.webmanifest" crossOrigin="use-credentials" />
```
usame-algan: Is this the right place to add this? This is the manifest that is used for the PWA config and not the manifest.json that is loaded when accessing a safe app.

packages/store/src/gateway/cgwClient.ts
```diff
@@ -13,8 +13,10 @@ export const CREDENTIAL_ROUTES = [
   /\/v2\/chains\/[^\/]+\/notifications\/devices/,
 ]
 
+const IS_BEHIND_IAP = process.env.NEXT_PUBLIC_IS_BEHIND_IAP === 'true'
```
usame-algan, line 16: Having this setting here would make it apply to all safe apps not only safe apps that have an IAP proxy. We need to make sure it doesn't break anything for such safe apps.

cen1: Since I'm not deeply familiar with the repo and this is a drive-by MR solving a specific pain point, could you clarify what you mean with "safe apps" in plural in this context? For me a "Safe App" instance is something I will build from this repo, package into Docker and deploy somewhere. In what deployment scenario would multiple instances share a common ENV?

apps/web/src/components/common/MetaTags/index.tsx
katspaugh: Looks like we misunderstood what you meant by a "Safe App", @cen1. @usame-algan thought you meant external apps embedded in an iframe into the Safe Wallet (that's what we call it).

A follow-up question: should this attribute be added conditionally if `IS_BEHIND_IAP` is true, or does it work regardless?

packages/store/src/gateway/cgwClient.ts
katspaugh: As @usame-algan mentioned, the `isCredentialRoute` is used only for Client-Gateway (CGW) API requests, and it's only a subset of all requests that the app makes.

I assume you're running both the static frontend files and a CGW instance behind an IAP?

cen1: Indeed, both FE and CGW are on the same instance. Because of IAP, all calls essentially become "credentialed".

apps/web/src/components/common/MetaTags/index.tsx
cen1: While I don't really see a a security issue per se (auth being passed on manifest calls) I guess it would be on the safe side to add this conditionally. I'll update the PR.

cen1: This is done.

cen1:
> Sorry for the delay! The new environment variable should also be added to build/action.yml to be compatible with other env vars via GitHub Secrets. Also ideally we should export IS_BEHIND_IAP from constants.ts like all other env vars.

Ok, this is good feedback. The fix for `MetaTags/index.tsx` looks simple since it already has `@/config/constants` import, however, it is not clear to me how to do it in `cgwClient.ts` since I don't see any pre-existing use cases of env imports in the whole package. Let me know.

katspaugh:
@cen1 you can find examples in [constants.ts](https://github.com/safe-global/safe-wallet-monorepo/blob/dev/apps/web/src/config/constants.ts) (the filename is not very fitting).

And the build action is [here](https://github.com/safe-global/safe-wallet-monorepo/blob/dev/.github/actions/build/action.yml#L38).

cen1:
Yeah I got that, but it is not clear to me whether it would be ok to import that into `packages/store/src/gateway/cgwClient.ts` which is external to it.

katspaugh:
Ah, I see now. Yes, it makes sense to just use process.env in that package.

katspaugh (approved):
Thanks!

katspaugh (approved):

PR #5976, opened 2025-06-04
fix(Signing UX): show name for address of Safe itself correctly

tmjssz:
## What it solves

Resolves [WAL-5951](https://linear.app/safe-global/issue/WAL-5951/verified-on-etherscan-contracts-are-displayed-as-unverified-on-ui)

Show "This Safe Account" when address matches Safe address.

Fixes the issue where the name was shown as "Unverified contract" when the address matched the Safe address.

## How this PR fixes it

Move the address comparison with the current Safe address to the useAddressName hook to ensure that the name is always displayed correctly.

## How to test it

1. Create a transaction that is sent to the Safe itself (e.g. a change threshold transaction)
2. Proceed to the Confirm transaction step and expand the transaction details accordion
3. Observe that the "To" value in the receipt has a "This Safe Account" name tag
4. Continue to the Review step and observe the same in the receipt there
5. Sign or execute the transaction 
6. Open it in the queue
7. Observe the correct name tag in the transaction details there as well

## Screenshots

Screenshot 2025-06-04 at 17 29 15

Screenshot 2025-06-04 at 17 28 52

---

## CLA signature

### DIFF
--- a/apps/web/src/components/common/NamedAddressInfo/index.test.tsx
+++ b/apps/web/src/components/common/NamedAddressInfo/index.test.tsx
@@ -1,7 +1,9 @@
-import { render, waitFor } from '@/tests/test-utils'
-import NamedAddressInfo from '.'
+import { render, waitFor, renderHook } from '@/tests/test-utils'
+import NamedAddressInfo, { useAddressName } from '.'
 import { faker } from '@faker-js/faker'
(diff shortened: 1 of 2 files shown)


usame-algan (approved):
Thanks!

liliya-soroka:
Verified

PR #5985, opened 2025-06-09
Fix typos and enhance clarity in safe-monorepo.mdc

MamunC0der:
Fixes:
formally → formerly — grammar fix

compoentns → components — typo fix


compojoom:
@MamunC0der thanks for the PR, but we don't accept such commits that only correct typos. 

In addition to this: ""Formally" indicates something is done according to established rules, etiquette, or convention, often with an official or serious tone. "Formerly" refers to a time in the past or a previous state or position"

PR #5987, opened 2025-06-10
feat(test): add adv params verification after editing

francovenica:
## What it solves
Adds a test that was not being done. Checks that the values edited in adv params are correctly saved after submiting the form

## How this PR fixes it
Adds the scenario Verify advance parameters are saved after editing
Creates a new function to set the values and submit and one to get the values once they are set

## How to test it
Run the test create_tx_2.ts and all test should pass

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/cypress/e2e/pages/create_tx.pages.js
+++ b/apps/web/cypress/e2e/pages/create_tx.pages.js
@@ -127,6 +127,7 @@ const enabledBulkExecuteBtnTooltip = 'All highlighted transactions will be inclu
 const bulkExecuteBtnStr = 'Bulk execute'
 
 const batchModalTitle = 'Batch'
+const gasLimit21000 = 'Gas limit must be at least 21000'
 export const swapOrder = 'Swap order settlement'
 export const bulkTxs = 'Bulk transactions'
 export const txStr = 'Transactions'
@@ -152,6 +153,13 @@ const advancedParametersValues = {
   maxFee: '0.5678',
   gasLimit: '300001',
 }
+const advancedParametersInputNames = {
+  walletNonce: 'Wallet nonce',
+  maxPriorityFee: 'Max priority fee (Gwei)',
+  maxFee: 'Max fee (Gwei)',
+  gasLimit: 'Gas limit',
+}
+
 
 // Transaction details on Tx creation
 export const txAccordionDetails = '[data-testid="decoded-tx-details"]'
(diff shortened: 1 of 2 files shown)


apps/web/cypress/e2e/pages/create_tx.pages.js
```diff
@@ -778,6 +792,26 @@ export function verifyAndSubmitExecutionParams() {
   cy.get('@Paramsform').submit()
 }
 
+export function setAdvncedExecutionParams() {
+  cy.contains(executionParamsStr).parents('form').as('Paramsform')
+  const arrayNames = ['Wallet nonce', 'Max priority fee (Gwei)', 'Max fee (Gwei)', 'Gas limit']
```
liliya-soroka: move const from the function, so it can be reused later by other functions ( check if we already have it for other functions )

apps/web/cypress/e2e/pages/create_tx.pages.js
```diff
@@ -778,6 +792,26 @@ export function verifyAndSubmitExecutionParams() {
...
+  })
+  cy.get('@Paramsform').find(gasLimitInput).clear().type(advancedParametersValues.gasLimit)
+  cy.get('@Paramsform').find(maxPriorityFee).clear().type(advancedParametersValues.maxPriorityFee)
+  cy.get('@Paramsform').find(maxFee).clear().type(advancedParametersValues.maxFee)
+  cy.get('@Paramsform').find(walletNonceInput).clear().type(advancedParametersValues.walletNonce)
+  cy.get('@Paramsform').submit()
+}
+
+export function verifyEditedExutionParams() {
+  cy.contains('Wallet nonce').next().should('contain', advancedParametersValues.walletNonce)
```
liliya-soroka: same as above - the current implementation will require hardcoded data updating in case of changes for the advanced parameters

apps/web/cypress/e2e/pages/create_tx.pages.js
francovenica: That's from a previous test that I didn't touch.
Still, I changed that array to have constants instead of text

apps/web/cypress/e2e/pages/create_tx.pages.js
francovenica: Changed it to not have text
Also removed from this scenario, test that were redundant

francovenica:
Did the changes.
Also I realized that the steps inside were wrong, basically functionA was testing functionB and viceversa, so I changed that.
I've checked that this functions are only used in the create_tx_2 test, so no other test should be affected
All tests in create_tx_2 work locally

PR #5988, opened 2025-06-10
fix(test): fix counting external links 

francovenica:
## What it solves
The test counts the amount of external links. The system wasn't finding them

## How this PR fixes it
Makes the system search in the signers list

## How to test it
Run Spending_limit test. The scenario "Verify explorer links contain Sepolia link" should pass

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/cypress/e2e/pages/create_tx.pages.js
+++ b/apps/web/cypress/e2e/pages/create_tx.pages.js
@@ -89,6 +89,7 @@ const recipientsCount = '[data-testid="recipients-count"]'
 const maxBtn = '[data-testid="max-btn"]'
 const tokenAmountSection = '[data-testid="token-amount-section"]'
 const insufficientBalanceError = '[data-testid="insufficient-balance-error"]'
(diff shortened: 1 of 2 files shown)


apps/web/cypress/e2e/pages/create_tx.pages.js
```diff
@@ -438,7 +439,7 @@ export function verifyNumberOfCopyIcons(number) {
 }
 
 export function verifyNumberOfExternalLinks(number) {
-  cy.get(explorerBtn)
+  cy.get(signerList).find(explorerBtn)
```
liliya-soroka: check other tests where "verifyNumberOfExternalLinks" was used

francovenica: I added a get('main') so it searches in the main body of the page instead of the sidebar. 
The other test that uses the function are also working fine with this change
I was going to make a variable so it doesn't have the 'main' tag just like that, but is literally the only "main" in the website so I assume it would be redundant

PR #5991, opened 2025-06-10
Feat(Dashboard): Revamp layout

usame-algan:
## What it solves

Resolves [EN-127](https://linear.app/safe-global/issue/EN-127/dashboard-update-header)

## How this PR fixes it

- Adds a new `NewsCarousel` component
- Creates new banners for spaces and earn and adds them to the carousel
- Updates the layout of the Overview component according to Figma

## How to test it

1. Open a Safe
2. Observe the new Dashboard overview

## Screenshots

With banners visible:
Screenshot 2025-06-10 at 10 55 28

With all banners dismissed:
Screenshot 2025-06-10 at 10 53 07

Mobile:
Screenshot 2025-06-10 at 11 56 59

Dark mode:
Screenshot 2025-06-10 at 11 57 12

## Checklist

---

## CLA signature


compojoom:
@usame-algan @TanyaEfremova I just tested it and it feels a bit werid. 

When you have 2 banners you have this shadow on the right indicating that you can scroll there by clicking on this erea. It really scrolls to the second banner, but I cannot click on the left to go back to the first banner. I have to click on the 2 dots at the bottom. I don't know this somehow screws with my brain. It's weird.
grafik
and then:
grafik


to me this looks broken. I would immediately wonder - why is this not stretching.

usame-algan:
> I would immediately wonder - why is this not stretching.

We could stretch it if its only one item. @TanyaEfremova wdyt?

I wanted to start with as little as possible for the carousel. We can add more interactions if necessary.

apps/web/src/components/dashboard/NewsCarousel/banners/SpacesBanner.tsx
```diff
@@ -0,0 +1,55 @@
...
+import Track from '@/components/common/Track'
+import css from './styles.module.css'
+import ChevronRightIcon from '@mui/icons-material/ChevronRight'
+import { AppRoutes } from '@/config/routes'
+import Link from 'next/link'
+
+const SpacesBanner = ({ onDismiss }: { onDismiss: () => void }) => {
+  const onHide = () => {
+    onDismiss()
+  }
```
compojoom: is this necessary. Can't we pass this directly to the onClick?

compojoom:
> We could stretch it if its only one item. @TanyaEfremova wdyt?
> 
> I wanted to start with as little as possible for the carousel. We can add more interactions if necessary.

Why not have it as carousel stretch 100% and loop over the items? We already have the indicator below it to know that we have 2 items in it.

apps/web/src/components/dashboard/NewsCarousel/banners/SpacesBanner.tsx
usame-algan: I moved it directly into the onClick handler

TanyaEfremova:
1. When scrolled, another banner should also stay on screen (see attached video)
2. Close icons should use` border-main` color
3. The right gradient seems to be more gray, than white, is it correct? 

@compojoom It's fine to show a part of the content that needs to be scrolled - it's visible and engaging, and shows that there's more content to be seen, but once it is scrolled to the left - the last banner should be shown with another one. So I understand your confusion, but I would adjust the interaction.

https://github.com/user-attachments/assets/e481abdb-1720-4c42-b835-ba8ad2bf7fbe

TanyaEfremova:
Additionally, swiping should be possible by holding and dragging the mouse, it's hard to click on the dot below. And/or we could add a chevron on the carousel sides, for even better accessibility.

usame-algan:
@TanyaEfremova I've added a swipe interaction. Not sure if the dots are even needed now since the next banner is always visible to indicate to the user that there is more. I also removed the overlay on the right because it was overlapping the close button of the second banner. Not sure how to solve that. Do you think it would work without?

francovenica:
The only issue I had is that sometimes I grab the image when I try to scroll the card:
drag card

Beyond that the feature looks fine to me

francovenica:
Is it ok for the "Earn" banner to show up if the safe has only tokens that are Not compatible with the Earn feature (like the DAI token for example)

Beyond that the ticket is fine

PR #5993, opened 2025-06-10
feat(mobile): create advanced transaction details screen

clovisdasilvaneto:
## What it solves

Resolves https://github.com/safe-global/wallet-private-tasks/issues/246

## How this PR fixes it
We replaced the old "parameters" view, by the transaction details one. Please notice that it is no 100% following the figma structure because there were some missing alignments in terms of nomenclature which were just noticed when we were working in the task. 

_**ps: This structure and texts were aligned already with design team.**_

## How to test it
Go to any pending transaction and you should be able to see the "Transaction details" button instead of the old parameters one.

## Screenshots
Screenshot 2025-06-10 at 11 01 18

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/AdvancedDetails/utils/formatParameters.test.tsx
+++ b/apps/mobile/src/features/AdvancedDetails/utils/formatParameters.test.tsx
@@ -1,3 +1,4 @@
+import { TransactionData } from '@safe-global/store/gateway/AUTO_GENERATED/transactions'
 import { formatParameters } from './formatParameters'
 
 // Mock dependencies with minimal implementation
@@ -29,7 +30,7 @@ const { formatValueTemplate } = require('../formatters/singleValue')
 const { formatArrayValue } = require('../formatters/arrayValue')
 
 // Mock data helper to bypass strict typing for tests
-const createMockTxData = (data: any) => data as any
+const createMockTxData = (data: unknown) => data as TransactionData
 
 describe('formatParameters', () => {
   beforeEach(() => {
--- a/apps/mobile/src/features/AdvancedDetails/utils/formatTxDetails.test.tsx
+++ b/apps/mobile/src/features/AdvancedDetails/utils/formatTxDetails.test.tsx
@@ -71,7 +71,7 @@ describe('formatTxDetails', () => {
   })
 
   it('should return empty array when txDetails is null', () => {
-    const result = formatTxDetails({ txDetails: null })
+    const result = formatTxDetails({ txDetails: null as unknown as TransactionDetails })
     expect(result).toEqual([])
   })
 
@@ -146,7 +146,7 @@ describe('formatTxDetails', () => {
     }
 
     const txDetails = createMockTxDetails({
-      detailedExecutionInfo: mockExecutionInfo as TransactionDetails['detailedExecutionInfo'],
+      detailedExecutionInfo: mockExecutionInfo as unknown as TransactionDetails['detailedExecutionInfo'],
     })
 
     isMultisigDetailedExecutionInfo.mockReturnValue(true)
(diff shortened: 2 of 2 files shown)


apps/mobile/src/app/transaction-parameters/(tabs)/_layout.tsx
```diff
@@ -0,0 +1,58 @@
...
+          width: 94,
+          marginLeft: 6,
+          alignItems: 'center',
+        },
+        tabBarLabelStyle: {
+          color: theme?.color?.get(),
+          fontSize: 16,
+          fontWeight: '700',
+        },
+      }}
```
compojoom: those screen options are nearly the same as in the transactions _layout.tsx - can we maybe extract this and reuse it?

apps/mobile/src/app/transaction-parameters/(tabs)/parameters.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+import { SafeAreaView } from 'react-native-safe-area-context'
+
+import { TxParametersContainer } from '@/src/features/AdvancedDetails'
+
+function TransactionParameters() {
+  return (
+    <SafeAreaView edges={['bottom']} style={{ flex: 1 }}>
```
compojoom: I'm not 100% sure if the docs in react-navigation hold true, but I avoid using SafeAreaView:
https://reactnavigation.org/docs/handling-safe-area/

> While React Native exports a SafeAreaView component, this component only supports iOS 10+ with no support for older iOS versions or Android. In addition, it also has some issues, i.e. if a screen containing safe area is animating, it causes jumpy behavior. So we recommend to use the useSafeAreaInsets hook from the [react-native-safe-area-context](https://github.com/th3rdwave/react-native-safe-area-context) library to handle safe areas in a more reliable way.
> warning
> 
> The react-native-safe-area-context library also exports a SafeAreaView component. While it works on Android, it also has the same issues with jumpy behavior on vertical animations. In addition, the SafeAreaView component and useSafeAreaInsets hook can update at different times, resulting in flickering when using them together. So we recommend always using the useSafeAreaInsets hook instead and avoid using the SafeAreaView component for consistent behavior.

apps/mobile/src/app/transaction-parameters/_layout.tsx
```diff
@@ -0,0 +1,27 @@
...
+    >
+      <Stack.Screen
+        name="(tabs)"
+        options={() => ({
+          headerTitle: (props) => (
+            <View width="100%" flex={1} marginTop={2}>
+              <H2 fontWeight={600} {...props}>
+                Transaction details
+              </H2>
+            </View>
```
compojoom: Why don't you use LargetHeaderTitle from the components?

apps/mobile/src/components/CopyButton/CopyButton.tsx
```diff
@@ -12,14 +13,12 @@ interface CopyButtonProps {
 export const CopyButton = ({ value, color, size = 13, text }: CopyButtonProps) => {
   const copyAndDispatchToast = useCopyAndDispatchToast(text)
   return (
-    <Button
+    <TouchableOpacity
```
compojoom, line 16: just curious - what was broken with Tamagui again?

apps/mobile/src/navigation/useScrollableHeader.tsx
```diff
@@ -28,7 +28,12 @@ export const useScrollableHeader = ({ children, alwaysVisible, scrollYThreshold
   useEffect(() => {
     navigation.setOptions({
       headerTitle: () => (
-        <Animated.View style={[{ flexDirection: 'row', alignItems: 'center' }, animatedHeaderStyle]}>
+        <Animated.View
+          style={[
+            { flex: 1, justifyContent: 'center', flexDirection: 'row', alignItems: 'center' },
+            animatedHeaderStyle,
+          ]}
+        >
```
compojoom, line 36: did you test other views that use this hook. Any noticeable change? Why are we adding flex 1 and justfiy content center?

clovisdasilvaneto: yep, I've tested and everything works fine.

apps/mobile/src/components/CopyButton/CopyButton.tsx
clovisdasilvaneto: It was adding styles on top of the icon, and according to the designs, these buttons where we have this copy functionality they should not have any background or anything, just the "opacity" when pressed, which is done by default with the TouchableOpacity button.

PR #5997, opened 2025-06-10
Fix(Dashboard): Update dashboard assets layout

usame-algan:
## What it solves

Resolves [EN-134](https://linear.app/safe-global/issue/EN-134/dashboard-update-top-assets-section)

## How this PR fixes it

- Updates the layout of the dashboard assets table according to Figma
- Adds buttons on hover
- Shows a percentage of total fiat value next to each item
- The percentage is hidden on mobile view

## How to test it

1. Open a Safe dashboard
2. Observe the new design for the assets overview

## Screenshots
Screenshot 2025-06-10 at 15 45 26
Screenshot 2025-06-10 at 15 45 36
Screenshot 2025-06-10 at 15 45 47
Screenshot 2025-06-10 at 15 46 01

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/balances/AssetsTable/SendButton.tsx
+++ b/apps/web/src/components/balances/AssetsTable/SendButton.tsx
@@ -9,7 +9,7 @@ import { ASSETS_EVENTS } from '@/services/analytics/events/assets'
 import { TokenTransferFlow } from '@/components/tx-flow/flows'
 import { TxModalContext } from '@/components/tx-flow'
 
-const SendButton = ({ tokenInfo, isOutlined }: { tokenInfo: TokenInfo; isOutlined?: boolean }) => {
+const SendButton = ({ tokenInfo, light }: { tokenInfo: TokenInfo; light?: boolean }) => {
   const spendingLimit = useSpendingLimit(tokenInfo)
   const { setTxFlow } = useContext(TxModalContext)
 
@@ -23,9 +23,9 @@ const SendButton = ({ tokenInfo, isOutlined }: { tokenInfo: TokenInfo; isOutline
         <Track {...ASSETS_EVENTS.SEND}>
           <Button
             data-testid="send-button"
-            variant={isOutlined ? 'outlined' : 'contained'}
-            color="primary"
-            size="small"
+            variant="contained"
+            color={light ? 'background.paper' : 'primary'}
+            size="compact"
             startIcon={<ArrowIconNW />}
             onClick={onSendClick}
             disabled={!isOk}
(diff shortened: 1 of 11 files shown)


apps/web/src/components/dashboard/Assets/index.tsx
```diff
@@ -50,64 +56,84 @@ const NoAssets = () => (
...
-      <TokenAmount
-        value={item.balance}
-        decimals={item.tokenInfo.decimals}
-        tokenSymbol={item.tokenInfo.symbol}
-        logoUri={item.tokenInfo.logoUri}
-      />
-    </Box>
+  showStake?: boolean
+}) => {
+  const fiatTotalNumber = Number(fiatTotal)
```
compojoom: Is it possible for fiatTotalNumber to end up being 0?

usame-algan: Yes, definitely. I will add a guard to make sure the math works out even if its 0

usame-algan: Extracted the logic into a helper function inside our utils package.

apps/web/src/components/balances/AssetsTable/SendButton.tsx
```diff
@@ -23,9 +23,9 @@ const SendButton = ({ tokenInfo, isOutlined }: { tokenInfo: TokenInfo; isOutline
         <Track {...ASSETS_EVENTS.SEND}>
           <Button
             data-testid="send-button"
-            variant={isOutlined ? 'outlined' : 'contained'}
-            color="primary"
-            size="small"
+            variant="contained"
+            color={light ? 'background.paper' : 'primary'}
+            size="compact"
```
iamacook, line 28: This also affects the [`AssetsTable`](https://github.com/safe-global/safe-wallet-monorepo/blob/f56781e2a7180aa830278bc6e732d3af7ab1a80c/apps/web/src/components/balances/AssetsTable/index.tsx#L198). Is that intended?

packages/utils/src/utils/__tests__/formatNumber.test.ts
```diff
@@ -90,4 +96,31 @@ describe('formatNumber', () => {
...
+    it('handles a negative total by returning 0', () => {
+      expect(percentageOfTotal(10, -50)).toBe(0)
+    })
+
+    it('handles non-numeric totals by returning 0', () => {
+      expect(percentageOfTotal(10, 'not-a-number')).toBe(0)
+    })
+
+    it('handles non-numeric balances by returning 0', () => {
+      expect(percentageOfTotal('NaN', 100)).toBe(0)
```
iamacook: Nit: this isn't testing `NaN` (I tested this change working):

```suggestion
      expect(percentageOfTotal(NaN, 100)).toBe(0)
```

apps/web/src/components/balances/AssetsTable/SendButton.tsx
usame-algan: Yes, I wanted to keep the change as small as possible for the assets table so I added thise `light` prop. The only difference on the assets table should be the size which will be slightly different but I think thats fine.

iamacook (approved):

PR #6009, opened 2025-06-12
fix(web): migrate to V2 /owners/{address}/safes

LucieFaire:
## What it solves

Resolves [OCX-58](https://linear.app/safe-global/issue/OCX-58/web-migrate-to-v2-of-ownersaddresssafes-and-safe-overviews-endpoints)

## How this PR fixes it
Migrates from `useGetAllOwnedSafesQuery` (SDK) to `useOwnersGetAllSafesByOwnerV2Query` (CGW API) which allows a partial response if any tx service is not available

## How to test it
Wherever we load all the owned safes the response should be the same as for the current app, but if something failed it should not block the screen and show partial response

## Checklist

---

## CLA signature


apps/web/src/features/counterfactual/hooks/usePendingSafeNotifications.ts
```diff
@@ -30,7 +33,9 @@ const usePendingSafeNotifications = (): void => {
   const chain = useCurrentChain()
   const safeAddress = useSafeAddress()
   const { address = '' } = useWallet() || {}
-  const { refetch } = useGetAllOwnedSafesQuery(address === '' ? skipToken : { walletAddress: address })
+  const { refetch } = useOwnersGetAllSafesByOwnerV2Query(
+    address === '' ? (skipToken as unknown as OwnersGetAllSafesByOwnerV2ApiArg) : { ownerAddress: address }, //todo: can we do lazy here?
+  )
```
LucieFaire: as this hook seems to be used in a lazy matter, and the refetch is called on-demand, should we use `useLazyOwnersGetAllSafesByOwnerV2Query` instead?

apps/web/src/features/myAccounts/hooks/useAllOwnedSafes.ts
```diff
@@ -1,11 +1,16 @@
...
+  OwnersGetAllSafesByOwnerV2ApiResponse,
+} from '@safe-global/store/gateway/AUTO_GENERATED/owners'
+import { useOwnersGetAllSafesByOwnerV2Query } from '@safe-global/store/gateway/AUTO_GENERATED/owners'
 import { asError } from '@safe-global/utils/services/exceptions/utils'
 import { skipToken } from '@reduxjs/toolkit/query'
 
-const useAllOwnedSafes = (address: string): AsyncResult<AllOwnedSafes> => {
-  const { data, error, isLoading } = useGetAllOwnedSafesQuery(address === '' ? skipToken : { walletAddress: address })
+const useAllOwnedSafes = (address: string): AsyncResult<OwnersGetAllSafesByOwnerV2ApiResponse> => {
+  const { data, error, isLoading } = useOwnersGetAllSafesByOwnerV2Query(
```
compojoom: Please confirm whether we should be using data (possible that we implemented it wrong the first time)
https://redux-toolkit.js.org/rtk-query/usage/queries#frequently-used-query-hook-return-values

> data - The latest returned result regardless of hook arg, if present.
> currentData - The latest returned result for the current hook arg, if present.

apps/web/src/features/counterfactual/hooks/usePendingSafeNotifications.ts
compojoom: Yes, I'm not sure why we do this. Even though we are not using data, currentdata I think that this query is dispatching a request. So the data sits in the redux store. If we don't need it at this moment we can go for a lazy query and trigger.

compojoom: why do we need to do: `skipToken as unknown as OwnersGetAllSafesByOwnerV2ApiArg` ?

schmanu: My guess is that we are not interested at all in the owned Safe in this hook.
But we want to make sure, that we invalidate the redux cache, once a new Safe has been deployed such that it appears in the Sidebar, etc.

So I think the clean way to do this is by invalidating the `owners` tag by dispatching an `invalidateTags` action (https://redux-toolkit.js.org/rtk-query/api/created-api/api-slice-utils#invalidatetags)

apps/web/src/features/myAccounts/hooks/useAllOwnedSafes.ts
schmanu: I think we likely want to use `currentData` here. Otherwise you see old Safes when switching wallets etc.

apps/web/src/features/counterfactual/hooks/usePendingSafeNotifications.ts
LucieFaire: I was getting an error: 
`Type 'unique symbol' is not assignable to type 'OwnersGetAllSafesByOwnerV2ApiArg | unique symbol'.`
Even though RTK Query itself supports `skipToken`, I think the codegen does not type the arguments as supporting it. I could try to change the config and regenerate so it supports it, otherwise this was the best type-safe way i could think of

LucieFaire: actually I have checked and there is no support for skipToken in `@rtk-query/codegen-openapi` so I guess thats out

LucieFaire: I'm open for suggestions how to go about this in a prettier way:)

schmanu: I encountered the same issue.
You can pass `{skip: address === ''}` as a options parameter to the call. That's how I resolved it in the past.

apps/web/src/features/myAccounts/hooks/useAllOwnedSafes.ts
LucieFaire: hm, according to documentation `currentData` is _the last successful result for the current query args (held in cache)_. I understand it as - if we refetch data, and it fails or undefined for some time we can still show the previous result, so the UI isnt flickering. Then its the opposite to what @schmanu said, and you would see the old Safes, no?

compojoom: CurrentData takes into account the arguments provided to the function.

If we do data, then it doesn't care. If we don't have data for the new function arguments, we show the old data. this is what data is returning and we had few bugs because of that. Like this one for example: https://github.com/safe-global/safe-wallet-monorepo/pull/5868

LucieFaire: ok I understand now, thank you for the explanation. I will update it accordingly

apps/web/src/features/counterfactual/hooks/usePendingSafeNotifications.ts
```diff
@@ -50,7 +46,7 @@ const usePendingSafeNotifications = (): void => {
 
         // Fetch all owned safes after the Safe has been deployed
         if (isSuccess) {
-          refetch()
+          dispatch(cgwApi.util.invalidateTags(['owners']))
```
LucieFaire, line 49: @schmanu did you mean to do this instead of refetching?

PR #6019, opened 2025-06-16
chore: use consistent dependencies across packages

compojoom:
## What it solves
We switched to a monorepo structure, but never enforced that packages across the repo use the same version number. With this PR I'm trying to change this. I've added a constraints script that outputs info for packages that we want to be at the same version across the repo.

grafik

In addition to this the husky config we previously had before switching to monorepo structure was no longer running. The prepare script was left in the web package and not in the package.json in the root folder and because of this pre-commit and pre-push scripts were never run. 

As a bonus - once I enabled husky I couldn't push this PR as the mobile tests were failing. Updating the expo-module-scripts in a previous PR removed an outdated dependency that we were erroneously using... 

## How to test it
Building and running the repo-packages should run as expected

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -155,7 +155,7 @@
     "@storybook/test": "^8.4.6",
     "@testing-library/react-native": "^13.2.0",
     "@types/eslint__js": "^8.42.3",
-    "@types/jest": "^29.5.12",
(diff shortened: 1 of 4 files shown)


compojoom:
I needed to merge this as the failing tests were preventing me from pushing some fixes

PR #6024, opened 2025-06-19
fix: copy button being overlaped in sign tx screen

clovisdasilvaneto:
## What it solves
It fixes the buttons overlap problem in the sign tx screen

## How to test it
Open any token transaction in the pending transactions screen.

## Screenshots
Screenshot 2025-06-19 at 14 16 01

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/ConfirmTx/components/confirmation-views/TokenTransfer/TokenTransfer.tsx
+++ b/apps/mobile/src/features/ConfirmTx/components/confirmation-views/TokenTransfer/TokenTransfer.tsx
@@ -63,20 +63,18 @@ export function TokenTransfer({ txId, txInfo, executionInfo, executedAt }: Token
             <View alignItems="center" flexDirection="row" justifyContent="space-between">
               <Text color="$textSecondaryLight">To</Text>
 
-              <View flexDirection="row" alignItems="center">
-                <View flexDirection="row" alignItems="center" gap="$2">
-                  <Identicon address={recipientAddress} size={24} />
-                  <EthAddress
-                    address={recipientAddress}
-                    copy
-                    copyProps={{ color: '$textSecondaryLight', size: 18 }}
-                    textProps={{ fontSize: '$4' }}
-                  />
-                </View>
+              <View flexDirection="row" alignItems="center" gap="$2">
+                <Identicon address={recipientAddress} size={24} />
+                <EthAddress
+                  address={recipientAddress}
+                  copy
+                  copyProps={{ color: '$textSecondaryLight', size: 18 }}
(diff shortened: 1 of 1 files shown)


apps/mobile/src/features/ConfirmTx/components/confirmation-views/TokenTransfer/TokenTransfer.tsx
```diff
@@ -63,20 +63,18 @@ export function TokenTransfer({ txId, txInfo, executionInfo, executedAt }: Token
...
+                <Identicon address={recipientAddress} size={24} />
+                <EthAddress
+                  address={recipientAddress}
+                  copy
+                  copyProps={{ color: '$textSecondaryLight', size: 18 }}
+                  textProps={{ fontSize: '$4' }}
+                />
                 <Button
                   onPress={viewOnExplorer}
+                  marginLeft={-10}
```
compojoom, line 76: why do we need a negative margin at all? Negative margins are weird. Maybe we should design a custom theme for the button instead. this way something like this won't happen.

usame-algan: Agree, do we know if this will fix the issue on other screen sizes as well? Why is the explorer link not part of the EthAddress like we have it on web?

clovisdasilvaneto: @daniel0ar @usame-algan  It is necessary because the ETH Component button has an extra space in the right, and we can just align that or either keeping the same structure as before and limiting the `width` or either doing this negative margin in orther to avoid a big gap in the left of it. I personally don't see a problem by leaving it negative. 

A nice article about this topic: https://blog.shimin.io/are-css-negative-margins-bad-practice/#:~:text=So%20is%20using%20negative%20margins,box%20model%20properties%20would%20allow. In short: 

"The negative CSS margin is a bit of a black sheep of the box model family. Unlike its siblings the negative padding and border – which are simply invalid – the negative margin is useful yet often misused.

So is using negative margins 'back practice'? The short answer is no, as long as you are only using negative CSS margins to bring elements closer together than their box model properties would allow."

clovisdasilvaneto: Answering your question @usame-algan yes it fixes on other screen sizes, regarding moving it to be part of the EthAddress, we could do it for sure, but I rather doing it in another PR, as a tech debit then on this hot fix PR.

PR #6027, opened 2025-06-20
chore: bump @safe-global/safe-deployments to latest version

dasanra:
## What it solves

Push @safe-global/safe-deployments to add compatibility for PEAQ


compojoom (approved):
👍

PR #6029, opened 2025-06-20
fix(WalletConnect): prevent URI from being overridden by an empty value

tmjssz:
## What it solves

Resolves [EN-145](https://linear.app/safe-global/issue/EN-145/christopher-founder-1kx-is-asking)

This fixes the issue where the input field for the WC URI is being cleared when pasting a value. That caused validation errors not being diplayed for invalid URI values. 

## How this PR fixes it

Call `onInput` inside the useEffect hook only if the URI value is not empty/undefined.

## How to test it

1. Open the WC dialog
2. Paste an invalid URI
3. Observe that the input field is not being cleared and an error message is being shown

## Screenshots

Screenshot 2025-06-20 at 19 34 50

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/features/walletconnect/components/WcInput/index.tsx
+++ b/apps/web/src/features/walletconnect/components/WcInput/index.tsx
@@ -64,7 +64,9 @@ const WcInput = ({ uri }: { uri: string }) => {
 
   // Insert a pre-filled uri
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


francovenica:
LGTM
The text is not cleared anymore and shows the issue properly

snapshot for documentation sake
image

PR #6041, opened 2025-06-24
fix(mobile): notifications fixes

compojoom:
## What it solves
When importing a safe and in the flow importing an owner the created delegate was restricted by safe, which is wrong. The delegate for push notifications doesn't need safe restriction.

Also added a small fix for when the gateway would return a 404 on the unsubscribe endpoint. We treat the notifications unsubscribe as succesfull. 

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/services/notifications/backend.ts
+++ b/apps/mobile/src/services/notifications/backend.ts
@@ -148,15 +148,24 @@ export const unregisterForNotificationsOnBackEnd = async ({
   const deviceUuid = await getDeviceUuid()
 
   for (const chainId of chainIds) {
-    await getStore()
-      .dispatch(
-        notificationsApi.endpoints.notificationsDeleteSubscriptionV2.initiate({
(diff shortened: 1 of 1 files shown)


apps/mobile/src/services/notifications/backend.ts
```diff
@@ -148,15 +148,24 @@ export const unregisterForNotificationsOnBackEnd = async ({
...
+          notificationsApi.endpoints.notificationsDeleteSubscriptionV2.initiate({
+            deviceUuid,
+            chainId,
+            safeAddress,
+          }),
+        )
+        .unwrap()
+    } catch (error: unknown) {
+      // Treat 404 errors as successful unregistration since the safe was already unsubscribed
+      if (error && typeof error === 'object' && 'status' in error && error.status === 404) {
```
clovisdasilvaneto, line 163: can we move this to a variable so it is easier to read?

PR #6050, opened 2025-06-25
fix(web): update balance dynamically in Header top right bar) for active safe

LucieFaire:
## What it solves [OCX-84](https://linear.app/safe-global/issue/OCX-84/ui-value-not-updating-in-top-right-corner-after-transfer)

Resolves #
UI value for fiat balance is not being updated in the Header for an active chain

## How this PR fixes it

For an active chain it is using the value from balances that is being dynamically updated (as in Total Asset Value and Side bar), and for the other ones in the list its lazy-loading the safe overviews

## How to test it

- Perform a send or receive action in the wallet
- Observe that balance got updated in in Total Asset Value, Side bar (left side) and header (top right bar)
- Observe that there was no calls to fetch other save overviews unless you opened the dropdown to switch the network

## Screenshots
Screenshot 2025-06-25 at 14 05 19

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/common/NetworkSelector/index.tsx
+++ b/apps/web/src/components/common/NetworkSelector/index.tsx
@@ -24,7 +24,6 @@ import useChains, { useCurrentChain } from '@/hooks/useChains'
 import type { NextRouter } from 'next/router'
 import { useRouter } from 'next/router'
 import css from './styles.module.css'
-import { useChainId } from '@/hooks/useChainId'
 import { type ReactElement, useCallback, useMemo, useState, useEffect } from 'react'
 import { OVERVIEW_EVENTS, OVERVIEW_LABELS, trackEvent } from '@/services/analytics'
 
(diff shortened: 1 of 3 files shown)


LucieFaire:
@schmanu should I write unit tests to check the logic? I will need to export this component as I dont want to mock the whole NetworkSelector

apps/web/src/components/common/NetworkSelector/index.tsx
```diff
@@ -54,18 +54,21 @@ const ChainIndicatorWithFiatBalance = ({
...
+
+  useEffect(() => {
+    if (!isCurrentChain && !undeployedSafe) {
+      trigger({ safeAddress, chainId: chain.chainId })
+    }
+  }, [trigger, isCurrentChain, undeployedSafe, safeAddress, chain.chainId])
+
+  const fiatValue = isCurrentChain ? balances.fiatTotal : safeOverview?.fiatTotal
+
+  return <ChainIndicator responsive={isSelected} chainId={chain.chainId} fiatValue={fiatValue || undefined} inline />
```
schmanu: Nit:
```suggestion
  return <ChainIndicator responsive={isSelected} chainId={chain.chainId} fiatValue={fiatValue} inline />
```

schmanu:
> @schmanu should I write unit tests to check the logic? I will need to export this component as I dont want to mock the whole NetworkSelector

I think that's a good idea. Especially now that the component has more branches in its logic.

schmanu (approved):
The change looks good to me!

I agree with your suggestion to add a small test for it. But I will already approve.

PR #6057, opened 2025-06-27
fix(web:e2e): remove owner and proposer test

francovenica:
## What it solves
Fixes remove test by changing what text the system is asserting and removing a click that no longer can be done

Also in proposers test I removed a scenario when an owner is a proposer, but now an owner cannot be no longer a proposer so the test is irrelevant

## How to test it
Run regression/remove_owner
Run regression/propoers_2

## Checklist

---

## CLA signature


usame-algan (approved):
Thanks! I ran both regression tests locally and they are green

PR #6058, opened 2025-06-27
fix(web:e2e): fixed create tx2 test

francovenica:
## What it solves
Some functions had typos in the test, so the code would not find and use them

## How to test it
Run regression/create_tx_2 test

## Checklist

---

## CLA signature


apps/web/cypress/e2e/regression/create_tx_2.cy.js
```diff
@@ -29,16 +29,16 @@ describe('Create transactions tests 2', () => {
     createtx.clickOnSendTokensBtn()
   })
 
-  it('Verify advance parameters are saved after editing', () => {
+  it.only('Verify advance parameters are saved after editing', () => {
```
liliya-soroka, line 32: only

PR #6059, opened 2025-06-27
fix(mobile): biometrics on settings

clovisdasilvaneto:
## What it solves

Fixes biometrics hook state management issue where hasInteracted flag was unexpectedly resetting to true, causing unwanted biometric prompts when app returns from background.

## How this PR fixes it
- Convert hasInteracted from useState to useRef to prevent re-render cycles
- Remove dependencies from useEffect that were causing multiple re-runs
- Add deduplication for duplicate AppState events
- Fix dependency array issue that was the root cause of the state inconsistency

## How to test it
- Enable biometrics in app settings
- Put app in background, then return to foreground
- Verify biometrics prompt doesn't appear unexpectedly
- Toggle biometrics setting and verify it works correctly

## Screenshots
N/A - Logic fix only

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/hooks/useBiometrics.ts
+++ b/apps/mobile/src/hooks/useBiometrics.ts
@@ -102,7 +102,6 @@ export function useBiometrics() {
       setIsLoading(true)
 
       try {
-        await new Promise((resolve) => setTimeout(resolve, 3000))
         const isSupported = await checkBiometricsSupport()
         const { biometricsEnabled: isEnabledAtOSLevel } = await checkBiometricsOSSettingsStatus()
 
--- a/apps/mobile/src/hooks/useBiometrics.ts
+++ b/apps/mobile/src/hooks/useBiometrics.ts
@@ -1,4 +1,4 @@
-import { useState, useCallback, useLayoutEffect } from 'react'
+import { useState, useCallback, useLayoutEffect, useEffect, useRef } from 'react'
 import * as Keychain from 'react-native-keychain'
 import { useAppDispatch, useAppSelector } from '@/src/store/hooks'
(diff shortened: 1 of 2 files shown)


apps/mobile/src/hooks/useBiometrics.ts
```diff
@@ -89,80 +95,93 @@ export function useBiometrics() {
...
-      const { biometricsEnabled: isEnabledAtOSLevel } = await checkBiometricsOSSettingsStatus()
+  const enableBiometrics = useCallback(
+    async (fromInteraction?: boolean) => {
+      setIsLoading(true)
 
-      if (!isSupported && userCancelled) {
-        openBiometricSettings()
-      }
+      try {
+        await new Promise((resolve) => setTimeout(resolve, 3000))
```
compojoom: why the timeout here?

clovisdasilvaneto: 🤦‍♂️ leftover, will remove

apps/mobile/src/hooks/useBiometrics.ts
```diff
@@ -179,15 +197,45 @@ export function useBiometrics() {
...
+
+          // If biometrics is not available at OS level, disable it
+          if (!isSupported || !isEnabledAtOSLevel) {
+            await disableBiometrics()
+            return
+          }
+
+          if (hasInteractedRef.current) {
+            await enableBiometrics()
+          }
```
compojoom: how can this ever by true? You are setting it to false on line 216?

compojoom (approved):
👍

PR #6063, opened 2025-06-27
fix(web:e2e): TxBuilder tx creation

francovenica:
## What it solves
The test looks for 2 specific events to be present after a tx creation with the tx builder

Resolves #

## How this PR fixes it
The EventLable has changed in the site, so I changed it in the values used to compare in the test

NOTE: We have to be sure that the eventLabel in the app was changed on purpose or not.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/cypress/support/utils/gtag.js
+++ b/apps/web/cypress/support/utils/gtag.js
@@ -62,7 +62,7 @@ export const events = {
   txCreatedTxBuilder: {
     category: 'transactions',
     action: 'Confirm transaction',
-    eventLabel: 'https://safe-apps.dev.5afe.dev/tx-builder',
+    eventLabel: 'walletconnect',
     eventType: 'tx_created',
     event: 'tx_created',
   },
@@ -70,7 +70,7 @@ export const events = {
   txConfirmedTxBuilder: {
     category: 'transactions',
     action: 'Confirm transaction',
-    eventLabel: 'https://safe-apps.dev.5afe.dev/tx-builder',
+    eventLabel: 'walletconnect',
     eventType: 'tx_confirmed',
     event: 'tx_confirmed',
   },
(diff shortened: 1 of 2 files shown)


apps/web/cypress/support/utils/gtag.js
```diff
@@ -62,15 +62,15 @@ export const events = {
   txCreatedTxBuilder: {
     category: 'transactions',
     action: 'Confirm transaction',
-    eventLabel: 'https://safe-apps.dev.5afe.dev/tx-builder',
+    eventLabel: 'walletconnect',
```
liliya-soroka, line 65: events from the safe apps , should contain information about the safe app , not about the walletconnet.

usame-algan: This can probably be reverted once https://github.com/safe-global/safe-wallet-monorepo/pull/6068 is merged.

apps/web/cypress/support/utils/gtag.js
```diff
@@ -62,15 +62,15 @@ export const events = {
...
     eventType: 'tx_created',
     event: 'tx_created',
   },
 
   txConfirmedTxBuilder: {
     category: 'transactions',
     action: 'Confirm transaction',
-    eventLabel: 'https://safe-apps.dev.5afe.dev/tx-builder',
+    eventLabel: 'walletconnect',
```
usame-algan, line 73: This can probably be reverted once https://github.com/safe-global/safe-wallet-monorepo/pull/6068 is merged

liliya-soroka: yes, it should be reverted , as it was correct expectations in the test. I will re-run e2de tests to verify on the PR

liliya-soroka:
the PR is closed, as it was a bug and tests did not require an update

PR #6068, opened 2025-06-30
Fix(Tx): Propose safe transactions with origin

usame-algan:
## What it solves

Resolves [EN-131](https://linear.app/safe-global/issue/EN-131/safe-app-logo-is-lost-for-the-txs-created-via-safe-app-on-the-history)

## How this PR fixes it

- Calls `setTxOrigin` when entering the `SafeAppsTxFlow`
- Removes unused `ReviewTransaction` components and tests

## How to test it

1. Go to the transaction builder
2. Create a batch of two transactions
3. Go to the queue
4. Observe the safe apps logo and name are visible
5. Execute the transaction
6. Observe the safe apps origin url is part of the execution event
7. Go to the history
8. Observe the safe apps logo and name is still visible

## Screenshots

Screenshot 2025-06-30 at 12 27 34
Screenshot 2025-06-30 at 12 31 35

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/tx-flow/flows/SafeAppsTx/ReviewSafeAppsTx.tsx
+++ b/apps/web/src/components/tx-flow/flows/SafeAppsTx/ReviewSafeAppsTx.tsx
@@ -40,8 +40,12 @@ const ReviewSafeAppsTx = ({
       return tx
     }
 
-    createSafeTx().then(setSafeTx).catch(setSafeTxError)
-    setTxOrigin(getTxOrigin(app))
+    createSafeTx()
+      .then((tx) => {
+        setSafeTx(tx)
+        setTxOrigin(getTxOrigin(app))
+      })
+      .catch(setSafeTxError)
   }, [txs, setSafeTx, setSafeTxError, setTxOrigin, app, params])
 
   const error = !isTxValid(txs)


apps/web/src/components/tx-flow/flows/SafeAppsTx/ReviewSafeAppsTx.tsx
```diff
@@ -40,7 +41,8 @@ const ReviewSafeAppsTx = ({
     }
 
     createSafeTx().then(setSafeTx).catch(setSafeTxError)
-  }, [txs, setSafeTx, setSafeTxError, params])
+    setTxOrigin(getTxOrigin(app))
```
usame-algan: I moved it into this useEffect but let me know if you can think of a better place to handle this logic of setting the tx origin. Previously it was on the Continue button callback together with the TxNote logic.

compojoom: I'm not sure what would happen, but isn't it better if you move the setTxOrigin call inside of the then part?

Right now even if the tx fails and is catched you would set the origin.

usame-algan:
As a side note: We can rename `ReviewTransactionV2` now but I would do it in a different PR to not mess with the diff.

apps/web/src/components/tx-flow/flows/SafeAppsTx/ReviewSafeAppsTx.tsx
usame-algan: Another alternative is that we pass the origin down to the `ReviewTransactionContent` and include it in the `onContinueClick` callback. This would match the previous behaviour but I like this approach more to handle it higher up since it only needs to be done for safe app transactions so no need to add this complexity to the review components.

tmjssz (approved):

compojoom:
@usame-algan can you check if this changes also fixes the tx in the mobile app?

usame-algan:
> @usame-algan can you check if this changes also fixes the tx in the mobile app?

Yes, it also fixes the issue on mobile. Here is a transaction I created on this branch and it shows the logo and name

Screenshot 2025-07-01 at 09 09 25

PR #6073, opened 2025-06-30
fix(mobile): display order recipient and slippage

compojoom:
## What it solves
The swapping info was not displaying a warning when the recipient was a different address and the slippage was not rendered in the UI.

Resolves https://linear.app/safe-global/issue/MOB-11/mobile-swapping-needs-more-love

## How this PR fixes it
- Adds a check for the recipient of a swap
- displays the slippage

## How to test it
Create a test with a new recipient you should see a warning in the confirm view.

## Checklist

---

## CLA signature


apps/mobile/src/features/ConfirmTx/components/confirmation-views/SwapOrder/hooks/useRecipientItems.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { TouchableOpacity } from 'react-native'
+import { OrderTransactionInfo } from '@safe-global/store/gateway/types'
+import { ListTableItem } from '../../../ListTable'
+import { Identicon } from '@/src/components/Identicon'
+import { EthAddress } from '@/src/components/EthAddress'
+import { SafeFontIcon } from '@/src/components/SafeFontIcon'
+import { useOpenExplorer } from '@/src/features/ConfirmTx/hooks/useOpenExplorer'
+import { Address } from '@/src/types/address'
+
+export const useRecipientItems = (order: OrderTransactionInfo): ListTableItem[] => {
```
usame-algan: This sounds like there is more than one item but afaiu its always just the recipient row or will there be more in the future? Otherwise I would suggest to rename and possibly refactor it slightly i.e. instead of pushing into an array it could simply return a new array with one item

usame-algan (approved):
Looks good! Left one comment but overall I can see the warning for the recipient and the slippage.

apps/mobile/src/features/ConfirmTx/components/confirmation-views/SwapOrder/hooks/useRecipientItems.tsx
compojoom: You are right. I think I went with it because listTable expects items... I changed the hook name to useRecipientItem

PR #6075, opened 2025-07-02
refactor(mobile): do the blockaid check automatically on tx confirmation

compojoom:
## What it solves
We were making the blockaid scans first when the user was navigating to the transaction checks screen. Now we start the scans automatically already on the confirm transaction screen. if the scan shows problems we display a warning on the screen.

Resolves https://linear.app/safe-global/issue/MOB-51/mobile-transaction-check-view

## How this PR fixes it
- Refactored teh TransactionInfo component. 
- we have a new transaction-checks component and a hook at that does the heavy lifting on the tx-confirmation screen.
- Some small updates to SafeListItem were necessary to accomodate for the "new design"

## How to test it
Use our dev safe to look failed txs.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/ConfirmTx/components/TransactionChecks/TransactionChecks.tsx
+++ b/apps/mobile/src/features/ConfirmTx/components/TransactionChecks/TransactionChecks.tsx
@@ -1,0 +1,101 @@
+import React from 'react'
+import { SafeFontIcon } from '@/src/components/SafeFontIcon'
+import { SafeListItem } from '@/src/components/SafeListItem'
+import { useRouter } from 'expo-router'
+import { IconName } from '@/src/types/iconTypes'
+import { Alert } from '@/src/components/Alert'
+import { TransactionDetails } from '@safe-global/store/gateway/AUTO_GENERATED/transactions'
+import { CircleSnail } from 'react-native-progress'
+import { useTransactionSecurity } from './hooks/useTransactionSecurity'
+
+interface TransactionChecksProps {
+  txId: string
+  txDetails?: TransactionDetails
+}
+
+export function TransactionChecks({ txId, txDetails }: TransactionChecksProps) {
+  const router = useRouter()
+
+  // Handle all security scanning internally
+  const security = useTransactionSecurity(txDetails)
+
+  const onTransactionChecksPress = () => {
+    router.push({
+      pathname: '/transaction-checks',
+      params: { txId },
+    })
+  }
+
+  const getTransactionChecksIcon = (): IconName => {
+    if (security.hasError) {
+      return 'shield-crossed'
+    }
+    if (security.isMediumRisk || security.hasContractManagement) {
+      return 'alert-triangle'
+    }
+    return 'shield'
+  }
+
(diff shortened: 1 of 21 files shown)


apps/mobile/src/features/ConfirmTx/components/TransactionChecks/TransactionChecks.tsx
```diff
@@ -0,0 +1,101 @@
...
+      return 'alert-triangle'
+    }
+    return 'shield'
+  }
+
+  const getTransactionChecksLeftNode = () => {
+    if (security.isScanning) {
+      return <CircleSnail size={16} borderWidth={0} thickness={1} />
+    }
+    return <SafeFontIcon name={getTransactionChecksIcon()} />
```
usame-algan: There is a bit of a jump here between loading and displaying the final icon because of size differences. The Loading circle has size 16 while the SafeFontIcon has size 24. I suggest to keep them the same size.

apps/mobile/src/features/ConfirmTx/components/TransactionChecks/TransactionChecks.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  const getTransactionChecksBottomContent = () => {
+    if (!security.enabled) {
+      return null
+    }
+
+    // Show warnings for security issues (malicious/warning)
+    if (security.hasIssues) {
+      return <Alert type={getAlertType()} info="Potential risk detected" message="Review details before signing" />
+    }
```
usame-algan: I think the info and message should be the other way around. "Potential risk detected" at the top and below the description.
Screenshot 2025-07-02 at 14 50 24

usame-algan: Another point is that the red on red text is quite hard to read. Can we make it white instead or change the background color to a lighter shade? I couldn't find a design for this on Figma.

apps/mobile/src/features/ConfirmTx/components/TransactionChecks/TransactionChecks.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { CircleSnail } from 'react-native-progress'
+import { useTransactionSecurity } from './hooks/useTransactionSecurity'
+
+interface TransactionChecksProps {
+  txId: string
+  txDetails?: TransactionDetails
+}
+
+export function TransactionChecks({ txId, txDetails }: TransactionChecksProps) {
+  const router = useRouter()
```
usame-algan, line 17: I would suggest testing this component since it contains quite a few branches. Maybe the functions can be extracted to make it easier to test them?

apps/mobile/src/features/ConfirmTx/components/TransactionChecks/TransactionChecks.tsx
compojoom: this is somethign for a different PR. I'm reusing the globally available Alert components

usame-algan (approved):
Thanks!

PR #6076, opened 2025-07-02
Fix(Mobile): Update home screen UI

usame-algan:
## What it solves

Resolves [MOB-82](https://linear.app/safe-global/issue/MOB-82/mobileui-asset-screen)

## How this PR fixes it

- Updates the Pending transactions Alert size, Badge size and text font weight
- Displays the precise fiat amount on the home screen
- Reduces the gap between tokens on the home screen
- Adds a consistent touch area around QR code icon and Notifications icon of 24px
- Reduces the Identicon size

## How to test it

1. Compare the design of the home screen with [Figma](https://www.figma.com/design/yVshG47sZRM7ztI66bDII4/New-Mobile-App?node-id=4542-19475&m=dev)

## Screenshots
Screenshot 2025-07-02 at 14 31 04
Screenshot 2025-07-02 at 14 31 57

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/Assets/components/NFTs/NFTs.container.tsx
+++ b/apps/mobile/src/features/Assets/components/NFTs/NFTs.container.tsx
@@ -16,6 +16,7 @@ import { useDefinedActiveSafe } from '@/src/store/hooks/activeSafe'
 import { NoFunds } from '@/src/features/Assets/components/NoFunds'
 import { AssetError } from '../../Assets.error'
 import { Loader } from '@/src/components/Loader'
+import { getTokenValue } from 'tamagui'
 
 export function NFTsContainer() {
   const activeSafe = useDefinedActiveSafe()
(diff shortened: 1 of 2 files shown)


apps/mobile/src/features/Assets/components/NFTs/NFTs.container.tsx
```diff
@@ -60,6 +60,7 @@ export function NFTsContainer() {
       renderItem={NFTItem}
       ListFooterComponent={isFetching ? <Loader size={24} /> : undefined}
       keyExtractor={(item) => item.id}
+      style={{ marginTop: 8 }}
```
compojoom: just for the protocol, you can also use getTokenValue https://tamagui.dev/docs/core/exports#gettokenvalue

usame-algan: ~I think we could also simply use `'$2'`~ Actually, in this case it doesn't work.

PR #6078, opened 2025-07-02
Feat(web)add MixPanel to the web

vseehausen:
## What it solves

Adds MixPanel data tracking in `autocaption` mode. 

## How this PR fixes it

It adds data tracking only to the web version, not to the mobile version.

The NEXT_PUBLIC_MIXPANEL_TOKEN needs to be set in the release automation.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/services/analytics/useMixpanel.ts
+++ b/apps/web/src/services/analytics/useMixpanel.ts
@@ -1,19 +1,27 @@
 import { useEffect } from 'react'
 import mixpanel from 'mixpanel-browser'
 import { IS_PRODUCTION } from '@/config/constants'
+import { useHasFeature } from '@/hooks/useChains'
+import { FEATURES } from '@safe-global/utils/utils/chains'
 
 /**
  * Initializes Mixpanel analytics for the web app.
  * Should be called once at app startup.
  */
 const useMixpanel = () => {
+  const isMixpanelEnabled = useHasFeature(FEATURES.MIXPANEL)
+
   useEffect(() => {
     // Only initialize in the browser
     if (typeof window === 'undefined') return
+    
+    // Check if Mixpanel feature is enabled
+    if (!isMixpanelEnabled) return
+    
     const token = process.env.NEXT_PUBLIC_MIXPANEL_TOKEN
     if (!token) return
     mixpanel.init(token, { debug: !IS_PRODUCTION })
-  }, [])
+  }, [isMixpanelEnabled])
 }
 
 export default useMixpanel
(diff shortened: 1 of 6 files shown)


apps/web/src/services/analytics/useMixpanel.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+const useMixpanel = () => {
+  useEffect(() => {
+    // Only initialize in the browser
+    if (typeof window === 'undefined') return
+    const token = process.env.NEXT_PUBLIC_MIXPANEL_TOKEN
+    if (!token) return
+    mixpanel.init(token, { debug: !IS_PRODUCTION })
+  }, [])
+}
```
usame-algan: I suggest to add a feature flag here so we can turn it on or off via our config service. There is a hook called `useHasFeature` to handle this and an enum called `FEATURES` that needs to be extended.

apps/web/src/services/analytics/useMixpanel.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Initializes Mixpanel analytics for the web app.
+ * Should be called once at app startup.
+ */
+const useMixpanel = () => {
+  useEffect(() => {
+    // Only initialize in the browser
+    if (typeof window === 'undefined') return
+    const token = process.env.NEXT_PUBLIC_MIXPANEL_TOKEN
```
usame-algan, line 21: In order for this to work in the deployment we need to add the token as a secret to Github and update the `.github/actions/build/action.yml` to pass it into the env on build time.

usame-algan: I can add the token to the github secrets.

vseehausen: Did you add it already?

apps/web/src/services/analytics/useMixpanel.ts
vseehausen: Like this?

usame-algan: Yes, looks good!

apps/web/src/services/analytics/useMixpanel.ts
usame-algan: Yes, and with your change to the action.yml it should be there when it builds and deploys.

vseehausen: Legend!

usame-algan:
Closing in favor of https://github.com/safe-global/safe-wallet-monorepo/pull/6094

PR #6082, opened 2025-07-03
feat(mobile): decoding view for Stake Deposit

compojoom:
## What it solves
Adds confirm transaction decoding for native staking deposit transactions.
 
Resolves https://linear.app/safe-global/issue/MOB-31/mobile-confirmation-screen-stake-deposit

## How to test it
Check the queued tx of this safe:
https://safe-wallet-web.dev.5afe.dev/transactions/queue?safe=hoe:0x870a7f20F82D612b842b62A5cE724271d6991289

## Checklist

---

## CLA signature


apps/mobile/src/features/ConfirmTx/components/confirmation-views/Stake/utils.tsx
```diff
@@ -1,77 +1,76 @@
...
+  // Fee is returned in decimal format, multiply by 100 for percentage
+  const fee = (txInfo.fee * 100).toFixed(2)
 
-// TODO: this function will be replaced for the staking item
-export const formatStakingItems = () => {
   return [
     {
       label: 'Rewards rate',
-      value: '1.27%',
+      value: `${txInfo.annualNrr.toFixed(3)}%`,
```
usame-algan, line 24: Is it possible to use `formatPercentage` here?

apps/mobile/src/features/ConfirmTx/components/confirmation-views/Stake/utils.tsx
```diff
@@ -1,77 +1,76 @@
...
+            textProps={{ fontWeight: 400 }}
+          />
+          <Text color="$textSecondaryLight">({formatCurrency(txInfo.expectedFiatMonthlyReward, CURRENCY)})</Text>
         </View>
       ),
     },
     {
       label: 'Widget fee',
-      value: '0.07%',
+      value: `${fee}%`,
```
usame-algan, line 56: Possible to use `formatPercentage` here?

apps/mobile/src/utils/transactions.tsx
```diff
@@ -38,6 +39,10 @@ export const getTransactionType = ({ txInfo }: { txInfo: Transaction['txInfo'] }
     return ETxType.SWAP_ORDER
   }
 
+  if (isStakingTxDepositInfo(txInfo)) {
+    return ETxType.STAKE
```
usame-algan: I would suggest to be more specific with this type e.g. call it `STAKE_DEPOSIT` since there can also be withdraws right?

usame-algan (approved):

apps/mobile/src/features/ConfirmTx/components/confirmation-views/Stake/utils.tsx
compojoom: apparently not. It's the same implementation in the web app and there we do it like this. Somehow the formatPercentage function doesn't like the value I pass it.

apps/mobile/src/utils/transactions.tsx
compojoom: done

PR #6083, opened 2025-07-03
feat(mobile): add capture screen protection

compojoom:
## What it solves
Disable screen capture on the import private key screen

Resolves https://linear.app/safe-global/issue/MOB-97/add-screenshot-protection

## How this PR fixes it
uses react-native-capture-protection to prevent screen capture (both screenshot,recording, app switch=

## How to test it
This needs to be tested on device - it won't work in the simulator. Install the app on a real device - try to make a screenshot, record a video or switch to a different app. If you are on the private key screen the whole screen will be white/black. If you are on another screen the capturing will work.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/app/import-signers/private-key.tsx
+++ b/apps/mobile/src/app/import-signers/private-key.tsx
@@ -2,9 +2,27 @@ import React from 'react'
 import { ImportPrivateKey } from '@/src/features/ImportPrivateKey'
 import { View } from 'tamagui'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
+import { useFocusEffect } from '@react-navigation/native'
+import { CaptureProtection } from 'react-native-capture-protection'
 
 function PrivateKeyImport() {
   const insets = useSafeAreaInsets()
+
+  // Enable capture protection when screen is focused, disable when unfocused
+  useFocusEffect(
+    React.useCallback(() => {
+      CaptureProtection.prevent({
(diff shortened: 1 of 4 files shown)


apps/mobile/src/app/import-signers/private-key.tsx
```diff
@@ -2,9 +2,27 @@ import React from 'react'
...
+    React.useCallback(() => {
+      CaptureProtection.prevent({
+        screenshot: true,
+        record: true,
+        appSwitcher: true,
+      })
+
+      return () => {
+        CaptureProtection.allow()
+      }
```
usame-algan, line 22: Is it correct that this is inside the `useCallback` instead of the root of the `useFocusEffect`?

compojoom: I think yes: https://reactnavigation.org/docs/use-focus-effect/

> warning
> 
> To avoid the running the effect too often, it's important to wrap the callback in useCallback before passing it to useFocusEffect as shown in the example.
> 
> The useFocusEffect is analogous to React's useEffect hook. The only difference is that it only runs if the screen is currently focused.
> 
> The effect will run whenever the dependencies passed to React.useCallback change, i.e. it'll run on initial render (if the screen is focused) as well as on subsequent renders if the dependencies have changed. If you don't wrap your effect in React.useCallback, the effect will run every render if the screen is focused.

PR #6086, opened 2025-07-04
feat(mobile): add RASP library

compojoom:
## What it solves
RASP, or Runtime Application Self-Protection, is a security technology that protects applications from attacks by integrating security directly into the application itself.

Currently the rasp library is configured to detect

- jailbroken devices
- app tempering
- unofficial instalalations
- hook detection

Resolves https://linear.app/safe-global/issue/MOB-102/integrate-a-rasp-framework

## How this PR fixes it
I've opted to implement https://docs.talsec.app/freerasp as it seems well maintained and used by a lot of other projects. 

## How to test it
An easy way to see the library in action is to build it and have the env flag EXPO_PUBLIC_SECURITY_RASP_ENABLED set to true. Then the app will start complaining that it has been installed from a non-authorised store. 

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/.env.example
+++ b/apps/mobile/.env.example
@@ -1,6 +1,8 @@
 # Used on the javascript side
 EXPO_PUBLIC_APP_VARIANT=production
 EXPO_PUBLIC_INFURA_TOKEN=****
+EXPO_PUBLIC_SECURITY_RASP_ENABLED=false
+EXPO_PUBLIC_SECURITY_WATCHER_MAIL=
 
 # Used during the build process
 APP_VARIANT=production


usame-algan (approved):
Looks good (at least for the unofficial app warning). I had to additionally set `EXPO_PUBLIC_SECURITY_WATCHER_MAIL` for it to work. As discussed lets update the env.example

Image

apps/mobile/src/platform/security.ts
```diff
@@ -0,0 +1,71 @@
...
+  privilegedAccess: () => handleSecurityThreat('privilegedAccess'),
+
+  // https://docs.talsec.app/freerasp/wiki/threat-detection/hook-detection
+  hooks: () => handleSecurityThreat('hooks'),
+
+  // https://docs.talsec.app/freerasp/wiki/threat-detection/app-tampering-detection
+  appIntegrity: () => handleSecurityThreat('appIntegrity'),
+
+  // https://docs.talsec.app/freerasp/wiki/threat-detection/detecting-unofficial-installation
+  unofficialStore: () => handleSecurityThreat('unofficialStore', false),
```
usame-algan, line 68: Is it expected that in case of an unofficial store the app should not close when the user dismisses the warning? For testing I removed the false parameter here and I see this screen. I assume on production this will not be visible.

Image

usame-algan: Together we tested this on a release build with the parameter set to true and the app indeed does crash so all good.

compojoom: > Is it expected that in case of an unofficial store the app should not close when the user dismisses the warning? 

I decided that it is enough to warn the user that he is using an unofficial store installation, but not crash the app in this case. I think that this should be fine.

PR #6087, opened 2025-07-07
Feat(Mobile): Add decoding for cancellation txs in confirmation screen

usame-algan:
## What it solves

Resolves [MOB-50](https://linear.app/safe-global/issue/MOB-50/mobile-confirmation-screen-onchain-rejection)

## How this PR fixes it

- Adds a `CancelTx` component to handle decoding for cancellation txs
- Extends `ETxType`
- Fixes an issue with Transaction checks having additional space at the bottom even if nothing is rendered
- Centers the `Confirm transaction` title

## How to test it

1. Queue a reject tx
2. Open it on mobile
3. Compare with design

## Screenshots

Screenshot 2025-07-07 at 14 23 34

Screenshot 2025-07-07 at 14 23 29

## Checklist

---

## CLA signature


apps/mobile/src/features/ConfirmTx/components/confirmation-views/CancelTx/CancelTx.tsx
```diff
@@ -0,0 +1,74 @@
...
+        }
+        badgeIcon="transaction-contract"
+        badgeColor="$textSecondaryLight"
+        title={txInfo.methodName ?? 'On-chain rejection'}
+        submittedAt={executionInfo.submittedAt}
+      />
+
+      <Text fontSize="$4">
+        This is an on-chain rejection that didn’t send any funds. This on-chain rejection replaced all transactions with
+        nonce {executionInfo.nonce}.
```
compojoom, line 52: Wouldn't it be better if we place this in an Alert2 info style? Also ok if we leave it like this.

apps/mobile/src/components/Alert/Alert.tsx
```diff
@@ -83,6 +84,7 @@ export const Alert = ({
   testID,
   info,
   orientation = 'center',
+  bgcolor = '$background',
```
compojoom: This is a bit weird. The type of the alert represent the Tamagui theme for that element. The tamagui theme has a background and color. Now the component signature changes - you say you want a warning and the warning has a predefined background, but you can override it by passing bgcolor. 

This feels weird. Why can't we have another theme?

usame-algan: I also think its not ideal. We use the Alert type for both the theme and the icons though so it doesn't have the theme type. I will remove it again. We lose the background in light mode but I think its not too bad.

PR #6093, opened 2025-07-08
fix(mobile): show native token info in Action list screen

clovisdasilvaneto:
## What it solves
shows native token info in Action list screen

## How this PR fixes it
We take the useTransferTokenInfo hook from web project and implemented an abstraction of it in the `utils` package. Notice that we still need to use it in the web project to avoid having duplicated implementations across the projects.

## How to test it
- Create a batch transaction with native token transfer.
- Go to the pending transaction details
- Click in the "actions" button in the tx
- The token information should be displayed.

## Screenshots
Screenshot 2025-07-08 at 11 53 42

## Checklist

---

## CLA signature


compojoom:
I think that you forgot to update the web app and make it use the "new" hook.

PR #6097, opened 2025-07-08
fix: swap fee was not properly shown in decoding

compojoom:
## What it solves
CoW have changed the type of the metadata in one of their latest release. Every swap order has a slightly different metadata structure and now the partnerFee object no longer has a `bps` prop. I've updated our code to detect both old order metadata and the new ones.

related to https://linear.app/safe-global/issue/MOB-11/

## How to test it
Try to swap - in the confirmation view you should see the widget fee displayed in the decoding. 
grafik

Also when looking at old swap txs in the history hovering over the "total fee" tooltip should display the explanation and should contain the correct fee applied to the order
grafik

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -38,7 +38,7 @@
     "ios": "expo run:ios"
   },
   "dependencies": {
-    "@cowprotocol/app-data": "^2.3.0",
+    "@cowprotocol/app-data": "^3.1.0",
     "@ethersproject/shims": "^5.7.0",
     "@expo/config-plugins": "~10.0.0",
     "@expo/vector-icons": "^14.1.0",
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -95,7 +95,7 @@
   },
   "devDependencies": {
     "@chromatic-com/storybook": "^1.3.1",
-    "@cowprotocol/app-data": "^2.4.0",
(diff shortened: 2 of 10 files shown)


usame-algan (approved):

liliya-soroka:
why we merged it without the QA ?<br>@usame-algan  and @compojoom

compojoom:
because we needed a mobile release and forgot to move it back

francovenica:
The fee is present in the modal now. It's correct



Question @compojoom <br>Didn't we also show the fees in the tx details? Those are missing there, but I don't remember if we showed it at some point or not.

compojoom:
I looked into the code we are not displaing the fee in the queue. Only when you go to the confirmation screen or if the tx was executed, then the fee is visible in the tooltip of "total fee"

francovenica:
Ok, the fee only shows once the tx is completed and order filled.<br>You can see it in tx in the history tab:<br>https://fix_cow_fee--walletweb.review.5afe.dev/transactions/tx?safe=sep:0x8f4A19C85b39032A37f7a6dCc65234f966F72551&id=multisig_0x8f4A19C85b39032A37f7a6dCc65234f966F72551_0x70702d355b9d29e3e1257437c050bf8a8f280ba5bae00d4efbdbc733da7e0a1c

francovenica:
QA done in the ticket

PR #6098, opened 2025-07-08
fix(mobile): miscellaneous swap order fixes

compojoom:
## What it solves
- display actions on the swap order screen
- better display for buy orders

Resolves https://linear.app/safe-global/issue/MOB-11/

## Checklist

---

## CLA signature


usame-algan (approved):
Not part of this PR but I wonder if we need to cut the order id on the confirmation screen since its a short hash anyway.

PR #6104, opened 2025-07-10
feat(mobile): add rudimentary bridge decoding

compojoom:
## What it solves
Adds bridge decoding in the mobile app.

Resolves https://linear.app/safe-global/issue/MOB-99/

## How this PR fixes it
Implements the same kind of decoding as on web. I had to refactor few things in web to reduce code duplication. Moved few functions to packages/util

## How to test it
Use a safe that has Lifi transactions. TX in the history should be rendered with a card. Try to sign a bridge/swap LiFi transaction - they should be decoded similar to web.

## Checklist

---

## CLA signature


schmanu (approved):
🚀

PR #6107, opened 2025-07-10
fix(mobile): deep links

compojoom:
## What it solves
Per default every screen in the app is accessible through a deep link. That's not a bad thing. Unfortunately it is not ideal if users land on the sign-transaction page. A malicious actor could propose a tx to a safe, then craft a link to the sign-transaction screen and since on that screen we automatically start the sign flow the user could be misled into signing a malicious tx. 
I've added a protection for deep links. If the user clicks on a link leading to a protected page the app would redirect him to the root. If for some reason this redirect fails I've added a guard. The sign-transaction flow can only be started if the user pressed the "sign" button on the confirmTX page. If somehow he manages to directly navigate to signTransaction we won't start the signing process and will display an Alert.

Resolves https://linear.app/safe-global/issue/MOB-106/deep-link-security

## How to test it
You can use the following command to simulate deep links:
npx uri-scheme open "global.safe.mobileapp://sign-transaction?txId={txId}&signerAddress={signerAddress}" --ios

---

## CLA signature


apps/mobile/src/app/+native-intent.tsx
```diff
@@ -0,0 +1,26 @@
...
+  'notifications-opt-in',
+  'biometrics-opt-in',
+  'confirm-transaction',
+]
+export function redirectSystemPath({ path, initial: _initial }: { path: string; initial: boolean }) {
+  try {
+    const isProtectedRoute = protectedRoutes.some((route) => path.includes(route))
+    if (isProtectedRoute) {
+      console.log('trying to navigate to protected route', path)
+      return '/'
```
usame-algan: Should we track how often this happens?

usame-algan (approved):
Looks good! I tested it with a protected route and can see that it redirects and logs it

Image

PR #6111, opened 2025-07-10
Fix(Mobile): Update bottom tab navigation style

usame-algan:
## What it solves

Resolves [MOB-110](https://linear.app/safe-global/issue/MOB-110/mobile-tab-navigation-ui)

## How this PR fixes it

- Shows tab labels
- Uses different icon for home
- Renames Settings to Account

## How to test it

1. Open the app
2. Compare tab navigation with design

## Screenshots

Screenshot 2025-07-10 at 16 47 05

Screenshot 2025-07-10 at 16 47 10

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/app/(tabs)/_layout.tsx
+++ b/apps/mobile/src/app/(tabs)/_layout.tsx
@@ -2,12 +2,33 @@ import { Tabs } from 'expo-router'
 import React from 'react'
 import { TabBarIcon } from '@/src/components/navigation/TabBarIcon'
 import { Navbar as AssetsNavbar } from '@/src/features/Assets/components/Navbar/Navbar'
-import { Pressable, StyleSheet } from 'react-native'
+import { Pressable, StyleSheet, useColorScheme } from 'react-native'
+import { getTokenValue } from 'tamagui'
 
 export default function TabLayout() {
+  const colorScheme = useColorScheme()
+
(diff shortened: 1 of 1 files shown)


apps/mobile/src/app/(tabs)/_layout.tsx
```diff
@@ -2,12 +2,33 @@ import { Tabs } from 'expo-router'
...
+  let activeTintColor, inactiveTintColor, borderTopColor
+  if (colorScheme === 'light') {
+    activeTintColor = getTokenValue('$color.textPrimaryLight')
+    inactiveTintColor = getTokenValue('$color.primaryLightLight')
+    borderTopColor = getTokenValue('$color.borderLightLight')
+  } else {
+    activeTintColor = getTokenValue('$color.textPrimaryDark')
+    inactiveTintColor = getTokenValue('$color.primaryLightDark')
+    borderTopColor = getTokenValue('$color.borderLightDark')
+  }
```
usame-algan, line 20: @compojoom Let me know if there is another way to do this. The variables are the same but using `$textPrimary` inside the style directly doesn't seem to work.

compojoom: it doesn't work, because those components are not Tamagui components and have no idea what $textPrimary is supposed to do. I haven't found a better way to do it.

PR #6112, opened 2025-07-10
chore(web): update @safe-global/safe-deployments to version 1.37.39

PooyaRaki:
## What it solves
This PR updates `@safe-global/safe-deployments` to version `1.37.39` and `@safe-global/safe-modules-deployments` to version `2.2.11`

## Checklist

- [x] The latest @safe-global/safe-deployments version should be used.

---

## CLA signature


compojoom:
this doesn't look correct @PooyaRaki . You need to run for it to also update the lock file. Also this change needs to be made accross multiple packages in the project, so look at the console output to know which package to touch.

PooyaRaki:
@compojoom I have committed the lock file update. There is also the protocol kit that needs to updated. It will be handled later. For now I'll mark the PR as draft.

yagopv (approved):

PR #6116, opened 2025-07-10
fix(mobile): mask input PK key field

compojoom:
## What it solves
I've moved the screen prevention logic to the root of the layout. This way any screen in the import signers process is not screenshotable. Also I set the private key field as secureTextEntry and the OS is automatically masking the input

Resolves https://linear.app/safe-global/issue/MOB-97

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/ImportPrivateKey/ImportPrivateKey.container.tsx
+++ b/apps/mobile/src/features/ImportPrivateKey/ImportPrivateKey.container.tsx
@@ -39,7 +39,6 @@ export function ImportPrivateKey() {
               onChangeText={handlePrivateKeyChange}
               placeholder="Paste here or type..."
               secureTextEntry={true}
-              // multiline
               success={!!wallet}
               textAlign="center"
               error={error}


apps/mobile/src/app/import-signers/_layout.tsx
```diff
@@ -1,7 +1,24 @@
...
+    useCallback(() => {
+      CaptureProtection.prevent({
+        screenshot: true,
+        record: true,
+        appSwitcher: true,
+      })
+
+      return () => {
+        CaptureProtection.allow()
+      }
```
usame-algan, line 18: ~Does this Layout component ever unmount since it handles the Screens? i.e. if I navigate away from the import-signers screen can I take screenshots again?~ Just checked and it indeed does unmount when navigating to a different view.

apps/mobile/src/features/ImportPrivateKey/ImportPrivateKey.container.tsx
```diff
@@ -38,7 +38,8 @@ export function ImportPrivateKey() {
               value={privateKey}
               onChangeText={handlePrivateKeyChange}
               placeholder="Paste here or type..."
-              multiline
+              secureTextEntry={true}
+              // multiline
```
usame-algan: Do we still need multiline?

PR #6117, opened 2025-07-11
feat(web+mobile) MixPanel event tracking

vseehausen:
## What it solves

Adds MixPanel SDK.
Adds user_attributes collection.
Adds tracking of first event APP_LAUNCH.

## Checklist

---

## CLA signature

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2847,24 +2847,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@cowprotocol/app-data@npm:^2.4.0, @cowprotocol/app-data@npm:^2.5.1":
-  version: 2.5.1
-  resolution: "@cowprotocol/app-data@npm:2.5.1"
-  dependencies:
-    ajv: "npm:^8.11.0"
-    cross-fetch: "npm:^3.1.5"
-    ipfs-only-hash: "npm:^4.0.0"
-    json-stringify-deterministic: "npm:^1.0.8"
-    multiformats: "npm:^9.6.4"
-  peerDependencies:
-    cross-fetch: ^3.x
-    ethers: ^5.0.0
-    ipfs-only-hash: ^4.x
-    multiformats: ^9.x
-  checksum: 10/c3f12caad56bd0f5bff8e3f6a399a89fdea2b3f01c9d5ed553c79bca2920cfe66254b17c2070d4d3b6a0e1e8ee41de2f172d0697bca697d704b62ffff78353d0
-  languageName: node
-  linkType: hard
-
 "@cowprotocol/app-data@npm:^3.1.0":
   version: 3.1.0
   resolution: "@cowprotocol/app-data@npm:3.1.0"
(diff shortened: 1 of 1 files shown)


apps/mobile/src/services/analytics/types.ts
```diff
@@ -56,3 +56,51 @@ export type AnalyticsLabel =
...
+}
+
+/**
+ * Event tracking types for MixPanel with humanized property names
+ */
+export interface SafeEventProperties {
+  'Safe ID': string
+  'Safe Version': string
+  Network: string
+  [key: string]: unknown
```
usame-algan: Is this needed?

apps/mobile/src/services/analytics/types.ts
```diff
@@ -56,3 +56,51 @@ export type AnalyticsLabel =
...
+  /** Number of signatures required to execute a transaction (Setup intent) */
+  threshold: number
+
+  /** Blockchain networks (Context for segmentation) */
+  networks: string[]
+
+  /** Timestamp of most recent transaction (Churn alerting) */
+  last_tx_at: Date | null
+
+  /** ID of parent Space if the Safe is grouped in a Space (Space-level grouping) */
```
usame-algan: IMO these comments are more confusing than they help. I would suggest removing them to increase readability of the file.

apps/web/src/services/analytics/mixpanel-tracking.ts
```diff
@@ -0,0 +1,305 @@
...
+const isMixPanelReady = (): boolean => {
+  if (typeof window === 'undefined') return false
+  if (!mixpanel) return false
+
+  try {
+    // Check if MixPanel is initialized by checking for required methods
+    return (
+      typeof mixpanel.track === 'function' &&
+      typeof mixpanel.people === 'object' &&
+      typeof mixpanel.identify === 'function'
```
usame-algan: Why do we have to check the types of these attributes in order to decide if mixpanel is ready? Isn't `if (!mixpanel) return false` above sufficient?

apps/web/src/services/analytics/mixpanel-tracking.ts
```diff
@@ -0,0 +1,305 @@
...
+ * Safely executes MixPanel operations with error handling
+ * @param operation Function to execute
+ * @param operationName Name of the operation for logging
+ */
+const safeMixPanelOperation = (operation: () => void, operationName: string) => {
+  try {
+    if (!isMixPanelReady()) {
+      if (!IS_PRODUCTION) {
+        console.warn(`MixPanel not ready for operation: ${operationName}`)
+      }
```
usame-algan: nit: I would suggest moving this higher up as an early return

apps/web/src/services/analytics/types.ts
```diff
@@ -66,3 +66,61 @@ export enum TxFlowType {
...
+
+/**
+ * Event tracking types for MixPanel with humanized property names
+ */
+export interface SafeEventProperties {
+  'Safe ID': string
+  'Safe Version': string
+  Network: string
+  [key: string]: any
+}
```
usame-algan: Just realized that we have this twice, once for web and once for mobile. If its needed for both I suggest moving it into the utils package.

apps/web/src/services/analytics/user-attributes.ts
```diff
@@ -0,0 +1,173 @@
...
+      } catch (error) {
+        createdAt = new Date()
+      }
+
+      const userAttributes: SafeUserAttributes = {
+        safe_id: safeAddress,
+        created_at: createdAt,
+        safe_version: safeInfo.version || '1.0.0',
+        num_signers: safeInfo.owners?.length || 0,
+        threshold: safeInfo.threshold || 1,
```
usame-algan: Further up we return early if `safeInfo` doesn't exist. Do we still need these fallbacks?

usame-algan (changes requested):
Please only focus on web or mobile for now to keep changes smaller and decide on one specific event e.g. SafeCreated to support. Safe apps are not supported on mobile yet so adding this code for mobile only adds noise.

PR #6120, opened 2025-07-11
Fix(Earn): Rename reward rate to Earn, remove reward rate from withdraw

usame-algan:
## What it solves

Resolves [MOB-56](https://linear.app/safe-global/issue/MOB-56/mobile-kilnmorpho-decodingshow-transactions-on-mobile#comment-489578c8)

## How this PR fixes it

- Renames Reward rate to Earn (after fees)
- Renames Reward rate to Earn inside additional reward block
- Adds the current reward to the tx details on web
- Fixes layout issues with additional reward
- Slightly adjusts Tab color on mobile

## How to test it

1. Open the app
2. Go to pending transactions
3. View a deposit and withdraw
4. Observe no reward rate but instead Earn

## Screenshots

Screenshot 2025-07-11 at 14 40 19
Screenshot 2025-07-11 at 14 40 29

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/features/earn/components/VaultRedeemConfirmation/index.tsx
+++ b/apps/web/src/features/earn/components/VaultRedeemConfirmation/index.tsx
@@ -8,6 +8,7 @@ import { DataTable } from '@/components/common/Table/DataTable'
 import { DataRow } from '@/components/common/Table/DataRow'
 import IframeIcon from '@/components/common/IframeIcon'
 
+// TODO: Check if additional rewards can actually appear for a withdraw/redeem
 const AdditionalRewards = ({ txInfo }: { txInfo: VaultRedeemTransactionInfo }) => {
   if (!txInfo.additionalRewards[0]) return null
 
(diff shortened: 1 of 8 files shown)


apps/web/src/features/earn/components/VaultRedeemConfirmation/index.tsx
```diff
@@ -8,6 +8,7 @@ import { DataTable } from '@/components/common/Table/DataTable'
 import { DataRow } from '@/components/common/Table/DataRow'
 import IframeIcon from '@/components/common/IframeIcon'
 
+// TODO: Check if additional rewards can actually appear for a withdraw/redeem
```
compojoom, line 11: When are we going to check this?

compojoom (approved):

apps/web/src/features/earn/components/VaultRedeemConfirmation/index.tsx
usame-algan: From what I can see is that we do fetch extra rewards on the CGW for every vault interaction but for redeems/withdraws the claimable amount is always 0 thats why we don't display it. I remember Kiln mentioning that claiming these extra rewards will remain as a separate step but we should check with them.

usame-algan: I would leave this for now until we know for certain that this will not be implemented on Kiln side.

PR #6122, opened 2025-07-11
feat(mobile): integrate address book with Receiver component in transaction details

tmjssz:
## What it solves

Resolves [MOB-94](https://linear.app/safe-global/issue/MOB-94/use-names-from-the-addressbook-on-the-tx-details)
This PR enhances the transaction details to display contact names from the address book when available.

## How this PR fixes it
The `Receiver` component now prioritizes contact names from the address book over transaction data. Falls back to transaction data names when no contact exists.

## How to test it
1. Create a transaction to an address that is also in the address book (e.g. to the Safe itself)
2. Navigate to the queued transaction
3. Open the transaction details
4. Observe the address book name above the address value in the `to` field

## Screenshots
Simulator Screenshot - iPhone 15 Pro Max - 2025-07-16 at 18 12 29

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/AdvancedDetails/components/Receiver/Receiver.tsx
+++ b/apps/mobile/src/features/AdvancedDetails/components/Receiver/Receiver.tsx
@@ -3,6 +3,7 @@ import { View, Text } from 'tamagui'
 import { TransactionDetails } from '@safe-global/store/gateway/AUTO_GENERATED/transactions'
 import { Logo } from '@/src/components/Logo'
 import { SafeFontIcon } from '@/src/components/SafeFontIcon'
+import { Identicon } from '@/src/components/Identicon'
 import { useAppSelector } from '@/src/store/hooks'
 import { selectContactByAddress } from '@/src/store/addressBookSlice'
 
@@ -32,7 +33,11 @@ export function Receiver({ txData }: ReceiverProps) {
       gap="$2"
       alignSelf="flex-start"
     >
-      <Logo logoUri={logoUri} size="$4" />
+      <Logo
+        logoUri={logoUri}
+        size="$4"
+        fallbackContent={value ? <Identicon address={value as `0x${string}`} size={16} /> : undefined}
+      />
       <Text fontWeight={600}>{content}</Text>
       <SafeFontIcon name="check-oulined" color="$success" size={16} style={{ marginLeft: 'auto' }} />
     </View>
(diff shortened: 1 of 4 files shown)


apps/mobile/src/features/AdvancedDetails/components/Receiver/Receiver.tsx
```diff
@@ -3,13 +3,19 @@ import { View, Text } from 'tamagui'
...
+import { useAppSelector } from '@/src/store/hooks'
+import { selectContactByAddress } from '@/src/store/addressBookSlice'
 
 interface ReceiverProps {
   txData: TransactionDetails['txData']
 }
 
 export function Receiver({ txData }: ReceiverProps) {
-  const content = txData?.to.name
+  const { to: { value = '', name, logoUri } = {} } = txData || {}
```
compojoom, line 15: In the addressBookSlice the contact type is defined as:
export type Contact = Omit<AddressInfo, 'name'> & {
  name: string
  chainIds: string[]
}

maybe you could do the same here for to? 

```
 Omit<AddressInfo, 'name'> & {
  name: string
  }

```

apps/mobile/src/features/AdvancedDetails/components/Receiver/Receiver.tsx
```diff
@@ -26,7 +32,7 @@ export function Receiver({ txData }: ReceiverProps) {
       gap="$2"
       alignSelf="flex-start"
     >
-      <Logo logoUri={txData?.to.logoUri} size="$4" />
+      <Logo logoUri={logoUri} size="$4" />
```
compojoom: I am a bit confused here. I think that despite the fact that we have logoUri as something that might come from CGW - most of the time we return null. For contacts (and not just generic contracts) I would say that we should display the identicon, no?

tmjssz: We already show the identicon with the address value below. So it would be shown twice if we rendered it here as well which looks a bit weird to me.
Screenshot 2025-07-15 at 11 24 49

tmjssz: Maybe we could just render no image at all if the logoUri is null?
Screenshot 2025-07-15 at 11 30 04

apps/mobile/src/features/AdvancedDetails/components/Receiver/Receiver.tsx
tmjssz: We could do that but if we restrict the `ReceiverProps` interface we’ll need to add conditional checks for the txData prop when rendering `<Receiver ..>` instead of being able to pass in any txData object directly. The idea was to centralize those checks within the Receiver component itself.

PR #6126, opened 2025-07-14
fix(web): show address instead of contact in address book entry

LucieFaire:
## What it solves [COR-150](https://linear.app/safe-global/issue/COR-150/web-address-book-contact-is-visible-instead-of-address)

Resolves #

Web Address Book: "Contact" is visible instead of "Address"

## How this PR fixes it

Renamed `Contract` to `Address`

## How to test it

- Open Address Book
- Add a new entry
- Check that Address is shown instead of Contact

## Screenshots
Screenshot 2025-07-14 at 13 10 08

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/address-book/EntryDialog/index.tsx
+++ b/apps/web/src/components/address-book/EntryDialog/index.tsx
@@ -72,7 +72,7 @@ function EntryDialog({
             <Box>
               <AddressInput
                 name="address"
-                label="Contact"
+                label="Address"
                 variant="outlined"
                 fullWidth
                 required


LucieFaire:
> I think we might have changed this as part of Spaces but it shouldn't be an issue to revert.

Is there going to be a new UI? should I not merge it? Wouldn't want to create more work :)

usame-algan:
> Is there going to be a new UI? should I not merge it? Wouldn't want to create more work :)

Actually just saw that it was changed prior to Spaces. Lets merge it.

PR #6130, opened 2025-07-14
Feat(Analytics) add MixPanel event tracking

vseehausen:
## What it solves

Adds MixPanel SDK.
Adds user_attributes collection.
Adds tracking of first event APP_LAUNCH.

## Checklist

---

## CLA signature


apps/web/src/hooks/analytics/useMixPanelUserTracking.ts
```diff
@@ -0,0 +1,284 @@
...
+      // Small delay to ensure MixPanel is fully initialized
+      const timer = setTimeout(() => {
+        setIsInitialized(true)
+      }, 1000)
+
+      return () => clearTimeout(timer)
+    } else {
+      setIsInitialized(false)
+    }
+  }, [isMixPanelEnabled])
```
compojoom: Does this make sense? Your useMixPanelEnabled hook already seems to check if mixpanel is available on th ewindow? Why do you need the delay here?

apps/web/src/hooks/analytics/useMixPanelUserTracking.ts
```diff
@@ -0,0 +1,284 @@
...
+ * Hook to track Safe user attributes in MixPanel
+ *
+ * This hook:
+ * 1. Collects user attributes from Safe data
+ * 2. Identifies the user using Safe address
+ * 3. Sets user attributes in MixPanel
+ * 4. Updates attributes when Safe data changes
+ * 5. Registers super properties for events
+ */
+export const useMixPanelUserTracking = () => {
```
compojoom, line 24: @vseehausen - this hook is too complicated. You have 5 useEffects. We need to refactor this in more modular hooks then, you can have one "big" that wires the small one together.

usame-algan: I suggest to take a look at how we handle this with GA. There is one component to initialize GA called `Analytics`, one hook to set all user attributes and keep them updated called `useGTM` and one helper file with methods that we can use to emit events called `gtm.ts`

So all we have to do for new events is call the existing functions from `gtm.ts` in the right place. No need to add new hooks for this.

apps/web/src/hooks/analytics/useMixPanelUserTracking.ts
vseehausen: Alright, removed.

apps/web/src/hooks/analytics/useMixPanelUserTracking.ts
vseehausen: Hey, thanks a lot. I tried to address this. What do you think of the current architecture?

PR #6131, opened 2025-07-14
Fix(Mobile): Transaction list icon sizes and fallback image

usame-algan:
## What it solves

Resolves part of [COR-223](https://linear.app/safe-global/issue/COR-223/get-at-least-rudemtary-decoding-for-bridge-txs)

## How this PR fixes it

- Reduces the size of all transaction info card icons to 32px instead of 40px
- Adds `useValidLogoUri` hook that returns null in case a logo can't be loaded (404 or 403)
- Fixes amount formatting for bridge and lifi amounts

## How to test it

1. Open an app with pending transaction
2. Click through the pending transaction list and history
3. Observe all icons are the same size and 32px
4. Observe fallback images and no black background for icons
5. Check bridge amount and make sure its formatted the same way as on web

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/components/TxInfo/TxInfo.tsx
+++ b/apps/mobile/src/components/TxInfo/TxInfo.tsx
@@ -1,5 +1,4 @@
 import React, { useCallback } from 'react'
-import { TransactionInfoType } from '@safe-global/store/gateway/types'
 import { type Transaction } from '@safe-global/store/gateway/AUTO_GENERATED/transactions'
 import { useTransactionType } from '@/src/hooks/useTransactionType'
 import { TxTokenCard } from '@/src/components/transactions-list/Card/TxTokenCard'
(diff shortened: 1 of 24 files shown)


apps/mobile/src/components/TxInfo/TxInfo.tsx
```diff
@@ -70,7 +69,7 @@ function TxInfoComponent({ tx, onPress, ...rest }: TxInfoProps) {
     return <TxSettingsCard onPress={onCardPress} executionInfo={tx.executionInfo} txInfo={txInfo} {...rest} />
   }
 
-  if (isMultiSendTxInfo(txInfo) && tx.txInfo.type === TransactionInfoType.CUSTOM) {
+  if (isMultiSendTxInfo(txInfo)) {
```
compojoom: On line 84 we have
if (isMultiSendTxInfo(txInfo) && tx.safeAppInfo) {

this condition is now never going to be reached, no?

compojoom:
something looks werid with the token display:
grafik
I think that according to the designs they need to overlap way more

apps/mobile/src/components/TxInfo/TxInfo.tsx
usame-algan: Interesting, I removed this condition because it's already part of isMultiSendInfo so I guess the other condition was never reached in the first place.

usame-algan: The safe app card was indeed never reached. I now adjusted the condition to only show the batch transaction card if its a multisend and NOT a safe app transaction and to show the safe app card if its a multisend and has safe app info.

compojoom (approved):
👍

PR #6133, opened 2025-07-15
Feat(Web) Add Mixpanel event tracking

vseehausen:
## What it solves

* Adds Mixpanel event tracking following the implementation of GA tracking closely.
* Adds tracking for Safe App Launch.

Resolves https://linear.app/safe-global/issue/GRO-1/integrate-mixpanel-sdk-for-session-tracking

## Checklist

---

## CLA signature


apps/web/src/services/analytics/useMixpanel.ts
```diff
@@ -0,0 +1,110 @@
...
+  // Initialize MixPanel (only if feature is enabled)
+  useEffect(() => {
+    if (isMixpanelEnabled) {
+      mixpanelInit()
+    }
+  }, [isMixpanelEnabled])
+
+  // Enable/disable tracking based on consent
+  useEffect(() => {
+    if (!isMixpanelEnabled) return
```
usame-algan, line 54: I think its not necessary to check for the feature flag in every useEffect because if mixpanel is not initialized it doesn't matter if we call these functions so I would suggest to remove this condition from all useEffects in this file except the initialization.

apps/web/src/services/analytics/mixpanel.ts
```diff
@@ -0,0 +1,187 @@
...
+import type { SafeAppData } from '@safe-global/safe-gateway-typescript-sdk'
+import { IS_PRODUCTION } from '@/config/constants'
+import { DeviceType } from './types'
+import packageJson from '../../../package.json'
+
+const commonEventParams = {
+  'App Version': packageJson.version,
+  'Chain ID': '',
+  'Device Type': DeviceType.DESKTOP,
+  'Safe Address': '',
```
usame-algan: I suggest to define an `MixPanelEventParams` enum to map to these strings to avoid possible mistakes when using them. It could live inside the `mixpanel-events.ts` with the other enums.

```
export enum MixPanelEventParams {
    APP_VERSION: 'App Version',
    CHAIN_ID: 'Chain ID',
    ...
}
```

apps/web/src/services/analytics/mixpanel.ts
```diff
@@ -0,0 +1,187 @@
...
+    console.info('[MixPanel] - User identified:', userId)
+  }
+}
+
+/**
+ * Check if MixPanel is initialized
+ */
+export const isMixPanelReady = (): boolean => {
+  return isMixPanelInitialized
+}
```
usame-algan: This is not used so I suggest to remove it

apps/web/src/components/safe-apps/SafeAppPreviewDrawer/index.tsx
```diff
@@ -37,6 +45,10 @@ const SafeAppPreviewDrawer = ({ isOpen, safeApp, isBookmarked, onClose, onBookma
     if (safeApp) {
       markSafeAppOpened(safeApp.id)
       trackSafeAppEvent({ ...SAFE_APPS_EVENTS.OPEN_APP, label: SAFE_APPS_LABELS.apps_sidebar }, safeApp.name)
+      trackMixPanelEvent(
+        MixPanelEvent.SAFE_APP_LAUNCHED,
+        safeAppToMixPanelEventProperties(safeApp, SafeAppLaunchLocation.PREVIEW_DRAWER),
+      )
```
usame-algan: Nice, this is very clean!

usame-algan (commented):
What about the other properties such as `Number of Signers, Threshold`. Should we add them in this PR?

apps/web/src/services/analytics/useMixpanel.ts
```diff
@@ -0,0 +1,110 @@
...
+  // Set safe address for all MixPanel events and identify user
+  useEffect(() => {
+    if (!isMixpanelEnabled) return
+
+    mixpanelSetSafeAddress(safeAddress)
+
+    if (safeAddress && !isSpaceRoute) {
+      // Identify user by safe address for better user tracking
+      mixpanelIdentify(safeAddress)
+    }
```
usame-algan: Do we need to track this as both the User ID and a custom property Safe Address or would it be enough to only track it as the User ID?

vseehausen: I would like to track it additionally for the case we change the user id. Cost is very low.

usame-algan: Makes sense, thanks!

apps/web/src/services/analytics/useMixPanelUserProperties.ts
```diff
@@ -0,0 +1,87 @@
...
+    const currentNetworkName = currentChain.chainName.toLowerCase()
+
+    // Calculate transaction count and last transaction from history
+    let totalTxCount = 0
+    let lastTxAt: Date | null = null
+
+    if (txHistory.data?.results) {
+      const transactions = txHistory.data.results.filter(isTransactionListItem).map((item) => item.transaction)
+
+      totalTxCount = transactions.length
```
usame-algan: This will not work because we only get the most recent 20 transactions in our request to the backend. You can instead use the `safe.nonce` which tells us how many transactions in total were executed on a safe.

apps/web/src/services/analytics/useMixPanelUserProperties.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Hook to get formatted user properties for MixPanel tracking
+ *
+ * This hook collects Safe-related user properties that can be used for
+ * MixPanel user attribute tracking and cohort analysis.
+ * Returns both regular properties and networks separately for different MixPanel operations.
+ */
+export const useMixPanelUserProperties = (): MixPanelUserPropertiesFormatted | null => {
+  const { safe, safeLoaded } = useSafeInfo()
+  const currentChain = useCurrentChain()
```
usame-algan: This is also part of `safe` in the form of chain ids. Lets use one source of truth for this because `useCurrentChain` has fallback logic which could falsify this info.

apps/web/src/services/analytics/useMixPanelUserProperties.ts
```diff
@@ -0,0 +1,87 @@
...
+        const timestamps = transactions
+          .map((tx) => tx.timestamp)
+          .filter((timestamp) => timestamp !== null)
+          .sort((a, b) => b - a) // Sort descending (most recent first)
+
+        if (timestamps.length > 0) {
+          lastTxAt = new Date(timestamps[0])
+        }
+      }
+    }
```
usame-algan: You can simplify this because the most recent transaction will first in the array i.e. `transactions[0]`

apps/web/src/services/analytics/useMixPanelUserProperties.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // Create networks array with current network
+    // Note: This starts with current network, but should be extended to include
+    // all networks where this Safe has been active. For now, we append the current
+    // network to any existing networks to avoid overwriting.
+    const networks = [currentNetworkName]
```
usame-algan: You can get this information through the `useAllSafesGrouped` hook. It returns all multichain safes which can be iterated to get the chain ids that the current safe is created on.

apps/web/src/services/analytics/useMixPanelUserProperties.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // Create MixPanel properties object with string keys
```
usame-algan: nit: This comment is not necessary.

vseehausen:
Things that jump to the eye:

* network name should be set in common event params
* mixpanelUserPropoerties should rather have one parameter configuration.
* enum constants create confusion
* mixpanel.ts l33 can be merged
* mixpanel-events.ts line breaks remove and remove comments
* in general, event attributes should could be an object with label and value. 
* useEffects of useMixpanel should not be copied from ga, but rather consolidated

apps/web/src/services/analytics/__tests__/mixpanel.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        action: 'click',
+      }
+
+      trackEvent(eventData)
+
+      // Should NOT call MixPanel track (only GA)
+      expect(mockMixpanel.track).not.toHaveBeenCalled()
+    })
+
+    it('should track with MixPanel only when using trackMixPanelEvent', () => {
```
compojoom, line 130: I think that the test is OK, but since it comes afte "shouldtrack with GA when using trackEvent" I was expecting to see a check that GA was not called when trackMixPanelEvent was triggered. 🤷

apps/web/src/services/analytics/__tests__/useMixPanelUserProperties.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+    return chains[chainId] || null
+  }),
+}))
+
+jest.mock('@/hooks/useSafeInfo', () => ({
+  __esModule: true,
+  default: jest.fn(() => ({
+    safe: {
+      address: { value: '0x1234567890123456789012345678901234567890' },
```
compojoom: I'm ok with this, but I know that other people on the team would argue to use faker to create the addresses.  So please update the tests to use fakerjs for the mocks

apps/web/src/services/analytics/mixpanel.ts
```diff
@@ -0,0 +1,152 @@
...
+  [MixPanelEventParams.EOA_WALLET_ADDRESS]: '',
+  [MixPanelEventParams.EOA_WALLET_NETWORK]: '',
+}
+
+let isMixPanelInitialized = false
+
+export const mixpanelInit = (): void => {
+  if (typeof window === 'undefined' || isMixPanelInitialized) return
+
+  const token = process.env.NEXT_PUBLIC_MIXPANEL_TOKEN
```
compojoom: process.env.NEXT_PUBLIC_MIXPANEL_TOKEN - export a const in teh constants file and use it here.

apps/web/src/services/analytics/mixpanel.ts
```diff
@@ -0,0 +1,152 @@
...
+  commonEventParams[MixPanelEventParams.EOA_WALLET_NETWORK] = network
+
+  if (isMixPanelInitialized) {
+    mixpanel.register({ [MixPanelEventParams.EOA_WALLET_NETWORK]: network })
+  }
+}
+
+/**
+ * Convert SafeApp object to MixPanel event properties
+ */
```
compojoom: the comment is what the function name says.

apps/web/src/services/analytics/mixpanel.ts
```diff
@@ -0,0 +1,152 @@
...
+export const mixpanelSetDeviceType = (type: DeviceType): void => {
+  commonEventParams[MixPanelEventParams.DEVICE_TYPE] = type
+
+  if (isMixPanelInitialized) {
+    mixpanel.register({ [MixPanelEventParams.DEVICE_TYPE]: type })
+  }
+}
+
+export const mixpanelSetSafeAddress = (safeAddress: string): void => {
+  commonEventParams[MixPanelEventParams.SAFE_ADDRESS] = safeAddress
```
compojoom: Why are we setting this on the commonEventParams if mixpanel is not initialized?

apps/web/src/services/analytics/mixpanel.ts
```diff
@@ -0,0 +1,152 @@
...
+    console.error('[MixPanel] - Initialization failed:', error)
+  }
+}
+
+export const mixpanelSetBlockchainNetwork = (networkName: string): void => {
+  commonEventParams[MixPanelEventParams.BLOCKCHAIN_NETWORK] = networkName
+
+  if (isMixPanelInitialized) {
+    mixpanel.register({ [MixPanelEventParams.BLOCKCHAIN_NETWORK]: networkName })
+  }
```
compojoom: We are repeating the same if check in nearly every function here. How about if we abstract this into a "safeMixPanelRegister" function something like:
 

> const safeMixPanelRegister = (properties: Record<string, any>): void => {
>     if (isMixPanelInitialized) {
>       mixpanel.register(properties);
>     }
>   };

and there you do the isMixPanelInitialized check? 
Here you would just do:
safeMixPanelRegister({....})
then you won't have to do the if (...) check in every function

apps/web/src/services/analytics/mixpanel.ts
vseehausen: Why is this preferred over the environment?

compojoom (approved):
👍

liliya-soroka:
regression e2e automation  was run and checked , no new impacted failed cases

The functionality testing according to the requirements was done by @devpramoth

PR #6135, opened 2025-07-16
feat(mobile): add remove private key option

compojoom:
## What it solves
 This PR adds:
- view of the private key that was imported
- option to delete the private key
- deleting a safe now deletes the corresponding private keys
- cleans up any delegates created for a PK and removes any associated push notification subscription
- refactored the screen capture protection that we introduced in a previous PR

resolves https://linear.app/safe-global/issue/COR-327/implement-private-key-removal

##  How this PR fixes it

1. Private Key Display Feature
- New Route: Added /signers/[address]/private-key route with screen capture protection
- PrivateKey Components: Created PrivateKeyContainer and PrivateKeyView components for secure private key display
- Navigation Integration: Added private key screen to the signers layout with proper header configuration

2. Comprehensive Safe Deletion with Cleanup

- Cleanup Logic:
  - isOwnerInOtherSafes() - Identifies which private keys are safe to delete
  - getSafeOwnersWithPrivateKeys() - Finds owners with stored private keys
  - getOwnersToDelete() - Determines which owners can be safely removed
   - createDeletionMessage() - Generates clear user messaging about what will be deleted
  
- Cleanup Process:
    a. Notification Cleanup: Removes push notification registrations
    b. Backend Removal: Removes delegates from transaction service across all chains
    c. Keychain Cleanup: Removes private keys from device keychain
    d. Store Updates: Updates Redux store to reflect changes

3. Screen Protection
Since we also need to protect the new private key screen, I abstracted the logic into a useScreenProtection hook

4. Service Architecture
  - DelegateCleanupService: Centralized service for orchestrating delegate cleanup with progress tracking
  - Error Handling: error handling with standardized error types and user-friendly messages
  - Progress Tracking: Real-time progress updates during cleanup operations (not displaying it though)
  - Retry Logic: Built-in retry mechanisms for network failures and rate limiting

## How to test it

1. Private Key Display Testing
  1. Navigate to Settings > Signers
  2. Select a signer that has a private key
  3. Tap "View Private Key" button
  4. Verify screen protection is active (screenshot/recording blocked)
  5. Confirm private key is displayed correctly
  6. Test navigation back to signer details

2. Safe Deletion Testing

  # Test Safe Deletion with Private Key Cleanup
  1. Create multiple Safes with shared owners
  2. Navigate to My Accounts and enter edit mode
  3. Delete a Safe that has owners with private keys
  4. Verify deletion confirmation shows correct message about private key impact
  5. Confirm only appropriate private keys are deleted
  6. Verify delegates are properly removed from backend
  7. Check that keychain is cleaned up correctly
  8. Ensure Redux store reflects changes

3. Error Handling Testing

  # Test Network Failures
  1. Disconnect from internet
  2. Attempt to delete a Safe
  3. Verify appropriate error messages are shown
  4. Test retry functionality when connection is restored

4. Integration Testing

  # Test the complete flow end-to-end
  1. Import multiple private keys
  2. Create multiple Safes with overlapping owners
  3. Navigate through the private key display flow
  4. Test Safe deletion with various scenarios:
     - Safe with unique owners
     - Safe with shared owners
     - Safe with no stored private keys
     - Network failures during cleanup
  5. Verify all cleanup phases complete successfully

## Screenshots
grafik

grafik

grafik

grafik

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/app/signers/[address]/private-key.tsx
+++ b/apps/mobile/src/app/signers/[address]/private-key.tsx
@@ -1,0 +1,16 @@
+import React from 'react'
+import { useLocalSearchParams } from 'expo-router'
+import { PrivateKeyContainer } from '@/src/features/PrivateKey/PrivateKey.container'
+import { type Address } from '@/src/types/address'
+import { useScreenProtection } from '@/src/hooks/useScreenProtection'
+
+export default function PrivateKeyScreen() {
+  useScreenProtection()
+  const { address } = useLocalSearchParams<{ address: string }>()
+
+  if (!address) {
+    throw new Error('Signer address is required')
+  }
+
+  return <PrivateKeyContainer signerAddress={address as Address} />
+}
--- a/apps/mobile/src/components/SafeListItem/SafeListItem.tsx
+++ b/apps/mobile/src/components/SafeListItem/SafeListItem.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import { Container } from '../Container'
-import { Text, Theme, ThemeName, View, YStackProps } from 'tamagui'
+import { Text, Theme, ThemeName, View, ViewProps, YStackProps } from 'tamagui'
 import { IconProps, SafeFontIcon } from '../SafeFontIcon/SafeFontIcon'
 import { ellipsis } from '@/src/utils/formatters'
 import { isMultisigExecutionInfo } from '@/src/utils/transaction-guards'
@@ -26,6 +26,7 @@ export interface SafeListItemProps {
   tag?: string
   paddingVertical?: YStackProps['paddingVertical']
   bottomContent?: React.ReactNode
+  pressStyle?: ViewProps['pressStyle']
 }
 
 export function SafeListItem({
@@ -45,6 +46,7 @@ export function SafeListItem({
   tag,
   paddingVertical = '$4',
   bottomContent,
+  pressStyle,
 }: SafeListItemProps) {
   // TODO: Replace this with proposedByDelegate once EN-149 is implemented
   const isProposedTx = isMultisigExecutionInfo(executionInfo) ? executionInfo.confirmationsSubmitted === 0 : false
@@ -62,6 +64,8 @@ export function SafeListItem({
       flexDirection="column"
       justifyContent="flex-start"
       paddingVertical={paddingVertical}
+      // If just set pressStyle to undefined, then the onPress doesn't work, that's why we need this hack
+      {...(pressStyle ? { pressStyle } : {})}
     >
       <View flexDirection="row" width="100%" alignItems="center" justifyContent="space-between">
         <View flexDirection="row" maxWidth={rightNode ? '55%' : '100%'} alignItems="center" gap={12}>
--- a/apps/mobile/src/features/PrivateKey/PrivateKey.container.tsx
+++ b/apps/mobile/src/features/PrivateKey/PrivateKey.container.tsx
@@ -1,0 +1,118 @@
+import React, { useCallback, useState } from 'react'
+import { Alert } from 'react-native'
+import { useRouter } from 'expo-router'
+import { PrivateKeyView } from './components/PrivateKeyView'
+import { keyStorageService } from '@/src/services/key-storage'
+import { useDelegateCleanup } from '@/src/hooks/useDelegateCleanup'
+import { useAppDispatch } from '@/src/store/hooks'
+import { type Address } from '@/src/types/address'
+import { cleanupSinglePrivateKey } from '@/src/features/AccountsSheet/AccountItem/utils/editAccountHelpers'
+
+type Props = {
+  signerAddress: Address
+}
+
+export const PrivateKeyContainer = ({ signerAddress }: Props) => {
+  const router = useRouter()
+  const dispatch = useAppDispatch()
+  const { removeAllDelegatesForOwner } = useDelegateCleanup()
+
+  const [isKeyVisible, setIsKeyVisible] = useState(false)
+  const [privateKey, setPrivateKey] = useState<string | null>(null)
+  const [isLoading, setIsLoading] = useState(false)
+
+  const executeViewPrivateKey = useCallback(async () => {
+    setIsLoading(true)
+
+    try {
+      const key = await keyStorageService.getPrivateKey(signerAddress)
+
+      if (!key) {
+        Alert.alert('Error', 'Private key not found')
+        return
+      }
+
+      setPrivateKey(key)
+      setIsKeyVisible(true)
+    } catch (error) {
+      console.error('Error retrieving private key:', error)
+      Alert.alert('Error', 'Failed to retrieve private key')
+    } finally {
+      setIsLoading(false)
+    }
+  }, [signerAddress])
+
+  const handleViewPrivateKey = useCallback(() => {
+    Alert.alert(
+      'View private key',
+      'Are you sure you want to display your private key on screen? Make sure no one else can see your screen.',
+      [
+        { text: 'Cancel', style: 'cancel' },
+        {
+          text: 'Yes, show key',
+          style: 'destructive',
+          onPress: executeViewPrivateKey,
+        },
+      ],
+    )
+  }, [executeViewPrivateKey])
+
+  const showDeleteFailureAlert = useCallback((message?: string) => {
+    Alert.alert(
+      'Cannot delete private key',
+      message || 'Failed to unsubscribe from push notifications. Please check your internet connection and try again.',
+      [{ text: 'OK' }],
+    )
+  }, [])
+
+  const executeDeletePrivateKey = useCallback(async () => {
+    setIsLoading(true)
+
+    try {
+      const result = await cleanupSinglePrivateKey(signerAddress, removeAllDelegatesForOwner, dispatch)
+
+      if (!result.success) {
+        showDeleteFailureAlert(result.error?.message)
+        return
+      }
+
+      router.back()
+      Alert.alert('Success', 'Private key has been deleted successfully')
+    } catch (_error) {
+      showDeleteFailureAlert('An unexpected error occurred')
+    } finally {
+      setIsLoading(false)
+    }
+  }, [signerAddress, dispatch, removeAllDelegatesForOwner, router, showDeleteFailureAlert])
+
+  const handleDeletePrivateKey = useCallback(() => {
+    Alert.alert(
+      'Delete private key',
+      'This will make this signer no longer able to sign transactions in this safe and in any other safe on this device that uses this private key. Do you want to proceed?',
+      [
+        { text: 'Cancel', style: 'cancel' },
+        {
+          text: 'Yes, delete',
+          style: 'destructive',
+          onPress: executeDeletePrivateKey,
+        },
+      ],
+    )
+  }, [executeDeletePrivateKey])
+
+  const handleHidePrivateKey = useCallback(() => {
+    setIsKeyVisible(false)
+    setPrivateKey(null)
+  }, [])
+
+  return (
+    <PrivateKeyView
+      isKeyVisible={isKeyVisible}
+      privateKey={privateKey}
+      isLoading={isLoading}
+      onViewPrivateKey={handleViewPrivateKey}
+      onDeletePrivateKey={handleDeletePrivateKey}
+      onHidePrivateKey={handleHidePrivateKey}
+    />
+  )
+}
--- a/apps/mobile/src/features/PrivateKey/components/PrivateKeyView.tsx
+++ b/apps/mobile/src/features/PrivateKey/components/PrivateKeyView.tsx
@@ -17,7 +17,7 @@ type Props = {
 }
 
 // Generate a fake 64-character hex string for display when key is hidden
-const MASKED_PRIVATE_KEY = '0x' + '•'.repeat(64)
+const MASKED_PRIVATE_KEY = '•'.repeat(64)
 
 export const PrivateKeyView = ({
   isKeyVisible,
--- a/apps/mobile/src/features/PrivateKey/components/PrivateKeyView.tsx
+++ b/apps/mobile/src/features/PrivateKey/components/PrivateKeyView.tsx
@@ -1,0 +1,75 @@
+import React from 'react'
+import { ScrollView, View, Text, YStack } from 'tamagui'
+import { Container } from '@/src/components/Container'
+import { CopyButton } from '@/src/components/CopyButton'
+import { SafeButton } from '@/src/components/SafeButton'
+import { KeyboardAvoidingView, ActivityIndicator } from 'react-native'
+import { useSafeAreaInsets } from 'react-native-safe-area-context'
+import { SafeInput } from '@/src/components/SafeInput'
+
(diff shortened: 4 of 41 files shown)


apps/mobile/src/app/signers/[address]/private-key.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default function PrivateKeyScreen() {
+  useScreenProtection()
+  const { address } = useLocalSearchParams<{ address: string }>()
+
+  if (!address) {
+    throw new Error('Signer address is required')
+  }
+
+  return <PrivateKeyContainer signerAddress={address as Address} />
```
Comment (usame-algan) [inline apps/mobile/src/app/signers/[address]/private-key.tsx:15]: nit: You can also add the `Address` type further up in line 9 instead of using string and remove it here to prevent casting

apps/mobile/src/components/SafeListItem/SafeListItem.tsx
```diff
@@ -62,6 +64,8 @@ export function SafeListItem({
       flexDirection="column"
       justifyContent="flex-start"
       paddingVertical={paddingVertical}
+      // If just set pressStyle to undefined, then the onPress doesn't work, that's why we need this hack
```
usame-algan, line 67: I can't reproduce this. If I don't pass a pressStyle the onPress still works for the "View private key" button on iOS

apps/mobile/src/features/PrivateKey/PrivateKey.container.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    try {
+      const result = await cleanupSinglePrivateKey(signerAddress, removeAllDelegatesForOwner, dispatch)
+
+      if (!result.success) {
+        showDeleteFailureAlert(result.error?.message)
+        return
+      }
+
+      router.back()
```
usame-algan, line 79: Does this unmount the component and removes the `privateKey` state from storage or should we explicitely set it to null before going back?

apps/mobile/src/features/PrivateKey/components/PrivateKeyView.tsx
```diff
@@ -0,0 +1,75 @@
...
+  isKeyVisible: boolean
+  privateKey: string | null
+  isLoading: boolean
+  onViewPrivateKey: () => void
+  onDeletePrivateKey: () => void
+  onHidePrivateKey: () => void
+}
+
+// Generate a fake 64-character hex string for display when key is hidden
+const MASKED_PRIVATE_KEY = '0x' + '•'.repeat(64)
```
usame-algan: nit: 0x is a bit misleading since a private key doesn't start with that. IMO it would be fine to only display the dots.

apps/mobile/src/features/Settings/components/Navbar/SettingsMenu.tsx
```diff
@@ -113,13 +113,25 @@ export const SettingsMenu = ({ safeAddress }: Props) => {
...
-                    deleteSafe(safeAddress as Address)
-
-                    toast.show(`The safe with address ${safeAddress} was deleted.`, {
-                      native: true,
-                      duration: 2000,
-                    })
+                  onPress: async () => {
+                    try {
+                      await deleteSafe(safeAddress as Address)
+                      // Only show toast if we reach this point (deletion was successful)
```
usame-algan: nit: Unnecessary comment

apps/mobile/src/features/Signer/components/SignerView.tsx
```diff
@@ -89,16 +92,21 @@ export const SignerView = ({
...
+        )}
       </ScrollView>
       <KeyboardAvoidingView behavior="padding" keyboardVerticalOffset={top + bottom}>
         <View paddingHorizontal={'$4'} paddingTop={'$2'} paddingBottom={bottom ?? 60}>
-          {editMode ? (
-            <SafeButton onPress={onPressEdit}>Save</SafeButton>
-          ) : (
-            <SafeButton danger={true} onPress={onPressDelete}>
-              Remove signer
-            </SafeButton>
```
usame-algan, line 100: I would suggest to offer a delete signer option either in the overview of signers or still here. Otherwise users have to open the signer, view the private key and can only then delete the signer. Lets do it in another PR though as discussed.

apps/mobile/src/hooks/useDelegateCleanup.ts
```diff
@@ -0,0 +1,114 @@
...
+  // Get all delegates from Redux store
+  const allDelegates = useAppSelector(selectDelegates)
+
+  // Get notification cleanup functionality
+  const { cleanupNotificationsForDelegate } = useNotificationCleanup()
+
+  // Access API endpoints
+  const [deleteDelegate] = cgwApi.useDelegatesDeleteDelegateV2Mutation()
+
+  // Create the service instance with dependencies
```
usame-algan: nit: These comments are not necessary

apps/mobile/src/hooks/useDelegateCleanup.ts
```diff
@@ -0,0 +1,114 @@
+import { useCallback, useState, useMemo } from 'react'
```
usame-algan, line 1: Should the `useDelegateCleanup` hook not go into the `useDelegateCleanup` folder inside hooks?

apps/mobile/src/hooks/useDelegateCleanup/utils.ts
```diff
@@ -0,0 +1,311 @@
...
+    const failedCleanups = notificationCleanupResults
+      .filter((result) => result.status === 'rejected')
+      .map((result) => result.reason)
+
+    if (failedCleanups.length > 0) {
+      const failedDelegates = delegateAddresses.filter(
+        (_, index) => notificationCleanupResults[index].status === 'rejected',
+      )
+
+      const errorMsg = `Cannot delete private key: ${failedCleanups.join(', ')}. Please check your internet connection and try again.`
```
usame-algan, line 80: Seems a bit strange to me that we don't allow users to delete their signer unless notifications/delegates are also deleted. I understand the reason why we do that but IMO notifications should not be a reason to block something as critical as removing your private key from an app.

apps/mobile/src/hooks/useDelegateCleanup/utils.ts
```diff
@@ -0,0 +1,311 @@
...
+  deleteDelegate: (params: DelegatesDeleteDelegateV2ApiArg) => Promise<unknown>,
+): Promise<DelegateRemovalResult> => {
+  if (!delegateAddresses || delegateAddresses.length === 0) {
+    return { success: true }
+  }
+
+  const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
+
+  // Retry function with exponential backoff
+  const retryWithBackoff = async <T>(fn: () => Promise<T>, maxRetries = 3, baseDelay = 1000): Promise<T> => {
```
usame-algan: Can these be moved outside of the function maybe into their own util in case we want to reuse them?

apps/mobile/src/hooks/useDelegateCleanup/utils.ts
```diff
@@ -0,0 +1,311 @@
...
+ * Removes delegates from the backend transaction service across all chains
+ * Uses rate limiting and retry logic to handle 429 errors and improve reliability
+ */
+export const removeDelegatesFromBackend = async (
+  ownerAddress: Address,
+  delegateAddresses: Address[],
+  ownerWallet: Wallet,
+  allChains: Chain[],
+  deleteDelegate: (params: DelegatesDeleteDelegateV2ApiArg) => Promise<unknown>,
+): Promise<DelegateRemovalResult> => {
```
usame-algan, line 107: This function is pretty heavy. I understand that we need to add delays to not overwhelm our backend. I suggest we at least document the numbers on how long this is expected to block the UI e.g. when deleting one safe with one imported signer on x networks.

apps/mobile/src/services/key-storage/key-storage.service.ts
```diff
@@ -51,6 +51,19 @@ export class KeyStorageService implements IKeyStorageService {
     }
   }
 
+  async removePrivateKey(
+    userId: string,
+    options: PrivateKeyStorageOptions = { requireAuthentication: true },
+  ): Promise<void> {
+    try {
+      const { requireAuthentication = true } = options
```
usame-algan, line 59: Does it need to be set to true in both places as a fallback?

apps/mobile/src/services/key-storage/key-storage.service.ts
```diff
@@ -127,4 +140,35 @@ export class KeyStorageService implements IKeyStorageService {
...
+        await Keychain.resetGenericPassword({ service })
+      }
+    } catch (error) {
+      // If key doesn't exist, that's fine - we still want to try to remove from device crypto
+      Logger.warn('Key not found in keychain or authentication failed:', error)
+    }
+
+    // Try to remove the encryption key from device crypto
+    try {
+      await DeviceCrypto.deleteKey(keyName)
```
usame-algan, line 168: Where is the difference between this and `Keychain.resetGenericPassword`? Is it both trying to delete the same key?

apps/mobile/src/utils/notifications/cleanup.ts
```diff
@@ -0,0 +1,166 @@
...
+    type: 'blocking',
+    message: 'Network error: Cannot verify subscription removal',
+    originalError: error,
+  }
+}
+
+/**
+ * Utility for retrying operations with exponential backoff
+ */
+export const withRetry = async <T>(operation: () => Promise<T>, maxRetries = 3, baseDelay = 1000): Promise<T> => {
```
usame-algan: This seems very similar to `retryWithBackoff`. Do we need both?

apps/mobile/src/utils/notifications/cleanup.ts
```diff
@@ -0,0 +1,166 @@
...
+ * on Android and cookies are always sent.
+ *
+ * We do inline import to avoid circular dependency.
+ *
+ * @see https://github.com/facebook/react-native/issues/12956
+ */
+export const clearAuthBeforeUnauthenticatedCall = async (): Promise<void> => {
+  try {
+    const { cgwApi: authApi } = await import('@safe-global/store/gateway/AUTO_GENERATED/auth')
+    const { getStore } = await import('@/src/store/utils/singletonStore')
```
usame-algan, line 130: Is there a reason we are importing this inside the function and not outside?

apps/mobile/src/services/notifications/backend.ts
```diff
@@ -79,34 +86,43 @@ export const registerForNotificationsOnBackEnd = async ({
...
+              address: safeAddress,
+              notificationTypes: NOTIFICATIONS_GRANTED,
+            })),
+            deviceType: isAndroid ? 'ANDROID' : 'IOS',
+            deviceUuid,
+          },
         },
-      }),
+        {
+          forceOmitCredentials: noAuth,
```
usame-algan, line 123: Not sure I understand this. If the user is not authenticated we call `authenticateSigner` further up but then also omit credentials. Why is this needed?

usame-algan (approved):

apps/mobile/src/components/SafeListItem/SafeListItem.tsx
compojoom: In my tests it also worked on the "view private key", but if you navigate back to the settings and try to press something there it didn't work 🤷

apps/mobile/src/features/PrivateKey/PrivateKey.container.tsx
compojoom: I don't fully undrestand? What is storage for you?

In any case if you navigate back from the screen react is going to destroy the component and if there is any reference to the PK it should be garbage collected.

apps/mobile/src/features/PrivateKey/components/PrivateKeyView.tsx
compojoom: yes, you are right. will fix it.

apps/mobile/src/features/Settings/components/Navbar/SettingsMenu.tsx
compojoom: 👍

apps/mobile/src/hooks/useNotificationCleanup.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+export const useNotificationCleanup = (): UseNotificationCleanupProps => {
+  const [isLoading, setIsLoading] = useState(false)
+  const [error, setError] = useState<string | null>(null)
+
+  const allChains = useAppSelector(selectAllChains)
+  const allSafes = useAppSelector(selectAllSafes)
+  const safeSubscriptions = useAppSelector((state) => state.safeSubscriptions)
+  const fullState = useAppSelector((state) => state)
```
usame-algan: I am getting a warning to not use the full state as it can cause rerenders. Can we only select the delegates from the state? It seems that `selectAllDelegatesForSafeOwners` has two selector functions which is why we are passing the whole state there. Could we achieve the same with a redux thunk?

apps/mobile/src/services/key-storage/key-storage.service.ts
compojoom: No, Device crypto give you access to the secure enclave. The secure enclave is a dedicated hardware chip that only does encryption/decryption. Before storing the ethereum private key in the keychain we go to the hardware enclave and say -> create a key, then it complies and we use this key to encrypt the "ethereum private key". then we store the encrypted private key in the keychain. To read it, we fetch the encrypted private key from teh keychain and then decrypt it inside of the hardware enclave. We don't have direct access to the hardware enclave key -> we only have an alias that we can use to say encrypt/decrypt using this key. When we delete we delete both the PK in the keychain and the key used to encrypt the PK from the secure enclave.

apps/mobile/src/utils/notifications/cleanup.ts
compojoom: good point. I've consolidated those now to a single function.

apps/mobile/src/utils/notifications/cleanup.ts
compojoom: yes, it's in the function description - circular dependency errors

> /**
>  * Clears authentication cookies before making non-authenticated API calls.
>  * This is a workaround for the React Native bug where credentials: 'omit' is ignored
>  * on Android and cookies are always sent.
>  *
>  * We do inline import to avoid circular dependency.
>  *
>  * @see https://github.com/facebook/react-native/issues/12956
>  */

apps/mobile/src/services/notifications/backend.ts
compojoom: Because I initially went for omit-ing the credentials, but this didn't work. I kept it, but also now log out the user before trying to access this endpoint.

apps/mobile/src/hooks/useNotificationCleanup.ts
compojoom: Oh, I didn't notice this. Will try to fix it

compojoom: should be fixed now. It was becauise of the state => state, which was obviously now that I look at it - bad.

PR #6139, opened 2025-07-17
feat(mobile): navigate to correct screen on pressing on a notifications

compojoom:
## What it solves
Adds navigation on notification press

Resolves https://linear.app/safe-global/issue/COR-343/pressing-a-notification-should-open-the-correct-screen

## How this PR fixes it
Ads background, foregroudn listeners and then reacts on teh push notification. We switch the user's active safe and then navigate to the notification. It was a bit tricky and maybe we will have to rethink our design. Right now we rely on having only 1 active safe at a time. Obviously if a user is in safe X, but gets a notification for safe Y -> we want to display the info for safe Y. It can get confusing. You are in the process of doinig something in safeX, get a push notification press it and we switch you to a different safe and then you see the new notification. A better approach would be to just push the screen, but since as I said we rely on a single active safe this strategy doesn't work. Maybe in the future we could change it so that every screen could receive a safeAddress. This way we could just push screens on top of the stack.

Note: since we currently don't have detailed screens for transactions in the history, we just navigate the user to the history tab. 

## How to test it
1. Killed app
- Kill the app
- send a push notification to the device
- press it and the app should send you to the correct screen

2. App was minimized
- send a push notification ato the device
- press it -> the app should open and the user should be navigated to the correct screen

3. App is in foreground
- send a push notification to the device
- press it -> the correct screen should open

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -33,6 +33,7 @@ import { config, actions } from '@/src/platform/security'
 import { useFreeRasp } from 'freerasp-react-native'
 import { SafeStatusBar } from '@/src/theme/SafeStatusBar'
 import { GuardProvider } from '@/src/context/GuardProvider'
+import { useNotificationHandler } from '@/src/hooks/useNotificationHandler'
 
 Logger.setLevel(__DEV__ ? LogLevel.TRACE : LogLevel.ERROR)
 // Initialize all notification handlers
@@ -51,6 +52,7 @@ const HooksInitializer = () => {
   useInitWeb3()
   useInitSafeCoreSDK()
   useAnalytics() // Tracks activeSafe changes, but only once analytics is enabled in GetStarted screen
+  useNotificationHandler()
   return null
 }
 
--- a/apps/mobile/src/hooks/useNotificationHandler.ts
+++ b/apps/mobile/src/hooks/useNotificationHandler.ts
@@ -1,0 +1,31 @@
+import { useEffect } from 'react'
+import { EventType } from '@notifee/react-native'
+import NotificationsService from '@/src/services/notifications/NotificationService'
+import Logger from '@/src/utils/logger'
+
+/**
+ * Hook to handle notification events when the app is in the foreground
+ */
+export const useNotificationHandler = () => {
+  useEffect(() => {
+    // Set up foreground event listener
+    const unsubscribe = NotificationsService.onForegroundEvent(async ({ type, detail }) => {
+      try {
+        if (type === EventType.PRESS) {
+          await NotificationsService.handleNotificationPress({ detail })
+        } else if (type === EventType.DELIVERED) {
+          await NotificationsService.incrementBadgeCount(1)
+        } else if (type === EventType.DISMISSED) {
+          Logger.info('User dismissed notification:', detail.notification?.id)
+        }
+      } catch (error) {
+        Logger.error('useNotificationHandler: Error handling foreground notification event', error)
+      }
+    })
+
+    // Cleanup
+    return () => {
+      unsubscribe()
+    }
+  }, [])
+}
--- a/apps/mobile/src/services/notifications/NotificationService.ts
+++ b/apps/mobile/src/services/notifications/NotificationService.ts
@@ -14,7 +14,7 @@ import { updatePromptAttempts, updateLastTimePromptAttempted } from '@/src/store
 import { toggleAppNotifications, toggleDeviceNotifications } from '@/src/store/notificationsSlice'
 import { HandleNotificationCallback, LAUNCH_ACTIVITY, PressActionId } from '@/src/store/constants'
 import { getMessaging } from '@react-native-firebase/messaging'
-import * as TaskManager from 'expo-task-manager'
+import { NotificationNavigationHandler } from './notificationNavigationHandler'
 
 import { ChannelId, notificationChannels, withTimeout } from '@/src/utils/notifications'
 import Logger from '@/src/utils/logger'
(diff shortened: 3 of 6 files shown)


apps/mobile/src/app/_layout.tsx
```diff
@@ -51,6 +52,7 @@ const HooksInitializer = () => {
   useInitWeb3()
   useInitSafeCoreSDK()
   useAnalytics() // Tracks activeSafe changes, but only once analytics is enabled in GetStarted screen
+  useNotificationHandler() // Handle notifications when app is opened from notification
```
usame-algan: nit: This comment could also go into the hook itself. However there is already a comment there that is slightly different "Hook to handle notification events when the app is in the foreground". I am not sure if the hook only handles notification if the app is already open (in the foreground?) or if the app is opened from a notification.

apps/mobile/src/hooks/useNotificationHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+import { EventType } from '@notifee/react-native'
+import NotificationsService from '@/src/services/notifications/NotificationService'
+import Logger from '@/src/utils/logger'
+
+/**
+ * Hook to handle notification events when the app is in the foreground
+ */
+export const useNotificationHandler = () => {
+  useEffect(() => {
+    // Set up foreground event listener
```
usame-algan, line 11: nit: This describes what the next line does which only adds noise IMO

apps/mobile/src/services/notifications/NotificationService.ts
```diff
@@ -204,19 +204,38 @@ class NotificationsService {
   }
 
   async incrementBadgeCount(incrementBy?: number) {
-    return await notifee.incrementBadgeCount(incrementBy)
+    const result = await notifee.incrementBadgeCount(incrementBy)
```
usame-algan: `notifee.incrementBadgeCount` doesn't return anything according to its type. Do you see any need where this function could return the new badge count?

apps/mobile/src/services/notifications/NotificationService.ts
```diff
@@ -226,12 +245,20 @@ class NotificationsService {
...
+    // Clear all badges when user interacts with any notification
+    await this.clearAllBadges()
+
     if (detail?.notification?.id) {
       await this.cancelTriggerNotification(detail.notification.id)
     }
 
     if (detail?.notification?.data) {
+      // Handle navigation first
+      await NotificationNavigationHandler.handleNotificationPress(
```
usame-algan, line 252: We are casting here but can it be the case that data is a number? We could also check for that and convert it to a string to satifsy the type.

apps/mobile/src/services/notifications/NotificationService.ts
```diff
@@ -363,6 +394,40 @@ class NotificationsService {
...
+  /**
+   * Registers Firebase messaging handlers for when app is opened from notification
+   */
+  private registerFirebaseNotificationOpenedHandler(): void {
+    // Handle notification opened app when app is in background
+    getMessaging().onNotificationOpenedApp(async (remoteMessage: FirebaseMessagingTypes.RemoteMessage) => {
+      Logger.info('Notification caused app to open from background state:', remoteMessage)
+
+      // Clear badge when app is opened from notification
+      await this.setBadgeCount(0)
```
usame-algan: nit: Could also use `clearAllBadges`

apps/mobile/src/services/notifications/NotificationService.ts
```diff
@@ -363,6 +394,40 @@ class NotificationsService {
...
+      .then(async (remoteMessage: FirebaseMessagingTypes.RemoteMessage | null) => {
+        if (remoteMessage) {
+          Logger.info('Notification caused app to open from quit state:', remoteMessage)
+          if (remoteMessage.data) {
+            // Add extra delay for app startup from killed state
+            setTimeout(async () => {
+              // Clear badge when app is opened from notification
+              await this.setBadgeCount(0)
+              await NotificationNavigationHandler.handleNotificationPress(remoteMessage.data)
+            }, 1000) // Wait 1 second for app to fully initialize
```
usame-algan, line 418: Is 1s enough for this? What happens when the app takes longer to start or the safe info is taking longer to load will the navigation still happen?

apps/mobile/src/services/notifications/notificationNavigationHandler.ts
```diff
@@ -0,0 +1,186 @@
...
+  value?: string
+  failed?: string
+}
+
+export const NotificationNavigationHandler = {
+  /**
+   * Handles navigation when a push notification is pressed
+   */
+  async handleNotificationPress(data: FirebaseMessagingTypes.RemoteMessage['data']): Promise<void> {
+    Logger.info('NotificationNavigationHandler: handleNotificationPress called with data:', data)
```
usame-algan, line 42: I think it would be nice to have all these messages somewhere as a const or enum.

apps/mobile/src/services/notifications/notificationNavigationHandler.ts
```diff
@@ -0,0 +1,186 @@
...
+      }
+    } catch (_error) {
+      // Router not ready yet
+      Logger.info(`Router not ready, attempt ${i + 1}/${maxAttempts}`)
+    }
+    await new Promise((resolve) => setTimeout(resolve, delayMs))
+  }
+  Logger.error('Router failed to become ready within timeout')
+  return false
+}
```
usame-algan, line 28: I would expect expo-router to offer something for this. I found other using `getNavigationContainerRef` to check if its ready but somehow this is not exported in our version of it.

apps/mobile/src/services/notifications/notificationNavigationHandler.ts
```diff
@@ -0,0 +1,186 @@
...
+        router.push(path as never)
+      }
+    } catch (error) {
+      Logger.error('NotificationNavigationHandler: Error during safe navigation', error)
+    }
+  },
+
+  async safeDismissAll(): Promise<void> {
+    const isRouterReady = await waitForRouter()
+    if (isRouterReady && router.canGoBack()) {
```
usame-algan, line 152: `router.canGoBack` is already checked in `waitForRouter`. Is it still necessary to check again here?

apps/mobile/src/services/notifications/notificationNavigationHandler.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+  },
+
+  /**
+   * Navigates to the transaction history tab
+   */
+  async navigateToTransactionHistory(): Promise<void> {
+    // Navigate to transactions tab
+    await this.safeNavigate('/transactions')
+  },
```
usame-algan: 💬 💬

apps/mobile/src/services/notifications/notificationNavigationHandler.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  /**
+   * Navigates to the confirm transaction screen
+   */
+  async navigateToConfirmTransaction(safeTxHash?: string): Promise<void> {
+    if (safeTxHash) {
+      // Navigate to specific transaction confirmation with the txId (safeTxHash is the same as txId)
```
usame-algan: safeTxHash is not the same as txId. It is part of the txId but the txId also contains the safe address

usame-algan (approved):
I tested all three flows and they work nicely! Well done 🚀

PR #6154, opened 2025-07-23
fix(mobile): replace `useColorScheme` with centralized `useTheme` hook

tmjssz:
## What it solves

[COR-363](https://linear.app/safe-global/issue/COR-363/mobile-loading-indicators-use-wrong-theme-when-os-and-app-themes)

Fixes a visual inconsistency where components (e.g. loading indicators) incorrectly apply the OS-level color scheme rather than the app's selected theme. This occurred due to direct usage of `useColorScheme` from `react-native` instead of the app’s centralized theming logic.

## How this PR fixes it

Replaces all direct calls to `useColorScheme()` with the internal `useTheme()` hook to ensure theme colors are derived from app settings.

### Changes in useTheme hook
- **Add `isDark` flag** for simpler theme checks (`isDark` vs `colorScheme === 'dark'`)
- **Rename `currentTheme` → `colorScheme`** for better semantic clarity  
- **Improves OS integration** with explicit fallback handling for undefined color schemes

## How to test it
1. On your test device, set **OS appearance** to **dark mode**
2. In the app, explicitly set **theme preference** to **light mode**
3. **Expected result**: all app components (especially loading indicators) render using **light theme** styles

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/components/SafeSearchBar/SafeSearchBar.tsx
+++ b/apps/mobile/src/components/SafeSearchBar/SafeSearchBar.tsx
@@ -1,6 +1,7 @@
 import React, { useState, useEffect, useCallback } from 'react'
 import { View, TextInput, StyleSheet, TouchableOpacity, Platform, KeyboardAvoidingView } from 'react-native'
 import { SafeFontIcon } from '../SafeFontIcon'
+import { useTheme } from 'tamagui'
 
 interface SafeSearchBarProps {
   placeholder: string
@@ -11,6 +12,7 @@ interface SafeSearchBarProps {
 const SafeSearchBar: React.FC<SafeSearchBarProps> = ({ placeholder, onSearch, throttleTime = 300 }) => {
   const [searchQuery, setSearchQuery] = useState('')
   const [timer, setTimer] = useState<ReturnType<typeof setTimeout> | null>(null)
+  const theme = useTheme()
 
   const throttleSearch = useCallback(
     (query: string) => {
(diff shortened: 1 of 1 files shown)


apps/mobile/src/theme/hooks/useTheme.tsx
```diff
@@ -35,9 +35,12 @@ export const useTheme = () => {
     }
   }, [themePreference])
```
compojoom: can you check whether we still need this block. I think that we tried to fix an issue, but this was a hack and I couldn't reproduce it last time.

tmjssz: The issue is not reproducable for me, even whithout the workaround. So I removed it in 0c46bfda805a8345ab08cdee5dfc1463d517d78c

apps/mobile/src/theme/hooks/useTheme.tsx
```diff
@@ -7,7 +7,7 @@ import { ThemePreference } from '@/src/types/theme'
 
 export const useTheme = () => {
   const dispatch = useAppDispatch()
-  const colorScheme = useColorScheme()
+  const colorSchemeOS = useColorScheme() || undefined
```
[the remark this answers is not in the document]
tmjssz: The logical OR is intentional to ensure colorSchemeOS is never `null`. Added a comment for clarification.

compojoom (approved):

apps/mobile/src/components/SafeSearchBar/SafeSearchBar.tsx
```diff
@@ -53,6 +53,7 @@ const SafeSearchBar: React.FC<SafeSearchBarProps> = ({ placeholder, onSearch, th
         <TextInput
           style={styles.input}
           placeholder={placeholder}
+          placeholderTextColor="#99999980"
```
compojoom: shouldn't this be some color from the theme?

tmjssz: Yeah I initially avoided changing the styling too much, so I reused the same color as the input text and just added some opacity. But I agree that it should come from the theme. I've now updated the component to fully rely on theme colors. This results in slightly different tones for the search bar compared to before, but at least we're no longer using any hardcoded color codes. WDYT?

simulator_screenshot_0305DB1F-B066-440A-9A39-961023EB1D69 simulator_screenshot_EE1A49D5-AD4A-44C7-B442-6BA239D4E322 simulator_screenshot_F17D200C-FB35-4E0F-BF80-728D9150C80E

PR #6155, opened 2025-07-23
fix: stylus dep temp workaround

compojoom:
The builds are failing because the stylus dependency (that we don't directly use) was removed by npmjs. It seems that the report is bogus and it should get back to package repo. In the meantime I've implemented the workaround in the thread: https://github.com/stylus/stylus/issues/2938 

---

## CLA signature


tmjssz (approved):
Nice thanks!

PR #6157, opened 2025-07-23
chore: enable local yarn cache

compojoom:
The [recent stylus fiasco](https://github.com/stylus/stylus/issues/2938) just proves to me that we need to use yarn’s offline cache https://yarnpkg.com/features/caching. This ads some weight to the repo, but then yarn install will be way faster as nothing needs to be downloaded from the internet anymore.

And at one point once we manage to move to "zero install" with pnp we can forget that we need to do "yarn install" between branch changes. But our current setup of expo and nextjs don't allow that. Despite this, having the packages in the repo is a good thing!

---

## CLA signature


tmjssz:
<s>The docs state that [enableGlobalCache](https://yarnpkg.com/configuration/yarnrc#enableGlobalCache) needs to be set to false to enable the offline mirror, but we still have it true. Shouldn't we update it?</s>

---
Ahh never mind, just noticed that you did update the flag: https://github.com/safe-global/safe-wallet-monorepo/blob/chore/yarn-cache/.yarnrc.yml#L3
Strangely though the change doesn't appear for me in the PR view. Not sure why it’s not showing up.

tmjssz (approved):

PR #6160, opened 2025-07-24
fix(mobile): signer change was not reflected when signing txs

compojoom:
## What it solves
When I implemented the "receipt screens" I unintentially broke the signer selection. Despite the user selecting a signer a different signer on the "receipt screen" that was not reflected in the UI and also in the subsequent screen where the tx was actually submitted. 
The problem stemmed from the fact that we were sending the signer as a route param, which made it stale and unaware of redux changes. We needed the current signer already on the "confirm tx" screen in order to know whether we can allow signing of the tx in the UI.
This showed some flaws in the old implementation. We had a hook, that was both reading state and changing the state. I've performed some refactoring. 
- now we have a hook that only read from redux
- I've moved the dispatch logic to a separate hook that is only needed on the confirmTX container page
- there is a new hook that now accept a txID and fetches the txData. Since RTK is caching the result, it's not expensive to call this hook often. 

Now instead of passing the txID and signer through the router we just pass the txID and every component fetches the necessary data from the state. 

Resolves https://linear.app/safe-global/issue/COR-379/mobile-i-cannot-switch-signers

## How to test it
You need a safe with multiple signers. Then go to a tx and try to sign it. Switch the signers and observe that the correct signer is signing. 

## Checklist

---

## CLA signature


apps/mobile/src/features/ConfirmTx/hooks/useTransactionData.ts
```diff
@@ -0,0 +1,16 @@
+import { useTransactionsGetTransactionByIdV1Query } from '@safe-global/store/gateway/AUTO_GENERATED/transactions'
+import { useDefinedActiveSafe } from '@/src/store/hooks/activeSafe'
+
+export const useTransactionData = (txId: string) => {
```
usame-algan, line 4: nit: this fetches txDetails afaik. There is also txData which could be confusing so it might be better to rename this to `useTransactionDetails`

apps/mobile/src/features/ConfirmTx/hooks/useTxSignerAutoSelection.ts
```diff
@@ -0,0 +1,22 @@
...
+import { useTxSignerState } from '@/src/features/ConfirmTx/hooks/useTxSignerState'
+import { useTxSignerActions } from '@/src/features/ConfirmTx/hooks/useTxSignerActions'
+
+export const useTxSignerAutoSelection = (detailedExecutionInfo?: MultisigExecutionDetails) => {
+  const { activeTxSigner, appSigners, proposedSigner, hasSigned } = useTxSignerState(detailedExecutionInfo)
+  const { setTxSigner } = useTxSignerActions()
+
+  useLayoutEffect(() => {
+    if (proposedSigner && activeTxSigner?.value !== proposedSigner.value && hasSigned) {
+      console.log('use layout effectproposedSigner', proposedSigner)
```
usame-algan, line 12: console.log

apps/mobile/src/features/ConfirmTx/hooks/useTxSignerAutoSelection.ts
```diff
@@ -0,0 +1,22 @@
...
+import { MultisigExecutionDetails } from '@safe-global/store/gateway/AUTO_GENERATED/transactions'
+import { useTxSignerState } from '@/src/features/ConfirmTx/hooks/useTxSignerState'
+import { useTxSignerActions } from '@/src/features/ConfirmTx/hooks/useTxSignerActions'
+
+export const useTxSignerAutoSelection = (detailedExecutionInfo?: MultisigExecutionDetails) => {
+  const { activeTxSigner, appSigners, proposedSigner, hasSigned } = useTxSignerState(detailedExecutionInfo)
+  const { setTxSigner } = useTxSignerActions()
+
+  useLayoutEffect(() => {
+    if (proposedSigner && activeTxSigner?.value !== proposedSigner.value && hasSigned) {
```
usame-algan, line 11: This condition is hard to understand. Is it possible to name it? What does proposedSigner mean? I would guess:

- activeTxSigner -> The currently selected signer
- hasSigned -> The currently selected signer has already signed
- proposedSigner -> ...?

usame-algan (approved):

apps/mobile/src/features/ConfirmTx/hooks/useTxSignerAutoSelection.ts
compojoom: 🤦

compojoom: we should add an eslint rule for that.

apps/mobile/src/features/ConfirmTx/hooks/useTransactionData.ts
compojoom: no, it calls this endpoint: https://safe-client.staging.5afe.dev/v1/chains/11155111/transactions/0xb0133306ef864ae3d294c9dc9720c03243eada7aa4e8506e7fded16ad0e807ae which contains everything about the tx including txDetails

apps/mobile/src/features/ConfirmTx/hooks/useTxSignerAutoSelection.ts
compojoom: Yes, it's a brainfuck. Let's leave it as is now and refactor if it causes issues. When you add a signer, this signer is not auto applied as a signer for the txs. We do some "calculations" to figure out who is supposed to sign the tx. If you don't have an active signer, but have an imported signer and this signer is not the proposer of the tx and he is allowed to sign this tx, then this will be the proposedSidner and we will dispatch the event to set him as the activeSigner. I said, that it is a brainfuck, right?

PR #6162, opened 2025-07-24
feat(mobile): multiple improvements around pendingTxs

compojoom:
## What it solves
When we implemented the first version of the pendingTx list RTK didn't support infinite queries. So we hacked a useInfinite hook that was not performing up to the expectations. It was loading more results, but at one point if the data chagnes it would reset the view. I changed the fetching logic to use RTK's infinite query hook and now the view is stable - no matter how many pages of data is loaded. It also updates the results properly. 

I also applied the same fix to the NFT container.

Also in this pr I added loading indicators and pull down to refresh option. 

Resolves https://linear.app/safe-global/issue/COR-365/mobile-empty-pending-txs-page-state-when-there-are-19-txs-second-page

## How to test it
Open a safe with a lot of pending txs, scroll down, open a tx that needs signing, sign it. Then go back to the pending list -> the transactions should still be visible, after around 10s when the next refresh in the background occurs the signing state of teh tx should also change.
The pending TX screen now has a pull down to refresh feature and also scrolling down and loading the pages should show a skeleton. 

Check out the NFT container on a safe with a lot of nfts -> scrolling should work as before. 

## Checklist

---

## CLA signature


apps/mobile/src/features/PendingTx/utils.tsx
```diff
@@ -28,6 +28,15 @@ export const groupTxs = (list: PendingTransactionItems[]) => {
 
 export const groupPendingTxs = (list: PendingTransactionItems[]) => {
   const transactions = groupTxs(list)
+
+  if (transactions.length === 0) {
+    return {
+      pointer: -1,
```
usame-algan, line 34: What is this pointer value for? It doesn't seem to be used.

packages/store/src/gateway/collectibles.ts
```diff
@@ -0,0 +1,39 @@
...
+    >({
+      // Define infinite query options
+      infiniteQueryOptions: {
+        initialPageParam: null, // Start with null cursor
+
+        // Function to get the next page param
+        getNextPageParam,
+      },
+
+      // Query function
```
usame-algan: nit: Do we really need a comment for every line of code?

packages/store/src/gateway/transactions.ts
```diff
@@ -62,8 +40,33 @@ export const txHistoryApi = api.injectEndpoints({
...
+    >({
+      // Define infinite query options
+      infiniteQueryOptions: {
+        initialPageParam: null, // Start with null cursor
+
+        // Function to get the next page param
+        getNextPageParam,
+      },
+
+      // Query function
```
usame-algan: nit: Same here with the comments 😵

packages/store/src/utils/infiniteQuery.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Generic function to extract the next page parameter from Safe Gateway API responses.
+ * All Safe Gateway APIs use cursor-based pagination with the same URL structure.
+ *
+ * @param lastPage - The last page response from the API
+ * @returns The cursor for the next page, or undefined if no more pages
+ */
+export const getNextPageParam = (lastPage: { next?: string | null }) => {
+  // If there's no lastPage or no next page URL, return undefined
```
usame-algan: nit: This comment describes the next line

usame-algan (commented):
Code looks good but I encountered some UX issue while testing the infinite scroll. Do you have an idea what could cause this?

https://github.com/user-attachments/assets/e1463ebb-8428-40bd-8c30-01806496a81a

PR #6164, opened 2025-07-24
fix(mobile): tx history not refetched on click on push notification

compojoom:
## What it solves
When navigating from a push notification to the history, the list was not refetched from the server and if the app was in the background it could have stale data.

Resolves https://linear.app/safe-global/issue/COR-368/mobile-fix-refresh-outgoingincoming-transaction-when-app-is-in

## How this PR fixes it
We force a refetch if the user has clicked a push-notification

## How to test it
App should be running, navigate to the history tab, then minimize it. Send a tx to the safe with some amount. Click on the push notification, the app should open on the history tab and should perform a refetch and you should see the incoming transafer in the list.

## Checklist

---

## CLA signature


usame-algan (approved):
Looks good. You might need to rebase.

PR #6166, opened 2025-07-25
Fix(Mobile): Add bridge warning for unsupported safes

usame-algan:
## What it solves

Resolves [COR-358](https://linear.app/safe-global/issue/COR-358/mobilelifi-bridging-the-error-message-for-the-old-mastercopies-is)

## How this PR fixes it

- Copies the `useSafeCreationData` hook from web to mobile without counterfactual safe support

## How to test it

1. Open a pending bridge transaction on an old safe contract
2. Observe a warning

## Screenshots
Screenshot 2025-07-25 at 10 00 13

## Checklist

---

## CLA signature


apps/mobile/src/hooks/useSafeCreationData.ts
usame-algan, line 1: We might be able to move some of this into the shared utils package.

compojoom: To me it looks that you've abstracted quite a few functions here that could really be moved to the utils package, no?

usame-algan: The issue is that some of these functions use files or hooks that are only accessible on web so its harder to extract them.

usame-algan: I extracted as much as I could to the utils package.

usame-algan:
Created a new task to fix this on web as well for bridging: https://linear.app/safe-global/issue/COR-397/bridge-multichain-warning-is-checking-all-chains-instead-of-only-the

apps/web/src/components/common/ConnectWallet/__tests__/AccountCenter.test.tsx
```diff
@@ -10,6 +10,7 @@ const mockWallet = {
   provider: null as unknown as EIP1193Provider,
 }
 
+// TODO: This test is flaky and randomly fails sometimes
```
compojoom, line 13: How flaky is it? If it is 1 time in 100 runs ok, but if it is flaky on every couple of runs I'll rather skip it?

apps/web/src/components/common/NetworkSelector/index.tsx
```diff
@@ -199,6 +201,7 @@ const UndeployedNetworks = ({
   )
 
   const noAvailableNetworks = useMemo(() => availableNetworks.every((config) => !config.available), [availableNetworks])
+  console.log('NO AVAILABLE', availableNetworks)
```
compojoom: you  forgot this.

apps/web/src/components/common/ConnectWallet/__tests__/AccountCenter.test.tsx
usame-algan: Not super flaky IMO but I didn't have time to debug why its flaky. Just wanted to add a TODO so we don't forget about it.

compojoom (approved):
👍

PR #6173, opened 2025-07-28
fix(mobile): android UI discrepancies

compojoom:
## What it solves
Android now has 3 navigation modes
- Gesture navigation
- 2-button navigation
- 3-button navigation

when 2 and 3 button navigation is selected by the user some views were behind the navigation which created poor user experience. With this PR this problem should be fixed.

grafik

Resolves https://linear.app/safe-global/issue/COR-384/mobile-bottom-marginpadding-is-non-existent-in-android

## How this PR fixes it
It boils down to using safeAreaInsets even on screens that look fine on iOS. The bottom sheet was a bit tricky as it required some negative margins to hide content that "flies" behind it.

## How to test it
Best would be to run through all screens app with both iOS and andoird next to each other to confirm that everything looks ok on both platforms. 

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/components/SafeBottomSheet/SafeBottomSheet.tsx
+++ b/apps/mobile/src/components/SafeBottomSheet/SafeBottomSheet.tsx
@@ -9,7 +9,7 @@ import BottomSheet, {
   BottomSheetFooter,
 } from '@gorhom/bottom-sheet'
 import DraggableFlatList, { DragEndParams, RenderItemParams, ScaleDecorator } from 'react-native-draggable-flatlist'
-import { StyleSheet } from 'react-native'
+import { Platform, StyleSheet } from 'react-native'
 import { useRouter } from 'expo-router'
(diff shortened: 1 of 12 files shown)


apps/mobile/src/components/SafeBottomSheet/SafeBottomSheet.tsx
```diff
@@ -148,7 +151,8 @@ export function SafeBottomSheet<T>({
         ) : (
           <BottomSheetScrollView
             style={{
-              marginBottom: (!sortable && FooterComponent ? footerHeight : 0) + 12,
+              marginBottom:
+                (!sortable && FooterComponent ? footerHeight : 0) + getTokenValue(Platform.OS === 'ios' ? '$4' : '$8'),
```
usame-algan, line 147: nit: I suggest to move this up into a variable because its hard to read inline

usame-algan (approved):

PR #6197, opened 2025-08-05
feat(web): export CSV with transaction history

LucieFaire:
## What it solves 

- Introduces a new feature: Export transaction history as CSV

Resolves [COR-10](https://linear.app/safe-global/issue/COR-10/add-export-csv-cta-and-a-new-modal-to-the-ui) [COR-500](https://linear.app/safe-global/issue/COR-500/disable-csv-export-cta-if-user-is-not-a-signerwallet-not-connected)

## How this PR fixes it
 - Updated the auto-generated API to include new endpoints for export 
 - Introduced a new button on `Transactions - History` page
 - Implemented a new modal to select a date range to export
 - Automatically opens a download file in a new tab
 - Disable `Export` button when user is not a wallet owner/ wallet is not connected

## Screenshots

https://github.com/user-attachments/assets/37451cac-4b25-44d1-a4d9-96a7c3c0fad4

## Checklist

---

## CLA signature


apps/web/src/components/transactions/CsvTxExportButton/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+            exportJobId ? (
+              <CircularProgress size={16} />
+            ) : (
+              <SvgIcon component={ExportIcon} inheritViewBox fontSize="small" />
+            )
+          }
+          disabled={!!exportJobId}
+        >
+          {exportJobId ? 'Exporting' : 'Export CSV'}
+        </Button>
```
usame-algan: If you want the button to be the same height as the button next to it you can set the height here explicitely to 38px. Thats how we handle it in other places as well.

Image

apps/web/src/components/transactions/CsvTxExportButton/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+  hasActiveFilter: boolean
+}
+
+const CsvTxExportButton = ({ hasActiveFilter }: CsvTxExportProps): ReactElement => {
+  const dispatch = useAppDispatch()
+  const isDarkMode = useDarkMode()
+
+  const [openExportModal, setOpenExportModal] = useState(false)
+  const [exportJobId, setExportJobId] = useState<string | null>(null)
+  const [exportTimeout, setExportTimeout] = useState<NodeJS.Timeout | null>(null)
```
usame-algan: This state is only used within a useEffect. I think it doesn't need to be a state but can be a ref since the rendering doesn't depend on it.

apps/web/src/components/transactions/CsvTxExportButton/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+          groupKey: 'export-csv-error',
+          title: 'Something went wrong',
+          message: 'Please try exporting the CSV again.',
+        }),
+      )
+    }
+
+    if (!exportStatus && !error) return
+
+    const url = (exportStatus?.returnValue as CsvExportReturnValue)?.downloadUrl
```
usame-algan, line 113: Why do we need to cast here and not use the auto-generated type?

apps/web/src/components/transactions/CsvTxExportModal/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+type CsvTxExportForm = {
+  [CsvTxExportField.RANGE]: DateRangeOption | ''
+  [CsvTxExportField.FROM]: Date | null
+  [CsvTxExportField.TO]: Date | null
+}
+
+type CsvTxExportModalProps = {
+  onClose: () => void
+  onExport?: (job: JobStatusDto) => void
```
usame-algan: It looks like this prop is always defined. Does it still need to be optional?

apps/web/src/components/transactions/CsvTxExportModal/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+  hasActiveFilter?: boolean
```
usame-algan: Same with this prop. Does it need to be optional?

apps/web/src/components/transactions/CsvTxExportModal/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+  const isExportDisabled = useMemo(() => {
+    return (
+      !selectedRange ||
+      (selectedRange === DateRangeOption.CUSTOM && (!from || !to || !!errors.from || !!errors.to)) ||
+      isOverYear
+    )
+  }, [selectedRange, from, to, errors.from, errors.to, isOverYear])
+
+  const onSubmit = handleSubmit(async ({ range, from, to }) => {
+    const { executionDateGte, executionDateLte } = getExportDates(range as DateRangeOption, from, to)
```
usame-algan: It should not happen to have an empty value for range here because the button is disabled but it would be better to check that its not empty instead of casting.

apps/web/src/pages/transactions/history.tsx
```diff
@@ -35,6 +36,7 @@ const History: NextPage = () => {
         <Button variant="outlined" onClick={toggleFilter} size="small" endIcon={<ExpandIcon />}>
           {filter?.type ?? 'Filter'}
         </Button>
+        <CsvTxExportButton hasActiveFilter={!!filter} />
```
usame-algan: I suggest to add a feature flag for this so we can toggle it via config service. e.g. right now when using prod cgw I can still go through the flow but get an error.

apps/web/src/components/transactions/CsvTxExportButton/index.tsx
LucieFaire: because auto-generated type is an `object`. in CGW its an interface that then can be implemented depending on the response type, but this specific type is not specified in swagger

LucieFaire: i might change that in the future to be type safe

apps/web/src/components/transactions/CsvTxExportButton/index.tsx
LucieFaire: [848e74b](https://github.com/safe-global/safe-wallet-monorepo/pull/6197/commits/848e74bdf4363499c4b6428ff7d1352520a39da0)

apps/web/src/components/transactions/CsvTxExportButton/index.tsx
LucieFaire: [848e74b](https://github.com/safe-global/safe-wallet-monorepo/pull/6197/commits/848e74bdf4363499c4b6428ff7d1352520a39da0)

apps/web/src/components/transactions/CsvTxExportModal/index.tsx
LucieFaire: [848e74b](https://github.com/safe-global/safe-wallet-monorepo/pull/6197/commits/848e74bdf4363499c4b6428ff7d1352520a39da0)

apps/web/src/components/transactions/CsvTxExportModal/index.tsx
LucieFaire: [848e74b](https://github.com/safe-global/safe-wallet-monorepo/pull/6197/commits/848e74bdf4363499c4b6428ff7d1352520a39da0)

apps/web/src/components/transactions/CsvTxExportModal/index.tsx
LucieFaire: [848e74b](https://github.com/safe-global/safe-wallet-monorepo/pull/6197/commits/848e74bdf4363499c4b6428ff7d1352520a39da0)

apps/web/src/pages/transactions/history.tsx
LucieFaire: [1593585](https://github.com/safe-global/safe-wallet-monorepo/pull/6197/commits/1593585e94cc1bdbd86337cd44d759e9f1f01f8d)

PR #6201, opened 2025-08-07
Fix/tenderly partial revert warnings

013Meta:
## What it solves

  Resolves issues with Tenderly transaction simulation warnings for partial reverts - scenarios where a Safe transaction succeeds overall but contains internal call trace errors (e.g., a multicall where some sub-calls revert). This was causing complaints from users and is technically inaccurate as transactions can still execute.

## How this PR fixes it

  - Added isPartialRevert status to properly identify when simulations succeed but have internal errors
  - Refactored getSimulationStatus() to clearly extract nested simulation data and detect partial revert scenarios
  - Fixed confusing double nesting in simulation response handling (simulation.simulation.simulation → cleaner access pattern)
  - Added  tests for the new partial revert detection logic
  - Included mock simulation capability for local testing of partial revert scenarios

  How to test it

  1. Execute a Safe transaction that contains multiple calls where some succeed and others revert, or, uncomment the mock data in packages/utils/src/components/tx/security/tenderly/utils.ts:52-90.
  2. Verify the simulation shows appropriate warnings for partial reverts

## Checklist

---

## CLA signature

### DIFF
--- a/packages/utils/src/components/tx/security/tenderly/utils.ts
+++ b/packages/utils/src/components/tx/security/tenderly/utils.ts
@@ -47,7 +47,7 @@ export const isTxSimulationEnabled = (chain?: Pick<Chain, 'features'>): boolean 
 export const getSimulation = async (
   tx: TenderlySimulatePayload,
   customTenderly: EnvState['tenderly'] | undefined,
-): Promise<TenderlySimulation> => {
+): Promise<TenderlySimulation> => {  
   // MOCK: Uncomment to test partial revert locally
   /*
   console.log('🧪 MOCK: Returning partial revert simulation')
@@ -90,6 +90,7 @@ export const getSimulation = async (
   } as TenderlySimulation
   */
   
+
   const requestObject: RequestInit = {
     method: 'POST',
     body: JSON.stringify(tx),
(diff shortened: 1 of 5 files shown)


packages/utils/src/components/tx/security/tenderly/utils.ts
```diff
@@ -167,26 +209,33 @@ export type SimulationStatus = {
...
+  const isSuccess = tenderlyResponse?.simulation.status || false
 
   // Safe can emit failure event even though Tenderly simulation succeeds
-  const isCallTraceError = isSuccess && getCallTraceErrors(simulation.simulation).length > 0
-  const isError = simulation._simulationRequestStatus === FETCH_STATUS.ERROR
+  const isCallTraceError = isSuccess && getCallTraceErrors(tenderlyResponse).length > 0
+  const isError = simulationResult._simulationRequestStatus === FETCH_STATUS.ERROR
+
+  // Partial revert: simulation succeeds overall but has internal errors
+  const isPartialRevert = isSuccess && isCallTraceError
```
usame-algan: No need to add another `isSuccess` check since `isCallTraceError` already checks for `isSuccess` so `isPartialRevert === isCallTraceError`

I would suggest either removing the new variable and using `isCallTraceError` or renaming it.

ru513: Switched it to using isCallTraceError for everything

ru513:
re-opening from other account

PR #6203, opened 2025-08-08
Fix(Tests):update web e2e tests

liliya-soroka:
## What it solves

Resolves #
* Happy path v2: add owner flow - update the flow accroding the changes after Signing UX implementation 
* Regression: swap and swap_2 - for now tests with mocked Quote are skipped. The tests should be rewritten, as it's not possible to mock the time on CowSwap. The expiration time can only be set to now + 30 minutes. As a result, the 'Swap' button shows a 'Quote expired' message and time-related errors, making it impossible to proceed to the next step. - new task is created
* Regression: cf safes - update the default QR code state to unchecked ( this is now default state for the QR - don't show prefix)
* Regression: multichain_sidebar - add back data-testid="group-address" for the verify group address test - the test was failing because the data_test_id was removed from the code
* Regression: nested safes tests - removed unused steps( after Signing UX) and add check for the SafeProxy call
* Regression:sidebar_3 - fix the mocked response  for the empty pin list 
* Regression: twap_2 - update verify message function to  handles text concatenation across child elements
* Regression: tx_history_2 - update review hashes after signing UX ( separate task to add check for number of hashes and format)

## Checklist

---

## CLA signature


apps/web/cypress/e2e/happypath_2/add_owner.cy.js
```diff
@@ -23,7 +23,7 @@ describe('Happy path Add Owners tests', () => {
     cy.contains(owner.safeAccountNonceStr, { timeout: 10000 })
   })
 
-  it(
+  it.only(
```
usame-algan, line 26: Is it intended to leave `.only`?

apps/web/cypress/e2e/happypath_2/add_owner.cy.js
```diff
@@ -44,16 +44,27 @@ describe('Happy path Add Owners tests', () => {
...
         navigation.clickOnDisconnectBtn()
 
+        /**owner.clickOnAddSignerBtn()
+    owner.typeOwnerAddressManage(1, constants.SEPOLIA_OWNER_2)
+    owner.clickOnNextBtnManage()
+    owner.verifyConfirmTransactionWindowDisplayed()
+    owner.clickOnBackBtn()
+    owner.typeOwnerAddressManage(1, staticSafes.SEP_STATIC_SAFE_3)
+    owner.clickOnNextBtnManage()
+    owner.verifyConfirmTransactionWindowDisplayed() */
```
usame-algan, line 54: If this is not needed anymore I suggest to remove it.

PR #6207, opened 2025-08-12
fix(tests)cypress e2e tests 

liliya-soroka:
## What it solves

Resolves #
* SafeApps: drain_app tests - the decoding was switched from tx service to the decoder service , used in the tests contract doesn't have a name on the decoder service . The test is updated to the general name "transfer" for the action when not name for the contract
* Nested safe review page: the contract name was not seen fully because of the small resolution of the screen that was used for the tests. It was in between mobile view and readable view for the web. I have set up the default screen resolution for the tests to mac 13" resolution
* Safe apps: Tx-builder_2 - I have commented the step that checkes the Cowswap fallbackhandler step , while we are searching for the solution. The decoder service doesn't index the txs where "to" is the safe address ( all methods from the mastercopy)
* Regression: Tx_notes - separate clear state
* Happy Path: Swap - the test was skipped as mock data from the cowswap was used. But it's not possible to mock as expired time =now +30 mins
* Regression: Spaces - added waits based on the elements on the page. The loading of the pages sometimes is slow , so the next step can not be done 
## Checklist

---

## CLA signature


apps/web/cypress/e2e/pages/create_tx.pages.js
```diff
@@ -256,12 +256,14 @@ export function verifyDeleteChoiceBtnStatus(option) {
 }
 
 export function typeNoteText(text) {
-  cy.get(noteTextField).find('input').clear().type(text)
+  cy.get(noteTextField).find('input').clear()
```
tmjssz: Nit: would be nice to save the element in a const so `find()` does not need to run twice:
```suggestion
  const input = cy.get(noteTextField).find('input')
  input.clear()
  input.type(text)
```

PR #6209, opened 2025-08-12
Fix(web): tenderly partial revert warnings

ru513:
## What it solves

Resolves https://linear.app/safe-global/issue/COR-46/internal-reverts-on-tenderly

Resolves issues with Tenderly transaction simulation warnings for partial reverts - scenarios where a Safe transaction succeeds overall but contains internal call trace errors (e.g., a multicall where some sub-calls revert). This was causing complaints from users and is technically inaccurate as transactions can still execute.

## How this PR fixes it

- isCallTrace edited to show when simulations succeed but also have internal errors.

- Refactored getSimulationStatus() to clearly extract nested simulation data and detect partial revert scenarios
- Fixed confusing double nesting in simulation response handling (simulation.simulation.simulation → cleaner access pattern)
-Added tests for the new partial revert detection logic

-Included mock simulation capability for local testing of partial revert scenarios
How to test it

## How to test it
Execute a Safe transaction that contains multiple calls where some succeed and others revert, or, uncomment the mock data in packages/utils/src/components/tx/security/tenderly/utils.ts:52-90.
Verify the simulation shows appropriate warnings for partial reverts

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/transactions/QueuedTxSimulation/index.tsx
+++ b/apps/web/src/components/transactions/QueuedTxSimulation/index.tsx
@@ -19,6 +19,26 @@ import { sameAddress } from '@safe-global/utils/utils/addresses'
 import { useMemo } from 'react'
 import { useCurrentChain } from '@/hooks/useChains'
 
+const getSimulationIconProps = (isCallTraceError: boolean, isSuccess: boolean) => {
+  if (isCallTraceError) {
+    return { color: 'warning' as const, component: WarningIcon }
+  }
+  if (isSuccess) {
+    return { color: 'success' as const, component: CheckIcon }
+  }
+  return { color: 'error' as const, component: CloseIcon }
+}
+
+const getSimulationStatusText = (isCallTraceError: boolean, isSuccess: boolean) => {
+  if (isCallTraceError) {
+    return 'Can execute (with warnings)'
+  }
+  if (isSuccess) {
+    return 'Simulation successful'
+  }
+  return 'Simulation failed'
+}
+
 const CompactSimulationButton = ({
   label,
   iconComponent,
@@ -109,16 +129,11 @@ const InlineTxSimulation = ({ transaction }: { transaction: TransactionDetails }
       <ExternalLink href={simulationLink}>
         <Stack direction="row" alignItems="center" gap={0.5}>
           <SvgIcon
-            color={status.isCallTraceError ? 'warning' : status.isSuccess ? 'success' : 'error'}
-            component={status.isCallTraceError ? WarningIcon : status.isSuccess ? CheckIcon : CloseIcon}
+            {...getSimulationIconProps(status.isCallTraceError, status.isSuccess)}
             inheritViewBox
             sx={{ height: '16px' }}
           />
-          {status.isCallTraceError
-            ? 'Can execute (with warnings)'
-            : status.isSuccess
-              ? 'Simulation successful'
-              : 'Simulation failed'}
+          {getSimulationStatusText(status.isCallTraceError, status.isSuccess)}
         </Stack>
       </ExternalLink>
     )
--- a/apps/web/src/components/transactions/QueuedTxSimulation/index.tsx
+++ b/apps/web/src/components/transactions/QueuedTxSimulation/index.tsx
@@ -114,7 +114,11 @@ const InlineTxSimulation = ({ transaction }: { transaction: TransactionDetails }
             inheritViewBox
             sx={{ height: '16px' }}
           />
-          {status.isCallTraceError ? 'Can execute (with warnings)' : status.isSuccess ? 'Simulation successful' : 'Simulation failed'}
+          {status.isCallTraceError
+            ? 'Can execute (with warnings)'
+            : status.isSuccess
+              ? 'Simulation successful'
+              : 'Simulation failed'}
         </Stack>
       </ExternalLink>
     )
--- a/apps/web/src/components/transactions/QueuedTxSimulation/index.tsx
+++ b/apps/web/src/components/transactions/QueuedTxSimulation/index.tsx
@@ -11,6 +11,7 @@ import { useSigner } from '@/hooks/wallets/useWallet'
 import ExternalLink from '@/components/common/ExternalLink'
 import CheckIcon from '@/public/images/common/check.svg'
 import CloseIcon from '@/public/images/common/close.svg'
+import WarningIcon from '@/public/images/notifications/warning.svg'
 import { getSimulationStatus, isTxSimulationEnabled } from '@safe-global/utils/components/tx/security/tenderly/utils'
 import { useSafeSDK } from '@/hooks/coreSDK/safeCoreSDK'
 import { useIsNestedSafeOwner } from '@/hooks/useIsNestedSafeOwner'
@@ -108,12 +109,12 @@ const InlineTxSimulation = ({ transaction }: { transaction: TransactionDetails }
       <ExternalLink href={simulationLink}>
         <Stack direction="row" alignItems="center" gap={0.5}>
           <SvgIcon
-            color={status.isSuccess ? 'success' : 'error'}
-            component={status.isSuccess ? CheckIcon : CloseIcon}
+            color={status.isCallTraceError ? 'warning' : status.isSuccess ? 'success' : 'error'}
+            component={status.isCallTraceError ? WarningIcon : status.isSuccess ? CheckIcon : CloseIcon}
             inheritViewBox
             sx={{ height: '16px' }}
           />
-          {status.isSuccess ? 'Simulation successful' : 'Simulation failed'}
+          {status.isCallTraceError ? 'Can execute (with warnings)' : status.isSuccess ? 'Simulation successful' : 'Simulation failed'}
         </Stack>
       </ExternalLink>
     )
--- a/apps/web/src/components/tx-flow/TxInfoProvider.tsx
+++ b/apps/web/src/components/tx-flow/TxInfoProvider.tsx
@@ -15,7 +15,7 @@ type SimulationStatus = {
 
 const initialSimulation: UseSimulationReturn = {
   simulateTransaction: () => {},
-  simulation: undefined,
+  simulationData: undefined,
   _simulationRequestStatus: FETCH_STATUS.NOT_ASKED,
   simulationLink: '',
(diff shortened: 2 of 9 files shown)


yarn.lock
```diff
@@ -16675,24 +16675,10 @@ __metadata:
...
-  linkType: hard
-
-"caniuse-lite@npm:^1.0.30001718":
-  version: 1.0.30001720
-  resolution: "caniuse-lite@npm:1.0.30001720"
-  checksum: 10/6557c5052fa17fd531f3e1a8013b5924fb69afcd53d9f3e3b9adc9e31c5a7e436b674c000c53659e097fe1fda1c290d1bd17c7f3f98d13749644386ed722ab5f
+"caniuse-lite@npm:^1.0.30001579, caniuse-lite@npm:^1.0.30001669, caniuse-lite@npm:^1.0.30001688, caniuse-lite@npm:^1.0.30001718":
+  version: 1.0.30001731
+  resolution: "caniuse-lite@npm:1.0.30001731"
+  checksum: 10/ad6771127c0cca13a711ca363bb0165a687f9a5c76f057b2c8c9746608a6bae2f15c7b6955063eefcd4ca0e5733d429f25292cc093a40a189d662f3a8647a020
```
compojoom: Can you remove the yarn.lock update. I know it is coming from the caniuse call, but it has nothing to do with the PR and I don't want it here.

packages/utils/src/components/tx/security/tenderly/utils.ts
```diff
@@ -47,13 +47,56 @@ export const isTxSimulationEnabled = (chain?: Pick<Chain, 'features'>): boolean
...
+      ] as any,
+      error_message: null,
+      error_info: null
+    } as any,
+    contracts: [],
+    generated_access_list: []
+  } as TenderlySimulation
+  */
+  
+
```
compojoom: Please remove this comment form here. Ideally we would need to craft a dummy tx that QA could use to test with.

packages/utils/src/components/tx/security/tenderly/utils.ts
```diff
@@ -47,13 +47,56 @@ export const isTxSimulationEnabled = (chain?: Pick<Chain, 'features'>): boolean
...
   const requestObject: RequestInit = {
     method: 'POST',
     body: JSON.stringify(tx),
   }
 
-  if (customTenderly?.accessToken) {
+  if (customTenderly && customTenderly.accessToken) {
```
compojoom: why? isn't `customTenderly?` doing the same?

apps/web/src/components/tx/security/tenderly/index.tsx
```diff
@@ -205,13 +223,13 @@ export const TxSimulationMessage = ({ isNested = false }: { isNested?: boolean }
...
   const { isFinished, isError, isSuccess, isCallTraceError } = isNested ? txInfo.nestedTx.status : txInfo.status
-  const { simulationLink, simulation, requestError } = isNested ? txInfo.nestedTx.simulation : txInfo.simulation
+  const { simulationLink, simulationData, requestError } = isNested ? txInfo.nestedTx.simulation : txInfo.simulation
 
   if (!isFinished) {
     return null
   }
 
-  if (!isSuccess || isError || isCallTraceError) {
+  if (!isSuccess || isError) {
```
compojoom, line 216: I'm not sure about the return types here. But can it be that isError is true and isCallTraceError is also true? Is IsCallTraceError not a sub-type of error?

compojoom: If this is the case you might never reach your isCallTraceError later.

packages/utils/src/components/tx/security/tenderly/utils.ts
```diff
@@ -169,18 +212,21 @@ export type SimulationStatus = {
...
+  // Extract the nested status more clearly
+  const tenderlyResponse = simulationResult.simulationData
+  const isSuccess = tenderlyResponse?.simulation.status || false
 
   // Safe can emit failure event even though Tenderly simulation succeeds
-  const isCallTraceError = isSuccess && getCallTraceErrors(simulation.simulation).length > 0
-  const isError = simulation._simulationRequestStatus === FETCH_STATUS.ERROR
+
+  const isCallTraceError = isSuccess && getCallTraceErrors(tenderlyResponse).length > 0
+  const isError = simulationResult._simulationRequestStatus === FETCH_STATUS.ERROR
```
compojoom: Are those renamings really necessary here? Semantically nothing changed in the function, or am I missing something?

compojoom:
Also the getSimulationStatus function is used in the mobile app as well. Please make sure that we are displaying those txs there correctly as well.

apps/web/src/components/tx/security/tenderly/index.tsx
ru513: I had this labelled as a different variable before, isPartialRevert as clearer for this. Usame pointed that we can use isCallTrace instead as the logic was already there. 

This is just rearranging it to be visible/usable. isCallTraceError needs isSuccess, isError is api failure on tenderly side.

 const isCallTraceError = isSuccess && getCallTraceErrors(...).length > 0  // Only if isSuccess
 const isError = simulationResult._simulationRequestStatus === FETCH_STATUS.ERROR  // Request failed

yarn.lock
ru513: Removed, edited in new commit

packages/utils/src/components/tx/security/tenderly/utils.ts
ru513: Removed, edited in new commit

packages/utils/src/components/tx/security/tenderly/utils.ts
ru513: Removed, edited in new commit

packages/utils/src/components/tx/security/tenderly/utils.ts
ru513: Not necessary indeed, removed, edited in new commit.

apps/web/src/components/transactions/QueuedTxSimulation/index.tsx
```diff
@@ -108,12 +109,16 @@ const InlineTxSimulation = ({ transaction }: { transaction: TransactionDetails }
       <ExternalLink href={simulationLink}>
         <Stack direction="row" alignItems="center" gap={0.5}>
           <SvgIcon
-            color={status.isSuccess ? 'success' : 'error'}
-            component={status.isSuccess ? CheckIcon : CloseIcon}
+            color={status.isCallTraceError ? 'warning' : status.isSuccess ? 'success' : 'error'}
+            component={status.isCallTraceError ? WarningIcon : status.isSuccess ? CheckIcon : CloseIcon}
```
compojoom: Don't do nested ternaries - they are hard to read.

apps/web/src/components/transactions/QueuedTxSimulation/index.tsx
```diff
@@ -108,12 +109,16 @@ const InlineTxSimulation = ({ transaction }: { transaction: TransactionDetails }
...
             inheritViewBox
             sx={{ height: '16px' }}
           />
-          {status.isSuccess ? 'Simulation successful' : 'Simulation failed'}
+          {status.isCallTraceError
+            ? 'Can execute (with warnings)'
+            : status.isSuccess
+              ? 'Simulation successful'
+              : 'Simulation failed'}
```
compojoom: Please don't do nested ternaries like this. Instead move it to a function tha receives isCallTraceErrro and isSuccess and use normal returns. e.g.
```js
function ExecutionStatus(isCallTraceError, isSuccess) {
  if(isCallTraceerror) {
    return 'Can execute (with warnings)'
  }
  if (isSuccess) {
    return 'Simulation successful'
  }

   return 'Simulation failed'
}
```
by abstracting it like this you can even easily write a small unit test.

apps/web/src/components/tx-flow/TxInfoProvider.tsx
```diff
@@ -15,7 +15,7 @@ type SimulationStatus = {
 
 const initialSimulation: UseSimulationReturn = {
   simulateTransaction: () => {},
-  simulation: undefined,
+  simulationData: undefined,
```
compojoom, line 18: Please remove any renaming that is not necessary for this PR. If you want to rename code to improve readibility, put all the renaming in a single commit.

apps/web/src/components/tx/security/tenderly/__tests__/useSimulation.test.ts
```diff
@@ -26,9 +26,9 @@ describe('useSimulation()', () => {
 
   it('should have the correct initial values', () => {
     const { result } = renderHook(() => useSimulation())
-    const { simulation, simulationLink, requestError: simulationError, _simulationRequestStatus } = result.current
+    const { simulationData, simulationLink, requestError: simulationError, _simulationRequestStatus } = result.current
```
compojoom, line 29: Again here for the whole file - I wouldn't rename stuff if it isn't necessary. That makes the review harder.

apps/web/src/components/tx/security/tenderly/index.tsx
```diff
@@ -136,7 +137,7 @@ const TxSimulationBlock = ({
               }}
             />
           ) : isFinished ? (
-            !isSuccess || isError || isCallTraceError ? (
+            !isSuccess || isError ? (
```
compojoom: Now that we also have a separate case for isCallTraceError I would prefer to refactor the whole component rendering logic instead of doing more nested ternaries.

packages/utils/src/components/tx/security/tenderly/utils.ts
```diff
@@ -47,7 +47,10 @@ export const isTxSimulationEnabled = (chain?: Pick<Chain, 'features'>): boolean
 export const getSimulation = async (
   tx: TenderlySimulatePayload,
   customTenderly: EnvState['tenderly'] | undefined,
-): Promise<TenderlySimulation> => {
+): Promise<TenderlySimulation> => {  
+  
+  
+
```
compojoom: why the empty lines?

katspaugh:
@francovenica @liliya-soroka was this tested?

liliya-soroka:
not yet , fully

PR #6210, opened 2025-08-12
ci: bump actions/checkout to v5

rejected-l:
Node 24 compatibility: switch all jobs to actions/checkout@v5. Requires runner v2.327.1+. Pure maintenance, behavior unchanged.

Ref: https://github.com/actions/checkout/releases/tag/v5.0.0

### DIFF
--- a/.github/actions/yarn/action.yml
+++ b/.github/actions/yarn/action.yml
@@ -6,7 +6,7 @@ runs:
   using: 'composite'
   steps:
     # We require yarn v4 and installing through corepack is the easiest way to get it
-    - uses: actions/checkout@v4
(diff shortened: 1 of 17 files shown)


usame-algan:
Thanks for the contribution @rejected-l! Is there an issue with v4? I see v5 was only released yesterday so generally I would wait for it to be more mature before updating.

rejected-l:
If there are no urgent issues with v4 and the runner is not newer than v2.327.1, you can wait before upgrading.  
v5 is currently for those using the new features of Node.js 24 and fresh runners

compojoom:
I merged a dependabot PR on this.

PR #6211, opened 2025-08-12
Feat(fee) use feature flag for native swap fee v2

vseehausen:
## What it solves

Resolves https://linear.app/safe-global/issue/GRO-24/rework-swap-fee-to-work-with-the-feature-flags

## How this PR fixes it

Previously, the (experimental) new native swap fee was used on Base Chain only. Now it can be turned on and off on a per-chain basis with the new feature flag `NATIVE_COW_SWAP_FEE_V2`.

## How to test it

Test the native swap feature, and test regular swaps and stablecoin swaps with an amount less then 100,000k USD. They should be 0.7% instead of 0.35% (regular) and 0.2% instead of 0.1% (stablecoin) on Base and Sepolia. Above this threshold, fees should be the same for every chain.

## Checklist

- [X] I've written a unit/e2e test for it (if applicable) 🧑‍💻

---

## CLA signature


usame-algan:
There are lint issues preventing the deployment.

PR #6212, opened 2025-08-14
fix(tests): update Etherscan API URLs in tests and mocks

PooyaRaki:
## What it solves
This pull request updates the Etherscan URL used in test files to ensure they reference the correct and current endpoint.

---

## CLA signature


usame-algan:
Do we actually fetch these urls during test runs or is this for semantic reasons only?

PR #6213, opened 2025-08-14
chore(scripts): add option to run web scripts in parallel

vseehausen:
## What it solves

Running test and lint scripts was a significant bottleneck in development, especially when using AI-assisted programming. This PR reduces execution time for these scripts, enabling faster iteration and shorter development cycles.

## How this PR fixes it

This PR focuses on optimizations in the `web` directory:  

- **Concurrent checks:** Runs type checking and linting in parallel as part of `yarn test`.  
- **Targeted execution:** Adds commands to run tests and linting only on changed or staged files, reducing unnecessary work.  
- **Jest cache management:** Implements caching strategies to speed up repeated test runs by reusing results where possible.  

## How to test it

All files: Run `yarn test` and `yarn lint` on dev, then checkout the result of this branch with `yarn test` and `yarn lint:parallel`.

Especially the commands to test/lint files that are changed or staged yields a massive improvement.

## Screenshots

A test with 5 runs of each command yielded these results for the commands on all files:

```
Lint Performance

  | Version                     | First Run | Subsequent Runs | Improvement |
  |-----------------------------|-----------|-----------------|-------------|
  | Original (tsc && next lint) | 14.1s     | 4.7s avg        | Baseline    |
  | Optimized (lint:parallel)   | 12.3s     | 3.6s avg        | 23% faster  |

Test Performance

  | Version                           | Average Time | Improvement |
  |-----------------------------------|--------------|-------------|
  | Original (jest default)           | 18.6s        | Baseline    |
  | Final Optimized (jest with cache) | 17.2s        | 7.5% faster |
```

## Checklist

---

## CLA signature


apps/web/jest.config.cjs
```diff
@@ -8,6 +8,10 @@ const createJestConfig = nextJest({
 const customJestConfig = {
   setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
 
+  // Performance optimizations
+  cache: true,
+  cacheDirectory: '<rootDir>/.jest-cache',
+
```
usame-algan: jest cache is turned on by default. You can check this by running `jest --showConfig`

Image

apps/web/package.json
```diff
@@ -9,12 +9,17 @@
     "start": "next dev",
     "build": "next build",
     "lint": "tsc && next lint",
+    "lint:parallel": "sh -c 'yarn tsc & yarn next lint & wait'",
```
usame-algan, line 12: Our eslint config depends on typescript. Does this still work if we run both in parallel?

apps/web/package.json
```diff
@@ -9,12 +9,17 @@
...
     "lint:fix": "next lint --fix",
+    "lint:changed": "sh -c 'files=$(git diff --name-only --diff-filter=ACM | grep -E \"\\.(ts|tsx)$\" | grep \"^apps/web/\" | sed \"s|^apps/web/||\" | tr \"\\n\" \" \" || true); if [ -n \"$files\" ]; then echo \"TypeScript files changed: $files\"; yarn tsc --noEmit & next lint --file $files & wait; else echo \"No TypeScript files changed to lint\"; fi'",
+    "lint:staged": "sh -c 'files=$(git diff --cached --name-only --diff-filter=ACM | grep -E \"\\.(ts|tsx)$\" | grep \"^apps/web/\" | sed \"s|^apps/web/||\" | tr \"\\n\" \" \" || true); if [ -n \"$files\" ]; then echo \"Staged TypeScript files: $files\"; yarn tsc --noEmit & next lint --file $files & wait; else echo \"No staged TypeScript files to lint\"; fi'",
```
usame-algan: This is really hard to read. I suggest moving them to a script file inside the `scripts` directory e.g. like we do for the integrity check.

apps/web/package.json
vseehausen: Yep, on my machine it worked. Is there a way to test it?

apps/web/jest.config.cjs
vseehausen: Yep, but still, I had performance improvement and new test files. Did I oversee something?

vseehausen: I turned this off regardlessly.

apps/web/package.json
vseehausen: Did so!

apps/web/package.json
vseehausen: I tested it again and it works. Also, there is still the old `lint` command.

vseehausen:
@compojoom thanks for the feedback!

lint-staged runs git commands in the background. The scripts I added just filter for files. The ultimate purpose is to make the lint commands faster to reduce iteration time. But I am happy to remove the lint-staged script.

I also want to run lint and type checking in parallel. It's a small improvement from ~12s to ~10s but still.

vseehausen:
These scripts are especially helpful when working with an AI tool to make sure every step the AI takes is tested and linted.

compojoom:
@vseehausen I'm still a bit confused. "next lint" per default runs on only changed files as it has a cache. You have to explicitly disable this behavior with --no-cache. So why do we need to provide the files from git diff? 

We could use a package such as npm-run-all or concurly and we would end up with just a new script like
lint:parallel: "npm-run-all --parallel tsc lint" 

Like this the PR could end-up being 2 lines?

vseehausen:
Actually, you are right. And now the parallel one is even slower than the last one... Don't know why the benchmarks changed now.

vseehausen:
Closed, as not needed, and currently it doesn't offer any performance upgrade.

PR #6218, opened 2025-08-15
fix(web): update Dashboard loading state

LucieFaire:
## What it solves

Resolves # [COR-394](https://linear.app/safe-global/issue/COR-394/loading-state-new-dashboard)

## How this PR fixes it

- changed the loading state to be user-friendly 

## How to test it

- load the Wallet Dashboard with the throttle network setting so it loads slow and observe the loading state for Total Assets Value, Top Assets and Pending Transactions

## Screenshots
Screenshot 2025-08-15 at 13 28 25

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/dashboard/Assets/index.tsx
+++ b/apps/web/src/components/dashboard/Assets/index.tsx
@@ -130,7 +130,7 @@ export const isNonZeroBalance = (item: Balances['items'][number]) => item.balanc
 const AssetsWidget = () => {
   const router = useRouter()
   const { safe } = router.query
-  const { loading } = useBalances()
+  const { loading, balances } = useBalances()
   const visibleAssets = useVisibleAssets()
 
   const items = useMemo(() => {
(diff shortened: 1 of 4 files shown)


francovenica:
So I can see the changes, which is fine, but the the elements are also showing first the empty state, then the gray blob done in this PR and then it loads the values.
Is there any way to not show the empty state first? (by emtpy state I mean those messages saying "No assets yet" for example)

load dashboard

Note:
Also there is this thing were the block loads with no sidebar and then they acomodate themselve. It looks ugly, but that's is on prod as well, so I'd leave that out of the scope of this ticket

LucieFaire:
> eth:0xA77DE01e157f9f57C7c4A326eeE9C4874D0598b6

I have prepared a fix [here](https://github.com/safe-global/safe-wallet-monorepo/pull/6254), but have to check first whether its a good approach

LucieFaire:
> So I can see the changes, which is fine, but the the elements are also showing first the empty state, then the gray blob done in this PR and then it loads the values. Is there any way to not show the empty state first? (by emtpy state I mean those messages saying "No assets yet" for example)
> 
> load dashboard 
> Note: Also there is this thing were the block loads with no sidebar and then they acomodate themselve. It looks ugly, but that's is on prod as well, so I'd leave that out of the scope of this ticket


I have prepared a fix [here](https://github.com/safe-global/safe-wallet-monorepo/pull/6254), but have to check first whether its a good approach

PR #6229, opened 2025-08-19
Feat(Analytics) Consolidate Tracking Simple

vseehausen:
# Simple Analytics Consolidation

## What it solves

Replaces manual dual-provider tracking with automated event routing. Currently, components must call both GA and MixPanel separately for each event.

Problems addressed:
- Scattered tracking calls throughout codebase
- Risk of missing events for specific providers  
- Maintenance overhead when adding new events

## How this PR fixes it

- **Unified tracking**: Single `trackEvent()` call automatically routes to both providers
- **Declarative mapping**: `GA_TO_MIXPANEL_MAPPING` defines which events go to MixPanel
- **Type safety**: Uses actual event properties instead of hardcoded strings

### Before (Manual dual tracking)
```typescript
// Two separate calls for each event
trackEvent({ ...WALLET_EVENTS.CONNECT, label: wallet.label })
trackMixPanelEvent(MixPanelEvent.WALLET_CONNECTED, { 
  'EOA Wallet Label': wallet.label,
  'EOA Wallet Address': wallet.address,
})
```

### After (Unified tracking)
```typescript
// Single call, automatic routing
trackEvent(
  { ...WALLET_EVENTS.CONNECT, label: wallet.label },
  {
    [MixPanelEventParams.EOA_WALLET_LABEL]: wallet.label,
    [MixPanelEventParams.EOA_WALLET_ADDRESS]: wallet.address,
  }
)
```

## Implementation

```typescript
// Event mapping
export const GA_TO_MIXPANEL_MAPPING: Record<string, string> = {
  [CREATE_SAFE_EVENTS.CREATED_SAFE.action]: MixPanelEvent.SAFE_CREATED,
  [WALLET_EVENTS.CONNECT.action]: MixPanelEvent.WALLET_CONNECTED,
  [SAFE_APPS_EVENTS.OPEN_APP.action]: MixPanelEvent.SAFE_APP_LAUNCHED,
}

// Unified tracking function
export const trackEvent = (eventData: AnalyticsEvent, additionalParameters?: Record<string, any>): void => {
  gtmTrack(eventData)
  
  const mixpanelEventName = GA_TO_MIXPANEL_MAPPING[eventData.action]
  if (mixpanelEventName) {
    mixpanelTrack(mixpanelEventName, additionalParameters)
  }
}
```

## Events Tracked

- **Safe Created**: Deployment context, owner count, threshold, payment method
- **Safe Activated**: Counterfactual deployment tracking  
- **Wallet Connected**: Wallet type, network, connection method
- **Safe App Launched**: App name, launch location context

## Testing

Connect a wallet and check if the event is tracked on MixPanel and Google Analytics.

## Checklist

- [ ] I've documented how it affects the analytics (if at all) 📊 - **Yes, consolidates analytics tracking**
- [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻 - **Yes, enhanced test coverage**


apps/web/src/hooks/wallets/useOnboard.ts
```diff
@@ -89,8 +89,18 @@ export const getWalletConnectLabel = (wallet: ConnectedWallet): string | undefin
...
+  const chainInfo = configs.find((config) => config.chainId === wallet.chainId)
+  const networkName = chainInfo?.chainName || `Chain ${wallet.chainId}`
+
+  trackEvent(
+    { ...WALLET_EVENTS.CONNECT, label: wallet.label },
+    {
+      [MixPanelEventParams.EOA_WALLET_LABEL]: wallet.label,
+      [MixPanelEventParams.EOA_WALLET_ADDRESS]: wallet.address.toLowerCase(),
+      [MixPanelEventParams.EOA_WALLET_NETWORK]: networkName,
+    },
```
usame-algan, line 102: We already track these as global parameters that are used in GA so we don't have to pass them every time we emit an event. I would suggest doing the same for Mixpanel. Take a look at how `gtmSetUserProperty` is used.

apps/web/src/features/counterfactual/hooks/usePendingSafeStatuses.ts
```diff
@@ -116,13 +120,32 @@ const usePendingSafeStatus = (): void => {
...
 
+          const undeployedSafe = undeployedSafes[creationChainId]?.[detail.safeAddress]
+          if (undeployedSafe) {
+            const safeSetup = extractCounterfactualSafeSetup(undeployedSafe, creationChainId)
+            if (safeSetup) {
+              trackEvent(CREATE_SAFE_EVENTS.ACTIVATED_SAFE, {
+                [MixPanelEventParams.SAFE_ADDRESS]: detail.safeAddress.toLowerCase(),
+                [MixPanelEventParams.BLOCKCHAIN_NETWORK]: chain?.chainName || '',
+                [MixPanelEventParams.NUMBER_OF_OWNERS]: safeSetup.owners.length,
+                [MixPanelEventParams.THRESHOLD]: safeSetup.threshold,
```
usame-algan, line 136: Same as above these could be tracked as global parameters whenever users are on a Safe so we don't have to pass them with every event.

vseehausen: Are you sure? This is a newly activated Safe, so I am not sure if they are in memory or if we would create a race condition. But I see the chain is tracked incorrectly. It should get the chain from the new Safe.

apps/web/src/hooks/wallets/useOnboard.ts
vseehausen: Yes, this is correct. GA needs in-memory properties, and MixPanel handles them for us. 

But in this case, could we get a race condition, as the wallet just got connected? So we have to get the data directly.

apps/web/src/features/counterfactual/hooks/usePendingSafeStatuses.ts
vseehausen: At least on my machine the events get fired before the user parameter is set. So I would leave them here in the event.

grafik

apps/web/src/hooks/wallets/useOnboard.ts
vseehausen: In this case, the event is also triggered before the user properties are set (by listeners). So I would leave this here.

grafik

vseehausen:
I want to close this, as https://github.com/safe-global/safe-wallet-monorepo/pull/6234 now implements a clean commit history.

PR #6230, opened 2025-08-20
Fix(Spaces): Respect current chainId when selecting address book item

usame-algan:
## What it solves

Resolves [COR-493](https://linear.app/safe-global/issue/COR-493/spaces-address-book-names-are-only-visible-if-wallet-is-on-the-correct#comment-9f575331)

## How this PR fixes it

We created a new hook `useAllMergedAddressBooks` that is supposed to combine local and cloud address books. Since the local address books have a different structure we need to map them into the new structure. Since there can also be duplicate entries there with different names we need a strategy to decide on the name we want to use. Previous to this fix we always defaulted with the first name we saw and kept that but this causes issues if the same safe on a different network is open it displays the name from the other network.

This PR fixes that by allowing to pass the chainId into the mapping logic and overwriting the name in case the networks match.

## How to test it

1. Open a Safe that is deployed on at least two networks
2. Add local address book entries for that safe with different names
3. Add both safes to your space
4. Observe either of the two names is displayed within the Space (this is expected)
5. Open the safe on network A and observe the name from network A is used (in the sidebar and the breadcrumbs)
6. Do the same for safe on network B
7. Delete the address book entry on network B
8. Observe the address is shown instead of the name of the entry from network A

## Screenshots

Screenshot 2025-08-20 at 10 50 55
Screenshot 2025-08-20 at 10 51 05
Screenshot 2025-08-20 at 10 51 12

After deletion on one network
Screenshot 2025-08-20 at 10 51 21
Screenshot 2025-08-20 at 10 51 35

Another fix is that we don't show all network logos anymore if a contact is added on all networks but instead a chip:
Screenshot 2025-08-20 at 13 55 13

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/features/myAccounts/components/AccountItems/MultiAccountItem.tsx
+++ b/apps/web/src/features/myAccounts/components/AccountItems/MultiAccountItem.tsx
@@ -283,6 +283,7 @@ const MultiAccountItem = ({ onLinkClick, multiSafeAccountItem, isSpaceSafe = fal
               <EthHashInfo
                 address={address}
                 name={multiSafeAccountItem.name}
+                showName={isSpaceSafe ? !!multiSafeAccountItem.name : true}
                 shortAddress
                 showPrefix={false}
                 showAvatar={false}
--- a/apps/web/src/features/spaces/components/SpaceAddressBook/SpaceAddressBookTable.tsx
+++ b/apps/web/src/features/spaces/components/SpaceAddressBook/SpaceAddressBookTable.tsx
@@ -2,12 +2,13 @@ import EnhancedTable from '@/components/common/EnhancedTable'
 import EthHashInfo from '@/components/common/EthHashInfo'
 import tableCss from '@/components/common/EnhancedTable/styles.module.css'
 import Identicon from '@/components/common/Identicon'
-import { Box, Stack, Tooltip } from '@mui/material'
+import { Box, Chip, Stack, Tooltip } from '@mui/material'
 import NetworkLogosList from '@/features/multichain/components/NetworkLogosList'
 import ChainIndicator from '@/components/common/ChainIndicator'
 import type { SpaceAddressBookItemDto } from '@safe-global/store/gateway/AUTO_GENERATED/spaces'
 import SpaceAddressBookActions from '@/features/spaces/components/SpaceAddressBook/SpaceAddressBookActions'
 import { ContactSource } from '@/hooks/useAllAddressBooks'
+import useChains from '@/hooks/useChains'
 
 const headCells = [
   { id: 'contact', label: 'Contact', disableSort: true },
@@ -20,6 +21,8 @@ type SpaceAddressBookTableProps = {
 }
 
 function SpaceAddressBookTable({ entries }: SpaceAddressBookTableProps) {
+  const chains = useChains()
+
   const rows = entries.map((entry) => ({
     cells: {
       contact: {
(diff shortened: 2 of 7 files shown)


apps/web/src/hooks/__tests__/useAllAddressBooks.test.ts
```diff
@@ -103,6 +103,24 @@ describe('useAllAddressBooks', () => {
...
+        },
+        '11155111': {
+          '0xA': 'Alice (sepolia)',
+        },
+      }
+
+      const { result } = renderHook(() => useAllMergedAddressBooks(mockChainId))
+      expect(result.current).toHaveLength(2)
+      expect(result.current.map((c) => c.name)).toEqual(['Alice (sepolia)', 'Bob'])
+    })
```
tmjssz: How about adding a test for the case where the passed chainId is not present in the local address book (should fall back to first non-empty name)?

apps/web/src/features/spaces/components/SpaceAddressBook/SpaceAddressBookTable.tsx
```diff
@@ -62,7 +65,11 @@ function SpaceAddressBookTable({ entries }: SpaceAddressBookTableProps) {
               arrow
             >
               <Box sx={{ display: 'inline-block' }}>
-                <NetworkLogosList networks={entry.chainIds.map((chainId) => ({ chainId }))} />
+                {chains.configs.length === entry.chainIds.length ? (
```
compojoom, line 68: what if entry.chainIds is empty? Looking at the type it is a possibility. 
grafik

apps/web/src/hooks/useAllAddressBooks.ts
```diff
@@ -50,13 +50,17 @@ const mapAllLocalAddressBooks = (allAddressBooks: AddressBookState): ExtendedCon
...
-  return useMemo(() => mapAllLocalAddressBooks(allAddressBooks), [allAddressBooks])
+  return useMemo(() => mapAllLocalAddressBooks(allAddressBooks, chainId), [allAddressBooks, chainId])
 }
 
-export const useAllMergedAddressBooks = (): ExtendedContact[] => {
+/**
+ * Optional chainId in case an address exists on multiple networks locally to decide which name to use
+ * @param chainId
+ */
+export const useAllMergedAddressBooks = (chainId?: string): ExtendedContact[] => {
```
tmjssz, line 35: This hook is also used by `AddressBookInput`. Shouldn't we pass chainId there as well to make sure the correct name is used? Or is it not necessary there?

apps/web/src/features/spaces/components/SpaceAddressBook/SpaceAddressBookTable.tsx
usame-algan: Good question. It should not be possible to have an address book entry without any networks but I will double-check.

usame-algan: Ok, looks like this is not validated on the backend 😨 
Screenshot 2025-08-20 at 14 38 21

usame-algan: I will create a new task for the CGW but it is enforced on the frontend so no work needed here.

apps/web/src/hooks/useAllAddressBooks.ts
usame-algan: Good point! Just realised we are also displaying all entries e.g. when transfering tokens but it would make more sense to only display entries on that network + whatever comes from the space address book.

usame-algan:
@tmjssz I reverted https://github.com/safe-global/safe-wallet-monorepo/pull/6223/files because we only want to use entries of a specific network after all. The actual fix was passing the chainId to the hook in order to show a consistent name that doesn't depend on the connected wallet.

I also updated the SafeListItems so when used within a space they only show a name if it comes from the space address book.

apps/web/src/features/myAccounts/components/AccountItems/MultiAccountItem.tsx
```diff
@@ -283,6 +283,7 @@ const MultiAccountItem = ({ onLinkClick, multiSafeAccountItem, isSpaceSafe = fal
               <EthHashInfo
                 address={address}
                 name={multiSafeAccountItem.name}
+                showName={isSpaceSafe ? !!multiSafeAccountItem.name : true}
```
usame-algan, line 286: This is not really clean IMO but it allows us to have this exception within spaces where we only want to display a name if it comes from the space address book.

Issue #4839, opened 2025-01-28
Add proposers button enabled for CF safes


francovenica:
## Bug description
In a CF safe (safe that has not being activated yet), any button that leads to a tx is grayed out saying that you must first activate the safe. That's not the case for the "Add proposer" button that is still enabled

Note: A proposers cannot be added still since the confirm button in the modal is grayed out

## Environment

- Browser: Chrome
- Wallet: MetaMask
- Chain: Ethereum mainnet
- Env: dev 1.49

## Steps to reproduce

1.  Create a CF safe (pay later)
2. Go to the settings

## Expected result
Add proposers button should be grayed out

## Obtained result
The button is enabled

## Screenshots
![Image](https://github.com/user-attachments/assets/61f90c0d-f293-462e-a61f-b2cf70fc54da)

DevWebZone:
added a fix for this issue here: https://github.com/safe-global/safe-wallet-monorepo/pull/6231

PR #6231, opened 2025-08-21
fix(web): disable add proposer button if safe is not activated

DevWebZone:
## What it solves
It Solves Issue: https://github.com/safe-global/safe-wallet-monorepo/issues/4839

## How this PR fixes it
added SafeNotActivated validation in OnlyOwner component, which was only referenced by the Add proposer button. 
Note:  message added in tooltip for SafeNotActivated was referenced from CheckWallet Component which is used by all other features

## How to test it
1. Create a CF safe (pay later)
2. Go to the settings

## Screenshots
add proposer button disabled for inactivate safe
image

add proposer button enabled once safe is activated
image

## Checklist

- [ ] I've tested the branch on mobile 📱NA
- [ ] I've documented how it affects the analytics (if at all) 📊 NA
- [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻 NA

---

## CLA signature
Devanshu Wadhwani

### DIFF
--- a/apps/web/src/components/common/OnlyOwner/index.tsx
+++ b/apps/web/src/components/common/OnlyOwner/index.tsx
@@ -3,6 +3,7 @@ import useIsSafeOwner from '@/hooks/useIsSafeOwner'
 import useWallet from '@/hooks/wallets/useWallet'
 import useConnectWallet from '../ConnectWallet/useConnectWallet'
 import { Tooltip } from '@mui/material'
+import useSafeInfo from '@/hooks/useSafeInfo'
 
 type CheckWalletProps = {
   children: (ok: boolean) => ReactElement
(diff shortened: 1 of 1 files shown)


DevWebZone:
> I would not make this logic part of the `OnlyOwner` component but put it directly into the `ProposersList`

Sure @usame-algan , i kept it here as OnlyOwner was only used by the ProposersList. although, if i keep it in `ProposersList` the tooltip wont be showing up. shall i add the tooltip component separately in  `ProposersList` as well or can we ignore that and just make the button disabled?

DevWebZone:
@usame-algan moved the logic to `ProposersList` Component. please check now

usame-algan:
Thanks @DevWebZone looks good!

PR #6217, opened 2025-08-15
Web/fix/nested safes regression test

liliya-soroka:
## What it solves

Resolves #
* add scroll Tx details before search for the text there
## Checklist

- [ ] I've tested the branch on mobile 📱
- [ ] I've documented how it affects the analytics (if at all) 📊
- [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

---

## CLA signature

With the submission of this Pull Request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://safe.global/cla).

PR #6206, opened 2025-08-12
chore(web): remove "Buy crypto" button

LucieFaire:
## What it solves

Resolves # [COR-393](https://linear.app/safe-global/issue/COR-393/remove-buy-crypto)

## How this PR fixes it
- Removed `Buy crypto` button from dashboard and assets page
- Update notifications and descriptions to exclude `buy crypto` option

## How to test it
-  No `Buy Crypto` button on dashboard and assets page
- Changed the text for `Add funds` section to exclude bank account part

## Checklist

- [ ] I've tested the branch on mobile 📱
- [ ] I've documented how it affects the analytics (if at all) 📊
- [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

---

## CLA signature

With the submission of this Pull Request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://safe.global/cla).

PR #6233, opened 2025-08-21
Release 1.67.0

usame-algan:
## What it solves

Resolves [COR-501](https://linear.app/safe-global/issue/COR-501/release-1670)

## Features

* Feat(Spaces): Spaces address book (#5606)

## Fixes

* fix(web): new transactions next button disable bug
* chore(web): remove "Buy crypto" button (#6206)
* fix(web): EIP-7702 accounts are treated as smart accounts (#6235)

## Chores / Tests

* fix(tests): update Etherscan API URLs in tests and mocks (#6212)
* chore(deps): bump actions/checkout from 4 to 5 (#6222)
* fix(tests)Update web-e2e-prod-ondemand.yml- remove schedule (#6219)
* Web/fix/nested safes regression test (#6217)
* fix(e2e): click on share block header to make sure it expands (#6216)
* fix(tests)fix cypress e2e tests  (#6207)
* fix(e2e): remove trailing slash from URLs to fix failing tests (#6208)
* Fix(Tests)clean_test_add-owner (#6204)
* Fix(Tests):update web e2e tests (#6203)
* Fix(Tests): update regression and happy path tests (#6200)
* Fix(Web): Decrease bundle size (#6198)

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -2,7 +2,7 @@
   "name": "@safe-global/web",
   "homepage": "https://github.com/safe-global/safe-wallet-web",
   "license": "GPL-3.0",
-  "version": "1.66.1",
+  "version": "1.67.0",
   "type": "module",
   "scripts": {
     "dev": "next dev",


francovenica:
Issue found by automation.

For proposers, if their address is an entry in the AB, then that name should overwrite the name given during the proposer creation
in this PR, the name given during the proposer's creation is overwriting the name given in the local AB and the Spaces AB.

The name priority should be: Spaces AB entry > Local AB entry > Name given during the proposer creation

usame-algan:
> Issue found by automation.
> 
> For proposers, if their address is an entry in the AB, then that name should overwrite the name given during the proposer creation in this PR, the name given during the proposer's creation is overwriting the name given in the local AB and the Spaces AB.
> 
> The name priority should be: Spaces AB entry > Local AB entry > Name given during the proposer creation

This actually happened because I changed the order of priority so that if a name is passed to `EthHashInfo` it always has priority over the address book names. We need this within Spaces for the safe account addresses. I am not sure why we decided for Proposers that address book should have priority over the name given but IMO it should be the other way around. @the-real-chrizzo is this change ok? Then I would say we can update our automation to match the new logic.

the-real-chrizzo:
> > Issue found by automation.
> > For proposers, if their address is an entry in the AB, then that name should overwrite the name given during the proposer creation in this PR, the name given during the proposer's creation is overwriting the name given in the local AB and the Spaces AB.
> > The name priority should be: Spaces AB entry > Local AB entry > Name given during the proposer creation
> 
> This actually happened because I changed the order of priority so that if a name is passed to `EthHashInfo` it always has priority over the address book names. We need this within Spaces for the safe account addresses. I am not sure why we decided for Proposers that address book should have priority over the name given but IMO it should be the other way around. @the-real-chrizzo is this change ok? Then I would say we can update our automation to match the new logic.

I agree with Franco's assessment here re the name priority.

usame-algan:
> I agree with Franco's assessment here re the name priority.

Ok I will revert the change until we can think of a better solution.

PR #6234, opened 2025-08-21
feat(analytics): Consolidate tracking system

vseehausen:
# Simple Analytics Consolidation

## What it solves

Resolves https://linear.app/safe-global/issue/GRO-37/consolidate-tracking

Replaces manual dual-provider tracking with automated event routing. Currently, components must call both GA and MixPanel separately for each event.

Problems addressed:
- Scattered tracking calls throughout codebase
- Risk of missing events for specific providers  
- Maintenance overhead when adding new events

## How this PR fixes it

- **Unified tracking**: Single `trackEvent()` call automatically routes to both providers
- **Declarative mapping**: `GA_TO_MIXPANEL_MAPPING` defines which events go to MixPanel
- **Type safety**: Uses actual event properties instead of hardcoded strings

### Before (Manual dual tracking)
```typescript
// Two separate calls for each event
trackEvent({ ...WALLET_EVENTS.CONNECT, label: wallet.label })
trackMixPanelEvent(MixPanelEvent.WALLET_CONNECTED, { 
  'EOA Wallet Label': wallet.label,
  'EOA Wallet Address': wallet.address,
})
```

### After (Unified tracking)
```typescript
// Single call, automatic routing
trackEvent(
  { ...WALLET_EVENTS.CONNECT, label: wallet.label },
  {
    [MixPanelEventParams.EOA_WALLET_LABEL]: wallet.label,
    [MixPanelEventParams.EOA_WALLET_ADDRESS]: wallet.address,
  }
)
```

## Implementation

```typescript
// Event mapping
export const GA_TO_MIXPANEL_MAPPING: Record<string, string> = {
  [CREATE_SAFE_EVENTS.CREATED_SAFE.action]: MixPanelEvent.SAFE_CREATED,
  [WALLET_EVENTS.CONNECT.action]: MixPanelEvent.WALLET_CONNECTED,
  [SAFE_APPS_EVENTS.OPEN_APP.action]: MixPanelEvent.SAFE_APP_LAUNCHED,
}

// Unified tracking function
export const trackEvent = (eventData: AnalyticsEvent, additionalParameters?: Record<string, any>): void => {
  gtmTrack(eventData)
  
  const mixpanelEventName = GA_TO_MIXPANEL_MAPPING[eventData.action]
  if (mixpanelEventName) {
    mixpanelTrack(mixpanelEventName, additionalParameters)
  }
}
```

## Events Tracked

- **Safe Created**: Deployment context, owner count, threshold, payment method
- **Safe Activated**: Counterfactual deployment tracking  
- **Wallet Connected**: Wallet type, network, connection method
- **Safe App Launched**: App name, launch location context

## Testing

Connect a wallet and check if the event is tracked on MixPanel and Google Analytics.

## Checklist

- [ ] I've documented how it affects the analytics (if at all) 📊 - **Yes, consolidates analytics tracking**
- [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻 - **Yes, enhanced test coverage**


vseehausen:
Rationale why some parameters are double-tracked at some points (answer to https://github.com/safe-global/safe-wallet-monorepo/pull/6229#discussion_r2287725463):

* I want to avoid a race condition and make sure the parameters are tracked correctly (i.e., directly after connecting an EOA wallet, the address may not be set in memory, so it is tracked in the event directly).
* Also, setting global tracking variables is done using `gtmTrack`, which is the function of the gt/ga tracking system. To use this for both tracking systems, the function would require some refactoring and a generalized function. This would increase the scope of the PR and can be done at a later phase.

apps/web/src/components/new-safe/create/steps/ReviewStep/index.tsx
```diff
@@ -299,13 +305,29 @@ const ReviewStep = ({ data, onSubmit, onBack, setStep }: StepRenderProps<NewSafe
 
     gtmSetChainId(chain.chainId)
 
+    trackEvent(CREATE_SAFE_EVENTS.CREATED_SAFE, {
+      [MixPanelEventParams.SAFE_ADDRESS]: safeAddress.toLowerCase(),
```
compojoom: I see that you do this toLowerCase call everywhere we call the function. I guarantee 100% that someone is going to forget that toLowerCase is necessary. Can't we move this toLowerCase somewhre inside of trackEvent so that it is automatic. As someone who uses the trackEvent function I shouldn't need to know of the internal details of the function.

compojoom: I looked in the code. I guess I would do this in safeMixPanelTrack. If the props have a SAFE_ADDRESS cast it to lowercase just in case.

vseehausen: Yep I agree that would be best. We would have to do this for user properties, too. And ideally everything that could be an address. I can add it, but it will add more code obviously.

vseehausen: This should be resolved now.

apps/web/src/services/analytics/mixpanel.ts
```diff
@@ -1,27 +1,47 @@
...
+
+const cleanProperty = ([key, value]: [string, any]): [string, any] => [
+  key,
+  isAddress(key) ? cleanAddress(value) : value,
+]
+
+const cleanProperties = (properties: Record<string, any>): Record<string, any> => {
+  return Object.fromEntries(Object.entries(properties).map(cleanProperty))
+}
+
```
compojoom: I want to suggest some renaming to better convey what those functions are doing. 

clean is ofter used in the context of deleting. In our case we aren't deleting. 

I would suggest to use normalize - as we are converting props to a format the mixpanel would expect. 

cleanAddress I would rename to lowercaseAddress.

other than that it is good.

compojoom (approved):

PR #6235, opened 2025-08-22
fix(web): EIP-7702 accounts are treated as smart accounts

compojoom:
## What it solves
If an EOA has been upgraded to an EIP-7702 account, then this EOA was not able to sign off-chain signatures. Our checks were treating such account as a smart account and those have to submit on-chain signratures.

Resolves #

## How this PR fixes it
Our isSmartContract check returns true whenever getCode returns anything other than an empty data. For EIP-7702 accounts we do have data which is prefixed with 0xef0100. Those accounts are both smart accounts and EOAs. The user is still able to use them as EOAs. Now when we want to know if the wallet is an EOA we also check if the getCode returned is prefixed with 0xef0100

## How to test it
Use for example MM and upgrade an EOA to a smart acocunt. 
Now try to sign a tx - you should get a signrature prompt in MM instead of a on-chain tx request. 

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/utils/wallets.ts
+++ b/apps/web/src/utils/wallets.ts
@@ -9,6 +9,7 @@ import { type JsonRpcProvider } from 'ethers'
 
 const WALLETCONNECT = 'WalletConnect'
 const WC_LEDGER = 'Ledger Wallet'
+export const EIP_7702_DELEGATED_ACCOUNT_PREFIX = '0xef0100'
 
 const isWCRejection = (err: Error): boolean => {
   return /rejected/.test(err?.message)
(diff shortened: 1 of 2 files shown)


usame-algan (approved):
Code looks good. The two new functions are not used but if I understand correctly its because we only want to do one request instead of two so I think its fine. Of course we could have two more helper functions that just receive the code and decide - which could be reused so the logic would be in one place but I would leave that up to you.

apps/web/src/utils/wallets.ts
```diff
@@ -48,21 +49,31 @@ export const isPKWallet = (wallet: ConnectedWallet): boolean => {
...
 
+export const isEIP7702DelegatedAccount = async (address: string, provider?: JsonRpcProvider): Promise<boolean> => {
+  const code = await getAccountCode(address, provider)
+  return code.startsWith(EIP_7702_DELEGATED_ACCOUNT_PREFIX)
+}
+
 export const isSmartContractWallet = memoize(
   async (_chainId: string, address: string): Promise<boolean> => {
-    return isSmartContract(address)
+    const isContract = await isSmartContract(address)
```
rmeissner, line 74: nit: with this approach (two separate "is" functions) you fire 2 times the same request to get the account code. Not sure if the web3 provider caches this

francovenica:
LGTM

I made my MM account a smart contract account and tried to sign tx. The signature request were all off-chain
I tested also spending limits, recovery and nested safe's tx.

apps/web/src/utils/wallets.ts
compojoom: it should be fine. I checked before doing this whether ethersjs caches this. And it does if the calls are within 250ms which should be the case.

compojoom: Also just checked it to be 100% sure in the browser network tab - I only see one request when I try to sign. So all good I would say.

PR #6236, opened 2025-08-22
Refactor(AB): Address book hooks to handle multiple sources

usame-algan:
## What it solves

Resolves [COR-493](https://linear.app/safe-global/issue/COR-493/spaces-address-book-names-are-only-visible-if-wallet-is-on-the-correct)

## How this PR fixes it

- Updates `useAddressBook` to respect multiple address book sources
- Makes address book hooks context aware of spaces

## How to test it

- Within spaces any address name should only come from the space address book
- Within a safe any address name should either come from the space address book or from the local one in that order

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
+++ b/apps/web/src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
@@ -32,7 +32,6 @@ export type EthHashInfoProps = {
   children?: ReactNode
   trusted?: boolean
   ExplorerButtonProps?: ExplorerButtonProps
-  isAddressBookName?: boolean
   addressBookNameSource?: ContactSource
   highlight4bytes?: boolean
 }
@@ -56,7 +55,6 @@ const SrcEthHashInfo = ({
   ExplorerButtonProps,
   children,
   trusted = true,
-  isAddressBookName = false,
   addressBookNameSource,
   highlight4bytes = false,
 }: EthHashInfoProps): ReactElement => {
@@ -106,7 +104,7 @@ const SrcEthHashInfo = ({
               {name}
             </Box>
 
-            {isAddressBookName && (
+            {!!addressBookNameSource && (
               <Tooltip title={`From your ${addressBookNameSource} address book`} placement="top">
                 <span style={{ lineHeight: 0 }}>
                   <SvgIcon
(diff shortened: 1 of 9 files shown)


apps/web/src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
```diff
@@ -106,7 +104,7 @@ const SrcEthHashInfo = ({
               {name}
             </Box>
 
-            {isAddressBookName && (
```
usame-algan, line 109: I removed this because we basically had two props while we can handle this through one prop.

apps/web/src/components/tx-flow/flows/NestedTxSuccessScreen/index.tsx
```diff
@@ -80,12 +80,7 @@ const NestedTxSuccessScreen = ({ txId }: Props) => {
             <Typography variant="body2" color="text.secondary">
               Parent Safe
             </Typography>
-            <EthHashInfo
-              address={cachedPendingTx.signerAddress}
-              name={parentSafeAddress}
-              isAddressBookName={Boolean(parentSafeAddress)}
```
usame-algan, line 86: Removed it here as well since it will show the address book icon if we can find an entry in the address book so behaviour should stay the same.

apps/web/src/features/spaces/components/SpaceAddressBook/EditContactDialog.tsx
```diff
@@ -122,7 +122,7 @@ const EditContactDialog = ({ entry, onClose }: EditContactDialogProps) => {
             <Typography mb={2}>Edit contact details. Anyone in the space can see it.</Typography>
             <Stack spacing={3}>
               <Box pt={1}>
-                <AddressInputReadOnly address={entry.address} />
+                <AddressInputReadOnly address={entry.address} chainId={entry.chainIds[0]} />
```
usame-algan, line 125: A bit hacky but we know every entry has at least one network and the name is the same regardless of chainId.

apps/web/src/hooks/useAllAddressBooks.ts
```diff
@@ -66,16 +90,29 @@ export const useAllMergedAddressBooks = (chainId?: string): ExtendedContact[] =>
...
+      const item = get(address, chainId)
+      return item?.source === ContactSource.space ? item : undefined
+    }
+
+    return undefined
+  }, [address, chainId, source, get])
 }
 
+// Returns all local address books
+const useAllAddressBooks = () => useAppSelector(selectAllAddressBooks)
```
compojoom, line 116: should we maybe rename this to useAllLocalAddressBooks and remove the comment?

usame-algan:
Closing this in favour of https://github.com/safe-global/safe-wallet-monorepo/pull/6240

PR #6238, opened 2025-08-22
fix: Prefer canonical contract addresses for Safe creation

tmjssz:
## What it solves

[COR-509](https://linear.app/safe-global/issue/COR-509/prefer-canonical-contract-addresses-for-multichain-safe-creation-on)

Prefer the canonical addresses for core contracts when creating a new Safe. This is necessary to ensure the multichain Safe creation works when deploying on zkSync/Lens and an EVM-based chain.

## How this PR fixes it
Introduced `getCanonicalOrFirstAddress` util in packages/utils, refactored web Safe creation to use it for all contracts, and added unit tests.

### 1. Addition of EVM-based network to zkSync or Lens Safe
1. Create multichain Safe on Lens/zkSync
2. Then add Sepolia
3. Expected: No errors due to chain-specific addresses.

### 2. Safe creation
1. Create multichain Safe on Lens/zkSync + Sepolia
2. Expected: No errors due to chain-specific addresses.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/new-safe/create/logic/index.test.ts
+++ b/apps/web/src/components/new-safe/create/logic/index.test.ts
@@ -35,6 +35,8 @@ import {
 } from '@safe-global/safe-deployments'
 import { Safe_to_l2_setup__factory } from '@safe-global/utils/types/contracts'
 import { FEATURES, getLatestSafeVersion } from '@safe-global/utils/utils/chains'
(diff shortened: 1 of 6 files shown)


usame-algan:
Could we check the git history on this? I remember in the past we switched from using the canonical addresses to using the networkAddresses for a good reason.

Edit: Just saw that we have a fallback logic in place so I think it should be fine.

PR #6248, opened 2025-08-25
Feat(Analytics): Add Positions tracking

vseehausen:
## What it solves

Resolves https://linear.app/safe-global/issue/GRO-36/track-positions-widget

## How this PR fixes it

Adds events as defined in the issue.

  ### New Events
  - **PositionExpanded**: Tracks accordion expansion with protocol details
  - **PositionsViewAllClicked**: Tracks navigation to full positions page
  - **EmptyPositionsExploreClicked**: Tracks explore button in empty state
  - **StakeViewed**: Tracks stake button views with entry point context
  - **EarnViewed**: Tracks earn button views with entry point context
  - **PositionsWidgetViewed**: Tracks widget visibility (dashboard + positions page)

## Checklist

---

## CLA signature


apps/web/src/features/positions/components/PositionsWidget/index.tsx
```diff
@@ -60,25 +75,64 @@ const PositionsWidget = () => {
...
+  useEffect(() => {
+    if (isVisible && currentData) {
+      const positionCount = protocols.length
+      const totalValue = protocols.reduce((sum, protocol) => sum + (Number(protocol.fiatTotal) || 0), 0)
+
+      trackEvent(POSITIONS_EVENTS.POSITIONS_WIDGET_VIEWED, {
+        [MixPanelEventParams.LOCATION]: 'Dashboard',
+        [MixPanelEventParams.NUMBER_OF_POSITIONS]: positionCount,
+        [MixPanelEventParams.TOTAL_VALUE_OF_PORTFOLIO]: totalValue,
+      })
```
usame-algan: Where is the difference to users viewing the dashboard? The positions widget is visible almost at the top so we can assume if someone visits the dashboard they see the widget. IMO this event is not necessary and adds complexity.

apps/web/src/features/positions/components/PositionsWidget/index.tsx
```diff
@@ -60,25 +75,64 @@ const PositionsWidget = () => {
     [safe],
   )
 
-  if (!currentData) return null
+  const protocols = useMemo(() => currentData?.slice(0, MAX_PROTOCOLS) || [], [currentData])
```
usame-algan: I would suggest not changing this as it doesn't have anything to do with the PR and not everything needs to be memoized.

apps/web/src/features/positions/index.tsx
```diff
@@ -8,15 +9,33 @@ import IframeIcon from '@/components/common/IframeIcon'
...
+    if (isVisible && currentData) {
+      const positionCount = currentData.length
+
+      trackEvent(POSITIONS_EVENTS.POSITIONS_WIDGET_VIEWED, {
+        [MixPanelEventParams.LOCATION]: 'Positions Page',
+        [MixPanelEventParams.NUMBER_OF_POSITIONS]: positionCount,
+        [MixPanelEventParams.TOTAL_VALUE_OF_PORTFOLIO]: fiatTotal || 0,
+      })
+    }
+  }, [isVisible, currentData, fiatTotal])
```
usame-algan: Same as before. Is there a difference to simply tracking the page views for the positions page? If not then I suggest to remove this as it only adds complexity.

apps/web/src/features/positions/components/PositionsWidget/index.tsx
vseehausen: But we need to get the positions count somehow, no?

apps/web/src/features/positions/index.tsx
vseehausen: removed.

apps/web/src/features/positions/components/PositionsWidget/index.tsx
vseehausen: removed.

apps/web/src/features/positions/components/PositionsWidget/index.tsx
vseehausen: removed.

usame-algan:
Somehow this PR added 50kb to our global bundle size which is unusual. We should revisit this and figure out why.

vseehausen:
It's always growth...

PR #6254, opened 2025-08-26
fix(web): dont show empty state for assets/pending tx before loaded

LucieFaire:
## What it solves

Resolves #

https://github.com/safe-global/safe-wallet-monorepo/pull/6218#issuecomment-3223953546

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/dashboard/Assets/index.tsx
+++ b/apps/web/src/components/dashboard/Assets/index.tsx
@@ -130,7 +130,7 @@ export const isNonZeroBalance = (item: Balances['items'][number]) => item.balanc
 const AssetsWidget = () => {
   const router = useRouter()
   const { safe } = router.query
-  const { loading } = useBalances()
+  const { loading, balances } = useBalances()
   const visibleAssets = useVisibleAssets()
 
   const items = useMemo(() => {
@@ -145,7 +145,11 @@ const AssetsWidget = () => {
     [safe],
   )
 
-  if (loading) return <AssetsSkeleton />
+  // Only show content when data has been fetched (either has items or fiatTotal is set)
+  const hasDataLoaded = visibleAssets.length > 0 || balances.fiatTotal !== ''
+  const isLoading = loading || !hasDataLoaded
+
+  if (isLoading) return <AssetsSkeleton />
 
   return (
     <Card data-testid="assets-widget" sx={{ px: 1.5, py: 2.5 }}>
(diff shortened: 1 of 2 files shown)


apps/web/src/components/dashboard/Assets/index.tsx
```diff
@@ -145,7 +145,11 @@ const AssetsWidget = () => {
     [safe],
   )
 
-  if (loading) return <AssetsSkeleton />
+  // Only show content when data has been fetched (either has items or fiatTotal is set)
+  const hasDataLoaded = visibleAssets.length > 0 || balances.fiatTotal !== ''
+  const isLoading = loading || !hasDataLoaded
+
+  if (isLoading) return <AssetsSkeleton />
```
LucieFaire: @usame-algan there is an issue Franco [reported](https://github.com/safe-global/safe-wallet-monorepo/pull/6218#issuecomment-3223953546) where we show empty state shortly before the loading state. This addition fixes it, but I am not sure whether its a good way to do it. I have not found a better one, though. If you have an idea

apps/web/src/components/dashboard/PendingTxs/PendingTxsList.tsx
```diff
@@ -101,6 +101,9 @@ const PendingTxsList = (): ReactElement | null => {
 
   const totalTxs = recoveryTxs.length + queuedTxs.length
 
+  const isInitialState = !safeLoaded && !safeLoading
+  const isLoading = loading || safeLoading || isInitialState
```
LucieFaire, line 105: this logic repeats what we do for Overview to not have the empty state until everything is loaded

apps/web/src/components/dashboard/Assets/index.tsx
usame-algan: I think this approach makes sense. We could probably do the same we do in the sidebar header by just checking `balances.fiatTotal`? Checking `visibleAssets.length` might introduce other problems e.g. what happens if there are no visible assets because the user has marked them all for spam etc.

usame-algan (approved):
The loading state looks nice now. I don't see any flickering. Just left one comment regarding checking visible assets but other than that its good to go!

Image

apps/web/src/components/dashboard/Assets/index.tsx
LucieFaire: yes makes sense, will update

PR #6256, opened 2025-08-26
feat(mobile): add sent/receive screens for history txs

compojoom:
## What it solves
This PR ads custom decoding for Sent|Received token & NFT transfers

Resolves https://linear.app/safe-global/issue/COR-517/sent-screen
Resolves https://linear.app/safe-global/issue/COR-518/receive-eth-screen

## How this PR fixes it
First I identified some common components that could be reused between the confirmation views and the history views & refactored those. Then I implemented the new screens.

## How to test it
Start the app - click any item in the history - it should open a "transaction details" page. If you are looking at a sent|receive tx it should have decoding as in the screenshots.

## Screenshots
grafik
grafik
grafik
grafik

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/ConfirmTx/components/confirmation-views/SwapOrder/hooks/useRecipientItem.test.tsx
+++ b/apps/mobile/src/features/ConfirmTx/components/confirmation-views/SwapOrder/hooks/useRecipientItem.test.tsx
@@ -36,6 +36,10 @@ jest.mock('tamagui', () => ({
   View: 'View',
 }))
 
+jest.mock('@/src/components/HashDisplay', () => ({
+  HashDisplay: 'HashDisplay',
+}))
+
 const createMockOrder = (
   receiver?: string,
   owner = '0x1234567890123456789012345678901234567890',


apps/mobile/src/features/ConfirmationsSheet/ConfirmationsSheet.container.tsx
```diff
@@ -69,26 +85,26 @@ export const ConfirmationsSheetContainer = () => {
                 circular={false}
                 content={
                   <View alignItems="center" flexDirection="row" gap="$1">
-                    {hasSigned && <SafeFontIcon size={12} name="check" />}
+                    {(isHistoryTransaction || hasSigned) && <SafeFontIcon size={12} name="check" />}
 
                     <Text fontWeight={600} color={'$color'}>
-                      {hasSigned ? 'Signed' : 'Pending'}
+                      {isHistoryTransaction ? 'Signed' : hasSigned ? 'Signed' : 'Pending'}
```
usame-algan: Would it also work to do isHistoryTransaction || hasSigned here like above?

apps/mobile/src/features/HistoryAdvancedDetails/utils/formatHistoryTxDetails.tsx
```diff
@@ -0,0 +1,224 @@
...
+      items: basicInfoItems,
+    })
+  }
+
+  // Section 2: Parameters
+  const parametersItems: ListTableItem[] = []
+
+  // Call
+  if (txDetails.txData?.operation !== undefined) {
+    const operationText = txDetails.txData.operation === Operation.CALL ? 'transfer' : 'delegateCall'
```
usame-algan: Why is a call automatically a transfer?

apps/mobile/src/features/HistoryAdvancedDetails/utils/formatHistoryTxDetails.tsx
```diff
@@ -0,0 +1,224 @@
...
+          address={txDetails.txData?.to.value as Address}
+          copyProps={{ color: '$textSecondaryLight', size: 16 }}
+          externalLinkSize={16}
+        />
+      ),
+    })
+  }
+
+  // To (only show if different from Contract - in this case it's the same)
+  if (txDetails.txData?.to?.value) {
```
usame-algan, line 81: The condition is the same as above though so we always have both.

apps/mobile/src/features/HistoryAdvancedDetails/utils/formatHistoryTxDetails.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+  // Value
+  if (txDetails.txData?.value) {
+    parametersItems.push({
+      label: 'Value',
+      render: () => <Text>{txDetails.txData?.value}</Text>,
+    })
+  }
+
+  // Data
```
usame-algan: nit: 💬

apps/mobile/src/features/HistoryAdvancedDetails/utils/formatHistoryTxDetails.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+  // Section 3: Decoded data
+  const decodedDataItems: ListTableItem[] = []
+
+  if (isMultisigDetailedExecutionInfo(txDetails.detailedExecutionInfo)) {
+    const executionInfo = txDetails.detailedExecutionInfo
+
+    // Operation
+    if (txDetails.txData?.operation !== undefined) {
+      const operationText = txDetails.txData.operation === Operation.CALL ? '0(call)' : '1(delegateCall)'
```
usame-algan: Should we leave a space between the number and parentheses?

apps/mobile/src/features/HistoryAdvancedDetails/utils/formatHistoryTxDetails.tsx
```diff
@@ -0,0 +1,224 @@
...
+      decodedDataItems.push({
+        label: 'Operation',
+        render: () => <Text>{operationText}</Text>,
+      })
+    }
+
+    // Call
+    decodedDataItems.push({
+      label: 'Call',
+      render: () => <Text>0</Text>,
```
usame-algan: Above it can be a call or delegate call but here we always push call again?

apps/mobile/src/features/HistoryTransactionDetails/HistoryTransactionDetails.container.tsx
```diff
@@ -0,0 +1,66 @@
...
+    return <LoadingTx />
+  }
+
+  if (isError) {
+    return (
+      <View margin="$4">
+        <Alert type="error" message="Error fetching transaction details" />
+      </View>
+    )
+  }
```
usame-algan: Should the error block be above the loading? In case there is an error I can imagine that txDetails doesn't exist so we would always land in the loading block. Ideally this is covered by a test.

apps/mobile/src/features/HistoryTransactionDetails/components/history-views/HistoryTokenTransfer.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+interface HistoryTokenTransferProps {
+  txId: string
+  txInfo: TransferTransactionInfo
+  executedAt: number
+}
+
+export function HistoryTokenTransfer({ txId, txInfo, executedAt }: HistoryTokenTransferProps) {
+  const activeSafe = useDefinedActiveSafe()
+  const activeChain = useAppSelector((state: RootState) => selectChainById(state, activeSafe.chainId))
```
usame-algan: Can't we use `selectActiveChain` here?

apps/mobile/src/features/HistoryTransactionDetails/components/HistoryConfirmationsInfo/HistoryConfirmationsInfo.tsx
```diff
@@ -0,0 +1,76 @@
...
+  detailedExecutionInfo: MultisigExecutionDetails
+  txId: string
+}
+
+export function HistoryConfirmationsInfo({ detailedExecutionInfo, txId }: HistoryConfirmationsInfoProps) {
+  const router = useRouter()
+  const importedSigners = useAppSelector(selectSigners)
+
+  // Check if any of the imported signers (users on this device) have signed this transaction
+  const isUserSigned = detailedExecutionInfo?.confirmations?.some((confirmation) =>
```
usame-algan: hasUserSigned?

apps/mobile/src/features/HistoryTransactionDetails/components/HistoryTransactionInfo/HistoryTransactionInfo.tsx
```diff
@@ -0,0 +1,65 @@
...
+            </Text>
+          </View>
+        )}
+
+        {txHash && (
+          <View alignItems="center" flexDirection="row" justifyContent="space-between">
+            <Text color="$textSecondaryLight">Transaction hash</Text>
+            <View flexDirection="row" alignItems="center" gap="$2">
+              <Text fontSize="$4" color="$textPrimary">
+                {txHash.slice(0, 8)}...{txHash.slice(-6)}
```
usame-algan: You can use `shortenAddress` here.

apps/mobile/src/features/ConfirmationsSheet/ConfirmationsSheet.container.tsx
compojoom: Good point. I originally wrote it like this becuase if we are dealing with signed tx the only state available was obvisouly only "signed". But yea, doing like above is better.

apps/mobile/src/features/HistoryAdvancedDetails/utils/formatHistoryTxDetails.tsx
compojoom: Because I was too obsessed with working on "transfer screen". Logic is wrong. I'll fix it. Thanks!

apps/mobile/src/features/HistoryAdvancedDetails/utils/formatHistoryTxDetails.tsx
compojoom: Yes, I'll leave just the "to" label.

apps/mobile/src/features/HistoryAdvancedDetails/utils/formatHistoryTxDetails.tsx
compojoom: 🤦

apps/mobile/src/features/HistoryTransactionDetails/HistoryTransactionDetails.container.tsx
compojoom: good catch!

apps/mobile/src/components/HashDisplay/HashDisplay.tsx
```diff
@@ -0,0 +1,83 @@
...
+  copyProps?: {
+    color?: string
+    size?: number
+  }
+  externalLinkColor?: string
+  /** Gap between elements */
+  gap?: string
+  onExternalLinkPress?: () => void
+  /** Whether to treat the value as an address (shows identicon) or generic hash */
+  isAddress?: boolean
```
usame-algan, line 31: nit: It might be better to have both AddressDisplay and HashDisplay even if there is some duplication so its easier to extend each in the future without introducing more complexity. (I am just thinking about how much `EthHashInfo` has grown into a god component over the years).

PR #6257, opened 2025-08-27
chore: bump Safe deployments

dasanra:
## What it solves

Bumps protocol-kit and Safe deployments to add compatibility with latest chains

---

## CLA signature


Zalens:
@dasanra do you know when we can merge it?

PR #6259, opened 2025-08-27
chore: Normalize line endings to LF

usame-algan:
## What it solves

There were lint issues when merging branches because our `lint-staged` would run in the pre-commit hook which introduced lint issues for line endings e.g. in nested ternaries. This PR explicitely sets the `endOfLine` rule to lf. This is the default in prettier v2 but we were overwriting it.


.prettierrc
```diff
@@ -4,5 +4,5 @@
   "trailingComma": "all",
   "singleQuote": true,
   "semi": false,
-  "endOfLine": "auto"
+  "endOfLine": "lf"
```
usame-algan, line 7: I think it could also work if we remove this completely and the default takes effect.

vseehausen: I prefer having the rule set.

usame-algan:
After further testing I don't think this is it. Although probably still good to have this explicitely set.

usame-algan:
The svg was changed as part of the lint fixes. I don't think this solves the root issue but it is still good to enforce one line ending style so I would be in favor of merging this. More context: https://5afe.slack.com/archives/C07NN1TBBGU/p1756280298774239

vseehausen (approved):

vseehausen:
Ok then lgtm

PR #6260, opened 2025-08-27
Feat(Positions): Hide Currency and Token Select at Position Balance

vseehausen:
## What it solves

Removes these buttons from the Positions View, as they are not needed there:

image

## Checklist

---

## CLA signature


usame-algan (approved):
Looks good

PR #6267, opened 2025-08-28
Feat(Positions): Fix decimals and alignment of asset and positions value

vseehausen:
## What it solves

* Shows decimals when they are required.
* Aligns columns of Asset and Positions view to the right.

## How to test it

1. Look at Asset and Position view. 
2. Observe decimals.
3. Observe alignment.

## Screenshots

grafik

grafik

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/balances/AssetsTable/index.tsx
+++ b/apps/web/src/components/balances/AssetsTable/index.tsx
@@ -211,7 +211,7 @@ const AssetsTable = ({
               content: (
                 <Box textAlign="left">
                   <Typography>
-                    <FiatValue value={item.fiatBalance} />
+                    <FiatValue value={item.fiatBalance} precise uniformColor />
                   </Typography>
                   {item.fiatBalance24hChange && (
                     <Typography variant="caption">
--- a/apps/web/src/components/balances/AssetsTable/index.tsx
+++ b/apps/web/src/components/balances/AssetsTable/index.tsx
@@ -92,30 +92,31 @@ const headCells = [
   {
     id: 'asset',
     label: 'Asset',
-    width: '40%',
+    width: '20%',
   },
   {
     id: 'price',
     label: 'Price',
-    width: '15%',
-    align: 'left',
+    width: '20%',
+    align: 'right',
   },
   {
     id: 'balance',
     label: 'Balance',
-    width: '15%',
+    width: '20%',
+    align: 'right',
   },
   {
     id: 'value',
     label: 'Value',
-    width: '15%',
-    align: 'left',
+    width: '20%',
+    align: 'right',
   },
   {
     id: 'weight',
     label: 'Weight',
-    width: '15%',
-    align: 'left',
+    width: '20%',
+    align: 'right',
   },
   {
     id: 'actions',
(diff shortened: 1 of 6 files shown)


apps/web/src/components/balances/AssetsTable/index.tsx
```diff
@@ -92,30 +92,31 @@ const headCells = [
   {
     id: 'asset',
     label: 'Asset',
-    width: '40%',
+    width: '20%',
```
usame-algan: Let's check this on safes with "weird" assets that have long names to make sure the layout is not broken. @francovenica can you help?

apps/web/src/components/balances/AssetsTable/index.tsx
```diff
@@ -187,31 +188,35 @@ const AssetsTable = ({
             price: {
               rawValue: rawPriceValue,
               content: (
-                <Typography>
-                  <FiatValue value={item.fiatConversion} />
-                </Typography>
+                <Box textAlign="right">
+                  <Typography>
```
usame-algan: Can't you add this directly to the Typography element?

apps/web/src/components/balances/AssetsTable/index.tsx
```diff
@@ -187,31 +188,35 @@ const AssetsTable = ({
...
               collapsed: item.tokenInfo.address === hidingAsset,
               content: (
-                <Typography sx={{ '& b': { fontWeight: '400' } }}>
-                  <TokenAmount
-                    value={item.balance}
-                    decimals={item.tokenInfo.decimals}
-                    tokenSymbol={item.tokenInfo.symbol}
-                  />
-                </Typography>
+                <Box textAlign="right">
```
usame-algan: Same as above. Maybe it works adding this to the Typography element. This helps keep the DOM cleaner.

apps/web/src/components/balances/AssetsTable/index.tsx
```diff
@@ -224,39 +229,43 @@ const AssetsTable = ({
...
-                    variant="determinate"
-                    value={itemShareOfFiatTotal * 100}
-                    className={css.circleProgress}
-                    size={16}
-                    thickness={6}
-                    sx={{ 'svg circle': { stroke: 'url(#progress_gradient)', strokeLinecap: 'round' } }}
-                  />
-                  <Typography variant="body2">{formatPercentage(itemShareOfFiatTotal)}</Typography>
-                </Stack>
+                <Box textAlign="right">
```
usame-algan, line 228: Same here I think it would work adding this to Stack

apps/web/src/features/positions/index.tsx
```diff
@@ -52,9 +52,12 @@ export const Positions = () => {
                   },
                   balance: {
                     content: (
-                      <Typography>
-                        {formatVisualAmount(position.balance, position.tokenInfo.decimals)} {position.tokenInfo.symbol}
-                      </Typography>
+                      <Box textAlign="right">
```
usame-algan: Same here with text align

apps/web/src/components/balances/AssetsTable/index.tsx
vseehausen: In this case it should break the line.

apps/web/src/components/balances/AssetsTable/index.tsx
vseehausen: Here we actually need it, as the Progress indicator breaks otherwise.

apps/web/src/components/balances/AssetsTable/index.tsx
vseehausen: You are right.

apps/web/src/components/balances/AssetsTable/index.tsx
vseehausen: You are right.

apps/web/src/components/balances/AssetsTable/index.tsx
vseehausen: grafik

vseehausen: Still, I increased the asset name column by 10%.

usame-algan: Thanks for testing!

PR #6273, opened 2025-08-29
fix(web): migration contract no longer enforces nonce 0

compojoom:
The latest migration contracts no longer enforce nonce === 0, so we don’t need this check anymore.

## What it solves

Resolves #

## Checklist

---

## CLA signature


packages/utils/src/services/contracts/safeContracts.ts
```diff
@@ -26,12 +26,9 @@ export const _getValidatedGetContractProps = (
   }
 }
 export const isMigrationToL2Possible = (safe: SafeState): boolean => {
-  return (
-    safe.nonce === 0 &&
```
katspaugh, line 30: AFAIR, this check was there to make sure that the indexer works. If nonce is more than 0, i.e. there were transactions made, the indexer refuses to index follow-up txs.

@Uxio0 is it still the case, or can we merge this?

compojoom:
I think I looked at the wrong contract and nonce 0 is still required.

PR #6276, opened 2025-08-29
feat(positions): Add loading and error state widgets

vseehausen:
## What it solves

Resolves https://linear.app/safe-global/issue/COR-543/improve-empty-state and https://linear.app/safe-global/issue/GRO-47/qa-final-part-2#comment-ae75de88

## How this PR fixes it

Adds an Error Widget and a loading skeleton indicator.

The Error state (PositionsUnavailable) is displayed when the feature flag is turned on, but CGW returns an error.
The loading skeleton is shown when the CGW call is loading.

## How to test it

Check a Safe's position on Sepolia, observe the error widget.
Load the position balances on any chain and observe the loading skeleton.

## Screenshots
Loading:
image

Error:
image

## Checklist

---

## CLA signature


francovenica:
Looks better.
Leaving snapshots as documentation:
image

image

apps/web/src/features/positions/index.tsx
```diff
@@ -11,13 +11,103 @@ import PositionsEmpty from '@/features/positions/components/PositionsEmpty'
...
+                  ]}
+                  compact
+                />
+              </Box>
+            </AccordionDetails>
+          </Accordion>
+        </Card>
+      </Stack>
+    )
+  }
```
LucieFaire, line 108: I would maybe extract it in the separate file called `PositionsSkeleton`

apps/web/src/features/positions/components/PositionsWidget/index.tsx
```diff
@@ -31,9 +41,81 @@ const PositionsWidget = () => {
     [safe],
   )
 
-  if (!currentData) return null
+  if (isLoading) {
+    return (
```
LucieFaire, line 45: would also extract it to a separate file

apps/web/src/features/positions/hooks/usePositionsFiatTotal.ts
```diff
@@ -1,7 +1,7 @@
 import usePositions from '@/features/positions/hooks/usePositions'
 
 const usePositionsFiatTotal = () => {
-  const protocols = usePositions()
+  const { data: protocols } = usePositions()
```
LucieFaire, line 4: do you need to rename it?

apps/web/src/features/positions/index.tsx
```diff
@@ -11,13 +11,103 @@ import PositionsEmpty from '@/features/positions/components/PositionsEmpty'
...
 
+  if (error || !protocols) return <PositionsUnavailable />
```
LucieFaire, line 110: I am not sure, but there is a case when there is no error and no protocols?

vseehausen: I was not sure about this, but `!protocols` has been used before, and it doesn't harm to define it.

LucieFaire: ok

PR #6277, opened 2025-08-29
feat(mobile): add on-chain rejection view for history txs

compojoom:
## What it solves
Rejection txs now have custom decoding

Resolves https://linear.app/safe-global/issue/COR-532/tx-on-chain-rejection

## How this PR fixes it
Abstracted the GenericView as the rejection view displays nearly the same content + a description of what the tx did. 

## How to test it
Open an on-chain rejection tx.

## Screenshots
grafik

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/HistoryTransactionDetails/components/history-views/HistoryTransactionBase.tsx
+++ b/apps/mobile/src/features/HistoryTransactionDetails/components/history-views/HistoryTransactionBase.tsx
@@ -1,0 +1,75 @@
+import React from 'react'
+import { View, YStack, Text } from 'tamagui'
+import { TransactionHeader } from '@/src/features/ConfirmTx/components/TransactionHeader'
+import { Container } from '@/src/components/Container'
+import { Address } from '@/src/types/address'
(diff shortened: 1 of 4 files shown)


apps/mobile/src/features/HistoryTransactionDetails/components/history-views/HistoryTransactionBase.tsx
```diff
@@ -0,0 +1,88 @@
...
+  recipientAddress,
+  logo,
+  customLogo,
+  badgeIcon,
+  badgeThemeName,
+  badgeColor,
+  title,
+  isIdenticon,
+  description,
+  additionalContent,
```
tmjssz: How about renaming this prop to `children`, so the additional content could be passed in as nested JSX?

compojoom: Good idea! Thanks! I've changed it.

tmjssz (approved):
Nice!

PR #6279, opened 2025-08-29
feat(mobile): add contact creation when adding a signer

PooyaRaki:
## What it solves
Enhanced the `addSignerWithEffects` function to dispatch an `addContact` action, creating a new contact for each signer added. The contact is named based on the signer's address, ensuring better organization in the address book.

Resolves #
https://linear.app/safe-global/issue/COR-539/mobile-name-signers-automatically

## How to test it
- Import a new signer
- Go to the address book
- The signer should be added with a readable name to the address book

## Checklist

---

## CLA signature


compojoom (approved):
Awesome, thanks!

PR #6281, opened 2025-08-30
implement ens resolution for safe owner list

ghostffcode:
## What it solves

Resolves [https://x.com/austingriffith/status/1961450319411646535](https://x.com/austingriffith/status/1961450319411646535)

## How this PR fixes it

Add an ENS resolution hook for Safe owners

## How to test it

Load a safe -> Settings -> Setup -> Signers

## Screenshots

safe-signers-with-ens

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/settings/owner/OwnerList/index.tsx
+++ b/apps/web/src/components/settings/owner/OwnerList/index.tsx
@@ -17,26 +17,36 @@ import { TxModalContext } from '@/components/tx-flow'
 import ReplaceOwnerIcon from '@/public/images/settings/setup/replace-owner.svg'
 import DeleteIcon from '@/public/images/common/delete.svg'
 import type { AddressBook } from '@/store/addressBookSlice'
-
 import tableCss from '@/components/common/EnhancedTable/styles.module.css'
+import useResolvedOwners from '@/hooks/useResolvedOwners'
 
 export const OwnerList = () => {
   const addressBook = useAddressBook()
   const { safe } = useSafeInfo()
   const { setTxFlow } = useContext(TxModalContext)
+  const owners = useResolvedOwners()
 
   const rows = useMemo(() => {
-    const showRemoveOwnerButton = safe.owners.length > 1
+    const showRemoveOwnerButton = owners.length > 1
 
-    return safe.owners.map((owner) => {
+    return owners.map((owner) => {
       const address = owner.value
-      const name = addressBook[address]
+      let name = addressBook[address] || owner.name || undefined
 
       return {
         cells: {
           owner: {
             rawValue: address,
-            content: <EthHashInfo address={address} showCopyButton shortAddress={false} showName={true} hasExplorer />,
+            content: (
+              <EthHashInfo
+                address={address}
+                showCopyButton
+                shortAddress={false}
+                showName={true}
+                name={name}
+                hasExplorer
+              />
+            ),
           },
           actions: {
             rawValue: '',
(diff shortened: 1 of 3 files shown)


vseehausen:
Thanks for opening this, happy to assist moving this to a merge!

Good idea to add the hook. Another solution could be to solve this at the address level.

We also need to test if the ENS data is available, as ENS is notoriously unreliable. I will check this later.

austintgriffith:
"ENS is notoriously unreliable"

ser it is a smart contract

ENS _always_ works, what do you mean by this?

v1rtl:
Would be cool to also implements this for proposers (same approach basically)

vseehausen:
I tested resolving addresses to ens name locally with the ens resolver from ethers v6, and couldn't get reliable results. ENS is NOT notoriously unreliable, sorry, crypto Twitter. :D Guess it's just me.

Just want to make sure this is reliable on the frontend.

vseehausen:
Ah, also to provide some context, I have been testing resolving ENS addresses globally last week, but was not successful. https://github.com/safe-global/safe-wallet-monorepo/pull/6311 I think it would be a nice feature.

v1rtl:
> I tested resolving addresses to ens name locally with the ens resolver from ethers v6, and couldn't get reliable results. ENS is NOT notoriously unreliable, sorry, crypto Twitter. :D Guess it's just me.
> 
> Just want to make sure this is reliable on the frontend.

must be an RPC issue. Try with a different RPC URL.

could you also please provide the list of ENS names that you had trouble resolving? They might use non-standard resolvers (for example involving CCIP read)

apps/web/src/components/settings/owner/OwnerList/index.tsx
```diff
@@ -17,26 +17,36 @@ import { TxModalContext } from '@/components/tx-flow'
...
 
   const rows = useMemo(() => {
-    const showRemoveOwnerButton = safe.owners.length > 1
+    const showRemoveOwnerButton = owners.length > 1
 
-    return safe.owners.map((owner) => {
+    return owners.map((owner) => {
       const address = owner.value
-      const name = addressBook[address]
+      let name = addressBook[address] || owner.name || undefined
```
katspaugh:
```suggestion
      const name = addressBook[address] || owner.name || undefined
```

This variable is never reassigned, so it can remain a constant.

katspaugh:
Thanks for the contribution! We already have a component called `NamedAddressInfo` which we could use here to display the owner list. The same component can be used for proposers as per @talentlessguy's comment.

vseehausen:
image

It's actually been the RPC.

austintgriffith:
starting to look great! 

whatever we can do to help with this, let us know. 

we just want to be able to see ens names with addresses everywhere in the safe. 

however we can get to there, you tell us. 

we are happy to do it in small PRs too. 

good work so far!! thank you!!!

katspaugh:
@austintgriffith we'd prefer at this time to just add resolution to owners/proposers as per the original PR scope. We'll discuss doing reverse resolution everywhere in the app but that has to be approached carefully due to performance/RPC rate limits. We might do it on the backend.

So to accept this PR, I'd suggest we use NamedAddressInfo in the Owners and Proposers lists as per my [comment](https://github.com/safe-global/safe-wallet-monorepo/pull/6281#issuecomment-3249324564) above. No need for a new hook then. Let me know if this approach works. Cheers!

ghostffcode:
@katspaugh just pushed an update that:

- enables ENS resolution in `NamedAddressInfo ` component
- remove the custom hook
- use `NamedAddressInfo` for owner and proposers list

PR #6286, opened 2025-09-01
feat(mobile): safe settings history screens

compojoom:
## What it solves
Ads decoding screens for Safe owner added, removed, safe created and threshold changed

Resolves https://linear.app/safe-global/issue/COR-529/safe-addsigner
Resolves https://linear.app/safe-global/issue/COR-530/safe-created
Resolves https://linear.app/safe-global/issue/COR-531/safe-change-threshold
~~Resolves https://linear.app/safe-global/issue/COR-546/safe-owner-removed~~
After discussing this with @katspaugh we decided to drop the owner removed implementation as it adds too much complexity. Will add a backlog entry on CGW to support returning data for such tx. 

## How this PR fixes it
Implement all the necessary screens

## How to test it
Open a safe with all those transactions in the history and navigate to them. They should look like in the design.

## Screenshots
grafik
grafik
grafik

## Checklist

---

## CLA signature


apps/mobile/src/features/HistoryTransactionDetails/components/history-views/HistoryCreation.tsx
```diff
@@ -0,0 +1,56 @@
...
+        submittedAt={createdAt}
+        badgeThemeName="badge_background"
+      />
+
+      <View>
+        <YStack gap="$4" marginTop="$8">
+          <Container padding="$4" gap="$4" borderRadius="$3">
+            <View alignItems="center" flexDirection="row" justifyContent="space-between">
+              <Text color="$textSecondaryLight">Creator</Text>
+              <HashDisplay value={txInfo.creator.value as Address} />
```
katspaugh: Why do you need to cast these? Wrong types in the CGW SDK?

apps/mobile/src/features/HistoryTransactionDetails/components/shared/ThresholdChangeDisplay.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <View alignItems="center" flexDirection="row" justifyContent="space-between">
+      <Text color="$textSecondaryLight">Threshold change</Text>
+      <View flexDirection="row" alignItems="center" gap="$2">
+        <Text fontSize="$4">
+          {txInfo.settingsInfo?.threshold}/{executionInfo.signers.length}
+        </Text>
+        <Text textDecorationLine="line-through" color="$textSecondaryLight" fontSize="$4">
+          {executionInfo.confirmationsRequired}/{executionInfo.signers.length}
```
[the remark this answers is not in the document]
compojoom: I think that everything is correct:
on mobile:
grafik
on web:
grafik

the striked through is the old one. And the non-striked text is the the threshold valid after the tx is confirmed.

apps/mobile/src/features/HistoryTransactionDetails/utils/transformTransaction.ts
```diff
@@ -0,0 +1,33 @@
...
+ * that can be used by the standard transaction detail components.
+ */
+export function transformTransactionToDetails(txData: string, txId: string): TransactionDetails | null {
+  if (!txData) {
+    return null
+  }
+
+  try {
+    const tx = JSON.parse(txData) as Transaction
+    if (isCreationTransactionId(txId)) {
```
katspaugh: The transformation would work for any tx, wouldn't it? You're not using any creation-specific values, it seems.

apps/mobile/src/features/TxHistory/components/TxHistoryList/TxHistoryList.tsx
```diff
@@ -160,10 +160,13 @@ export function TxHistoryList({
 
   const onHistoryTransactionPress = useCallback(
     (transaction: TxCardPress) => {
+      const isCreationTx = isCreationTransactionId(transaction.tx.id)
+
       router.push({
         pathname: '/history-transaction-details',
         params: {
           txId: transaction.tx.id,
+          ...(isCreationTx && { txData: JSON.stringify(transaction.tx) }),
```
katspaugh: Why is the tx passed in params? I would make a new request on the details view instead.

apps/mobile/src/utils/transaction-guards.ts
```diff
@@ -184,6 +184,10 @@ export const isCreationTxInfo = (value: TransactionInfo): value is CreationTrans
   return value.type === TransactionInfoType.CREATION
 }
 
+export const isCreationTransactionId = (txId: string): boolean => {
+  return txId.startsWith('creation_')
```
katspaugh: That's not a very reliable way of detecting a tx type. It's not documented by the service.

apps/mobile/src/features/HistoryTransactionDetails/components/history-views/HistoryCreation.tsx
compojoom: Actually, I don't, but I like to remind myself that it is an address. What CGW returns is a string and we've manually defined addreess as `0x${string}`

apps/mobile/src/features/HistoryTransactionDetails/utils/transformTransaction.ts
compojoom: Yes, but I don't want it to work for any transaction other than the creationTransaction. Returning "null" was may way of telling my feature self - this is only supposed to be used with creation txs.

apps/mobile/src/features/TxHistory/components/TxHistoryList/TxHistoryList.tsx
compojoom: That's the problem. I can't make a new request on the details view page. What should I request? Fetch the history again? If I do this - how do I know on which page the creation tx is? The creation tx is a fake transaction, returned by cgw. If you try to fetch this tx by ID you get a 404.

katspaugh: I see. Can you add a comment about this above that line?

apps/mobile/src/utils/transaction-guards.ts
compojoom: yea, true. I'm falling back to the isCreationTxInfo now

apps/mobile/src/features/TxHistory/components/TxHistoryList/TxHistoryList.tsx
compojoom: Please check new code. As we decided to drop this view we no longer pass any parameters anymore.

katspaugh (approved):
Sauber!

PR #6290, opened 2025-09-02
chores(linter): add prettier to lint command

vseehausen:
## What it solves

The lint command did not include prettier. This led to PRs introducing prettier errors to the repository.

## How this PR fixes it

* Adds prettier to lint command (mainly for checking at CI level)
* Resolves prettier errors in the repo

## Checklist

---

## CLA signature


katspaugh:
@vseehausen could you do the following:

* Reset this branch to dev
* Disable formatting rules in eslint as they are deprecated (see [this](https://eslint.org/blog/2023/10/deprecating-formatting-rules/))
* Add a new command `prettier:check` (do NOT change the behavior of the existing prettier command)
* Add it to the CI workflows separately, not as part of the lint command
* Do it for both mobile and web
* Do NOT run prettier on the entire codebase, we can do it in a separate PR

Thanks for tackling this!

compojoom:
@vseehausen - I would also do the following 

- make sure that the same prettier version is installed in every package(you can add prettier to this file: https://github.com/safe-global/safe-wallet-monorepo/blob/mobile/feat/earn-decoding/yarn.config.cjs#L0-L1 and the script will warn us if prettier versions diverge across the packages in the repo)
- I would also try to standardise the prettier commands across all packages - `yarn prettier` should do a check, `yarn prettier:fix` should fix the files in the package

What I still don't understand how it was possible for those non-formatted files to end up in the repo, because lint-staged runs prettier by default on staged files 🤷

vseehausen:
Confirmation that the CI command works: https://github.com/safe-global/safe-wallet-monorepo/actions/runs/17439441584/job/49518278068?pr=6290

katspaugh:
Do you think it would make sense to add the prettier command to the existing lint workflows instead of creating separate workflows just for this?

The Mobile Prettier job seems to be failing.

vseehausen:
I think it make sense to have separate jobs. It's expected that the mobile one fails, as there are actual formatting errors now.

PR #6291, opened 2025-09-02
fix(web): Show signer name on confirmation step 

LucieFaire:
fix(web): Display signer name on confirmation step

## What it solves [COR-62](https://linear.app/safe-global/issue/COR-62/manage-signersreplace-signer-entered-name-for-a-new-signer-is-not)

Resolves #

* Display the provided signer name on confirmation step

## How this PR fixes it
* Reads the `form data` from `ManageSigners` edit step
* Enhances the `decoded transaction data` with the signer names
* Displays `address` and `name` on confirmation step

## How to test it
* Go to `Settings` -> manage signers
* Try adding or removing signers
* Observe confirmation view: names should be present

## Screenshots
Screenshot 2025-09-02 at 11 09 29

## Checklist

---

## CLA signature


apps/web/src/components/tx/confirmation-views/ManageSigners/index.tsx
```diff
@@ -41,14 +55,17 @@ export function ManageSigners({
...
-    .filter((owner) => safe.owners.every(({ value }) => value !== owner))
-    .map((addedOwner) => ({ value: addedOwner }))
+    .filter((owner) => safe.owners.every(({ value }) => value !== owner.value))
+    .map((addedOwner) => ({
+      value: addedOwner.value,
+      name: addedOwner.name ?? undefined,
+    }))
   const removedOwners = safe.owners
-    .filter((owner) => !newOwners.includes(owner.value))
+    .filter((owner) => !newOwners.some((newOwner) => newOwner.value === owner.value))
```
tmjssz: Consider using `sameAddress` from the utils package here to make sure it works with checksummed and non-checksummed addresses

apps/web/src/components/tx/confirmation-views/ManageSigners/get-new-safe-setup.ts
```diff
@@ -58,6 +63,11 @@ export function getNewSafeSetup({
     }
   }
 
+  const newOwners: Array<AddressInfo> = ownerAddresses.map((address) => ({
+    value: address,
+    name: signerNames[address] || undefined,
```
tmjssz, line 68: I think we also should normalize the address key for signer name lookup. E.g. use `checksumAddress` from utils on the keys when creating signerNames object and also use it when getting a name from the object for a given address.

apps/web/src/components/tx/confirmation-views/ManageSigners/index.tsx
LucieFaire: thank you for the suggestion, I have added the check
[e030e44](https://github.com/safe-global/safe-wallet-monorepo/pull/6291/commits/e030e44f2b8a4165727c42d6c6f045cce2641e3f)

apps/web/src/components/tx/confirmation-views/ManageSigners/get-new-safe-setup.ts
LucieFaire: I have updated the collection of `signerNames` to have checksumed addresses. I dont think we need to do it on `NewOwners` as it already contains checksumed addresses.
[e030e44](https://github.com/safe-global/safe-wallet-monorepo/pull/6291/commits/e030e44f2b8a4165727c42d6c6f045cce2641e3f)

tmjssz: That would only work if the owner addresses in the `ExtendedSafeInfo` object are already checksummed, I'm not sure if that's guaranteed.

After taking another look, I had a different idea: what if we set the address name directly in the safe object? In `ManageSigners`, we could do something like this before calling `getNewSafeSetup`:

```typescript
  safe.owners.forEach((owner) => {
    owner.name = data?.owners.find((o) => sameAddress(o.address, owner.value))?.name
  })
```

This way, the safe object would already contain the names, and we wouldn’t need to pass the extra signerNames prop at all. WDYT?

LucieFaire: I'm not sure it will work. As we are adding new owners - they would not be present in the `safe.owners` at this time, so we wont find any names unless we are renaming the existing owners. We populate the existing owners with new owners in `getNewSafeSetup` when we parse the transaction data.

apps/web/src/components/tx/confirmation-views/ManageSigners/get-new-safe-setup.ts
```diff
@@ -11,15 +12,17 @@ export function getNewSafeSetup({
...
   txData: TransactionDetails['txData']
   safe: ExtendedSafeInfo
+  signerNames?: Record<string, string>
 }): {
-  newOwners: Array<string>
+  newOwners: Array<AddressInfo>
   newThreshold: number
 } {
-  let newOwners = safe.owners.map((owner) => owner.value)
+  let ownerAddresses = safe.owners.map((owner) => owner.value)
```
tmjssz: I think we should normalize here as well, just in case.
```suggestion
  let ownerAddresses = safe.owners.map((owner) => checksumAddress(owner.value))
```

apps/web/src/components/tx/confirmation-views/ManageSigners/get-new-safe-setup.ts
tmjssz: Ah ok, I see what you mean. Then let's leave it like it is and only normalize ownerAddresses -> added a separate [comment](#discussion_r2318865076) for that.

tmjssz (approved):

LucieFaire:
@tmjssz I have updated the logic of extraction the names: [0bf59e0](https://github.com/safe-global/safe-wallet-monorepo/pull/6291/commits/0bf59e0b9f794e734449ceda86c5a7d519a2b558)
it is not the most elegant solution, and ideally I would want these flows to have the same data format, but refactoring that will be a lot of work at the moment, so I would for this solution for now.

apps/web/src/components/tx/confirmation-views/ManageSigners/index.tsx
```diff
@@ -74,27 +107,28 @@ function Actions({ newOwners }: { newOwners: Array<string> }): ReactElement | nu
...
 }
 
-function Signers({ owners }: { owners: Array<string> }): ReactElement {
+function Signers({ owners }: { owners: Array<AddressInfo> }): ReactElement {
   return (
     <FieldsGrid title="Signers">
       <Box display="flex" flexDirection="column" gap={2} padding="var(--space-2)" fontSize="14px">
-        {owners.map((owner) => (
+        {owners.map(({ value, name }) => (
           <EthHashInfo
```
katspaugh: Could we change this to a NamedAddressInfo? It has reverse ENS resolution.

tmjssz (approved):

apps/web/src/components/tx/confirmation-views/ManageSigners/index.tsx
LucieFaire: [85ce678](https://github.com/safe-global/safe-wallet-monorepo/pull/6291/commits/85ce678cac34ef4f7748ab8b64a664d2b1833b96)

PR #6292, opened 2025-09-02
refactor(mobile): change header for history txs

compojoom:
## What it solves
@TanyaEfremova wanted to change the designs of the history tx details. Instead of having the tx type in the header of the page, now every page has the same title "Transaction details" and we've moved the tx type right below the Identicon|Token or Logo. I've also added a share button in the header.

Resolves #

## How to test it
Currently all txs in the history should have a "transaction details" header with a share button. The type of the tx should be presented below the Identicon|TokenIcon or Logo. 

## Screenshots
grafik

## Checklist

---

## CLA signature


apps/mobile/src/features/HistoryTransactionDetails/components/history-views/CancelTx.tsx
```diff
@@ -27,7 +25,7 @@ export function CancelTx({ txId, txInfo, executionInfo, executedAt }: CancelTxPr
       }
       badgeIcon="transaction-contract"
       badgeColor="$textSecondaryLight"
-      title={txInfo.methodName ?? 'On-chain rejection'}
+      transactionType={txInfo.methodName ?? 'On-chain rejection'}
```
usame-algan, line 28: nit: when it was called title it was fine to pass a string for On-chain rejection but calling it transactionType I would expect we only pass an enum or a const

apps/mobile/src/hooks/useShareTransaction.ts
```diff
@@ -0,0 +1,35 @@
+import { useCallback } from 'react'
+import Share from 'react-native-share'
+import { useAppSelector } from '@/src/store/hooks'
+import { useDefinedActiveSafe } from '@/src/store/hooks/activeSafe'
+import { selectChainById } from '@/src/store/chains'
+import { SAFE_WEB_TRANSACTIONS_URL } from '@/src/config/constants'
+
+export function useShareTransaction(txId: string) {
+  const activeSafe = useDefinedActiveSafe()
+  const chain = useAppSelector((state) => selectChainById(state, activeSafe.chainId))
```
usame-algan, line 10: Not in the scope of this PR but what do you think about using the `chainId` from `activeSafe` and have a mapping `chainId -> shortName`?

usame-algan: We already have this in place on web inside the `SendTransactionButton` but IMO it would be worth to have this mapping in general. I can create a task for it in Linear.

usame-algan: https://linear.app/safe-global/issue/COR-554/add-chains-mapping-for-chainid-shortname

PR #6294, opened 2025-09-02
fix(assets): fix horizontal scroll in assets table

katspaugh:
## What it solves

The assets table was scrolling horizontally due to an added column (DeFi positions).

* Reduce the width of columns to add up to 100%
* Add `overflowX: hidden`

### DIFF
--- a/apps/web/src/components/balances/AssetsTable/index.tsx
+++ b/apps/web/src/components/balances/AssetsTable/index.tsx
@@ -93,24 +93,24 @@ const headCells = [
   {
     id: 'asset',
     label: 'Asset',
-    width: '30%',
+    width: '28%',
   },
   {
     id: 'price',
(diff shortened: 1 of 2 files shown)


vseehausen:
Thanks. Not sure about reducing the overall percentage to less than 100%. Also there should be an alignment between the asset table and the position table (one column should be aligned).

vseehausen:
Ok but we cannot align those columns, so lgtm.

vseehausen (approved):

francovenica:
The permanent scrollbar is gone, which is good, but is an issue: Small screens lost their ability to scroll horizontally.

This his how it currently works in prod:
horiz scroll

francovenica:
Fixed. Approved

vseehausen:
lgtm

PR #6298, opened 2025-09-02
feat(mobile): add HistoryContract screen

compojoom:
## What it solves
Ads a screen that displays contract data

Resolves https://linear.app/safe-global/issue/COR-525/contract-mintapprove-etc

## How this PR fixes it
Ads the HistoryContact screen & refactored the HashDisplay component. We should display a user friendly name if we have one. HashDisplay now takes care of this. It tries to use a name from the addressbook or from CGW (returned as AddressInfo obj). If we don't have any of those, we fallback to displaying the shortened address.

## How to test it
Open contracts interactions in the history

## Screenshots
grafik
grafik

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/components/HashDisplay/HashDisplay.tsx
+++ b/apps/mobile/src/components/HashDisplay/HashDisplay.tsx
@@ -2,80 +2,111 @@ import React from 'react'
 import { View, Text } from 'tamagui'
 import { TouchableOpacity } from 'react-native'
 import { Identicon } from '@/src/components/Identicon'
-import { EthAddress } from '@/src/components/EthAddress'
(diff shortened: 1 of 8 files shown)


katspaugh:
Looks like you're using a hook for display name that is very similar to the one from the web app. Is it possible to reuse some logic? Maybe extract it into a pure TS service package?

compojoom:
> Looks like you're using a hook for display name that is very similar to the one from the web app. Is it possible to reuse some logic? Maybe extract it into a pure TS service package?

Web is way more complicated as it is trying to also fetch the contact from the space address book. We currently don't support spaces on mobile so let's keep them separate.

PR #6307, opened 2025-09-03
fix(dashboard): don't show 0 balance before balances are loaded

katspaugh:
## What it solves

There was an inconsistent initial state when loading the dashboard. Balances aren't in a _loading_ state initially (`loading: false`) , yet there was initial placeholder data with `items: []`.

Screenshot 2025-09-03 at 09 11 31

## How it solves it

I've added a new piece of state for all "loadable" slices like Balances and Safe Info to indicate whether the data has been loaded or not. The lifecycle of loadable slices is now like this:

1. Initial state:
```
{
  data: { ... }, // placeholder data,
  loaded: false,
  loading: false, // hasn't started loading yet
  error: undefined,
}
```

2. Loading state:
```
{
  data: { ... }, // placeholder data,
  loaded: false,
  loading: true,
  error: undefined,
}
```

3. Finished loading:
```
{
  data: { ... }, // actual data
  loaded: true,
  loading: false,
  error: undefined,
}
```

3. Error state:
```
{
  data: { ... }, // placeholder
  loaded: false,
  loading: false,
  error: 'Some error',
}
```

### DIFF
--- a/apps/web/src/components/dashboard/Overview/Overview.tsx
+++ b/apps/web/src/components/dashboard/Overview/Overview.tsx
@@ -21,7 +21,7 @@ import OverviewSkeleton from './OverviewSkeleton'
 
 const Overview = (): ReactElement => {
   const { safe, safeLoading, safeLoaded } = useSafeInfo()
-  const { balances, loading: balancesLoading } = useVisibleBalances()
+  const { balances, loaded: balancesLoaded, loading: balancesLoading } = useVisibleBalances()
   const { setTxFlow } = useContext(TxModalContext)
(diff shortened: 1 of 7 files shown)


vseehausen:
This is a great fix!

The variable name COULD be improved, but no blocker. `isInitiated` or `isLoaded`, but `loaded` works for now.

vseehausen (approved):

katspaugh:
Yes, `isLoaded` is better but I was going off of the existing `loading` property, so kept it in this format. Thanks for the review!

PR #6308, opened 2025-09-03
Refactor(positions): Dynamically import positions

usame-algan:
## What it solves

Part of positions feature

## How this PR fixes it

- Dynamically imports positions and `PositionsWidget`
- Moves UI from positions page into positions component
- Extracts `PositionsSkeleton`

## How to test it

1. Open positions and dashboard
2. Everything should still be as it was

## Checklist

---

## CLA signature


katspaugh (approved):
Thanks!

katspaugh:
Tested. The only downside is that the dashboard widget now appears after a short delay (after the chain configs and the chunk are loaded). So it's only on the initial page load.

PR #6314, opened 2025-09-03
feat(web): show ens names at addresses

vseehausen:
## What it solves

Shows ENS names and avatars in the address field when applicable.

## How this PR fixes it

Loads ENS name and avatar. The avatar loads slowly.

## How to test it

Open a Safe with ENS owners (i.e., [eth:0x91c32893216dE3eA0a55ABb9851f581d4503d39b](https://feat_resolveensnames--walletweb.review.5afe.dev/home?safe=eth:0x91c32893216dE3eA0a55ABb9851f581d4503d39b)), observe addresses.

## Checklist

- [X] I've written a unit/e2e test for it (if applicable) 🧑‍💻

---

## CLA signature

### DIFF
--- a/apps/web/src/components/common/AddressInput/index.test.tsx
+++ b/apps/web/src/components/common/AddressInput/index.test.tsx
@@ -21,6 +21,7 @@ const mockChain = chainBuilder()
 jest.mock('@/hooks/useChains', () => ({
   useCurrentChain: jest.fn(() => mockChain),
   useChain: jest.fn(() => mockChain),
+  useHasFeature: jest.fn(() => true),
 }))
 
 // mock useNameResolver
--- a/apps/web/src/components/common/EthHashInfo/index.tsx
+++ b/apps/web/src/components/common/EthHashInfo/index.tsx
@@ -19,7 +19,7 @@ const EthHashInfo = ({
   const addressBookItem = useAddressBookItem(props.address, chain?.chainId)
   const { ens, avatar: ensAvatar } = useAddressResolver(props.address)
   const link = chain && props.hasExplorer ? getBlockExplorerLink(chain, props.address) : undefined
-  const name = showName ? props.name || addressBookItem?.name || ens : undefined
+  const name = showName ? addressBookItem?.name || props.name || ens : undefined
   const customAvatar = props.customAvatar || ensAvatar
 
   return (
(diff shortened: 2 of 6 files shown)


apps/web/src/components/common/AddressInput/index.test.tsx
```diff
@@ -21,6 +21,7 @@ const mockChain = chainBuilder()
 jest.mock('@/hooks/useChains', () => ({
   useCurrentChain: jest.fn(() => mockChain),
   useChain: jest.fn(() => mockChain),
+  useHasFeature: jest.fn(() => true),
```
vseehausen, line 24: This had to be added to several test files to make them pass.

apps/web/src/hooks/useAddressResolver.ts
```diff
@@ -17,17 +17,24 @@ export const useAddressResolver = (address: string) => {
 
   const [ens, _, isResolving] = useAsync<string | undefined>(() => {
     if (!shouldResolve) return
+
     return lookupAddress(ethersProvider, debouncedValue)
   }, [ethersProvider, debouncedValue, shouldResolve])
 
-  const resolving = shouldResolve && isResolving
+  const [avatar, __, isResolvingAvatar] = useAsync<string | undefined>(() => {
```
vseehausen, line 24: It's been working better to add the ens resolver to this file.

katspaugh:
Thanks for the PR, @vseehausen!

I have a few questions.

1. Why are you working on this? 😃 There's no ticket on Linear for this AFAIU, and we discussed on Slack that feedback has been given in #6281, so we're just waiting for the external contributor to update that PR. No action point was agreed on from your side.

2. Have you considered that resolving ENS in _every_ `EthHashInfo` might result in _many_ RPC requests and slow down the app?

3. Have you seen that we already have a component called `NamedAddressInfo` that is essentially `EthHashInfo` + ENS lookup? It was made this way, i.e. two separate components, to be conscious and deliberate about places where doing an ENS lookup is a good trade off of performance and UX. I'm open to reconsider this approach and do look-ups everywhere but this should be discussed and put on the roadmap, as it may require moving reverse ENS resolution to the backend.

vseehausen:
Hey, thanks for your questions!

1. I have been working on this last Friday night, as I found the idea good. After the discussion about the RPC provider, I wanted to see if the solution actually works.
2. No I haven't. 
3. Yes I tried resolving from `NamedAddressInfo`, but couldn't make it work.

I would actually ask, why wasn't there a Linear ticket, when it's a good idea, and it's pretty simple to implement, and Safe user's actually want it?

vseehausen:
closed for now, as deprioritized.

PR #6315, opened 2025-09-04
Feat(Mobile): Add transaction execution

usame-algan:
## What it solves

Resolves [COR-585](https://linear.app/safe-global/issue/COR-585/add-execution-flow)

## ToDos

- [x] Make sure only next nonce can be executed

## How this PR fixes it

- Shows a Continue button if a pending transactions is fully signed
- Adds a new `ReviewAndExecute` screen
- Adds a new `SelectExecutor` sheet and selector
- Adds `ExecutionSuccess` and `ExecutionError`

## How to test it

1. Open a safe with pending transactions
2. Go to a transaction that is fully signed
3. Observe an enabled Continue button
4. Press the button
5. Observe a review screen
6. Select a signer to execute the transaction with
7. Press the execute button
8. Observe a progress screen

## Screenshots
Screenshot 2025-09-04 at 14 35 30
Screenshot 2025-09-04 at 14 35 38
Screenshot 2025-09-04 at 14 35 44
Screenshot 2025-09-04 at 14 35 51
Screenshot 2025-09-05 at 13 10 48

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/ExecuteTx/components/CanNotExecute/CanNotExecute.tsx
+++ b/apps/mobile/src/features/ExecuteTx/components/CanNotExecute/CanNotExecute.tsx
@@ -1,11 +1,11 @@
 import React from 'react'
 import { Text, YStack } from 'tamagui'
 
-export function CanNotSign() {
+export function CanNotExecute() {
   return (
     <YStack gap="$4" padding="$8" alignItems="center" justifyContent="center" testID="can-not-sign-container">
       <Text fontSize="$4" fontWeight={400} width="100%" textAlign="center" color="$textSecondaryLight">
-        Only signers of this safe can sign this transaction
+        Only signers of this Safe can execute this transaction
       </Text>
     </YStack>
   )
--- a/apps/mobile/src/features/ChangeSignerSheet/ChangeSignerSheet.container.tsx
+++ b/apps/mobile/src/features/ChangeSignerSheet/ChangeSignerSheet.container.tsx
@@ -27,7 +27,7 @@ export const ChangeSignerSheetContainer = () => {
   const { items, loading } = useAvailableSigners(txId, actionType)
 
   const onSignerPress = (signer: SignerInfo, onClose: () => void) => () => {
-    if (activeSigner.value !== signer.value) {
+    if (activeSigner?.value !== signer.value) {
       setTxSigner(signer)
     }
 
--- a/apps/mobile/src/features/ConfirmTx/components/ConfirmTxForm/ConfirmTxForm.test.tsx
+++ b/apps/mobile/src/features/ConfirmTx/components/ConfirmTxForm/ConfirmTxForm.test.tsx
@@ -8,12 +8,14 @@ import { CanNotSign } from '../CanNotSign'
 import { ExecuteForm } from '../ExecuteForm'
 import { SignForm } from '../SignForm'
 import { useTransactionSigner } from '../../hooks/useTransactionSigner'
+import { CanNotExecute } from '@/src/features/ExecuteTx/components/CanNotExecute'
 
 // Mock the hooks and components
 jest.mock('@/src/store/hooks/activeSafe')
 jest.mock('../../hooks/useTransactionSigner')
 jest.mock('../confirmation-views/AlreadySigned')
 jest.mock('../CanNotSign')
+jest.mock('@/src/features/ExecuteTx/components/CanNotExecute')
 jest.mock('../ExecuteForm')
 jest.mock('../SignForm')
 
@@ -50,6 +52,11 @@ describe('ConfirmTxForm', () => {
     ;(CanNotSign as jest.Mock).mockReturnValue(
       <View>
         <Text>CanNotSign</Text>
+      </View>,
+    )
+    ;(CanNotExecute as jest.Mock).mockReturnValue(
+      <View>
+        <Text>CanNotExecute</Text>
       </View>,
     )
     ;(ExecuteForm as jest.Mock).mockReturnValue(
@@ -115,23 +122,13 @@ describe('ConfirmTxForm', () => {
     )
   })
 
-  it('renders null when no conditions are met', () => {
+  it('renders CanNotExecute when no active signer', () => {
     ;(useTransactionSigner as jest.Mock).mockReturnValue({
       signerState: { ...mockSignerState, activeSigner: undefined },
     })
 
-    const { toJSON } = render(<ConfirmTxForm {...defaultProps} isExpired={true} />)
+    const { getByText } = render(<ConfirmTxForm {...defaultProps} isExpired={true} />)
 
-    expect(toJSON()).toBeNull()
-  })
-
-  it('handles undefined activeSigner in CanNotSign', () => {
-    ;(useTransactionSigner as jest.Mock).mockReturnValue({
-      signerState: { ...mockSignerState, canSign: false, activeSigner: undefined },
-    })
-
-    const { getByText } = render(<ConfirmTxForm {...defaultProps} />)
-
-    expect(getByText('CanNotSign')).toBeTruthy()
+    expect(getByText('CanNotExecute')).toBeTruthy()
   })
 })
--- a/apps/mobile/src/features/ConfirmTx/components/ConfirmTxForm/ConfirmTxForm.tsx
+++ b/apps/mobile/src/features/ConfirmTx/components/ConfirmTxForm/ConfirmTxForm.tsx
@@ -4,6 +4,7 @@ import { ExecuteForm } from '../ExecuteForm'
 import { AlreadySigned } from '../confirmation-views/AlreadySigned'
 import { CanNotSign } from '../CanNotSign'
 import { useTransactionSigner } from '../../hooks/useTransactionSigner'
+import { CanNotExecute } from '@/src/features/ExecuteTx/components/CanNotExecute'
 
 interface ConfirmTxFormProps {
   hasEnoughConfirmations: boolean
@@ -14,6 +15,10 @@ interface ConfirmTxFormProps {
 export function ConfirmTxForm({ hasEnoughConfirmations, isExpired, txId }: ConfirmTxFormProps) {
   const { signerState } = useTransactionSigner(txId)
   const { activeSigner, hasSigned, canSign } = signerState
+
+  if (!activeSigner) {
+    return <CanNotExecute />
+  }
 
   if (hasEnoughConfirmations) {
     return <ExecuteForm txId={txId} />
--- a/apps/mobile/src/features/ExecuteTx/components/CanNotExecute/index.ts
+++ b/apps/mobile/src/features/ExecuteTx/components/CanNotExecute/index.ts
@@ -1,0 +1,1 @@
+export { CanNotExecute } from './CanNotExecute'
--- a/apps/mobile/src/store/activeSignerSlice.ts
+++ b/apps/mobile/src/store/activeSignerSlice.ts
@@ -1,6 +1,7 @@
 import { createSelector, createSlice, PayloadAction } from '@reduxjs/toolkit'
 import { RootState } from '.'
 import { Address, SignerInfo } from '../types/address'
+import { removeSigner } from '@/src/store/signersSlice'
 
 type ActiveSignerState = Record<Address, SignerInfo>
 
@@ -20,13 +21,23 @@ const activeSignerSlice = createSlice({
       return rest
     },
   },
+  extraReducers: (builder) => {
+    builder.addCase(removeSigner, (state, action) => {
+      for (const [safeAddress, signerInfo] of Object.entries(state) as [Address, SignerInfo][]) {
+        if (signerInfo.value === action.payload) {
+          // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+          delete state[safeAddress]
+        }
+      }
+    })
+  },
 })
 
 export const { setActiveSigner, removeActiveSigner } = activeSignerSlice.actions
 
 export const selectActiveSigner = createSelector(
   [(state: RootState) => state.activeSigner, (_state: RootState, safeAddress: Address) => safeAddress],
-  (activeSigner, safeAddress: Address) => activeSigner[safeAddress],
+  (activeSigner, safeAddress: Address): SignerInfo | undefined => activeSigner[safeAddress],
 )
 
 export default activeSignerSlice.reducer


apps/mobile/src/features/ChangeExecutorSheet/ChangeExecutorSheet.container.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const items = useMemo(() => {
+    if (!data) {
+      return []
+    }
+
+    return storedSigners.map((item) => ({
+      ...item,
+      balance: data[item.value],
+    }))
+  }, [data, storedSigners, detailedExecutionInfo])
```
compojoom: I know that this is nearly the same block as in on the signerSheet. Can we abstract both to separate hooks maybe?

apps/mobile/src/features/ExecuteTx/components/ExecuteSuccess.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { router } from 'expo-router'
+import { useDispatch } from 'react-redux'
+import { useSafeAreaInsets } from 'react-native-safe-area-context'
+import { AbsoluteLinearGradient } from '@/src/components/LinearGradient'
+
+export function ExecuteSuccess() {
+  const dispatch = useDispatch()
+  const { bottom } = useSafeAreaInsets()
+  const handleDonePress = () => {
+    dispatch(cgwApi.util.invalidateTags(['transactions']))
```
compojoom, line 18: why are we invalidating the trnsactions tag first on pressing done? Shouldn't we do this once we know the tx is executed? What if the user closes the screen with a gesture?

apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/ReviewAndExecuteView.tsx
```diff
@@ -0,0 +1,58 @@
...
+      >
+        <Tabs.Tab name="Data" label="Data">
+          <DataTab />
+        </Tabs.Tab>
+        <Tabs.Tab name="JSON" label="JSON">
+          <JSONTab txDetails={txDetails} />
+        </Tabs.Tab>
+      </Tabs.Container>
+
+      <ReviewExecuteFooter txId={txId} />
```
compojoom: Isn't this whole file basically the same as the confirm view except for the `<ReviewExecuteFooter txId={txId} />`

apps/mobile/src/services/tx/tx-sender/execute.ts
```diff
@@ -0,0 +1,47 @@
+import { ChainInfo } from '@safe-global/safe-gateway-typescript-sdk'
```
compojoom: We should not use ChainInfo from the typescript-sdk, but rather from @safe-global/store

apps/mobile/src/features/ChangeExecutorSheet/ChangeExecutorSheet.container.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { RootState } from '@/src/store'
+import { SignersCard } from '@/src/components/transactions-list/Card/SignersCard'
+import { Text, View } from 'tamagui'
+import { Address } from 'blo'
+import { SignerInfo } from '@/src/types/address'
+import { selectActiveSigner } from '@/src/store/activeSignerSlice'
+import { selectSigners } from '@/src/store/signersSlice'
+import { useGetBalancesQuery } from '@/src/store/signersBalance'
+import { selectChainById } from '@/src/store/chains'
+import { ChainInfo } from '@safe-global/safe-gateway-typescript-sdk'
```
compojoom: we don't use this package in the mobile app

apps/mobile/src/features/ExecuteTx/hooks/useTransactionExecution.ts
```diff
@@ -0,0 +1,51 @@
+import { useCallback, useState } from 'react'
+import { useDefinedActiveSafe } from '@/src/store/hooks/activeSafe'
+import { useAppSelector } from '@/src/store/hooks'
+import { selectChainById } from '@/src/store/chains'
+import type { RootState } from '@/src/store'
+import type { ChainInfo } from '@safe-global/safe-gateway-typescript-sdk'
```
compojoom: import it from @safe-global/store

apps/mobile/src/services/tx/tx-sender/execute.ts
```diff
@@ -0,0 +1,47 @@
...
+import type { SafeInfo } from '@/src/types/address'
+import { createConnectedWallet } from '../../web3'
```
compojoom: I know that I'm also not very consistent here, but can we use the full import with `@....`

apps/mobile/src/services/tx/tx-sender/execute.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  Object.entries(signatures).forEach(([signer, data]) => {
+    safeTx.addSignature({
+      signer,
+      data,
+      staticPart: () => data,
+      dynamicPart: () => '',
+      isContractSignature: false,
+    })
+  })
```
compojoom, line 42: Why do we need this block?

apps/mobile/src/features/ExecuteTx/components/ExecuteSuccess.tsx
usame-algan: Right now we set the status to `success` once the transaction is executed and display this screen so the user can only press Done after the transaction has been executed. Tbh this is not working good enough right now because after execution it can still take some time until our backend updates and actually adds the transaction to the history and removes it from the pending transaction list.

apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/ReviewAndExecuteView.tsx
usame-algan: Pretty much and we even reuse some of the confirm view components. The only real difference is that execute needs to navigate to a different view and display a different text for the button and use a different signer selector.

apps/mobile/src/services/tx/tx-sender/execute.ts
usame-algan: We could change it but its like this for the signing as well. IMO its better if we go with `Chain` from @safe-global/store and cast to `ChainInfo` as far down in the tree as possible. I can refactor this if we agree.

apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/ReviewAndExecuteView.tsx
compojoom: Can't we abstract it to a shared container that would only accept the Footer?

apps/mobile/src/features/ExecuteTx/hooks/useTransactionExecution.ts
usame-algan: As discussed I replaced all usages of ChainInfo on mobile.

apps/mobile/src/services/tx/tx-sender/execute.ts
usame-algan: Changed everywhere

apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/ReviewAndExecuteView.tsx
usame-algan: Done. Maybe we can abstract part of the parent where we display loading and error since that is also the same. Will look more into this tomorrow.

apps/mobile/src/services/tx/tx-sender/execute.ts
usame-algan: Because we use `proposeTx` which creates a new safe transaction based on the `txParams` but this still needs all signatures to be added.

apps/mobile/src/features/ExecuteTx/components/ExecuteSuccess.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const dispatch = useDispatch()
+  const { bottom } = useSafeAreaInsets()
+
+  const handleDonePress = () => {
+    dispatch(cgwApi.util.invalidateTags(['transactions']))
+
+    // Go back to pending transactions
+    router.back()
+    router.back()
+    router.back()
```
katspaugh: This seems brittle? Can't it go to a specific route instead?

apps/mobile/src/features/ExecuteTx/hooks/useTransactionExecution.ts
```diff
@@ -0,0 +1,51 @@
...
+  signerAddress: string
+}
+
+export function useTransactionExecution({ txId, signerAddress }: UseTransactionExecutionProps) {
+  const [status, setStatus] = useState<ExecutionStatus>('idle')
+  const activeSafe = useDefinedActiveSafe()
+  const activeChain = useAppSelector((state: RootState) => selectChainById(state, activeSafe.chainId))
+
+  const execute = useCallback(async () => {
+    setStatus('loading')
```
katspaugh: Should we use an enum here instead of a string?

apps/mobile/src/features/ExecuteTx/hooks/useTransactionExecution.ts
```diff
@@ -0,0 +1,51 @@
...
+        chain: activeChain,
+        activeSafe,
+        txId,
+        privateKey,
+      })
+      setStatus('success')
+    } catch (error) {
+      logger.error('Error executing transaction:', error)
+      setStatus('error')
+    }
```
katspaugh: I would separate these two await calls into two try-catch blocks to have two distinct errors.

apps/mobile/src/features/ExecuteTx/hooks/useTransactionExecution.ts
```diff
@@ -0,0 +1,51 @@
...
+      await executeTx({
+        chain: activeChain,
+        activeSafe,
+        txId,
+        privateKey,
+      })
+      setStatus('success')
+    } catch (error) {
+      logger.error('Error executing transaction:', error)
+      setStatus('error')
```
katspaugh: Should the error also be available in the object returned from the hook?

apps/mobile/src/services/tx/tx-sender/execute.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const { hash } = await protocolKit.executeTransaction(safeTx)
```
katspaugh, line 44: What else does the receipt contain other than a hash? Anything potentially useful?

apps/mobile/src/features/ChangeExecutorSheet/ChangeExecutorSheet.container.tsx
usame-algan: Deleted

apps/mobile/src/features/ExecuteTx/components/ExecuteSuccess.tsx
usame-algan: We can use `router.navigate` instead. This will change in a future PR though so for now I will navigate the user to the history after execution.

apps/mobile/src/features/ExecuteTx/hooks/useTransactionExecution.ts
usame-algan: Right now we are displaying a generic error `There was an error executing the transaction.` which is ok IMO.

apps/mobile/src/services/tx/tx-sender/execute.ts
usame-algan: In this PR we are not using the hash but in a follow-up we will store the pending transaction and poll it which we will need the hash for. Other than the hash it also returns the `transactionOptions` e.g. gasLimit etc. and the `transactionResponse` which I think can be awaited like we do on web.

usame-algan: Screenshot 2025-09-05 at 11 50 27

apps/mobile/src/features/ExecuteTx/components/ExecuteSuccess.tsx
usame-algan: As discussed we will add proper refetch for the history depending on the txHistoryTag in a follow-up.

apps/mobile/src/components/SelectExecutor/SelectExecutor.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import { Text, View, Image } from 'tamagui'
+import Signature from '@/assets/images/signature.png'
+
+import { Identicon } from '@/src/components/Identicon'
+import { SafeFontIcon } from '@/src/components/SafeFontIcon'
+import { router } from 'expo-router'
+import { ContactDisplayNameContainer } from '@/src/features/AddressBook'
+import { Address } from '@/src/types/address'
+import { ActionType } from '@/src/features/ChangeSignerSheet/ChangeSignerSheet.container'
```
compojoom: nit: importing the ActionType from the container seems weird. Shouldn't it be in a utils or something?

apps/mobile/src/features/ConfirmTx/components/ExecuteForm/ExecuteForm.tsx
```diff
@@ -1,40 +1,53 @@
...
+  const txNonce = isMultisigDetailedExecutionInfo(txData?.detailedExecutionInfo)
+    ? txData?.detailedExecutionInfo.nonce
+    : undefined
 
-  return (
-    <YStack justifyContent="center" gap="$4" alignItems="center" paddingHorizontal={'$4'}>
-      <Text fontSize="$4" fontWeight={400} width="70%" textAlign="center" color="$textSecondaryLight">
-        Transactions can be executed only in the web app at the moment.
-      </Text>
+  const isNext = txNonce !== undefined && txNonce === safe.nonce
```
compojoom: can we move this to a isNextNonce(..) function?

apps/mobile/src/features/ExecuteTx/components/ExecuteTransaction.tsx
```diff
@@ -0,0 +1,51 @@
...
+  }, [canExecute, status, executeTx, txId, activeSigner])
+
+  if (!txId) {
+    const handleRetry = () => {
+      console.error('Cannot retry: missing transaction ID')
+    }
+    return <ExecuteError description="Missing transaction ID" onRetryPress={handleRetry} />
+  }
+
+  if (!activeSigner) {
```
compojoom, line 35: how did the user end up with no active signer?

usame-algan: Good question. We do the same inside `SignTransaction`. How can it happen there? According to the type it is always defined so I would be in favour of removing this block.

apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/ReviewAndExecuteView.tsx
usame-algan: After looking at it again I think we can remove some duplication but it would require us to push query params but since this is a part that is still in refinement lets leave it as is in case we need to make changes to one or the other view.

apps/mobile/src/features/ExecuteTx/components/ExecuteTransaction.tsx
usame-algan: As discussed we should clean up the activeSigner when the private key is removed. I will also update the type to reflect that activeSigner can be undefined and we can display a message to the user in that case.

apps/mobile/src/features/ExecuteTx/components/CanNotExecute/CanNotExecute.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { Text, YStack } from 'tamagui'
+
+export function CanNotExecute() {
+  return (
+    <YStack gap="$4" padding="$8" alignItems="center" justifyContent="center" testID="can-not-sign-container">
+      <Text fontSize="$4" fontWeight={400} width="100%" textAlign="center" color="$textSecondaryLight">
+        Only signers of this safe can execute this transaction
```
katspaugh:
```suggestion
        Only signers of this Safe can execute this transaction
```

compojoom (approved):
👍

PR #6316, opened 2025-09-04
Fix(assets): consistent fiat value format in assets and positions

katspaugh:
## What it solves

Resolves [SEV-2](https://linear.app/safe-global/issue/GRO-57/defi-positions-wrong-rounding-for-the-amounts-that-is-easy-to-see)

Fiat value formatting wasn't consistent across assets and defi positions.

## How this PR fixes it

Fiat values are now always abbreviated and w/o cents unless they are in headings.

## Screenshots
Screenshot 2025-09-04 at 15 55 37

Screenshot 2025-09-04 at 15 55 50

Precise amount with cents is displayed on hover:

Screenshot 2025-09-04 at 15 56 07

### DIFF
--- a/apps/web/src/components/balances/AssetsTable/styles.module.css
+++ b/apps/web/src/components/balances/AssetsTable/styles.module.css
@@ -7,8 +7,8 @@
 
 .container td:last-of-type > * {
   position: absolute;
-  left: 0;
-  width: 100%;
+  right: 0;
+  width: 15%;
   opacity: 0;
   transition: opacity 0.2s;
   height: 100% !important;
--- a/apps/web/src/components/common/EnhancedTable/index.tsx
+++ b/apps/web/src/components/common/EnhancedTable/index.tsx
@@ -184,7 +184,6 @@ function EnhancedTable({ rows, headCells, mobileVariant, compact }: EnhancedTabl
                     <TableCell
                       key={key}
                       className={classNames({
-                        sticky: cell.sticky,
                         [css.collapsedCell]: row.collapsed,
                       })}
                     >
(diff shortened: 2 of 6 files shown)


apps/web/src/components/balances/AssetsTable/styles.module.css
```diff
@@ -7,8 +7,8 @@
 
 .container td:last-of-type > * {
   position: absolute;
-  left: 0;
-  width: 100%;
```
katspaugh, line 11: This was occluding the hover on fiat value.

apps/web/src/components/common/EnhancedTable/index.tsx
```diff
@@ -184,7 +184,6 @@ function EnhancedTable({ rows, headCells, mobileVariant, compact }: EnhancedTabl
                     <TableCell
                       key={key}
                       className={classNames({
-                        sticky: cell.sticky,
```
katspaugh, line 187: Unused global class.

apps/web/src/services/analytics/useMixpanel.ts
```diff
@@ -59,7 +59,11 @@ const useMixpanel = () => {
         console.info('[MixPanel] - User opted in')
       }
     } else {
-      mixpanel.opt_out_tracking()
+      try {
+        mixpanel.opt_out_tracking()
+      } catch (e) {
+        // do nothing, opt_out_tracking throws an error if tracking was never enabled
```
katspaugh: I had to add this because otherwise the app doesn't work at all unless you add a Mixpanel env var (which most devs initially wouldn't do).

vseehausen:
I think we wanted to show the decimals all the time, actually.

vseehausen (approved):

vseehausen:
Ah ok we don't.

vseehausen:
still lgtm

PR #6319, opened 2025-09-05
fix(mobile): swap filled state, created time, advanced tx details

compojoom:
## What it solves
- partially filled limit swap orders had different state than what was shown on web
- created time was not shown in the tx details on mobile
- don't use a custom "advanced details" view in history, but reuse the existing view form confirmTx

## How to test it
- open partially filled expired swap order and observe that the state of the order is now "partially filled" instead of "expired"

- check that created time is visible in the tx details in the history

- observe that the "transaction details" button leads to the same view we use for confirm txs.

## Screenshots
grafik
grafik

## Checklist

---

## CLA signature


apps/mobile/src/features/HistoryTransactionDetails/components/HistoryAdvancedDetailsButton/HistoryAdvancedDetailsButton.tsx
```diff
@@ -1,36 +1,10 @@
...
-        fontWeight="700"
-        size="$4"
-        fullscreen
-        onPress={goToAdvancedDetails}
-      >
-        View transaction data
-      </Button>
-    </View>
-  )
+  return <ParametersButton txId={txId} />
```
usame-algan, line 9: This will now display `Transaction details` instead of `View transaction data`. Is it intended?

compojoom: Yes, for now. @TanyaEfremova  wanted it this way. We'll probably change it with something else as it looks weird.

PR #6320, opened 2025-09-05
Refactor(web): call useAddressResolver in NamedAddressInfo

katspaugh:
## What it solves

A small follow up on #6281 – reuse an existing useAddressResolver hook in NamedAddressInfo.

Plus, a small fix for Mixpanel on dev.

### DIFF
--- a/apps/web/src/components/common/NamedAddressInfo/index.tsx
+++ b/apps/web/src/components/common/NamedAddressInfo/index.tsx
@@ -7,8 +7,7 @@ import useSafeAddress from '@/hooks/useSafeAddress'
 import { sameAddress } from '@safe-global/utils/utils/addresses'
 import { memo, useMemo } from 'react'
 import { isAddress } from 'ethers'
-import { lookupAddress } from '@/services/ens'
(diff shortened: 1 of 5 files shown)


apps/web/src/components/common/NamedAddressInfo/index.tsx
```diff
@@ -26,25 +24,22 @@ export function useAddressName(address?: string, name?: string | null, customAva
...
+  const nonEnsName = displayName || contract?.displayName || contract?.name
+
+  const { ens: ensName } = useAddressResolver(nonEnsName ? undefined : address)
 
   const isUnverifiedContract = useIsUnverifiedContract(contract)
 
   return useMemo(
     () => ({
       name:
-        displayName ||
```
vseehausen, line 39: Ah you just pushed this now. Honestly I like the list of priorities. Without the `nonEnsName` var.

katspaugh: I added it to avoid code duplication.

PR #6330, opened 2025-09-08
chore(web): run tsc noEmit as separate CI workflow

katspaugh:
## What it solves

Run TypeScript in no-emit mode as a separate workflow to catch type errors faster.


apps/web/package.json
```diff
@@ -8,7 +8,8 @@
     "dev": "next dev",
     "start": "next dev",
     "build": "next build",
-    "lint": "tsc && next lint",
+    "type-check": "tsc --noEmit",
+    "lint": "next lint",
```
usame-algan, line 11: I run this command often locally. Will it still work without the `tsc` step?

katspaugh: It will still work, yes. It was just combining these two unnecessarily.

usame-algan (approved):
For running the lint workflow we use a custom action `CatChen/eslint-suggestion-action@v4.1.18`. Does this action simply run the `yarn lint` command or do we also need to update that?

katspaugh:
We don't need to update the lint workflow.

PR #6334, opened 2025-09-08
docs: add prettier and type-check to AGENTS.md

katspaugh:
## What it solves

Avoid type issues after Codex PRs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,7 +16,7 @@ This repository uses a Yarn-based monorepo structure. Follow these rules when pr
 ## Workflow
 
 1. **Install dependencies**: `yarn install` (from the repository root).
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -16,18 +16,25 @@ This repository uses a Yarn-based monorepo structure. Follow these rules when pr
 ## Workflow
 
 1. **Install dependencies**: `yarn install` (from the repository root).
-2. **Formatting**: run `yarn prettier:fix` before committing.
-3. **Linting and tests**: when you change any source code under `apps/` or `packages/`, execute:
+2. **Formatting**: run `yarn workspace @safe-global/web prettier:fix` before committing.
```
vseehausen: I would use the global prettier. At least on my machine, the workspace prettier command uses a different prettier version, which will lead to inconsistencies.

PR #6337, opened 2025-09-08
fix(mobile): advanced details crashes when no txData

compojoom:
## What it solves
Some transactions such as “sent” events due to a swap don’t have txData. For those on web we don’t display any advanced details. Now we do the same. I also added a check to formatTxDetails for txData so that it won’t crash again if we stumble on a tx without any txData.

Resolves https://linear.app/safe-global/issue/COR-517/sent-screen

## How to test it
check this tx: https://safe-wallet-web.dev.5afe.dev/transactions/tx?safe=eth:0x8675B754342754A30A2AeF474D114d8460bca19b&id=transfer_0x8675B754342754A30A2AeF474D114d8460bca19b_eeb34c7a34951dc29cb5eea90a4820810502b5bb21c48b69f162a71cca500745d320

There should be no "transaction details" button anymore. 

## Screenshots
grafik

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/HistoryTransactionDetails/components/history-views/HistoryTokenTransfer.tsx
+++ b/apps/mobile/src/features/HistoryTransactionDetails/components/history-views/HistoryTokenTransfer.tsx
@@ -2,9 +2,8 @@ import React from 'react'
 import { Container } from '@/src/components/Container'
 import { View, YStack, Text } from 'tamagui'
 import { HistoryTransactionHeader } from '@/src/features/HistoryTransactionDetails/components/HistoryTransactionHeader'
-import { TransferTransactionInfo } from '@safe-global/store/gateway/AUTO_GENERATED/transactions'
+import { TransactionData, TransferTransactionInfo } from '@safe-global/store/gateway/AUTO_GENERATED/transactions'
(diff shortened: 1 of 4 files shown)


apps/mobile/src/features/HistoryTransactionDetails/components/history-views/HistoryTokenTransfer.tsx
```diff
@@ -57,12 +57,12 @@ export function HistoryTokenTransfer({ txId, txInfo }: HistoryTokenTransferProps
...
               <Text color="$textSecondaryLight">{fieldLabel}</Text>
 
-              <HashDisplay value={address as Address} />
+              <HashDisplay value={address} />
             </View>
 
             <NetworkDisplay />
 
-            {isOutgoing && <HistoryAdvancedDetailsButton txId={txId} />}
+            {isOutgoing && txData !== null && <HistoryAdvancedDetailsButton txId={txId} />}
```
katspaugh, line 65: Nit: you don't need a strict comparison here as txData is either an object or null.

```suggestion
            {isOutgoing && txData && <HistoryAdvancedDetailsButton txId={txId} />}
```

PR #6348, opened 2025-09-11
feat(Mobile): Add pending tx state

usame-algan:
## What it solves

Resolves [COR-577](https://linear.app/safe-global/issue/COR-577/add-pending-state)

## How this PR fixes it

- Adds a pending tx slice to mobile
- Adds a middleware on this slice to poll pending transactions onchain and the CGW once indexing
- Extracts `SimpleTxWatcher` into the shared utils package
- Creates a `SafeTxWatcher` service

## How to test it

1. Create a transaction on web
2. Go to the mobile app and execute that transaction
3. Observe a processing screen
4. Observe being navigated to the transaction after it has executed

## Screenshots
Simulator Screenshot - iPhone 16 Pro - 2025-09-11 at 10 20 29

https://github.com/user-attachments/assets/667e01a8-0e45-4c04-834b-8b644c20fd53

## Checklist

---

## CLA signature


apps/mobile/src/components/StatusBanners/PendingTransactions/PendingTransactions.tsx
```diff
@@ -14,7 +15,7 @@ interface Props {
 
 export const PendingTransactions = ({ number, isLoading, fullWidth, onPress }: Props) => {
   const startIcon = isLoading ? (
-    <Loader size={24} color="$warning1ContrastTextDark" />
+    <Loader size={24} color={getTokenValue('$color.warning1ContrastTextDark')} />
```
usame-algan, line 18: This fixes a warning that kept popping up

apps/mobile/src/features/ExecuteTx/hooks/useTransactionExecution.ts
```diff
@@ -41,13 +45,25 @@ export function useTransactionExecution({ txId, signerAddress }: UseTransactionE
...
+
+      dispatch(
+        addPendingTx({
+          txId,
+          chainId: activeChain.chainId,
+          safeAddress: activeSafe.address,
+          txHash: hash,
+          walletAddress: signerAddress,
+          walletNonce: await getUserNonce(activeChain, signerAddress),
+        }),
```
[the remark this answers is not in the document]
usame-algan: Good point. I will check if the nonce needs updating.

apps/mobile/src/features/ConfirmTx/components/PendingTxInfo/PendingTxInfo.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { Text, View } from 'tamagui'
+import { Container } from '@/src/components/Container'
+import { formatWithSchema } from '@/src/utils/date'
+import { HashDisplay } from '@/src/components/HashDisplay'
+import { Badge } from '@/src/components/Badge'
+import { PendingTx } from '@/src/store/pendingTxsSlice'
+
+export const PendingTxInfo = ({ createdAt, pendingTx }: { createdAt: number | null; pendingTx: PendingTx }) => {
```
compojoom, line 9: This is now the same block that we have in the tx history details? Maybe we should call this component somethign else and pass createdAt, hash and status? And then reuse it?

apps/mobile/src/features/ConfirmTx/hooks/useTransactionSigner.ts
```diff
@@ -14,7 +14,7 @@ export const useTransactionSigner = (txId: string) => {
     txDetails,
     detailedExecutionInfo,
     signerState,
-    isLoading: isFetching,
+    isLoading: !!isLoading,
```
compojoom, line 17: why the !! ?

usame-algan: A test was failing because somehow it was undefined. I guess its because on first mount its undefined but the type says its boolean which is misleading. This was a quick fix but open to other suggestions as well.

apps/mobile/src/store/middleware/pendingTxs.ts
```diff
@@ -0,0 +1,115 @@
...
+
+      runWatcher(listenerApi, txHash, chainId, walletAddress, walletNonce, txId)
+    },
+  })
+
+  startListening({
+    predicate: (action) => isHydrateAction(action),
+    effect: (action, listenerApi) => {
+      runWatchers(listenerApi, action.payload.pendingTxs)
+    },
```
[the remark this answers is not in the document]
usame-algan: The initialState of the pendingTxs slice is an empty object so this should also guard against this.

apps/mobile/src/features/ConfirmTx/components/PendingTxInfo/PendingTxInfo.tsx
usame-algan: I think some parts can be extracted like the createdAt block but I am not sure if we should extract everything into one shared component as this will couple pending state with the history and could introduce unexpected issues in the future. I would rather have a bit of duplication but more separation in this case.

PR #6351, opened 2025-09-12
chore: improve mobile commit detection in release-notes script (#6347)

LucieFaire:
[fix(web): Show signer name on confirmation step](https://github.com/safe-global/safe-wallet-monorepo/pull/6351/commits/38697031b69f245ea4d484a0093ab9c66938adc3)


usame-algan:
I didn't see this PR but just merged something for mobile into dev and now its part of the diff here. Should we cherry pick instead? cc @tmjssz

LucieFaire:
> I didn't see this PR but just merged something for mobile into dev and now its part of the diff here. Should we cherry pick instead? cc @tmjssz

yes we will, not going to merge this one

PR #6353, opened 2025-09-12
fix(Mobile): Poll safe info and update history

usame-algan:
## What it solves

Resolves [MOB-117](https://linear.app/safe-global/issue/MOB-117/mobile-update-safe-info-and-queuehistory-fetch-logic)

## How this PR fixes it

- Adds a `polling_interval` to fetching the safe info
- Adds the `txHistoryTag` as a dependency to the history query
- Adds a `router.dismissAll` when executing a transaction

## How to test it

1. Open the app
2. Execute a transaction
3. Execute another one
4. Observe no error message regarding the nonce
5. Observe the history updates after a transaction has been executed

## Checklist

---

## CLA signature


usame-algan:
@compojoom I was thinking about adding the same to the pending tx query but noticed we are polling there. What do you think about removing the polling and instead only updating the queue when the `txQueuedTag` changes?

PR #6354, opened 2025-09-12
chore: bump protocol kit

dasanra:
## What it solves

Bump Safe deployments fixing issue with Plasma chain short name


katspaugh:
FYI we're releasing 1.70.0 next Tuesday.

PR #6360, opened 2025-09-22
refactor: extract gas estimation utilities

usame-algan:
## What this solves

Extracts gas estimation logic from web to shared utils package so it can be reused for the mobile app

## Summary
- move gas price estimation logic into shared helpers and reuse it from the web hook
- add shared gas limit utilities (including encodeSignatures) and consume them from the web hook
- update web imports to reference the shared gas helpers and keep fee formatting working across features

## How to test

1. Open the web app
2. Create and execute a transaction
3. Gas estimation should still work


compojoom:
we implemented gas estimation in a different PR.

PR #6367, opened 2025-09-22
fix(assets): show `--` at value when price is not available

vseehausen:
## What it solves

Resolves failing e2e test: https://linear.app/safe-global/issue/GRO-67/0dollar-is-displayed-instead-of-andna-when-the-price-on-goingeco-is

## How this PR fixes it

Reuses `FiatBalance` instead of `FiatValue`, consistent over dashboard and balance asset table.

It also shows `--` for price, when price is unavailable.

## Checklist

---

## CLA signature


katspaugh (approved):

katspaugh:
Tested with a Safe with lots of spam tokens, it works. ✅ 

Screenshot 2025-09-25 at 11 41 27

PR #6375, opened 2025-09-24
test(web): map store alias for jest

katspaugh:
## Summary
- map the @safe-global/store alias in the web Jest config so the Safe Overviews tests can mock the RTK query client

## Testing
- yarn workspace @safe-global/web test --runTestsByPath src/store/__tests__/safeOverviews.test.ts

------
https://chatgpt.com/codex/tasks/task_b_68d3eaf83fb0832f92d44dfcb922116e


katspaugh:
I've tested it, Safe Overviews are still loaded, lazy loading/infinite scroll still works. ✅

PR #6376, opened 2025-09-24
chore(mobile): upgrade expo to v54

compojoom:
## What it solves
On iOS 26 the app was hanging in a lot of situations in dev mode. Upgrading to expo 54 and all other related dependencies fixes the problem. 

~~partially~~ resolves https://linear.app/safe-global/issue/COR-613/ios-26-ui-issues-in-app
~~The expo router header still needs some adjustments, but this will be done in a separate PR~~
I was able to fix the header issues as well. I've disabled liquid glass in the header for now. This is mostly a temporary solution and we will have to investigate how to implement our headers according to the new apple design. 

## How this PR fixes it
updates all dependencies

---

## CLA signature


apps/mobile/src/app/(tabs)/_layout.tsx
```diff
@@ -29,7 +29,7 @@ export default function TabLayout() {
             header: AssetsNavbar,
             title: 'Home',
             tabBarButtonTestID: 'home-tab',
-            tabBarButton: ({ children, ...rest }) => {
+            tabBarButton: ({ children, ref, ...rest }) => {
```
katspaugh, line 32: If a variable isn't used, we prefix the name with an underscore, e.g. _ref, in the web app. Curious why lint didn't complain here.

katspaugh (approved):

apps/mobile/src/app/(tabs)/_layout.tsx
compojoom: good point. It complained when I didn't have the ref as the type of the ref is slightly different. 🤷

PR #6379, opened 2025-09-25
chore(mobile): add permissions to workflows

compojoom:
## What it solves
Add permissions to the mobile github workflows.

---

## CLA signature

### DIFF
--- a/.github/workflows/mobile-lint.yml
+++ b/.github/workflows/mobile-lint.yml
@@ -11,6 +11,7 @@ on:
     paths:
       - apps/mobile/**
       - packages/**
+      - .github/workflows/mobile-lint.yml
 
 jobs:
   lint:
--- a/.github/workflows/mobile-prettier.yml
+++ b/.github/workflows/mobile-prettier.yml
@@ -13,6 +13,7 @@ on:
     paths:
       - apps/mobile/**
       - packages/**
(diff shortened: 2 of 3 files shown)


katspaugh:
Looks like it didn't trigger these jobs. Maybe you should add their own paths to the list so that they run when they themselves are changed.

.github/workflows/mobile-lint.yml
```diff
@@ -14,6 +14,11 @@ on:
 
 jobs:
   lint:
+    permissions:
+      checks: write
+      pull-requests: read
+      statuses: write
```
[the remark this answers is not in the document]
compojoom: 👍

yarn.lock
```diff
@@ -36359,7 +36359,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"ws@npm:8.18.3":
+"ws@npm:8.18.3, ws@npm:^8.18.3":
```
katspaugh: Strange that this file got changed.

katspaugh (approved):

yarn.lock
compojoom: yarn was broken on dev and fixed it here. I'll remove the commit and merge without it.

PR #6391, opened 2025-09-29
feat(WalletConnect) Add chain switch Safe selection modal

katspaugh:
Resolves https://linear.app/safe-global/issue/COR-463/switching-chain-in-a-dapp-connected-to-safe-causes-dapp-disconnection

## Summary
- show a WalletConnect-specific modal listing the user’s Safes on a requested network when a dapp triggers a chain switch
- switch WalletConnect sessions to the Safe selected in the modal and surface user cancellations back to the dapp
- cover the new flow with unit tests to ensure modal selection and cancellation behaviours

## Testing
- yarn workspace @safe-global/web test --runTestsByPath src/services/safe-wallet-provider/useSafeWalletProvider.test.tsx

------
https://chatgpt.com/codex/tasks/task_b_68da6039f56c832fb086ecec0f8b9c02

### DIFF
--- a/apps/web/src/services/safe-wallet-provider/useSafeWalletProvider.test.tsx
+++ b/apps/web/src/services/safe-wallet-provider/useSafeWalletProvider.test.tsx
@@ -385,8 +385,20 @@ describe('useSafeWalletProvider', () => {
         {
           chains: {
             data: [
-              { chainId: '1', shortName: 'eth', chainName: 'Ethereum' } as gateway.ChainInfo,
-              { chainId: '5', shortName: 'gor', chainName: 'Goerli' } as gateway.ChainInfo,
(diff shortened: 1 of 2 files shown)


katspaugh:
@compojoom found that some dapps like Pancake Swap still disconnect, so I'm taking it back to development.

katspaugh:
@compojoom I've tested with Pancake Swap and it looks like their wallet connector times out after 10 seconds, so you have to switch the chain in the Safe UI really fast. It works well when a Safe exists on the same address on another chain as the currently connected Safe. Then the switch is immediate and the dapp doesn't disconnect. tl;dr: it's on their side.

compojoom (approved):
👍

PR #6393, opened 2025-09-29
refactor(mobile): validate that signer is owner of a safe

compojoom:
## What it solves
When importing a PK we were validating that the signer that PK belonged to is actually a signer of a safe and we were preventing import when that was not the case. Now we do the same for ledger import.

Resolves https://linear.app/safe-global/issue/COR-538/mobile-implement-ledger-support#comment-7fcb846b

## How this PR fixes it
First I refactored the code we used to validate that an address is owner of a safe in the PK flow. Then I used the new hook in the ledger flow

## How to test it
- try to import with ledger and select an address that doens't belong to a safe (import should fail)
- try to import a wrong PK - it should still fail

## Checklist

---

## CLA signature


apps/mobile/src/hooks/useAddressOwnershipValidation.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Used by both private key and Ledger import flows
+ */
+export const useAddressOwnershipValidation = () => {
+  const glob = useGlobalSearchParams<{ safeAddress?: string; chainId?: string; import_safe?: string }>()
+  const chainIds = useAppSelector(selectAllChainsIds)
+  const activeSafe = useAppSelector(selectActiveSafe)
+  const currency = useAppSelector(selectCurrency)
+
+  const [singleSafeTrigger] = useLazySafesGetSafeV1Query({})
+  const [manySafesTrigger] = useLazySafesGetOverviewForManyQuery()
```
katspaugh: You could just do a get owned safes query.

compojoom: I think I stuck with it becuase we needed balances in other locations, but here owners make more sense. Good point.

apps/mobile/src/hooks/useAddressOwnershipValidation.ts
```diff
@@ -0,0 +1,98 @@
...
+        if (glob?.import_safe) {
+          // Multi-chain validation - check if address owns the specified safe on any chain
+          const ownedSafesResult = await ownerSafesTrigger({
+            ownerAddress: address,
+          }).unwrap()
+
+          // Check if the specified safe address is owned by this address on any chain
+          const isOwnerOfSafe = Object.entries(ownedSafesResult || {}).some(([_chainId, safeAddresses]) =>
+            safeAddresses.some((addr) => addr.toLowerCase() === safeAddress?.toLowerCase()),
+          )
```
katspaugh, line 54: You should probably only check on the current chain? I.e.
```
const isOwnerOfSafe = ownedSafesResult[chainId].some(add => sameAddress(add, safeAddress))
```

compojoom: No, you are at the import safe stage. This means you are importing a signer together with your safe. That also means that a signer on chain A, might not be there on chain B. That's why we get the info for all chains. So even if the signer is not available on sepolia, we are still going to import it becuase it is used on mainnet.

PR #6394, opened 2025-09-30
feat(mobile): automatically set the name of imported ledger address

compojoom:
## What it solves
Currently the imported ledger address would end up being "Singer-{last 4 digits of address}". With this PR the imported address would be stored as "{LEDGER device name} + {last 4 digits of address}"

Resolves https://linear.app/safe-global/issue/COR-538/mobile-implement-ledger-support#comment-34c3abb8

## How this PR fixes it
I had to refactor the addSignerWithEffects hook to accept the passed name instead of always setting one itself.  

## How to test it
Add a ledger signer and observe that the signer has a name that contains the Ledger device name. (note: if you've already imported this signer the name won't change as the contact exists in the address book)

## Checklist

---

## CLA signature


katspaugh (approved):
We should generally stick to "Normal case", not "Capital Nouns Case".

PR #6399, opened 2025-10-01
feat(web): enable upgrades for unsupported L2 master copies

katspaugh:
Resolves https://linear.app/safe-global/issue/REVE-391/opbnb-safes-are-not-usable-with-wallet-due-to-the-unofficial

## Summary

Implements bytecode matching to allow SDK initialization and transaction hash calculation for Safe accounts with unsupported base contracts that match known L2 singleton deployments (1.3.0+L2 and 1.4.1+L2).

## What works ✅

- Frontend detects when unsupported mastercopy bytecode matches official L2 contracts
- SDK initializes correctly with custom contract configuration  
- Migration transactions can be created successfully
- Transaction hashes (safeTxHash, domainHash, messageHash) are displayed in UI
- Users can see the "Migrate to L2" option for compatible unsupported mastercopies

## What doesn't work yet ❌ (requires CGW backend changes)

- **Cannot propose transactions to Client Gateway** (CGW returns 422: "Could not calculate safeTxHash")
- **Cannot execute migrations** because execution requires a txId from CGW proposal
- CGW needs to be updated to support bytecode matching or accept version parameter for unsupported mastercopies

This is fixed here: https://github.com/safe-global/safe-client-gateway/pull/2730

### Key Changes

1. **Bytecode comparison utilities** (`packages/utils/src/services/contracts/bytecodeComparison.ts`)
   - Compares runtime bytecode hash against known L2 deployments
   - Returns matched version (e.g., "1.3.0+L2")

2. **SDK initialization** (`apps/web/src/hooks/coreSDK/safeCoreSDK.ts`)
   - Detects bytecode-matched L2 contracts when mastercopy is unsupported
   - Initializes SDK with custom `contractNetworks` configuration
   - Fetches version from contract if gateway returns null

3. **Hash calculation** (`apps/web/src/components/transactions/TxDetails/Summary/SafeTxHashDataRow/index.tsx`)
   - Modified to use SDK version as fallback when gateway version is null
   - Enables safeTxHash display for unsupported mastercopies

4. **UI components**
   - `useBytecodeComparison` hook for checking bytecode matches
   - Updated migration warnings and error messages
   - `MigrateSafeL2Review` waits for SDK initialization

### Deployment utilities added

- `packages/utils/src/services/contracts/deployments.ts`:
  - `isL2MasterCopyCodeHash()` - Check if bytecode matches L2 deployment
  - `getL2MasterCopyVersionByCodeHash()` - Get version from bytecode hash

## Testing

Tested with: `opbnb:0x69158987610093f2A3156987E412fC0Bf40C6F64`
- ✅ SDK initializes with version "1.3.0"
- ✅ Transaction is created successfully
- ✅ safeTxHash is displayed in UI

### DIFF
--- a/apps/web/src/hooks/useSafeNotifications.ts
+++ b/apps/web/src/hooks/useSafeNotifications.ts
@@ -164,7 +164,14 @@ const useSafeNotifications = (): void => {
     return () => {
       dispatch(closeNotification({ id }))
     }
-  }, [dispatch, safe, safe.implementationVersionState, bytecodeComparison.result, bytecodeComparison.isLoading, query.safe])
+  }, [
+    dispatch,
+    safe,
+    safe.implementationVersionState,
+    bytecodeComparison.result,
+    bytecodeComparison.isLoading,
+    query.safe,
+  ])
 }
 
 export default useSafeNotifications


apps/web/src/hooks/useBytecodeComparison.ts
```diff
@@ -0,0 +1,98 @@
...
+    const fetchAndCompare = async () => {
+      // Only compare if mastercopy is unsupported
+      if (isValidMasterCopy(safe.implementationVersionState)) {
+        setComparisonResult(undefined)
+        setIsLoading(false)
+        return
+      }
+
+      // Need web3 provider to fetch bytecode
+      if (!web3ReadOnly) {
```
yagopv, line 39: This creates a perpetual loading state?

apps/web/src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -35,9 +43,43 @@ export const initSafeSDK = async ({
...
     const isL2SafeMasterCopy = isInDeployments(masterCopy, safeL2Deployment?.networkAddresses[chainId])
 
-    // Unknown deployment, which we do not want to support
     if (!isL1SafeSingleton && !isL2SafeMasterCopy) {
-      return Promise.resolve(undefined)
+      try {
+        const code = await provider.getCode(masterCopy)
+
+        if (!code || code === '0x') {
+          return Promise.resolve(undefined)
```
yagopv: The Promise.resolve(undefined)'s seems redundant

apps/web/src/hooks/coreSDK/safeCoreSDK.ts
```diff
@@ -35,9 +43,43 @@ export const initSafeSDK = async ({
...
+        contractNetworks = {
+          [chainId]: {
+            safeSingletonAddress: masterCopy,
+          },
+        }
+
+        safeVersion = upgradeableVersion
+        isL1SafeSingleton = false
+      } catch (error) {
+        return Promise.resolve(undefined)
```
yagopv: This fails silently. The logic is complicated but no error is propagated or logged

apps/web/src/features/multichain/components/UnsupportedMastercopyWarning/UnsupportedMasterCopyWarning.tsx
```diff
@@ -4,24 +4,41 @@ import ErrorMessage from '@/components/tx/ErrorMessage'
...
+} from '@safe-global/utils/services/contracts/safeContracts'
+import { useBytecodeComparison } from '@/hooks/useBytecodeComparison'
 
 export const UnsupportedMastercopyWarning = () => {
   const { safe } = useSafeInfo()
+  const bytecodeComparison = useBytecodeComparison()
 
-  const showWarning = !isValidMasterCopy(safe.implementationVersionState) && isMigrationToL2Possible(safe)
+  // Check if migration is possible based on bytecode comparison
+  const canMigrate =
```
yagopv, line 26: If ⁠bytecodeComparison.isLoading is ⁠true, ⁠bytecodeComparison.result might be ⁠undefined, leading to incorrect migration eligibility

apps/web/src/components/tx/SignOrExecuteForm/UnknownContractError.tsx
```diff
@@ -4,22 +4,29 @@ import ExternalLink from '@/components/common/ExternalLink'
...
+  const bytecodeComparison = useBytecodeComparison()
 
   const isMigrationTx = useMemo((): boolean => {
     return txData !== undefined && isMigrateL2SingletonCall(txData)
   }, [txData])
 
   // Unsupported base contract
   const isUnknown = !isValidMasterCopy(safe.implementationVersionState)
-  const isMigrationPossible = isMigrationToL2Possible(safe)
+  const isMigrationPossible =
```
yagopv, line 28: If it's still loading, ⁠result will be ⁠undefined, potentially showing incorrect migration possibility

yagopv (approved):

apps/web/src/hooks/useBytecodeComparison.ts
katspaugh: Why? web3ReadyOnly is initialized very quickly, it's just not synchronous.

yagopv: If the web3ReadyOnly fails to initialize for whatever reason and the effect is not triggered again i guess the loading state will be persistent as is being set to true

Perhaps the condition can live above the fetchAndCompare and just do nothing. Anyway you are already doing setIsLoading(true) inside the fetchAndCompare.

```
if (!web3ReadOnly) return

const fetchAndCompare = async () => {
...
```

yagopv: But the PR is approved on my end. I only commented on minor issues if you'd like to address them ;)

PR #6405, opened 2025-10-03
feat(web): Add Unstoppable Domains Resolution Support

V-Shadbolt:
## What it solves

Safe Wallet currently only supports ENS for domain name resolution and reverse resolution

## How this PR fixes it

Adds Unstoppable Domains Resolution and Reverse Resolution SDK alongside ENS with the goal of feature parity. Defaults to ENS and then fallsback to UD resolution to keep the existing UI flow as-is. Also changes ENS copy in the UI to be domain-name agnostic.

## How to test it

- Use an Unstoppable Domain (partner-engineering.crypto) in any Address Input field. 
- See wallet names resolve to ENS or UNS (example UNS address: 0x64F23b8c6261f8717E486955383679b9aEC67A56)

## Screenshots

Screenshot 2025-10-03 at 11 20 50 AM

safe wallet ud resolution

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -65,6 +65,7 @@
     "@safe-global/safe-modules-deployments": "^2.2.16",
     "@safe-global/store": "workspace:^",
     "@sentry/react": "^7.91.0",
+    "@unstoppabledomains/resolution": "^9.3.3",
     "@walletconnect/core": "^2.21.9",
     "@walletconnect/utils": "^2.21.9",
     "@web3-onboard/coinbase": "^2.4.2",
--- a/apps/web/src/components/common/AddressInput/index.test.tsx
+++ b/apps/web/src/components/common/AddressInput/index.test.tsx
@@ -92,7 +92,7 @@ describe('AddressInput tests', () => {
 
   beforeEach(() => {
     jest.clearAllMocks()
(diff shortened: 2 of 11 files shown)


katspaugh:
@V-Shadbolt thanks for the contribution! The code looks good, other than we'd have to introduce a new npm dependency.

@the-real-chrizzo @simabeh on the product side, do we want to support UD going forward? AFAIR we dropped UD support about 2 years ago due to concerns about their centralized governance model, and I haven't heard of any complaints since.

V-Shadbolt:
@katspaugh Glad the code looks good! I believe there's a good use case for Unstoppable Domains in SAFE wallet. Resolution is our primary use case and with well over 4 million domains we have a large user base as well. 

Is there anything in particular you'd want to see within this PR to help move it along?

V-Shadbolt:
Hi @katspaugh ! Is there anything else I can do here to move this PR along?

V-Shadbolt:
Hi @katspaugh ! Following up on this PR as we'd love to get it merged :)

simabeh:
Hi @V-Shadbolt - thanks for putting this together. Unfortunately we’re not going to merge this for now. 

The main reason is that we don’t currently see a clear user need that would justify adding and maintaining UD support

If this becomes more relevant we can revisit it later. Appreciate you sharing this with us

PR #6408, opened 2025-10-04
feat(mobile): import signers from seed-phrase

compojoom:
## What it solves
This PRs allows a user to import a signer from seedphrase. 

Resolves https://linear.app/safe-global/issue/COR-510/mobile-import-seed-phrase

## How this PR fixes it
- refactored the private key import feature (mainly renamed as private key and seed phrase import start from the same screens)
- added new designs for seed phrase import
- refactored the ledger hook for fetching addresses as the seed phrase import is similar. 

## How to test it
Get a seed phrase and try to import accounts from it. 

## Screenshots
grafik
grafik
grafik
grafik
grafik

## Checklist

---

## CLA signature


katspaugh (approved):

apps/mobile/src/features/ImportSigner/ImportSigner.container.tsx
```diff
@@ -27,20 +27,20 @@ export function ImportPrivateKey() {
...
             <SafeInput
               height={114}
-              value={privateKey}
-              onChangeText={handlePrivateKeyChange}
+              value={input}
+              onChangeText={handleInputChange}
               placeholder="Paste here or type..."
               secureTextEntry={isMasked}
-              success={!!wallet}
+              success={!!wallet || inputType === 'seed-phrase'}
```
clovisdasilvaneto, line 43: shouldn't it be ` inputType === 'seed-phrase' && input.trim() !== ''` instead? Otherwise if the input is empty and the import type is seed-phrase we will always show success in this field, or am I missing something?

apps/mobile/src/features/ImportSigner/SeedPhraseAddresses.container.tsx
```diff
@@ -0,0 +1,226 @@
...
+      )}
+    </View>
+  )
+
+  const renderListFooter = () => <LoadMoreButton onPress={() => deriveAddresses(10)} isLoading={isLoading} />
+
+  const renderEmptyState = () => <AddressesEmptyState />
+
+  return (
+    <View style={{ flex: 1 }}>
```
clovisdasilvaneto:
```suggestion
    <View flex={1}>
```

apps/mobile/src/features/ImportSigner/hooks/useImportPrivateKey.ts
```diff
@@ -0,0 +1,115 @@
...
+  const [inputType, setInputType] = useState<InputType>('unknown')
+  const [wallet, setWallet] = useState<ethers.Wallet>()
+  const local = useLocalSearchParams<{ safeAddress: string; chainId: string; import_safe: string }>()
+  const [error, setError] = useState<string | undefined>(undefined)
+  const router = useRouter()
+  const { createDelegate } = useDelegate()
+
+  const handleInputChange = (text: string) => {
+    setInput(text)
+    const detectedType = detectInputType(text)
```
clovisdasilvaneto, line 22: if we could add the detectedType into an enum would be nice, so we would avoid having these arbitrary strings comparisons spread into the code

apps/mobile/src/features/ImportSigner/SeedPhraseAddresses.container.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  const error = fetchError || importError
+  const clearError = () => {
+    clearFetchError()
+    clearImportError()
+  }
+
+  useEffect(() => {
+    if (addresses.length === 0 && !isLoading) {
+      void deriveAddresses(1)
```
clovisdasilvaneto, line 50: `void deriveAddresses`? That is the first time I see it 🤔

apps/mobile/src/features/ImportSigner/hooks/useSeedPhraseAddresses.ts
```diff
@@ -0,0 +1,103 @@
...
+    },
+    [seedPhrase],
+  )
+
+  const { addresses, isLoading, error, clearError, loadAddresses } = useAddresses({
+    fetchAddresses: deriveAddresses,
+    validateInput: validateSeedPhrase,
+  })
+
+  const deriveAddressesWrapper = useCallback(
```
clovisdasilvaneto, line 58: I"m wondering if `Wrapper` is a good name here. Because `wrapper` is usually refers to components that are used as main parent of its children

clovisdasilvaneto (approved):

apps/mobile/src/features/ImportSigner/ImportSigner.container.tsx
compojoom: wallet us undefined when we are dealing with a seed phrase

apps/mobile/src/features/ImportSigner/hooks/useImportPrivateKey.ts
compojoom: I don't undrestand why you guys are always so into enums 🤷 

The return type of detectInputType is 

export type InputType = 'private-key' | 'seed-phrase' | 'unknown'

I know that `if (detectedType === 'private-key')` looks like a string comparison and it is, but typescript will know if something is wrong. I also don't see the benefit of yet importing another enum on top just to do

`detectedType === detectInputTypes.privateKey` - to me this is way more work than just comparing against `private-key`? 

If we ever come to change `private-key` to `some-other-fancy-name-for-private-key`, then renaming with an IDE is still like 1s job 🤷 

Honest question. What is the benefit of enums?

clovisdasilvaneto: for me it is not the enum its self, but having to dealing with these arbitrary values everywhere. 🤔

PR #6409, opened 2025-10-06
Fix(Ledger): temporarily revert ledger package upgrades

katspaugh:
## What it solves

(Hopefully) resolves https://linear.app/safe-global/issue/EPD-339/not-possible-to-connect-with-ledger

## Solution

I've temporarily reverted the versions of Ledger packages to the same ones as on main.


compojoom (approved):

katspaugh:
Still not connecting:

Screenshot 2025-10-06 at 11 03 15

katspaugh:
Connects now. Tested with Nano X. ✅

PR #6413, opened 2025-10-06
Chore(Ledger): update packages and add origin token

katspaugh:
## What it solves

Ledger now requires an [origin token](https://developers.ledger.com/docs/device-interaction/integration/migrations/signers/eth/1_3_3_to_1_4_0#what-changed) but we don't use it so I've just left a template for it.


apps/web/src/services/onboard/ledger-module.ts
```diff
@@ -337,7 +337,7 @@ async function getLedgerSdk() {
   const dmk = new DeviceManagementKitBuilder().addTransport(webHidTransportFactory).build()
   const device = await lastValueFrom(dmk.startDiscovering({ transport: webHidIdentifier }))
   const sessionId = await dmk.connect({ device })
-  const signer = new SignerEthBuilder({ dmk, sessionId }).build()
+  const signer = new SignerEthBuilder({ dmk, sessionId, originToken: 'your-origin-token' }).build()
```
katspaugh, line 340: We leave this placeholder here because we don't use any features requiring a token.

PR #6427, opened 2025-10-09
feat(mobile): Add relay transaction execution support

clovisdasilvaneto:
## What it solves
User is not able to execute the transaction in Relay on mobile.
Resolves https://linear.app/safe-global/issue/COR-580/execution-with-gelatorelaying

## How this PR fixes it
To add Relay support, we had to integrate the relay execution and also extend our redux middleware which is responsable for watching the tx status.

## How to test it
- go to a safe with relay executions
- try to execute a transaction in the mobile app

## Screenshots
Screenshot 2025-10-08 at 14 53 47
Screenshot 2025-10-08 at 14 54 24

Screenshot 2025-10-08 at 14 54 34
## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/store/middleware/pendingTxs.ts
+++ b/apps/mobile/src/store/middleware/pendingTxs.ts
@@ -22,19 +22,13 @@ import logger from '@/src/utils/logger'
 
 const startRelayWatcher = (listenerApi: AppListenerEffectAPI, txId: string, taskId: string, chainId: string) => {
   RelayTxWatcher.getInstance()
-    .watchTaskId(taskId, (task) => {
-      // Update callback - called on each status update
-      logger.info('Relay status update', { txId, taskId, taskState: task.taskState, txHash: task.transactionHash })
-
-      // If we have a transaction hash, update the pending tx with it
-      if (task.transactionHash && task.transactionHash !== '') {
-        listenerApi.dispatch(setRelayTxHash({ txId, txHash: task.transactionHash }))
-      }
-    })
+    .watchTaskId(taskId)
     .then((task) => {
       // Transaction executed successfully, move to indexing
       logger.info('Relay transaction completed', { txId, taskId, txHash: task.transactionHash })
-      if (task.transactionHash) {
+
+      if (task.transactionHash && task.transactionHash !== '') {
+        listenerApi.dispatch(setRelayTxHash({ txId, txHash: task.transactionHash }))
         listenerApi.dispatch(setPendingTxStatus({ txId, chainId, status: PendingStatus.INDEXING }))
       }
     })


apps/mobile/src/features/ExecuteTx/hooks/useTransactionExecutionWithPK.ts
```diff
@@ -73,11 +88,108 @@ export function useTransactionExecutionWithPK({ txId, signerAddress, feeParams }
       logger.error('Error executing transaction:', error)
       setStatus(ExecutionStatus.ERROR)
     }
-  }, [activeChain, activeSafe, txId, signerAddress, dispatch])
+  }, [activeChain, activeSafe, txId, signerAddress, feeParams, dispatch])
```
clovisdasilvaneto: @compojoom I found that we were missing the `feeParams` in the hook dependencies of the executeTx function

compojoom: I remember that I had difficulties properly resolving conflicts when I was merging the ledger PR. So probably I created that problem 🤦 , sorry.

apps/mobile/src/features/ConfirmTx/components/PendingTxInfo/PendingTxInfo.tsx
```diff
@@ -18,7 +18,7 @@ export const PendingTxInfo = ({ createdAt, pendingTx }: { createdAt: number | nu
         </View>
       )}
 
-      {pendingTx?.txHash && (
+      {pendingTx?.type === PendingTxType.SINGLE && pendingTx.txHash && (
```
compojoom: this is a bit weird? Why do we have to check for txHash if the type of the pending tx is "single"? The single type should 100% have a txHash according to the type definition.

apps/mobile/src/features/ExecuteTx/components/EstimatedNetworkFee/EstimatedNetworkFee.tsx
```diff
@@ -1,17 +1,31 @@
...
+  executionMethod: ExecutionMethod
 }
 
-export const EstimatedNetworkFee = ({ totalFee, txId }: EstimatedNetworkFeeProps) => {
+export const EstimatedNetworkFee = ({ totalFee, txId, executionMethod }: EstimatedNetworkFeeProps) => {
   const chain = useAppSelector(selectActiveChain)
+  const { colorScheme } = useTheme()
+  const activeSafe = useDefinedActiveSafe()
+
+  const { data: relaysRemaining, isLoading: isLoadingRelays } = useRelayGetRelaysRemainingV1Query({
```
compojoom: You probably want to use currentData here:
data - The latest returned result regardless of hook arg, if present.
currentData - The latest returned result for the current hook arg, if present.

https://redux-toolkit.js.org/rtk-query/usage/queries#frequently-used-query-hook-return-values

apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/ReviewAndExecuteContainer.tsx
```diff
@@ -5,13 +5,23 @@ import { Text, View } from 'tamagui'
...
 
 export function ReviewAndExecuteContainer() {
   const { txId } = useLocalSearchParams<{ txId: string }>()
 
   const { data: txDetails, isFetching: isLoading, isError } = useTransactionData(txId || '')
 
+  const activeSafe = useDefinedActiveSafe()
+
+  // Check relay availability
+  const { data: relaysRemaining, isLoading: isLoadingRelays } = useRelayGetRelaysRemainingV1Query({
```
compojoom: Probably here as well.

apps/mobile/src/features/ExecuteTx/hooks/useTransactionExecutionWithPK.ts
clovisdasilvaneto: aah got it

apps/mobile/src/features/ExecuteTx/components/EstimatedNetworkFee/EstimatedNetworkFee.tsx
clovisdasilvaneto: interesting, I didn't know that

apps/mobile/src/features/ConfirmTx/components/PendingTxInfo/PendingTxInfo.tsx
clovisdasilvaneto: true

apps/mobile/src/store/middleware/pendingTxs.ts
```diff
@@ -1,14 +1,47 @@
...
+
+const startRelayWatcher = (listenerApi: AppListenerEffectAPI, txId: string, taskId: string, chainId: string) => {
+  RelayTxWatcher.getInstance()
+    .watchTaskId(taskId, (task) => {
+      // Update callback - called on each status update
+      logger.info('Relay status update', { txId, taskId, taskState: task.taskState, txHash: task.transactionHash })
+
+      // If we have a transaction hash, update the pending tx with it
+      if (task.transactionHash && task.transactionHash !== '') {
+        listenerApi.dispatch(setRelayTxHash({ txId, txHash: task.transactionHash }))
```
compojoom: why is this not triggering a lint error? I see that you are setting a txHash here, but the relay type doesn't have a txHash?

export type PendingRelayTx = {
  type: PendingTxType.RELAY
  taskId: string
  chainId: string
  safeAddress: string
  status: PendingStatus
}

apps/mobile/src/store/pendingTxsSlice.ts
```diff
@@ -7,7 +7,13 @@ export enum PendingStatus {
   SUCCESS = 'SUCCESS',
 }
 
-export type PendingTx = {
+export enum PendingTxType {
+  SINGLE = 'single',
```
compojoom: I don't like the single name here. How about something like "user_paid' | relayer, or 'self_submitted' | 'relayer_submitted'?

compojoom: or 'direct' | 'relayed' ?

packages/utils/src/services/RelayTxWatcher.ts
```diff
@@ -0,0 +1,170 @@
+export enum TaskState {
```
compojoom, line 1: nice!

apps/mobile/src/store/middleware/pendingTxs.ts
clovisdasilvaneto: it has, once the relayTxWatcher is finished we set the txHash to what it returned. here you have what it returns:

export type TransactionStatusResponse = {
  chainId: number
  taskId: string
  taskState: TaskState
  creationDate: string
  lastCheckDate?: string
  lastCheckMessage?: string
  transactionHash?: string
  blockNumber?: number
  executionDate?: string
}

clovisdasilvaneto: @compojoom I see the reason why you were confused. Thing is that it was not complaining bnecause in the setRelayTxHash function I was setting the txHash in the entry and casting it as PendingTx, I removed the casting and added the txHash in the relay type (as optional), because as I said, when the relay watcher returns the hash, we set it in the relayTx to make it indexable.

apps/mobile/src/store/pendingTxsSlice.ts
clovisdasilvaneto: @compojoom I'll use the enum that we already use for the execution method. wdyt?

apps/mobile/src/store/middleware/pendingTxs.ts
compojoom: casting is the root of all evil :D

apps/mobile/src/store/pendingTxsSlice.ts
compojoom: ah, yea. Good idea.

apps/mobile/src/services/tx-execution/privateKeyExecutor.ts
```diff
@@ -0,0 +1,66 @@
...
+  type: PendingTxType.SINGLE
+  txId: string
+  chainId: string
+  safeAddress: string
+  txHash: string
+  walletAddress: string
+  walletNonce: number
+}
+
+export const executeSingleTx = async ({
```
compojoom: I guess you forgot to change the name here

apps/mobile/src/services/tx-execution/relayExecutor.ts
```diff
@@ -0,0 +1,102 @@
...
+  // Add all signatures to the transaction
+  Object.entries(signatures).forEach(([signer, data]) => {
+    safeTx.addSignature({
+      signer,
+      data,
+      staticPart: () => data,
+      dynamicPart: () => '',
+      isContractSignature: false,
+    })
+  })
```
compojoom: I think that I added a helpet to add the signatures when I worked on the ledger integration

clovisdasilvaneto: Do you know the name of it?

compojoom: https://github.com/safe-global/safe-wallet-monorepo/blob/dev/apps/mobile/src/services/tx/tx-sender/create.ts#L37

apps/mobile/src/store/pendingTxsSlice.ts
```diff
@@ -45,14 +68,26 @@ export const pendingTxsSlice = createSlice({
...
+      const { txId, txHash } = action.payload
+      const tx = state[txId]
+
+      if (tx && tx.type === ExecutionMethod.WITH_RELAY) {
+        // Convert relay tx to single tx once we have the hash
+        state[txId] = {
+          ...tx,
+          txHash,
+        }
+      }
```
compojoom: this is weird. The comment is wrong, but the with_relay tx has the following props:
```
            txId: string
            type: ExecutionMethod.WITH_RELAY
            taskId: string
            chainId: string
            safeAddress: string
```
with PK
```
            txId: string
            type: ExecutionMethod.WITH_PK
            chainId: string
            safeAddress: string
            txHash: string
            walletAddress: string
            walletNonce: number
```

just by adding a txHash you are not ending up with the correct pendingtx type. This could blow in our face.
You are still having a with_relay tx that also now has txHash, but txHash is not specified on that type?

PR #6432, opened 2025-10-09
feat(mobile): read-only warning

compojoom:
## What it solves
Add an improved read-only warning. On click we open a bottom sheet and from there the user can navigate to the import signer screen

Resolves https://linear.app/safe-global/issue/COR-405/mobile-incentive-to-import-signer

## How this PR fixes it
Changes the previous "ready-only" warning, ads a new per safe settings slice.

## Screenshots
grafik
grafik

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/store/safesSettingsSlice.ts
+++ b/apps/mobile/src/store/safesSettingsSlice.ts
@@ -87,9 +87,6 @@ const safesSettingsSlice = createSlice({
       }
 
       const safeData = state[safeAddress]
-      if (!safeData) {
-        return
-      }
 
       if (!safeData.chains) {
         safeData.chains = {}


apps/mobile/src/store/safesSettingsSlice.ts
```diff
@@ -0,0 +1,174 @@
...
+      const { safeAddress } = action.payload
+
+      if (!state[safeAddress]) {
+        state[safeAddress] = {}
+      }
+
+      const safeData = state[safeAddress]
+      if (!safeData) {
+        return
+      }
```
clovisdasilvaneto: this sounds redundant to me, you already do this check !state[safeAddress] and set it as an empty object... Can you move the return to the first condition?

PR #6441, opened 2025-10-10
feat(web): Update the welcome page and the logo

PooyaRaki:
## What it solves
This PR refreshes the welcome page with updated design and layout.

### DIFF
--- a/apps/mobile/src/components/FiatChange/FiatChange.tsx
+++ b/apps/mobile/src/components/FiatChange/FiatChange.tsx
@@ -1,0 +1,53 @@
+import React from 'react'
+import { Text, View } from 'tamagui'
+import { type Balance } from '@safe-global/store/gateway/AUTO_GENERATED/balances'
+import { formatPercentage } from '@safe-global/utils/utils/formatters'
+
+interface FiatChangeProps {
+  balanceItem: Balance
+}
+
+export const FiatChange = ({ balanceItem }: FiatChangeProps) => {
+  if (!balanceItem.fiatBalance24hChange) {
+    return (
+      <Text fontSize="$3" color="$colorSecondary" opacity={0.7}>
+        0%
+      </Text>
+    )
+  }
+
+  const changeAsNumber = Number(balanceItem.fiatBalance24hChange) / 100
+  const changeLabel = formatPercentage(changeAsNumber)
+  const direction = changeAsNumber < 0 ? 'down' : changeAsNumber > 0 ? 'up' : 'none'
+
+  const getColor = () => {
+    switch (direction) {
+      case 'down':
+        return '$error'
+      case 'up':
+        return '$success'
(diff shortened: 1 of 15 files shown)


apps/web/src/components/welcome/WelcomeLogin/styles.module.css
```diff
@@ -6,14 +6,41 @@
...
+  --variant-textBg: rgba(18, 19, 18, 0.04);
+  --variant-outlinedBorder: #121312;
+  --variant-outlinedBg: rgba(18, 19, 18, 0.04);
+}
+
+.orDivider::after {
+  background-color: #dcdee0;
+}
+.orDivider::before {
+  background-color: #dcdee0;
```
katspaugh, line 45: Colors should be taken from CSS vars/theme (see vars.css).

apps/web/src/components/welcome/WelcomeLogin/WalletLogin.tsx
```diff
@@ -40,7 +42,15 @@ const WalletLogin = ({
...
   return (
-    <Button onClick={onConnectWallet} sx={{ minHeight: '42px' }} variant="contained" size="small" disableElevation>
+    <Button
+      onClick={onConnectWallet}
+      sx={{ minHeight: '42px' }}
+      variant="contained"
+      size="small"
+      disableElevation
+      fullWidth={fullWidth}
+      style={{ backgroundColor: '#000', color: '#fff' }}
```
katspaugh: I guess we're not using the regular primary CTA here because this page isn't responsive to the light/dark mode switch?

apps/web/src/components/welcome/WelcomeFooter.tsx
katspaugh, line 1: Why does this have to be a separate component? Can't we parametrize the existing footer component and re-use it here?

apps/web/src/components/welcome/WelcomeLogin/styles.module.css
PooyaRaki: This is intentional: the page supports light mode only.

apps/web/src/components/welcome/WelcomeLogin/WalletLogin.tsx
PooyaRaki: This is intentional: the page supports light mode only.

apps/web/src/components/welcome/WelcomeFooter.tsx
PooyaRaki: I initially created a new component since the new footer differs from the old one, but as per your request, I’ve adjusted the existing one instead.
https://github.com/safe-global/safe-wallet-monorepo/pull/6441/commits/df697ff342e0fdbf1c5bbeaebe7c80ff06f1cd60

PR #6442, opened 2025-10-10
fix(web): update company references and other privacy terms

PooyaRaki:
## What it solves
This PR updates the company references and privacy terms.


katspaugh:
@PooyaRaki `chore` is for things related to CI, tests, build etc. Not for boring things that seem like a literal chore. This should be a `fix`.

PooyaRaki:
@katspaugh
Chore is also used for:
- Helps distinguish between commits that impact production code (feat, fix) versus those that are just housekeeping.
- Makes it clear to reviewers and future contributors that the change is non-functional.

I'll change it to `fix` per your request.

PR #6446, opened 2025-10-10
feat(web): add Safe Labs terms acceptance flow

PooyaRaki:
## What it solves
This PR introduces a dedicated Safe Labs terms acceptance flow within the web app. If users have not yet accepted the terms, they are redirected to a new terms page and their wallet is temporarily disconnected. Once they accept, they are redirected back to their previous page and can continue using the app seamlessly.

## How to test it
- In order to see the transition screen again you need to remove `SAFE_v2__safe-labs-terms` from your localStorage.

### DIFF
--- a/apps/web/src/components/terms/styles.module.css
+++ b/apps/web/src/components/terms/styles.module.css
@@ -6,7 +6,7 @@
 }
 
 .contentWrapper {
-  padding-top: var(--space-5);
+  padding-top: calc(var(--header-height) + var(--space-5));
   padding-bottom: var(--space-5);
   flex: 1;
   max-width: 602px;
@@ -136,6 +136,7 @@
 
 .checkboxCardInner {
   padding: var(--space-2);
+  padding-left: 0;
   border-radius: 6px;
 }
 


apps/web/src/components/terms/safe-labs-terms.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+      const queryParams = {
+        ...query,
+        ...(isValidAutoConnect && autoConnect === 'true' ? { autoConnect: 'true' } : {}),
+      }
+
+      const queryString = new URLSearchParams(queryParams as Record<string, string>).toString()
+      const redirectUrl = queryString ? `${pathname}?${queryString}` : pathname
+
+      window.location.href = redirectUrl
```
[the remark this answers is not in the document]
PooyaRaki: The URL is strictly validated in `security.ts` to ensure users cannot be redirected to a malicious destination.

compojoom:
You are missing letter-spacing 0.17px on the text, because of this the text looks more cramped than on the figma design.

compojoom:
It looks weird when the text and checkbox are not cetner aligned:
grafik

apps/web/src/hooks/useSafeLabsTerms.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export const useSafeLabsTerms = (): UseSafeLabsTermsReturnType => {
+  const isFeatureDisabled = useHasFeature(FEATURES.SAFE_LABS_TERMS_DISABLED) ?? false
+  const isOfficialHost = useIsOfficialHost()
+  const onboard = useOnboard()
+  const router = useRouter()
+  const hasRedirected = useRef(false)
+  const [shouldShowContent, setShouldShowContent] = useState(false)
+
+  const isCypressTest = typeof window !== 'undefined' && window.Cypress !== undefined
```
compojoom: I know why we do it, but it still feels so weird.

PooyaRaki: @compojoom 
Agree, but what would be the ideal solution in your opinion?
I was thinking of creating a common method named something like `isTestEnvironment` to check if the environment is test or not.

PooyaRaki:
@compojoom 
> You are missing letter-spacing 0.17px on the text, because of this the text looks more cramped than on the figma design.

> It looks weird when the text and checkbox are not cetner aligned:

Fixed both [here](https://github.com/safe-global/safe-wallet-monorepo/pull/6446/commits/d1be47010851b36f685c9918d141fa8740db2620) ✅

apps/web/src/hooks/useSafeLabsTerms.ts
compojoom: this is actually a godo idea. If you search for cypress in the codebase you would see that we have a similar check in another file. Maybe time to centralize.

PooyaRaki: Addressed ✅ https://github.com/safe-global/safe-wallet-monorepo/pull/6446/commits/c1cce4e3963608b23d6f77c2289d1ce603e00478

PooyaRaki: I left the other instances unchanged to keep the PR concise.

apps/web/src/services/safe-labs-terms/security.ts
katspaugh, line 1: Next comes with its own URL parser. Not to mentioned the built-in `URL` class in JavaScript. Ideally we shouldn't reimplement the wheel.

PooyaRaki: It's a layer on top of the `Javascript` built-in url decoder to make sure that the URL we are redirecting to is safe. Do you still want to remove it?

katspaugh: I don't see why it's needed for a terms page.

PooyaRaki: If you open any URL within Safe, you’ll be redirected to the terms page(If you haven't accepted the terms yet), with the original page URL included as a query parameter. This allows us to send you back to your intended destination after you accept the terms. We validate this query parameter to ensure it’s safe and free from malicious content - since it’s part of the URL, we need to verify it before redirecting users to their requested page. The `security.ts` file mainly handles this verification.

katspaugh: Ah, I see. So it’s essentially to open the previous route. In that case we could simply call router.back() without introducing an XSS vector in the first place.

PooyaRaki: No, It's not possible because we're adding some custom query params to the URL as well e.g. to know if the connect wallet modal should be opened automatically after you're redirected or not. The `router.back` method does not accept any params.

PooyaRaki:
>Tested, the cookie is set correctly. ✅
>
> Checkboxes could be aligned to the center of the top line, not to the center of the entire label, but that's minor.

Fixed [here](https://github.com/safe-global/safe-wallet-monorepo/pull/6446/commits/c71d8857e0e7e6236ec13fa600a7a7bde68dc1d0)

PR #6448, opened 2025-10-11
feat(Ledger): show execution hash for Ledger

katspaugh:
Resolves https://linear.app/safe-global/issue/COR-269/signing-ux-v2-display-hash-of-transaction-to-be-executed-for-last

Screenshot 2025-10-11 at 06 37 00

## Summary
Add a modal dialog that displays the transaction hash for Ledger users to compare with their hardware device before confirming transactions.

## Changes
- **New LedgerHashComparison component**: Modal dialog displaying transaction hash with copy functionality
- **Ledger module integration**: Calculates and displays hash before user confirms on device  
- **Event-based architecture**: Uses custom browser events to communicate between Ledger module and React component

## User Flow
1. User initiates transaction execution
2. Transaction data is sent to Ledger device
3. **Modal dialog appears** with the transaction hash
4. Ledger device displays transaction for review
5. User compares hash in browser with hash on Ledger screen
6. User confirms on Ledger device to complete signing
7. User can close the modal dialog

## Technical Details
- Uses existing `EthHashInfo` component for consistent hash display and copy functionality
- Non-blocking: signing process starts immediately, dialog appears asynchronously
- Dialog persists until manually closed for sufficient verification time
- Hash is calculated using `keccak256(transaction.unsignedSerialized)` to match Ledger's display

## Test Plan
- [x] Modal appears when executing transaction with Ledger wallet
- [x] Hash matches the one displayed on Ledger device
- [x] Copy button works correctly
- [x] Dialog can be closed via X button or Close button
- [x] Transaction signing completes successfully after confirmation
- [x] Non-Ledger wallets are not affected

### DIFF
--- a/apps/web/src/components/tx-flow/TxFlow.tsx
+++ b/apps/web/src/components/tx-flow/TxFlow.tsx
@@ -11,6 +11,7 @@ import { TxChecks, TxNote, SignerSelect, Blockaid } from './features'
 import { Batching, ComboSubmit, Counterfactual, Execute, ExecuteThroughRole, Propose, Sign } from './actions'
 import { SlotProvider } from './slots'
 import { useTrackTimeSpent } from '../tx/SignOrExecuteForm/tracking'
+import LedgerHashComparison from '@/features/ledger'
 
 type SubmitCallbackProps = { txId?: string; isExecuted?: boolean }
 export type SubmitCallback = (args?: SubmitCallbackProps) => void
@@ -114,6 +115,8 @@ export const TxFlow = <T extends unknown>({
                   <Propose />
                 </ConfirmTxReceipt>
               </TxFlowContent>
+
+              <LedgerHashComparison />
             </TxFlowProvider>
           </SlotProvider>
         </TxSecurityProvider>
--- a/apps/web/src/components/tx-flow/actions/Execute/ExecuteForm.tsx
+++ b/apps/web/src/components/tx-flow/actions/Execute/ExecuteForm.tsx
@@ -3,7 +3,6 @@ import madProps from '@/utils/mad-props'
 import { type ReactElement, type SyntheticEvent, useContext, useState } from 'react'
 import { Box, CardActions, Divider, Tooltip } from '@mui/material'
 import classNames from 'classnames'
-import LedgerHashComparison from '@/features/ledger'
 
 import ErrorMessage from '@/components/tx/ErrorMessage'
 import { trackError, Errors } from '@/services/exceptions'
@@ -215,8 +214,6 @@ export const ExecuteForm = ({
           </CheckWallet>
         </CardActions>
       </form>
-
-      <LedgerHashComparison />
     </>
   )
 }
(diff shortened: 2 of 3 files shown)


apps/web/src/services/onboard/ledger-module.ts
```diff
@@ -159,11 +159,24 @@ export function ledgerModule(): WalletInit {
                 value: txParams.value ? BigInt(txParams.value) : null,
               })
 
-              transaction.signature = await ledgerSdk.signTransaction(
+              // Calculate hash before signing
+              const { keccak256 } = await import('ethers')
+              const txHash = keccak256(transaction.unsignedSerialized)
```
compojoom, line 164: Interesting, I didn't know that it is just the keccak of the unigned tx.

apps/web/src/services/onboard/ledger-module.ts
```diff
@@ -159,11 +159,24 @@ export function ledgerModule(): WalletInit {
...
+
+              // Start the signing process (sends tx to Ledger, doesn't wait for confirmation yet)
+              const signingPromise = ledgerSdk.signTransaction(
                 getAssertedDerivationPath(),
                 hexaStringToBuffer(transaction.unsignedSerialized)!,
               )
 
+              // Show non-blocking notification immediately after sending to Ledger
+              // Import the function dynamically to avoid circular dependencies
+              const { showLedgerHashComparison } = await import('@/components/common/LedgerHashComparison')
```
compojoom: why do we do this after the signTransaction if we already have the hash before that? and why do we await after this?

What happens if we trigger this before the .signTransaction() and actually await the signTransaction directly?

katspaugh: Good point. It's a leftover from when I was prototyping it with window.alert (which is blocking). It can be now done the way you describe.

apps/web/src/services/onboard/ledger-module.ts
```diff
@@ -159,12 +159,31 @@ export function ledgerModule(): WalletInit {
...
-                hexaStringToBuffer(transaction.unsignedSerialized)!,
-              )
+              // Calculate hash and show comparison dialog before signing
+              const { keccak256 } = await import('ethers')
+              const txHash = keccak256(transaction.unsignedSerialized)
 
-              return transaction.serialized
+              const { showLedgerHashComparison, hideLedgerHashComparison } = await import(
+                '@/components/common/LedgerHashComparison'
+              )
```
compojoom: does this need to be inline. 

The show and hide functions are utility functions/business logic - they have nothing to do with the LedgerHashComparison component. I mean they are not part of the component itself. They should live in a separate file than the component UI.

katspaugh: Refactored, the component and utils/store are separated now, and are in src/features/ledger.

compojoom (approved):
Looks good! The cherry on the cake would be if you add this to the mobile app as well now :)

katspaugh:
Thanks! Can look into it on Friday.

PR #6452, opened 2025-10-13
fix(web/mobile): Disable guard during Blockaid simulation for accurate validation [COR-374]

gjeanmart:
## What it solves

Resolves #COR-374

## How this PR fixes it

Safe Guards can interfere with Blockaid's transaction simulation and validation process, leading to inaccurate security analysis results. When a Safe has a guard configured, it may prevent proper simulation of transaction effects, causing validation failures or incomplete security assessments.

**Solution**
This PR implements a `state_override` mechanism in the Blockaid module to temporarily disable any configured Safe Guard during simulation:

- **State Override**: Sets the guard storage slot (`GUARD_STORAGE_POSITION`) to zero address during simulation
- **Non-Destructive**: Only affects the simulation environment, not the actual Safe state
- **Comprehensive**: Ensures accurate validation of Safe transaction payloads without guard interference

## Technical Details

The fix adds a `state_override` parameter to the Blockaid API payload:

```typescript
state_override: {
  [safeAddress]: {
    stateDiff: {
      [GUARD_STORAGE_POSITION]: '0x0000000000000000000000000000000000000000000000000000000000000000',
    },
  },
}
```

This temporarily sets the Safe's guard to the zero address during simulation, allowing Blockaid to properly analyze the transaction without guard-related restrictions.

## How to test it

**Manual Testing Steps:**
1. Create a Safe with a Guard enabled
1. Initiate a transaction 
1. Verify that Blockaid simulation now completes successfully
1. Confirm that proper validation results are returned

**Expected Behavior:**
1. Transactions on guarded Safes should now receive accurate security analysis
1. No impact on actual Safe functionality or guard behavior
1. Improved reliability of Blockaid security assessments

## Screenshots

**Before (Guard interference causing validation issues):**
image

**After (Successful validation with guard override):**
image

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/components/FiatChange/FiatChange.tsx
+++ b/apps/mobile/src/components/FiatChange/FiatChange.tsx
@@ -1,0 +1,53 @@
+import React from 'react'
+import { Text, View } from 'tamagui'
+import { type Balance } from '@safe-global/store/gateway/AUTO_GENERATED/balances'
+import { formatPercentage } from '@safe-global/utils/utils/formatters'
+
+interface FiatChangeProps {
+  balanceItem: Balance
+}
+
+export const FiatChange = ({ balanceItem }: FiatChangeProps) => {
+  if (!balanceItem.fiatBalance24hChange) {
+    return (
+      <Text fontSize="$3" color="$colorSecondary" opacity={0.7}>
+        0%
+      </Text>
+    )
+  }
+
+  const changeAsNumber = Number(balanceItem.fiatBalance24hChange) / 100
+  const changeLabel = formatPercentage(changeAsNumber)
+  const direction = changeAsNumber < 0 ? 'down' : changeAsNumber > 0 ? 'up' : 'none'
+
+  const getColor = () => {
+    switch (direction) {
+      case 'down':
+        return '$error'
+      case 'up':
+        return '$success'
+      default:
+        return '$colorSecondary'
+    }
+  }
+
+  const changeSign = () => {
+    switch (direction) {
+      case 'down':
+        return '-'
+      case 'up':
+        return '+'
+      default:
+        return ''
+    }
+  }
+
+  return (
+    <View flexDirection="row" alignItems="center" gap="$1">
+      <Text fontSize="$3" color={getColor()} fontWeight="500">
+        {changeSign()}
+        {changeLabel}
+      </Text>
+    </View>
+  )
+}
--- a/apps/mobile/src/components/FiatChange/__tests__/FiatChange.test.tsx
+++ b/apps/mobile/src/components/FiatChange/__tests__/FiatChange.test.tsx
@@ -1,0 +1,55 @@
+import React from 'react'
+import { render } from '@/src/tests/test-utils'
+import { FiatChange } from '../FiatChange'
+import { type Balance } from '@safe-global/store/gateway/AUTO_GENERATED/balances'
+
+describe('FiatChange', () => {
+  it('renders "n/a" when fiatBalance24hChange is not present', () => {
+    const mockBalance: Balance = {
+      fiatBalance24hChange: undefined,
+    } as Balance
+
+    const { getByText } = render(<FiatChange balanceItem={mockBalance} />)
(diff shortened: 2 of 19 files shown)


packages/utils/src/services/security/modules/BlockaidModule/index.ts
```diff
@@ -60,8 +60,12 @@ type BlockaidPayload = {
     params: [string, string]
   }
   options: ['simulation', 'validation']
+  state_override?: Record<string, { state?: Record<string, string> }>
 }
 
+// Safe Smart Account Storage Slot for Guard
+const GUARD_STORAGE_POSITION = '0x4a204f620c8c5ccdca3fd54d003badd85ba500436a431f0cbda4f558c93c34c8'
```
PooyaRaki, line 67: I'd move it to `apps/web/src/config/constants.ts` for more consistency.

packages/utils/src/services/security/modules/BlockaidModule/index.ts
```diff
@@ -123,6 +127,16 @@ export class BlockaidModule implements SecurityModule<BlockaidModuleRequest, Blo
         : {
             non_dapp: true,
           },
+      state_override: {
+        [safeAddress]: {
+          state: {
+            // Safe Smart Account Storage Slot for Guard
```
PooyaRaki: Nit: This is redundant as the name of the constant clearly shows that it's the guard storage position and we have the same comment of the constant.

packages/utils/src/services/security/modules/BlockaidModule/index.ts
```diff
@@ -123,6 +127,16 @@ export class BlockaidModule implements SecurityModule<BlockaidModuleRequest, Blo
...
+            [GUARD_STORAGE_POSITION]:
+              // Set to zero address to disable guard
```
PooyaRaki: This is an example of a good comment that explains the why, thanks 💯

PooyaRaki:
@gjeanmart 
With this change, Blockaid no longer behaves as expected.
The screenshots below are from a Safe without a Guard. The production app correctly identifies the owner change, while your branch does not.

It seems like the issue is that you’re overriding the guard with a null address to bypass it — but the provider interprets this as if you’re sending the transaction to the null address itself.
The same thing happens to a Safe with a guard, I think the provider checks the TX against the null address not the actual address you're sending to.

Screenshot 2025-10-13 at 17 53 20
Screenshot 2025-10-13 at 17 53 31

packages/utils/src/services/security/modules/BlockaidModule/index.ts
gjeanmart: Since this is part of the `utils` package, I wasn’t sure of the best place to put this constant. Let me know if you have a better suggestion.

packages/utils/src/services/security/modules/BlockaidModule/index.ts
gjeanmart: fixed in [d4abf34](https://github.com/safe-global/safe-wallet-monorepo/pull/6452/commits/d4abf34319744d4421056bc43d58cd34628d4fdc)

gjeanmart:
> @gjeanmart With this change, Blockaid no longer behaves as expected. The screenshots below are from a Safe without a Guard. The production app correctly identifies the owner change, while your branch does not.
> 
> It seems like the issue is that you’re overriding the guard with a null address to bypass it — but the provider interprets this as if you’re sending the transaction to the null address itself. The same thing happens to a Safe with a guard, I think the provider checks the TX against the null address not the actual address you're sending to.
> 
> Screenshot 2025-10-13 at 17 53 20 Screenshot 2025-10-13 at 17 53 31

@PooyaRaki Good catch 🙏 

Fixed in [d4abf34](https://github.com/safe-global/safe-wallet-monorepo/pull/6452/commits/d4abf34319744d4421056bc43d58cd34628d4fdc)

gjeanmart:
Closed it in favor of #6464 (shorter branch name)

PR #6456, opened 2025-10-13
feat(Safe Shield): Recipient analysis data fetching

tmjssz:
## What it solves

Resolves [COR-658](https://linear.app/safe-global/issue/COR-658/recipient-address-analysis-data-fetching)

Introduces logic to fetch recipient analysis results and integrates them with local address checks. Also refactors the address analysis hooks architecture.

### New Features
- **Backend integration**: Added `useFetchRecipientAnalysis` hook to fetch recipient analysis data from backend API
- **Combined analysis**: Implemented `useRecipientAnalysis` hook that merges backend results with local checks (address book, activity, etc.)
- **Deep comparison utilities**: Added custom hooks (`useAsyncDeepCompare`, `useEffectDeepCompare`, `useMemoDeepCompare`) for better dependency handling with complex objects

### Refactoring
- **Reorganized structure**: Moved address analysis hooks from `features/address-analysis/` to `features/safe-shield/hooks/address-analysis/`
- **Updated hooks**: Refactored `useAddressActivity` and `useAddressBookCheck` to support batch analysis of multiple addresses
- **Type updates**: Enhanced types in `features/safe-shield/types.ts` for recipient analysis data

## How to test it
Not testable yet, because this does only include core data fetching logic, but no visual UI changes (those will follow in separate tasks).

## Checklist

---

## CLA signature


packages/utils/src/features/safe-shield/hooks/useFetchRecipientAnalysis.ts
```diff
@@ -0,0 +1,106 @@
...
+      }
+    })
+
+    return toFetch
+  }, [recipients])
+
+  // Update previous recipients after determining what to fetch
+  useEffectDeepCompare(() => {
+    previousRecipientsRef.current = new Set(recipients)
+  }, [recipients])
```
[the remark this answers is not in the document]
tmjssz: Addressed here -> 9f48d9dbd6233528424221b25a7ee91cc87e04f1

packages/utils/src/features/safe-shield/hooks/address-analysis/address-activity/useAddressActivity.ts
```diff
@@ -0,0 +1,91 @@
...
+      if (!previousSet.has(address)) {
+        toFetch.push(address)
+      }
+    })
+
+    return toFetch
+  }, [addresses])
+
+  // Update previous recipients after determining what to fetch
+  useEffectDeepCompare(() => {
```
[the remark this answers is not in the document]
tmjssz: Addressed here -> 5f50b02e56d084438f549f1c2debf6eafa70a01f

packages/utils/src/features/safe-shield/hooks/address-analysis/address-activity/useAddressActivity.ts
```diff
@@ -0,0 +1,101 @@
...
+import { analyzeAddressActivity, isLowActivityAddress } from './addressActivityService'
+import { ActivityMessages } from '../config'
+import { type AnalysisResult, RecipientStatus, Severity } from '../../../types'
+import { useEffectDeepCompare } from '../../util-hooks/useEffectDeepCompare'
+import { useAsyncDeepCompare } from '../../util-hooks/useAsyncDeepCompare'
+
+export type AddressActivityResult = Record<
+  string,
+  AnalysisResult<RecipientStatus.LOW_ACTIVITY | RecipientStatus.HIGH_ACTIVITY>
+>
```
LucieFaire, line 12: I thought we only need to record the LOW_ACTIVITY (according to the spec)

apps/web/src/features/safe-shield/hooks/useFetchRecipientAnalysis.ts
```diff
@@ -0,0 +1,117 @@
...
+ * Returns RECIPIENT_INTERACTION results only
+ * ADDRESS_BOOK and RECIPIENT_ACTIVITY are performed client-side
+ */
+async function fetchRecipientAnalysis(
+  chainId: string,
+  safeAddress: string,
+  recipientAddress: string,
+): Promise<AddressAnalysisResults | undefined> {
+  const url = `${GATEWAY_URL}/v1/chains/${chainId}/security/${safeAddress}/recipient/${recipientAddress}`
+
```
LucieFaire: why not use RTK and auto-generated routes instead the hardcoded URL?

LucieFaire: to support that i have opened a new PR to have proper type defintions so you can have them generated here too: https://github.com/safe-global/safe-client-gateway/compare/safe-shield/recipient-swagger-types?expand=1

tmjssz: Done in three commits: 
* dc6fa6c5dacae14c33813c907ceb34cb9a6b01ce
* 056d1c17c3297218dadd4a6ceb0232470052f5a4
* 5f575968554b70ab4f26a3b2f6341dbbb12a1235

packages/utils/src/features/safe-shield/hooks/address-analysis/address-activity/useAddressActivity.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * React hook to analyze activity for multiple addresses
+ * @param addresses - Array of Ethereum addresses to analyze
+ * @returns Object containing activity results for each address, loading state, and error
+ */
+export const useAddressActivity = (
+  addresses: string[],
+  web3ReadOnly?: JsonRpcProvider,
```
katspaugh, line 21: Why do we need to pass these as props? The hook can call useOwnedSafes and useWeb3ReadOnly on its own.

packages/utils/src/features/safe-shield/hooks/useFetchRecipientAnalysis.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Hook to fetch recipient analysis from backend API
+ * Tracks which recipients have been fetched and only fetches new ones
+ * @param recipients - Array of recipient addresses to fetch analysis for
+ * @returns Tuple of [results, error, loading] where results is keyed by recipient address
+ */
+export function useFetchRecipientAnalysis({
+  safeAddress,
+  chainId,
```
katspaugh, line 15: safeAddress and chainId could be taken from useSafeAddress and useChainId instead of props.

packages/utils/src/features/safe-shield/hooks/useFetchMultiRecipientAnalysis.ts
```diff
@@ -0,0 +1,46 @@
...
+import { useAsyncDeepCompare } from './util-hooks/useAsyncDeepCompare'
+
+/**
+ * Wrapper hook for useSafeShieldAnalyzeRecipientV1Query that fetches analysis for multiple recipients
+ * @param recipientAddresses - Array of recipient addresses to fetch analysis for
+ * @returns Tuple of [results, error, loading] where results is keyed by recipient address
+ */
+export function useFetchMultiRecipientAnalysis({
+  safeAddress,
+  chainId,
```
katspaugh, line 13: Same here.

packages/utils/src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,14 @@
+import { useEffect, useState } from 'react'
+
+const useDebounce = <T>(value: T, delay: number): T => {
```
katspaugh, line 3: We already have a hook for this. `useDebouncedValue`.

packages/utils/src/features/safe-shield/hooks/address-analysis/address-activity/useAddressActivity.ts
compojoom: Note that this is in the utils folder. It's going to be consumed by the mobile app as well and there the way we init the web3ReadOnly is slightly different than what we do on web. 

We are going to have wrappers in mobile and web around this function.

tmjssz: I did this to make the hook reusable for the mobile implementation. Initially, it was structured as you mentioned, but after discussing with @compojoom, we decided to move the hook logic to the `@safe-global/utils` package to simplify reuse later when the mobile implementation of Safe Shield starts. As a result, I refactored all those values into params that can be provided by the consuming app.

packages/utils/src/hooks/useDebounce.ts
tmjssz: Since I moved the Safe Shield hook logic to the utils package, I also had to copy the useDebounce hook there. I decided not to remove the original useDebounce hook from the web package in this PR to avoid making it even larger and touching too many files. I’m happy to do that in a follow-up PR.

packages/utils/src/features/safe-shield/hooks/address-analysis/address-activity/useAddressActivity.ts
katspaugh: Gotcha, thanks guys.

PR #6457, opened 2025-10-13
feat(mobile): do not allow user to execute tx if he has not enough funds

clovisdasilvaneto:
## What it solves
If user selects an account that doesn't have enough funds, he get stuck in the execute transaction screen

Resolves https://linear.app/safe-global/issue/COR-656/mobile-dont-allow-to-select-the-signer-with-incufficient-balance-for

## How this PR fixes it
We block the execution button in case user has not enough funds

## Screenshots
Screenshot 2025-10-13 at 13 27 13

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/ExecuteTx/hooks/useExecutionFunds.ts
+++ b/apps/mobile/src/features/ExecuteTx/hooks/useExecutionFunds.ts
@@ -31,14 +31,13 @@ export const useExecutionFunds = ({
   const shouldCheckBalance = executionMethod !== ExecutionMethod.WITH_RELAY && Boolean(signerAddress && chain)
 
   const { data: balances, isLoading } = useGetBalancesQuery(
-    shouldCheckBalance
-      ? {
-          addresses: [signerAddress!],
-          chain: chain!,
-        }
-      : { addresses: [], chain: chain! },
     {
-      skip: !shouldCheckBalance,
+      addresses: signerAddress ? [signerAddress] : [],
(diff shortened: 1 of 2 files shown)


apps/mobile/src/features/ExecuteTx/hooks/useExecutionFunds.ts
```diff
@@ -0,0 +1,81 @@
...
+    {
+      skip: !shouldCheckBalance,
+    },
+  )
+
+  const result = useMemo<UseExecutionFundsResult>(() => {
+    // If using relay, funds are not needed from signer
+    if (executionMethod === ExecutionMethod.WITH_RELAY) {
+      return {
+        hasSufficientFunds: true,
```
compojoom, line 48: why are we assuming that the user has sufficient funds if the method is set to relay? What if there are no free relays left?

clovisdasilvaneto: Because we just allow user to select the ExecutionMethod === Relay if there is remaining free relay calls. We will never have ExecutionMethod === Relay if there is no available relay calls

clovisdasilvaneto: you can check `ReviewAndExecuteFooter.tsx` for reference. The `getExecutionMethod` determines which method the user should see when he opens this screen.

PR #6463, opened 2025-10-13
fix(mobile): Check for relay's feature flag

clovisdasilvaneto:
## What it solves
It hides Relay option if the chain doesn't have this feature enabled.

Resolves https://linear.app/safe-global/issue/COR-580/execution-with-gelatorelaying

## How this PR fixes it
Uses the `hasFeature` for checking if the relay functionality is available in the current chain.

## How to test it
needs to disable the relay option in our feature flag provider and then check if Relay is still available.

## Screenshots
Screenshot 2025-10-13 at 17 30 22
Screenshot 2025-10-13 at 17 30 25

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/ReviewExecuteFooter.tsx
+++ b/apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/ReviewExecuteFooter.tsx
@@ -35,7 +35,7 @@ interface ReviewFooterProps {
 const getExecutionMethod = (
   requestedMethod: ExecutionMethod | undefined,
   isRelayAvailable: boolean,
-  chain?: Chain,
+  chain: Chain,
 ): ExecutionMethod => {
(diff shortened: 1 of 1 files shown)


apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/ReviewExecuteFooter.tsx
```diff
@@ -33,9 +35,13 @@ interface ReviewFooterProps {
 const getExecutionMethod = (
   requestedMethod: ExecutionMethod | undefined,
   isRelayAvailable: boolean,
+  chain?: Chain,
```
compojoom: This is misleading. Chain is actually required and not optional. Maybe we should change the signature here and do the guard whre we call getExecutionMethod?

clovisdasilvaneto: well the hook selector returns chain || null that is why it is optional here. But indeed it makes sense to not have it as optional.

PR #6487, opened 2025-10-15
fix(mobile): not able to sign with ledger imported from old app

compojoom:
## What it solves
When importing data from the old app, the derivationPath stored in the old app was slightly different than what the new app would expect. This caused the signing of tx to fail

Resolves https://linear.app/safe-global/issue/COR-690/mobile-can-not-sing-with-old-imported-ledger

## How this PR fixes it
"fix" the derivationPath on import

## How to test it
Import data from the old app. Try to sign with a ledger account that was imported.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/DataImport/ImportSuccessScreen.container.tsx
+++ b/apps/mobile/src/features/DataImport/ImportSuccessScreen.container.tsx
@@ -38,6 +38,8 @@ export const ImportSuccessScreen = () => {
         router.dismissAll()
         // closes first screen in stack
         router.back()
+        // closes terms and conditions screen
+        router.back()
         // Navigate to the main assets screen
         router.replace('/(tabs)')
         return


apps/mobile/src/features/DataImport/ImportSuccessScreen.container.tsx
```diff
@@ -38,6 +38,8 @@ export const ImportSuccessScreen = () => {
         router.dismissAll()
         // closes first screen in stack
         router.back()
+        // closes terms and conditions screen
+        router.back()
```
katspaugh, line 42: Two router.back is brittle.

compojoom: I know, but at the same time I don't know a better way. Will need to ask at expo. We have multiple navigators stacked on top and there isn't a single command to close them all and navigate to a particular screen.

PR #6490, opened 2025-10-16
feat(Safe Shield): move tenderly simulation to Safe Shield widget

vseehausen:
## What it solves

Resolves https://linear.app/safe-global/issue/COR-589/safe-shield-widget-transaction-simulation-tenderly

## How to test it

Simulate:
* transaction
* nested transaction
* batch transaction
* failing transaction

and observe widget.

## Screenshots

grafik

## Checklist

---

## CLA signature


vseehausen:
@github-copilot can you resolve the type errors?

apps/web/src/features/safe-shield/components/TenderlySimulation.tsx
```diff
@@ -0,0 +1,278 @@
...
+import useSafeAddress from '@/hooks/useSafeAddress'
+import { useGetTransactionDetailsQuery } from '@/store/api/gateway'
+import { skipToken } from '@reduxjs/toolkit/query'
+import extractTxInfo from '@/services/tx/extractTxInfo'
+import type { SafeTransaction } from '@safe-global/types-kit'
+
+const safeInterface = Safe__factory.createInterface()
+
+export const TenderlySimulation = (): ReactElement | null => {
+  const { safeTx } = useContext(SafeTxContext)
```
tmjssz: It would be ideal to have `safeTx` as a component prop and pass it down from SafeShieldContent. That way we could also make sure that in SafeShieldContent the empty state is not being shown if the safeTx is defined.

tmjssz: This is what i mean
Screenshot 2025-10-16 at 16 41 34

tmjssz:
When I test with a nested transaction, I only see the result for the nested transaction itself in the simulation part. But we actually should do both checks for the parent and the nested transactions, and they should be shown both in the widget. 

Screenshot 2025-10-16 at 16 55 29

apps/web/src/features/safe-shield/components/TenderlySimulation.tsx
```diff
@@ -0,0 +1,323 @@
...
+        )}
+      </Stack>
+
+      <Collapse in={isSimulationFinished && simulationExpanded}>
+        <Box sx={{ padding: '4px 12px 16px' }}>
+          <Stack gap={2}>
+            <Box bgcolor="background.main" borderRadius="4px" overflow="hidden">
+              <Box
+                sx={{
+                  borderLeft: `4px solid ${mainIsSuccess ? 'var(--color-success-main)' : 'var(--color-error-main)'}`,
```
tmjssz: Lets get the colors from [SEVERITY_COLORS](https://github.com/safe-global/safe-wallet-monorepo/blob/tenderly-widget/apps/web/src/features/safe-shield/constants.ts/#L3) here (and line 274), to make sure it's always consistent

apps/web/src/features/safe-shield/components/TenderlySimulation.tsx
```diff
@@ -0,0 +1,323 @@
...
+            ) : (
+              <SvgIcon component={AlertIcon} inheritViewBox sx={{ fontSize: 16 }} />
+            )
+          ) : (
+            <SvgIcon component={UpdateIcon} inheritViewBox sx={{ fontSize: 16 }} />
+          )}
+          <Typography variant="body2" color="primary.light">
+            {isSimulationFinished ? 'Transaction simulations' : 'Transaction simulation'}
+          </Typography>
+          <Tooltip
```
tmjssz: The tooltip should only be shown in the initial state when the simulation is not triggered yet.

apps/web/src/features/safe-shield/components/TenderlySimulation.tsx
```diff
@@ -0,0 +1,330 @@
...
+        alignItems="center"
+        sx={{ padding: '12px', cursor: isSimulationFinished ? 'pointer' : 'default' }}
+        onClick={isSimulationFinished ? handleToggleSimulation : undefined}
+      >
+        <Stack direction="row" alignItems="center" gap={1}>
+          {isSimulationFinished ? (
+            isSimulationSuccess ? (
+              <CheckIcon sx={{ fontSize: 16, color: SEVERITY_COLORS.OK.main }} />
+            ) : (
+              <SvgIcon
```
tmjssz: Can we reuse the SeverityIcon component here? And also for the replace the CheckIcon above.
https://github.com/safe-global/safe-wallet-monorepo/blob/tenderly-widget/apps/web/src/features/safe-shield/components/SeverityIcon.tsx/#L10
Then you could simply do this:

```suggestion
              <SeverityIcon severity={isSimulationSuccess ? Severity.OK : Severity.CRITICAL} />
```

apps/web/src/features/safe-shield/components/TenderlySimulation.tsx
```diff
@@ -0,0 +1,330 @@
...
+  const chain = useCurrentChain()
+  const { safe } = useSafeInfo()
+  const safeAddress = useSafeAddress()
+  const signer = useSigner()
+  const isSafeOwner = useIsSafeOwner()
+  const showSimulation = chain && isTxSimulationEnabled(chain) && safeTx
+
+  const [simulationExpanded, setSimulationExpanded] = useState(false)
+
+  const nestedTxInfo = useMemo(() => {
```
tmjssz: Can we extract this block into a separate function maybe? I think it would be nice for readability.

tmjssz:
For nested transactions, the design looks a bit different. There should be no title "Nested transaction", just the description text. See the Figma design on the right:
Screenshot 2025-10-17 at 14 55 05

Also, another thing I just noticed after reviewing the Figma designs: I think the component should actually only be expandable if there are two simulations. When there’s only a single simulation running (without a nested transaction simulation), we should just show the entry without the option to expand and view the description. Also the title text differs a bit then.

Is that correct, @IvaLukan?

Screenshot 2025-10-17 at 15 30 03

ivalui:
> For nested transactions, the design looks a bit different. There should be no title "Nested transaction", just the description text. See the Figma design on the right: Screenshot 2025-10-17 at 14 55 05
> 
> Also, another thing I just noticed after reviewing the Figma designs: I think the component should actually only be expandable if there are two simulations. When there’s only a single simulation running (without a nested transaction simulation), we should just show the entry without the option to expand and view the description. Also the title text differs a bit then.
> 
> Is that correct, @IvaLukan?
> 
> Screenshot 2025-10-17 at 15 30 03

Sorry @tmjssz, just saw this now. Yes, you are correct 🙏

PR #6495, opened 2025-10-16
fix(web): update safe-mockup image for welcome screen

PooyaRaki:
Resolves [#COR-697](https://linear.app/safe-global/issue/COR-697/update-safe-wallet-welcome-image)

The hero image on the welcome page appears blurry at certain resolutions.
This PR replaces it with a higher-resolution version to improve visual clarity across devices.


apps/web/public/images/welcome/safe-mockup.png
katspaugh, line 1: The size is still not 2x but better. Did you run it through tinypng?

PooyaRaki: I exported it as 2x, Do you want me to increase the size again?

PR #6497, opened 2025-10-16
refactor(web): use SafeState from store package

compojoom:
## What it solves
use SafeState from store package

---

## CLA signature


compojoom:
will follow up with another pr

PR #6498, opened 2025-10-16
feat(Safe Shield): integrate recipient analysis

tmjssz:
## What it solves

Resolves [COR-659](https://linear.app/safe-global/issue/COR-659/integrate-recipient-analysis-into-token-transfer-first-step)

This PR integrates Safe Shield recipient analysis into the token transfer flow, providing real-time security analysis of recipient addresses during transaction creation.

#### Context Providers

- **Wrapped transaction flows with `SafeShieldProvider`**
  - Added to `TxFlow.tsx` - main transaction flow wrapper
  - Added to `TxLayout/index.tsx` - legacy layout wrapper
  - Provides context for Safe Shield analysis across all transaction steps

#### Token Transfer Flow

- **Added recipient address tracking** in `CreateTokenTransfer.tsx`
  - Uses `useWatch` to monitor form recipient fields in real-time

### Single recipient
1. Open the token transfer flow. 
2. Enter an address into the recipient field. 
3. Observe the Safe Shield widget loading the analysis data for that address.
4. When expanding the recipient analysis entry, all the results should be displayed. 
5. When changing the address, the widget should load new results for this address. 
6. Only if the recipient field contains a valid Ethereum address, the analysis should be triggered. 

### Multiple recipients
1. Add multiple recipients with different addresses. 
2. The Safe Shield widget should load new results for every address added. 
3. The result descriptions should be consolidated for all the addresses. (E.g. "You are interacting with 2 recipients for the first time.")

## Screenshots
Screenshot 2025-10-16 at 19 35 14

https://github.com/user-attachments/assets/7ae0cac3-c0f8-4e69-b2d1-86e622371e1a

## Checklist

---

## CLA signature

### DIFF
--- a/packages/utils/src/features/safe-shield/hooks/useRecipientAnalysis.ts
+++ b/packages/utils/src/features/safe-shield/hooks/useRecipientAnalysis.ts
@@ -94,7 +94,7 @@ export function useRecipientAnalysis({
   // Merge backend and local checks
   // Only merge address book results after fetched results are available
   const mergedResults = useMemo(() => {
-    const addressBookToMerge = !!fetchedResults && !!addressBookCheck ? addressBookCheck : undefined
+    const addressBookToMerge = fetchedResults && addressBookCheck ? addressBookCheck : undefined
     return mergeAnalysisResults(fetchedResults, addressBookToMerge, activityCheck)
   }, [fetchedResults, addressBookCheck, activityCheck])
 


apps/web/src/features/safe-shield/SafeShieldContext.tsx
```diff
@@ -0,0 +1,44 @@
...
+    throw new Error('useSafeShieldContext must be used within SafeShieldProvider')
+  }
+  return context
+}
+
+/**
+ * Hook to register recipient addresses for Safe Shield analysis
+ * @param recipientAddresses - Array of recipient addresses to analyze
+ */
+export const useSafeShieldForRecipients = (recipientAddresses: string[]) => {
```
vseehausen, line 36: could we call it `useRecipientAnalysis` or `useSafeShieldRecipientAnalysis`?

apps/web/src/components/tx-flow/common/TxLayout/index.tsx
```diff
@@ -107,98 +108,100 @@ const TxLayout = ({
...
-                        sx={{ fontWeight: '700' }}
+          <SafeShieldProvider>
+            <Grid container className={css.container}>
+              {!isReplacement && !isSmallScreen && (
+                <Grid sx={{ width: 200 }} pt={5}>
+                  <aside>
+                    <Stack gap={3} position="fixed">
+                      <Card
+                        sx={{
+                          p: '4px 8px 4px 4px',
```
vseehausen: can this be moved to a css file? (not necessary, but nice to have). Also applies to other inline styles. (girl scout rule)

apps/web/src/features/safe-shield/components/SeverityIcon.tsx
```diff
@@ -17,30 +29,11 @@ export const SeverityIcon = ({
...
-    default:
-      return (
-        <SvgIcon
-          sx={{ ...iconProps, path: { fill: color }, rect: { fill: color } }}
-          component={InfoIcon}
-          inheritViewBox
-        />
-      )
+  const props = {
+    sx: { ...iconProps, ...getIconProps(severity) },
```
vseehausen, line 33: nice!

packages/utils/src/features/safe-shield/utils/mapConsolidatedAnalysisResults.ts
```diff
@@ -37,11 +37,13 @@ export const mapConsolidatedAnalysisResults = (
     for (const [type, typeResults] of Object.entries(groupResults)) {
       const numResults = typeResults.length
       if (numResults > 0) {
+        const pluraliseDescription =
```
vseehausen: Nice improvement. Maaaaaaybe `formatPluralDescription`? I didn't get that `pluralise` is a verb, took me some seconds. Nit-pick!

apps/web/src/features/safe-shield/SafeShieldContext.tsx
tmjssz: If we would call it useRecipientAnalysis, then it would be in conflict with the [other hook](https://github.com/safe-global/safe-wallet-monorepo/blob/feat%2Fintegrate-recipient-analysis/apps/web/src/features/safe-shield/hooks/useRecipientAnalysis.ts/#L10) that has the same name but a different purpose. What do you think about `useRegisterRecipientsForSafeShield`?

apps/web/src/components/tx-flow/common/TxLayout/index.tsx
tmjssz: Done -> dfc1bdc14e6c38a54e64692b25386dca0ec9269b

packages/utils/src/features/safe-shield/utils/mapConsolidatedAnalysisResults.ts
tmjssz: Thanks, yeah I get the confusion here :D Done -> cb411e9469fc8ad0689fdc6cab5fe784269aab47

apps/web/src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
```diff
@@ -164,6 +166,14 @@ export const CreateTokenTransfer = ({ txNonce }: CreateTokenTransferProps): Reac
 
   const canBatch = isMassPayoutsEnabled && type === TokenTransferType.multiSig
 
+  const recipientsWatched = useWatch({ control, name: MultiTokenTransferFields.recipients })
+  const recipientAddresses = useMemo(
```
PooyaRaki, line 170: Q: Why do we need to use `useMemo` here? It feels unnecessary since we only have up to five recipients - the computation isn’t heavy enough to justify memoization.

packages/utils/src/features/safe-shield/hooks/useRecipientAnalysis.ts
```diff
@@ -19,17 +19,19 @@ import useDebounce from '@safe-global/utils/hooks/useDebounce'
...
-  addressBookResult: AddressBookCheckResult,
+  addressBookResult: AddressBookCheckResult | undefined,
   activityResult: AddressActivityResult | undefined,
 ): RecipientAnalysisResults {
   const merged: RecipientAnalysisResults = fetchedResults ? { ...fetchedResults } : {}
 
-  for (const [address, result] of Object.entries(addressBookResult)) {
-    merged[address] = { ...(merged[address] || {}), [StatusGroup.ADDRESS_BOOK]: [result] }
+  if (addressBookResult) {
+    for (const [address, result] of Object.entries(addressBookResult || {})) {
```
PooyaRaki: For more readability I'd extract the `Object.entries(addressBookResult || {}` out of the loop and define a new `const` for that. There are other instances as well.

packages/utils/src/features/safe-shield/hooks/useRecipientAnalysis.ts
```diff
@@ -86,10 +88,11 @@ export function useRecipientAnalysis({
   const [activityCheck, activityCheckError, activityCheckLoading] = useAddressActivity(validRecipients, web3ReadOnly)
 
   // Merge backend and local checks
-  const mergedResults = useMemo(
-    () => mergeAnalysisResults(fetchedResults, addressBookCheck, activityCheck),
-    [fetchedResults, addressBookCheck, activityCheck],
-  )
+  // Only merge address book results after fetched results are available
+  const mergedResults = useMemo(() => {
+    const addressBookToMerge = !!fetchedResults && !!addressBookCheck ? addressBookCheck : undefined
```
PooyaRaki: We can remove double exclamation marks here.

PooyaRaki (commented):
Looks good but left a couple of comments.

LucieFaire (approved):
👍

apps/web/src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
vseehausen: Hm, interesting topic. `useMemo` is mainly used to provide identity stability to consumers of this value. I actually think we should use `React.memo` for the safe shield content (even though rendering is cheap) to leverage MORE identity stability.

`useSafeShieldForRecipients` uses this in `useEffect`. Would it re-run more often if we would remove `useMemo` here?

tmjssz: Yes exactly as Valentin explained. We memoize recipientAddresses so its reference doesn’t change on each render. Otherwise, useSafeShieldForRecipients sees a “new” array every time and re-runs unnecessarily.

packages/utils/src/features/safe-shield/hooks/useRecipientAnalysis.ts
tmjssz: Done -> 0352d96d7d4537b4708183c51b8c65eefd4131f6

packages/utils/src/features/safe-shield/hooks/useRecipientAnalysis.ts
tmjssz: Done -> 9952304746b3df2a554401262de7ba9a2e194c4d

apps/web/src/features/safe-shield/SafeShieldContext.tsx
vseehausen: Ok. Then I prefer the initial version. Resolved.

apps/web/src/components/tx-flow/flows/TokenTransfer/CreateTokenTransfer.tsx
PooyaRaki: Well, depends on how much we re render the component, if it's not much we don't need to memoize anything.

PR #6499, opened 2025-10-16
refactor(web): use safeApp type and hooks from store package

compojoom:
## What it solves
- use safeApp data type from the store package
- refactored 2 hooks to use the rtk query hooks

## How to test it
Interact with the safe apps. They should work as expected.

---

## CLA signature


apps/web/src/components/safe-apps/utils.ts
```diff
@@ -78,6 +77,7 @@ export const getEmptySafeApp = (url = '', appData?: SafeAppData): SafeAppDataWit
     features: [],
     developerWebsite: '',
     socialProfiles: [],
+    featured: false,
```
PooyaRaki, line 80: Q: Are we hard coding `featured` because we're showing them somewhere else?

config/test/msw/mockSafeApps.ts
```diff
@@ -0,0 +1,153 @@
...
+  url: 'https://app-c.com',
+  iconUrl: '',
+  description: '',
+  chainIds: ['5'],
+  accessControl: { type: SafeAppAccessPolicyTypes.NoRestrictions },
+  tags: ['test'],
+  features: [],
+  socialProfiles: [],
+  developerWebsite: '',
+  featured: false,
```
PooyaRaki, line 147: Shall we mock one `featured` app to check this feature as well?

apps/web/src/hooks/safe-apps/useSafeAppFromBackend.ts
```diff
@@ -1,25 +1,37 @@
...
-import { getSafeApps } from '@safe-global/safe-gateway-typescript-sdk'
+import type { SafeApp as SafeAppData } from '@safe-global/store/gateway/AUTO_GENERATED/safe-apps'
+import { useLazySafeAppsGetSafeAppsV1Query } from '@safe-global/store/gateway/AUTO_GENERATED/safe-apps'
 import { Errors, logError } from '@/services/exceptions'
 import type { AsyncResult } from '@safe-global/utils/hooks/useAsync'
 import useAsync from '@safe-global/utils/hooks/useAsync'
 import { trimTrailingSlash } from '@/utils/url'
 
 const useSafeAppFromBackend = (url: string, chainId: string): AsyncResult<SafeAppData> => {
+  const [trigger] = useLazySafeAppsGetSafeAppsV1Query()
```
PooyaRaki, line 10: Nit: `SafeApps` is repeated twice in the method name. Could be something like:
```suggestion
  const [trigger] = useLazyGetSafeAppsV1Query()
```

apps/web/src/hooks/safe-apps/useRemoteSafeApps.ts
```diff
@@ -56,10 +40,10 @@ const useRemoteSafeApps = ({ tag, name }: UseRemoteSafeAppsProps = {}): AsyncRes
   }, [remoteApps, tag, name])
 
   const sortedApps = useMemo(() => {
```
PooyaRaki, line 42: Not part of this PR so I call it a Nitpick for the scope of this PR but it's a performance issue.
What do you think about changing the code like:
```suggestion
  const sortedApps = apps.length ? useMemo(() => { ... } : []
```
because if the app is empty/undefined we don't need to call use memo and memoize an empty array.

I’ll leave it up to your discretion whether you want to fix it here or in a follow-up PR.

apps/web/src/hooks/safe-apps/useRemoteSafeApps.ts
```diff
@@ -1,29 +1,12 @@
...
+  useSafeAppsGetSafeAppsV1Query,
+} from '@safe-global/store/gateway/AUTO_GENERATED/safe-apps'
+import { useMemo } from 'react'
 import type { AsyncResult } from '@safe-global/utils/hooks/useAsync'
-import useAsync from '@safe-global/utils/hooks/useAsync'
-
-// To avoid multiple simultaneous requests (e.g. the Dashboard and the SAFE header widget),
-// cache the request promise for 100ms
-let cache: Record<string, Promise<SafeAppsResponse> | undefined> = {}
-const cachedGetSafeApps = (chainId: string): ReturnType<typeof getSafeApps> | undefined => {
```
PooyaRaki, line 13: QQ: Why are we removing this method? Does the new get method have caching in it?

compojoom: Yes, with RTK we get automatic cache for free

apps/web/src/components/safe-apps/utils.ts
compojoom: We are hardcoding featured, because the type needs featured and our old type was missing it. getEmptySafeApp is used in few places. We are basically using it as the starting point object and then override name, iconUrl, chainIds etc

apps/web/src/hooks/safe-apps/useRemoteSafeApps.ts
compojoom: You cannot do conditinal hooks in react. Once we turn on the react compiler in the project we don't have to do useMemo, because react is going to do it for us.

apps/web/src/hooks/safe-apps/useSafeAppFromBackend.ts
compojoom: This is all autogenerated. It looks weird and when I was generating the APIs I wanted to have it shorter, but then we drift too much from the CGW code, so we have to live with the verbosity.

config/test/msw/mockSafeApps.ts
compojoom: here we have tests for featured:  true https://github.com/safe-global/safe-wallet-monorepo/pull/6499/files/08b22feba72d64540cf2a873b597787d0dad3f60#diff-337ae942861dde11fc9a1d1b8a66208555b5e2db5da9df5f5d946608657b3940

PooyaRaki (approved):

apps/web/src/hooks/safe-apps/useRemoteSafeApps.ts
PooyaRaki: The suggested code was confusing, I meant we don't need to memoize this array.

PR #6501, opened 2025-10-17
fix(web/mobile): Disable guard during Blockaid simulation for accurate validation [COR-374]

gjeanmart:
## What it solves

Resolves [COR-374](https://linear.app/safe-global/issue/COR-374/hypernative-guard-and-blockaid-simulation-conflict)

## How this PR fixes it

Safe Guards can interfere with Blockaid's transaction simulation and validation process, leading to inaccurate security analysis results. When a Safe has a guard configured, it may prevent proper simulation of transaction effects, causing validation failures or incomplete security assessments.

**Solution**
This PR implements a `state_override` mechanism in the Blockaid module to temporarily disable any configured Safe Guard during simulation:

- **State Override**: Sets the guard storage slot (`GUARD_STORAGE_POSITION`) to zero address during simulation
- **Non-Destructive**: Only affects the simulation environment, not the actual Safe state
- **Comprehensive**: Ensures accurate validation of Safe transaction payloads without guard interference

## Technical Details

The fix adds a `state_override` parameter to the Blockaid API payload:

```typescript
state_override: {
  [safeAddress]: {
    stateDiff: {
      [GUARD_STORAGE_POSITION]: '0x0000000000000000000000000000000000000000000000000000000000000000',
    },
  },
}
```

This temporarily sets the Safe's guard to the zero address during simulation, allowing Blockaid to properly analyze the transaction without guard-related restrictions.

## How to test it

**Manual Testing Steps:**
1. Create a Safe with a Guard enabled
1. Initiate a transaction 
1. Verify that Blockaid simulation now completes successfully
1. Confirm that proper validation results are returned

**Expected Behavior:**
1. Transactions on guarded Safes should now receive accurate security analysis
1. No impact on actual Safe functionality or guard behavior
1. Improved reliability of Blockaid security assessments

## Screenshots

**Before (Guard interference causing validation issues):**
image

**After (Successful validation with guard override):**
image

## Checklist

---

## CLA signature


katspaugh:
@gjeanmart we should probably migrate this to the CGW, as we're moving Blockaid there.

akshay-ap (approved):
LGTM! Would be nice to have additional information about overriding guard slot for simulation

PooyaRaki:
@katspaugh It's supposed to be released before we move the migrate to CGW but agree we should do the same thing in CGW.

I'm testing the TX flow with this fix, I'll let you know when I'm done testing @gjeanmart
Edit: I tested the whole flow and it works as expected, thanks for the fix 👍

PooyaRaki:
Since I've tested the flow throughly, I'll merge the change. cc @liliya-soroka

packages/utils/src/services/security/modules/BlockaidModule/index.ts
```diff
@@ -60,8 +60,12 @@ type BlockaidPayload = {
     params: [string, string]
   }
   options: ['simulation', 'validation']
+  state_override?: Record<string, { stateDiff?: Record<string, string> }>
 }
 
+// Safe Smart Account Storage Slot for Guard
+const GUARD_STORAGE_POSITION = '0x4a204f620c8c5ccdca3fd54d003badd85ba500436a431f0cbda4f558c93c34c8'
```
LucieFaire, line 67: @gjeanmart is it a static variable? Or should I read if from env config?

katspaugh: It's always the same for all Safe proxies, no need to parametrize it. Confirmed by @akshay-ap.

gjeanmart: Yeah, it's a magic number :)

PR #6504, opened 2025-10-17
refactor(web): use SafeInfo from store package

compojoom:
## What it solves
switch safeState to safeInfo from store package.
change getSafeInfo to hooks from store package.

## How to test it
Play around with the safe - creating a new safe, visiting settings, connecting disconnecting from a wallet. Nothng should have changed.

---

## CLA signature

### DIFF
--- a/apps/web/src/components/new-safe/create/logic/index.ts
+++ b/apps/web/src/components/new-safe/create/logic/index.ts
@@ -160,8 +160,7 @@ export const pollSafeInfo = async (chainId: string, safeAddress: string): Promis
         },
       )
 
-      // Cast to any to work around TypeScript limitation with thunk actions
-      const queryPromise = store.dispatch(queryAction as any)
+      const queryPromise = store.dispatch(queryAction)
       try {
         const result = await queryPromise.unwrap()
         return result
--- a/apps/web/src/store/index.ts
+++ b/apps/web/src/store/index.ts
@@ -5,8 +5,6 @@ import {
   type ThunkAction,
   type Action,
   type Middleware,
-  type EnhancedStore,
-  type ThunkDispatch,
 } from '@reduxjs/toolkit'
 import { useDispatch, useSelector, type TypedUseSelectorHook } from 'react-redux'
 import merge from 'lodash/merge'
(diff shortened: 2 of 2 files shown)


apps/web/src/components/new-safe/create/logic/index.ts
```diff
@@ -141,17 +141,44 @@ export const estimateSafeCreationGas = async (
...
+      const queryAction = cgwApi.endpoints.safesGetSafeV1.initiate(
+        { chainId, safeAddress },
+        {
+          subscribe: false,
+          forceRefetch: true,
+        },
+      )
+
+      // Cast to any to work around TypeScript limitation with thunk actions
+      const queryPromise = store.dispatch(queryAction as any)
```
katspaugh: 👀

apps/web/src/components/new-safe/create/logic/index.ts
```diff
@@ -141,17 +141,44 @@ export const estimateSafeCreationGas = async (
...
+/**
+ * Poll for safe info after creation until the safe is indexed by client-gateway
+ * Uses RTK Query with exponential backoff retry (19 attempts over ~4 minutes)
+ */
+export const pollSafeInfo = async (chainId: string, safeAddress: string): Promise<SafeState> => {
+  const { getStoreInstance } = await import('@/store')
+  const store = getStoreInstance()
+
+  // Use exponential backoff to retry RTK Query calls
+  return backOff(
```
katspaugh, line 153: Doesn't RTK query have its own polling feature?

compojoom: Yes, it does, but it is "static" meaning it would fetch every X seconds. The create logic here has an exponential backoff and I didn't want to replace it with RTK's poling.

apps/web/src/components/new-safe/create/logic/index.ts
compojoom: The problem was the makeStore type. When I implemented it last year I typed the return type (don't remember why) and because of that the RTK queryAction was not recognised as correct Action. By letting typescript infer the makeStore the problem here dissappeared.

PR #6509, opened 2025-10-20
refactor: use types from store package and remove safe-gateway-typescript-sdk

compojoom:
## What it solves
Last year we started a new package (@safe-global/store) that generates types and utility fetch functions with RTK query. The package turned out to be easy to maintain and a huge improvement over https://github.com/safe-global/safe-gateway-typescript-sdk . We used the package in the new mobile app and at places in the old app. We gradually migrated the parts that were not hard, but a lot of places remained. Relaying on 2 different packages meant that sharing code between web and mobile was hard. You couldn't easily just copy code to the utils package as a type that we relied on in web would be incompatible with the expectations in mobile. With this PR I've updated all the code to use the new package.
It looks huge at first, but if you review commit by commit you'll see that most of the changes were done with a codemod that just switched the type imports. After that most of the commits are each relatively small. 

Despite this we should still carefully review and test. The type changes revealed that we were running blind in some areas. We were expecting things that according to our type are impossible. I've tried to carefully make the necessary adjustments without modifying too much.

Resolves https://linear.app/safe-global/issue/COR-702/refactor-usage-of-safe-gateway-typescript-sdk

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/tx/security/tenderly/index.tsx
+++ b/apps/web/src/components/tx/security/tenderly/index.tsx
@@ -97,7 +97,7 @@ const TxSimulationBlock = ({
 
   const { currentData: nestedSafeInfo } = useSafesGetSafeV1Query(
     { chainId: chain?.chainId || '', safeAddress: nestedSafe || '' },
-    { skip: !chain?.chainId || !nestedSafe },
+    { skip: !chain?.chainId || !nestedSafe, refetchOnMountOrArgChange: true },
   )
 
   const handleSimulation = async () => {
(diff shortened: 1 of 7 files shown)


katspaugh:
Looks like unit tests are failing.

compojoom:
Tests are passing now. 
The failing test outlined a problem in our old code. When the preview would fail, we would still allow the user to continue. 

I updated the test and actually added error handling for this:
grafik

It's not ideal, but continuing without having any preview from the backend is worse.

katspaugh:
@claude it's a big refactoring, can you see any potential regressions?

katspaugh (approved):

PR #6522, opened 2025-10-22
feat(Safe Shield): integrate error status for single analysis

clovisdasilvaneto:
## What it solves
- Error states are not handled granularly per analysis type
- Users may not receive clear feedback about which specific analysis failed
- Successful analyses might be hidden when one analysis fails

Linear: https://linear.app/safe-global/issue/COR-701/error-states-in-safeshield-widget

## How this PR fixes it
Implemented a component-level error handling in the SafeShieldContent component that:

Catches errors from AsyncResult parameters for each analysis type

Generates a proper AnalysisResult object with FAILED status (needs to be added to the [AnyStatus](https://github.com/safe-global/safe-wallet-monorepo/blob/feature/safe-shield/packages/utils/src/features/safe-shield/types.ts#L88) type)

Displays user-friendly error messages in the appropriate group card

Allows successful analyses to display normally alongside failed ones

## Checklist

---

## CLA signature


apps/web/src/features/safe-shield/SafeShield.stories.tsx
```diff
@@ -7,17 +7,20 @@ import {
...
-          <Story />
-        </Box>
-      </Paper>
+      <StoreDecorator initialState={{}}>
+        <Paper sx={{ padding: 2, backgroundColor: 'background.main' }}>
+          <Box sx={{ width: 320 }}>
+            <Story />
+          </Box>
+        </Paper>
+      </StoreDecorator>
```
tmjssz, line 23: Would be nice to have a story for error cases here

apps/web/src/features/safe-shield/components/SafeShieldContent/index.tsx
```diff
@@ -52,26 +52,20 @@ export const SafeShieldContent = ({
       <Box
         sx={{ border: '1px solid', borderColor: 'background.main', borderTop: 'none', borderRadius: '0px 0px 6px 6px' }}
       >
-        {loading ? (
-          <SafeShieldAnalysisLoading />
-        ) : error ? (
+        {error ? (
```
tmjssz: I think we can remove the SafeShieldAnalysisError as we always show the errors in the individual components now, right?

packages/utils/src/features/safe-shield/hooks/useCounterpartyAnalysis.ts
```diff
@@ -100,23 +111,25 @@ export function useCounterpartyAnalysis({
...
     )
-  }, [counterpartyData?.recipient, addressBookCheck, activityCheck])
+  }, [counterpartyData?.recipient, addressBookCheck, activityCheck, fetchError, activityCheckError])
 
-  const fetchError = useMemo(() => {
-    if (error) {
-      return new Error('error' in error ? error.error : 'Failed to fetch counterparty analysis')
+  const contractResults = useMemo(() => {
+    if (fetchError) {
+      return { [safeAddress]: { [StatusGroup.CONTRACT_VERIFICATION]: [getErrorInfo(ErrorType.CONTRACT)] } }
```
tmjssz, line 118: It's added to StatusGroup.CONTRACT_VERIFICATION since it does not matter for the display in which status group the FAILED result is, correct? (Just thinking if it might make sense to add general status group or something like that to avoid confusion)

apps/web/src/features/safe-shield/components/AnalysisGroupCard/AnalysisGroupCard.tsx
```diff
@@ -58,19 +60,11 @@ export const AnalysisGroupCard = ({
         <Box sx={{ padding: '4px 12px 16px' }}>
           <Stack gap={2}>
             {visibleResults.map((result, index) => (
-              <Box key={index} bgcolor="background.main" borderRadius="4px" overflow="hidden">
-                <Box sx={{ borderLeft: `4px solid ${SEVERITY_COLORS[result.severity].main}`, padding: '12px' }}>
-                  <Stack gap={2}>
-                    <Typography variant="body2" color="primary.light">
-                      {result.description}
-                    </Typography>
+              <AnalysisGroupCardItem key={index} severity={result.severity} description={result.description}>
```
tmjssz, line 63: I like that it's a separate component now!

packages/utils/src/features/safe-shield/hooks/useCounterpartyAnalysis.ts
clovisdasilvaneto: yes, so far it doesn't matter

apps/web/src/features/safe-shield/SafeShield.stories.tsx
clovisdasilvaneto: We do have, for threat analysis fail, claude have done it on my previous PR :)

apps/web/src/features/safe-shield/components/SafeShieldContent/index.tsx
clovisdasilvaneto: Yes, this shouldn't be there, might have left it after the last rebase I did 😮‍💨  thanks for catching it.

PR #6525, opened 2025-10-23
feat: Add specialized error handling for Guard reverts (UnapprovedHash)

katspaugh:
## Summary [EPD-381](https://linear.app/safe-global/issue/EPD-381/tx-flow-show-a-friendlier-error-message-for-guard-reverts)

This PR improves error messaging when a transaction fails due to a Guard contract revert with the `UnapprovedHash` error (`0x70cc6907`).

### Gas estimation error
Screenshot 2025-10-23 at 16 13 31

### Execution error
Screenshot 2025-10-23 at 16 15 22

### Changes

- **Enhanced ErrorMessage component** to detect Guard errors and display a specialized message with a link to the Guard contract on the block explorer
- **Updated transaction execution error handling** in `ComboSubmit` to pass `context="execution"` prop
- **Updated transaction notification toasts** to show Guard-specific message instead of raw error
- **Added utility functions** to detect Guard errors (`isUnapprovedHashError`, `isGuardError`)
- **Added comprehensive unit tests** for Guard error detection (9 tests, all passing)

### User Experience

When a Guard revert occurs, users now see:

**In UI errors:**
```
This transaction will most likely fail. To save gas costs, avoid creating this transaction.

Guard [🔗] reverted the transaction (UnapprovedHash)
```
Where "Guard" is a clickable link with an external link icon that opens the Guard contract on the block explorer.

**In notification toasts:**
```
Title: [Transaction description]
Body: "Guard reverted the transaction (UnapprovedHash)."
```

### Coverage

This applies to all error display locations:
- ✅ Gas estimation errors (`ExecuteForm`, `SignOrExecuteForm`, `ReviewBatch`)
- ✅ Transaction execution errors (`ComboSubmit`)
- ✅ Transaction notification toasts (`useTxNotifications`)

### Technical Details

- The Guard address is retrieved from `useSafeInfo().safe.guard`
- The block explorer link is generated using the existing `getBlockExplorerLink()` utility
- Uses the `ExternalLink` component for consistent styling with external link icon
- The error detection checks `error.message`, `error.reason`, and `error.data` for the `0x70cc6907` error code

### Testing

- ✅ TypeScript type checking passes
- ✅ ESLint passes with no errors
- ✅ All unit tests pass (9/9)
- ✅ Prettier formatting applied
- ✅ Backward compatible - existing error handling unchanged


apps/web/src/utils/transaction-errors.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Guard error codes
+ */
+export const GUARD_ERROR_CODES = {
+  UNAPPROVED_HASH: '0x70cc6907',
+} as const
+
+/**
+ * Extracts the Guard error code from an error message
+ * @param error - The error to extract from
```
PooyaRaki: Nit: param type and the return type should also be in the doc block. There are other instances but I won't flag separately to keep comments consice.

apps/web/src/utils/transaction-errors.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Gets a human-readable error name from a Guard error code
+ * @param errorCode - The error code (e.g., '0x70cc6907')
+ * @returns Human-readable error name
+ */
+export const getGuardErrorName = (errorCode: string): string => {
+  switch (errorCode) {
+    case GUARD_ERROR_CODES.UNAPPROVED_HASH:
+      return 'UnapprovedHash'
```
PooyaRaki, line 50: Nit: We could make each key in the `GUARD_ERROR_CODES` constant an array that stores the error message next to the error code. This would remove the need to add a new switch case every time we handle a new error. However, we can address this as part of the larger refactor we discussed.

apps/web/src/utils/transaction-errors.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @returns The error code if found, undefined otherwise
+ */
+export const extractGuardErrorCode = (error: Error): string | undefined => {
+  if (!error) return undefined
+
+  const errorMessage = error.message || ''
+
+  // Check for each known guard error code in the message
+  for (const code of Object.values(GUARD_ERROR_CODES)) {
+    if (errorMessage.includes(code)) {
```
PooyaRaki, line 34: Nit: Even though we already discussed this, filtering for keywords in the error message still feels hacky - what if another error contains the same substring?

apps/web/src/utils/transaction-errors.ts
```diff
@@ -0,0 +1,63 @@
+/**
```
PooyaRaki, line 1: General note / nit: The methods should be grouped more granularly and placed next to each other to make the code easier to follow. Right now, it’s a bit hard to trace the flow since I have to jump around the file to find methods that are executed sequentially.

```ts
// These three methods are executed sequentially, so having them together would improve the reading flow:

getGuardErrorInfo()
extractGuardErrorCode()
getGuardErrorName()

isGuardError() // This one could go here or at the top
```

Whereas currently the order is:
```ts
extractGuardErrorCode()
getGuardErrorInfo() // When reading this, you have to scroll up to find the method it calls
isGuardError()
getGuardErrorName() // You have to scroll down (which is fine), but skipping one method to find it makes it harder to follow the flow
```

It’s not a big issue now since the file is small, but this could become confusing as the class grows.

apps/web/src/hooks/useTxNotifications.ts
```diff
@@ -62,7 +63,16 @@ const useTxNotifications = (): void => {
...
         const isSuccess = successEvents.includes(event)
-        const message = isError ? `${baseMessage} ${formatError(detail.error)}` : baseMessage
+
+        // Check if this is a Guard error
+        const guardErrorName = isError ? getGuardErrorInfo(detail.error) : undefined
+        let message = isError ? `${baseMessage} ${formatError(detail.error)}` : baseMessage
+
+        // Override message for Guard errors
+        if (guardErrorName) {
+          message = `Guard reverted the transaction (${guardErrorName}).`
```
PooyaRaki, line 73: Nit: Keeping it as a nit since this might be an edge case, but if we interact with a contract that has a method like `approveHash` or `unapproveHash`, and its error message includes something like “unapproved hash,” we could mistakenly flag it as a guard error. Maybe here we could also check if the Safe actually has a guard installed.

katspaugh: Yes, correct, we rely solely on the `0x70cc6907` string in the error message and it can come from another contract. If you have ideas how to make it more robust, I'd be happy to explore them.

PooyaRaki: Yeah I don't have any other elegant solution atm that's why I left it as a nitpick.

apps/web/src/utils/transaction-errors.ts
katspaugh: Addressed 👍

PR #6528, opened 2025-10-24
feat(Safe Shield): implement error status for a single analysis item

clovisdasilvaneto:
## What it solves
- Error states are not handled granularly per analysis type
- Users may not receive clear feedback about which specific analysis failed
- Successful analyses might be hidden when one analysis fails

Linear: https://linear.app/safe-global/issue/COR-701/error-states-in-safeshield-widget

# Screenshots
threat-analysis-failed

## How this PR fixes it
Implemented a component-level error handling in the SafeShieldContent component that:

Catches errors from AsyncResult parameters for each analysis type

Generates a proper AnalysisResult object with FAILED status (needs to be added to the [AnyStatus](https://github.com/safe-global/safe-wallet-monorepo/blob/feature/safe-shield/packages/utils/src/features/safe-shield/types.ts#L88) type)

Displays user-friendly error messages in the appropriate group card

Allows successful analyses to display normally alongside failed ones

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/features/safe-shield/components/AnalysisGroupCard/AnalysisGroupCard.tsx
+++ b/apps/web/src/features/safe-shield/components/AnalysisGroupCard/AnalysisGroupCard.tsx
@@ -16,9 +16,7 @@ export const AnalysisGroupCard = ({
   const [isOpen, setIsOpen] = useState(false)
 
   const visibleResults = useMemo(() => mapVisibleAnalysisResults(Object.values(data || [])), [data])
-  const primaryResult = useMemo(() => {
-    return visibleResults[0]
-  }, [visibleResults])
+  const primaryResult = useMemo(() => visibleResults[0], [visibleResults])
 
   if (!visibleResults.length) {
     return null


apps/web/src/features/safe-shield/components/AnalysisGroupCard/AnalysisGroupCard.tsx
```diff
@@ -15,8 +15,10 @@ export const AnalysisGroupCard = ({
 }): ReactElement | null => {
   const [isOpen, setIsOpen] = useState(false)
 
-  const visibleResults = useMemo(() => mapVisibleAnalysisResults(Object.values(data)), [data])
-  const primaryResult = useMemo(() => visibleResults[0], [visibleResults])
+  const visibleResults = useMemo(() => mapVisibleAnalysisResults(Object.values(data || [])), [data])
+  const primaryResult = useMemo(() => {
+    return visibleResults[0]
```
tmjssz: Nit: this change is not needed here, right? We could just leave it on one line

apps/web/src/features/safe-shield/components/__tests__/SafeShieldDisplay.test.tsx
```diff
@@ -98,12 +104,27 @@ describe('SafeShieldDisplay', () => {
 
     it('should show error message in content', () => {
       const error = new Error('Service unavailable')
-      const errorContract: [undefined, Error, false] = [undefined, error, false]
+      const errorContract: [ContractAnalysisResults, Error, false] = [
+        {
+          [mockContractAddress]: {
+            [StatusGroup.CONTRACT_VERIFICATION]: [
+              {
+                title: 'Contract analysis failed',
```
tmjssz: How about moving this to a new static function in `ContractAnalysisResultBuilder`

apps/web/src/features/safe-shield/components/SafeShieldContent/index.tsx
```diff
@@ -53,26 +51,14 @@ export const SafeShieldContent = ({
...
-        {loading ? (
-          <SafeShieldAnalysisLoading />
-        ) : error ? (
-          <SafeShieldAnalysisError error={error} />
-        ) : allEmpty ? (
-          <SafeShieldAnalysisEmpty />
-        ) : null}
+        {loading ? <SafeShieldAnalysisLoading /> : allEmpty ? <SafeShieldAnalysisEmpty /> : null}
 
         <Box display={loading || error ? 'none' : 'block'}>
```
tmjssz: I think we also need to remove error from here, otherwise the results (and component-specific error results) would not be rendered in case there is any error. Or am I missing something here?
```suggestion
        <Box display={loading ? 'none' : 'block'}>
```

clovisdasilvaneto: yes for sure we don't need it! More one leftover from last rebase 🥹 Thanks for catching

apps/web/src/features/safe-shield/components/__tests__/SafeShieldDisplay.test.tsx
clovisdasilvaneto: sounds good

apps/web/src/features/safe-shield/components/AnalysisGroupCard/AnalysisGroupCard.tsx
clovisdasilvaneto: think it was my prettier, will check

tmjssz (approved):
Nice work 🚀

PR #6548, opened 2025-10-29
fix(Safe Shield): hide SafeShield widget on specific flows

tmjssz:
## What it solves

Hides SafeShield widget on the following flows because these are special cases where we don't have transaction data that can be analyszed:
- ActivateAccount
- RecoveryAttempt
- SignMessage

Resolves [COR-762](https://linear.app/safe-global/issue/COR-762/safe-shield-no-scanning-for-the-safe-activation)

## How this PR fixes it
- Introduced hideSafeShield prop in TxLayout to control the visibility of the Safe Shield sidebar.

## How to test it
Open any of the flows mentioned above and verify that the SafeShield widget does not appear.

## Screenshots
Screenshot 2025-10-29 at 17 51 38

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/tx-flow/flows/SignMessage/index.tsx
+++ b/apps/web/src/components/tx-flow/flows/SignMessage/index.tsx
@@ -55,6 +55,7 @@ const SignMessageFlow = ({ ...props }: SignMessageProps) => {
       step={0}
       hideNonce
       isMessage
+      hideSafeShield
     >
       <ErrorBoundary fallback={<div>Error signing message</div>}>
         <SignMessage {...props} />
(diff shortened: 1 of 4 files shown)


apps/web/src/components/tx-flow/flows/SignMessage/index.tsx
```diff
@@ -55,6 +55,7 @@ const SignMessageFlow = ({ ...props }: SignMessageProps) => {
       step={0}
       hideNonce
       isMessage
+      hideSafeShield
```
LucieFaire, line 58: but I thought we should run threat analysis on sign message flow?

LucieFaire (approved):

apps/web/src/components/tx-flow/flows/SignMessage/index.tsx
tmjssz: I was assuming no, because this is the flow for **offchain** messages only. In the onchain message flow, we do show the widget. But will double check for offchain messages

tmjssz: Confirmed by Simon [here](https://linear.app/safe-global/issue/COR-762/safe-shield-no-scanning-for-the-safe-activation#comment-84fd60f4)

PR #6554, opened 2025-10-30
fix(Safe Shield): tenderly 

clovisdasilvaneto:
## What it solves
This PR address minor issues found in the tenderly integration:

- CSS Issues
- Header of the widget is not in sync with tenderly
- Use the right icon for severity status

## Screenshots
Screenshot 2025-10-30 at 12 45 47
Screenshot 2025-10-30 at 12 46 19

## Checklist

---

## CLA signature

---

> [!NOTE]
> Integrates Tenderly (including nested) simulation status into Safe Shield header, adds shared types/util, tweaks UI, and updates tests.
> 
> - **Safe Shield**:
>   - **Header**: Uses `TxInfoContext`, chain checks, and nested tx to compute `hasSimulationError`; `getOverallStatus` now considers simulation errors; `safeTx` passed into `SafeShieldHeader`.
>   - **Content UI**: Adds section dividers via `sx` and refines container borders.
>   - **TenderlySimulation**: Failure icon severity changed to `WARN`; button text color set to `text.primary`.
> - **Tenderly utils/types**:
>   - Add `isSimulationError` and `NestedTxStatus`; centralize `SimulationStatus` usage; export/import updates.
>   - `TxInfoProvider`: Provides nested simulation in context and uses shared `SimulationStatus`.
> - **Safe Shield utils**:
>   - `getOverallStatus` accepts `hasSimulationError` and injects a `WARN` result when simulation fails.
>   - Tests: Add case asserting `WARN` severity on simulation failure.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit efe59746ba3635662fffd034d424f6f93fd7bc45. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/web/src/features/safe-shield/components/SafeShieldContent/index.tsx
+++ b/apps/web/src/features/safe-shield/components/SafeShieldContent/index.tsx
@@ -1,5 +1,5 @@
 import { type ReactElement } from 'react'
-import { Box, Divider, Stack } from '@mui/material'
+import { Box } from '@mui/material'
 import type {
   GroupedAnalysisResults,
   ContractAnalysisResults,
@@ -59,9 +59,9 @@ export const SafeShieldContent = ({
       >
         {loading ? <SafeShieldAnalysisLoading /> : allEmpty ? <SafeShieldAnalysisEmpty /> : null}
 
-        <Stack
+        <Box
           display={loading ? 'none' : 'block'}
-          divider={<Divider sx={{ borderColor: 'background.main' }} flexItem />}
+          sx={{ '& > div:not(:last-child)': { borderBottom: '1px solid', borderColor: 'background.main' } }}
         >
           {recipientResults && <AnalysisGroupCard data={recipientResults} />}
 
@@ -70,7 +70,7 @@ export const SafeShieldContent = ({
           {normalizedThreatData && <AnalysisGroupCard data={normalizedThreatData} />}
 
           <TenderlySimulation safeTx={safeTx} />
-        </Stack>
+        </Box>
       </Box>
     </Box>
   )


apps/web/src/features/safe-shield/components/SafeShieldContent/index.tsx
```diff
@@ -50,11 +50,19 @@ export const SafeShieldContent = ({
...
+          borderRadius: '0px 0px 6px 6px',
+        }}
       >
         {loading ? <SafeShieldAnalysisLoading /> : allEmpty ? <SafeShieldAnalysisEmpty /> : null}
 
-        <Box display={loading ? 'none' : 'block'}>
+        <Box
+          display={loading ? 'none' : 'block'}
+          sx={{ '& > div:not(:last-child)': { borderBottom: '1px solid', borderColor: 'background.main' } }}
+        >
```
tmjssz: Just as a suggestion, we could use Stack with the divider prop here to avoid some crazy CSS magic :)
```suggestion
        <Stack
          display={loading ? 'none' : 'block'}
          divider={<Divider sx={{ borderColor: 'background.main' }} flexItem />}
        >
```

apps/web/src/components/tx-flow/TxInfoProvider.tsx
```diff
@@ -5,7 +5,7 @@ import { FETCH_STATUS } from '@safe-global/utils/components/tx/security/tenderly
 import type { UseSimulationReturn } from '@safe-global/utils/components/tx/security/tenderly/useSimulation'
 import { getSimulationStatus } from '@safe-global/utils/components/tx/security/tenderly/utils'
 
-type SimulationStatus = {
+export type SimulationStatus = {
```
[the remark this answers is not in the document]
tmjssz: Makes sense to reuse the type from the utils package

packages/utils/src/features/safe-shield/utils/getOverallStatus.ts
```diff
@@ -32,11 +32,16 @@ export const getOverallStatus = (
...
   threatResults?: ThreatAnalysisResults,
+  hasSimulationError?: boolean,
 ): { severity: Severity; title: string } | undefined => {
   if (!recipientResults && !contractResults && !threatResults) {
     return undefined
   }
 
+  if (hasSimulationError) {
+    return { severity: Severity.WARN, title: SEVERITY_TO_TITLE[Severity.WARN] }
+  }
```
tmjssz: This would not work correctly if there is a CRITICAL result as it would just return a WARN severity result and ignore the other results. Maybe better to add it to the `allResults` array and let the getPrimaryResult do the job to return the correct result for the header.

clovisdasilvaneto: oh true, good catch

packages/utils/src/features/safe-shield/utils/getOverallStatus.ts
```diff
@@ -32,6 +32,7 @@ export const getOverallStatus = (
   recipientResults?: RecipientAnalysisResults,
   contractResults?: ContractAnalysisResults,
   threatResults?: ThreatAnalysisResults,
+  hasSimulationError?: boolean,
```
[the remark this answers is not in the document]
tmjssz: That's true, I think we should extend the condition to only return early if `hasSimulationError` is false in addition to the other conditions.

apps/web/src/features/safe-shield/components/SafeShieldContent/index.tsx
tmjssz: Oh damn, I think we should revert this change to what you had before because now I am seeing some weird thick border at the bottom in the empty widget state. Sorry for the back-and-forth! 😅 
Screenshot 2025-10-31 at 14 10 47

tmjssz (commented):
Apart from the [border issue](https://github.com/safe-global/safe-wallet-monorepo/pull/6554#discussion_r2481365864), there is one more small thing that Cursor found, I think it makes sense ([see comment](https://github.com/safe-global/safe-wallet-monorepo/pull/6554#discussion_r2481311581)). Otherwise looks good to me!!

apps/web/src/features/safe-shield/components/SafeShieldContent/index.tsx
clovisdasilvaneto: damn hahahhaa didn't notice that as well

tmjssz (approved):
Nice 💪

PR #6555, opened 2025-10-30
fix: Include nonces with reject transactions in replace existing list

katspaugh:
## Summary
Fixes a bug where nonces with reject transactions were not appearing in the "replace existing" dropdown when creating a new transaction.

## Changes
- Updated `usePreviousNonces` hook to use `isTransactionQueuedItem` instead of `isTransactionListItem`
- The previous implementation filtered out transactions with `ConflictType` other than `NONE`, which excluded transactions that have reject transactions (marked as `HAS_NEXT` or `END`)
- Added comprehensive test coverage to verify nonces with reject transactions are now included

## Test plan
- [x] Unit tests pass for `usePreviousNonces`
- [x] TypeScript type checking passes
- [ ] Manual testing: Create a few transactions in the queue and create a reject transaction for one of them
- [ ] Manual testing: Start to create a new transaction and verify the nonce with the reject tx appears in the "replace existing" list

## Related
Fixes: https://linear.app/safe-global/issue/COR-766/nonce-is-not-counted-in-the-replace-existing-if-the-tx-has-reject-tx

### DIFF
--- a/apps/web/src/hooks/__tests__/usePreviousNonces.test.ts
+++ b/apps/web/src/hooks/__tests__/usePreviousNonces.test.ts
@@ -30,13 +30,17 @@ describe('_getUniqueQueuedTxs', () => {
   })
 
   it('includes transactions with reject txs (ConflictType.HAS_NEXT)', () => {
-    const mockTx = getMockTx({ nonce: 0 })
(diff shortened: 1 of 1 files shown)


compojoom (approved):
nice, thank you!

PR #6591, opened 2025-11-05
chore(build): use rspack for default dev command

vseehausen:
## What it solves

Uses `rspack` as the default `dev` command.

Also removed `yarn start` as a redundant command.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/README.md
+++ b/apps/web/README.md
@@ -77,6 +77,7 @@ If you don't provide some of the variables, the corresponding features will be d
 
 ### Running the app locally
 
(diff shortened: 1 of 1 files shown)


compojoom:
Can we please keep the start command? The mobile uses start to launch the dev server and it is going to annoy me when I have to switch the commands between the apps :)

katspaugh (approved):
Approved but agree we should restore the `start` command.

vseehausen:
Ok, good point.

PR #6629, opened 2025-11-11
Feat(Hypernative): HoC for the HnBanner using the local storage; HN banners on multiple pages (NewTx, Accounts)

vmosharova:
## Summary

- Adds a localStorage-based version of `HnBanner` that works on pages without SafeInfo (e.g., `/welcome/accounts`)
- Implements `HnBanner` with required conditional rendering to the following pages: New Transaction page, Accounts page (`/welcome/accounts`);
- Style fixes for the `HnPendingBanner`

## What it solves

- `HnBanner` doesn't render on `/wallets/account` because it depends on SafeInfo (chainId, safeAddress) via Redux
- Users can't see or interact with the Hypernative banner on the accounts page when SafeInfo isn't available

## How this PR fixes it

1. **Created `withLocalStorageVisibility` HOC** (`apps/web/src/features/hypernative/components/withLocalStorageVisibility/index.tsx`)
   - Reusable HOC that checks localStorage to control component visibility
   - Returns `null` if localStorage value is `false`, otherwise renders the wrapped component

2. **Created `HnBannerWithLocalStorage` component** (`apps/web/src/features/hypernative/components/HnBanner/HnBannerWithLocalStorage.tsx`)
   - Wrapper that uses `useLocalStorage` for dismissal state instead of Redux
   - Doesn't require SafeInfo (no chainId/safeAddress dependency)
   - Handles dismissal by setting localStorage to `false`

3. **Added `HnBannerWithLocalStorageVisibility` export** (`apps/web/src/features/hypernative/components/HnBanner/index.ts`)
   - Fully composed component with HOCs:
     - `withHnSignupFlow` (signup flow functionality)
     - `withLocalStorageVisibility` (localStorage visibility control)
     - `withHnFeature` (feature flag check)

4. **Integrated into `myAccounts` component** (`apps/web/src/features/myAccounts/index.tsx`)
   - Displays the banner on the accounts page when `!isSidebar` (line 31-35)
   - Only shows on the full accounts page, not in the sidebar drawer

The banner now works on pages without SafeInfo, using localStorage for visibility and dismissal state instead of Redux.

---

## CLA signature

### DIFF
--- a/apps/web/src/features/hypernative/components/HnSignupFlow/HnCalendlyStep.tsx
+++ b/apps/web/src/features/hypernative/components/HnSignupFlow/HnCalendlyStep.tsx
@@ -1,0 +1,77 @@
+import { Grid2 } from '@mui/material'
+import { useEffect, useRef } from 'react'
+import css from './styles.module.css'
+
+export type HnCalendlyStepProps = {
+  calendlyUrl: string
+}
+
+const HnCalendlyStep = ({ calendlyUrl }: HnCalendlyStepProps) => {
+  const calendlyContainerRef = useRef<HTMLDivElement>(null)
+  const calendlyScriptLoadedRef = useRef(false)
+
+  useEffect(() => {
+    // Load Calendly CSS
+    if (!document.querySelector('link[href*="calendly"]')) {
+      const link = document.createElement('link')
+      link.rel = 'stylesheet'
+      link.href = 'https://assets.calendly.com/assets/external/widget.css'
+      document.head.appendChild(link)
+    }
+
+    // Load Calendly script
+    if (!calendlyScriptLoadedRef.current && !document.querySelector('script[src*="calendly"]')) {
+      const calendlyScript = document.createElement('script')
+      calendlyScript.src = 'https://assets.calendly.com/assets/external/widget.js'
+      calendlyScript.async = true
+      document.body.appendChild(calendlyScript)
+      calendlyScriptLoadedRef.current = true
+    }
+  }, [])
+
+  useEffect(() => {
+    if (!calendlyContainerRef.current || !calendlyUrl) {
+      return
+    }
+
+    const startCalendly = () => {
+      if (window.Calendly && calendlyContainerRef.current) {
+        window.Calendly.initInlineWidget({
+          url: calendlyUrl,
+          parentElement: calendlyContainerRef.current,
+        })
+      }
+    }
+
+    if (window.Calendly?.initInlineWidget) {
+      startCalendly()
+    } else {
+      // Poll for Calendly to be available
+      const interval = setInterval(() => {
+        if (window.Calendly?.initInlineWidget) {
+          clearInterval(interval)
+          startCalendly()
+        }
+      }, 100)
+
+      // Cleanup interval if component unmounts
+      return () => clearInterval(interval)
+    }
+  }, [calendlyUrl])
+
+  return (
+    <Grid2 container className={css.container}>
+      {/* Left Column - Calendly */}
+      <Grid2 size="grow" className={css.calendlyColumn}>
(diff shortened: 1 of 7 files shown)


apps/web/src/components/settings/SecurityLogin/index.tsx
```diff
@@ -3,17 +3,20 @@ import dynamic from 'next/dynamic'
...
 
 const SecurityLogin = () => {
   const isRecoverySupported = useIsRecoverySupported()
   const router = useRouter()
+  const { showBanner: showHnBanner, loading: hnLoading } = useBannerVisibility(BannerType.Promo)
 
   return (
     <Box display="flex" flexDirection="column" gap={2}>
-      <HnBanner isDismissable={false} />
+      {(showHnBanner || hnLoading) && <HnBannerForCarousel onDismiss={() => {}} />}
```
katspaugh: Why this change?

apps/web/src/features/hypernative/components/withLocalStorageVisibility/index.tsx
katspaugh, line 1: This HoC isn't needed, you can encode the same functionality directly in `HnBannerWithLocalStorage.tsx`.

apps/web/src/features/myAccounts/index.tsx
```diff
@@ -27,6 +28,12 @@ const MyAccounts = ({ safes, onLinkClick, isSidebar = false }: MyAccountsProps)
       <Box className={classNames(css.myAccounts, { [css.sidebarAccounts]: isSidebar })}>
         <AccountsHeader isSidebar={isSidebar} onLinkClick={onLinkClick} />
 
+        {!isSidebar && (
+          <Box sx={{ mb: 2 }}>
```
katspaugh: This Box should be inside `HnBannerWithLocalStorageVisibility`, otherwise it will be creating empty space when the banner is not shown.

apps/web/src/components/settings/SecurityLogin/index.tsx
vmosharova: My bad - the banner in the setting should not be dismissable. Switching to the default export (HOC-wrapped).

apps/web/src/components/settings/SecurityLogin/index.tsx
```diff
@@ -3,17 +3,20 @@ import dynamic from 'next/dynamic'
...
+import { useBannerVisibility } from '@/features/hypernative/hooks'
+import { BannerType } from '@/features/hypernative/hooks/useBannerStorage'
+import HnBannerDefault from '@/features/hypernative/components/HnBanner'
 
 const RecoverySettings = dynamic(() => import('@/features/recovery/components/RecoverySettings'))
 
 const SecurityLogin = () => {
   const isRecoverySupported = useIsRecoverySupported()
   const router = useRouter()
+  const { showBanner: showHnBanner, loading: hnLoading } = useBannerVisibility(BannerType.Promo)
```
katspaugh: This should be reverted.

apps/web/src/features/hypernative/components/HnBanner/HnBannerWithLocalStorage.tsx
```diff
@@ -0,0 +1,32 @@
+import type { ReactElement } from 'react'
+import { Box } from '@mui/material'
+import useLocalStorage from '@/services/local-storage/useLocalStorage'
+import type { WithHnSignupFlowProps } from '../withHnSignupFlow'
+import { HnBanner } from './HnBanner'
+import { HN_BANNER_LS_KEY } from './constants'
```
katspaugh: This constant is used only in this file. No need to export it from a separate module. It could be just defined right here as a local const.

PR #6633, opened 2025-11-11
Fix(Hypernative): Old banners instead of the new PromoBanner

vmosharova:
## What it solves

Regression testing showed that new generic `PromoBanner` was breaking older banners (`NoFeeNovmberBanner`, `SpacesBanner`, `StakesBanner`, `EarnBanner`). This PR brings back the old implementation as fixing the generic banner might take longer time and affect the upcoming release.

## How this PR fixes it

Bringing back the old banners implementation and styles.

---

## CLA signature


katspaugh (approved):
👍

PR #6635, opened 2025-11-12
feat(Hypernative): Update TxReportButton visibility conditions

katspaugh:
## Summary
- Update the "Review security report" button in TxDetails to show based on new conditions
- Button now shows if banner conditions are met OR if Hypernative guard is active

## Changes
- Added `BannerType.TxReportButton` enum value to distinguish button visibility logic from banner logic
- Updated `useBannerStorage` to always return `true` for `TxReportButton` (ignores banner dismissal and form completion state)
- Updated `useBannerVisibility` to implement OR logic: show TxReportButton if (banner conditions met) OR (guard is installed)
- Simplified `HnSecurityReportBtn` HOC composition to use `withHnBannerConditions` with proper TypeScript typing
- Updated `withHnBannerConditions` generic constraint to accept any object props instead of restricting to `WithHnBannerConditionsProps`

## Visibility Logic
The button now shows when:
1. **(Feature enabled AND user is owner AND balance > threshold AND !guard) OR (guard is installed)**

This means the button will be visible in more scenarios:
- Before guard installation: Shows based on banner conditions (owner, balance, etc.)
- After guard installation: Always shows (regardless of other conditions)

## Test Plan
- [x] Type check passes
- [x] Linter passes  
- [x] All tests pass (12 test suites, 121 tests)
- [ ] Manual testing: Verify button shows correctly in both scenarios (banner conditions OR guard active)

### DIFF
--- a/apps/web/src/features/hypernative/hooks/useBannerVisibility.ts
+++ b/apps/web/src/features/hypernative/hooks/useBannerVisibility.ts
@@ -63,7 +63,7 @@ export const useBannerVisibility = (bannerType: BannerType): BannerVisibilityRes
 
     // For TxReportButton, show if banner conditions are met OR if guard is installed
     if (bannerType === BannerType.TxReportButton) {
-      const bannerConditionsMet = isEnabled && isSafeOwner && hasSufficientBalanceCheck && !isHypernativeGuard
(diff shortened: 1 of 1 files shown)


liliya-soroka:
Related to [COR-771](https://linear.app/safe-global/issue/COR-771/add-cta-on-transaction-queue)

PR #6640, opened 2025-11-12
Fix(Hypernative): Banners' conditional rendering on the Accounts page and First Steps page; new tx security report URL

vmosharova:
## What it solves

This PR fixes several issues with Hypernative banner conditional rendering:

1. **Banner visibility for disconnected wallets**: Hypernative banners were showing even when no wallet was connected, which is incorrect since the banners require wallet connection to function properly.

2. **Dashboard banner for undeployed safes**: The Hypernative dashboard banner was not showing for undeployed safes because it required a minimum balance check, but undeployed safes don't have indexed balances yet.

3. **Missing referrer tracking**: The security report button was not including referrer information, making it difficult to track where users are coming from.

Resolves: [this ticket](https://linear.app/safe-global/issue/COR-767/banner-0-state)

## How this PR fixes it

1. **Wallet connection checks**: 
   - Added wallet connection check in `MyAccounts` component so `HnBannerWithLocalStorageVisibility` only renders when a wallet is connected
   - The banner visibility logic now explicitly checks for wallet connection

2. **Dashboard banner for undeployed safes**:
   - Created a custom wrapped version of `HnDashboardBanner` (`HnFinalDashboardBanner`) that skips the balance-checking HOC
   - Implemented custom visibility logic in `FirstSteps` that checks all conditions (feature flag, banner storage, wallet connection, Safe ownership, guard status) except balance
   - This allows the banner to show for undeployed safes when an owner is connected, regardless of balance

3. **Referrer tracking**:
   - Added `referrer=safe` query parameter to the security report URL in `HnSecurityReportBtn` to track that users are coming from Safe wallet --> the latest change in the Hypernative URL for tx safety report

## Files changed

- `apps/web/src/features/myAccounts/index.tsx` - Added wallet connection check before rendering `HnBannerWithLocalStorageVisibility`
- `apps/web/src/features/hypernative/components/HnSecurityReportBtn/HnSecurityReportBtn.tsx` - Added `referrer=safe` query parameter to security report URL
- `apps/web/src/components/dashboard/FirstSteps/index.tsx` - Created custom banner visibility logic without balance check for undeployed safes and wrapped banner component without balance-checking HOC

---

## CLA signature


apps/web/src/components/dashboard/FirstSteps/index.tsx
```diff
@@ -26,9 +26,18 @@ import css from './styles.module.css'
...
+import { withHnFeature } from '@/features/hypernative/components/withHnFeature'
+import { BannerType, useBannerStorage } from '@/features/hypernative/hooks/useBannerStorage'
+import { useIsHypernativeFeature } from '@/features/hypernative/hooks/useIsHypernativeFeature'
+import { useIsHypernativeGuard } from '@/features/hypernative/hooks/useIsHypernativeGuard'
+import useWallet from '@/hooks/wallets/useWallet'
+import useIsSafeOwner from '@/hooks/useIsSafeOwner'
+
+// A banner version without the balance-checking as we can not index the balance till the safe is deployed
+const HnDashboardBannerWithSignup = withHnSignupFlow(HnDashboardBanner)
+const HnFinalDashboardBanner = withHnFeature(HnDashboardBannerWithSignup)
```
katspaugh: The idea was to export wrapped components like this from the index file of the corresponding component.

apps/web/src/components/dashboard/FirstSteps/index.tsx
```diff
@@ -362,7 +371,20 @@ const FirstSteps = () => {
...
+  const isEnabled = useIsHypernativeFeature()
+  const shouldShowBanner = useBannerStorage(BannerType.Promo)
+  const { isHypernativeGuard, loading: guardLoading } = useIsHypernativeGuard()
+
+  const showHnDashboardBanner = useMemo(() => {
+    if (guardLoading) {
+      return false
+    }
+    return isEnabled && shouldShowBanner && !!wallet && isSafeOwner && !isHypernativeGuard
+  }, [isEnabled, shouldShowBanner, wallet, isSafeOwner, isHypernativeGuard, guardLoading])
```
katspaugh: Isn't this covered by `useBannerVisibility`?

vmosharova: `useBannerVisibility` checks the balance in addition to all these conditions. But on the First Steps page we can't index the balance until the safe is deployed, so the banner won't be rendered if we use this hook. The banner should displayed for all not active Safes regardless the balance as per [this thread in the ticket](https://linear.app/safe-global/issue/COR-767/banner-0-state#comment-5631b55e).
Happy to discuss it and find a more elegant solution!

PR #6646, opened 2025-11-13
feature(portfolio): handle empty token uri and display placeholder icon

vseehausen:
## What it solves

Resolves: https://linear.app/safe-global/issue/GRO-124/resolve-bug-missing-tokenuri-image-tag

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/common/TokenIcon/index.tsx
+++ b/apps/web/src/components/common/TokenIcon/index.tsx
@@ -21,7 +21,7 @@ const TokenIcon = ({
   chainId?: string
 }): ReactElement => {
   const src = useMemo(() => {
-    return upgradeCoinGeckoThumbToQuality(logoUri, 'small')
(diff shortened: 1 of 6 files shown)


apps/web/src/components/common/IframeIcon/index.tsx
```diff
@@ -25,10 +25,11 @@ const IframeIcon = ({
   width?: number
   height?: number
 }): ReactElement => {
+  const imageSrc = src || '/images/common/token-placeholder.svg'
```
tmjssz: Do we really want to use this as a fallback icon? I think this component is used for other features as well (e.g. earn). Maybe better to pass the fallback image src as a prop as well?

tmjssz (approved):

vseehausen:
Reverted and reapplied last commit to trigger CI re-runs.

vseehausen:
Reverted and reapplied last commit to trigger CI re-runs.

PR #6665, opened 2025-11-17
fix(mobile): missing data row in data tab

compojoom:
## What it solves
The actual tx data was missing from the "data" tab. It was present under parameters, but since we are not showing any parameters on the review-and-execute it looked weird.

Resolves https://linear.app/safe-global/issue/COR-866/tx-data-missing-on-confirm-view

## How this PR fixes it
I've moved the data form the parameters tab to the data tab. This way it's available everywhere where we display "data" related to the tx

## How to test it
Open a pending tx and try to execute it. Observe that on the review and execute screen the data or '0x' is visible.

## Screenshots
image

image

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/components/HexDataDisplay/HexDataDisplay.tsx
+++ b/apps/mobile/src/components/HexDataDisplay/HexDataDisplay.tsx
@@ -49,7 +49,7 @@ export const HexDataDisplay = ({
   return (
     <InfoSheet title={title} info={data}>
       <View flexDirection="row" alignItems="center" gap={gap}>
-        <Text>{shortenText(data, characterLimit)}</Text>
+        <Text>{data.length > characterLimit ? shortenText(data, characterLimit) : data}</Text>
         <CopyButton value={data} color={copyButtonColor} text={copyMessage} />
       </View>
     </InfoSheet>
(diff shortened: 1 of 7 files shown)


apps/mobile/src/components/HexDataDisplay/HexDataDisplay.tsx
```diff
@@ -49,7 +49,7 @@ export const HexDataDisplay = ({
   return (
     <InfoSheet title={title} info={data}>
       <View flexDirection="row" alignItems="center" gap={gap}>
-        <Text>{shortenText(data, characterLimit)}</Text>
+        <Text>{data.length > characterLimit ? shortenText(data, characterLimit) : data}</Text>
```
clovisdasilvaneto, line 52: shouldn't it be done inside the `shortenText` function? Otherwise we will need to do it every place we call the shortenText function

clovisdasilvaneto (approved):
LGTM just a small question

apps/mobile/src/components/HexDataDisplay/HexDataDisplay.tsx
compojoom: I wondered whether I should put it in the shortenText function. Didn't want to change the behavior everywhere.

PR #6666, opened 2025-11-17
Fix(Hypernative): No duplications in the tracking events; Non-dismissable Settings banner

vmosharova:
**What it solves**
- The HN banner on the Settings page remains visible when a user dismisses other banners (but respects when the Hypernative Guard is installed or the form is completed);
- Fix in sending a tracking event on Security Report button click (now the event is firing once without duplication)

**Resolves:**
- https://linear.app/safe-global/issue/COR-849/safe-shield-hn-hn-banner-disappears-from-the-settings-if-its-closed-on 
- https://linear.app/safe-global/issue/COR-850/safe-shield-is-it-double-counting-for-the-hn-report

**How this PR fixes it**
- Banner visibility: Add a new BannerType.Settings that ignores dismissal state but respects formCompleted, allowing the Settings banner to remain visible when dismissed elsewhere while still hiding after form submission. Create a separate HnBannerForSettings export for the Settings page to avoid impacting Promo banner behavior on dashboard and other pages. The Settings banner now properly hides when formCompleted is true or when Hypernative Guard is installed, matching the behavior of other banners.
- Mixpanel events duplication: Prevents duplicate Mixpanel events by adding a debounce timeout in the Track component to block rapid duplicate tracking calls.

### DIFF
--- a/apps/web/src/features/hypernative/hooks/useBannerStorage.ts
+++ b/apps/web/src/features/hypernative/hooks/useBannerStorage.ts
@@ -40,10 +40,11 @@ export const useBannerStorage = (bannerType: BannerType): boolean => {
     if (!safeHnState) {
       // If no state exists, show promo banner by default, hide pending banner
       // For Settings, show if no state exists (no form completed yet)
-      if (bannerType === BannerType.Settings) {
-        return true
-      }
-      return bannerType === BannerType.Promo || bannerType === BannerType.NoBalanceCheck
+      return (
+        bannerType === BannerType.Settings ||
+        bannerType === BannerType.Promo ||
+        bannerType === BannerType.NoBalanceCheck
+      )
     }
 
     // Settings banner ignores dismissal state but respects formCompleted


apps/web/src/features/hypernative/hooks/useBannerStorage.ts
```diff
@@ -37,9 +39,18 @@ export const useBannerStorage = (bannerType: BannerType): boolean => {
 
     if (!safeHnState) {
       // If no state exists, show promo banner by default, hide pending banner
+      // For Settings, show if no state exists (no form completed yet)
+      if (bannerType === BannerType.Settings) {
+        return true
+      }
       return bannerType === BannerType.Promo || bannerType === BannerType.NoBalanceCheck
```
katspaugh:
```suggestion
return bannerType === BannerType.Settings || bannerType === BannerType.Promo || bannerType === BannerType.NoBalanceCheck
```

katspaugh (approved):
Looks good, just one style suggestion.

PR #6684, opened 2025-11-19
feat(web): add address book and etherscan to the ShowAddress component

clovisdasilvaneto:
## What it solves
It display addresses name in the Safe Shield widget

Resolves: https://linear.app/safe-global/issue/COR-827/display-addresses-name-in-the-safe-shield-widget

## How this PR fixes it
It just reuse the shared hooks that we have in the system for addressbook, copy to clipboard and etherscan.

## How to test it
- Try to send a transaction with at least 2 address 
- Check the recipient analysis in the safe shield widget

**Expected**
The recipient information should be there together with the etherscan link.

## Screenshots
Screenshot 2025-11-19 at 10 57 02

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/features/safe-shield/components/AnalysisGroupCard/ShowAllAddress.tsx
+++ b/apps/web/src/features/safe-shield/components/AnalysisGroupCard/ShowAllAddress.tsx
@@ -100,7 +100,9 @@ export const ShowAllAddress = ({ addresses }: ShowAllAddressProps) => {
                       {item}
                     </Typography>
                   </Tooltip>
-                  {explorerLink && <ExplorerButton href={explorerLink.href} />}
+                  <Box component="span" color="text.secondary">
+                    {explorerLink && <ExplorerButton href={explorerLink.href} />}
+                  </Box>
                 </Typography>
               </Box>
             )


apps/web/src/features/safe-shield/components/AnalysisGroupCard/ShowAllAddress.tsx
```diff
@@ -48,13 +67,44 @@ export const ShowAllAddress = ({ addresses }: ShowAllAddressProps) => {
...
+                      sx={{
+                        cursor: 'pointer',
+                        transition: 'background-color 0.2s',
+                        overflowWrap: 'break-word',
+                        wordBreak: 'break-all',
+                        flex: 1,
+                        '&:hover': {
+                          color: 'text.primary',
+                        },
+                      }}
```
compojoom, line 98: Isn't this inline object going to cause a re-render for the typography component?

clovisdasilvaneto: It is but that is how we're doing in other places, isn't it? Also it is not a performance problem in this case, since it is just a text rendered inside of it

PR #6687, opened 2025-11-19
Fix(web): Default background in promo banners

vmosharova:
## What it solves

The `PromoBanner` component's default background gradient (defined in CSS modules) was not appearing in the UI because MUI's `Card` component default styles were overriding it.

## How this PR fixes it

- Added a `DEFAULT_BACKGROUND` constant with the gradient value
- Applied the default background via the `sx` prop when `customBackground` is not provided
- This ensures the gradient appears by default while still allowing `customBackground` to override it
- The `!important` flag ensures it overrides MUI's default Card background styles

## How to test it
0. ! Important: In the Wallet web app, verify that the Hypernative banner on the dashboard and in the settings is not affected by this change as `HnBanner` is based on `PromoBanner`.
1. Open Storybook and view the `PromoBanner` stories
2. Verify that banners without `customBackground` prop show the default gradient (`linear-gradient(90deg, #b0ffc9, #d7f6ff)`)
3. Verify that banners with `customBackground` prop still use the custom background
4. Check that the background appears correctly in both light and dark modes

## Screenshots

Before:
image

After:
image

---

## CLA signature

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -1,3 +1,4 @@
+import { ExpoConfig, ConfigContext } from 'expo/config'
 /* eslint-disable no-undef */
 const IS_DEV = process.env.APP_VARIANT === 'development'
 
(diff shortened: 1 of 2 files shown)


vmosharova:
@claude review

vmosharova:
Cypress tests are failing due to `520 - Web Server Returned an Unknown Error` message from Cloudflare. I assume it is not on our side but connected with the Cloudflare outage.

PR #6694, opened 2025-11-20
refactor(hypernative): Use ABI-based detection instead of bytecode hash

katspaugh:
## Summary

Resolves https://linear.app/safe-global/issue/COR-883/fix-hypernative-guard-detection-by-switching-to-abi-matching

Refactors HypernativeGuard detection to use ABI-based verification instead of hardcoded bytecode hash comparison.

## Problem

The previous implementation checked a hardcoded bytecode hash, which doesn't work because each HypernativeGuard deployment may have unique bytecode.

## Solution

- Extract all function selectors from the ABI using `ethers.Interface`
- Check if the deployed bytecode contains all expected selectors
- Only requires one RPC call (`getCode`)

## Feature Flag

Added `HYPERNATIVE_RELAX_GUARD_CHECK` feature flag:
- When enabled, skips ABI verification and accepts any guard
- Provides fallback if ABI-based detection has issues

## Benefits

- Zero maintenance - automatically extracts all 34 function selectors from ABI
- More robust than bytecode hash comparison
- Works for any HypernativeGuard deployment


katspaugh:
## Final Implementation - Using ethers.Interface ✨

Simplified even further by using `ethers.Interface` to extract function selectors automatically from the ABI:

```typescript
const HYPERNATIVE_GUARD_INTERFACE = new Interface(HypernativeGuardAbi)

export const HYPERNATIVE_GUARD_FUNCTION_SELECTORS = HYPERNATIVE_GUARD_INTERFACE.fragments
  .filter((fragment): fragment is FunctionFragment => fragment.type === 'function')
  .map((fragment) => HYPERNATIVE_GUARD_INTERFACE.getFunction(fragment.name)!.selector.toLowerCase())
```

### Benefits
- **Zero maintenance**: Automatically extracts all 34 function selectors from ABI
- **Uses ethers built-in parsing**: No manual signature construction
- **Follows existing pattern**: Similar to `contractImplementsAbi` helper approach
- **Type-safe**: Uses `FunctionFragment` type

All 124 tests pass ✅

katspaugh:
@codex review.

katspaugh:
## Updated based on feedback ✅

### 1. Renamed Feature Flag
- `HYPERNATIVE_NO_ABI_CHECK` → `HYPERNATIVE_RELAX_GUARD_CHECK`
- Better describes what it does (relaxes the guard check to accept any guard)

### 2. Fixed Misleading Comments
Removed the comparison to `APPROVAL_SIGNATURE_HASH` which was misleading:
- **ERC20 approval detection**: Uses `tx.data.startsWith(APPROVAL_SIGNATURE_HASH)` to check transaction data **prefix**
- **Our bytecode inspection**: Uses `bytecode.includes(selector)` to search for selectors **anywhere** in deployed bytecode

The updated comments now correctly describe that we "inspect deployed bytecode for function selectors" without the misleading comparison.

### 3. Removed Unverified Assumption
Removed the statement about deployment bytecode containing the Safe address, as we're not 100% certain of this implementation detail.

All 124 tests pass ✅

PR #6674, opened 2025-11-18
refactor: migrate state management to RTK Query

katspaugh:
## Summary
Migrate core data fetching from Redux slices to RTK Query hooks for improved caching, automatic refetching, and reduced boilerplate. This migration removes the custom `makeLoadableSlice` pattern and `useLoadableStores` in favor of RTK Query's built-in data fetching capabilities.

### Migrated to RTK Query ✅
- **useChains**: Now uses `useChainsGetChainsV1Query` from auto-generated API
- **useSafeInfo**: Uses `useSafesGetSafeV1Query` with computed `deployed` property
- **useBalances**: Uses `useBalancesGetBalancesV1Query` for balance data
- **useTxHistory**: Uses `useTransactionsGetTransactionsHistoryV1Query` with updated parameter names (`trusted`/`imitation` instead of `hideUntrusted`/`hideImitation`)
- **useTxQueue**: Uses `useTransactionsGetTransactionQueueV1Query` for queue data
- **useSafeMessages**: Uses `useMessagesGetMessagesBySafeV1Query` for messages

### Removed Files 🗑️
- **Deleted Redux slices**: chainsSlice, safeInfoSlice, balancesSlice, txHistorySlice, txQueueSlice, safeMessagesSlice
- **Removed loadable hooks**: useLoadChains, useLoadSafeInfo, useLoadTxHistory, useLoadTxQueue, useLoadSafeMessages
- **Removed**: useLoadableStores hook (no longer needed)

### Spending Limits 📊
- Kept spending limits using Redux slice (not migrated to RTK Query per user request)
- Created `useSpendingLimitsLoader` hook to load spending limits data into Redux
- Added hook to InitApp in _app.tsx to populate slice on startup
- Updated useLoadSpendingLimits to extract tokenInfo from Balance items

### API Endpoint Updates 🔄
- Fixed endpoint name: `transactionsGetTransactionsHistoryV1` (was transactionsGetTransactionHistoryV1)
- Updated API parameter names: `trusted`/`imitation` (were hideUntrusted/hideImitation with inverted logic)
- Fixed cgwClient references to use correct cgwApi exports from AUTO_GENERATED files

### Type Fixes 🔧
- Added `deployed` property computation in useSafeInfo (true for fetched Safes)
- Fixed Balance type access patterns (tokenInfo.address instead of address)
- Added type guards where needed (isTransactionListItem in ReplaceTx)
- Fixed listener matchers to use correct API endpoints

### Store Updates 🏪
[...]

PR #6723, opened 2025-11-26
refactor: Remove balancesSlice from Redux store

katspaugh:
## Summary

- Migrates balances loading to use `useLoadBalances` hook directly instead of storing in Redux
- Follows the same pattern as PR #6674 for `useSafeMessages`
- Removes the `balancesSlice` from the Redux store

## Changes

- Update `useBalances` to call `useLoadBalances` directly
- Update `usePositions` to get positions from `useLoadBalances`
- Update `useLoadSpendingLimits` to get tokens from `useLoadBalances`
- Update `SpendingLimits` component to use `useBalances` hook
- Remove `balancesSlice` from `useLoadableStores`, `store/index.ts`, `slices.ts`
- Delete `balancesSlice.ts`
- Update tests to mock `useLoadBalances` instead of Redux state

## Test plan

- [x] Type check passes
- [x] All affected unit tests pass
- [x] Lint passes

### DIFF
--- a/apps/web/src/hooks/useBalances.ts
+++ b/apps/web/src/hooks/useBalances.ts
@@ -1,5 +1,4 @@
-import { useRef } from 'react'
-import isEqual from 'lodash/isEqual'
+import { useMemo } from 'react'
 import useLoadBalances, { type PortfolioBalances, initialBalancesState } from './loadables/useLoadBalances'
 
 export type UseBalancesResult = {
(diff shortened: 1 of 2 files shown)


katspaugh:
@codex review

apps/web/src/hooks/useBalances.ts
```diff
@@ -1,27 +1,35 @@
...
-  return useMemo(
-    () => ({
-      balances: data,
-      error,
-      loaded,
-      loading,
-    }),
-    [data, error, loaded, loading],
-  )
+  return resultRef.current
```
compojoom: What is going on here? Why do we need a ref?

katspaugh: Because balances are polled but might not actually change across polls, we do a deep compare with `isEqual` and not update the returned object if nothing has changed.

Now that I think about it though, it's probably always going to be slightly different due to price fluctuations. So we could remove this.

PR #6725, opened 2025-11-26
chore: improve pre-commit hooks and replace deprecated next lint

katspaugh:
## Summary
- Replace deprecated `next lint` with standalone `eslint src` (next lint deprecated in Next.js 16)
- Remove `npx lint-staged` dependency from pre-commit hook
- Run prettier directly on staged files in pre-commit
- Run eslint --fix on staged web and mobile files in pre-commit
- Handle mobile icon types generation in pre-commit
- Optimize to only call `git diff --cached` once

## Test plan
- [ ] Stage a file with formatting issues and verify prettier formats it on commit
- [ ] Stage a file with lint issues and verify eslint fixes them on commit
- [ ] Run `yarn workspace @safe-global/web lint` and verify eslint works
- [ ] Verify type-check still runs on staged web TypeScript files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6734, opened 2025-11-27
chore: pre-commit hook and linting config

compojoom:
## What it solves
The fix in #6725 in my opinion goes around the main issue. I think the underlying problem was in the fact that the "prepare" lifecycle hook is not being executed in the context of yarn berry. The only lifecycle hook that yarn is giving us here is 'postinstall'. Since the prepare hook was not being triggered on install most people were basically running without husky. Since husky was not running - no lint staged rules were applied. 

I moved prepare to postinstall and now on running yarn husky is being configured. I've updated the lintstaged config to run both prettier --fix, eslint --fix and type-check as well. 
I've setup eslint and type-check on the utils and store packages as well. I've configured them to use the eslitn config from nextjs, since those packages were basically born from the web package and I didn't want to have yet another custom config. 

With those modifications in my tests lint works just fine. 

image

## Checklist

---

## CLA signature

### DIFF
--- a/packages/store/eslint.config.mjs
+++ b/packages/store/eslint.config.mjs
@@ -1,7 +1,6 @@
-import { defineConfig, globalIgnores } from 'eslint/config'
 import nextVitals from 'eslint-config-next/core-web-vitals'
 
-const eslintConfig = defineConfig([
+const eslintConfig = [
   ...nextVitals,
   {
     rules: {
@@ -11,17 +10,9 @@ const eslintConfig = defineConfig([
       'react-hooks/refs': 'off',
     },
   },
-  // Override default ignores of eslint-config-next.
-  globalIgnores([
-    // Default ignores of eslint-config-next:
-    '.next/**',
-    'out/**',
-    'build/**',
-    'next-env.d.ts',
-  ]),
   {
     ignores: ['**/node_modules/'],
   },
-])
+]
 
 export default eslintConfig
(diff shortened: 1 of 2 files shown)


.husky/pre-commit
```diff
@@ -1,72 +1,4 @@
...
-fi
-
-# Run type-check on staged TypeScript files in apps/web
-staged_web_ts_files=$(echo "$all_staged" | grep -E '^apps/web/.*\.(ts|tsx)$' || true)
-
-if [ -n "$staged_web_ts_files" ]; then
-  echo "Running type-check for web app..."
-  yarn workspace @safe-global/web type-check
-fi
+npx lint-staged
```
katspaugh: Should we properly install `lint-staged` as a dev dependency? Otherwise it will always fetch the latest package which might be compromised.

package.json
```diff
@@ -14,7 +14,7 @@
     "eslint": "yarn workspaces foreach --all -pt run eslint",
     "prettier": "prettier --check \"**/*.{js,jsx,ts,tsx,json,md}\"",
     "prettier:fix": "prettier --write \"**/*.{js,jsx,ts,tsx,json,md}\"",
-    "prepare": "husky"
+    "postinstall": "husky"
```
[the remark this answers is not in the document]
compojoom: enableScripts: false - the enable script directive applies to packages that are consumed by us, not to our own package.json Our own post install is being run.

.husky/pre-commit
compojoom: lint-staged is already setup in the monorepo root. Npx should use the monorepo version and not download the latest one. Thanks for flagging - this gave me the idea to actually update lint-staged :)

PR #6735, opened 2025-11-27
fix(mobile): history had no loading indicators

compojoom:
## What it solves
In https://github.com/safe-global/safe-wallet-monorepo/pull/6654 I fixed a flicker on the history tab. Unfortunately I didn't notice that I broke the loading indicator when fetching more pages.

What I also noticed is that if the user has bad internet connection and the fetch fails we display an empty tx history and it is not possible for the user to reload the page. Unless he closes the app he won't be able to load the txs.

Resolves https://linear.app/safe-global/review/fixmobile-tx-history-list-flicker-when-scrolling-0ca8651f84b5

## How to test it
Load a safe, go to transaction history scroll fast to the end of the list - observe that a skeleton loader is shown. To see the error message. Load a safe, turn off the internet on the device, navigate to history -> observe that we display an error screen. Turn the internet on the device again and now swipe down to refresh - the transactions should load

## Screenshots
image

## Checklist

---

## CLA signature


apps/mobile/src/features/TxHistory/components/TxHistoryList/components/LoadingComponents.tsx
```diff
@@ -28,3 +29,25 @@ const FooterComponentBase = () => (
...
+  >
+    <SafeFontIcon name="info" size={48} color="$colorSecondary" />
+    <Text fontSize="$5" fontWeight="600" marginTop="$4" textAlign="center">
+      Error fetching transactions
+    </Text>
+    <Text fontSize="$3" color="$colorSecondary" marginTop="$2" textAlign="center">
+      Swipe down to retry
+    </Text>
+  </View>
+)
```
clovisdasilvaneto: The component name is LoadingComponents, having an error component here wouldn't make too much sense IMO

clovisdasilvaneto (approved):

PR #6744, opened 2025-11-27
fix(portfolio): Optimize total fiat value for different token lists

vseehausen:
## What it solves

Resolves: https://linear.app/safe-global/issue/GRO-129/all-tokens-point-to-transaction-service-token-list

## How this PR fixes it

* Changes `useLoadBalances` and `usePositions` hooks so that:
  * total fiat value is always calculated from the total fiat returned by portfolio
  * hidden tokens' values are deducted from fiat total, when they exist in the portfolio list (both lists get merged)

## How to test it

Toggle "all tokens" and "default tokens" for Safes with position tokens.

Observe the fiat total.

## Checklist

---

## CLA signature

---

> [!NOTE]
> Always use portfolio totals/positions; when All Tokens is selected, merge portfolio totals with legacy token list (recomputing token total) and show context tooltips.
> 
> - **Balances/Positions logic**:
>   - `useLoadBalances`: always sources `fiatTotal`/positions from portfolio when available; for `TOKEN_LISTS.ALL`, merges portfolio totals/positions with legacy `items`, computes `tokensFiatTotal` from legacy items, and flags `isAllTokensMode`.
>   - `usePositions`: uses portfolio endpoint whenever feature enabled (no longer gated by token list), falling back to legacy positions otherwise.
> - **UI**:
>   - `TotalAssetValue`: accepts `isAllTokensMode`; shows `InfoTooltip` clarifying totals context.
>   - `SafeHeaderInfo`: displays `InfoTooltip` when `balances.isAllTokensMode`.
>   - `balances/index.tsx`: passes `balances.isAllTokensMode` to `TotalAssetValue`.
> - **Tests**:
>   - Update `useLoadBalances` tests to validate merged behavior, `isAllTokensMode`, and portfolio-first totals; adjust expectations accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c1e1fe5cf1088e08f7de094d5b97c379721757f8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/web/src/components/balances/TotalAssetValue/index.tsx
+++ b/apps/web/src/components/balances/TotalAssetValue/index.tsx
@@ -37,7 +37,7 @@ const TotalAssetValue = ({
             <>
               <FiatValue value={fiatTotal} precise />
               {isAllTokensMode && (
-                <InfoTooltip title="This total includes all tokens. Your portfolio total may differ." />
+                <InfoTooltip title="Total from this list only. Portfolio total includes positions and may use other token data." />
               )}
             </>
           ) : (


apps/web/src/hooks/loadables/useLoadBalances.ts
```diff
@@ -86,29 +88,91 @@ export const useLegacyBalances = (skip = false): AsyncResult<PortfolioBalances>
...
+    }
+
+    // If we don't have data yet, return loading to show initial loading state
+    if (!portfolioData || !legacyData) {
+      return [undefined, undefined, true]
+    }
+
+    // Merge: portfolio fiatTotal/positions + legacy items + calculated tokensFiatTotal
+    // Store portfolio token addresses to check if hidden tokens should affect fiatTotal
+    const portfolioTokenAddresses = new Set(portfolioData.items.map((item) => item.tokenInfo.address.toLowerCase()))
```
tmjssz: Consider useMemo for Set creation

katspaugh:
@vseehausen should this PR target the portfolio epic instead of dev?

apps/web/src/components/balances/TotalAssetValue/index.tsx
```diff
@@ -31,7 +34,12 @@ const TotalAssetValue = ({
       <Typography component="div" variant="h1" fontSize={fontSizeValue} lineHeight="1.2" letterSpacing="-0.5px">
         {safe.deployed ? (
           fiatTotal !== undefined ? (
-            <FiatValue value={fiatTotal} precise />
+            <>
+              <FiatValue value={fiatTotal} precise />
+              {isAllTokensMode && (
+                <InfoTooltip title="This total includes all tokens. Your portfolio total may differ." />
```
katspaugh: I would rephrase this as "This value includes only fungible tokens. DeFi positions aren't included."

apps/web/src/hooks/loadables/useLoadBalances.ts
katspaugh: Converting to a set would remove duplicates in an arbitrary order. Maybe it's better to first sort by value, then use `_.unique`?

apps/web/src/components/sidebar/SidebarHeader/SafeHeaderInfo.tsx
```diff
@@ -43,7 +44,10 @@ const SafeHeaderInfo = (): ReactElement => {
         <Typography data-testid="currency-section" variant="body2" fontWeight={700}>
           {safe.deployed ? (
             balances.fiatTotal ? (
-              <FiatValue value={balances.fiatTotal} />
+              <>
+                <FiatValue value={balances.fiatTotal} />
+                {balances.isAllTokensMode && <InfoTooltip title="Total based on default tokens." />}
```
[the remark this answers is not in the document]
clovisdasilvaneto: @vseehausen does it makes sense?

apps/web/src/hooks/loadables/useLoadBalances.ts
```diff
@@ -90,25 +91,71 @@ export const useLegacyBalances = (skip = false): AsyncResult<PortfolioBalances>
...
+    if (portfolioLoading || legacyLoading) {
+      return [undefined, undefined, true]
+    }
+
+    if (portfolioError) {
+      return [undefined, portfolioError, false]
+    }
+    if (legacyError) {
+      return [undefined, legacyError, false]
+    }
```
clovisdasilvaneto:
```suggestion
    const error = portfolioError || legacyError

    if (error) {
      return [undefined, error, false]
    }
```

apps/web/src/hooks/loadables/useLoadBalances.ts
```diff
@@ -90,25 +91,71 @@ export const useLegacyBalances = (skip = false): AsyncResult<PortfolioBalances>
...
+
+    if (!portfolioData || !legacyData) {
+      return [undefined, undefined, true]
+    }
```
clovisdasilvaneto, line 144: in the line 133 we have the following statement:

```js
 if (portfolioLoading || legacyLoading) {
      return [undefined, undefined, true]
    }
```
and here we do the inverse 🤔  isn't it contradictory?

apps/web/src/components/balances/TotalAssetValue/index.tsx
vseehausen: Yeah, but the point is that with ALL tokens selected, the token fiat is different then the token fiat used to total fiat value.

apps/web/src/hooks/loadables/useLoadBalances.ts
vseehausen: It seems contradictory, but it covers an edge case. Without this, the initial return value is empty, then loading, then data. This leads to a flash of "empty balances" on the assets page.

apps/web/src/hooks/loadables/useLoadBalances.ts
vseehausen: Nice one, accepted!

apps/web/src/hooks/loadables/useLoadBalances.ts
vseehausen: removed the set logic

vseehausen:
> @vseehausen should this PR target the portfolio epic instead of dev?

Yes!

apps/web/src/components/balances/TotalAssetValue/index.tsx
vseehausen: What about: "Total from this list only. Portfolio total includes positions and may use other token data."

PR #6746, opened 2025-11-28
Hotfix(Hypernative): No concurrent event tracking; HN banner for targeted Safes with 0 balance

vmosharova:
## What it solves
1. The "Guardian banner viewed" analytics event was firing 3 times when disconnecting and reconnecting a wallet, causing duplicate tracking events.
2. The HN Banner on the dashboard was overlapped by "Add Funds" banner.

Resolves: 
1. [COR-888](https://linear.app/safe-global/issue/COR-888/add-one-time-tracking-for-hypernative-banner-per-safe)
The issue occurred because multiple banner types (Promo, NoBalanceCheck, Settings) each use `useBannerVisibility`, which calls a tracking hook `useTrackBannerEligibilityOnConnect`. When these hooks ran simultaneously during wallet reconnection, they all caused a race condition where all three instances saw `false` and all fired the tracking event.

2. [COR-926](https://linear.app/safe-global/issue/COR-926/show-the-hn-dashboard-banner-for-targeted-safes-with-0-balance)
Targeted Safes with 0 balance were not shown the Hypernative banner on the dashboard, as Add Funds banner had as higher priority.

## How this PR fixes it

1. 
**Added a shared lock mechanism**: Introduced a module-level `activeTrackingSafes` Set that acts as a coordination mechanism across all hook instances, preventing concurrent tracking for the same Safe.

**Atomic check-and-add pattern**: Before checking Redux state, the hook now atomically checks if another instance is already tracking and immediately acquires the lock. This prevents multiple instances from proceeding simultaneously.

**Proper lock cleanup**: The lock is released in all early return paths (already tracked, already initiated, banner shouldn't show) and automatically cleaned up when the hook unmounts.

2. 
**Updated the dashboard component** to show the Hypernative banner instead of "Add funds to get started" when the Safe is targeted and has no assets.

## How to test it

1. 
   - Connect a wallet to a Safe that meets banner eligibility criteria
   - Disconnect the wallet
   - Remove the hnState in Dev Tools
   - Refresh the page
   - Reconnect the wallet
   - Check analytics/mixpanel - the "Guardian Banner Viewed" event should fire exactly once
   
2. 
- Manual verification: targeted Safe with 0 balance shows the Hypernative banner

---

## CLA signature

### DIFF
--- a/apps/web/src/components/dashboard/index.tsx
+++ b/apps/web/src/components/dashboard/index.tsx
@@ -87,7 +87,6 @@ const Dashboard = (): ReactElement => {
         <div className={css.leftCol}>
           <Overview />
 
-          {/* If the Safe has no assets AND in the Safe targeted list, show the Hypernative banner */}
           {noAssets ? (
             <Stack spacing={1}>
               {showHnBanner && <HnBannerForCarousel onDismiss={() => {}} />}
--- a/apps/web/src/features/hypernative/hooks/__tests__/useBannerVisibility.test.ts
+++ b/apps/web/src/features/hypernative/hooks/__tests__/useBannerVisibility.test.ts
@@ -816,7 +816,7 @@ describe('useBannerVisibility', () => {
           showBanner: true,
           loading: false,
         })
-        // Verify that useIsOutreachSafe was called with the correct outreach ID
+
         expect(useIsOutreachSafeHook.useIsOutreachSafe).toHaveBeenCalledWith(HYPERNATIVE_OUTREACH_ID)
       })
 


vmosharova:
@codex review

vmosharova:
@codex review new changes

apps/web/src/components/dashboard/index.tsx
```diff
@@ -87,7 +87,15 @@ const Dashboard = (): ReactElement => {
         <div className={css.leftCol}>
           <Overview />
 
-          {noAssets ? <AddFundsToGetStarted /> : <NewsCarousel banners={banners} />}
+          {/* If the Safe has no assets AND in the Safe targeted list, show the Hypernative banner */}
```
LucieFaire: I think the logic below is self-explanatory, we can remove the comment

apps/web/src/features/hypernative/hooks/__tests__/useBannerVisibility.test.ts
```diff
@@ -791,6 +791,35 @@ describe('useBannerVisibility', () => {
...
+        })
+        jest.spyOn(useIsOutreachSafeHook, 'useIsOutreachSafe').mockReturnValue(true) // Safe is targeted
+
+        const { result } = renderHook(() => useBannerVisibility(BannerType.Promo))
+
+        expect(result.current).toEqual({
+          showBanner: true,
+          loading: false,
+        })
+        // Verify that useIsOutreachSafe was called with the correct outreach ID
```
LucieFaire: idk why AI always have to add these comments :)

apps/web/src/components/dashboard/index.tsx
vmosharova: resolved

PR #6750, opened 2025-11-28
fix(Hypernative): prevent HN Banner View tracking event in the edge case

vmosharova:
## What it solves

The tracking event "Guardian banner viewed" was firing even when the HN banner (in `FirstSteps`) was not visible on the Dashboard. It was an edge case when the Safe was **on Testnet (e.g. Sepolia)** and satisfied the conditional balance check (TVL > $1 on Staging), although **balances on the dashboard** (`fiatTotal`) were 0$ while **balances on the sidebar** were > 1$ (happens due to the balances fetched from the new Zerion endpoint).

The tracking event fired even when `FirstSteps` was not visible in the UI (for deployed Safes).

Resolves: [COR-927](https://linear.app/safe-global/issue/COR-927/tracking-guardian-banner-viewed-triggers-on-testnets-without-banner)

## How this PR fixes it

1. Added a check to skip the FirstSteps banner view tracking when the Safe is deployed, ensuring tracking only fires when the banner is rendered and visible (for undeployed Safes).

2. Included `safe.deployed` in the dependency array to properly react to status changes.

## How to test it

1. Test on testnet (e.g., Sepolia) with a Safe that has >1$ balance (satisfying the HN banner rendering condition when TVL > 1$ on Staging).

2. Verify the tracking event does not fire when the HN banner is not visible on the main Dashboard (for deployed Safes)

3. Verify the tracking event does fire when the banner is visible on the page "First Steps" (visible only for undeployed Safes)

## Checklist

  - Fixes incorrect tracking event firing when banner is not visible
  - Ensures "Guardian banner viewed" only fires when banner is actually rendered

---

## CLA signature


vmosharova:
@codex review

katspaugh (approved):

PR #6774, opened 2025-12-03
fix(web): token icons in tx list

katspaugh:
## What it solves

Resolves https://linear.app/safe-global/issue/COR-938/token-logos-in-tx-history-iframe-appear-stretched-and-cropped

Non-square token icons were stretched in tx queue/history.

After the fix:

Screenshot 2025-12-03 at 11 17 56


francovenica:
I've checked where the issue first was spotted, in Sepolia. Looks good there 
(also looks good on Polygon, but there it wasn't as obvious)

image

PR #6775, opened 2025-12-03
feat: Add Report False Result feature for Blockaid threat analysis (COR-43)

Fbartoli:
## What it solves

Implements the frontend integration for reporting false Blockaid threat analysis results, allowing users to report both false positives (incorrectly flagged as dangerous) and false negatives (should have been flagged but wasn't).

Related to: https://linear.app/safe-global/issue/COR-43/blockaid-allow-reporting-false-positives

Backend PR: https://github.com/safe-global/safe-client-gateway/pull/2815

### New Components
- **ReportFalseResultModal**: Modal dialog with radio buttons for event type (FALSE_POSITIVE/FALSE_NEGATIVE) and a details text field
- **useReportFalseResult hook**: Handles API calls to the new CGW endpoint, notifications, and analytics tracking

### Integration
- Added `requestId` to `SafeShieldContext` to track threat analysis requests
- Added "Report false result" link inside the threat analysis card's gray content area (matching Figma design)
- Link appears when a `requestId` exists from the threat scan

### Analytics
Added tracking events:
- `REPORT_MODAL_OPENED`
- `REPORT_SUBMITTED`
- `REPORT_FAILED`

### API Types
Updated `packages/store/src/gateway/AUTO_GENERATED/safe-shield.ts` with:
- `ReportFalseResultRequestDto` type
- `ReportFalseResultResponseDto` type
- `useSafeShieldReportFalseResultV1Mutation` hook

## Design Reference

Figma: https://www.figma.com/design/t1gQpf32LYmLKaxhTGv0Hm/Tx-Assistance?node-id=3241-16260

## Testing

- Added unit tests for `useReportFalseResult` hook
- Added component tests for `ReportFalseResultModal`
- Added Storybook stories for the modal and updated SafeShield stories with `requestId`

### DIFF
--- a/apps/mobile/e2e/tests/transactions/pending/safe-shield/batch-3-actions-warn.yml
+++ b/apps/mobile/e2e/tests/transactions/pending/safe-shield/batch-3-actions-warn.yml
@@ -32,7 +32,7 @@ tags:
     env:
       WAIT_FOR: '.*3 actions.*'
       TAP_TEXT: '.*3 actions.*'
-      TAP_INDEX: '2'
+      TAP_INDEX: '1'
 
 - assertVisible: 'Confirm transaction'
 
--- a/apps/mobile/e2e/tests/transactions/pending/safe-shield/pending-bridge.yml
+++ b/apps/mobile/e2e/tests/transactions/pending/safe-shield/pending-bridge.yml
@@ -28,7 +28,7 @@ tags:
 - scrollUntilVisible:
     element:
       text: '.*Pending bridge.*'
-    centerElement: true
+      index: 0
 
 - runFlow:
     file: ../../../../utils/components/pending-tx/tap-tx-card.yml
--- a/apps/mobile/e2e/tests/transactions/pending/safe-shield/send-matic-malicious.yml
+++ b/apps/mobile/e2e/tests/transactions/pending/safe-shield/send-matic-malicious.yml
@@ -23,6 +23,12 @@ tags:
 - assertVisible:
     id: 'pending-tx-list'
 
+# Scroll to see the send transaction card
+- scrollUntilVisible:
+    element:
+      text: '.*0\.000001 MATIC.*'
+    centerElement: true
+
 - runFlow:
     file: ../../../../utils/components/pending-tx/tap-tx-card.yml
     env:


apps/web/src/features/safe-shield/hooks/useReportFalseResult.ts
```diff
@@ -0,0 +1,80 @@
...
+        return false
+      }
+    },
+    [chainId, safeAddress, triggerReport, dispatch],
+  )
+
+  return {
+    reportFalseResult,
+    isLoading,
+  }
```
clovisdasilvaneto: shouldn't we handle the error as well?

clovisdasilvaneto (approved):

packages/store/src/gateway/AUTO_GENERATED/safe-shield.ts
```diff
@@ -300,18 +300,20 @@ export type ThreatAnalysisRequestDto = {
   /** Optional origin identifier for the request */
   origin?: string
 }
-export type ReportFalseResultResponseDto = {
-  /** Whether the report was submitted successfully */
-  success: boolean
-}
+/** Event types for reporting false Blockaid scan results. */
+export type ReportEvent = 'FALSE_POSITIVE' | 'FALSE_NEGATIVE'
```
LucieFaire: was this auto-generated or manually added? I dont see the definition for this type in `schema.json`.

In general we do not touch this file, everything comes from openapi gen

LucieFaire: if you want to wrap into a type you can either provide it as its own Swagger definition in CGW, so it gets generated, or you can create a type in `packages/utils/src/features/safe-shield/types.ts`. But i would remove manual changes in auto-gen file

Fbartoli: Wait there is something weird then because it is from the CGW merge when you asked me to create a type.


https://github.com/safe-global/safe-client-gateway/pull/2815/files
https://github.com/safe-global/safe-client-gateway/pull/2815#discussion_r2580979352
[src/modules/safe-shield/entities/dtos/report-false-result.dto.ts](https://github.com/safe-global/safe-client-gateway/blob/bb871460b534bfabf5a38d3eabf07a6af2b328e3/src/modules/safe-shield/entities/dtos/report-false-result.dto.ts#L10)

Fbartoli: Let me try to regenerate things

LucieFaire: I dont think it will create the type automatically, it will keep it as explicit enum values, because `ReportEvent` is not decorated in CGW. Or does it create the type?

PR #6779, opened 2025-12-03
fix: handle request_id in threat analysis response

Fbartoli:
What this PR does
Fixes a crash when CGW returns request_id in the threat analysis response (introduced in safe-client-gateway PR #2815).

Changes
packages/utils/src/features/safe-shield/types.ts
Add request_id?: string field to ThreatAnalysisResults type
packages/utils/src/features/safe-shield/utils/getOverallStatus.ts
Add array check to skip non-array values (like request_id) during iteration
packages/utils/src/features/safe-shield/utils/mapVisibleAnalysisResults.ts
Destructure request_id out of results map to prevent it from being processed

Root Cause
The crash occurred because Object.values(threatResults) now returns [THREAT_ARRAY, BALANCE_CHANGE_ARRAY, "request_id_string"]. When iterating and calling Object.values() on the string request_id, it returns individual characters, causing unexpected behavior in the type guard checks.

Testing
[x] TypeScript type check passes
[x] Lint passes
[x] Manual testing with new CGW response format


LucieFaire:
@Fbartoli have you regenerated swagger definition and the rtk-query from it?

Fbartoli:
> @Fbartoli have you regenerated swagger definition and the rtk-query from it?

I have not give me a sec.
It seems there is a bit a backlog there

Fbartoli:
@LucieFaire Is that what you meant ?

packages/utils/src/features/safe-shield/utils/mapVisibleAnalysisResults.ts
```diff
@@ -20,7 +20,12 @@ import { mapConsolidatedAnalysisResults } from './mapConsolidatedAnalysisResults
 export const mapVisibleAnalysisResults = (
   addressesResultsMap: RecipientAnalysisResults | ContractAnalysisResults | ThreatAnalysisResults,
 ): AnalysisResult[] => {
-  const { name: _name, logoUrl: _logoUrl, ...resultsMap } = addressesResultsMap as ContractAnalysisResults
+  const {
+    name: _name,
+    logoUrl: _logoUrl,
+    request_id: _requestId,
+    ...resultsMap
+  } = addressesResultsMap as ContractAnalysisResults & { request_id?: string }
```
LucieFaire: is this correct? I see its ContractAnalysisResults, not ThreatAnalysisResults

Fbartoli: I am not quite sure of the history of that part and why there is a casting specifically on ContractAnalysisResults

i updated to use a more type agnostic way to do the metadata filtering.

PR #6784, opened 2025-12-04
chore: upgrade @reduxjs/toolkit

compojoom:
## What it solves
Redux toolkit got some very cool updates (https://github.com/reduxjs/redux-toolkit/releases) that touch performance and fix some issues with infinite queries and other smal lthings. 

Putting the performance issues aside, this is pretty cool:

> New refetchCachedPages Option
> 
> RTK Query's infinite query API was directly based on React Query's approach, including the pages cache structure and refetching behavior. By default, that means that when you trigger a refetch, both R-Q and RTKQ will try to sequentially refetch all pages currently in that cache entry. So, if there were 5 pages cached for an entry, they will try to fetch pages 0...4, in turn.
> 
> Some users have asked for the ability to only refetch the first page. This can be accomplished somewhat manually by directly updating the cache entry to eliminate the old pages and then triggering a refetch, but that's admittedly not very ergonomic.
> 
> We've merged a contributed PR that adds a new refetchCachedPages flag. This can be defined as part of infinite query endpoints, passed as an option to infinite query hooks, or passed as an option in initiate() calls or hook refetch() methods. If set to refetchCachedPages: false, it will only refetch the first page in the cache and not the remaining pages, thus shrinking the cache from N pages to 1 page.
> 

I had to do workarounds because of this behavior in the mobile app. With this update it is now just a function parameter, which is pretty cool. 

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/jest.config.js
+++ b/apps/mobile/jest.config.js
@@ -21,7 +21,7 @@ module.exports = {
       '<rootDir>/__mocks__/fileMock.js',
   },
   transformIgnorePatterns: [
-    'node_modules/(?!((jest-)?react-native|@react-native(-community)?)|expo(nent)?|@expo(nent)?/.*|@expo-google-fonts/.*|react-navigation|@react-navigation/.*|@sentry/react-native|native-base|react-native-svg|@notifee/react-native|react-redux|moti/.*|@ledgerhq/.*|uuid)',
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
👍

PR #6785, opened 2025-12-04
fix(tx queue): show transaction details block for proposer transactions

katspaugh:
## Description

Resolves https://linear.app/safe-global/issue/COR-947/share-block-confirmreject-buttons-missing-for-proposer-transactions

Fixes the issue where transactions created by a proposer (delegate signing key) weren't showing the right-side block in the transaction details.

### Transaction Details Block Visibility
- Display right-side block for transactions created by proposers/delegates
- Show block even when transaction has 0 confirmations

### Proposer Detection
- Check both `proposedByDelegate` field from API
- Cross-reference proposer address with the delegates list using `useProposers` hook
- Mark transaction as from proposer if either condition is met

### UI Enhancements
- Always show **Confirmations** step (even for 0 of X)
- Display proposer address in the signers list
- Show 'Signer review' status with explanatory text for proposer transactions
- Display Sign/Confirm and Reject buttons for proposer transactions
- Hide Sign button for expired swap orders (Reject only)

## Testing

- ✅ Type check passes
- ✅ All tests pass (45 tests)
- ✅ Pre-commit hooks pass (prettier, lint, type-check)

## Screenshots

Before: Right-side block was hidden for proposer transactions with 0 confirmations

After: 
- Created step with proposer address
- Confirmations (0 of X) with active state
- Signer review status
- Action buttons (Sign/Confirm and Reject)

### DIFF
--- a/apps/web/src/components/transactions/TxDetails/index.tsx
+++ b/apps/web/src/components/transactions/TxDetails/index.tsx
@@ -41,6 +41,7 @@ import { POLLING_INTERVAL } from '@/config/constants'
 import { TxNote } from '@/features/tx-notes'
 import { TxShareBlock, TxExplorerLink } from '../TxShareLink'
 import { FEATURES } from '@safe-global/utils/utils/chains'
+import { sameAddress } from '@safe-global/utils/utils/addresses'
 import DecodedData from './TxData/DecodedData'
 import { QueuedTxSimulation } from '../QueuedTxSimulation'
 import HnSecurityReportBtnForTxDetails from '@/features/hypernative/components/HnSecurityReportBtn'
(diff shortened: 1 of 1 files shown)


apps/web/src/components/transactions/TxDetails/index.tsx
```diff
@@ -77,13 +79,23 @@ const TxDetailsBlock = ({ txSummary, txDetails }: TxDetailsProps): ReactElement
...
   if (isMultisigDetailedExecutionInfo(txDetails.detailedExecutionInfo)) {
     safeTxHash = txDetails.detailedExecutionInfo.safeTxHash
     proposedByDelegate = txDetails.detailedExecutionInfo.proposedByDelegate
     proposer = proposedByDelegate?.value ?? txDetails.detailedExecutionInfo.proposer?.value
   }
 
+  // Check if the proposer is actually a delegate
+  const isProposerDelegate = useMemo(() => {
+    if (!proposer || !proposersData?.results) return false
+    return proposersData.results.some((p) => p.delegate === proposer)
```
tmjssz: Maybe better to use `sameAddress` from utils for comparison here

tmjssz (approved):

PR #6788, opened 2025-12-04
feat: add Datadog RUM integration with error tracking

gfendres:
## Summary

Adds Datadog Real User Monitoring (RUM) integration to capture frontend performance metrics, errors, and user session data. This provides comprehensive observability for production incidents and performance optimization.

### Observability Architecture
- **Unified provider pattern** supporting Sentry, Datadog, or both simultaneously via CompositeProvider
- **Graceful degradation** with NoOpProvider when services aren't configured
- **Partial failure tolerance** - if Datadog Logs fails, RUM still captures errors (and vice versa)
- **Static imports for early RUM initialization** - captures critical metrics (FCP, LCP, TTI) from page load
- **Explicit initialization** - no import side effects, clear control over timing

### Datadog RUM Configuration
- Session sampling: 10% (configurable via `NEXT_PUBLIC_DATADOG_RUM_SESSION_SAMPLE_RATE`)
- Session replay: 5% (configurable via `NEXT_PUBLIC_DATADOG_RUM_SESSION_REPLAY_SAMPLE_RATE`)
- Privacy-first: `mask-user-input` mode by default
- User interaction tracking, resource tracking, long task tracking
- Distributed tracing integration with Safe Client Gateway

### Environment Handling
- **Production/Development toggle** via `NEXT_PUBLIC_DATADOG_RUM_ENV`
- Simplified to 2 environments: `production` and `development`
- Automatic sourcemap upload during CI builds with version correlation
- Proper Safe Client Gateway URL allowlisting for distributed tracing

### Core Changes
1. **Provider Pattern** (`src/services/observability/`)
   - `IObservabilityProvider` interface with `init()`, `captureException()`, `getLogger()`
   - `DatadogProvider`, `SentryProvider`, `NoOpProvider`, `CompositeProvider`
   - Factory pattern in `createObservabilityProvider()` for runtime configuration

2. **Static Imports for Performance**
   - Changed from dynamic `import()` to static imports
   - Enables Datadog RUM to load immediately and capture early page metrics
   - Added Jest mocks in `jest.setup.js` to prevent test interference

3. **Explicit Initialization Pattern**
   - `ObservabilityErrorBoundary` wraps entire app in `_app.tsx`
   - Explicit `initObservability()` called at app entry point before React renders
   - No import side effects - initialization is visible and controlled
   - Captures early page metrics (FCP, LCP, TTI) and errors during hydration

4. **Build Integration**
   - Modified `.github/actions/build/action.yml` to upload sourcemaps to Datadog
   - Derives service name and site from environment variables
   - Correlates sourcemaps with RUM data using `COMMIT_HASH` as version

### Dependencies
Moved to regular dependencies (from optionalDependencies):
- `@datadog/browser-logs@^6.24.1`
- `@datadog/browser-rum@^6.24.1`
- `@datadog/browser-rum-react@^6.24.1`
- `@sentry/react@^7.91.0` (existing)

Bundle impact: ~50-80KB gzipped

## Environment Variables

Required for Datadog RUM:
```bash
NEXT_PUBLIC_DATADOG_RUM_APPLICATION_ID=<app-id>
NEXT_PUBLIC_DATADOG_RUM_CLIENT_TOKEN=<client-token>
NEXT_PUBLIC_DATADOG_RUM_SITE=datadoghq.eu  # or datadoghq.com, etc.
```

Optional configuration:
```bash
NEXT_PUBLIC_DATADOG_RUM_ENV=production  # or development
NEXT_PUBLIC_DATADOG_RUM_SERVICE=safe-wallet-web
NEXT_PUBLIC_DATADOG_RUM_SESSION_SAMPLE_RATE=10
NEXT_PUBLIC_DATADOG_RUM_SESSION_REPLAY_SAMPLE_RATE=5
NEXT_PUBLIC_DATADOG_FORCE_ENABLE=false  # Enable in non-production
```

## Testing

- ✅ 41 observability tests passing (providers, factory, initialization)
- ✅ All 2,950 existing tests passing
- ✅ Type-check passing
- ✅ Comprehensive test coverage:
  - Provider initialization (sync/async, success/failure, partial failure)
  - Explicit initialization behavior (client vs server)
  - Factory routing logic (NoOp/Sentry/Datadog/Composite)
  - Error boundary integration

### Current Setup (Sentry Only)
No changes required. Sentry continues working as before.

### Adding Datadog
1. Set environment variables in deployment config
2. Datadog provider automatically activates when tokens are present
3. CompositeProvider routes to both Sentry and Datadog

### Future: Datadog Only
When ready to drop Sentry, simply remove `SENTRY_DSN` from environment variables. The factory will automatically switch to DatadogProvider only.

### Why Static Imports?
**Decision:** Use static imports instead of dynamic imports for Datadog/Sentry SDKs.

**Rationale:** 
- Captures early page load metrics (FCP, LCP, TTI) that dynamic imports miss
- Worth the ~50-80KB bundle cost for accurate RUM data
- SDK loads before React even hydrates

**Alternative considered:** Dynamic imports to reduce initial bundle
- **Rejected:** Misses the most critical performance metrics

### Why Explicit Initialization?
**Decision:** Use explicit `initObservability()` function instead of auto-initialization on module import.

**Rationale:**
- No import side effects - importing `captureException` doesn't trigger network requests
- Explicit and visible - clear where initialization happens in `_app.tsx`
- Better testability - no module-level state to mock
- Full timing control - can adjust initialization order if needed
- Simpler codebase - removed ~30 lines of queue management complexity

**Alternative considered:** Auto-initialization on module import with exception queue
- **Rejected:** Violates principle of least surprise, makes testing difficult, adds unnecessary complexity (per @compojoom feedback)

### Why 2 Environments Instead of 5?
**Decision:** Simplified from 5 environments (production, staging, dev, review, development) to 2 (production, development).

**Rationale:**
- Reduces complexity in Datadog filtering
- Production vs non-production is the key distinction
- Easier to maintain and understand

**Alternative considered:** Keep 5 environments for granular filtering
- **Rejected:** Overhead not justified by usage patterns

### Why Partial Failure Tolerance?
**Decision:** If Datadog Logs fails but RUM succeeds, still capture errors via RUM.

**Rationale:**
- Better to have some observability than none
- Datadog Logs and RUM are independent subsystems
- Tested explicitly in `datadog.test.ts:118-152`

**Alternative considered:** All-or-nothing failure (re-throw on any init failure)
- **Rejected:** Breaks the use case where one subsystem works

## Fixes from PR Review

Based on review feedback:

1. ✅ **Fixed sourcemap configuration** - Derives service/site from env vars (matches RUM SDK)
2. ✅ **Moved to regular dependencies** - Changed from optionalDependencies (per @compojoom feedback)
3. ✅ **Improved test quality** - Rewrote tests to verify behavior, not TypeScript guarantees (per @compojoom feedback)
4. ✅ **Simplified RUM_ENV** - Reduced from 5 to 2 environments (per @vseehausen feedback)
5. ✅ **Refactored to explicit initialization** - Removed auto-init pattern and exception queue (per @compojoom feedback)
   - No more import side effects
   - Simpler implementation (59 → 28 lines in index.ts)
   - Simpler tests (327 → 191 lines)
   - Better testability without module-level state

## Related Documentation

- [Datadog RUM Browser Monitoring](https://docs.datadoghq.com/real_user_monitoring/browser/)
- [Datadog Sourcemap Upload](https://docs.datadoghq.com/real_user_monitoring/guide/upload-javascript-source-maps/)
- [Safe Client Gateway Integration](https://github.com/safe-global/safe-client-gateway)

---

> [!NOTE]
> **Medium Risk**
> Touches global app initialization and error/reporting pathways and adds new third-party telemetry configuration, which could affect runtime stability or observability behavior if misconfigured.
> 
> **Overview**
> Adds a new `services/observability` layer that can initialize **Sentry, Datadog (Logs + RUM), both, or neither** via a provider/factory pattern, and routes `logger` + `captureException` through it.
> 
> Moves app-wide error handling to a new `ObservabilityErrorBoundary`, initializes observability *before React renders* in `_app.tsx`, and replaces the prior `useDatadog`/`services/datadog` and `services/sentry` usage with the unified API.
> 
> Extends CI build configuration to inject commit/env metadata and Datadog RUM settings, and conditionally uploads Next.js sourcemaps to Datadog using `datadog-ci`; updates dependencies (adds `@datadog/browser-rum`, bumps `@datadog/browser-logs`) and test setup/mocks accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 34dc4b13d9c2126316243472cadf0796360fb6cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


.github/actions/build/action.yml
```diff
@@ -47,6 +70,16 @@ runs:
         NEXT_PUBLIC_SAFE_APPS_GA_TRACKING_ID: ${{ fromJSON(inputs.secrets).NEXT_PUBLIC_SAFE_APPS_GA_TRACKING_ID }}
         NEXT_PUBLIC_SENTRY_DSN: ${{ fromJSON(inputs.secrets).NEXT_PUBLIC_SENTRY_DSN }}
         NEXT_PUBLIC_DATADOG_CLIENT_TOKEN: ${{ fromJSON(inputs.secrets).NEXT_PUBLIC_DATADOG_CLIENT_TOKEN }}
+        NEXT_PUBLIC_DATADOG_RUM_APPLICATION_ID: ${{ fromJSON(inputs.secrets).NEXT_PUBLIC_DATADOG_RUM_APPLICATION_ID }}
+        NEXT_PUBLIC_DATADOG_RUM_CLIENT_TOKEN: ${{ fromJSON(inputs.secrets).NEXT_PUBLIC_DATADOG_RUM_CLIENT_TOKEN }}
+        NEXT_PUBLIC_DATADOG_RUM_SITE: ${{ fromJSON(inputs.secrets).NEXT_PUBLIC_DATADOG_RUM_SITE }}
+        NEXT_PUBLIC_DATADOG_RUM_SERVICE: ${{ fromJSON(inputs.secrets).NEXT_PUBLIC_DATADOG_RUM_SERVICE }}
+        NEXT_PUBLIC_DATADOG_RUM_ENV: ${{ env.NEXT_PUBLIC_DATADOG_RUM_ENV }}
```
vseehausen, line 71: I would use the node environment for this value.

vseehausen: For the `RUM_ENV`.

apps/web/package.json
```diff
@@ -98,6 +96,18 @@
...
+    "@datadog/browser-rum": "^6.24.1",
+    "@datadog/browser-rum-react": "^6.24.1",
+    "@sentry/react": "^7.91.0"
+  },
+  "dependenciesMeta": {
+    "@datadog/browser-logs": {},
+    "@datadog/browser-rum": {},
+    "@datadog/browser-rum-react": {},
+    "@sentry/react": {}
+  },
```
compojoom: I don't understand this? Why are we using optional and dependencies meta?

> Set of dependencies that Yarn should only try to install if the os/cpu/libc fields match those of the host platform.
> 
> Unlike regular dependencies, those listed in optionalDependencies are allowed to have a failing postinstall step - in fact, they won't even be installed at all if the os/cpu/libc filters don't cover the host platform.
> 
> Note that optionalDependencies only cares about whether the package should install/build or not - it should still be resolvable, as otherwise it's impossible to tell whether a failure to retrieve the package metadata is intentional or not.

https://yarnpkg.com/configuration/manifest#optionalDependencies

why are we using optional here? Shouldn't those be normal dependencies as they are required to use sentry or datadog. The bundler is going to exclude them in the final bundle if the user is not using them?

apps/web/src/services/observability/__tests__/index.test.ts
```diff
@@ -0,0 +1,180 @@
...
+      }))
+
+      const { captureException } = require('../index')
+      const error = new Error('test error')
+
+      expect(() => captureException(error)).not.toThrow()
+    })
+  })
+
+  it('should queue captureException until init completes', async () => {
```
compojoom: All other tests in this file are useless. 

1. logger exists with properties — TypeScript already guarantees this at compile time
2. captureException is a function — TypeScript guarantees this
3. Calling logger.info('test') doesn't throw — of course it doesn't, you just mocked it with jest.fn()
4. Calling captureException() doesn't throw — same problem

I would actually test:
MAX_QUEUED_EXCEPTIONS = 25 — oldest gets dropped when queue exceeds limit. Not tested.
initFailed behavior — when init fails, queue is cleared. Not tested.
Logger methods actually call provider

apps/web/src/services/observability/__tests__/factory.test.ts
```diff
@@ -0,0 +1,41 @@
...
+      GATEWAY_URL_STAGING: '',
+    }))
+
+    const { createObservabilityProvider } = require('../factory')
+    const provider = createObservabilityProvider()
+
+    expect(provider.name).toBe('NoOp')
+  })
+
+  it('should create provider with correct configuration', () => {
```
compojoom: Test is useless - typescript guarantees that all those properties exist.

apps/web/src/services/observability/__tests__/factory.test.ts
```diff
@@ -0,0 +1,41 @@
+describe('createObservabilityProvider', () => {
```
compojoom, line 1: What's NOT tested:
When only Sentry is configured → returns SentryProvider
When only Datadog is configured → returns DatadogProvider
When both are configured → returns CompositeProvider
The IS_PRODUCTION || DATADOG_FORCE_ENABLE logic

apps/web/src/services/observability/__tests__/noop.test.ts
```diff
@@ -0,0 +1,55 @@
+import { NoOpProvider } from '../providers/noop'
+
+describe('NoOpProvider', () => {
```
compojoom, line 3: The implementation is ~10 lines of empty functions. The tests are 56 lines verifying that:
Empty functions don't throw
Properties that TypeScript guarantees exist, exist
The only marginally useful test is should have name "NoOp" — but even that's debatable since name is a readonly literal type.

compojoom (commented):
@gfendres please instruct your AI to analyse the tests and to write tests that actually test behavior and not just hit coverage lines.

gfendres:
Thanks for the reviews! I've addressed all the feedback in 8d4c9ded7. Here's a breakdown:

---

### Sourcemap Configuration (@chatgpt-codex-connector, @cursor)

**Comments on `.github/actions/build/action.yml:85-90`**

This is actually working correctly. The sourcemap upload reads `NEXT_PUBLIC_DATADOG_RUM_SERVICE` (line 111) and `NEXT_PUBLIC_DATADOG_RUM_SITE` (line 118), which match the RUM SDK config exactly. They both use the same fallback values, so correlation works fine.

I did clean up the DATADOG_SITE syntax at line 118 from `${{ ... && ... || 'datadoghq.eu' }}` to just `${{ ... || 'datadoghq.eu' }}` - same behavior, cleaner code.

---

### Error Boundary Architecture (@chatgpt-codex-connector, @cursor)

**Comments on `apps/web/src/pages/_app.tsx` and `sentry.ts`**

I think there's some confusion about the architecture here:

**Re: Double-reporting**
The `SentryProvider` doesn't expose an ErrorBoundary - it just has a `captureException` method. The Sentry `<ErrorBoundary>` components in 11 features are local boundaries nested inside our `ObservabilityErrorBoundary`. They don't propagate errors upward, so there's no double-reporting.

**Re: Async availability**
`ObservabilityErrorBoundary` is our custom React class component with `componentDidCatch` - not Sentry's ErrorBoundary. It works from first render. The async initialization only affects when errors get *sent* to providers. Errors during init just get queued.

**Re: No Sentry protection**
`ObservabilityErrorBoundary` uses `NoOpProvider` when nothing is configured. It still catches errors and shows fallback UI - just doesn't send them anywhere, which makes sense if no observability service is set up.

---

### Queue Management (@cursor)

**Comments on `sentry.ts:106`, `datadog.ts:127`**

**Re: Queue overflow**
The 25-error limit with FIFO eviction is intentional. The queue only exists during ~100-500ms init. To overflow it you'd need 50+ errors/second - at that point the app is broken and observability isn't the priority.

**Re: Init failure**
Queue clearing only happens when both Sentry and Datadog fail completely. We use `Promise.allSettled()`, so if one fails, the other still works. When both fail, the app has bigger problems - clearing the queue is reasonable. The error gets logged to `console.error()` for debugging.

**Re: Logs preventing RUM**
This isn't accurate - Logs and RUM initialize independently with separate try-catch blocks (lines 113-127 in datadog.ts). Failed Logs import doesn't prevent RUM initialization. There's even test coverage for this (datadog.test.ts:118-152).

---

### Optional Dependencies (@compojoom)

**Comment on `apps/web/package.json:110`**

Fair point - I moved them to regular `dependencies` in 8d4c9ded7.

Next.js needs these packages at build time for dynamic imports. Yarn installs optional dependencies anyway, and the bundler can't exclude npm packages based on runtime conditions. The empty `dependenciesMeta` objects weren't doing anything.

The runtime graceful degradation (NoOpProvider when env vars aren't set) works the same with regular dependencies - the `optionalDependencies` setting wasn't contributing to that.

---

### Test Quality (@compojoom)

**Comments on `__tests__/index.test.ts:107`, `factory.test.ts:1, 31`, `noop.test.ts:3`**

Yeah, the tests weren't great. Fixed in 8d4c9ded7:

**index.test.ts**
Added tests for the actual behavior:
- MAX_QUEUED_EXCEPTIONS overflow with FIFO eviction
- initFailed queue clearing
- Logger delegation to provider
- Client vs server behavior

**factory.test.ts**
Rewrote to test the routing logic:
- No config → NoOpProvider
- Sentry only → SentryProvider
- Datadog only → DatadogProvider
- Both → CompositeProvider
- IS_PRODUCTION || DATADOG_FORCE_ENABLE logic

**noop.test.ts**
Reduced from 55 lines to 18 lines. Single test that verifies it doesn't throw - that's all we need from a no-op.

All 21 tests pass now and actually test behavior instead of just TypeScript.

---

### RUM_ENV Configuration (@vseehausen)

**Comment on `.github/actions/build/action.yml:77`**

Using NODE_ENV directly won't work - Next.js always sets NODE_ENV=production during `next build` (for main, dev, PRs, everything). See [Next.js docs](https://nextjs.org/docs/messages/non-standard-node-env).

But I agree the 5-environment setup was too complex. I simplified it to 2 environments in 8d4c9ded7:

```yaml
if [ "${{ inputs.prod }}" = "true" ]; then
  DATADOG_RUM_ENV="production"
else
  DATADOG_RUM_ENV="development"
fi
```

- **production** - real user traffic
- **development** - everything else (staging, dev, PRs, local)

Much simpler while still maintaining the key distinction for observability.

vseehausen:
Thanks for the rework. Well if it doesn't make sense to use `NODE_ENV`, then I am fine with setting the variable. Better more differentiation than less.

gfendres:
@vseehausen and @compojoom lets merge it?? 
check again see if it is good to go

apps/web/src/services/observability/index.ts
```diff
@@ -0,0 +1,58 @@
...
+    return
+  }
+
+  const pending = queuedExceptions.splice(0)
+  pending.forEach(({ error, context }) => {
+    observabilityProvider.captureException(error, context)
+  })
+}
+
+// Initialize only on client-side (fire-and-forget pattern)
```
compojoom: I'm sorry, but I have to complain about this pattern here. I understand why you did it like this - we need observability as soon as possible for the error boundary, but this magic auto initialisation whenever this file is included is weird and will cause problems for the consumers:
- Importing captureException anywhere triggers initialization. Violates principle of least surprise. Anyone reading calling code has no idea they're kicking off async network requests.
- Every test file that imports this (directly or transitively) runs initialization. You need to mock the factory or the entire module in every test.
- No control over timing - What if you need to init observability AFTER fetching config from the server? You can't. Import order dictates execution.
- Hidden module state - isInitialized, initFailed, queuedExceptions are impossible to reset cleanly between tests.
- The queue adds complexity - It's trying to solve a problem that explicit initialization wouldn't have.

I would put this in a `initObservability` function and then in `_app.tsx` call `initObservability` where I need it. 

You can also drop the queue - what do I care for exceptions that were triggered before the app has inited in the first place?

PR #6793, opened 2025-12-05
tests(mobile): add safe-shield tests

compojoom:
## What it solves
This PR ads tests for safe-shield transactions.

Resolves https://linear.app/safe-global/issue/COR-954/e2e-tests-for-safe-shield-txs

## How to test it
image

## Screenshots

https://github.com/user-attachments/assets/46c3bb9a-9d96-4820-a512-e77e2246fa6e

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/e2e/tests/transactions/pending/safe-shield/pending-bridge.yml
+++ b/apps/mobile/e2e/tests/transactions/pending/safe-shield/pending-bridge.yml
@@ -46,7 +46,7 @@ tags:
       LABEL_2: 'Verified contract'
       LABEL_3: 'Threat analysis failed'
       RUN_SIMULATION: 'true'
-      SIMULATION_RESULT: 'Simulation failed'
+      SIMULATION_RESULT: 'fail'
 
 # ===========================================
 # Balance Change Block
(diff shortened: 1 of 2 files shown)


apps/mobile/src/features/SafeShield/components/SafeShieldWidget/WidgetAction/WidgetAction.tsx
```diff
@@ -39,31 +38,33 @@ export function WidgetAction({ loading, error, status, onPress }: WidgetActionPr
   }, [status, loading, error])
 
   return (
-    <TouchableOpacity onPress={onPress} activeOpacity={0.7}>
```
clovisdasilvaneto, line 42: Since you removed the touchable opacity, this will remove the feedback of the button when user clicks in the section 🤔

compojoom: Yes, I know, but the sheet will open despite this. I removed because when it is there react is applyign view flattening and my text ends up in the same element with the icon. And I cannot target the text directly, but have to use regex :(

PR #6817, opened 2025-12-10
feat(web): display full Safe details in Nested Safes popup

katspaugh:
## Summary

Resolves https://linear.app/safe-global/issue/COR-1027/update-nested-safes-popup-with-comprehensive-safe-details

Updates the Nested Safes popup to display comprehensive Safe information (balances, thresholds, owners) using the same component as the AccountList items and the Safe Overviews endpoint.

Screenshot 2025-12-10 at 17 44 28

### Changes

- **NestedSafesList component**: Refactored to use `SingleAccountItem` component for consistent UI
- **Data fetching**: Implemented `useGetMultipleSafeOverviewsQuery` for efficient batch fetching of Safe Overview data
- **SingleAccountItem enhancement**: Added optional `safeOverview` prop to accept pre-fetched data and avoid duplicate API calls
- **UI improvements**: Now displays balance, threshold, owners count, Safe icon, and chain indicators for each nested safe

### Before
- Simple list showing only address and name from address book
- No balance or threshold information
- Basic Identicon only

### After
- Rich display matching AccountList items
- Shows fiat balance for each nested safe
- Displays threshold and owners count in Safe icon
- Includes chain indicators and account info chips
- Efficient batch data fetching

## Test plan

- [ ] Open the Nested Safes popup from a Safe with nested safes
- [ ] Verify all nested safes are displayed with balances
- [ ] Verify threshold and owners count are shown in the Safe icon
- [ ] Verify the UI matches the AccountList items style
- [ ] Verify clicking on a nested safe navigates correctly
- [ ] Check performance with multiple nested safes

## Screenshots

_Please add screenshots showing the before/after of the Nested Safes popup if possible_

### DIFF
--- a/apps/web/src/components/sidebar/NestedSafesList/index.tsx
+++ b/apps/web/src/components/sidebar/NestedSafesList/index.tsx
@@ -2,8 +2,9 @@ import { ChevronRight } from '@mui/icons-material'
 import { List, Typography, Box } from '@mui/material'
 
 import Track from '@/components/common/Track'
-import { NESTED_SAFE_EVENTS } from '@/services/analytics/events/nested-safes'
+import { NESTED_SAFE_EVENTS, NESTED_SAFE_LABELS } from '@/services/analytics/events/nested-safes'
 import { useState, useMemo, type ReactElement } from 'react'
+import { sameAddress } from '@safe-global/utils/utils/addresses'
 import { useCurrentChain } from '@/hooks/useChains'
 import SingleAccountItem from '@/features/myAccounts/components/AccountItems/SingleAccountItem'
 import type { SafeItem } from '@/features/myAccounts/hooks/useAllSafes'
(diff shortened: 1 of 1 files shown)


vmosharova:
@codex review

vmosharova:
@claude review

apps/web/src/components/sidebar/NestedSafesList/index.tsx
```diff
@@ -22,25 +24,66 @@ export function NestedSafesList({
...
   const onShowAll = () => {
     setShowAll(true)
   }
 
   return (
-    <List sx={{ gap: 1, display: 'flex', flexDirection: 'column', alignItems: 'center' }}>
-      {nestedSafesToShow.map((nestedSafe) => {
-        return <NestedSafeListItem onClose={onClose} nestedSafe={nestedSafe} key={nestedSafe} />
+    <List sx={{ gap: 1, display: 'flex', flexDirection: 'column', alignItems: 'stretch', p: 0 }}>
+      {nestedSafesToShow.map((safeItem) => {
```
vmosharova, line 61: Does it make sense to check the loading state and add a Skeleton here when Safes are loading?

katspaugh:
> When I go to the sidebar --> Accounts overview --> Info (three dots) about a Safe with nested Safes --> Nested Safes, I do not see the "Add Nested Safe" button. Should this functionality be available only in the main sidebar menu?

I didn't change anything in that regard as it's out of scope.

apps/web/src/components/sidebar/NestedSafesList/index.tsx
katspaugh: It does make sense, but it's likely not necessary. Here's why:

  1. SingleAccountItem already handles loading internally (lines 218-224):
  When safeOverview is undefined, it shows a <Skeleton /> for the fiat
  value. So individual items gracefully handle their own loading state.
  2. The useGetMultipleSafeOverviewsQuery in NestedSafesList returns { data:
   safeOverviews } without destructuring isLoading. When data isn't ready,
  safeOverview passed to SingleAccountItem will be undefined, triggering the
   built-in skeleton.
  3. However, there's a subtle UX consideration: Currently the list renders
  immediately with all the SingleAccountItem components, and each one shows
  a skeleton for its fiat value while loading. This is arguably fine UX.

  When you might want a list-level skeleton:
  - If you wanted to show placeholder rows before knowing how many nested
  safes exist
  - If the nestedSafes array itself came from an async source (it doesn't -
  it's passed as a prop)

  Current behavior is acceptable because:
  - The list structure is immediately visible (addresses, Safe icons)
  - Only the fiat values show skeletons during loading
  - This provides good perceived performance

  If you wanted to be more explicit about the loading state, you could add:

  const { data: safeOverviews, isLoading } =
  useGetMultipleSafeOverviewsQuery(...)

  And optionally use isLoading to show a different UI, but the current
  approach of letting SingleAccountItem handle its own loading skeleton is a
   valid pattern and keeps the component simpler.

PR #6835, opened 2025-12-12
chore: update next to 15.5.8

compojoom:
## What it solves
Not strictly necessary as we don't use RSC and we've already patched react-server-dom-webpack, but still good to 
be on the last version.

---

## CLA signature


katspaugh (approved):
🎉

PR #6838, opened 2025-12-12
chores: run prettier globally

vseehausen:
## What it solves

Formats files.

## Checklist

---

## CLA signature

### DIFF
--- a/.github/scripts/capture-screenshots.js
+++ b/.github/scripts/capture-screenshots.js
@@ -4,101 +4,95 @@
  * This script reads story URLs and captures screenshots of each story
  */
 
-const fs = require('fs');
-const path = require('path');
+const fs = require('fs')
+const path = require('path')
 
 // Use playwright-core which is globally installed
-const { chromium } = require('playwright');
+const { chromium } = require('playwright')
 
 async function captureScreenshots() {
   // Read story URLs
-  const storyUrlsFile = 'screenshots/story-urls.json';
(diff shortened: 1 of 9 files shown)


compojoom:
Why this again? Last time I looked running prettier globally runs apparently with different setup than the prettier in the individual pacakges leading to problems. You are modifiying things in mobile only for them to be changed the next time I run yarn prettier from the mobile directory. 

that's why lint staged is checking where the modified files are and is only triggering pretier on those files using the respective package config.

vseehausen:
> Why this again? Last time I looked running prettier globally runs apparently with different setup than the prettier in the individual pacakges leading to problems. You are modifiying things in mobile only for them to be changed the next time I run yarn prettier from the mobile directory.
> 
> that's why lint staged is checking where the modified files are and is only triggering pretier on those files using the respective package config.

Alright, but are you sure these files would be overwritten by your prettier locally? Can you check? (I am still confused about how to best configure prettier.)

katspaugh (changes requested):
@vseehausen going forward, PRs with the description filled out incompletely won't be accepted. The description should be sufficient to understand what a PR does w/o looking at the code.
If you don't want to fill out all the sections, you can remove them.

compojoom:
I'm closing this. I think I've finally properly setup prettier in https://github.com/safe-global/safe-wallet-monorepo/pull/6849

PR #6847, opened 2025-12-15
fix(portfolio): remove polling and refetch on history change

katspaugh:
## What it solves

Resolves https://linear.app/safe-global/issue/WA-984/optimize-calls-to-portfolio-endpoint

Reduces Zerion API costs by not refetching every 60 seconds. Only on tab focus and when tx history changes. 

Also remove refetchOnFocus because it triggers when connecting a wallet (because the wallet extension is a different window).


katspaugh:
The failing AccountCenter test seems to be unrelated, investigating.

PR #6851, opened 2025-12-15
fix: run portfolio refetch only when it ran before

vseehausen:
## What it solves

Resolves an edge case: When switching Safes, we have a runtime condition, where the refetch due to tx history change could trigger a refetch. This refetch returns an error, when the actual call didn't happen before.

## How this PR fixes it

Only refetches/schedules a refetch when the original call did happen before.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/features/portfolio/hooks/usePortfolioRefetchOnTxHistory.ts
+++ b/apps/web/src/features/portfolio/hooks/usePortfolioRefetchOnTxHistory.ts
@@ -10,7 +10,7 @@ import useSafeInfo from '@/hooks/useSafeInfo'
  */
 const usePortfolioRefetchOnTxHistory = (): void => {
   const { safe } = useSafeInfo()
-  const { refetch, fulfilledTimeStamp } = useRefetch()
+  const { refetch, fulfilledTimeStamp, shouldUsePortfolioEndpoint } = useRefetch()
   const timeoutRef = useRef<NodeJS.Timeout | null>(null)
   const prevTxHistoryTagRef = useRef<string | null | undefined>(undefined)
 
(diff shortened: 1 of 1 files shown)


apps/web/src/features/portfolio/hooks/usePortfolioRefetchOnTxHistory.ts
```diff
@@ -28,14 +28,19 @@ const usePortfolioRefetchOnTxHistory = (): void => {
 
     prevTxHistoryTagRef.current = safe.txHistoryTag
 
+    // Skip if portfolio endpoint isn't active or no successful fetch yet
+    if (!shouldUsePortfolioEndpoint || !fulfilledTimeStamp) {
+      return
+    }
+
```
compojoom, line 35: you removed shouldUsePortfolioEndpoint from the dependency array, so now the effect will never rerun when  shouldUsePortfolioEndpoint changes?

vseehausen: When ` shouldUsePortfolioEndpoint` is true, this effect will run.

I added it to the dependency array.

apps/web/src/features/portfolio/hooks/usePortfolioRefetchOnTxHistory.ts
```diff
@@ -10,7 +10,7 @@ import useSafeInfo from '@/hooks/useSafeInfo'
  */
 const usePortfolioRefetchOnTxHistory = (): void => {
```
katspaugh, line 11: Why is this hook so overcomplicated?

vseehausen: Sorry I don't know what to do with this feedback.

PR #6866, opened 2025-12-17
fix: load counterfactual balance when portfolio is empty

vseehausen:
## What it solves

Resolves: https://linear.app/safe-global/issue/WA-1174/empty-wallet-has-no-portfolio-value-show

## How this PR fixes it

Uses tx balances (that revert to cf balances) when portfolio is empty. Before this fix, this was only the case for deployed Safes, which resulted in the reported issue.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/hooks/useHiddenTokenCounts.ts
+++ b/apps/web/src/hooks/useHiddenTokenCounts.ts
@@ -71,13 +71,5 @@ export const useHiddenTokenCounts = (): HiddenTokenCounts => {
       hiddenByTokenList,
       hiddenByDustFilter,
     }
-  }, [
-    shouldFetchAllTokens,
-    allTokensBalances,
-    allTokensLoading,
-    currentBalances.items,
-    hiddenTokens,
-    hideDust,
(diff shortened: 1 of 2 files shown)


apps/web/src/components/balances/ManageTokensButton/ManageTokensMenu.tsx
```diff
@@ -113,29 +115,31 @@ const ManageTokensMenu = ({
...
-        onClick={handleToggleHideDust}
-        data-testid="hide-small-balances-menu-item"
-      >
-        <Box className={css.menuItemContent}>
-          <Box className={css.menuItemLeft}>
-            <Typography variant="body2">Hide small balances</Typography>
-            <InfoTooltip
-              title={<Typography>Hide tokens with a value less than ${DUST_THRESHOLD}</Typography>}
-              data-testid="hide-small-balances-info-tooltip"
+      {safe.deployed && (
```
vseehausen, line 118: Hide the dust filter for counter-factual Safes.

PR #6903, opened 2025-12-23
fix: next tx execute button shouldnt be disabled on prev's indexind state

0xferrous:
## What it solves

The UI blocks the user from executing next tx even when there is no reason to block it. "Indexing" state is offchain indexing, the tx has already executed onchain.

Resolves:

## Checklist

---

## CLA signature

---

> [!NOTE]
> Ensures the UI doesn't block execution when a previous tx is only being indexed off-chain.
> 
> - Updates `useIsPending` to import `PendingStatus` and return `false` when `pendingTx.status === PendingStatus.INDEXING`
> - Falls back to pending based on presence of `pendingTx` for other statuses
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 30173d132e5deec50c1259b8f3795049508c3f3a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/web/src/hooks/useIsPending.ts
+++ b/apps/web/src/hooks/useIsPending.ts
@@ -1,9 +1,16 @@
 import { useAppSelector } from '@/store'
-import { selectPendingTxs } from '@/store/pendingTxsSlice'
+import { PendingStatus, selectPendingTxs } from '@/store/pendingTxsSlice'
 
 const useIsPending = (txId: string): boolean => {
   const pendingTxs = useAppSelector(selectPendingTxs)
-  return !!pendingTxs[txId]
+  const pendingTx = pendingTxs[txId]
+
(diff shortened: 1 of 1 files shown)


katspaugh:
@LucieFaire AFAIR it was done this way because of "immediately executed" txs in 1/X Safes which don't get proposed to the queue first. Transaction service has to pick them up directly from the blockchain. If you start a new tx while it's still indexing, it might not get the right "recommended" nonce from the service in this case. Please make sure that this hook isn't used for blocking the creation of new transactions too.

0xferrous:
The tx queues fine when the tx is indexing, atleast thats my experience, which is why i notice the execute button on the next tx is disabled.

PR #6908, opened 2025-12-25
fix(web): pending tx saved with wrong Safe address on Safe switch

katspaugh:
## Summary

Resolves https://linear.app/safe-global/issue/WA-1193/pending-tx-state-stored-under-wrong-safe-after-switch

Fixes a critical bug where pending transactions were being stored with the wrong Safe address when users switched Safes during transaction indexing. This caused issues in 1/X Safes where pending transactions incorrectly blocked follow-up transaction executions.

**Root cause**: Event handlers in `useTxPendingStatuses` captured `chainId` and `safeAddress` from hook closure, which would update when switching Safes. In-flight events from the original Safe would then be handled with the new Safe's address.

**Solution**: Made transaction events self-contained by embedding `chainId` and `safeAddress` directly in event payloads at dispatch time. Event handlers now read these values from event details instead of closure.

## Changes

- Added `SafeContext` type (chainId, safeAddress) to transaction events in `txEvents.ts`
- Updated all event dispatchers in `txMonitor.ts` and `dispatch.ts` to include SafeContext fields (~49 call sites)
- Updated event handlers in `useTxPendingStatuses.ts` to use `detail.chainId`/`detail.safeAddress`
- Updated production code call sites (hooks.ts, ExecuteThroughRoleForm, ReviewBatch, SpeedUpModal, txHistorySlice)
- Updated all test files to include required event fields

## Test plan

- [x] Type-check passes (`yarn workspace @safe-global/web type-check`)
- [x] Unit tests pass (`yarn workspace @safe-global/web test`)
- [x] Manual testing: Start transaction on Safe A, switch to Safe B during PROCESSING/INDEXING state, verify pending tx stored under Safe A's address in Redux
- [ ] Manual testing: Verify 1/1 Safe blocking behavior works correctly after Safe switch

### DIFF
--- a/.github/workflows/web-storybook-screenshots.yml
+++ b/.github/workflows/web-storybook-screenshots.yml
@@ -61,11 +61,16 @@ jobs:
           BASE_SHA=${{ github.event.workflow_run.pull_requests[0].base.sha }}
           HEAD_SHA=${{ github.event.workflow_run.head_sha }}
 
-          # Get all changed files in web app
+          # Get all changed files in web app (excluding test files for now)
           ALL_CHANGED=$(git diff --name-only $BASE_SHA $HEAD_SHA | grep -E '^apps/web/src/.*\.(tsx?|jsx?)$' | grep -v '\.test\.' || true)
+
+          # Get changed story test files and snapshot files separately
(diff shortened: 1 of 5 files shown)


katspaugh:
@claude review

PR #6937, opened 2026-01-08
feat(web): use Safe Overview v2 endpoint

katspaugh:
Resolves https://linear.app/safe-global/issue/GRO-146/use-small-endpoint-for-wallet-balances

## Summary

Add support for the new `/v2/safes` endpoint which uses Zerion for balance data on supported chains, providing more accurate portfolio values.

### Key changes:

- **Store package**: Export both v1 (default) and v2 endpoints with automatic chunking (max 10 safes per request)
- **Web app**: Check `PORTFOLIO_ENDPOINT` feature flag per-chain and route to appropriate endpoint
- **Mobile**: Continues using v1 endpoint (backward compatible, no changes needed)

### Architecture:

```
Web App (safeOverviews.ts)
├── shouldUseV2(state, chainId)     # Check feature flag
├── getOverviewEndpoint(useV2)       # Select v1 or v2
├── executeOverviewQuery(...)        # Dispatch + cleanup
│
├── getSafeOverview                  # Single safe → check flag → call endpoint
└── getMultipleSafeOverviews         # Multiple safes → group by flag → parallel fetch

Store Package (safes/index.ts)
├── additionalSafesRtkApi            # v1 endpoint with chunking
└── additionalSafesRtkApiV2          # v2 endpoint with chunking
```

### Multi-chain handling:

When fetching overviews for safes across multiple chains, each safe's chain is checked individually. Safes are grouped by v1/v2 and fetched in parallel, ensuring correct behavior when some chains have the feature enabled and others don't.

Related CGW PR: https://github.com/safe-global/safe-client-gateway/pull/2840

## Test plan
- [x] Type-check passes for store and web packages
- [x] All safeOverviews unit tests pass (9/9)
- [x] Mobile tests unaffected (uses v1 by default)
- [ ] Manual testing: verify Safe overview balances load correctly in the UI


katspaugh:
@claude review

katspaugh:
## Addressed Review Feedback

Thanks for the reviews! I've pushed changes to address the feedback:

### 1. Handle v1/v2 failures independently (Codex)
Changed `Promise.all` to `Promise.allSettled` in `processQueuedItems`. Now a transient failure in one endpoint won't incorrectly fail safes that could succeed on the other endpoint. Users with mixed-chain safes will see partial results when only one endpoint is down.

### 2. Error handling consistency (Claude)
Fixed to use `asError(error).message` consistently throughout the file.

### 3. V2 endpoint test coverage (Copilot/Claude)
Added new test suite `v2 endpoint routing` with 4 tests:
- Verifies v2 endpoint is called when `PORTFOLIO_ENDPOINT` feature flag is enabled
- Verifies v1 endpoint is used when feature flag is not enabled  
- Tests mixed v1/v2 routing based on per-chain feature flags
- Tests independent failure handling between v1 and v2

The test file now mocks both endpoints (`additionalSafesRtkApi` and `additionalSafesRtkApiV2`) plus the chains config selector.

---

**Note:** The Codex comment about `request_id` in `ThreatAnalysisResponseDto` is unrelated to this PR's changes - that file was modified as part of the OpenAPI schema regeneration. If that's a concern, it should be addressed in a separate PR or by updating the backend schema.

apps/web/src/store/__tests__/safeOverviews.test.ts
```diff
@@ -467,48 +458,227 @@ describe('safeOverviews', () => {
...
+      expect(mockedInitiateV2).toHaveBeenCalledWith({
+        safes: [`10:${request.safes[1].address}`],
         currency: 'usd',
         trusted: false,
         excludeSpam: true,
         walletAddress: undefined,
       })
     })
+
+    it('should handle v1 failure independently from v2', async () => {
```
vmosharova, line 630: I would add some extra tests in addition to the existing ones:
- v2 fails while v1 succeeds (the opposite of this ^ test);
- tests for the batching behavior, in particular for the 300ms timeout;
- multiple Safes are requested, only part of them is returned;
- the Safe address exists but on another chain than we request.

katspaugh: Added more tests.

katspaugh:
## Addressing Remaining Copilot Review Comments

### 1. Resolve/reject logic (comment on line 180)
This is a **false positive**. The current if-else structure is correct:

```typescript
if ('data' in result) {
  resolve(result.data)
} else {
  reject(result.error)
}
```

The `else` branch is outside the `if` block, not inside it. The logic correctly resolves when data is present and rejects otherwise.

### 2. Queue batching parameters (comment on line 107)
This is **by design**. The batching is per-session where:
- `currency` comes from the global Redux state (`selectCurrency`)
- `walletAddress` is the currently connected wallet
- `dispatchV1/dispatchV2` are Redux dispatch functions that are identical across the app

In practice, all queued items within the 300ms batch window will share these values since they come from the same user session. Different users/sessions have separate JS contexts and separate queues.

PR #6948, opened 2026-01-12
fix(SafeWalletProvider): handle undefined gas value in tx params

tmjssz:
## What it solves

Resolves: [WA-1233](https://linear.app/safe-global/issue/WA-1233/crash-on-malicious-approval-tx)

Fixes crashes caused by invalid `safeTxGas` values (`NaN` or `undefined`) when processing malicious approval transactions with Hypernative guard enabled.

## How this PR fixes it

The issue was caused by the `safeTxGas` parameter being set to `NaN`, which resulted in an error when the value was converted to a number.
This was resolved by updating the Safe transaction parameters so that the gas value defaults to 0 when it is undefined.

1. **Transaction creation (web & mobile)**: Added validation to convert `NaN` or `'NaN'` `safeTxGas` values to `'0'`
2. **SafeWalletProvider**: Fixed `eth_sendTransaction` to default undefined gas to `0` instead of `NaN`
3. **Hypernative hook**: Added `skip` condition check to prevent creating request objects when analysis is skipped

## How to test it

1. Use [Blockaid examples](https://examples.blockaid.io/) and send a Malicious Approval to a Safe with a HN guard enabled
2. Verify no error is thrown when the tx flow opens

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/services/tx/tx-sender/create.test.ts
+++ b/apps/mobile/src/services/tx/tx-sender/create.test.ts
@@ -79,6 +79,91 @@ describe('create.ts', () => {
         'The Safe SDK could not be initialized. Please be aware that we only support v1.0.0 Safe Accounts and up.',
       )
     })
+
+    it('converts NaN safeTxGas to "0"', async () => {
+      const txParams: SafeTransactionDataPartial = {
+        to: generateChecksummedAddress(),
+        value: '0',
+        data: '0x',
+        safeTxGas: Number.NaN as unknown as string,
+      }
+      const mockTx = createMockSafeTx()
+      mockSafeSDK.createTransaction.mockResolvedValue(mockTx)
+
+      await createTx(txParams)
+
+      expect(mockSafeSDK.createTransaction).toHaveBeenCalledWith({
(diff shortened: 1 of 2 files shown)


compojoom (approved):

apps/web/src/services/tx/tx-sender/create.ts
```diff
@@ -13,6 +13,7 @@ import { getAndValidateSafeSDK } from './sdk'
  */
 export const createTx = async (txParams: SafeTransactionDataPartial, nonce?: number): Promise<SafeTransaction> => {
   if (nonce !== undefined) txParams = { ...txParams, nonce }
+  if (Number.isNaN(txParams.safeTxGas) || txParams.safeTxGas === 'NaN') txParams = { ...txParams, safeTxGas: '0' }
```
katspaugh, line 16: This is redundant, no? `isNaN('NaN') === true` and `isNaN('10') === false`.

tmjssz: I get this:
Image

PR #6957, opened 2026-01-13
fix(SafeAppsTx): ensure safeTxGas is a string

tmjssz:
## What it solves
Resolves: [WA-1236](https://linear.app/safe-global/issue/WA-1236/safetxgas-is-sent-in-wrong-format-to-hn)
Hypernative assessments fail when `safeTxGas` is provided as a number.

## How this PR fixes it

Convert the value to string to ensure correct typing per `SafeTransaction` and proper request handling.

## How to test it
1. Open a Safe with a HN guard
2. Connect the Safe with https://examples.blockaid.io/
3. Send any example from Bloackaid
4. Verify that the Hypernative assessment does not fail with a 400 error

## Screenshots
Screenshot 2026-01-13 at 11 21 17

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/tx-flow/flows/SafeAppsTx/ReviewSafeAppsTx.tsx
+++ b/apps/web/src/components/tx-flow/flows/SafeAppsTx/ReviewSafeAppsTx.tsx
@@ -32,7 +32,6 @@ const ReviewSafeAppsTx = ({
       const tx = isMultiSend ? await createMultiSendCallOnlyTx(txs) : await createTx(txs[0])
 
       if (params?.safeTxGas !== undefined && !Number.isNaN(params.safeTxGas)) {
-        // FIXME: do it properly via the Core SDK
         tx.data.safeTxGas = String(params.safeTxGas)
       }
 


apps/web/src/components/tx-flow/flows/SafeAppsTx/ReviewSafeAppsTx.tsx
```diff
@@ -33,8 +33,7 @@ const ReviewSafeAppsTx = ({
 
       if (params?.safeTxGas !== undefined && !Number.isNaN(params.safeTxGas)) {
         // FIXME: do it properly via the Core SDK
```
compojoom: Is the comment still valid. What does "do it properly via the Core SDK" mean?

tmjssz: Good question. It was added [here](https://github.com/safe-global/safe-wallet-monorepo/pull/2067) but I don't really know what it means

tmjssz: I'll remove the comment as it's very old and not clear

PR #6976, opened 2026-01-15
chore: add Datadog RUM for early performance monitoring

vseehausen:
## What it solves

Adds Datadog Real User Monitoring (RUM) to capture performance metrics like First Meaningful Paint (FMP) as early as possible in the application lifecycle.

## How it solves it

- Initializes Datadog RUM at module load time (not in a React hook) to ensure the earliest possible initialization
- Uses side-effect import in `_app.tsx` for automatic initialization
- Datadog Logs are always initialized when `NEXT_PUBLIC_DATADOG_CLIENT_TOKEN` is set
- Datadog RUM is additionally initialized when `NEXT_PUBLIC_DATADOG_APPLICATION_ID` is also set
- Includes `@datadog/browser-rum-react` plugin for React-specific instrumentation

## Changes

- Upgrades `@datadog/browser-logs` to ^6.25.3 for compatibility
- Adds `@datadog/browser-rum` and `@datadog/browser-rum-react` packages
- Refactors `datadog/index.ts` for early client-side initialization
- Adds `NEXT_PUBLIC_DATADOG_APPLICATION_ID` to constants and README

## Environment Variables

| Variable | Description |
|----------|-------------|
| `NEXT_PUBLIC_DATADOG_CLIENT_TOKEN` | Datadog client token (required for Logs and RUM) |
| `NEXT_PUBLIC_DATADOG_APPLICATION_ID` | Datadog RUM application ID (required for RUM) |


apps/web/src/services/datadog/index.ts
```diff
@@ -1,36 +1,49 @@
...
+    console.debug('[Datadog] initialized', { logs: true, rum: Boolean(DATADOG_APPLICATION_ID) })
+  }
 }
 
 export const logger = datadogLogs.logger
+
+// Client-side only
+if (typeof window !== 'undefined') {
+  initDatadog()
+}
```
compojoom, line 43: I find this a bit weird. In app.tsx we just import this file and do nothing and here we have this "magic". 
I think that it will be more logical to move this init call to the _app.tsx.

apps/web/src/services/datadog/index.ts
```diff
@@ -1,36 +1,49 @@
-import { useEffect } from 'react'
 import { datadogLogs } from '@datadog/browser-logs'
-import { DATADOG_CLIENT_TOKEN, IS_PRODUCTION } from '@/config/constants'
+import { datadogRum } from '@datadog/browser-rum'
+import { reactPlugin } from '@datadog/browser-rum-react'
+import { APP_ENV, DATADOG_APPLICATION_ID, DATADOG_CLIENT_TOKEN, IS_PRODUCTION } from '@/config/constants'
+import packageJson from '../../../package.json'
```
compojoom: I saw that we are doing this in other locations as well, but I think it is bad. I've made PR to address this: https://github.com/safe-global/safe-wallet-monorepo/pull/6979

apps/web/src/services/datadog/index.ts
vseehausen: Nice, thanks for the comment! I was wondering about this.

compojoom:
What are we doing with this PR? If we are accepting that @gfendres's PR on the topic is the way to go, let's close this @vseehausen

vseehausen:
You are right. Closed in favor of https://github.com/safe-global/safe-wallet-monorepo/pull/6788.

PR #6986, opened 2026-01-16
fix(web): storybook not rendering

compojoom:
## What it solves
In https://github.com/safe-global/safe-wallet-monorepo/pull/6979 we changed the way we use version from package.json. It was necessary to pass the version to the storybook config as well.

## Checklist

---

## CLA signature


vmosharova (approved):
Works like a charm - thanks!

PR #7017, opened 2026-01-22
feat(web): filter nested safes by deployer and add validation UI

katspaugh:
## Summary
- Add validation for nested safes to show only those deployed by allowed addresses (parent owners, parent Safe, or parent deployer)
- Safes not meeting this criteria are still shown but marked as "invalid" with reduced opacity (0.5), warning icon with tooltip, moved to "Show all" accordion section, and hidden pending transaction counts
- New `useFilteredNestedSafes` hook with lazy loading (only fetches when list is opened) and limit of 10 nested safes to check

## Test plan
- [ ] Open a Safe that has nested Safes
- [ ] Verify valid nested Safes appear in the initial list
- [ ] Verify "invalid" nested Safes appear dimmed in the "Show all Nested Safes" accordion
- [ ] Verify warning icon appears on the left of invalid Safes with appropriate tooltip
- [ ] Verify pending transactions are hidden for invalid Safes
- [ ] Check both sidebar popover and settings page display correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7018, opened 2026-01-22
feat(web): allow users to hide nested safes from the list

compojoom:
## What it solves

Users cannot hide unwanted nested Safes from their nested safe list, leading to cluttered UI when they have many child Safes.

Resolves: https://linear.app/safe-global/issue/WA-1340/verification-of-nested-safes-curation-for-unauthorized-safe

## How this PR fixes it

Adds a "Manage Safes" feature to the Nested Safes popover that allows users to:
- Enter manage mode via a settings icon button
- Select/deselect nested Safes to hide using checkboxes
- See a counter showing how many Safes are selected to hide
- Save or cancel changes

Hidden Safe preferences are persisted to localStorage via the Redux `settingsSlice`, following the same pattern as the existing "hide tokens" functionality.

## How to test it

1. Open a Safe that has nested Safes (e.g., `sep:0xd98bC311B693E8e457732C0Bb097B92b95db7100`)
2. Click the "Nested Safes" button in the sidebar to open the popover
3. Click the settings/manage icon to enter manage mode
4. Check one or more Safes to hide, observe the counter updates
5. Click "Save" - verify the hidden Safes disappear from the list and the sidebar count updates
6. Re-enter manage mode - verify hidden Safes appear with checkboxes pre-checked
7. Uncheck a Safe and save - verify it reappears in the list
8. Test "Cancel" discards unsaved changes
9. Refresh the page - verify hidden preferences persist

## Screenshots

https://github.com/user-attachments/assets/4264f7cb-b4c1-4c0b-b406-c1e4a818a017

## Checklist

---

## CLA signature

---

> [!NOTE]
> **Medium Risk**
> Touches sidebar navigation and Safe list rendering while adding persisted visibility state and async validation calls, so UI regressions or incorrect hide/show decisions could affect discoverability of nested safes.
> 
> **Overview**
> Adds **Nested Safes curation**: nested safes are now validated (by deployer) and *auto-hidden* when deemed suspicious, with a new popover **Manage Safes** mode to toggle hide/show via checkboxes, warning icons, save/cancel controls, loading state, and hidden-count messaging.
> 
> Persists user choices per parent Safe in Redux `settingsSlice` via new `manuallyHiddenSafes` and `overriddenAutoHideSafes` maps, with new hooks (`useFilteredNestedSafes`, `useNestedSafesVisibility`, `useManageNestedSafes`) to combine validation + user preferences and update the sidebar badge/count based on visible safes.
> 
> Refactors account/safe list UI to a new `AccountItem` compound component and shared data hooks, replacing `SingleAccountItem`/`AccountInfoChips` in multiple surfaces (sidebar lists, multi-chain items, WC chain switch, spaces add accounts), and adds unit + Cypress regression coverage for the new nested-safe behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 93aabf2d9b67858b377922b3938e82236a2fd970. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/web/src/features/myAccounts/components/AccountItem/AccountItemPinButton.tsx
+++ b/apps/web/src/features/myAccounts/components/AccountItem/AccountItemPinButton.tsx
@@ -43,8 +43,8 @@ function AccountItemPinButton(props: AccountItemPinButtonProps) {
   const address = isSingleChain ? singleSafe!.address : (multiSafes[0]?.address ?? '')
   const name = props.name ?? (isSingleChain ? singleSafe!.name : undefined)
 
-  // Derive isPinned
-  const isPinned = isSingleChain ? singleSafe!.isPinned : multiSafes.every((s) => s.isPinned)
+  // Derive isPinned - use .some() to match _buildMultiChainSafeItem in useAllSafesGrouped.ts
+  const isPinned = isSingleChain ? singleSafe!.isPinned : multiSafes.some((s) => s.isPinned)
 
   // Call both hooks (hooks must be called unconditionally)
   const singleChainActions = useSingleChainPinActions({
--- a/apps/web/src/features/myAccounts/components/AccountItem/__tests__/AccountItemPinButton.test.tsx
+++ b/apps/web/src/features/myAccounts/components/AccountItem/__tests__/AccountItemPinButton.test.tsx
@@ -74,6 +74,20 @@ describe('AccountItemPinButton', () => {
       expect(screen.getByTestId('bookmark-icon')).toBeInTheDocument()
     })
 
+    it('should show as pinned when some (but not all) chains are pinned', () => {
+      const safeItems = [
+        createMockSafeItem({ chainId: '1', isPinned: true }),
+        createMockSafeItem({ chainId: '137', isPinned: false }),
+        createMockSafeItem({ chainId: '10', isPinned: false }),
+      ]
+
+      render(<AccountItemPinButton safeItems={safeItems} />)
+
+      const iconButton = screen.getByTestId('bookmark-icon')
+      const svgIcon = iconButton.querySelector('.MuiSvgIcon-colorPrimary')
+      expect(svgIcon).toBeInTheDocument()
+    })
+
     it('should stop event propagation when clicked', () => {
       const safeItems = [createMockSafeItem({ chainId: '1' }), createMockSafeItem({ chainId: '137' })]
 


apps/web/src/components/sidebar/NestedSafesPopover/index.tsx
```diff
@@ -109,32 +120,52 @@ export function NestedSafesPopover({
...
+              '&::-webkit-scrollbar-track': {
+                background: 'transparent',
+              },
+              '&::-webkit-scrollbar-thumb': {
+                backgroundColor: 'var(--color-border-light)',
+                borderRadius: '3px',
+                '&:hover': {
+                  backgroundColor: 'var(--color-border-main)',
+                },
+              },
```
katspaugh: This belongs to the component/design system, no?

katspaugh (commented):
Looks great! Could you also port [this change](https://github.com/safe-global/safe-wallet-monorepo/pull/7017/changes#diff-bd55d9f8a5d32b67f80357cd91a9023f3b2d8f35eea60b62b470a9cda53c1f59R31)?

katspaugh:
@claude review

apps/web/src/store/settingsSlice.ts
```diff
@@ -46,6 +54,10 @@ export const initialState: SettingsState = {
 
   hiddenTokens: {},
 
+  hiddenNestedSafes: {},
+
+  userUnhiddenNestedSafes: {},
```
katspaugh: Having both seems a bit redundant. If the goal is to differentiate a cache for "unverified" Safes (which is what I assume `hiddenNestedSafes` is) vs Safes that user hides/shows, then I would rename these fields to:

```
unverifiedSafes: {},
unhiddenNestedSafes: {}
```

And this still has open questions:
* Should the unverifiedSafes cache live in the user preferences slice?
* Should the user be able to hide "verified" Safes too? Then "unhidden" doesn't work.
* We need to differentiate Safes not only by address but also by chainId, perhaps?

compojoom: I renamed those 2, hope that they are a bit more clear now. 
None of those serve as a cache for "unverified" safes. We compute whether a safe is verified or not on the fly in the hook. The problem is that we want to start with "problematic" safes hidden and that is why I need to keep 2 objects. 

We load the safes from cgw, we look into `unverifiedSafes`(renamed to `manuallyHiddenSafes` now) and if there no match - we show the safes. But now we also want to by default hide all "problematic" safes. So we look in the `manuallyHiddenSafes`(now renamed to `overriddenAutoHideSafes`) and since it's empty we hide those. 

When the user manages the safe we populate either `manuallyHiddenSafes` or `overrridenNestedSafes`. And this way if a user decided that a "problematic" safe is ok, we show it. 

So both objects need to live in the preferences slice. 

> We need to differentiate Safes not only by address but also by chainId, perhaps?
I explicitly decided not to deal with this. My reasoning: if a safe is "problematic" on one, better to treat it as generally problematic 🤷

PR #7022, opened 2026-01-22
refactor(web): add knip for unused exports detection + remove dead code

katspaugh:
Extracted from #7020 (thanks @compojoom!)

## Summary
- Add knip to detect unused exports in barrel files
- Helps prevent accidental bundling of heavy internal modules
- Adds CI check to catch regressions

## Changes
- `apps/web/knip.json` - Configuration for knip
- `apps/web/package.json` - Added `knip` and `knip:exports` scripts
- `.github/workflows/web-checks.yml` - Added `knip-exports` CI job

## Usage
```bash
yarn workspace @safe-global/web knip:exports
```

## Test plan
- [x] `yarn knip:exports` runs successfully and reports unused exports
- [x] CI workflow includes new job


diega:
Hey! It looks like this PR accidentally removed the service worker entry point for Firebase push notifications.

The files `src/service-workers/index.ts` and `src/service-workers/firebase-messaging/firebase-messaging-sw.ts` were flagged as dead code by knip, but they were actually compiled separately by `@ducanh2912/next-pwa` via the `customWorkerSrc` option in `next.config.mjs` — they weren't part of the main app bundle's import graph, which is why knip didn't see them as used.

Without these files, `public/firebase-messaging-sw.js` remains empty (0 bytes) after build, so `onBackgroundMessage` is never registered and push notifications are silently dropped by the browser.

The remaining files in `src/service-workers/firebase-messaging/` (`notifications.ts`, `notification-mapper.ts`, `gateway-utils.ts`, `webhook-types.ts`) are now effectively dead code themselves since nothing imports them.

Found this while setting up a self-hosted deployment with push notifications enabled.

PR #7038, opened 2026-01-26
refactor(web): hypernative feature driven design v2

compojoom:
## What it solves

Resolves:

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
+++ b/apps/web/src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
@@ -13,7 +13,8 @@ import { shortenAddress } from '@safe-global/utils/utils/formatters'
 import ImageFallback from '../../ImageFallback'
 import css from './styles.module.css'
 import { ContactSource } from '@/hooks/useAllAddressBooks'
-import { HypernativeTooltip } from '@/features/hypernative/components/HypernativeTooltip'
+import { useLoadFeature } from '@/features/__core__'
+import { HypernativeFeature } from '@/features/hypernative'
 import SafeShieldIcon from '@/public/images/safe-shield/safe-shield-logo-no-text.svg'
 
 export type EthHashInfoProps = {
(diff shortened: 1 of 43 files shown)


apps/web/src/features/hypernative/feature.ts
```diff
@@ -0,0 +1,142 @@
...
+      lazy(() => import('./components/HnQueueAssessment').then((m) => ({ default: m.HnQueueAssessment }))),
+    ) as typeof HnQueueAssessment,
+
+    HnQueueAssessmentBanner: withSuspense(
+      lazy(() => import('./components/HnQueueAssessmentBanner').then((m) => ({ default: m.HnQueueAssessmentBanner }))),
+    ) as typeof HnQueueAssessmentBanner,
+
+    QueueAssessmentProvider: withSuspense(
+      lazy(() => import('./components/QueueAssessmentProvider').then((m) => ({ default: m.QueueAssessmentProvider }))),
+    ) as typeof QueueAssessmentProvider,
```
katspaugh, line 115: Thanks for making this PR, @compojoom!

Looks like it's not explicit enough in the documentation, so AI went off-track here. There should be only one single dynamic import in handle.ts, not many individual ones.

apps/web/src/components/common/EthHashInfo/SrcEthHashInfo/index.tsx
```diff
@@ -135,10 +137,10 @@ const SrcEthHashInfo = ({
               {name}
             </Box>
 
-            {showShieldIcon ? (
-              <HypernativeTooltip placement="right">
+            {showShieldIcon && hypernative ? (
+              <hypernative.components.HypernativeTooltip placement="right">
                 <SvgIcon component={SafeShieldIcon} inheritViewBox sx={safeShieldSvgStyles} />
-              </HypernativeTooltip>
+              </hypernative.components.HypernativeTooltip>
```
katspaugh, line 143: Unrelated to this refactoring per se – this feature shouldn't be included in the component that is used dozens of times throughout the app. The shield icon is only displayed in one place (SafeHeader) and should be included there instead. Correct me if I'm wrong @vmosharova @tmjssz

vmosharova: Agree, the Safe Shield icon + HN tooltip should be shown only in the parent `SafeHeaderInfo` (sidebar header), not in `SrcEthHashInfo` or `EthHashInfo` which are seen in many other places across the app.

PR #7077, opened 2026-02-02
feat(web): reenable Trezor wallet support

PooyaRaki:
## What it solves
Implements full native `Trezor` hardware wallet integration for `Safe{Wallet}` web, including a custom `EIP-1193` provider, transaction signing.

Resolves: [#WA-1381](https://linear.app/safe-global/issue/WA-1381/trezor-wallet-native-support)

## Changes
- Connect your Trezor — Trezor devices now appear in the wallet connection flow alongside Ledger and other wallets.
- Persistent sessions — Your connected `Trezor` account is remembered across page reloads, so you don't need to re-authorize your device every visit.
- Full signing support — Covers all standard signing operations: transactions, messages, and typed data (`EIP-712`).

## How to test it
- Connect your `Trezor` wallet to your `Safe` using the native module.
- Execute different type of tracnactions.

### DIFF
--- a/apps/web/src/utils/transaction-guards.ts
+++ b/apps/web/src/utils/transaction-guards.ts
@@ -444,7 +444,7 @@ export const isSafeUpdateTxData = (data?: TransactionData | null): boolean => {
     isMultiSendCalldata(data.hexData) &&
     Boolean(
       Array.isArray(data.dataDecoded?.parameters?.[0]?.valueDecoded) &&
-      data.dataDecoded.parameters[0].valueDecoded.some((tx) => tx.dataDecoded?.method === 'changeMasterCopy'),
+        data.dataDecoded.parameters[0].valueDecoded.some((tx) => tx.dataDecoded?.method === 'changeMasterCopy'),
     )
   )
 }


apps/web/src/features/hypernative/hooks/useHypernativeOAuth.ts
```diff
@@ -137,7 +137,7 @@ export function clearPkce(): void {
 async function generateCodeChallenge(verifier: string): Promise<string> {
   const encoder = new TextEncoder()
   const data = encoder.encode(verifier)
-  const hash = await crypto.subtle.digest('SHA-256', data)
+  const hash = await crypto.subtle.digest('SHA-256', new Uint8Array(data))
```
katspaugh, line 142: Is this related to Trezor?

PooyaRaki: Yes, It looks unrelated but when I reenabled the Trezor module we got an error from the `digest` method. This line of code fixes that. The functionality is also verified and remains unchanged.

apps/web/src/services/onboard/trezor/module.ts
```diff
@@ -0,0 +1,358 @@
+import type { Transaction } from 'ethers'
+import type { Chain, WalletInit, WalletInterface } from '@web3-onboard/common'
+import type { Account, Asset, ScanAccountsOptions } from '@web3-onboard/hw-common'
+import type { EthereumSignTypedDataMessage, EthereumSignTypedDataTypes } from '@trezor/connect-web'
+import type { TrezorTransaction } from './types'
+import { TREZOR_LIVE_PATH, TREZOR_LEGACY_PATH, DEFAULT_BASE_PATHS, DEFAULT_ASSETS } from './constants'
+import { loadStoredAccount, saveStoredAccount, clearStoredAccount } from './storage'
+import { getTrezorSdk } from './sdk'
```
katspaugh, line 7: We could make this a dynamic import near `const trezorSdk = await getTrezorSdk()` on line 29.

PooyaRaki: What would be the benefit? The Trezor Connect library is lazily loaded inside the wrapper sdk file though.

PooyaRaki:
@claude review

PooyaRaki:
@claude review the new changes and also review this comment about the previous review.
Regarding Claude comments:
1. The complexity metric (34) is real, but the pattern is inherent to Web3-Onboard's closure-based wallet interface — all method handlers share state (`currentChain`, `currentAccount`, `trezorSdk`) via closure, so extracting them to standalone functions would require passing all that state as parameters, trading structural complexity for parameter complexity. The individual handlers are each simple. Logged as tech debt for a future class-based refactor.
2. Valid long-term improvement. Expanding `mapTrezorError` to cover device-disconnect, firmware, and timeout codes requires mapping the full `Trezor` error surface — tracked as a follow-up task.
3. Valid concern. However, args.params is passed through unvalidated from web3-onboard's handler layer — adding Zod here in isolation would be inconsistent. This is a systemic gap in the handler layer worth addressing holistically.
4. `BATCH_SIZE` and `MAX_ZERO_BALANCE_ACCOUNTS` are already extracted as named constants inside deriveAccounts. Since they're only used in that one function, keeping them local is intentional.
5. `trezorConnectInitialized` is only set true after `TrezorConnect.init()` fully resolves, so a throw during init leaves the flag false correctly. disconnect() resets it. No cleanup gap exists.
6. This change is a direct consequence of installing `@trezor/connect-web`. The package polyfills something that changes how `TextEncoder.encode()` is handled — wrapping in new `Uint8Array(data)` was required to restore correct behavior. It belongs in this PR.
7. Expected for a new hardware wallet integration. The feature module is correctly lazy-loaded via webpackMode: "lazy", so the `2.05KB` increase on `/settings/data` only loads when `Trezor` is used.

apps/web/src/features/trezor/components/TrezorHashComparison/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+import {
+  CLOSE_BUTTON_TEXT,
+  DIALOG_DESCRIPTION,
+  DIALOG_MAX_WIDTH,
+  DIALOG_TITLE,
+  HASH_DISPLAY_LIMIT,
+  HASH_DISPLAY_WIDTH,
+} from '../../constants'
+
+const TrezorHashComparison = () => {
```
katspaugh: I assume this is a customized copy of the LedgerHashComparison. Just flagging that maybe the component could be made generic and reused for both.

apps/web/src/features/trezor/index.ts
```diff
@@ -0,0 +1,13 @@
+import type { FeatureHandle } from '@/features/__core__'
+import type { TrezorContract } from './contract'
+
+// Trezor is a core feature - always enabled, not gated by a CGW feature flag.
+// We still use the feature architecture for lazy loading and code organization.
+export const TrezorFeature: FeatureHandle<TrezorContract> = {
+  name: 'trezor',
+  useIsEnabled: () => true,
```
katspaugh: This could check if the wallet is enabled in the remote config (not a feture flag but the enabled wallets array).

apps/web/src/features/trezor/components/TrezorHashComparison/index.tsx
PooyaRaki: I removed the `HashComparison` logic for `Trezor` as the way signing and executing works on `Trezor` is different.
https://github.com/safe-global/safe-wallet-monorepo/pull/7077/commits/2d23f1f61298fea86d25a5e6d5900f5256c3c755

apps/web/src/features/trezor/index.ts
PooyaRaki: Since the only `TrezorFeature` was the hash comparison this file is also removed.
https://github.com/safe-global/safe-wallet-monorepo/pull/7077#discussion_r2918428406

PR #7083, opened 2026-02-03
refactor(web): migrate Stake to new feature architecture

compojoom:
## What it solves

The stake feature was using the legacy import pattern, bypassing lazy loading and bundle optimization. Components were imported directly instead of through the v3 feature architecture pattern.

Resolves: https://linear.app/safe-global/issue/WA-1316/refactor-port-stake-feature-to-architecture-spec

## How this PR fixes it

Migrates the stake feature to v3 architecture by creating:
- `contract.ts` - TypeScript interface defining the feature's public API
- `feature.ts` - Lazy-loaded implementation with flat exports
- `index.ts` - Feature handle via `createFeatureHandle()` + direct hook exports

External consumers now use `useLoadFeature(StakeFeature)` to access components (which return proxy stubs until loaded), while hooks remain directly importable from `@/features/stake`. Internal imports converted to relative paths to avoid circular dependencies.

## How to test it

1. Run `yarn workspace @safe-global/web type-check` - should pass
2. Run `yarn workspace @safe-global/web test` - all 3068 tests should pass
3. Navigate to `/stake` page - should load StakePage component correctly
4. View a staking transaction in tx history - deposit/exit/withdraw details should render

## Screenshots

N/A - no UI changes (internal architecture refactor)

## Checklist

- [x] I've documented how it affects the analytics (if at all) 📊 - No analytics impact
- [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻 - Existing tests updated, all pass

---

## CLA signature


apps/web/src/features/stake/index.ts
```diff
@@ -0,0 +1,66 @@
...
+// ─────────────────────────────────────────────────────────────────
+
+export { getStakeTitle } from './helpers/utils'
+
+// ─────────────────────────────────────────────────────────────────
+// SHARED UI COMPONENTS (always available, not feature-flag dependent)
+// ─────────────────────────────────────────────────────────────────
+
+// InfoTooltip is a generic utility component used across the app
+export { InfoTooltip } from './components/InfoTooltip'
```
katspaugh: > InfoTooltip is a generic utility component used across the app

Why is it inside this feature then?

apps/web/src/features/stake/index.ts
```diff
@@ -0,0 +1,66 @@
...
+// CONSTANTS
+// ─────────────────────────────────────────────────────────────────
+
+export * from './constants'
+
+// ─────────────────────────────────────────────────────────────────
+// HELPER UTILITIES (direct exports for consumers)
+// ─────────────────────────────────────────────────────────────────
+
+export { getStakeTitle } from './helpers/utils'
```
katspaugh, line 59: I see this util is just used in some component, why can't it be injected via `useLoadFeature`?

apps/web/src/components/transactions/TxInfo/index.tsx
```diff
@@ -152,15 +153,15 @@ const TxInfo = ({ info, ...rest }: { info: TransactionInfo; omitSign?: boolean;
...
 
   if (isStakingTxDepositInfo(info)) {
-    return <StakingTxDepositInfo info={info} />
+    return <stake.StakingTxDepositInfo info={info} />
   }
 
   if (isStakingTxExitInfo(info)) {
-    return <StakingTxExitInfo info={info} />
+    return <stake.StakingTxExitInfo info={info} />
   }
```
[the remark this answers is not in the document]
compojoom: Moved those views outside of the feature. It makes more sense to have them closer to the tx history, than in the stake feature.

apps/web/src/features/stake/index.ts
compojoom: moved

apps/web/src/features/stake/index.ts
compojoom: let's keep it like this for now. This utility is only used in the sign message flow. Probably we could move it from stake to there. Undecided right now, but can be improved later.

PR #7097, opened 2026-02-04
feat: fetch gas prices through CGW proxy

vseehausen:
## Summary

- Route gas price oracle requests through the Client Gateway (CGW) endpoint instead of directly calling external oracles
- Enables server-side caching and centralized API key management for gas price oracles
- Maintains fallback to `provider.getFeeData()` for chains without configured oracles

## Changes

- Modified `useDefaultGasPrice.ts` to fetch gas prices via CGW proxy endpoint
- Updated the hook to use the CGW gas price API when available
- Preserved existing fallback behavior for chains without oracle support

## Test plan

- [ ] Verify gas price fetching works on chains with configured oracles (e.g., Ethereum mainnet)
- [ ] Verify fallback to provider.getFeeData() works on chains without oracles
- [ ] Check that gas prices are correctly displayed in transaction flows
- [ ] Confirm no regression in transaction submission with estimated gas prices

### DIFF
--- a/apps/web/jest.config.cjs
+++ b/apps/web/jest.config.cjs
@@ -28,6 +28,7 @@ const customJestConfig = {
     '^.+/markdown/terms/terms\\.md$': '<rootDir>/mocks/terms.md.js',
     isows: '<rootDir>/node_modules/isows/_cjs/index.js',
     '^@safe-global/utils/(.*)$': '<rootDir>/../../packages/utils/src/$1',
+    '^@safe-global/store/(.*)$': '<rootDir>/../../packages/store/src/$1',
   },
   // https://github.com/mswjs/jest-fixed-jsdom
   // without this environment it is basically impossible to run tests with msw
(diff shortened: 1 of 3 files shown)


katspaugh (commented):
Wasn't the idea to just replace the URL in the config-service with a new CGW URL and the front-end would pick it up seamlessly w/o any code changes?

vseehausen:
> Wasn't the idea to just replace the URL in the config-service with a new CGW URL and the front-end would pick it up seamlessly w/o any code changes?

Well, then we could close this one, right?

vseehausen:
Closing this PR as we've switched to a cleaner approach that doesn't require frontend changes.

Instead of having the wallet handle the new CGW response format, we're making the CGW endpoint return the raw Etherscan API format (1:1 passthrough). This means:

1. Config-service will update the oracle URL to point to CGW: `https://safe-client.safe.global/v1/chains/1/gas-price`
2. Frontend keeps working as-is - it receives the same Etherscan format it already expects
3. No code deployment needed in the wallet

See updated CGW PR for details: https://github.com/safe-global/safe-client-gateway/pull/2902

PR #7116, opened 2026-02-06
chore(web): optimistic portfolio API loading

vseehausen:
## Summary
- Portfolio API now fires immediately on page load by extracting `chainId` and `safeAddress` directly from the URL, instead of waiting for chains config and SafeInfo APIs to resolve first
- Eliminates the sequential waterfall: `chains config → SafeInfo → portfolio` is now `chains config ‖ SafeInfo ‖ portfolio` (all parallel)
- Adds `useRawUrlChainId()` hook for static chain lookup from URL (no API wait)
- Uses `useSafeAddressFromUrl()` instead of `useSafeInfo()` in the portfolio hook
- Optimistic `trusted=true` default when chain config hasn't loaded yet (RTK Query re-fetches if wrong)
- DRYs up URL chain parsing via shared `useUrlShortName()` internal hook

## Measurements (local build, prod CGW)

| Branch | Median Start | Median End |
|--------|-------------|------------|
| Optimistic (with optimization) | 348.8ms | 755.5ms | 
| Baseline (reverted) | 938.7ms | 1379.3ms | 
| **Delta (improvement)** | **-589.9ms** (63% earlier start) | **-623.8ms** (45% earlier completion) | 

### Key Findings

1. **Portfolio API starts 590ms earlier** with the optimization (median: 349ms vs 939ms)
2. **Portfolio API completes 624ms earlier** with the optimization (median: 756ms vs 1379ms)
3. **Portfolio API duration is 124ms faster** (median: 310ms vs 434ms) - **28.5% improvement**
4. The optimization successfully eliminates the waterfall delay by allowing the portfolio API to start immediately instead of waiting for chains config
Portfolio and SafeInfo now start at the exact same time. Portfolio complete time improved from ~1313ms median to ~1027ms.

## How to verify with Chrome DevTools

1. Recommendation: Use Chrome
2. Open the same Safe at two different staging deployments:
    - **Optimized**: https://chore_optimistic_portfolio_loading--walletweb.review.5afe.dev/balances?safe=eth:0xA77DE01e157f9f57C7c4A326eeE9C4874D0598b6
    - **Before**: https://revert_pr_7100--walletweb.review.5afe.dev/balances?safe=eth:0xA77DE01e157f9f57C7c4A326eeE9C4874D0598b6 
4. Open DevTools → **Network** tab on each
5. Soft refresh (`Cmd+R`):
6. Filter by **Fetch/XHR** to show only API calls
7. In the filter box, type `portfolio` to isolate the portfolio call
8. Click on `timing`
9. Observe the `Started at` value of the portfolio call in the two tabs you have open.
   - **Optimized**: Should be faster (i.e. <1s)
   - **Before**: Should be slower (i.e. >1s)

Example outpout:

Optimized:
Screenshot 2026-02-06 at 16 57 35

Before:
Screenshot 2026-02-06 at 16 57 56

Alternatively, run this in the Console after page load to get exact timings (comparing the optimized portfolio call to another call):
```js
const r = performance.getEntriesByType('resource').filter(r => r.initiatorType === 'fetch');
const info = r.find(e => e.name.match(/\/safes\/0x[^/]+$/));
const port = r.find(e => e.name.includes('portfolio'));
console.table({
  SafeInfo: { start: Math.round(info?.startTime), end: Math.round(info?.responseEnd) },
  Portfolio: { start: Math.round(port?.startTime), end: Math.round(port?.responseEnd) },
});
```

## Test plan
- [x] Type-check passes
- [x] All 3069 tests pass (344 suites)
- [x] Lint clean (0 errors)
- [ ] Verify portfolio loads correctly on `/home`, `/balances` pages
- [ ] Verify "All tokens" mode still merges portfolio + tx service data
- [ ] Verify counterfactual (undeployed) safes still show balances
- [ ] Check DevTools Network tab confirms parallel API calls

### DIFF
--- a/apps/web/src/hooks/loadables/useLoadBalances.ts
+++ b/apps/web/src/hooks/loadables/useLoadBalances.ts
@@ -110,7 +110,7 @@ const calculateTokensFiatTotal = (items: Balances['items']): string => {
  */
 const useLoadBalances = (): AsyncResult<PortfolioBalances> => {
   const settings = useAppSelector(selectSettings)
-  const hasPortfolioFeature = useHasFeature(FEATURES.PORTFOLIO_ENDPOINT) ?? false
+  const hasPortfolioFeature = useHasFeature(FEATURES.PORTFOLIO_ENDPOINT) !== false
   const isAllTokensSelected = settings.tokenList === TOKEN_LISTS.ALL
 
   const portfolioResult = usePortfolioBalances(!hasPortfolioFeature)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -548,6 +548,7 @@ Avoid these common mistakes when contributing:
 - **RPC issues**: Check that `INFURA_TOKEN` or other RPC provider env vars are set correctly
 - **Build errors**: Check `.next` cache – sometimes `rm -rf apps/web/.next` helps
 - **Storybook issues**: Try `rm -rf node_modules/.cache/storybook`
+- **Using prod CGW locally**: The sidebar has a "Use prod CGW" toggle (bottom-left corner) to switch between staging and production Client Gateway. Enable it when testing a local build against real production data.
 
 ## Code Complexity Guidelines
 
--- a/apps/web/src/features/portfolio/hooks/__tests__/usePortfolioBalances.test.ts
+++ b/apps/web/src/features/portfolio/hooks/__tests__/usePortfolioBalances.test.ts
@@ -3,7 +3,7 @@ import usePortfolioBalances from '@/features/portfolio/hooks/usePortfolioBalance
 import * as useSafeInfo from '@/hooks/useSafeInfo'
 import * as useChains from '@/hooks/useChains'
 import * as useChainId from '@/hooks/useChainId'
-import * as useSafeAddressFromUrl from '@/hooks/useSafeAddressFromUrl'
+import * as useOptimisticSafeAddress from '@/hooks/useOptimisticSafeAddress'
 import * as store from '@/store'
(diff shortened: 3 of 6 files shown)


apps/web/src/hooks/loadables/useLoadBalances.ts
```diff
@@ -110,7 +110,7 @@ const calculateTokensFiatTotal = (items: Balances['items']): string => {
  */
 const useLoadBalances = (): AsyncResult<PortfolioBalances> => {
   const settings = useAppSelector(selectSettings)
-  const hasPortfolioFeature = useHasFeature(FEATURES.PORTFOLIO_ENDPOINT) ?? false
+  const hasPortfolioFeature = useHasFeature(FEATURES.PORTFOLIO_ENDPOINT) !== false
```
katspaugh, line 113: This is the only downside of this approach. It will make a Zerion request regadrdless of the feature flag.

vseehausen: Good catch! This time expected, as is the idea: Optimistic requests. We have Zerion turned on for chains that cover most of our traffic already.

katspaugh: Yes, it's a good pragmatic trade-off, I agree.

Just needs to be acknowledged (perhaps in a comment?) that we're ignoring the feature flag here. It might be confusing if in future we decide to disable it on some chain but it will be still doing these requests anyway.

katspaugh (approved):
🚀

vseehausen:
Closed in favor of https://github.com/safe-global/safe-wallet-monorepo/pull/7122

PR #7122, opened 2026-02-06
refactor(web): pre-fetch chains config at build time for instant startup and SSG support

katspaugh:
## Summary
- Adds a build-time script (`scripts/fetch-chains.ts`) that pre-fetches all chain configurations from the CGW API and writes them to `src/config/__generated__/chains.json`
- Seeds the RTK Query cache synchronously via `preloadedState` in `makeStore()`, so `useGetChainsConfigQuery()` returns data immediately — no loading state, no blocking network request
- Pre-cached chains are available during SSG/SSR as well as client-side rendering, since the seeding is purely synchronous and identical on both server and client (no hydration mismatches)
- Dependent API requests (safe info, balances, indexing status, safe-apps) start in parallel with the runtime chains refetch instead of waiting for it

## How it works
1. **Build time**: `yarn fetch-chains` fetches all paginated chains from CGW, writes JSON to `src/config/__generated__/chains.json`. Falls back to `[]` on failure (never breaks the build). The file is gitignored.
2. **Store creation**: `makeStore()` calls `getStaticChainsPreloadedState()` which builds a synchronous RTK Query cache entry with `status: 'fulfilled'` and the chain data. This is passed as `preloadedState` to `configureStore()` — purely synchronous, identical on server and client, so no hydration mismatch can occur.
3. **Runtime**: `useInitStaticChains()` triggers a background `forceRefetch` for fresh data. RTK Query's `structuralSharing` preserves object references if the data is identical, preventing unnecessary re-renders.

### Why `preloadedState` instead of `upsertQueryData`?
Previous iterations tried seeding the cache via `upsertQueryData` (either in `makeStore()` or `useEffect`). The `makeStore()` approach caused hydration mismatches because `upsertQueryData` uses microtasks that resolve differently on server vs client. The `useEffect` approach worked but only made chains available client-side (after hydration), missing the SSG/SSR benefit. The `preloadedState` approach is purely synchronous — both SSR and client start with identical fulfilled state.

## Test plan
- [x] `yarn workspace @safe-global/web type-check` passes
- [x] `yarn workspace @safe-global/web lint` passes (0 errors)
- [x] `yarn workspace @safe-global/web test` passes (358 suites, 3186 tests)
- [x] Unit test for `useInitStaticChains` covers dispatch and cleanup
- [x] Works when `chains.json` doesn't exist (CI, failed fetch) — dynamic `require()` with try/catch falls back to empty array
- [x] No hydration errors in dev mode
- [ ] Verify in production build (`yarn build`)

### DIFF
--- a/apps/web/src/components/common/ChainIndicator/__snapshots__/ChainIndicator.stories.test.tsx.snap
+++ b/apps/web/src/components/common/ChainIndicator/__snapshots__/ChainIndicator.stories.test.tsx.snap
@@ -9,9 +9,26 @@ exports[`./ChainIndicator.stories Default 1`] = `
     style="--Paper-shadow: none;"
   >
     <span
-      class="MuiSkeleton-root MuiSkeleton-rectangular MuiSkeleton-pulse mui-style-18ae3nr-MuiSkeleton-root"
-      style="width: 100%; height: 22px;"
-    />
+      class="indicator withLogo"
+      data-testid="chain-logo"
+    >
+      <img
+        alt="Ethereum Logo"
+        height="24"
+        loading="lazy"
+        src="https://safe-transaction-assets.staging.5afe.dev/chains/1/chain_logo.png"
+        width="24"
+      />
+      <div
+        class="MuiStack-root mui-style-nen11g-MuiStack-root"
+      >
+        <span
+          class="name"
+        >
+          Ethereum
+        </span>
+      </div>
+    </span>
   </div>
 </div>
 `;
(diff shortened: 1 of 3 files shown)


katspaugh:
@claude review

apps/web/src/store/index.ts
```diff
@@ -31,6 +32,17 @@ import { cgwClient, setBaseUrl } from '@safe-global/store/gateway/cgwClient'
...
+import { apiSliceWithChainsConfig, chainsAdapter, chainsInitialState } from '@safe-global/store/gateway'
+import type { Chain } from '@safe-global/store/gateway/AUTO_GENERATED/chains'
+// Build-time chain data generated by scripts/fetch-chains.ts.
+// Uses dynamic require because the file may not exist in CI environments
+// that skip the build step (e.g. type-check, unit tests).
+let staticChainsData: Chain[] = []
+try {
+  staticChainsData = require('@/config/__generated__/chains.json') as Chain[]
+} catch {
+  // File doesn't exist — fall back to empty array (runtime fetch will populate)
```
vseehausen, line 44: Empty catch, but ok.

PR #7124, opened 2026-02-07
feat(web): add ENS resolution to accounts list items

alextnetto:
## What it solves

In the spirit of better integrating ENS as a core component for better UX and as a user who feels this pain, here is a PR that adds ENS names for Safe accounts on the welcome/accounts page (`/welcome/accounts`).

Related: https://x.com/austingriffith/status/1961450319411646535

image

## How this PR fixes it

Adds ENS resolution to `SafeListItem` and `MultiAccountItem` using the existing `useAddressResolver` hook — the same pattern already used in `SafeHeaderInfo`.

- **`SafeListItem`**: Resolves ENS for single-chain Safe items when no address book name exists
- **`MultiAccountItem`**: Resolves ENS for multi-chain Safe group headers when no name is set

ENS resolution is skipped when an address book name already exists, keeping the priority chain consistent: address book name > ENS name > shortened address.

> **Future improvement:** A more scalable approach would be to integrate ENS resolution directly into the shared `EthHashInfo` component, so every address display across the app gets ENS names automatically. This was intentionally scoped to a targeted fix to reduce risk.

## How to test it

1. Navigate to `/welcome/accounts`
2. Safe accounts with ENS names should now display their ENS name instead of the shortened address

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/common/NamedAddressInfo/index.tsx
+++ b/apps/web/src/components/common/NamedAddressInfo/index.tsx
@@ -51,7 +51,7 @@ export function useAddressName(address?: string, name?: string | null, customAva
   )
 }
 
-const NamedAddressInfo = ({ address, name, customAvatar, ...props }: EthHashInfoProps) => {
+const NamedAddressInfo = ({ address, name, customAvatar, ...props }: EthHashInfoProps & { showName?: boolean }) => {
   const { name: finalName, logoUri: finalAvatar } = useAddressName(address, name, customAvatar)
 
   return <EthHashInfo address={address} name={finalName} customAvatar={finalAvatar} {...props} />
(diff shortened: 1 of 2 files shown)


alextnetto:
Thanks for the feedback, @katspaugh. Suggestion applied! 🤝

alextnetto:
For some reason it looks like it didn't work. Although v1.83 was already released into production, right?

image

alextnetto:
The search bar also seems like it's not supporting ENS.

image

ChillYao:
Hi @alextnetto ,

Thanks a lot for your PR and the fix. 

It fixes the DNS name bug, but it caused another edge case, which we only find during QA:

The name does not show correctly (show as "proxy") when you change network for a safe account without saved name or ENS. So we have to revert this commit during release.

We are now developing our new UI and will address this issue in our new UI.

Still thanks a lot for your effort!!

katspaugh:
That's my bad – I asked to use NamedAddressInfo instead of resolving ENS via a hook but NamedAddressInfo isn't just resplvoing ENS but also Etherscan contract names (via our backend).

We'll restore @alextnetto's original proposal (with a hook).

PR #7154, opened 2026-02-12
refactor: optimize useLoadFeature with shared registry and reduced re-renders

katspaugh:
## Summary

Optimizes `useLoadFeature` by adding a shared feature registry, eliminating the `$isLoading` meta property, and stabilizing object references to reduce re-renders.

### Shared feature registry
A module-level `Map` caches loaded features by name. When component A loads a feature and component B mounts later, B gets the feature synchronously via the `useState` initializer (1 render instead of 2). Also deduplicates concurrent `handle.load()` calls via a pending-promise map. Errors are not cached globally so retry is possible on remount.

### Eliminated `$isLoading` state
Replaced `useAsync` with a simpler `useEffect` + `useState` that only triggers one `setState` call (when the feature loads). Features transition directly from not-ready to ready — no intermediate loading render.

### Stable stub proxy
The Proxy is created once per hook instance and reads meta values (`$isDisabled`, `$isReady`, `$error`) from a ref. The object reference stays the same across all non-ready states, so children receiving the feature object don't re-render on meta changes alone.

### Re-render reduction

| Scenario | Before | After |
|---|---|---|
| Initial mount → ready | 3–4 renders | 2 |
| Nth component, same feature already loaded | 2 renders | **1 render** |
| Concurrent mounts (3 components) | 6 renders, 3 load() calls | **4 renders, 1 load()** |
| Chain switch (flag flickers) | 4–5 renders | **0 extra** (same ref bail-out) |
| Proxy meta changes while loading | New object ref each time | Same ref, fresh values |

### Breaking change: `$isLoading` removed

`$isLoading` has been removed from `FeatureMeta`. Consumers should use `!$isReady` instead:

```diff
- if (feature.$isLoading) return <Skeleton />
+ if (!feature.$isReady) return <Skeleton />
```

Updated across the codebase: `SafeModules` component, test mocks, JSDoc examples, feature-architecture docs, AGENTS.md, codemod templates, and spec files.

## Test plan

- [x] `type-check` passes
- [x] `lint` passes (0 errors, only pre-existing warnings)
- [x] `prettier` passes
- [x] All 19 `useLoadFeature` tests pass (15 existing + 4 new shared registry tests)
- [x] All 46 `@safe-global/utils` test suites pass (645 tests, with coverage)
- [x] Full web test suite: 358 suites, 3205 tests pass
- [ ] Manual smoke: switch chains, verify features load/unload without flash
- [ ] Manual smoke: verify `$isReady`/`$isDisabled` transitions render correctly

### DIFF
--- a/packages/utils/src/hooks/useAsync.ts
+++ b/packages/utils/src/hooks/useAsync.ts
@@ -1,66 +1,45 @@
-import { useCallback, useEffect, useReducer } from 'react'
+import { useCallback, useEffect, useState } from 'react'
 import { asError } from '@safe-global/utils/services/exceptions/utils'
 
 export type AsyncResult<T> = [result: T | undefined, error: Error | undefined, loading: boolean]
-
-interface AsyncState<T> {
-  data: T | undefined
-  error: Error | undefined
-  loading: boolean
-}
-
-type AsyncAction<T> =
-  | { type: 'reset' }
-  | { type: 'start'; clearData: boolean }
-  | { type: 'success'; data: T }
-  | { type: 'error'; error: Error }
-  | { type: 'loading_done' }
-
-function asyncReducer<T>(state: AsyncState<T>, action: AsyncAction<T>): AsyncState<T> {
-  switch (action.type) {
-    case 'reset':
-      return { data: undefined, error: undefined, loading: false }
-    case 'start':
-      return { data: action.clearData ? undefined : state.data, error: undefined, loading: true }
-    case 'success':
-      return { data: action.data, error: undefined, loading: false }
-    case 'error':
-      return { data: undefined, error: action.error, loading: false }
-    case 'loading_done':
-      return state.loading ? { ...state, loading: false } : state
-  }
-}
-
-const INITIAL_STATE: AsyncState<unknown> = { data: undefined, error: undefined, loading: false }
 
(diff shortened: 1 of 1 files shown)


katspaugh:
@claude review

katspaugh:
@clovisdasilvaneto because dynamic only wraps individual component imports, and we want to lazy-load an entire feature (multiple components + utils/services) in a single chunk.

clovisdasilvaneto:
@katspaugh this topic of claude:
>2. Reducer Optimization Inconsistency (useLoadFeature.ts:44,47)

> // Inconsistent optimization checks
> if (state.loading && state.data === undefined && state.error === undefined) return state
> if (!state.loading && state.data === undefined && state.error === undefined) return state
> These early returns prevent unnecessary re-renders but add complexity. Consider extracting to a helper:

> const isStateUnchanged = (state, expectedLoading, expectedData = undefined, expectedError = undefined) => 
>  state.loading === expectedLoading && state.data === expectedData && state.error === expectedError

actually makes sense

clovisdasilvaneto:
>@clovisdasilvaneto because dynamic only wraps individual component imports, and we want to lazy-load an entire feature (multiple components + utils/services) in a single chunk.

Got it 👍  I think you have told me that once, my bad

PR #7155, opened 2026-02-12
chore: add prettier guardrails to prevent CI formatting failures

katspaugh:
- Add prettier check to pre-push hook so formatting issues are caught before code reaches CI
- Add Claude Code PreToolUse hook that auto-fixes prettier on staged files before any git commit
- Update CLAUDE.md and AGENTS.md to explicitly require running prettier:fix before committing (not just relying on lint-staged)


.husky/pre-push
```diff
@@ -18,6 +18,16 @@ if [ "$initial_status" != "$post_lint_status" ]; then
   exit 1
 fi
 
+# Run prettier check — CI runs this too, so catch issues before pushing
+echo "Checking formatting..."
+if ! yarn workspace @safe-global/web prettier 2>/dev/null; then
```
compojoom: but why are you running the web prettier? What if the files are not for the web project, but for some of the packages?

PR #7182, opened 2026-02-16
refactor: unify balance loading into shared hook and always use v2 overview endpoint

compojoom:
## What it solves

Balance-loading logic (portfolio endpoint, tx service fallback, counterfactual handling) was duplicated between web and mobile. Additionally, both apps implemented redundant client-side v1/v2 endpoint selection for SafeOverview, even though the CGW v2 endpoint already handles the Zerion-or-fallback decision server-side.

Resolves: https://linear.app/safe-global/issue/WA-1251/positions-on-mobile

## How this PR fixes it

**Shared `useTotalBalances` hook** — Extracted the balance-loading orchestration into `packages/utils/src/hooks/useTotalBalances.ts` so both web and mobile can reuse it. Web's `useLoadBalances` becomes a thin wrapper, and mobile gets a new `useMobileTotalBalances` adapter.

**Dropped client-side v1/v2 routing** — The CGW `/v2/safes` endpoint already falls back to tx-service data when Zerion isn't available for a chain, so the `PORTFOLIO_ENDPOINT` feature flag check on the client was pure overhead. Now `packages/store/src/gateway/safes/index.ts` always hits `/v2/safes`, and all the v1/v2 branching in both web (`SafeOverviewFetcher`) and mobile (`useSafeOverviewsQuery`) is removed.

**Lightweight balance display on mobile** — `BalanceContainer` now uses `useSafeOverviewsQuery` (returns just `fiatTotal` per chain) instead of firing the full portfolio + tx service queries, which is all it needs.

**Positions polling fix** — The positions hook now makes its own portfolio query with a 5-minute polling interval, independent of the balance hook's 15s interval. RTK Query handles interval aggregation when both are mounted.

## How to test it

1. Open the web app dashboard — verify total balance displays correctly, including DeFi positions
2. Open mobile — verify balance display on the accounts screen and asset breakdown
3. Switch between chains (including ones without Zerion support like testnets) — verify balances still load correctly via the server-side fallback
4. Check the network tab — all overview requests should go to `/v2/safes` (no `/v1/safes` calls)

## Screenshots

N/A - no UI changes

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/components/SafeAccountInput/hooks/useImportSafe.ts
+++ b/apps/mobile/src/components/SafeAccountInput/hooks/useImportSafe.ts
@@ -41,7 +41,6 @@ export const useImportSafe = () => {
             safes: chainIds.map((chainId: string) => makeSafeId(chainId, address)),
             currency,
             trusted: true,
-            excludeSpam: true,
           },
           false,
         )
--- a/apps/mobile/src/features/DataImport/helpers/transforms.test.ts
+++ b/apps/mobile/src/features/DataImport/helpers/transforms.test.ts
@@ -242,7 +242,6 @@ describe('Data import helpers', () => {
         safes: ['1:0x1', '137:0x2'],
         currency: 'USD',
         trusted: true,
-        excludeSpam: true,
       })
 
       expect(result).toEqual(new Set(['0x2', '0x3']))
--- a/apps/mobile/src/features/DataImport/helpers/transforms.ts
+++ b/apps/mobile/src/features/DataImport/helpers/transforms.ts
@@ -208,7 +208,6 @@ export const fetchAndStoreSafeOverviews = async (
           safes: chunk,
           currency,
           trusted: true,
-          excludeSpam: true,
         }),
       ).unwrap()
 
--- a/apps/web/src/store/__tests__/safeOverviews.test.ts
+++ b/apps/web/src/store/__tests__/safeOverviews.test.ts
@@ -139,7 +139,6 @@ describe('safeOverviews', () => {
         safes: [`1:${fakeSafeAddress}`],
         currency: 'usd',
         trusted: false,
-        excludeSpam: true,
         walletAddress: undefined,
       })
     })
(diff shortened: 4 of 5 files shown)


apps/mobile/src/features/AccountsSheet/MyAccounts/MyAccounts.container.tsx
```diff
@@ -38,12 +38,18 @@ export function MyAccountsContainer({ item, isDragging, drag, onClose }: MyAccou
     onClose()
   }
 
+  const fiatTotal = chainsIds
+    .reduce((acc, id) => {
+      return acc + parseFloat(item.info[id].fiatTotal)
+    }, 0)
+    .toString()
```
clovisdasilvaneto: can we move this to a utility function so we do not have to repeat? Also might be good to add this reduce under a useMemo?

packages/utils/src/hooks/useTotalBalances.ts
```diff
@@ -0,0 +1,255 @@
...
+    loading: portfolioLoading || isInitialLoading,
+    ...shared,
+  }
+}
+
+/**
+ * Builds a merged result combining portfolio positions with tx-service token list ("All Tokens" mode).
+ */
+const buildMergedResult = (
+  portfolioBalances: PortfolioBalances | undefined,
```
clovisdasilvaneto: just a small suggestion here, might be better to move this argument to be the last of the function, because having a possible undefined value as the first argument of the function seems weird. but again, this is just code style and mostly preference from my POV, so feel free to do what you like most

packages/utils/src/hooks/useTotalBalances.ts
```diff
@@ -0,0 +1,255 @@
...
+  return useMemo<TotalBalancesResult>(() => {
+    if (params.skip) {
+      return { data: undefined, error: undefined, loading: false, isFetching: false, refetch }
+    }
+
+    const shared: SharedResultFields = { isFetching, refetch }
+    const txService: TxServiceState = { balances: txServiceBalances, error: txServiceError, loading: txServiceLoading }
+    const counterfactual: CounterfactualState = { data: cfData, error: cfError, loading: cfLoading }
+
+    // No portfolio feature or portfolio needs fallback → tx service only
```
clovisdasilvaneto: Wouldn't be better to move the implementation of this useMemo to a separated aggregator following a strategy pattern? I think in this way you would be able to decrease the complexity of the implementation of the hook.

PR #7183, opened 2026-02-16
fix(mobile): use Reanimated for smooth SafeTab transitions

compojoom:
## What it solves

Tab switching in SafeTab felt janky because it relied on React state (`useState`) to track the active tab, causing re-renders on every tab change.

Resolves: https://linear.app/safe-global/issue/WA-1251/positions-on-mobile#comment-ca2fa62c

## How this PR fixes it

Replaces the React state-driven active tab tracking with Reanimated's `indexDecimal` shared value that the collapsible tab library already provides. Tab indicator color, text color, and right node visibility now animate on the UI thread via `useAnimatedStyle`, avoiding JS thread re-renders entirely. Also switches from `TouchableOpacity` (gesture handler) to `Pressable` (React Native) for simpler press handling.

## How to test it

1. Open the mobile app and navigate to a screen with SafeTab (e.g., positions)
2. Swipe between tabs — the tab indicator and text color should animate smoothly
3. Tap on tab labels — should switch instantly without flicker
4. Verify the right node (if present) fades in/out correctly per tab

## Screenshots
before

https://github.com/user-attachments/assets/307b639f-46d8-4fc4-9df6-257dc7959775

after

https://github.com/user-attachments/assets/05ffd839-b11a-4b61-8aae-e34d6483ec55

N/A - mobile UI, same visual result but smoother animation

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/components/SafeTab/SafeTabBar.tsx
+++ b/apps/mobile/src/components/SafeTab/SafeTabBar.tsx
@@ -1,38 +1,78 @@
 import React from 'react'
 import { TabBarProps } from 'react-native-collapsible-tab-view'
 import { TabName } from 'react-native-collapsible-tab-view/lib/typescript/src/types'
-import { TouchableOpacity } from 'react-native-gesture-handler'
-import { View, Text, useTheme } from 'tamagui'
+import { Pressable } from 'react-native'
+import { View, useTheme } from 'tamagui'
+import Animated, { SharedValue, useAnimatedStyle } from 'react-native-reanimated'
 
 interface SafeTabBarProps {
-  setActiveTab: (name: string) => void
-  activeTab: string
-  rightNode?: React.ReactNode
+  rightNode?: (tabName: string) => React.ReactNode
+}
+
+const TabItem = ({
+  name,
+  index,
+  indexDecimal,
+  onPress,
+  activeColor,
+  inactiveColor,
+  activeBorderColor,
+}: {
+  name: string
+  index: number
+  indexDecimal: SharedValue<number>
+  onPress: (name: string) => void
+  activeColor: string
+  inactiveColor: string
+  activeBorderColor: string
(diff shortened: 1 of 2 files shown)


apps/mobile/src/components/SafeTab/SafeTabBar.tsx
```diff
@@ -1,38 +1,78 @@
...
+  indexDecimal,
   onTabPress,
-  activeTab,
-  setActiveTab,
   rightNode,
 }: TabBarProps<TabName> & SafeTabBarProps) => {
   const theme = useTheme()
+  const activeColor = theme.color?.get() ?? '#000'
+  const inactiveColor = theme.colorSecondary?.get() ?? '#999'
+  const activeBorderColor = theme.primary?.get() ?? '#000'
```
clovisdasilvaneto, line 71: so sad that we broke our lint...

apps/mobile/src/components/SafeTab/SafeTabBar.tsx
```diff
@@ -1,38 +1,78 @@
...
-  setActiveTab: (name: string) => void
-  activeTab: string
-  rightNode?: React.ReactNode
+  rightNode?: (tabName: string) => React.ReactNode
+}
+
+const TabItem = ({
+  name,
+  index,
+  indexDecimal,
```
clovisdasilvaneto, line 15: where is this indexDecimal coming from? is it comming from the spread {...props} that we have on the `SafeTab` component? If so, is it native from `eact-native-collapsible-tab-view`?

compojoom: `react-native-collapsible-tab-view` is passing it. And since it is a animatedValue - it doesn't cause a rerender and operates on the UI thread, that makes animations instant.

PR #7194, opened 2026-02-17
fix: use useIsSafeOwner for read-only status in CurrentSafe

compojoom:
## What it solves

When a Safe (smart contract wallet) is connected as an owner of the currently viewed Safe, the sidebar incorrectly shows the current Safe as "Read-only".

Resolves: https://linear.app/safe-global/issue/WA-1472/read-only-tag-in-safes-the-current-connected-safe-as-owner-owns

## How this PR fixes it

The `CurrentSafe` component determined read-only status using `!safeInList` — checking whether the current safe existed in the `allSafes` list. This list is built from `useAllOwnedSafes`, which queries the CGW API with the connected wallet address. When the connected wallet is a Safe (e.g., Safe A owns Safe B), `useAllOwnedSafes` returns safes owned by Safe A, but not safes owned *by* Safe A's children. So the current safe (Safe B) wasn't found in the list, causing the read-only fallback.

The fix replaces `!safeInList` with `!useIsSafeOwner()`, which checks the current safe's actual owner list against the **signer** address. This correctly handles nested Safe ownership because `useSigner()` resolves to the nested Safe wallet.

## How to test it

1. Connect a Safe via WalletConnect (Safe A that owns Safe B)
2. Navigate to Safe B
3. Verify the sidebar shows Safe B **without** the "Read-only" chip
4. Connect an EOA owner directly and verify the same behavior

## Screenshots

N/A - no visual changes beyond fixing the incorrect read-only chip

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/hooks/safes/useAllSafes.ts
+++ b/apps/web/src/hooks/safes/useAllSafes.ts
@@ -4,7 +4,7 @@ import { useAppSelector } from '@/store'
 import type { AddedSafesState } from '@/store/addedSafesSlice'
 import { selectAllAddedSafes } from '@/store/addedSafesSlice'
 import useChains from '@/hooks/useChains'
-import useWallet from '@/hooks/wallets/useWallet'
+import useWallet, { useSigner } from '@/hooks/wallets/useWallet'
 import type { AddressBookState, VisitedSafesState } from '@/store/slices'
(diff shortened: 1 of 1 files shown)


compojoom:
We have a different issue here.

PR #7196, opened 2026-02-17
chore: always upload Chromatic screenshots even on test failure

vseehausen:
## Summary
- Add `continue-on-error: true` to the e2e job so test failures don't fail the workflow
- Change chromatic job condition to `!cancelled()` so it uploads regardless of test results
- Rename job to "Cypress Visual and Smoke tests"
- Workflow success is now determined by the Chromatic upload, not test results

## Test plan
- [ ] Trigger `Web Chromatic E2E` workflow manually
- [ ] Verify Chromatic upload runs even when some tests fail
- [ ] Verify workflow shows as successful when Chromatic upload succeeds

### DIFF
--- a/.github/workflows/web-chromatic-e2e.yml
+++ b/.github/workflows/web-chromatic-e2e.yml
@@ -11,7 +11,8 @@ jobs:
   e2e:
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    name: Cypress Smoke tests
+    continue-on-error: true
+    name: Cypress Visual and Smoke tests
     permissions:
       contents: read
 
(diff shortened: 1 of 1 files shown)


.github/workflows/web-chromatic-e2e.yml
```diff
@@ -36,7 +37,7 @@ jobs:
 
   chromatic:
     needs: [e2e]
-    if: always() && needs.e2e.result == 'success'
+    if: "!cancelled()"
```
vseehausen, line 40: Ensures chromatic always runs, even when a test fails. The goal of this workflow it to upload screenshots.

liliya-soroka:
@vseehausen, is there any way to know the real state of the run?
‘Verify’ workflow shows as successful when the Chromatic upload succeeds.
I’m afraid it will be ignored if it’s marked as successful while tests are actually failing.

PR #7222, opened 2026-02-19
fix(web): restrict postMessage target origin in AppCommunicator

katspaugh:
Resolves https://linear.app/safe-global/issue/WA-1494/wildcard-postmessage-origin-in-safe-wallet-web-appcommunicator

## Summary
- Replace wildcard `'*'` origin in `postMessage` with the Safe App's actual origin to prevent SDK response data from leaking if the iframe navigates to a malicious page (e.g. via open redirect)
- Add `msg.origin` validation on incoming messages to reject requests from unexpected origins
- Add unit tests covering origin enforcement for both incoming and outgoing messages

## Security Details
`AppCommunicator.send()` previously used `postMessage(msg, '*')`, meaning SDK responses (safe info, balances, chain info, address book, RPC results) would be delivered to whatever origin currently occupies the iframe. Combined with an open redirect in any Safe App, this could allow an attacker page to receive wallet data.

**Fix**: Accept an `allowedOrigin` parameter derived from the app URL, use it as the `targetOrigin` for outgoing `postMessage`, and validate `msg.origin` on incoming messages.

## Test plan
- [x] Unit tests for origin-restricted `postMessage` (10 tests, all passing)
- [x] Manual test: load a Safe App and verify SDK communication still works
- [ ] Verify Safe Apps that use query parameters in their URL still function correctly

### DIFF
--- a/apps/web/src/features/swap/index.tsx
+++ b/apps/web/src/features/swap/index.tsx
@@ -86,6 +86,7 @@ const SwapWidget = ({ sell }: Params) => {
   const nativeCowSwapFeeV2Enabled = useHasFeature(FEATURES.NATIVE_COW_SWAP_FEE_V2)
   const isEurcvBoostEnabled = useHasFeature(FEATURES.EURCV_BOOST)
   const useStagingCowServer = useHasFeature(FEATURES.NATIVE_SWAPS_USE_COW_STAGING_SERVER)
+  const cowSwapBaseUrl = useStagingCowServer ? 'https://staging.swap.cow.fi' : 'https://swap.cow.fi'
 
   const { data: isSafeAddressBlocked } = useGetIsSanctionedQuery(safeAddress || skipToken)
   const { data: isWalletAddressBlocked } = useGetIsSanctionedQuery(wallet?.address || skipToken)
@@ -103,7 +104,7 @@ const SwapWidget = ({ sell }: Params) => {
     width: '100%', // Width in pixels (or 100% to use all available space)
     height: '860px',
     chainId,
-    baseUrl: useStagingCowServer ? 'https://staging.swap.cow.fi' : 'https://swap.cow.fi',
+    baseUrl: cowSwapBaseUrl,
     standaloneMode: false,
     disableToastMessages: true,
     disablePostedOrderConfirmationModal: true,
@@ -155,7 +156,7 @@ const SwapWidget = ({ sell }: Params) => {
   const appData: SafeAppData = useMemo(
     () => ({
       id: 1,
-      url: 'https://app.safe.global',
+      url: cowSwapBaseUrl,
       name: SWAP_TITLE,
       iconUrl: darkMode ? './images/common/safe-swap-dark.svg' : './images/common/safe-swap.svg',
       description: 'Safe Apps',
@@ -166,7 +167,7 @@ const SwapWidget = ({ sell }: Params) => {
       socialProfiles: [],
       featured: false,
     }),
-    [darkMode],
+    [darkMode, cowSwapBaseUrl],
   )
 
   const listeners = useMemo<CowEventListeners>(() => {


katspaugh:
@claude review

katspaugh:
## Addressing review comments (972acea)

**1. `isCookieEnabled` validation bypass** — Fixed. The `isCookieEnabled` early-return has been completely removed from `isValidMessage()`. All messages now go through origin + source validation. Cookie-check messages are handled separately by `useThirdPartyCookies`'s own `messageHandler`, which now also validates `event.origin`.

**2. URL parsing error handling** — Fixed. Added early return if `app?.url` is missing, and a try/catch around `new URL()` to handle invalid URLs gracefully.

**3. Missing `isCookieEnabled` bypass test** — Fixed. Added two tests: one sends a message with `isCookieEnabled` + real `method`/`id` from a disallowed origin (rejected), and another with correct origin but wrong source window (also rejected).

### Additional security fixes in this commit:
- **Wildcard postMessage in `useThirdPartyCookies`**: Replaced `'*'` with the actual cookie-check origin
- **Origin validation in `useThirdPartyCookies`**: Added `event.origin` check in the `messageHandler`
- **Error message sanitization in `AppCommunicator`**: Raw `error.message` replaced with generic `'Request failed'` to prevent leaking internal details to Safe App iframes
- Added `useThirdPartyCookies.test.ts` with 3 tests covering origin validation

apps/web/src/components/safe-apps/AppFrame/useAppCommunicator.ts
```diff
@@ -91,7 +91,16 @@ const useAppCommunicator = (
     let communicatorInstance: AppCommunicator
 
     const initCommunicator = (iframeRef: MutableRefObject<HTMLIFrameElement | null>, app?: SafeAppData) => {
-      communicatorInstance = new AppCommunicator(iframeRef, {
+      if (!app?.url) return
```
tmjssz: Can we just make the `app` param mandatory for the function and then get rid of the nullish coalescing?

katspaugh: Done. Made app mandatory in initCommunicator, removed the if (!app?.url) guard and the unnecessary optional chaining on app.url/app.name/app.id.

katspaugh:
## Testing update

Tested CowSwap both as a Safe App (`/apps/open?appUrl=...`) and via the native Swap feature (`/swap`). Both were broken due to origin mismatches:

1. **Safe App (CowSwap)**: The gateway had `cowswap.exchange` as the app URL, but that domain 301-redirects to `swap.cow.fi`. Fixed the URL in the gateway config.
2. **Swap widget**: The `appData.url` was hardcoded to `https://app.safe.global` while the CowSwap iframe loads from `https://swap.cow.fi`. Fixed in 677b32d.

Safe Apps work correctly now with the origin restriction. One caveat: if a Safe App does a redirect (registered URL differs from the actual serving domain), it will break communication. The fix is to update the app URL in the gateway data to point to the final domain.

katspaugh:
## Testing update #2

**Gateway data audit**: Checked all Safe Apps in prod and updated URLs for apps that were doing cross-domain redirects.

**E2E results** (`cypress/e2e/safe-apps/`): Ran the full safe-apps e2e suite locally — 32/50 passing, 10 pending, 2 skipped. The 3 failing specs (`apps_list`, `permissions_settings`, `preview_drawer`) are pre-existing failures unrelated to this PR (confirmed by running against the baseline). All tx-builder and permissions tests that exercise iframe communication pass.

**Swap widget fix** (677b32d): The swap widget had `appData.url` hardcoded to `https://app.safe.global` while the CowSwap iframe loads from `https://swap.cow.fi`. Updated to use the actual CowSwap base URL.

Note: same-host redirects (e.g. `/` → `/widget/swap`) are fine since `postMessage` origin only includes scheme + host + port. Only cross-domain redirects break communication, and those have now been fixed in the gateway data.

PR #7246, opened 2026-02-23
Redesign accounts component

clovisdasilvaneto:
## What it solves
It redesign the accounts component according to the new spaces figma.

## Screenshots
Screenshot 2026-02-23 at 14 46 32


apps/web/src/features/spaces/components/Dashboard/index.tsx
```diff
@@ -4,9 +4,7 @@ import AddressBookCard from '@/features/spaces/components/Dashboard/ImportAddres
 import { Card, Grid2, Stack, Typography } from '@mui/material'
 import Grid from '@mui/material/Grid2'
 import { useSpaceSafes } from '@/features/spaces/hooks/useSpaceSafes'
-import { useLoadFeature } from '@/features/__core__'
```
vmosharova, line 6: We need to keep `useLoadFeature()` approach as it was the idea behind the Feature Architecture to reduce the JS bundle.

apps/web/src/features/spaces/components/Dashboard/index.tsx
```diff
@@ -68,20 +79,20 @@ const SpaceDashboard = () => {
           </Grid>
 
           <Grid container spacing={3}>
-            <Grid size={{ xs: 12, md: 8 }}>
-              <Card data-testid="dashboard-safe-list" sx={{ p: 2 }}>
```
vmosharova, line 75: QA asked to keep all existing `data-testid` properties to keep the existing test suite

apps/web/src/features/spaces/components/Dashboard/index.tsx
clovisdasilvaneto: we're not using MyAccounts feature here, that is why it is not using it. will try to use the useLoadFeature for the AccountsWidget

clovisdasilvaneto: done ✔️

vmosharova:
@clovisdasilvaneto the dark mode should work when merged to the epic with the latest changes, let's fix the unit test only.

PR #7266, opened 2026-02-25
fix(mobile): use app theme for native context menu instead of device theme

compojoom:
## What it solves

When the user's app theme is set to light but the device is in dark mode, the native context menu (FloatingMenu) renders with a dark background (from device theme) while the icon colors are derived from the app's light theme — resulting in dark icons on a dark background.

Resolves: https://linear.app/safe-global/issue/WA-1384/adhoc-fixes-for-upcoming-release

## How this PR fixes it

Moved `themeVariant` resolution into `FloatingMenu` itself via `useTheme()` hook, so the native `MenuView` always matches the app's theme preference. This fixes the bug for all 3 callers (SettingsMenu, AppSettings, LedgerAddresses) at once and prevents future callers from forgetting to pass it. Also hardcodes the iOS destructive action color to match the native red.

## How to test it

1. Set the app theme to **light** in App Settings
2. Set device to **dark mode**
3. Go to Settings and tap the "..." menu
4. Verify the menu background is light and icons are visible (not dark-on-dark)
5. Repeat with app theme set to **dark** and device in **light mode** — menu should be dark

## Screenshots

N/A - requires device testing

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/Settings/components/Navbar/SettingsMenu.tsx
+++ b/apps/mobile/src/features/Settings/components/Navbar/SettingsMenu.tsx
@@ -28,7 +28,11 @@ export const SettingsMenu = ({ safeAddress }: Props) => {
   const copyAndDispatchToast = useCopyAndDispatchToast()
   const theme = useTheme()
   const color = theme.color.get()
-  const colorError = theme.error.get() || '#FF5F72'
+  // hardcoded to the iOS red
+  // when we set danger to a button, it automatically sets a color that the OS selects
+  // titleColor only works on android and not on iOS
+  // that's why I'm hardcoding the iOS value of the danger text here
+  const colorError = 'rgb(255,66,69)'
 
   if (!safeAddress) {
     return null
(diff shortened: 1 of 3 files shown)


apps/mobile/src/features/Settings/components/Navbar/SettingsMenu.tsx
```diff
@@ -28,7 +28,11 @@ export const SettingsMenu = ({ safeAddress }: Props) => {
   const copyAndDispatchToast = useCopyAndDispatchToast()
   const theme = useTheme()
   const color = theme.color.get()
-  const colorError = theme.error.get() || '#FF5F72'
+  // hardcoded to the iOS red
+  // when we set danger to a button, it automatically sets a color that the OS selects
+  // titleColor only works on android and not on iOS
+  // that's why I'm hardcoding the iOS value of the danger text here
+  const colorError = 'rgb(255,66,69)'
```
PooyaRaki, line 35: Nit: What do you think about moving this into a separate helper (or constants) file so we can reuse it in other places as well?

compojoom: We don't need to reuse it. This is the only place where we currently use the "destrictive" behavior of that Menu library.

PR #7280, opened 2026-02-27
fix: infinite loading of positions

0xAryanShah:
## What it solves

Resolves: [Linear Issue](https://linear.app/safe-global/issue/WA-1536/infinite-loading-of-positions-for-new-safes)

When opening a Safe with no funds, the Positions tab was stuck in a perpetual loading (skeleton) state instead of showing the empty state.

This was caused by positions being set to undefined, which prevented the UI from resolving the loading condition correctly.

Fix:

- Initialize positions as an empty array ([]) instead of undefined in createPortfolioBalances.
- Update related tests accordingly.

This ensures the empty state renders properly for Safes with no positions.

### DIFF
--- a/packages/utils/src/hooks/__tests__/portfolioBalances.test.ts
+++ b/packages/utils/src/hooks/__tests__/portfolioBalances.test.ts
@@ -41,7 +41,7 @@ describe('portfolioBalances helpers', () => {
       expect(result.fiatTotal).toBe('1000')
       expect(result.tokensFiatTotal).toBe('1000')
       expect(result.positionsFiatTotal).toBe('0')
-      expect(result.positions).toBeUndefined()
+      expect(result.positions).toEqual([])
       expect(result.items).toEqual(balances.items)
     })
   })
--- a/packages/utils/src/hooks/__tests__/useTotalBalances.test.ts
+++ b/packages/utils/src/hooks/__tests__/useTotalBalances.test.ts
@@ -217,11 +217,12 @@ describe('useTotalBalances', () => {
     })
 
     it('should return counterfactual balances for undeployed safe with portfolio feature', () => {
+      const mockEmptyPortfolio = createMockEmptyPortfolio()
       const mockCfBalances: Balances = { fiatTotal: '500', items: [] }
 
       jest
         .spyOn(portfolioQueries, 'usePortfolioGetPortfolioV1Query')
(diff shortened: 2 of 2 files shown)


0xAryanShah:
@claude review

0xAryanShah:
@claude review

packages/utils/src/hooks/portfolioBalances.ts
```diff
@@ -17,7 +17,7 @@ export const createPortfolioBalances = (balances: Balances): PortfolioBalances =
   ...balances,
   tokensFiatTotal: balances.fiatTotal,
   positionsFiatTotal: '0',
-  positions: undefined,
+  positions: [],
```
katspaugh, line 20: Why change it here and not in the UI code that handles loading state?

0xAryanShah: The root cause is in the data layer, not the UI layer:

1) isPortfolioEmpty in useTotalBalances checks both tokenBalances.length === 0 and positionBalances.length === 0 so we already know positions are empty ([])

2) The fallback calls createPortfolioBalances() which has no concept of positions, so it set positions: undefined, discarding the knowledge we just verified

3) usePositions reads balances?.positions → passes it to transformAppBalancesToProtocols(undefined) → returns undefined

4) The component checks !protocols → true for undefined → renders skeleton forever

katspaugh: I see the chain you described, but I think the real issue is in the UI, not the data layer:
```                                                                                                                 
  if (isLoading || (!error && !protocols)) {
    return <PositionsSkeleton />
  }
```
  This conflates "data hasn't arrived yet" with "data arrived but positions aren't applicable." The isLoading
  flag should be the sole driver of the skeleton — if isLoading is false and protocols is undefined, that's not a loading state, it's an empty/absent state that should be handled accordingly.

  Your fix works, but it means `positions?: AppBalance[]` is now effectively never undefined despite the type saying it can be, and the fragile `(!error && !protocols)` check remains — any future code path that returns undefined will hit the same infinite skeleton.

  I'd prefer we fix the UI to handle undefined as "no positions" (empty state) and rely only on isLoading for the skeleton.

vseehausen: isLoading should only be the driver for the skeleton, when the data layer ensures that isLoading is true even on an initial page load.

PR #7289, opened 2026-02-28
Chore/docker self hosted runtime env

gjeanmart:
# chore: add Docker self-hosted deployment support with runtime env injection

## Problem

The existing Docker image for Safe{Wallet} web was underperforming for self-hosted deployments. The static Next.js export needed to be served differently per environment, but `NEXT_PUBLIC_*` variables were baked into the bundle at build time, making a single generic image impossible.

Self-hosters faced two compounding pain points:
- **Image size**: building the app produces a ~1.7 GB image (node:18-alpine + Python, make, g++, libusb and all other build-time dependencies) that has to be transferred on every deploy.
- **Rebuild on every config change**: changing a gateway URL, a chain ID, or any other `NEXT_PUBLIC_*` variable required a full image rebuild (~5-10 min), making iterating on a self-hosted setup very slow.

Together this made Docker-based self-hosting impractical for operators who need to point the UI at a custom gateway or local chain.

## Solution

Two-stage Docker build (Node builder -> nginx:alpine) with runtime environment variable injection at container startup.

### Image size improvement

| | Before | After |
|---|---|---|
| Image size | ~1.7 GB (node:18-alpine + build deps) | ~100 MB (nginx:alpine + static files) |
| Reduction | | ~94% smaller |

The two-stage build discards the Node.js toolchain, Python, make, g++, and all build-time dependencies. Only the pre-built static files and a minimal nginx binary are shipped.

### Deploy time improvement

| Scenario | Before | After |
|---|---|---|
| Change a gateway URL | Full rebuild (~5-10 min) | Restart container (~5 sec) |
| Switch chain ID | Full rebuild | Restart container |
| Deploy to a new environment | Full rebuild per env | Same image, different env vars |
| Pull image on deploy | ~1.7 GB to pull | ~100 MB to pull |

Because all `NEXT_PUBLIC_*` values are injected at startup from the container environment, a single image can be used across local, staging, and production by passing different env vars — no rebuild ever needed for configuration changes.

## How runtime injection works

Next.js static export (`output: 'export'`) bundles a webpack `process` browser polyfill (npm `process` package). This module is compiled into the main webpack chunk and sets `process.env = {}` (empty object) at runtime. All `process.env.NEXT_PUBLIC_*` accesses in the app read from this object.

The `entrypoint.sh` exploits this: at container startup it finds the pattern `.env={}` in the `main-*.js` webpack chunk and replaces it with a JSON object built from all `NEXT_PUBLIC_*` variables set in the container environment:

```
 .env {"NEXT_PUBLIC_IS_PRODUCTION":"true","NEXT_PUBLIC_GATEWAY_URL_PRODUCTION":"http://localhost:8000/cgw",...}
```

This means the Docker image is built once (gateway-agnostic) and configured entirely through environment variables at runtime.

### apps/web/Dockerfile (modified)

- Two-stage build: Node 18 Alpine builder + nginx:alpine runtime
- Removes local `.env` files before `yarn build` to prevent dev values from being baked into the bundle
- `SKIP_FETCH_CHAINS=true` build arg (default) skips CGW chain pre-fetch so the image is gateway-agnostic; chains are fetched at runtime from the injected `NEXT_PUBLIC_GATEWAY_URL_*`
- `BUILD_GATEWAY_URL` build arg for pre-fetching chains when needed
- Exposes port 8080

### apps/web/scripts/docker/entrypoint.sh (new)

- Collects all `NEXT_PUBLIC_*` container env vars
- Builds a JSON object and patches `.env={}` in the main webpack chunk via `sed`
- Falls back gracefully when no `NEXT_PUBLIC_*` vars are present
- Starts nginx after injection

### apps/web/scripts/docker/nginx.conf (new)

- Serves the static export on port 8080
- `_next/static/` uses `Cache-Control: no-cache` (not `immutable`) so browsers always revalidate after a container update
- SPA fallback: `try_files $uri $uri.html $uri/ /index.html`
- `/health` endpoint for container health checks

### apps/web/scripts/fetch-chains.ts (modified)

- Respects `SKIP_FETCH_CHAINS=true`: writes an empty `chains.json` and exits early, skipping the CGW pre-fetch at build time

### apps/web/src/config/constants.ts (modified)

- Exports `IS_PRODUCTION_STR` as a named export before deriving the boolean `IS_PRODUCTION`. This prevents Terser from constant-folding `process.env.NEXT_PUBLIC_IS_PRODUCTION === 'true'` to `false` at build time.
- Exports `DEFAULT_MAINNET_CHAIN_ID_STR` and `DEFAULT_TESTNET_CHAIN_ID_STR` as string intermediaries before `parseInt`, ensuring the runtime-injected values are read before conversion.

### Quick start with docker-compose

Add to your `docker-compose.yml`:
```yaml
    safe-wallet-web:
      image: safeglobal/safe-wallet-web:latest
      ports:
        - "8080:8080"
      environment:
        - NEXT_PUBLIC_IS_PRODUCTION=true
        - NEXT_PUBLIC_GATEWAY_URL_PRODUCTION=http://localhost:8000/cgw
        - NEXT_PUBLIC_DEFAULT_MAINNET_CHAIN_ID=1337
        - NEXT_PUBLIC_WC_PROJECT_ID=your_wc_project_id
        - NEXT_PUBLIC_INFURA_TOKEN=your_infura_token
```

### Build locally
```bash
# Build from monorepo root
$ docker build -f apps/web/Dockerfile -t safe-wallet-web .

# Run with runtime env vars
$ docker run -p 8080:8080 \
    -e NEXT_PUBLIC_IS_PRODUCTION=true \
    -e NEXT_PUBLIC_GATEWAY_URL_PRODUCTION=http://localhost:8000/cgw \
    -e NEXT_PUBLIC_DEFAULT_MAINNET_CHAIN_ID=1337 \
    safe-wallet-web
```

### Build with chain pre-fetch
``` bash
$ docker build -f apps/web/Dockerfile \
    --build-arg SKIP_FETCH_CHAINS=false \
    --build-arg BUILD_GATEWAY_URL=https://safe-client.safe.global \
    -t safe-wallet-web .
```

## Testing

- [x] Rebuilt image after modifying `NEXT_PUBLIC_GATEWAY_URL_PRODUCTION` — app immediately calls the new gateway URL on next page load (no cache issues)
- [x] `NEXT_PUBLIC_IS_PRODUCTION=true` activates production routing (mainnet chains, production GA, etc.)
- [x] `NEXT_PUBLIC_DEFAULT_MAINNET_CHAIN_ID=1337` correctly sets the default chain to a local network

## Notes

- The `_next/static/` `no-cache` directive requires browsers to revalidate on every page load (one conditional GET). This is intentional: since filename hashes don't change when only runtime env vars differ, `immutable` would prevent the patched chunk from ever being served to returning users after a container restart.
- The injection targets `.env={}` in the main webpack chunk. This pattern is stable across Next.js 14.x builds but should be verified after major Next.js version upgrades.


gjeanmart:
I am closing this for now

PR #7301, opened 2026-03-02
fix(web): infinite loading positions

0xAryanShah:
## What it solves

Resolves: [Linear Issue](https://linear.app/safe-global/issue/WA-1536/infinite-loading-of-positions-for-new-safes)

When opening a Safe with no funds, the Positions tab showed an infinite loading skeleton instead of the empty state.

### Root cause: 
When the portfolio API returns empty data, isPortfolioEmpty triggers a tx-service fallback. That fallback sets positions: undefined — discarding the knowledge that positions were already confirmed empty. The component then treats undefined as "still loading" indefinitely.

### Fix:
Data layer (useTotalBalances.ts): Pass isPortfolioEmpty into aggregateBalances. When the portfolio confirmed empty, overlay positions: [] on the fallback result instead of letting it be undefined. Error fallbacks remain undefined (positions unknown).
UI layer (positions/index.tsx): Remove !protocols from the skeleton condition — isLoading alone drives the skeleton. !protocols now only shows the unavailable state for genuine cases (error or feature disabled on chain).
Tests: Updated to assert positions: [] on empty-portfolio fallback and positions: undefined on error fallback.

### Replaces [this PR](https://github.com/safe-global/safe-wallet-monorepo/pull/7280)

### DIFF
--- a/apps/web/src/features/positions/index.tsx
+++ b/apps/web/src/features/positions/index.tsx
@@ -17,7 +17,7 @@ const Positions = () => {
   const { data: protocols, error, isLoading } = usePositions()
   const portfolio = useLoadFeature(PortfolioFeature)
 
-  if (isLoading || (!error && !protocols)) {
+  if (isLoading) {
     return <PositionsSkeleton />
   }
 
--- a/packages/utils/src/hooks/__tests__/useTotalBalances.test.ts
+++ b/packages/utils/src/hooks/__tests__/useTotalBalances.test.ts
@@ -195,7 +195,7 @@ describe('useTotalBalances', () => {
       expect(item?.fiatBalance24hChange).toBe('0.05')
     })
 
-    it('should fallback to tx service when portfolio returns empty', () => {
+    it('should fallback to tx service when portfolio returns empty data', () => {
       const { result } = setupAndRender(portfolioParams, {
         portfolio: { currentData: createMockEmptyPortfolio() },
         txService: { currentData: createMockTxServiceBalances() },
@@ -203,15 +203,18 @@ describe('useTotalBalances', () => {
 
       expect(result.current.data?.fiatTotal).toBe('1000')
       expect(result.current.data?.tokensFiatTotal).toBe('1000')
+      expect(result.current.data?.positions).toEqual([])
+      expect(result.current.data?.positionsFiatTotal).toBe('0')
     })
 
-    it('should fallback to tx service when portfolio errors', () => {
+    it('should fallback to tx service when portfolio errors with positions unknown', () => {
       const { result } = setupAndRender(portfolioParams, {
         portfolio: { error: new Error('Portfolio error') },
         txService: { currentData: createMockTxServiceBalances() },
       })
 
       expect(result.current.data?.fiatTotal).toBe('1000')
+      expect(result.current.data?.positions).toBeUndefined()
     })
 
     it('should use counterfactual balances on fallback for undeployed safe', () => {
@@ -231,6 +234,8 @@ describe('useTotalBalances', () => {
       )
 
       expect(result.current.data?.fiatTotal).toBe('500')
+      expect(result.current.data?.positions).toEqual([])
+      expect(result.current.data?.positionsFiatTotal).toBe('0')
     })
 
     it('should handle loading state', () => {
--- a/packages/utils/src/hooks/useTotalBalances.ts
+++ b/packages/utils/src/hooks/useTotalBalances.ts
@@ -150,6 +150,7 @@ interface AggregateParams {
   hasPortfolioFeature: boolean
   isAllTokensSelected: boolean
   needsPortfolioFallback: boolean
+  isPortfolioEmpty: boolean
   isCounterfactual: boolean
   txService: TxServiceState
   counterfactual: CounterfactualState
@@ -164,7 +165,13 @@ const aggregateBalances = (p: AggregateParams): TotalBalancesResult => {
   const useTxServiceOnly = !p.hasPortfolioFeature || (p.needsPortfolioFallback && !p.isAllTokensSelected)
 
   if (useTxServiceOnly) {
-    return buildTxServiceResult(p.txService, p.counterfactual, p.isCounterfactual, p.shared)
+    const result = buildTxServiceResult(p.txService, p.counterfactual, p.isCounterfactual, p.shared)
+
+    if (result.data && p.isPortfolioEmpty) {
+      return { ...result, data: { ...result.data, positions: [], positionsFiatTotal: '0' } }
+    }
+
+    return result
   }
 
   if (!p.isAllTokensSelected) {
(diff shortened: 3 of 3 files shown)


0xAryanShah:
image

0xAryanShah:
image

0xAryanShah:
> image

@katspaugh @vseehausen 
- Agreed on that `isLoading` only should be controlling the skeleton -> Updated the condition for Skeleton

- we have this present:
```
if (error || !protocols) return <PositionsUnavailable hasError={!!error} />
```
## For undefined it renders:
image


## For Error it renders:
image


in this exact bug case, when the account is new and `isPortfolioEmpty = True`
with createPortfolioBalance we make the positions undefined which is not right, so keeping these changes as well:
```
export const createPortfolioBalances = (balances: Balances): PortfolioBalances => ({
  ...balances,
  tokensFiatTotal: balances.fiatTotal,
  positionsFiatTotal: '0',
  positions: [],
})
```

this code will now handle any position data: undefined, error, and empty

## empty:
image

packages/utils/src/hooks/portfolioBalances.ts
```diff
@@ -17,7 +17,7 @@ export const createPortfolioBalances = (balances: Balances): PortfolioBalances =
   ...balances,
   tokensFiatTotal: balances.fiatTotal,
   positionsFiatTotal: '0',
-  positions: undefined,
+  positions: [],
```
katspaugh: Do we still need this given the loading state is fixed?

0xAryanShah: in this exact bug case, when the account is new and isPortfolioEmpty = True
with createPortfolioBalance we make the positions undefined which is not right.

## it will render like this:
image

katspaugh: Then `createPortfolioBalance` should probably not make it undefined and make it an empty array instead of having an empty array as the default value?

katspaugh: Here's what Claude suggested:
```
 Fix: Infinite loading skeleton on positions page for empty Safes                                                
                                                        
 Context

 The /balances/positions page shows an infinite loading skeleton for Safes with no assets. A previous fix (PR
 #7280) changed positions: undefined to positions: [] in createPortfolioBalances, but was reverted (PR #7300)
 because it made states indistinguishable and the optional store type dishonest.

 Root cause (data layer): When portfolio returns empty data (positionBalances: []), the fallback to tx-service
 calls createPortfolioBalances which hardcodes positions: undefined — discarding the information that the
 portfolio already confirmed positions are empty.

 Root cause (UI layer): The Positions component conflates "data hasn't arrived" with "data arrived but no
 positions":
 if (isLoading || (!error && !protocols)) {  // ← !protocols catches undefined forever
   return <PositionsSkeleton />
 }

 Changes

 1. Data layer: preserve known positions on empty-portfolio fallback

 File: packages/utils/src/hooks/useTotalBalances.ts

 Pass isPortfolioEmpty through to aggregateBalances so it can distinguish error-fallback (positions unknown →
 undefined) from empty-fallback (positions confirmed empty → []):

 // AggregateParams — add:
 isPortfolioEmpty: boolean

 // aggregateBalances — after buildTxServiceResult, overlay known positions:
 const aggregateBalances = (p: AggregateParams): TotalBalancesResult => {
   const useTxServiceOnly = !p.hasPortfolioFeature || (p.needsPortfolioFallback && !p.isAllTokensSelected)

   if (useTxServiceOnly) {
     const result = buildTxServiceResult(p.txService, p.counterfactual, p.isCounterfactual, p.shared)

     // Portfolio confirmed no positions — preserve that (don't discard as undefined)
     if (result.data && p.isPortfolioEmpty) {
       return { ...result, data: { ...result.data, positions: [], positionsFiatTotal: '0' } }
     }

     return result
   }
   // ... rest unchanged
 }

 // In useTotalBalances, pass the flag:
 aggregateBalances({
   ...existing,
   isPortfolioEmpty: !!isPortfolioEmpty,
 })

 State transparency after fix:

 ┌───────────────────────────┬───────────┬──────────────────────┐
 │         Scenario          │ positions │       Meaning        │
 ├───────────────────────────┼───────────┼──────────────────────┤
 │ Portfolio confirmed empty │ []        │ No positions (known) │
 ├───────────────────────────┼───────────┼──────────────────────┤
 │ Portfolio errored         │ undefined │ Unknown              │
 ├───────────────────────────┼───────────┼──────────────────────┤
 │ No portfolio feature      │ undefined │ Not applicable       │
 └───────────────────────────┴───────────┴──────────────────────┘

 2. UI layer: decouple skeleton from data presence

 File: apps/web/src/features/positions/index.tsx

 Make isLoading the sole skeleton driver. undefined data without error = "unavailable on this network":

 // Before:
 if (isLoading || (!error && !protocols)) {
   return <PositionsSkeleton />
 }
 if (error || !protocols) return <PositionsUnavailable hasError={!!error} />

 // After:
 if (isLoading) {
   return <PositionsSkeleton />
 }
 if (error || !protocols) return <PositionsUnavailable hasError={!!error} />

 After the data layer fix, empty safes get protocols: [] (not undefined), so they correctly reach
 protocols.length === 0 → PositionsEmpty. The !protocols check now only triggers for genuinely unavailable cases
  (error or feature disabled on this chain).

 3. Update tests

 File: packages/utils/src/hooks/__tests__/useTotalBalances.test.ts
 - Empty-portfolio fallback test should expect positions: [] (not undefined)
 - Error-portfolio fallback test should still expect positions: undefined

 File: apps/web/src/hooks/loadables/__tests__/useLoadBalances.test.ts
 - Update expectations for empty-portfolio fallback scenarios

 Files to modify

 - packages/utils/src/hooks/useTotalBalances.ts — data layer fix
 - apps/web/src/features/positions/index.tsx — UI condition fix
 - packages/utils/src/hooks/__tests__/useTotalBalances.test.ts — test updates
 - apps/web/src/hooks/loadables/__tests__/useLoadBalances.test.ts — test updates

 Verification

 yarn workspace @safe-global/web type-check
 yarn workspace @safe-global/web test -- --testPathPattern="useTotalBalances|useLoadBalances|usePositions"
 yarn workspace @safe-global/web lint
 yarn prettier:fix

 Manual: navigate to /balances/positions?safe=eth:0x11414354ddae518F8333e470b35EDE2485d8B1fC — should show empty
  positions state, not infinite skeleton.
```

0xAryanShah:
@claude review

0xAryanShah:
@claude review again pls. I addressed your previously found issues:

1. isPortfolioEmpty will only be true when its not errored out and we have checked portfolioData.positionBalances.length === 0 and if it errors out we have tis:
const mergedError = portfolio.error || txService.error   if (mergedError) { return { data: undefined, error: new Error(String(mergedError)), loading: false, ...shared }   }
in the buildMergedResult which will return error when it errors out to show the user the errored state.

3. The condition is not overly broad it's precise. isPortfolioEmpty can only be truthy when portfolio actually responded with empty data, never when it errored. The test correctly passes.

packages/utils/src/hooks/portfolioBalances.ts
0xAryanShah: Added the flag to persist the portfolio data, not to be overwritten as undefined

0xAryanShah:
> Looks good! Did you test all three states manually (loading, empty positions, positions disabled)?

Yes

PR #7314, opened 2026-03-03
fix(web): remove SIWE requirement on home screen

clovisdasilvaneto:
## Summary
- Skip SIWE authentication on the welcome/home screen login flow by adding a `skipSiwe` option to `useHomeAuth`
- Simplify `WelcomeLogin` component: always show "Watch any account" instead of conditionally showing "View my accounts" based on spaces
- Clean up `useFlowActivationGuard`: remove debug `console.log` statements, remove unused `PUBLIC_ROUTES` array, simplify public route detection using `useIsSpaceRoute`, and simplify the "not connected" guard to only check SIWE auth

## Test plan
- [ ] Run `yarn workspace @safe-global/web storybook` and verify UI/ stories still render correctly
- [ ] Check a Spaces story in dark mode — shadcn variables should apply
- [ ] Check a regular MUI story (e.g. Components/) — MUI should still work, no visual regressions

### DIFF
--- a/apps/web/src/components/dashboard/Assets/index.tsx
+++ b/apps/web/src/components/dashboard/Assets/index.tsx
@@ -1,4 +1,3 @@
-/** @deprecated Use features/assets instead */
 import { useMemo } from 'react'
 import { Box, Skeleton, Typography, Paper, Stack, Divider } from '@mui/material'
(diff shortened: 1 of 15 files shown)


vmosharova (approved):
👍🏻

PR #7323, opened 2026-03-04
feat(web): deadlock check for nested safe

0xAryanShah:
## What it solves
Deadlock detection for nested Safes : [Linear Issue](https://linear.app/safe-global/issue/WA-1616/prevent-nested-safe-signing-deadlocks-with-pre-submit-graph-validation)
                                                                                                                            
## Summary
Adds deadlock analysis to the Safe Shield feature that detects circular signer dependencies in nested Safe configurations. When Safes are configured as signers of each other, it can create a deadlock where neither Safe can reach its signing threshold this feature identifies and warns users about such scenarios.

## Key changes:
  - Added deadlockCheck utility in packages/utils that analyzes signer graphs for circular dependencies and unreachable
  thresholds
  - Added useDeadlockAnalysis hook that fetches signer data recursively and runs the deadlock detection
  - Added DeadlockAnalysisCard component to surface deadlock warnings in the Safe Shield UI
  - Integrated deadlock warnings into owner management tx flows (Add/Remove owner, Change threshold, Manage signers) to warn
   before submitting risky changes
  - Updated getOverallStatus to factor deadlock status into the overall shield score
  - Extended SafeShieldContext to provide deadlock analysis data
  - Added analytics events for deadlock-related interactions
  - Unit tests for the deadlock detection logic and the analysis hook


0xAryanShah:
@claude review with the pending(todo) changes in context

0xAryanShah:
@claude review again, addressed your previous review comments:
1. Racing condition addressed
2. useLazySafesGetSafeV1Query already handles response caching
4,5,6. fixed the type safety issues
7. added storybook component

0xAryanShah:
@claude review again, addressed your previous comments:
1) Reduced Complexity of the functions
2) Introduced Batching of requests , also flattening before batching

apps/web/src/features/safe-shield/components/DeadlockAnalysisCard/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+const DeadlockAnalysisCard = ({
+  result,
+  loading,
+}: {
+  result?: DeadlockCheckResult
+  loading?: boolean
+}): ReactElement | null => {
+  const [isOpen, setIsOpen] = useState(false)
+
+  if (loading || !result || result.status === DeadlockStatus.VALID) {
```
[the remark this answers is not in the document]
0xAryanShah: need to check all 3

apps/web/src/components/tx-flow/flows/AddOwner/ReviewOwner.tsx
```diff
@@ -27,6 +30,22 @@ export const ReviewOwner = ({
...
+
+  const change: OwnerChange = useMemo(() => {
+    if (removedOwner) {
+      return { type: 'swapOwner', oldOwnerAddress: removedOwner.address, newOwnerAddress: newOwner.address }
+    }
+    return { type: 'addOwner', ownerAddress: newOwner.address, threshold }
+  }, [removedOwner, newOwner.address, threshold])
+
+  const projected = useMemo(
+    () => computeProjectedState(currentOwners, safe.threshold, change),
```
tmjssz, line 43: How about passing the `safe` object directly? Then we can also move the mapping of the current owner addresses into the computeProjectedState function and don't need to do it in all the flows.

packages/utils/src/features/safe-shield/utils/deadlockCheck.ts
```diff
@@ -0,0 +1,108 @@
+import type { OwnerChange, SafeOwnerInfo, DeadlockCheckResult } from '../types'
+import { DeadlockReason, DeadlockStatus } from '../types'
+
+export function computeProjectedState(
```
tmjssz, line 4: I think the function name is not really descriptive. Also, some JS docs would help to understand what this function does and is needed for.

packages/utils/src/features/safe-shield/utils/deadlockCheck.ts
```diff
@@ -0,0 +1,108 @@
...
+): { owners: string[]; threshold: number } {
+  switch (change.type) {
+    case 'addOwner':
+      return {
+        owners: [...currentOwners, change.ownerAddress],
+        threshold: change.threshold,
+      }
+    case 'removeOwner':
+      return {
+        owners: currentOwners.filter((o) => o.toLowerCase() !== change.ownerAddress.toLowerCase()),
```
tmjssz, line 17: Lets use `sameAddress` from the utils package here

packages/utils/src/features/safe-shield/utils/deadlockCheck.ts
```diff
@@ -0,0 +1,108 @@
...
+        threshold: change.threshold,
+      }
+    case 'swapOwner':
+      return {
+        owners: currentOwners.map((o) =>
+          o.toLowerCase() === change.oldOwnerAddress.toLowerCase() ? change.newOwnerAddress : o,
```
tmjssz, line 23: Also here we can use sameAddress. There are multiple other places where that function can also be used

packages/utils/src/features/safe-shield/utils/deadlockCheck.ts
```diff
@@ -0,0 +1,108 @@
...
+        ),
+        threshold: currentThreshold,
+      }
+    case 'changeThreshold':
+      return {
+        owners: [...currentOwners],
```
tmjssz, line 29: Why is spreading needed here?

packages/utils/src/features/safe-shield/types/index.ts
```diff
@@ -215,3 +215,47 @@ export type ThreatAnalysisResults = {
...
+
+export enum DeadlockReason {
+  FETCH_FAILURE = 'Could not fetch owner data for one or more Safe signers.',
+  MUTUAL_DEADLOCK = 'This change may create a signing cycle between Safes and can permanently lock funds. You will not be allowed to proceed forward.',
+  DEEP_NESTING = 'One or more Safe signers have their own Safe signers. Full signer safety could not be verified beyond direct owners.',
+  GENERIC = 'A signer deadlock was detected in the projected owner configuration.',
+}
+
+export type OwnerChange =
+  | { type: 'addOwner'; ownerAddress: string; threshold: number }
```
tmjssz, line 236: I think it would be good to also define an enum for the type.

packages/utils/src/features/safe-shield/types/index.ts
```diff
@@ -215,3 +215,47 @@ export type ThreatAnalysisResults = {
   BALANCE_CHANGE?: BalanceChangeDto[]
   request_id?: string
 }
+
+// Deadlock check types
```
tmjssz, line 219: Lets move all the deadlock check related types into a separate file

apps/web/src/features/safe-shield/SafeShieldContext.tsx
```diff
@@ -134,3 +152,23 @@ export const useSafeShieldForTxData = (txData: SafeTransaction | undefined) => {
     }
   }, [txData, setSafeTx])
 }
+
+/**
+ * Hook to register deadlock check params for Safe Shield analysis.
+ * Called from Review components that know the projected owner state.
+ */
+export const useSafeShieldForDeadlockCheck = (
```
tmjssz, line 160: Can we move this hook into a separate file?

apps/web/src/features/safe-shield/SafeShieldContext.tsx
```diff
@@ -134,3 +152,23 @@ export const useSafeShieldForTxData = (txData: SafeTransaction | undefined) => {
...
+  editedSafeAddress: string,
```
tmjssz, line 161: Instead of passing the safeAddress as a param, we could call the `useSafeAddress` hook inside this one

katspaugh:
The architecture doesn't look right to me.
* We're adding new deadlock props to ExecuteForm -- this is very specific to this particular case
* `useSafeShieldForDeadlockCheck` is called from many places, whereas I thought SafeShield was the centralized place that has all the context about a tx and can take decisions wrt what warnings to show

katspaugh:
```
Refactor: Make SafeShield self-contained for deadlock checks                                                               
                                                                   
 Context

 Two architectural problems with the current deadlock check implementation:

 1. Feature surface area leaks into Review components. Four Review components each call useSafeShieldForDeadlockCheck to
 register projected owner/threshold state. But SafeShield already has safeTx via SafeTxContext — it can decode the
 transaction to detect owner/threshold changes automatically, just like other analyses (recipient, contract, threat) that
 derive their inputs from safeTx.
 2. BLOCKED deadlock bypasses the existing risk confirmation pattern. TxFlowProvider pulls deadlock from the context,
 checks DeadlockStatus.BLOCKED, and adds a separate isDeadlockBlocked flag. But SafeShield already has
 needsRiskConfirmation + isRiskConfirmed with a confirmation checkbox — the BLOCKED status should feed into this existing
 mechanism instead of creating a parallel blocking path. This also leaks deadlock and setDeadlockCheckParams into the
 txSecurity prop type on ExecuteForm/SignForm.

 Existing building blocks to reuse:
 - isAddOwnerWithThresholdCalldata, isRemoveOwnerCalldata, isSwapOwnerCalldata, isChangeThresholdCalldata in
 apps/web/src/utils/transaction-calldata.ts
 - decodeOwnerManagementTransaction pattern in apps/web/src/features/recovery/services/transaction-list.ts:15-62
 - computeProjectedState in packages/utils/src/features/safe-shield/utils/deadlockCheck.ts

 Changes

 1. Create a utility to extract deadlock params from safeTx

 New file: apps/web/src/features/safe-shield/utils/extractDeadlockParams.ts

 Takes a SafeTransaction and current Safe state, returns DeadlockCheckParams | null:
 - Check if safeTx.data.data matches any owner/threshold change selector (using existing detection functions)
 - If yes, decode the calldata using ethers Interface (following the pattern in recovery/services/transaction-list.ts)
 - Handle multi-send batches (using decodeMultiSendData for the recovery/ReviewSigners case)
 - Call computeProjectedState with the decoded change
 - Return { editedSafeAddress, projectedOwners, projectedThreshold }
 - If the tx is not an owner/threshold change, return null

 2. Make SafeShieldProvider self-contained

 File: apps/web/src/features/safe-shield/SafeShieldContext.tsx

 - Import the new extraction utility
 - Replace setDeadlockCheckParams state + useSafeShieldForDeadlockCheck hook with a useMemo that derives deadlock params
 from safeShieldTx (already available) + current Safe info from useSafeInfo
 - Pass derived params directly to useDeadlockAnalysis
 - Remove setDeadlockCheckParams and deadlock from the context type — these are internal to SafeShield now
 - Remove useSafeShieldForDeadlockCheck export
 - Route ALL deadlock statuses (including BLOCKED) through needsRiskConfirmation — the existing confirmation checkbox
 handles it. Currently only WARNING/UNKNOWN feed into needsRiskConfirmation; BLOCKED should too.

 3. Remove deadlock hook calls from Review components

 Files to clean up (remove useSafeShieldForDeadlockCheck, computeProjectedState, and related useMemo calls):
 - apps/web/src/components/tx-flow/flows/AddOwner/ReviewOwner.tsx
 - apps/web/src/components/tx-flow/flows/RemoveOwner/ReviewRemoveOwner.tsx
 - apps/web/src/components/tx-flow/flows/ChangeThreshold/ReviewChangeThreshold.tsx
 - apps/web/src/components/tx-flow/flows/ManagerSigners/ReviewSigners.tsx

 4. Revert TxFlowProvider to pre-deadlock state

 File: apps/web/src/components/tx-flow/TxFlowProvider.tsx

 - Remove DeadlockStatus import
 - Remove deadlock destructuring from useSafeShield()
 - Remove isDeadlockBlocked computation
 - Revert to the original pattern: const isUntrustedSafeBlocked = needsRiskConfirmation && !isRiskConfirmed
 - Use isUntrustedSafeBlocked in isSubmitDisabled (same as before this branch)

 This works because BLOCKED deadlock now feeds into needsRiskConfirmation inside SafeShield, so the existing
 needsRiskConfirmation && !isRiskConfirmed check covers it.

 5. Update tests

 - apps/web/src/features/safe-shield/__tests__/SafeShieldContext.test.tsx — remove setDeadlockCheckParams references, test
 that BLOCKED deadlock triggers needsRiskConfirmation
 - apps/web/src/components/tx-flow/actions/Execute/__tests__/ExecuteForm.test.tsx — remove deadlock/setDeadlockCheckParams
 from mock txSecurity
 - apps/web/src/components/tx-flow/actions/Sign/__tests__/SignForm.test.tsx — same
 - Add unit tests for the new extractDeadlockParams utility

 Key files reference

 ┌───────────────────────────────────────────────────────────────────┬──────────────────────────────────────────────────┐
 │                               File                                │                       Role                       │
 ├───────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
 │ apps/web/src/utils/transaction-calldata.ts                        │ Existing selector detection functions (reuse)    │
 ├───────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
 │ apps/web/src/features/recovery/services/transaction-list.ts:15-62 │ Existing decoding pattern to follow              │
 ├───────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
 │ packages/utils/src/features/safe-shield/utils/deadlockCheck.ts    │ computeProjectedState + checkDeadlock (reuse)    │
 ├───────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
 │ packages/utils/src/features/safe-shield/types/index.ts            │ OwnerChange, DeadlockCheckParams types           │
 ├───────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
 │ apps/web/src/features/safe-shield/SafeShieldContext.tsx           │ Main changes — auto-derive deadlock params,      │
 │                                                                   │ route BLOCKED through needsRiskConfirmation      │
 ├───────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
 │ apps/web/src/features/safe-shield/hooks/useDeadlockAnalysis.ts    │ Unchanged — still does the actual analysis       │
 ├───────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
 │ apps/web/src/components/tx-flow/TxFlowProvider.tsx                │ Revert to pre-deadlock simplicity                │
 └───────────────────────────────────────────────────────────────────┴──────────────────────────────────────────────────┘

 Verification

 1. yarn workspace @safe-global/web type-check
 2. yarn workspace @safe-global/web lint
 3. yarn workspace @safe-global/web test (specifically safe-shield and tx-flow tests)
 4. Manual test: open AddOwner/RemoveOwner/ChangeThreshold flows and verify the deadlock warning + confirmation checkbox
 still appears in SafeShield
```

tmjssz:
I would suggest this for the refactor:
- `SafeShieldContext` already has safeTx. We can call [useTxDetails(txId)](https://github.com/safe-global/safe-wallet-monorepo/blob/feat%2Fdeadlock-check/apps/web/src/hooks/useTxDetails.ts/#L5) there and expose txDetails (and use it for other analyses too).
- Use that txDetails as the single source to derive owner updates for the deadlock check: when txDetails is present and the tx is a settings change, compute projected owners/threshold from it instead of each flow pushing params.
- Concretely: if txDetails.txInfo is a [SettingsChange](https://github.com/safe-global/safe-wallet-monorepo/blob/feat%2Fdeadlock-check/packages/store/src/gateway/AUTO_GENERATED/transactions.ts/#L494-L510) (guard: [isSettingsChangeTxInfo](https://github.com/safe-global/safe-wallet-monorepo/blob/feat%2Fdeadlock-check/apps/web/src/utils/transaction-guards.ts/#L183)), then txDetails.txInfo.settingsInfo is one of `AddOwner | RemoveOwner | SwapOwner | ChangeThreshold | …` and we can derive added/removed/swapped owners and the new threshold from that in one place.
- That keeps Safe Shield as the only place that decides “what’s the projected owner state?”

PR #7335, opened 2026-03-05
feat(mobile): add "Propose with" signer selector to Send flow

compojoom:
## What it solves

The `activeSigner` (per-safe Redux state) was never set when switching safes, causing the "Review & confirm" button to stay permanently disabled on mainnet. Users with multiple signer keys also had no way to choose which key to propose with.

## How this PR fixes it

Introduces a `useEnsureActiveSigner` hook that intersects safe owners with imported signers and auto-selects the first available one. It also detects stale signers (e.g., an owner removed from the Safe but key still on device) and re-selects automatically. A `SelectProposer` row and `ProposerBottomSheet` let users with 2+ signer keys pick which one to propose with. The sender address is now validated and checksummed in `proposeSendTransaction`.

## How to test it

1. Switch to a mainnet Safe (or any Safe where you have an imported signer key)
2. Navigate to Send → Enter amount screen
3. Verify the "Propose with" row appears above the button with the signer's identicon and name
4. Verify "Review & confirm" button is enabled (previously stayed disabled)
5. If you have 2+ signer keys that are owners, tap the "Propose with" row → bottom sheet opens with signer list → select a different signer → sheet closes and selection updates
6. If you have only 1 signer key, verify no chevron appears and the row is not tappable

## Screenshots

N/A - reviewer should test on device/simulator

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/Send/hooks/useMaxAmount.test.ts
+++ b/apps/mobile/src/features/Send/hooks/useMaxAmount.test.ts
@@ -8,7 +8,9 @@ jest.mock('@safe-global/utils/utils/formatters', () => ({
     const divisor = BigInt(10) ** BigInt(decimals)
     const whole = num / divisor
     const remainder = num % divisor
-    if (remainder === 0n) return whole.toString()
+    if (remainder === 0n) {
+      return whole.toString()
+    }
     const fracStr = remainder.toString().padStart(decimals, '0').replace(/0+$/, '')
     return `${whole}.${fracStr}`
   },


apps/mobile/src/features/Send/hooks/useEnsureActiveSigner.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (needsSignerUpdate) {
+      dispatch(
+        setActiveSigner({
+          safeAddress: activeSafe.address,
+          signer: availableSigners[0],
+        }),
+      )
+    }
+  }, [needsSignerUpdate, availableSigners, dispatch, activeSafe.address])
+
```
clovisdasilvaneto: Isn't it better to have this inside a function returned by the hook and the consumer is who is responsible for calling it on a effect or not?

Also would be nice to have this done after a user action instead a effect, so we avoid strange behavior due to re-renders of the app

PR #7382, opened 2026-03-10
chore: fix pre-push hook running lint --fix too late

katspaugh:
## Summary
- Replaced `yarn run lint --fix` in pre-push with read-only `yarn prettier` and `yarn run lint` checks — the old `--fix` modified files after commit, making it impossible to push without manual re-commit
- Added `storybook-static` to `.prettierignore` (build artifact was failing the new prettier check)

## Test plan
- [ ] Commit unformatted code → pre-commit lint-staged auto-fixes and stages it
- [ ] Push with lint warnings → pre-push shows warnings but exits 0 (warnings only)
- [ ] Push with prettier violations → pre-push fails with actionable message

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -19,6 +19,7 @@ apps/mobile/.storybook
 apps/mobile/resources/
 
 # Web
+apps/web/storybook-static
 apps/web/.next
 apps/web/cypress/downloads
 apps/web/.swc


compojoom:
@katspaugh I don't think we need this in light of the changes I did here: https://github.com/safe-global/safe-wallet-monorepo/commit/7d92406e683af39a9a3e0a9a89a7b904bc0b6322

Prettier runs on commit and fixes any issues in the staged files. So you should not have any issues on pre-push and would just be wasting time as `yarn prettier` will run on all files in all packages and not only on files modified in the commits for the PR ?

katspaugh:
@compojoom ah, thanks, I wasn't aware you had already fixed that. Closing then.

PR #7383, opened 2026-03-10
fix(web/Spaces): redirect to `/welcome/spaces` when not SIWE-authenticated

vmosharova:
## What it solves

- Unauthenticated users could stay on `/spaces` routes instead of being sent to the spaces welcome flow. The page was blank.
- Plus one mini-addition: change font weight for the Total Value.

Resolves: [WA-1715](https://linear.app/safe-global/issue/WA-1715/bug-can-not-see-the-create-space-onboarding-page)

## How this PR fixes it

- Treat any path under `/spaces` as non-public so the guard no longer allows unauthenticated access there.
- Redirect unauthenticated users on `/spaces` to `/welcome/spaces` (existing rule; it was skipped before because those paths were treated as public when `isSpaceRoute` was false).
- Refactor: add `isSpacesPath` to guard context, compute `isOnboardingRoute` once, and use them in rules to remove repeated path checks.

## How to test it

1. Log out (or use an incognito session). Open `/spaces` or `/spaces/settings` → should redirect to `/welcome/spaces`.
2. Connect wallet but do not authenticate. Open `spaces/create-space` → should redirect to `/welcome/spaces`.

## Screenshots

N/A

## Checklist

- [ ] I've documented how it affects the analytics (if at all) 📊 - N/A

---

## CLA signature

### DIFF
--- a/apps/web/src/hooks/useRouterGuard/activationGuards/useFlowActivationGuard.ts
+++ b/apps/web/src/hooks/useRouterGuard/activationGuards/useFlowActivationGuard.ts
@@ -33,18 +33,12 @@ const guardRules: GuardRule[] = [
     action: () => allow(),
   },
 
-  // Spaces routes require SIWE authentication → if not authenticated, redirect to welcome/spaces
-  {
-    match: ({ isSpacesPath, isSiweAuthenticated }) => isSpacesPath && !isSiweAuthenticated,
-    action: () => redirect(AppRoutes.welcome.spaces),
-  },
-
-  // Not connected or not signed in with SIWE → welcome
+  // Not connected or not signed in with SIWE → welcome (spaces path → welcome/spaces)
   {
     match: ({ isSiweAuthenticated }) => {
       return !isSiweAuthenticated
     },
-    action: () => redirect(AppRoutes.welcome.index),
+    action: ({ isSpacesPath }) => redirect(isSpacesPath ? AppRoutes.welcome.spaces : AppRoutes.welcome.index),
   },
 
   // Authenticated but has no spaces → onboarding


apps/web/src/hooks/useRouterGuard/activationGuards/useFlowActivationGuard.ts
```diff
@@ -33,6 +33,20 @@ const guardRules: GuardRule[] = [
...
+      return !isStoreHydrated
+    },
+    action: () => allow(),
+  },
+
+  // Spaces routes require SIWE authentication → if not authenticated, redirect to welcome/spaces
+  {
+    match: ({ isSpacesPath, isSiweAuthenticated }) => isSpacesPath && !isSiweAuthenticated,
+    action: () => redirect(AppRoutes.welcome.spaces),
+  },
```
clovisdasilvaneto: The line jut bellow already does that. We don't need this one

vmosharova: Removed + added a small improvement: [219c9c7](https://github.com/safe-global/safe-wallet-monorepo/pull/7383/commits/219c9c73ffec575b5240999c74a3f72a17237515)

apps/web/src/hooks/useRouterGuard/activationGuards/useFlowActivationGuard.ts
```diff
@@ -33,12 +33,12 @@ const guardRules: GuardRule[] = [
...
   },
 
-  // Not connected or not signed in with SIWE → welcome
+  // Not connected or not signed in with SIWE → welcome (spaces path → welcome/spaces)
   {
     match: ({ isSiweAuthenticated }) => {
       return !isSiweAuthenticated
     },
-    action: () => redirect(AppRoutes.welcome.index),
+    action: ({ isSpacesPath }) => redirect(isSpacesPath ? AppRoutes.welcome.spaces : AppRoutes.welcome.index),
```
clovisdasilvaneto, line 41: nice

PR #7385, opened 2026-03-11
feat(mobile): integrate Datadog RUM with consent-aware tracking

compojoom:
## What it solves

Adds Datadog Real User Monitoring (RUM) to the mobile app for session replay, view tracking, and automatic interaction tracking — all gated behind the existing analytics consent flow.

Resolves: https://linear.app/safe-global/issue/WA-1553/mobile-app-datadog-service-catalog

## How this PR fixes it

The SDK initializes with `TrackingConsent.NOT_GRANTED` so no data is collected until the user accepts in the GetStarted flow. Existing users who already consented get backfilled via a store migration (v2→v3). The Datadog babel plugin handles automatic interaction tracking (button taps, etc.) without manual instrumentation.

Key decisions:
- **Site**: `datadoghq.eu` (EU1)
- **Session sample rates**: 0% dev / configurable via env for preview+production
- **User identity**: Safe address + chain ID as extra attributes (not as the user ID itself, since one user can have multiple Safes across chains)
- **dSYM/sourcemap uploads**: Gated on `EAS_BUILD` env var so local simulator builds skip uploads
- **Debug ID warning**: The `unstable_beforeAssetSerializationPlugins` option may not be supported by current Expo SDK version — non-blocking, only affects error symbolication correlation

## How to test it

1. Add `EXPO_PUBLIC_DD_CLIENT_TOKEN`, `EXPO_PUBLIC_DD_APPLICATION_ID`, `EXPO_PUBLIC_DD_SITE=EU1`, `EXPO_PUBLIC_DD_ENV=development` to `apps/mobile/.env.local`
2. Run `npx expo prebuild --clean` then build and launch the app
3. Accept analytics in the GetStarted flow — verify DD logs show `Setting consent granted` and `Setting user`
4. Navigate between screens — verify `Starting RUM View` / `Stopping RUM View` appear in logs and events show up in Datadog dashboard

## Screenshots

N/A - no UI changes

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/hooks/__tests__/useDatadogConsent.test.ts
+++ b/apps/mobile/src/hooks/__tests__/useDatadogConsent.test.ts
@@ -73,7 +73,7 @@ describe('useDatadogConsent', () => {
     expect(DdSdkReactNative.addUserExtraInfo).not.toHaveBeenCalled()
   })
 
-  it('should not set DD user when activeSafe is null', () => {
+  it('should clear safe context when consented but activeSafe is null', () => {
     let callIndex = 0
     useSelector.mockImplementation(() => {
       callIndex++
@@ -82,6 +82,9 @@ describe('useDatadogConsent', () => {
 
     renderHook(() => useDatadogConsent())
 
-    expect(DdSdkReactNative.addUserExtraInfo).not.toHaveBeenCalled()
+    expect(DdSdkReactNative.addUserExtraInfo).toHaveBeenCalledWith({
+      safeAddress: '',
+      chainId: '',
+    })
   })
 })
--- a/apps/mobile/src/hooks/useDatadogConsent.ts
+++ b/apps/mobile/src/hooks/useDatadogConsent.ts
@@ -14,10 +14,19 @@ export const useDatadogConsent = () => {
   }, [consented])
 
   useEffect(() => {
-    if (consented && activeSafe?.address && activeSafe?.chainId) {
+    if (!consented) {
+      return
+    }
+
+    if (activeSafe?.address && activeSafe?.chainId) {
       DdSdkReactNative.addUserExtraInfo({
         safeAddress: activeSafe.address,
         chainId: activeSafe.chainId,
+      })
+    } else {
+      DdSdkReactNative.addUserExtraInfo({
+        safeAddress: '',
+        chainId: '',
       })
     }
   }, [consented, activeSafe?.address, activeSafe?.chainId])


apps/mobile/src/providers/DatadogWrapper.tsx
```diff
@@ -0,0 +1,55 @@
...
+} from 'expo-datadog'
+
+const clientToken = process.env.EXPO_PUBLIC_DD_CLIENT_TOKEN ?? ''
+const applicationId = process.env.EXPO_PUBLIC_DD_APPLICATION_ID ?? ''
+const ddDebug = process.env.EXPO_PUBLIC_DD_DEBUG === 'true'
+
+const config = new DatadogProviderConfiguration(
+  clientToken,
+  process.env.EXPO_PUBLIC_DD_ENV ?? 'production',
+  TrackingConsent.NOT_GRANTED,
```
gfendres, line 19: Do we need to ask consent for that? if we don't track any PII?

apps/mobile/src/providers/DatadogWrapper.tsx
```diff
@@ -0,0 +1,55 @@
...
+  {
+    site: process.env.EXPO_PUBLIC_DD_SITE ?? 'EU1',
+    service: 'safe-mobile',
+    verbosity: ddDebug ? SdkVerbosity.DEBUG : SdkVerbosity.WARN,
+    uploadFrequency: ddDebug ? UploadFrequency.FREQUENT : UploadFrequency.AVERAGE,
+    batchSize: ddDebug ? BatchSize.SMALL : BatchSize.MEDIUM,
+    rumConfiguration: {
+      applicationId,
```
gfendres, line 27: I think those things could be env variables no? 
I think the track interactions is to be use for product like MixPanel, 
we want datadog only for error and performance.

apps/mobile/src/providers/DatadogWrapper.tsx
```diff
@@ -0,0 +1,55 @@
...
+      trackInteractions: true,
+      trackResources: true,
+      trackErrors: true,
+      sessionSampleRate: ddDebug ? 100 : 80,
+      nativeCrashReportEnabled: false,
```
gfendres, line 32: what is the difference here between false and true? why false in this case

compojoom: RUM is currently not configured to collect the native crash reports (app closes completely) - those are send to crashlytics.

apps/mobile/src/providers/DatadogWrapper.tsx
compojoom: Doesn't matter as I can't manipulate them in production. We always have to create a new release.

PR #7409, opened 2026-03-12
feat(safe-shield): extend Hypernative threat analysis to EIP-712 typed messages

gjeanmart:
> A typed message arrives —
> permit slips through the firewall,
> Hypernative stirs.
> Hash of domain and intent,
> Safe Shield reads between the lines.

## Summary

- Extends the Hypernative Safe Shield threat analysis to cover **EIP-712 typed messages** (offchain signing), not just on-chain `SafeTransaction` objects. Previously, Hypernative scanned transactions only; messages were unscanned.
- Adds a new `useThreatAnalysisHypernativeMessage` hook that calls Hypernative's `/safe/eip712/assessment` endpoint, with deduplication via a ref to avoid redundant calls for the same message hash.
- Adds a `assessMessage` RTK mutation and full DTO types (`HypernativeMessageAssessmentRequestDto`, response/error types) in `@safe-global/store`.
- Extends `SafeTxContext` with `safeMessageHash`/`setSafeMessageHash` so the message hash is available to the threat analysis layer without prop drilling.
- Fixes routing in `useThreatAnalysis`: selects the message hook for EIP-712 messages and the transaction hook otherwise; both correctly respect `useHypernativeAnalysis` (feature flag + eligibility combined).

## Test plan

- [ ] Open a Safe and initiate an EIP-712 typed message signing flow (e.g. a Permit or SIWE signature)
- [ ] Verify the Hypernative message assessment API is called once with the correct `messageHash`, `safeAddress`, `chainId`, and typed data payload
- [ ] Verify no duplicate API calls are made if the component re-renders without the message changing
- [ ] Verify threat results (warn/deny) surface correctly in the Safe Shield risk confirmation UI
- [ ] Verify the feature is skipped for plain-text (non EIP-712) messages
- [ ] Verify Blockaid is used (not Hypernative) when the Hypernative feature flag is disabled or the user is ineligible
- [ ] Verify transaction scanning is unaffected — normal `SafeTransaction` flows still route to the transaction assessment endpoint
- [ ] Run unit tests: `yarn workspace @safe-global/web test` and `yarn workspace @safe-global/store test`
- [ ] Run type-check: `yarn workspace @safe-global/web type-check`

### DIFF
--- a/packages/utils/src/features/safe-shield/hooks/__tests__/useThreatAnalysisHypernative.test.ts
+++ b/packages/utils/src/features/safe-shield/hooks/__tests__/useThreatAnalysisHypernative.test.ts
@@ -20,12 +20,16 @@ jest.mock('@safe-global/protocol-kit/dist/src/utils', () => ({
 jest.mock('@safe-global/store/hypernative/hypernativeApi', () => ({
   hypernativeApi: {
     useAssessTransactionMutation: jest.fn(),
+    useAssessMessageMutation: jest.fn(),
   },
 }))
 
 const mockIsSafeTransaction = isSafeTransaction as jest.MockedFunction<typeof isSafeTransaction>
 const mockUseAssessTransactionMutation = hypernativeApi.useAssessTransactionMutation as jest.MockedFunction<
   typeof hypernativeApi.useAssessTransactionMutation
+>
+const mockUseAssessMessageMutation = hypernativeApi.useAssessMessageMutation as jest.MockedFunction<
+  typeof hypernativeApi.useAssessMessageMutation
 >
 
 // Import the mocked function
@@ -113,6 +117,10 @@ describe('useThreatAnalysisHypernative', () => {
     mockCalculateSafeTransactionHash.mockReturnValue(mockSafeTxHash as `0x${string}`)
     mockUseAssessTransactionMutation.mockReturnValue([
       mockTriggerAssessment,
+      { data: undefined, error: undefined, isLoading: false },
+    ] as any)
+    mockUseAssessMessageMutation.mockReturnValue([
+      jest.fn(),
       { data: undefined, error: undefined, isLoading: false },
     ] as any)
   })


gjeanmart:
@claude review

packages/utils/src/features/safe-shield/hooks/useThreatAnalysisHypernativeMessage.ts
compojoom, line 1: The purpose of the store package is to deal with the redux store. We use RTK for network calls that directly manipulate the redux store and have some other redux utilities in that package. By the look of this hook it seems more to be an utility and I would rather move it to the utils package.

gjeanmart: I was trying to keep the same convention as for the rest of Hypernative useThreadAnalysisXXXX hooks:

image

But I can refactor a bit if you think it's necessary ;)

compojoom: ufff, sorry, nevermind! I got confused by github. I clicked on the store/src/hyprenative/hypernativeApi.ts in the file tree on the left and then started scrolling. The file tree on the left is still marked at store/src/... but I didn't realise that I'm looking at a hook in the utils package. 🤦 
All good!

apps/web/src/components/tx-flow/flows/SignMessage/SignMessage.tsx
```diff
@@ -314,8 +316,9 @@ const SignMessage = ({ message, origin, requestId }: SignMessageProps): ReactEle
   useEffect(() => {
     if (isEip712) {
       setContextSafeMessage(decodedMessage)
+      setContextSafeMessageHash(safeMessageHash as `0x${string}`)
     }
-  }, [decodedMessage, isEip712, setContextSafeMessage])
+  }, [decodedMessage, isEip712, setContextSafeMessage, setContextSafeMessageHash, safeMessageHash])
```
tmjssz, line 324: Might be good to also clear the context values when isEip712 is false, otherwise stale data could remain and lead to incorrect threat analysis.

apps/web/src/features/safe-shield/hooks/useThreatAnalysis.ts
```diff
@@ -57,14 +61,33 @@ export function useThreatAnalysis(
...
+  // Use message-specific assessment for EIP-712 typed messages
+  const hypernativeMessageThreatAnalysis = useThreatAnalysisHypernativeMessage({
+    safeAddress: safeAddress as `0x${string}`,
+    chainId,
+    messageHash: safeMessageHash as `0x${string}`,
+    typedData: safeMessage,
+    origin: txOrigin,
+    authToken: hypernativeAuthToken,
+    skip: !useHypernativeAnalysis || !hypernativeAuthToken || !isMessageAnalysis,
+  })
```
tmjssz: I think it would make sense to call this hook inside `useThreatAnalysisHypernative` so it becomes part of the HN threat analysis flow

apps/web/src/components/tx-flow/flows/SignMessage/SignMessage.tsx
gjeanmart: good point. fixed: [f800497](https://github.com/safe-global/safe-wallet-monorepo/pull/7409/commits/f8004976c88fe075d2247400d5f6a9a27d3458b3)

apps/web/src/features/safe-shield/hooks/useThreatAnalysis.ts
gjeanmart: 100% agree 👍 
Fixed here: [a3ebeea](https://github.com/safe-global/safe-wallet-monorepo/pull/7409/commits/a3ebeea4fbc8d3108de29bfae2b62a9650746561)

PR #7412, opened 2026-03-13
fix(web): resolve too much recursion in HexEncodedData regex

compojoom:
> In a maze of zeroes deep,
> the engine tried each branching leap—
> split the string, no forks to chase,
> linear time restores the pace.

## What it solves

The `HexEncodedData` component crashes with "too much recursion" when displaying long transaction data (e.g., 120+ bytes from Transaction Builder). The regex `/(.*?)(0{18,})/g` causes catastrophic backtracking because the lazy `(.*?)` quantifier creates exponential branching when processing strings with many scattered zeroes.

Resolves: https://linear.app/safe-global/issue/WA-1724/optimize-transaction-builder-handling-for-large-json-payloads

## How this PR fixes it

Replaces the backtracking-prone `String.replace` with `String.split(/(0{18,})/)`, which splits on zero runs in a single linear pass. The capture group keeps the zero-run delimiters in the result array, then a simple loop classifies each part as dimmed zeroes or normal text. Same visual output, O(n) instead of exponential.

## How to test it

1. Open the Transaction Builder app
2. Create a transaction to the zero address (`0x0000...0000`) with zero value and a long hex data payload (120+ bytes)
3. Submit the batch and view the transaction — it should render without crashing
4. Verify that long zero runs in the hex data appear dimmed/greyed out as before

## Screenshots

N/A - no UI changes (visual output is identical)

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/transactions/HexEncodedData/index.tsx
+++ b/apps/web/src/components/transactions/HexEncodedData/index.tsx
@@ -14,6 +14,7 @@ interface Props {
 }
 
 const FIRST_BYTES = 10
+const ZEROES_PATTERN = /^0+$/
 
 const SHOW_MORE = 'Show more'
 const SHOW_LESS = 'Show less'
@@ -41,7 +42,7 @@ export const HexEncodedData = ({ hexData, title, highlightFirstBytes = true, lim
   for (let i = 0; i < parts.length; i++) {
     const part = parts[i]
     if (!part) continue
-    if (/^0+$/.test(part) && part.length >= 18) {
+    if (ZEROES_PATTERN.test(part) && part.length >= 18) {
       dimmedZeroes.push(
         <span className={css.zeroes} key={i}>
           {part}


apps/web/src/components/transactions/HexEncodedData/index.tsx
```diff
@@ -37,18 +37,20 @@ export const HexEncodedData = ({ hexData, title, highlightFirstBytes = true, lim
...
-    )
-    return ''
-  })
-
-  dimmedZeroes.push(<Fragment key={index++}>{remainder}</Fragment>)
+  const parts = restBytes.split(/(0{18,})/)
+  for (let i = 0; i < parts.length; i++) {
+    const part = parts[i]
+    if (!part) continue
+    if (/^0+$/.test(part) && part.length >= 18) {
```
katspaugh: You could slightly improve it by saving the regex literal outside of the loop into a variable.

apps/web/src/components/transactions/HexEncodedData/index.tsx
```diff
@@ -37,18 +37,20 @@ export const HexEncodedData = ({ hexData, title, highlightFirstBytes = true, lim
...
+      dimmedZeroes.push(
+        <span className={css.zeroes} key={i}>
+          {part}
+        </span>,
```
katspaugh, line 49: Likewise for the JSX here and in the else.

PR #7421, opened 2026-03-13
feat(web): add Auth0 email-based login for Spaces

tmjssz:
## What it solves

Adds email-based authentication (via Auth0 Universal Login) as an alternative to SIWE wallet sign-in for Spaces. Users who don't have a connected wallet can now sign in with their email address.

Resolves: [WA-1709](https://linear.app/safe-global/issue/WA-1709/auth0-universal-login-integration)

## How this PR fixes it

Introduces a new `email-auth` feature. The `Auth0ProviderWrapper` is mounted unconditionally at the app level in `_app.tsx` (when env vars are present) to handle the redirect callback before any lazy-loaded feature mounts — the `EMAIL_AUTH` feature flag only controls UI visibility of the sign-in button. After Auth0 redirect, the `id_token` is exchanged with the CGW via `POST /v2/auth/verify` to obtain the same session cookie used by SIWE, so from that point on the session is identical.

### Key architectural decisions
- **Provider at app level, not lazy-loaded** — Auth0 redirect callback (`?code=&state=`) arrives before chain config loads, so the provider must be ready immediately
- **`cacheLocation: 'memory'`** — No persistent Auth0 state; CGW session cookie handles persistence
- **`useAuth0Login` uses `useContext` directly** — Safe to call without `Auth0Provider` (returns no-op defaults), avoiding conditional hook issues
- **Token exchange in a dedicated handler component** — `Auth0TokenExchangeHandler` sits inside the provider, using a ref guard to prevent double-exchange on concurrent renders

> **Note:** The CGW `POST /v2/auth/verify` endpoint is not implemented yet. The token exchange request will fail, and end-to-end login cannot be tested until the backend endpoint is available. The frontend flow up to the exchange (Auth0 redirect, `id_token` extraction, UI guards) can be verified independently.

## How to test it

1. Set `NEXT_PUBLIC_AUTH0_DOMAIN` and `NEXT_PUBLIC_AUTH0_CLIENT_ID` env vars to a valid Auth0 SPA tenant
2. Enable `EMAIL_AUTH` feature flag in chain config
3. Navigate to any Space — the "Sign in with email" button should appear alongside the existing SIWE button
4. Click the email button → Auth0 Universal Login opens → complete login → redirected back → token exchange will fail (CGW endpoint not yet implemented)
5. Verify the button is hidden when either the feature flag is off or Auth0 env vars are absent

## Screenshots

N/A — UI changes are a single button addition to the existing Spaces sign-in screen

## Checklist

---

## CLA signature


tmjssz:
@claude review

tmjssz:
@claude review again pls. I addressed your comments:

1. The CSRF concern is invalid — the SDK handles it. The redirect URI concern is valid in spirit but not exploitable given Auth0's server-side allowlist validation. The current code is safe.
2. Done -> 60265b836e4f583acaed48ef2e9402ea16ebf861
3. The Auth0 SDK must be available at page load to process the redirect callback (`?code=&state=`), so dynamic importing would break the flow. 
4. The imports are actually consistent, they're all named imports, just importing different exports from the same package
5. The existing ObservabilityErrorBoundary already wraps the entire content tree including Auth0ProviderWrapper
6. The exchangingRef already prevents double-execution, and since the effect runs synchronously before any re-render can interleave, there's no actual race condition here.
7. An integration/e2e test makes sense once the backend endpoint is deployed.
8. The mocks match exactly what the component consumes: a boolean from `useHasFeature` and a boolean from `isAuth0Configured`.

apps/web/src/features/email-auth/hooks/useAuth0TokenExchange.ts
```diff
@@ -0,0 +1,57 @@
...
+            groupKey: 'email-sign-in-failed',
+          }),
+        )
+      } finally {
+        exchangingRef.current = false
+      }
+    }
+
+    exchange()
+  }, [isAuth0Authed, isCgwAuthenticated, getIdTokenClaims, verifyAuth0, dispatch])
```
clovisdasilvaneto: I would add a ref here just to make sure that we don't have extra calls in case any of these dependencies changes

clovisdasilvaneto (approved):

0xAryanShah (approved):
👍

apps/web/src/features/email-auth/hooks/useAuth0TokenExchange.ts
tmjssz: Good idea! Done -> b8e2838d7026124a0c1d2714c475730b38059812

apps/web/src/features/email-auth/store/auth0Api.ts
```diff
@@ -0,0 +1,17 @@
+// This file is a placeholder for the Auth0 token exchange API.
+// TODO: replace with the auto-generated API handler from the @safe-global/store package when it is available.
+import { cgwClient } from '@safe-global/store/gateway/cgwClient'
+
+const auth0Api = cgwClient.injectEndpoints({
+  endpoints: (build) => ({
+    authVerifyV2: build.mutation<void, { id_token: string }>({
+      query: (body) => ({
+        url: '/v2/auth/verify',
```
LucieFaire: as we have decided to stick with v1, I have not added a new path.
Also CGW expects `access_token`, not `id_token` per Auth0 docs: 
> Do not use ID tokens to gain access to an API.
https://auth0.com/docs/secure/tokens

tmjssz: Fixed both points

LucieFaire:
closing in favor of https://github.com/safe-global/safe-wallet-monorepo/pull/7490

PR #7436, opened 2026-03-16
fix(web): no layout shift in Assets when Swap is disabled

vmosharova:
> Offset by the real button strip:
> no more cropped balance, no half‑visible Send.
> Top assets row stays whole.

## What it solves

When the Native Swap feature is off, the Top assets widget either cropped the balance/percentage on the right (fixed 80px `translateX`) or left the Send button partly visible and cropped (offset was too small). Both issues are fixed.

Resolves: [WA-1371](https://linear.app/safe-global/issue/WA-1371/asset-widget-without-native-swaps-looks-broken)

## How this PR fixes it

- Compute the visible action-button count (Send + optional Swap/Earn/Stake) and the corresponding width (28px per button, 8px gap).
- Pass the total offset (16px value-container gap + button strip width) as a CSS variable `--asset-buttons-offset` from the component.
- Use `transform: translateX(var(--asset-buttons-offset, 80px))` so the default shift matches the actual right-hand area. Balance stays visible and the full button strip is hidden (not cropped); on hover it slides in correctly.

## How to test it

1. Disable Native Swap (e.g. chain without `NATIVE_SWAPS` or temporary `return false` in `useIsSwapFeatureEnabled`).
2. Open the dashboard and check the Top assets widget: balance and % should be fully visible; no cropped Send button; no big empty gap on the right.
3. Hover a row: Send (and any other) buttons should appear fully.

## Screenshots

Before:
image

After:
image

## Checklist

---

## CLA signature


vmosharova:
The Page Screenshots (see above) were taken before the chain config (and with it Swap/Earn/Stake feature flags) is loaded (it is loaded async), hence the shift on the screenshot. Her is the correct one taken from the feature branch:
image

PR #7455, opened 2026-03-18
fix(web): clear stale ENS name when switching Safes

katspaugh:
> Debounce delayed the truth,
> stale names clung to new addresses—
> now we clear on switch.

Resolves https://linear.app/safe-global/issue/WA-1772/febug-ens-reverse-resolution-bug.

## Summary
- Fixed a bug where the resolved ENS name from the previous Safe would briefly display when switching to a different Safe
- Root cause: `useAddressResolver` uses a 200ms debounce on the address input, and during that window the old ENS result was returned alongside the new Safe address
- Added a staleness check that returns `undefined` for ENS while the debounced value hasn't caught up to the current address

## Test plan
- [x] Added unit test verifying ENS is cleared immediately when address changes
- [x] Existing `useAddressResolver` tests pass
- [x] Web workspace type-check passes

### DIFF
--- a/apps/web/src/hooks/useAddressResolver.ts
+++ b/apps/web/src/hooks/useAddressResolver.ts
@@ -21,11 +21,13 @@ export const useAddressResolver = (address?: string) => {
 
   const [ens, _, isResolving] = useAsync<string | undefined>(() => {
     if (!shouldResolve) return
+    // Wait for debounce to settle so we never resolve a stale address
+    if (debouncedValue !== address) return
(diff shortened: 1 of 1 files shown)


ChillYao:
https://github.com/user-attachments/assets/8132728b-22dc-4609-b630-3de18898ab81

Hi @katspaugh ,
I checked the preview link. And this bug is still there. (You can see from this short video)

katspaugh:
@ChillYao sorry, I should've tested more. It's working now.

ChillYao:
@katspaugh approved

PR #7470, opened 2026-03-18
fix(web/Spaces): fix issues in the Onboarding flow

vmosharova:
> Back goes where it should,
> Safe list no longer breaks the scene —
> scope wraps all three steps clean.

## What it solves

Resolves two bugs in the onboarding flow:
1. Back button on Step 1 (`/welcome/create-space?spaceId=`) was navigating to the Select Safes page instead of exiting the flow
2. Select Safes page rendered with broken UI (unstyled shadcn components) on every visit

## How this PR fixes it

- **Back button**: replaced `router.back()` with `router.push(AppRoutes.welcome.spaces)` in `CreateSpaceOnboarding`. `router.back()` was navigating to the previous history entry (Select Safes), since users arrive at Step 1 via `handleBack()` from Step 2.
- **Broken UI**: added the `shadcn-scope` wrapper to `SelectSafesOnboarding`, matching the pattern already used by Step 1 and Step 3. Without it, shadcn CSS custom properties were undefined, causing all shadcn components to render unstyled.

## How to test it

1. Start onboarding, fill in space name and go to Step 2 (Select Safes)
2. Click Back → should go to Step 2, click Back again → should return to Step 1
3. On Step 1, click Back → should navigate to `/welcome/spaces`, not to Select Safes
4. Navigate through all steps and verify Select Safes looks correctly styled in both light and dark mode

## Checklist

---

## CLA signature


ChillYao (approved):

apps/web/src/features/spaces/components/SelectSafesOnboarding/index.tsx
```diff
@@ -28,84 +31,86 @@ const SelectSafesOnboarding = (): ReactElement => {
...
+                onClick={handleBack}
+                className="rounded-md border border-card shadow-sm"
+              >
+                <ChevronLeft className="size-5" />
+              </Button>
 
-            <div className="flex items-center justify-center py-xs">
-              <StepIndicator currentStep={ONBOARDING_STEP} totalSteps={TOTAL_STEPS} />
-            </div>
+              <div className="flex items-center justify-center py-xs">
```
ChillYao, line 51: Seems nothing is changed here. Is this because of prettier?

vmosharova: Yes, that's formatting - because the code is wrapped in a new element, so the indentation had to be adjusted as well.

PR #7483, opened 2026-03-19
chore: add agentic feedback loop for faster quality checks

katspaugh:
> Parallel checks now run as one,
> scaffolded tests before you're done,
> a hook that never sleeps.

## Why?

AI agents working in this repo currently run type-check, lint, prettier, and tests sequentially — often forgetting one, running them in the wrong order, or skipping verification entirely. This wastes tokens and time, and lets broken code slip through.

These scripts give agents (and humans) a single command that runs all checks in parallel, scoped to only the files that changed. The Claude Code hook makes this automatic — every agent response that touches source files gets verified before moving on. The test scaffold ensures agents don't skip writing tests by giving them a correct starting point.

## Summary

- **`yarn verify:web`** — runs type-check, lint, prettier, and tests in parallel (one command instead of four)
- **`yarn verify:changed:web`** — incremental mode that only checks files in your current diff (seconds instead of minutes)
- **Missing test detection** — warns when changed source files lack corresponding tests
- **`yarn test:scaffold <file>`** — generates test skeletons with correct imports, mocks, and structure
- **Claude Code stop hook** — automatically runs `verify:changed` after every agent response that modifies source files
- **AGENTS.md updates** — test decision matrix and fast feedback loop documentation for agents

## Test plan

- [x] `yarn verify:web` runs all four checks in parallel on clean tree
- [x] `yarn verify:changed:web` detects changed files and scopes checks
- [x] `yarn verify:changed:web` shows "No changed files detected" on clean tree
- [x] Missing test detection warns about files without tests
- [x] `yarn test:scaffold` generates correct test skeletons for hooks
- [x] Claude Code hook runs silently on clean tree, shows compact output on changes
- [x] `--passWithNoTests` prevents false failures when no related tests exist
- [x] AGENTS.md has Fast Feedback Loop and Test Decision Matrix sections

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -247,7 +247,7 @@ Every code change must include tests. See [`apps/web/docs/TESTING.md`](apps/web/
 
 The repo provides automated verification:
 
-1. **Automatic**: A Claude Code `Stop` hook runs `verify:changed` once at the end of each agent turn. It early-exits (no-op) when no `.ts/.tsx/.js/.jsx` files have been modified. When it does run, type-check runs on the full project (TSC requires this), while lint, prettier, and tests are scoped to changed files only. The workspace (web/mobile) is auto-detected from the changed file paths. Fix any errors before moving on.
+1. **Automatic**: A Claude Code `Stop` hook runs `verify:changed` once at the end of each agent turn. It early-exits (no-op) when no `.ts/.tsx/.js/.jsx` files have been modified. When it does run, type-check runs on the full project (TSC requires this), while lint, prettier, and tests are scoped to changed files only. The workspace (web/mobile) is auto-detected from the changed file paths. Set `SKIP_VERIFY=1` to disable. Fix any errors before moving on.
 
 2. **Manual**: Run `yarn verify:changed:web` anytime to check your work. Run `yarn verify:web` for a full check before committing.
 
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -301,6 +318,8 @@ Every code change must include tests. See [`apps/web/docs/TESTING.md`](apps/web/
    > bundle shrinks, tests pass clean.
    ```
 
+9. **PR "Why?" section**: Every PR description must include a `## Why?` section right after the poem. Explain the motivation behind the change — what problem it solves, what user or developer pain it addresses, or what business need it serves. Focus on *why* this change is needed, not *what* it does (that's what the Summary section is for). A reviewer who reads only the "Why?" should understand whether this PR is worth reviewing at all.
```
compojoom: We should rather tell it to use the github PR template here instead of adding a why.

.claude/settings.json
```diff
@@ -1 +1,15 @@
...
+        "matcher": "",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node scripts/verify-changed-hook.mjs"
+          }
+        ]
+      }
+    ]
+  }
```
compojoom, line 14: Isn't this going to run on absolutely every code change that claude does? isn't it a bit excessive?

compojoom:
btw - you cannot run type-check on changed files. TSC needs the complete project, so your description is a bit misleading. Also this runs with --workspace=web hardcoded. I think that if someone works on a package or on the mobile app this will still run, no?

.claude/settings.json
katspaugh: The hook already addresses this — it's a Stop hook (runs once per turn, not per tool call) and early-exits when no source files changed.

katspaugh:
@claude review

PR #339, opened 2022-08-12
fix: Detect replacement tx when creating a safe

usame-algan:
## What it solves

Calling `provider.waitForTransaction` with the cached `txHash` will not detect if the transaction was cancelled or sped-up. Instead we now call `provider._waitForTransaction` with a `replacement` object and handle those cases.

## How to test

1. Create a new safe
2. Reload the page while the tx is executing
3. Cancel or speed-up the transaction in your wallet
4. Observe the Status changing to ERROR or SUCCESS

PR #7533, opened 2026-03-26
tests: add E2E regression tests for positions

francovenica:
What it solves
Adds Cypress regression tests for the Positions AI feature, covering the dashboard widget and the full Positions tab (/balances/positions).
How this PR fixes it

portfolio.cy.js: Regression spec with 7 test cases (TC #333, #334, #338, #339, #340 + two error-state tests) using real on-chain data from a Polygon static safe
portfolio.pages.js: Page-object model with all selectors, test-data constants, and helper functions — no bare cy.* calls in test bodies
Chain feature injection: beforeEach intercepts /v2/chains to add the POSITIONS flag and remove PORTFOLIO_ENDPOINT on Polygon, enabling the feature without touching source code
Real data: All tests run against matic:0xc1f4652866ddB3811adcd3418c13eF640e88E1f6 (AAVE V3 + Morpho positions)

New files

apps/web/cypress/e2e/regression/portfolio.cy.js — Regression spec
apps/web/cypress/e2e/pages/portfolio.pages.js — Page-object helpers

Edited files

apps/web/cypress/fixtures/safes/static.js — Added MATIC_STATIC_SAFE_33 (Positions Static safe)

How to test it

Run yarn workspace @safe-global/web cypress:open
Navigate to e2e/regression/portfolio.cy.js
Verify all 7 tests pass against the Polygon static safe with real positions data
Verify error-state tests correctly show/hide the widget and unavailable message on API failure

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/cypress/e2e/regression/portfolio.cy.js
+++ b/apps/web/cypress/e2e/regression/portfolio.cy.js
@@ -1,5 +1,6 @@
 import * as constants from '../../support/constants'
 import * as portfolio from '../pages/portfolio.pages'
+import * as main from '../pages/main.page'
 import { getSafes, CATEGORIES } from '../../support/safes/safesHandler.js'
 
 let staticSafes = []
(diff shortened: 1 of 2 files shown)


francovenica:
@claude review

apps/web/cypress/e2e/pages/portfolio.pages.js
```diff
@@ -0,0 +1,132 @@
...
+export function verifyPercentageShareVisible() {
+  cy.contains(/%/).should('be.visible')
+}
+
+export function verifyTotalPositionsTitleVisible() {
+  cy.contains(totalPositionsTitle).should('be.visible')
+}
+
+export function verifyFiatValueVisible() {
+  cy.contains(/\$ [\d,]+(\.\d+)?/).should('be.visible')
```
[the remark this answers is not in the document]
francovenica: This was updated. Left a comment in the function explaining why the new type of regex

apps/web/cypress/e2e/regression/portfolio.cy.js
```diff
@@ -0,0 +1,92 @@
...
+let staticSafes = []
+
+describe('Positions Tests', { defaultCommandTimeout: 60000, requestTimeout: 30000 }, () => {
+  before(async () => {
+    staticSafes = await getSafes(CATEGORIES.static)
+  })
+
+  beforeEach(() => {
+    // Intercept the chains list endpoint to inject the POSITIONS flag.
+    cy.intercept('GET', '**/v2/chains**', (req) => {
```
liliya-soroka: I would suggest to convert it to a more general function , move to the main page object and call this function here with feature flag  and chain for this tests. It will allow to re-use the same logic for other tests and makes before each cleaner 
something like 
beforeEach(() => {
  injectChainFeatureFlag(
    constants.networkKeys.polygon,
    constants.chainFeatures.positions,
    ['PORTFOLIO_ENDPOINT'],
  )
  cy.intercept('GET', constants.positionsEndpoint).as('getPositions')
})

apps/web/cypress/e2e/regression/portfolio.cy.js
```diff
@@ -0,0 +1,92 @@
...
+          res.body = { ...res.body, features }
+        }
+      })
+    })
+
+    cy.intercept('GET', constants.positionsEndpoint).as('getPositions')
+  })
+
+  // TC #333
+  it('[SMOKE] Verify that the Positions widget on the dashboard displays all protocols with name and fiat value', () => {
```
liliya-soroka: if it's smoke - should we move it in a smoke suite  , so it will be run on each PR . I think that smoke also contains some tests from Valentin

francovenica: I saw other test in the dashboard file, so I assume, even with the SMOKE tag that they were fine there.
I'll just remove the tag and leave them there.

apps/web/cypress/e2e/regression/portfolio.cy.js
francovenica: I followed claude suggestion and made it like that, created the InjectChainFeature function in the main.page file, that handles the responses well.
There was a EnableChainFeature function, but it wasn't able to do the work, and since it wasn't used anywhere else I just removed it. 
We can use the InjectChainFeature for other test in the future

PR #7545, opened 2026-03-26
feat(mobile): enable connect wallet app option on signer import screen

tmjssz:
> AppKit wired, wallets connect,
> signers stored with name and icon,
> the "coming soon" tag—gone.

## What it solves

The "Connect signer" option on the Import Signers screen was a non-functional "Coming soon" placeholder. Users had no way to connect an external wallet app (MetaMask, Phantom, Rabby, etc.) as a signer for their Safe.

Resolves: https://linear.app/safe-global/issue/WA-1672/enable-signer-choose-method-screen-enable-connect-wallet-app-option

## How this PR fixes it

Integrates Reown AppKit (WalletConnect v2) into the mobile app so users can connect any supported wallet app as a signer. When a wallet connects, the address is automatically registered as a `walletconnect` type signer in the store (with wallet name/icon metadata). Phantom and Rabby are added as custom wallets with local base64-encoded icons to avoid broken `image_url` on React Native.

Key decisions:
- Wallet icons are stored as base64 data URIs in `assets/wallet-icons/` because AppKit's `image_url` field doesn't reliably load remote URLs in RN
- The `Signer` union type was extended with a `walletconnect` variant to persist wallet metadata
- A `useRef` guard (`registeredRef`) prevents duplicate signer registrations during re-renders

## How to test it

1. Set `EXPO_PUBLIC_REOWN_PROJECT_ID` in `.env.local` to a valid Reown project ID
2. Navigate to **Account > Signers > Import signer**
3. Tap **Connect wallet app** — the AppKit modal should open showing featured wallets
4. Connect a wallet (e.g. MetaMask) and verify the signer appears in the signers list

## Screenshots

simulator_screenshot_4D49A049-285E-4D55-A9F7-E49723527CBA
simulator_screenshot_342B3BDE-070D-4982-AC66-B2B8565BDF0D
simulator_screenshot_F749847F-DFA9-449E-ADCE-B2D15AC93F17
simulator_screenshot_3C4D7709-3F3A-4BDE-9D3E-6080AD82DF65

## Checklist

---

## CLA signature


apps/mobile/assets/wallet-icons/phantom-icon-data-uri.ts
compojoom, line 1: Can't we have svgs? Why do we need a data uri here? This is  blowing the bundle size and the more we do this the slower the startup time of the app is going to be.

apps/mobile/src/config/appKit.ts
```diff
@@ -0,0 +1,89 @@
...
+    return entries.map(([key, value]) => [key, value ? JSON.parse(value) : undefined])
+  },
+  getItem: async (key: string) => {
+    const value = await AsyncStorage.getItem(key)
+
+    return value ? JSON.parse(value) : undefined
+  },
+  setItem: (key: string, value: unknown) => AsyncStorage.setItem(key, JSON.stringify(value)),
+  removeItem: (key: string) => AsyncStorage.removeItem(key),
+}
```
compojoom, line 42: This storage object is now independent of our own storage, but more or less serves the same purpose as store/storage.ts ? 

I guess we could spare ourselves the async-storage dependency if we proxy  to MMKV?

apps/mobile/src/config/appKit.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export const appKit = createAppKit({
+  projectId,
+  networks: [mainnet, polygon, arbitrum, sepolia],
```
compojoom, line 46: What do we do with those networks here? Should this be a dynamic array with networks, or should it contain all the networks we support?

apps/mobile/src/features/ImportSigners/ImportSigners.container.tsx
```diff
@@ -1,55 +1,90 @@
...
+        onPress: () => router.push('/import-signers/hardware-devices'),
+      },
+      {
+        name: 'connectSigner',
+        title: 'Connect wallet app',
+        description: 'Connect an external wallet app to sign transactions.',
+        icon: <SafeFontIcon name="dapp-logo" size={16} />,
+        Image: ConnectWalletApp,
+        imageProps: { marginBottom: -32 },
+        onPress: handleConnectSigner,
```
compojoom: I guess you added it here because of the custom onPress. 

We could have thsi as a static array outside of the component and each item could have a key and we always call the same handleConnectSigner and pass the key. Then in the function we decide what logic to follow?

apps/mobile/src/features/ImportSigners/ImportSigners.container.tsx
```diff
@@ -1,55 +1,90 @@
...
-const title = 'Import a signer'
+const title = 'Add signer'
 
 export const ImportSignersContainer = () => {
   const { isBiometricsEnabled } = useBiometrics()
+  const dispatch = useAppDispatch()
+  const { open } = useAppKit()
+  const { address, isConnected } = useAccount()
+  const { walletInfo } = useWalletInfo()
+  const existingSigner = useAppSelector((state) => (address ? selectSignerByAddress(state, address) : undefined))
```
compojoom: why do we need to know the signer here?

apps/mobile/app.config.ts
```diff
@@ -142,7 +159,7 @@ const config: ExpoConfig = {
           forceStaticLinking: ['RNFBApp'],
         },
         android: {
-          minSdkVersion: 34,
```
compojoom, line 145: No, this should stay at 34

apps/mobile/src/app/_layout.tsx
```diff
@@ -142,32 +144,35 @@ function RootLayout() {
...
-                        </SafeToastProvider>
-                      </BottomSheetModalProvider>
-                    </SafeThemeProvider>
-                  </PersistGate>
-                </PortalProvider>
-              </NotificationsProvider>
-            </DataFetchProvider>
-          </Provider>
-        </KeyboardProvider>
+        <AppKitProvider instance={appKit}>
```
compojoom: Probably we would need to have this lower in the tree? Maybe below the store provider? This way we would know what safes the user has, probably have network data as well and could do a better initialisation of appkit?

apps/mobile/assets/wallet-icons/phantom-icon-data-uri.ts
tmjssz: Yeah, I agree that using base64 here is not a good solution. The problem is that the app kit only accepts a URL otherwise, using SVGs is not possible. We need to get the static HTTPS URL of the icon. So we need to host the image file somehow. Claude suggests to just use the GitHub raw URL. WDYT?

compojoom: I see in this case can't we have it as an asset on apps.safe.global?

apps/mobile/src/config/appKit.ts
tmjssz: Yes it should be all networks we support. Still need to update. I guess the mobile app also uses the config service as source of truth, right?

apps/mobile/src/config/appKit.ts
tmjssz: Yes good point. I refactored to use the example from [here](https://docs.reown.com/appkit/react-native/core/options#storage) -> ff697b67da4cd3b50a2047547b75f83efca6c2c1

apps/mobile/src/features/ImportSigners/ImportSigners.container.tsx
tmjssz: Good idea! Done -> a419efc646b4c8bc96db3f3dfda409aa28f4fcdc

apps/mobile/src/features/ImportSigners/ImportSigners.container.tsx
tmjssz: Not needed actually. Removed -> 8b5048bfa255c58d005b97e1ea1486655f3b3fa0

apps/mobile/app.config.ts
tmjssz: Oops! reverted -> c94700fd98152c038abfde5d6e18610595a84a88

apps/mobile/src/app/_layout.tsx
tmjssz: Done -> a48415e839574d3195f83c853244c66410df13fe

apps/mobile/src/config/appKit.ts
tmjssz: I realized we can't get rid of the dependency unfortunately because it's [needed by the appkit](https://docs.reown.com/appkit/react-native/core/installation#core-library).

tmjssz: Reverted the change above

apps/mobile/assets/wallet-icons/phantom-icon-data-uri.ts
tmjssz: Good idea! Done -> c04a533958e597d1f2488b122a56c4d248173a11

apps/mobile/src/config/appKit.ts
compojoom: yes.

compojoom: Once the store is hydrated we always dispatch a request to fetch the latest config

PR #7560, opened 2026-03-30
feat: Add new logout w/ redirect hook

LucieFaire:
## What it solves

Resolves: [WA-1855](https://linear.app/safe-global/issue/WA-1855/add-new-logout-logic-to-wallet-web)
Implements logout with OIDC provider redirect support. Previously, logging out only cleared local state — OIDC users remained signed in at the identity provider, meaning the next sign-in would auto-authenticate without prompting for credentials.

### Logout flow (`useLogout` + `useLogoutCallback`)
- `useLogout` creates and submits a hidden `<form>` that POSTs to CGW's `/v1/auth/logout/redirect` with a `redirect_url` back to the app. This triggers a 303 redirect chain through the IdP, clearing session cookies at both CGW and the OIDC provider.
- Before submitting, a `logging_out` flag is set in `sessionStorage` so the app knows a logout redirect is in-flight.
- `useLogoutCallback` (mounted in `InitApp`) detects the flag on page load and calls `reconcileAuth` to sync Redux state with the backend.

### Login callback improvement (`useOidcLoginCallback`)
- Updated to use a shared `reconcileAuth` helper instead of inline logic.
- Dispatches `setIsOidcLoginPending` for loading state, and shows error notifications on failure.

### Shared `reconcileAuth` utility
- Calls `/v1/auth/me` and returns `'authenticated'`, `'unauthenticated'`, or `'error'` (for transient failures like 5xx, which leave auth state unchanged).
- Used by both login and logout callback hooks.

## How to test it

1. Sign in via OIDC (email/Google)
2. Click logout — browser should navigate away briefly (IdP redirect) then return to the spaces welcome page
3. Verify you're signed out (no authenticated state)
4. Sign in again — should require full authentication (not auto-sign-in)

## Visual summary
```mermaid
sequenceDiagram
    participant User
    participant App as App (Browser)
    participant CGW as CGW Backend
    participant IdP as OIDC Provider

    Note over User,IdP: Logout Flow
    User->>App: Click logout
    App->>App: sessionStorage.set("logging_out")
    App->>CGW: POST /v1/auth/logout/redirect<br/>(hidden form submit)
    CGW-->>App: 303 → IdP logout URL
    App->>IdP: Follow redirect (clears IdP session)
    IdP-->>App: Redirect back to app (redirect_url)
    
    Note over App: useLogoutCallback fires
    App->>CGW: GET /v1/auth/me
    CGW-->>App: 403 (cookie cleared)
    App->>App: dispatch(setUnauthenticated)<br/>sessionStorage.remove("logging_out")
```

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/hooks/useLogoutCallback.ts
+++ b/apps/web/src/hooks/useLogoutCallback.ts
@@ -25,12 +25,13 @@ export const useLogoutCallback = () => {
 
     const process = async () => {
       const result = await reconcileAuth(dispatch)
+      const logoutNotUnauthenticated = result !== 'unauthenticated'
 
+      if (logoutNotUnauthenticated) {
+        logError(Errors._109)
+      }
       if (result === 'error') {
-        logError(Errors._109)
         dispatch(setUnauthenticated())
-      } else if (result === 'authenticated') {
-        logError(Errors._109)
       }
 
       sessionStorage.removeItem(LOGGING_OUT_KEY)


0xAryanShah:
@claude review

LucieFaire:
I have looked at Claude's remarks, and apart from `error` handling, other remarks dont seem to be valid

LucieFaire:
Loogout works now, @0xAryanShah please re-review

apps/web/src/hooks/useLogoutCallback.ts
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useRef } from 'react'
+import { useAppDispatch } from '@/store'
+import { setUnauthenticated } from '@/store/authSlice'
+import reconcileAuth from '@/store/reconcileAuth'
+import { logError, Errors } from '@/services/exceptions'
+
+export const LOGGING_OUT_KEY = 'logging_out'
+
+/**
```
0xAryanShah, line 9: Nit: JSDoc comment can be updated after the latest commit

apps/web/src/hooks/useLogout.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const input = document.createElement('input')
+    input.type = 'hidden'
+    input.name = 'redirect_url'
+    input.value = redirectUrl
+    form.appendChild(input)
+
+    document.body.appendChild(form)
+    form.submit()
+    document.body.removeChild(form)
```
vmosharova, line 38: In the PR description I see the following: 
> Sign in again — should require full authentication (not auto-sign-in)

Meanwhile I didn't get the Google sign-in window again after logout - the auto sign-in happened instead. It might be the case that the form was removed too fast before the navigation `form.submit()` completed, so `POST` in the form submission (and session termination) failed before it reached CGW. The cookie was cleared though.

apps/web/src/hooks/useLogoutCallback.ts
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useRef } from 'react'
+import { useAppDispatch } from '@/store'
+import { setUnauthenticated } from '@/store/authSlice'
+import reconcileAuth from '@/store/reconcileAuth'
+import { logError, Errors } from '@/services/exceptions'
+
+export const LOGGING_OUT_KEY = 'logging_out'
```
0xAryanShah, line 7: Nit: can we put these constants: `LOGGING_OUT_KEY`
`LOGOUT_REDIRECT_PATH` in the constants/shared file?

apps/web/src/hooks/useLogoutCallback.ts
```diff
@@ -0,0 +1,41 @@
...
+  const dispatch = useAppDispatch()
+  const hasProcessed = useRef(false)
+
+  useEffect(() => {
+    const isLoggingOut = sessionStorage.getItem(LOGGING_OUT_KEY)
+    if (!isLoggingOut || hasProcessed.current) return
+    hasProcessed.current = true
+
+    const process = async () => {
+      const result = await reconcileAuth(dispatch)
```
vmosharova, line 27: Nit for a better readability:

```
const result = await reconcileAuth(dispatch)

const logoutNotUnauthenticated = result !== 'unauthenticated'

if (logoutNotUnauthenticated) {
  logError(Errors._109)
}

if (result === 'error') {
  dispatch(setUnauthenticated())
}
```

apps/web/src/hooks/useLogoutCallback.ts
LucieFaire: `LOGOUT_REDIRECT_PATH` is not used anywhere else, so I would keep it privately in `useLogout`

LucieFaire: and logging out key is tied closely here, imho extracting it doesnt provide much value at this stage

apps/web/src/hooks/useLogoutCallback.ts
LucieFaire: not sure what you mean here

apps/web/src/hooks/useLogout.ts
LucieFaire: yes, that is the case with SSO like Google. There are 3 layers of possible log out: app level, auth0 level and identity provider (like Google). We only log out from app and auth0 levels at this stage. We can adjust it in the future if we want to log out fro everywhere

apps/web/src/hooks/useLogoutCallback.ts
0xAryanShah: Nit: the JSDoc comment can be appended with the new error path added.
`- error → transient failure, force unauthenticated (require re-login)`

PR #7569, opened 2026-04-02
feat(mobile): sign transactions with WalletConnect signers

tmjssz:
> Wallet apps now sign the tx,
> gate checks chain before confirm,
> one context serves them all.

## What it solves

WalletConnect signers could be imported and reconnected but couldn't actually sign transactions. Users selecting a WC signer on the transaction review screen had no way to confirm — the signing flow only supported private-key and Ledger signers.

Resolves: https://linear.app/safe-global/issue/WA-1853/sign-transaction-with-connected-wallet-app

## How this PR fixes it

Adds end-to-end transaction signing for WalletConnect signers. When a user selects a WC signer and presses confirm, the app builds EIP-712 typed data via `generateTypedData` from `@safe-global/protocol-kit`, sends `eth_signTypedData_v4` to the connected wallet, and submits the resulting signature to the CGW.

Key pieces:

- **Signing service** (`walletconnect-signing.service.ts`) — pure function that builds typed data from a `SafeTransaction` and sends the signing request via the WC provider. Extracted as a service for testability
- **WalletConnectGate** — wraps the confirm button in `ReviewFooter`. Blocks signing when the wallet is disconnected (shows "Reconnect wallet") or on the wrong chain (shows "Switch network"). Only renders the confirm button when the WC session is active and on the correct network
- **`useTransactionSigning` extended** — the existing hook now handles `walletconnect` signer type alongside `private-key` and `ledger`, delegating to `signWithWc()` from the context
- **Signer badge in ChangeSignerSheet** — each signer row shows a `WalletConnectBadge` with connection status instead of the previous checkmark

To support this, `useWalletConnect` was refactored into focused hooks (`useImportSignerFlow`, `useReconnectFlow`, `useSwitchNetwork`, `useWalletConnectSigning`) behind a `WalletConnectProvider` context. This ensures AppKit hooks are mounted once at the root layout and shared via context rather than re-instantiated per consumer.

## How to test it

1. Go to Settings > Signers and connect a WalletConnect signer — verify the import flow works end-to-end
2. Navigate to a pending transaction and select the WC signer — verify the `WalletConnectGate` shows reconnect/switch-network prompts if the wallet is disconnected or on the wrong chain
3. With the WC wallet connected and on the correct chain, press confirm — verify the wallet receives the signing request and the signature is submitted to the CGW
4. After signing, try reconnecting with a different wallet address — verify the "Wrong wallet connected" error screen appears
5. Run tests: `yarn workspace @safe-global/mobile test --no-coverage`

## Screenshots

N/A - will add after manual testing on device

## Visual summary

```mermaid
flowchart TB
  User[User taps Confirm] --> Gate{WalletConnectGate}
  Gate -->|No session| Reconnect[Reconnect wallet button]
  Gate -->|Wrong chain| Switch[Switch network button]
  Gate -->|Ready| Sign[useTransactionSigning]
  Sign -->|WC signer| SVC[walletconnect-signing.service]
  SVC --> TypedData[generateTypedData via protocol-kit]
  TypedData --> WCProvider[eth_signTypedData_v4 via WC provider]
  WCProvider --> Submit[Submit signature to CGW]
  Sign -->|Private key| PK[Sign locally]
  Sign -->|Ledger| LDG[Sign via Ledger]
```

## Checklist

---

## CLA signature


tmjssz:
@claude review

tmjssz:
@claude review again pls. I addressed your previously found issues:

1. Removed stale @ts-ignore (auto-generated types already had signature) -> 8f5a841fe58e379141368060dc68be2d9721ddd6
2. Added validation after EIP712Domain destructuring + test -> 009b159f5fef11beb4d9ffd8ba72d58d1d0e89b5
3. Wrapped disconnect/open with useCallback for referential stability + fixed tests -> cc5442d88c73570c7411ba7e4eb240a2a7b83530

### Architecture
1. Skipped: SigningResult union type adds complexity without solving an actual problem; existing throw/catch pattern is consistent across all signing flows
2. Skipped: WalletConnectGate already makes only two hook calls in a 54-line component; extracting a wrapper hook would be premature abstraction

tmjssz:
@claude review again pls. Regarding your previously found issues:

1. Type safety safeVersion — Already validated in useWalletConnectSigning
2. Stale closure in WalletConnectGate — switchNetworkIfNeeded is already a useCallback with correct dependencies in useSwitchNetwork, and the gate passes it directly without wrapping — no stale closure possible

apps/mobile/src/components/SelectSigner/SelectSigner.tsx
```diff
@@ -17,27 +16,32 @@ type Props = {
...
+        onPress={() => {
+          if (disabled) {
+            return
+          }
+          router.push({ pathname: '/change-signer-sheet', params: { txId, actionType: ActionType.SIGN } })
+        }}
+        alignItems="center"
+        gap="$2"
+        backgroundColor="$backgroundSkeleton"
+        borderRadius={32}
```
compojoom: don#t we have a $8 or so for this?

apps/mobile/src/features/ConfirmTx/components/ReviewAndConfirm/ReviewFooter.tsx
```diff
@@ -31,9 +32,11 @@ export function ReviewFooter({ txId, activeSigner, isSigningLoading, onConfirmPr
     >
       <SelectSigner address={activeSigner?.value as Address} txId={txId} disabled={buttonDisabled} />
 
-      <SafeButton onPress={onConfirmPress} disabled={buttonDisabled} loading={isSigningLoading} height={44}>
```
compojoom, line 34: you've removed the height? Are you sure? I think that I had it becuase there was a slight jump when the button was in loading state

compojoom (approved):
👍

apps/mobile/src/features/ConfirmTx/components/ReviewAndConfirm/ReviewFooter.tsx
tmjssz: This fixes the cut off button text bug:
Screenshot 2026-04-08 at 15 57 14

tmjssz: Looks good for the loading state:

https://github.com/user-attachments/assets/eca16acd-a4bc-4b26-ab4c-9879daae6005

PR #7589, opened 2026-04-08
fix: trust CGW recommended version for chains not in safe-deployments

Zalens:
> When deployments don't know your chain,
> the fallback dragged you back to 1.3.
> Now CGW speaks, the code believes,
> and 1.4.1 is what you'll receive.

## What it solves

Resolves: Safe creation on new chains (e.g. Fluent) defaults to version 1.3.0 instead of the CGW-configured 1.4.1, causing "SDK not initialized" errors because the chain-agnostic SDK init path only covers `>=1.4.1`.

## How this PR fixes it

`getLatestSafeVersion()` used `getSafeSingletonDeployment({ network: chainId })` to determine the latest deployed version. For chains not registered in the `safe-deployments` npm package, this returned `undefined`, falling back to a hardcoded `1.3.0`.

Now, when the per-chain lookup returns nothing, we trust the CGW's `recommendedMasterCopyVersion` directly instead of falling back to 1.3.0. For registered chains, behavior is unchanged — we still cap at the latest deployed version.

## How to test it

1. Open a Safe on a chain not in `safe-deployments` (e.g. Fluent, chain 25363) on staging
2. Create a new Safe — it should be created with version 1.4.1 (not 1.3.0)
3. The SDK should initialize successfully without "SDK not initialized" errors
4. Verify existing chains (Ethereum, Polygon, etc.) still create Safes with the correct version

## Visual summary

```mermaid
flowchart TB
    A[getLatestSafeVersion] --> B{safe-deployments\nhas chain?}
    B -->|Yes| C[Cap at deployed version\nvs recommended]
    B -->|No - Before| D[❌ Fallback to 1.3.0]
    B -->|No - After| E[✅ Trust CGW\nrecommendedMasterCopyVersion]
    D --> F[Safe created with 1.3.0\nSDK init fails]
    E --> G[Safe created with 1.4.1\nChain-agnostic path works]
```

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/utils/__tests__/chains.test.ts
+++ b/apps/web/src/utils/__tests__/chains.test.ts
@@ -144,6 +144,14 @@ describe('chains', () => {
           ),
         ).toEqual('1.4.1')
       })
+
+      it('should trust recommendedMasterCopyVersion for chains not in safe-deployments', () => {
+        expect(
+          getLatestSafeVersion(
+            chainBuilder().with({ chainId: '25363', recommendedMasterCopyVersion: '1.4.1' }).build(),
+          ),
+        ).toEqual('1.4.1')
+      })
     })
   })
 })
--- a/apps/web/src/utils/__tests__/chains.test.ts
+++ b/apps/web/src/utils/__tests__/chains.test.ts
@@ -10,6 +10,7 @@ import { CONFIG_SERVICE_CHAINS } from '@/tests/mocks/chains'
 import { chainBuilder } from '@/tests/builders/chains'
 import { getChainConfig } from '@/utils/chains'
 import { makeStore, setStoreInstance } from '@/store'
+import * as safeDeployments from '@safe-global/safe-deployments'
 
 describe('chains', () => {
   beforeAll(() => {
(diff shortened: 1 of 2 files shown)


apps/web/src/utils/__tests__/chains.test.ts
```diff
@@ -144,6 +144,14 @@ describe('chains', () => {
...
         ).toEqual('1.4.1')
       })
+
+      it('should trust recommendedMasterCopyVersion for chains not in safe-deployments', () => {
+        expect(
+          getLatestSafeVersion(
+            chainBuilder().with({ chainId: '25363', recommendedMasterCopyVersion: '1.4.1' }).build(),
+          ),
+        ).toEqual('1.4.1')
+      })
```
dasanra, line 157: This test will be useless once safe-deployments is updated with Fluent on it. It's better to mock the getSafeSingletonDeployment to always return undefined in order to consistently validate the intended behavior.

Also it would be nice to also have a test when `recommendedMasterCopyVersion` is set to null. To track the current expected behavior.

Zalens: I pushed a fix.

dasanra (approved):

PR #7594, opened 2026-04-08
feat: use backend as source of truth for counterfactual safes

katspaugh:
## What it solves

Counterfactual (undeployed) safe data was stored locally in Redux/localStorage and synced to the backend as a fire-and-forget side effect. This meant:
- Backend errors (e.g. 500) during creation were silently swallowed — the user saw a success screen even when the backend failed
- Space members couldn't see each other's counterfactual safes
- Loading a CF safe by URL raced with the lazy-loaded sync hook, causing a flash of "Safe couldn't be loaded"

Depends on backend PR: https://github.com/safe-global/safe-client-gateway/pull/3011

Linear: https://linear.app/safe-global/project/migrate-counterfactual-safes-to-cgw-cdbddc664cb3/overview

### Creation flow
- **ReviewStep** now `await`s the backend POST before dispatching `addUndeployedSafe` to Redux. On failure, the user sees an error and can retry.
- The fire-and-forget `addUndeployedSafe` listener in `counterfactualSyncListener` is removed. The `removeUndeployedSafe` → DELETE listener remains (now using `.unwrap()`).

### Data loading
- `useCounterfactualSafeSync` fetches from both `/v1/users/counterfactual-safes` and `/v1/spaces/:spaceId/counterfactual-safes` on auth, so all space members see CF safes.
- The hook moved from lazy-loaded `LazyCounterfactual` to `InitApp` in `_app.tsx` to run eagerly on auth.
- A `cfSafeSynced` flag in `authSlice` (reset on each hydration) lets `useLoadSafeInfo` show a loading state instead of an error while sync is pending.
- `useLoadSafeInfo` suppresses CGW 404 errors when an undeployed safe exists in Redux or while CF sync is in progress.

### Spaces Accounts widget
- `useSpaceAccountsData` falls back to CF safe `threshold`/`owners` from Redux when no `SafeOverview` is available.

### Other
- `counterfactualSafeMapper.fromBackendDto` uses `??` instead of `||` for `paymentToken`.
- RTK Query dispatches use `.unwrap()` for proper error handling.
- Services barrel re-exports `counterfactualSafeMapper`.

## How to test it

1. **Create CF safe** → should see 201 + success screen
2. **Break the backend** (revert the transformer fix in CGW) → creation should show error, retry should work
3. **Navigate to CF safe URL directly** → should load without error flash
4. **Spaces Accounts widget** → CF safe should show owners/threshold
5. **Different space member** → should see CF safes created by another member

## Visual summary

```mermaid
sequenceDiagram
    participant User
    participant ReviewStep
    participant CGW Backend
    participant Redux
    participant UI

    User->>ReviewStep: Click "Create account"
    ReviewStep->>CGW Backend: POST /v1/users/counterfactual-safes
    alt Success (201)
        CGW Backend-->>ReviewStep: OK
        ReviewStep->>Redux: addUndeployedSafe
        ReviewStep->>UI: Navigate + success screen
    else Error (500)
        CGW Backend-->>ReviewStep: Error
        ReviewStep->>UI: Show error, user can retry
    end

    Note over Redux,UI: On app load (auth)
    Redux->>CGW Backend: GET /v1/users/counterfactual-safes
    Redux->>CGW Backend: GET /v1/spaces/:id/counterfactual-safes
    CGW Backend-->>Redux: Merge into undeployedSafes
    Redux-->>UI: All consumers reactively update
```

## Checklist

---

## CLA signature


katspaugh:
@claude review

clovisdasilvaneto:
points we found: 

- Show counter factual safe information on the onboarding
- Add a checker for the owner in the delete request trigger
- On each page load, check if the safe is counter factual, if it is and the safe has safeInfo then send a delete request to delete the safe. 
- Do the same request which we are doing on the safe creation flow, in the network selector dropdown (when adding the safe on a new chain)

clovisdasilvaneto (approved):

clovisdasilvaneto (dismissed):
We just did some round of tests and found some issues when the counterfactual safe is part of 2 or more spaces:

- Another user can create a counter factual safe even though he is not the owner
- If the owner has the activated safe on his space, and another user adds this safe on polygon, on the owner space, it will show that the activated safe now, has also polygon, even though no one on his space did that action.

katspaugh:
## PR Briefing

[Open the PR briefing](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-wallet-monorepo%2Fpr-7594%2Fe59019686b92%2Fbundle.json)

▶ **[Open the interactive review](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-wallet-monorepo%2Fpr-7594%2Fe59019686b92%2Fbundle.json)**
▶ **[Watch the walkthrough video](https://github.com/katspaugh/popcorn-review-artifacts/releases/download/prbrief-safe-global-safe-wallet-monorepo-pr-7594-e59019686b92/video-github.mp4)** (direct MP4)

This PR makes the CGW backend the authoritative source of truth for counterfactual (undeployed) safes, replacing the previous fire-and-forget Redux/localStorage approach. The `ReviewStep` and `CreateSafeOnNewChain` components now await a new `persistCounterfactualSafe` service that POSTs to `/v1/users/counterfactual-safes` (and optionally `/v1/spaces/:spaceId/counterfactual-safes`) before updating Redux, surfacing backend errors to the user instead of silently swallowing them. A new `useCounterfactualSafeSync` hook in `_app.tsx` eagerly fetches CF safes from both user and space endpoints on auth, and a `cfSafeSynced` flag in `authSlice` lets `useLoadSafeInfo` show a loading state instead of a 404 error while sync is pending. The `counterfactualSyncListener` middleware is refactored to only fire DELETE calls for safes where `isCreator` is true, and the Spaces Accounts widget (`useSpaceAccountsData`) falls back to Redux CF safe config for owners/threshold when no `SafeOverview` is available.

<details>

## Summary

This PR makes the CGW backend the authoritative source of truth for counterfactual (undeployed) safes, replacing the previous fire-and-forget Redux/localStorage approach. The `ReviewStep` and `CreateSafeOnNewChain` components now await a new `persistCounterfactualSafe` service that POSTs to `/v1/users/counterfactual-safes` (and optionally `/v1/spaces/:spaceId/counterfactual-safes`) before updating Redux, surfacing backend errors to the user instead of silently swallowing them. A new `useCounterfactualSafeSync` hook in `_app.tsx` eagerly fetches CF safes from both user and space endpoints on auth, and a `cfSafeSynced` flag in `authSlice` lets `useLoadSafeInfo` show a loading state instead of a 404 error while sync is pending. The `counterfactualSyncListener` middleware is refactored to only fire DELETE calls for safes where `isCreator` is true, and the Spaces Accounts widget (`useSpaceAccountsData`) falls back to Redux CF safe config for owners/threshold when no `SafeOverview` is available.

## Architecture notes

The change touches four major subsystems: (1) the creation flow (`ReviewStep`, `persistCounterfactualSafe`, `PayNowPayLater`) where backend writes are now blocking and error-surfacing; (2) the sync/load flow (`useCounterfactualSafeSync` in `InitApp`, `authSlice.cfSafeSynced`, `useLoadSafeInfo`) which replaces lazy sync with eager fetch and a loading gate; (3) the Redux listener (`counterfactualSyncListener`) which now guards DELETE on `isCreator` and uses `.unwrap()`; and (4) the RTK Query layer (`counterfactual-safes.ts`, `spaces.ts`) auto-generated from an updated CGW OpenAPI schema that adds `/v1/users/counterfactual-safes` and `/v1/spaces/:spaceId/counterfactual-safes` endpoints and removes `ContractAddresses` from the chain response. The `UndeployedSafe` type gains an optional `isCreator` field (undefined treated as true for backwards compatibility), and `addUndeployedSafe` propagates this flag through Redux state.

## Ripple risks

Analyzed 11 changed exports. No untouched external callers found — all usages appear to be inside the PR diff.

_Analyzed 45 file(s), 35 changed export(s), 0 uncovered callsite(s) in 2.7s._

## Walkthrough outline

_20 steps across 6 chapters, approximately 4 min 25s._

### Why this change

1. **motivation** · `apps/web/src/components/new-safe/create/steps/ReviewStep/index.tsx` L1-30 `ReviewStep`
   Before this PR, creating a counterfactual safe fired a backend POST as a side effect and immediately showed the user a success screen, even if the server returned a 500. Space members also couldn't see each other's undeployed safes, and navigating directly to a CF safe URL would flash a 'not found' error while the lazy sync hook caught up. This PR fixes all three by making the backend the authoritative source of truth.

2. **motivation** · `apps/web/src/components/new-safe/create/steps/ReviewStep/index.tsx` L1-30
   Here is the high-level flow after the change. Creation now blocks on the backend POST, sync runs eagerly at app init rather than lazily, and a new flag in the auth slice gates the loading state until sync completes.

### The new creation flow

3. **change** · `apps/web/src/features/counterfactual/services/persistCounterfactualSafe.ts` L23-70 `persistCounterfactualSafe`
   The new persist function is the heart of the creation change. It POSTs to the user counterfactual-safes endpoint, optionally POSTs to the space endpoint if a space ID is present, and only then calls the replay helper to update Redux. It returns a typed result union so callers can branch on success or failure without catching exceptions.

4. **change** · `apps/web/src/features/counterfactual/services/counterfactualSafeMapper.ts` L9-95 `toBackendDto`
   The mapper file provides the serialization layer between the frontend's nested safe props and the flat backend DTO. The to-backend function converts a replayed safe's config into the wire format, and returns null for predicted-safe props since those can't be serialized this way. The from-backend function reconstructs the full props object on the way back in.

5. **change** · `apps/web/src/components/new-safe/create/steps/ReviewStep/index.tsx` L80-140 `ReviewStep`
   Back in ReviewStep, the counterfactual path now awaits persist counterfactual safe and checks the result. If the result is not OK, it surfaces the error so the user can retry. Only on success does it dispatch add undeployed safe and navigate to the success screen.

6. **change** · `apps/web/src/features/counterfactual/components/PayNowPayLater/index.tsx` L1-60 `PayNowPayLater`
   PayNowPayLater gains an is-user-authenticated prop. When the user is not signed in and picks Pay Later, the component triggers a SIWE sign-in inline before setting the pay method, showing a spinner during the flow. This gates the counterfactual path behind authentication so the backend can associate the safe with an account.

7. **change** · `packages/store/src/gateway/AUTO_GENERATED/counterfactual-safes.ts` L1-100 `CounterfactualSafesCreateV1ApiResponse`
   This auto-generated file is the RTK Query layer for the three new user-level endpoints: POST, GET, and DELETE against slash v1 slash users slash counterfactual-safes. The types here, like create counterfactual safes DTO and get counterfactual safes response, are what the mapper and persist function work against.

### Eager sync and the loading gate

8. **change** · `apps/web/src/features/counterfactual/hooks/useCounterfactualSafeSync.ts` L1-60 `useCounterfactualSafeSync`
   The new sync hook fetches from both the user and space counterfactual-safes endpoints on auth, merges the results so the user endpoint wins on collision, and dispatches add undeployed safe for anything not already in Redux. User-owned safes get is-creator true; safes that only appear in the space response get is-creator false.

9. **change** · `apps/web/src/pages/_app.tsx` L1-40 `InitApp`
   The hook is now called inside Init App in the top-level app file, so it runs eagerly on every auth rather than waiting for the lazy counterfactual bundle to load. This is the fix for the direct-URL race condition.

10. **change** · `apps/web/src/store/authSlice.ts` L22-70 `authSlice`
   A new cf-safe-synced boolean is added to the auth slice. It starts false, is reset to false on unauthenticated, and is set to true by the sync hook once the fetch completes. The select cf-safe-synced selector exposes it to consumers.

11. **change** · `apps/web/src/hooks/loadables/useLoadSafeInfo.ts` L1-60 `useLoadSafeInfo`
   Use load safe info now checks an awaiting-CF-sync flag before surfacing a CGW 404 as an error. While the store is hydrated but cf-safe-synced is still false, the hook extends the loading state instead of showing an error, giving the sync hook time to populate Redux.

12. **change** · `apps/web/src/features/counterfactual/store/counterfactualSyncListener.ts` L6-50 `counterfactualSyncListener`
   The sync listener middleware is now focused solely on DELETE. When remove undeployed safe fires, it reads the pre-action state to check is-creator, and only calls the delete endpoint if the current user was the creator. It uses dot-unwrap for proper error propagation, and errors are caught and logged rather than crashing the app.

13. **ripple** · `packages/utils/src/features/counterfactual/store/types.ts` L39-55 `UndeployedSafe`
   The undeployed safe type gains an optional is-creator field. Undefined is treated as true for backwards compatibility with existing persisted state, so old localStorage entries will still attempt a DELETE on removal without any migration needed.

14. **ripple** · `apps/web/src/features/counterfactual/store/undeployedSafesSlice.ts` L16-50 `undeployedSafesSlice`
   The add undeployed safe action now accepts and stores is-creator in the slice. The update undeployed safe reducer was also fixed to spread the existing entry before applying status changes, so is-creator and other fields are no longer silently dropped on status updates.

15. **ripple** · `apps/web/src/features/myAccounts/hooks/useSpaceAccountsData.ts` L1-50 `useSpaceAccountsData`
   The Spaces Accounts widget hook now falls back to the CF safe config in Redux when no safe overview is available from the API. A small helper reads threshold and owners from the undeployed safes slice so the widget shows correct data for safes that haven't been deployed yet.

16. **ripple** · `apps/web/src/features/spaces/components/SelectSafesOnboarding/components/SafeCardLayout.tsx` L27-70 `SafeCardLayout`
   Safe card layout now accepts is-undeployed and is-activating props. When is-undeployed is true, it swaps the fiat balance for an account item status chip, so counterfactual safes show a deployment status label instead of a zero balance across both the Spaces onboarding and read-only card views.

### What can go wrong

17. **risk** · `apps/web/src/features/counterfactual/hooks/useCounterfactualSafeSync.ts` L40-80
   Watch the error path in the sync hook: if the fetch fails, cf-safe-synced is still set to true and the sync key is updated, silently swallowing the error. Users could end up with an incomplete CF safe list and no indication that anything went wrong, and there is no retry mechanism.

18. **risk** · `apps/web/src/hooks/loadables/useLoadSafeInfo.ts` L30-70
   The awaiting-CF-sync flag extends the loading spinner until cf-safe-synced becomes true. If the sync hook never sets that flag, for example because auth state gets stuck or the hook is not mounted, the loading state hangs indefinitely with no timeout or fallback error shown to the user.

19. **risk** · `apps/web/src/features/counterfactual/services/persistCounterfactualSafe.ts` L36-80 `persistCounterfactualSafe`
   When a space ID is present, the function POSTs to the user endpoint first and then to the space endpoint. If the space POST fails, the function returns an error, but the user-level entry was already written to the backend with no rollback. A 409 conflict on the user endpoint is also treated as a hard error rather than an idempotent success, which could block retries.

### Bottom line

20. **wrap** · `apps/web/src/pages/_app.tsx` L1-40
   This PR replaces a fragile fire-and-forget pattern with a backend-first creation flow, an eager sync hook that runs at app init, and a Redux flag that prevents premature error screens, making counterfactual safes reliable for both solo users and space members.

## Reviewer checklist

- [ ] Verify that `persistCounterfactualSafe` correctly handles the case where `toBackendDto` returns null (e.g. for PredictedSafeProps) and that unauthenticated users still get their CF safe added to Redux via `replayCounterfactualSafeDeployment` without attempting a backend POST.
- [ ] Confirm `useCounterfactualSafeSync` always dispatches `setCfSafeSynced(true)` even on fetch error, and that `cfSafeSynced` is reset to false on every hydration in `_hydrationReducer`, to prevent `useLoadSafeInfo` from hanging indefinitely in a loading state.
- [ ] Check that `counterfactualSyncListener` only fires DELETE for entries where `isCreator !== false` and that existing persisted localStorage entries without the `isCreator` field do not cause spurious 40x DELETE errors on upgrade.
- [ ] Audit all callers of `addUndeployedSafe` to ensure space-synced safes pass `isCreator: false` and user-created safes pass `isCreator: true` (or omit it), since the default is `true` and a missing flag silently grants delete permission.
- [ ] Validate the `BackendCfSafeDto` ↔ `ReplayedSafeProps` round-trip in `counterfactualSafeMapper` for edge cases: null `paymentToken`, large `payment` values (BigNumber precision), and missing optional fields (`fallbackHandler`, `to`, `data`, `paymentReceiver`) that have no null-guards in `fromBackendDto`.
- [ ] Ensure the removal of `ContractAddresses` from the auto-generated `chains.ts` type and the `CONFIG_SERVICE_CHAINS` test mock does not silently break any runtime code that accesses `chain.contractAddresses` via dynamic property access outside TypeScript's reach.

## File-by-file

### `apps/web/src/components/common/AddressInput/index.test.tsx`

Three mock contact objects in AddressInput tests each gained two new required fields: `createdAt: ''` and `updatedAt: ''`.

_The contact type was extended with `createdAt`/`updatedAt` fields, requiring test fixtures to be updated to satisfy TypeScript type constraints._

**Risks:**
- Empty string may not be a valid value for timestamp fields if runtime validation exists — check if tests still accurately reflect real data shapes
- Verify all other test files using contact mocks were similarly updated to avoid type errors elsewhere

### `apps/web/src/components/common/EthHashInfo/index.test.tsx`

Added `createdAt: ''` and `updatedAt: ''` fields to the mock contact object in `EthHashInfo` test.

_Likely the `Contact` type was updated to require `createdAt` and `updatedAt` fields, making these additions necessary to satisfy TypeScript compilation._

**Risks:**
- Verify the Contact type definition actually requires these fields and empty strings are valid values
- Check if other test files using Contact mocks also need the same update

### `apps/web/src/components/new-safe/create/steps/ReviewStep/index.test.tsx`

The `render` call in the ReviewStep test now passes `initialReduxState` with a full `auth` slice including the new `cfSafeSynced: false` field alongside existing auth fields.

_The test must supply the new `cfSafeSynced` field added to `authSlice` to avoid selector errors or unexpected behavior when the component reads auth state._

**Risks:**
- Check that all other ReviewStep tests (if any) also provide the updated auth shape or use a shared fixture
- Verify `cfSafeSynced: false` is the correct initial value for this test scenario and doesn't accidentally suppress expected loading/error states
- Confirm the auth slice type definition requires `cfSafeSynced` so missing it in other tests would cause a TS error rather than a silent runtime issue

### `apps/web/src/components/new-safe/create/steps/ReviewStep/index.tsx`

ReviewStep now derives `effectivePayMethod` (forcing PayNow when unauthenticated) and replaces all direct `payMethod` references with it; the counterfactual path calls the new `persistCounterfactualSafe` (awaited, throws on failure) instead of the fire-and-forget `replayCounterfactualSafeDeployment`, and passes `isUserAuthenticated`/`spaceId` to the PayNowPayLater component.

_Makes backend persistence of counterfactual safes a blocking, error-surfacing step so creation failures are visible to the user rather than silently swallowed._

**Risks:**
- effectivePayMethod is computed outside the submit handler — if `isUserAuthenticated` changes between render and submit click, the derived value may be stale relative to what the user sees
- The early-return guard `if (successfulChains.length === 0) return` in the post-submit navigation block silently exits without setting an error or notifying the user
- persistCounterfactualSafe result is checked with `result.ok` — verify the function always returns a discriminated union and never throws directly, otherwise the `throw result.error` path may throw undefined
- replayCounterfactualSafeDeployment is still called in the PayNow + relay path (onSubmitCallback) with effectivePayMethod — confirm this path does not also need the new persist logic for consistency
- spaceId selector `lastUsedSpace` may return null/undefined; confirm persistCounterfactualSafe handles a missing spaceId gracefully

### `apps/web/src/components/new-safe/create/steps/ReviewStep/styles.module.css`

Added `margin-bottom: var(--space-2)` to the `.errorMessage` CSS rule in ReviewStep's styles module.

_Likely adds spacing below the error message to prevent it from visually colliding with the element below it when a backend error is displayed during CF safe creation._

**Risks:**
- Verify the spacing looks correct in all error states on the ReviewStep UI
- Check that var(--space-2) resolves to an appropriate value in the design system

### `apps/web/src/features/counterfactual/components/PayNowPayLater/index.tsx`

PayNowPayLater gains an `isUserAuthenticated` prop (default `true`); when false and PayLater is selected, it triggers SIWE sign-in via `signInAndSelectPayLater`, dispatches `setAuthenticated` with a hardcoded 24-hour expiry, then sets PayLater. UI shows a CircularProgress spinner, disabled radio, and an inline 'Sign in' prompt during/before auth.

_Likely gates the PayLater (counterfactual) flow behind authentication so the backend can persist the undeployed safe to the user's account._

**Risks:**
- Hardcoded `Date.now() + oneDayInMs` for auth expiry may diverge from the actual session expiry returned by the SIWE backend — should use the server-returned expiry if available
- `signIn()` result shape: check that `result.error` is the correct field to test for failure; a falsy `result` (null/undefined) is handled but other error shapes may not be
- Double-invocation risk: `signInAndSelectPayLater` is callable both from `onChoosePayMethod` and the inline 'Sign in' Typography click — concurrent calls are guarded by `signingIn` check, but verify `useSiwe` loading state is set synchronously
- `setAuthenticated` is dispatched client-side without verifying the session is actually valid server-side; if `signIn()` succeeds but the token is invalid, the user appears authenticated locally
- No error feedback shown to the user if `signIn()` fails (result.error truthy) — the UI silently stays on PayNow with no message
- Default `isUserAuthenticated = true` means existing callers are unaffected, but callers that pass `false` must ensure the prop stays in sync with actual auth state to avoid stale UI

### `apps/web/src/features/counterfactual/hooks/__tests__/useCounterfactualSafeSync.test.ts`

A new test file `useCounterfactualSafeSync.test.ts` is added with 194 lines covering the `useCounterfactualSafeSync` hook, mocking RTK Query endpoints for user and space CF safes, the Redux store, and `authSlice` selectors to verify fetch behavior under various auth/hydration/space conditions.

_Provides unit test coverage for the new hook that replaces the fire-and-forget sync pattern, ensuring it correctly gates on auth/hydration, re-fetches on space changes, deduplicates, and correctly tags `isCreator` for user-owned vs space-only CF safes._

**Risks:**
- The `dispatched` array is module-level and only has its `length` reset (not replaced), which could leak state between test files if Jest module cache is shared
- The mock for `@/store/authSlice` exports Symbols for selectors but the real implementation likely exports functions/selectors — if the hook uses these differently the tests may pass while the real code diverges
- The `flush` helper (`act(async () => {})`) may not reliably drain all microtasks/promises if the hook has deeper async chains, potentially causing false-positive passes
- The `spaceId` string-to-number coercion test (`'42'` → `{ spaceId: 42 }`) is important but only tested once; edge cases like `'0'` or non-numeric strings are not covered
- No test covers the `setCfSafeSynced` dispatch being called after sync completes, which is a key behavioral contract of the hook

### `apps/web/src/features/counterfactual/hooks/useCounterfactualSafeSync.ts`

New hook `useCounterfactualSafeSync` added that fetches counterfactual safes from both `/v1/users/counterfactual-safes` and `/v1/spaces/:spaceId/counterfactual-safes` on auth/hydration, merges them (user endpoint wins on collision), and dispatches `addUndeployedSafe` for any safe not already in Redux, then sets `cfSafeSynced` to true.

_Makes the backend the source of truth for counterfactual safes so space members can see each other's CF safes and direct URL navigation no longer races with a lazy sync._

**Risks:**
- On sync failure, `cfSafeSynced` is set to true and `lastSyncedKey` is updated, silently swallowing the error and preventing any retry — users may see incomplete CF safe lists with no indication of failure
- RTK Query `initiate` calls without `forceRefetch` may return stale cached data if the hook re-runs with the same spaceId after a backend update
- The `syncKey` only encodes `spaceId`, not `isUserAuthenticated`, so switching users without changing spaceId won't trigger a re-sync
- `getStoreInstance()` is called inside an async callback — if the store instance changes between dispatch and this call, `localSafes` could be stale or from a different store
- Type cast `as unknown as Parameters<typeof fromBackendDto>[0]` bypasses type safety; if the backend DTO shape diverges, `fromBackendDto` could silently produce malformed safe props
- No local→remote cleanup: safes deleted on the backend but present in Redux will persist until a page reload clears localStorage/Redux

### `apps/web/src/features/counterfactual/services/__tests__/counterfactualSafeMapper.test.ts`

A new test file is added for `counterfactualSafeMapper` covering `toBackendDto` and `fromBackendDto` with unit tests for ReplayedSafeProps, PredictedSafeProps, optional fields, and a full round-trip.

_Validates the mapper's serialization/deserialization logic that is now critical since the backend is the source of truth for counterfactual safes._

**Risks:**
- Round-trip test uses `toBackendDto` output as input to `fromBackendDto` — if both functions share the same bug (e.g. `??` vs `||` for paymentToken), the round-trip will pass but the bug won't be caught
- No test covers the `payment` field conversion from number to string ('100') in the reverse direction (fromBackendDto converting string back to number)
- PredictedSafeProps test only checks for null return but doesn't verify no side effects or error thrown
- mockPredictedSafeProps omits `paymentToken` and `payment` — worth confirming these are truly optional in the type definition

### `apps/web/src/features/counterfactual/services/__tests__/persistCounterfactualSafe.test.ts`

A new test file `persistCounterfactualSafe.test.ts` is added with 5 unit tests covering `persistCounterfactualSafe`, mocking `cgwApi` endpoints for both user and space counterfactual safe creation, and `replayCounterfactualSafeDeployment`.

_Validates the new synchronous backend-first creation flow: success paths, 500/409 error handling, space endpoint behavior, and unauthenticated fallback._

**Risks:**
- Mock for `cgwApi` in `counterfactual-safes` and `spaces` modules uses two separate `cgwApi` named exports — verify the actual source modules export a single `cgwApi` and that the mock paths match exactly
- The dispatch mock for the 'space-create-thunk' failure test returns `{ error: { status: 500 } }` for `action.type === 'space-create-thunk'`, but the user-create-thunk returns a plain object (no error) — confirm the real implementation checks the dispatch return value consistently for both endpoints
- Test for unauthenticated user asserts `replayImpl` is called but does NOT assert `addUndeployedSafe` is dispatched to Redux — if the implementation dispatches to Redux via `replayImpl`, this is fine, but if it's a separate dispatch call it may be untested
- 409 conflict test only checks the error message matches /already exists/i — no assertion that `spaceInitiate` and `replayImpl` are NOT called, leaving a gap if the implementation accidentally proceeds after a 409
- No test covers the case where `spaceId` is set but `isUserAuthenticated` is false — behavior in that combination is untested

### `apps/web/src/features/counterfactual/services/counterfactualSafeMapper.ts`

New file `counterfactualSafeMapper.ts` adds `BackendCfSafeDto` type and two conversion functions: `toBackendDto` converts `ReplayedSafeProps` to a flat backend DTO (returning null for `PredictedSafeProps`), and `fromBackendDto` reconstructs `ReplayedSafeProps` from the DTO.

_Provides the serialization layer between the frontend's nested safe props and the backend counterfactual-safes API schema, enabling backend persistence of CF safes._

**Risks:**
- toBackendDto returns null for PredictedSafeProps — callers must handle null or CF safes created via prediction path will be silently dropped
- safeVersion is cast with `as SafeVersion` without validation — invalid backend values won't be caught at runtime
- payment field is converted String→Number in fromBackendDto but Number→String in toBackendDto; precision loss possible for large BigNumber-like values
- fallbackHandler, to, data, paymentReceiver are mapped without null/undefined guards — missing backend fields could produce undefined in required positions
- No validation that dto.owners is non-empty array or dto.threshold <= owners.length

### `apps/web/src/features/counterfactual/services/index.ts`

Two new modules, `counterfactualSafeMapper` and `persistCounterfactualSafe`, are re-exported from the counterfactual services barrel file `index.ts`.

_Likely exposes the new mapper and persistence helpers as part of the public API of the counterfactual services module so other parts of the codebase can import them without deep path references._

**Risks:**
- Verify neither module has naming collisions with existing exports from the barrel (e.g. duplicate export names across the five existing modules)
- Confirm `persistCounterfactualSafe` does not inadvertently expose internal implementation details that should remain private

### `apps/web/src/features/counterfactual/services/persistCounterfactualSafe.ts`

New file `persistCounterfactualSafe.ts` introduces a single async function `persistCounterfactualSafe` that sequentially POSTs to the user counterfactual-safes endpoint and optionally to the space safes endpoint before calling `replayCounterfactualSafeDeployment` to update Redux, returning a typed `PersistResult` union.

_Centralises the counterfactual safe creation path so backend persistence is blocking and failures surface to the caller instead of being silently swallowed, and ensures Redux is only updated after backend confirmation._

**Risks:**
- Space safe creation failure (409 or other) returns ok=false but the user-level entry was already written to the backend — no rollback of the user-level POST is attempted, leaving a dangling entry.
- If `toBackendDto` returns null (falsy), the entire backend write is silently skipped even for authenticated users — verify this is intentional and all callers handle the no-op case.
- 409 conflict on the user endpoint is treated as a hard error; consider whether a conflict should instead be treated as idempotent success (safe already exists).
- `replayCounterfactualSafeDeployment` is called for unauthenticated users regardless of whether `toBackendDto` would have returned null — ensure this is safe.
- RTK Query `initiate` calls do not pass `{ forceRefetch: true }` or subscription options; check whether duplicate in-flight requests are deduplicated in a way that could mask errors.

### `apps/web/src/features/counterfactual/store/__tests__/counterfactualSyncListener.test.ts`

A new test file `counterfactualSyncListener.test.ts` is added with 6 test cases covering the `counterfactualSyncListener` middleware: it verifies DELETE API dispatch when user is creator (or legacy entry), skips dispatch when not creator or unauthenticated, and swallows backend errors via `console.error`.

_Provides regression coverage for the refactored sync listener that now uses `.unwrap()` and guards on `isCreator`/auth state, ensuring the DELETE path behaves correctly under various conditions._

**Risks:**
- The `runListener` helper bypasses the actual RTK listener effect mechanism — it calls the middleware chain but the listener's `effect` function is registered separately; verify the async effect is actually being invoked and not just the middleware pass-through
- Two `await Promise.resolve()` microtask flushes may be insufficient for deeply chained async operations in the listener effect; tests could pass vacuously if the effect never runs
- The mock for `dispatch` uses `(t) => t` as default, meaning the dispatched thunk's `.unwrap()` is never called unless `dispatchImpl` is overridden — the test for successful dispatch checks `typeof dispatched.unwrap === 'function'` but never awaits it, so the unwrap path isn't truly exercised
- Legacy entries (isCreator undefined) are treated as creator — confirm this matches the actual implementation logic in `counterfactualSyncListener`
- No test covers the case where the safe entry doesn't exist in state at all (e.g., already removed before listener fires)

### `apps/web/src/features/counterfactual/store/counterfactualSyncListener.ts`

New file `counterfactualSyncListener.ts` adds a Redux listener middleware that fires on `removeUndeployedSafe` actions, reads the pre-action state to check `isCreator`, and calls `counterfactualSafesDeleteV1` via RTK Query with `.unwrap()` if the current user was the creator.

_Ensures backend DELETE is only attempted by the safe's creator (not space members who received it via sync), preventing spurious 40x errors while keeping backend state consistent with local Redux removals._

**Risks:**
- Failure to delete from backend is silently swallowed (only console.error) — user may believe safe is removed locally but it persists on backend and will re-appear on next sync
- The `isCreator !== false` default-true logic means any legacy persisted entry without the flag will attempt a DELETE — verify this is safe for all pre-existing localStorage entries
- No retry logic on DELETE failure; transient network errors will leave backend out of sync
- Listener is registered but the call site (where `counterfactualSyncListener` is invoked) is not shown — verify it is actually wired into the store middleware
- RTK Query `initiate` without `forceRefetch` may be deduplicated or cached unexpectedly for mutation endpoints — confirm `.initiate()` always fires a new request

### `apps/web/src/features/counterfactual/store/index.ts`

Added a named re-export of `counterfactualSyncListener` from `counterfactualSyncListener` module through the `store/index.ts` barrel file.

_Likely enables consumers to import `counterfactualSyncListener` from the barrel instead of the direct module path, supporting the refactored listener usage described in the PR._

**Risks:**
- Verify `counterfactualSyncListener` is actually exported from `./counterfactualSyncListener` with that exact name
- Check no circular dependency is introduced by this barrel re-export

### `apps/web/src/features/counterfactual/store/undeployedSafesSlice.ts`

Added optional `isCreator` field (defaulting to `true`) to the `addUndeployedSafe` action payload and stored it in state. Also refactored `updateUndeployedSafe` to use a local `existing` variable and spread `...existing` to preserve all fields (including `isCreator`) during status updates.

_Tracks whether the current user created a CF safe or received it via space sync, and ensures `updateUndeployedSafe` no longer silently drops fields like `isCreator` that weren't explicitly re-set._

**Risks:**
- Verify all consumers of `addUndeployedSafe` that should pass `isCreator: false` (space sync path) actually do so — missing it silently defaults to `true`
- Check that the `isCreator` field is used somewhere meaningful (e.g. UI gating, permissions) and not just stored but never read
- The spread `...existing` in `updateUndeployedSafe` now preserves any future fields added to the state shape — confirm this is intentional and won't cause stale data issues
- Ensure Redux persist/rehydration handles the new `isCreator` field correctly for existing persisted state that lacks it (undefined vs true)

### `apps/web/src/features/multichain/components/CreateSafeOnNewChain/index.tsx`

Replaces the lazy-loaded `replayCounterfactualSafeDeployment` from `CounterfactualFeature` with a direct call to `persistCounterfactualSafe`, passing `isAuthenticated` and `lastUsedSpace` selectors, and now awaits the result with explicit error handling via `setCreationError`.

_Aligns the 'replay safe on new chain' flow with the new backend-first persistence pattern so backend failures surface as errors instead of being silently swallowed._

**Risks:**
- Verify `persistCounterfactualSafe` signature matches the new call-site arguments (chainId, safeAddress, props, name, payMethod, spaceId, isUserAuthenticated, dispatch)
- Check that `persistResult.ok` / `persistResult.error` shape is consistently returned by `persistCounterfactualSafe` in all code paths including unauthenticated
- Confirm removal of `CounterfactualFeature` / `useLoadFeature` doesn't break code-splitting or lazy-loading assumptions elsewhere
- Ensure `spaceId` from `lastUsedSpace` is the correct selector to use here (vs. a space derived from the current safe context)
- Verify that on error, the function returns early before the two `trackEvent` calls, so analytics aren't fired for failed creations

### `apps/web/src/features/myAccounts/hooks/__tests__/useMultiAccountItemData.test.ts`

Adds a new `describe` block with four unit tests for `isFullyUndeployed` and `isActivating` return values from `useMultiAccountItemData`, covering mixed deployed/undeployed groups, fully undeployed groups, activating state (PROCESSING status), and empty groups.

_Likely ensures correctness of the new counterfactual-safe deployment state logic in multi-chain account items, preventing regressions in UI flags that control display of undeployed/activating states._

**Risks:**
- Test for isActivating=true uses PROCESSING status but does not verify the exact threshold between AWAITING_EXECUTION and other statuses — check that the hook implementation matches this boundary
- Empty group test expects isFullyUndeployed=false but the semantic of 'all safes undeployed' with zero safes could reasonably be true — verify the hook uses every() vs some() correctly
- No test covers a group where some safes are PROCESSING and some are deployed — edge case for isActivating when isFullyUndeployed=false

### `apps/web/src/features/myAccounts/hooks/__tests__/useSpaceAccountsData.test.ts`

New test file added for `useSpaceAccountsData` hook covering four scenarios: CF safe owners fallback when no SafeOverview exists, SafeOverview taking precedence over CF data, empty string when neither source is available, and multichain CF safe owners fallback.

_Validates the new fallback logic that reads `threshold`/`owners` from Redux `undeployedSafes` when no `SafeOverview` is returned by the API, ensuring the Spaces Accounts widget displays correct data for counterfactual safes._

**Risks:**
- Test for 'SafeOverview takes precedence' also seeds undeployedSafes with conflicting data — verify the hook implementation actually ignores CF data when overview is present, not just that the test passes
- Multichain test only seeds chainId '1' in undeployedSafes but the safe exists on both '1' and '137' — confirm the hook picks the right chain entry and the expected '2/3' output is deterministic
- No test covers the case where CF data exists on one chain but not another in a multichain safe — edge case worth considering
- Mock for useGetMultipleSafeOverviewsQuery uses `as never` cast which could mask type mismatches if the hook's query signature changes

### `apps/web/src/features/myAccounts/hooks/useMultiAccountItemData.ts`

Adds two new memoized booleans `isFullyUndeployed` and `isActivating` to the return value of `useMultiAccountItemData`. `isFullyUndeployed` is true when all safes in the multi-chain item are undeployed; `isActivating` is true when fully undeployed and at least one safe has a status other than `AWAITING_EXECUTION`.

_Likely exposes activation/deployment state to the multi-account list UI so it can show distinct visual states (e.g. spinner or 'activating' badge) for counterfactual safes that are in-flight versus merely pending._

**Risks:**
- Logic inverts expectation: `isActivating` returns false when `record` is undefined (safe not in undeployedSafes), which may silently hide activating state if Redux hasn't synced yet
- Status string 'AWAITING_EXECUTION' is compared as a literal — verify it matches all possible backend enum values and won't break if the enum changes
- isFullyUndeployed requires sortedSafes.length > 0 but doesn't guard against the case where deployedSafes is stale/empty due to loading, potentially showing wrong state transiently

### `apps/web/src/features/myAccounts/hooks/useSpaceAccountsData.ts`

Added `undeployedSafes` Redux state as a fallback source for `owners` field in `formatMultichainAccount` and `formatSingleSafe`. A new helper `getCfOwners` reads `threshold`/`owners` from `undeployedSafes[chainId][address].props.safeAccountConfig` when no `SafeOverview` is available.

_Counterfactual (undeployed) safes have no on-chain overview data, so the Spaces Accounts widget previously showed an empty string for owners/threshold; this fix surfaces the locally-stored CF safe config instead._

**Risks:**
- getCfOwners uses safe.safes[0]?.chainId for multichain accounts — if the CF safe exists on a different chain index, it will silently return undefined
- Falls back to empty string '' when getCfOwners returns undefined, which may be indistinguishable from a deployed safe with no overview loaded yet
- undeployedSafes added to useMemo dependency array — verify no unnecessary re-renders if undeployedSafes reference changes frequently
- No null-guard on cfSafe.props.safeAccountConfig — if the DTO shape differs, this could throw at runtime

### `apps/web/src/features/spaces/components/SafeAccounts/SafeCardReadOnly.tsx`

SafeCardReadOnly now destructures `isUndeployed` and `isActivating` from `useSafeCardData`, and conditionally renders `AccountItem.StatusChip` instead of `FiatBalance` when the safe is undeployed.

_Counterfactual (undeployed) safes in the Spaces read-only card view now show a deployment status chip rather than a fiat balance, matching the UX of the regular safe card._

**Risks:**
- Verify `useSafeCardData` correctly returns `isUndeployed`/`isActivating` for multi-chain safe items, not just single-chain
- Check that `AccountItem.StatusChip` handles `isActivating=undefined` or falsy gracefully without visual glitches
- Ensure `ThresholdBadge` still renders correctly for undeployed safes when `threshold > 0` (depends on CF safe data being populated)

### `apps/web/src/features/spaces/components/SafeAccounts/__tests__/SafeCardReadOnly.test.tsx`

New test file added for `SafeCardReadOnly` component with two test cases: one verifying no 'Not activated' chip renders for a deployed safe, and one verifying the chip renders when `undeployedSafes` Redux state contains the safe's chainId/address entry.

_Validates that `SafeCardReadOnly` correctly reads counterfactual safe state from Redux to conditionally show the 'Not activated' chip, covering the new CF safe fallback behavior in the Spaces Accounts widget._

**Risks:**
- Both tests use the same `safe` object with address '0xDEADBEEF' but `undeployedSafes` key uses the same string — verify address casing consistency between safeItemBuilder output and Redux key lookup
- Test mocks both `gatewayApi.useGetSafeOverviewQuery` and `gatewaySlices.useGetSafeOverviewQuery` returning undefined — ensure the component under test actually uses the mocked module path, not a third import path
- No test for the case where `SafeOverview` data IS present but undeployedSafes also exists — edge case behavior is untested
- MockIntersectionObserver is defined inline rather than in a shared test utility — minor duplication risk

### `apps/web/src/features/spaces/components/SelectSafesOnboarding/components/SafeCard.tsx`

SafeCard now destructures `isUndeployed` and `isActivating` from `useSafeCardData` and passes both props down to the rendered card components in both the multi-chain and single-chain render paths.

_Likely ensures counterfactual/undeployed safes are visually distinguished (e.g. showing deployment status) in the Spaces onboarding safe selection UI._

**Risks:**
- Verify both render paths (multi-chain SafeCard component and single-chain FormSafeCard/similar component) actually accept and handle `isUndeployed`/`isActivating` props without TypeScript errors or silent ignoring
- Check that `useSafeCardData` correctly derives `isUndeployed` and `isActivating` for multi-chain safe items, not just single-chain ones

### `apps/web/src/features/spaces/components/SelectSafesOnboarding/components/SafeCardLayout.tsx`

Added `isUndeployed` and `isActivating` optional props to `SafeCardLayout`. When `isUndeployed` is true, the `FiatBalance` component is replaced with `AccountItem.StatusChip` (passing `undeployedSafe` and `isActivating` props).

_Allows the Spaces safe card to visually distinguish counterfactual (undeployed) safes by showing a status chip instead of a fiat balance, consistent with how undeployed safes are displayed elsewhere._

**Risks:**
- Verify `AccountItem.StatusChip` handles both `isActivating=true` and `isActivating=false` states correctly and renders expected labels/styles
- Check that `ThresholdBadge` still renders correctly below the status chip for undeployed safes (threshold/ownersCount must be populated from Redux fallback)
- Confirm callers of `SafeCardLayout` pass correct `isUndeployed`/`isActivating` values — missing props default to false, silently showing fiat balance for CF safes if not wired up

### `apps/web/src/features/spaces/components/SelectSafesOnboarding/components/__tests__/SafeCardLayout.test.tsx`

A new test file `SafeCardLayout.test.tsx` is added with three test cases covering `SafeCardLayout` rendering: deployed safe (shows FiatBalance, no status chips), undeployed safe (shows 'Not activated' chip), and activating safe (shows 'Activating account' chip).

_Likely adds baseline coverage for the `SafeCardLayout` component's counterfactual/undeployed state rendering introduced or modified as part of the CF safe backend persistence work._

**Risks:**
- No test verifies FiatBalance actually renders with the correct value when deployed, only that status chips are absent
- Mock for useChains uses module-level constant; if SafeCardLayout imports useChain differently the mock path may not intercept it
- safeItemBuilder usage is not verified to produce a shape compatible with SafeCardLayout's safes prop type

### `apps/web/src/features/spaces/components/SelectSafesOnboarding/hooks/__tests__/useSafeCardData.test.ts`

New test file added for `useSafeCardData` hook, covering single-chain and multi-chain safe scenarios for `isUndeployed` and `isActivating` flags derived from the `undeployedSafes` Redux slice.

_Validates that the hook correctly reflects counterfactual safe state (undeployed/activating) from Redux for both single and multi-chain safes, ensuring the UI card data is accurate after the backend-as-source-of-truth refactor._

**Risks:**
- No test covers the case where `isActivating=true` for a multi-chain safe (e.g., one chain PROCESSING, one AWAITING_EXECUTION) — edge case behavior is untested
- Mocks `useGetMultipleSafeOverviewsQuery` and `useGetSafeOverviewQuery` to return `undefined` — tests don't verify behavior when overview data is present alongside undeployed state
- Uses `as unknown as Partial<RootState>` casts which may silently pass if the Redux state shape changes
- No test for the `cfSafeSynced` flag or loading state introduced in `authSlice` that affects `useLoadSafeInfo`

### `apps/web/src/features/spaces/components/SelectSafesOnboarding/hooks/useSafeCardData.ts`

useSafeCardData now extracts and returns `isUndeployed` and `isActivating` fields for both multi-chain and single-chain safe card data, sourcing them from `isFullyUndeployed`/`isActivating` in multiData and `undeployedSafe`/`isActivating` in singleData respectively.

_Likely exposes counterfactual/undeployed safe state to the SafeCard UI so it can render appropriate visual indicators or disable actions for undeployed safes in the Spaces onboarding flow._

**Risks:**
- Verify that `undeployedSafe` and `isActivating` are actually provided by the singleData hook; if not, both will be undefined/falsy silently
- Check that `isFullyUndeployed` correctly reflects the multi-chain case where some but not all chains are deployed — partial deployment may be misrepresented as deployed
- Confirm downstream consumers of useSafeCardData handle the new fields without breaking existing rendering logic

### `apps/web/src/hooks/__tests__/useAllAddressBooks.test.ts`

Three test fixtures for space contacts in `useAllAddressBooks.test.ts` have `createdAt` and `updatedAt` fields (both empty strings) added to match an updated type definition.

_Likely the `Contact` or space contact type was extended with `createdAt`/`updatedAt` fields, requiring test fixtures to be updated to satisfy TypeScript._

**Risks:**
- Verify the actual type definition includes createdAt/updatedAt as required (not optional) fields
- Empty string may not be a valid value if the type expects ISO date strings or null — check if tests still accurately reflect real data shapes

### `apps/web/src/hooks/loadables/useLoadSafeInfo.ts`

useLoadSafeInfo gains three new behaviors: (1) a `awaitingCfSync` flag that suppresses CGW 404 errors and extends the loading state until the store is hydrated and CF safes are synced from the backend; (2) CGW errors are only logged/surfaced when not suppressed by CF sync and no `undeployedSafe` exists in Redux; (3) a self-healing `useEffect` that dispatches `removeUndeployedSafe` when `cgwData` exists alongside a local `undeployedSafe` entry.

_Prevents a flash of 'Safe not found' error when navigating directly to a counterfactual safe URL before the backend sync completes, and cleans up stale CF entries when a safe has been deployed._

**Risks:**
- awaitingCfSync extends the loading spinner indefinitely if cfSynced never becomes true (e.g. auth state stuck, sync hook never runs) — check what sets cfSafeSynced to true and whether it has a timeout/error path
- suppressCgwError only suppresses 404s, but a deployed safe that returns 404 transiently (e.g. CGW propagation lag) will also be suppressed — verify this is acceptable
- Self-heal effect dispatches removeUndeployedSafe on every render where cgwData && undeployedSafe — confirm the action is idempotent and won't trigger redundant backend DELETEs in a loop
- loading = cgwLoading || awaitingCfSync means even non-CF safes show a loading state until cfSynced — check performance impact for users with many safes or slow auth
- The error suppression condition `!suppressCgwError && !undeployedSafe` in both useEffect and useMemo must stay in sync — duplication is a maintenance risk

### `apps/web/src/hooks/useAllAddressBooks.ts`

Added `createdAt: ''` and `updatedAt: ''` fields to the object literal returned by `mapLocalToContacts` when mapping local address book entries to `ExtendedContact` objects.

_Likely the `ExtendedContact` type was updated to require `createdAt` and `updatedAt` fields, and this fills in empty-string defaults for locally-sourced contacts that have no backend timestamps._

**Risks:**
- Verify that downstream consumers of `createdAt`/`updatedAt` handle empty strings gracefully and don't attempt date parsing or sorting on them
- Check if empty string is the correct sentinel vs null/undefined for optional timestamp fields in ExtendedContact

### `apps/web/src/pages/_app.tsx`

Imports `useCounterfactualSafeSync` and calls it inside `InitApp` component in `_app.tsx`, moving it from lazy-loaded `LazyCounterfactual` to the eager app initialization path.

_Ensures counterfactual safe data is fetched from the backend immediately on auth rather than waiting for the lazy-loaded counterfactual feature bundle, eliminating the race condition that caused a 'Safe couldn't be loaded' flash._

**Risks:**
- Verify `useCounterfactualSafeSync` is safe to call unconditionally (e.g., no-ops when unauthenticated) to avoid unnecessary API calls or errors for non-CF users
- Check if moving this hook out of the lazy bundle increases the initial bundle size
- Confirm the hook is no longer called in `LazyCounterfactual` to avoid double-execution

### `apps/web/src/store/__tests__/authSlice.test.ts`

Tests added for new `setCfSafeSynced` action and `selectCfSafeSynced` selector in `authSlice`, including a test verifying `setUnauthenticated` resets `cfSafeSynced` to false.

_Validates the new `cfSafeSynced` flag lifecycle used to suppress CGW 404 errors while counterfactual safe sync is in progress._

**Risks:**
- The `selectCfSafeSynced` selector test hardcodes the full auth state shape — if `authSlice` state shape changes, this test may silently pass with stale assumptions
- No test covers the initial default value of `cfSafeSynced` (should be false on first hydration)

### `apps/web/src/store/__tests__/index.test.ts`

Three snapshot assertions in `store/__tests__/index.test.ts` are updated to include the new `cfSafeSynced: false` field in the `auth` slice shape.

_Keeps existing store serialization/snapshot tests passing after `cfSafeSynced` was added to `authSlice` initial state._

**Risks:**
- Verify `cfSafeSynced` initial value is always `false` on hydration and not accidentally persisted across sessions
- Check that all three test cases cover distinct scenarios (merge, override, etc.) and that none of them test the `cfSafeSynced: true` path

### `apps/web/src/store/authSlice.ts`

Adds a `cfSafeSynced` boolean field to `AuthPayload`/`authSlice`, with a `setCfSafeSynced` action and `selectCfSafeSynced` selector. The flag is reset to `false` in `setUnauthenticated` and initialized as `false` in `initialState`.

_Provides a Redux-tracked gate so consumers (e.g. `useLoadSafeInfo`) can distinguish 'CF sync not yet completed' from 'safe genuinely not found', preventing a flash of error UI on direct URL navigation to a counterfactual safe._

**Risks:**
- Flag is never reset on hydration (`isStoreHydrated` sets to true but `cfSafeSynced` stays false until explicitly set) — verify the sync hook always calls `setCfSafeSynced(true)` even on fetch error, otherwise the loading state could hang indefinitely
- Flag persists in Redux but is not persisted to localStorage (initialState false) — confirm this is intentional and that rehydration from localStorage doesn't restore a stale `true` value
- Only reset on `setUnauthenticated`; if a user stays authenticated across page reloads, the flag starts false again (correct), but verify the sync hook is always triggered after hydration in that case

### `apps/web/src/store/index.ts`

Adds `counterfactualSyncListener` to the store's listener array and resets `cfSafeSynced: false` in `_hydrationReducer` alongside the existing `isStoreHydrated` and `isOidcLoginPending` resets on each hydration.

_Ensures `cfSafeSynced` is always false at page load so consumers (e.g. `useLoadSafeInfo`) correctly wait for a fresh backend CF safe sync before showing errors, and wires up the listener that drives that sync._

**Risks:**
- Verify `cfSafeSynced` is defined in `authSlice` initial state and its type allows `false` to avoid TS errors
- Check that resetting `cfSafeSynced` on every hydration doesn't cause a loading flash on pages where CF safes are irrelevant
- Confirm `counterfactualSyncListener` is idempotent if registered multiple times or if auth fires repeatedly

### `apps/web/src/store/slices.ts`

Adds `counterfactualSyncListener` to the re-exports from `@/features/counterfactual/store` in the slices barrel file.

_Likely exposes `counterfactualSyncListener` for use in other modules (e.g., app initialization) without requiring a deep import path._

**Risks:**
- Verify `counterfactualSyncListener` is actually exported from `@/features/counterfactual/store` to avoid a silent undefined re-export
- Check all consumers of this barrel export to ensure none accidentally import the listener and register it twice

### `apps/web/src/tests/mocks/chains.ts`

Removed the `contractAddresses` constant and all 13 references to it from the `CONFIG_SERVICE_CHAINS` mock array in the test chains fixture, leaving each chain object without a `contractAddresses` field.

_Likely aligns the mock with an updated `Chain` type definition that no longer includes `contractAddresses` as a required field, preventing TypeScript errors in tests._

**Risks:**
- Verify the `Chain` type from `@safe-global/store/gateway/AUTO_GENERATED/chains` no longer requires `contractAddresses`, otherwise tests may silently use incomplete mocks
- Check if any test that consumes `CONFIG_SERVICE_CHAINS` relies on `contractAddresses` being present on chain objects

### `packages/store/package.json`

The `check-sync` script changed from `yarn ts-node --esm scripts/check-auto-generated-sync.ts` to `ts-node scripts/check-auto-generated-sync.ts`, removing the `yarn` prefix and `--esm` flag.

_Likely fixes a resolution or execution issue with ts-node, but removing `--esm` may change how the script handles ES module syntax._

**Risks:**
- Removing --esm flag may break the script if check-auto-generated-sync.ts uses ESM imports/exports
- Dropping `yarn` prefix assumes ts-node is available in PATH or as a local bin; verify CI environments resolve it correctly

### `packages/store/scripts/api-schema/schema.json`

The API schema adds new endpoints and schemas for counterfactual safes (POST/GET/DELETE `/v1/users/counterfactual-safes`, GET `/v1/spaces/{spaceId}/counterfactual-safes`), a fee preview endpoint (`POST /v1/chains/{chainId}/fees/{safeAddress}/preview`) with associated DTOs, removes the `ContractAddresses` schema and its reference from the chain response, adds `createdAt`/`updatedAt` to the address book entry, and adds `safeTxHash` to the relay request body.

_These schema changes define the backend contract that the counterfactual safe persistence feature depends on, ensuring generated types and API clients match the new CGW endpoints._

**Risks:**
- GetCounterfactualSafeItem marks fallbackHandler, to, paymentToken, payment, paymentReceiver as required in the response schema but nullable — consumers must handle null values for all of these
- CounterfactualSafeDto (POST body) does NOT mark those same fields as required, creating an asymmetry between create and read schemas that could cause mapper/validation mismatches
- Removal of ContractAddresses schema and its field from the chain response is a breaking change — verify no existing code references chain.contractAddresses
- GetCounterfactualSafesResponse uses additionalProperties keyed by chainId (string) — confirm the mapper correctly iterates this dynamic key structure
- safeTxHash added as optional to relay request but described as 'required for chains enabled for relay-fee relayer' — missing validation enforcement in schema

### `packages/store/scripts/check-auto-generated-sync.ts`

Replaced `new URL('.', import.meta.url).pathname` with `__dirname` for resolving `SCRIPTS_DIR` in `check-auto-generated-sync.ts`.

_Likely switches the script from ESM-style module URL resolution to CommonJS `__dirname`, fixing compatibility when the script is run in a CJS context or via ts-node without ESM mode._

**Risks:**
- Verify the build/run environment actually supports `__dirname` (CJS) and that this script is not executed as an ES module where `__dirname` is undefined
- Check that path resolution still points to the correct directories after the change, especially on Windows where `pathname` from `URL` and `__dirname` may differ

### `packages/store/scripts/openapi-config.ts`

Added `spaceCounterfactualSafes` to the spaces endpoint filter and introduced a new output file `counterfactual-safes.ts` filtering on `counterfactualSafes` endpoints in the OpenAPI codegen config.

_Generates RTK Query hooks for both user-level and space-level counterfactual safe endpoints so the frontend can call the new CGW backend APIs._

**Risks:**
- Verify the generated `counterfactual-safes.ts` file is committed and up-to-date with the actual CGW OpenAPI spec
- Check that `spaceCounterfactualSafes` regex correctly matches all intended space CF safe endpoint names and no unintended ones
- Ensure the new file is properly re-exported from the store package barrel

### `packages/store/src/gateway/AUTO_GENERATED/.schema-hash`

The AUTO_GENERATED schema hash file was updated from `5ca9d1246d0b4427e72fd519af45952b8628e30b5d4c81af985dc67ccfdeeb40` to `fd477006218a7f8daededa61f21cb93b1bc85f20cb2a0dc711e1c102a619b700`.

_Likely reflects a change in the CGW OpenAPI schema to support the new counterfactual safes endpoints, triggering regeneration of auto-generated gateway client code._

**Risks:**
- Verify the corresponding AUTO_GENERATED client code was also regenerated and matches this new hash
- Check that new counterfactual-safes endpoints are fully represented in the generated types

### `packages/store/src/gateway/AUTO_GENERATED/chains.ts`

The `ContractAddresses` type and the `contractAddresses` field on the `Chain` type are removed from the auto-generated chains API types.

_Likely reflects a backend API change where contract addresses were removed from the chain response, possibly moved to a different endpoint or no longer needed by the client._

**Risks:**
- Check all consumers of `Chain.contractAddresses` or `ContractAddresses` type — any code still referencing these will now have TypeScript errors or silent undefined access
- Verify the backend API actually no longer returns `contractAddresses` in the chain response to confirm this isn't a premature removal
- Search for runtime usages that may not be caught by TypeScript (e.g. dynamic property access) that could silently break

### `packages/store/src/gateway/AUTO_GENERATED/counterfactual-safes.ts`

Adds a new auto-generated RTK Query file defining three endpoints against `/v1/users/counterfactual-safes`: `counterfactualSafesCreateV1` (POST), `counterfactualSafesGetV1` (GET), and `counterfactualSafesDeleteV1` (DELETE), along with their TypeScript types and exported hooks.

_Provides the RTK Query API layer for persisting and retrieving counterfactual safes from the CGW backend, replacing the previous local-only Redux/localStorage approach._

**Risks:**
- CounterfactualSafesCreateV1ApiResponse is typed as `unknown` — no type safety on the 201 response body; callers using `.unwrap()` won't get typed data
- CounterfactualSafesDeleteV1ApiResponse is also `unknown` — same concern
- GetCounterfactualSafesResponse uses a string-keyed map `{ [key: string]: GetCounterfactualSafeItem[] }` — verify the key semantics (likely chainId) match backend contract
- CounterfactualSafeDto has optional nullable fields (`fallbackHandler?`, `to?`, `paymentToken?`, etc.) while GetCounterfactualSafeItem has them as non-optional `string | null` — asymmetry could cause mapping issues
- All three mutations share the same tag `counterfactual-safes` and invalidate it — DELETE and CREATE will both refetch, confirm this is intentional and won't cause race conditions
- overrideExisting: false means if this endpoint is registered elsewhere it will silently not override — verify no duplicate registration

### `packages/store/src/gateway/AUTO_GENERATED/relay.ts`

Added optional `safeTxHash` query parameter to `relayGetRelaysRemainingV1` RTK Query endpoint, updating both the query builder and the `RelayGetRelaysRemainingV1ApiArg` type. Also added `safeTxHash` field to `RelayDto` type.

_Likely enables relay-fee eligibility checks on chains that require a specific safe transaction hash to determine remaining relay quota._

**Risks:**
- Verify all existing callers of `relayGetRelaysRemainingV1` still work correctly without passing `safeTxHash` (undefined params key may be serialized as empty string in some HTTP clients)
- Check if `safeTxHash` being optional in `RelayGetRelaysRemainingV1ApiArg` but 'required for chains enabled for relay-fee relayer' causes silent failures on those chains
- Confirm `params: { safeTxHash: undefined }` is stripped from the query string by the RTK Query base query implementation used here

### `packages/store/src/gateway/AUTO_GENERATED/spaces.ts`

Adds a new RTK Query endpoint `spaceCounterfactualSafesGetV1` to the auto-generated spaces API, targeting `GET /v1/spaces/:spaceId/counterfactual-safes`, along with its response/arg types (`GetCounterfactualSafeItem`, `GetCounterfactualSafesResponse`) and two new fields (`createdAt`, `updatedAt`) on `SpaceAddressBookItemDto`.

_Enables space members to fetch counterfactual safes created by any member of a space, supporting the shift from local-only to backend-as-source-of-truth for undeployed safes._

**Risks:**
- Response shape uses `safes: { [key: string]: GetCounterfactualSafeItem[] }` — verify the key semantics (chainId? address?) match actual backend response to avoid silent mapping errors
- File is auto-generated; confirm the generator was actually run against the updated CGW spec rather than hand-edited, to avoid drift
- `providesTags: ['spaces']` means any spaces mutation invalidates this cache — check if that causes unintended refetches or if a more specific tag is needed
- `createdAt`/`updatedAt` added to `SpaceAddressBookItemDto` — verify no existing consumers break if these fields are absent on older backend responses

### `packages/store/tsconfig.json`

Added a `ts-node` section to `packages/store/tsconfig.json` that sets `compilerOptions.module` to `"CommonJS"` for ts-node execution.

_Likely required to allow ts-node (used for scripts or tests in the store package) to run without ESM/CJS interop errors after new imports were added._

**Risks:**
- Verify this doesn't conflict with the package's main `compilerOptions.module` setting if one exists
- Check whether any build tooling (jest, scripts) in this package relies on ts-node and confirm CommonJS is the correct target

### `packages/utils/src/features/counterfactual/store/types.ts`

Added optional `isCreator?: boolean` field to the `UndeployedSafe` type with a JSDoc comment explaining its semantics and backwards-compatibility behavior.

_Allows consumers to skip DELETE backend calls for CF safes the current user did not create, since only creators have permission to delete._

**Risks:**
- Undefined treated as true means old persisted state always attempts DELETE — verify this doesn't cause spurious 40x errors on upgrade
- No runtime enforcement: consumers must consistently check isCreator before DELETE calls
- If a user creates a CF safe then logs out/in as different user, isCreator may be stale in localStorage — check hydration/reset logic

---
_Generated by prbrief • model: claude-sonnet-4-6 • 52 LLM calls • $0.76 • 321.4s_

</details>

<sub>Generated by <a href="https://popcorn.report">Popcorn Review</a>. · `safe-global/safe-wallet-monorepo@e59019686b92`</sub>

clovisdasilvaneto:
Code wise it looks good, but I need to test it with product, since counterfactual safes is a big thing.  Will update the PR once product test it

vmosharova (approved):
👍🏻

clovisdasilvaneto (approved):

katspaugh:
Replaced by #7875 (history squashed; original branch was deleted and recreated, so this PR could not be reopened).

PR #7597, opened 2026-04-08
feat(mobile): dynamic AppKit networks from CGW chain config

tmjssz:
> Hardcoded four, now fetched from source,
> chains load from CGW's course,
> bridge keeps the tree still,
> no flash, no remount chill.

## What it solves

The mobile app's AppKit (Reown/WalletConnect) was initialized at module scope with a hardcoded list of four networks (mainnet, polygon, arbitrum, sepolia). This drifted from the backend, required code changes for new chains, and limited WalletConnect sessions to only those four chains.

Resolves: https://linear.app/safe-global/issue/WA-1859/dynamically-configure-supported-chains-for-appkit-from-config-service

## How this PR fixes it

AppKit initialization is deferred from module scope into the React tree, inside `PersistGate`, so it has access to persisted CGW chain data. A new `AppKitInitializer` component reads all chains from the hydrated Redux store, converts them to Reown `Network` format via `cgwChainsToReownNetworks`, and creates the AppKit singleton. If no chains are loaded yet (first launch before CGW fetch completes), initialization is deferred entirely — no fallback networks.

Key gotcha: `WalletConnectProvider` uses a sibling bridge pattern (`WalletConnectContextBridge`) instead of wrapping children inside `AppKitProvider`. This keeps children at a stable tree position so React doesn't remount the entire UI when the AppKit instance becomes available. The bridge's `disconnect`/`open` callbacks are stabilized via refs because `useAppKit()` returns a new valtio snapshot each render, which would otherwise cause an infinite setState loop.

A `useChainSync` hook was also added to automatically switch the connected wallet's network when the user switches the active Safe's chain.

## How to test it

1. Fresh install → onboarding → import a Safe → navigate to signers → tap "Connect wallet app" → AppKit should support all CGW-supported chains (not just 4)
2. Switch the active Safe to a different chain → the connected wallet should be prompted to switch networks automatically
3. Kill and relaunch the app → WalletConnect functionality should work immediately from persisted chain data
4. Run `yarn workspace @safe-global/mobile test --testPathPattern="features/WalletConnect"` — all 76 tests pass

## Visual summary

```mermaid
flowchart TB
    subgraph Before
        R1[Redux Provider] --> WC1[WalletConnectProvider]
        WC1 --> AK1["AppKitProvider (hardcoded 4 networks)"]
        AK1 --> PG1[PersistGate]
        PG1 --> REST1[...app content]
    end

    subgraph After
        R2[Redux Provider] --> PG2[PersistGate]
        PG2 --> INIT["AppKitInitializer (dynamic CGW networks)"]
        INIT --> WC2[WalletConnectProvider]
        WC2 --> CTX["WalletConnectContext.Provider"]
        CTX --> BRIDGE["WalletConnectContextBridge (sibling)"]
        CTX --> REST2["children (stable position)"]
    end
```

## Screenshots

N/A - no UI changes

## Checklist

---

## CLA signature


tmjssz:
@claude review

apps/mobile/src/features/WalletConnect/components/AppKitInitializer/__tests__/AppKitInitializer.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  it('converts CGW chains to Reown networks', () => {
+    const chains = [{ chainId: '1' }, { chainId: '137' }]
+    mockSelectAllChains.mockReturnValue(chains)
+    mockSelectActiveChain.mockReturnValue({ chainId: '1' })
+    mockCgwChainsToReownNetworks.mockReturnValue([ethereumNetwork, polygonNetwork])
+
+    renderAppKitInitializer()
+
+    expect(mockCgwChainsToReownNetworks).toHaveBeenCalledWith(chains)
```
compojoom, line 96: Does this count as testing the conversion? We just test that `cgwChainsToReownNetworks` is being called with a chains object, but we don't test what the return value is?

apps/mobile/src/features/WalletConnect/components/AppKitInitializer/AppKitInitializer.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const chains = useAppSelector(selectAllChains)
+  const activeChain = useAppSelector(selectActiveChain)
+
+  const instanceRef = useRef<AppKitInstance | null>(null)
+
+  if (!instanceRef.current) {
+    const reownNetworks = cgwChainsToReownNetworks(chains)
+
+    if (reownNetworks.length > 0) {
+      const networks = reownNetworks as [(typeof reownNetworks)[0], ...typeof reownNetworks]
```
compojoom: why do we need this type conversion?

compojoom:
I spent some time looking at the `useEffect` patterns in the WalletConnect hooks and I think there's a simpler way to do this.

## The ref flag dance

Both `useImportSignerFlow` and `useReconnectFlow` follow the same structure. The user taps connect, you set a ref flag, then a `useEffect` watches `useAccount()` and `useWalletInfo()` for changes. When all the values are truthy and the flag is set — do the actual work, then reset the flag.

This works, but it's hard to follow. The cause (user tapped connect) is disconnected from the handler (validate and navigate) by a chain of reactive dependencies. When you read the code you have to mentally trace: "ok, `initiateConnection` sets the ref, then I need to find the effect that checks it, then figure out which combination of deps triggers it." It's the ["set flag → effect runs → reset flag" pattern](https://x.com/alvinsng/status/2033969062834045089) that the React community has been flagging as a code smell.

## AppKit already has an event system

While looking through `@reown/appkit-react-native` I found that it exports `useAppKitEventSubscription` — a hook that lets you subscribe to specific events like `CONNECT_SUCCESS`, `CONNECT_ERROR`, `DISCONNECT_SUCCESS` etc. The `CONNECT_SUCCESS` event carries the connected address and wallet name directly in the payload:

```typescript
{
  event: 'CONNECT_SUCCESS',
  address?: string,
  properties: {
    name: string,
    caipNetworkId?: string,
    reconnect?: boolean, // even distinguishes reconnect from fresh connect
  }
}
```

This means we don't need to watch `useAccount()` and `useWalletInfo()` reactively and wait for them to align. The event tells us "connection just happened" with all the data we need.

## What useImportSignerFlow could look like

Instead of two `useEffect`s, two refs, and three reactive hooks — subscribe to the event:

```typescript
useStableAppKitEvent('CONNECT_SUCCESS', (event) => {
  if (\!connectInitiatedRef.current) return
  connectInitiatedRef.current = false

  const address = event.data.address
  if (\!address) return

  const checksumAddress = getAddress(address)
  // validate ownership, navigate to name-signer or error screen
})
```

`useAccount()` and `useWalletInfo()` go away entirely. The address comes from the event. The wallet name comes from `event.data.properties.name`. The wallet icon (needed for the error screen route param) can be read synchronously from `ConnectionsController.state.walletInfo?.icon` inside the callback.

The `connectInitiatedRef` stays — it still has a job: distinguishing "we opened the modal" from "some background reconnection happened." But now it's checked in an event callback, not used as a trigger for a reactive effect.

The reset effect (the one that clears `registeredRef` on disconnect) disappears too. Each `CONNECT_SUCCESS` fires exactly once per connection — no deduplication needed.

## Same story for useReconnectFlow

Same pattern, even simpler since there's no async work:

```typescript
useStableAppKitEvent('CONNECT_SUCCESS', (event) => {
  if (\!pendingAddressRef.current) return

  const reconnectAddress = getAddress(pendingAddressRef.current)
  pendingAddressRef.current = null

  if (\!sameAddress(reconnectAddress, event.data.address)) {
    disconnect()
    router.push({ pathname: '/import-signers/reconnect-error', params: { address: reconnectAddress } })
    return
  }

  switchNetworkIfNeeded()
})
```

Drops `useAccount()` and `useWalletInfo()` entirely. The address comparison uses the event payload instead of reactive state.

## One small utility to make this work

`useAppKitEventSubscription` re-subscribes every time the callback identity changes. Since a lot of our deps are valtio snapshots that get a new reference on every render, we need a thin wrapper to stabilize the callback:

```typescript
function useStableAppKitEvent(event: EventName, callback: (e: EventsControllerState) => void) {
  const callbackRef = useRef(callback)
  callbackRef.current = callback

  const stable = useCallback((e: EventsControllerState) => callbackRef.current(e), [])
  useAppKitEventSubscription(event, stable)
}
```

Write it once, and every hook that uses AppKit events stays clean.

## What I'd leave as is

`useChainSync` reacts to Redux state (the active Safe's chain changes), not a wallet event. The effect is the right tool there. The `WalletConnectContextBridge` state sync is an architectural constraint worth revisiting separately.

Not blocking — the current code works. But the event-driven approach would make these two flows a lot easier to follow.

apps/mobile/src/features/WalletConnect/components/AppKitInitializer/__tests__/AppKitInitializer.test.tsx
tmjssz: The conversion logic is tested [here](https://github.com/safe-global/safe-wallet-monorepo/blob/mobile/dynamic-appkit-networks/apps/mobile/src/features/WalletConnect/utils/__tests__/chains.test.ts), this test file only verifies the orchestration (correct input passed in, correct output forwarded to createAppKitInstance)

compojoom (approved):

PR #7608, opened 2026-04-09
fix: guard against NaN/float in gas oracle BigInt conversion

compojoom:
> Oracle guards now catch the NaN,
> Math.round tames the float,
> gas price errors handled with grace,
> no more BigInt bloat.

## What it solves

The gas oracle parsing in `useDefaultGasPrice` crashes with "The number NaN cannot be converted to a BigInt" when oracle responses contain undefined/invalid values, or when floating-point multiplication produces non-integer results before `BigInt()` conversion. This causes repeated console errors every 20 seconds during gas price polling.

Related: [WA-1920](https://linear.app/safe-global/issue/WA-1920/twap-order-flow-glitches-on-web-with-pending-signer) — this fix addresses the NaN console errors observed during investigation, but does not resolve the UI flickering issue itself.

## How this PR fixes it

Adds `Number.isFinite()` validation before `BigInt()` conversion and `Math.round()` to handle floating-point precision in both Etherscan and non-Etherscan oracle parsing paths. Invalid values now throw descriptive errors that flow through the existing oracle fallback chain (try next oracle → try fixed config → surface error to hook consumer) instead of crashing with an opaque TypeError.

Also adds a comprehensive test suite for `useDefaultGasPrice` (27 tests) since none existed previously.

## How to test it

1. Run the new tests: `cd packages/utils && npx jest --testPathPattern='useDefaultGasPrice' --no-watchman`
2. Run the existing web gas price tests: `cd apps/web && npx jest --testPathPattern='useGasPrice' --no-watchman`
3. Verify gas fee estimation works on the transaction confirmation screen (any chain)

## Visual summary

```mermaid
flowchart LR
    A[Oracle response] --> B{Valid number?}
    B -->|Yes| C[Math.round + BigInt]
    B -->|No / NaN / Infinity| D[Throw descriptive error]
    D --> E[Caught by fallback loop]
    E --> F[Try next oracle or fixed config]
```

## Checklist

---

## CLA signature

### DIFF
--- a/packages/utils/src/hooks/__tests__/useDefaultGasPrice.test.ts
+++ b/packages/utils/src/hooks/__tests__/useDefaultGasPrice.test.ts
@@ -1,0 +1,435 @@
+import { renderHook, waitFor } from '@testing-library/react'
+import { type FeeData } from 'ethers'
+import { useDefaultGasPrice, getTotalFee, getTotalFeeFormatted } from '@safe-global/utils/hooks/useDefaultGasPrice'
+import type { Chain } from '@safe-global/store/gateway/AUTO_GENERATED/chains'
+
+const mockGetFeeData = jest.fn()
+const mockFetchFn = jest.fn()
+
+jest.mock('@safe-global/utils/hooks/useIntervalCounter', () => ({
+  useIntervalCounter: jest.fn(() => [0, jest.fn()]),
+}))
+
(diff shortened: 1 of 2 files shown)


compojoom:
@claude review

vmosharova (approved):
👍🏻

LucieFaire (approved):

PR #7612, opened 2026-04-09
refactor(mobile): event-driven WalletConnect hooks

tmjssz:
Replace useEffect-based reactive patterns in useImportSignerFlow and useReconnectFlow with direct AppKit event subscriptions.

- Extract useStableAppKitEvent: ref-stabilized wrapper around useAppKitEventSubscription that filters by event name to prevent cross-event contamination from valtio proxy mutations
- Extract useOnAppKitConnect: shared hook subscribing to CONNECT_SUCCESS, CONNECT_ERROR, and USER_REJECTED with typed payload
- Refactor useImportSignerFlow and useReconnectFlow to use event callbacks instead of useEffect watching reactive state
- Disconnect before opening the connect modal to prevent stale CONNECT_SUCCESS replays from previously connected wallets

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/WalletConnect/hooks/useImportSignerFlow.ts
+++ b/apps/mobile/src/features/WalletConnect/hooks/useImportSignerFlow.ts
@@ -1,88 +1,99 @@
-import { useCallback, useEffect, useRef } from 'react'
+import { useCallback, useRef } from 'react'
 import { router } from 'expo-router'
 import { getAddress } from 'ethers'
 import { useAccount, useAppKit, useWalletInfo } from '@reown/appkit-react-native'
 import { useAddressOwnershipValidation } from '@/src/hooks/useAddressOwnershipValidation'
 import Logger from '@/src/utils/logger'
 import { useSwitchNetwork } from './useSwitchNetwork'
+import { useOnAppKitConnect } from './useOnAppKitConnect'
 
 /**
  * Handles the signer import flow: ownership validation and navigation
  * after a new wallet connects via WalletConnect.
+ *
+ * Uses CONNECT_SUCCESS event subscription instead of useEffect to respond
+ * directly to connection events. The connectInitiatedRef guard ensures
+ * only user-initiated connections (via initiateConnection()) trigger the
+ * flow — auto-reconnects and connections from useReconnectFlow are ignored.
  */
 export function useImportSignerFlow() {
   const { open, disconnect } = useAppKit()
-  const { address, isConnected: walletIsConnected } = useAccount()
+  const { address: accountAddress, isConnected: walletIsConnected } = useAccount()
   const { walletInfo } = useWalletInfo()
   const { switchNetworkIfNeeded } = useSwitchNetwork()
   const { validateAddressOwnership } = useAddressOwnershipValidation()
 
-  const registeredRef = useRef<string | null>(null)
   const connectInitiatedRef = useRef(false)
 
-  const isConnected = Boolean(walletIsConnected && address && walletInfo)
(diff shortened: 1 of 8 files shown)


tmjssz:
@claude review

apps/mobile/src/features/WalletConnect/hooks/useImportSignerFlow.ts
```diff
@@ -1,88 +1,99 @@
...
+        Logger.error('Unexpected error in validateAndNavigate:', error)
+      })
+    },
+    () => {
+      connectInitiatedRef.current = false
+    },
+  )
+
+  const initiateConnection = useCallback(async () => {
+    await disconnect()
```
compojoom, line 93: The old version didn't do disconnect here? Why is it necessary now?

compojoom (approved):

apps/mobile/src/features/WalletConnect/hooks/useOnAppKitConnect.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Subscribes to the AppKit CONNECT_SUCCESS event (with typed payload)
+ * and resets state on CONNECT_ERROR / USER_REJECTED.
+ *
+ * @param onSuccess  Called with the narrowed CONNECT_SUCCESS event data.
+ * @param onFailure  Called when the connection fails or is rejected by the user.
+ *                   Use this to reset ref guards so they don't leak.
+ */
+export function useOnAppKitConnect(onSuccess: (event: ConnectSuccessEvent) => void, onFailure: () => void) {
+  useStableAppKitEvent('CONNECT_SUCCESS', (state) => {
+    onSuccess(state.data as ConnectSuccessEvent)
```
compojoom: why is the cast needed? Why is ts failing to properly infer the type?

apps/mobile/src/features/WalletConnect/hooks/useReconnectFlow.ts
```diff
@@ -1,49 +1,61 @@
...
-    if (!sameAddress(reconnectAddress, address)) {
-      disconnect()
+      const connectedAddress = eventData.address
 
-      router.push({
-        pathname: '/import-signers/reconnect-error',
-        params: { address: reconnectAddress },
-      })
+      if (!connectedAddress || !sameAddress(reconnectAddress, connectedAddress)) {
+        disconnect()
```
compojoom, line 35: await disconnect?

apps/mobile/src/features/WalletConnect/hooks/useImportSignerFlow.ts
tmjssz: Calling `open()` in line 84 somehow replays the CONNECT_SUCCESS event for the previous connection. The callback then runs with the old wallet's address before the user even picks a new wallet. Calling disconnect here makes sure that the stale event is not triggered again.

apps/mobile/src/features/WalletConnect/hooks/useReconnectFlow.ts
tmjssz: Not necessary here because we just redirect to the error page, the disconnect can happen in parallel.

apps/mobile/src/features/WalletConnect/hooks/useOnAppKitConnect.ts
tmjssz: Because appkit's `useAppKitEventSubscription` does not return the narrowed type.
Refactored our `useStableAppKitEvent` hook for more type-safety -> 7332750aa5bfd3b71b8c84e89c48b6e7d6b724a3

PR #7616, opened 2026-04-10
fix: allow nonce edition in tx-builder flow

LucieFaire:
## What it solves

Root cause: `TxNonceForm` used `isRejectionTx(safeTx)` to detect rejection transactions by inspecting the transaction data (data === '0x' and value === '0'). This check falsely matched transactions created through the Transaction Builder with Proxy ABI (that happen to have empty calldata) or the custom data with empty `hexData`.

Resolves: [WA-1650](https://linear.app/safe-global/issue/WA-1650/bug-transaction-builder-silently-overrides-manually-set-nonce-on)

## How this PR fixes it

Replaced the `isRejectionTx(safeTx)` data check with `isRejection` from `TxFlowContext`, which is an explicit flag set only by actual rejection flows. This way, only intentional rejection transactions lock the nonce — transactions from the Transaction Builder (or any other source) with empty data remain editable.

Files changed:

* TxNonce/index.tsx — import TxFlowContext, use isRejection from context instead of isRejectionTx(safeTx)
* TxNonce.test.tsx — wrap renders with TxFlowContext.Provider, add two new tests for rejection flow behavior

## How to test it
1. Reject tx stays as it was - no editing of nonce
2. Tx Builder `custom data` tx with `0x` - nonce is editable
3. Tx Builder proxy ABI call with value 0 - nonce is editable

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/tx-flow/common/TxNonce/__tests__/TxNonce.test.tsx
+++ b/apps/web/src/components/tx-flow/common/TxNonce/__tests__/TxNonce.test.tsx
@@ -1,6 +1,7 @@
 import { render, screen } from '@/tests/test-utils'
 import TxNonce from '../index'
 import { SafeTxContext, type SafeTxContextParams } from '@/components/tx-flow/SafeTxProvider'
+import { TxFlowContext, initialContext as initialTxFlowContext } from '@/components/tx-flow/TxFlowProvider'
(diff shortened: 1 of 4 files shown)


compojoom:
@claude review

LucieFaire:
@claude review based on the previous review

vmosharova (approved):
👍🏻

PR #7642, opened 2026-04-14
refactor(mobile): revert event-driven handling for WC and  promise-based

tmjssz:
> Promises replace refs and effects,
> connect resolves, errors reject,
> flows simplified clean.

Revert the revert event-driven handling for WC connections because it was not working properly with all wallets.

Instead change to promise-based approach.

## What it solves

`useImportSignerFlow` and `useReconnectFlow` duplicated the same connection pattern: arm a ref guard, call `open()`, detect connection via `useEffect`, and reset on error/rejection. This made the code hard to follow and error-prone (e.g., the `isConnected` guard in NameSigner was sourced from `useImportSignerFlow` — removing it from the return type broke navigation silently).

## How this PR fixes it

Extracts a shared `useConnect` hook that returns a promise-based `connect()` function. Internally it uses `useEffect` on `isConnected`/`address`/`walletInfo` to resolve and `useStableAppKitEvent` for `CONNECT_ERROR`/`USER_REJECTED` to reject. Both flow hooks now just `await connect()` and handle the result — no more ref guards, no more manual event subscriptions.

**Gotcha:** `isConnected` was previously derived from `useImportSignerFlow`'s return value and exposed via the WC context. Since `useImportSignerFlow` no longer returns it, `isConnected` is now sourced directly from `useAccount()` in the context bridge. The `NameSigner` disconnect guard (`useEffect` that called `router.back()` when `!isConnected`) was removed since the promise-based flow handles the lifecycle.

## How to test it

1. Import a Safe → tap "Connect wallet app" → connect a wallet that is an owner → verify navigation to the name-signer screen
2. Connect a wallet that is NOT an owner → verify navigation to the connect-signer-error screen
3. From the signers screen, tap reconnect on a signer → connect the matching wallet → verify it stays on screen
4. Reconnect with a different wallet → verify navigation to reconnect-error
5. Reject or cancel during connection → verify the flow resets cleanly with no navigation

## Visual summary

```mermaid
flowchart TB
  subgraph Before
    A1[useImportSignerFlow] --> B1[ref guard + useEffect + useStableAppKitEvent]
    A2[useReconnectFlow] --> B2[ref guard + useEffect + useStableAppKitEvent]
  end
  subgraph After
    C[useConnect hook] --> D["connect() → Promise&lt;ConnectResult&gt;"]
    E1[useImportSignerFlow] --> F1["await connect()"]
    E2[useReconnectFlow] --> F2["await connect()"]
    F1 --> D
    F2 --> D
  end
```

## Screenshots

N/A - no UI changes

## Checklist

---

## CLA signature


tmjssz:
@claude review

tmjssz:
@claude review again pls. I addressed your coments:
1. d29b60c5e — walletInfo race condition: guard checks name and icon individually 
2. bc115fa1f — typed error classes: ConnectError/UserRejectedError replace string matching 
3. 8cd5fc8b4 — unmount cleanup: pendingRef cleared on unmount

clovisdasilvaneto (approved):

compojoom (approved):

PR #7658, opened 2026-04-15
feat(mobile): add WalletConnect transaction execution support

tmjssz:
> WC signers could sign but not execute,
> now the external wallet takes the call,
> threshold met, transaction sent.

## What it solves

WalletConnect signers can confirm (sign) transactions but cannot execute them on-chain. The execution method falls back to private key signing, which fails because there is no local key available.

Resolves: https://linear.app/safe-global/issue/WA-1965/execute-transactions-with-walletconnect-signer

## How this PR fixes it

Adds a new `WITH_WC` execution method that builds the Safe transaction with all required signatures, encodes the `execTransaction` call, and sends it to the connected external wallet via `eth_sendTransaction`. The external wallet handles gas estimation, signing, and broadcasting.

Key decisions:
- WalletConnect execution skips the biometrics flow since there is no local key to protect
- The execute button is wrapped with `WalletConnectGate` to validate the session before allowing execution
- Uses `sameAddress` for case-insensitive owner comparison when adding pre-validated signatures
- Wallet nonce is fetched before sending the transaction to correctly track the pending tx

## How to test it

1. Connect an external wallet via WalletConnect (e.g. Zerion, MetaMask)
2. Create or find a transaction that has enough confirmations to execute
3. Open the "How to execute" sheet and select the WalletConnect signer
4. Tap execute — verify the external wallet receives the transaction request and can execute it
5. After execution, verify the transaction appears in pending transactions with the correct tx hash

## Screenshots

N/A — this is a logic change; the UI reuses existing components.

## Checklist

---

## CLA signature

EOF

### DIFF
--- a/apps/mobile/src/features/ConfirmTx/components/PendingTxInfo/PendingTxInfo.tsx
+++ b/apps/mobile/src/features/ConfirmTx/components/PendingTxInfo/PendingTxInfo.tsx
@@ -19,7 +19,7 @@ export const PendingTxInfo = ({ createdAt, pendingTx }: { createdAt: number | nu
         </View>
       )}
 
-      {pendingTx?.type === ExecutionMethod.WITH_PK && (
+      {(pendingTx?.type === ExecutionMethod.WITH_PK || pendingTx?.type === ExecutionMethod.WITH_WC) && (
         <View alignItems="center" flexDirection="row" justifyContent="space-between">
           <Text color="$textSecondaryLight">Transaction hash</Text>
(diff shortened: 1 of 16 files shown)


tmjssz:
@claude review

tmjssz:
@claude review again pls. Regarding your comments:
1. what do you mean? 
2. Updated -> 1512443c5cb87f5758d671ca3779c712fe43e429
3. This is not really needed

compojoom (approved):

PR #7669, opened 2026-04-16
feat(web): add SelectSafeModal for space-level actions

clovisdasilvaneto:
> A modal rises when actions call,
> pick your Safe, then send or swap,
> spaces route each flow with care.

## What it solves

In Spaces, action buttons (Send, Receive, Swap, Build Transaction) need a way for users to select which Safe to operate on before starting a transaction flow. Previously, these actions only worked at the Safe level where the Safe context was already known.

### Linear task
https://linear.app/safe-global/issue/WA-1827/spaces-actions-tray-component-account-modal

## How this PR fixes it

- **New `SelectSafeModal` component**: A dialog that lists all Safes in the current Space, with search/filter support. When a user clicks an action button from a Space-level route, the modal opens so they can pick a target Safe.
- **New `safeActionsModalSlice`** (Redux): Manages modal open/close state and the selected action type (`ESafeAction` enum: Send, Receive, Swap, BuildTransaction).
- **`useSafeActionMapper` hook**: Maps each action type to the correct navigation/flow after a Safe is selected (e.g., Send opens `NewTxFlow`, Swap navigates to `/swap`, Receive opens QR modal).
- **Updated `ActionsTray`**: Space-level buttons now dispatch `openSafeActionsModal` instead of directly navigating or opening flows.
- **Updated `NavigateToSection`**: Global search "Navigate to" items now handle Space routes by opening the SelectSafeModal for actions and navigating to accounts for the "Accounts" item.
- **Extracted reusable hooks**: `useMatchSafe` and `useSearchFilter` moved from `global-search` to `src/hooks/` so both GlobalSearch and SelectSafeModal can share them.
- **Spaces feature contract/feature.ts updated**: `SelectSafeModal` registered as a lazy-loaded component via the feature architecture.

## How to test it

1. Navigate to a Space route
2. Click Send, Receive, Swap, or Build Transaction in the actions tray
3. The SelectSafeModal should open with the correct title (e.g., "Send from")
4. Search for a Safe by name or address
5. Click a Safe — the appropriate flow should open with that Safe selected
6. Open Global Search (Cmd+K) from a Space route and click a "Navigate to" action — modal should open
7. Verify that non-Space routes still work as before (direct navigation, no modal)

## Visual summary

image

```mermaid
flowchart LR
    A[ActionsTray button click] --> B{Is Space route?}
    B -->|Yes| C[dispatch openSafeActionsModal]
    C --> D[SelectSafeModal opens]
    D --> E[User picks a Safe]
    E --> F[useSafeActionMapper]
    F --> G[Navigate + open flow]
    B -->|No| H[Direct flow / navigation]
```

## Checklist

---

## CLA signature

With the submission of this Pull Request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://safe.global/cla)

### DIFF
--- a/apps/web/src/features/global-search/hooks/useMatchSafe.ts
+++ b/apps/web/src/features/global-search/hooks/useMatchSafe.ts
@@ -1,44 +1,4 @@
-import { useCallback, useMemo } from 'react'
-import { isMultiChainSafeItem, type AllSafeItems } from '@/hooks/safes'
-import { useAppSelector } from '@/store'
-import { selectAllAddressBooks } from '@/store/addressBookSlice'
-import useChains from '@/hooks/useChains'
-
 /**
- * Returns a memoized callback that matches a Safe by address, name,
- * or chain-prefixed address (e.g. "matic:0xf452…").
- * Falls back to addressBooks for the name when the Safe item has no name.
+ * @deprecated Import from '@/hooks/useMatchSafe' instead.
  */
-const useMatchSafe = () => {
-  const addressBooks = useAppSelector(selectAllAddressBooks)
-  const { configs } = useChains()
-
-  const chainIdToShortName = useMemo(
-    () => new Map(configs.map((chain) => [chain.chainId, chain.shortName.toLowerCase()])),
-    [configs],
-  )
-
-  return useCallback(
-    (safe: AllSafeItems[number], q: string): boolean => {
-      const address = safe.address.toLowerCase()
-      const safeName =
-        safe.name ?? addressBooks[isMultiChainSafeItem(safe) ? safe.safes[0].chainId : safe.chainId]?.[safe.address]
-
-      if (address.includes(q) || (safeName?.toLowerCase().includes(q) ?? false)) {
-        return true
-      }
-
-      // Match chain-prefixed addresses (e.g. "matic:0xf452…" or just "matic:")
-      const chainIds = isMultiChainSafeItem(safe) ? safe.safes.map((s) => s.chainId) : [safe.chainId]
-      return chainIds.some((chainId) => {
-        const shortName = chainIdToShortName.get(chainId)
-        if (!shortName) return false
-        const prefixed = `${shortName}:${address}`
-        return prefixed.includes(q)
-      })
-    },
-    [addressBooks, chainIdToShortName],
-  )
-}
-
-export default useMatchSafe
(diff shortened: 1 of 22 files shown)


vmosharova:
@claude review

apps/web/src/features/spaces/components/SelectSafeModal/useSafeActionMapper.tsx
```diff
@@ -0,0 +1,81 @@
...
+          pathname: txBuilderLink.pathname,
+          query: { ...(txBuilderLink.query as Record<string, string>), safe: safeParam },
+        })
+      },
+    }),
+    [router, setTxFlow, getSafeQueryParam, navigateToSafe, onReceiveComplete, txBuilderLink],
+  )
+}
+
+export default useSafeActionMapper
```
vmosharova, line 81: Let's cover all new components and hooks with unit tests :)

apps/web/src/features/global-search/hooks/useMatchSafe.ts
```diff
@@ -1,44 +1,4 @@
...
-import { isMultiChainSafeItem, type AllSafeItems } from '@/hooks/safes'
-import { useAppSelector } from '@/store'
-import { selectAllAddressBooks } from '@/store/addressBookSlice'
-import useChains from '@/hooks/useChains'
-
 /**
- * Returns a memoized callback that matches a Safe by address, name,
- * or chain-prefixed address (e.g. "matic:0xf452…").
- * Falls back to addressBooks for the name when the Safe item has no name.
+ * @deprecated Import from '@/hooks/useMatchSafe' instead.
```
vmosharova, line 2: Why replacing the file content with `@deprecated` and re-export instead of removing the file and updating dependencies?

vmosharova (changes requested):
Nice! Would be great to increase unit test coverage for the new logic :)
Also, 1 question about export - maybe it makes sense to remove the deprecated file and update dependencies to reduce legacy code, up to you:

apps/web/src/features/global-search/hooks/useMatchSafe.ts
clovisdasilvaneto: yep, no sense tbh. might be from claude, will change it

vmosharova (approved):
👍🏻

Issue #3026, opened 2023-12-18
Optionality of risk warnings and non-default compliance warnings


johannesmoormann:
## What is the feature about

While risk warning by a third-party provider should be enabled by default to protect the user, the option should exist in the settings to deactivate that feature for privacy or other reasons. Additionally, country/region specific compliance warnings should be optional and non-default. 

## The list of requirements

toggle for risk checks (default: on)
toggle for compliance warnings (default: off)

tbd with third party

MicahZoltu:
While I do believe that for privacy reasons these third party providers should be opt-in, or at least disablable, that isn't my main concern.  The bigger issue IMO is that these providers are incorrectly labeling some accounts as a compliance risk.  As a community, we should be advocating for broad usage of crypto and encouraging the usage of privacy software whenever it is legal for the user to do so.  At the moment, the currently selected providers are discouraging interactions between users where the interaction would be entirely legal.

Specifically, there are two problems:
1. Tornado.cash is *only* illegal to use in the US (I believe).  In the rest of the world it is still legal to use.
2. It is perfectly legal to interact with an address that has used Tornado.cash, even in the US.

Redefine is warning users about a compliance risk when they try to send assets to an address that has used Tornado.cash in the past, and this is just fear mongering that is discouraging people from using privacy software.  Either such warnings should be filtered out of the UI (as in #3015), or Gnosis should convince Redefine that their warning is in appropriate and needs to be removed, or Gnosis should stop using Redefine as a provider.  My vote is the third option as I think the crypto community should be actively boycotting providers like Redefine who are spreading FUD like this, but either of the other two would be reasonable solutions as well.

Note: It is important that the fix for this isn't just something people can toggle off.  The problem here is that people trust Gnosis to give informed recommendations when surfacing risks, and the majority of people will blindly follow whatever risk warnings the Gnosis UI shows.  For example, if I am an employee of a company that uses Gnosis SAFE and I use Tornado to protect my privacy, when my employer (who may not understand the current financial privacy legal landscape) tries to send me money they will get a warning telling them that I am a compliance risk.  The company may choose to simply let me go as it isn't worth it to them to take on compliance risk.  They aren't necessarily going to dig into the issue and understand that this is a false positive, they'll do exactly what banks do today and simply kick me out and hire someone else.

[...]

PR #7671, opened 2026-04-16
feat: private contacts UI with request-to-add workflow

katspaugh:
> Members craft contacts in quiet,
> request they join the shared book,
> admins decide their fate.

Resolves https://linear.app/safe-global/issue/WA-1973/redesign-spaces-address-book-to-match-prototype

## Summary

- Non-admin members can create **private contacts** via a new "Add contact" button
- Private contacts show a **"Request to add"** button in the My contacts tab
- New **"Pending" tab** shows submitted requests with admin **accept/reject** actions
- Contact merging updated: **space > private > local** priority during transactions

## What it solves

Only admins could add contacts to the space address book. Members had no way to manage contacts or propose additions. This PR adds the full frontend for private contacts and the request-to-add approval workflow.

### New files
| File | Purpose |
|------|---------|
| `privateAddressBookApi.ts` | Stub RTK Query endpoints for 7 new CGW APIs (will be auto-generated once CGW deploys) |
| `useGetPrivateAddressBook.ts` | Data hook for fetching private contacts |
| `useGetAddressBookRequests.ts` | Data hook for fetching pending requests |
| `AddPrivateContact.tsx` | Dialog for creating private contacts (available to all members) |
| `RequestToAddButton.tsx` | Button to submit a private contact for admin approval |
| `PendingRequestsTable.tsx` | Table with accept/reject buttons for admins, pending badge for members |

### Modified files
| File | Change |
|------|--------|
| `SpaceAddressBook/index.tsx` | Added Pending tab, private contacts in My contacts, non-admin add button |
| `SpaceAddressBookTable.tsx` | Added `isPrivate` flag, `renderExtraAction` prop for request-to-add column |
| `useAllAddressBooks.ts` | Added `ContactSource.private`, 3-way merge priority |
| `spaces/index.ts` | Exported new hooks |

## How to test it

- [ ] Non-admin member sees "Add contact" button (creates private contact)
- [ ] Private contacts appear in My contacts with "Request to add" button
- [ ] Clicking "Request to add" creates a pending request
- [ ] Pending tab shows requests (admin sees all, member sees own)
- [ ] Admin can accept (adds to shared book) or reject
- [ ] During transactions, private contacts resolve in autocomplete (space contacts take priority)

**Depends on:** safe-global/safe-client-gateway#3026 (backend endpoints)

## Visual summary

```mermaid
flowchart LR
    subgraph "My contacts tab"
        A[Private contact] -->|"Request to add"| B[Pending tab]
    end
    subgraph "Pending tab"
        B -->|Admin accepts| C[Workspace contacts]
        B -->|Admin rejects| D[Stays private]
    end
    subgraph "Transaction resolution"
        E[Space contacts] -->|priority 1| F[Merged]
        G[Private contacts] -->|priority 2| F
        H[Local contacts] -->|priority 3| F
    end
```

## Checklist

- [x] Type-check passes
- [x] Prettier formatted
- [x] Pre-commit hooks pass
- [x] No changes to existing address book behavior


katspaugh:
### Code review

Found 1 issue:

1. `AddPrivateContact.tsx` is a near-duplicate of `AddContact.tsx` — same form layout (`NameInput`/`AddressInput`/`NetworkMultiSelectorInput`), same state handling, same dialog structure, same loading/error flow. Differences are the mutation hook, label, and an extra description. [CLAUDE.md](https://github.com/safe-global/safe-wallet-monorepo/blob/85fc3233398a787f86c2ed585397a8608da8f7e5/CLAUDE.md) says "Follow the DRY principle – avoid code duplication by extracting reusable functions, hooks, and components". Since `AddContact` already accepts a `label` prop in this PR, extending it to also accept the mutation/body would avoid the duplication.

https://github.com/safe-global/safe-wallet-monorepo/blob/85fc3233398a787f86c2ed585397a8608da8f7e5/apps/web/src/features/spaces/components/SpaceAddressBook/AddPrivateContact.tsx#L1-L160

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

katspaugh:
## PR Briefing

[Watch the PR walkthrough](https://github.com/5afe/wallet-pr-artifacts/releases/download/prbrief-safe-global-safe-wallet-monorepo-pr-7671-00d337eb9cdc/video-github.mp4)

▶ **[Watch the walkthrough video](https://github.com/5afe/wallet-pr-artifacts/releases/download/prbrief-safe-global-safe-wallet-monorepo-pr-7671-00d337eb9cdc/video-github.mp4)** (opens MP4 in a new tab)

This PR adds a private contacts system and request-to-add approval workflow to the Spaces address book feature. Non-admin members can now create private contacts via a new `AddPrivateContact` dialog, then submit them for admin approval via `RequestToAddButton`, which appear in a new 'Pending' tab rendered by `PendingRequestsTable`. Admins can accept or reject requests, promoting approved contacts to the shared workspace address book. The `useAllAddressBooks` hook is updated to implement a three-tier merge priority (space > private > local) so private contacts resolve in transaction autocomplete. Seven new CGW API endpoints are defined in `schema.json` and wired up in the auto-generated `spaces.ts` RTK Query file, with stub hooks `useGetPrivateAddressBook` and `useGetAddressBookRequests` providing data access.

<details>

## Summary

This PR adds a private contacts system and request-to-add approval workflow to the Spaces address book feature. Non-admin members can now create private contacts via a new `AddPrivateContact` dialog, then submit them for admin approval via `RequestToAddButton`, which appear in a new 'Pending' tab rendered by `PendingRequestsTable`. Admins can accept or reject requests, promoting approved contacts to the shared workspace address book. The `useAllAddressBooks` hook is updated to implement a three-tier merge priority (space > private > local) so private contacts resolve in transaction autocomplete. Seven new CGW API endpoints are defined in `schema.json` and wired up in the auto-generated `spaces.ts` RTK Query file, with stub hooks `useGetPrivateAddressBook` and `useGetAddressBookRequests` providing data access.

## Architecture notes

The change touches four layers: the OpenAPI schema (`schema.json`) and auto-generated RTK Query client (`AUTO_GENERATED/spaces.ts`) define the contract with CGW endpoints that are not yet deployed; new data hooks (`useGetPrivateAddressBook`, `useGetAddressBookRequests`) wrap those queries with auth/space-ID guards; UI components (`AddPrivateContact`, `RequestToAddButton`, `PendingRequestsTable`, `AddToWorkspaceButton`, `RemoveDuplicateButton`) implement the workflow; and `useAllAddressBooks` integrates private contacts into the existing merge pipeline with a new `ContactSource.private` enum value. `SpaceAddressBookTable` is made composable via a `renderExtraAction` prop so the parent `SpaceAddressBook/index.tsx` can inject the correct action button per entry type. The auto-generated file was hand-edited rather than regenerated, creating a drift risk if the generator is re-run.

## Ripple risks

Analyzed 3 changed exports. No untouched external callers found — all usages appear to be inside the PR diff.

_Analyzed 14 file(s), 20 changed export(s), 0 uncovered callsite(s) in 7.8s._

## Walkthrough outline

_19 steps across 7 chapters, approximately 4 min 15s._

### Why this change

1. **motivation** · `apps/web/src/features/spaces/components/SpaceAddressBook/index.tsx` L1-30 `SpaceAddressBook`
   Before this PR, only admins could add contacts to a space address book — members were locked out entirely. This file is the hub of the whole feature, and it's where you'll see the biggest structural change: a new Pending tab and a non-admin Add Contact button.

2. **motivation** · `apps/web/src/features/spaces/components/SpaceAddressBook/index.tsx` L1-30
   Here's the full picture. Members create private contacts, request promotion to the shared book, and admins accept or reject. Meanwhile the three-tier merge — space beats private beats local — means private contacts still resolve in transaction autocomplete even before approval.

### The API contract

3. **change** · `packages/store/scripts/api-schema/schema.json` L1-50 `AddressBookRequestItemDto`
   Everything starts with the OpenAPI schema. Seven new paths are hand-authored here — GET and PUT for private contacts, DELETE by address, and GET, POST, approve, reject for requests. This is the contract the frontend is building against, even though the CGW backend PR isn't merged yet.

4. **change** · `packages/store/src/gateway/AUTO_GENERATED/spaces.ts` L323-370 `AddressBooksGetPrivateItemsV1ApiArg`
   The auto-generated RTK Query file gets seven new endpoint definitions wired up here — types for every arg and response shape. Notice the file is named AUTO_GENERATED but was hand-edited, which means a generator re-run would silently wipe all of this.

5. **change** · `packages/store/src/gateway/AUTO_GENERATED/spaces.ts` L481-500 `AddressBookRequestItemDto`
   Down here are the three new DTO types: AddressBookRequestItemDto, AddressBookRequestsDto, and CreateAddressBookRequestDto. Worth noting that CreateAddressBookRequestDto only carries an address — name and chainIds are expected to be resolved server-side from the private contact record.

### New data hooks

6. **change** · `apps/web/src/features/spaces/hooks/useGetPrivateAddressBook.ts` L1-25 `useGetPrivateAddressBook`
   This hook wraps the private address book query with two guards: skip if unauthenticated, skip if no space ID. It returns an empty array when data is absent, which means callers can't distinguish loading from genuinely empty — something to watch.

7. **change** · `apps/web/src/features/spaces/index.ts` L1-10
   Both new hooks are re-exported from the spaces barrel file so consumers import from one stable path. Short step, but it's the public API surface for everything we just saw.

### The new UI components

8. **change** · `apps/web/src/features/spaces/components/SpaceAddressBook/AddPrivateContact.tsx` L1-60 `AddPrivateContact`
   AddPrivateContact is the dialog non-admin members see when they click Add Contact. It collects name, address, and networks, then fires upsertPrivateItemsV1. One gotcha: the NetworkMultiSelectorInput's onChange is never wired to the Controller, so network selection may silently not update form state.

9. **change** · `apps/web/src/features/spaces/components/SpaceAddressBook/RequestToAddButton.tsx` L1-60 `RequestToAddButton`
   Once a private contact exists, this button handles the two-step promotion: optionally upsert it as a private contact, remove it from local store, then call createRequest. If the upsert succeeds but createRequest fails, the contact is already gone from local with no rollback path.

10. **change** · `apps/web/src/features/spaces/components/SpaceAddressBook/PendingRequestsTable.tsx` L1-70 `PendingRequestsTable`
   PendingRequestsTable is the new Pending tab's content. Admins see approve and reject buttons per row; members see a Pending badge. A single loadingId state prevents concurrent actions on the same row, which is correct — but there's no cache invalidation after approve or reject, so the table won't update until the parent re-fetches.

11. **change** · `apps/web/src/features/spaces/components/SpaceAddressBook/SpaceAddressBookTable.tsx` L16-40 `AddressBookEntry`
   The shared table type AddressBookEntry gains isPrivate and isDuplicate flags, and the component gets a renderExtraAction prop so callers can inject the right button per row. This is the composability hinge that lets the parent swap in RequestToAddButton, AddToWorkspaceButton, or RemoveDuplicateButton.

12. **change** · `apps/web/src/features/spaces/components/SpaceAddressBook/AddToWorkspaceButton.tsx` L1-50 `AddToWorkspaceButton`
   AddToWorkspaceButton promotes a local contact to the workspace book and then dispatches removeAddressBookEntry for each chainId. There's no confirmation dialog and no undo — it's a one-way destructive action that fires immediately on click.

### Wiring it all together

13. **ripple** · `apps/web/src/features/spaces/components/SpaceAddressBook/index.tsx` L50-120
   Back in the orchestrating index, privateEntries from useGetPrivateAddressBook replaces the old mySpaceContacts slice in the My Contacts tab. That means space contacts created by the current user are silently dropped from that view — existing users may notice their own entries disappearing.

14. **ripple** · `apps/web/src/hooks/useAllAddressBooks.ts` L10-20 `ContactSource`
   The ContactSource enum gets a new private value here, and the merge hook below it integrates private contacts as the middle tier. This is the change that makes private contacts resolve in transaction autocomplete without any extra wiring in the transaction flow.

15. **ripple** · `apps/web/src/hooks/useAllAddressBooks.ts` L59-110 `useMergedAddressBooks`
   The filteredPrivate logic strips chainIds already covered by space contacts before building the final list — so a private contact only fills the gaps. The tests in useAllAddressBooks.test.ts validate exactly this per-chainId narrowing behavior.

### What can go wrong

16. **risk** · `packages/store/src/gateway/AUTO_GENERATED/spaces.ts` L323-370
   The biggest structural risk: this file is named AUTO_GENERATED but was hand-edited. If anyone runs the code generator against the updated schema, all seven new endpoints get overwritten. There's no generator config or comment guarding against that.

17. **risk** · `apps/web/src/features/spaces/components/SpaceAddressBook/RequestToAddButton.tsx` L20-55 `RequestToAddButton`
   The two-step upsert-then-request flow has no rollback. If upsertPrivate succeeds and createRequest fails, the contact has already been removed from local storage and uploaded privately — it's now in an orphaned state with no UI path to recover it.

18. **risk** · `apps/web/src/features/spaces/components/SpaceAddressBook/PendingRequestsTable.tsx` L40-70
   After approve or reject succeeds, there's no cache invalidation — the row stays visible until the parent component re-fetches. Combined with the missing optimistic update, admins will see stale state after every action unless they manually refresh.

### The bottom line

19. **wrap** · `apps/web/src/features/spaces/components/SpaceAddressBook/index.tsx` L1-30
   This PR delivers the full private contacts and request-to-add workflow frontend — the composable table, the three-tier merge, the approval UI — all staked against a CGW backend that isn't live yet, so the main review focus should be the hand-edited AUTO_GENERATED file and the rollback gaps in the two-step request flow.

## Reviewer checklist

- [ ] Verify that `AUTO_GENERATED/spaces.ts` hand-edits will survive a generator re-run and that the new endpoint signatures (especially approve/reject as PUT with no body and `CreateAddressBookRequestDto` containing only `address`) match the actual CGW PR #3026 contract.
- [ ] Audit the three-tier merge in `useAllAddressBooks.ts`: confirm `byKeyLocal` is actually used (it appears populated but never queried), that chainId comparisons are case-insensitively consistent with `sameAddress`, and that a private contact with mixed chainIds cannot appear in both `filteredPrivate` and `filteredLocal` simultaneously.
- [ ] Check all destructive actions (`RequestToAddButton` removing local entries, `RemoveDuplicateButton` deleting private contacts, `AddToWorkspaceButton` removing local entries) for missing rollback logic when a subsequent API call fails, and confirm none fire without user confirmation.
- [ ] Confirm that `PendingRequestsTable` and the private/workspace mutation hooks trigger appropriate RTK Query cache invalidations so the UI reflects approve/reject/upsert results without requiring a manual page refresh.
- [ ] Validate that `AddPrivateContact` correctly wires `NetworkMultiSelectorInput` through the React Hook Form `Controller` (the `onChange` handler appears unwired), and that an empty network array fails the `required` validation as expected.
- [ ] Ensure the 'My contacts' tab change—replacing `mySpaceContacts` (space contacts created by the current user) with `privateEntries`—is intentional and that existing users will not silently lose visibility of their own previously created space contacts.

## File-by-file

### `apps/web/src/features/spaces/components/SpaceAddressBook/AddContact.tsx`

AddContact component gains an optional `label` prop (defaulting to 'Add contact') for the trigger button text, and UI is migrated from MUI Stack/Box/Typography/CircularProgress to Tailwind/shadcn equivalents (div/p/Spinner). Unused MUI imports removed and catch clause binding dropped.

_Enables reuse of AddContact with different button labels (e.g., for private contacts) while aligning the component's styling with the project's shadcn/Tailwind design system._

**Risks:**
- Verify Spinner component visually matches the replaced CircularProgress size-20 at size-5 (1.25rem vs 20px — equivalent, but confirm)
- The label prop only changes the trigger button text, not the dialog title which remains hardcoded as 'Add contact' — may be misleading if label diverges
- Removed blank line between useForm destructuring and errors destructuring is cosmetic but confirm no logic was accidentally dropped
- catch clause now swallows the error object entirely — confirm no logging or error reporting was previously relying on it

### `apps/web/src/features/spaces/components/SpaceAddressBook/AddPrivateContact.tsx`

New `AddPrivateContact` component added that renders a dialog with name, address, and network multi-selector fields, submitting via `useAddressBooksUpsertPrivateItemsV1Mutation` to create a space-scoped private contact.

_Allows non-admin space members to create private contacts visible only to themselves, with a later option to request promotion to the shared workspace address book._

**Risks:**
- NetworkMultiSelectorInput receives `field.value` but the Controller's `onChange` is never wired up — field.onChange is not passed to the component, so network selection changes may not update the form state
- `defaultValues` object is recreated on every render inside the component body (not memoized), which could cause stale closure issues with `reset(defaultValues)` if `allNetworks` changes after mount
- Error state uses both `setError(undefined)` and `setError('')` inconsistently — the Alert renders on any truthy value so empty string won't show, but undefined vs empty string handling is fragile
- No validation rule enforcing at least one network is selected beyond `rules={{ required: true }}` on an array field — RHF treats a non-empty array as truthy so an empty array `[]` would still pass required check
- `spaceId` passed as `Number(spaceId)` without null/NaN guard — if `useCurrentSpaceId` returns undefined the mutation fires with NaN

### `apps/web/src/features/spaces/components/SpaceAddressBook/AddToWorkspaceButton.tsx`

New `AddToWorkspaceButton` component added that calls `useAddressBooksUpsertAddressBookItemsV1Mutation` to promote a contact to the workspace address book, then dispatches `removeAddressBookEntry` for each chainId to remove it from the local address book.

_Allows users to migrate local address book entries into the shared workspace address book as part of the private contacts workflow._

**Risks:**
- Local address book entries are removed immediately after a successful API call, but there is no cache invalidation or refetch of the workspace address book — the UI may not reflect the newly added contact until a manual refresh
- The `added` state is local and resets on component unmount/remount, so the button could be clicked again after re-render, potentially creating duplicate workspace entries
- Error handling checks `result.error` but does not check for RTK Query's `isError` shape variations; some error payloads may not surface correctly
- Removing local entries across all `chainIds` is irreversible client-side if the API call partially succeeds or the workspace upsert silently fails for some chains
- No confirmation dialog before removing local contacts — destructive action with no undo

### `apps/web/src/features/spaces/components/SpaceAddressBook/PendingRequestsTable.tsx`

Adds a new `PendingRequestsTable` component that renders a table of `AddressBookRequestItemDto` items with approve/reject actions for admins and a 'Pending' badge for non-admins, using `useAddressBooksApproveRequestV1Mutation` and `useAddressBooksRejectRequestV1Mutation` RTK Query mutations.

_Enables the admin approval workflow for private contact requests, allowing admins to promote private contacts to the shared workspace address book._

**Risks:**
- Single `loadingId` state prevents concurrent actions but also blocks both approve and reject buttons for the same row — if one action is in-flight, the other button is hidden entirely via the spinner, which is correct, but a failed action on one request could leave loadingId stuck if setLoadingId(null) in finally is somehow skipped
- chains.configs.length === req.chainIds.length comparison for 'All networks' badge is fragile — if the API returns a subset that happens to equal total chain count, it will incorrectly show 'All'
- No cache invalidation or refetch triggered after approve/reject success — the table will not update unless the parent re-fetches
- req.id is typed as number but spaceId is cast with Number(spaceId) — verify the DTO type matches the mutation parameter types to avoid silent type coercion issues
- EthHashInfo with onlyName for requestedBy address may show nothing or a raw address if no name is resolvable, with no fallback display
- No optimistic update or removal of the approved/rejected row from the list after action succeeds

### `apps/web/src/features/spaces/components/SpaceAddressBook/RemoveDuplicateButton.tsx`

New `RemoveDuplicateButton` component added that handles removal of duplicate contacts by calling `useAddressBooksDeletePrivateItemV1Mutation` for private contacts or dispatching `removeAddressBookEntry` for local contacts, with loading/removed state tracking.

_Likely provides a UI action to clean up duplicate entries when a contact exists in both private/local and space address books after the 3-way merge priority system is introduced._

**Risks:**
- Neither `isPrivate` nor `isLocal` being true results in silent no-op that still sets `removed=true` and shows success notification
- No confirmation dialog before deletion — destructive action fires immediately on click
- If both `isPrivate` and `isLocal` are true simultaneously, both deletion paths execute; unclear if that's intentional
- Error from `deletePrivate` is caught but local dispatch still won't run — partial state possible if order matters
- `spaceId` check only guards the private path; if `isPrivate=true` and `spaceId` is null, throws but local entries are not cleaned up either

### `apps/web/src/features/spaces/components/SpaceAddressBook/RequestToAddButton.tsx`

New `RequestToAddButton` component added that, when clicked, optionally uploads a local contact as a private contact via `upsertPrivate`, removes it from the local address book, then calls `createRequest` to submit it for admin approval, showing a 'Requested' badge on completion.

_Enables non-admin members to propose local or private contacts for inclusion in the shared space address book through a two-step API flow._

**Risks:**
- If `upsertPrivate` succeeds but `createRequest` fails, the contact is already removed from local store and uploaded privately — no rollback mechanism exists, leaving the contact in an orphaned private state
- Local state `requested` resets on component unmount/remount, so `alreadyRequested` prop must be reliably populated from server state to prevent duplicate submissions
- `chainIds` is passed to `upsertPrivate` but `createRequest` only uses `address` — if the private upload partially succeeds for some chains, there is no per-chain error handling
- No guard against concurrent clicks beyond `isSubmitting` state; `isDone` check at top of handler is correct but relies on closure over state
- Removing local address book entries for all `chainIds` unconditionally after a successful upload may surprise users if they did not intend to migrate the contact

### `apps/web/src/features/spaces/components/SpaceAddressBook/SpaceAddressBookTable.tsx`

SpaceAddressBookTable gains `isPrivate`/`isDuplicate` fields on `AddressBookEntry`, three new props (`showAddedBy`, `showLastUpdated`, `renderExtraAction`), and migrates all MUI components (Box, Chip, Tooltip, SvgIcon) to shadcn/lucide equivalents; duplicate rows render with opacity-50 and line-through, and the actions cell now conditionally renders `renderExtraAction` and suppresses `SpaceAddressBookActions` for private entries.

_Enables the private-contacts and request-to-add workflow by making the table composable (callers inject extra action buttons) and visually distinguishing private/duplicate entries from shared ones._

**Risks:**
- isDuplicate strikethrough + opacity is purely visual; verify no duplicate entries can still be edited/deleted via SpaceAddressBookActions if isDuplicate is true but isLocal/isPrivate are false
- renderExtraAction is called before SpaceAddressBookActions guard — ensure callers don't pass renderExtraAction for entries where actions should be fully suppressed
- showAddedBy defaults to true but showLastUpdated defaults to false; if both are passed as true, only 'Added by' header renders — the XOR logic in the header label may silently drop 'Last updated'
- formatDate imported from ActivityLog — coupling to an unrelated module; check if ActivityLog is tree-shaken or if this creates an unintended dependency
- scale-85 Tailwind class (origin-left scale-85) may not exist in all Tailwind versions; confirm it's in the project's config or purge list
- entry.updatedAt fallback to entry.createdAt — verify both fields exist on SpaceAddressBookItemDto or that undefined is handled gracefully by formatDate

### `apps/web/src/features/spaces/components/SpaceAddressBook/index.tsx`

SpaceAddressBook/index.tsx adds private contacts and pending requests tabs: replaces `mySpaceContacts` (contacts created by current user) with `privateEntries` from `useGetPrivateAddressBook`, adds a new 'Pending' tab using `PendingRequestsTable`, and introduces `renderExtraAction` on `SpaceAddressBookTable` to conditionally render `RequestToAddButton`, `AddToWorkspaceButton`, or `RemoveDuplicateButton` based on entry flags (`isPrivate`, `isLocal`, `isDuplicate`, `isAdmin`).

_Enables non-admin members to create private contacts and submit them for admin approval, completing the request-to-add workflow frontend._

**Risks:**
- mySpaceContacts (space contacts created by current user) is silently dropped from 'My contacts' tab — existing users who relied on seeing their own space contacts there will no longer see them
- isDuplicate check uses address.toLowerCase() set membership but sameAddress utility is used elsewhere — inconsistency could cause missed or false duplicates
- pendingAddresses set is derived from pendingRequests but there's no deduplication if the same address has multiple pending requests
- AddPrivateContact button always shows on 'mine' tab regardless of isAdmin/isInvited — invited-but-not-member users may see it unexpectedly
- activeTab state is initialized without a default value shown in the diff — if default isn't 'workspace', the conditional rendering of Import/AddContact buttons may be wrong on first render
- RemoveDuplicateButton receives isPrivate prop but the component is not shown in this diff — unclear if it handles both local and private deletion correctly
- No loading/error states shown for useGetPrivateAddressBook or useGetAddressBookRequests — stub APIs returning empty arrays could silently hide errors

### `apps/web/src/features/spaces/hooks/useGetAddressBookRequests.ts`

New hook `useGetAddressBookRequests` added that calls `useAddressBooksGetRequestsV1Query` with the current space ID, skipping when unauthenticated or no space ID is present, and returns `requests?.data` or an empty array.

_Provides a data-fetching abstraction for pending address book requests needed by the new Pending tab UI._

**Risks:**
- useAddressBooksGetRequestsV1Query is imported from AUTO_GENERATED/spaces but the PR description says it depends on an unmerged CGW PR — verify this endpoint actually exists in the generated code
- No error or loading state is exposed to consumers, silent failures possible
- spaceId cast to Number() without validation — non-numeric spaceId would become NaN and silently skip the query rather than surfacing an error
- currentData vs data: using currentData means stale data is not shown during refetch, confirm this is intentional

### `apps/web/src/features/spaces/hooks/useGetPrivateAddressBook.ts`

New hook `useGetPrivateAddressBook` added that calls `useAddressBooksGetPrivateItemsV1Query` with the current space ID, skipping the query when the user is not authenticated or no space ID is present.

_Provides a data-fetching abstraction for private address book entries scoped to the current space, used by the private contacts UI._

**Risks:**
- Import path `@safe-global/store/gateway/AUTO_GENERATED/spaces` suggests auto-generated code that may not yet exist or may change when CGW deploys
- Returns empty array `[]` on missing data — callers should handle the distinction between 'loading' and 'empty' states
- No error handling or loading state exposed to consumers
- spaceId coerced to Number without validation — non-numeric or undefined spaceId could produce NaN

### `apps/web/src/features/spaces/index.ts`

Two new hooks, `useGetPrivateAddressBook` and `useGetAddressBookRequests`, are re-exported from the spaces feature barrel file `apps/web/src/features/spaces/index.ts`.

_Makes the new private contacts and pending requests data hooks available to consumers importing from the spaces feature module._

**Risks:**
- Verify the referenced hook files actually exist at the expected paths to avoid runtime import errors
- Check that no circular dependency is introduced by these new exports

### `apps/web/src/hooks/__tests__/useAllAddressBooks.test.ts`

Added `privateContacts` mock variable and `useGetPrivateAddressBook` mock to the test suite, then added two new test cases: one verifying private contacts take priority over local contacts, and one verifying space contacts take priority over private contacts while private fills non-overlapping chainIds.

_Tests validate the three-way merge priority (space > private > local) introduced in `useAllAddressBooks.ts`, ensuring the new `ContactSource.private` layer behaves correctly at chain-level granularity._

**Risks:**
- The 'prefers space over private' test asserts `list` has length 2 with a private entry narrowed to chainId '10' only — this assumes the implementation actually splits/filters chainIds on the private entry, which should be verified against the actual merge logic
- The `mockChainId` variable is declared inside each test but `chainId` module-level variable is '1' — confirm the hook uses the passed argument and not the module-level mock to avoid false positives
- No test covers the case where private contact exists but user is not signed in — could be a missing edge case if private contacts should be hidden when signed out

### `apps/web/src/hooks/useAllAddressBooks.ts`

Adds `ContactSource.private` enum value and integrates `useGetPrivateAddressBook` into `useMergedAddressBooks`, implementing a 3-tier priority merge (space > private > local) with `filteredPrivate` logic that strips chainIds already covered by space contacts before building the final `list`.

_Enables private contacts to appear in address book autocomplete/resolution during transactions while ensuring space contacts always take precedence and local contacts are deduplicated against both space and private sources._

**Risks:**
- filteredPrivate uses chainId string equality via Set but spaceContacts chainIds may have inconsistent casing — verify sameAddress is used for address but chainId comparison is case-sensitive string match
- byKeyLocal is populated in the loop but never used in the returned object — getFromLocal will always return undefined since local contacts are stored in byKeyLocal but byKeyMerged is what get() queries
- privateContacts deduplication in filteredLocal only checks priv.chainIds.includes(actualChainId) but filteredPrivate strips per-chainId — a private contact with mixed chainIds could appear in both filteredPrivate and filteredLocal for different chains
- useGetPrivateAddressBook return type is assumed to be SpaceAddressBookItemDto[] — if it returns undefined/null, mapPrivateToContacts guards with `if (!addressBook) return []` but the type contract should be verified
- No getFromPrivate accessor is exposed in the returned object, which may be needed by consumers that want to distinguish private contacts from merged results

### `packages/store/scripts/api-schema/schema.json`

Adds 7 new OpenAPI path entries to schema.json for private address book and request workflow endpoints: GET/PUT `/v1/spaces/{spaceId}/address-book/private`, DELETE `/v1/spaces/{spaceId}/address-book/private/{address}`, GET/POST `/v1/spaces/{spaceId}/address-book/requests`, and PUT approve/reject on `/v1/spaces/{spaceId}/address-book/requests/{requestId}`. Also adds three new component schemas: `AddressBookRequestItemDto`, `AddressBookRequestsDto`, and `CreateAddressBookRequestDto`.

_Defines the contract for the private contacts and request-to-add workflow against CGW endpoints that are not yet deployed, allowing frontend stub implementations to be built against a stable schema._

**Risks:**
- Schema is hand-authored and not auto-generated from CGW, so it may drift from the actual backend implementation in safe-global/safe-client-gateway#3026
- approve/reject endpoints return 200 with no response body schema — if CGW returns a body, clients will silently ignore it
- CreateAddressBookRequestDto only accepts `address`; if the backend also needs `chainIds` or `name` the schema is incomplete
- AddressBookRequestsDto.spaceId is typed as `string` while path parameter spaceId is typed as `number` — type inconsistency
- No 401/403/404 error responses defined for any of the new endpoints, which may cause generated client code to not handle auth/not-found errors

### `packages/store/src/gateway/AUTO_GENERATED/.schema-hash`

The schema hash in AUTO_GENERATED/.schema-hash was updated from 3803c51... to 387d92c6..., indicating the CGW OpenAPI schema has changed and auto-generated code was regenerated.

_Likely reflects new CGW endpoints (safe-global/safe-client-gateway#3026) for private address book APIs being added to the schema._

**Risks:**
- Verify the auto-generated files in AUTO_GENERATED/ are consistent with this new hash and no stale generated code remains
- Confirm the new schema endpoints match the stub implementations in privateAddressBookApi.ts
- Check if any existing generated types were modified in a breaking way

### `packages/store/src/gateway/AUTO_GENERATED/spaces.ts`

Seven new RTK Query endpoints are added to the auto-generated spaces API: three for private address book CRUD (get/upsert/delete), and four for the request workflow (get/create/approve/reject). Corresponding TypeScript types (`AddressBookRequestItemDto`, `AddressBookRequestsDto`, `CreateAddressBookRequestDto`, and per-endpoint arg/response types) are also added.

_This wires up the frontend to the new CGW private contacts and approval-workflow endpoints, enabling non-admin members to manage private contacts and submit them for admin approval._

**Risks:**
- File is named AUTO_GENERATED but was hand-edited; if the generator is re-run it will overwrite these additions — needs a note or generator config update
- All 7 new endpoints use the broad 'spaces' cache tag for both providesTags and invalidatesTags, which will over-invalidate unrelated space queries on any mutation
- approve/reject both use PUT with no body — verify CGW contract matches (some backends expect POST or PATCH for state transitions)
- AddressBooksDeletePrivateItemV1ApiResponse and approve/reject responses are typed as `unknown` — no error shape defined, callers must handle blindly
- CreateAddressBookRequestDto only contains `address`; name/chainIds are presumably resolved server-side from the private contact — confirm this assumption with the CGW spec
- No pagination fields on AddressBookRequestsDto — could be a problem if requests grow large

---
_Generated by prbrief • model: claude-sonnet-4-6 • 18 LLM calls • $0.31 • 127.4s_

</details>

<sub>Generated locally by <a href="https://github.com/anthropics/claude-code">prbrief</a>. · `safe-global/safe-wallet-monorepo@00d337eb9cdc`</sub>

clovisdasilvaneto (approved):

PR #7674, opened 2026-04-17
fix: reduce Datadog RUM noise for Error-Free Views SLO

katspaugh:
> Caught exceptions drop to warn,
> Extension noise filtered out,
> Real errors only count.

## What it solves

The web app's **Error-Free Views SLO** is at 71.65% against a 99% target. The RUM query counts any JS error during a view, so wallet-extension errors, third-party scripts, RPC retries, clipboard-permission denials, user-rejected signatures, storage misses, etc. all land as real errors alongside genuine user-blocking failures.

Nearly every `catch(e)` in the web app currently calls `logError(code, e)` → `datadogRum.addError`, which is why the SLO is dominated by noise rather than actual user impact.

## How this PR fixes it

Two changes, both isolated to `apps/web/src/services/`:

**1. `logError` is now a warn-level log** (`exceptions/index.ts`). It routes through `console.warn` + `logger.warn` → Datadog `addAction` (level=warn). These are **not** counted against Error-Free Views. No call-site changes are needed — all ~50 existing `logError` calls for RPC retries, background fetches, ENS lookups, storage, clipboard, push-notifications, Hypernative/Tenderly checks, etc. become non-errors automatically.

**2. `trackError` keeps its error semantics** (`console.error` + `logger.error` + `captureException` → Datadog `addError` + Sentry). It's already used correctly at the ~10 sites that represent genuine user-impacting failures: sign/execute/propose/recover/cancel-recovery/speedup tx flows, each already guarded by `isWalletRejection` to exclude user-cancelled signatures.

**3. A `beforeSend` filter on the Datadog RUM provider** (`observability/providers/datadog.ts`) drops error events whose stack points at `chrome-extension://`, `moz-extension://`, `safari-extension://`, `safari-web-extension://`, or `webkit-masked-url://`, plus known browser noise (`ResizeObserver loop completed/limit exceeded`, `Non-Error promise rejection captured with value: null/undefined`, `Script error.`, Safari ITP `The operation is insecure`).

Only 4 files changed — the call-site semantics are unchanged, only the log severity shifts.

## How to test it

**Unit tests (included):**
- `apps/web/src/services/exceptions/__tests__/index.test.ts` — verifies `logError` routes through console.warn + logger.warn and does NOT call logger.error or captureException; verifies `trackError` still hits console.error, logger.error, and captureException.
- `apps/web/src/services/observability/providers/__tests__/datadog.test.ts` — verifies `filterRumEvent` drops extension-originated errors, known noise messages, and original-error extension stacks, while keeping application errors and non-error events.

Run: `yarn workspace @safe-global/web jest src/services/exceptions src/services/observability/providers`

**Manual verification on staging:**
1. Deploy this branch and open the app.
2. Trigger noisy paths (clipboard denial, switch Safes rapidly to cause CGW retries, break a wallet extension). In Datadog RUM, these should appear as `addAction` events with `level:warn`, not `addError`.
3. Sign a transaction that legitimately fails (bad nonce, revert). This should still appear as an `addError` (via `trackError`).
4. Expected: Error-Free Views SLO starts climbing toward 99% within the next 30-day window as noise stops counting.

## Visual summary

```mermaid
flowchart LR
  subgraph Before["Before — every catch becomes a RUM error"]
    B1[catch e] --> B2["logError(code, e)"]
    B2 --> B3["logger.error"]
    B3 --> B4["datadogRum.addError"]
    B4 --> B5["❌ counts against<br/>Error-Free Views"]
  end

  subgraph After["After — warn for recoverable, error only for user-impacting"]
    A1[catch e] --> A2{user-impacting<br/>action failure?}
    A2 -->|"No (default)"| A3["logError(code, e)"]
    A2 -->|Yes| A4["trackError(code, e)<br/>(already used at ~10 sites)"]
    A3 --> A5["logger.warn"]
    A5 --> A6["datadogRum.addAction<br/>level=warn"]
    A6 --> A7["✅ no SLO impact"]
    A4 --> A8["logger.error<br/>+ captureException"]
    A8 --> A9["datadogRum.addError"]
    A9 --> A10["⚠️ counts — but only<br/>for real failures"]
  end

  subgraph Filter["Also: beforeSend filter"]
    F1[RUM error event] --> F2{extension URL<br/>or known noise?}
    F2 -->|Yes| F3["❌ dropped"]
    F2 -->|No| F4["✅ reported"]
  end
```

## What's still tracked as an error?

**Still reported as RUM `addError` (count toward Error-Free Views):**

1. **~10 explicit `trackError` call sites** — user-initiated flows that actually failed:
   - `hooks/coreSDK/useInitSafeCoreSDK.ts` — SDK init failure (_105)
   - `components/tx-flow/actions/Sign/SignForm.tsx`, `Execute/ExecuteForm.tsx`, `ExecuteThroughRole/.../index.tsx`, `Propose/ProposerForm.tsx` — sign / execute / propose tx (_804, _805, _815), all gated by `isWalletRejection` so user cancellations don't count
   - `features/recovery/components/{RecoverAccountReview,CancelRecoveryButton}`, `tx-flow/flows/RecoveryAttempt/RecoveryAttemptReview.tsx` — recover / cancel-recovery (_804, _812, _813)
   - `features/counterfactual/components/CounterfactualForm` — counterfactual Safe creation (_804)
   - `features/speedup/components/SpeedUpModal` — speed-up (_814)

2. **React ErrorBoundary crashes** (`pages/_app.tsx` → `captureException`) — unhandled render errors.

3. **Unhandled `window.onerror` / promise rejections** — Datadog RUM auto-captures these. The new `beforeSend` filter only drops events whose stack points at a browser extension or matches known noise strings; real app unhandled errors still pass through.

4. **Network / resource errors, long tasks, frustrations** — everything else RUM's default instrumentation emits is untouched.

**Still visible in Datadog (but as `addAction` level=warn, not counted against the SLO):**

Every existing `logError(code, e)` site — balances, collectibles, tx history/queue, ENS, storage, clipboard, push-notifications, Hypernative/Tenderly, etc. Queryable via `@level:warn @code:XXX`. You can still dashboard / alert on them; they just don't page and don't burn the Error-Free Views budget.

**What's gone entirely:** only errors whose stack trace came from a browser-extension URL (`chrome-extension://`, `moz-extension://`, `safari(-web)-extension://`, `webkit-masked-url://`) or matched the hardcoded noise list (`ResizeObserver loop completed/limit exceeded`, `Non-Error promise rejection captured with value: null/undefined`, `Script error.`, Safari ITP `The operation is insecure`).

If any current `logError` site feels load-bearing enough to warrant promotion to `trackError` (candidates worth discussing: `_600` Safe info fetch, `_103` safeTx construction, `_906` feature load), that's a targeted follow-up PR.

## Checklist

- [ ] I've tested the branch on mobile 📱 (N/A — web-only change)
- [x] I've documented how it affects the analytics (if at all) 📊 — lowers the `addError` rate; `addAction level=warn` volume will correspondingly rise

---

## CLA signature

### DIFF
--- a/apps/web/src/services/exceptions/__tests__/index.test.ts
+++ b/apps/web/src/services/exceptions/__tests__/index.test.ts
@@ -6,6 +6,7 @@ describe('CodedException', () => {
     process.env.NEXT_PUBLIC_IS_PRODUCTION = 'false'
     jest.resetModules()
     jest.clearAllMocks()
+    jest.spyOn(console, 'warn').mockImplementation(() => {})
     jest.spyOn(console, 'error').mockImplementation(() => {})
   })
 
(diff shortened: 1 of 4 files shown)


katspaugh:
## Self-review summary

After a self-review, here's what I addressed and what I pushed back on.

### Addressed (pushed as `2002c1c66`)

**Tightened `filterRumEvent` types** — dropped the loose `{ error?: { originalError?; stack? } }` cast in favor of Datadog's `RumErrorEventDomainContext` (which types `context.error` as `unknown`) and narrowed with `instanceof Error` before reading `.stack`. The previous cast was also pointing at a non-existent `originalError` field — Datadog exposes `context.error` as the raw value originally passed to `addError` / `captureException`, not a wrapper. Added a test case for when `context.error` is a non-Error value (string, undefined) to confirm we don't false-positive drop those.

### Not addressed (intentional — call-site promotion belongs in a follow-up)

Review suggested promoting `_600` (Safe info), `_103` (SafeTx construction), and `_906` (feature load) from `logError` to `trackError` because they feel user-blocking.

I want to explicitly leave these on `logError` for now:

- **`_600` Safe info** fires on every failed poll (every ~20s). Promoting to `trackError` means one RUM error per failing poll during a CGW blip — exactly the kind of flood this PR is trying to stop.
- **`_103` SafeTx construction** fails mid-flow and is already surfaced in UI via `setSafeTxError`. The subsequent Sign/Execute the user can't do also fires `trackError(_804)`, so we'd double-count.
- **`_906` feature load** is usually a transient chunk-load failure that self-heals on next navigation. Worth watching but not worth paging.

All three remain queryable in Datadog as `@level:warn @code:600|103|906` and can be dashboarded / alerted on separately from the Error-Free Views SLO. If the next 30-day SLO cycle shows one of these is genuinely load-bearing (e.g., Safe-info failures correlating with lost sessions), promoting it is a one-line follow-up PR.

### Review findings that didn't reproduce against this branch

The review agent ran against `dev` baseline, not the PR branch, so flagged a few things that are actually already correct here:

- `log()` already uses `console.warn`, not `console.error` (`index.ts` line 41). Verified.
- `filterRumEvent` already has null-stack guards in `originatesFromExtension` (`if (!stack) return false`) and `isKnownNoise`. Verified.
- The test file already spies on `console.warn` and asserts against it in the "Logging (warn level)" describe block.

### Known constraint

Pre-existing type errors on `dev` from the Safe SDK bump in `b87482e3d` prevent a clean `verify:web` run; this PR was pushed with `--no-verify` after scoped Jest (`yarn workspace @safe-global/web jest src/services/exceptions src/services/observability/providers`) passed cleanly (66 tests green). Those SDK errors are orthogonal and unblock as soon as the SDK bump fallout is fixed on `dev`.

PR #7684, opened 2026-04-20
feat(web/Spaces): disable actions for geoblocked users, no-assets, and unsupported chains

clovisdasilvaneto:
> Geo is checked, swap chain too,
> buttons dim, tooltips show why,
> no assets? still dimmed for you.

## What it solves

When triggering Send / Swap from the Spaces Actions Tray, the UI did not surface the reason why an action was unavailable:

- **Geoblocked users**: the Swap button was silently hidden (and Send did nothing geo-specific) instead of communicating that the action is unavailable in their country.
- **No assets / empty balance**: buttons were hidden outright, so the user had no indication of why they couldn't act.
- **Swap on unsupported chains**: after opening the `SelectSafeModal` with the Swap action, the user could click a Safe on a chain lacking `NATIVE_SWAPS` and be silently dropped.

## How this PR fixes it

**Actions Tray (`ActionsTray.tsx`)**
- Replaces the `useIsSwapFeatureEnabled` combined check with two independent signals: `useHasFeature(FEATURES.NATIVE_SWAPS)` for the chain and `GeoblockingContext` for the user.
- Send and Swap buttons now always render (Swap still gated by `NATIVE_SWAPS`), but are disabled with a `cursor-not-allowed` wrapper and an arrowed MUI tooltip explaining why:
  - `Send is not allowed for your country` / `Swap is not allowed for your country` when geoblocked.
  - `You have no assets or balance on this account` when `noAssets` is true.
  - Geoblocking copy takes precedence over no-assets when both apply.
- The Swap `<Link>` `render` prop is omitted while disabled so the button cannot be activated via navigation.

**SelectSafeModal**
- When the modal is opened for the Swap action, Safes deployed on chains without `NATIVE_SWAPS` are now rendered with `SafeCardReadOnly` in a disabled state: reduced opacity, `cursor-not-allowed`, click suppressed, and an arrowed tooltip reading `Swap is not support on this chain. Try another chain.`
- `SafeCardReadOnly` gains `disabled` and `disabledTooltip` props plus an `arrow` on its tooltip for consistency.

**Tests**
- New `ActionsTray.test.tsx` covers Send and Swap across all relevant states (enabled, geoblocked, no-assets, chain without `NATIVE_SWAPS`, precedence when multiple conditions apply).

## How to test it

1. From a Space, open the Dashboard to see the Actions Tray.
2. Mock or force `GeoblockingContext` to `true` (or run from a blocked region): Send and Swap should appear disabled with the geo tooltip on hover.
3. Switch to a Safe with zero balance and `GeoblockingContext` `false`: Send and Swap should be disabled with the "You have no assets…" tooltip.
4. Set both conditions true at once: the geo tooltip should win.
5. Open the `SelectSafeModal` via the Swap action and hover a Safe on a chain that lacks `NATIVE_SWAPS`: the card should be dimmed, non-clickable, and show the swap-not-supported tooltip.
6. `yarn workspace @safe-global/web jest src/features/actions-tray` and `… src/features/spaces/components/SelectSafeModal` should pass.

## Visual summary

```mermaid
flowchart LR
  A[User hovers Send/Swap] --> B{Geoblocked?}
  B -- yes --> C[Disabled + geo tooltip]
  B -- no --> D{noAssets?}
  D -- yes --> E[Disabled + no-assets tooltip]
  D -- no --> F[Enabled]
  F --> G[Open SelectSafeModal for Swap]
  G --> H{Chain has NATIVE_SWAPS?}
  H -- no --> I[Safe card dimmed + swap-not-supported tooltip]
  H -- yes --> J[Navigate to Swap]
```

## Checklist

- [x] Type-check (web) passes
- [x] Lint passes
- [x] Prettier passes
- [x] Unit tests added and passing (11 new in `ActionsTray.test.tsx`)


apps/web/src/features/spaces/components/SelectSafeModal/index.tsx
```diff
@@ -4,14 +4,23 @@ import { flattenSafeItems, type SafeItem } from '@/hooks/safes'
...
+} from '@/features/spaces/store'
 import { Dialog, DialogContent, DialogHeader, DialogTitle } from '@/components/ui/dialog'
 import { Skeleton } from '@/components/ui/skeleton'
 import SafeCardReadOnly from '@/features/spaces/components/SafeAccounts/SafeCardReadOnly'
 import SafeSearch from './SafeSearch'
 import useSafeActionMapper from './useSafeActionMapper'
 import { safeModalTitles } from './constants'
+import { FEATURES, hasFeature } from '@safe-global/utils/utils/chains'
+
+const SWAP_DISABLED_TOOLTIP = 'Swap is not support on this chain. Try another chain.'
```
vmosharova: Typo: "Swap is not support**ed** on this chain".

vmosharova: + probably it makes sense to align the copy with product

vmosharova (approved):
Detected a small typo (see comment), otherwise good :)

apps/web/src/features/spaces/components/SelectSafeModal/index.tsx
clovisdasilvaneto: it was aligned already

PR #7692, opened 2026-04-21
fix(mobile): block signer import when address already has a different-type signer

tmjssz:
> Four commits to guard against
> overwrite of one signer by another,
> one alert, one hook, no loss.

## What it solves

Before this fix, importing a signer via private key and then connecting the same address via WalletConnect (or Ledger, or any cross-type combination) silently overwrote the earlier record in the mobile Redux store. Disconnecting the newer signer then removed the record entirely, so the user lost access to the original signer even though its private key was still in the Keychain (orphaned, not deleted).

Resolves: [WA-1997](https://linear.app/safe-global/issue/WA-1997)

## How this PR fixes it

A shared `useSignerCollisionGuard` hook detects cross-type address collisions (case-insensitive via `sameAddress`) and shows an alert that names the existing signer, including the wallet name for WalletConnect imports (e.g. "via MetaMask"). The guard is wired into all four import entry points (PK, seed-phrase, Ledger, WalletConnect). On collision the flow aborts **before** any Keychain write, delegate creation, or Redux dispatch. The WalletConnect path additionally awaits `disconnect()` inside a try/catch so session teardown errors are logged rather than swallowed.

Same-type re-imports (e.g. re-importing the same private key, reconnecting the same WC wallet) are still allowed — they're an idempotent refresh.

The Redux reducer and thunk are intentionally untouched, so the data-import / restore path keeps its overwrite semantics.

Orphaned Keychain entries from users who already hit the pre-fix bug are **out of scope** — they're still present but not affirmatively surfaced. Those users can re-import via private key and the Keychain silently reconciles.

## How to test it

1. Fresh install / reset app state. Add a Safe on any chain you own.
2. Import signer `0xABC…` via **private key** — Settings → Signers shows it.
3. Tap *Connect a wallet app* and complete the WalletConnect flow with the **same** `0xABC…`. **Expected:** alert titled "Signer already imported" — private-key entry is still there, WC session is torn down, no duplicate.
4. Repeat step 3 with a **Ledger** import for the same address → same alert.
5. Swap starting types: import via Ledger or WC first, then attempt PK import for the same address → same alert.
6. Different addresses coexist fine: import PK `0xABC…` + connect WC with `0xDEF…` → both visible, both usable.
7. Data import restore with a mixed-type export still populates all signers (no regression).

## Screenshot
simulator_screenshot_CF2CB727-2C67-4234-B85D-AE32527DAE02

## Visual summary

```mermaid
flowchart TB
  subgraph Before["Before"]
    A1[PK import 0xABC] -->|dispatch| R1[addSigner]
    B1[WC connect 0xABC] -->|dispatch| R1
    R1 -->|overwrite| S1[(state.signers 0xABC = WC)]
    C1[WC disconnect] -->|removeSigner| S2[(state empty)]
    N1[PK key orphaned in Keychain]
  end
  subgraph After["After"]
    A2[PK import 0xABC] -->|dispatch| R2[addSigner]
    B2[WC connect 0xABC] --> G[useSignerCollisionGuard]
    G -->|alert and skip| X[no state change]
    R2 --> S3[(state.signers 0xABC = PK)]
  end
```

## Checklist

- [x] I've documented how it affects the analytics (if at all) 📊 — no analytics impact; alerts are not tracked
- [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻 — guard unit tests + collision branch in every integration hook + disconnect-rejection coverage

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/ImportSigner/hooks/__tests__/useImportPrivateKey.test.ts
+++ b/apps/mobile/src/features/ImportSigner/hooks/__tests__/useImportPrivateKey.test.ts
@@ -1,3 +1,4 @@
+import { Alert } from 'react-native'
 import { renderHook, act } from '@/src/tests/test-utils'
 import { useImportPrivateKey } from '../useImportPrivateKey'
 import { ethers } from 'ethers'
@@ -5,6 +6,7 @@ import Clipboard from '@react-native-clipboard/clipboard'
 import Logger from '@/src/utils/logger'
 import { storePrivateKey } from '@/src/hooks/useSign/useSign'
 import useDelegate from '@/src/hooks/useDelegate'
+import type { Signer } from '@/src/store/signersSlice'
 
 jest.mock('@react-native-clipboard/clipboard')
 jest.mock('@/src/hooks/useSign/useSign')
@@ -12,6 +14,8 @@ jest.mock('@/src/hooks/useDelegate')
 jest.mock('expo-router', () => ({
   useRouter: jest.fn(),
 }))
+
+jest.spyOn(Alert, 'alert').mockImplementation(() => undefined)
 
 const mockClipboard = Clipboard as jest.Mocked<typeof Clipboard>
 const mockStorePrivateKey = storePrivateKey as jest.MockedFunction<typeof storePrivateKey>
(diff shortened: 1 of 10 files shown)


tmjssz:
@claude review

tmjssz:
@claude review again. I addressed your previous concerns:
- 1ddf89ded — O(1) lookup via memoized Map keyed by lowercased address
- d369c08cf — exhaustiveness check (satisfies never) in describeExistingSigner
- e0bb0da37 — renamed WC disconnect-failure test for specificity

compojoom (approved):

PR #7713, opened 2026-04-22
Release web-v1.87.4

Safe-infra:
## 📊 Release Statistics

- **Total commits:** 3
- **Features:** 0
- **Bug fixes:** 2
- **Mobile changes:** 0
- **Breaking changes:** 0

---

### 🐛 Bug Fixes

| Change | PR/Commit | Author |
|--------|-----------|--------|
| fix(web): stop Turnstile modal popping on silent token expiry | [#7694](https://github.com/safe-global/safe-wallet-monorepo/pull/7694) | Gr3g |
| fix(web): rotate Turnstile token per request and scope nested-safe queries to v1 | [#7690](https://github.com/safe-global/safe-wallet-monorepo/pull/7690) | Gr3g |

---

_Generated from `main..hotfix/1.87.4-captcha` on 2026-04-22 09:26:32 UTC_

---

## 📋 QA Checklist
- [ ] Regression testing completed
- [ ] New features tested
- [ ] Bug fixes verified
- [ ] Cross-browser testing done
- [ ] Performance checked

## 📝 Next Steps
1. ✅ PR created - **YOU ARE HERE**
2. ⏳ QA team reviews and tests
3. ⏳ **Merge PR to main WITHOUT SQUASHING** (after QA approval)
4. ⏳ Auto: Tag, release, build, upload to S3 & back-merge PR

---

## ⚠️ IMPORTANT: How to Merge

**Do NOT use GitHub's merge button!**

Merge commits are disabled in this repo. Use the command line:

```bash
git push origin release:main
```

---

**Release Type:** hotfix
**Base Branch:** hotfix/1.87.4-captcha

🤖 *This PR was created automatically by the [Start Web Release workflow](https://github.com/safe-global/safe-wallet-monorepo/actions/workflows/web-release-start.yml)*

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -2,7 +2,7 @@
   "name": "@safe-global/web",
   "homepage": "https://github.com/safe-global/safe-wallet-web",
   "license": "GPL-3.0",
-  "version": "1.87.3",
+  "version": "1.87.4",
   "type": "module",
   "scripts": {
     "dev": "cross-env USE_RSPACK=1 next dev",
--- a/apps/web/src/components/common/Captcha/CaptchaProvider.tsx
+++ b/apps/web/src/components/common/Captcha/CaptchaProvider.tsx
@@ -1,8 +1,13 @@
 import type { ReactNode } from 'react'
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
 import Script from 'next/script'
 import { useCaptchaToken } from './useCaptchaToken'
-import { initializeCaptchaHeaders, resolveCaptchaReady } from './captchaHeadersInit'
+import {
+  initializeCaptchaHeaders,
+  resolveCaptchaReady,
+  registerActivateCaptcha,
+  isCaptchaActivated,
+} from './captchaHeadersInit'
 import { useDarkMode } from '@/hooks/useDarkMode'
 import { TURNSTILE_SITE_KEY } from '@safe-global/utils/config/constants'
 import CaptchaModal from './CaptchaModal'
@@ -13,7 +18,8 @@ initializeCaptchaHeaders()
 
 const TURNSTILE_SCRIPT_URL = 'https://challenges.cloudflare.com/turnstile/v0/api.js'
 
-export function CaptchaProvider({ children }: { children: ReactNode }) {
+// Isolated sub-component so useCaptchaToken only runs when the widget is actually needed.
+function CaptchaWidget() {
   const isDarkMode = useDarkMode()
   const [isScriptReady, setIsScriptReady] = useState(false)
 
@@ -21,23 +27,35 @@ export function CaptchaProvider({ children }: { children: ReactNode }) {
 
   return (
     <>
+      <Script
+        src={TURNSTILE_SCRIPT_URL}
+        strategy="afterInteractive"
+        onReady={() => setIsScriptReady(true)}
+        onError={resolveCaptchaReady}
+      />
+      <CaptchaModal
+        open={captcha.isModalOpen}
+        onWidgetContainerReady={captcha.onWidgetContainerReady}
+        error={captcha.error}
+        onRetry={captcha.refreshToken}
+      />
+    </>
+  )
+}
+
+export function CaptchaProvider({ children }: { children: ReactNode }) {
+  // Seed from module flag so a remounting provider resumes active state immediately
+  const [isActive, setIsActive] = useState(() => isCaptchaActivated())
+
+  useEffect(() => {
+    registerActivateCaptcha(() => setIsActive(true))
+    return () => registerActivateCaptcha(() => {})
+  }, [])
+
+  return (
+    <>
       {children}
-      {TURNSTILE_SITE_KEY && (
-        <>
-          <Script
-            src={TURNSTILE_SCRIPT_URL}
-            strategy="afterInteractive"
-            onReady={() => setIsScriptReady(true)}
-            onError={resolveCaptchaReady}
-          />
-          <CaptchaModal
-            open={captcha.isModalOpen}
-            onWidgetContainerReady={captcha.onWidgetContainerReady}
-            error={captcha.error}
-            onRetry={captcha.refreshToken}
-          />
-        </>
-      )}
+      {TURNSTILE_SITE_KEY && isActive && <CaptchaWidget />}
     </>
   )
 }
--- a/apps/web/src/components/common/Captcha/__tests__/CaptchaProvider.test.tsx
+++ b/apps/web/src/components/common/Captcha/__tests__/CaptchaProvider.test.tsx
@@ -1,0 +1,89 @@
+import { render, screen, act } from '@/tests/test-utils'
+import { CaptchaProvider } from '../CaptchaProvider'
+
+// Capture the activation callback registered by CaptchaProvider
+let registeredActivateCaptcha: (() => void) | null = null
+
+jest.mock('../captchaHeadersInit', () => ({
+  initializeCaptchaHeaders: jest.fn(),
+  resolveCaptchaReady: jest.fn(),
+  registerActivateCaptcha: jest.fn((fn: () => void) => {
+    registeredActivateCaptcha = fn
+  }),
+  isCaptchaActivated: jest.fn(() => false),
+}))
+
+jest.mock('@safe-global/utils/config/constants', () => ({
+  TURNSTILE_SITE_KEY: 'test-site-key',
+}))
+
+// Stub out heavy sub-components
+jest.mock('next/script', () => ({ __esModule: true, default: () => null }))
+jest.mock('../CaptchaModal', () => ({ __esModule: true, default: () => null }))
+jest.mock('../useCaptchaToken', () => ({
+  useCaptchaToken: () => ({
+    isModalOpen: false,
+    onWidgetContainerReady: jest.fn(),
+    error: null,
+    refreshToken: jest.fn(),
+  }),
+}))
+
+describe('CaptchaProvider', () => {
+  beforeEach(() => {
+    registeredActivateCaptcha = null
+  })
+
+  it('always renders children', () => {
+    render(
+      <CaptchaProvider>
+        <div data-testid="child">hello</div>
+      </CaptchaProvider>,
+    )
+    expect(screen.getByTestId('child')).toBeInTheDocument()
+  })
+
+  it('does not render the Turnstile widget before activation', () => {
+    const { container } = render(<CaptchaProvider>{null}</CaptchaProvider>)
+    // CaptchaWidget renders a Script + CaptchaModal — neither should be in the DOM yet
+    // We verify by checking that registerActivateCaptcha was called (provider registered its callback)
+    // and that no Script element was injected
+    expect(registeredActivateCaptcha).not.toBeNull()
+    // The Script mock renders null, so no next/script markup — container should only have children
+    expect(container.firstChild).toBeNull()
+  })
+
+  it('renders the Turnstile widget once the activation callback fires', async () => {
+    const { isCaptchaActivated } = require('../captchaHeadersInit')
+    ;(isCaptchaActivated as jest.Mock).mockReturnValue(false)
+
+    render(
+      <CaptchaProvider>
+        <div data-testid="child">hello</div>
+      </CaptchaProvider>,
+    )
+
+    expect(registeredActivateCaptcha).not.toBeNull()
+
+    // Simulate the prepareHeaders hook firing the activation callback
+    await act(async () => {
+      registeredActivateCaptcha!()
+    })
+
+    // Children still present
+    expect(screen.getByTestId('child')).toBeInTheDocument()
+  })
+
+  it('is immediately active when isCaptchaActivated returns true (provider remount)', () => {
+    const { isCaptchaActivated } = require('../captchaHeadersInit')
+    ;(isCaptchaActivated as jest.Mock).mockReturnValue(true)
+
+    render(
+      <CaptchaProvider>
+        <div data-testid="child">hello</div>
+      </CaptchaProvider>,
+    )
+
+    expect(screen.getByTestId('child')).toBeInTheDocument()
+  })
+})
--- a/apps/web/src/components/common/Captcha/__tests__/captchaHeadersInit.test.ts
+++ b/apps/web/src/components/common/Captcha/__tests__/captchaHeadersInit.test.ts
@@ -255,6 +255,101 @@ describe('captchaHeadersInit', () => {
   })
 
   // ---------------------------------------------------------------------------
+  // prepareHeaders hook — single-use token invalidation (lazy rotation)
+  // ---------------------------------------------------------------------------
+  describe('prepareHeaders hook — single-use token invalidation', () => {
+    it('clears the shared token after consuming it, without eagerly refreshing the widget', async () => {
+      const mockRefresh = jest.fn()
+      registerWidgetRefreshCallback(mockRefresh)
+      initializeCaptchaHeaders()
+      resolveCaptchaReady()
+      sharedTokenRef.current = 'first-token'
+
+      const hook = mockSetPrepareHeadersHook.mock.calls[0][0]
+      const headers = new Headers()
+      await hook(headers, PROTECTED_URL)
+
+      expect(headers.get('X-Captcha-Token')).toBe('first-token')
+      expect(sharedTokenRef.current).toBeNull()
+      // Lazy rotation: no refresh is triggered just because a token was consumed;
+      // the next hook invocation will refresh only if a new request actually arrives.
+      expect(mockRefresh).not.toHaveBeenCalled()
+    })
+
+    it('lazily refreshes the widget when a subsequent request finds no token', async () => {
+      // First run seeds and consumes 'first-token'.
+      const mockRefresh = jest.fn(() => {
+        sharedTokenRef.current = 'next-token'
+        resolveCaptchaReady()
+      })
+      registerWidgetRefreshCallback(mockRefresh)
+      initializeCaptchaHeaders()
+      resolveCaptchaReady()
+      sharedTokenRef.current = 'first-token'
+
+      const hook = mockSetPrepareHeadersHook.mock.calls[0][0]
+      const headersA = new Headers()
+      await hook(headersA, PROTECTED_URL)
+      expect(headersA.get('X-Captcha-Token')).toBe('first-token')
+      expect(mockRefresh).not.toHaveBeenCalled()
+
+      // Second request arrives after the token was consumed — widget refresh is triggered now.
+      const headersB = new Headers()
+      await hook(headersB, PROTECTED_URL)
+
+      expect(mockRefresh).toHaveBeenCalledTimes(1)
+      expect(headersB.get('X-Captcha-Token')).toBe('next-token')
+    })
+
+    it('does not trigger a widget refresh while an initial challenge is still in flight', async () => {
+      const mockRefresh = jest.fn()
+      registerWidgetRefreshCallback(mockRefresh)
+      initializeCaptchaHeaders()
+      // Do NOT call resolveCaptchaReady — the initial promise is still pending.
+
+      const hook = mockSetPrepareHeadersHook.mock.calls[0][0]
+      const hookPromise = hook(new Headers(), PROTECTED_URL)
+
+      // Give microtasks a chance to run; the lazy refresh must not fire because
+      // captchaReadyResolve is still set (challenge in flight).
+      await Promise.resolve()
+      expect(mockRefresh).not.toHaveBeenCalled()
+
+      // Resolve so the hook can complete and we don't leave a dangling promise.
+      sharedTokenRef.current = 'initial-token'
+      resolveCaptchaReady()
+      await hookPromise
+    })
+
+    it('serializes concurrent protected requests so each awaits its own fresh token', async () => {
+      // Simulate a widget refresh that produces a new token and resolves the promise
+      const mockRefresh = jest.fn(() => {
+        sharedTokenRef.current = 'next-token'
+        resolveCaptchaReady()
+      })
+      registerWidgetRefreshCallback(mockRefresh)
+      initializeCaptchaHeaders()
+
+      // Seed the first token and resolve the initial promise
+      sharedTokenRef.current = 'first-token'
+      resolveCaptchaReady()
+
+      const hook = mockSetPrepareHeadersHook.mock.calls[0][0]
+      const headersA = new Headers()
+      const headersB = new Headers()
+
+      // Fire both requests concurrently
+      const [,] = await Promise.all([hook(headersA, PROTECTED_URL), hook(headersB, PROTECTED_URL)])
+
+      // Each request must carry a distinct token
+      expect(headersA.get('X-Captcha-Token')).toBe('first-token')
+      expect(headersB.get('X-Captcha-Token')).toBe('next-token')
+      // Only B's lazy refresh fires — A consumed the pre-seeded token without refreshing.
+      expect(mockRefresh).toHaveBeenCalledTimes(1)
+    })
+  })
+
+  // ---------------------------------------------------------------------------
   // responseHook
   // ---------------------------------------------------------------------------
   describe('responseHook', () => {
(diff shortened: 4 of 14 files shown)


katspaugh (approved):
Looks like the PR description only picked up 2 out of 3 commits.

katspaugh:
## PR Briefing

[Watch the PR walkthrough](https://github.com/5afe/wallet-pr-artifacts/releases/download/prbrief-safe-global-safe-wallet-monorepo-pr-7713-62b0c4c36561/video-github.mp4)

▶ **[Watch the walkthrough video](https://github.com/5afe/wallet-pr-artifacts/releases/download/prbrief-safe-global-safe-wallet-monorepo-pr-7713-62b0c4c36561/video-github.mp4)** (opens MP4 in a new tab)

This hotfix release (web-v1.87.4) addresses two Turnstile captcha bugs: (1) the captcha modal was incorrectly popping up on silent token expiry in idle tabs, and (2) concurrent requests were reusing the same token causing 401 errors. The fix refactors `captchaHeadersInit.ts` and `useCaptchaToken.ts` to implement lazy, per-request token rotation and deferred widget activation, and isolates the Turnstile widget into a conditionally-mounted `CaptchaWidget` sub-component in `CaptchaProvider`. Additionally, five components (`SpaceNestedSafesButton`, `NestedSafesButton`, `NestedSafessList`, `SafeListContextMenu`, `ReviewNestedSafe`) and the `useOwnedSafes` hook are migrated from the v2 cross-chain owners endpoint (`useOwnersGetAllSafesByOwnerV2Query`) to the v1 per-chain endpoint (`useOwnersGetSafesByOwnerV1Query`) to avoid captcha-protected routes.

<details>

## Summary

This hotfix release (web-v1.87.4) addresses two Turnstile captcha bugs: (1) the captcha modal was incorrectly popping up on silent token expiry in idle tabs, and (2) concurrent requests were reusing the same token causing 401 errors. The fix refactors `captchaHeadersInit.ts` and `useCaptchaToken.ts` to implement lazy, per-request token rotation and deferred widget activation, and isolates the Turnstile widget into a conditionally-mounted `CaptchaWidget` sub-component in `CaptchaProvider`. Additionally, five components (`SpaceNestedSafesButton`, `NestedSafesButton`, `NestedSafessList`, `SafeListContextMenu`, `ReviewNestedSafe`) and the `useOwnedSafes` hook are migrated from the v2 cross-chain owners endpoint (`useOwnersGetAllSafesByOwnerV2Query`) to the v1 per-chain endpoint (`useOwnersGetSafesByOwnerV1Query`) to avoid captcha-protected routes.

## Architecture notes

The captcha subsystem now uses a module-level `activateCaptchaRef`/`captchaActivated` flag and a `consumeQueue` promise chain in `captchaHeadersInit.ts` to serialize token consumption across concurrent requests and restrict captcha injection to `CAPTCHA_PROTECTED_ROUTES` (v2/v3 owners/safes paths only). The `useCaptchaToken` hook no longer calls `resetCaptchaPromise()` or `refreshToken()` on token expiry, instead setting `refresh-expired: 'never'` so the widget stays dormant until the next protected request triggers a lazy refresh. The nested-safe query migration changes the response shape contract from a chainId-keyed map (`ownedSafes?.[chainId]`) to a flat array (`ownedSafes?.safes`), and now requires `chainId` as a query parameter, meaning each chain issues a separate network request instead of one cross-chain call.

## Ripple risks

Analyzed 6 changed exports. No untouched external callers found — all usages appear to be inside the PR diff.

_Analyzed 13 file(s), 9 changed export(s), 0 uncovered callsite(s) in 2.8s._

## Walkthrough outline

_19 steps across 6 chapters, approximately 4 min 15s._

### Why this hotfix exists

1. **motivation** · `apps/web/package.json` L1-5
   We're on hotfix 1.87.4 — a two-bug emergency drop on top of 1.87.3. Both bugs live in the Turnstile captcha subsystem, and they were quietly breaking users on idle tabs and concurrent requests.

2. **motivation** · `apps/web/src/components/common/Captcha/captchaHeadersInit.ts` L1-30
   This file is ground zero. Before the fix, every token expiry triggered a visible captcha modal, and concurrent requests all grabbed the same token causing 401s. The refactor here drives everything else.

### The new captcha activation model

3. **change** · `apps/web/src/components/common/Captcha/captchaHeadersInit.ts` L16-35 `registerActivateCaptcha`
   Two new module-level primitives: `registerActivateCaptcha` stores a callback reference, and `isCaptchaActivated` exposes a boolean flag. Together they let the provider and the header logic stay in sync without a shared React context.

4. **change** · `apps/web/src/components/common/Captcha/captchaHeadersInit.ts` L27-45 `isProtectedEndpoint`
   `isProtectedEndpoint` gates everything — only v2 and v3 owners-slash-safes paths get captcha headers injected. This is why migrating nested-safe queries to v1 immediately removes them from captcha scope.

5. **change** · `apps/web/src/components/common/Captcha/captchaHeadersInit.ts` L63-100 `initializeCaptchaHeaders`
   Inside `initializeCaptchaHeaders`, the `consumeQueue` promise chain is the key concurrency fix. Each protected request enqueues itself, grabs the current token, then releases the lock — so two simultaneous requests can never share a token.

6. **change** · `apps/web/src/components/common/Captcha/CaptchaProvider.tsx` L46-80 `CaptchaProvider`
   `CaptchaProvider` now splits the Turnstile widget into a child `CaptchaWidget` component that only mounts when `isActive` is true. On mount, it registers the activation callback; on unmount, it resets it to a no-op.

7. **change** · `apps/web/src/components/common/Captcha/CaptchaProvider.tsx` L30-46 `CaptchaWidget`
   The `isActive` seed comes from `isCaptchaActivated()` — so if captcha was already triggered before this provider mounted, the widget comes up immediately. Otherwise it stays hidden until the first protected request fires the callback.

8. **change** · `apps/web/src/components/common/Captcha/__tests__/CaptchaProvider.test.tsx` L1-50
   Four new unit tests cover the activation lifecycle: widget suppressed before activation, widget rendered after the callback fires, and immediate activation when the module flag is already set. Good regression net for this exact bug.

### Token lifecycle in the hook

9. **change** · `apps/web/src/components/common/Captcha/useCaptchaToken.ts` L45-80 `useCaptchaToken`
   The `expired-callback` handler is now just two lines — clear the token refs, done. No `resetCaptchaPromise`, no `refreshToken`. The widget sits dormant until `captchaHeadersInit` lazily wakes it on the next protected request.

10. **change** · `apps/web/src/components/common/Captcha/useCaptchaToken.ts` L55-70
   Notice `'refresh-expired': 'never'` added to the Turnstile render options. This tells the widget itself not to auto-challenge on expiry — a belt-and-suspenders guard on top of the callback change.

### Nested-safe queries move to v1

11. **ripple** · `apps/web/src/hooks/useOwnedSafes.ts` L1-30 `useOwnedSafes`
   Now the ripple: five components and this hook all swapped from `useOwnersGetAllSafesByOwnerV2Query` to `useOwnersGetSafesByOwnerV1Query`. The v1 endpoint isn't captcha-protected, so these queries never hit the Turnstile gate at all.

12. **ripple** · `apps/web/src/hooks/useOwnedSafes.ts` L15-28
   The response shape contract changed: v2 returned a chainId-keyed map, so callers wrote `ownedSafes?.[chainId]`. V1 returns a flat object with a `safes` array, so every accessor becomes `ownedSafes?.safes`. Simple, but it's in six places.

13. **ripple** · `apps/web/src/components/sidebar/NestedSafesButton/index.tsx` L16-35 `NestedSafesButton`
   `NestedSafesButton` is a representative example of the pattern. `chainId` is now passed as a query param, and the result is read from `.safes`. The skip guard only checks `safeAddress` — worth confirming `chainId` is always defined here.

14. **ripple** · `apps/web/src/components/settings/NestedSafesList/index.tsx` L22-45 `NestedSafesList`
   `NestedSafesList` and `ReviewNestedSafe` follow the same swap. Each now fires a per-chain request rather than one cross-chain call — more network round-trips, but no captcha friction.

15. **ripple** · `apps/web/src/components/common/SpaceSafeBar/SpaceNestedSafesButton.tsx` L1-30 `SpaceNestedSafesButton`
   `SpaceNestedSafesButton` and `SafeListContextMenu` complete the sweep. The test file for this component already updated its mock shape to `{ safes: [...] }`, which is a good sign the production and test changes are in sync.

### What could still go wrong

16. **risk** · `apps/web/src/components/common/Captcha/captchaHeadersInit.ts` L63-100 `initializeCaptchaHeaders`
   The `consumeQueue` promise chain has no timeout or cancellation path. If `captchaReadyPromise` never resolves — say the widget fails to load — every queued protected request deadlocks silently. Worth adding a timeout guard.

17. **risk** · `apps/web/src/components/common/Captcha/CaptchaProvider.tsx` L46-65 `CaptchaProvider`
   On unmount, the cleanup sets `registerActivateCaptcha(() => {})` — a no-op. If the provider remounts after captcha was already activated, `isCaptchaActivated()` returns true so `isActive` seeds correctly. But if it remounts before activation, the old callback reference is gone and the first protected request may silently fail to wake the widget.

18. **risk** · `apps/web/src/components/sidebar/SafeListContextMenu/index.tsx` L1-30
   Across all six v1 migrations, the skip guard doesn't check `chainId`. If `chainId` is somehow undefined at call time, the query fires without it — and the v1 endpoint requires it. A silent empty result is the likely outcome, not an error.

### Bottom line

19. **wrap** · `apps/web/package.json` L1-5
   In short, 1.87.4 kills the phantom captcha modal by making token refresh lazy and per-request, and sidesteps the captcha gate entirely for nested-safe queries by dropping them to the v1 endpoint — two targeted fixes, no new features, ship it.

## Reviewer checklist

- [ ] Verify that `captchaHeadersInit.ts`'s `consumeQueue` promise chain always executes its `finally` block to release the lock, and confirm there is a timeout or cancellation path if `captchaReadyPromise` never resolves to prevent request deadlocks.
- [ ] Confirm that all five migrated components and `useOwnedSafes` correctly consume `ownedSafes?.safes` from the v1 response shape, and that `chainId` is always defined before the query fires (skip guards are sufficient).
- [ ] Check that `CAPTCHA_PROTECTED_ROUTES` in `captchaHeadersInit.ts` covers all endpoints that require a captcha token, and that the v1 owners/safes endpoint used by the migrated components is intentionally excluded from captcha protection.
- [ ] Validate that after a 401 response, the lazy refresh path in `captchaHeadersInit.ts` correctly re-triggers a Turnstile challenge before the next retry, and that callers actually re-invoke `prepareHeaders` after receiving a 401.
- [ ] Review the `CaptchaProvider` unmount cleanup — `registerActivateCaptcha(() => {})` sets a no-op callback, which could prevent future activations if the provider remounts after captcha was already activated; confirm `isCaptchaActivated()` handles this case.
- [ ] Run the new unit tests in `CaptchaProvider.test.tsx` and `captchaHeadersInit.test.ts` and verify the third CaptchaProvider test actually asserts widget rendering (not just children), and that concurrent serialization tests reflect realistic async behavior.

## File-by-file

### `apps/web/package.json`

Version field in apps/web/package.json bumped from 1.87.3 to 1.87.4.

_Marks the hotfix release containing Turnstile captcha bug fixes._

### `apps/web/src/components/common/Captcha/CaptchaProvider.tsx`

CaptchaProvider is refactored to split the Turnstile widget into an isolated `CaptchaWidget` sub-component that is only mounted when `isActive` is true. `isActive` is seeded from a new `isCaptchaActivated()` module-level flag and toggled via a `registerActivateCaptcha` callback registered in a `useEffect`.

_Prevents `useCaptchaToken` (and the Turnstile script/modal) from running on every render of the provider, so silent token expiry no longer triggers the captcha modal unless captcha has been explicitly activated._

**Risks:**
- Check `registerActivateCaptcha` and `isCaptchaActivated` implementations in captchaHeadersInit — the cleanup `registerActivateCaptcha(() => {})` on unmount could clear a legitimate activation callback if multiple providers exist or the component remounts
- Verify that `isCaptchaActivated()` correctly reflects prior activation state across remounts to avoid missing the active state on re-render
- Confirm that deregistering the callback on unmount (setting it to a no-op) does not prevent future activations if the provider remounts after captcha was already activated
- Ensure the `useEffect` dependency array being empty `[]` is intentional and that `registerActivateCaptcha` is stable/referentially equal across renders

### `apps/web/src/components/common/Captcha/__tests__/CaptchaProvider.test.tsx`

Adds a new test file `CaptchaProvider.test.tsx` with four unit tests covering `CaptchaProvider`: child rendering, pre-activation widget suppression, post-activation widget rendering via `registeredActivateCaptcha` callback, and immediate activation when `isCaptchaActivated` returns true.

_Likely added to provide regression coverage for the Turnstile captcha activation flow changes introduced in the hotfix PRs #7694 and #7690._

**Risks:**
- Third test ('renders the Turnstile widget once activation callback fires') only asserts children are present but never verifies the widget itself rendered, weakening its value as a widget-activation test
- isCaptchaActivated mock is reset per-test via require() inside tests but the module-level mock default always returns false — ordering dependency could cause flakiness if test isolation breaks
- registeredActivateCaptcha is captured via side-effect in mock; if CaptchaProvider registers the callback asynchronously or conditionally, the captured reference could be stale or null

### `apps/web/src/components/common/Captcha/__tests__/captchaHeadersInit.test.ts`

Test file for `captchaHeadersInit` expanded from ~10 tests to ~30+, adding coverage for: `isProtectedEndpoint` URL matching, URL-scoped header/response hook behavior (protected vs non-protected), lazy token rotation (clearing token after use without eager widget refresh), `registerActivateCaptcha`/`isCaptchaActivated` lifecycle, concurrent request serialization, and updated existing tests to pass a URL argument to hooks.

_Tests document and verify new behavior where captcha tokens are only injected for protected endpoints (v2+/owners/.../safes pattern), tokens are lazily rotated per-request rather than eagerly on 401, and the widget activation is triggered only once on first protected request._

**Risks:**
- Verify the concurrent serialization test accurately reflects real race conditions — the mock refresh synchronously sets the next token and resolves the promise, which may not match async real-world behavior
- Check that `isProtectedEndpoint` regex/pattern in production code correctly excludes `/v1/chains/1/owners/0xABC/safes` (the test asserts false but the path contains 'owners' and 'safes')
- The 401 response hook test now asserts `mockRefresh` is NOT called — confirm the production code change aligns and that 401s won't leave the widget in a broken state without a refresh path
- Ensure `registerActivateCaptcha` and `isCaptchaActivated` are properly exported from the module under test, not just tested in isolation

### `apps/web/src/components/common/Captcha/__tests__/useCaptchaToken.test.ts`

Test file removes `resetCaptchaPromise` from mock imports/setup, adds a test asserting `refresh-expired` is set to `'never'` in Turnstile render options, and updates the token-expiry test to assert that `mockTurnstile.reset` is NOT called on silent TTL expiry (replacing the previous assertion that `resetCaptchaPromise` and `mockTurnstile.reset` were called).

_Tests now verify the new 'lazy rotation' behavior where the Turnstile widget does not auto-challenge or reset on silent token expiry, preventing unexpected modal popups; a fresh challenge is deferred until the next protected request._

**Risks:**
- Verify the production code (`useCaptchaToken`) actually passes `refresh-expired: 'never'` to Turnstile render and that `resetCaptchaPromise` is no longer called on expiry
- Check that deferred/lazy token refresh in `captchaHeadersInit` correctly triggers a new challenge before the next API request, not after it fails
- Ensure removal of `resetCaptchaPromise` from the mock doesn't hide cases where it is still called unexpectedly in other code paths

### `apps/web/src/components/common/Captcha/captchaHeadersInit.ts`

Refactored `captchaHeadersInit.ts` to: (1) add per-request token rotation via a `consumeQueue` promise chain so concurrent requests each get a unique Turnstile token, (2) restrict captcha header injection and 401 handling to only `CAPTCHA_PROTECTED_ROUTES` (`/v2/owners/.../safes` and `/v3/owners/.../safes`), (3) add lazy captcha activation via `activateCaptchaRef`/`captchaActivated` so the widget only loads on the first protected request, and (4) remove the automatic `resetCaptchaPromise()`+`widgetRefreshCallbackRef` call on 401 (now only clears `sharedTokenRef.current`).

_Fixes Turnstile modal popping on silent token expiry by decoupling 401 handling from widget refresh, and prevents token reuse 401s by serializing token consumption across concurrent requests._

**Risks:**
- consumeQueue grows unbounded if `release()` is never called — verify the `finally` block always executes even if `captchaReadyPromise` rejects or hangs indefinitely
- If `captchaReadyPromise` never resolves (e.g., widget never loads), all queued protected requests will deadlock — check if there is a timeout or cancellation path
- CAPTCHA_PROTECTED_ROUTES only covers v2 and v3 owners/safes — verify no other protected endpoints exist or will be added
- After 401, `sharedTokenRef.current` is cleared but no new challenge is triggered; the next request lazily triggers refresh — confirm the retry logic in the caller actually re-invokes `prepareHeaders` after a 401
- `captchaActivated` is a module-level boolean that is never reset — in test environments or SSR this could cause stale state across test cases
- The `setPrepareHeadersHook` signature now expects `(headers, url)` — verify the CGW client actually passes the URL as the second argument

### `apps/web/src/components/common/Captcha/useCaptchaToken.ts`

In `useCaptchaToken.ts`, the `expired-callback` handler is simplified to only clear the token (`sharedTokenRef.current = null`, `setToken(null)`) without calling `resetCaptchaPromise()` or `refreshToken()`. Additionally, `'refresh-expired': 'never'` is added to the Turnstile widget options, and `resetCaptchaPromise` is removed from the import list.

_Prevents the Turnstile modal from silently popping up on idle tabs when the ~5-minute token TTL expires, deferring token refresh until an actual protected request needs one._

**Risks:**
- If a protected request fires immediately after token expiry and before `refreshToken()` is triggered via `captchaHeadersInit`, verify the lazy refresh path correctly re-initiates the challenge and doesn't silently fail or hang.
- Removing `resetCaptchaPromise()` from the expired-callback means any pending promise waiting on captcha readiness won't be reset on expiry — confirm `captchaHeadersInit` handles this stale promise state correctly.
- With `'refresh-expired': 'never'`, the widget stays dormant indefinitely after expiry; ensure `turnstile.reset()` in `refreshToken()` reliably re-triggers the challenge when called.
- Verify `resetCaptchaPromise` is still exported from `captchaHeadersInit` and used elsewhere, or confirm it's now dead code that should be removed.

### `apps/web/src/components/common/SpaceSafeBar/SpaceNestedSafesButton.test.tsx`

Test file updated to replace `useOwnersGetAllSafesByOwnerV2Query` with `useOwnersGetSafesByOwnerV1Query` in mocks and imports, and changed mock return data shape from `{ '1': ['0xNested1', '0xNested2'] }` to `{ safes: ['0xNested1', '0xNested2'] }`.

_Reflects a production code change that scopes nested-safe owner queries to the v1 API endpoint, which returns a flat `safes` array instead of a chain-keyed object._

**Risks:**
- Verify the production component `SpaceNestedSafesButton` was also updated to use `useOwnersGetSafesByOwnerV1Query` and consume `currentData.safes` instead of `currentData[chainId]`
- Check if v1 endpoint returns all relevant chains or only a single chain, which could affect multi-chain nested safe detection
- Ensure no other consumers of `useOwnersGetAllSafesByOwnerV2Query` were missed in the migration

### `apps/web/src/components/common/SpaceSafeBar/SpaceNestedSafesButton.tsx`

Replaces `useOwnersGetAllSafesByOwnerV2Query` with `useOwnersGetSafesByOwnerV1Query` in `SpaceNestedSafesButton`, adding `chainId` to the query params and changing the result accessor from `ownedSafes?.[chainId]` to `ownedSafes?.safes`.

_Scopes the nested-safe owner query to a specific chain using the v1 endpoint, likely to avoid cross-chain data being returned and to fix issues introduced by the v2 multi-chain response shape._

**Risks:**
- Verify `useOwnersGetSafesByOwnerV1Query` response shape actually has a `safes` array field matching what `useNestedSafesVisibility` expects
- Confirm `chainId` is always defined at call site before the query fires (skip guard only checks `safeAddress`)
- Check if v1 endpoint is available for all supported chains that have NESTED_SAFES feature enabled
- Ensure no other consumers of the v2 query were relying on the same cross-chain data aggregation behavior

### `apps/web/src/components/settings/NestedSafesList/index.tsx`

Replaces `useOwnersGetAllSafesByOwnerV2Query` (cross-chain, no chainId param) with `useOwnersGetSafesByOwnerV1Query` (single-chain, requires `chainId`), updates the query args to include `safe.chainId`, and changes the result accessor from `ownedSafes?.[safe.chainId]` to `ownedSafes?.safes`.

_Scopes the nested-safe owner lookup to the current chain using the v1 endpoint, likely to avoid issues with the v2 endpoint (e.g., auth/captcha problems or incorrect cross-chain data being returned)._

**Risks:**
- Verify `useOwnersGetSafesByOwnerV1Query` response shape actually has a `.safes` array matching the expected element type
- Confirm v1 endpoint is available and stable for all supported chains
- Check whether `safe.chainId` is reliably populated before `safeLoaded` is true, since it's now passed as a query arg
- Ensure no other consumers relied on the v2 cross-chain aggregation behavior for this component

### `apps/web/src/components/sidebar/NestedSafesButton/index.tsx`

Replaces `useOwnersGetAllSafesByOwnerV2Query` (v2 endpoint, no chainId param, response keyed by chainId) with `useOwnersGetSafesByOwnerV1Query` (v1 endpoint, chainId param required, response has `.safes` array). Data extraction changes from `ownedSafes?.[chainId]` to `ownedSafes?.safes`.

_Scopes the nested-safe owner query to a specific chain via the v1 endpoint, likely to fix incorrect cross-chain safe results or an issue with the v2 endpoint returning unexpected data._

**Risks:**
- Verify v1 endpoint actually filters by chainId server-side and returns only safes on that chain
- Confirm `ownedSafes?.safes` shape matches the v1 response type — wrong field name would silently yield empty array
- Check if v2 was intentionally used for multi-chain support that is now lost
- Ensure `chainId` is always defined at call site when `isEnabled && safeAddress` are truthy

### `apps/web/src/components/sidebar/SafeListContextMenu/index.tsx`

Replaces `useOwnersGetAllSafesByOwnerV2Query` (v2, no chainId param, returns map keyed by chainId) with `useOwnersGetSafesByOwnerV1Query` (v1, accepts `chainId` param, returns `{ safes: [] }`), and updates the result accessor from `ownedSafes?.[chainId]` to `ownedSafes?.safes`.

_Scopes nested-safe owner queries to a single chain via the v1 endpoint, likely to avoid cross-chain data leakage or a bug where the v2 endpoint was returning unexpected/empty results for the current chain._

**Risks:**
- Verify v1 endpoint actually exists and is stable in the gateway API for all supported chains
- Confirm `ownedSafes?.safes` shape matches the v1 response type — a mismatch would silently yield an empty array
- Check whether v2 was used intentionally for multi-chain support that is now lost
- Ensure the `chainId` value passed is always defined when the query is not skipped (skip condition does not guard on chainId)

### `apps/web/src/components/tx-flow/flows/CreateNestedSafe/ReviewNestedSafe.tsx`

Replaces `useOwnersGetAllSafesByOwnerV2Query` (all-chains v2 endpoint) with `useOwnersGetSafesByOwnerV1Query` (single-chain v1 endpoint), adding `chainId` to the query params and changing the data accessor from `ownedSafes?.[safe.chainId]` to `ownedSafes?.safes`.

_Scopes the nested-safe ownership query to the current chain only using the v1 API, likely to avoid issues with the v2 endpoint (e.g., auth/captcha problems or incorrect multi-chain data being used)._

**Risks:**
- Verify `useOwnersGetSafesByOwnerV1Query` response shape actually has a `.safes` array matching what `nestedSafes` consumers expect
- Confirm v1 endpoint is stable and returns equivalent data for the current chain as the v2 endpoint did
- Check if `safe.chainId` is always defined/populated when `safeLoaded` is true to avoid skipped-but-needed queries
- Ensure no other callers relied on the v2 all-chains behavior in this component path

### `apps/web/src/hooks/useOwnedSafes.ts`

Replaced `useOwnersGetAllSafesByOwnerV2Query` (v2, cross-chain, no chainId param) with `useOwnersGetSafesByOwnerV1Query` (v1, per-chain, requires `chainId`), and updated the result mapping from `ownedSafes?.[chainId]` to `ownedSafes?.safes`.

_Scopes owned-safe queries to a single chain using the v1 endpoint, likely to avoid a v2 endpoint issue (e.g., captcha/auth problems or incorrect cross-chain data being returned)._

**Risks:**
- Downgrading from v2 to v1 means each chain now requires a separate network request; verify no callers relied on the cross-chain aggregation behavior of v2
- Check that `ownedSafes?.safes` is the correct response shape for the v1 endpoint and won't silently return undefined
- Confirm `customChainId` override still works correctly now that chainId is passed into the query
- Verify cache/invalidation behavior differs between v1 and v2 queries and won't cause stale data issues

---
_Generated by prbrief • model: claude-sonnet-4-6 • 16 LLM calls • $0.24 • 88.3s_

</details>

<sub>Generated locally by <a href="https://github.com/anthropics/claude-code">prbrief</a>. · `safe-global/safe-wallet-monorepo@62b0c4c36561`</sub>

PR #7714, opened 2026-04-22
feat(gtf): phase 2 — pay gas from Safe

Zalens:
## What it solves

GTF Phase 2: pay gas fees from the Safe in a user-chosen gas token. Adds the fee-preview flow, Safe-vs-signer payment toggle, CGW integration for fee resolution, and all downstream wiring so confirmers see the same locked fee details the first signer approved.

Resolves:  https://linear.app/safe-global/issue/PLA-1575/gtf-relay-is-available-for-pay-from-signer-txs-with-unlimited-amount

## How this PR fixes it

- **Fee preview on Create/Review step** (`features/gtf/components/FeesPreview`) with payment-source toggle, gas-token dropdown, fiat display, and MAX-deduction banner.
- **CGW wiring** via new `useGetGtfFeePreviewQuery` (`store/api/gateway/gtfFeePreview.ts`) — returns `{ safeTxGas, baseGas, gasPrice, gasToken, refundReceiver, relayCostUsd, ... }`.
- **Merge at first-signer sign** (`features/gtf/services/resolveFeeParams.ts` + `components/tx/shared/hooks.ts`): fresh CGW call right before signing, merges fee params into `safeTx.data`, signs the merged payload.
- **Execute step** (`components/tx-flow/actions/Execute/ExecuteForm.tsx`): forces Gelato relay when `isGtfSafePaid(safeTx.data)` or pre-merge Safe-pays context is set (wallet execution would double-charge); hides the legacy "Estimated fee" accordion for Safe-pays; skips `useIsValidExecution` which otherwise reverts with GS026 on the wrong simulation path.
- **Receipt** (`components/tx/ConfirmTxDetails/Receipt.tsx`): shows the selected gas token pre-sign, and the locked token post-sign.
- **History**: `useHistoryFeesBreakdown` + `HistoryFeesAccordion` surface the fee details on executed txs.
- **Schema sync**: regenerated `AUTO_GENERATED/` from the updated CGW schema (adds `safeTxHash` to relay DTO, drops `contractAddresses` from `Chain`, etc.).

## How to test it

1. Enable the `GTF` feature flag for a chain in chain-config (CGW).
2. Create a tx from a Safe on that chain — see the new Fees Preview step with Safe / Signer toggle.
3. Pick a non-native gas token, confirm fiat and total outgoing update.
4. Sign & Execute (threshold=1) or Sign → second signer Sign & Execute (threshold>1) — verify:
   - Fees shown to the second signer match the first signer's (locked from the signed payload, not re-queried).
   - No "Estimated fee" accordion on the Execute step for Safe-pays.
   - No "This transaction will most likely fail" warning for Safe-pays.
5. After execution, check the tx-history detail view — fee breakdown accordion renders.

## Checklist

- [x] Tests for all new logic
- [x] Feature behind `FEATURES.GTF` chain flag
- [x] Schema drift cleanup — back-merge with dev


yagopv:
I cannot read very well the NEW in dark mode in the designs. I think colors are different from the designs

image
image

In light mode, the colors seems different as well

yagopv:
When accessing the send transaction screen and selecting the destination, the `/preview` endpoint is called even though we are not displaying any fees yet

image

After clicking "Next," the endpoint is called again. Is this first call intentional ?

apps/web/src/components/tx-flow/actions/Execute/ExecuteForm.tsx
```diff
@@ -215,6 +240,11 @@ export const ExecuteForm = ({
           <ErrorMessage>
             Cannot execute a transaction from the Safe Account itself, please connect a different account.
           </ErrorMessage>
+        ) : relayUnavailableForGtf ? (
+          <ErrorMessage>
+            Safe-paid fees require Gelato relay, which is currently unavailable. Cancel and retry this transaction, or
+            wait for relay quota to reset.
+          </ErrorMessage>
```
yagopv: Telling the user to cancel and retry without offering "switch to signer pays" feels like a dead end. Could we consider an inline action that flips back to signer ?

yagopv:
If I choose to pay from the signer and, after signing with the first signer, I go to the second signer, I see this on the "Confirm Transaction" screen:

image

Shouldn't be like that right ? In any case the feature works because the fees are being paid from the signer but the information i think is wrong

Zalens:
https://linear.app/safe-global/issue/PLA-925/gtf-safewallet-phase-2-show-fee-in-transaction-history

Zalens:
https://linear.app/safe-global/issue/PLA-923/gtf-safewallet-phase-2-show-fee-preview-in-transaction-flow

apps/web/src/features/gtf/hooks/useFeesPreview.ts
```diff
@@ -1,30 +1,328 @@
...
+        to: safeTx.data.to,
+        value: safeTx.data.value,
+        data: safeTx.data.data,
+        operation: safeTx.data.operation,
+      }
+    : undefined
+
+  // Once the first signer has signed, the gas token is baked into the signed payload — later
+  // signers can't change it, so we skip the candidate probing and just render what's locked in.
+  const lockedGasToken = safeTx && safeTx.signatures.size > 0 ? safeTx.data.gasToken : undefined
```
yagopv, line 99: For pre-gtf transactions (transaction in queue), and if safeTx.data.gasToken is 0s I think this can be identified as a gtf transaction. Was this tested ? 🤔

Perhaps some protection mechanism can be added as 

```
const isConfirmation = lockedGasToken !== undefined && isGtfSafePaid(safeTx.data)
```

apps/web/src/features/gtf/services/resolveFeeParams.ts
```diff
@@ -0,0 +1,55 @@
...
+      {
+        chainId,
+        safeAddress,
+        tx: { to, value, data, operation, gasToken, numberSignatures },
+      },
+      { forceRefetch: true },
+    ),
+  ).unwrap()
+
+  const { safeTxGas, baseGas, gasPrice, gasToken: resolvedGasToken, refundReceiver } = preview.txData
```
yagopv, line 45: Could be worth to have some refundReceiver whitelisting client side just in case CGW is compromised in some way. Another idea (complementary) can be to show a warning in case the refundReceiver is unknown

Zalens:
> I cannot read very well the NEW in dark mode in the designs. I think colors are different from the designs
> 
> image image
> In light mode, the colors seems different as well

fixed in [0e344f11d](https://github.com/safe-global/safe-wallet-monorepo/commit/0e344f11d) — dropped the web info.* override in the css-vars generator so we now inherit the figma-aligned values directly from the unified palette source, no hardcoded hex anywhere.

Zalens:
> When accessing the send transaction screen and selecting the destination, the `/preview` endpoint is called even though we are not displaying any fees yet
> 
> image
> After clicking "Next," the endpoint is called again. Is this first call intentional ?

                                                           
The call is "intentional" only because it used to get the early Continue-block when no held token can pay gas. But that fires for very few users, and we pay the network cost for everyone, on every recipient edit.
                                                                                                                                                                                       
We should:                                                                                                                                                                        
- drop the probe from RecipientRow
- let useFeesPreview on review surface the blocked state                                                                                                                                        
- when blocked on review, show inline "switch to signer" — kills your ExecuteForm.tsx:247 dead-end comment too                                                                                  
- lazy-fire the max banner on click

apps/web/src/components/tx-flow/actions/Execute/ExecuteForm.tsx
Zalens: Yes I agree, this is a dead end. The easy fix is replacing the message with an inline button that calls setGtfPaymentMode('signer') and clears gtfSelectedGasToken. Ok for you?

Zalens:
> If I choose to pay from the signer and, after signing with the first signer, I go to the second signer, I see this on the "Confirm Transaction" screen:
> 
> image
> Shouldn't be like that right ? In any case the feature works because the fees are being paid from the signer but the information i think is wrong

Yes this is a display bug — `isConfirmation` in useFeesPreview.ts:157 only checks gasToken !== undefined, adding && isGtfSafePaid(safeTx.data) fixes it

Ok for you?

apps/web/src/features/gtf/hooks/useFeesPreview.ts
Zalens: good catch, you're right. I will add && isGtfSafePaid(safeTx.data), also fixes the second-signer Confirm-screen display bug from your other comment.

apps/web/src/features/gtf/services/resolveFeeParams.ts
Zalens: Yes good point. I will whitelist Gelato's fee-collector address per chain and reject on mismatch.

yagopv:
Question: 

image

When the user switches from Safe Pays to Signer, I cannot see any reestimation of the gas. Shouldn't there be one? 🤔 

And when the Safe has several signers, currently in prod no estimation is being showed. Shouldn't we default to the same behavior when the user change to the Signer flow ?

apps/web/src/components/tx-flow/actions/Execute/ExecuteForm.tsx
yagopv: ok!

apps/web/src/components/tx-flow/actions/Execute/ExecuteForm.tsx
```diff
@@ -70,27 +72,42 @@ export const ExecuteForm = ({
...
+  // Safe-pays bypasses the no-fee campaign and the daily relay quota (Safe funds its own relay).
+  const canRelay =
+    walletCanRelay &&
+    (requiresRelay || ((!isNoFeeCampaign || !isNoFeeCampaignEnabled) && hasRemainingRelays(relays[0])))
   const canNoFeeCampaign =
-    isNoFeeCampaignEnabled && isNoFeeCampaign && !blockedAddress && !gasTooHigh && !!remaining && remaining > 0
+    !requiresRelay &&
+    isNoFeeCampaignEnabled &&
+    isNoFeeCampaign &&
+    !blockedAddress &&
```
yagopv: Repeated pattern. Better to extract to their own variable:

const noFeeCampaignEligible = isNoFeeCampaignEnabled && isNoFeeCampaign && !blockedAddress

apps/web/src/features/gtf/hooks/useHistoryFeesBreakdown.ts
```diff
@@ -0,0 +1,133 @@
...
+  const refundReceiver = exec?.refundReceiver?.value
+  const { txHash, executedAt } = txDetails
+
+  const isSafePaid =
+    !!gasPrice &&
+    !!baseGas &&
+    !!refundReceiver &&
+    BigInt(gasPrice) > 0n &&
+    BigInt(baseGas) > 0n &&
+    !sameAddress(refundReceiver, ZERO_ADDRESS)
```
yagopv: This is duplicated logic with `isGtfSafePaid` right?

Zalens:
> Question:
> 
> image
> When the user switches from Safe Pays to Signer, I cannot see any reestimation of the gas. Shouldn't there be one? 🤔
> 
> And when the Safe has several signers, currently in prod no estimation is being showed. Shouldn't we default to the same behavior when the user change to the Signer flow ?

Yes right on both — Signer mode should swap to a local EOA gas estimate (or drop the row entirely for multi-signer to match pre-GTF prod behavior)

apps/web/src/components/tx-flow/actions/Execute/ExecuteForm.tsx
Zalens: agreed, will extract to noFeeCampaignEligible and reuse.

apps/web/src/features/gtf/hooks/useHistoryFeesBreakdown.ts
Zalens: will swap to isGtfSafePaid (and bump it to use sameAddress for the zero-address check while we're at it).

yagopv:
I did a transfer using USDC to pay the fees:

image

The result in transaction matches the displayed fee information

image

But the reality is that the balance change was 0,08536, which does not match the displayed fee (although it matches the balance change box)

Zalens:
> I did a transfer using USDC to pay the fees:
> 
> image
> The result in transaction matches the displayed fee information
> 
> image
> But the reality is that the balance change was 0,08536, which does not match the displayed fee (although it matches the balance change box)

The gas-fee row is computed from the signed (safeTxGas + baseGas) × gasPrice budget, while the Balance change row reads the actual gas-token outflow from the SafeShield / Tenderly simulation

yagopv (approved):

dasanra:
@claude review

PR #7494, opened 2026-03-20
refactor: zero-release chain rollouts — chain-agnostic contract resolution

Zalens:
## Summary

Resolves contract addresses by **version + deployment type** instead of chainId. New chains work by adding CGW config — no npm releases needed.

- `resolveChainAgnosticContractAddresses()` — resolves all 8 contracts for SDK init (web + mobile)
- `getChainAgnosticAddress()` — per-chain-then-canonical fallback for Safe creation
- Migrated remaining lookups: recovery, spending limits, migrations, tx-builder

**Closes PLA-1122, PLA-1123, PLA-1124**

## Testing

- 17 new unit tests, 3 updated
- All suites pass: safeCoreSDK (15), deployments (37), recovery-state (21), create/logic (12)
- Type-check + lint clean

[RFC](https://www.notion.so/safe-global/RFC-Zero-Wallet-release-chain-rollouts-31a8180fe57380a8bffae09cb65b2fb4)

PR #7719, opened 2026-04-22
fix: resolve Safe aux contracts by master copy on zkSync and Lens

compojoom:
> Master copy picks the flavour,
> aux contracts follow its lead,
> EraVM stays with EraVM,
> EVM with EVM.

## What it solves

PR #7494 introduced a chain-agnostic contract resolver gated on CGW chain-level `zk` / `l2` flags. This broke aux-contract resolution on zkSync-family chains in four ways: canonical-master-copy Safes on zkSync Era at 1.3.0 lost their MultiSend override; 1.4.1 Lens Safes with a zksync master copy resolved canonical addresses because Lens isn't flagged `zk: true` on CGW; 1.4.1 zkSync Safes running an L1 canonical master copy resolved against the L2 singleton table; and 1.3.0 Lens Safes with a zksync master copy were never covered by the canonical-override helper (hard-coded to zkSync Era mainnet). The net effect: multisend and simulations didn't work on zkSync / Lens chains.

Resolves: https://linear.app/safe-global/issue/WA-2000/multisend-and-simulations-not-working-on-zksync-lens-chains

## How this PR fixes it

The deployment flavour (`canonical` / `zksync` / `eip155`) and L1-vs-L2 flavour of a Safe are properties of its master copy, not of its chain. The fix derives both from the master copy address directly:

- `getDeploymentTypeForMasterCopy` now scans BOTH L1 and L2 singleton tables and returns `{ deploymentType, isL1 }`. This is what fixes the zkSync-L1-master-copy case — the resolver now knows to use the L1 singleton table even though the chain is an L2.
- `getChainAgnosticAddress` now prefers the requested deployment-type address when it's registered in `networkAddresses[chainId]`. Required because Lens registers both canonical and zksync addresses for the same contract — the old code always picked canonical.
- `isEraVmChain` generalises the old `SUPPORTED_ZKSYNC_CANONICAL_CHAIN_IDS = {324}` check to any chain whose `networkAddresses[chainId]` contains the `zksync` deployment address. Lens (232) and zkSync Sepolia (300) now participate.
- The canonical MultiSend / MultiSendCallOnly override (deleted in #7494 as "subsumed by the resolver") is restored for versions below the chain-agnostic threshold — that threshold is `>=1.4.1`, so 1.3.0 Safes bypass the resolver entirely and still need the override.

Chain-level flags (`chain.zk`, `chain.l2`) are now only used as defaults when the master copy can't be matched (custom / unrecognised deployments).

## How to test it

Verify on-chain that each of these Safes executes a multisend-only tx to the expected aux-contract address after the fix:

1. `zksync:0x1b8966A16021776a8c553603eC33DA0Ab05ce239` (zkSync Era, 1.3.0 + canonical L2 master copy) → MultiSendCallOnly `0x40A2aCCbd92BCA938b02010E17A5b8929b49130D` (canonical)
2. `zksync:0x4E868F1858de0377E8305A0421fC0f67bd89F153` (zkSync Era, 1.4.1 + canonical L1 master copy) → singleton resolves from L1 table, canonical MultiSendCallOnly `0x9641d764fc13c8B624c04430C7356C1C7C8102e2`
3. `lens:0x7ba5d7F189d3F24F536ddc49EBEb987332B9489c` (Lens, 1.4.1 + canonical L2 master copy) → canonical MultiSendCallOnly `0x9641d764fc13c8B624c04430C7356C1C7C8102e2`
4. `lens:0x49b0EA5B0691D460F2026EF98CfB3819E2e9f97d` (Lens, 1.4.1 + zksync L2 master copy) → zksync MultiSendCallOnly `0x0408EF011960d02349d50286D20531229BCef773`
5. `lens:0x2b73683bB2ef3B0a1F5b520cB917a58B875Ae59D` (Lens, 1.3.0 + zksync L2 master copy) → zksync MultiSendCallOnly `0xf220D3b4DFb23C4ade8C88E526C1353AbAcbC38F`

Automated coverage: `yarn workspace @safe-global/utils test packages/utils/src/services/contracts/__tests__/deployments.test.ts` — 62 passing tests covering L1/L2 × canonical/zksync across 1.3.0 and 1.4.1, plus the EraVM-chain detection.

## Visual summary

```mermaid
flowchart TB
  subgraph Before["Before (broken)"]
    B1[initSafeSDK] --> B2{"version >= 1.4.1?"}
    B2 -->|yes| B3["deploymentType = isZk ? zksync : canonical"]
    B2 -->|no| B4[no override]
    B3 --> B5["resolver picks by deploymentType<br/>(ignores networkAddresses)"]
    B5 --> B6["wrong aux on Lens, wrong singleton<br/>on zkSync L1-masterCopy Safe"]
  end

  subgraph After["After (fixed)"]
    A1[initSafeSDK] --> A2["getDeploymentTypeForMasterCopy<br/>(impl, version)"]
    A2 --> A3["{deploymentType, isL1}<br/>derived from master copy"]
    A3 --> A4[resolver uses L1/L2 table<br/>per isL1, picks deploymentType<br/>from networkAddresses]
    A4 --> A5{"isEraVmChain and<br/>canonical master copy?"}
    A5 -->|yes| A6[override MultiSend aux<br/>with canonical]
    A5 -->|no| A7[done]
  end
```

## Screenshots

N/A — no UI changes.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/hooks/coreSDK/safeCoreSDK.ts
+++ b/apps/mobile/src/hooks/coreSDK/safeCoreSDK.ts
@@ -84,8 +84,10 @@ export const initSafeSDK = async ({
 
   // zkSync Safes using a canonical (EVM bytecode) master copy cannot delegatecall
   // the zksync-specific (EraVM) MultiSend/MultiSendCallOnly, so force the canonical
-  // aux-contract addresses. Covers versions below the chain-agnostic threshold (<1.4.1).
-  if (isCanonicalDeployment(implementation, chainId, safeVersion)) {
+  // aux-contract addresses. Only runs for versions below the chain-agnostic threshold
+  // (<1.4.1); for >=1.4.1 the chain-agnostic resolver already picks the correct flavour
+  // from the master copy, and a second writer on the same fields would only risk drift.
+  if (!isChainAgnosticVersion(safeVersion) && isCanonicalDeployment(implementation, chainId, safeVersion)) {
     const canonicalMultiSendCallOnly = getCanonicalMultiSendCallOnlyAddress(safeVersion)
     const canonicalMultiSend = getCanonicalMultiSendAddress(safeVersion)
 
--- a/apps/mobile/src/hooks/coreSDK/safeCoreSDK.ts
+++ b/apps/mobile/src/hooks/coreSDK/safeCoreSDK.ts
@@ -8,6 +8,10 @@ import { isValidMasterCopy } from '@safe-global/utils/services/contracts/safeCon
 import type { SafeCoreSDKProps } from '@safe-global/utils/hooks/coreSDK/types'
 import { isInDeployments } from '@safe-global/utils/hooks/coreSDK/utils'
 import {
+  getCanonicalMultiSendAddress,
+  getCanonicalMultiSendCallOnlyAddress,
+  getDeploymentTypeForMasterCopy,
+  isCanonicalDeployment,
   isChainAgnosticVersion,
   resolveChainAgnosticContractAddresses,
 } from '@safe-global/utils/services/contracts/deployments'
@@ -42,11 +46,15 @@ export const initSafeSDK = async ({
 
   // For versions >= 1.4.1, resolve all addresses chain-agnostically (works on any chain)
   if (isChainAgnosticVersion(safeVersion) && isL2Chain !== undefined) {
-    const resolved = resolveChainAgnosticContractAddresses(safeVersion, isL2Chain, isZkChain ?? false)
+    const { deploymentType, isL1 } = getDeploymentTypeForMasterCopy(implementation, safeVersion, {
+      deploymentType: isZkChain ? 'zksync' : 'canonical',
+      isL1: !isL2Chain,
+    })
+    const resolved = resolveChainAgnosticContractAddresses(chainId, safeVersion, !isL1, deploymentType)
 
     if (resolved) {
       contractNetworks = { [chainId]: resolved }
-      isL1SafeSingleton = !isL2Chain
+      isL1SafeSingleton = isL1
     }
   }
 
@@ -72,6 +80,23 @@ export const initSafeSDK = async ({
 
   if (isLegacyVersion(safeVersion)) {
     isL1SafeSingleton = true
+  }
+
+  // zkSync Safes using a canonical (EVM bytecode) master copy cannot delegatecall
+  // the zksync-specific (EraVM) MultiSend/MultiSendCallOnly, so force the canonical
+  // aux-contract addresses. Covers versions below the chain-agnostic threshold (<1.4.1).
+  if (isCanonicalDeployment(implementation, chainId, safeVersion)) {
+    const canonicalMultiSendCallOnly = getCanonicalMultiSendCallOnlyAddress(safeVersion)
+    const canonicalMultiSend = getCanonicalMultiSendAddress(safeVersion)
+
+    contractNetworks = {
+      ...contractNetworks,
+      [chainId]: {
+        ...contractNetworks?.[chainId],
+        ...(canonicalMultiSendCallOnly && { multiSendCallOnlyAddress: canonicalMultiSendCallOnly }),
+        ...(canonicalMultiSend && { multiSendAddress: canonicalMultiSend }),
+      },
+    }
   }
 
   const safeSDK = await Safe.init({
--- a/apps/web/src/components/new-safe/create/logic/index.test.ts
+++ b/apps/web/src/components/new-safe/create/logic/index.test.ts
@@ -374,7 +374,13 @@ describe('create/logic', () => {
       spy.mockRestore()
     })
 
-    it('falls back to first network address when canonical not present for chain', () => {
+    it('prefers the canonical deployment address over networkAddresses[0] when canonical is not registered for the chain', () => {
+      // Chain's networkAddresses list omits the canonical address entirely — this
+      // happens on zk-stack chains that register only the EraVM (zksync) flavour.
+      // Returning networkAddresses[0] here would give the WRONG bytecode flavour
+      // for a canonical Safe (EVM master copy delegatecalling EraVM aux reverts),
+      // so getChainAgnosticAddress must return the canonical address from
+      // deployments.canonical instead and log a warning.
       const chain = chainBuilder()
         .with({ chainId: '1' })
         .with({ features: [FEATURES.COUNTERFACTUAL] as any })
@@ -387,12 +393,12 @@ describe('create/logic', () => {
       }
 
       const canonical = faker.finance.ethereumAddress()
-      const firstAddress = faker.finance.ethereumAddress()
+      const otherFlavour = faker.finance.ethereumAddress()
 
       const mockDeployment: SingletonDeploymentV2 = {
         version: '1.4.1',
         contractName: 'CompatibilityFallbackHandler',
-        networkAddresses: { [chain.chainId]: [firstAddress] },
+        networkAddresses: { [chain.chainId]: [otherFlavour] },
         deployments: {
           canonical: { address: canonical },
         },
@@ -402,12 +408,15 @@ describe('create/logic', () => {
       const spy = jest
         .spyOn(safeDeploymentHandlers, 'getCompatibilityFallbackHandlerDeployments')
         .mockReturnValueOnce(mockDeployment)
+      const warnSpy = jest.spyOn(console, 'warn').mockImplementation()
 
       const result = createNewUndeployedSafeWithoutSalt('1.4.1', safeSetup, chain)
 
-      expect(result.safeAccountConfig.fallbackHandler).toEqual(firstAddress)
+      expect(result.safeAccountConfig.fallbackHandler).toEqual(canonical)
+      expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('does not register the canonical address'))
 
       spy.mockRestore()
+      warnSpy.mockRestore()
     })
   })
 })
--- a/apps/web/src/hooks/coreSDK/safeCoreSDK.ts
+++ b/apps/web/src/hooks/coreSDK/safeCoreSDK.ts
@@ -10,7 +10,11 @@ import { isInDeployments } from '@safe-global/utils/hooks/coreSDK/utils'
 import type { SafeCoreSDKProps } from '@safe-global/utils/hooks/coreSDK/types'
 import { keccak256 } from 'ethers'
 import {
+  getCanonicalMultiSendAddress,
+  getCanonicalMultiSendCallOnlyAddress,
+  getDeploymentTypeForMasterCopy,
   getL2MasterCopyVersionByCodeHash,
+  isCanonicalDeployment,
   isChainAgnosticVersion,
   isL2MasterCopyCodeHash,
   resolveChainAgnosticContractAddresses,
@@ -37,13 +41,22 @@ export const initSafeSDK = async ({
   let isL1SafeSingleton = chainId === chains.eth
   let contractNetworks: ContractNetworksConfig | undefined
 
-  // For versions >= 1.4.1, resolve all addresses chain-agnostically (works on any chain)
+  // For versions >= 1.4.1, resolve all addresses chain-agnostically (works on any chain).
+  // Derive deployment type AND L1/L2 flavour from the master copy so that Safes on
+  // zk chains with a canonical master copy get canonical aux contracts (and vice
+  // versa), and Safes on L2 chains running an L1 master copy resolve against the
+  // L1 singleton table. Chain-level flags are only used as defaults when the master
+  // copy can't be matched (e.g. custom / unregistered deployments).
   if (isChainAgnosticVersion(safeVersion) && isL2Chain !== undefined) {
-    const resolved = resolveChainAgnosticContractAddresses(safeVersion, isL2Chain, isZkChain ?? false)
+    const { deploymentType, isL1 } = getDeploymentTypeForMasterCopy(implementation, safeVersion, {
+      deploymentType: isZkChain ? 'zksync' : 'canonical',
+      isL1: !isL2Chain,
+    })
+    const resolved = resolveChainAgnosticContractAddresses(chainId, safeVersion, !isL1, deploymentType)
 
     if (resolved) {
       contractNetworks = { [chainId]: resolved }
-      isL1SafeSingleton = !isL2Chain
+      isL1SafeSingleton = isL1
     }
   }
 
@@ -87,7 +100,12 @@ export const initSafeSDK = async ({
         }
 
         // Merge custom mastercopy with chain-agnostic auxiliary addresses
-        const baseAddresses = resolveChainAgnosticContractAddresses(upgradeableVersion, true, isZkChain ?? false)
+        const baseAddresses = resolveChainAgnosticContractAddresses(
+          chainId,
+          upgradeableVersion,
+          true,
+          isZkChain ? 'zksync' : 'canonical',
+        )
         contractNetworks = {
           [chainId]: {
             ...baseAddresses,
@@ -106,6 +124,23 @@ export const initSafeSDK = async ({
 
   if (isLegacyVersion(safeVersion)) {
     isL1SafeSingleton = true
+  }
+
+  // zkSync Safes using a canonical (EVM bytecode) master copy cannot delegatecall
+  // the zksync-specific (EraVM) MultiSend/MultiSendCallOnly, so force the canonical
+  // aux-contract addresses. Covers versions below the chain-agnostic threshold (<1.4.1).
+  if (isCanonicalDeployment(implementation, chainId, safeVersion)) {
+    const canonicalMultiSendCallOnly = getCanonicalMultiSendCallOnlyAddress(safeVersion)
+    const canonicalMultiSend = getCanonicalMultiSendAddress(safeVersion)
+
+    contractNetworks = {
+      ...contractNetworks,
+      [chainId]: {
+        ...contractNetworks?.[chainId],
+        ...(canonicalMultiSendCallOnly && { multiSendCallOnlyAddress: canonicalMultiSendCallOnly }),
+        ...(canonicalMultiSend && { multiSendAddress: canonicalMultiSend }),
+      },
+    }
   }
 
   if (undeployedSafe) {
--- a/apps/web/src/hooks/coreSDK/safeCoreSDK.ts
+++ b/apps/web/src/hooks/coreSDK/safeCoreSDK.ts
@@ -128,8 +128,10 @@ export const initSafeSDK = async ({
 
   // zkSync Safes using a canonical (EVM bytecode) master copy cannot delegatecall
   // the zksync-specific (EraVM) MultiSend/MultiSendCallOnly, so force the canonical
-  // aux-contract addresses. Covers versions below the chain-agnostic threshold (<1.4.1).
-  if (isCanonicalDeployment(implementation, chainId, safeVersion)) {
+  // aux-contract addresses. Only runs for versions below the chain-agnostic threshold
+  // (<1.4.1); for >=1.4.1 the chain-agnostic resolver already picks the correct flavour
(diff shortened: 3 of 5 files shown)


katspaugh:
## PR Briefing

[Watch the PR walkthrough](https://github.com/5afe/wallet-pr-artifacts/releases/download/prbrief-safe-global-safe-wallet-monorepo-pr-7719-9ac9f626e2e0/video-github.mp4)

▶ **[Watch the walkthrough video](https://github.com/5afe/wallet-pr-artifacts/releases/download/prbrief-safe-global-safe-wallet-monorepo-pr-7719-9ac9f626e2e0/video-github.mp4)** (opens MP4 in a new tab)

This PR fixes broken MultiSend and simulation flows on zkSync Era and Lens chains by changing how aux-contract addresses (MultiSend, MultiSendCallOnly) are resolved. Previously, `initSafeSDK` in both `apps/web` and `apps/mobile` derived the deployment flavour (`canonical` vs `zksync`) and L1/L2 table selection from CGW chain-level flags (`zk`, `l2`), which failed when a Safe's master copy flavour disagreed with those flags. The fix introduces `getDeploymentTypeForMasterCopy` in `packages/utils/src/services/contracts/deployments.ts` to scan both L1 and L2 singleton tables and return `{deploymentType, isL1}` directly from the master copy address. Additionally, `isEraVmChain` generalises EraVM detection beyond the hardcoded zkSync Era chain ID 324, and a canonical MultiSend override is restored for Safe versions below 1.4.1 that bypass the chain-agnostic resolver entirely.

<details>

## Summary

This PR fixes broken MultiSend and simulation flows on zkSync Era and Lens chains by changing how aux-contract addresses (MultiSend, MultiSendCallOnly) are resolved. Previously, `initSafeSDK` in both `apps/web` and `apps/mobile` derived the deployment flavour (`canonical` vs `zksync`) and L1/L2 table selection from CGW chain-level flags (`zk`, `l2`), which failed when a Safe's master copy flavour disagreed with those flags. The fix introduces `getDeploymentTypeForMasterCopy` in `packages/utils/src/services/contracts/deployments.ts` to scan both L1 and L2 singleton tables and return `{deploymentType, isL1}` directly from the master copy address. Additionally, `isEraVmChain` generalises EraVM detection beyond the hardcoded zkSync Era chain ID 324, and a canonical MultiSend override is restored for Safe versions below 1.4.1 that bypass the chain-agnostic resolver entirely.

## Architecture notes

The core logic lives in `packages/utils/src/services/contracts/deployments.ts`, which now exports `getDeploymentTypeForMasterCopy`, `isEraVmChain`, and a revised `resolveChainAgnosticContractAddresses` whose signature changed from `isZk: boolean` to `{chainId, deploymentType}` — all call sites in `apps/web/src/hooks/coreSDK/safeCoreSDK.ts` and `apps/mobile/src/hooks/coreSDK/safeCoreSDK.ts` must be updated accordingly. `getChainAgnosticAddress` now prefers the requested deployment-type address when registered in `networkAddresses[chainId]`, falling back to `networkAddresses[0]`, which changes resolution behaviour for chains like Lens that register both canonical and zksync addresses. A post-resolution canonical override block in both `initSafeSDK` implementations can overwrite addresses already set by `resolveChainAgnosticContractAddresses`, so the interaction between these two blocks for 1.4.1+ Safes with canonical master copies must be verified as intentionally idempotent. Chain-level flags (`isZkChain`, `isL2Chain`) are now only used as fallback defaults inside `getDeploymentTypeForMasterCopy` when the master copy address is unrecognised.

## Ripple risks

Analyzed 7 changed exports. No untouched external callers found — all usages appear to be inside the PR diff.

_Analyzed 4 file(s), 11 changed export(s), 0 uncovered callsite(s) in 2.0s._

## Walkthrough outline

_18 steps across 6 chapters, approximately 4 min 15s._

### Why this broke

1. **motivation** · `packages/utils/src/services/contracts/deployments.ts` L32-35 `DeploymentType`
   The previous PR introduced a chain-agnostic contract resolver that picked MultiSend addresses based on CGW chain-level flags like zk and l2. The problem is that a Safe's master copy can be a canonical build on a zk chain, or a zksync build on a chain that isn't flagged zk at all, so the chain flag and the actual deployment flavour can disagree. This PR fixes that by deriving the flavour directly from the master copy address.

2. **motivation** · `packages/utils/src/services/contracts/deployments.ts` L355-372 `MasterCopyFlavour`
   Two new types anchor the redesign. Deployment type captures whether a Safe was deployed with the canonical, zksync, or eip155 factory. Master copy flavour bundles that with an is-L1 flag so callers know which singleton table to consult. Everything downstream flows from these two pieces of information.

### The new master-copy scanner

3. **change** · `packages/utils/src/services/contracts/deployments.ts` L373-403 `getDeploymentTypeForMasterCopy`
   Here is the heart of the fix: get deployment type for master copy. It receives the implementation address and the Safe version, then scans the L2 singleton table first and the L1 table second, returning both the deployment type and whether the match came from the L1 table. If nothing matches it falls back to chain-level flags, which are now just a safety net for unrecognised custom deployments.

4. **change** · `packages/utils/src/services/contracts/deployments.ts` L373-403 `getDeploymentTypeForMasterCopy`
   Notice the L2 table is checked before the L1 table. If an address somehow appeared in both, L2 would win. That ordering is intentional because most zkSync-family Safes use L2 singletons, but it is worth keeping in mind for any edge case where the same address is registered in both tables.

5. **change** · `packages/utils/src/services/contracts/deployments.ts` L110-145 `getChainAgnosticAddress`
   Get chain agnostic address also changed. Previously it always picked the canonical address from network addresses. Now it first checks whether the requested deployment-type address is registered for the specific chain ID, and only falls back to the zero-index entry if it is not. This is what fixes Lens, which registers both canonical and zksync addresses for the same contract.

6. **change** · `packages/utils/src/services/contracts/deployments.ts` L404-440 `resolveChainAgnosticContractAddresses`
   Resolve chain agnostic contract addresses now accepts chain ID and deployment type as explicit parameters instead of a single boolean is-zk flag. Internally it passes both to get chain agnostic address so the right entry is selected from network addresses. All call sites must be updated to match this new signature.

7. **change** · `packages/utils/src/services/contracts/deployments.ts` L300-325 `getCanonicalMultiSendCallOnlyAddress`
   The canonical MultiSend and MultiSendCallOnly address helpers are restored here. They were deleted in the previous PR on the assumption that the chain-agnostic resolver covered everything, but Safes on version 1.3.0 bypass that resolver entirely, so they still need an explicit override. The helpers strip any plus-suffix from the version string before querying safe-deployments.

### EraVM detection grows up

8. **change** · `packages/utils/src/services/contracts/deployments.ts` L253-274 `isEraVmChain`
   Is EraVM chain replaces the old hard-coded set that only contained zkSync Era mainnet chain ID 324. Instead of a static list, it asks whether the chain's network addresses entry contains the zksync deployment address for the given version. Lens on chain 232 and zkSync Sepolia on chain 300 now qualify automatically without any manual list maintenance.

9. **change** · `packages/utils/src/services/contracts/deployments.ts` L275-299 `isCanonicalDeployment`
   Is canonical deployment is updated in the same spirit: it now queries the global singleton tables without a network filter, so it recognises a canonical master copy regardless of which chain it is running on. This is what lets the post-resolution override block in init safe SDK know when to force canonical MultiSend addresses.

10. **ripple** · `apps/web/src/hooks/coreSDK/safeCoreSDK.ts` L24-80 `initSafeSDK`
   In the web app's init safe SDK, the call to resolve chain agnostic contract addresses now passes chain ID and a deployment type derived from get deployment type for master copy rather than the raw is-zk chain flag. The is-L1 safe singleton field is also now set from the master copy scan result, which matters for chains running an L1 master copy on an L2 network.

11. **ripple** · `apps/web/src/hooks/coreSDK/safeCoreSDK.ts` L80-120 `initSafeSDK`
   After the chain-agnostic block, a new post-resolution block checks is canonical deployment and, when true, overwrites multi-send address and multi-send call-only address with the canonical values. For 1.3.0 Safes this is the primary resolution path; for 1.4.1 Safes with a canonical master copy it should be idempotent because the resolver already picked the same addresses.

12. **ripple** · `apps/mobile/src/hooks/coreSDK/safeCoreSDK.ts` L21-80 `initSafeSDK`
   The mobile app mirrors the same changes. Get deployment type for master copy is called with the implementation address and version, and the result drives both the resolver call and the canonical override block. The same helper functions are imported from the shared utils package, so behaviour should be identical across platforms.

13. **ripple** · `packages/utils/src/services/contracts/__tests__/deployments.test.ts` L1-50 `getDeploymentTypeForMasterCopy`
   The test file grew by nearly 200 lines. It covers get deployment type for master copy across L1 and L2 tables, canonical and zksync flavours, and both 1.3.0 and 1.4.1 versions. All resolve chain agnostic contract addresses call sites in the tests are updated to the new chain-ID-plus-deployment-type signature, and is canonical deployment tests for chains 300 and 232 are flipped to reflect their new EraVM status.

### What can go wrong

14. **risk** · `apps/web/src/hooks/coreSDK/safeCoreSDK.ts` L80-120 `initSafeSDK`
   The biggest risk is the interaction between the chain-agnostic resolver and the canonical override block. For a 1.4.1 Safe with a canonical master copy, both blocks run and both write to the same contract-networks fields. If is canonical deployment ever returns true when it should not, it silently overwrites a correctly resolved zksync address with a canonical one. Reviewers should confirm the two blocks are truly idempotent for every supported combination.

15. **risk** · `packages/utils/src/services/contracts/deployments.ts` L110-145 `getChainAgnosticAddress`
   Get chain agnostic address now falls back to the zero-index entry in network addresses when the requested deployment type is not registered for the chain. If a chain lists addresses in an unexpected order, the fallback could return the wrong one. This is subtle because the function does not signal which path it took, so callers have no way to detect a silent fallback.

16. **risk** · `packages/utils/src/services/contracts/deployments.ts` L373-403 `getDeploymentTypeForMasterCopy`
   When the master copy address is not found in either singleton table, get deployment type for master copy falls back to chain-level flags and returns canonical as the deployment type with is-L1 set to false. For a custom or unrecognised deployment on an L2 chain this default is reasonable, but it means the old chain-flag bugs can still surface for non-standard Safes. The fallback path is tested, but reviewers should confirm the default is acceptable for production custom deployments.

17. **risk** · `packages/utils/src/services/contracts/__tests__/deployments.test.ts` L1-50 `isEraVmChain`
   The tests assert specific addresses for Lens chain 232 and zkSync Sepolia chain 300 that are pulled from the safe-deployments package at test time. If the version of safe-deployments pinned in CI does not register the zksync deployment address for those chains, is EraVM chain will return false and the canonical override will not fire, silently regressing to the old behaviour. Worth double-checking the package version in the lock file.

### Bottom line

18. **wrap** · `packages/utils/src/services/contracts/deployments.ts` L373-403 `getDeploymentTypeForMasterCopy`
   The bottom line: by scanning the singleton tables directly instead of trusting chain-level flags, this PR makes MultiSend and simulation work correctly on zkSync Era and Lens regardless of whether the chain is flagged zk, and it does so in a way that generalises to any future chain that registers a zksync deployment address.

## Reviewer checklist

- [ ] Verify every call site of `resolveChainAgnosticContractAddresses` has been updated to the new `{chainId, deploymentType}` signature and no old `isZk` boolean usage remains in either `apps/web` or `apps/mobile`.
- [ ] Confirm that for 1.4.1+ Safes with a canonical master copy, the post-resolution canonical override block in `initSafeSDK` is truly idempotent (i.e., it writes the same addresses that `resolveChainAgnosticContractAddresses` already set, not different ones).
- [ ] Check that the second call to `resolveChainAgnosticContractAddresses` inside the upgradeable-version block in `apps/web/src/hooks/coreSDK/safeCoreSDK.ts` also uses the master-copy-derived `deploymentType` rather than the old chain-flag-based value.
- [ ] Cross-check the hardcoded canonical and zksync addresses for Lens (chainId 232) and zkSync Sepolia (chainId 300) asserted in `deployments.test.ts` against the actual `networkAddresses` entries in the `safe-deployments` package version pinned in CI.
- [ ] Confirm `getDeploymentTypeForMasterCopy` correctly handles the case where a master copy address appears in both L1 and L2 singleton tables, and that the L2-first priority is intentional and documented.
- [ ] Manually test or trace the five on-chain Safe examples listed in the PR description to confirm each resolves to the expected MultiSendCallOnly address after the fix.

## File-by-file

### `apps/mobile/src/hooks/coreSDK/safeCoreSDK.ts`

In `initSafeSDK`, the call to `resolveChainAgnosticContractAddresses` is updated to derive `deploymentType` and `isL1` from the master copy address via `getDeploymentTypeForMasterCopy` instead of chain-level flags, and a new block after the existing L1 singleton override forces canonical `multiSendAddress`/`multiSendCallOnlyAddress` when `isCanonicalDeployment` returns true.

_Fixes aux-contract resolution for zkSync/Lens Safes whose master copy flavour (canonical vs zksync, L1 vs L2) doesn't match the chain-level `isZkChain`/`isL2Chain` flags, preventing wrong MultiSend addresses being used in multisend and simulation flows._

**Risks:**
- The canonical override block runs unconditionally after the chain-agnostic block, so for versions >=1.4.1 it may overwrite addresses already correctly set by `resolveChainAgnosticContractAddresses` — verify `isCanonicalDeployment` is false for those cases or that the override is intentionally idempotent.
- Fallback defaults passed to `getDeploymentTypeForMasterCopy` still use chain-level `isZkChain`/`isL2Chain`; confirm these defaults are correct when the master copy is unrecognised.
- Mobile-specific: ensure the same `getCanonicalMultiSendAddress`/`getCanonicalMultiSendCallOnlyAddress` helpers are available and return the same values as in the web app (no platform-specific bundle differences).
- `contractNetworks?.[chainId]` spread could silently drop fields if `contractNetworks` is undefined at that point — check that the spread handles undefined correctly across all code paths.

### `apps/web/src/hooks/coreSDK/safeCoreSDK.ts`

In `initSafeSDK`, the call to `resolveChainAgnosticContractAddresses` now passes `chainId` and a `deploymentType` derived from `getDeploymentTypeForMasterCopy(implementation, safeVersion, ...)` instead of raw `isL2Chain`/`isZkChain` flags. A new post-resolution block calls `isCanonicalDeployment` and, when true, overwrites `multiSendAddress`/`multiSendCallOnlyAddress` in `contractNetworks` with canonical addresses from `getCanonicalMultiSendAddress`/`getCanonicalMultiSendCallOnlyAddress`.

_Aux-contract addresses (MultiSend, MultiSendCallOnly) are now derived from the master copy's actual deployment flavour rather than chain-level flags, fixing broken multisend/simulation on zkSync and Lens chains where the chain flag and master copy flavour disagree._

**Risks:**
- The canonical override block runs unconditionally after the chain-agnostic block, so it can silently overwrite addresses already set by `resolveChainAgnosticContractAddresses` — verify the two blocks don't conflict for 1.4.1+ Safes where `isCanonicalDeployment` is also true
- `getDeploymentTypeForMasterCopy` fallback defaults (`isZkChain ? 'zksync' : 'canonical'` and `isL1: !isL2Chain`) are still chain-flag-based; confirm these defaults are safe for unrecognised/custom master copies
- The second call to `resolveChainAgnosticContractAddresses` (inside the upgradeable-version block) still uses `isZkChain ? 'zksync' : 'canonical'` without consulting the master copy — check if this path also needs the master-copy-derived type
- `isL1SafeSingleton` is now set to `isL1` from `getDeploymentTypeForMasterCopy`; confirm this value is correct for L2 chains running an L1 master copy, as it affects singleton selection in the SDK
- Verify `getCanonicalMultiSendAddress`/`getCanonicalMultiSendCallOnlyAddress` return `undefined` gracefully for versions where no canonical address exists, and that the spread conditional handles that correctly

### `packages/utils/src/services/contracts/__tests__/deployments.test.ts`

The test file adds 194 new lines covering three new exported functions (`getDeploymentTypeForMasterCopy`, `isEraVmChain`) and updates all `resolveChainAgnosticContractAddresses` call sites to pass an explicit `chainId` and `deploymentType` string instead of the old boolean `isZk` flag. Several `isCanonicalDeployment` tests are also revised to reflect that chains 300 (zkSync Sepolia) and 232 (Lens) are now treated as EraVM chains.

_The tests document and enforce the new contract-resolution logic that derives deployment flavour from the master copy address rather than chain-level flags, which is the core correctness guarantee of the fix._

**Risks:**
- Hardcoded addresses (L2_CANONICAL_141, L2_ZKSYNC_141, etc.) must exactly match what safe-deployments publishes; any mismatch silently passes if the implementation also uses the same wrong address
- UNREGISTERED_CHAIN_ID '99999999' is used as a fallback-path fixture — verify the implementation truly has no networkAddresses entry for it, otherwise tests exercise the wrong code path
- isCanonicalDeployment for chain '300' flipped from false to true; confirm the production safe-deployments package version used in CI actually registers the zksync deployment address for chainId 300
- Lens chainId 232 canonical address '0x29fcB43b46531BcA003ddC8FCB67FFE91900C762' and zksync address '0x610fcA2e0279Fa1F8C00c8c2F71dF522AD469380' are asserted without a source reference — should be cross-checked against safe-deployments networkAddresses['232']
- getDeploymentTypeForMasterCopy with unknown address returns {deploymentType:'canonical', isL1:false} as default — test covers this but reviewer should confirm the default is safe for unrecognised custom deployments on L2 chains
- No test covers the interaction between getDeploymentTypeForMasterCopy and the EraVM canonical-override path (the restored MultiSend override for <1.4.1); integration coverage may be missing

### `packages/utils/src/services/contracts/deployments.ts`

Refactors `deployments.ts` to derive deployment type (`canonical`/`zksync`/`eip155`) and L1/L2 flavour from the master copy address rather than chain-level flags. Adds `getDeploymentTypeForMasterCopy`, `isEraVmChain`, and `MasterCopyFlavour`; rewrites `getChainAgnosticAddress` to prefer the requested deployment-type address when registered in `networkAddresses[chainId]`; updates `resolveChainAgnosticContractAddresses` to accept `chainId` and `deploymentType` instead of `isZk`; and strips `+L2` metadata suffixes before passing versions to safe-deployments getters.

_Fixes aux-contract resolution (MultiSend, MultiSendCallOnly) on zkSync-family chains (Lens, zkSync Sepolia) where CGW chain flags (`zk`, `l2`) were insufficient to determine the correct deployment table, causing multisend and simulations to break._

**Risks:**
- All callers of `resolveChainAgnosticContractAddresses` must now pass `chainId` and `deploymentType` (not `isZk`) — verify every call site is updated and no old signature remains.
- `getDeploymentTypeForMasterCopy` iterates L2 table before L1; if an address appears in both tables (unlikely but possible), L2 wins — confirm this priority is intentional.
- `isEraVmChain` uses `getSafeL2SingletonDeployments`/`getSafeSingletonDeployments` (plural, no `network` filter) and checks `networkAddresses[chainId]` — confirm these V2 getters always populate `networkAddresses` for all registered chains.
- `getChainAgnosticAddress` now returns `networkAddresses[0]` as fallback when the deployment-type address isn't registered for the chain; if a chain lists multiple addresses in a different order than expected, the wrong address could be returned.
- Version suffix stripping (`version.split('+')[0]`) is now applied in multiple places — ensure consistency and that no caller passes a version that legitimately contains `+` for a different reason.
- `isCanonicalDeployment` now uses `getSafeL2SingletonDeployments`/`getSafeSingletonDeployments` without `network` filter, so it matches the canonical address globally rather than per-chain — verify this doesn't produce false positives on chains where the canonical address coincidentally matches something else.
- Exported `DeploymentType` and `MasterCopyFlavour` types are now part of the public API — check downstream consumers for compatibility.

---
_Generated by prbrief • model: claude-sonnet-4-6 • 6 LLM calls • $0.21 • 97.3s_

</details>

<sub>Generated locally by <a href="https://github.com/anthropics/claude-code">prbrief</a>. · `safe-global/safe-wallet-monorepo@9ac9f626e2e0`</sub>

katspaugh (commented):
Strings `zksync` and `canonical` are repeated in many places as literals. It would be good to make them constants or an enum.

compojoom:
We are exporting a union type `export type DeploymentType = 'canonical' | 'eip155' | 'zksync'`
I don't see a benefit of having yet another constant. TS is able to protect us from wrong types.

vmosharova (approved):
👏🏻 Approved!

PR #7721, opened 2026-04-22
fix(mobile): surface error when wallet can't honor the active Safe's chain

tmjssz:
> Wallet connects but can't sign,
> promise hangs, user stares blind —
> now switch, then alert.

## What it solves

When a user connects a wallet via WalletConnect that doesn't support the active Safe's chain, the `connect()` promise hung forever — no error, no feedback, no escape. Reown's AppKit fires neither `CONNECT_ERROR` nor `USER_REJECTED` in that scenario.

Resolves: [WA-1996](https://linear.app/safe-global/issue/WA-1996/no-error-if-there-is-an-issue-with-connection)

## How this PR fixes it

The `CONNECT_SUCCESS` handler in `useConnect` now compares Reown's `caipNetworkId` against the active Safe's chain. On mismatch it attempts `switchNetwork`; if the switch throws, it disconnects and rejects with a new `UnsupportedChainError`. An 8-second timeout covers the less-obvious Reown behavior where `switchNetwork` resolves without actually switching (silent no-op when internal state is stale).

A chain gate in the resolver `useEffect` prevents the complementary race: without it, state updates from `useAccount()` could satisfy the pending promise on the wrong chain before the switch attempt settled.

Consumers (`useImportSignerFlow`, `useReconnectFlow`) catch `UnsupportedChainError` and show a shared "Unsupported network" alert. `chainId` comparisons normalize via `parseInt` so zero-padded or hex-encoded chainIds upstream don't silently fail the match.

## How to test it

1. Open the mobile app and select a Safe on a chain your test wallet doesn't support (e.g., a Safe on Base if your wallet only has Ethereum mainnet enabled).
2. Tap to import a signer or reconnect → scan the WalletConnect QR code → approve in your wallet.
3. Expect an "Unsupported network" alert; tapping OK should leave the import/reconnect flow in a recoverable state.
4. Repeat with a wallet that **does** support the Safe's chain but is currently on a different one — expect the wallet to prompt a network switch; approving should proceed normally.

## Screenshots

N/A — the new UI is a standard `Alert.alert` dialog.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/WalletConnect/hooks/__tests__/useConnect.test.ts
+++ b/apps/mobile/src/features/WalletConnect/hooks/__tests__/useConnect.test.ts
@@ -1,41 +1,53 @@
 import { renderHook, act } from '@/src/tests/test-utils'
-import { useConnect, ConnectError, UserRejectedError } from '../useConnect'
+import { useConnect, ConnectError, UnsupportedChainError, UserRejectedError } from '../useConnect'
 
 const mockOpen = jest.fn()
+const mockDisconnect = jest.fn().mockResolvedValue(undefined)
+const mockSwitchNetwork = jest.fn().mockResolvedValue(undefined)
 
 const mockWalletState = {
   address: undefined as string | undefined,
   isConnected: false,
   walletInfo: undefined as { name: string; icon: string } | undefined,
+  chain: undefined as { caipNetworkId: string } | undefined,
 }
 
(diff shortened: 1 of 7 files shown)


tmjssz:
@claude review

compojoom (approved):

compojoom:
@claude review

PR #7559, opened 2026-03-30
feat(mobile): add signer ownership validation, error screen, and design updates

tmjssz:
> Validate before you name,
> flat list, no sections remain,
> red glow warns the wrong.

## What it solves

Adds ownership validation to the WalletConnect signer import flow so non-owners are blocked before reaching the naming screen. Also updates ImportSuccess, ConnectSignerError, and signers list screens to match Figma designs.

Resolves:
- https://linear.app/safe-global/issue/WA-1839/name-your-signer-screen
- https://linear.app/safe-global/issue/WA-1673/signers-list-screen
- https://linear.app/safe-global/issue/WA-1668/wallet-app-connected-success-screen
- https://linear.app/safe-global/issue/WA-1840/signer-not-an-owner-error-screen

## How this PR fixes it

The key architectural decision is moving ownership validation from the NameSigner screen (on Continue tap) to ImportSigners.container (immediately after WalletConnect connection). A `connectInitiatedRef` guards against pre-existing WalletConnect sessions triggering navigation on mount. Non-owners are routed to the new ConnectSignerError screen instead of the naming screen.

The signers list was converted from SectionList to FlatList, removing "My signers" / "Not imported signers" section headers. The `signersGroup` prop is still passed through to SignersListItem so it can determine whether a signer is imported (for context menu actions).

SectionTitle description color was changed globally to `$colorSecondary` — this affects all SectionTitle consumers app-wide.

## How to test it

1. Open the app → navigate to Signers → tap "Add signer" → "Connect wallet app"
2. Connect a wallet that IS an owner → should navigate to the "Name your signer" screen
3. Connect a wallet that is NOT an owner → should show the red error screen "Can't sign with this wallet" with a close X button
4. After importing a signer, verify the success screen shows green gradient, signer card, and "Done" button
5. On the signers list, verify items are shown as a flat list (no section headers), with title "Your signers" and "Add signer" button

## Visual summary

```mermaid
flowchart LR
    A[Connect Wallet] --> B{Ownership Check}
    B -->|Owner| C[Name Signer Screen]
    B -->|Not Owner| D[Error Screen]
    C --> E[Import Success]
    D --> F[Dismiss]
```

## Screenshots

simulator_screenshot_A3174696-DC63-48C4-9744-14B282B37577
simulator_screenshot_F3CFC60A-0586-46F8-926E-EF0C60C5491A
simulator_screenshot_67A4134B-1F46-49E8-8E6D-3948B35C268A
[...]

PR #7729, opened 2026-04-23
fix(mobile): validate owner against fetched Safe overviews during add-Safe flow

tmjssz:
> Add-Safe owner imports
> validated against the list already shown —
> WalletConnect works again.

## What it solves

Resolves: [WA-2041](https://linear.app/safe-global/issue/WA-2041)

In the mobile "Add new Safe" onboarding flow, any attempt to import an owner via WalletConnect, Ledger, or seed phrase always resolved to `isOwner: false` and routed the user to the `connect-signer-error` screen — even when the wallet was genuinely an owner of the Safe being added.

## How this PR fixes it

Introduced in #7559, the `pendingSafe` branch of `useAddressOwnershipValidation` queried the cross-chain owner-index endpoint `/v2/owners/{address}/safes` and checked whether `pendingSafe.address` appeared in the returned map. That check never matched in practice.

The fix replaces that branch with a local membership check against the Safe overviews already fetched by the owners-list screen (`AddSignersForm.container.tsx`) via `useSafesGetOverviewForManyQuery` + `extractSignersFromSafes`. Same query args as the container — `{ safes, currency: 'USD', trusted: true, excludeSpam: true }` — so RTK Query serves the response from cache. The `activeSafe` branch is untouched.

Two small supporting pieces to keep this robust:

- **`useLazySafeOverviews`** — a lazy wrapper over `useLazySafesGetOverviewForManyQuery` that normalizes `currency.toUpperCase()` in one place. The consumers (this hook, `useImportSafe`) can't silently drift on casing.
- **`overviewQueryArgs.ts`** — a shared `OverviewQueryArgs` type + `normalizeOverviewArgs(args)` used by both the eager `useSafeOverviewsQuery` and the new lazy wrapper, so their cache keys stay identical.

## How to test it

1. Pick a Safe on Sepolia (or any supported chain) that has a MetaMask-owned address as an owner.
2. "Add Safe" → QR-scan the Safe address → enter a name → proceed to the owners list.
3. Tap "Import owner" → WalletConnect → connect the owning MetaMask wallet.
   **Expected**: navigation to `/import-signers/name-signer` (NOT the error screen).
4. Repeat with a wallet that is NOT an owner → **expected**: `connect-signer-error` screen.
5. Optionally repeat with Ledger and seed-phrase imports.
6. Post-onboarding smoke: import an owner for an already-activated Safe (the `activeSafe` branch) — unchanged behavior.

## Affected flows

- Add Safe → owners list → Import owner (WalletConnect / Ledger / seed phrase)

## Blast radius

- **`useAddressOwnershipValidation`** (mobile) — same public shape, semantics changed only for the `pendingSafe` branch. Consumers: `useImportSignerFlow` (WC), `useImportLedgerAddress`, `useImportSeedPhraseAddress`, `LoadingImport.container`. Also includes small internal cleanup: dropped a dead `if (!pendingSafe && !activeSafe)` early-return, dropped a dead `!overviews` guard, extracted a local `findOwner(safes, address)` helper to remove duplication between the two branches.
- **`useImportSafe`** (mobile, `SafeAccountInput`) — switched to the new `useLazySafeOverviews` wrapper. Side effects: (a) currency sent to `/v2/safes` is now uppercase, matching what the owners-list container already sent (one test assertion updated); (b) `excludeSpam: true` is now included in the query args so cache keys line up; (c) `currency` added to the debounced callback's `useCallback` dep array — pre-existing stale-currency bug surfaced during review.
- **New file** `apps/mobile/src/hooks/services/useLazySafeOverviews.ts` — tiny wrapper; keeps a stable trigger reference via `useCallback`, infers extra trigger args from the underlying RTK Query hook signature.
- **New file** `apps/mobile/src/hooks/services/overviewQueryArgs.ts` — shared `OverviewQueryArgs` type + `normalizeOverviewArgs` function used by both overview wrappers.
- **New file** `apps/mobile/src/hooks/services/__tests__/useLazySafeOverviews.test.ts` — pins the two load-bearing invariants (currency uppercased; trigger ref stable across re-renders — the latter guards against an OOM regression seen during development).
- No slice / state-shape / schema changes. No feature-flag changes. Web app unaffected.

## Risks / not checked

- Did not verify the behavior when the CGW `/v2/safes` response is degraded or the Safe is unindexed on every chain; hook falls through to `isOwner: false`, same as when legitimately not an owner.
- **Accepted multi-chain tradeoff**: ownership is determined by union across all chains where the Safe is deployed (matches what the owners-list screen shows). If a Safe has diverged owner sets per chain, a wallet that is only an owner on one of them can be imported as a signer, even though `AddSignersForm.container.tsx` may then activate a different chain. The contract still enforces signatures, so this is a UX-grade inconsistency, not a security issue. Alternative semantics (intersection / first-chain-only) were considered and rejected as strictly worse for the common case.

## Visual summary

Before: `pendingSafe` branch → `GET /v2/owners/{address}/safes` → match `pendingSafe.address` in returned chainId-keyed map (always false).

After: `pendingSafe` branch → `GET /v2/safes?safes=<chainId:addr>…` (same args as owners-list container, cache-hit) → `extractSignersFromSafes` → case-insensitive match against connected address.

```mermaid
flowchart LR
  subgraph Before
    A1[validateAddressOwnership<br/>pendingSafe branch] --> B1["/v2/owners/{addr}/safes"]
    B1 --> C1[Match pendingSafe.address<br/>in chainId→addresses map]
    C1 -->|always false in practice| D1[isOwner: false]
  end
  subgraph After
    A2[validateAddressOwnership<br/>pendingSafe branch] --> W[useLazySafeOverviews<br/>same args as container]
    W --> B2["/v2/safes (cache hit)"]
    B2 --> E[extractSignersFromSafes]
    E --> F{sameAddress match}
    F -->|yes| G[isOwner: true]
    F -->|no| H[isOwner: false]
  end
```

## Checklist

- [x] I've documented how it affects the analytics (if at all) 📊 _(no analytics impact)_

---

## CLA signature

---

[Safe Engineering](https://github.com/5afe/safe-engineering-plugin) Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/mobile/src/components/SafeAccountInput/hooks/useImportSafe.test.tsx
+++ b/apps/mobile/src/components/SafeAccountInput/hooks/useImportSafe.test.tsx
@@ -240,7 +240,7 @@ describe('useImportSafe', () => {
       })
 
       await waitFor(() => {
-        expect(requestedCurrency).toBe('USD')
+        expect(requestedCurrency).toBe('usd')
       })
     })
   })
--- a/apps/mobile/src/components/SafeAccountInput/hooks/useImportSafe.test.tsx
+++ b/apps/mobile/src/components/SafeAccountInput/hooks/useImportSafe.test.tsx
@@ -240,7 +240,7 @@ describe('useImportSafe', () => {
       })
 
       await waitFor(() => {
-        expect(requestedCurrency).toBe('usd')
+        expect(requestedCurrency).toBe('USD')
       })
     })
   })
--- a/apps/mobile/src/components/SafeAccountInput/hooks/useImportSafe.ts
+++ b/apps/mobile/src/components/SafeAccountInput/hooks/useImportSafe.ts
@@ -41,6 +41,7 @@ export const useImportSafe = () => {
             safes: chainIds.map((chainId: string) => makeSafeId(chainId, address)),
             currency,
             trusted: true,
+            excludeSpam: true,
           },
           false,
         )
--- a/apps/mobile/src/components/SafeAccountInput/hooks/useImportSafe.ts
+++ b/apps/mobile/src/components/SafeAccountInput/hooks/useImportSafe.ts
@@ -5,7 +5,7 @@ import { isValidAddress } from '@safe-global/utils/utils/validation'
 import { parsePrefixedAddress } from '@safe-global/utils/utils/addresses'
 import { useAppSelector } from '@/src/store/hooks'
 import { selectAllChainsIds } from '@/src/store/chains'
-import { useLazySafesGetOverviewForManyQuery } from '@safe-global/store/gateway/safes'
+import { useLazySafeOverviews } from '@/src/hooks/services/useLazySafeOverviews'
 import { FormValues } from '@/src/features/ImportReadOnly/types'
 import debounce from 'lodash/debounce'
 import { selectCurrency } from '@/src/store/settingsSlice'
@@ -26,7 +26,7 @@ export const useImportSafe = () => {
     trigger: triggerInput,
     formState: { isValid },
   } = useFormContext<FormValues>()
-  const [trigger, result] = useLazySafesGetOverviewForManyQuery()
+  const [trigger, result] = useLazySafeOverviews()
 
   const inputAddress = watch('safeAddress')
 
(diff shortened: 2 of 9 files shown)


tmjssz:
@claude review

compojoom (approved):

PR #7733, opened 2026-04-23
fix(web): Adjust accounts page alignement

clovisdasilvaneto:
> Cards align on a grid,
> content breathes in equal cells —
> add accounts, bold green.

Closes [WA-2006](https://linear.app/safe-global/issue/WA-2006/accounts-page-inside-space).

## What it solves

The Accounts page (inside Space) had two UX issues flagged in WA-2006:

1. **Uneven content distribution inside each account card.** Threshold, chain badge, and pending-tx cells were floating as loose flex children with ad-hoc margins (`mr-8`, `ml-auto`, `pl-1/2`), so they clumped toward the right edge instead of splitting the row evenly.
2. **"Add accounts" CTA wasn't primary.** It rendered with the outline variant, so it didn't read as the main action on the page.

## How this PR fixes it

**Equal content distribution (`SafeCardReadOnly.tsx`)**

- Replaced the flex layout with a 5-column grid `minmax(0,20vw) 1fr 1fr 1fr auto` at ≥786px — identifier block, threshold, chain, pending-tx, and trailing (balance + menu) each get their own track and justify to center. Below 786px the card keeps the simple flex row.
- Moved the threshold and pending-tx cells into dedicated grid cells with fixed widths (`w-16`, `w-24`) and centered content, instead of trailing the chain badge with arbitrary margins.
- Dropped the dead `sm:grid-cols-[minmax(0,20vw)_1fr_auto]` rule that was immediately overridden.

**Primary "Add accounts" CTA (`AddAccounts`, `SafeAccounts`, `button.tsx`)**

- `SpaceSafeAccounts` now passes `buttonVariant="default"` to `AddAccounts` so the CTA uses the primary variant.
- The default button variant in `button.tsx` now tints inline SVGs green in light mode and inherits `primary-foreground` in dark mode, matching the design token spec.
- `AddAccounts` wraps the button in a `shadcn-scope` so the dark-mode class propagates correctly to the variant colors.

## How to test it

- [ ] Open the Accounts page inside a Space with several Safe cards visible.
- [ ] At ≥786px: verify each card splits content into equal-width cells — identifier / threshold / chain / pending-tx / balance + menu — with no rightward clumping.
- [ ] Below 786px: card stays as a flex row with only identifier, chain, balance, and menu; threshold and pending-tx cells hide.
- [ ] Verify the "Add accounts" button reads as primary (filled background, green plus icon in light mode) in both light and dark themes.
- [ ] Disabled state (non-admin): button still renders with the primary variant and the correct tooltip.

## Visual summary

```mermaid
flowchart LR
    subgraph Before
      A1[identifier] --> B1["chain + pending + balance + menu (clumped right via ml-auto / mr-8)"]
    end
    subgraph After
      A2["identifier (minmax 20vw)"] --> T["threshold (1fr)"]
      T --> C["chain (1fr)"]
      C --> P["pending-tx (1fr)"]
      P --> E["balance + menu (auto)"]
    end
```

## Checklist

- [x] I've documented how it affects the analytics (if at all) 📊 — no analytics impact, UI-only change
- [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻 — layout-only change; existing suite covers rendering

---

## CLA signature


apps/web/src/features/spaces/components/SafeAccounts/SafeCardReadOnly.tsx
```diff
@@ -91,15 +91,18 @@ const SafeCardReadOnly = ({
         ref={elementRef as React.Ref<HTMLDivElement>}
         onClick={isClickable ? onClick || handleCardClick : undefined}
         className={cn(
-          'box-border flex w-full min-w-0 max-w-full items-center gap-1.5 rounded-3xl border-2 border-card bg-card py-4 pl-3 pr-3 transition-colors sm:gap-2 sm:pl-6 sm:pr-6',
+          'box-border flex w-full min-w-0 max-w-full items-center gap-4 rounded-3xl border-2 border-card bg-card py-4 pl-3 pr-3 transition-colors sm:gap-6 sm:pl-6 sm:pr-6',
+          // Grid layout activates at ≥786px; below that the card stays as a simple flex row.
+          'min-[786px]:grid min-[786px]:items-center min-[786px]:justify-items-center',
+          'min-[786px]:grid-cols-[minmax(0,20vw)_1fr_1fr_1fr_auto]',
```
vmosharova: Let's put these css classes in a separate module to keep the component more readable

vmosharova (changes requested):
A very small comment about CSS structure - otherwise good!

vmosharova (approved):
👍🏻

PR #7734, opened 2026-04-23
feat: optimize chain selector

ChillYao:
## What it solves
https://linear.app/safe-global/issue/WA-2049/fix-chain-selector-not-showing-error-message
Resolves: the "All networks" section inside the Space safe bar's chain selector currently lists **every** chain the Safe is not yet deployed on. It does not check whether the Safe can actually be added to those chains — so users can click a chain that silently fails later because the Safe's master copy is outdated, the creation data can't be fetched, or the target chain is incompatible (e.g. zkSync / Lens). The click leads to a broken add-network flow instead of a clear message up front.

## How this PR fixes it

- Extract a shared `useAddNetworkState(safeAddress, deployedChainIds)` hook under `@/features/multichain`. It composes `useSafeCreationData` + `useCompatibleNetworks` + the `hasMultiChainAddNetworkFeature` gate, and returns a single state object: `loading`, `availableNetworks`, `unavailableReason` (`'safe-specific' | 'outdated-mastercopy' | null`), `error`, `isFeatureEnabled`.
- Introduce `AllNetworksSection`, which owns all the visual states previously inlined in `ChainSelectorBlock`:
  - feature disabled on current chain → render nothing
  - `safe-specific` → muted info row "Adding another network is not possible for this Safe." Info icon becomes a tooltip trigger when a raw error is available
  - `outdated-mastercopy` → muted info row explaining the mastercopy is outdated
  - loading → centered spinner
  - available networks → the existing "All networks" accordion, but each chain row respects `available`; incompatible chains render as a disabled button with "Not available" and cannot fire `onAddNetwork`
- `ChainSelectorBlock` no longer receives `availableChains`. It now receives `safeAddress` + `deployedChainIds` and delegates the whole section to `AllNetworksSection`.
- `useSpaceChainSelector` drops its naive `availableChains` derivation (the one that just diffed `chainConfigs` against `deployedChainIds`), which was the source of the over-permissive list.
- Analytics preserved: `SHOW_ALL_NETWORKS` fires only on expand (not collapse); `ADD_NEW_NETWORK` fires with the `top_bar` label on a successful click.

## How to test it

1. `yarn workspace @safe-global/web dev`, open a Space with a Safe that is **not** multichain-eligible (e.g. a Safe created from an old master copy, or one on a chain where creation data can't be resolved — zkSync / Lens work).
2. Open the Space safe bar chain selector.
   - Expected: the "All networks" accordion is replaced by a muted info row "Adding another network is not possible for this Safe." / "…outdated mastercopy…".
   - Hover the info icon — if the underlying error has a message, a tooltip shows it.
3. Open the selector for a healthy Safe on a supported chain.
   - Expected: "All networks" accordion appears with only the chains that are compatible + feature-enabled + not already deployed. Clicking a chain opens the add-network dialog.
4. If any chain in the list is technically compatible but not deployable (`available: false`), its row is disabled and shows "Not available".
5. While creation data is still resolving, the accordion area shows a spinner instead of a stale list.
6. Storybook: `yarn workspace @safe-global/web storybook` → `Features/Spaces/SpaceChainSelector` for the visual states.

Run:

```bash
yarn workspace @safe-global/web test src/features/multichain/hooks/__tests__/useAddNetworkState.test.ts
yarn workspace @safe-global/web test src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.test.tsx
yarn workspace @safe-global/web test src/components/common/SpaceSafeBar
```

## Affected flows

- Space safe bar → chain selector dropdown → "All networks" accordion (primary).
- Space safe bar → "Add network" entry point into `CreateSafeOnNewChain` (the dialog itself is unchanged; only the set of chain IDs that can reach it is tightened).

## Blast radius

- **New public exports from `@/features/multichain`**: `useAddNetworkState` and the `AddNetworkState` / `AddNetworkUnavailableReason` / `AvailableNetwork` types. Only consumed by `AllNetworksSection`; no existing consumer changes.
- **`ChainSelectorBlock` prop contract changed**: `availableChains` removed, `safeAddress` + `deployedChainIds` added. Only caller is `SpaceChainSelector` in this repo (verified via search).
- **`useSpaceChainSelector` return shape changed**: `availableChains` removed. Only consumed by `SpaceChainSelector`.
- **`useSafeCreationData`**: now fires for every open of the Space chain selector (previously only when the user opened `CreateSafeOnNewChain`). This hook hits RPCs across deployed chains; expect one extra round of reads per Safe whenever the selector renders. It runs lazily inside a Portal-rendered `DropdownMenuContent`, so it only executes when the dropdown is opened.
- **Analytics events**: `OVERVIEW_EVENTS.SHOW_ALL_NETWORKS` and `OVERVIEW_EVENTS.ADD_NEW_NETWORK` (unchanged names, same trigger points).
- **No mobile impact**: changes are scoped to `apps/web` (Space safe bar + multichain feature hook); no touched files live in `packages/**`.

## Risks / not checked

- Did not verify behaviour against a real outdated-mastercopy Safe on mainnet — relied on mocking `useSafeCreationData` / `useCompatibleNetworks` for tests.
- Did not test the RPC cost of `useSafeCreationData` running on every selector open for Safes deployed across many chains; if this becomes a problem we should memoise by `(safeAddress, deployedChainIds)` or hoist the hook so it only fires on first open.
- Did not exercise the non-web surface (mobile); shared packages were not touched so no regression expected there.
- Did not test the Chromatic visual diff locally — relying on CI Chromatic to catch theming regressions in `AllNetworksSection`.
- Did not verify behaviour when `hasMultiChainAddNetworkFeature` is feature-flagged off on the origin chain in production config (covered by unit tests only).

## Visual summary
Screenshot 2026-04-24 at 10 58 23
Screenshot 2026-04-24 at 10 58 44
Screenshot 2026-04-24 at 13 16 42
Screenshot 2026-04-24 at 13 16 59
Screenshot 2026-04-24 at 13 24 49
Screenshot 2026-04-24 at 13 25 01

```mermaid
flowchart TB
  subgraph Before
    A1[ChainSelectorBlock] --> B1["availableChains = chainConfigs − deployedChainIds<br/>(naive, inline in useSpaceChainSelector)"]
    B1 --> C1["All networks accordion<br/>lists every non-deployed chain"]
    C1 --> D1[Click → add-network dialog → silently fails<br/>for incompatible Safes]
  end
  subgraph After
    A2[ChainSelectorBlock] --> S[AllNetworksSection]
    S --> H[useAddNetworkState]
    H --> H1[useSafeCreationData]
    H --> H2[useCompatibleNetworks]
    H --> H3[hasMultiChainAddNetworkFeature]
    S --> R{unavailableReason?}
    R -- "safe-specific" --> M1[Muted info row<br/>+ tooltip on raw error]
    R -- "outdated-mastercopy" --> M2[Muted info row]
    R -- "null & loading" --> M3[Spinner]
    R -- "null & available" --> M4["Accordion<br/>disabled row for available=false"]
  end
```

## Checklist

---

## CLA signature


ChillYao:
@claude review

ChillYao:
@claude review

ChillYao:
@claude where are you?

vmosharova:
@claude review

ChillYao:
@claude review

katspaugh:
## PR Briefing

[Open the PR briefing](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-wallet-monorepo%2Fpr-7734%2Fc51ca4d62224%2Fbundle.json)

▶ **[Open the interactive review](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-wallet-monorepo%2Fpr-7734%2Fc51ca4d62224%2Fbundle.json)**
▶ **[Watch the walkthrough video](https://github.com/katspaugh/popcorn-review-artifacts/releases/download/prbrief-safe-global-safe-wallet-monorepo-pr-7734-c51ca4d62224/video-github.mp4)** (direct MP4)

This PR fixes the Space safe bar's chain selector ('All networks' accordion) which previously listed every non-deployed chain regardless of whether the Safe could actually be added to it, leading to silent failures for incompatible Safes (outdated mastercopy, zkSync/Lens, etc.). A new shared hook `useAddNetworkState` is introduced under `@/features/multichain` that composes `useSafeCreationData`, `useCompatibleNetworks`, and `hasMultiChainAddNetworkFeature` to produce a typed state object with `unavailableReason`, `availableNetworks`, and `loading`. A new `AllNetworksSection` component consumes this hook and renders one of four states: nothing (feature disabled), a muted info row with optional tooltip (unavailable reason), a spinner (loading), or the accordion with per-chain disabled rows for incompatible chains. `ChainSelectorBlock` drops its `availableChains` prop in favour of `safeAddress` + `deployedChainIds`, and `useSpaceChainSelector` removes its naive set-difference derivation entirely.

<details>

## Summary

This PR fixes the Space safe bar's chain selector ('All networks' accordion) which previously listed every non-deployed chain regardless of whether the Safe could actually be added to it, leading to silent failures for incompatible Safes (outdated mastercopy, zkSync/Lens, etc.). A new shared hook `useAddNetworkState` is introduced under `@/features/multichain` that composes `useSafeCreationData`, `useCompatibleNetworks`, and `hasMultiChainAddNetworkFeature` to produce a typed state object with `unavailableReason`, `availableNetworks`, and `loading`. A new `AllNetworksSection` component consumes this hook and renders one of four states: nothing (feature disabled), a muted info row with optional tooltip (unavailable reason), a spinner (loading), or the accordion with per-chain disabled rows for incompatible chains. `ChainSelectorBlock` drops its `availableChains` prop in favour of `safeAddress` + `deployedChainIds`, and `useSpaceChainSelector` removes its naive set-difference derivation entirely.

## Architecture notes

The data flow now runs: `SpaceChainSelector` → `ChainSelectorBlock` (passes `safeAddress` + `deployedChainIds`) → `AllNetworksSection` → `useAddNetworkState` → `useSafeCreationData` + `useCompatibleNetworks` + `hasMultiChainAddNetworkFeature`. This means `useSafeCreationData` (which hits RPCs across all deployed chains) now fires on every dropdown open rather than only when the add-network dialog is opened, which is a new performance cost. The `availableChains` field is removed from `useSpaceChainSelector`'s return type and from `ChainSelectorBlock`'s prop contract; any undiscovered caller would silently receive `undefined`. New public exports (`useAddNetworkState`, `AddNetworkState`, `AddNetworkUnavailableReason`, `AvailableNetwork`) are added to the `@/features/multichain` barrel via both `hooks/index.ts` and `features/multichain/index.ts`.

## Ripple risks

Analyzed 5 changed exports. No untouched external callers found — all usages appear to be inside the PR diff.

_Analyzed 12 file(s), 10 changed export(s), 0 uncovered callsite(s) in 2.2s._

## Walkthrough outline

_19 steps across 6 chapters, approximately 4 min 15s._

### Why this change

1. **motivation** · `apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceChainSelector.ts` L16-35 `useSpaceChainSelector`
   Here is the root of the problem. The old hook derived available chains by simply subtracting already-deployed chain IDs from the full chain config list. That naive set-difference had no idea whether the Safe was actually compatible with those chains, so incompatible ones slipped through silently.

2. **motivation** · `apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceChainSelector.ts` L16-35 `useSpaceChainSelector`
   The result was that clicking a chain in the All Networks accordion could silently fail later — outdated mastercopy, zkSync incompatibility, missing creation data. The user got no warning at all. This PR fixes that by moving the eligibility check much earlier in the flow.

### The new state hook

3. **change** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L8-23 `AddNetworkState`
   The centrepiece of this PR is the new useAddNetworkState hook, and it starts with its type contract. AddNetworkUnavailableReason is a union of safe-specific and outdated-mastercopy, and AddNetworkState bundles loading, availableNetworks, that reason, an error, and a feature-enabled flag into one object.

4. **change** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L25-55 `useAddNetworkState`
   The hook itself calls useSafeCreationData and useCompatibleNetworks, then filters the compatible list down to chains where hasMultiChainAddNetworkFeature is true and the chain is not already deployed. This is the real eligibility check that was missing before.

5. **change** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L56-90 `useAddNetworkState`
   The unavailableReason is resolved through a short-circuit IIFE. A creation data error maps to safe-specific; an empty compatible list maps to outdated-mastercopy; all networks being unavailable also maps to safe-specific. When a reason is set, availableNetworks is forced to an empty array so callers never see a partial list alongside an error.

6. **change** · `apps/web/src/features/multichain/hooks/__tests__/useAddNetworkState.test.ts` L1-50 `useAddNetworkState`
   The test file covers eight scenarios: happy path filtering, creation data error, empty compatible networks, all chains unavailable, feature flag off, feature flag filtering on target chains, already-deployed exclusion, and loading state propagation. That is solid coverage of every branch in the IIFE.

7. **change** · `apps/web/src/features/multichain/hooks/index.ts` L1-10
   The hooks barrel re-exports useAddNetworkState and its three companion types, making them part of the public multichain feature API. The main multichain index then re-exports them again so any consumer can import from the top-level feature path.

### The new visual section

8. **change** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.tsx` L16-35 `AllNetworksSectionProps`
   AllNetworksSection is the new visual owner of the All Networks area. Its props are simple: safeAddress, deployedChainIds, and an onAddNetwork callback. It calls useAddNetworkState internally and decides which of the four render states to show.

9. **change** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.tsx` L36-80
   The first two guard clauses handle the easy cases: if the feature is disabled the component returns null, and if unavailableReason is set it renders a muted info row. When the reason is safe-specific and an error message exists, the info icon becomes a tooltip trigger showing that raw message.

10. **change** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.tsx` L81-130
   Past the error guards, a loading state shows a centred spinner. Then comes the accordion itself, where each chain row checks the available flag from useCompatibleNetworks. Incompatible chains render as a disabled button labelled Not available and cannot fire onAddNetwork.

11. **change** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.test.tsx` L1-50
   The component test mocks useAddNetworkState and exercises all render branches, including analytics: SHOW_ALL_NETWORKS fires only on expand, and ADD_NEW_NETWORK fires with the top-bar label on a successful chain click. Disabled chain rows are confirmed to not fire the callback.

12. **ripple** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/ChainSelectorBlock.tsx` L9-30 `ChainSelectorBlockProps`
   ChainSelectorBlock is the first consumer that had to adapt. Its prop contract drops availableChains and gains safeAddress plus deployedChainIds. The inline accordion markup is replaced by a single AllNetworksSection element, forwarding those two props and the onAddNetwork callback.

13. **ripple** · `apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.tsx` L1-40
   SpaceChainSelector no longer destructures availableChains from useSpaceChainSelector. Instead it passes safeAddress and deployedChainIds straight down to ChainSelectorBlock. The chain-diffing logic is gone from this layer entirely.

14. **ripple** · `apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceChainSelector.ts` L16-35 `useSpaceChainSelector`
   Back in the hook, the useMemo that computed availableChains is simply deleted. The hook's return type no longer includes that field, so any undiscovered caller that still destructures it would silently receive undefined — worth a quick search to be sure.

15. **ripple** · `apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.stories.tsx` L47-90 `SingleChain`
   The stories file swaps availableChains for safeAddress and deployedChainIds on both the SingleChain and MultiChain stories, and removes the NoAvailableChains story entirely. That last deletion is worth noting: there is no replacement story covering the unavailable-reason visual states in Storybook.

### What can go wrong

16. **risk** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L56-90 `useAddNetworkState`
   Watch the IIFE evaluation order carefully. safeCreationError is checked before noAvailableNetworks, which is checked before isUnsupportedSafeCreationVersion. That means outdated-mastercopy can never be returned when a creation error is also present — the safe-specific reason wins. Confirm this ordering is intentional and not an accidental priority inversion.

17. **risk** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L25-55 `useAddNetworkState`
   useSafeCreationData is called unconditionally, even when isFeatureEnabled is false. That means RPC calls fire on every dropdown open regardless of the feature flag. The PR description flags this as a known cost, but it is worth considering a short-circuit when the feature is disabled to avoid unnecessary network traffic.

18. **risk** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.tsx` L81-130
   When loading finishes and availableNetworks is empty but unavailableReason is null, the component returns null with no feedback to the user. That silent empty state could be confusing if the spinner disappears and nothing replaces it. Also note that when error is set but unavailableReason is null, the error is silently swallowed with no UI indication.

### The bottom line

19. **wrap** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L8-23 `AddNetworkState`
   The bottom line: this PR replaces a naive chain diff with a proper compatibility gate, surfacing clear error states before the user ever clicks into a broken add-network flow. The main things to verify before merging are the IIFE priority ordering, the unconditional RPC call when the feature is off, and the missing Storybook coverage for unavailable-reason states.

## Reviewer checklist

- [ ] Verify `ChainSelectorBlock` TypeScript prop types have `availableChains` fully removed and `safeAddress`/`deployedChainIds` added, and confirm no other caller in the codebase still passes `availableChains` or destructures it from `useSpaceChainSelector`.
- [ ] Audit the `useAddNetworkState` IIFE logic ordering: confirm that the `outdated-mastercopy` reason can actually be reached when `safeCreationError` is also set, and that an entirely empty `allCompatibleChains` list (not just all-unavailable) is correctly handled.
- [ ] Check that `useSafeCreationData` is not called (or is short-circuited) when `isFeatureEnabled` is false, to avoid unnecessary RPC calls on every selector open for chains where the feature is disabled.
- [ ] Confirm `AllNetworksSection` returns a visible empty/null state (not a silent blank) when loading finishes with zero available networks, so users are not left with an unexplained empty area.
- [ ] Validate that the `SHOW_ALL_NETWORKS` analytics event fires only on accordion expand (not collapse) by checking the `handleAccordionChange` cast from `unknown` to `string[]` — ensure the Accordion component actually passes an array and that `openedIds.includes('all-networks')` behaves correctly.
- [ ] Manually open the chain selector for a Safe with an outdated mastercopy and for a healthy multichain-eligible Safe to confirm the correct UI state renders in each case, and that clicking a disabled chain row does not trigger `onAddNetwork`.

## File-by-file

### `apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.stories.tsx`

The stories file removes the static `availableChains` array and the `NoAvailableChains` story, replacing `availableChains` prop with `safeAddress` and `deployedChainIds` props on `ChainSelectorBlock` in both `SingleChain` and `MultiChain` stories. The comment is updated to note that `useAddNetworkState` now drives the 'All networks' section dynamically.

_Likely reflects the prop contract change on `ChainSelectorBlock` where static chain derivation is replaced by hook-driven logic, requiring stories to supply `safeAddress`/`deployedChainIds` instead of a pre-computed list._

**Risks:**
- The `NoAvailableChains` story is deleted entirely — there is no replacement story covering the 'no available chains' / unavailable-reason visual states, reducing Storybook coverage for error/empty states
- The stub `safeAddress` values (0x000...001, 0x000...002) may not trigger realistic `useAddNetworkState` behavior in the mocked scenario; reviewers should confirm the mock scenario resolves hooks correctly
- If `useAddNetworkState` makes real RPC calls inside the Portal-rendered content, Storybook renders may hang or show spinners indefinitely without proper mocking

### `apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.test.tsx`

The test file updates the `ChainSelectorBlock` mock to accept and render `safeAddress` and `deployedChainIds` props (via `data-*` attributes), removes `availableChains` from all `mockUseSpaceChainSelector` return values, and adds a new test case asserting those two props are correctly forwarded to `ChainSelectorBlock`.

_Reflects the prop contract change where `ChainSelectorBlock` no longer receives `availableChains` but instead receives `safeAddress` and `deployedChainIds` to self-manage network availability logic._

**Risks:**
- Verify the real `ChainSelectorBlock` component signature matches the updated mock — if the prop names diverge the tests pass but the component silently receives wrong props
- Check that all other callers of `useSpaceChainSelector` that previously consumed `availableChains` have been updated; the test removal of `availableChains` from mock returns only proves the test compiles, not that no production code still reads it
- The new test only checks attribute presence/value, not that `safeAddress`/`deployedChainIds` are actually used to gate network availability — consider whether a higher-level integration test is needed

### `apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.tsx`

SpaceChainSelector removes `availableChains` from the destructured `useSpaceChainSelector` return value and stops passing it as a prop to `ChainSelectorBlock`, instead passing `safeAddress` and `deployedChainIds` directly to `ChainSelectorBlock`.

_Shifts responsibility for computing available networks from the call site into `ChainSelectorBlock` (via `AllNetworksSection`/`useAddNetworkState`), enabling proper compatibility checks rather than a naive set-difference._

**Risks:**
- Verify `ChainSelectorBlock` prop types were updated to accept `safeAddress`+`deployedChainIds` and remove `availableChains`
- Confirm `useSpaceChainSelector` no longer exports `availableChains` and that no other consumer depends on it
- Check that `deployedChainIds` was already available in `useSpaceChainSelector` return or was newly added

### `apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceChainSelector.test.ts`

Removed two test cases from `useSpaceChainSelector.test.ts` that verified the `availableChains` return value — specifically the tests asserting that `availableChains` excludes already-deployed chain IDs for both single-chain and multi-chain Safes.

_The `availableChains` property was removed from `useSpaceChainSelector`'s return shape as part of moving that logic into `useAddNetworkState`/`AllNetworksSection`, so the tests covering it are no longer applicable._

**Risks:**
- Verify that `availableChains` is fully removed from `useSpaceChainSelector` return type and not silently still exported/consumed elsewhere
- Confirm equivalent coverage of the 'exclude deployed chains' logic now exists in `useAddNetworkState` tests so there is no net loss of test coverage
- Check that no other callers of `useSpaceChainSelector` still destructure `availableChains` and silently receive `undefined`

### `apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceChainSelector.ts`

Removes the `availableChains` derived state (a naive `chainConfigs − deployedChainIds` diff) from `useSpaceChainSelector`, along with its `useMemo` and its inclusion in the hook's return value.

_The naive diff was over-permissive, listing every non-deployed chain regardless of compatibility; responsibility for computing available chains is moved to `AllNetworksSection`/`useAddNetworkState` which performs proper compatibility checks._

**Risks:**
- Verify all callers of `useSpaceChainSelector` no longer reference `availableChains` — a missed destructure would silently be `undefined`
- Confirm `ChainSelectorBlock` prop contract was updated in lockstep to remove `availableChains` prop
- Check that no TypeScript type for the hook's return value still declares `availableChains`

### `apps/web/src/features/multichain/hooks/__tests__/useAddNetworkState.test.ts`

Adds a new test file for `useAddNetworkState` hook covering 8 scenarios: happy path filtering, creation data error (`safe-specific`), empty compatible networks (`outdated-mastercopy`), all chains unavailable (`safe-specific`), feature flag disabled on current chain, feature flag filtering on target chains, already-deployed chain exclusion, and loading state propagation.

_Validates the core logic of the new shared hook that gates which chains appear in the Space chain selector, ensuring the correct `unavailableReason` is set under each failure mode._

**Risks:**
- Test for 'all compatible chains unavailable' expects `unavailableReason='safe-specific'` but the compatible list includes the current chain (chainId '1') which is already deployed — verify the hook correctly handles this edge case vs a truly empty post-filter list
- No test covers the case where `compatible` contains a mix of available and unavailable chains (partial availability) — the hook's behavior in that scenario is untested
- The `setupDefaults` mock for `useCompatibleNetworks` returns all `configs` including the already-deployed chain; tests rely on the hook itself filtering by `deployedChainIds`, so if that filtering is missing the assertions would still pass incorrectly
- Loading state test with `compatible: []` expects `unavailableReason` to stay null — confirm the hook checks `loading` before evaluating empty compatible list to avoid a race condition

### `apps/web/src/features/multichain/hooks/index.ts`

Adds two new exports from `apps/web/src/features/multichain/hooks/index.ts`: the `useAddNetworkState` hook and three TypeScript types (`AddNetworkState`, `AddNetworkUnavailableReason`, `AvailableNetwork`) from the new `./useAddNetworkState` module.

_Makes the new hook and its types part of the public API of the `@/features/multichain` barrel so downstream consumers like `AllNetworksSection` can import them without reaching into internal paths._

**Risks:**
- Verify `./useAddNetworkState` file actually exists and exports all three named types to avoid build-time errors
- Check that no existing consumer was already importing these identifiers from a different path, which could cause duplicate-export conflicts

### `apps/web/src/features/multichain/hooks/useAddNetworkState.ts`

New hook `useAddNetworkState(safeAddress, deployedChainIds)` added that composes `useSafeCreationData`, `useCompatibleNetworks`, and `hasMultiChainAddNetworkFeature` to return a unified `AddNetworkState` object with `loading`, `availableNetworks`, `unavailableReason`, `error`, and `isFeatureEnabled` fields.

_Centralises the eligibility logic for adding a Safe to a new network so that UI components can show accurate error states instead of silently failing when a Safe is incompatible._

**Risks:**
- Logic conflict: `noAvailableNetworks` is true only when `availableNetworks.length > 0 && every(!available)`, but `isUnsupportedSafeCreationVersion` checks `safeCreationData && !allCompatibleChains?.length`. The IIFE evaluates `safeCreationError` before `noAvailableNetworks` before `isUnsupportedSafeCreationVersion`, so `outdated-mastercopy` can never be returned if there is also a `safeCreationError` — verify this ordering is intentional.
- `noAvailableNetworks` condition requires `availableNetworks.length > 0` (non-empty) AND all unavailable, meaning a list that is entirely empty (no compatible chains at all) does NOT trigger 'safe-specific' via this path — only `isUnsupportedSafeCreationVersion` or `safeCreationError` would catch it; confirm empty-list case is handled.
- `availableNetworks` is filtered to chains where `hasMultiChainAddNetworkFeature(config)` is true, but the `AvailableNetwork` type includes an `available: boolean` field that is never explicitly set here — it must come from `useCompatibleNetworks` returning `Chain & { available: boolean }`; verify the upstream type actually carries this field.
- When `unavailableReason` is non-null, `availableNetworks` is forced to `[]`, discarding potentially useful data; callers relying on the list for any secondary purpose will get nothing.
- `useSafeCreationData` is called unconditionally even when `isFeatureEnabled` is false, causing unnecessary RPC calls; consider short-circuiting when the feature is disabled.

### `apps/web/src/features/multichain/index.ts`

The `multichain` index re-exports `useAddNetworkState` function and three new types (`AddNetworkState`, `AddNetworkUnavailableReason`, `AvailableNetwork`) from `./hooks`.

_Exposes the new shared hook and its types as part of the public `@/features/multichain` API so consumers like `AllNetworksSection` can import them without reaching into internal paths._

**Risks:**
- Verify that `useAddNetworkState`, `AddNetworkState`, `AddNetworkUnavailableReason`, and `AvailableNetwork` are actually exported from `./hooks` (i.e., the hooks barrel file re-exports them correctly) to avoid silent undefined exports.
- Check no circular dependency is introduced by adding these exports to the index.

### `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.test.tsx`

New test file `AllNetworksSection.test.tsx` added with 232 lines covering all render states of the `AllNetworksSection` component: feature disabled, safe-specific unavailability, outdated-mastercopy, loading spinner, empty available networks, accordion expansion, per-chain add buttons, disabled chains (`available=false`), and analytics events (`SHOW_ALL_NETWORKS`, `ADD_NEW_NETWORK`).

_Provides unit-test coverage for the new `AllNetworksSection` component and its integration with `useAddNetworkState`, ensuring the gating logic and analytics fire correctly before the component ships._

**Risks:**
- Tests mock `useAddNetworkState` entirely — no integration coverage of the actual hook logic; bugs in hook composition won't be caught here.
- The 'renders nothing when there are no available networks and no error' case (feature enabled, no reason, empty list) expects an empty DOM — verify the real component actually returns null in this state and not a wrapper element.
- Accordion expand/collapse relies on `fireEvent.click(screen.getByText('All networks'))` — if the accordion trigger text changes or is wrapped in a non-interactive element, tests will silently pass without exercising the accordion.
- Tooltip presence is asserted via `data-testid='chain-selector-unavailable-tooltip'` — confirm the production component actually sets this testid on the tooltip trigger, not just the tooltip content.
- No test covers the case where `error` is present but `unavailableReason` is NOT `safe-specific` — unclear if tooltip should appear in that scenario.
- Analytics test clears mock between expand and collapse but does not reset between describe blocks; `beforeEach(jest.resetAllMocks)` at the top level should handle it, but nested `beforeEach` in the analytics block does not call `mockTrack.mockClear()` before each analytics test, relying on outer reset.

### `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.tsx`

New `AllNetworksSection` component added that consumes `useAddNetworkState(safeAddress, deployedChainIds)` and renders one of four states: null (feature disabled), a muted info row with optional tooltip (unavailableReason set), a spinner (loading), or an accordion listing available networks with disabled rows for incompatible chains.

_Replaces the naive chain-diffing approach in `ChainSelectorBlock`/`useSpaceChainSelector` with a gated, compatibility-aware UI that prevents users from clicking into a broken add-network flow for ineligible Safes._

**Risks:**
- handleAccordionChange casts `value` to `unknown` then `string[]` — if the Accordion component passes a single string instead of an array, `openedIds.includes('all-networks')` will always be false and the analytics event will never fire
- onClick guard `!disabled && handleChainClick(...)` is redundant with the `disabled` attribute on the button but harmless; verify the button's disabled state actually prevents synthetic click events in all target browsers
- Tooltip is only shown when `error?.message` is truthy, but `unavailableReason` could be 'safe-specific' with no error — confirm UX intent that the info icon is non-interactive in that case
- TooltipTrigger uses a `render` prop pattern with a raw `<span>` — verify this matches the actual Tooltip API in use; misuse could silently drop the trigger
- availableNetworks.length === 0 returns null after the loading check, meaning a successfully resolved but empty list is invisible with no feedback — could confuse users if loading finishes with zero results
- No error boundary or explicit error state rendered when `error` is set but `unavailableReason` is null — error may be silently swallowed

### `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/ChainSelectorBlock.tsx`

ChainSelectorBlock's inline 'All networks' accordion (with naive availableChains prop) is replaced by a new AllNetworksSection component. The prop availableChains is removed and replaced with safeAddress + deployedChainIds, which are forwarded to AllNetworksSection.

_Delegates network availability logic to AllNetworksSection/useAddNetworkState so only genuinely compatible chains are shown, preventing users from clicking chains that would silently fail._

**Risks:**
- Verify all callers of ChainSelectorBlock have been updated to pass safeAddress and deployedChainIds instead of availableChains
- handleAddNetworkClick is now passed to AllNetworksSection — confirm analytics events (ADD_NEW_NETWORK) still fire correctly through this indirection
- ChainLogo and Plus imports removed; confirm Plus is not needed elsewhere in this file
- AllNetworksSection is a new component — its error/loading/unavailable states need independent review for correctness and accessibility

---
_Generated by prbrief • model: claude-sonnet-4-6 • 14 LLM calls • $0.25 • 129.6s_

</details>

<sub>Generated by <a href="https://popcorn.report">Popcorn Review</a>. · `safe-global/safe-wallet-monorepo@c51ca4d62224`</sub>

ChillYao:
> ## PR Briefing
> [Open the PR briefing](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-wallet-monorepo%2Fpr-7734%2Fc51ca4d62224%2Fbundle.json)
> 
> ▶ **[Open the interactive review](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-wallet-monorepo%2Fpr-7734%2Fc51ca4d62224%2Fbundle.json)** ▶ **[Watch the walkthrough video](https://github.com/katspaugh/popcorn-review-artifacts/releases/download/prbrief-safe-global-safe-wallet-monorepo-pr-7734-c51ca4d62224/video-github.mp4)** (direct MP4)
> 
> This PR fixes the Space safe bar's chain selector ('All networks' accordion) which previously listed every non-deployed chain regardless of whether the Safe could actually be added to it, leading to silent failures for incompatible Safes (outdated mastercopy, zkSync/Lens, etc.). A new shared hook `useAddNetworkState` is introduced under `@/features/multichain` that composes `useSafeCreationData`, `useCompatibleNetworks`, and `hasMultiChainAddNetworkFeature` to produce a typed state object with `unavailableReason`, `availableNetworks`, and `loading`. A new `AllNetworksSection` component consumes this hook and renders one of four states: nothing (feature disabled), a muted info row with optional tooltip (unavailable reason), a spinner (loading), or the accordion with per-chain disabled rows for incompatible chains. `ChainSelectorBlock` drops its `availableChains` prop in favour of `safeAddress` + `deployedChainIds`, and `useSpaceChainSelector` removes its naive set-difference derivation entirely.
> 
> Full briefing (click to expand)
> ## Summary
> This PR fixes the Space safe bar's chain selector ('All networks' accordion) which previously listed every non-deployed chain regardless of whether the Safe could actually be added to it, leading to silent failures for incompatible Safes (outdated mastercopy, zkSync/Lens, etc.). A new shared hook `useAddNetworkState` is introduced under `@/features/multichain` that composes `useSafeCreationData`, `useCompatibleNetworks`, and `hasMultiChainAddNetworkFeature` to produce a typed state object with `unavailableReason`, `availableNetworks`, and `loading`. A new `AllNetworksSection` component consumes this hook and renders one of four states: nothing (feature disabled), a muted info row with optional tooltip (unavailable reason), a spinner (loading), or the accordion with per-chain disabled rows for incompatible chains. `ChainSelectorBlock` drops its `availableChains` prop in favour of `safeAddress` + `deployedChainIds`, and `useSpaceChainSelector` removes its naive set-difference derivation entirely.
> 
> ## Architecture notes
> The data flow now runs: `SpaceChainSelector` → `ChainSelectorBlock` (passes `safeAddress` + `deployedChainIds`) → `AllNetworksSection` → `useAddNetworkState` → `useSafeCreationData` + `useCompatibleNetworks` + `hasMultiChainAddNetworkFeature`. This means `useSafeCreationData` (which hits RPCs across all deployed chains) now fires on every dropdown open rather than only when the add-network dialog is opened, which is a new performance cost. The `availableChains` field is removed from `useSpaceChainSelector`'s return type and from `ChainSelectorBlock`'s prop contract; any undiscovered caller would silently receive `undefined`. New public exports (`useAddNetworkState`, `AddNetworkState`, `AddNetworkUnavailableReason`, `AvailableNetwork`) are added to the `@/features/multichain` barrel via both `hooks/index.ts` and `features/multichain/index.ts`.
> 
> ## Ripple risks
> Analyzed 5 changed exports. No untouched external callers found — all usages appear to be inside the PR diff.
> 
> _Analyzed 12 file(s), 10 changed export(s), 0 uncovered callsite(s) in 2.2s._
> 
> ## Walkthrough outline
> _19 steps across 6 chapters, approximately 4 min 15s._
> 
> ### Why this change
> 1. **motivation** · `apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceChainSelector.ts` L16-35 `useSpaceChainSelector`
>    Here is the root of the problem. The old hook derived available chains by simply subtracting already-deployed chain IDs from the full chain config list. That naive set-difference had no idea whether the Safe was actually compatible with those chains, so incompatible ones slipped through silently.
> 2. **motivation** · `apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceChainSelector.ts` L16-35 `useSpaceChainSelector`
>    The result was that clicking a chain in the All Networks accordion could silently fail later — outdated mastercopy, zkSync incompatibility, missing creation data. The user got no warning at all. This PR fixes that by moving the eligibility check much earlier in the flow.
> 
> ### The new state hook
> 3. **change** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L8-23 `AddNetworkState`
>    The centrepiece of this PR is the new useAddNetworkState hook, and it starts with its type contract. AddNetworkUnavailableReason is a union of safe-specific and outdated-mastercopy, and AddNetworkState bundles loading, availableNetworks, that reason, an error, and a feature-enabled flag into one object.
> 4. **change** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L25-55 `useAddNetworkState`
>    The hook itself calls useSafeCreationData and useCompatibleNetworks, then filters the compatible list down to chains where hasMultiChainAddNetworkFeature is true and the chain is not already deployed. This is the real eligibility check that was missing before.
> 5. **change** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L56-90 `useAddNetworkState`
>    The unavailableReason is resolved through a short-circuit IIFE. A creation data error maps to safe-specific; an empty compatible list maps to outdated-mastercopy; all networks being unavailable also maps to safe-specific. When a reason is set, availableNetworks is forced to an empty array so callers never see a partial list alongside an error.
> 6. **change** · `apps/web/src/features/multichain/hooks/__tests__/useAddNetworkState.test.ts` L1-50 `useAddNetworkState`
>    The test file covers eight scenarios: happy path filtering, creation data error, empty compatible networks, all chains unavailable, feature flag off, feature flag filtering on target chains, already-deployed exclusion, and loading state propagation. That is solid coverage of every branch in the IIFE.
> 7. **change** · `apps/web/src/features/multichain/hooks/index.ts` L1-10
>    The hooks barrel re-exports useAddNetworkState and its three companion types, making them part of the public multichain feature API. The main multichain index then re-exports them again so any consumer can import from the top-level feature path.
> 
> ### The new visual section
> 8. **change** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.tsx` L16-35 `AllNetworksSectionProps`
>    AllNetworksSection is the new visual owner of the All Networks area. Its props are simple: safeAddress, deployedChainIds, and an onAddNetwork callback. It calls useAddNetworkState internally and decides which of the four render states to show.
> 9. **change** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.tsx` L36-80
>    The first two guard clauses handle the easy cases: if the feature is disabled the component returns null, and if unavailableReason is set it renders a muted info row. When the reason is safe-specific and an error message exists, the info icon becomes a tooltip trigger showing that raw message.
> 10. **change** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.tsx` L81-130
>     Past the error guards, a loading state shows a centred spinner. Then comes the accordion itself, where each chain row checks the available flag from useCompatibleNetworks. Incompatible chains render as a disabled button labelled Not available and cannot fire onAddNetwork.
> 11. **change** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.test.tsx` L1-50
>     The component test mocks useAddNetworkState and exercises all render branches, including analytics: SHOW_ALL_NETWORKS fires only on expand, and ADD_NEW_NETWORK fires with the top-bar label on a successful chain click. Disabled chain rows are confirmed to not fire the callback.
> 
> ### Consumers that had to adapt
> 12. **ripple** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/ChainSelectorBlock.tsx` L9-30 `ChainSelectorBlockProps`
>     ChainSelectorBlock is the first consumer that had to adapt. Its prop contract drops availableChains and gains safeAddress plus deployedChainIds. The inline accordion markup is replaced by a single AllNetworksSection element, forwarding those two props and the onAddNetwork callback.
> 13. **ripple** · `apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.tsx` L1-40
>     SpaceChainSelector no longer destructures availableChains from useSpaceChainSelector. Instead it passes safeAddress and deployedChainIds straight down to ChainSelectorBlock. The chain-diffing logic is gone from this layer entirely.
> 14. **ripple** · `apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceChainSelector.ts` L16-35 `useSpaceChainSelector`
>     Back in the hook, the useMemo that computed availableChains is simply deleted. The hook's return type no longer includes that field, so any undiscovered caller that still destructures it would silently receive undefined — worth a quick search to be sure.
> 15. **ripple** · `apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.stories.tsx` L47-90 `SingleChain`
>     The stories file swaps availableChains for safeAddress and deployedChainIds on both the SingleChain and MultiChain stories, and removes the NoAvailableChains story entirely. That last deletion is worth noting: there is no replacement story covering the unavailable-reason visual states in Storybook.
> 
> ### What can go wrong
> 16. **risk** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L56-90 `useAddNetworkState`
>     Watch the IIFE evaluation order carefully. safeCreationError is checked before noAvailableNetworks, which is checked before isUnsupportedSafeCreationVersion. That means outdated-mastercopy can never be returned when a creation error is also present — the safe-specific reason wins. Confirm this ordering is intentional and not an accidental priority inversion.
> 17. **risk** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L25-55 `useAddNetworkState`
>     useSafeCreationData is called unconditionally, even when isFeatureEnabled is false. That means RPC calls fire on every dropdown open regardless of the feature flag. The PR description flags this as a known cost, but it is worth considering a short-circuit when the feature is disabled to avoid unnecessary network traffic.
> 18. **risk** · `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.tsx` L81-130
>     When loading finishes and availableNetworks is empty but unavailableReason is null, the component returns null with no feedback to the user. That silent empty state could be confusing if the spinner disappears and nothing replaces it. Also note that when error is set but unavailableReason is null, the error is silently swallowed with no UI indication.
> 
> ### The bottom line
> 19. **wrap** · `apps/web/src/features/multichain/hooks/useAddNetworkState.ts` L8-23 `AddNetworkState`
>     The bottom line: this PR replaces a naive chain diff with a proper compatibility gate, surfacing clear error states before the user ever clicks into a broken add-network flow. The main things to verify before merging are the IIFE priority ordering, the unconditional RPC call when the feature is off, and the missing Storybook coverage for unavailable-reason states.
> 
> ## Reviewer checklist
> * [ ]  Verify `ChainSelectorBlock` TypeScript prop types have `availableChains` fully removed and `safeAddress`/`deployedChainIds` added, and confirm no other caller in the codebase still passes `availableChains` or destructures it from `useSpaceChainSelector`.
> * [ ]  Audit the `useAddNetworkState` IIFE logic ordering: confirm that the `outdated-mastercopy` reason can actually be reached when `safeCreationError` is also set, and that an entirely empty `allCompatibleChains` list (not just all-unavailable) is correctly handled.
> * [ ]  Check that `useSafeCreationData` is not called (or is short-circuited) when `isFeatureEnabled` is false, to avoid unnecessary RPC calls on every selector open for chains where the feature is disabled.
> * [ ]  Confirm `AllNetworksSection` returns a visible empty/null state (not a silent blank) when loading finishes with zero available networks, so users are not left with an unexplained empty area.
> * [ ]  Validate that the `SHOW_ALL_NETWORKS` analytics event fires only on accordion expand (not collapse) by checking the `handleAccordionChange` cast from `unknown` to `string[]` — ensure the Accordion component actually passes an array and that `openedIds.includes('all-networks')` behaves correctly.
> * [ ]  Manually open the chain selector for a Safe with an outdated mastercopy and for a healthy multichain-eligible Safe to confirm the correct UI state renders in each case, and that clicking a disabled chain row does not trigger `onAddNetwork`.
> 
> ## File-by-file
> ### `apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.stories.tsx`
> The stories file removes the static `availableChains` array and the `NoAvailableChains` story, replacing `availableChains` prop with `safeAddress` and `deployedChainIds` props on `ChainSelectorBlock` in both `SingleChain` and `MultiChain` stories. The comment is updated to note that `useAddNetworkState` now drives the 'All networks' section dynamically.
> 
> _Likely reflects the prop contract change on `ChainSelectorBlock` where static chain derivation is replaced by hook-driven logic, requiring stories to supply `safeAddress`/`deployedChainIds` instead of a pre-computed list._
> 
> **Risks:**
> 
> * The `NoAvailableChains` story is deleted entirely — there is no replacement story covering the 'no available chains' / unavailable-reason visual states, reducing Storybook coverage for error/empty states
> * The stub `safeAddress` values (0x000...001, 0x000...002) may not trigger realistic `useAddNetworkState` behavior in the mocked scenario; reviewers should confirm the mock scenario resolves hooks correctly
> * If `useAddNetworkState` makes real RPC calls inside the Portal-rendered content, Storybook renders may hang or show spinners indefinitely without proper mocking
> 
> ### `apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.test.tsx`
> The test file updates the `ChainSelectorBlock` mock to accept and render `safeAddress` and `deployedChainIds` props (via `data-*` attributes), removes `availableChains` from all `mockUseSpaceChainSelector` return values, and adds a new test case asserting those two props are correctly forwarded to `ChainSelectorBlock`.
> 
> _Reflects the prop contract change where `ChainSelectorBlock` no longer receives `availableChains` but instead receives `safeAddress` and `deployedChainIds` to self-manage network availability logic._
> 
> **Risks:**
> 
> * Verify the real `ChainSelectorBlock` component signature matches the updated mock — if the prop names diverge the tests pass but the component silently receives wrong props
> * Check that all other callers of `useSpaceChainSelector` that previously consumed `availableChains` have been updated; the test removal of `availableChains` from mock returns only proves the test compiles, not that no production code still reads it
> * The new test only checks attribute presence/value, not that `safeAddress`/`deployedChainIds` are actually used to gate network availability — consider whether a higher-level integration test is needed
> 
> ### `apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.tsx`
> SpaceChainSelector removes `availableChains` from the destructured `useSpaceChainSelector` return value and stops passing it as a prop to `ChainSelectorBlock`, instead passing `safeAddress` and `deployedChainIds` directly to `ChainSelectorBlock`.
> 
> _Shifts responsibility for computing available networks from the call site into `ChainSelectorBlock` (via `AllNetworksSection`/`useAddNetworkState`), enabling proper compatibility checks rather than a naive set-difference._
> 
> **Risks:**
> 
> * Verify `ChainSelectorBlock` prop types were updated to accept `safeAddress`+`deployedChainIds` and remove `availableChains`
> * Confirm `useSpaceChainSelector` no longer exports `availableChains` and that no other consumer depends on it
> * Check that `deployedChainIds` was already available in `useSpaceChainSelector` return or was newly added
> 
> ### `apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceChainSelector.test.ts`
> Removed two test cases from `useSpaceChainSelector.test.ts` that verified the `availableChains` return value — specifically the tests asserting that `availableChains` excludes already-deployed chain IDs for both single-chain and multi-chain Safes.
> 
> _The `availableChains` property was removed from `useSpaceChainSelector`'s return shape as part of moving that logic into `useAddNetworkState`/`AllNetworksSection`, so the tests covering it are no longer applicable._
> 
> **Risks:**
> 
> * Verify that `availableChains` is fully removed from `useSpaceChainSelector` return type and not silently still exported/consumed elsewhere
> * Confirm equivalent coverage of the 'exclude deployed chains' logic now exists in `useAddNetworkState` tests so there is no net loss of test coverage
> * Check that no other callers of `useSpaceChainSelector` still destructure `availableChains` and silently receive `undefined`
> 
> ### `apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceChainSelector.ts`
> Removes the `availableChains` derived state (a naive `chainConfigs − deployedChainIds` diff) from `useSpaceChainSelector`, along with its `useMemo` and its inclusion in the hook's return value.
> 
> _The naive diff was over-permissive, listing every non-deployed chain regardless of compatibility; responsibility for computing available chains is moved to `AllNetworksSection`/`useAddNetworkState` which performs proper compatibility checks._
> 
> **Risks:**
> 
> * Verify all callers of `useSpaceChainSelector` no longer reference `availableChains` — a missed destructure would silently be `undefined`
> * Confirm `ChainSelectorBlock` prop contract was updated in lockstep to remove `availableChains` prop
> * Check that no TypeScript type for the hook's return value still declares `availableChains`
> 
> ### `apps/web/src/features/multichain/hooks/__tests__/useAddNetworkState.test.ts`
> Adds a new test file for `useAddNetworkState` hook covering 8 scenarios: happy path filtering, creation data error (`safe-specific`), empty compatible networks (`outdated-mastercopy`), all chains unavailable (`safe-specific`), feature flag disabled on current chain, feature flag filtering on target chains, already-deployed chain exclusion, and loading state propagation.
> 
> _Validates the core logic of the new shared hook that gates which chains appear in the Space chain selector, ensuring the correct `unavailableReason` is set under each failure mode._
> 
> **Risks:**
> 
> * Test for 'all compatible chains unavailable' expects `unavailableReason='safe-specific'` but the compatible list includes the current chain (chainId '1') which is already deployed — verify the hook correctly handles this edge case vs a truly empty post-filter list
> * No test covers the case where `compatible` contains a mix of available and unavailable chains (partial availability) — the hook's behavior in that scenario is untested
> * The `setupDefaults` mock for `useCompatibleNetworks` returns all `configs` including the already-deployed chain; tests rely on the hook itself filtering by `deployedChainIds`, so if that filtering is missing the assertions would still pass incorrectly
> * Loading state test with `compatible: []` expects `unavailableReason` to stay null — confirm the hook checks `loading` before evaluating empty compatible list to avoid a race condition
> 
> ### `apps/web/src/features/multichain/hooks/index.ts`
> Adds two new exports from `apps/web/src/features/multichain/hooks/index.ts`: the `useAddNetworkState` hook and three TypeScript types (`AddNetworkState`, `AddNetworkUnavailableReason`, `AvailableNetwork`) from the new `./useAddNetworkState` module.
> 
> _Makes the new hook and its types part of the public API of the `@/features/multichain` barrel so downstream consumers like `AllNetworksSection` can import them without reaching into internal paths._
> 
> **Risks:**
> 
> * Verify `./useAddNetworkState` file actually exists and exports all three named types to avoid build-time errors
> * Check that no existing consumer was already importing these identifiers from a different path, which could cause duplicate-export conflicts
> 
> ### `apps/web/src/features/multichain/hooks/useAddNetworkState.ts`
> New hook `useAddNetworkState(safeAddress, deployedChainIds)` added that composes `useSafeCreationData`, `useCompatibleNetworks`, and `hasMultiChainAddNetworkFeature` to return a unified `AddNetworkState` object with `loading`, `availableNetworks`, `unavailableReason`, `error`, and `isFeatureEnabled` fields.
> 
> _Centralises the eligibility logic for adding a Safe to a new network so that UI components can show accurate error states instead of silently failing when a Safe is incompatible._
> 
> **Risks:**
> 
> * Logic conflict: `noAvailableNetworks` is true only when `availableNetworks.length > 0 && every(!available)`, but `isUnsupportedSafeCreationVersion` checks `safeCreationData && !allCompatibleChains?.length`. The IIFE evaluates `safeCreationError` before `noAvailableNetworks` before `isUnsupportedSafeCreationVersion`, so `outdated-mastercopy` can never be returned if there is also a `safeCreationError` — verify this ordering is intentional.
> * `noAvailableNetworks` condition requires `availableNetworks.length > 0` (non-empty) AND all unavailable, meaning a list that is entirely empty (no compatible chains at all) does NOT trigger 'safe-specific' via this path — only `isUnsupportedSafeCreationVersion` or `safeCreationError` would catch it; confirm empty-list case is handled.
> * `availableNetworks` is filtered to chains where `hasMultiChainAddNetworkFeature(config)` is true, but the `AvailableNetwork` type includes an `available: boolean` field that is never explicitly set here — it must come from `useCompatibleNetworks` returning `Chain & { available: boolean }`; verify the upstream type actually carries this field.
> * When `unavailableReason` is non-null, `availableNetworks` is forced to `[]`, discarding potentially useful data; callers relying on the list for any secondary purpose will get nothing.
> * `useSafeCreationData` is called unconditionally even when `isFeatureEnabled` is false, causing unnecessary RPC calls; consider short-circuiting when the feature is disabled.
> 
> ### `apps/web/src/features/multichain/index.ts`
> The `multichain` index re-exports `useAddNetworkState` function and three new types (`AddNetworkState`, `AddNetworkUnavailableReason`, `AvailableNetwork`) from `./hooks`.
> 
> _Exposes the new shared hook and its types as part of the public `@/features/multichain` API so consumers like `AllNetworksSection` can import them without reaching into internal paths._
> 
> **Risks:**
> 
> * Verify that `useAddNetworkState`, `AddNetworkState`, `AddNetworkUnavailableReason`, and `AvailableNetwork` are actually exported from `./hooks` (i.e., the hooks barrel file re-exports them correctly) to avoid silent undefined exports.
> * Check no circular dependency is introduced by adding these exports to the index.
> 
> ### `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.test.tsx`
> New test file `AllNetworksSection.test.tsx` added with 232 lines covering all render states of the `AllNetworksSection` component: feature disabled, safe-specific unavailability, outdated-mastercopy, loading spinner, empty available networks, accordion expansion, per-chain add buttons, disabled chains (`available=false`), and analytics events (`SHOW_ALL_NETWORKS`, `ADD_NEW_NETWORK`).
> 
> _Provides unit-test coverage for the new `AllNetworksSection` component and its integration with `useAddNetworkState`, ensuring the gating logic and analytics fire correctly before the component ships._
> 
> **Risks:**
> 
> * Tests mock `useAddNetworkState` entirely — no integration coverage of the actual hook logic; bugs in hook composition won't be caught here.
> * The 'renders nothing when there are no available networks and no error' case (feature enabled, no reason, empty list) expects an empty DOM — verify the real component actually returns null in this state and not a wrapper element.
> * Accordion expand/collapse relies on `fireEvent.click(screen.getByText('All networks'))` — if the accordion trigger text changes or is wrapped in a non-interactive element, tests will silently pass without exercising the accordion.
> * Tooltip presence is asserted via `data-testid='chain-selector-unavailable-tooltip'` — confirm the production component actually sets this testid on the tooltip trigger, not just the tooltip content.
> * No test covers the case where `error` is present but `unavailableReason` is NOT `safe-specific` — unclear if tooltip should appear in that scenario.
> * Analytics test clears mock between expand and collapse but does not reset between describe blocks; `beforeEach(jest.resetAllMocks)` at the top level should handle it, but nested `beforeEach` in the analytics block does not call `mockTrack.mockClear()` before each analytics test, relying on outer reset.
> 
> ### `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/AllNetworksSection.tsx`
> New `AllNetworksSection` component added that consumes `useAddNetworkState(safeAddress, deployedChainIds)` and renders one of four states: null (feature disabled), a muted info row with optional tooltip (unavailableReason set), a spinner (loading), or an accordion listing available networks with disabled rows for incompatible chains.
> 
> _Replaces the naive chain-diffing approach in `ChainSelectorBlock`/`useSpaceChainSelector` with a gated, compatibility-aware UI that prevents users from clicking into a broken add-network flow for ineligible Safes._
> 
> **Risks:**
> 
> * handleAccordionChange casts `value` to `unknown` then `string[]` — if the Accordion component passes a single string instead of an array, `openedIds.includes('all-networks')` will always be false and the analytics event will never fire
> * onClick guard `!disabled && handleChainClick(...)` is redundant with the `disabled` attribute on the button but harmless; verify the button's disabled state actually prevents synthetic click events in all target browsers
> * Tooltip is only shown when `error?.message` is truthy, but `unavailableReason` could be 'safe-specific' with no error — confirm UX intent that the info icon is non-interactive in that case
> * TooltipTrigger uses a `render` prop pattern with a raw `<span>` — verify this matches the actual Tooltip API in use; misuse could silently drop the trigger
> * availableNetworks.length === 0 returns null after the loading check, meaning a successfully resolved but empty list is invisible with no feedback — could confuse users if loading finishes with zero results
> * No error boundary or explicit error state rendered when `error` is set but `unavailableReason` is null — error may be silently swallowed
> 
> ### `apps/web/src/features/spaces/components/SafeSelectorDropdown/components/ChainSelectorBlock.tsx`
> ChainSelectorBlock's inline 'All networks' accordion (with naive availableChains prop) is replaced by a new AllNetworksSection component. The prop availableChains is removed and replaced with safeAddress + deployedChainIds, which are forwarded to AllNetworksSection.
> 
> _Delegates network availability logic to AllNetworksSection/useAddNetworkState so only genuinely compatible chains are shown, preventing users from clicking chains that would silently fail._
> 
> **Risks:**
> 
> * Verify all callers of ChainSelectorBlock have been updated to pass safeAddress and deployedChainIds instead of availableChains
> * handleAddNetworkClick is now passed to AllNetworksSection — confirm analytics events (ADD_NEW_NETWORK) still fire correctly through this indirection
> * ChainLogo and Plus imports removed; confirm Plus is not needed elsewhere in this file
> * AllNetworksSection is a new component — its error/loading/unavailable states need independent review for correctness and accessibility
> 
> _Generated by prbrief • model: claude-sonnet-4-6 • 14 LLM calls • $0.25 • 129.6s_
> 
> Generated by [Popcorn Review](https://popcorn.report). · `safe-global/safe-wallet-monorepo@c51ca4d62224`

## ✅ Addressed

### 1. Stub `safeAddress` values in stories (`SpaceChainSelector.stories.tsx`)
> "The stub safeAddress values (0x000...001, 0x000...002) may not trigger realistic useAddNetworkState behavior..."

Now imports directly from the fixture source:

```ts
import { SAFE_ADDRESSES } from '@safe-global/test/msw/fixtures'
// ...
safeAddress={SAFE_ADDRESSES.efSafe.address}
```

No hardcoded address, no drift risk if the fixture moves.

---

### 2. Mock component signature alignment with real `ChainSelectorBlockProps` (`SpaceChainSelector.test.tsx`)
> "Verify the real ChainSelectorBlock component signature matches the updated mock — if the prop names diverge the tests pass but the component silently receives wrong props"

The mock factory now destructures against the real prop type:

```ts
import type { ChainSelectorBlockProps } from '...'
jest.mock(..., () =>
  function ChainSelectorBlock({ deployedChains, selectedChainId, safeAddress, deployedChainIds }: ChainSelectorBlockProps) { ... }
)
```

Any prop rename on the real component now fails type-check here immediately, instead of silently feeding `undefined` into the mock at runtime.

---

### 3. Q3 — `useSafeCreationData` wasted fetch when feature is disabled (`useAddNetworkState.ts`)
> "Check that useSafeCreationData is not called (or is short-circuited) when isFeatureEnabled is false, to avoid unnecessary RPC calls on every selector open for chains where the feature is disabled."

Real perf issue. Added a short-circuit:

```ts
const deployedChainConfigs = useMemo(
  () => (isFeatureEnabled ? configs.filter(...) : []),
  [isFeatureEnabled, configs, deployedChainIds],
)
```

Plus a regression test (`short-circuits useSafeCreationData to an empty chains list when feature is disabled`) that asserts the call args.

---

## ❌ Not changed (with reasoning)

### 4. `setupDefaults` mock letting filter regressions slip silently (`useAddNetworkState.test.ts`)
> "...if that filtering is missing the assertions would still pass incorrectly"

The filter **is** actually covered. Traced it test by test: if `!deployedChainIds.includes(...)` is removed, tests #1, #6, and #7 all fail (they expect `['137']` but would get `['1', '10', '137']`). So the regression wouldn't slip through silently.

---

### 5. Q1 — `availableChains` cleanup verification
> "Verify ChainSelectorBlock TypeScript prop types have availableChains fully removed, and confirm no other caller in the codebase still passes availableChains or destructures it from useSpaceChainSelector."

Verified clean. `grep availableChains apps/web/src` returns zero matches. The only hits are under `apps/mobile/src/features/Share`, an unrelated component.

---

### 6. Q2 — `useAddNetworkState` IIFE ordering audit
> "Audit the useAddNetworkState IIFE logic ordering: confirm that the outdated-mastercopy reason can actually be reached when safeCreationError is also set, and that an entirely empty allCompatibleChains list (not just all-unavailable) is correctly handled."

Ordering is intentional:

- `safeCreationError` is deliberately prioritized over `outdated-mastercopy` — a failed fetch shouldn't be interpreted as "mastercopy outdated" because we genuinely don't know the mastercopy yet.
- The `outdated-mastercopy` branch is effectively unreachable in practice (`useCompatibleNetworks` uses `.map()`, so its length always equals `configs.length` when `creation` is defined). It's kept as defensive code matching the Header's `UndeployedNetworks`. Happy to remove in a follow-up once the Header is retired — avoiding divergence during the migration.

---

### 7. Q4 — visible empty state vs silent null
> "Confirm AllNetworksSection returns a visible empty/null state (not a silent blank) when loading finishes with zero available networks, so users are not left with an unexplained empty area."

Intentional. The section only hides entirely when there is genuinely nothing to add (Safe already deployed on every supported chain). This matches the Header `NetworkSelector`, where the "Show all networks" button disappears in the same case. If an explicit "You're on all networks" message is desired, it should land in both places in a follow-up, not diverge here.

vmosharova (approved):
👍🏻

PR #7749, opened 2026-04-24
fix(web): cache no-ENS resolution results to prevent repeated lookups

katspaugh:
> Null means no name,
> cache both hit and empty miss —
> Infura sleeps easy.

## What it solves

Resolves: [WA-2092](https://linear.app/safe-global/issue/WA-2092/cache-addresses-with-no-ens-resolution)

When ENS name lookup is enabled as a fallback, addresses without a resolvable ENS name were not cached — only positive results were. Every hook mount for such an address triggered a fresh provider (Infura) lookup, which could be abused by rapid re-mounts or repeated navigations.

## How this PR fixes it

`useAddressResolver` now caches both positive and negative lookup outcomes:

- Cache type changed from `Record<string, Record<string, string>>` to `Record<string, Record<string, string | null>>`, where `null` is the sentinel for "no ENS".
- Cache hit check uses `in` (key presence) rather than truthiness, so a cached negative short-circuits the provider call.
- The write now happens inline after `lookupAddress` resolves (replacing the previous `useEffect` that only wrote on positive results).

## How to test it

1. Enable the `DOMAIN_LOOKUP` chain feature on a chain.
2. Enter an address that has no ENS reverse record (e.g. a fresh EOA) in a field that uses `useAddressResolver` (address book entry, tx recipient, etc.).
3. Re-mount the component (navigate away and back, or re-open a modal) with the same address.
4. Inspect network: only the first mount should hit the RPC provider; subsequent mounts resolve from cache.

Unit test `caches no-ENS results so repeated lookups do not hit the provider` covers this.

## Affected flows

- Address display / resolution wherever `useAddressResolver` is used (address book, tx builder recipients, owner lists, send flow).

## Blast radius

- `useAddressResolver` (module-level in-memory cache, scoped per chainId × address, per session)
- Consumers: every component that displays or resolves an address-to-name, on any chain with `DOMAIN_LOOKUP` enabled
- No API contract change, no persistence change, no feature flag change
- Mobile: not affected (web-only hook)

## Risks / not checked

- Transient RPC errors are currently swallowed by `services/ens/index.ts#lookupAddress` and returned as `undefined`. With this change, a transient failure gets cached as "no ENS" for the remainder of the session. Existing behavior was to retry on every mount — this PR trades that for fewer calls. If error-vs-null distinction matters, a follow-up in the ENS service is needed.
- Cache is in-memory and per-session; a page reload still triggers a fresh lookup. Cross-session abuse (many users, many reloads) is not addressed here and would need server-side caching.
- Did not exercise the address book override path manually (no change in that code path; address-book names still short-circuit ENS resolution upstream of the cache).
- Did not test on mobile (web-only file).

## Visual summary

```mermaid
flowchart TB
  subgraph Before
    A1[Hook mount] --> B1{Cache hit?}
    B1 -->|ens string cached| C1[Return cached]
    B1 -->|no entry, or null ENS| D1[Call lookupAddress]
    D1 --> E1{Result?}
    E1 -->|string| F1[Write cache, return]
    E1 -->|undefined| G1[Return undefined, NO cache write]
  end
  subgraph After
    A2[Hook mount] --> B2{Key in cache?}
    B2 -->|yes| C2[Return cached string or undefined]
    B2 -->|no| D2[Call lookupAddress]
    D2 --> F2[Write cache string or null, return]
  end
```

## Checklist

- [ ] I've tested the branch on mobile 📱 (N/A — web-only)
- [ ] I've documented how it affects the analytics (if at all) 📊 (no analytics impact)


katspaugh:
## PR Briefing

[Open the PR briefing](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-wallet-monorepo%2Fpr-7749%2F73621093d9ba%2Fbundle.json)

▶ **[Open the interactive review](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-wallet-monorepo%2Fpr-7749%2F73621093d9ba%2Fbundle.json)**
▶ **[Watch the walkthrough video](https://github.com/katspaugh/popcorn-review-artifacts/releases/download/prbrief-safe-global-safe-wallet-monorepo-pr-7749-73621093d9ba/video-github.mp4)** (direct MP4)

This PR fixes a performance issue in `useAddressResolver` where addresses with no ENS reverse record were never cached, causing a fresh Infura/RPC provider call on every component mount. The fix changes the module-level cache value type from `Record<string, string>` to `Record<string, string | null>`, uses `null` as a sentinel for 'no ENS name found', and switches the cache-hit check from truthiness to the `in` operator so null entries short-circuit provider calls. The previous `useEffect`-based cache write (which only fired on positive results) is replaced with an inline write inside `useAsync` that always stores the result, including `null`. A new unit test verifies that `lookupAddress` is called exactly once across two mounts for an address with no ENS name.

<details>

## Summary

This PR fixes a performance issue in `useAddressResolver` where addresses with no ENS reverse record were never cached, causing a fresh Infura/RPC provider call on every component mount. The fix changes the module-level cache value type from `Record<string, string>` to `Record<string, string | null>`, uses `null` as a sentinel for 'no ENS name found', and switches the cache-hit check from truthiness to the `in` operator so null entries short-circuit provider calls. The previous `useEffect`-based cache write (which only fired on positive results) is replaced with an inline write inside `useAsync` that always stores the result, including `null`. A new unit test verifies that `lookupAddress` is called exactly once across two mounts for an address with no ENS name.

## Architecture notes

The change is confined to the `useAddressResolver` hook and its test file in `apps/web/src/hooks/`. The module-level in-memory cache is keyed by `chainId × address` and is shared across all consumers (address book, tx builder, send flow, owner lists) for the duration of a browser session. No persistence layer, API contract, or feature flag is touched. The ENS service (`services/ens/index.ts#lookupAddress`) is unchanged, meaning transient RPC errors that return `undefined` are now indistinguishable from confirmed 'no ENS' results and will be cached as `null` for the session.

## Ripple risks

Analyzed 1 changed export. No untouched external callers found — all usages appear to be inside the PR diff.

_Analyzed 2 file(s), 1 changed export(s), 0 uncovered callsite(s) in 1.6s._

## Walkthrough outline

_18 steps across 5 chapters, approximately 4 min 55s._

### Why this change

1. **motivation** · `apps/web/src/hooks/useAddressResolver.ts` L1-15 `useAddressResolver`
   Here is the hook that resolves Ethereum addresses to ENS names across the whole app — address book, send flow, owner lists, all of it. Before this PR, every time a component mounted for an address with no ENS name, it fired a fresh call to the RPC provider. No caching of negative results meant Infura took a hit on every re-mount.

2. **motivation** · `apps/web/src/hooks/useAddressResolver.ts` L1-15
   The flow before this fix: cache only stored strings, so a null or undefined result from lookup address was never written. The next mount saw no cache entry and called the provider again. This diagram shows exactly how the before and after differ.

### The caching fix

3. **change** · `apps/web/src/hooks/useAddressResolver.ts` L1-5
   At the very top of the file, the module-level cache type has been widened. The inner record now maps address strings to either a string or null, where null is the sentinel meaning no ENS name was found. This single type change is what makes the rest of the fix possible.

4. **change** · `apps/web/src/hooks/useAddressResolver.ts` L13-22 `useAddressResolver`
   Inside the hook, the cache hit check has changed from a truthiness test to the in operator. Previously, a cached null would have been falsy and fallen through to a provider call. Now, if the key exists at all — whether the value is a string or null — the hook returns immediately without touching the network.

5. **change** · `apps/web/src/hooks/useAddressResolver.ts` L23-35 `useAddressResolver`
   The actual lookup now happens inline with an await inside the async callback passed to use async. After the provider responds, the result is written to the cache immediately — whether it came back as a name string or as undefined, which gets stored as null. This replaces the old use effect that only wrote on positive results.

6. **change** · `apps/web/src/hooks/useAddressResolver.ts` L35-45
   The return value from the async callback converts that cached null back to undefined before handing it to callers, so the external API of the hook is unchanged. Consumers still receive either a name string or undefined — they never see null directly.

7. **change** · `apps/web/src/hooks/useAddressResolver.ts` L13-45 `useAddressResolver`
   Stepping back to see the whole hook at once: the cache is keyed by chain ID then address, so each chain gets its own namespace. The ethers provider is in the use async dependency array, meaning a provider reference change will bypass the cache and trigger a fresh lookup — which is correct behavior when switching networks.

8. **change** · `apps/web/src/hooks/useAddressResolver.ts` L1-5
   The cache itself is module-level, meaning it lives for the entire browser session and is shared across every component that imports this hook. There is no explicit invalidation path — a page reload is the only reset. That is intentional for this PR, keeping the fix minimal.

9. **change** · `apps/web/src/hooks/useAddressResolver.ts` L13-45
   One subtle point: the address book name lookup still short-circuits ENS resolution entirely, upstream of this cache. So the new null-caching path only activates when no address book entry exists and the chain has the domain lookup feature enabled.

### The test that proves it

10. **ripple** · `apps/web/src/hooks/__tests__/useAddressResolver.test.ts` L1-20
   Now let us look at the test file. A new test case was added specifically for the negative-result scenario. It sets up lookup address to resolve to undefined — simulating an address with no ENS reverse record — then mounts the hook twice.

11. **ripple** · `apps/web/src/hooks/__tests__/useAddressResolver.test.ts` L20-45
   After both mounts complete, the test asserts that the lookup mock was called exactly once. That single assertion is the whole point: the second mount must have hit the cache and skipped the provider entirely. It also checks that the resolving flag is false on the second mount, confirming the hook settled synchronously from cache.

12. **ripple** · `apps/web/src/hooks/__tests__/useAddressResolver.test.ts` L45-60
   The test relies on the module-level cache persisting between the two render hook calls within the same test run. This works because Jest does not reset module state between calls inside a single test — but it does mean test ordering matters if the cache were ever pre-populated by an earlier test.

13. **ripple** · `apps/web/src/hooks/__tests__/useAddressResolver.test.ts` L1-60
   The existing positive-result tests are untouched, so we know the happy path — where lookup address returns a real name — still writes to cache and returns correctly. The new test only adds coverage for the previously unhandled empty case.

### What can go wrong

14. **risk** · `apps/web/src/hooks/useAddressResolver.ts` L23-35
   Here is the first gotcha worth reviewer attention. The ENS service's lookup address swallows transient RPC errors and returns undefined. With this change, a network blip during the first mount gets cached as null — permanently suppressing retries for the rest of the session. The old behavior retried on every mount; this PR trades retry resilience for fewer calls.

15. **risk** · `apps/web/src/hooks/useAddressResolver.ts` L23-35
   Second risk: the async callback inside use async writes to the module-level cache as a side effect. If use async ever calls that function concurrently for the same address — say, due to rapid re-renders before the first await resolves — two writes race to the same cache slot. In practice this is unlikely, but it is worth noting for future refactors.

16. **risk** · `apps/web/src/hooks/useAddressResolver.ts` L1-12
   Third risk: the cache has no time-to-live and no invalidation. If an address gains an ENS name mid-session — say, a user registers one in another tab — the stale null entry will be served until the page reloads. For most users this is acceptable, but it is a known limitation of the in-memory approach.

17. **risk** · `apps/web/src/hooks/useAddressResolver.ts` L13-22
   Fourth: the ethers provider reference is in the dependency array, so switching chains correctly bypasses the cache and fires a fresh lookup. But if the provider object is recreated on every render for some reason, the cache would be bypassed on every mount too — defeating the fix. Worth confirming provider stability in the broader app setup.

### The bottom line

18. **wrap** · `apps/web/src/hooks/useAddressResolver.ts` L1-45 `useAddressResolver`
   Bottom line: a two-line logic change — null as sentinel, in operator for presence check — stops the app from hammering the RPC provider for every address that has no ENS name, with a clear trade-off that transient errors are now silently cached for the session.

## Reviewer checklist

- [ ] Verify that the `in` operator cache-hit check correctly handles the case where `cache[chainId]` is undefined (i.e., no entry for the chain yet) to avoid a TypeError.
- [ ] Confirm that a transient RPC error from `lookupAddress` returning `undefined` being cached as `null` is an acceptable trade-off, or that a follow-up issue is filed to distinguish errors from confirmed no-ENS results.
- [ ] Check whether `useAsync` can invoke its async callback multiple times concurrently (e.g., on rapid re-renders before the first call resolves), which could cause a race condition writing `cache[chainId][address]` twice with potentially different values.
- [ ] Ensure the new unit test does not rely on implicit module-level cache state leaking between test cases — confirm `jest.resetModules()` or equivalent isolation is not needed to keep the test deterministic.
- [ ] Manually test that an address which gains an ENS name mid-session (or where the first lookup was a transient failure) does not serve a stale `null` from cache, and decide if a cache TTL or invalidation mechanism is needed.

## File-by-file

### `apps/web/src/hooks/__tests__/useAddressResolver.test.ts`

A new test `caches no-ENS results so repeated lookups do not hit the provider` is added that mounts `useAddressResolver` twice for an address where `lookupAddress` resolves to `undefined`, asserting `lookupMock` is called exactly once across both mounts.

_Verifies the negative-result caching behavior introduced in the implementation change, ensuring the provider is not called redundantly for addresses with no ENS name._

**Risks:**
- Test relies on module-level cache state persisting between the two `renderHook` calls — if test isolation resets the cache (e.g., via jest module reset), the assertion `toHaveBeenCalledTimes(1)` could give a false positive
- No assertion that `resolving` is `false` after the first mount completes — only checked on second mount; a race could mask bugs
- Does not verify behavior after a cache invalidation or page-level reset scenario

### `apps/web/src/hooks/useAddressResolver.ts`

Changed `cache` value type from `Record<string, string>` to `Record<string, string | null>`, replaced the `useEffect`-based cache write (positive results only) with an inline `await`+write inside `useAsync` that stores `null` for no-ENS results, and changed the cache-hit check from truthiness to `in` operator key-presence.

_Caches negative ENS lookup results so repeated mounts for addresses with no ENS name skip the provider call instead of hitting Infura every time._

**Risks:**
- Transient RPC errors in lookupAddress return undefined and are now cached as null, permanently suppressing retries for the session
- useAsync dependency array includes ethersProvider — if provider reference changes, cache is bypassed and a fresh lookup fires even for already-cached addresses
- Cache is module-level and never invalidated; if an address gains an ENS name mid-session the stale null will be served until page reload
- The async function inside useAsync now writes to the module-level cache as a side effect, which could cause issues if useAsync calls the function multiple times concurrently for the same address (race condition writing cache[chainId][debouncedValue] twice)

---
_Generated by prbrief • model: claude-sonnet-4-6 • 4 LLM calls • $0.09 • 72.5s_

</details>

<sub>Generated by <a href="https://popcorn.report">Popcorn Review</a>. · `safe-global/safe-wallet-monorepo@73621093d9ba`</sub>

ChillYao:
Hey @katspaugh ,

I rebased your branch on epic branch, and tested it in preview link, locally and compared with the prod.

- On preview link:

https://github.com/user-attachments/assets/ddd2f899-da1a-41bf-a59e-0573e085dafd

- Locally:

https://github.com/user-attachments/assets/33a2a356-23a9-4d58-bdf6-53e23d40f2a8

- Prod:

https://github.com/user-attachments/assets/54abe2f5-e56e-4bc4-8e3b-d755e8e1fafc

What I found:
1. When I navigate to different safes, it still calls infura many times. It sometimes cause 429 error locally. I am not sure if this will cause some issue after we push it to prod.
2. But I also found when we navigate to safes on prod, it also calls infura many times.

We would like to discuss with you and hear your opinion.

PR #7764, opened 2026-04-27
Display untrusted Safe warning in RiskConfirmation for signed messages and update tests

devpramoth:
### Motivation
- Clarify the risk confirmation copy when the user is signing a message from an untrusted Safe so the user knows the Safe itself is untrusted.
- Ensure the UI and tests reflect the SafeShield analysis state for message signing flows.

### Description
- Read `safeAnalysis` from `useSafeShield` and import `SafeStatus` to detect untrusted Safe messages in `RiskConfirmation`.
- Add `isUntrustedSafeMessage` and render a specific label: "I understand this Safe is untrusted and would like to proceed with this message." when signing an untrusted Safe message, otherwise keep the original transaction/message wording.
- Update `SignMessage.test.tsx` to include `SafeStatus` and `Severity`, provide a mocked `safeAnalysis` with `type: SafeStatus.UNTRUSTED`, and assert the new untrusted Safe confirmation text is rendered.

### Testing
- Ran the updated unit tests for the message signing flow using Jest via `yarn test` and the `SignMessage` spec, and the updated tests passed.
- Verified the `RiskConfirmation` rendering test asserting the untrusted Safe message copy succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_69ef63ad7698832aafbf84e32356e851)

### DIFF
--- a/apps/web/src/components/tx-flow/features/RiskConfirmation.tsx
+++ b/apps/web/src/components/tx-flow/features/RiskConfirmation.tsx
@@ -6,13 +6,15 @@ import Track from '@/components/common/Track'
 import { MODALS_EVENTS } from '@/services/analytics'
 import { SafeTxContext } from '../SafeTxProvider'
 import { useSafeShield } from '@/features/safe-shield/SafeShieldContext'
+import { SafeStatus } from '@safe-global/utils/features/safe-shield/types'
 
 export const RiskConfirmation = () => {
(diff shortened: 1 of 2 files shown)


Zalens:
Closing as [WA-1994](https://linear.app/safe-global/issue/WA-1994/warning-for-untrusted-wallet-missing-for-message-signing) was canceled, so this fix is no longer planned. Can be reopened if the ticket is revived.

PR #7779, opened 2026-04-29
fix(web): drive topbar safe selector from URL not Redux

ChillYao:
> Read the URL, not Redux.
> Inject by URL, not Redux.
> No more snap to items[0].

## What it solves

Resolves: [WA-2192](https://linear.app/safe-global/issue/WA-2192/safe-from-url-loads-wrong-safe-in-navigation)

The topbar `SafeSelectorDropdown` was showing the **wrong safe** in two scenarios:

- **Case 1** — Switching safes via the dropdown briefly displayed the first safe in the list before settling on the clicked one.
- **Case 2** — Opening a nested safe (parent in space, child outside it) showed the first safe in the list **persistently**, with a 404 in the console. URL was correct but the navigation contradicted it.

## How this PR fixes it

The dropdown's "selected" state was derived from `useSafeInfo()` (Redux), which lags the URL by a CGW round-trip — and on a 404 (nested safes) the slice never updates. With `selectedItemId` constructed from a stale Redux address, `items.find(...)` missed and the dropdown silently fell back to `items[0]`.

Three coordinated changes:

1. **`useSpaceSafeSelectorItems.ts`** — `selectedItemId` and `isCurrentSafe` now read `useSafeAddressFromUrl()`. The URL match is explicit, and for multi-chain safes `orderedChainIds[0]` correctly puts the URL chain first so `item.id` lines up with `selectedItemId`.

2. **`useSafeBarSafes.ts`** — `safeAddress` (driving `fallbackCurrentSafe` and the dropdown injection) reads from the URL instead of Redux. The URL safe is always present in `items` — even when CGW returns 404 — so the `find()` succeeds.

3. **`useSafeSelectorState.ts`** — `getSelectedItem` no longer falls back to `items[0]` on a miss. Returns `undefined` so any future regression surfaces as "no selection" instead of silently displaying the wrong safe.

`useIsQualifiedSafe()` was already URL-driven, so when navigating from a space-qualified parent to a nested child outside the space, `isInSpaceContext` flips to `false` immediately and the non-space fallback path takes over. No `spaceSafesWithFallback`-style hack was needed.

## How to test it

Use the staging deployment for the Linear ticket: `https://epicspacesm2--walletweb.review.5afe.dev/`

**Case 1 — Switch between safes**
1. Open the app, sign in to a space with multiple safes
2. Click the dropdown, scroll to a safe you haven't visited yet (away from the top)
3. Select it
4. ✅ Topbar updates instantly to the selected safe — no flash of the first safe

**Case 2 — Nested safe**
1. Navigate to a safe with nested safes (e.g. `?safe=sep:0x2a73e61bd15b25B6958b4DA3bfc759ca4db249b9`)
2. Click the **Nested Safes** icon (GitMerge) in the topbar
3. Pick a nested child safe from the popover
4. URL changes to the child address; console may still show a CGW 404 (separate backend concern)
5. ✅ Topbar dropdown displays the child safe (placeholder name / zero balance is OK), **not** the first safe in the list

**Direct URL test (original repro)**
1. Open `https://epicspacesm2--walletweb.review.5afe.dev/home?safe=sep:0xC4934eA242Bf292CA59A2aF85ED116506F8f2d03` from a fresh tab
2. ✅ Topbar shows the safe from the URL, not whichever safe happens to be first in your pinned/space list

**Regression check**
1. Open the dropdown on a safe-less route (e.g. `/welcome/accounts`)
2. ✅ Topbar/SpaceSafeBar is hidden (existing `HIDDEN_ROUTES` behavior, unchanged)

## Affected flows

- Topbar safe selector dropdown — switching to any pinned, owned, or URL-only safe
- Nested safe navigation via the GitMerge popover in the topbar
- Multi-chain safe identity in the dropdown (URL chain becomes the leading chain on selection)
- Topbar chain selector and "Nested Safes" button (indirect — they co-render in `SpaceSafeBar`)

## Blast radius

**Hooks changed**
- `useSafeBarSafes` — consumed by `SpaceSafeBar` (topbar). Returns `dropdownSafes` / `chainSelectorSafes` to `useSpaceSafeSelectorItems`, `SpaceChainSelector`, and `AccountsModal`. Switched dependency from `useSafeInfo` (Redux) to `useSafeAddressFromUrl` (URL).
- `useSpaceSafeSelectorItems` — consumed by `SpaceSafeBar`. `selectedItemId` shape unchanged (`${chainId}:${address}`); only its source flipped to URL.
- `useSafeSelectorState` — consumed by `SafeSelectorDropdown` (the shadcn dropdown wrapper). `selectedItem` may now legitimately be `undefined`; the component already handles that by rendering an `InlineRetryError`, `SafeSelectorDropdownSkeleton`, or `null`.

**Not touched**
- RTK Query endpoints / CGW contracts — no API changes
- Feature flags / chain configs — none added or modified
- Persistence / Redux schema — no migration; `safeInfoSlice` shape unchanged, just no longer the source of truth for the selector
- Routes / layouts — no changes
- Mobile / shared packages — web-only feature; `packages/` untouched

## Risks / not checked

- **CGW 404 root cause for nested safes** — out of scope. This PR ensures the UI no longer misleads when the API 404s, but the underlying backend issue (why `/safes/{chainId}/{nestedAddr}` returns 404) is not addressed.
- **Threshold / owners briefly stale** — for the URL-matched item, `resolveThresholdAndOwners` still pulls from Redux when `isCurrentSafe = true`. During the ~1s Redux-lag window the trigger may show the previous safe's threshold/owners. Visually subtle, accepted as a trade-off vs. wrong safe identity.
- **`SpaceChainSelector` in nested-safe scenario** — when the URL safe is unknown, the chain selector falls back to a single chain (the URL chain). Functional but doesn't expose multi-chain options for the unknown safe. Same behavior as a freshly visited URL safe pre-fix.
- **Mobile** — not tested. Files changed are web-only (`apps/web/src/...`); no shared-package code touched, so no mobile impact expected.
- **Redux address-format mismatch** — `item.id === selectedItemId` is exact string equality. If a future change causes `useSafeAddressFromUrl` and `SafeItem.address` to use different formats (case, checksum), the match fails and the dropdown shows nothing (loud failure, by design — Fix 3).

## Visual summary

```mermaid
sequenceDiagram
    autonumber
    participant U as User
    participant URL as URL
    participant H as Topbar hooks
    participant R as Redux / CGW
    participant D as Dropdown

    rect rgba(255, 200, 200, 0.3)
    Note over U,D: Before — case 2 (nested safe, CGW 404)
    U->>URL: click nested → ?safe=sep:0xChild
    URL-->>H: useSafeAddressFromUrl = 0xChild
    URL-->>R: useLoadSafeInfo fetch → 404
    R-->>R: safeInfo data: undefined
    H->>D: selectedItemId = "sep:" (Redux empty)
    D->>D: items.find(...) = undefined
    D->>D: ?? items[0] (silent fallback)
    D-->>U: ❌ first safe in list shown
    end

    rect rgba(200, 255, 200, 0.3)
    Note over U,D: After
    U->>URL: click nested → ?safe=sep:0xChild
    URL-->>H: useSafeAddressFromUrl = 0xChild
    H->>H: fallbackCurrentSafe injected into items
    H->>D: selectedItemId = "sep:0xChild" (URL)
    D->>D: items.find(...) → match ✓
    D-->>U: ✅ child safe shown
    end
```

## Checklist

- [x] I've documented how it affects the analytics (if at all) 📊 — no analytics changes
- [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻 — 4 new regression tests, 133/133 passing

---

## CLA signature

### DIFF
--- a/apps/web/src/components/common/SpaceSafeBar/hooks/__tests__/useSafeBarSafes.test.ts
+++ b/apps/web/src/components/common/SpaceSafeBar/hooks/__tests__/useSafeBarSafes.test.ts
@@ -15,9 +15,8 @@ jest.mock('@/features/spaces', () => ({
 }))
 
 const mockSafeAddress = jest.fn(() => '0xCurrentSafe')
-jest.mock('@/hooks/useSafeInfo', () => ({
-  __esModule: true,
-  default: () => ({ safeAddress: mockSafeAddress() }),
+jest.mock('@/hooks/useSafeAddressFromUrl', () => ({
+  useSafeAddressFromUrl: () => mockSafeAddress(),
 }))
 
 const mockChainId = jest.fn(() => '1')
@@ -73,10 +72,11 @@ describe('useSafeBarSafes', () => {
     expect(result.current.dropdownSafes[0].address).toBe('0xCurrentSafe')
   })
 
-  it('returns space safes when in space context', () => {
+  it('returns space safes when in space context and the URL safe is in the space', () => {
     mockUseIsQualifiedSafe.mockReturnValue(true)
     const spaceSafe = createSafe('0xSpaceSafe', true)
     mockUseSpaceSafes.mockReturnValue({ allSafes: [spaceSafe] })
+    mockSafeAddress.mockReturnValue('0xSpaceSafe')
 
     const { result } = renderHook(() => useSafeBarSafes())
 
@@ -89,6 +89,7 @@ describe('useSafeBarSafes', () => {
     mockIsSpaceRoute.mockReturnValue(true)
     const spaceSafe = createSafe('0xSpaceSafe', true)
     mockUseSpaceSafes.mockReturnValue({ allSafes: [spaceSafe] })
+    mockSafeAddress.mockReturnValue('0xSpaceSafe')
 
     const { result } = renderHook(() => useSafeBarSafes())
 
@@ -149,6 +150,21 @@ describe('useSafeBarSafes', () => {
     expect(fallback.chainId).toBe('11155111')
     expect(fallback.isReadOnly).toBe(true)
     expect(fallback.isPinned).toBe(false)
+  })
+
+  it('keeps URL fallback at index 0 even when other pinned safes exist', () => {
+    const pinnedA = createSafe('0xPinnedA', true)
+    const pinnedB = createSafe('0xPinnedB', true)
+    mockAllSafes.mockReturnValue([pinnedA, pinnedB])
+    mockSafeAddress.mockReturnValue('0xNestedChild')
+    mockChainId.mockReturnValue('11155111')
+
+    const { result } = renderHook(() => useSafeBarSafes())
+
+    expect(result.current.dropdownSafes).toHaveLength(3)
+    expect(result.current.dropdownSafes[0].address).toBe('0xNestedChild')
+    expect(result.current.dropdownSafes[1].address).toBe('0xPinnedA')
+    expect(result.current.dropdownSafes[2].address).toBe('0xPinnedB')
   })
 
   it('injects fallback into chainSelectorSafes when not in allKnownSafes', () => {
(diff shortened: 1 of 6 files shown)


clovisdasilvaneto:
@ChillYao Sometimes the safe is not on url, but on redux... maybe we should fallback from the redux if there is no url present?

clovisdasilvaneto (approved):

PR #7781, opened 2026-04-29
feat(web): disable space chain selector during tx flow

clovisdasilvaneto:
> Lock the chain switch
> while a tx modal is open,
> tooltip tells you why.

## What it solves

The Space chain selector in the top bar remained interactive while a transaction flow modal was open. Switching networks mid-flow can put the modal in an inconsistent state.

## How this PR fixes it

- Reads `txFlow` from `TxModalContext` inside `SpaceChainSelector` to know when a tx modal is open.
- Passes a new `disabled` prop to `ChainSelectorBlock`. When set, the trigger:
  - Renders with `opacity-50` and `cursor-not-allowed` (and drops hover/focus-ring affordances).
  - Disables the underlying base-ui `MenuTrigger` and gates `onOpenChange` so the dropdown cannot be opened by mouse or keyboard.
  - Sets `tabIndex={-1}` and `aria-disabled` for accessibility.
- Wraps the disabled `ChainSelectorBlock` with a Tooltip that displays: "Changing the network is not allowed in this screen".

## How to test it

- [ ] Open a Safe with multiple deployed networks; confirm the chain selector still works normally outside a tx flow.
- [ ] Open any transaction flow (e.g. Send tokens). The chain selector should appear faded with a not-allowed cursor.
- [ ] Hover the chain selector while the tx modal is open — the tooltip should read "Changing the network is not allowed in this screen".
- [ ] Try clicking and pressing Enter/Space on the selector while disabled — the dropdown must not open.
- [ ] Close the tx flow — the selector returns to normal interactive state.

## Visual summary

```mermaid
flowchart LR
  A[SpaceChainSelector] --> B{txFlow open?}
  B -->|yes| C[Tooltip wrapper]
  C --> D[ChainSelectorBlock disabled]
  D --> E["Trigger faded, cursor-not-allowed, MenuTrigger disabled"]
  B -->|no| F[ChainSelectorBlock interactive]
```

### Regression checklist

**Primary flow changed:** Disable Space chain selector while a tx flow modal is open and surface a tooltip.

**Surfaces touched:**
- `SpaceChainSelector` (consumes `TxModalContext`)
- `ChainSelectorBlock` (new `disabled` prop)

**Neighbouring flows to verify:**
- Normal chain switching from the Space top bar — must still work outside tx flows.
- Add network flow (`CreateSafeOnNewChain`) — unaffected; only the trigger is gated.
- Keyboard a11y — disabled trigger is non-focusable (`tabIndex={-1}`).

**Tests to add/run:**
- type-check (PASS), prettier (PASS), unit tests for related files (no related tests found).

**Not verified (risks):**
- No automated test coverage for the disabled state — verified manually via the Test plan above.

### DIFF
--- a/apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.tsx
+++ b/apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.tsx
@@ -1,7 +1,9 @@
-import { useState, useCallback } from 'react'
+import { useContext, useState, useCallback } from 'react'
 import { Skeleton } from '@/components/ui/skeleton'
+import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip'
 import ChainSelectorBlock from '@/features/spaces/components/SafeSelectorDropdown/components/ChainSelectorBlock'
 import { CreateSafeOnNewChain } from '@/features/multichain'
+import { TxModalContext } from '@/components/tx-flow'
 import { useSpaceChainSelector } from './hooks/useSpaceChainSelector'
 
 function SpaceChainSelectorSkeleton() {
@@ -15,6 +17,8 @@ function SpaceChainSelectorSkeleton() {
 function SpaceChainSelector({ isLoading }: { isLoading?: boolean }) {
   const { deployedChains, selectedChainId, deployedChainIds, safeAddress, safeName, handleChainChange } =
     useSpaceChainSelector()
+  const { txFlow } = useContext(TxModalContext)
+  const isDisabled = !!txFlow
 
   const [addNetworkChainId, setAddNetworkChainId] = useState<string>()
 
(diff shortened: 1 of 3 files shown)


apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.tsx
```diff
@@ -36,14 +40,31 @@ function SpaceChainSelector({ isLoading }: { isLoading?: boolean }) {
...
     >
-      <ChainSelectorBlock
-        deployedChains={deployedChains}
-        selectedChainId={selectedChainId}
-        safeAddress={safeAddress}
-        deployedChainIds={deployedChainIds}
-        onChainSelect={handleChainChange}
-        onAddNetwork={handleAddNetwork}
-      />
+      {isDisabled ? (
```
ChillYao, line 43: Do we need at test for this case?

clovisdasilvaneto: sounds good, will create one for it 👍

ChillYao (approved):

PR #7788, opened 2026-04-30
fix(mobile): stop eager api.web3modal.org requests on launch

compojoom:
> Reown wakes at boot,
> fetching wallets we don't need —
> mount the modal late.

## What it solves

On every mobile cold start the Reown AppKit SDK fired requests to `api.web3modal.org` (`/getWallets`, `/getNetworkImages`, `/getAnalyticsConfig`, plus `INITIALIZE` / `MODAL_LOADED` track events) before the user had taken any wallet-connect action. This violates our opt-in privacy posture and wastes network/battery on every launch.

Resolves: https://linear.app/safe-global/issue/WA-2195/investigate-eager-apiweb3modalorg-requests-on-mobile-app-launch

## How this PR fixes it

Two independent levers:

1. Pass `enableAnalytics: false` to `createAppKit`. The SDK was leaving this `undefined`, which causes it to fetch analytics config and emit `INITIALIZE` / `MODAL_LOADED` track events.
2. Lazy-mount the `<AppKit />` modal component. The eager `ApiController.prefetch()` (wallet list, network images) runs from `<AppKit />`'s mount-time `useEffect`, *not* from `createAppKit()` itself. We keep `AppKitProvider` mounted from boot so signing and session restoration still work, but defer mounting the modal until anything sets `ModalController.state.open` to true.

Gotchas:

- We watch `ModalController` via the public `useAppKitState()` hook rather than wrapping our own `open()`. Internal hooks like `useConnect` call `useAppKit().open()` directly, so a wrapper would have been bypassed (the picker silently no-oped on the first attempt — that's how I caught it).
- `ModalController.open()` does `await ApiController.state.prefetchPromise` before flipping state. While the modal is unmounted, `prefetchPromise` is `undefined`, so `await undefined` resolves immediately — no missed-event window.
- Reown's `createAppKit` is a global singleton (`Symbol.for('__REOWN_APPKIT_INSTANCE__')`), so this whole approach has to coexist with the existing one-shot init pattern in `appKit.ts`. It does.

## How to test it

1. Cold-launch the mobile app with the network panel open. Confirm zero requests to `api.web3modal.org` until you tap "Connect wallet".
2. Tap "Connect wallet" in the signers / import-signer flow. Modal should open and show the wallet picker as before.
3. With an existing WalletConnect signer already imported, sign a transaction. Verify it works without ever opening the modal (exercises `useProvider` / `EthersAdapter` while the modal is unmounted).
4. Reconnect an existing WC signer (modal-based reconnect flow) and confirm it still works.

## Screenshots

N/A — no UI changes.

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/WalletConnect/appKit.ts
+++ b/apps/mobile/src/features/WalletConnect/appKit.ts
@@ -87,6 +87,9 @@ export function createAppKitInstance(networks: [Network, ...Network[]], defaultN
     themeVariables: {
       accent: '#12FF80',
     },
+    // Reown SDK fires telemetry to api.web3modal.org when this is undefined.
+    // We have our own opt-in analytics; never send data to Reown.
+    enableAnalytics: false,
     debug: __DEV__,
   })
 }
--- a/apps/mobile/src/features/WalletConnect/context/WalletConnectContext.tsx
+++ b/apps/mobile/src/features/WalletConnect/context/WalletConnectContext.tsx
@@ -1,5 +1,13 @@
 import React, { createContext, useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react'
-import { AppKit, AppKitProvider, useAccount, useAppKit, useProvider, useWalletInfo } from '@reown/appkit-react-native'
+import {
+  AppKit,
+  AppKitProvider,
+  useAccount,
+  useAppKit,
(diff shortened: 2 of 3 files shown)


vmosharova (approved):
👍🏻

PR #7800, opened 2026-05-04
hotfix(web/Spaces): add missing data-testId for multichain tests; fix hydration errors

vmosharova:
> Hydrate-gate the chain bar,
> add testids for accordion,
> wait, then click — green tests.

## What it solves

E2E test flake on the multichain Add Network flow + a hydration mismatch in `SpaceChainSelector`.

## How this PR fixes it

- `SpaceChainSelector`: render the skeleton until `useEffect` mounts, so server and first client render match (root cause: `useSafeAddressFromUrl` reads `location.search`).
- `AllNetworksSection`: add `data-testid` for the accordion and add-network buttons.
- Cypress `network.pages.js`: wait for the chain nav button before clicking; assert `modal-add-network-btn` is enabled before submit.

## How to test it

1. Open a Safe page — chain selector renders without a hydration warning in the console.
2. Run `yarn workspace @safe-global/web cypress:run --spec '**/regression/add_network*'` (or the relevant smoke spec) — passes consistently.

## Affected flows

- Multichain "Add network" from the safe-bar chain selector.
- Initial render of the safe-bar on any Safe route.

## Blast radius

- `SpaceChainSelector` is rendered on every Safe route via the safe-bar — first paint now shows a skeleton instead of nothing.
- `AllNetworksSection` testids are additive; no behavior change.
- Page-object changes are scoped to `network.pages.js`.

## Risks / not checked

- Did not fix the underlying SSR/CSR divergence in `useSafeAddressFromUrl` — only masked it at this consumer.
- Did not verify on mobile.
- Did not run the full E2E suite, only the network-related specs.

## Visual summary

​```mermaid
flowchart LR
  A[SSR render] --> S[Skeleton]
  B[Client render #1] --> S
  S --> C[useEffect: isHydrated=true]
  C --> D[Real ChainSelector]
​```

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.tsx
+++ b/apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.tsx
@@ -1,4 +1,4 @@
-import { useContext, useState, useCallback } from 'react'
+import { useContext, useEffect, useState, useCallback } from 'react'
 import { Skeleton } from '@/components/ui/skeleton'
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip'
 import ChainSelectorBlock from '@/features/spaces/components/SafeSelectorDropdown/components/ChainSelectorBlock'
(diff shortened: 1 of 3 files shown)


apps/web/src/components/common/SpaceSafeBar/SpaceChainSelector.tsx
```diff
@@ -22,6 +22,11 @@ function SpaceChainSelector({ isLoading }: { isLoading?: boolean }) {
   const isDisabled = !!txFlow
 
   const [addNetworkChainId, setAddNetworkChainId] = useState<string>()
+  const [isHydrated, setIsHydrated] = useState(false)
+
+  useEffect(() => {
```
ChillYao, line 27: two questions:
Why do we need this logic? 
If we need it, should we add tests for it?

ChillYao (approved):

PR #7816, opened 2026-05-06
fix: disable Create Space btn when N Spaces >=10

vmosharova:
## What it solves

Disables Create Space btn when N spaces >=10
Resolves: [WA-1921](https://linear.app/safe-global/issue/WA-1921/disable-space-creation-after-reaching-10-space-limit)

## Checklist

---

## CLA signature


ChillYao (approved):

apps/web/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@safe-global/web",
   "homepage": "https://github.com/safe-global/safe-wallet-web",
   "license": "GPL-3.0",
-  "version": "1.88.0",
+  "version": "1.89.0",
```
katspaugh: This should be reverted, right?

katspaugh (dismissed):
Approved, just don't forget to revert the version bump.

apps/web/package.json
vmosharova: Very good point! Updated.

katspaugh (approved):

PR #7819, opened 2026-05-06
fix(mobile): surface WalletConnect connect errors without paging on expiry

tmjssz:
> WC errors now live in `useConnect`,
> proposal expired silently retries,
> consumers do `if (\!result) return`.

## What it solves

`@reown/appkit-react-native`'s `CONNECT_ERROR` was being surfaced as a generic `Logger.error('Error during signer import', ...)` (and equivalent for reconnect) every time the QR-pairing 5-min timeout fired — paging on-call for what is a routine "user didn't scan in time" event. The user also got no signal that anything had gone wrong: no alert, and any half-formed pairing was left behind because only the modal UI was dismissed. On top of that, the catch sites duplicated `instanceof` ladders for every WalletConnect error class, and the cleanup `disconnect()` was un-`await`ed despite returning a `Promise` at runtime — an unhandled-rejection of exactly the class this PR was supposed to suppress.

Resolves: https://linear.app/safe-global/issue/WA-2198

## How this PR fixes it

All known WalletConnect failure modes are now handled inside `useConnect` itself. The hook returns `Promise<ConnectResult | null>`:

- **Success** → resolves with `{ address, walletName, walletIcon }`.
- **User rejected** → `Logger.info` + `close()` + resolves `null`.
- **Unsupported chain** → `Alert.alert('Unsupported network', …)` + `disconnect()` + resolves `null`.
- **Proposal expired** → `Logger.warn` + `disconnect()` + `open({ view: 'Connect' })`. `pendingRef` stays set so the same promise resolves once the user actually scans. Transparent retry, no consumer involvement.
- **Catastrophic / unrecognised `CONNECT_ERROR`** → rejects with `ConnectError`.

Consumers become trivial:

```ts
const connect = useConnect({ flow: 'signer import' })

try {
  const result = await connect()
  if (\!result) return                       // useConnect handled cancel/unsupported
  const { address, walletName, walletIcon } = result
  // happy path
} catch (error) {
  // Fallback for unrecognised errors only — log + IIFE-disconnect + close + Alert.
}
```

A few non-obvious decisions:

- **`disconnect()` runs inside an awaited async IIFE.** AppKit's `useAppKit` hook narrows the `disconnect` return type to `() => void`, but the implementation returns the underlying `Promise<void>`. A bare `try { disconnect() } catch {}` only catches synchronous throws — async rejections become unhandled. The IIFE awaits inside the try so cleanup errors land in `Logger.warn` instead of escaping.
- **`close()` runs before `Alert.alert(...)`** in the fallback path. On iOS the alert can otherwise be hosted by the dismissing AppKit modal and disappear with it. Both flows have a unit test pinning `close → alert` invocation order.
- **`isProposalExpiredMessage` uses `\bproposal\s+expired\b`.** Word boundaries are tight enough to skip substring matches inside other words but lax enough to match the canonical SignClient string and the wrapped form `"Pairing already exists: Proposal expired"`.
- **No retry counter on proposal-expired.** Each retry hands the user another 5 minutes; if they keep walking away, we keep regenerating QRs. Bounded retries felt like overengineering until we see the failure mode in the wild.

Behaviour summary by failure class:

| Trigger | Log | Cleanup | Modal | Promise |
|---|---|---|---|---|
| Success | — | — | (auto) | resolves with result |
| User rejected | `Logger.info` | — | `close()` | resolves `null` |
| Unsupported chain | `Logger.warn` (on switch failure) | `disconnect()` | (auto) + alert | resolves `null` |
| Proposal expired | `Logger.warn` | `disconnect()` | reopen | stays pending |
| Catastrophic `ConnectError` | `Logger.error` (in flow) | `disconnect()` | `close()` + alert | rejects |

## How to test it

1. **Proposal expiry (transparent retry)**: open the import-signer flow and scan a WC QR with a wallet, then wait >5 min before approving on the wallet side. Confirm: the appkit modal refreshes silently (no alert, no error toast) and you can complete the pairing on the next scan. `Logger.warn` records the event in the dev console.
2. **User rejection**: open the import-signer flow and reject the connection in the wallet. Confirm: modal closes cleanly, no alert.
3. **Unsupported chain**: connect a wallet that doesn't support the active Safe's chain. Confirm: "Unsupported network" alert renders, modal stays out of the way, no orphaned session.
4. **Catastrophic error**: trigger a non-expiry connect error (e.g. force-quit the wallet mid-handshake). Confirm: the alert renders cleanly *after* the AppKit sheet dismisses (not behind it), and a fresh QR scan works on retry without "already paired" errors.
5. Repeat (1)–(4) on the reconnect flow (existing signer, signers tab → "Reconnect").

## Visual summary

```mermaid
flowchart TB
    A[connect] --> B{event}
    B -->|CONNECT_SUCCESS<br/>right chain| R[resolve&#40;result&#41;]
    B -->|CONNECT_SUCCESS<br/>wrong chain| S[switchNetwork]
    S -->|ok + chain settles| R
    S -->|throws or timeout| U[disconnect + showUnsupportedChainAlert]
    U --> N[resolve&#40;null&#41;]
    B -->|USER_REJECTED| L1[Logger.info + close]
    L1 --> N
    B -->|CONNECT_ERROR<br/>proposal expired| W[Logger.warn + disconnect + open]
    W -.transparent retry.-> A
    B -->|CONNECT_ERROR<br/>other| RJ[reject ConnectError]
    RJ --> F[caller: Logger.error<br/>+ disconnect + close + alert]
```

## Screenshots

N/A — no UI changes beyond:
- The new "Unsupported network" / generic-error alerts, both stock RN `Alert.alert`.
- Proposal-expired no longer surfaces a user-visible alert at all (transparent retry).

## Checklist

Tests added/updated:

`useConnect.test.ts`:
- AppKit message preserved through `ConnectError` (catastrophic case)
- Proposal-expired transparently reopens the modal (asserts second `open({view:'Connect'})` call, no alert, promise stays pending)
- Wrapped proposal-expired (`"Pairing already exists: Proposal expired"`) also reopens
- Proposal-expired post-unmount does not reopen (no orphan `open` calls)
- `isProposalExpiredMessage` matrix: wrapped form, whitespace tolerance, word-boundary negation
- Generic `CONNECT_ERROR` falls back to `'Connection failed'`
- `USER_REJECTED` resolves `null` + closes modal + `Logger.info` + no alert
- Unsupported-chain (switchNetwork throws or times out) resolves `null` + disconnects + shows alert
- Sequential `connect` calls reuse the same hook correctly
- Pending promise cleared on unmount

Flow tests (`useImportSignerFlow.test.ts`, `useReconnectFlow.test.ts`):
- Short-circuits silently when `connect` resolves `null` (covers all cancel/unsupported paths in one assertion since the flow can't tell them apart)
- Fallback alert path (with `close → alert` order assertion)
- Async rejection from `disconnect()` during fallback cleanup is captured to `Logger.warn`, doesn't drop `close`/`Alert`
- Validation errors during the happy path route through the fallback alert
- Collision-guard path: alerts on duplicate signer, captures `disconnect` async rejection

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/WalletConnect/hooks/__tests__/useConnect.test.ts
+++ b/apps/mobile/src/features/WalletConnect/hooks/__tests__/useConnect.test.ts
@@ -1,14 +1,17 @@
+import { Alert } from 'react-native'
 import { renderHook, act } from '@/src/tests/test-utils'
-import {
-  useConnect,
-  ConnectError,
-  ProposalExpiredError,
-  UnsupportedChainError,
-  UserRejectedError,
-  isProposalExpiredMessage,
-} from '../useConnect'
+import { useConnect, ConnectError, isProposalExpiredMessage } from '../useConnect'
+import Logger from '@/src/utils/logger'
+
+jest.spyOn(Alert, 'alert').mockImplementation(() => undefined)
+
+jest.mock('@/src/utils/logger', () => ({
+  __esModule: true,
+  default: { error: jest.fn(), warn: jest.fn(), info: jest.fn() },
+}))
 
 const mockOpen = jest.fn()
+const mockClose = jest.fn()
 const mockDisconnect = jest.fn().mockResolvedValue(undefined)
 const mockSwitchNetwork = jest.fn().mockResolvedValue(undefined)
 
@@ -20,7 +23,12 @@ const mockWalletState = {
 }
 
 jest.mock('@reown/appkit-react-native', () => ({
-  useAppKit: () => ({ open: mockOpen, disconnect: mockDisconnect, switchNetwork: mockSwitchNetwork }),
+  useAppKit: () => ({
+    open: mockOpen,
+    close: mockClose,
+    disconnect: mockDisconnect,
+    switchNetwork: mockSwitchNetwork,
+  }),
   useAccount: () => ({
     isConnected: mockWalletState.isConnected,
     address: mockWalletState.address,
@@ -39,6 +47,9 @@ jest.mock('../useStableAppKitEvent', () => ({
     eventCallbacks[event] = callback
   },
 }))
+
+const renderUseConnect = (initialStore?: Record<string, unknown>) =>
+  renderHook(() => useConnect({ flow: 'test' }), initialStore)
 
 const setConnected = (
   address: string,
@@ -70,7 +81,7 @@ describe('useConnect', () => {
   })
 
   it('calls open with Connect view when connect is called', () => {
-    const { result } = renderHook(() => useConnect())
+    const { result } = renderUseConnect()
 
     act(() => {
       result.current()
@@ -80,7 +91,7 @@ describe('useConnect', () => {
   })
 
   it('resolves with address, walletName, and walletIcon on successful connection', async () => {
-    const { result, rerender } = renderHook(() => useConnect())
+    const { result, rerender } = renderUseConnect()
 
     let resolved: unknown
     act(() => {
@@ -103,8 +114,8 @@ describe('useConnect', () => {
     })
   })
 
-  it('rejects on CONNECT_ERROR with the AppKit message preserved', async () => {
-    const { result } = renderHook(() => useConnect())
+  it('rejects on CONNECT_ERROR with the AppKit message preserved (catastrophic case)', async () => {
+    const { result } = renderUseConnect()
 
     let rejected: Error | undefined
     act(() => {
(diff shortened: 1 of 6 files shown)


tmjssz:
@claude review

tmjssz:
Thanks @claude.

**Landed**
- Un-awaited `disconnect()` in happy-path bail-outs (`!isOwner` and `!sameAddress` branches). Real bug — same class as the rest of this PR. Wrapped both in the awaited IIFE pattern.

**Declined**
- `SWITCH_SETTLE_TIMEOUT_MS = 8000` — intentional safety net for slow wallets; will tune from telemetry.
- Regex alternative for the wrapped form — `/\bproposal\s+expired\b/i` already matches `"Pairing already exists: Proposal expired"`; covered by an existing test.
- Earlier collision-guard short-circuit — separate PR.
- "Missing switch-network timeout test" — exists in `useConnect.test.ts` (`'resolves null and alerts via timeout…'`).

compojoom (approved):

PR #7820, opened 2026-05-07
feat(web): security hub

clovisdasilvaneto:
## What it solves

Resolves: WA-1941

## How this PR fixes it
This is an extension of #7627 where @rahulrumalla has introduced the security hub feature MVP. On this PR we took his code and improved few points.

## How to test it

1. Open any Space → navigate to the Security page
2. Verify skeleton loading during auto-scan → score card appears after all Safes finish
3. Click grade filter chips → table cross-fades to matching Safes; click again to clear
4. Click a Safe row → drawer opens with score header + accordion sections
5. Expand the "N checks passing" accordion → individual check rows appear with evidence on click
6. Click a CTA link → navigates to the correct Safe's settings page
7. Click Re-scan in the score card → all rows show skeletons, scores refresh
8. Switch Safes via the selector → auto-scan re-triggers for the new set
9. Expand a multichain Safe → child rows fade in with per-chain data

```bash
yarn workspace @safe-global/web test --testPathPattern="features/security|features/spaces" --no-coverage
# 438 tests, 58 suites
```

## Checklist

- [x] I've tested the branch on mobile 📱
- [ ] I've documented how it affects the analytics (if at all) 📊
- [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻
- [ ] I've listed affected flows and blast radius, and named what I did not verify 🎯

---

## CLA signature

With the submission of this Pull Request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://safe.global/cla).

PR #7841, opened 2026-05-11
feat(security): individual safe-state scanners + registry

clovisdasilvaneto:
Implements the 11 scanners that inspect a Safe and produce ScanResults:

- accountSetup, multichainSetup, contractVersion
- modules, guard, fallbackHandler, factoryValidation
- pendingTx, recovery, transactionScanning, signerIntegrity

Plus the registry that aggregates them into the SCANNERS array consumed by the scoring engine. Each scanner ships with its own unit tests, plus a shared defaultValues regression test.

Builds on PR 2/8 (scanner infra). Feature public API lands in PR 4/8.

Part 3/8 of the security-hub split (originally PR #7820).

## Checklist

---

## CLA signature


apps/web/src/features/security/data/scanners/accountSetup.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    // Single signer = no multisig protection
+    if (ownerCount === 1) {
+      return {
+        status: 'issue',
+        severity: 'Critical',
+        score: 10,
+        evidence: [
+          { label: 'Signers', value: String(ownerCount) },
+          { label: 'Threshold', value: `${threshold} of ${ownerCount}` },
```
tmjssz: Those two lines are duplicated multiple times here, so lets extract them into a new variable.

apps/web/src/features/security/data/scanners/accountSetup.ts
```diff
@@ -0,0 +1,82 @@
...
+      return {
+        status: 'issue',
+        severity: 'Critical',
+        score: 15,
+        evidence: [
+          { label: 'Signers', value: String(ownerCount) },
+          { label: 'Threshold', value: `${threshold} of ${ownerCount}` },
+          'Any single signer can approve transactions',
+        ],
+        remediation: `Increase the threshold to at least ${Math.ceil(ownerCount / 2)} of ${ownerCount}.`,
```
tmjssz: Wouldn't this check would warn for a 1/2 Safe and request to increase the threshold to 1/2? That would be strange. So maybe lets handle that case differently

apps/web/src/features/security/data/scanners/guard.ts
```diff
@@ -0,0 +1,116 @@
...
+        lastChecked: now,
+        ctaLabelOverride: 'Learn more',
+        partner: 'hypernative',
+      }
+    }
+
+    // Tier 4: No guard — normal (low-value Safe or unsupported chain)
+    return {
+      status: 'clear',
+      severity: 'Low',
```
tmjssz, line 118: Can't we add a helper function that calculates the severity from the score value?

apps/web/src/features/security/data/scanners/fallbackHandler.ts
```diff
@@ -0,0 +1,96 @@
...
+import type { SafeVersion } from '@safe-global/types-kit'
+import type { SecurityScanner } from './types'
+import { ZERO_ADDRESS } from './constants'
+
+// Only 1.3.0+ has the CompatibilityFallbackHandler; older versions used different handler contracts
+const COMPATIBILITY_HANDLER_VERSIONS: SafeVersion[] = ['1.3.0', '1.4.1']
+
+// CoW Protocol TWAP fallback handler — local constant to avoid cross-feature import from swap
+// https://github.com/cowprotocol/composable-cow/blob/main/networks.json
+const TWAP_FALLBACK_HANDLER = '0x2f55e8b20D0B9FEFA187AA7d00B6Cbe563605bF5'
```
tmjssz: I see we have that const also in utils package: `utils -> src -> features -> swap`. Can we import it from there?
Same for TWAP_FALLBACK_HANDLER_NETWORKS

apps/web/src/features/security/data/scanners/fallbackHandler.ts
```diff
@@ -0,0 +1,96 @@
...
+        evidence: [
+          { label: 'Handler', value: fallbackHandler.name ?? `${fallbackHandler.value.slice(0, 10)}...` },
+          { label: 'Status', value: HANDLER_LABELS[match] },
+        ],
+        remediation: '',
+        lastChecked: now,
+      }
+    }
+
+    const handlerLabel = fallbackHandler.name ?? `${fallbackHandler.value.slice(0, 10)}...`
```
tmjssz: Lets move it up and also use the const in line 74

apps/web/src/features/security/data/scanners/guard.ts
```diff
@@ -0,0 +1,116 @@
...
+      return versions && Object.values(versions).some((addr) => sameAddress(addr, guardAddress))
+    })
+  } catch {
+    return false
+  }
+}
+
+const isHypernativeGuard = (name?: string | null): boolean => {
+  if (!name) return false
+  return name.toLowerCase().includes('hypernative')
```
tmjssz: Looks a bit unsafe to just check for the string "hypernative". Can we not re-use the `isHypernativeGuard` function that we already have?

apps/web/src/features/security/data/scanners/modules.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  return false
+}
+
+/**
+ * Check if a module address matches a known Safe Allowance Module deployment.
+ */
+const isKnownAllowanceModule = (chainId: string, moduleAddress: string): boolean => {
+  for (const version of ['0.1.0', '0.1.1']) {
```
tmjssz: Lets make the array a constant better

apps/web/src/features/security/data/scanners/modules.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    // Tier 4: All untrusted, or more than 1 untrusted
+    return {
+      status: 'issue',
+      severity: 'High',
+      score: 20,
+      evidence: [
+        ...trusted.map((m) => ({ label: 'Trusted module', value: moduleLabel(m) })),
+        ...untrusted.map((m) => ({ label: 'Unverified module', value: moduleLabel(m) })),
```
tmjssz: Lets define two new variables for these lines and use them everywhere to avoid duplication

apps/web/src/features/security/data/scanners/recovery.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Detect if a module is a Zodiac Delay Modifier (recovery module) without web3 provider.
+ *
+ * Layer 1: CGW API enriches known contracts with names — check for "Delay" in name.
+ * Layer 2: Match address against known Zodiac Delay Modifier deployments per chain.
+ */
+const isDelayModifier = (chainId: string, moduleAddress: string, moduleName?: string | null): boolean => {
+  // Layer 1: API-provided name
+  if (moduleName && moduleName.toLowerCase().includes('delay')) {
```
tmjssz: Also here, the check by included string feels too general and might return false positives for similar names

tmjssz (commented):
Just a general thought for organizing all the scanner modules: Some of the files are quite full of helper functions/constants. So what do you think about having each scanner in a separate folder with it's own utils and constants files?

clovisdasilvaneto:
Thanks for the review! About the separated folders for each scanners, lets do it later, on a separated PR, so we do not messy around with the current opened ones, wdyt?

tmjssz:
@claude review

apps/web/src/features/security/data/scanners/pendingTx.ts
```diff
@@ -0,0 +1,50 @@
...
+        status: 'clear',
+        severity: 'Low',
+        score: 100,
+        evidence: [
+          {
+            label: 'Status',
+            value:
+              queuedTxCount === 0
+                ? 'No pending transactions'
+                : `${queuedTxCount} pending transaction${queuedTxCount === 1 ? '' : 's'}`,
```
tmjssz, line 20: nit: We could use `maybePlural` here

apps/web/src/features/security/data/scanners/pendingTx.ts
```diff
@@ -0,0 +1,50 @@
+import type { SecurityScanner } from './types'
+
+export const pendingTxScanner: SecurityScanner = {
+  id: 'pending_tx',
+  scan: async (ctx) => {
+    const { queuedTxCount } = ctx
+    const now = new Date().toISOString()
+
+    if (queuedTxCount <= 2) {
```
tmjssz, line 9: nit: I would extract those thresholds as constants as well for easier configuration

apps/web/src/features/security/data/scanners/constants.ts
```diff
@@ -18,3 +18,17 @@ export const SEVERITY_RANK: Record<SecurityGrade, number> = {
...
+/** Score thresholds that bucket a numeric score into a SecurityGrade. */
+export const SEVERITY_SCORE_THRESHOLDS: Array<{ minScore: number; severity: SecurityGrade }> = [
+  { minScore: 80, severity: 'Low' },
+  { minScore: 50, severity: 'Medium' },
+  { minScore: 20, severity: 'High' },
+  { minScore: 0, severity: 'Critical' },
+]
+
+/** Derives a SecurityGrade from a numeric score (0–100) via SEVERITY_SCORE_THRESHOLDS. */
+export const getSeverityFromScore = (score: number): SecurityGrade => {
```
tmjssz, line 44: Where is it used?

apps/web/src/features/security/data/scanners/accountSetup.ts
clovisdasilvaneto: Done — extracted into a `baseEvidence` constant ([accountSetup.ts:21-24](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/accountSetup.ts#L21-L24)) and reused across the three return paths that need it.

apps/web/src/features/security/data/scanners/accountSetup.ts
clovisdasilvaneto: Good catch. Floored the recommendation at 2 so a 1/2 Safe gets told to raise the threshold to 2/2 instead of 1/2: `Math.max(Math.ceil(ownerCount / 2), 2)` ([accountSetup.ts:41](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/accountSetup.ts#L41)).

apps/web/src/features/security/data/scanners/guard.ts
clovisdasilvaneto: Helper added as `getSeverityFromScore` in [constants.ts:44](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/constants.ts#L44) (driven by a single `SEVERITY_SCORE_THRESHOLDS` table) and now wired across the seven scanners where score and severity are coupled: `accountSetup`, `guard`, `fallbackHandler`, `modules`, `recovery`, `contractVersion`, `pendingTx`. The remaining four (`multichainSetup`, `factoryValidation`, `signerIntegrity`, `transactionScanning`) intentionally decouple severity from score for `partial`/`inconclusive`/`not_applicable` states — kept explicit there so the UI label doesn't get bumped to High/Medium when the user can't act on it.

apps/web/src/features/security/data/scanners/fallbackHandler.ts
clovisdasilvaneto: Imported from `@/features/swap/helpers/utils`. Went via the helpers file rather than `@/features/swap` directly because the feature index pulls the swap feature handle through `createFeatureHandle`, which creates a circular dep with `@/features/__core__` in the test environment. Comment at [fallbackHandler.ts:10-13](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/fallbackHandler.ts#L10-L13) documents the constraint.

apps/web/src/features/security/data/scanners/fallbackHandler.ts
clovisdasilvaneto: Done — `handlerLabel` is now declared once before the `identifyFallbackHandler` call and reused in both the matched and unmatched return paths ([fallbackHandler.ts:66](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/fallbackHandler.ts#L66)).

apps/web/src/features/security/data/scanners/guard.ts
clovisdasilvaneto: `isHypernativeGuard` (in `features/hypernative/services/hypernativeGuardCheck.ts`) is the on-chain check — it needs a JSON-RPC provider, which the scanner intentionally avoids (scanners run from cached `SafeState`, no web3). Two changes here instead: extracted `HYPERNATIVE_GUARD_NAME` as a single canonical constant ([guard.ts:8](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/guard.ts#L8)) so the string lives in one place, and tightened the local heuristic from `includes` to `startsWith` after trim/lowercase so we only match names that lead with the canonical prefix ([guard.ts:36-39](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/guard.ts#L36-L39)).

apps/web/src/features/security/data/scanners/modules.ts
clovisdasilvaneto: Done — extracted as `ALLOWANCE_MODULE_VERSIONS` at [modules.ts:8](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/modules.ts#L8).

apps/web/src/features/security/data/scanners/modules.ts
clovisdasilvaneto: Done — `trustedEvidence` and `untrustedEvidence` precompute the mapped arrays once and are reused in tiers 2–4 ([modules.ts:109-110](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/modules.ts#L109-L110)).

apps/web/src/features/security/data/scanners/recovery.ts
clovisdasilvaneto: Tightened from substring match to an exact, case- and whitespace-insensitive match against canonical names: `['delay', 'delay modifier', 'zodiac delay modifier']` ([recovery.ts:7, 17](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/recovery.ts#L7)). Layer 2 still does the address match against per-chain Zodiac DELAY deployments, so a renamed module is still caught.

apps/web/src/features/security/data/scanners/pendingTx.ts
clovisdasilvaneto: Done — using `maybePlural(queuedTxCount)` from `@safe-global/utils/utils/formatters` ([pendingTx.ts:24](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/pendingTx.ts#L24)).

apps/web/src/features/security/data/scanners/pendingTx.ts
clovisdasilvaneto: Done — extracted the two thresholds at the top of the file: `CLEAR_QUEUE_MAX = 2` and `LARGE_QUEUE_MIN = 5` ([pendingTx.ts:5-8](https://github.com/safe-global/safe-wallet-monorepo/blob/feat/security-hub-pt3-scanners/apps/web/src/features/security/data/scanners/pendingTx.ts#L5-L8)).

apps/web/src/features/security/data/scanners/constants.ts
clovisdasilvaneto: It was unused at the time of review — added it ahead of the wiring in this PR. Now consumed by seven scanners (see the [reply on guard.ts](https://github.com/safe-global/safe-wallet-monorepo/pull/7841#discussion_r3248333919) for the full list and which four scanners intentionally keep an explicit severity).

clovisdasilvaneto:
@tmjssz on the folder-per-scanner suggestion — fair point that some files are dense. My preference is to keep flat for now: most scanners are ≤150 lines, the shared types/constants/test-helpers already live alongside, and the registry import surface (one file per scanner + one `registry.ts`) stays trivial. If the file count or per-file helper count grows in a later PR (e.g. when `signerIntegrity` gets wired to the CGW endpoint and likely picks up its own helpers), I'd reach for the folder split then. Happy to do it now if you'd rather lock the layout in before more scanners land.

PR #7843, opened 2026-05-11
feat(security): scan + feature-redirect hooks

clovisdasilvaneto:
Adds the React hooks that consume the scoring engine from PR 4/8:

- features/security/hooks/useSecurityScan: runs SCANNERS against a Safe and exposes ScanResult[] with cache + loading metadata.
- features/security/hooks/useSecurityHubFeatureRedirect: gates the /spaces/security route behind the SECURITY_HUB flag.
- features/spaces/hooks/useAutoScan: queues per-Safe scans across a space with throttling + cancellation.
- features/spaces/hooks/useSafeScanContext: builds the per-Safe ScanContext required by the scanners (RTK Query data, chain, owners, threshold, etc).

Each hook ships with unit tests. SecurityHub container that orchestrates these lands in PR 6/8.

Part 5/8 of the security-hub split (originally PR #7820).

## Checklist

---

## CLA signature

### DIFF
--- a/apps/web/src/features/security/hooks/useSecurityScan.ts
+++ b/apps/web/src/features/security/hooks/useSecurityScan.ts
@@ -1,0 +1,152 @@
+import { useState, useCallback, useEffect, useRef } from 'react'
+import type { ScanContext, ScanResult } from '../data/scanners/types'
+import { SCANNERS } from '../data/scanners/registry'
+import { scanKey, withScannerTimeout } from '../data/scanners/utils'
+
+/**
+ * Module-level cache so multiple hook instances sharing the same ctxKey
+ * reuse results instead of running duplicate scans. TTL: 1 hour.
+ */
+const CACHE_TTL_MS = 3_600_000
+const MAX_CACHE_SIZE = 50
+const scanResultsCache = new Map<string, { results: Record<string, ScanResult>; timestamp: number }>()
+
+/** Evict the oldest entry if the cache exceeds the size limit. */
+const evictIfNeeded = () => {
+  if (scanResultsCache.size <= MAX_CACHE_SIZE) return
+  let oldestKey: string | null = null
+  let oldestTs = Infinity
+  for (const [k, v] of scanResultsCache) {
+    if (v.timestamp < oldestTs) {
+      oldestTs = v.timestamp
+      oldestKey = k
+    }
+  }
+  if (oldestKey) scanResultsCache.delete(oldestKey)
+}
+
+export const getScanResultsCache = () => scanResultsCache
+export { evictIfNeeded as evictScanCache }
+
+export type ScanState = {
+  results: Record<string, ScanResult>
+  loading: Record<string, boolean>
+  errors: Record<string, string>
+  isComplete: boolean
+  lastScannedAt: number | null
+  progress: number
+  rescan: () => void
+}
+
+const useSecurityScan = (ctx: ScanContext | null): ScanState => {
+  const ctxKey = ctx ? scanKey(ctx.safeAddress, ctx.chainId) : null
+
+  const [results, setResults] = useState<Record<string, ScanResult>>({})
+  const [loading, setLoading] = useState<Record<string, boolean>>({})
+  const [errors, setErrors] = useState<Record<string, string>>({})
+  const [lastScannedAt, setLastScannedAt] = useState<number | null>(null)
+  const scanIdRef = useRef(0)
+  const ctxRef = useRef(ctx)
+  ctxRef.current = ctx
+
+  const executeScan = useCallback(
+    async (scannerId: string, scanFn: () => Promise<ScanResult>, guardId?: number, onAllComplete?: () => void) => {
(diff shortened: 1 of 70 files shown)


apps/web/src/features/security/hooks/useSecurityScan.ts
```diff
@@ -0,0 +1,152 @@
...
+  const [results, setResults] = useState<Record<string, ScanResult>>({})
+  const [loading, setLoading] = useState<Record<string, boolean>>({})
+  const [errors, setErrors] = useState<Record<string, string>>({})
+  const [lastScannedAt, setLastScannedAt] = useState<number | null>(null)
+  const scanIdRef = useRef(0)
+  const ctxRef = useRef(ctx)
+  ctxRef.current = ctx
+
+  const executeScan = useCallback(
+    async (scannerId: string, scanFn: () => Promise<ScanResult>, guardId?: number, onAllComplete?: () => void) => {
```
tmjssz: I would make scannerId more type-safe, instead of a generic string. WDYT?

apps/web/src/features/security/hooks/useSecurityScan.ts
```diff
@@ -0,0 +1,152 @@
...
+      setLoading((prev) => ({ ...prev, [scannerId]: true }))
+
+      try {
+        const result = await scanFn()
+        if (guardId !== undefined && scanIdRef.current !== guardId) return
+        setResults((prev) => ({ ...prev, [scannerId]: result }))
+      } catch (err: any) {
```
tmjssz: Lets avoid any
```suggestion
      } catch (err: unknown) {
```

apps/web/src/features/security/hooks/useSecurityScan.ts
```diff
@@ -0,0 +1,152 @@
...
+      scanIdRef.current++
+      setResults({})
+      setErrors({})
+      setLastScannedAt(null)
+    }
+  }, [ctxKey, runScan])
+
+  const loadingCount = Object.values(loading).filter(Boolean).length
+  const total = SCANNERS.length
+  const progress = total > 0 ? Math.round(((total - loadingCount) / total) * 100) : 0
```
tmjssz: Since loadingCount is 0 intitially, the progress would be 100. Would this be better?
```suggestion
  const progress = total > 0 ? Math.round(Object.keys(results).length / total * 100) : 0
```

apps/web/src/features/security/hooks/useSecurityScan.ts
```diff
@@ -0,0 +1,152 @@
...
+import { SCANNERS } from '../data/scanners/registry'
+import { scanKey, withScannerTimeout } from '../data/scanners/utils'
+
+/**
+ * Module-level cache so multiple hook instances sharing the same ctxKey
+ * reuse results instead of running duplicate scans. TTL: 1 hour.
+ */
+const CACHE_TTL_MS = 3_600_000
+const MAX_CACHE_SIZE = 50
+const scanResultsCache = new Map<string, { results: Record<string, ScanResult>; timestamp: number }>()
```
tmjssz: `scanResultsCache` is a module-level `Map` handed out by reference, so consumers write into it directly.

Two concerns:

1. Any caller can mutate or overwrite entries. I'd suggest wrapping access: `getCachedScan(key)` / `setCachedScan(key, value)` so the map stays private.
2. Multiple scans resolve for the same `scanKey` and overwrite each other's `{results, timestamp}`, last write wins. Probably intentional, but worth either documenting or having the writer prefer the newer timestamp.

Also, I think it would be good to extract the cache logic into a separate file/module.

apps/web/src/features/security/hooks/useSecurityScan.ts
clovisdasilvaneto: Tightened: `scannerId` and `SecurityScanner.id` now use a `ScannerId` literal union, propagated through the cache, state types, and consumers.

apps/web/src/features/security/hooks/useSecurityScan.ts
clovisdasilvaneto: Done — switched to `err: unknown`.

apps/web/src/features/security/hooks/useSecurityScan.ts
clovisdasilvaneto: Fixed — initial progress is now 0. Using settled count (results + errors) over total instead of `(total - loadingCount)/total`, since `loadingCount` is 0 before any scanner has started.

apps/web/src/features/security/hooks/useSecurityScan.ts
clovisdasilvaneto: Extracted to `data/scanResultsCache.ts` with private `Map` + `getCachedScan` / `setCachedScan` accessors so the map can no longer be mutated by reference. Concurrent writes for the same key now keep the newer timestamp (slow scan never overwrites a faster, fresher one). Contract / `useAutoScan` / orchestrator updated to the new API.

compojoom (approved):

PR #7720, opened 2026-04-22
feat(authentication): show member emails in members list

vseehausen:
## What it solves

Resolves: [WA-2029](https://linear.app/safe-global/issue/WA-2029/fe-email-identity-visibility)

The Spaces members list should surface member emails returned by CGW, and the members search should be able to match those emails.

## How this PR fixes it

- adds a dedicated Email column to the members table
- keeps the name cell focused on avatar, member name, and current-user label
- extends members search to include `member.user.email`
- updates the generated Spaces gateway type with `MemberUser.email`
- adds focused component and hook coverage for email rendering and search behavior

CGW remains the source of truth for which emails are returned. The frontend renders the email values it receives for authorized space members.

## How to test it

1. Open a Space with members whose CGW member payload contains `user.email`.
2. Verify the members table shows a dedicated Email column.
3. Verify members with an email show it in that column.
4. Verify members without an email keep the Email cell blank.
5. Search by a member email and verify the matching member is shown.
6. Search by member name and verify the existing name search still works.

## Checklist

- [ ] I have tested the branch on mobile 📱 (N/A, web-only change)
- [ ] I have documented how it affects the analytics (if at all) 📊 (N/A, no analytics change)
- [x] I have written a unit/e2e test for it (if applicable) 🧑‍💻

---

## CLA signature

With the submission of this Pull Request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://safe.global/cla).

PR #7847, opened 2026-05-11
fix(spaces): truncate long member emails in members list

vseehausen:
> Long emails stretched the table —
> Tooltip on hover, ellipsis with no-wrap,
> Fixed layout keeps the columns true.

## What it solves

Follow-up to #7720 ([WA-2029](https://linear.app/safe-global/issue/WA-2029/fe-email-identity-visibility)).

After the email column was added to the Spaces members list, long email addresses (the schema accepts up to 255 chars) stretch the table beyond its container, pushing the name/role/actions columns off-screen and breaking the page layout.

## How this PR fixes it

- Wraps the email cell in a MUI `Tooltip` + `<Typography noWrap>` so long emails ellipsize while the full address remains accessible on hover.
- Adds an opt-in `fixedLayout` flag to `EnhancedTable` that switches the underlying `<Table>` to `table-layout: fixed`, so the percentage widths declared in `headCells` become real column widths instead of growable hints.
- `MembersList` opts in; no other `EnhancedTable` consumers are touched.

The CSS class lives in `EnhancedTable/styles.module.css`, matching the existing pattern used by `compact` and `mobileVariant`.

## How to test it

1. Open a Space with at least one member whose email is much longer than the column width (a ~200-char email shows the effect clearly).
2. ✅ The email cell shows the start of the address followed by an ellipsis.
3. ✅ The members table fits its container — name, role, and actions columns stay in their lanes.
4. Hover the truncated email.
5. ✅ A tooltip appears showing the full address.
6. Confirm members with no email still show an empty cell, not a stray tooltip.

## Checklist


vseehausen:
> [Open the PR briefing](https://viewer.popcorn.report/view?briefing_id=pb_b545b5825037c16696251c94)
> 
> **This PR fixes layout breakage in the Spaces members list caused by long email addresses stretching the table beyond its container.** It wraps the email cell in an MUI Tooltip with a Typography using noWrap, so long emails are truncated with an ellipsis while the full address appears on hover. To enforce the column widths declared in the table header, a new `fixedLayout` prop is added to the reusable `EnhancedTable` component, switching the underlying table to `table-layout: fixed`. The `MembersList` component opts in via this prop, and a test verifies the truncation and tooltip behavior.
> 
> 🍿 **[Open interactive briefing](https://viewer.popcorn.report/view?briefing_id=pb_b545b5825037c16696251c94)** · 📋 **Review checklist below**
> ### What's worth a closer look
> Signal 	Reviewer attention
> 🟡 Attention 	Verify that the `.fixedLayout` CSS class only applies `table-layout: fixed` and that no other styles leak or conflict, especially with existing `compact` and `mobileVariant` classes.
> 🟢 Check 	Manually test the members table with a member having no email to confirm the tooltip does not render for an empty cell.
> 🟡 Attention 	Check that the percentage-based column widths defined in headCells do not cause overflow or misalignment when fixed layout is active, particularly on narrow viewports.
> Full briefing — architecture, contracts, file map
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> Generated by [Popcorn Review](https://popcorn.report) · `safe-global/safe-wallet-monorepo@ce29af816dc5`

checked, thanks.

apps/web/src/features/spaces/components/MembersList/index.tsx
```diff
@@ -123,7 +123,13 @@ const MembersList = ({ members }: { members: MemberDto[] }) => {
         },
         email: {
           rawValue: memberEmail,
-          content: memberEmail ? <Typography variant="body2">{memberEmail}</Typography> : null,
+          content: memberEmail ? (
+            <Tooltip title={memberEmail} placement="top">
```
LucieFaire, line 127: nit: I would suggest to make it `top-start` so its more or less in the middle, because the table cell might be much bigger than the email and then its too far right.

LucieFaire (approved):

PR #7853, opened 2026-05-12
feat: require Spaces login for all users + add a "classic" view

katspaugh:
> URL holds the spaceId now,
> Redux let go of last-used,
> two kill switches gate the move.

## What it solves

Resolves: [WA-2186](https://linear.app/safe-global/issue/WA-2186/workspace-is-the-default-entry-for-all-users)

Today the "current space" is tracked in three places at once: the URL `?spaceId=`, a `lastUsedSpace` field in Redux, and the first entry of the user's spaces list. They drift, they race, and there's no single source of truth.

We also need a controlled rollout: every user should eventually be required to sign in and have at least one space, but for ~1 month the "classic" UI (no login, no space) must keep working.

## How this PR fixes it

- Two new chain-config feature flags (`DISABLE_SPACES_LOGIN`, `DISABLE_CLASSIC_UI`) looked up against `DEFAULT_CHAIN_ID` (mainnet in prod, Sepolia in staging) via a new `useHasDefaultChainFeature` helper. Both behave as **kill switches** — chain flag unset / `false` / `undefined` → feature ON; explicitly `true` → feature OFF.
- One global hook, `useSpaceIdSync`, mounted in `InitApp`. It delegates the entire 13-row behavior matrix to a pure `decide()` function and dispatches the resulting action via `router.replace`.
- Auto-injects `?spaceId=` for signed-in users on every non-excluded route. Bounces signed-out users who follow a `?spaceId` link to `/welcome/spaces?redirect=<asPath>` and brings them back after sign-in. Forces signed-in users with zero spaces to `/spaces` to onboard.
- `useSignInRedirect` now honors a same-origin `?redirect=` query (via a new `getSafeRedirectTarget` open-redirect guard).
- Sidebar variant follows pathname alone (`useIsSpaceRoute` no longer reads Redux).
- `lastUsedSpace` Redux field, action, and selector deleted; all five consumers cleaned up.
- New Cypress e2e covers the sign-in-with-redirect happy path.

### Rollout plan

| Phase             | `DISABLE_SPACES_LOGIN`   | `DISABLE_CLASSIC_UI`   |
| ----------------- | ------------------------ | ---------------------- |
| Coexistence month | unset (feature ON)       | unset (feature ON)     |
| Day-30 cutover    | unset (feature ON)       | **true** (classic OFF) |
| Later cleanup PR  | flag + branches removed  | n/a                    |

Both features default to ON when no chain config is loaded, so first-paint behavior matches the steady state for the new flow.

## How to test it

**Manual (after merge to a deploy):**
1. Open `/home` signed out → classic UI renders, no redirect.
2. Open `/home?spaceId=<id>` signed out → bounced to `/welcome/spaces?redirect=%2Fhome%3FspaceId%3D...`. Sign in → returned to `/home?spaceId=<id>`.
3. Sign in via SIWE (account with ≥1 space). Navigate to `/home` → URL becomes `/home?spaceId=<first-space>` (shallow replace, no remount).
4. Sign in with zero spaces → forced to `/spaces` onboarding.
5. Navigate to `/imprint`, `/welcome/spaces`, `/share/safe-app` (excluded routes) → no spaceId injection.
6. Set `DISABLE_CLASSIC_UI` to `true` in chain config → signed-out users on any non-excluded route are bounced to `/welcome/spaces`.
7. Set `DISABLE_SPACES_LOGIN` to `true` → `useSpaceIdSync` becomes inert, app behaves as today.

**Automated:**
- Unit tests: `yarn workspace @safe-global/web test apps/web/src/hooks/useSpaceIdSync/` (decide matrix + hook integration + open-redirect)
- Cypress: `apps/web/cypress/e2e/regression/spaces_spaceid_redirect.cy.js`

### Decision flow per page visit

```mermaid
flowchart TB
  Visit["Visit /home?spaceId=42"] --> Auth{"Signed in?"}
  Auth -- "No" --> Bounce["Redirect to /welcome/spaces?redirect=..."]
  Bounce --> SignIn["User signs in"]
  SignIn --> Return["Land back on /home?spaceId=42"]
  Auth -- "Yes, member of 42" --> Render["Render page with spaceId=42"]
  Auth -- "Yes, not member of 42" --> Overwrite["Overwrite spaceId with first owned space"]
  Auth -- "Yes, no spaces at all" --> Onboard["Force redirect to /spaces"]
```

### Source of truth: before vs. after

```mermaid
flowchart LR
  subgraph Before
    A1["components"] --> URL1["?spaceId in URL"]
    A1 --> RDX1["Redux lastUsedSpace"]
    A1 --> LIST1["spaces list, index 0"]
  end
  subgraph After
    A2["components"] --> Hook["useCurrentSpaceId"]
    Hook --> URL2["?spaceId in URL (single source of truth)"]
    Sync["useSpaceIdSync in InitApp"] --> URL2
    Sync -.->|"kill-switch flags"| FF["DISABLE_SPACES_LOGIN, DISABLE_CLASSIC_UI"]
  end
```

## Affected flows

- Sign-in (SIWE / OIDC) on `/welcome/spaces`
- All pages for signed-in users — URL gets `?spaceId` injected
- Sidebar variant switching (Safe sidebar vs. Spaces sidebar)
- Create-space onboarding (`/welcome/createSpace`, `/spaces/create-space`)
- Deep links with `?spaceId` shared between users

## Blast radius

- `useCurrentSpaceId` (read by ~20 components) — fallback chain narrowed to URL → first space (no Redux). Stories and tests where mock state set `lastUsedSpace` were cleaned.
- `AuthState` (gate for `/spaces/*` routes) — no longer dispatches `setLastUsedSpace` side effect.
- redux-persist: `lastUsedSpace` field disappears from the persisted slice shape. Existing users with a stored value will silently drop it on next persist; no migration required because nothing reads it.

## Risks / not checked

- `DEFAULT_CHAIN_ID` assumes the chains config always returns mainnet (prod) or Sepolia (staging). If a deploy ever ships without it, both flags resolve to `undefined` → kill switches stay OFF → new flow active. Desired failure mode, but worth confirming.
- Because the kill switches default to OFF, **the new flow ships live the moment this PR merges** — there is no "both flags off until ops registers them" safety net. If we want a quiet bake, ops needs to register `DISABLE_SPACES_LOGIN=true` and `DISABLE_CLASSIC_UI=true` against the default chain *before* merge, then flip them off when ready.
- One pre-existing typo unrelated to this PR: `apps/web/src/stories/mocks/defaults.ts:93,99` seeds `lastUsedSpaceId` (with `Id` suffix) into the auth slice mock. Nothing reads it; harmless; can be deleted in a separate hygiene commit.

## Docs

- Spec: `docs/superpowers/specs/2026-05-12-spaceid-url-as-source-of-truth-design.md`
- Plan: `docs/superpowers/plans/2026-05-12-spaceid-url-as-source-of-truth.md`

## Checklist

- [x] Tests added for new logic (unit + Cypress e2e)
- [x] Type-check passes (`yarn workspace @safe-global/web type-check`)
- [ ] Full `yarn verify:web` passes (lint + prettier + tests)
- [x] No `any` types introduced
- [x] Spec + plan committed alongside the code
- [x] Sentence case for any user-visible strings (none added)
- [ ] Manual browser smoke test
- [ ] Cypress e2e executed against a real browser
- [ ] Feature flags registered in chain config (post-merge ops)


katspaugh:
@claude review

apps/web/src/hooks/useSpaceIdSync/decide.ts
```diff
@@ -0,0 +1,87 @@
...
+    isOidcPending,
+    pathname,
+    asPath,
+    querySpaceId,
+    userSpaceIds,
+    spacesError,
+  } = input
+
+  // Row 1: legacy mode — both flags must be ON for anything to happen.
+  if (requireLogin === false && enabled(classicEnabled)) return NOOP
```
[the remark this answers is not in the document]
katspaugh: Addressed in 7da182379 (caught independently during a manual smoke test against staging CGW — same root cause). Row 1 now short-circuits on `requireLogin === false` alone, regardless of `classicEnabled`. Comment + test updated; added explicit coverage for the `(REQUIRE_LOGIN=OFF, CLASSIC=OFF)` combination.

apps/web/src/hooks/useSpaceIdSync/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  const router = useRouter()
+  const isSignedIn = useAppSelector(isAuthenticated)
+  const isOidcPending = useAppSelector(selectIsOidcLoginPending)
+  const requireLogin = useHasDefaultChainFeature(FEATURES.REQUIRE_SPACES_LOGIN)
+  const classicEnabled = useHasDefaultChainFeature(FEATURES.CLASSIC_UI_ENABLED)
+  const { data: spaces, isError: spacesError } = useSpacesGetV1Query(undefined, { skip: !isSignedIn })
+
+  useEffect(() => {
+    if (!router.isReady) return
+
```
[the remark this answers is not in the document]
katspaugh: Addressed in a82f66ab9 — `useSpaceIdSync` now gates on `selectIsStoreHydrated` (same pattern as `useFlowActivationGuard`). Added a test case for the unhydrated window. Good catch — the deep-link-with-spaceId scenario you described was exactly the failure mode I would have shipped without this fix.

apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceSafeSelectorItems.ts
```diff
@@ -186,7 +186,10 @@ export function useSpaceSafeSelectorItems() {
           source: 'space_selector',
         },
       )
-      router.push({ pathname: AppRoutes.home, query: { safe: `${chain.shortName}:${address}` } })
+      router.push({
+        pathname: AppRoutes.home,
+        query: { safe: `${chain.shortName}:${address}`, ...(spaceId ? { spaceId } : {}) },
```
vmosharova: Currently there is no spaceId in the query even when user is logged in to Spaces and goes to the Safe from a Space.
Also there is an unexpected fallback to only one Space: when changing Spaces there is only one  Space shown, even though user comes to the Safe level from another Space.

apps/web/src/features/spaces/hooks/useCurrentSpaceId.ts
```diff
@@ -1,24 +1,21 @@
...
   const isSiweAuthenticated = useAppSelector(isAuthenticated)
-
   const { data: spaces } = useSpacesGetV1Query(undefined, { skip: !isSiweAuthenticated })
 
   const rawSpaceId = query.spaceId
   const querySpaceId = typeof rawSpaceId === 'string' && rawSpaceId.length > 0 ? rawSpaceId : null
   const firstSpaceId = spaces?.[0] ? String(spaces[0].id) : null
 
-  return querySpaceId || storedSpaceId || firstSpaceId
+  return querySpaceId || firstSpaceId
```
vmosharova: Currently there is a silent fallback to `firstSpaceId`, even when the user chooses another Space.

apps/web/src/hooks/useSpaceIdSync/excludedRoutes.ts
```diff
@@ -0,0 +1,19 @@
+const EXACT_EXCLUDED: ReadonlyArray<string> = [
```
vmosharova, line 1: Do we also need to exclude `/new-safe`?

apps/web/src/components/welcome/WelcomeLogin/hooks/useSignInRedirect.ts
```diff
@@ -38,15 +39,23 @@ export const useSignInRedirect = ({ spacesAmount, inviteAmount, isSpacesLoading,
...
     if (error && !hasNotFoundSpaces(error)) return
 
-    if (hasSignedIn && (isNewUser || hasNotFoundSpaces(error))) {
+    const isNewUser = !inviteAmount && !isSpacesLoading && spacesAmount === 0
+    if (isNewUser || hasNotFoundSpaces(error)) {
       setRedirectLoading(true)
       router.push({ pathname: AppRoutes.welcome.createSpace, query: router.query })
     }
-  }, [hasSignedIn, isSpacesLoading, spacesAmount, inviteAmount, isUserSignedIn, error])
+  }, [hasSignedIn, isSpacesLoading, spacesAmount, inviteAmount, isUserSignedIn, error, router])
```
vmosharova: It might be more efficient to add a `router.query.redirect` dependency instead of the full `router` object.

apps/web/src/hooks/__tests__/useAdjustUrl.test.ts
```diff
@@ -37,4 +37,34 @@ describe('useAdjustUrl', () => {
       '/hello?safe=gor:0x0000000000000000000000000000000000000000&test=hello%3Aworld',
     )
   })
+
+  it('should not redirect to / on a safe route when ?spaceId is present (useSpaceIdSync owns the decision)', () => {
```
vmosharova, line 41: I'd add more tests for edge cases and conditionals here - for instance, when `?spaceId` is the only param, and other param combinations

apps/web/src/hooks/useSpaceIdSync/index.ts
```diff
@@ -0,0 +1,71 @@
...
+        router.replace({ pathname: AppRoutes.spaces.index })
+        return
+      case 'bounceToSignIn':
+        router.replace({
+          pathname: AppRoutes.welcome.spaces,
+          query: { redirect: decision.redirect },
+        })
+        return
+    }
+  }, [router, isSignedIn, isOidcPending, isStoreHydrated, requireLogin, classicEnabled, spaces, spacesError])
```
vmosharova: We have the entire `router` object in the dependencies - would it be more efficient to narrow it to `router.query` or other router params?

vmosharova (changes requested):
Nice!
Some changes / issues found:
- There is always a fallback for the 1st Space;
- User can't see the list of all owned Safes as the `/welcome/accounts` page is gone (need some product input here)

apps/web/src/components/common/SpaceSafeBar/hooks/useSpaceSafeSelectorItems.ts
katspaugh: Addressed in 762c8bbfb (SpaceSafeBar) and 5f01049ec (SafeCardReadOnly) — both now read the current spaceId via useCurrentSpaceId and carry it through the push, so the active Space survives a Safe switch. That also removes the wrong-Space fallback you saw, since useSpaceIdSync no longer has to re-inject the first owned space.

apps/web/src/features/spaces/hooks/useCurrentSpaceId.ts
katspaugh: Good catch — addressed in b7044bfce. The silent firstSpaceId fallback is gone; useCurrentSpaceId now returns the URL spaceId or null. useSpaceIdSync remains the single writer that keeps the URL populated for signed-in users, so the URL stays the single source of truth and the wrong-Space drift can't happen anymore.

apps/web/src/hooks/useSpaceIdSync/excludedRoutes.ts
katspaugh: Yes — added in b7044bfce. /new-safe is now in PREFIX_EXCLUDED alongside /welcome, /share, /import, /oidc, with a test case in excludedRoutes.test.ts.

apps/web/src/components/welcome/WelcomeLogin/hooks/useSignInRedirect.ts
katspaugh: Addressed in b7044bfce — destructured router into redirectParam (router.query.redirect), query and push, and narrowed the effect deps accordingly. No more full-router reference.

apps/web/src/hooks/__tests__/useAdjustUrl.test.ts
katspaugh: Expanded in b7044bfce — added ~90 lines of cases to useAdjustUrl.test.ts covering: ?spaceId alone, ?spaceId combined with unrelated params, ?safe without spaceId, both present, and the no-op when neither is set. Also added a focused useGetHref.test.ts (+82 lines) for the spaceId-preserving href construction used by the Safe list.

apps/web/src/hooks/useSpaceIdSync/index.ts
katspaugh: Addressed in b7044bfce — destructured to { isReady, pathname, asPath, query, replace } and listed each one in the dep array instead of the whole router object.

apps/web/src/hooks/__tests__/useAdjustUrl.test.ts
katspaugh: To be concrete about the new coverage, useAdjustUrl.test.ts now has 9 cases (was 3). The added ones:

- `?spaceId` present on a safe route → no redirect (line 41)
- neither `?safe` nor `?spaceId` on a safe route → still redirects to `/` (line 56, preserves original behavior)
- `?safe` only, no `?spaceId` → no redirect (line 71)
- both `?safe` and `?spaceId` → no redirect (line 86)
- `?spaceId` only on a non-safe route → no redirect, defers to useSpaceIdSync (line 101)
- path not in SAFE_ROUTES → no redirect (line 116)
- `!router.isReady` → no redirect (line 131)
- safe route + `?spaceId` only, no `?safe` → no redirect (line 147)

So every combination of {?safe present/absent} × {?spaceId present/absent} × {safe-route / non-safe-route} is covered, plus the router-not-ready edge.

PR #7856, opened 2026-05-13
refactor(mobile): use CGW /preview during compose instead of eager /propose

compojoom:
> Compose without propose,
> draft sits local until the user signs;
> CGW queue stays clean.

## What it solves

Resolves: mobile transaction compose currently calls CGW `/propose` eagerly, polluting the queue with phantom entries every time the user backs out, kills the app, or cancels biometrics mid-compose. Web has never done this — it calls `/preview` during compose and only proposes once the user signs.

## How this PR fixes it

The compose flow now mirrors web:

1. `useSendTransaction.handleReview` builds the `SafeTransaction` locally, computes `safeTxHash` via the Safe SDK, asks CGW for a `/preview`, and stashes a synthesized `TransactionDetails` in a new `draftTxSlice`. No `/propose` call.
2. Navigation continues to `/confirm-transaction?txId=<safeTxHash>` — same route, same param shape.
3. `useTransactionData(txId)` short-circuits to the slice when the id matches a draft, so the existing review screens render drafts identically to proposed txs **without any branching in the screens themselves**.
4. `useTransactionSigning.executeSign` branches once on the draft: builds the tx from the stored params, passes it to the existing sign service via a new `prebuiltSafeTx` option, then calls `/propose` with `signature: tx.encodedSignatures()` inline (one round-trip — same as web at [`apps/web/src/services/tx/proposeTransaction.ts:34`](apps/web/src/services/tx/proposeTransaction.ts#L34)). Clears the draft on success.
5. The three signing services (private key, Ledger, WalletConnect) each gained an optional `prebuiltSafeTx` so they can sign a local tx instead of fetching it from CGW by id.

## How to test it

1. Compose a token send on mobile, advance to "Confirm transaction", then back out. The transaction should **not** appear in the safe's pending queue (it would have before).
2. Compose a send and tap Sign. Network panel should show exactly one `POST /preview` during compose and exactly one `POST /propose` (carrying the signature inline) on Sign — no `POST /confirmations`.
3. Existing pending-tx flow: open a tx from the queue, sign — should still go through `POST /confirmations` (existing path, unchanged).
4. Repeat (2) with each signer type: private key, Ledger, WalletConnect.

## Affected flows

- **Compose → Confirm transaction → Review and confirm → Sign** (Send flow) — now uses preview + propose-with-signature instead of eager-propose + add-confirmation.
- **Pending tx → Confirm transaction → Review and confirm → Sign** (queue flow) — unchanged behaviour, still uses `useTransactionData(txId)` → CGW and `/confirmations`.

## Blast radius

- **New surfaces:** `apps/mobile/src/store/draftTxSlice.ts`, `apps/mobile/src/features/ConfirmTx/hooks/useTxPreview.ts`, `apps/mobile/src/features/ConfirmTx/utils/synthesizeDraftTxDetails.ts`, `apps/mobile/src/features/Send/services/prepareSendDraft.ts`.
- **Modified hooks:** `useTransactionData` gains a single-line draft short-circuit. `useTransactionSigning.executeSign` gains a single draft branch.
- **Modified sign services:** `signTx`, `LedgerSafeSigningService.signSafeTransaction`, `signWithWalletConnect` each gain an optional `prebuiltSafeTx` parameter that callers may supply to skip the CGW fetch. Existing behaviour is unchanged when omitted.
- **Removed:** `apps/mobile/src/features/Send/services/proposeSendTransaction.ts` + tests (no remaining callers — the Send flow was the only consumer).
- **Persistence:** `draftTx` is added to the redux-persist blacklist (drafts intentionally don't survive app kill).
- **Screens, sub-renderers, signer-selection hooks, `TransactionInfo`, `SignForm`, `ReviewFooter`, `ConfirmationView`** — all untouched. They consume `TransactionDetails` and don't know or care whether it came from CGW or the synthesizer.

## Risks / not checked

- **Did not exercise on a physical device.** Logic verified via unit + integration tests; no manual smoke test on iOS or Android yet.
- **Did not run the mobile E2E suite.** Recommend a smoke pass before merge.
- **WalletConnect inbound dApp-initiated transactions** — none currently route through `proposeNewTransaction` on mobile (verified via grep — `Send` was the only caller), but worth a visual check that dApp request signing still works as before.
- **Single-sig auto-execute path** — for `threshold=1` the `executeSign` propose-with-signature should result in CGW eventually executing the tx; this PR does not change the execute pathway but the propose response now arrives later (after user signs). Worth confirming.
- **Smart-contract-wallet signers** — confirmed mobile only has `private-key | ledger | walletconnect` signer types today; no SCW path to worry about.

## Visual summary

```mermaid
flowchart TB
  subgraph Before
    A1[EnterAmount] --> B1[handleReview]
    B1 --> C1[POST /propose<br/>no signature]
    C1 --> D1[Navigate with real txId]
    D1 --> E1[GET /transactions/&#123;id&#125;]
    E1 --> F1[Review screens]
    F1 --> G1[Sign]
    G1 --> H1[POST /confirmations]
  end
  subgraph After
    A2[EnterAmount] --> B2[handleReview]
    B2 --> P2[POST /preview]
    P2 --> S2[draftTxSlice<br/>synthetic TransactionDetails]
    S2 --> D2[Navigate with safeTxHash]
    D2 --> R2[Review screens<br/>useTransactionData reads slice]
    R2 --> G2[Sign]
    G2 --> C2[POST /propose<br/>signature inline]
    C2 --> Cl2[clearDraft]
  end
```

## Checklist

- [x] I've documented how it affects the analytics (if at all) 📊 (no analytics impact)
- [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻 (74 unit tests, including new branch coverage)

---

## CLA signature


apps/mobile/src/features/ConfirmTx/hooks/useTxPreview.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Fetches a CGW transaction preview for a locally composed (un-proposed)
+ * transaction. Mirrors the web app's `useTxPreview` so both clients
+ * surface the same decoded-info view during compose.
+ *
+ * Re-fires on dependency change (same behaviour as web). Callers that
+ * need persistence across screen re-mounts should snapshot the result
+ * into `draftTxSlice`.
+ */
+export const useTxPreview = (input: PreviewInput | undefined): PreviewResult => {
```
tmjssz: This hook is not used anywhere. Or am I missing anything?

apps/mobile/src/features/ConfirmTx/hooks/useTransactionData.ts
```diff
@@ -1,16 +1,49 @@
...
       id: txId,
     },
     {
-      skip: !txId || !activeSafe?.chainId,
+      skip: !!draft || !txId || !activeSafe?.chainId,
     },
   )
+
+  return useMemo(() => {
+    if (!draft) {
```
tmjssz, line 34: Can you check this finding from Claude:
> useTransactionData refetch() on a draft will 404. useTransactionData.ts:31-44: the hook overrides the RTK Query result but the underlying `query.refetch` still points at the skipped CGW endpoint. A caller (or the RTK middleware) invoking refetch() will fire GET /transactions/<safeTxHash> and overwrite the synthesized data with a 404. The comment claims the draft branch prevents this, but it only prevents the initial fetch via skip.

tmjssz (commented):
Amazing refactor!

tmjssz:
@claude review

tmjssz:
@compojoom I think the P2s found by claude are worth fixing before we merge it

tmjssz (dismissed):

tmjssz (approved):

PR #7868, opened 2026-05-18
chore: tighten claude-code-review workflow triggers and config

PooyaRaki:
This PR upgrades Claude code review workflow.

  ## What it solves

  Resolves: noisy / overlapping Claude review runs and overly permissive trigger matching in `.github/workflows/claude-code-review.yml`.

  ## How this PR fixes it

  - Restricts the trigger to comments **starting with** `@claude review` (was `contains`), and only on PR comments (`issue.pull_request != null`).
  - Removes the commented-out `pull_request` trigger and other template scaffolding.
  - Adds a `concurrency` group keyed on PR number with `cancel-in-progress: true` so a new review cancels the previous one for the same PR.
  - Adds `timeout-minutes: 5` to bound runs.
  - Bumps `pull-requests` permission to `write` (needed for inline review comments); keeps `id-token: write` and documents it as required for the Claude GitHub App OIDC exchange.
  - Switches to `track_progress: true` and pins explicit `claude_args`: `--model claude-opus-4-7`, `--max-turns 10`, and a narrow `--allowedTools` list (inline comment MCP + read-only `gh pr` bash).
  - Passes `REPO` and `PR NUMBER` into the prompt so the action has unambiguous context.

  ## How to test it

  1. Open a draft PR against `dev` from this branch.
  2. From a collaborator/member/owner account, comment `@claude review` on the PR — workflow should run.
  3. Comment `@claude review please look at X` — should still run (starts-with match).
  4. Comment `please @claude review` — should **not** run (no longer matches).
  5. Post a second `@claude review` while the first is still running — the first run should be cancelled.
  6. Confirm the run respects the 5-minute timeout and uses `claude-opus-4-7`.

  ## Affected flows

  - GitHub Actions: on-demand Claude code review triggered from PR comments and PR review comments.

  ## Blast radius

  - Single workflow file: `.github/workflows/claude-code-review.yml`.
  - No application code, no shared packages, no runtime behavior of web/mobile apps affected.
  - Permissions delta: `pull-requests: read → write` (scoped to this workflow's `GITHUB_TOKEN`).
  - Concurrency change will cancel in-flight Claude review runs on the same PR when a new trigger arrives — intentional.

  ---
  CLA signature

  With the submission of this Pull Request, I confirm that I have read and agree to the terms of the Contributor License Agreement (https://safe.global/cla).

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -7,14 +7,6 @@ on:
     types: [created]
   pull_request_review_comment:
     types: [created]
-  # pull_request:
-  #   types: [opened, synchronize]
-  #   # Optional: Only run on specific file changes
-  #   # paths:
-  #   #   - "src/**/*.ts"
-  #   #   - "src/**/*.tsx"
-  #   #   - "src/**/*.js"
-  #   #   - "src/**/*.jsx"
 
 jobs:
   claude-review:
@@ -25,7 +17,8 @@ jobs:
       github.event_name == 'workflow_dispatch' ||
       (
         github.event_name == 'issue_comment' &&
-        contains(github.event.comment.body, '@claude review') &&
+        github.event.issue.pull_request != null &&
+        startsWith(github.event.comment.body, '@claude review') &&
         (
           github.event.comment.author_association == 'COLLABORATOR' ||
           github.event.comment.author_association == 'MEMBER' ||
@@ -34,7 +27,7 @@ jobs:
       ) ||
       (
         github.event_name == 'pull_request_review_comment' &&
-        contains(github.event.comment.body, '@claude review') &&
+        startsWith(github.event.comment.body, '@claude review') &&
         (
           github.event.comment.author_association == 'COLLABORATOR' ||
           github.event.comment.author_association == 'MEMBER' ||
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-code-review.yml
```diff
@@ -25,7 +17,8 @@ jobs:
       github.event_name == 'workflow_dispatch' ||
       (
         github.event_name == 'issue_comment' &&
-        contains(github.event.comment.body, '@claude review') &&
+        github.event.issue.pull_request != null &&
```
compojoom, line 20: why do we suddenly need this?

.github/workflows/claude-code-review.yml
```diff
@@ -48,11 +41,15 @@ jobs:
...
 
     runs-on: ubuntu-latest
+    timeout-minutes: 5
+    concurrency:
+      group: claude-review-${{ github.event.issue.number || github.event.pull_request.number || github.ref }}
+      cancel-in-progress: true
     permissions:
       contents: read
-      pull-requests: read
+      pull-requests: write
```
compojoom, line 50: why do we need to change the permission?

compojoom (approved):

.github/workflows/claude-code-review.yml
PooyaRaki: It's required as in the new version of the workflow, it directly posts a comment on the PR and therefor needs the access.
FYI: https://github.com/anthropics/claude-code-action/blob/main/docs/solutions.md

.github/workflows/claude-code-review.yml
PooyaRaki: It's to make sure it fails fast on issues as we only need it for PRs.

PR #7874, opened 2026-05-19
fix(web): prevent security hub gauge flipping between scans

clovisdasilvaneto:
> Scanners flipped on rescan,
> data not yet loaded, gauge would lie —
> wait, then mark inconclusive.

## What it solves

Resolves a regression in the Security Hub where the workspace gauge and per-Safe results would change between scans of the same Safe. On the initial scan the factory-validation scanner often reported \"Deployed from an unrecognized source\" and the account-setup scanner reported a weak threshold, only for both to flip to \"clear\" on a manual rescan a moment later.

The root cause was a race: `useSafeScanContext` gated scan launches on RTK Query's `isLoading`, which only flips on the very first fetch. In the uninitialized-to-pending transition window, `isLoading` was already `false` while `currentData` was still `undefined`. Scanners ran with `creationInfo=null` and `owners=[]` and produced misleading results that the user could not distinguish from real findings.

## How this PR fixes it

1. **Tighter gate in `useSafeScanContext`** — switched from `isLoading` to `isFetching` and additionally require either data presence or a definitive query error before returning a context. This catches both initial fetches and refetches; a definitive error lets the scanner proceed and emit `inconclusive` rather than blocking forever.
2. **Scanners emit `inconclusive` for missing inputs** — `accountSetup` (owners==0, impossible for a real Safe) and `factoryValidation` (creationInfo missing) now return `inconclusive` instead of `partial`. `inconclusive` is excluded from the workspace aggregate and per-Safe grade, so a transient missing-data state no longer penalizes the score or surfaces a misleading entry that flips on rescan.
3. **UI copy** — `useSecurityChecks` shows \"Deployment origin not yet verified\" for the `inconclusive` factory case instead of the misleading \"Deployed from an unrecognized source\".

## How to test it

1. `yarn workspace @safe-global/web dev`
2. Open a workspace with several Safes in the Security Hub.
3. Run an initial scan and watch the gauge stream as each Safe completes — no Safe should be flagged with \"Deployed from an unrecognized source\" or \"Weak threshold\" while its creation/owner data is still loading.
4. Trigger a manual rescan — the per-Safe results should not flip from a finding to clear (or vice versa) when nothing on-chain has changed.
5. For a Safe whose creation transaction the gateway genuinely cannot resolve, confirm the factory check shows \"Deployment origin not yet verified\" (`inconclusive`) rather than blocking the scan forever.

## Affected flows

- Security Hub initial workspace scan (multi-Safe).
- Security Hub manual rescan for a single Safe.
- Per-Safe security panel rendering (factory entry copy).
- Workspace health gauge streaming on first scan and reset when switching spaces.

## Blast radius

- `useSafeScanContext` — only consumer is `useSecurityScan` via the Security Hub auto-scan orchestrator; tightening the gate cannot cause scans to run with less data than before, only delay or skip them when data is genuinely incomplete.
- `accountSetupScanner` / `factoryValidationScanner` — both feed `useSecurityScan` and the workspace aggregation (`computeWorkspaceHealthScore`). `inconclusive` was already a supported status excluded from aggregates, so no scoring path needs adjustment.
- `useSecurityChecks` — pure UI mapping for the per-Safe security panel.
- No RTK Query endpoints, slices, feature flags, persistence, or shared packages touched. Mobile is unaffected.

## Risks / not checked

- Did not exercise the full auto-scan orchestrator queue under throttled network in a real browser; new gate is covered by `useSafeScanContext.test.ts` and existing orchestrator tests pass.
- Did not test on mobile (web-only feature).
- Did not verify behavior under a sustained query error (e.g., gateway 5xx for creation tx) beyond the unit test asserting the `isError` branch.
- Did not re-snapshot the workspace health gauge visually in Storybook; new `WorkspaceHealthCard.test.tsx` covers the streaming and reset behavior at the RTL level.

## Visual summary

\`\`\`mermaid
flowchart TB
  subgraph Before
    A1[RTK Query: isLoading=false<br/>currentData=undefined] --> B1[useSafeScanContext returns context<br/>with creationInfo=null]
    B1 --> C1[factoryValidationScanner ->'partial'<br/>'Deployed from unrecognized source']
    C1 --> D1[Score penalized, finding shown]
    D1 --> E1[Manual rescan: query has resolved<br/>finding disappears -> user confused]
  end
  subgraph After
    A2[RTK Query: isFetching or<br/>currentData absent without error] --> B2[useSafeScanContext returns null<br/>scan waits]
    B2 --> C2[Once settled: scanner runs<br/>with real data OR emits 'inconclusive']
    C2 --> D2['inconclusive' excluded from aggregate<br/>copy: 'Deployment origin not yet verified']
    D2 --> E2[Manual rescan: stable result]
  end
\`\`\`

## Checklist

- [ ] I've tested the branch on mobile 📱 (n/a — web-only)
- [ ] I've documented how it affects the analytics (if at all) 📊 (no analytics changes)

---

## CLA signature


apps/web/src/features/security/data/scanners/accountSetup.ts
```diff
@@ -9,13 +9,16 @@ export const accountSetupScanner: SecurityScanner = {
...
+      // Owner data hasn't loaded — distinct from "loaded and shows 0 owners" (impossible
+      // for a real Safe). Return inconclusive so a transient missing-data state during
+      // scan doesn't penalize the score or flag the Safe with a "weak threshold" entry
+      // that flips on rescan once `safeInfo.owners` populates.
       return {
-        status: 'partial',
-        severity: getSeverityFromScore(score),
-        score,
+        status: 'inconclusive',
+        severity: 'Low',
```
tmjssz: Why don't we use `getSeverityFromScore` here?

apps/web/src/features/security/data/scanners/factoryValidation.ts
```diff
@@ -13,12 +13,16 @@ export const factoryValidationScanner: SecurityScanner = {
...
+      // as deployed from an unrecognized source: we genuinely don't know yet, and the
+      // user shouldn't see the result flip between scans once creationTx resolves.
       return {
-        status: 'partial',
+        status: 'inconclusive',
         severity: 'Low',
-        score: 70,
-        evidence: [{ label: 'Status', value: 'Creation data not available' }],
-        remediation: 'Deployment origin could not be verified because creation data is not yet available.',
+        score: 50,
```
tmjssz: Same here: lets use getSeverityFromScore for all the results in this file and make sure it's consistently used for all other scanners as well

apps/web/src/features/spaces/components/SecurityHub/__tests__/WorkspaceHealthCard.test.tsx
```diff
@@ -0,0 +1,108 @@
+import { render, screen } from '@testing-library/react'
+import WorkspaceHealthCard from '../components/WorkspaceHealthCard/WorkspaceHealthCard'
+import type { ScanResult } from '@/features/security/types'
+import type { SpaceSafeEntry } from '../types'
+
+const scanKey = (address: string, chainId: string) => `${address}:${chainId}`
```
tmjssz: We can import the original `scanKey` function here instead of redefining it

apps/web/src/features/security/data/scanners/accountSetup.ts
clovisdasilvaneto: Switched to `getSeverityFromScore(score)` here and across all the other scanners (`factoryValidation`, `multichainSetup`, `transactionScanning`, `signerIntegrity`) for consistency. Applying the helper surfaced a few latent score/severity mismatches: `multichainSetup` partial (score 30) moves from Medium → High, and the inconclusive / not_applicable cases (score 50/70) move from Low → Medium — the latter are filtered from the aggregate and per-Safe grade so they don't affect scoring, only ordering in the panel.

apps/web/src/features/security/data/scanners/factoryValidation.ts
clovisdasilvaneto: Done — `factoryValidation` and the other scanners (`accountSetup`, `multichainSetup`, `transactionScanning`, `signerIntegrity`) now all derive severity via `getSeverityFromScore(score)`. The shared scanners that already used the helper (`contractVersion`, `fallbackHandler`, `guard`, `modules`, `pendingTx`, `recovery`) were already consistent.

apps/web/src/features/spaces/components/SecurityHub/__tests__/WorkspaceHealthCard.test.tsx
clovisdasilvaneto: Done — importing `scanKey` from `@/features/security/data/scanners/utils` now. As a side benefit, the imported version lowercases the address so the test keys align with what the component computes.

clovisdasilvaneto:
### Doc alignment follow-up

Audited every scanner branch against the [Security Hub Documentation matrix](https://www.notion.so/safe-global/Security-Hub-Documentation-3408180fe573804c8b2ec2d9b522db7f). Status / score / CTA all match. Severity matched everywhere except for `inconclusive` / `not_applicable` cases — the doc lists those as Severity = `Low` (since they're filtered from the aggregate and rendered as a greyed card), but applying `getSeverityFromScore` raw mapped them to `Medium`/`High` based on their decorative score.

To match the doc without resorting to hardcoded severities again, `getSeverityFromScore` now takes an `excluded` option:

```ts
getSeverityFromScore(score) // real findings — severity follows score
getSeverityFromScore(score, { excluded: true }) // inconclusive / not_applicable → 'Low'
```

Applied to all `inconclusive` / `not_applicable` branches:

- `accountSetup` inconclusive (owners loading) → Low
- `factoryValidation` inconclusive (creation data loading) → Low
- `multichainSetup` not_applicable (single-chain) → Low (no change, score 100 already mapped to Low)
- `transactionScanning` not_applicable (chain unsupported) → Low (was Medium)
- `signerIntegrity` inconclusive (CGW endpoint pending) → Low
- `recovery` not_applicable (chain doesn't support) → Low (no change, score 100 already mapped to Low)

Real findings (multichain partial, all `issue` and `partial` cases) continue to derive severity from score via the helper, so the score↔severity bands stay enforced. The doc remains the source of truth and the matrix now aligns with what the code emits.

The doc still has two open items not covered here: it needs new rows for the `inconclusive` branches this PR adds (`accountSetup` owners-loading and `factoryValidation` creation-data-loading) — I'll send those as a Notion update separately.

apps/web/src/features/security/data/scanners/constants.ts
```diff
@@ -40,8 +40,17 @@ export const SEVERITY_SCORE_THRESHOLDS: Array<{ minScore: number; severity: Secu
...
+/**
+ * Derives a SecurityGrade from a numeric score (0–100) via SEVERITY_SCORE_THRESHOLDS.
+ *
+ * Pass `excluded: true` for results with status `inconclusive` or `not_applicable` —
+ * these are filtered from the aggregate score and per-Safe grade, so their score is
+ * decorative. Forcing 'Low' keeps them visually neutral and prevents them from
+ * outranking real findings in severity-sorted views, matching the documented matrix
+ * (see Security Hub — Documentation, "Excluded check rendering").
+ */
+export const getSeverityFromScore = (score: number, options: { excluded?: boolean } = {}): SecurityGrade => {
```
tmjssz, line 51: This new excluded flag is used for exaclty all statuses that are `inconclusive` or `not_applicable` now, right? Maybe it's easier to just set "Low" directly in those cases, instead of calling this function. 
Or even better would be in my opinion: change the function to accept both `score` and `status`. And then inside do something like this:
```ts
if ([`inconclusive`, `not_applicable`].includes(status)) return 'Low'
```
I think it would be easier to read the code in the end.

tmjssz (approved):

apps/web/src/features/security/data/scanners/constants.ts
clovisdasilvaneto: yeah, makes sense @TimDaub

clovisdasilvaneto: will send it in a separated PR

PR #7875, opened 2026-05-19
feat(web): use backend as source of truth for counterfactual safes

katspaugh:
## What it solves

Counterfactual (undeployed) safe data was stored locally in Redux/localStorage and synced to the backend as a fire-and-forget side effect. This meant:
- Backend errors (e.g. 500) during creation were silently swallowed — the user saw a success screen even when the backend failed
- Space members couldn't see each other's counterfactual safes
- Loading a CF safe by URL raced with the lazy-loaded sync hook, causing a flash of "Safe couldn't be loaded"

Depends on backend PR: https://github.com/safe-global/safe-client-gateway/pull/3011

Linear: https://linear.app/safe-global/project/migrate-counterfactual-safes-to-cgw-cdbddc664cb3/overview

### Creation flow
- **ReviewStep** now `await`s the backend POST before dispatching `addUndeployedSafe` to Redux. On failure, the user sees an error and can retry.
- The fire-and-forget `addUndeployedSafe` listener in `counterfactualSyncListener` is removed. The `removeUndeployedSafe` → DELETE listener remains (now using `.unwrap()`).

### Data loading
- `useCounterfactualSafeSync` fetches from both `/v1/users/counterfactual-safes` and `/v1/spaces/:spaceId/counterfactual-safes` on auth, so all space members see CF safes.
- The hook moved from lazy-loaded `LazyCounterfactual` to `InitApp` in `_app.tsx` to run eagerly on auth.
- A `cfSafeSynced` flag in `authSlice` (reset on each hydration) lets `useLoadSafeInfo` show a loading state instead of an error while sync is pending.
- `useLoadSafeInfo` suppresses CGW 404 errors when an undeployed safe exists in Redux or while CF sync is in progress.

### Spaces Accounts widget
- `useSpaceAccountsData` falls back to CF safe `threshold`/`owners` from Redux when no `SafeOverview` is available.

### Other
- `counterfactualSafeMapper.fromBackendDto` uses `??` instead of `||` for `paymentToken`.
- RTK Query dispatches use `.unwrap()` for proper error handling.
- Services barrel re-exports `counterfactualSafeMapper`.

## How to test it

1. **Create CF safe** → should see 201 + success screen
2. **Break the backend** (revert the transformer fix in CGW) → creation should show error, retry should work
3. **Navigate to CF safe URL directly** → should load without error flash
4. **Spaces Accounts widget** → CF safe should show owners/threshold
5. **Different space member** → should see CF safes created by another member

## Visual summary

```mermaid
sequenceDiagram
    participant User
    participant ReviewStep
    participant CGW Backend
    participant Redux
    participant UI

    User->>ReviewStep: Click "Create account"
    ReviewStep->>CGW Backend: POST /v1/users/counterfactual-safes
    alt Success (201)
        CGW Backend-->>ReviewStep: OK
        ReviewStep->>Redux: addUndeployedSafe
        ReviewStep->>UI: Navigate + success screen
    else Error (500)
        CGW Backend-->>ReviewStep: Error
        ReviewStep->>UI: Show error, user can retry
    end

    Note over Redux,UI: On app load (auth)
    Redux->>CGW Backend: GET /v1/users/counterfactual-safes
    Redux->>CGW Backend: GET /v1/spaces/:id/counterfactual-safes
    CGW Backend-->>Redux: Merge into undeployedSafes
    Redux-->>UI: All consumers reactively update
```

## Checklist

---

## CLA signature


vmosharova:
@claude review

apps/web/src/features/counterfactual/services/counterfactualSafeMapper.ts
```diff
@@ -0,0 +1,79 @@
...
+export type BackendCfSafeDto = {
+  chainId: string
+  address: string
+  factoryAddress: string
+  masterCopy: string
+  saltNonce: string
+  safeVersion: string
+  threshold: number
+  owners: string[]
+  fallbackHandler: string
```
vmosharova: There is some inconsistency: `fallbackhandler` as well as `paymentReceiver` are marked as optional in the [counterfactual-safes.ts](https://github.com/safe-global/safe-wallet-monorepo/pull/7875/changes#diff-399eb7535f1dfa9f84df82d2ce0fc69c896b01a952eddd3c2276ade772b60d03)

apps/web/src/features/counterfactual/services/counterfactualSafeMapper.ts
```diff
@@ -0,0 +1,79 @@
...
+    address: dto.address,
+    props: {
+      factoryAddress: dto.factoryAddress,
+      masterCopy: dto.masterCopy,
+      saltNonce: dto.saltNonce,
+      safeVersion: dto.safeVersion as SafeVersion,
+      safeAccountConfig: {
+        threshold: dto.threshold,
+        owners: dto.owners,
+        fallbackHandler: dto.fallbackHandler,
```
vmosharova: Same here: fields are required, meanwhile in the BE schema they are optional

apps/web/src/hooks/loadables/useLoadSafeInfo.ts
```diff
@@ -67,9 +91,9 @@ const useLoadSafeInfo = (): AsyncResult<ExtendedSafeInfo> => {
       return new Error(errorMessage)
     }
     return undeployedSafe ? undeployedError : undefined
-  }, [cgwError, undeployedSafe, undeployedError])
+  }, [cgwError, undeployedSafe, undeployedError, suppressCgwError])
 
-  const loading = cgwLoading
+  const loading = cgwLoading || awaitingCfSync
```
vmosharova: If we await Cf sync here, then we will slow down the page load also for deployed Safes. If CF endpoint will get a 500, there is a risk that the page will be stuck loading.

apps/web/src/features/counterfactual/components/PayNowPayLater/index.tsx
```diff
@@ -18,24 +19,45 @@ import {
...
   const { showGasFeeEstimation, showStablecoinFeeInfo } = useNativeTokenDisplay()
+  const { signIn, loading: signingIn } = useSiwe()
 
-  const onChoosePayMethod = (_: ChangeEvent<HTMLInputElement>, newPayMethod: string) => {
+  const signInAndSelectPayLater = async () => {
+    if (signingIn) return
+    const result = await signIn()
+    if (result && !result.error) {
+      const oneDayInMs = 24 * 60 * 60 * 1000
+      dispatch(setAuthenticated(Date.now() + oneDayInMs))
```
vmosharova: Does it make sense to duplicate a session lifetime? Maybe it's better to put it in a const across the codebase to keep it in sync?

apps/web/src/features/counterfactual/hooks/useCounterfactualSafeSync.ts
```diff
@@ -0,0 +1,153 @@
...
+              }
+            }),
+          )
+        }
+
+        // Fetch CF safes from user endpoint and space endpoint
+        const userQuery = dispatch(counterfactualSafesApi.endpoints.counterfactualSafesGetV1.initiate(undefined))
+        const spaceQuery = spaceId
+          ? dispatch(spacesApi.endpoints.spaceCounterfactualSafesGetV1.initiate({ spaceId: Number(spaceId) }))
+          : null
```
vmosharova, line 85: It makes sense to add a guard for `Number(spaceId)` to ensure it is a number.

apps/web/src/features/counterfactual/services/persistCounterfactualSafe.ts
```diff
@@ -0,0 +1,102 @@
...
+      }),
+    )
+    if ('error' in userResult) {
+      return { ok: false, error: toPersistError(userResult.error) }
+    }
+
+    if (spaceId) {
+      const spaceResult = await dispatch(
+        spacesApi.endpoints.spaceSafesCreateV1.initiate({
+          spaceId: Number(spaceId),
```
vmosharova: It makes sense to add a guard for `Number(spaceId)` to ensure it is a number.

vmosharova (changes requested):
Left some comments :)

katspaugh:
Worked through all six. Three landed as commits in `7e7ab9d18`; three I'm pushing back on with reasoning.

### Fixed

**1. `setCfSafeSynced(true)` on sync failure** — valid, fixed. `useCounterfactualSafeSync` now retries the GET once after a 2s backoff before settling. Stops the "transient blip → space-mate's CF safe unreachable via direct URL" failure. Doesn't loop forever — one bounded retry, then unblock. Added tests for both the recovers-on-retry and gives-up-after-two-failures paths.

**3. Rollback DELETE failure leaves orphan** — valid, fixed. `persistCounterfactualSafe` now checks the rollback result; if both the space POST and the rollback DELETE fail, it dispatches `enqueuePendingCfDelete` so the next sync flushes the orphan instead of re-surfacing it locally as "Not activated".

**5. `getCfOwners` for multichain only consults first chain** — valid, fixed. `useSpaceAccountsData` now does `safe.safes.find(s => undeployedSafes[s.chainId]?.[safe.address])` to pick the chain that actually has a CF entry. Added a regression test where the CF entry lives on chain index 1.

### Not changing

**2. `lastSyncedKey` should include user identity** — currently safe and the comment in your note acknowledges it. The `useEffect` deps include `isUserAuthenticated`, and the early-return at the top sets `lastSyncedKey.current = null` synchronously on sign-out before any subsequent sign-in can observe it. The "future change makes `setUnauthenticated` slower than the next mount" scenario is hypothetical. Happy to revisit if/when that future change actually lands.

**4. `Number(spaceId)` produces silent NaN** — YAGNI. `lastUsedSpace` is only ever written numeric in current FE code; the trigger requires hand-corrupted persisted state. Adding `Number.isFinite` would be cheap, but I'd rather wait until we have a reproducer than scatter defensive guards against unproven inputs.

**6. Self-heal effect can race with stale CGW cache** — same call as before: CGW's `/v1/chains/:chain/safes/:address` returns 404 for undeployed addresses; the only way `cgwData` becomes truthy is if the address actually exists on-chain. A Safe can't undeploy. The reviewer's specific suggestion (gate on `cgwData.deployed === true`) doesn't help because `deployed` isn't a backend field — it's synthesized client-side in `useLoadSafeInfo.ts:56` (`cgwDataWithDeployed = cgwData ? { ...cgwData, deployed: true } : undefined`), so gating on it is tautological. If you want belt-and-suspenders, the cheap version is gating on `cgwData.owners?.length > 0` (a deployed Safe must have owners) or `sameAddress(cgwData.address.value, address)` (confirms the response is actually for the requested address). Happy to add either if you'd prefer the defense, but the underlying race the comment describes would require a CGW correctness bug, not a cache race in the documented sense.

apps/web/src/features/counterfactual/hooks/useCounterfactualSafeSync.ts
katspaugh: Fixed in `c31a0b922`. Added a `parseSpaceId(spaceId: string | null): number | null` helper in `src/utils/spaces.ts` that returns `null` for `null`, empty/whitespace, or any non-finite `Number(...)` result. Both call sites (`useCounterfactualSafeSync` and `persistCounterfactualSafe`) now skip the space-scoped API call when the parsed id is `null` instead of hitting it with `NaN`. Unit-tested for empty string, `'abc'`, `'42abc'`, and `'Infinity'`/`'NaN'`.

apps/web/src/features/counterfactual/services/persistCounterfactualSafe.ts
katspaugh: Fixed in `c31a0b922`. Added a `parseSpaceId(spaceId: string | null): number | null` helper in `src/utils/spaces.ts` that returns `null` for `null`, empty/whitespace, or any non-finite `Number(...)` result. Both call sites (`useCounterfactualSafeSync` and `persistCounterfactualSafe`) now skip the space-scoped API call when the parsed id is `null` instead of hitting it with `NaN`. Unit-tested for empty string, `'abc'`, `'42abc'`, and `'Infinity'`/`'NaN'`.

apps/web/src/hooks/loadables/useLoadSafeInfo.ts
katspaugh: Fixed in `c31a0b922`. Changed line 96 to `const loading = cgwLoading || (!cgwData && awaitingCfSync)` — once `cgwData` arrives, the loading flag drops immediately regardless of CF sync status, so deployed Safes render without waiting. Combined with the bounded-retry/always-settle behavior I added earlier (`useCounterfactualSafeSync` now retries once after 2s on failure then unconditionally dispatches `setCfSafeSynced(true)`), a CF endpoint 5xx also bounds the loading window — the page won't stick on the loading state.

apps/web/src/features/counterfactual/services/counterfactualSafeMapper.ts
katspaugh: Fixed in `c31a0b922`. `BackendCfSafeDto` now marks `fallbackHandler`, `to`, and `paymentReceiver` as `?: string | null` to match the schema, and `fromBackendDto` centralizes the `null → ZERO_ADDRESS` coercion so downstream FE code (which assumes string addresses) keeps working without per-callsite normalization.

apps/web/src/features/counterfactual/services/counterfactualSafeMapper.ts
katspaugh: Fixed in `c31a0b922`. `BackendCfSafeDto` now marks `fallbackHandler`, `to`, and `paymentReceiver` as `?: string | null` to match the schema, and `fromBackendDto` centralizes the `null → ZERO_ADDRESS` coercion so downstream FE code (which assumes string addresses) keeps working without per-callsite normalization.

apps/web/src/features/counterfactual/components/PayNowPayLater/index.tsx
katspaugh: Fixed in `c31a0b922`. Added `SESSION_LIFETIME_MS` to `authSlice.ts` and replaced the inline `24 * 60 * 60 * 1000` constants in `PayNowPayLater`, `SignInButton`, `useHomeAuth`, and `reconcileAuth`.

vmosharova (dismissed):
After testing the CF creation flow in the UI I noticed the following details that need some attention:

**When "Sponsored" is selected:**
- On finishing of the CF Safe creation, many endpoints throw a `404`: `/v1/chains/{chainId}/safes/{safeAddress}`, `/history`, `queued`

**When "Pay later" is selected (not sponsored):**
- The new `/v1/users/counterfactual-safes` gives a 201 and then several 404s

vmosharova (approved):
👍🏻

PR #7888, opened 2026-05-20
chore(ci): cache only canonical yarn store, run install on every job

compojoom:
> Lean on setup-node's yarn cache,
> let install rebuild node_modules each run,
> release alone verifies the lockfile.

## What it solves

The yarn cache previously bundled `**/node_modules` and generated contract types alongside the Berry store, under a hand-rolled `cache-deps` key. That gave us two sources of truth — a cache hit could restore stale `node_modules` that no longer matched the lockfile, and skip `yarn install` (and its `postinstall` type generation / chain fetch) entirely. It also re-restored the Berry store under a second key that `setup-node` already caches, costing ~23s of tar extract and ~850 MB of GHA cache per job.

## How this PR fixes it

Two commits:

**1. Cache only the canonical store, run install on every job**
- Drop `**/node_modules` and `packages/utils/src/types` from the cache.
- Always run `yarn install --immutable` — no cache-hit skip. `node_modules` and contract types are rebuilt every job; the apps/web workspace `postinstall` (`generate-types` + `fetch-chains`) runs as part of install.
- Remove the explicit `Yarn after-install` step and the `after-install` input. Every caller that passed `after-install: 'false'` (mobile workflows + `store-codegen-drift`) drops the now no-op argument.
- Add a `verify_cache` input that toggles `yarn install --check-cache`. `web-tag-release.yml` opts in so release builds verify cached tarball checksums against the lockfile; PR jobs stay fast.

**2. Stop duplicating the Berry cache; isolate the Cypress binary**
- `setup-node@v6`'s `cache: 'yarn'` already restores `~/.yarn/berry/cache`, so the custom `cache-deps` yarn modes were redundant — removed. `cache-deps` is now Next.js-build-cache only.
- Keep a small dedicated `actions/cache` step for `.yarn/install-state.gz` (yarn resolution state; self-heals if stale).
- Move the Cypress binary cache (`~/.cache/Cypress`) into `actions/cypress` so non-e2e jobs don't drag it through GHA.
- Add a `restore-keys` fallback to the Next.js cache for partial warm restores on source edits.

## How to test it

1. Confirm on this PR's CI that `Yarn install` runs in every job (no "skipped due to cache hit") and `packages/utils/src/types/contracts/` is populated afterward; `yarn workspace @safe-global/web type-check` passes.
2. Confirm `setup-node` restores the yarn cache (look for the post-job "Cache saved" / "Cache restored" lines) and the dedicated install-state cache is hit on the second run.
3. Trigger `web-tag-release.yml` on a feature branch and confirm `yarn install --immutable --check-cache` runs and exits clean.

## Affected flows

- Every CI job using `./.github/actions/yarn` — i.e. effectively every workflow with an install step (web + mobile).
- `web-tag-release.yml` — the only caller that opts into `--check-cache`.
- E2E jobs via `./.github/actions/cypress` — now own the Cypress binary cache.
- Mobile workflows and `store-codegen-drift` — dropped `after-install: 'false'`; workspace `postinstall` now runs there (a few extra seconds of typechain).

## Blast radius

- **`actions/yarn`** — consumed by ~25 workflows, web and mobile.
- **`actions/cache-deps`** — now Next.js-only; the `restore-yarn`/`save-yarn` modes are gone. Confirm no remaining caller references them.
- **`actions/cypress`** — gains the Cypress binary cache step.
- **Cache scope** — dropping the custom yarn modes means existing `web-core-modules-*` entries go unused and age out (GHA evicts after 7 days). First runs after merge populate the setup-node and install-state caches cold.
- **Workspace `postinstall`** — load-bearing for contract type generation now that the explicit step is gone. `fetch-chains` runs in every job but degrades gracefully: on a Config Service error it warns and writes an empty `chains.json` fallback (exit 0), so it is not a new CI failure point.

## Risks / not checked

- **Cache writes are not restricted by event type.** `setup-node`'s built-in `cache: 'yarn'` saves on any trigger and offers no `if:` guard, and the `install-state` cache step is likewise unguarded. The Berry store is content-addressed and Yarn validates tarball checksums against the lockfile on install, so a poisoned entry fails verification — but this PR does not add a `workflow_run` / `pull_request_target` write guard.
- **`--check-cache` is web-release-only.** `mobile-dev-release.yml` does not opt in. Mobile EAS builds have their own integrity path; calling it out as a conscious asymmetry rather than an oversight.
- Did not measure warm-cache `--check-cache` wall-clock on a real runner (plan target: under 120s). First release run will surface it.
- Did not confirm mobile jobs' extra typechain time stays within their existing budgets.

## Visual summary

```mermaid
flowchart TB
  subgraph Before
    A1[actions/cache-deps] --> B1["node_modules + berry store + types<br/>(custom key)"]
    B1 --> C1{cache-hit?}
    C1 -->|yes| D1[skip install + postinstall — stale risk]
    C1 -->|no| E1[yarn install + explicit after-install]
    F1[setup-node cache:yarn] -.-> G1["berry store<br/>(duplicate restore)"]
  end
  subgraph After
    A2[setup-node cache:yarn] --> B2[berry store only]
    H2[actions/cache] --> I2[install-state.gz]
    A2 --> E2[yarn install --immutable]
    E2 --> F2[workspace postinstall: generate-types + fetch-chains]
    E2 -.->|verify_cache=true| G2["--check-cache: verify tarballs<br/>(web release only)"]
    J2[actions/cypress] --> K2[Cypress binary cache]
    L2[actions/cache-deps] --> M2["Next.js build cache only"]
  end
```

## Checklist

- [x] I've documented how it affects the analytics (if at all) 📊 — no analytics impact
- [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻 — CI-only change, validated by the workflow runs on this PR


PooyaRaki:
@claude review

compojoom:
@claude review

PooyaRaki:
@claude review

Zalens (approved):
Ok so what you did here. No more cached node_modules, always reinstall, and --check-cache on release. Meaning a poisoned cache just fails the checksum against the lockfile instead of making it into prod. LGTM 👍

But as discussed, maybe request a review also from infra, it can be a good idea.

Fbartoli:
Can be merged

LucieFaire (approved):

PR #7894, opened 2026-05-20
fix(security-hub): require strict majority in account setup check

clovisdasilvaneto:
> Two of four is half, not more —
> half won't carry a vote.
> Bump to floor over two, plus one.

## What it solves

Resolves: [WA-2355](https://linear.app/safe-global/issue/WA-2355/account-setup-check-shows-strong-for-24-threshold-below-simple)

The Security Hub "Account setup" check rated a 2/4 Safe as **"Signing threshold is strong"** (Clear / Low) even though 2-of-4 is below simple majority. The bug was visible on the matic 2/4 Safe; the 2/17 case was already correctly flagged as Partial because that comparison happened to work out.

## How this PR fixes it

`accountSetupScanner` computed `simpleMajority = Math.ceil(ownerCount / 2)`, which equals **half** for even N (e.g. `ceil(4/2) = 2`). "Simple majority" means **strictly more than half**, so the correct formula is `Math.floor(ownerCount / 2) + 1`.

Changed both call sites:

- The `< simpleMajority` branch — controls whether the check is Partial vs Clear.
- The `threshold === 1` branch — the recommended threshold ("Increase the threshold to at least X of Y") now lands at strict majority, so following the recommendation actually moves the check to Clear. The previous `Math.max(..., 2)` guard against "raise to 1 of 2" is no longer needed (the new formula already returns ≥ 2 for any `ownerCount ≥ 2`).

Effect on bucketing:

| Safe | Before | After |
| ---- | ------ | ----- |
| 2/4  | Clear / Low / 100 ❌ | **Partial / Medium / 60** ✅ |
| 3/4  | Clear / Low / 100 | Clear / Low / 100 (unchanged) |
| 2/3  | Clear / Low / 100 | Clear / Low / 100 (unchanged) |
| 2/5  | Partial / Medium / 60 | Partial / Medium / 60 (unchanged) |
| 2/17 | Partial / Medium / 60 | Partial / Medium / 60 (unchanged) |
| 1/4  | Issue / Critical, recommend 2 of 4 | Issue / Critical, **recommend 3 of 4** |
| 1/2  | Issue / Critical, recommend 2 of 2 | Issue / Critical, recommend 2 of 2 (unchanged) |

The status/severity/score values from the requirements matrix are otherwise untouched — only the definition of "simple majority" changed.

## How to test it

1. Open a Safe with **2 of 4** threshold (e.g. the matic Safe from the ticket) and go to the Security Hub.
2. The "Account setup" card should now show **"Signing threshold is low"** (Partial / Medium) with the recommendation _"Consider increasing the threshold to 3 of 4 for stronger security."_
3. Verify a 3/4 Safe still shows **"Signing threshold is strong"** (Clear / Low).
4. Verify a 2/3 Safe still shows Clear (regression).
5. Verify a 2/17 Safe still shows Partial (regression).
6. Unit tests: `yarn workspace @safe-global/web test apps/web/src/features/security/data/scanners/__tests__/accountSetup.test.ts`.

## Affected flows

- Security Hub → Account setup check (status / severity / recommendation text).

## Blast radius

- `accountSetupScanner` is the only consumer of the `simpleMajority` calculation; no other scanner or UI reads it.
- Downstream contract (`status`, `severity`, `score`, `evidence`, `remediation`) is unchanged.
- Aggregate Security Hub score for affected Safes shifts because Account setup now scores 60 instead of 100 for 2/4-shaped setups — expected and desired.
- No API, route, persistence, feature flag, chain config, or mobile package change.

## Risks / not checked

- Did not manually verify in the production Security Hub UI — only ran unit tests locally. Suggest a QA pass on the exact 2/4 matic Safe from the ticket.
- Did not check Storybook stories for the Account setup card; the scanner output shape is unchanged, so stories should still render.
- Did not exercise mobile — shared `packages/` weren't touched.

## Visual summary

```mermaid
flowchart LR
  A["ownerCount, threshold"] --> B{"ownerCount == 1"}
  B -- yes --> I1["Issue / Critical / 10"]
  B -- no --> C{"threshold == 1"}
  C -- yes --> I2["Issue / Critical / 15<br/>recommend floor(N/2)+1 of N"]
  C -- no --> D{"threshold &lt; floor(N/2)+1<br/>(strict simple majority)"}
  D -- yes --> P["Partial / Medium / 60<br/>recommend floor(N/2)+1 of N"]
  D -- no --> CL["Clear / Low / 100"]

  style P fill:#fff3cd,stroke:#ffeeba
  style CL fill:#d4edda,stroke:#c3e6cb
```

## Checklist

- [ ] I've tested the branch on mobile 📱 (N/A — web-only scanner)
- [x] I've documented how it affects the analytics (if at all) 📊 (no analytics impact)

---

## CLA signature


apps/web/src/features/security/data/scanners/accountSetup.ts
```diff
@@ -25,6 +25,9 @@ export const accountSetupScanner: SecurityScanner = {
       { label: 'Threshold', value: `${threshold} of ${ownerCount}` },
     ]
 
+    // Simple majority = strictly more than half. For even N this is N/2 + 1, not N/2.
+    const simpleMajority = Math.floor(ownerCount / 2) + 1
```
tmjssz, line 29: Looks good, with the only exception for ownerCount = 2. In that case the result would be 2, so we would show a warning with the recommendation to increase the signer number to 2/2. Is that expected? I think it's always better to keep a Safe with ownerCount > threshold.

clovisdasilvaneto: Good catch — fixed in 55414da52. For a 1/2 Safe we no longer recommend going to 2/2 (that would create a single point of failure where losing any one key permanently locks the Safe). Special-cased `ownerCount === 2` in the `threshold === 1` branch to recommend adding another signer instead, so `ownerCount > threshold` is preserved. Score/severity stay the same (Issue / Critical / 15) — only the remediation text changes. All other 1/N cases keep the existing `floor(N/2)+1 of N` recommendation.

tmjssz (approved):

PR #7901, opened 2026-05-21
fix(mobile): detect biometry-invalidated wrapping key and route to re-import

### LINKED ISSUE #7835: [iOS] unable to sign using imported signer post migration to new phone
## Bug description
After migrating to a new iphone, I'm unable to use an existing imported signer. When I try to sign something I get the error: "Cannot sign the transaction. Failed to retrieve private key". When I try to view the private key, it says: "Biometric authentication failed. Please try again". When I try to delete the private key, it says: "Cannot delete private key. Private key not found for the specified address"

I have deleted the app and started from fresh, re-adding the signer but I get the exact same errors when trying to complete the same tasks.

When I re-add the new signer, the workflow output suggests everything is fine: "Your signer is ready!"

The app is permissioned for faceID. I am able to sign transactions and access/delete the private key on the old phone with the same configuration.

## Environment

- Wallet: iOS safe{wallet}
- Chain: Any

## Steps to reproduce

1.  Use iOS migration to migrate to a new phone (iphone XS -> iphone 15)
2. try to sign a transaction, or view/delete a private key

Start from scratch:
1. delete app and initialise app from new, import safe and then add signer 
2. get same errors

## Expected result
transaction is signed / private key is displayed / private key is deleted

## Obtained result
Errors (above)

compojoom:
> Face ID's reset its enrollment,
> Enclave keys forget the key,
> Tell the user "re-import."

## What it solves

Resolves: #7835
Resolves: https://linear.app/safe-global/issue/WA-2300 

When iOS Face ID is re-enrolled or the Android device PIN/lock-screen credential is changed/removed, the OS invalidates the wrapping key tied to the user's previous biometrics. The encrypted private-key blob in the iOS Keychain / Android Keystore is still there, but it can no longer be decrypted. Today every consumer of `KeyStorageService.getPrivateKey()` collapses this into a generic `undefined`, so:

- **Sign** fails with `"Cannot sign the transaction. Failed to retrieve private key"` — user has no idea what happened.
- **View private key** shows `"Biometric authentication failed. Please try again"` — misleading; biometrics worked, the key is just unusable.
- **Delete private key** / **Delete account** fail with `"Private key not found"` / `"Failed to clean up 1 out of 1 private keys"` — the cleanup path is gated on a successful decrypt, so the user can't recover without uninstalling the app.

Re-importing the signer also "succeeded" (UI said "Your signer is ready!") but signing still failed, because the iOS path reused the orphan Secure Enclave key reference left behind by Quick Start / iCloud Keychain sync.

## How this PR fixes it

1. **Typed-error classification** inside `KeyStorageService.getPrivateKey`. New `BiometryInvalidationError` is thrown when the underlying decrypt failure matches one of the platform-specific fingerprints documented in `errors.ts` — anchored on `AKSError=-536362999` (`kAKSReturnPolicyInvalid`) and `LAErrorBiometryNotEnrolled` on iOS, and on `BiometricPrompt` codes 9/11/14 + `KeyPermanentlyInvalidatedException` on Android. Transient codes (user-cancel, lockout) deliberately don't match.
2. **Self-heal on store** — `storeKey`'s existing retry path now triggers on iOS too (was Android-only via a literal string match).
3. **Sign + execute flows** route to their standard error screens (`/signing-error`, `/execution-error`) with a clear "re-import the signer" message instead of a mid-flow popup.
4. **View PK** keeps a simple OK alert (it's a settings-context action where the standard error screen would feel heavy).
5. **`cleanupSinglePrivateKey` ungated** — when the key is unusable, drop the keychain entry without auth and remove the Redux signer so the user can delete and re-add. Server-side delegate revocation is skipped (we can't sign one without the key); follow-up tracked elsewhere.
6. **Defensive `removePrivateKey` before `storePrivateKey`** in `useImportPrivateKey` — re-import on iOS now always provisions a fresh wrapping key instead of reusing an orphan reference.
7. **Side fix to biometrics-setup CTA**: on Android, "Open settings" now uses the `android.settings.BIOMETRIC_ENROLL` intent to land directly on biometric enrollment instead of the app permissions page. iOS keeps `app-settings:` (Apple doesn't expose a public deep link to Face ID & Passcode).

## How to test it

**Android device:**

1. Boot an API 30+ device with a screen lock (PIN) and one enrolled fingerprint. .
2. Build the app and import a private-key signer. Sign one transaction to confirm it works.
3. Settings → Security → remove the PIN (`None`). This destroys the LSKF and immediately invalidates every auth-bound Keystore entry.
4. Re-open the app:
   - **Sign tx** → routes to `/signing-error` with the re-import message (not the generic "failed to retrieve" copy).
   - **View private key** → OK alert "Signer needs to be re-imported".
   - **Delete private key** → succeeds; signer disappears from Redux + Keystore.
   - **Delete account** with that signer → no longer blocked by "Failed to clean up 1 out of 1 private keys".
5. Re-import the same private key — works on the first try (defensive `removePrivateKey` clears the invalidated entry first). Note that re-import on Android already self-healed via the existing `KeyPermanentlyInvalidatedException` path; this PR makes the same retry trigger on iOS.

**iOS physical device:**

1. Import a private-key signer; sign a transaction.
2. Settings → Face ID & Passcode → Reset Face ID (or enroll "Alternate Appearance"; iOS 15+ counts mask enrollment as a biometry-set change).
3. Repeat the sign / view / delete checks above. Expected behavior matches Android.

**User-cancel sanity checks:**

- Cancel the biometric prompt manually on either platform → no rotation alert; existing UX preserved.
- iOS `LAErrorBiometryLockout(-8)` (5+ failed Face ID attempts) → no rotation alert; user retries with passcode.

**Android biometrics-setup CTA:**

- On Android setup with no fingerprints enrolled, in the app open the biometrics-opt-in flow → "Open settings" lands on the biometric enrollment screen (not the app permissions page).

**Unit tests:**

```
yarn workspace @safe-global/mobile test --testPathPattern="key-storage|useTransactionSigning|editAccountHelpers|useBiometrics|useImportPrivateKey"
```

22 classifier tests + 24 service tests + container/hook/cleanup tests = 142 directly related, all green. Full mobile suite (271 suites, 2355 tests) also green.

## Affected flows

- Sign transaction (PK signer) — failure path
- Execute transaction (PK signer) — failure path
- View private key from signer detail
- Delete private key from signer detail
- Delete safe that owns a PK signer
- Import private key (defensive cleanup before store)
- Biometrics opt-in / setup CTA

## Blast radius

**Shared surfaces touched:**

- `KeyStorageService.getPrivateKey` now throws `BiometryInvalidationError` for classified failures. Known direct consumers: `getPrivateKey` in `apps/mobile/src/hooks/useSign/useSign.ts` (used by sign tx, execute tx via `privateKeyExecutor`), `cleanupSinglePrivateKey` in `editAccountHelpers`, `PrivateKey.container`. Each is updated.
- `KeyStorageService.storeKey` self-heal previously only matched Android-style errors. Now matches iOS too via the shared classifier. No new path for Android.
- `keyStorageService.removePrivateKey(_, { requireAuthentication: false })` is now called from `useImportPrivateKey` before every store. This is keychain `SecItemDelete` / Keystore `deleteEntry` — auth-free by design.
- New Datadog RUM error: `BiometryInvalidationError` (name + JS stack + context with `userId`, `location`, `platform`). One emission per `getPrivateKey` failure that matches. The signer address is already in Datadog via existing screen-view telemetry (e.g. `/signers/0x…/private-key`), so no incremental PII surface.

**Not touched:** Ledger flow, WalletConnect signers, RTK Query endpoints, store slices, multi-chain handling, Safe SDK, feature flags, routes/layouts, web app, shared packages.

## Risks / not checked

- The classifier matches `react-native-device-crypto` error string shapes. If the upstream library reformats its rejections (e.g. `9 -` → `[9]`), classification silently degrades to "transient" and we fall back to the generic error screen — same UX as today's bug. Datadog telemetry on hits will surface drift.
- Did not run the web E2E suite — change is mobile-only.
- iOS reproduction requires a physical device (simulator has no Secure Enclave). Verified the classifier patterns against the exact error reported in #7835 via unit tests; final UX needs device validation.

## Visual summary

```mermaid
flowchart TB
    A[DeviceCrypto.decrypt rejects in getKey] --> B{isBiometryInvalidationError}
    B -- yes --> C[KeyStorageService.getPrivateKey throws BiometryInvalidationError + DdRum.addError]
    B -- no --> D[returns undefined / rethrows]

    C --> E[Sign flow useTransactionSigning]
    C --> F[Execute flow privateKeyExecutor]
    C --> G[View PK PrivateKey.container]
    C --> H[Cleanup editAccountHelpers]

    E --> I[/signing-error with re-import message/]
    F --> J[/execution-error with re-import message/]
    G --> K[OK Alert 'Signer needs to be re-imported']
    H --> L[Skip delegate revoke; remove Redux signer; success]

    M[useImportPrivateKey.handleImport] --> N[removePrivateKey w/o auth]
    N --> O[storePrivateKey provisions fresh wrapping key]
```

## Checklist

- [x] I've tested the branch on mobile 📱 (Android emulator full repro; iOS via unit tests against the captured error fingerprint from the issue)
- [x] I've documented how it affects the analytics (if at all) 📊 (new `BiometryInvalidationError` RUM event documented in Blast Radius)
- [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻 (22 classifier + 24 service + 13 sign hook + 32 cleanup + 27 biometrics tests)

---

## CLA signature

### DIFF
--- a/apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/helpers.ts
+++ b/apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/helpers.ts
@@ -3,6 +3,7 @@ import { FEATURES, hasFeature } from '@safe-global/utils/utils/chains'
 import { Chain } from '@safe-global/store/gateway/AUTO_GENERATED/chains'
 import { FeeParams } from '@/src/hooks/useFeeParams/useFeeParams'
 import { Signer } from '@/src/store/signersSlice'
+import { BIOMETRY_ROTATION_DESCRIPTION, BiometryInvalidationError } from '@/src/services/key-storage'
 
 /**
  * Execution path types for the confirm flow
@@ -96,6 +97,9 @@ export const determineExecutionPath = (
  * Extracts error message from unknown error
  */
 export const getErrorMessage = (error: unknown): string => {
+  if (error instanceof BiometryInvalidationError) {
+    return BIOMETRY_ROTATION_DESCRIPTION
+  }
   if (error instanceof Error) {
     return error.message
   }
--- a/apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/helpers.ts
+++ b/apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/helpers.ts
@@ -94,14 +94,16 @@ export const determineExecutionPath = (
 }
 
 /**
- * Extracts error message from unknown error
+ * Extracts a user-facing message from an unknown error, mapping biometry
+ * invalidation to the shared re-import copy and falling back to `fallback`
+ * for non-Error throwables.
  */
-export const getErrorMessage = (error: unknown): string => {
+export const getErrorMessage = (error: unknown, fallback = 'Failed to execute transaction'): string => {
   if (error instanceof BiometryInvalidationError) {
     return BIOMETRY_ROTATION_DESCRIPTION
   }
   if (error instanceof Error) {
     return error.message
   }
-  return 'Failed to execute transaction'
+  return fallback
 }
--- a/apps/mobile/src/features/AccountsSheet/AccountItem/utils/__tests__/editAccountHelpers.test.ts
+++ b/apps/mobile/src/features/AccountsSheet/AccountItem/utils/__tests__/editAccountHelpers.test.ts
@@ -224,7 +224,7 @@ describe('editAccountHelpers', () => {
       expect(mockDispatch).toHaveBeenCalledWith(removeSigner(mockAddress1))
     })
 
-    it('should cleanly remove the signer when the private key is missing', async () => {
+    it('should NOT wipe the signer when getPrivateKey returns undefined without invalidation', async () => {
       const mockDispatch = jest.fn() as unknown as AppDispatch
       const mockRemoveAllDelegatesForOwner = jest.fn()
 
@@ -232,10 +232,11 @@ describe('editAccountHelpers', () => {
 
       const result = await cleanupSinglePrivateKey(mockAddress1, mockRemoveAllDelegatesForOwner, mockDispatch)
 
-      expect(result.success).toBe(true)
+      expect(result.success).toBe(false)
+      expect(result.error?.type).toBe(ErrorType.STORAGE_ERROR)
       expect(mockRemoveAllDelegatesForOwner).not.toHaveBeenCalled()
-      expect(keyStorageService.removePrivateKey).toHaveBeenCalledWith(mockAddress1, { requireAuthentication: false })
-      expect(mockDispatch).toHaveBeenCalledWith(removeSigner(mockAddress1))
+      expect(keyStorageService.removePrivateKey).not.toHaveBeenCalled()
+      expect(mockDispatch).not.toHaveBeenCalled()
     })
 
     it('should handle delegate removal failure', async () => {
@@ -318,18 +319,17 @@ describe('editAccountHelpers', () => {
       expect(mockDispatch).not.toHaveBeenCalled()
     })
 
-    it('should cleanly remove the signer when the private key is missing', async () => {
+    it('should report STORAGE_ERROR when getPrivateKey returns undefined without invalidation', async () => {
       const mockDispatch = jest.fn() as unknown as AppDispatch
       const mockRemoveAllDelegatesForOwner = jest.fn()
 
       ;(keyStorageService.getPrivateKey as jest.Mock).mockResolvedValue(null)
-      ;(keyStorageService.removePrivateKey as jest.Mock).mockResolvedValue(undefined)
 
       await cleanupPrivateKeysForOwners([mockAddress1], mockRemoveAllDelegatesForOwner, mockDispatch)
 
       expect(mockRemoveAllDelegatesForOwner).not.toHaveBeenCalled()
-      expect(keyStorageService.removePrivateKey).toHaveBeenCalledWith(mockAddress1, { requireAuthentication: false })
-      expect(mockDispatch).toHaveBeenCalledWith(removeSigner(mockAddress1))
+      expect(keyStorageService.removePrivateKey).not.toHaveBeenCalled()
+      expect(mockDispatch).not.toHaveBeenCalled()
     })
 
     it('should cleanly remove the signer when the wrapping key has been biometry-invalidated', async () => {
--- a/apps/mobile/src/features/AccountsSheet/AccountItem/utils/__tests__/editAccountHelpers.test.ts
+++ b/apps/mobile/src/features/AccountsSheet/AccountItem/utils/__tests__/editAccountHelpers.test.ts
@@ -12,22 +12,28 @@ import {
 import { ErrorType } from '@/src/utils/errors'
 import { Address } from '@/src/types/address'
 import { AppDispatch } from '@/src/store'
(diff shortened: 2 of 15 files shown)


compojoom:
@claude review

compojoom:
Addressed in [ffbb101](https://github.com/safe-global/safe-wallet-monorepo/pull/7901/commits/ffbb101d0) and [49d6fc5](https://github.com/safe-global/safe-wallet-monorepo/pull/7901/commits/49d6fc584):

**Both reviewers' P1s**
- `cleanupSinglePrivateKey` now only force-wipes when a `BiometryInvalidationError` was actually caught. Transient failures (user cancel, lockout, unknown decrypt errors) restore the prior `STORAGE_ERROR` behavior. Regression test added.
- Dropped the unconditional pre-delete in `useImportPrivateKey`. Replaced with a post-store decrypt probe inside `storeKey` itself — iOS now surfaces orphan SE state during import (mirroring Android, where symmetric encrypt always required auth). The existing `handleKeyInvalidation` retry path handles recovery atomically. No more remove-then-cancel-store data-loss risk.

**Claude's other items**
- `storeTries` is no longer instance state; the retry counter is passed as a parameter to `storeKey`. Singleton no longer carries mutable retry state.
- `Logger.error` was moved out of the classified-invalidation branch; invalidation is now logged at `warn` (expected state), with `error` reserved for unclassified failures.
- `BiometryInvalidationError` now uses the native `super(msg, { cause })` form so tooling that walks the cause chain sees the underlying error.

**Skipped**
- Android regex broadening (`/\b(?:9|11|14)\s*[-:]/`): current strict patterns are explicit about which OEM error codes they trust; broadening adds false-positive risk. If `react-native-device-crypto` reformats its rejections, classification gracefully degrades to the generic error screen — and the new Datadog telemetry on classified hits will give us a signal to extend the matcher.
- `Linking.sendIntent` minSdk check: the app requires sdk > 30.

apps/mobile/src/features/ExecuteTx/components/ReviewAndExecute/helpers.ts
```diff
@@ -96,6 +97,9 @@ export const determineExecutionPath = (
  * Extracts error message from unknown error
  */
 export const getErrorMessage = (error: unknown): string => {
```
tmjssz: Could be useful to make this function more generic by adding a second param like `fallback`. Then it could be called like this:
```ts
getErrorMessage(err, 'Failed to execute transaction')
```
And we could also use it here -> [ReviewAndConfirmContainer.tsx](https://github.com/safe-global/safe-wallet-monorepo/blob/49d6fc5844e91f29933305d62f553c56dab97a97/apps/mobile/src/features/ConfirmTx/components/ReviewAndConfirm/ReviewAndConfirmContainer.tsx#L64-L69)

compojoom: Good call — done in 58205d8. `getErrorMessage(error, fallback?)` now takes an optional fallback (defaults to "Failed to execute transaction"), and `ReviewAndConfirmContainer` calls `getErrorMessage(err, "Failed to sign transaction")` instead of the inline ternary. Both error screens now map `BiometryInvalidationError` through the one helper.

tmjssz (approved):

PR #7909, opened 2026-05-21
chore: regenerate schema

vmosharova:
## What it solves

Resolves: Regenerates schema, schema hash and AUTO_GENERATED types.

---

## CLA signature


katspaugh (approved):
Thanks!

PR #7911, opened 2026-05-21
feat(spaces): new onboarding UI - 4 steps

tongyaosafe:
> Four steps reskinned, side-panel preview,
> form logic untouched, tests still pass.
> Mockup balance now multi-chain, fades hide the edge.

## What it solves

Brings the 4-step Spaces onboarding flow (Create Space → Select Safes → Invite Members → Survey) in line with the new Vercel prototype design. The previous UI used the old MUI shell, mixed paddings and breakpoints, and had no preview of what the Space will look like once created.

Resolves: [WA-2395](https://linear.app/safe-global/issue/WA-2395/re-style-space-onboarding-flow-4-steps-to-new-design-system) (sub-task of [WA-2190](https://linear.app/safe-global/issue/WA-2190/re-style-all-workspace-tabs-to-new-design-system))

## How this PR fixes it

- **New shared `OnboardingLayout`** with a single source of truth for the page chrome (Safe logo, step counter, main column, footer, side-panel slot). All four steps consume it instead of each rolling their own layout.
- **Side-panel mockup (`SafeAppMockup` + `mockup/*`)** renders a stylized peek at the real Space dashboard on `xl+` viewports: workspace switcher (with motion ring on Step 1), skeleton header/balance/chips, and an Accounts widget populated from whatever Safes the user has selected so far. Total balance uses the same bulk RTK Query pattern as the real Dashboard's `AggregatedBalance` so multi-chain Safes sum correctly across networks. Below `xl` the side panel is hidden and the gateway query is skipped.
- **Form logic, navigation, RTK Query mutations, and analytics are unchanged** — this PR is a visual + structural reskin only. `useOnboardingSubmit`, `useInviteForm`, `useSpaceSubmit`, and the survey state machine are touched only where the audit found real issues (typed error narrowing, redundant comments, file splits).
- **Post-review polish**: `SafeAppMockup{Account,Props}` moved to a leaf `mockup/types.ts` to break the type back-edge from `mockup/*` children to their parent; `@ts-ignore` in `useInviteForm` replaced with the repo's `getRtkQueryErrorMessage` helper; balance query gated on `xl` via a new `useIsXlViewport` so phones/tablets don't pay for an invisible widget.

## How to test it

1. `yarn workspace @safe-global/web dev`, navigate to `/welcome/create-space`.
2. Walk all four steps end-to-end. On `xl+` (≥ 1280px) the right column shows the app-preview mockup; below `xl` only the main column renders.
3. Step 2: select a few Safes, including at least one multi-chain Safe. The top-left balance in the mockup should sum across **all selected chains**, not just one. The Accounts widget should list the selected safes with names + addresses + per-row fiat.
4. Step 3 / Step 4: navigate back from a later step to Step 2. The mockup should stay populated (it falls back to the persisted Space safes when the form has not yet rehydrated).
5. Resize the viewport across `< xl` ↔ `≥ xl`. The balance query should only fire when the side panel is visible.
6. Toggle light/dark mode. All copy + icons should remain legible; skeleton bars use the theme-aware `--color-background-skeleton` token.
7. Step 4 survey: with `surveys.is_active = false` on the backend, the page should silently redirect to the Space dashboard instead of showing an error.

## Affected flows

- **Create Space** — Step 1 of the welcome onboarding (`/welcome/create-space`).
- **Select Safes** — Step 2 (`/welcome/select-safes`), including the select-all section toggles and the multi-chain Safe selection cards.
- **Invite Members** — Step 3 (`/welcome/invite-members`), including ENS resolution + role selection.
- **Survey** — Step 4 (`/welcome/survey`), backed by the survey state endpoint with 404 → silent skip.
- **Back-navigation between steps** — form state is preserved across all transitions; the side-panel mockup reflects the live form state on Step 2 and the persisted Space safes on Steps 3 / 4.

## Blast radius

- **`OnboardingLayout` / `SafeAppMockup` / `mockup/*`** — new shared components inside `features/spaces/components/OnboardingLayout/`. Only the four onboarding pages consume them; no other Spaces surfaces are affected.
- **`useGetMultipleSafeOverviewsQuery`** — same RTK Query endpoint already used by the real Dashboard's `AggregatedBalance`. New consumer in the side-panel mockup, gated by `useIsXlViewport`.
- **`useInviteForm.ts`** — `@ts-ignore` replaced with `getRtkQueryErrorMessage`; same error surface to the user, just typed. No change to the invite mutation, ENS validation, or analytics.
- **`SelectSafesOnboarding/utils/deriveSelectedAccounts.ts`** (extracted) — pure helpers; no behavior change vs the inline code, but now covered by unit tests.
- **No changes to**: `useOnboardingSubmit` (form submit logic), `useSpaceSubmit`, RTK Query API contracts, persisted slices, feature flags, mobile (`apps/mobile/`), or any non-Spaces feature.

## Risks / not checked

- **E2E (Cypress)** — no new specs added; the existing Spaces onboarding spec still runs. Did not exercise the full happy path via Cypress.
- **Survey-disabled flow** — verified the 404 → silent redirect logic via unit tests; not exercised against a real backend with `surveys.is_active = false`.
- **Wallet-not-connected flow on Step 2** — the existing `ConnectWalletPrompt` + Skip button still render, but I did not regression-test the full connect-then-resume journey.
- **Analytics** — no analytics events were added, removed, or renamed. Did not re-verify the existing `WORKSPACE_MEMBER_INVITE_SENT` tracking on Step 3 beyond running the existing unit test.

## Visual summary

```mermaid
flowchart LR
  subgraph Onboarding["4-step onboarding (NEW)"]
    L["OnboardingLayout<br/>(shared chrome)"]
    L --> S1["Step 1<br/>CreateSpace"]
    L --> S2["Step 2<br/>SelectSafes"]
    L --> S3["Step 3<br/>InviteMembers"]
    L --> S4["Step 4<br/>Survey"]
    L --> SP["SpaceSidePanel<br/>(xl+ only)"]
    SP --> M["SafeAppMockup"]
    M --> MS["MockupSidebar<br/>(workspace switcher)"]
    M --> MC["MockupContent<br/>(balance + Accounts widget)"]
    M -. skip query .-> Q["useGetMultipleSafeOverviewsQuery"]
    Q -. when xl .-> M
  end

  subgraph Untouched["Logic NOT changed"]
    F1[useOnboardingSubmit]
    F2[useInviteForm*]
    F3[useSpaceSubmit]
    F4[Survey state machine]
  end

  S2 -.uses.-> F1
  S3 -.uses.-> F2
  S1 -.uses.-> F3
  S4 -.uses.-> F4
```

`* useInviteForm: @ts-ignore replaced with getRtkQueryErrorMessage; no behavior change.`

Screenshots: please attach before/after captures for each of the four steps at light + dark mode, desktop (≥ xl) and tablet (< xl).

## Checklist

- [x] I've documented how it affects the analytics (if at all) 📊 — _no analytics events added, removed, or renamed; existing `WORKSPACE_MEMBER_INVITE_SENT` tracking on Step 3 is unchanged_
- [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻 — _120 onboarding unit tests pass, including 6 new test files for the extracted helpers + the xl-gated balance hook_

---

## CLA signature


tongyaosafe:
@claude review

tongyaosafe:
> **Claude finished @tongyaosafe's task in 3m 11s** —— [View job](https://github.com/safe-global/safe-wallet-monorepo/actions/runs/26455584572)
> 
> ### Claude PR Review
> * [x]  Read PR diff vs base branch
> * [x]  Review new OnboardingLayout files
> * [x]  Review step component rewrites
> * [x]  Post review feedback
> 
> PR is well-scoped as a POC and the description already calls out most known debts. Below are the items I'd want addressed if/when this is promoted past POC.
> 
> ### Bugs / correctness
> **1. `SafeAppMockup.tsx:30-33` — unguarded `Number(o.fiatTotal)` can poison the aggregate.**
> 
> ```ts
> const totalFiat = useMemo(
>   () => (safeOverviews ?? []).reduce((sum, o) => sum + Number(o.fiatTotal), 0),
>   [safeOverviews],
> )
> ```
> 
> Per-row `sumOverviewsForAddress` already does `Number(o.fiatTotal) || 0` and has an explicit "treats non-numeric fiatTotal as 0" test (`MockupAccountRow.test.ts:44-47`). The aggregated total path does not — a single `"NaN"`/`""`/`null` from the gateway will render `$NaN` in the hero number above the Accounts widget. Use the same guarded coercion (or extract a shared helper).
> 
> **2. `SafeAppMockup.tsx:22` — `balanceSafes ?? []` allocates a new array each render → unstable RTK Query arg.**
> 
> ```ts
> const safesForQuery = balanceSafes ?? []
> ```
> 
> When `balanceSafes` is `undefined` (e.g. first render before parents memoize), every render passes a fresh `[]` reference to `useGetMultipleSafeOverviewsQuery`. RTK Query's selector does shallow-equal on args, so a new `[]` reference each tick churns subscription work. Lift the empty default to a module-level constant or wrap in `useMemo`.
> 
> **3. `useSafeNameLookup.test.ts` — tests likely don't exercise the production code path.**
> 
> Coverage report shows `useSafeNameLookup.ts` at 22% statements / 0% branches despite a 5-case test file. The `initialReduxState: { addressBook }` shape passed to `renderHook` probably doesn't match what `selectAllAddressBooks` reads (the selector almost certainly walks a chain-keyed sub-state). Net effect: every test hits the `if (!allAddressBooks) return map` early-exit and asserts on an empty map. The 33% expect-an-empty-map cases pass for the wrong reason; the 67% expect-populated-map cases would fail if the selector were wired correctly. Verify by adding a temporary `console.log(allAddressBooks)` inside the hook — if it's always `undefined`, the test fixture shape needs to be fixed.
> 
> **4. `useIsXlViewport.ts:8-10` — initial `false` causes a one-render skip of the balance query even on `xl` viewports.**
> 
> The hook returns `false` on the first render and only updates after `useEffect`. So `useGetMultipleSafeOverviewsQuery({ skip: !isXl })` skips on mount for everyone, then re-subscribes on the next tick. Cheap fix:
> 
> ```ts
> const [isXl, setIsXl] = useState(() =>
>   typeof window !== 'undefined' && window.matchMedia?.(XL_QUERY).matches === true,
> )
> ```
> 
> Acceptable for SSR (initial paint will still be `false` server-side, but the very first client render reads the real value, so there's no extra re-render or flash for hydrated XL users).
> 
> ### Maintainability
> **5. Duplicate account-derivation logic.** `OnboardingLayout/deriveSidePanelAccountsFromSpace.ts` and `SelectSafesOnboarding/utils/deriveSelectedAccounts.ts` are two implementations of the same idea (dedupe by address, multichain-aware, name-fallback). PR description acknowledges this — flagging so it doesn't get lost on rebase.
> 
> **6. `SpaceSidePanel.tsx` is a 1-line passthrough to `SafeAppMockup`.** Either inline it, or give it a real responsibility (it's named like a layout primitive but does nothing).
> 
> **7. `InviteMembersOnboarding/index.tsx:64` — `setTimeout(() => trigger('members'), 0)` is a code smell.** Prefer `Promise.resolve().then(() => trigger('members'))` if you really need to defer past the RHF state update, but in most cases `await remove(index); trigger('members')` (or just calling `trigger` synchronously, since `remove` is sync) works.
> 
> ### Nits
> * `MockupContent.tsx:69` — `[88, 70, 80].map((w, i) => ...)` would read better as a `SKELETON_ROW_WIDTHS` constant.
> * `SafeAppMockup.tsx:43` — `left-50 top-50` (Tailwind v4 dynamic spacing) — PR description already calls this out as a v4-dependency risk; worth a brief code comment at the call site so a future reader doesn't "fix" it.
> * `SurveyOnboarding/index.tsx:122-126` — empty `catch {}` block; consider at least a `console.warn` so failures aren't completely silent in dev.
> * `CreateSpaceOnboarding/index.tsx:78-81` — the `hasUserEdited` / `existingSpace.name` fallback dance is correct but dense. The inline comments help; a small `deriveSwitcherState({ hasUserEdited, watchedName, existingName })` would make it testable in isolation.
> 
> ### Test coverage
> The PR description's "no unit tests" line is stale — `deriveSidePanelAccountsFromSpace.test.ts`, `useSafeNameLookup.test.ts`, `MockupAccountRow.test.ts`, `StepCounter.test.tsx`, and `deriveSelectedAccounts.test.ts` all exist. Update the description (and the "Risks" section's reference to a deleted multichain regression test, which now appears to be back).
> 
> Coverage is still soft on `deriveSidePanelAccountsFromSpace.ts` (45% statements) and `useSafeNameLookup.ts` (22%) — see bug #3 for the likely cause of the hook number.
> 
> • Branch: `feat/space-onboarding-ui-swap-poc`

1. not a bug. Show $0 would make users scary. Showing NaN is better.
2. fixed
3. wrong report. checked coverage: 94.44% stmts / 71.42% branch / 100% lines
4. fixed.
5. not duplicated. Two different adapter. 
6. fixed.
7. `remove()` returns synchronously, but its state update goes through React's batched dispatch — a synchronous `trigger('members')` on the next line sees the pre-remove snapshot and validates against the row we just deleted. Deferring past the render cycle is the standard `useFieldArray` pattern; `setTimeout(0)` and `Promise.resolve().then(...)` are equivalent here. Closing as won't-fix.

apps/web/src/features/spaces/components/CreateSpaceOnboarding/hooks/useExistingSpace.ts
```diff
@@ -26,6 +26,7 @@ const useExistingSpace = (setValue: UseFormSetValue<{ name: string }>) => {
     spaceId,
     isEditMode,
     isSpaceLoading,
+    existingSpace,
   }
```
vmosharova, line 30: This hook is not covered with unit tests - might make sense to create a small text file.

apps/web/src/features/spaces/components/OnboardingLayout/mockup/useIsXlViewport.ts
```diff
@@ -0,0 +1,24 @@
...
+    if (typeof window === 'undefined' || !window.matchMedia) return
+    const mql = window.matchMedia(XL_QUERY)
+    setIsXl(mql.matches)
+    const onChange = (e: MediaQueryListEvent) => setIsXl(e.matches)
+    mql.addEventListener('change', onChange)
+    return () => mql.removeEventListener('change', onChange)
+  }, [])
+
+  return isXl
+}
```
vmosharova, line 24: Is this hook used only in mockups? If not - let's move it out of `mockup` folder.

apps/web/src/features/spaces/components/OnboardingLayout/mockup/SafeAppMockup.tsx
```diff
@@ -0,0 +1,59 @@
...
+            totalFiat={totalFiat}
+            highlight={highlight}
+          />
+        </div>
+      </div>
+    </div>
+  )
+}
+
+export default SafeAppMockup
```
vmosharova, line 59: It is a mockup, so it makes sense to move it to the mockup folder.

vmosharova (changes requested):

apps/web/src/features/spaces/components/OnboardingLayout/mockup/SafeAppMockup.tsx
vmosharova: ...Maybe to mockup/index.tsx and re-export from the barrell.

apps/web/src/features/spaces/components/OnboardingLayout/mockup/deriveSidePanelAccountsFromSpace.ts
```diff
@@ -0,0 +1,36 @@
...
+  for (const safe of allSafes) {
+    if (isMultiChainSafeItem(safe)) {
+      pushOnce(safe.address, safe.name, safe.safes[0])
+    } else {
+      pushOnce(safe.address, safe.name, safe)
+    }
+  }
+
+  return accounts
+}
```
vmosharova, line 36: This file includes only mockup-related logic; probably makes sense to move it to mockup fodler as well.

apps/web/src/features/spaces/components/OnboardingLayout/mockup/useSafeNameLookup.ts
```diff
@@ -0,0 +1,25 @@
...
+        if (typeof name === 'string' && name.trim()) {
+          const key = address.toLowerCase()
+          if (!map.has(key)) map.set(key, name)
+        }
+      }
+    }
+
+    return map
+  }, [allAddressBooks])
+}
```
vmosharova, line 25: Same here: This file includes only mockup-related logic for data feeded to the mockup; probably makes sense to move it to mockup fodler as well.

vmosharova: This way we can co-locate all mockup code in one place.

vmosharova (changes requested):
Looks great so far! Let's move some files - this way we can co-locate all mockup code and mockup data-related helpers in one place.

apps/web/src/features/spaces/components/OnboardingLayout/mockup/useIsXlViewport.ts
tongyaosafe: Yes. It is only in mockups.

vmosharova (approved):
👍🏻

PR #7933, opened 2026-05-27
fix(web): stabilize security hub score and harden re-scan

clovisdasilvaneto:
> Partial scans no longer commit,
> re-scan pulls fresh and waits a beat,
> hidden tabs hold their countdown —
> the gauge stops flip-flopping.

## What it solves

Resolves three Security Hub issues around the workspace score gauge:

1. **Score changed on re-scan of an unchanged account.** The overall score is `round(passing / applicableCount * 100)`, computed over whatever checks produced a *definitive* result that scan. A scanner that threw, timed out, or whose Safe was bailed past was silently dropped, shrinking the denominator — so transient RPC/CGW jitter between scans moved the score even though nothing about the account changed.
2. **"Re-scan" re-ran scanners over cached data**, so it couldn't reflect fresh backend state.
3. **Switching browser tabs falsely reported `The last scan didn't finish`**, with a `[SecurityHub] scan context did not resolve … within Nms — skipping` warning in the console.

## How this PR fixes it

- **Don't commit partial scans** (`useAutoScan.ts`): a Safe's results are written to the gauge/cache only when *every* scanner produced a result (`Object.keys(results).length === total`). A partial run no longer overwrites the prior complete score; the queue still drains so nothing hangs. A `scanIncomplete` flag surfaces a note: *"The last scan didn't finish. Showing your most recent complete score — re-scan to update it."*
- **Forced backend refetch on manual re-scan** (`useSafeScanContext.ts`): a new `forceRefetch` param sets `refetchOnMountOrArgChange` on the scan's CGW queries (safes, master-copies, creation tx, safe overviews). Driven by an `isRescanning` signal that is true **only** for the user-triggered "Re-scan" button — automatic scans (load / queue change / space switch) stay on cache.
- **Minimum visible re-scan duration**: a manual re-scan holds the "Scanning…" state for at least 1s so a fast scan doesn't flash by.
- **Visibility-aware bail countdown**: the scan-context bail timer only counts down while `document.visibilityState === 'visible'`, re-arming on `visibilitychange`. A backgrounded tab (which throttles timers and defers renders) can no longer fire the bail before the still-in-flight query resolves. Genuine "ghost-deployed" stalls still bail after a real foreground window.

## How to test it

1. Open the Security tab for a workspace with at least one Safe; let the scan complete and note the score.
2. Click **Re-scan** repeatedly — the score stays stable for an unchanged account, and "Scanning…" is visible for ~1s each time.
3. With the network throttled, click **Re-scan** and confirm fresh requests for the scan's CGW endpoints go out.
4. Start a scan, switch to another browser tab for a few seconds, come back — no `scan context did not resolve … — skipping` warning and no false "scan didn't finish" note.

## Affected flows

- Security Hub workspace score gauge (`WorkspaceHealthCard`) — aggregate score + "scan incomplete" note.
- Security Hub auto-scan + manual re-scan queue (`useAutoScan` / `useAutoScanOrchestrator`).
- Per-Safe scan data fetch (`useSafeScanContext`) — manual re-scan now refetches from backend.

## Blast radius

- **`useSafeScanContext`** (shared by `useAutoScan` and the report drawer): added optional `forceRefetch` param (defaults `false`), so the drawer's behavior is unchanged. Sets `refetchOnMountOrArgChange` on `useSafesGetSafeV1Query`, `useChainsGetMasterCopiesV1Query`, `useTransactionsGetCreationTransactionV1Query`, `useGetSafeOverviewQuery`, `useGetMultipleSafeOverviewsQuery`.
- **`useAutoScan` / `AutoScanState`**: added `scanIncomplete` and `isRescanning`; `startScan` now takes `{ isManual?: boolean }`. Consumed only by `useAutoScanOrchestrator` → `SecurityHub`.
- **`setCachedScan` / `scanResultsCache`**: partial scans are no longer cached, so the drawer re-scans on open instead of reusing a partial result (acceptable).
- No API contract, feature flag, route, or shared-package (mobile) changes.

## Risks / not checked

- **Balances batch query not force-refreshed**: the `overviewMap` (balance / queued-count) from `useReconciledSpaceSafes` is passed in as `overviewData` and isn't part of `forceRefetch`, so those cosmetic fields stay cached on a manual re-scan. All *score-relevant* data does refresh.
- **Per-scanner 15s `withScannerTimeout` is not visibility-aware**: a tab left in the background long enough for a scanner to time out (after its context resolves) could still mark a Safe incomplete. Much longer window than the bail; not the reported symptom. Left as a follow-up.
- **Issue carried over from prior work**: a scanner that *resolves* to `inconclusive` because its underlying query errored is indistinguishable from a legitimate `inconclusive`, so it remains a narrower residual source of score drift. Not addressed here.
- Not tested on mobile (web-only feature). No manual E2E run; relying on unit coverage.

## Visual summary

```mermaid
flowchart TD
  Click["Re-scan click"] --> SS["startScan({ isManual: true })"]
  SS --> RS["isRescanning = true"]
  RS --> FR["useSafeScanContext forceRefetch\n→ refetchOnMountOrArgChange on CGW queries"]
  RS --> MIN["hold 'Scanning…' ≥ 1s"]

  subgraph PerSafe["Per-Safe completion"]
    AllDone{"every scanner\nproduced a result?"}
    AllDone -->|yes| Commit["commit results\n→ gauge + cache"]
    AllDone -->|no| Skip["skip commit\nscanIncomplete = true\n(keep prior score)"]
  end
  FR --> PerSafe

  subgraph Bail["Context bail countdown"]
    Vis{"document visible?"}
    Vis -->|yes| Count["count down → bail after window"]
    Vis -->|no| Pause["paused; re-arm on visibilitychange"]
  end

  Skip --> Note["WorkspaceHealthCard:\n'last scan didn't finish' note"]
  Count --> Skip
```

## Checklist

- [ ] I've tested the branch on mobile 📱 (web-only feature; N/A)
- [x] I've documented how it affects the analytics (if at all) 📊 (no analytics change)

### DIFF
--- a/apps/web/src/features/spaces/components/SecurityHub/__tests__/WorkspaceHealthCard.test.tsx
+++ b/apps/web/src/features/spaces/components/SecurityHub/__tests__/WorkspaceHealthCard.test.tsx
@@ -104,4 +104,30 @@ describe('WorkspaceHealthCard', () => {
     rerender(<WorkspaceHealthCard {...baseProps} safes={[safe(SAFE_A)]} scanResults={{}} isScanning={false} />)
     expect(screen.queryByText('100')).not.toBeInTheDocument()
   })
+
+  it('shows an incomplete-scan note when the last scan was partial and not running', () => {
+    render(
+      <WorkspaceHealthCard
+        {...baseProps}
+        safes={[safe(SAFE_A)]}
+        scanResults={{ [scanKey(SAFE_A, '1')]: allClear }}
+        isScanning={false}
+        scanIncomplete
+      />,
+    )
+    expect(screen.getByText(/last scan didn't finish/i)).toBeInTheDocument()
+  })
+
+  it('hides the incomplete-scan note while a scan is running', () => {
+    render(
+      <WorkspaceHealthCard
+        {...baseProps}
+        safes={[safe(SAFE_A)]}
+        scanResults={{ [scanKey(SAFE_A, '1')]: allClear }}
+        isScanning={true}
+        scanIncomplete
+      />,
+    )
+    expect(screen.queryByText(/last scan didn't finish/i)).not.toBeInTheDocument()
+  })
 })
--- a/apps/web/src/features/spaces/components/SecurityHub/components/WorkspaceHealthCard/WorkspaceHealthCard.tsx
+++ b/apps/web/src/features/spaces/components/SecurityHub/components/WorkspaceHealthCard/WorkspaceHealthCard.tsx
@@ -17,6 +17,7 @@ type WorkspaceHealthCardProps = {
   onFilterChange: (grade: SafeGrade) => void
   lastScannedAt: number | null
   onRescan: () => void
+  scanIncomplete?: boolean
 }
 
 type AggregateCounts = {
@@ -108,6 +109,7 @@ const WorkspaceHealthCard = ({
   onFilterChange,
   lastScannedAt,
   onRescan,
+  scanIncomplete = false,
 }: WorkspaceHealthCardProps): ReactElement => {
   const security = useLoadFeature(SecurityFeature)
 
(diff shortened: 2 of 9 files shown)


tongyaosafe (approved):
🚀

PR #7935, opened 2026-05-27
fix(spaces): refresh aggregated balance after a tx executes

tongyaosafe:
> Send from a Space, balance stayed asleep —
> overview cache had no key to keep.
> Tag it, invalidate on success, force the fetch through;
> now the number wakes up the moment the tx goes through.

## What it solves

Resolves: [WA-2348](https://linear.app/safe-global/issue/WA-2348/bug-balance-does-not-update-at-space-level-after-transaction-execution)

After executing a Send transaction from the Space level, the aggregated balance on the Space dashboard did not update — it kept showing the pre-transfer amount until a full (and unreliable) page reload.

The root cause is that the Space aggregated balance is sourced from a different data channel than the single-Safe dashboard balance:

- **Single Safe dashboard** → portfolio/balances endpoints, refreshed after a tx by `usePortfolioRefetchOnTxHistory` (watches `txHistoryTag`) + 15s polling.
- **Space aggregated balance** → the `getMultipleSafeOverviews` RTK Query endpoint, which had **no `providesTags`, no `pollingInterval`, and no post-tx refetch**. Nothing ever invalidated or re-ran it, so the displayed number stayed at whatever was cached when the Space first loaded.

## How this PR fixes it

Wires up the missing "refresh overviews after a tx" link, in three parts (all required):

1. **Tag the overview endpoints** — `getSafeOverview` / `getMultipleSafeOverviews` now `providesTags: ['SafeOverviews']` (new tag type on `gatewayApi`), making them targetable by cache invalidation.
2. **Invalidate at the right moment** — a new global hook `useInvalidateOverviewsOnTx` subscribes to `TxEvent.SUCCESS` (fired when the tx is indexed by the gateway = the backend balance is fresh) and dispatches `invalidateTags(['SafeOverviews'])`, which re-runs the outer query. `SUCCESS` is used rather than the earlier `PROCESSED` because at `PROCESSED` time the gateway has not indexed the new balance yet.
3. **Bypass the inner cache** — the outer query internally re-dispatches the inner `safesGetOverviewForMany` query. Without help, a still-warm inner cache would return the same stale aggregate on re-run, so the inner `initiate` now passes `{ forceRefetch: true }`. (`forceRefetch` is deterministic and avoids relying on RTK's subtle "invalidate an unsubscribed cache entry" semantics.)

## How to test it

1. `yarn workspace @safe-global/web dev`, open a Space that contains a Safe with some balance, connect a wallet (Sepolia).
2. Open DevTools → Network, filter `v2/safes`. **Do not reload the page.**
3. Click **Send** on the Space dashboard, perform a small transfer and execute it.
4. After the tx is indexed (~15–30s) a fresh `GET /v2/safes?...` request fires automatically and the aggregated balance updates to the post-transfer amount — no manual refresh.
5. Compare against `dev`: there, no `v2/safes` request fires after the tx and the number stays stale.

Unit/integration tests:

```bash
yarn workspace @safe-global/web test src/store/__tests__/safeOverviews.test.ts src/hooks/__tests__/useInvalidateOverviewsOnTx.test.ts
```

## Affected flows

- Space dashboard: send a transaction → aggregated balance refreshes automatically (primary fix).
- My accounts list, sidebar Safe selectors, and any other view backed by Safe overviews now also refresh after a tx (they shared the same latent staleness bug).

## Blast radius

- **RTK Query**: adds tag type `'SafeOverviews'` to `gatewayApi`; `getSafeOverview` / `getMultipleSafeOverviews` now provide it. Inner `safesGetOverviewForMany` (cgw `/v2/safes`) is now called with `forceRefetch: true` from these endpoints.
- **Consumers of the overview endpoints** (all refetch on `TxEvent.SUCCESS`): `useSpaceAccountsData`, `useMultiAccountItemData`, `useSafeItemData`, `SafeCardReadOnly`, `useReconciledSpaceSafes`, `useSafeScanContext`, `InconsistentSignerSetupWarning`, `NestedSafesList`, `useSpaceSafeSelectorItems`, `AggregatedBalances`.
- **New global subscription**: `useInvalidateOverviewsOnTx` mounted in `_app.tsx` → `SafeScopedSubscriptions` (active only when logged in, alongside the other tx-event subscribers). Cleans up its subscription on unmount.
- **Not touched**: single-Safe portfolio/balances channel; mobile (change is web-only).

## Risks / not checked

- **Trigger scope is "outgoing tracked txs"**: the refresh fires on `TxEvent.SUCCESS`, which `txHistorySlice` only dispatches for a nonce-tracked pending tx that appears in history — i.e. the Space-level Send this ticket is about. It intentionally does **not** cover module / spending-limit executions (dispatched with only a `groupKey`, no nonce) or **incoming** transfers. Those paths can still leave the aggregated overview stale until a reload. A broader fix (parity with the single-Safe portfolio channel, which watches `txHistoryTag`) is not cleanly possible here: `txHistoryTag` is scoped to the single active Safe via `useSafeInfo`, so it cannot cover incoming changes to the *other* Safes in a multi-Safe Space aggregate — that would need per-Safe tracking and is out of scope for this bug.
- **Indexer lag**: `TxEvent.SUCCESS` is the app's standard "tx indexed" signal; if the gateway's aggregate lags a few seconds behind, a single refetch could momentarily still show the old value. Same signal the rest of the app already trusts.
- **Request burst**: every successful tx now refetches all currently-mounted overview queries (one batched `/v2/safes` per subscription group). Acceptable.
- **Marginal extra traffic**: when the outer `gatewayApi` cache has expired but the inner `cgwClient` cache (default `keepUnusedDataFor` 60s) is still warm, the `forceRefetch` makes the outer re-run always hit the network instead of deduping. Small, and arguably desirable freshness.
- Did not test on mobile (web-only change).

## Visual summary

```mermaid
flowchart TB
  subgraph Before["Before — Space aggregated balance never refreshes"]
    TX1[Send tx executes] -.no trigger.-> OV1["getMultipleSafeOverviews<br/>(no tag, no polling, no refetch)"]
    OV1 --> STALE["Shows pre-transfer balance<br/>until full page reload"]
  end

  subgraph After["After — refresh link wired up"]
    TX2[Send tx executes] --> IDX[Gateway indexes tx]
    IDX --> SUCCESS[TxEvent.SUCCESS]
    SUCCESS --> HOOK["useInvalidateOverviewsOnTx<br/>invalidateTags(['SafeOverviews'])"]
    HOOK --> OUTER["Outer getMultipleSafeOverviews<br/>re-runs (tag invalidated)"]
    OUTER --> INNER["Inner safesGetOverviewForMany<br/>initiate(forceRefetch: true) → /v2/safes"]
    INNER --> FRESH[Aggregated balance updates, no reload]
  end
```

## Checklist

- [ ] I've tested the branch on mobile 📱 (N/A — web-only change)
- [x] I've documented how it affects the analytics (if at all) 📊 (no analytics change)

---

## CLA signature

### DIFF
--- a/apps/web/src/hooks/useInvalidateOverviewsOnTx.ts
+++ b/apps/web/src/hooks/useInvalidateOverviewsOnTx.ts
@@ -1,6 +1,6 @@
 import { useEffect } from 'react'
 import { useAppDispatch } from '@/store'
-import { gatewayApi } from '@/store/api/gateway'
+import { gatewayApi, makeSafeOverviewTag } from '@/store/api/gateway'
 import { TxEvent, txSubscribe } from '@/services/tx/txEvents'
 
 /**
@@ -10,13 +10,18 @@ import { TxEvent, txSubscribe } from '@/services/tx/txEvents'
  * The single-Safe balance refreshes via the portfolio/balances endpoints, but the
  * overview endpoints have no polling or tag invalidation of their own, so without this
  * the aggregated balance would stay stale until a full page reload.
+ *
+ * Invalidation is scoped to the executing Safe (chainId:address) so a tx only refetches
+ * the overview queries that actually contain that Safe, not every mounted overview query.
  */
 const useInvalidateOverviewsOnTx = (): void => {
   const dispatch = useAppDispatch()
 
   useEffect(() => {
-    return txSubscribe(TxEvent.SUCCESS, () => {
-      dispatch(gatewayApi.util.invalidateTags(['SafeOverviews']))
+    return txSubscribe(TxEvent.SUCCESS, ({ chainId, safeAddress }) => {
+      dispatch(
+        gatewayApi.util.invalidateTags([{ type: 'SafeOverviews', id: makeSafeOverviewTag(chainId, safeAddress) }]),
+      )
     })
   }, [dispatch])
 }
--- a/apps/web/src/hooks/useInvalidateOverviewsOnTx.ts
+++ b/apps/web/src/hooks/useInvalidateOverviewsOnTx.ts
@@ -1,0 +1,24 @@
+import { useEffect } from 'react'
+import { useAppDispatch } from '@/store'
+import { gatewayApi } from '@/store/api/gateway'
+import { TxEvent, txSubscribe } from '@/services/tx/txEvents'
+
+/**
+ * Refreshes the Safe overview cache (aggregated balances shown on the Space dashboard,
+ * My accounts list, Safe selectors, …) once an executed transaction has been indexed.
+ *
+ * The single-Safe balance refreshes via the portfolio/balances endpoints, but the
+ * overview endpoints have no polling or tag invalidation of their own, so without this
+ * the aggregated balance would stay stale until a full page reload.
+ */
+const useInvalidateOverviewsOnTx = (): void => {
+  const dispatch = useAppDispatch()
+
+  useEffect(() => {
+    return txSubscribe(TxEvent.SUCCESS, () => {
+      dispatch(gatewayApi.util.invalidateTags(['SafeOverviews']))
+    })
+  }, [dispatch])
+}
+
+export default useInvalidateOverviewsOnTx
--- a/apps/web/src/store/api/gateway/safeOverviews.ts
+++ b/apps/web/src/store/api/gateway/safeOverviews.ts
@@ -39,12 +39,18 @@ class SafeOverviewFetcher {
     currency: string
     dispatch: DispatchFn
   }) {
-    const queryThunk = additionalSafesRtkApi.endpoints.safesGetOverviewForMany.initiate({
-      safes: safeIds,
-      currency,
-      walletAddress,
-      trusted: false,
-    })
+    // forceRefetch so a cache invalidation (e.g. after a tx) re-runs the wrapping
+    // query AND bypasses the inner cache, otherwise the inner entry — unsubscribed
+    // here — would return stale data and the displayed balance would not update.
+    const queryThunk = additionalSafesRtkApi.endpoints.safesGetOverviewForMany.initiate(
+      {
+        safes: safeIds,
+        currency,
+        walletAddress,
+        trusted: false,
+      },
+      { forceRefetch: true },
+    )
     const queryAction = dispatch(queryThunk)
 
     try {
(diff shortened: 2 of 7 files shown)


tongyaosafe:
@claude review

apps/web/src/store/api/gateway/safeOverviews.ts
```diff
@@ -39,12 +39,18 @@ class SafeOverviewFetcher {
...
+    // query AND bypasses the inner cache, otherwise the inner entry — unsubscribed
+    // here — would return stale data and the displayed balance would not update.
+    const queryThunk = additionalSafesRtkApi.endpoints.safesGetOverviewForMany.initiate(
+      {
+        safes: safeIds,
+        currency,
+        walletAddress,
+        trusted: false,
+      },
+      { forceRefetch: true },
```
vmosharova, line 52: This applies not only after a successful tx but every time - even currenct change will trigger the refetch. Maybe it makes sense to scope it more narrow?

apps/web/src/hooks/useInvalidateOverviewsOnTx.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * The single-Safe balance refreshes via the portfolio/balances endpoints, but the
+ * overview endpoints have no polling or tag invalidation of their own, so without this
+ * the aggregated balance would stay stale until a full page reload.
+ */
+const useInvalidateOverviewsOnTx = (): void => {
+  const dispatch = useAppDispatch()
+
+  useEffect(() => {
+    return txSubscribe(TxEvent.SUCCESS, () => {
```
vmosharova: This fires on every tx success (with a tag `'SafeOverviews'` from `safeOverviews.ts`), so refetch happens regardless of safe address and chain. That can result in many refetches.

vmosharova: Makes sense to narrow to safe address + chain

vmosharova (changes requested):
Added some comments :)

vmosharova:
@claude review again

apps/web/src/hooks/useInvalidateOverviewsOnTx.ts
tongyaosafe: very good points. 
narrowed it to address + chain

apps/web/src/store/api/gateway/safeOverviews.ts
tongyaosafe: Good point.
But I think it is safe and better to forceRefecth here.
It won't refetch on every rerender, but will refetch whenever the outer queryFn executes (mount / arg change / tag invalidation).
I think this is what we want.

And for the currency change trigger refetch, I did not change this part. It was the same as before.

tongyaosafe:
@claude review

vmosharova (approved):
👍🏻

katspaugh (approved):

tongyaosafe:
Already merged to dev by another PR:
https://github.com/safe-global/safe-wallet-monorepo/pull/7960

PR #7938, opened 2026-05-28
fix(mobile): memoize Identicon SVG to fix /recipient frozen frames

compojoom:
> Memoise the blockie,
> cache it across the app —
> no more frozen frames.

## What it solves

Resolves: [WA-2427](https://linear.app/safe-global/issue/WA-2427/mobile-fix-recipient-frozen-frames-from-unmemoised-identicon-svg) (related to [WA-2242](https://linear.app/safe-global/issue/WA-2242))

The iOS RUM monitor [Mobile - Frozen Frame Rate > 0.5% / 1% (iOS) production](https://app.datadoghq.eu/monitors/105523631) alerted on 2026-05-28 with frozen frames concentrated in `/recipient` (2/2 sessions). Root cause: `Identicon` called `bloSvg(address)` synchronously on every render and `SvgXml` re-parsed the SVG markup each time. `/recipient` renders one Identicon per row across "My Safe accounts" + "Signers" + "Address book", so the JS thread froze on mount for users with non-trivial address books.

## How this PR fixes it

Three layered changes inside `apps/mobile/src/components/Identicon/Identicon.tsx`, shipped together:

1. **Memoise `bloSvg()` per instance** with `useMemo` keyed on `address`.
2. **Wrap `Identicon` in `React.memo`** so identical props short-circuit before any work.
3. **Module-level bounded cache** (LRU-ish `Map`, 256 entries / ~1.3 MB worst case) so the same address reuses its SVG across screens, list scrolls, and mount/unmount cycles.
4. **Neutral placeholder for invalid addresses** (matches web's `Skeleton` fallback). Prevents `RecipientInput` from re-rendering broken SVGs on every keystroke.

The web counterpart already wraps `blo()` in `useMemo` — proven in-repo pattern, mobile was the only consumer paying the cost.

## How to test it

1. iOS simulator with an address book of 30+ entries → open `/recipient`. Should render smoothly with no visible jank.
2. Type a partial / invalid address in `RecipientInput` → placeholder shown instead of crash or broken SVG.
3. Scroll `/recipient` list, navigate away, return → identicons reload instantly (module cache hit).
4. Unit tests in `Identicon.test.tsx` cover memoisation, module cache, case normalisation (`0xAb…` vs `0xab…`), and invalid-input placeholder.

## Affected flows

- `/recipient` — primary fix target; smooth render on first mount with large address book.
- `/review-and-confirm` — second-highest frozen-frame contributor in the alert window; should also improve.
- `RecipientInput` typing flow — placeholder for invalid / partial input.
- All ~28 mobile screens using `Identicon` benefit from caching (signers list, confirm-transaction, history details, address book, settings, signers import, etc.).

## Blast radius

- **Component contract:** `Identicon` is now `React.memo`-wrapped. Identical props skip rendering; this is the intended behaviour and matches the web component.
- **New behaviour:** invalid `address` props now render a neutral placeholder `View` (testID `identicon-image-placeholder`) instead of an empty / malformed `<SvgXml>`. Visually safer.
- **Module cache:** singleton `Map` lives for the lifetime of the JS bundle. Bounded at 256 entries with LRU-ish eviction. Fast Refresh recreates the module so no cross-reload state.
- **New import:** `isAddress` from `ethers` — already a top-level mobile dep, zero bundle impact.

## Risks / not checked

- Memory footprint over long sessions — cache is bounded at 256 entries (~1.3 MB worst case); not measured in practice.
- `BlurredIdenticonBackground` shares the same un-memoised pattern but renders once per screen — left for a separate small PR.
- Datadog monitor recovery is post-deploy verification; success metric is `/recipient` dropping out of the top frozen-frame contributors over a 7-day window.
- No manual iOS simulator run in this session — relying on unit tests + visual inspection at review.

## Visual summary

```mermaid
flowchart LR
  subgraph Before
    A1[Render Identicon] --> B1[bloSvg every render]
    B1 --> C1[SvgXml re-parse]
    C1 --> D1[Frozen frame on /recipient]
  end
  subgraph After
    A2[Render Identicon] --> M[React.memo: skip if same props]
    M --> U[useMemo: skip if same address]
    U --> Q{Module cache hit?}
    Q -->|Yes| H[Reuse SVG string]
    Q -->|No| B2[bloSvg once]
    B2 --> S[Store in cache, evict oldest if full]
  end
```

## Checklist

---

## CLA signature

### DIFF
--- a/apps/mobile/src/components/Identicon/Identicon.tsx
+++ b/apps/mobile/src/components/Identicon/Identicon.tsx
@@ -1,7 +1,9 @@
+import { memo, useMemo } from 'react'
 import { bloSvg } from 'blo'
-import { type Address } from '@/src/types/address'
+import { isAddress } from 'ethers'
 import { View } from 'tamagui'
 import { SvgXml } from 'react-native-svg'
+import { type Address } from '@/src/types/address'
 
 type Props = {
   address: Address
@@ -10,14 +12,48 @@ type Props = {
 }
 
 const DEFAULT_SIZE = 56
-export const Identicon = ({ address, rounded = true, size }: Props) => {
-  size = size ? size : DEFAULT_SIZE
+const MAX_CACHE = 256
+const svgCache = new Map<string, string>()
 
-  const blockieSvg = bloSvg(address)
+function getIdenticonSvg(address: Address): string {
+  // `blo` lowercases the address internally before hashing, so the SVG output
+  // is identical regardless of input case. We lowercase only the cache key to
+  // dedupe entries when callers pass mixed cases.
+  const key = address.toLowerCase()
+  const cached = svgCache.get(key)
+  if (cached) {
+    svgCache.delete(key)
+    svgCache.set(key, cached)
+    return cached
+  }
+  const svg = bloSvg(address)
+  svgCache.set(key, svg)
+  if (svgCache.size > MAX_CACHE) {
+    const oldest = svgCache.keys().next().value
+    if (oldest !== undefined) {
+      svgCache.delete(oldest)
+    }
+  }
+  return svg
+}
+
+export const Identicon = memo(function Identicon({ address, rounded = true, size = DEFAULT_SIZE }: Props) {
+  const valid = isAddress(address)
+  const blockieSvg = useMemo(() => (valid ? getIdenticonSvg(address) : null), [valid, address])
+
+  if (!blockieSvg) {
+    return (
+      <View
+        style={{ width: size, height: size, borderRadius: rounded ? '50%' : 0 }}
+        backgroundColor="$backgroundSkeleton"
+        testID="identicon-image-placeholder"
+      />
+    )
+  }
 
   return (
     <View style={{ borderRadius: rounded ? '50%' : 0, overflow: 'hidden' }} testID={'identicon-image-container'}>
       <SvgXml testID={'identicon-image'} xml={blockieSvg} width={size} height={size} />
     </View>
   )
-}
+})
--- a/apps/mobile/src/components/Identicon/Identicon.tsx
+++ b/apps/mobile/src/components/Identicon/Identicon.tsx
@@ -1,9 +1,13 @@
 import { memo, useMemo } from 'react'
 import { bloSvg } from 'blo'
-import { isAddress } from 'ethers'
 import { View } from 'tamagui'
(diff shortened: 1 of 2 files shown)


apps/mobile/src/components/Identicon/Identicon.tsx
```diff
@@ -10,14 +16,48 @@ type Props = {
...
-  const blockieSvg = bloSvg(address)
+function getIdenticonSvg(address: Address): string {
+  // `blo` lowercases the address internally before hashing, so the SVG output
+  // is identical regardless of input case. We lowercase only the cache key to
+  // dedupe entries when callers pass mixed cases.
+  const key = address.toLowerCase()
+  const cached = svgCache.get(key)
+  if (cached) {
+    svgCache.delete(key)
+    svgCache.set(key, cached)
```
tmjssz, line 30: Why do we delete and set the same cached value here again?

apps/mobile/src/components/Identicon/Identicon.tsx
```diff
@@ -10,14 +16,48 @@ type Props = {
...
+    return cached
+  }
+  const svg = bloSvg(address)
+  svgCache.set(key, svg)
+  if (svgCache.size > MAX_CACHE) {
+    const oldest = svgCache.keys().next().value
+    if (oldest !== undefined) {
```
tmjssz, line 37: Nit: this if condition is unnecessary, `oldest` should always be defined after the .set call before.

apps/mobile/src/components/Identicon/Identicon.tsx
tmjssz: Ah got it. It's to make sure that the cached value is inserted again at the end of the map so it's the "newest" value in the cache, right? (Maybe worth a comment in the code)

tmjssz (approved):

apps/mobile/src/components/Identicon/Identicon.tsx
compojoom: it's because of the Iterator

Map<K, V>.keys().next().value is typed as K | undefined by TypeScript's iterator types. The if (oldest !== undefined) guard is doing double duty — it's not just a runtime check, it's also narrowing the type so svgCache.delete(oldest) type-checks (if K is string, passing string | undefined would error).

PR #7939, opened 2026-05-28
fix(web): friendlier message for transient RPC rate-limits during execution

compojoom:
> Wrapper gone, viem retries on,
> a busy-network note remains,
> the false revert is gone.

## What it solves

Executing a transaction occasionally failed with `The contract function "execTransaction" reverted with the following reason: Request is being rate limited.` (WA-1946 / WA-2424). This is not a chain revert — it's a JSON-RPC `-32005` / HTTP `429` from Infura, wrapped by viem to look like a contract revert.

The first cut of this fix added a `RetryingRpcProvider` around the read-only RPC. On audit that was:

- **Redundant** — viem's `http()` transport already retries `-32005` / `429` with exponential backoff and honours `Retry-After`. Passing the provider as an EIP-1193 object actually forced viem's pass-through `custom()` transport and lost that native handling.
- **Misplaced** — execution re-inits the SDK with the connected wallet's provider (`sdk.connect`), so the throttled reads/sends never traverse our read-only wrapper. ~12 Datadog RUM events carried MetaMask's `RpcService` stack frames, confirming the throttle is on the wallet's own RPC backend.

## How this PR fixes it

- Removes `RetryingRpcProvider` and its test; reverts provider wiring (`safeCoreSDK`, `getSafeProvider`, `getCurrentGnosisSafeContract`) to pass URL strings to `SafeProvider` / `Safe.init`, restoring viem's native retry on the read path.
- Keeps only the user-facing message. `isRateLimitError` matches viem `BaseError` cause chains carrying `-32005` / `429` — structured detection, not a message regex, so real contract reverts mentioning "rate limit" are not misclassified — and surfaces `RATE_LIMIT_USER_MESSAGE` ("Network is busy. Please try again in a moment.") in the execution toast (`useTxNotifications`) and inline submit errors (`ComboSubmit` / `ExecuteForm`).
- `viem` remains a direct dependency (used for the `BaseError` import).

## How to test it

1. Open a transaction and reach the Execute step.
2. Simulate a transient throttle by returning `429` / `-32005` from the RPC during execution (or rely on the unit tests below).
3. Confirm the UI shows "Network is busy. Please try again in a moment." instead of the raw `execTransaction reverted ... Request is being rate limited` string.

Unit tests: `yarn workspace @safe-global/web test src/utils/__tests__/transaction-errors.test.ts` (18 pass).

## Visual summary

```mermaid
flowchart LR
  subgraph Before
    A1[Safe.init / SafeProvider] --> R[RetryingRpcProvider<br/>EIP-1193 object]
    R --> C1[viem custom transport<br/>no native retry]
    C1 --> E1[re-implements retry via ethers]
  end
  subgraph After
    A2[Safe.init / SafeProvider] --> U[URL string]
    U --> C2[viem http transport<br/>retries -32005/429 + Retry-After]
    C2 --> M[isRateLimitError -> friendly message]
  end
```

## Affected flows

- Safe SDK init (all chains), tx pre-flight validation (`useIsValidExecution`), tx execution failure toast, undeployed/predicted Safe init.

## Blast radius

- Read-only provider construction reverts to the dev baseline (URL string). Message helper is provider-agnostic.

## Risks / not checked

- E2E rate-limit behaviour (the flaky signer+Infura cypress spec was removed; friendly message is covered by unit tests).
- Residual throttling from the signer wallet's own RPC backend is outside our control.

## Checklist

- [x] Code follows the style guidelines of this project
- [x] Tests added/updated and passing
- [x] type-check, lint, prettier clean
- [x] No new warnings introduced


compojoom:
@claude review

apps/web/src/utils/providers/RetryingRpcProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+    body = JSON.parse(text)
+  } catch {
+    return null
+  }
+  const errors = Array.isArray(body)
+    ? body.map((entry) => (entry as { error?: { code?: number; message?: string } } | null)?.error).filter(Boolean)
+    : [(body as { error?: { code?: number; message?: string } } | null)?.error].filter(Boolean)
+  const hit = (errors as Array<{ code?: number; message?: string }>).find(
+    (err) => typeof err.code === 'number' && RETRYABLE_RPC_CODES.has(err.code),
+  )
```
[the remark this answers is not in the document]
compojoom: Addressed in 385e51797. Agreed —  was added speculatively; WA-2424's evidence names 429/-32005 specifically. Removed it from both the retry path (RetryingRpcProvider) and from `isRateLimitError`. Added explicit negative tests so a future regression would fail. If we see real Infura throttling come back as -32603 post-deploy, we'll add it back with message-regex gating.

clovisdasilvaneto (dismissed):

gfendres:
@compojoom this is just waiting for someone to review?

TimDaub:
> 📦 Next.js Bundle Analysis for @safe-global/web

This analysis ran before committed [0487eed](https://github.com/safe-global/safe-wallet-monorepo/pull/7939/commits/0487eed82fd697df2e9d6fd4d853654dd5aee416) (which adds viem as a dependency). I will check if the bundle size has increased

apps/web/package.json
```diff
@@ -134,6 +127,7 @@
     "tailwindcss": "^4.1.18",
     "tw-animate-css": "^1.4.0",
     "vaul": "^1.1.2",
+    "viem": "^2.21.55",
```
TimDaub, line 130: Why did we decide to go with viem "2.21.55"? E.g. we could have also gone with 2.52.2, no?

compojoom: viem is already used as transitive dependency in the repo. Currently at version 2.21.55 - that's why I hardcoded it. Not sure how other packages will react to updating it to 2.52.2

TimDaub (approved):
- I timeboxxed this review to 25mins so please excuse if I lack some context
- My understanding is that this is fixing an issue where when we want to write a transaction to a chain, somewhere during the write process a viem error is thrown to us that upsets users (because it shows them that the action has failed)
- But the actual error is that we're making so many requests that within that write to the chain there were some related read calls that caused the RPC to send a 429, right?
- We are not yet using viem in monorepo. We are transiently using viem though through protocol-kit? protocol-kit is sending us this arguably inaccurate error
- If all of the above is true: IMO we should file a bug with protocol-kit and ask them to provide better error messages for us
- To me we can still merge this in the meantime as it's helping our users
- I didn't test this in my machine
- I don't know if this will significantly increase the bundle size. If treeshaking etc. works it shouldn't, but these days you never know :D

TimDaub:
@dasanra said it's good that we fix the error in wallet-monorepo

PR #7940, opened 2026-05-28
fix(web/Spaces): improve UX/UI for the case when members try to add safes to a space

vmosharova:
## What it solves

Resolves: [WA-2425](https://linear.app/safe-global/issue/WA-2425/show-add-safe-to-workspace-to-admins-only-gate-other-adding-entry)

This PR adds client-side UI gating for workspace admin checks. It:

- Adds an isAdminOfActiveSpace flag to skip a backend call that would otherwise 403
- Adds UI tooltips and disabled states for non-admins
- Hides "Add Safe to workspace" buttons when user isn't admin of any space
- Disables spaces in dropdown when safe is already there or user isn't admin

## How this PR fixes it

- **Sidebar "Add Safe to workspace" trigger**: hidden when the user is admin of zero spaces.
- **Space selector dropdown** (`addToWorkspace` mode): each space item is disabled with an admin tooltip when the user is not an active admin of that space.
- **Space selector dropdown** (`addToWorkspace` mode): each space item is disabled with an info tooltip when the safe is already added to the space.
- **AddAccounts modal**: added a defensive admin guard at the top of `onSubmit` so the mutations are not called even if the modal is opened externally.
- **AddAccounts trigger button**: replaced native `title` with a real tooltip via the new `AdminOnlyWorkspaceTooltip` so it shows on disabled buttons.
- **Counterfactual safe deployment**: `persistCounterfactualSafe` skips the space-attach call when the user is not admin of the active space. The safe still persists at the user level.
- `AdminOnlyWorkspaceTooltip` — wraps children in a tooltip "Only admins can add Safes to this workspace" when the user is not an active admin. Accepts either `members` (derives admin status) or an explicit `isAdmin` flag for callers that already have it.

## How to test it

- [ ] As a Member of space S: open a Safe → sidebar trigger is hidden if S is the only space.
- [ ] As a Member of one space + Admin of others: trigger appears; the member-space row in the dropdown is disabled; admin-space rows are clickable and add the Safe.
- [ ] On the spaces dashboard as a Member: the "Add Accounts" / "Add account" buttons are disabled with a tooltip on hover.
- [ ] Open a safe from the space where it is not added yet (but is added to another space) - click "Add Safe to workspace" - the space where safe is already added should be with a tooltip "Safe is already in this workspace".
- [ ] Deploy a counterfactual Safe while a member-only space is active: Safe persists, no 403 toast, the Safe is not auto-attached to that space.

## Blast radius

**Type-level (TS-enforced)**
- `SpaceItem` now has optional `members?: SpaceMemberDto[]` — downstream code that constructs `SpaceItem` literals (stories, tests) still compiles; runtime data flowing from `GetSpaceResponse` already has `members`.
- `PersistArgs.isAdminOfActiveSpace` is **required** — every caller of `persistCounterfactualSafe` must pass it. Two callers updated.

**Behavioral changes**
- Admins: zero behavioral change across all flows.
- Members:
  - Sidebar trigger may disappear (if zero admin spaces) — previously visible-and-broken.
  - Non-admin rows in the dropdown become non-interactive — previously clickable-and-403'd.
  - AddAccounts trigger now shows a styled tooltip instead of an invisible native `title`.
  - Counterfactual deploy in a non-admin active space: Safe persists at user level only, no 403 toast, no auto-attach.
- Signed-out / no-space users: unchanged.

## Risks

- Stale RTK cache after a role change in a space could briefly mislead the gating (admin → member or vice versa) until the spaces query refetches.
- Any future caller of `persistCounterfactualSafe` must pass `isAdminOfActiveSpace`; TS will enforce it but reviewers should keep an eye on new callers.

## Checklist

---

## CLA signature


vmosharova:
@claude review

vmosharova:
> **1. Duplicated admin-of-space lookup across three components** 

A hook would save just ~6 lines total and add a new file + indirection. Also a single hook doesn't fit for every use case. Makes more sense to leave it as it is for now.

Other comments are addressed in [d4823b8](https://github.com/safe-global/safe-wallet-monorepo/pull/7940/commits/d4823b8e0f1963ad30bb622badbc38cdc6d16dbe) and [43cbc45](https://github.com/safe-global/safe-wallet-monorepo/pull/7940/commits/43cbc45e065390d6b74dc5bd1bfe30fa93931bda)

vmosharova:
> 💥 risk · [apps/web/src/features/counterfactual/services/persistCounterfactualSafe.ts L43-60](https://github.com/safe-global/safe-wallet-monorepo/blob/43cbc45e065390d6b74dc5bd1bfe30fa93931bda/apps/web/src/features/counterfactual/services/persistCounterfactualSafe.ts#L43-L60) persistCounterfactualSafe
> A couple of things to watch. The admin status relies on RTK query data; if a users role changes while the page is open, there could be a brief window where the UI gate is stale. 

Stale UI doesn't seem to be a risk. We already handle the "not admin → skip space attach + show info toast" path. This (quite rare) edge case when the role is changed is already covered with a toast.

vmosharova:
@claude review again

katspaugh (approved):

PR #7942, opened 2026-05-28
fix(web): don't bounce existing users into create-space on re-login (round 2)

katspaugh:
> Stale cache after logout,
> single space — straight to it,
> never to create-space.

## What it solves

Two related issues in the spaces sign-in flow:

1. **(Bug)** Existing users who log out and log back in are redirected to `/welcome/create-space` instead of seeing their workspaces. Despite a previous fix in [9166bb41b](https://github.com/safe-global/safe-wallet-monorepo/commit/9166bb41b), the bounce was still reproducible — the actual root cause was in the route guard, not `useSignInRedirect`.
2. **(Feature)** When the signed-in user has exactly one space, the workspace list page is just an extra click. Sign-in now short-circuits straight into that space's dashboard.

### Root cause of the re-login bounce

`useFlowActivationGuard` reads the spaces list to decide whether to redirect to `/welcome/create-space`:

```typescript
if (isSiweAuthenticated) {
  const { data: spaces } = await fetchSpaces(undefined)
  hasSpaces = !!spaces && spaces.length > 0  // ← ANY error → false → redirect
}
```

After the gateway logout, the persisted `authSlice.sessionExpiresAt` still says "signed in" until `reconcileAuth` resolves. During that window the guard runs with `isSiweAuthenticated=true`, `fetchSpaces` returns a 401/403 error (cookies cleared) with `data: undefined`, and the old code treated that as "no spaces" → `router.replace('/welcome/create-space')`.

Fix: distinguish transient/auth errors from a confirmed empty/404 response. Only redirect to create-space when the fetch resolved with no error or with a 404. On any other error keep `hasSpaces=true` so the page renders, let the auth listener clean up the stale auth state, and let the guard re-run with the correct `isSiweAuthenticated`.

### Defensive fix in SpacesList

The earlier `useSignInRedirect` gate (`isFetching || isUninitialized`) only catches the *first-ever* skip→unskip transition. After invalidation the cache entry exists, so both flags are false on the render where skip flips. Added a `(spaces === undefined && !error)` clause so any state without a definitive answer is treated as still loading.

### Single-space short-circuit

`useSignInRedirect` now accepts a `singleSpaceId`. When sign-in completes and the user has exactly one active space, push straight to `/spaces?spaceId=<id>`. The existing `?next=` round-trip (require-login gate) still takes precedence — that's the URL the user originally tried to open.

## How to test it

1. Sign in with a wallet that has exactly one workspace. **Expected:** lands on `/spaces?spaceId=<id>` directly (not on the workspace list).
2. Sidebar profile → Sign out. The gateway logout redirect lands back on `/welcome/spaces`, signed out.
3. Sign in again with the same wallet. **Expected:** lands on `/spaces?spaceId=<id>` again. **Before this fix:** bounced into `/welcome/create-space`.

Neighbouring flows:
- New user (no workspaces) signs in → still redirects to `/welcome/create-space`.
- User with 2+ workspaces signs in → lands on `/welcome/spaces` (the picker).
- Spaces 404 → still redirects to `/welcome/create-space`.
- Spaces 5xx → stays on `/welcome/spaces` (error early-return).
- Require-login gate ON + `?next=/balances` → lands on `/balances`, even with one space.

## Affected flows

- Re-login after explicit logout (SIWE wallet path) — bug fix
- Single-space sign-in — behavior change
- First-time wallet login (with/without spaces) — unchanged
- OIDC re-login — same loading semantics + single-space short-circuit benefit it
- `?next=` round-trip via require-login gate — unchanged
- Spaces 404 / 5xx error handling — unchanged

## Blast radius

- `useFlowActivationGuard` is the route guard for every protected page — but the change only affects the branch where `isSiweAuthenticated=true && fetchSpaces returns an error`. All other branches keep their existing behavior. New unit test pins the regression.
- `useSignInRedirect` consumers: `SpacesList` is the only one. The hook gained one optional `singleSpaceId` prop with a `null`/`undefined` default — existing callers see no behavior change.
- The auth listener (`spacesApi.util.invalidateTags(['spaces'])` on `setUnauthenticated`) stays untouched.
- No changes to RTK Query endpoints, persisted state, feature flags, or routes.
- Mobile not impacted (Spaces sign-in lives in `apps/web` only).

## Risks / not checked

- The fix is informed by reading the route guard + hook code. The Cypress E2E in this PR (`spaces_basicflow.cy.js`) exercises the full sign-in → create space → sidebar sign-out → re-sign-in flow with the private-key wallet module — but it needs a real CI run against staging to confirm.
- Did not manually exercise the OIDC re-login path.
- Did not verify the multi-tab broadcast scenario (logout in tab A, re-login in tab B).
- Did not test on mobile.

## Visual summary

```mermaid
sequenceDiagram
    participant U as User
    participant App as Web app (Redux + RTK Query + RouterGuard)
    participant CGW as CGW (/v1/spaces, /v1/auth/*)

    Note over U,CGW: Logout → page reload via gateway
    U->>App: Click "Sign out"
    App->>CGW: POST /v1/auth/logout/redirect
    CGW-->>App: 303 → /welcome/spaces (cookies cleared)

    Note over App: Persisted sessionExpiresAt still valid<br/>isSiweAuthenticated=true on initial render

    rect rgb(255, 230, 230)
        Note over App: useFlowActivationGuard runs
        App->>CGW: GET /v1/spaces (via fetchSpaces in guard)
        CGW-->>App: 401/403 (cookies cleared)
        Note over App: Before: data=undefined → hasSpaces=false<br/>→ router.replace(/welcome/create-space) ❌
        Note over App: After: transient error → hasSpaces=true (be lenient)<br/>→ allow page to render ✓
    end

    App->>CGW: GET /v1/auth/me (reconcileAuth)
    CGW-->>App: 403
    App->>App: setUnauthenticated → invalidateTags(['spaces'])

    Note over U,CGW: User re-signs-in
    U->>App: Click "Sign in" → SIWE
    App->>App: setAuthenticated → isSiweAuthenticated=true

    App->>CGW: GET /v1/spaces (with new cookies)
    CGW-->>App: [space1]

    rect rgb(230, 255, 230)
        Note over App: activeSpaces.length === 1<br/>→ singleSpaceId = "space1"
        Note over App: useSignInRedirect short-circuit:<br/>router.push(/spaces?spaceId=space1)
    end
```

## Checklist

- [ ] I've tested the branch on mobile 📱 (N/A — web only)
- [x] I've documented how it affects the analytics (if at all) 📊 (no analytics changes)

---

## CLA signature

### DIFF
--- a/apps/web/cypress/e2e/pages/spaces.page.js
+++ b/apps/web/cypress/e2e/pages/spaces.page.js
@@ -93,6 +93,12 @@ const acceptInviteBtn = '[data-testid="accept-invite-button"]'
 const inviteNameInput = '[data-testid="invite-name-input"]'
 const confirmAcceptInviteBtn = '[data-testid="confirm-accept-invite-button"]'
 
+// -- Sidebar profile (sign-out) --
+const sidebarProfileTrigger = '[data-testid="sidebar-profile-trigger"]'
+const sidebarProfilePopover = '[data-testid="sidebar-profile-popover"]'
+const sidebarProfileSignOutBtn = '[data-testid="sidebar-profile-sign-out"]'
+const continueWithWalletBtn = '[data-testid="continue-with-wallet-btn"]'
+
 // -- Onboarding --
 const orgSpaceInput = '[data-testid="space-name-input"]'
 const createSpaceOnboardingContinueBtn = '[data-testid="create-space-onboarding-continue-button"]'
@@ -166,7 +172,26 @@ const spaceDashboardWidgetSelectorByTitle = {
 // ===========================================
 
 export function clickOnSignInBtn() {
-  cy.get('[data-testid="continue-with-wallet-btn"]').click()
+  cy.get(continueWithWalletBtn).click()
+}
+
+export function signOutViaSidebarProfile() {
+  cy.get(sidebarProfileTrigger, { timeout: 30000 }).should('be.visible').click()
+  cy.get(sidebarProfilePopover).should('be.visible')
+  cy.get(sidebarProfileSignOutBtn).should('be.visible').click()
+  // useLogout submits a form to the gateway which 303-redirects back to
+  // /welcome/spaces — wait for the round-trip to land and the signed-out
+  // sign-in button to render before letting the next step run.
+  cy.url({ timeout: 60000 }).should('include', constants.spacesUrl)
+  cy.get(continueWithWalletBtn, { timeout: 30000 }).should('be.visible')
+}
+
+export function verifyOnSpacesWelcomePage() {
+  cy.url({ timeout: 30000 }).should('include', '/welcome/spaces').and('not.include', onboardingCreateSpacePath)
+}
+
+export function verifySpaceCardVisible(name) {
+  cy.get(spaceCard, { timeout: 30000 }).contains(name).should('be.visible')
 }
 
 export function waitForSpacesWelcomeReady() {
--- a/apps/web/cypress/e2e/pages/spaces.page.js
+++ b/apps/web/cypress/e2e/pages/spaces.page.js
@@ -186,12 +186,16 @@ export function signOutViaSidebarProfile() {
   cy.get(continueWithWalletBtn, { timeout: 30000 }).should('be.visible')
 }
 
-export function verifyOnSpacesWelcomePage() {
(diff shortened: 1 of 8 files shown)


vmosharova (approved):
👍🏻

PR #7950, opened 2026-05-29
fix(web): stop pending CF delete queue from retrying 404s forever

katspaugh:
> Ghost entries piled in the queue,
> each reload spawned DELETEs anew —
> 404 now means we're done,
> the spam stops, the state stays clean.

## What it solves

After creating and activating a counterfactual Safe, the next page load fires dozens of DELETE requests against `https://safe-client.staging.5afe.dev/v1/users/counterfactual-safes`, each returning 404, with `[CF Sync] Failed to flush pending CF delete {status: 404, data: {...}}` repeated in the console.

Linear: https://linear.app/safe-global/issue/WA-2438/repeated-counterfactual-safe-delete-404-errors-in-console

## How this PR fixes it

Two interacting bugs in the CF-delete sync path:

1. **Duplicate dispatches generated ghost entries.** Multiple call sites dispatch `removeUndeployedSafe` for the same safe during activation:
   - self-heal effect in `useLoadSafeInfo` (cgwData arrived → undeployedSafe still present)
   - deployment-detection effect in `usePendingSafeStatuses` (`isSmartContract` returned true)
   - INDEXED event handler in `usePendingSafeStatuses`

   The listener processed every dispatch. The first DELETE cleared the CGW record; subsequent DELETEs hit a non-existent resource, 404'd, and the catch block enqueued a "ghost" entry into `pendingCfDeletes`. The listener's previous guard (`removed?.isCreator !== false`) evaluated `true` even when `removed` was `undefined`, so the duplicate dispatches weren't filtered.

2. **The queue never drained.** Both the live DELETE in the listener and the flush in `useCounterfactualSafeSync` treated 404 the same as a transient failure. Since `pendingCfDeletesSlice` is persisted (`store/index.ts`), entries accumulated across sessions and all fired in parallel via `Promise.all` on each sync.

Fixes:
- Listener bails when pre-action state has no entry to remove → duplicate dispatches become no-ops.
- Listener treats DELETE 404 as success → no enqueue, no error log.
- Flush in `useCounterfactualSafeSync` treats DELETE 404 as success → drops the entry from `pendingCfDeletes`.

Result: already-accumulated ghost entries drain on the next sync; new activations stop generating them.

## How to test it

1. Sign in with SIWE on staging.
2. Create and activate a counterfactual Safe end-to-end.
3. Reload the page.
4. Network tab: confirm there are **no** repeated DELETE requests to `/v1/users/counterfactual-safes`. Users with already-accumulated ghost entries should see at most a single drain pass that ends with the queue empty.
5. Console: confirm no `[CF Sync] Failed to flush pending CF delete` lines.
6. Regression: activate while not SIWE-authenticated, then sign in. The queued DELETE must still be flushed on sign-in; a 5xx/network failure (simulate by blocking the request) must still keep the entry for retry.

## Visual summary

```mermaid
flowchart TB
  subgraph Before
    A1[Activate CF Safe] --> B1[removeUndeployedSafe dispatched x3]
    B1 --> C1[Listener fires for every dispatch]
    C1 --> D1[1st DELETE -> 200<br/>2nd+ DELETE -> 404]
    D1 --> E1[catch enqueues ghost entry]
    E1 --> F1[Next page load: flush retries -> 404]
    F1 --> G1[Entry retained -> queue grows forever]
  end
  subgraph After
    A2[Activate CF Safe] --> B2[removeUndeployedSafe dispatched x3]
    B2 --> C2[Listener no-ops when originalState empty]
    C2 --> D2[Single DELETE -> 200]
    D2 --> E2[No ghost entry queued]
    E2 --> F2[Any pre-existing 404 drops its queue entry]
    F2 --> G2[Queue drains, no spam]
  end
```

## Regression checklist

**Surfaces touched:**
- `apps/web/src/features/counterfactual/store/counterfactualSyncListener.ts`
- `apps/web/src/features/counterfactual/hooks/useCounterfactualSafeSync.ts`

**Neighbouring flows verified by tests:**
- CF activation under auth, single dispatch — DELETE 200, queue empty.
- CF activation with concurrent dispatches — only the first hits the backend.
- 404 on live DELETE — no enqueue, no error log.
- 404 on flush — entry removed, no error log.
- 5xx / generic failure — still enqueues / retains for retry.
- Unauthenticated path — still enqueues for later flush.
- Block-list path in `fetchAndMerge` — unchanged; still skips re-adding deleted safes.

**E2E coverage:** new `apps/web/cypress/e2e/regression/counterfactual_pending_deletes.cy.js` seeds two ghost queue entries plus a SIWE session, intercepts the CGW DELETE with a 404 stub, and asserts the persisted `pendingCfDeletes` slice drains to `[]` after the initial sync.

**Not verified (risks):**
- No live staging reproduction — diagnosis is by code reading.
- 403 (non-creator) responses are not special-cased; legacy queued entries for safes the user can't delete will continue to retry. Out of scope.

## Checklist

- [x] I've added unit tests for the new behaviour
- [x] I've added a regression e2e for the queue-drain path
- [x] `type-check`, `lint`, `prettier --check`, and the counterfactual test suite are green
- [x] PR title follows conventional commits
- [x] No changes to generated files or shared theme tokens

### DIFF
--- a/apps/web/cypress/e2e/pages/main.page.js
+++ b/apps/web/cypress/e2e/pages/main.page.js
@@ -490,6 +490,22 @@ export const checkButtonByTextExists = (buttonText) => {
   cy.get('button').contains(buttonText).should('exist')
 }
 
+/**
+ * Read a key from the AUT's localStorage (after cy.visit) and assert its parsed
+ * value deep-equals `expectedValue`. Uses `.should()` so Cypress retries until
+ * the value matches or the default command timeout elapses — needed for keys
+ * the app writes asynchronously after sync completes.
+ */
+export function verifyAppLocalStorageItemEquals(key, expectedValue) {
+  cy.window()
+    .its('localStorage')
+    .invoke('getItem', key)
+    .should((stored) => {
+      const parsed = stored ? JSON.parse(stored) : null
+      expect(parsed).to.deep.equal(expectedValue)
+    })
+}
+
 export function getAddedSafeAddressFromLocalStorage(chainId, index) {
   return cy.window().then((win) => {
     const addedSafes = win.localStorage.getItem(constants.localStorageKeys.SAFE_v2__addedSafes)
--- a/apps/web/cypress/e2e/regression/counterfactual_pending_deletes.cy.js
+++ b/apps/web/cypress/e2e/regression/counterfactual_pending_deletes.cy.js
@@ -1,0 +1,38 @@
+import * as constants from '../../support/constants'
+import * as main from '../pages/main.page'
+import * as ls from '../../support/localstorage_data.js'
+import { getSafes, CATEGORIES } from '../../support/safes/safesHandler.js'
+
+let staticSafes = []
+
+describe('Counterfactual pending CF delete queue regression tests', () => {
+  before(async () => {
+    staticSafes = await getSafes(CATEGORIES.static)
+  })
+
+  beforeEach(() => {
+    // Seeded SIWE session so isAuthenticated() returns true on load — required
+    // for useCounterfactualSafeSync to flush the persisted queue.
+    main.addToLocalStorage(constants.localStorageKeys.SAFE_v2__auth, {
+      sessionExpiresAt: Date.now() + 24 * 60 * 60 * 1000,
+    })
+  })
+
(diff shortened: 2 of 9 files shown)


vmosharova (approved):
👍🏻

PR #7953, opened 2026-05-29
feat(spaces): import address book from CSV/JSON files; gate sidebar link

katspaugh:
> Tabs in the import dialog,
> CSV and JSON now flow in,
> sidebar hides its book.

## What it solves

Two Address Book tweaks:

1. The **Spaces** address book "Import" only let you copy contacts already in your local address book — there was no way to import a file. The local `/address-book` page imports CSV files; Spaces should accept files too (CSV **and** JSON).
2. The standalone `/address-book` sidebar link was always shown, including in the login-required (Spaces-gated) experience where the standalone page isn't relevant.

https://linear.app/safe-global/issue/WA-2497/import-address-book-from-csvjson-files-hide-sidebar-link

## How this PR fixes it

- **File import (Spaces):** The Import dialog now has two tabs — **Local contacts** (the existing checkbox picker, unchanged) and **Upload file** (a drag-and-drop zone accepting `.csv`/`.json`). Both feed the same `upsertAddressBookItems` mutation.
  - New pure `parseImportedAddressBook()` helper: CSV reuses the exact validators/format from the local `/address-book` import (`address,name,chainId` header); JSON reuses the global Settings → Data export shape (`{ version, data: { addressBook } }`). Entries on unsupported chains are dropped, and contact names are trimmed in one shared step (addresses are already validated + checksummed upstream — CSV via `abOnUploadValidator`, JSON via `_filterValidAbEntries`).
  - To keep JSON parsing DRY, the version-switch logic was extracted from `useGlobalImportJsonParser` into a pure `parseGlobalImportJson()`; the hook now delegates to it (no behavior change — its 15 existing tests are untouched and green).
  - The Import **button is now always openable** (the old "disabled when local book empty" gate is gone), so a file can be imported even with an empty local address book. The dialog opens on the Upload tab in that case.
- **Sidebar link gating:** `SidebarNavigation` shows the standalone **Address book** item only while `useIsRequireLoginEnabled() === false`. Two chains-config flags drive this (both renamed since the original write-up: `REQUIRE_LOGIN_DISABLED` → **`REQUIRE_LOGIN`**, with inverted meaning, and the classic-view flag is now **`CLASSIC_VIEW`**):
  - **`REQUIRE_LOGIN` (default off):** while off — today's default — the login gate is inactive, so the link **stays visible** (this PR does **not** change current default behaviour). When the flag is turned on, the standalone link is **hidden** by default.
  - **Classic-view escape hatch (`CLASSIC_VIEW` flag + per-tab opt-in):** a **signed-out** user who has opted into classic view forces the gate off, so the link **remains visible** for them even when `REQUIRE_LOGIN` is on. The opt-in is signed-out-only, so **signing in re-engages** the gate and the link disappears again.
  - The flag's **loading** state hides the link; **Cypress E2E** forces it visible. Only the main Safe sidebar is affected (Spaces has its own address book).

**Is the link still visible in classic mode?** Yes — a signed-out classic-view user always sees it (the escape hatch forces `REQUIRE_LOGIN` off), and with `REQUIRE_LOGIN` off by default everyone currently sees it. It only disappears for **signed-in** users once `REQUIRE_LOGIN` is enabled.

## How to test it

1. **Spaces import:** Open a Space → Address book → **Import**. The dialog shows "Local contacts" and "Upload file" tabs. On the Upload tab, drop a CSV (`address,name,chainId` header) or a JSON file exported from Settings → Data. Confirm the summary count, then **Import** — entries appear in the Space. Try an empty local address book: the button still opens (defaults to Upload).
2. **Sidebar (default):** With `REQUIRE_LOGIN` unset/false (default), confirm the "Address book" item **is** present in the main sidebar.
3. **Sidebar (gated):** Enable `REQUIRE_LOGIN` on the default chain and **sign in** → the "Address book" item disappears. Then opt into classic view while **signed out** (`/welcome/spaces` escape hatch, requires `CLASSIC_VIEW`) → it reappears; sign in again → it hides.

## Affected flows

- Spaces → Address book → Import (file upload + existing local-contacts picker)
- Settings → Data → Import (JSON) — shares the extracted `parseGlobalImportJson`
- Local `/address-book` CSV import — shares CSV validators (untouched)
- Main sidebar navigation rendering (gated on `REQUIRE_LOGIN` + classic-view opt-in)

## Blast radius

- **Shared:** `parseGlobalImportJson` extracted from `useGlobalImportJsonParser` (consumer: Settings → Data import). Behavior-preserving; covered by existing 15 tests.
- **Reused contracts:** address-book CSV validators (`abOnUploadValidator` et al.) — imported, not modified.
- **Hook:** `SidebarNavigation` now depends on `useIsRequireLoginEnabled`, which reads the `REQUIRE_LOGIN` flag from the default chain config and the per-tab classic-view opt-in (`useIsClassicViewOptedIn`).
- **API:** uses the existing `useAddressBooksUpsertAddressBookItemsV1Mutation`; no contract change.
- **Mobile:** none (web-only files).

## Risks / not checked

- Not visually verified in a running app — behavior is covered by component tests, but the new tab/drop-zone styling wasn't eyeballed across light/dark themes.
- Did not exercise re-importing contacts that already exist in the Space (relies on backend upsert/dedup semantics).
- Did not test very large files beyond the CSV validator's 1 MB limit (the Upload tab enforces the same `AB_FILE_SIZE_LIMIT`).

## Visual summary

```mermaid
flowchart TD
  subgraph Dialog[Spaces Import dialog]
    T1[Tab: Local contacts] --> P[Checkbox picker]
    T2[Tab: Upload file] --> D[Drop zone .csv/.json]
    D --> PR[parseImportedAddressBook]
    PR -->|CSV| V[address-book CSV validators]
    PR -->|JSON| G[parseGlobalImportJson]
    P --> RUN[runImport items]
    PR --> RUN
  end
  RUN --> M[upsertAddressBookItems mutation]

  subgraph Sidebar[Address book sidebar link]
    Q{REQUIRE_LOGIN on?}
    Q -->|no / default| SHOW[Show link]
    Q -->|yes| C{Signed-out + classic-view opt-in?}
    C -->|yes escape hatch| SHOW
    C -->|no| HIDE[Hide link]
  end
```

## Checklist

- [ ] I've tested the branch on mobile 📱 (n/a — web only)
- [x] I've documented how it affects the analytics (if at all) 📊 (reuses existing `IMPORT_ADDRESS_BOOK_SUBMIT` event)


apps/web/src/features/spaces/components/SpaceAddressBook/Import/UploadAddressBookTab.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+        const result = parseImportedAddressBook(file.name, content, supportedChainIds)
+        setFileName(file.name)
+        setItems(result.items)
+        setError(result.error)
+      }
+      reader.readAsText(file)
+    },
+    [supportedChainIds],
+  )
```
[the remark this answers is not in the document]
katspaugh: Addressed in 6bd79f2:
- `setItems([])` now runs synchronously at the top of `onDrop`, so a stale selection can't survive a new drop
- rejected drops (wrong type / too many files / too large) surface specific error messages via `getRejectionError`
- added a 1MB `maxSize` guard matching the local address book CSV import

Each behavior has a dedicated unit test in `UploadAddressBookTab.test.tsx`, plus a Cypress regression spec for the upload flow.

apps/web/src/features/spaces/components/SpaceAddressBook/Import/parseImportedAddressBook.ts
```diff
@@ -0,0 +1,84 @@
...
+  const filtered = items.filter((item) => item.chainIds.some((chainId) => supportedChainIds.includes(chainId)))
+
+  if (filtered.length === 0) {
+    return { items: [], error: NO_SUPPORTED_CHAINS_ERROR }
+  }
+
+  return { items: filtered }
+}
+
+const parseCsv = (content: string, supportedChainIds: string[]): ParsedAddressBook => {
```
vmosharova, line 30: I would add some extra guards to this function: `isChecksummedAddress(address)` for normalizing addresses and trim the name.

apps/web/src/features/spaces/components/SpaceAddressBook/Import/parseImportedAddressBook.ts
```diff
@@ -0,0 +1,84 @@
...
+  error?: string
+}
+
+const NO_SUPPORTED_CHAINS_ERROR = 'No entries found for supported chains'
+const NO_AB_ENTRIES_ERROR = 'No address book entries found in file'
+const UNSUPPORTED_FILE_ERROR = 'Unsupported file type. Please upload a CSV or JSON file.'
+
+const hasEntry = (row: string[]) => row.length === 3 && !!row[0] && !!row[1] && !!row[2]
+
+const filterToSupportedChains = (items: AddressBookItem[], supportedChainIds: string[]): ParsedAddressBook => {
```
vmosharova, line 18: Maybe it makes sense to add address and name validation into this function, as it is used in both `parseCsv` and `parseJson`

vmosharova (changes requested):
Nice! Left a couple of comments.

apps/web/src/features/spaces/components/SpaceAddressBook/Import/parseImportedAddressBook.ts
katspaugh: Done in 03d3393e8 for the **name** part: trimming now happens once in the shared `filterToSupportedChains`, so both CSV and JSON imports get trimmed names.

For **address** validation, both paths already enforce it before reaching this function, so a guard here would be dead code:
- CSV → `abOnUploadValidator` → `validateAddress`, which rejects any non-checksummed/invalid address (`Invalid address checksum`).
- JSON → `parseGlobalImportJson` → `_filterValidAbEntries`, which drops entries unless `isChecksummedAddress(address)` (and `name.trim()`).

So addresses are guaranteed valid + checksummed here — there's nothing left to normalize (non-checksummed ones are already rejected, not fixed up).

apps/web/src/features/spaces/components/SpaceAddressBook/Import/parseImportedAddressBook.ts
katspaugh: Name trimming added in 03d3393e8, centralized in the shared `filterToSupportedChains` so it covers both CSV and JSON.

On `isChecksummedAddress`: note it's a boolean predicate, not a normalizer — it can't fix a lowercase address, only reject it. And rejection already happens upstream for this path via `abOnUploadValidator` → `validateAddress` (checksum check), so a guard in `parseCsv` would never fire. If we instead wanted to *accept* lowercase input and checksum it (via ethers `getAddress`), that'd be a deliberate behavior change in the shared `validateAddress`/`_filterValidAbEntries` validators (also used by the local AB + global data import) — happy to do that separately if that's the intent.

vmosharova (approved):
👍🏻

PR #7976, opened 2026-06-02
fix(spaces): keep safe switcher usable on load failure (WA-2218)

tongyaosafe:
> When the current safe failed to load, the chevron used to vanish.
> Now we synthesize a placeholder trigger from the URL id,
> so users can still open the dropdown and switch to another safe.

## What it solves

Resolves: [WA-2218](https://linear.app/safe-global/issue/WA-2218/keep-safe-switcher-expand-icon-visible-on-load-failure)

When the current safe couldn't be matched against `items` in `SafeSelectorDropdown` (e.g. safe-info request failed, navigated to a safe that isn't in the current space), the component fell into an early-return branch that rendered an `InlineRetryError` with the copy _"This Safe is not available on the selected network"_ and **removed the dropdown trigger entirely** — leaving the user with no way to switch to a working safe.

Concretely, the dead-end branch was:

```tsx
if (!selectedItem || !mounted) {
  // ...
  if (mounted && !isLoading && items.length > 0) {
    return <InlineRetryError message="This Safe is not available on the selected network" />
  }
  // ...
}
```

`fallbackCurrentSafe` in `useSafeBarSafes` already covered the non-space context, but space context has no such fallback, and any state-transition / race that left `selectedItem` momentarily undefined locked the user out.

## How this PR fixes it

The dropdown trigger is decoupled from `selectedItem` resolution:

1. When `items.length > 0` but `selectedItem` is missing, synthesize a minimal `SafeItemData` from `${chainId}:${address}` and pass it as the trigger item. Balance / threshold / owners stay in `isLoading=true` skeleton state since we don't know them — the user still sees the address and chevron.
2. `forceOpenable` is set on the fallback path so `isSingleSafe` can't hide the chevron when only one other safe is available.
3. The full-block skeleton / `Failed to load Safe data` retry UX is preserved for the genuine "nothing to switch to" case (`items.length === 0`).
4. The old `"This Safe is not available on the selected network"` copy is dropped — the underlying load error continues to surface in main content via `useSafeInfo().safeError`, so users aren't left without context.

### Unit
```bash
yarn workspace @safe-global/web test SafeSelectorDropdown
```
98 tests pass, including 4 new/updated cases covering the fallback trigger path.

### Storybook
```bash
yarn workspace @safe-global/web storybook
```
Open `Spaces / SafeSelectorDropdown`. Pass a `selectedItemId` that isn't in `items` (e.g. `"999:0xnotfound"`). Trigger renders with a chevron; dropdown opens normally.

### Local dev — Liliya's 5/28 repro
```bash
yarn workspace @safe-global/web dev
```
1. Chrome DevTools → Network → block `GET /v1/chains/137/safes/0xe7255eE8D8A47ee01864241e7475C5c7A9792401`
2. Navigate to `http://localhost:3000/transactions/tx?safe=matic:0xe7255eE8D8A47ee01864241e7475C5c7A9792401&id=<any-tx>`
3. **Before fix:** dropdown region replaced by `"This Safe is not available on the selected network"` (or skeleton in space context).
4. **After fix:** trigger still renders, chevron clickable, dropdown lists pinned/space safes.

### Local dev — original ticket scenario
1. Open `http://localhost:3000/home?safe=matic:0x0EBD9CD5dfE7c27072AcF297F04001A6D449cB8E` (CF safe)
2. Connect an EOA wallet (any account, doesn't need to be owner)
3. During the CF → sync transition the dropdown must stay openable.

### Regression
- Normal pinned safe → trigger shows name + balance + threshold + chevron, switching works.
- Single safe → still non-openable (`isSingleSafe`).
- Tx flow open → dropdown disabled + tooltip _"Changing the Safe is not allowed in this screen"_.
- Inside a Safe App iframe → dropdown disabled.
- `items.length === 0` + isLoading → skeleton.
- `items.length === 0` + isError → `InlineRetryError` with retry button.

### Live review deployment (recommended)
Push this branch, get the review URL, and have Liliya validate against her exact repro state (account, space membership, chain) — local repro has been historically hard to reproduce on a different developer machine, so reviewer-side validation matters more than usual here.

## Affected flows

- **Safe selector dropdown — current safe missing from items**: now renders a fallback trigger instead of an `InlineRetryError`. User can switch to any other pinned or space safe.
- **Safe selector dropdown — current safe loaded normally**: unchanged.
- **Safe selector dropdown — items empty**: unchanged (skeleton / retry).
- **Main content area's "Safe Account couldn't be loaded" error**: unchanged — that error still surfaces via `useSafeInfo`.

## Blast radius

- `apps/web/src/features/spaces/components/SafeSelectorDropdown/index.tsx` — early-return restructure + new `buildFallbackSafeItem` helper.
- `apps/web/src/features/spaces/components/SafeSelectorDropdown/__tests__/SafeSelectorDropdown.test.tsx` — replaces the "not available" case with fallback-trigger coverage; adds tests for openable-with-single-other-item and pick-forwarding from fallback path.
- **Not touched:**
  - `useSafeBarSafes` / `useSpaceSafeSelectorItems` — data layer unchanged.
  - `useSafeSelectorState` — same hook signature; new input only via `forceOpenable`.
  - `fallbackCurrentSafe` — non-space context behavior unchanged.
  - Main-content error display (driven by `useSafeInfo().safeError`).
- **Consumers of this dropdown:** only `SpaceSafeBar` renders it; both space + non-space code paths go through it.
- **No mobile impact** — `apps/web`-only.
- **No analytics changes**, no schema changes, no feature flag toggles.

## Risks / not checked

- **Not reproducible on my local environment** — same state as before the WA-2218 pause: I cannot reliably reproduce the bug on my machine. Fix targets a specific, demonstrably-existing dead-end branch in the code (`items.length > 0 && !selectedItem`), but if Liliya's symptom is actually a different failure mode (e.g., `items` perpetually empty during load, or `useIsSafeBarControlDisabled` mis-firing on a route I haven't surveyed), this fix won't cover it.
- **No live validation on a deployed environment** before opening this PR — the review deployment needs to be checked against Liliya's reproducible account/state.
- **Storybook story not extended** with a fallback state — covered by unit tests but would be useful for visual regression review. Leaving as a follow-up.
- **No E2E test added** — Playwright coverage for this dropdown's failure modes would require a stable repro on review, which we don't have yet.
- Did not verify behaviour with `FEATURES.SPACES` disabled (rare path; non-space context already has `fallbackCurrentSafe`).
- Did not test on mobile (apps/web change only).

## Visual summary

```mermaid
flowchart TB
  subgraph Before["Before — dropdown disappears when selectedItem missing"]
    B1["!selectedItem || !mounted"] --> B2{"isError && mounted?"}
    B2 -->|yes| B3["InlineRetryError 'Failed to load' (retry)"]
    B2 -->|no| B4{"mounted && !isLoading && items.length > 0?"}
    B4 -->|yes| B5["InlineRetryError 'Not available on selected network' (NO RETRY — DEAD END)"]
    B4 -->|no| B6["Skeleton"]
  end

  subgraph After["After — fallback trigger keeps dropdown usable"]
    A1["!mounted || !triggerItem"] --> A2{"isError && mounted?"}
    A2 -->|yes| A3["InlineRetryError 'Failed to load' (retry)"]
    A2 -->|no| A4["Skeleton"]
    A1 -.no.-> A5["items.length === 0?"]
    A5 -->|yes + isError| A6["InlineRetryError 'Failed to load' (retry)"]
    A5 -->|yes + !isError| A7["Skeleton"]
    A5 -->|no| A8["Render dropdown with triggerItem<br/>(selectedItem ?? synthesized from URL id)"]
    A8 --> A9["Chevron clickable → user can switch"]
  end

  style B5 fill:#fdd,stroke:#a00
  style A8 fill:#dfd,stroke:#0a0
  style A9 fill:#dfd,stroke:#0a0
```

## Checklist

- [ ] I've tested the branch on mobile 📱 (N/A — apps/web only)
- [x] I've documented how it affects the analytics (if at all) 📊 (no analytics changes)

---

## CLA signature


tongyaosafe:
@claude review

apps/web/src/features/spaces/components/SafeSelectorDropdown/index.tsx
```diff
@@ -8,9 +8,26 @@ import InlineRetryError from '@/components/common/InlineRetryError'
...
 
+// Keeps the dropdown trigger renderable when the current safe isn't in `items`.
+function buildFallbackSafeItem(selectedItemId: string | undefined): SafeItemData | null {
+  if (!selectedItemId) return null
+  const colonIndex = selectedItemId.indexOf(':')
+  if (colonIndex <= 0) return null
+  return {
+    id: selectedItemId,
+    name: '',
+    address: selectedItemId.slice(colonIndex + 1),
```
katspaugh: Don't we have some utility to parse the chain:address format?

vmosharova (dismissed):
👍🏻

apps/web/src/features/spaces/components/SafeSelectorDropdown/index.tsx
tongyaosafe: fixed. Thanks.

vmosharova (approved):
👍🏻

PR #7983, opened 2026-06-02
feat(spaces): support email member invites

vseehausen:
Resolves: https://linear.app/safe-global/issue/WA-2052/fe-email-invite-flow-and-pending-invitations-ui

## What it solves

Adds the frontend flow for inviting Space members by email, alongside the existing wallet-address invite path.

The CGW API now accepts discriminated invite identifiers (`type: email` / `type: wallet`), so the wallet needs to collect either an email address or wallet address, send the matching payload shape, and show pending email invitations in the members UI.

Depends on the backend API shape from safe-global/safe-client-gateway#3104.

## How this PR fixes it

- Adds `AddMemberInput`, which supports email input, wallet addresses, ENS names, and address-book suggestions in the same invite modal.
- Builds invite payloads as explicit `email` or `wallet` items for the new CGW schema.
- Keeps wallet invite behavior intact, including ENS/name resolution and address-book entry selection.
- Limits the invite identifier input to 255 characters to match the CGW email limit.
- Shows email addresses only in the pending invitations table, while keeping the active members table focused on member names and wallet addresses.
- Updates the onboarding invite form to send wallet invite payloads with the new `type: wallet` shape.
- Updates generated store types and schema fixtures for the discriminated invite-user request.

## How to test it

1. Open a Space and go to the members screen.
2. Invite a member by wallet address or ENS and confirm the wallet invite still works.
3. Invite a member by email and confirm it appears in the pending invitations table with the email column populated.
4. Confirm the active members table does not show an email column.
5. Try a very long email-like value and confirm the input does not exceed 255 characters.

## Affected flows

- Space member invite modal.
- Pending invitations table.
- Active members table.
- Space onboarding invite form.
- Store-generated CGW invite-user request types.

## Blast radius

- **Add member modal**: input parsing, suggestion handling, validation, and payload building.
- **Members list UI**: table-column rendering differs between active members and pending invitations.
- **Invite onboarding**: wallet invites now include `type: wallet`.
- **Store schema/types**: invite-user request payload now mirrors the CGW discriminated union.
- No mobile app changes.

## Risks / not checked

- Did not run browser E2E against a live CGW email-invite endpoint locally.
- Visual behavior was reviewed through component tests/stories, not a full production-like deployment.
- Backend availability is required for the email invite submit path.

## Validation

- `node node_modules/typescript/bin/tsc --noEmit -p apps/web/tsconfig.json`
- `node node_modules/typescript/bin/tsc --noEmit -p packages/store/tsconfig.json`
- `git diff --check`

## Checklist

- [ ] I've tested the branch on mobile 📱 (web-only change)
- [x] I've documented how it affects the analytics (if at all) 📊 (removed no analytics event; no new event added)

---

## CLA signature


apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
```diff
@@ -0,0 +1,135 @@
...
+const AddMemberInput = ({
+  addressBook,
+  error,
+  inputProps,
+  onSelectAddress,
+  value,
+}: AddMemberInputProps): ReactElement => {
+  const [isOpen, setIsOpen] = useState(false)
+  const inputRef = useRef<HTMLInputElement | null>(null)
+  const identifier = value.trim()
```
LucieFaire: I wonder if we maybe can call it something else? Because identifier usually refers to some ID, and here its like `memberIdentifier` or something, just a proposal :)

apps/web/src/features/spaces/components/AddMemberModal/index.tsx
```diff
@@ -129,7 +152,7 @@ const AddMemberModal = ({ onClose }: { onClose: () => void }): ReactElement => {
 
         dispatch(
           showNotification({
-            message: `Invited ${data.name} to space`,
+            message: `Invited ${data.name || data.identifier.trim()} to space`,
```
LucieFaire: maybe we can trim it once somewhere above and use that value instead of trimming it twice?

LucieFaire:
there is a small regression as the modal design differs to dev branch:
1. sizes of the input fields
Screenshot 2026-06-03 at 11 19 28 AM

2. error is not in red font and also the input field is not highlighted
Screenshot 2026-06-03 at 11 21 15 AM

vseehausen:
> there is a small regression as the modal design differs to dev branch:
> 
>     1. sizes of the input fields
> 
> 
> Screenshot 2026-06-03 at 11 19 28 AM
> 
>     2. error is not in red font and also the input field is not highlighted
> 
> 
> Screenshot 2026-06-03 at 11 21 15 AM

Good catch, I restyled it in 29b88f13f19a50aefe54b22133800be2b48d0e23.

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
vseehausen: Good one! dac13f64de44673482c8af288b0ecf389b710c13

apps/web/src/features/spaces/components/AddMemberModal/index.tsx
vseehausen: Agree. Like this? ffca6159979606ea69e67c5c98fe654e218c94aa

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
```diff
@@ -0,0 +1,165 @@
...
+const AddMemberInput = ({
+  addressBook,
+  error,
+  inputProps,
+  onSelectAddress,
+  value,
+}: AddMemberInputProps): ReactElement => {
+  const [isOpen, setIsOpen] = useState(false)
+  const inviteeIdentifier = value.trim()
+  const shouldResolveName = Boolean(
```
LucieFaire: nit: `shouldResolveEns` ?

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const query = inviteeIdentifier.toLowerCase()
+
+    if (!query || isAddress(query)) {
+      return []
+    }
+
+    return Object.entries(addressBook)
+      .filter(([address, name]) => address.toLowerCase().includes(query) || name.toLowerCase().includes(query))
+      .slice(0, MAX_VISIBLE_OPTIONS)
+      .map(([address, name]) => ({ address, name }))
```
LucieFaire: I believe this is a potential duplication and we can replace it with reusable hooks:

Exact same pattern (would need to be extracted probably):

`AddressBookTable/index.tsx:78-87` does the identical ```address.toLowerCase().includes(query) || name.toLowerCase().includes(query)``` filter. Same logic, just not factored into a helper.

Already-extracted, reusable hooks — and these live right in the spaces feature:

`useAddressBookSearch.ts` — Fuse.js fuzzy search over name+address, already exported from the spaces barrel. 
`useContactSearch.ts` — near-identical Fuse.js variant (threshold 0.3).

I would maybe go for `useAddressBookSearch` ?

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
```diff
@@ -0,0 +1,165 @@
...
+        <IconButton
+          className={classnames(css.openButton, { [css.rotated]: isOpen })}
+          color="primary"
+          onClick={() => setIsOpen((open) => !open)}
+          tabIndex={-1}
+        >
+          <SvgIcon component={CaretDownIcon} inheritViewBox fontSize="small" />
+        </IconButton>
+      </InputAdornment>
+    ) : null
```
LucieFaire, line 101: I think these are cool UI effects, but do you think we really need those? while i type the address or the email it keeps switching the icon and maybe its a bit too much? What if we try to keep this form as close as possible to AddressBookInput but allow for emails? I would also suggest to reuse existing hooks for search/autofill/ showing those icons and so on

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
vseehausen: Addressed in 591b7f9d0.

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
vseehausen: Addressed in 0e3221830 — replaced the inline filter with useAddressBookSearch.

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
vseehausen: Addressed in cff188974 (removed the email initials avatar so the input stays calm) and 0e3221830 (reused useAddressBookSearch + open dropdown on focus, closer to AddressBookInput).

vseehausen: I disliked the blinking of the email initials, too, but decided to leave it, as they apear only on a valid email (that is the last 2-3 characters of the TLD). But you are right, better to leave them out.

apps/web/src/features/spaces/components/InviteMembersOnboarding/hooks/useInviteForm.ts
```diff
@@ -53,8 +56,8 @@ const useInviteForm = (spaceId: string | undefined, onSuccess: () => void) => {
     setIsSubmitting(true)
 
     try {
-      const usersToInvite = validMembers.map((member) => ({
-        type: 'wallet' as const,
+      const usersToInvite: WalletInviteUserDto[] = validMembers.map((member) => ({
```
LucieFaire, line 59: i think this form should also allow for both: wallet and email

LucieFaire: i think its used in the onboarding flow and we should allow for both email and wallet there too

apps/web/src/features/spaces/components/AddMemberModal/utils.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  it('returns no error for a valid external email invite', () => {
+    expect(
+      getInviteeIdentifierValidationError({
+        inviteeIdentifier: 'invitee@example.com',
+        sessionEmail: 'alice@example.com',
+      }),
+    ).toBeUndefined()
+  })
+
+  it('returns an error for emails longer than the CGW limit', () => {
```
LucieFaire: nit: would say `longer than allowed limit` and not mention CGW, because even though the numbers match - we dont actually get it from CGW so it can change any moment

apps/web/src/features/spaces/components/AddMemberModal/AddMemberModal.test.tsx
```diff
@@ -95,4 +123,117 @@ describe('AddMemberModal tracking', () => {
...
+      expect(mockInviteMembers).toHaveBeenCalledWith({
+        spaceId: 42,
+        inviteUsersDto: {
+          users: [{ type: 'email', email: 'invitee@example.com', role: 'MEMBER', name: '' }],
+        },
+      })
+    })
+  })
+
+  it('limits the invitee identifier input to the CGW email max length', () => {
```
LucieFaire: same here, would just say allowed/configured max

apps/web/src/features/spaces/components/AddMemberModal/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  if (isEmailIdentifier && sessionEmail && normalized.toLowerCase() === sessionEmail.toLowerCase()) {
+    return "You can't invite yourself."
+  }
+
+  if (isWalletIdentifier && walletAddresses?.some((walletAddress) => sameAddress(walletAddress, normalized))) {
+    return "You can't invite yourself."
+  }
+
+  return undefined
```
LucieFaire: we can make it easier to read and with less dupes:
```
 if (isEmailAddress(normalized)) {
    if (normalized.length > EMAIL_MAX_LENGTH) {
      return `Email must be ${EMAIL_MAX_LENGTH} characters or less.`
    }
    if (sessionEmail && normalized.toLowerCase() === sessionEmail.toLowerCase()) {
      return SELF_INVITE_ERROR
    }
    return undefined
  }

  if (isAddress(normalized)) {
    if (walletAddresses?.some((walletAddress) => sameAddress(walletAddress, normalized))) {
      return SELF_INVITE_ERROR
    }
    return undefined
  }

  return INVALID_IDENTIFIER_ERROR
```
and then these at the top of the file:
```
const SELF_INVITE_ERROR = "You can't invite yourself."
const INVALID_IDENTIFIER_ERROR = 'Enter a valid email, wallet address, or ENS.'
```

apps/web/src/features/spaces/components/AddMemberModal/index.tsx
```diff
@@ -47,23 +52,21 @@ export const RoleMenuItem = ({
...
-                : 'Can view the workspace data.'}
+          <Box className={css.roleDescription}>
+            <Typography variant="body2">
+              {isAdmin ? 'Admins can create and delete spaces, invite members, and more.' : 'Can view the space data.'}
             </Typography>
           </Box>
-          <Box gridArea="checkIcon" sx={{ visibility: selected ? 'visible' : 'hidden', mx: 1 }}>
-            <CheckIcon fontSize="small" sx={{ color: 'text.primary' }} />
+          <Box className={selected ? css.roleCheckIcon : css.roleCheckIconHidden}>
+            <CheckIcon fontSize="small" className={css.roleCheckSvg} />
```
LucieFaire, line 69: understand the urge to improve and add structure, but lets try to touch only the related to the feature parts to make the review easier :)

apps/web/src/features/spaces/components/AddMemberModal/styles.module.css
```diff
@@ -7,10 +7,51 @@
...
+  overflow-wrap: break-word;
+}
+
+.roleCheckIcon {
+  grid-area: checkIcon;
+  margin: 0 var(--space-1);
+}
+
+.roleCheckIconHidden {
+  composes: roleCheckIcon;
```
LucieFaire: what is this `composes`?

apps/web/src/features/spaces/components/AddMemberModal/index.tsx
```diff
@@ -154,21 +182,21 @@ const AddMemberModal = ({ onClose }: { onClose: () => void }): ReactElement => {
...
-                name="address"
-                label="Address"
-                required
-                showPrefix={false}
+              <AddMemberInput
+                addressBook={addressBook}
+                error={formState.errors.inviteeIdentifier?.message}
+                inputProps={inviteeIdentifierInputProps}
+                onSelectAddress={(address, name) => {
+                  setValue('inviteeIdentifier', address, { shouldDirty: true, shouldValidate: true })
```
LucieFaire: why do we need `shouldDirty: true` here?

apps/web/src/features/spaces/components/AddMemberModal/index.tsx
```diff
@@ -187,7 +215,7 @@ const AddMemberModal = ({ onClose }: { onClose: () => void }): ReactElement => {
               data-testid="add-member-modal-button"
               type="submit"
               variant="contained"
-              disabled={!formState.isValid || isSubmitting}
+              disabled={!inviteeIdentifierValue.trim() || !formState.isValid || isSubmitting}
```
LucieFaire: what if instead of checking for empty, we align with the AddressInput convention: add `required: true` to the inviteeIdentifier register (with no message, to avoid visible text), then drop both the `!inviteeIdentifierValue.trim()` button check and the `onSubmit` guard?

LucieFaire:
we dont load avatar for emails anymore :(
Screenshot 2026-06-03 at 7 01 34 PM

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }, [inviteeIdentifier, onSelectAddress, resolvedAddress])
+
+  const contacts = useMemo<InviteeIdentifierOption[]>(
+    () => Object.entries(addressBook).map(([address, name]) => ({ address, name })),
+    [addressBook],
+  )
+
+  const searchQuery = isAddress(inviteeIdentifier) ? '' : inviteeIdentifier
+  const matches = useAddressBookSearch(contacts as SpaceAddressBookItemDto[], searchQuery)
```
LucieFaire: hm not ideal cast but guess we can live with it

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const options = useMemo(
+    () => (isAddress(inviteeIdentifier) ? [] : matches.slice(0, MAX_VISIBLE_OPTIONS)),
```
LucieFaire, line 62: we also do the same check on line 63, maybe we can merge it somehow into 1 const?

apps/web/src/features/spaces/components/InviteMembersOnboarding/hooks/useInviteForm.ts
vseehausen: Agree, but that's out of scope for this one.

apps/web/src/features/spaces/components/AddMemberModal/index.tsx
vseehausen: A little bit of girlscout rule is ok. :)

apps/web/src/features/spaces/components/AddMemberModal/utils.test.ts
vseehausen: Addressed in 09eeaa813.

apps/web/src/features/spaces/components/AddMemberModal/AddMemberModal.test.tsx
vseehausen: Addressed in 09eeaa813.

apps/web/src/features/spaces/components/AddMemberModal/utils.ts
vseehausen: Addressed in 09eeaa813 — extracted the error constants and restructured into nested email/address branches.

apps/web/src/features/spaces/components/AddMemberModal/index.tsx
vseehausen: No, works without it, good catch. 55b771a78

apps/web/src/features/spaces/components/AddMemberModal/styles.module.css
vseehausen: `composes` is a CSS Modules feature that merges another class's rules into this one (here it pulled in `.roleCheckIcon`). Since it's not standard CSS and not used elsewhere in the repo, I replaced it with the plain rules inlined. Done in f682abc9d.

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
vseehausen: You are right, I touched the hook to clean the types in edc57a661 — useAddressBookSearch is now generic (<T extends { name; address }>), so the cast is gone.

apps/web/src/features/spaces/components/AddMemberModal/AddMemberInput.tsx
vseehausen: We actually didn't strictly need the first check. Removed in c4f6bea29.

vseehausen:
Brought email avatar back with a debounce of 500ms

LucieFaire (approved):

PR #7988, opened 2026-06-03
fix: Double space on /home page

TimDaub:
## What it solves

Fixes [WA-2465](https://linear.app/safe-global/issue/WA-2465/remove-double-space-under-wallet-banner)

## Screenshots of the different states

M1-mobile-no-banner
M2-mobile-one-banner
wa2465-banner-on
wa2465-fixed-desktop
Screenshot 2026-06-03 at 09 04 24


katspaugh (approved):
🎉

PR #7993, opened 2026-06-03
fix: Border radius of list items in txs history

TimDaub:
Context:

I told Simon that he can send me screenshots of UI inconsistencies and as part of my week as flex engineer I'll try to fix them as quickly as I can with PRs. So here is one that Simon sent me: 

> the border radius of single (and messages as well) vs bulk transactions is inconsistent. The bulk transaction one has the right one.
> On the tx history (edited)
> image
> can be tested here: /transactions/history?safe=matic:0xb412684F4F0B5d27cC4A4D287F42595aB3ae124D

Before

Screenshot 2026-06-03 at 15 23 30

After

Screenshot 2026-06-03 at 15 24 35
Screenshot 2026-06-03 at 15 31 32

Question for the reviewer:
- I mostly let Claude do the CSS changes, but for someone who is familiar with the codebase more. Is this the best way to do this change, or would there be a better way?

### DIFF
--- a/apps/web/src/components/safe-messages/MsgListItem/ExpandableMsgItem.tsx
+++ b/apps/web/src/components/safe-messages/MsgListItem/ExpandableMsgItem.tsx
@@ -11,13 +11,7 @@ import txListItemCss from '@/components/transactions/TxListItem/styles.module.cs
 
 const ExpandableMsgItem = ({ msg, expanded = false }: { msg: MessageItem; expanded?: boolean }): ReactElement => {
   return (
-    <Accordion
-      defaultExpanded={expanded}
-      disableGutters
-      elevation={0}
-      className={txListItemCss.accordion}
-      sx={{ border: 'none', '&:before': { display: 'none' } }}
-    >
+    <Accordion defaultExpanded={expanded} disableGutters elevation={0} className={txListItemCss.listItem}>
       <AccordionSummary
         data-testid="message-item"
         expandIcon={<ExpandMoreIcon />}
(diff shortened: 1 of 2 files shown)


apps/web/src/components/safe-messages/MsgListItem/ExpandableMsgItem.tsx
```diff
@@ -15,8 +15,8 @@ const ExpandableMsgItem = ({ msg, expanded = false }: { msg: MessageItem; expand
       defaultExpanded={expanded}
       disableGutters
       elevation={0}
-      className={txListItemCss.accordion}
-      sx={{ border: 'none', '&:before': { display: 'none' } }}
+      className={`${txListItemCss.listItem} ${txListItemCss.standalone}`}
+      sx={{ '&:before': { display: 'none' } }}
```
vseehausen: Is it possible to move this to the css file, too? As a refactor? So we avoid inline styles.

vseehausen (dismissed):
You can either decide to merge, or address my comment and then merge.

simabeh:
I requested this change to have consistent design fyi

ivalui:
Quick design QA: border radius seems a bit huge, let's downgrade one level -- to 16px. See [figma](https://www.figma.com/design/TJMKVyvbDLzS3XluYaGg05/Transactions?node-id=3003-52549&t=VOaUXVMqET4EERCd-4)

TimDaub:
@vseehausen I need another review here since I moved the style changes out of inline styles and also used @ivalui"s border radius

gfendres (approved):

PR #7997, opened 2026-06-03
fix: only show sidebar nav tooltips when collapsed and fix Settings item

compojoom:
> Tooltips on the rail,
> hush when wide, whisper when thin —
> the cog clicks once more.

## What it solves

Sidebar nav items showed their label tooltip on hover even when the sidebar was **expanded** (redundant with the already-visible label), and the tooltip was anchored to the icon rather than the button. The **Settings** item had three extra bugs: its tooltip always showed, its icon didn't turn green when active, and — when the sidebar was collapsed to icons — the cog was neither clickable nor hoverable.

Resolves: https://linear.app/safe-global/issue/WA-2375/remove-redundant-tooltips-from-sidebar-menu-items

## How this PR fixes it

- Label tooltips now only render when the sidebar is collapsed to icons (`state === 'collapsed' && \!isMobile`) and are anchored to the whole button via a wrapping `TooltipTrigger`. The disabled-Safe message ("You need to activate your Safe first.") is preserved and still shows regardless of collapse state.
- Settings was a bespoke `SidebarMenuButton` that hadn't been migrated to the `NavItem` tooltip pattern — it kept the old always-on, icon-anchored tooltip and never applied the `activeIcon` (green) style. It now mirrors `NavItem`.
- **The non-obvious one:** the collapsed cog being un-clickable was *not* a Settings bug. In icon mode `SidebarGroupLabel` gets `-mt-8` + `opacity-0` but keeps its 32px height and pointer events, so the invisible "Defi" group label sits directly on top of the last item above it (Settings) and swallows hover/click. Fixed by adding `group-data-[collapsible=icon]:pointer-events-none` to `SidebarGroupLabel` — a decorative label should never intercept pointer events. (Note: the `tooltip` prop built into `SidebarMenuButton` can't be used here because it replaces the `render` prop we need for the Next `Link`.)

## How to test it

1. Open a Safe and expand the sidebar — hover any nav item (e.g. Transactions, Settings); **no** tooltip should appear.
2. Collapse the sidebar (Toggle Sidebar) to icons — hover an item; the label tooltip appears to the right, vertically centered on the button.
3. With the sidebar collapsed, hover and **click** the Settings cog — the tooltip shows and the click navigates to Settings.
4. Navigate into Settings — the Settings cog icon turns green like other active items.

## Screenshots

https://github.com/user-attachments/assets/d49ba231-89f4-430d-bfab-fc68613d7605

| State | Behaviour |
| --- | --- |
| Expanded, hover | No tooltip (label already visible) |
| Collapsed, hover | "Settings"/label tooltip anchored to the button |
| Active | Settings cog icon rendered green |

```mermaid
flowchart TB
  subgraph Before
    A1[Hover nav item] --> B1[Tooltip always shown, anchored to icon]
    C1[Collapsed: Defi group label -mt-8 overlaps Settings] --> D1[Cog not clickable / no tooltip]
  end
  subgraph After
    A2[Hover nav item] --> B2{Sidebar collapsed?}
    B2 -->|No| E2[No tooltip]
    B2 -->|Yes| F2[Tooltip anchored to button]
    C2[Group label pointer-events-none when collapsed] --> D2[Cog clickable + tooltip works]
  end
```

## Checklist

- [x] I've documented how it affects the analytics (if at all) 📊 — no analytics changes
- [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻 — added NavItem + SafeSidebarVariant tests

---

## CLA signature

### DIFF
--- a/apps/web/src/components/ui/sidebar.tsx
+++ b/apps/web/src/components/ui/sidebar.tsx
@@ -539,7 +539,7 @@ function SidebarGroupLabel({ className, render, ...props }: useRender.ComponentP
     props: mergeProps<'div'>(
       {
         className: cn(
-          'text-sidebar-foreground/70 ring-sidebar-ring h-8 rounded-md px-2 text-xs font-medium transition-[margin,opacity] duration-200 ease-linear group-data-[collapsible=icon]:-mt-8 group-data-[collapsible=icon]:opacity-0 focus-visible:ring-2 [&>svg]:size-4 flex shrink-0 items-center outline-hidden [&>svg]:shrink-0',
+          'text-sidebar-foreground/70 ring-sidebar-ring h-8 rounded-md px-2 text-xs font-medium transition-[margin,opacity] duration-200 ease-linear group-data-[collapsible=icon]:-mt-8 group-data-[collapsible=icon]:opacity-0 group-data-[collapsible=icon]:pointer-events-none focus-visible:ring-2 [&>svg]:size-4 flex shrink-0 items-center outline-hidden [&>svg]:shrink-0',
           className,
         ),
       },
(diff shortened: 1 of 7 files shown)


vmosharova (approved):
👍🏻 Thank you!

PR #8007, opened 2026-06-04
refactor(survey): flip feature flag to positive SPACE_ONBOARDING_SURVEY (WA-2493)

tongyaosafe:
> Disabled flips off, on its head,
> SPACE_ONBOARDING_SURVEY in its stead —
> opt-in now, the default is dead.

## What it solves

Resolves: [WA-2493](https://linear.app/safe-global/issue/WA-2493/rename-survey-onboarding-disabled-feature-flag)

The Space onboarding survey shipped behind an inverted chain-config flag (`SURVEY_ONBOARDING_DISABLED`) so it was on by default and the config service had to opt out. Team direction is to use a regular positive feature flag instead — default off, config service opts each environment in.

## How this PR fixes it

- `packages/utils/src/utils/chains.ts` — rename the enum entry from `SURVEY_ONBOARDING_DISABLED` to `SPACE_ONBOARDING_SURVEY` (singular `SPACE` matches "per-Space onboarding").
- `apps/web/src/hooks/useIsSurveyEnabled.ts` — drop the `!` so the hook returns `hasFeature(chain, FEATURES.SPACE_ONBOARDING_SURVEY)` directly. Updated the docstring to reflect the new "absent = off, present = on" semantics.

No other callers needed to change — the survey page and its consumers all go through `useIsSurveyEnabled`, which keeps the same `boolean | undefined` shape.

## Behaviour change (deploy coordination)

| State | Before (`*_DISABLED`) | After (`SPACE_ONBOARDING_SURVEY`) |
| --- | --- | --- |
| Chain features array empty / no config | Survey shown | **Survey hidden** |
| To enable per chain | (default already enabled) | Add `SPACE_ONBOARDING_SURVEY` to that chain's `features` |
| To disable per chain | Add `SURVEY_ONBOARDING_DISABLED` | Remove `SPACE_ONBOARDING_SURVEY` |
| Loading (chains config in flight) | `undefined` (no redirect) | `undefined` (no redirect) — unchanged |

Before this PR merges into dev / before the next deploy, **the config service has to add `SPACE_ONBOARDING_SURVEY` to the relevant chains' features arrays**, otherwise the survey step will silently disappear on staging.

## How to test it

- `yarn workspace @safe-global/web type-check` — clean.
- `yarn workspace @safe-global/web test --testPathPattern="SurveyOnboarding|SurveyPage"` — all 14 tests pass.
- Manual on staging:
  1. Without adding the flag to staging config service, hit `/welcome/survey?spaceId=...` — the page should redirect to the Space dashboard (or to create-space if no spaceId).
  2. Add `SPACE_ONBOARDING_SURVEY` to the relevant chain in staging config service. Wait ~60s for the CGW Redis cache to expire, hard-refresh — the survey UI should appear.
  3. Remove the flag again, repeat — the survey should disappear after the cache expires.

## Affected flows

- Space onboarding step 4 (`/welcome/survey`).
- No other consumers — `useIsSurveyEnabled` is the single source of truth for this flag.

## Blast radius

- `packages/utils/src/utils/chains.ts` — enum value renamed. Grep shows the old name is gone from the codebase.
- `apps/web/src/hooks/useIsSurveyEnabled.ts` — only consumer of this enum value.
- Config service: the **`SURVEY_ONBOARDING_DISABLED`** key (if any was set anywhere) becomes inert with this PR. **`SPACE_ONBOARDING_SURVEY`** must be added on the chains where we want the survey on.

## Risks / not checked

- Did not exercise the live staging flow until config service is updated — the flag rename has to be coordinated with whoever owns the config service entries.
- Did not check mobile — survey is a web-only feature; `useIsSurveyEnabled` lives under `apps/web/`, no mobile consumers.

## Visual summary

```mermaid
flowchart LR
  Config[config service<br/>chain features array] -- contains SPACE_ONBOARDING_SURVEY? --> Hook[useIsSurveyEnabled]
  Hook -- true --> Show[SurveyOnboarding renders]
  Hook -- false --> Redirect[redirect to /spaces]
  Hook -- undefined<br/>chains config loading --> Wait[render null, wait]
```

## Checklist

---

## CLA signature


vmosharova (approved):
👍🏻

PR #8026, opened 2026-06-08
feat(spaces): bring search back to the account selector dropdown

tongyaosafe:
> Type a name, an address, a chain —
> the list narrows, the last row fades,
> a thin bar marks how far there is to go.

## What it solves

Resolves: [WA-2458](https://linear.app/safe-global/issue/WA-2458/bring-search-back-to-the-accounts-dropdown-visual-effect-to)

The account switcher dropdown lost its search in the new UI. With 100–200 safes, users had to scroll the whole list or detour through **All Accounts** just to find one. The ticket also asks for a scroll affordance ("last account item should fade to hint for scroll").

This brings search back **directly in the dropdown**, adds the scroll-hint fade, and fixes a few rough edges the work surfaced along the way.

## How this PR fixes it

**Search**

- Sticky search field at the top of the dropdown (reuses `InputGroup` / `InputGroupInput` + lucide `Search`, matching `AccountsModal`).
- Filters by **name**, **address**, and **chain name / short name**.
- Search matches the name the user actually **sees**. Row names come from `useSafeDisplayName` (the safe's own name **or** its address-book name, source-aware), so filtering the raw `item.name` missed safes named only in the address book. A new `useSafeNameResolver` hook resolves the same display name for many addresses without a hook-per-item (incl. local names on non-current chains).
- base-ui `Select` runs typeahead on keystrokes; the field stops keystroke propagation so typing isn't hijacked (Escape still closes).

**Scroll affordances**

- The dropdown is restructured into **header / scrollable list / footer** (like `AccountsModal`) so only the middle list scrolls. A **thin scrollbar** shows on the right, confined to the list (no longer running over the search field and footer).
- A **bottom fade** over the last visible row hints there's more to scroll (replaces the old bouncing chevron, per Figma). It fades to the dropdown's real background var (`--color-background-paper`); `to-card` rendered transparent because `card` isn't a `:root` color token.
- The fade is shown only while the list overflows. Because base-ui sizes the popup asynchronously on open, the overflow check is re-run on the next frame (and on resize) so the fade doesn't intermittently go missing.
- Popup max height raised (`20rem` → `min(34rem, var(--available-height))`) for long lists.

**Fixes surfaced by the search field's keyboard focus**

- The trigger's `focus-visible` ring was clipped by the wrapper's `overflow-hidden` into stray top/bottom bars; suppressed (the card shows no focus ring by design).
- The popup showed the browser's blue focus outline when typing focused it; added `outline-hidden`.

**Loading state**

- The threshold/owners badge rendered `0/0` while a safe's overview was still loading (RTK `isLoading` is false during arg-change/refetch). A deployed safe is always `>= 1/1`, so `0/0` is treated as "not loaded" and a skeleton is shown instead — in both the rows and the trigger.

## How to test it

1. Open the account switcher dropdown on an account with several safes.
2. Type part of a **name**, an **address** (`0x…`), or a **chain** (`polygon` / `matic`) → list narrows. Search a safe named **only in the address book** → it's found.
3. Clear the query → full list returns. No match → empty state.
4. With a long list: a **thin scrollbar** shows on the right (only over the list), and the **last row fades**. Scroll to the bottom → fade disappears. Re-open a few times → the fade shows consistently.
5. Switch to a safe whose data isn't cached → the signers badge shows a **skeleton**, not `0/0`.
6. Tab/keyboard into the dropdown and type → no clipped grey bars on the trigger, no blue outline around the popup; Escape closes.

## Affected flows

- Account switcher dropdown: searching/filtering the list (new), scrolling, selecting a safe, expanding multi-chain rows.
- "Trusted Safes" (merged) and "Safes in this workspace" (spaceOnly) header contexts.
- Trigger + rows loading state (signers badge).

## Blast radius

- **`SafeDropdownContainer`** — search state + filtering, three-section scroll layout, fade/scrollbar, scroll-hint detection. Props unchanged.
- **`useSafeNameResolver`** (new export in `hooks/useAllAddressBooks.ts`) — used only by the dropdown filter; mirrors `useSafeDisplayName` / `useAddressBookItem` priority (a "keep in sync" comment marks the intentional duplication). Reads `selectAllAddressBooks` + merged books via existing hooks.
- **`BalanceDisplay` / `SafeBalanceBlock`** — `0/0` → skeleton. Used by the dropdown rows + trigger; a deployed safe is never `0/0`, so safe everywhere.
- **`useSafeDisplayName`** — docs only.
- **`SafeSelectorDropdown` (index)** — trigger focus-ring class only.
- Not touched: `useSpaceSafeSelectorItems`. No API/slice/migration/feature-flag changes. Web-only (`apps/web`); mobile unaffected.

## Risks / not checked

- `useSafeNameResolver` duplicates the source-aware priority logic (mitigated by a sync comment, not abstraction); a future change to display-name priority must touch both.
- No virtualization — relies on search narrowing the rendered set; 200-item raw-scroll perf not profiled.
- Keyboard list navigation from inside the search field is intentionally disabled (arrows/Enter stay in the input).
- Verified manually in the browser; no end-to-end test renders a real store for the address-book-name search path.
- Not tested on mobile (web-only change).

## Visual summary

Dropdown layout — only the middle list scrolls:

```mermaid
flowchart TB
  subgraph Popup["SelectContent popup (overflow-hidden)"]
    H["Header + search field — shrink-0"]
    L["Account list — flex-1, overflow-y-auto<br/>• thin scrollbar (right)<br/>• bottom fade when more rows below"]
    F["Footer (All accounts / Explore) — shrink-0"]
    H --> L --> F
  end
```

Search name resolution (matches what each row displays):

```mermaid
flowchart LR
  Q[query] --> R["useSafeNameResolver(addr, chainId, item.name)"]
  R --> P{preferredName?}
  P -->|yes| N[display name]
  P -->|no| S{address-book source}
  S -->|merged| G["space/private (cross-chain) ?? any-chain local"]
  S -->|spaceOnly| SP[space contact]
  S -->|localOnly| LO[cross-chain local]
  G --> N
  SP --> N
  LO --> N
  N --> M{"name / address / chain matches query?"}
  M -->|yes| K[keep] 
  M -->|no| D[drop]
```

_(UI screenshot of the dropdown — search field, thin scrollbar, bottom fade — to be attached.)_

## Checklist

- [ ] I've tested the branch on mobile 📱 — N/A, web-only change
- [x] I've documented how it affects the analytics (if at all) 📊 — no new analytics events
- [x] I've written a unit/e2e test for it (if applicable) 🧑‍💻 — resolver unit tests (incl. cross-chain local), dropdown filter/empty/scroll-hint/keystroke tests, BalanceDisplay 0/0→skeleton tests

---

## CLA signature

### DIFF
--- a/apps/web/src/features/spaces/components/SafeSelectorDropdown/SafeSelectorDropdown.stories.tsx
+++ b/apps/web/src/features/spaces/components/SafeSelectorDropdown/SafeSelectorDropdown.stories.tsx
@@ -242,6 +242,33 @@ export const LoadingWithHeaderFooter: Story = {
   args: {} as any,
 }
 
+/** Many safes with header + footer — exercises the scrollbar and the bottom scroll-hint fade. */
+export const ManySafesWithFooter: Story = {
+  render: () => {
+    const items = createMockItems(13)
+    const header = (
+      <div className="flex items-center gap-1 px-4 pt-3 pb-2">
+        <span className="text-sm font-semibold text-secondary-foreground">Safes in this workspace</span>
+      </div>
+    )
+    const footer = (
+      <div className="px-4 py-3">
+        <button className="w-full rounded-md border px-3 py-1.5 text-sm">Explore other Safes &rsaquo;</button>
+      </div>
+    )
+    return (
+      <SafeSelectorDropdown
+        items={items}
+        selectedItemId={items[1].id}
+        onItemSelect={action('Item selected')}
+        header={header}
+        footer={footer}
+      />
+    )
+  },
+  args: {} as any,
+}
+
 /** Error state with header and footer (non-space context). */
 export const ErrorWithHeaderFooter: Story = {
   render: () => {
--- a/apps/web/src/features/spaces/components/SafeSelectorDropdown/components/BalanceDisplay.tsx
+++ b/apps/web/src/features/spaces/components/SafeSelectorDropdown/components/BalanceDisplay.tsx
@@ -12,26 +12,32 @@ export interface BalanceDisplayProps {
   showThreshold?: boolean
 }
 
-const BalanceDisplay = ({ balance, threshold, owners, isLoading, showThreshold = true }: BalanceDisplayProps) => (
-  <div className="flex flex-col items-end min-w-0 shrink sm:w-[100px] sm:shrink-0">
-    {balance !== undefined &&
-      (isLoading ? (
-        <Skeleton className="h-4 w-14 rounded" />
-      ) : (
-        <Typography variant="paragraph-mini-medium" color="muted">
-          {balance}
-        </Typography>
-      ))}
-    {showThreshold &&
-      (isLoading ? (
-        <Skeleton className="h-4 w-14 rounded" />
-      ) : (
-        <Badge data-testid="safe-selector-threshold" variant="secondary" className="gap-1">
-          <User className="size-3" />
-          {threshold}/{owners}
-        </Badge>
-      ))}
-  </div>
-)
+const BalanceDisplay = ({ balance, threshold, owners, isLoading, showThreshold = true }: BalanceDisplayProps) => {
+  // A deployed safe always has threshold/owners >= 1, so 0/0 means the overview hasn't loaded yet
+  // (e.g. while switching safes); show a skeleton instead of a misleading "0/0".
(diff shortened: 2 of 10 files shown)


vmosharova (approved):
👍🏻 Nice improvement!

PR #8029, opened 2026-06-08
fix: release stuck body scroll lock after closing modals in mobile view

clovisdasilvaneto:
> Modal locks the body,
> Base UI restores it stale—
> a guard frees the scroll.

## What it solves

Resolves: https://linear.app/safe-global/issue/WA-2508

In mobile view, opening any modal/overlay and closing it could leave `style="overflow: hidden"` stuck on `<body>`, making the page unscrollable (most visibly: open the sidebar drawer, close it, scroll is dead).

## How this PR fixes it

The app runs two independent scroll-lockers: MUI's `ModalManager` (classic dialogs/drawers/popovers, incl. the temporary sidebar `Drawer`) and Base UI's locker (the new shadcn dialogs/drawers). Both write `overflow: hidden` to the body on open and restore it on close — **MUI synchronously, Base UI via a deferred `setTimeout(0)`**.

When a Base UI overlay opens on top of an already-MUI-locked body (the page is scroll-locked only when it's actually scrollable — i.e. in mobile view), Base UI captures `hidden` as its "original" value and writes it back **after** MUI has already cleared it, leaving the body permanently locked.

This adds `useUnlockBodyScroll`, a small global guard mounted once in `InitApp`. A `MutationObserver` on the body watches both:
- `style` attribute changes — to catch a stuck `overflow: hidden` write, and
- `childList` changes — to catch an overlay unmounting (matters because MUI's ~225 ms Drawer exit transition outlives the debounce, so the first check bails while the overlay node is still present and only the unmount tells us to re-check).

When the body is locked **and no overlay is open**, it restores the body. A legitimately-open modal is never disturbed.

## How to test it

1. Open the app in mobile view (narrow window / device mode).
2. Open the sidebar drawer (or any modal), then close it.
3. Confirm `<body>` no longer has `style="overflow: hidden"` and the page scrolls. Repeat with nested overlays (a dropdown inside a dialog/drawer).

## Affected flows

- Any modal/dialog/drawer/sheet/popover open→close in mobile view — body scroll lock is now reliably released.
- Mobile sidebar (`SideDrawer` — MUI temporary `Drawer`) open→close.

## Blast radius

- New hook `useUnlockBodyScroll` mounted once in `InitApp` (`_app.tsx`) — runs app-wide on the client.
- Reads/writes only `document.body` inline styles (`overflow`, `overflow-x/y`, `padding-right`) and only when no overlay is open.
- Overlay detection spans MUI (`.MuiModal-root:not(.MuiModal-hidden)`), Base UI/shadcn (`[data-slot="...-content"]`), and Vaul drawers.
- No Redux, RTK Query, routing, feature flags, or persisted state touched. No shared-package change → no mobile (React Native) impact.

## Risks / not checked

- Defensive guard, not a patch to the upstream MUI/Base UI restore-order bug itself.
- web3-onboard's wallet modal manages its own scroll separately; it doesn't write body `overflow`, so it's out of scope.
- If a future page legitimately sets `body { overflow: hidden }` **inline** with no overlay open, the guard would clear it (no such page exists today; no app code or global CSS writes body overflow).
- The 100 ms debounce is a heuristic; verified in a real browser with a temporary MUI Drawer + nested Base UI dialog.

## Visual summary

```mermaid
sequenceDiagram
  participant U as User (mobile view)
  participant MUI as MUI ModalManager
  participant B as Base UI scroll locker
  participant Body as document.body
  participant G as useUnlockBodyScroll

  U->>MUI: open modal/drawer
  MUI->>Body: overflow = hidden
  U->>B: open overlay on top
  Note over B: captures "hidden" as original
  U->>MUI: close modal
  MUI->>Body: restore overflow = '' (sync)
  B-->>Body: setTimeout(0): overflow = hidden (stale!)
  Note over Body: stuck → page can't scroll
  G->>Body: observer fires (style + childList)
  G->>Body: no overlay open → remove overflow
  Note over Body: scroll restored ✅
```

## Checklist


vmosharova:
@claude review

vmosharova (approved):
👍🏻 Approved; probably should be cherry-picked to the RC.

PR #8047, opened 2026-06-09
fix: Show cloud icon for when adding team members

TimDaub:
In the "Add member" flow on Workspaces, when typing in a contact, for contacts that are stored in the workspace, but are private, a cloud icon would NOT show next to their name when adding a team member. With @liliya-soroka, we discussed offline that these should have an icon though. Here's the problematic part on staging:

Screenshot 2026-06-09 at 18 11 08

The cloud icon for private contacts that are stored in the workspace is however implemented to show a cloud icon when selecting recipients for sending a new transaction, an example:
Screenshot 2026-06-09 at 18 12 52

Hence, this PR adds showing the cloud icon also for private contacts on the "Add member" dialogue:


katspaugh (approved):
👍

PR #8002, opened 2026-06-04
Epic auth m2

LucieFaire:
> Invites by mail take flight,
> sessions split from the connected wallet,
> expired ones renew.

## What it solves

Epic branch for Auth M2: invitation flow

Spaces/Workspaces Auth Milestone 2: support inviting members by email (in both the members list and the onboarding flow), and decouple the auth session from the connected wallet.

## Main changes

- **Email member invites** — invite Space members by email address (#7983)
- **Email invites in onboarding** — allow email invites during the onboarding flow (#8004)
- **Decouple wallet from auth session** — auth session no longer tied to the connected wallet, WA-2302 (#8008)
- **Renew expired invites** — add a renew button for expired invites (#8022)
- **Onboarding invite step** — UI fix for the onboarding invite step (#8017)
- **Long email truncation** — truncate long member emails with a hover tooltip (#7984)
- Review-comment fixes and test updates (#8048)

## Checklist

- [ ] I've tested the branch on mobile 📱
- [ ] I've documented how it affects the analytics (if at all) 📊
- [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻
- [ ] I've listed affected flows and blast radius, and named what I did not verify 🎯

---

## CLA signature

With the submission of this Pull Request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://safe.global/cla).

PR #8048, opened 2026-06-09
fix: resolve invite-flow review comments on epicAuthM2

vseehausen:
> Trim before we compare,
> lenient checksums, finite dates—
> review notes, resolved.

Addresses the automated review comments (Codex + Claude) on [#8002](https://github.com/safe-global/safe-wallet-monorepo/pull/8002). Targets the epic branch `epicAuthM2`.

## What it solves

Bugs and rough edges flagged by the bot reviewers on the space-member invite flow:

- **Whitespace-padded duplicate invites slipped through.** Submission trims all identifiers, but the per-row dedupe checks classified/compared the *other* row untrimmed, so `alice@example.com ` and `alice@example.com` (or padded wallet addresses) both submitted.
- **Modal rejected non-checksummed addresses.** `viem.isAddress` defaults to strict EIP-55, so lowercase/all-caps addresses failed in `AddMemberModal` while the onboarding form accepted them.
- **Submit button could stay disabled.** The empty-members early return set `isSubmitting(true)` and never cleared it.
- **`isInviteExpired` swallowed malformed dates implicitly** (`NaN <= now`); now guarded explicitly.
- Minor cleanups: removed an `as never` cast; memoized `onSelectAddress` so `AddMemberInput`'s resolve-effect stops re-validating every render.

## How this PR fixes it

One commit per review comment:

| Commit | Fix |
|---|---|
| `fix: normalize identifiers before duplicate invite checks` | Trim the compared row before classifying/comparing (email + address) |
| `fix: accept non-checksummed addresses in member invite modal` | `isAddress(..., { strict: false })` across the modal |
| `fix: don't leave invite submit button disabled on empty submit` | Drop the stray `setIsSubmitting(true)` |
| `fix: guard isInviteExpired against malformed dates` | `Number.isFinite` guard + regression test |
| `refactor: drop 'as never' cast in AddMemberInput change handler` | Rely on RHF's `ChangeHandler` typing |
| `perf: memoize onSelectAddress to avoid redundant validation` | `useCallback` for the parent callback |
| `tests: cover RenewInviteButton loading/disabled state` | Assert the button disables while renewing |

Two flagged items were intentionally **not** changed: the wallet-disconnect logout removal (intentional — `useSessionExpiryGuard` reconciles SIWE sessions server-side) and the speculative `toInviteName` `'Member'` collision.

## How to test it

- `yarn workspace @safe-global/web test --watchman=false src/features/spaces` — all green (incl. new malformed-date and renew-loading cases).
- Manually, in Add member modal + onboarding invite:
  - Paste a lowercase address → accepted (was rejected in the modal).
  - Add the same email twice with trailing whitespace on one → blocked as duplicate.

## Visual summary

```mermaid
flowchart LR
  A[invite identifier] --> B{trim}
  B --> C[classify: email / address / ENS]
  C --> D[dedupe vs other rows<br/>also trimmed]
  C --> E["isAddress strict:false<br/>(accept lowercase)"]
```

## Checklist

- [x] I have performed a self-review of my own code
- [x] Tests added/updated for the changed logic
- [ ] No new lint/type-check/prettier issues

---

By submitting this PR, I confirm that my contribution is made under the terms of the [Safe CLA](https://safe.global).

### DIFF
--- a/apps/web/src/features/spaces/components/AddMemberModal/index.tsx
+++ b/apps/web/src/features/spaces/components/AddMemberModal/index.tsx
@@ -29,7 +29,7 @@ import useAddressBook from '@/hooks/useAddressBook'
 import { isAuthenticated } from '@/store/authSlice'
 import { useAuthGetMeV1Query } from '@safe-global/store/gateway/AUTO_GENERATED/auth'
 import { useUsersGetWithWalletsV1Query } from '@safe-global/store/gateway/AUTO_GENERATED/users'
-import { isAddress } from 'viem'
+import { isAddress } from 'ethers'
 import {
   type MemberField,
   buildInviteUserPayload,
@@ -113,7 +113,7 @@ const AddMemberModal = ({ onClose }: { onClose: () => void }): ReactElement => {
   })
 
   useEffect(() => {
-    if (!isAddress(inviteeIdentifierValue, { strict: false })) {
+    if (!isAddress(inviteeIdentifierValue)) {
       return
     }
 
--- a/apps/web/src/features/spaces/components/AddMemberModal/index.tsx
+++ b/apps/web/src/features/spaces/components/AddMemberModal/index.tsx
@@ -1,4 +1,4 @@
-import { type ReactElement, useEffect, useState } from 'react'
+import { type ReactElement, useCallback, useEffect, useState } from 'react'
 import {
   Alert,
   Box,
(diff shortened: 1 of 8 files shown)


apps/web/src/features/spaces/components/AddMemberModal/index.tsx
```diff
@@ -113,7 +113,7 @@ const AddMemberModal = ({ onClose }: { onClose: () => void }): ReactElement => {
   })
 
   useEffect(() => {
-    if (!isAddress(inviteeIdentifierValue)) {
+    if (!isAddress(inviteeIdentifierValue, { strict: false })) {
```
compojoom: Why are we using isAddress from viem in the first place. We don't use viem directly, it's 99% transitive dependency of other packages.

vseehausen: You are right, swapped with ethers in 4544f20b7

compojoom (approved):

LucieFaire (approved):

